PR #13719, opened 2023-04-16
Makes a sub-type for catwalk supports that are auto-connectable

CaioFrancisco:
## About the PR 

Makes a new turf for catwalk supports, which auto-connects to each other.

## Why's this needed? 

#13643 (i should have finished this PR a long time ago, but eh)


CaioFrancisco:
You're the one reviewing the other PR, so I'm gonna request you to do this one as well.

TobleroneSwordfish (approved):
Looks good to me, the fact that update_neighbors isn't defined on any of the parents is cursed but not your fault.

PR #13737, opened 2023-04-18
Melt doors open with a C saber

Chatauscours:
[FEATURE] [GAME OBJECTS]

## About the PR 
This PR makes it so you can melt airlocks open with a c saber on help intent.
Melted airlocks are forced open and disabled forever (similarly to the emag).
For the melting to be done, there is an actionbar during which you cannot move or attack.
The duration of the action bar scales with door HP and matches the give or take time needed to bash the door open normally. 
Airlocks that are not hackable or emaggable (azones, listening post, ...) are immune to this.

## Why's this needed? 
It's very cool and badass and a little bit of a buff to the c saber (bashing the door open takes about 20 seconds currently).
https://www.youtube.com/watch?v=K48M2S7bkSA&t=22s

## Changelog 

```changelog
(u)Chatauscours
(*) Powered cyalume sabers on help intent are now able to melt doors open, like the cool guys do it in movies.
```

### DIFF
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -1594,23 +1594,22 @@ About the new airlock wires panel:
 		return src.Attackhand(user)
 	else if (ispryingtool(C))
 		src.unpowered_open_close()
-	else if (C.type == /obj/item/sword && user.a_intent == INTENT_HELP && !( src.operating ))
+	else if (istype(C, /obj/item/sword) && (user.a_intent == INTENT_HELP) && !src.operating)
 		var/obj/item/sword/S = C
 		if (!S.active)
-			boutput(user, "<span class='alert'>[S] needs to be on to melt the [src]</span>")
+			boutput(user, "<span class='alert'>[S] needs to be on to melt the [src]!</span>")
 			return
 		if (src.cant_emag)
 			boutput(user, "<span class='alert'>[src] is too resistant to melt from the saber!</span>")
 			return
-		else if (!src.density)
+		if (!src.density)
 			boutput(user, "<span class='alert'>You don't need to break open [src], as it's already open!</span>")
 			return
-		else
-			var/positions = src.get_welding_positions(user)
-			user.visible_message("<span class='alert'>[user] begins melting [src] open with [S]!</span>",\
-				"<span class='alert'>You begin melting [src] open with your [S.name]!</span>")
-			playsound(user.loc, 'sound/items/Welder2.ogg', 40, 1)
-			actions.start(new /datum/action/bar/private/welding(user, src, 10 SECONDS, /obj/machinery/door/airlock/proc/sword_open, null, null, positions[1], positions[2]),user)
+		var/positions = src.get_welding_positions(user)
+		user.visible_message("<span class='alert'>[user] begins melting [src] open with [S]!</span>",\
+			"<span class='alert'>You begin melting [src] open with your [S.name]!</span>")
+		playsound(user.loc, 'sound/items/Welder2.ogg', 40, 1)
+		actions.start(new /datum/action/bar/private/welding(user, src, src.health/45 SECONDS, /obj/machinery/door/airlock/proc/sword_open, null, null, positions[1], positions[2]),user)
 	else
 		..()
 	return
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -1594,7 +1594,11 @@ About the new airlock wires panel:
 		return src.Attackhand(user)
 	else if (ispryingtool(C))
 		src.unpowered_open_close()
-	else if (C.type == /obj/item/sword && C:can_reflect() && user.a_intent == INTENT_HELP && !( src.operating ))
+	else if (C.type == /obj/item/sword && user.a_intent == INTENT_HELP && !( src.operating ))
+		var/obj/item/sword/S = C
+		if (!S.active)
+			boutput(user, "<span class='alert'>[S] needs to be on to melt the [src]</span>")
+			return
 		if (src.cant_emag)
 			boutput(user, "<span class='alert'>[src] is too resistant to melt from the saber!</span>")
 			return
@@ -1603,8 +1607,8 @@ About the new airlock wires panel:
 			return
 		else
 			var/positions = src.get_welding_positions(user)
-			user.visible_message("<span class='alert'>[user] is melting [src] open with a cyalume saber!</span>",\
-				"<span class='alert'>You are melting [src] open with your cyalume saber!</span>")
+			user.visible_message("<span class='alert'>[user] begins melting [src] open with [S]!</span>",\
+				"<span class='alert'>You begin melting [src] open with your [S.name]!</span>")
 			playsound(user.loc, 'sound/items/Welder2.ogg', 40, 1)
 			actions.start(new /datum/action/bar/private/welding(user, src, 10 SECONDS, /obj/machinery/door/airlock/proc/sword_open, null, null, positions[1], positions[2]),user)
 	else
@@ -1626,8 +1630,7 @@ About the new airlock wires panel:
 /obj/machinery/door/airlock/proc/sword_open()
 	src.operating = -1
 	playsound(src, 'sound/machines/airlock_break_very_temp.ogg', 40, 1)
-	var/obj/item/scrap/S = new /obj/item/scrap
-	S.set_loc(src.loc)
+	new /obj/item/scrap(src.loc)
 	play_animation("opening")
 	src.UpdateIcon(1)
 	src.set_density(0)
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -1630,15 +1630,7 @@ About the new airlock wires panel:
 	src.operating = -1
 	playsound(src, 'sound/machines/airlock_break_very_temp.ogg', 40, 1)
 	new /obj/item/scrap(src.loc)
-	play_animation("opening")
-	src.UpdateIcon(1)
-	src.set_density(0)
-	if (!istype(src, /obj/machinery/door/airlock/glass))
-		if (ignore_light_or_cam_opacity)
-			src.set_opacity(0)
-		else
-			src.RL_SetOpacity(0)
-	src.UpdateIcon()
+	open(TRUE)
 
 /obj/machinery/door/airlock/proc/unpowered_open_close()
 	if (!src || !istype(src))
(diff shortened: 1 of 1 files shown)


code/obj/machinery/door/airlock.dm
```diff
@@ -1610,6 +1623,21 @@ About the new airlock wires panel:
...
+	S.set_loc(src.loc)
+	play_animation("opening")
+	src.UpdateIcon(1)
+	src.set_density(0)
+	if (!istype(src, /obj/machinery/door/airlock/glass))
+		if (ignore_light_or_cam_opacity)
+			src.set_opacity(0)
+		else
+			src.RL_SetOpacity(0)
+	src.UpdateIcon()
```
TobleroneSwordfish, line 1639: why not just call the `open` proc like emags do?

code/obj/machinery/door/airlock.dm
```diff
@@ -1610,6 +1623,21 @@ About the new airlock wires panel:
 		src.welded = null
 	src.UpdateIcon()
 
+/obj/machinery/door/airlock/proc/sword_open()
+	src.operating = -1
+	playsound(src, 'sound/machines/airlock_break_very_temp.ogg', 40, 1)
+	var/obj/item/scrap/S = new /obj/item/scrap
+	S.set_loc(src.loc)
```
TobleroneSwordfish, line 1630:
```suggestion
	new /obj/item/scrap(src.loc)
```

code/obj/machinery/door/airlock.dm
```diff
@@ -1594,6 +1594,19 @@ About the new airlock wires panel:
...
+		if (src.cant_emag)
+			boutput(user, "<span class='alert'>[src] is too resistant to melt from the saber!</span>")
+			return
+		else if (!src.density)
+			boutput(user, "<span class='alert'>You don't need to break open [src], as it's already open!</span>")
+			return
+		else
+			var/positions = src.get_welding_positions(user)
+			user.visible_message("<span class='alert'>[user] is melting [src] open with a cyalume saber!</span>",\
+				"<span class='alert'>You are melting [src] open with your cyalume saber!</span>")
```
TobleroneSwordfish, line 1607:
```suggestion
			user.visible_message("<span class='alert'>[user] begins melting [src] open with [C]!</span>",\
				"<span class='alert'>You begin melting [src] open with your [C.name]!</span>")
```

code/obj/machinery/door/airlock.dm
```diff
@@ -1594,6 +1594,19 @@ About the new airlock wires panel:
 		return src.Attackhand(user)
 	else if (ispryingtool(C))
 		src.unpowered_open_close()
+	else if (C.type == /obj/item/sword && C:can_reflect() && user.a_intent == INTENT_HELP && !( src.operating ))
```
TobleroneSwordfish, line 1597: probably makes more sense to check `active` directly, also `:` bad!

TobleroneSwordfish (changes requested):
⭐ wars

flrsh:
20 seconds to bash open a door seems reasonable to me, especially when doors are often one of the only things between you and your victim. I don't think csabers need this buff. However, adding an additional traitor item that does this would be fine imo.

Chatauscours:
This is mainly for the reference rather than balance.
Also, I forgot to mention it, but 20 seconds is for the armoury doors, we are looking at about 12-14 seconds for regular doors.
That said, if this mini balance change is unwelcome, the length of the melting could be increased to match the bashing time.

flrsh:
> This is mainly for the reference rather than balance. Also, I forgot to mention it, but 20 seconds is for the armoury doors, we are looking at about 12-14 seconds for regular doors. That said, if this mini balance change is unwelcome, the length of the melting could be increased to match the bashing time.

Me who never watched Star Wars :sheltersweat:

Increasing the length of the melting time to be equivalent works for me!

code/obj/machinery/door/airlock.dm
```diff
@@ -1594,6 +1594,23 @@ About the new airlock wires panel:
 		return src.Attackhand(user)
 	else if (ispryingtool(C))
 		src.unpowered_open_close()
+	else if (C.type == /obj/item/sword && user.a_intent == INTENT_HELP && !( src.operating ))
```
Mister-Moriarty, line 1597:
```suggestion
	else if (istype(C, /obj/item/sword) && (user.a_intent == INTENT_HELP) && !src.operating)
```

code/obj/machinery/door/airlock.dm
```diff
@@ -1594,6 +1594,23 @@ About the new airlock wires panel:
...
+		var/obj/item/sword/S = C
+		if (!S.active)
+			boutput(user, "<span class='alert'>[S] needs to be on to melt the [src]</span>")
+			return
+		if (src.cant_emag)
+			boutput(user, "<span class='alert'>[src] is too resistant to melt from the saber!</span>")
+			return
+		else if (!src.density)
```
Mister-Moriarty, line 1605:
```suggestion
		if (!src.density)
```
Redundant `else`; the above `if` statement has an early return.

code/obj/machinery/door/airlock.dm
```diff
@@ -1594,6 +1594,23 @@ About the new airlock wires panel:
 		return src.Attackhand(user)
 	else if (ispryingtool(C))
 		src.unpowered_open_close()
+	else if (C.type == /obj/item/sword && user.a_intent == INTENT_HELP && !( src.operating ))
+		var/obj/item/sword/S = C
+		if (!S.active)
+			boutput(user, "<span class='alert'>[S] needs to be on to melt the [src]</span>")
```
Mister-Moriarty, line 1600:
```suggestion
			boutput(user, "<span class='alert'>[S] needs to be on to melt the [src]!</span>")
```

code/obj/machinery/door/airlock.dm
```diff
@@ -1594,6 +1594,23 @@ About the new airlock wires panel:
...
+			return
+		else if (!src.density)
+			boutput(user, "<span class='alert'>You don't need to break open [src], as it's already open!</span>")
+			return
+		else
+			var/positions = src.get_welding_positions(user)
+			user.visible_message("<span class='alert'>[user] begins melting [src] open with [S]!</span>",\
+				"<span class='alert'>You begin melting [src] open with your [S.name]!</span>")
+			playsound(user.loc, 'sound/items/Welder2.ogg', 40, 1)
+			actions.start(new /datum/action/bar/private/welding(user, src, 10 SECONDS, /obj/machinery/door/airlock/proc/sword_open, null, null, positions[1], positions[2]),user)
```
Mister-Moriarty, line 1613:
```suggestion
		var/positions = src.get_welding_positions(user)
		user.visible_message("<span class='alert'>[user] begins melting [src] open with [S]!</span>",\
			"<span class='alert'>You begin melting [src] open with your [S.name]!</span>")
		playsound(user.loc, 'sound/items/Welder2.ogg', 40, 1)
		actions.start(new /datum/action/bar/private/welding(user, src, 10 SECONDS, /obj/machinery/door/airlock/proc/sword_open, null, null, positions[1], positions[2]),user)
```
Redundant `else`; the above `if` statement has an early return.

Mister-Moriarty (changes requested):

Chatauscours:
> Me who never watched Star Wars :sheltersweat:
> 
> Increasing the length of the melting time to be equivalent works for me!

I've made it so the time for melting now scales with the amount of HP a door has, so we always land on times similar to bashing.

code/obj/machinery/door/airlock.dm
TobleroneSwordfish: Maybe pass a `force` param to `open` then? It's not great to duplicate the open behaviour here

Chatauscours:
Added a force param to the `open` proc as Leah suggested. Set to `FALSE` by default, if a `TRUE` is given when the proc is called, then all other conditions get bypassed.

code/obj/machinery/door/airlock.dm
```diff
@@ -1666,8 +1688,8 @@ About the new airlock wires panel:
 
 	return
 
-/obj/machinery/door/airlock/open()
-	if (!src.density || src.welded || src.locked || src.operating == 1 || (!src.arePowerSystemsOn()) || (src.status & NOPOWER) || src.isWireCut(AIRLOCK_WIRE_OPEN_DOOR))
+/obj/machinery/door/airlock/open(force = FALSE) //force bypasses doors being welded, bolted, etc
+	if (force == FALSE && (!src.density || src.welded || src.locked || src.operating == 1 || (!src.arePowerSystemsOn()) || (src.status & NOPOWER) || src.isWireCut(AIRLOCK_WIRE_OPEN_DOOR)))
```
TobleroneSwordfish, line 1692:
```suggestion
	if (!force && (!src.density || src.welded || src.locked || src.operating == 1 || (!src.arePowerSystemsOn()) || (src.status & NOPOWER) || src.isWireCut(AIRLOCK_WIRE_OPEN_DOOR)))
```

code/obj/machinery/door/airlock.dm
```diff
@@ -1594,6 +1594,22 @@ About the new airlock wires panel:
 		return src.Attackhand(user)
 	else if (ispryingtool(C))
 		src.unpowered_open_close()
+	else if (istype(C, /obj/item/sword) && (user.a_intent == INTENT_HELP) && !src.operating)
```
TobleroneSwordfish, line 1597:
```suggestion
	else if ((C.type == /obj/item/sword) && (user.a_intent == INTENT_HELP) && !src.operating)
```
Unless you want the discount variant to also melt through doors, but that would mess with the melt vs bash timings since it's a lot weaker. (yes it sucks that csabers are just `/obj/item/sword`)

TobleroneSwordfish (changes requested):

flappybatpal:
Why do csabers get an additional mechanic that katanas and red saws can't do? These are all 7 TC items.

TobleroneSwordfish (approved):
Code good, the concept seems fine to me since you're essentially trading the stamina you'd lose bashing the door for doing a lengthy interruptible action bar.

virvatuli-gh:
Agreed with Tarm, this makes zero sense conceptually, c-sabres are not laser swords and shouldn't be, the whole gimmick is that they're Excessively Not From Star Wars And That's Funny

ju45he:
Instead it should perma ban you for attempting to melt doors with something that obviously isn't hot.

HatNJacket:
Something to keep in mind is that bashing open a door is very loud and distinct. Assumedly, this animation would be near silent, or at least silent after the initial cutting noise. I feel like giving the c-saber the ability to open any door without much noise on top of being _a c-saber_ might be a bit worrying. If this was to go through, maybe give it a distinct sound and have it be audible throughout the animation?

PR #10922, opened 2022-09-20
New Command Traitor Item: AI Laser Upgrade Modules

Wisemonster:
[GAME OBJECTS] [FEATURE] [INPUT WANTED] [EXPERIMENTAL]

## About the PR 
This pr adds a new item for traitorous heads of staff, an upgrade module that lets the AI use a laser ability from their cameras. The laser module shoots a 20 damage phaser shot, with a 10 second cool down, that costs 6 tc.

## Why's this needed? 
AI ability upgrade modules have been in the code for a while now, and I figured this ability wouldn't really fit as anything other than a traitor item. Plus gives traitor members of command more control of how a rogue AI can help them.

## Changelog

```changelog
(u)Wisemonster
(*)Adds a new command traitor item: An AI laser upgrade module that costs 6 tc.

PR #9158, opened 2022-06-15
AI Abilities

Azrun:
[experimental]
## About the PR 
Friend computer would like to help ~hurt~ you.

Provides _framework_ for giving abilities via `/obj/item/aiModule` to attached AIs.

* Taser Module
* Laser Module
https://streamable.com/m0ff3u
* Metal Foam Launcher Module
https://streamable.com/5uakvf
* Teleporter Module (Easy Send and Receive)
https://streamable.com/5a6pvl
* Camera and Silicon Repair Module
* Security Vision
* Doctor Vision
https://streamable.com/ulj322

## Why's this needed? 
Gimmickry and possible future integration into general play.  Would allow for crew to have some control over how they want the AI to best serve the station.  Additionally a vector for rogue AIs to have something other to do than explode lights and electrify/bolt doors.

PR #13750, opened 2023-04-19
New Command Traitor Item: AI Proto-Teleman module

Wisemonster:
[GAME OBJECTS] [FEATURE] [INPUT WANTED]

## About the PR 
Adds a new item for traitorous captains and RDs, the prototype teleporter module. This module let's AIs send things to and from the science teleporter, with some restrictions on delivery locations.

## Why's this needed? 
See #10922

## Changelog 

```changelog
(u)Wisemonster
(*)Adds a new command traitor item: A prototype teleporter module that costs 6 tc. Available for the captain and research director.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -1003,6 +1003,16 @@ This is basically useless for anyone but miners.
 	job = list("Captain", "Head of Personnel", "Research Director", "Medical Director", "Chief Engineer")
 	can_buy = UPLINK_TRAITOR
 
+/datum/syndicate_buylist/traitor/ai_teleport
+	name = "AI Proto-Teleman Module"
+	item = /obj/item/aiModule/ability_expansion/proto_teleman
+	cost = 6
+	vr_allowed = FALSE
+	not_in_crates = TRUE
+	desc = "An AI module that upgrades any AI connected to the installed law rack access to a targeted teleport system using the on-station science teleporter"
+	job = list("Captain", "Research Director")
+	can_buy = UPLINK_TRAITOR
+
 /////////////////////////////////////////// Surplus-exclusive items //////////////////////////////////////////////////
 
 ABSTRACT_TYPE(/datum/syndicate_buylist/surplus)


nefarious6th:
Demo and ability: https://github.com/goonstation/goonstation/pull/9158

This is unbalanced, and by a lot.
- easy murder with no trace left for security or others to determine causes (I can move someone in seconds into the burn chamber without interacting with them)
- very little counterplay; when teleportation becomes quickened like this, there is nothing to be done to counter it. 
- it's low-cost for what it does but beyond that it's not balanced for compatibility with the laser module also. If you trap someone in an access-restricted area, you then also can get rid of them in a matter of seconds. this goes for any sort of trapping and shooting. More on the specific workout of the timings in the mechanical finesse stuff below; but the two examples I can think of would be 1.) I trap someone in the teleporter room and over a duration of time, I shoot them while they are unable to escape, alternatively 2.) I teleport someone to the AI upload while the turrets are active and set to lethal, and they can take a maximum of three laser hits in a matter of seconds (two 80-damage ones from turrets, one 20-damage from module). You can move someone from full health to crit by doing so with just one turret and the module.
- likewise with the balancing, I dont think this is balanced in relation to AI engineering shells either. make a confined space and leave a shell there, teleport someone to it, transfer to shell and kill them, repeat.
- I dont want to see people pre-emptively mass-cut ai cameras to deal with this. That's sort of one tradeoff for the laser module, you can get shot by a camera but still be in the area and run up to it and cut it. With this, you can't run up to the camera to disable it because you aren't there anymore. Notably, cutting the camera in the teleporter itself doesn't do anything to hinder the AI, since the ability doesn't require interfacing with any of the systems there.

some mechanical finesses related to the above:
- from what I am able to see, there's no currently coded restrictions on delivery locations other than the inherent "whatever the AI camera can/can't see" (and maybe even not that?). you will have AIs teleport people into the (closed) crusher in disposals.
- if time is the biggest factor in counter play abilities, it's worth acknowledging that there is no delay between the teleport and receive functions on this. it's a hypothetically instantaneous set of actions (reality is it'd be like 2 seconds for someone to move the mouse up there and click). 2 seconds probably doesnt allow the person who's been teleported to the pad to at least figure out what's up and move off the pad, much less escape the teleporter if the blast doors are down. you can very, very easily trap someone indefinitely if they don't have the access to get through the door. The cooldown for the abilities when used is 10 seconds and to break a door with an extinguisher, it's a lot longer than that. even if the AI can't follow up with physical action to get rid of a crewmember, you can repeatedly teleport and re-teleport them around to continue disrupting them.
- cooldown as 10 seconds poses similar concerns as above; easy to pick people off without much lag time between killing/isolation. 

Finally, a story to sort of illustrate:
I am an engineer. I am on Nadir and the station is flooding with acid at the edges, from maintenance. People are moving to the medical hallway with no idea what's going on. I'm on the right side of the entrance to medbay, hall doors are electrified and bolted, and now I know what's going on. The AI shoots me with the laser module, and I figure out where the shot comes from and go to clip the camera. I then have to clip all the other cameras around medbay because clipping one doesnt matter when their coverage is redundant with others. And the AI, an experienced player, sees this and uses the teleporter to teleport me away and trap me in the teleporter room. I try removing the window and grill but in the process am re-teleported back to the tele pad. I instead opt to wrench the floor under the teleporter, move a tile, pry the floor, move a tile and see the teleporter activate where I was two tiles before, and snip the wire. Then I return to the window and removing it. The person playing AI, upon attempting to teleport me that third time, had the coordinates to send me into the ocean readied. The delay of manual entry gave me time to fight back.

Wisemonster:
Hmm, it seems that although there are some restrictions coded in (can't tele into space, areas without atmos, solid objects), there aren't actually enough restrictions to keep this from being very broken. Also, I think AI abilities are bugged atm, because the abilities aren't actually going on cooldown.

PR #13759, opened 2023-04-20
Makes it so underage drinking relies on sec records and not true age

Chatauscours:
[STATION SYSTEMS] [QOL]

## About the PR 
Basically makes it check sec records for the age of a person, istead of checking the bioholder age.

## Why's this needed? 
Because it makes no sense cameras know how old that stoaway is, also lets security twist records a little bit if an underage drinking problem becomes excessive.

## Changelog 

None needed


DisturbHerb:
I would personally recommend a minor changelog for this one.

Tarmunora (approved):
ambivalent but code looks fine

PR #13764, opened 2023-04-20
Window damage overlays

Flaborized:
[FEAT]

## About the PR 
Adds damage overlay for windows. At 90% health a little crack appears, then a medium crack at 60%, and a big crack at 85% (to show you're just about to break it)

Plasmaglass looked weird with the default opacity so they have special reduced alpha (since they're so common they get a hardcoded fix)

Only applies to the auto walls!! so the big ones in the above screenshot. 

## Why's this needed? 

damage icons are cool! helps visually illustrate how messed up windows are at a glance, and makes destroyed areas look a bit cooler.

## Changelog 

```changelog
(u)Flaborized
(+)Full tile windows now have damaged sprites.
```


code/obj/window.dm
```diff
@@ -31,6 +31,8 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
 	var/reinf = 0 // cant figure out how to remove this without the map crying aaaaa - ISN
 	var/deconstruct_time = 1 SECOND
 	var/image/connect_image = null
+	var/image/damage_image = null
+	var/damage_severity = null //used to determine which damage overlay to use if using them
```
TobleroneSwordfish, line 35: this probably doesn't need to be a var since it's recalculated every time?

TobleroneSwordfish (approved):
Yes!

TheTyrant25:
Screenshot_20230421_080628_Firefox
I assume that by 85% you mean damage and not health (i.e. 15% health)? Because the way you put 90, 60, 85 is confusing at first glance

code/obj/window.dm
Flaborized: I moved the var down to the area where it's actually used (doesn't make sense to be with the others that are used *everywhere*) but I like using it as a variable cuz I can't think of a way to do it that's more readable (though it probably exists!)

Flaborized: wait no figured it out like instantly after leaving this comment and changed it from using a variable to Not 👍

Flaborized:
> Screenshot_20230421_080628_Firefox I assume that by 85% you mean damage and not health (i.e. 15% health)? Because the way you put 90, 60, 85 is confusing at first glance

yeah I meant 15%!! my brain do math

code/obj/window.dm
```diff
@@ -176,6 +177,7 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
 			qdel(src)
 		else if (src.health == 0 && !nosmash)
 			smash()
+		update_icon()
```
pali6, line 180: I wish we had some linter check for this. Basically `update_icon` is the internal proc that you should override and `UpdateIcon` is the external proc that you should call. It does some additional bookkeeping and calls signals that may listen for icon updates.
```suggestion
		UpdateIcon()
```

code/obj/window.dm
```diff
@@ -189,6 +191,7 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
 		src.health = clamp(src.health - amount, 0, src.health_max)
 		if (src.health == 0)
 			smash()
+		update_icon()
```
pali6, line 194:
```suggestion
		UpdateIcon()
```

code/obj/window.dm
```diff
@@ -202,6 +205,7 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
...
```
pali6, line 208:
```suggestion
		UpdateIcon()
```

code/obj/window.dm
```diff
@@ -213,6 +217,7 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
...
```
pali6, line 220:
```suggestion
		UpdateIcon()
```

code/obj/window.dm
```diff
@@ -231,6 +236,7 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
 				qdel(src)
 			else
 				smash()
+		update_icon()
```
pali6, line 239:
```suggestion
		UpdateIcon()
```

code/obj/window.dm
```diff
@@ -836,6 +845,26 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
 		for (var/obj/grille/G in orange(1,src))
 			G.UpdateIcon()
 
+	proc/update_damage_overlay()
+		var/health_percentage = health/health_max
+		if (!src.damage_image)
+			src.damage_image = image('icons/obj/window_damage.dmi')
+			src.damage_image.appearance_flags = RESET_COLOR | RESET_ALPHA
```
pali6, line 852: This makes it so when scaled / transformed this overlay doesn't get bilinearly interpolated which looks blurry and doesn't fit with pixel art.
```suggestion
			src.damage_image.appearance_flags = PIXEL_SCALE | RESET_COLOR | RESET_ALPHA
```

pali6 (changes requested):

BatElite:
What does regular glass look like with the reduced opacity? I feel it might look nicer

code/obj/window.dm
Flaborized: ah yeah I forgot this! thanks!! (for the code review in general 🐝 )

Flaborized:
> What does regular glass look like with the reduced opacity? I feel it might look nicer




like this. (this is at the same value as plasmaglass)

I prefer the higher alpha version since I think it being much more visible is and contrast-y is more functional, and it works mostly with plasmaglass since the white overlay contrasts very well with the more saturated plasma glass.

pali6 (approved):

PR #9366, opened 2022-06-27
Initial framework/system for datumization of antags

Ilysen:
[CLEANLINESS]

## About the PR 

TODO - I'm about to eat dinner as I write this but I really want to get the PR up for diffs to mark as a draft. I'll fill this out in a few!

## Why's this needed? 

TODO

## Changelog

```changelog
(u)Ilysen
(*)Began a refactoring/update of the antagonist system in the code. This should result in only a few user-facing changes (ideally!) but it might cause bugs, unforeseen interactions, or other changes. Please keep an eye out while playing and report bugs on GitHub!
```

PR #13774, opened 2023-04-21
Datumises Mindhacks

Mister-Moriarty:
[Gamemodes] [Internal] [Code Quality]

## About the PR:
Mindhacks have been fully migrated onto the new datum system, which only includes implanted and admin-created mindhacks.
Additionally removes the `remove_mindhack_status()` system in favour of the new datums.

## Why's this needed?
Same rationale as #9366.


TobleroneSwordfish (changes requested):
- Currently this lets mindhacks cut out their own implant, which they should not be able to do. The mindhack antag status removal should be in the implant's `on_remove` and the surgery code should skip the mindhack implant if the patient is the surgeon.
- The mindhacked status effect should be removed on the antagonist removal so they stay synched, currently if you revive a dead mindhack they would still have the status effect but not the antag datum.
- Being cloned with a mindhack cloning pod results in this confusing mess
image
along with a popup that says you're not a mindhack, a status effect that says you are and no antag datum.

TobleroneSwordfish (approved):

PR #13365, opened 2023-03-15
Mindhack rebalance

TobleroneSwordfish:
[BALANCE] [FEATURE] [FREEZE-EXEMPT] [CONTROVERSIAL] [FEEDBACK] [WIKI]
## About the PR 
All security officers (not assistants or detectives) get the anti-mindhack health implant currently reserved for HoSes.

Adds a new 2 TC traitor item available to traitors, spy-thieves and nuke-op custom class uplinks: implant remover.
After a 3 second public action bar and obnoxiously loud noise, the implant remover purges all non-syndicate implants from the target, dealing 10 burn damage per implant removed.

https://i.imgur.com/0qwdjA2.mp4

The sprite and sound effect are placeholder, if this looks like getting merged me or someone else can make an actual sprite for it.
## Why's this needed? 
One traitor immediately mindhacking the entire security team generally doesn't make for very entertaining rounds, this aims to make mindhacking security officers much more challenging but still possible if you can remove the implant. Cloned security officers will also be vulnerable to mindhacking unless they are able to recover the implant from their corpse.

Standard disclaimer that although I'm a dev, this is far from guaranteed to be merged.

## Changelog 

```changelog
(u)LeahTheTech
(*)All security officers (not assistants) now receive anti-mindhack implants.
(*)New 2TC traitor item: implant remover. Removes all non-syndicate implants from a person after a short delay.
```

PR #13103, opened 2023-02-16
Health implants no longer send the coords of the implantee.

Wisemonster:
[BALANCE] [STATION SYSTEMS] [INPUT WANTED]

## About the PR 
Changes the health implants to no longer send the cords of the implantee. This applies to the normal and security ones.

## Why's this needed? 
Sec currently has a lot of things that let them know when an officer is in trouble, including announcing over coms, a button that sends an alert, and 2 alerts from their health implants (1 when they are in crit, another when they die.) This aims to give antags that ambush a sec off some more breathing room before sec finds them.

## Changelog 

```changelog
(u)Wisemonster
(*)The health implants no longer send the coordinates of the implantee on an alert. The area is still sent.
```

PR #13782, opened 2023-04-21
Gives all security officers mind protection implants

TobleroneSwordfish:
[BALANCE] [WIKI]
## About the PR 
All security officers now get the previously HoS exclusive mind protection health implant, which now destroys itself on death. 
The description of both regular and deluxe mindhack implants in the traitor buy menu has been updated to mention mind protection implants.
The implant is visible when scanned using any kind of health analyzer.

This is a simpler remake of #13365, taking into account feedback received on it.

## Why's this needed? 
- Mindhacks on classic are almost never used on anything other than security.
- The entire security team being mindhacked or murdered by the ones who are happens so often it's practically a meme at this point.
- #13742 and #13103 nerfed security's access to immediate information about antag activity, and b0ac5c386b94bfa64ede0ac3559967e3cf719185 and other adjustments nerfed their weaponry slightly. These changes should hopefully allow for the addition of mind protection implants without making security too oppressive when highly populated.

## Changelog 

```changelog
(u)LeahTheTech
(*)All security officers including NTSCs (not assistants) now receive mind protection health implants. Said implants now self-destruct upon death.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -2536,7 +2536,7 @@ ABSTRACT_TYPE(/datum/job/special/halloween/critter)
 	cant_spawn_as_rev = 1
 	receives_badge = 1
 	receives_miranda = 1
-	receives_implant = /obj/item/implant/health/security
+	receives_implant = /obj/item/implant/health/security/anti_mindhack
 	slot_back = list(/obj/item/storage/backpack/NT)
 	slot_belt = list(/obj/item/storage/belt/security/ntsc) //special secbelt subtype that spawns with the NTSO gear inside
 	slot_jump = list(/obj/item/clothing/under/misc/turds)
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -447,7 +447,7 @@ ABSTRACT_TYPE(/datum/job/security)
 	allow_spy_theft = 0
 	cant_spawn_as_con = 1
 	cant_spawn_as_rev = 1
-	receives_implant = /obj/item/implant/health/security
+	receives_implant = /obj/item/implant/health/security/anti_mindhack
 	receives_disk = 1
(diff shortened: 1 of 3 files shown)


ThatFiveGuys:
Whats the point of them self destructing on death? Cloning removes implants anyways and if its for people to use sr with them if theyre going to be that prepared they can grab a scapel alongside it

Wisemonster:
> Whats the point of them self destructing on death? Cloning removes implants anyways and if its for people to use sr with them if theyre going to be that prepared they can grab a scapel alongside it

Might be to prevent sec from just harvesting the implant from their previous body and re-implanting themselves with it.

TobleroneSwordfish:
Yep, it means a security officer who gets killed and cloned is now vulnerable to mindhacking, rather than being able to retrieve the implant from their corpse. This means killing a security officer still achieves something even if they immediately get cloned.

Chatauscours:
Good change IMO, however I believe it would be a good idea to also give the implant to the NTSCs, as they are already THE perfect mindhack target.

TobleroneSwordfish:
added

Tarmunora (approved):

Sord213:
HoS should (imo) get a single implant in their locker or the armory lockers in case officers die, but not have large numbers to replenish the entire team.

TobleroneSwordfish:
I think it's good to keep it to just the ones they start with for now, that way you know for (almost) certain that any secoff killed and cloned won't have one, and you don't end up trying to implant a random crewmember only to find the HoS gave them an anti-mindhack implant.

ThatFiveGuys:
I still prefer my approach (though if i were to make a change Id make you gasp on crit + make the threshold crit), but i think this is a fine solution. May reopen my previous pr but we will see about that

PR #13784, opened 2023-04-21
Adds doorway from engineering core to mechanics on donut3

ThatFiveGuys:
## About the PR 
This pr adds a shortcut to donut3 to get to mechanics faster
image

## Why's this needed? 
Mechanics and engineering has been merged, maps should act like that as well.

## Changelog 

Not really needed tbh

### DIFF
--- a/maps/donut3.dmm
+++ b/maps/donut3.dmm
@@ -25896,10 +25896,6 @@
 	},
 /turf/simulated/floor,
 /area/station/hallway/primary/southeast)
-"hMu" = (
-/obj/table/auto,
-/turf/simulated/floor/yellow,
-/area/station/engine/elect)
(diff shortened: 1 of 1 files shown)


Sord213 (changes requested):
Could you use a non-glass airlock, so the room is still pretty good for crime without being able to peek in too easily from eng.

Cherman0:
Also so the room does not become horribly irradiated whenever theres a big singularity.

ThatFiveGuys:
> Could you use a non-glass airlock, so the room is still pretty good for crime without being able to peek in too easily from eng.

Done! Swapped the door from mechanics to a non glass door one as well as that seems like a good idea

PR #13785, opened 2023-04-22
Adds atmos barriers to external doors on cogmap2

DanaTarSUS:
## About the PR 
Adds doorlinked atmos barriers to most external doors on cogmap2. In addition, the emergency disposal mass driver on the southern end of research should now have breathable air at the start of the round.

## Why's this needed? 
Parity change similar to #13751, helps to prevent rooms from being depressurized because a door was left open.

## Changelog 

```changelog
(u)DeeDotVeeA
(+)Atmos barriers can now be found on most external airlocks on cogmap2
```

### DIFF
--- a/maps/cogmap2.dmm
+++ b/maps/cogmap2.dmm
@@ -2166,6 +2166,11 @@
 	name = "cargo driver"
 	},
 /obj/machinery/door/poddoor/pyro,
+/obj/forcefield/energyshield/perma/doorlink{
+	desc = "A door-linked force field that prevents gasses from passing through it.";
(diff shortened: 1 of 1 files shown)


RelentlessGarbage:
Kinda funny how all the airlocks are becoming obsolete in their function as airlocks now that there are door shields, but I think this is a good idea overall. I find that Cog2's mining department is especially prone to fully depressurizing, since everyone is running around in space suits all the time and running off of internals, and those double doors tend to stay open, so if they get door shields then that's a very good thing.

PR #13023, opened 2023-02-08
Add storage datums

FlameArrow57:
[INTERNAL][REWORK][CODE QUALITY]

## About the PR 
This PR adds storage datums, an alternative to storage components proposed in #860.

Storage datums are something that can be added to atoms to give them storage functionality, allowing anything to be a storage for items.

Items of type /obj/item/storage have their storage variables changed to just be an interface with their storage datums and their procs removed.

Bibles are reworked from /obj/item/storage/bible to /obj/item/bible, as Flourish did

Storage datums act pretty close to the same as /obj/item/storage does at the moment

This PR started from the work that Flourish initially did

Ex.
https://streamable.com/utdjca

## Why's this needed? 
It allows you to turn anything into a storage, which would allow storage functionality on items that aren't of a storage item type, helping with code quality and allowing new features

Fixes #7002, per comment from MetricDuck, fixes #11017, fixes #12090, fixes #12677

PR #13789, opened 2023-04-22
Add glasses equipped and unequipped checks for slot_glasses

FlameArrow57:
[INTERNAL][CLEANLINESS]

## About the PR 
This PR adds checks for glasses items for equipping and unequipping them to make sure that they actually are in the slot_glasses slot before applying their effects.

## Why's this needed? 
More meant for #13023 since that allows storage of items put in pockets, which relies on equipped() and unequipped(). Currently items equipped to pocket slots do not call equipped(), same for unequipping

### DIFF
--- a/code/WorkInProgress/ud5/urs_dungeon.dm
+++ b/code/WorkInProgress/ud5/urs_dungeon.dm
@@ -183,15 +183,15 @@
 
 
 	equipped(var/mob/user, var/slot)
-		..()
+		if (!..())
+			return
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
This is probably going to break a lot more things like magboots, stompers and such. Is it possible to just preserve the old behaviour of not calling `equipped` for pocket slots?

FlameArrow57:
Yeah

PR #13801, opened 2023-04-22
Don't hit mechcomp on setting custom configs

monkeymilesmooYT:
## About the PR 
This PR makes it so that changing the configs of a component doesn't hit it with the multitool.

## Why's this needed? 

No more being clumsy and smacking a clock when you try to change the time.


monkeymilesmooYT:
oh shoot I accidentally uploaded the wrong PR X_X

monkeymilesmooYT:
dumbass moment

PR #13803, opened 2023-04-22
Don't hit mechcomp on edit custom signal

monkeymilesmoo:
## About the PR 
Fixes hitting mechcomp with multitool when you edit a custom signal

## Why's this needed? 
Stop being clumsy and smacking your mechcomp!


TobleroneSwordfish:
Does this affect fingerprints left on stuff?

CheffieGithub:
Please label your PRs

monkeymilesmoo:
It might affect fingerprints, I'm not sure, but if it does, wouldn't the other functions also not leave fingerprints already?

TobleroneSwordfish:
Ah I see, it was backwards.

PR #13807, opened 2023-04-23
Adds Generic AIHolders 

CheffieGithub:
## About the PR 
Adds a new file in AI for general AIHolders that could be used for multiple critters, currently has aggressive wandering / scavenging but this is more of a base for future stuff like ranged attacking.

## Why's this needed? 
No more copy paste AIHolders with the same content


TobleroneSwordfish (approved):
looks good to me, less boilerplate

PR #13812, opened 2023-04-23
Updates chameleon gear and gives it hides from examine

CheffieGithub:
## About the PR 
Updates the roboticist and rancher disguise to use their own belt.
Updates the medical doctor disguise to use their labcoat.
Updates 0/1 to TRUE/FALSE.
Updates outersuit and hat to use hides from examine flags.

## Why's this needed? 
Some disguises are currently outdated and all lack hides from examine flags which is a dead givaway on certain ones


TobleroneSwordfish (approved):
Why is the chameleon suit like this, it's a maintenance nightmare!

Tarmunora (approved):
good lord this item is a nightmare

PR #11905, opened 2022-11-09
adds chemical barrels to the sciene fabricator

colossusqw:
[LABEL][balance][chemistry][feature]
## About the PR 
This PR adds a chemical barrel as a science fabricator option. The barrel can't be held but can be pulled, and stores a large amount of chemicals, shooting the barrel heats up the reagents inside so be careful handling explosives. It requires cobryl to manufacture, so science will need to ask for mining help after two barrels.

## Why's this needed? 
Gives science a integrated option to fabricate chems in larger quantities that isn't stealing the watercoolers from the station or stealing and emptying the fuel on every fuel tank.
It makes up for it's large volume by being extremely obvious and undwieldy.

```changelog
(u)Hans
(*)Science can now fabricate unwieldy reagent barrels to aid in chemical production.
```

PR #13814, opened 2023-04-23
The Automatic Chemical Dispenser

monkeymilesmoo:
[Chemistry] [Game-Objects] [Feature]

## About the PR 
dispenser
This PR adds a new machine, The Automatic Chemical Dispenser. It is able to be purchased by qm and chemists are able to set it up by following the package's included manual. The ACD is connectable to either a chemicompiler, in which case the reservoir number is set through mechcomp config, or to a fluid canister to store a large amount of base chemicals.

## Why's this needed? 
Allows for chemists to make a large scale production of chemicals. Also gives the ChemiCompiler more use since the reservoirs don't have to manually be filled up every time it is run, which is one of the reasons most people don't use it much.

## Changelog 

```changelog
(u)monkeymilesmoo
(*)New machine: The Automatic ChemDispenser which can be bought by qm and connect to components.
```


wolfolotl:
I'm pretty sure chemists not having the ability to manufacture most chemicals in large amounts is by design

monkeymilesmoo:
They can already do what this machine does, it just takes a long time. Plus, for someone to set up an actual large-scale operation would take a bit of effort since it would involve the chemicompiler.

Exester509:
I think this would remain unused most of the time due to it needing chemicompiler nerdery, and if you can set a proper factory with a chemicompiler already, this would be redundant.

I honestly don't see it as that big of a balance concern considering the chemical records are already decent at mass production of chemicals, and few need a steady, unstopped production.

monkeymilesmoo:
It probably would unused most of the time along with the compiler, but having it gives chem nerds who want to set up an automatic factory the option to do so without having to do the recording and manually doing half of the job themselves.

nefarious6th:
maybe see dev comments on #11905 for matters related to design philosophy with not being able to large-batch make chemicals

I dont feel a ways about this or the other PR, I do think that having both in together probably circumvents some essential limitation on what people can reasonably do with their gameplay that might make chem very unbalanced? that's for someone else to decide, I just thought there was interesting proximity between both PRs and what they offer chem

monkeymilesmoo:
Hmm, I do have a feature in this to be able to fill a fluid canister, which is similar to that. If wanted, I could just remove that and only have the ChemiCompiler since it can only output 100u at a time and the beaker would fill up inside the reservoir.

Exester509:
> It probably would unused most of the time along with the compiler, but having it gives chem nerds who want to set up an automatic factory the option to do so without having to do the recording and manually doing half of the job themselves.

That was a reason why the old chemgroup system got removed. It was just a roundstart copypaste and you could start mass producing generally powerful, sometimes secret chems at an alarming rate without even having to go through all the trouble of at least making it once. You could compress considerably long formulas into exactly two clicks.

And again, the crew doesn't really need any mass production of chemicals, there simply isn't a necessity. Giving chemists the ability to make a factory this easily without this really being needed isn't something I'm a fan of. Some mechanics are intentionally hard, either as a way to prevent abuse or to gatekeep newcomers who have barely done any experimentation from outputting something similar to what more experienced players have dumped tons of hours into optimizing

PR #13828, opened 2023-04-24
Moves Cogmap2 mechanics to engineering

ThatFiveGuys:
## About the PR 
This pr moves mechanics closer to the core of engineering
image
image
image

## Why's this needed? 
Mechanics and Engineers are the same job now, the maps should treat them that way.

## Changelog 

```changelog
(u)Ikea
(*)On cogmap2, mechanics has been moved closer to the teg. Expect some minor shuffling around of things.
```

### DIFF
--- a/maps/cogmap2.dmm
+++ b/maps/cogmap2.dmm
@@ -22843,22 +22843,6 @@
 /obj/disposalpipe/segment/transport,
 /turf/space,
 /area/space)
-"beZ" = (
-/obj/grille/catwalk{
-	dir = 4
-	},
-/obj/machinery/atmospherics/pipe/simple/color_pipe/green_pipe{
-	dir = 4
-	},
-/obj/disposalpipe/segment/food{
-	dir = 1;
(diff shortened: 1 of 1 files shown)


ThatFiveGuys:
Bleh think my strongdmm is out of date leading to some whacky mapping behavior, Ill fix all of that  stuff later tm

Tarmunora (changes requested):
After some discussion in admin channels, we've decided we'd rather see something more along the lines of a transit chute between cog2 mech/engineering rather than moving the department.

TobleroneSwordfish:
To add some context:
Moving mechanics out of there leaves mining very isolated, which is a shame since cog 2 mining is one of the more busy mining departments right now, being across the hall from one of the main uses of their work. Also the new area is more cramped than the old one, having less floor space for mechcomp machinery.

ThatFiveGuys:
Hmm okay gonna close this pr and make a new one with pipe

Issue #13636, opened 2023-04-10
Monkeys picking up body bags


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Monkeys can pick up closed body bags with items/corpses inside and give them to you

### Steps to reproduce

1.Put items/corpses in a body bag
2.Let a monkey pick it up
3.Wait until the monkey offers you the body bag

### Expected Behavior

Monkeys should be unable to pick up filled body bags just like you yourself can't

### Additional Information & Screenshots

Reported by: Null200
On server: Goonstation 1 Classic: Heisenbee
Active test merges: nothing
Round log date: 2023-04-10-19-36
Reported on: 2023-04-10 19:58:56
Map: COGMAP

Retrino:
shoul d be fixed in #13842

Issue #13793, opened 2023-04-22
Monkey can pick up bulky items that arent supposed to be picked up


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Monkey can pick up Old Piano

### Steps to reproduce

1.Unbolt a piano from chaplin with a screwdriver
2.move the piano near a monkey
3. wait or beat the monkey
4. watch it pick up the old piano

### Expected Behavior

Idk I think monkeys aint hulk right :))))))

### Additional Information & Screenshots

https://cdn.discordapp.com/attachments/890223070999101480/1099197844780490813/image.png

Reported by: SenterNoir
On server: Goonstation 1 Classic: Heisenbee
Active test merges: #13023, #13667, and #13773
Round log date: 2023-04-22-04-39
Reported on: 2023-04-22 05:02:56
Map: OSHAN

frawhst:
why are pianos items

frawhst:
oh god if it's an instrument subtype

frawhst:
we live in a society
https://github.com/goonstation/goonstation/blob/b44c91d64485f604024d60500af4ea37049fe944/code/obj/item/instruments.dm#L271

DisturbHerb:
apes together strong

RelentlessGarbage:
What happens if the monkey throws the old piano at you?

RelentlessGarbage:
Drat. I was hoping that it'd collapse dramatically, as if it had fallen off a building.

monkeymilesmoo:
![](https://cdn.discordapp.com/attachments/890226559691157524/1100220966744891442/NINTCHDBPICT000453604234-e1543852310346.png)

CheffieGithub:
Fixed by #13842

PR #13842, opened 2023-04-25
Fix monkeys picking up huge objects

monkeymilesmoo:
## About the PR 
Hopefully monkeys shouldn't be going around grabbing super huge objects now
should fix:
https://github.com/goonstation/goonstation/issues/13636
https://github.com/goonstation/goonstation/issues/13793

## Why's this needed? 
There's no way a monkey is stronger than me :,(

## Changelog 

```changelog
(u)monkeymilesmoo
(+)Monkeys are no longer bodybuilders
```

### DIFF
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -2979,7 +2979,7 @@
 		src.mind.transfer_to(newbody)
 	else //Oh welp, still need to move that key!
 		newbody.key = src.key
-
+	qdel(src)
 	////////////Now play the degibbing animation and move them to the turf.////////////////
 
 	var/atom/movable/overlay/animation = new(reappear_turf)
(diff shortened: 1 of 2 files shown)


Laboredih123:
accept it. monkey stronger. stop trying to stop monkey

monkeymilesmoo:
😢

monkeymilesmoo:
I can't code for nothin X_X

CheffieGithub:
W_CLASS_GIGANTIC is used for items that can't be picked up at all. Bulky is used for items which don't fit in backpacks

Tarmunora (approved):
hm. I think a more generic solution would be better, but this is a good enough hotpatch

PR #13848, opened 2023-04-26
Allow equip hotkey to auto-sheathe swords

FlameArrow57:
[GAME OBJECTS][PLAYER ACTIONS][QoL]

## About the PR 
This PR allows using the equip hotkey while having a sword in hand to auto-sheathe it

Also implements it in a generic way that allows you to do similar things for other items.

## Why's this needed? 
Faster, otherwise you have to manually click the sheath to sheathe the sword which is annoying

## Changelog 
```changelog
(u)FlameArrow57
(+)Swords will now auto-sheathe if using the equip hotkey while they're held in hand.
```

### DIFF
--- a/code/datums/hud/human.dm
+++ b/code/datums/hud/human.dm
@@ -390,7 +390,7 @@
 			if ("equip")
 				var/obj/item/I = master.equipped()
 				if (I)
-					if (I.try_specific_equip())
+					if (I.try_specific_equip(user))
 						return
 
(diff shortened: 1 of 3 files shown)


code/datums/hud/human.dm
```diff
@@ -387,6 +390,9 @@
 			if ("equip")
 				var/obj/item/I = master.equipped()
 				if (I)
+					if (I.try_specific_equip())
```
TobleroneSwordfish, line 393: shouldn't we be passing `user` here?

FlameArrow57: Oops yeah

Tarmunora (approved):

TobleroneSwordfish (approved):

PR #13852, opened 2023-04-26
allow all canisters to spawn empty

TheTyrant25:
[QOL] [MAPPING] [INTERNAL]
## About the PR
Adds a new var to each canister which lets it start empty.
## Why's this needed?
When mapping azones, it was previously not possible to spawn in plasma canisters or any other kind that were empty, as the gas contents are set in `New()`. This new var lets you start them off empty for derelict areas and such.


BatElite:
If we're talking azones, isn't it just as easy to varedit a regular blank canister rather than do all this?

Azrun (commented):
Consider use of `src.air_contents.vacuum()`

TheTyrant25:
> If we're talking azones, isn't it just as easy to varedit a regular blank canister rather than do all this?

nope, it isn't, because the contents get overwritten in New() to be full.

TheTyrant25:
> Consider use of `src.air_contents.vacuum()`

will do

BatElite:
> nope, it isn't, because the contents get overwritten in New() to be full.

No I mean, you take a `/obj/machinery/portable_atmospherics/canister/empty` and you change the name, icon_state, casecolor

Tarmunora (approved):
I do wonder if a more general solution would be better, e.g.: `var/spawn_filled = TRUE` (and associated handling in `New()`) on the canister parent, letting mapmakers spawn any canister as empty with a single var override

TheTyrant25:
yeah that was an odd choice by me. Changed it to use a var instead: `isempty`

ZeWaka:
You don't need to vacuum if there's no contents...

TheTyrant25:
i wasn't sure, played it safe. I'll remove that line then

PR #13855, opened 2023-04-26
Swaps Out Capsaicin Grenades In the Sec Vendor For Flashbangs

Wisemonster:
[BALANCE] [INPUT WANTED] [GAME OBJECTS]

## About the PR 
This pr swaps out the capsaicin grenades in the sec weapons vendor for flashbangs.

## Why's this needed? 
Since I reworked the protection for capsaicin, capsaicin grenades have become incredibly strong for how abundant they can be. This keeps the strength of the grenades, but limits them to the armory to keep them from being so abundant. I replaced them with flashbangs due to them not being nearly as strong, but still having some utility.

## Changelog 

```changelog
(u)Wisemonster
(*)Replaced the capsaicin grenades in the security weapon vendor with flashbangs.
```

### DIFF
--- a/code/obj/item/storage/weapon_security.dm
+++ b/code/obj/item/storage/weapon_security.dm
@@ -123,6 +123,9 @@
 	icon_state = "flashbang"
 	spawn_contents = list(/obj/item/chem_grenade/flashbang = 7)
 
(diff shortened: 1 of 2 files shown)


Tarmunora (approved):
I'm a bit wary of putting flashbangs back into the spotlight again, but I think it's probably a good change

PR #13863, opened 2023-04-27
Mobifies Repairbots

CheffieGithub:
## About the PR 
Mobifies ancient repair bots and its subtypes.
They will only attack people who aren't incapacitated and are immune to stuns but have little health.

## Why's this needed? 
Object critter bad.


kremlin-:
unsubscribe

TobleroneSwordfish (changes requested):
Please don't just shove the shade changes in here, it makes it much harder to review and merge.

CheffieGithub:
shade changes gone

TobleroneSwordfish (changes requested):
Killing the security variant seems to consistently cause this runtime from `hands.dm`:
image
Other than that this looks good, will require secret changes when merged.

PR #13882, opened 2023-04-28
Hacked Gas Extractors can craft N20 canisters

Piesuu:
[OBJECTS] [QOL]

## About the PR 
Hacked Gas extractors can now craft N20 canisters, the datum was there but it wasn't implemented

## Why's this needed? 
Makes N20 canisters more available and easier to get, some maps have one in the armory and others lack it altogether

## Changelog 

```changelog
(u)Piesuu
(+)N20 canisters can now be crafted in hacked gas extractors
```


CaioFrancisco:
By the way, as it currently is, a N2O requires the following materials to be manufactured:

- 0.3 Metal
- 1 Koshmarite

I feel like there should be some molitz in there as well, for the "oxide" part of the "nitrous oxide", but that may be part of another PR.

DimWhat:
This feels like a bad idea.

ThatFiveGuys:
N20 only kinda works because its not available, whereever its deployed it inevitably leads to lots of players falling into it and lings and other antags thriving. Being able to do this across the station seems like itd be way too benefical for too little work, this is definitely a bad idea

ThatFiveGuys:
Co2 is a lot slower to work then no2, personally i think it could be made more obvious to new players by adding an aphyxiation warning but n2o is definitely more dangerous then co2 despite the increased visibility.

M-Earthfire:
Personally, it requires a rather uncommon recource, is restricted to engineering and more visible than CO2. Given the almost non-existing sleep traps involving CO2, it does not sound too far off to make it avaible in a niche way

Sbmhawk:
CO2 is already available to produce, and I personally believe it is more meta than n2o. This allows for non-lethal sleep traps to be done easily, which I think is good.

PR #13886, opened 2023-04-28
Prevents Abominations From Being Mindhacked

Wisemonster:
[BALANCE] [INPUT WANTED] [GAME OBJECTS] [MUTANTRACES]

## About the PR 
Prevents abominations from being mindhacked. Does not affect changelings.

## Why's this needed? 
Considering abominations are these massive, loud threats, being able to shut them down with 1 click is really bad. This makes it so a random sec off/traitor with a mind hack on them can't just negate an active abomination.

## Changelog 

```changelog
(u)Wisemonster
(*)Abominations can no longer be mindhacked.
```


TobleroneSwordfish:
Kind of an odd exception for what would be a very rare funny edge case, don't see the need for it. Feel free to re-open if the mindhacked abomination meta happens somehow.

PR #13887, opened 2023-04-28
Oshan Cargo Redesign

DanaTarSUS:
[Mapping] [QoL] [INPUT]
## About the PR 

This pr completely alters the layout of the cargo area on Oshan, as well as Southwest Maints, Central Warehouse, and The Crew Lounge, all located within the southwest corner of the station. A screenshot of the changes can be found here:
https://gyazo.com/a4e2206c0f40e8956782f8a85b16b4d9

## Why's this needed? 

Oshan's cargo area currently has a problem where it's far too easy for QM players to shut themselves off in the backroom and not interact with the rest of the station while they do their work. This new layout fixes this problem by making it easier for both crew and cargo to interact with each other without trying to converse through the backroom shutters.

Additionally, this PR helps to turn the crew lounge found near this end of the station in a room that "almost kinda" resembles a lounge instead of a single sad looking table and stools.

## Changelog 

```changelog
(u)DeeDotVeeA
(*)Oshan Laboratory's Quartermasters have become so rich from reselling metal sheets to the market that they were able to purchase a brand new Cargo area. Good for them!
```

### DIFF
--- a/maps/oshan.dmm
+++ b/maps/oshan.dmm
@@ -140,35 +140,33 @@
 	},
 /area/station/medical/medbay/lobby)
 "aax" = (
+/obj/firedoor_spawn,
+/obj/decal/stripe_delivery,
 /obj/machinery/door/airlock/pyro/glass{
 	dir = 4;
-	name = "Supply Lobby"
+	name = "Quartermaster's Office"
 	},
+/obj/access_spawn/cargo,
 /obj/firedoor_spawn,
-/obj/decal/stripe_delivery,
 /turf/simulated/floor,
-/area/station/crew_quarters/market{
-	name = "Supply Lobby"
-	})
+/area/station/quartermaster/office)
 "aay" = (
-/obj/machinery/door/airlock/pyro/glass{
-	name = "Supply Lobby"
+/turf/simulated/floor/yellow/side{
+	dir = 1
 	},
-/obj/firedoor_spawn,
-/obj/decal/stripe_delivery,
-/turf/simulated/floor,
 /area/station/crew_quarters/market{
 	name = "Supply Lobby"
 	})
 "aaz" = (
-/obj/machinery/door/airlock/pyro/glass{
-	name = "Supply Lobby"
+/obj/disposalpipe/segment{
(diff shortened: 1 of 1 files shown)


nefarious6th:
just some really minor things
- did you happen to re-place a stock exchange PC somewhere accessible to the public? can't tell because of shared icon with req computers, sorry in advance 😔
- transit chutes are usually pink if they're meant to move you across the station; look to the crew quarters kitchen northeast of arrivals on Donut 3 to grab a copy of the pink chute subtype
- these two pipes going through exterior walls might be tricky to fix in the event of a break since removing the walls to access the pipes will cause flooding


crew lounge looks like a fun spot; I like the ability of people waiting at cargo to see both the intake/outgoing belts and the front desk at the same time

DanaTarSUS:
> * did you happen to re-place a stock exchange PC somewhere accessible to the public? can't tell because of shared icon with req computers, sorry in advance 😔

I did not end up re-adding stock exchange pc's. The feedback I got early on when debating to remove that area was that it was severely underused. That said, I can easily get rid of the 3x3 random room and add back some stock exchange stuff to that room if need be.

DanaTarSUS:
> * transit chutes are usually pink if they're meant to move you across the station; look to the crew quarters kitchen northeast of arrivals on Donut 3 to grab a copy of the pink chute subtype
> * these two pipes going through exterior walls might be tricky to fix in the event of a break since removing the walls to access the pipes will cause flooding

Yeah, I can go back and fix both the labeling and color of those chutes. Will probably make another edit soon.

DanaTarSUS:
Made another edit to simplify the pipes leading between either side of cargo. I wasn't aware you could get away with only using a single pipeline for a two way system. 
image
image

flrsh (changes requested):
I'm not interested in this redesign. If people feel like QMs are too inaccessible when they're in the lower room, I feel like that could be addressed just by adding windows or something, instead of reorganizing the entire section of the map.

Also, the area with the tables was never meant to be crew quarters, it was simply an area with tables.

PR #13888, opened 2023-04-28
Change how items that can fit on your back are equipped through the equip hotkey

FlameArrow57:
[PLAYER ACTIONS][QoL]

## About the PR 
This PR makes it so that using the equip hotkey with an item that can fit on your back, while wearing a backpack, will no longer swap them, but the item will be added to the backpack storage instead if you have the storage hud open.

This is how belts work too.

## Why's this needed? 
This change is mostly aimed at guns that can be held on your back, but figured the change can be done for other items that can do that too.

It's still common that people are expecting guns that fit on your back to be added to your backpack when using the equip hotkey, but its unexpectedly swapped with your backpack instead.

## Changelog 
```changelog
(u)FlameArrow57
(+)Items that can fit on your back will no longer auto-swap with worn backpacks when using the equip hotkey.
```

### DIFF
--- a/code/datums/hud/human.dm
+++ b/code/datums/hud/human.dm
@@ -341,7 +341,11 @@
 					autoequip_slot(slot_ears, ears)
 					autoequip_slot(slot_wear_mask, wear_mask)
 					autoequip_slot(slot_head, head)
-					autoequip_slot(slot_back, back)
+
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
I think this is probably more intuitive behaviour, but I am in the habit of switching air tanks onto my back with the hotkey 🤷

PR #13890, opened 2023-04-29
Adds A Define Related To Testing Antags/Weapon Vendors/Uplinks

Wisemonster:
[INTERNAL] [QOL]

## About the PR 
This pr adds a define that sets the points for various items very high. These include antag abilities, weapon vendors, uplinks, blob/wraith, and gang lockers. This does not include werewolf bodies eaten, and flock compute. I did not have these affected due to the werewolf not needing bodies eaten to use abilities, and flock compute influences the relay spawning.

## Why's this needed? 
Makes testing antags that rely on ability points much easier/faster to test, and makes testing changes to weapon vendors and uplinks easier/faster to test as well.


TobleroneSwordfish (approved):
approve, people can be angry at someone else for committing to build.dm :)

PR #13891, opened 2023-04-29
Adds a security ejection chute to the shuttles

Jak27:
## About the PR 

Adds ejection chutes that will safely kick out any staff assistants trying to squirm their way into the security room on the shuttle.
Thankfully it just goes into the main room of the shuttle, and not space.

First time trying my hand at mapping, let me know if I goofed up somehow, thanks.

Capture

## Why's this needed? 

Attempting to throw out some unruly security visitors usually will only result in more flooding in through the open door. Also, this is pretty funny.

## Changelog 

```changelog
(u)Jak27
(+)The security room on the shuttle now feature safer ways to boot out staff assistants.
```

### DIFF
--- a/assets/maps/shuttles/east/eastbase.dmm
+++ b/assets/maps/shuttles/east/eastbase.dmm
@@ -392,6 +392,15 @@
 	icon_state = "floor2"
 	},
 /area/shuttle/escape/centcom)
+"pP" = (
+/obj/disposalpipe/trunk/regular/east,
+/obj/machinery/disposal/small/west{
+	name = "ejection chute"
+	},
+/turf/unsimulated/floor/shuttle{
+	icon_state = "floor4"
+	},
+/area/shuttle/escape/centcom)
 "qb" = (
 /obj/decal/fakeobjects/shuttleengine{
 	dir = 8;
@@ -543,6 +552,15 @@
 /obj/decal/tile_edge/stripe/extra_big,
 /turf/unsimulated/floor/shuttlebay,
 /area/centcom/outside)
+"uJ" = (
+/obj/disposalpipe/trunk/regular/west,
+/obj/disposaloutlet/east{
+	pixel_x = 14
+	},
+/turf/simulated/wall/auto/shuttle{
+	icon_state = "3"
+	},
+/area/shuttle/escape/centcom)
 "uY" = (
 /obj/decal/tile_edge/stripe/extra_big{
 	icon_state = "delivery2"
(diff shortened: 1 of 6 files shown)


TheTyrant25:
does that actually work..?

TobleroneSwordfish:
Seems like an extremely easy way to for crimers to escape? Especially since you can get in chutes while handcuffed iirc. Also should probably be a red security chute rather than a grey disposals one.

Jak27:
> Seems like an extremely easy way to for crimers to escape? Especially since you can get in chutes while handcuffed iirc. Also should probably be a red security chute rather than a grey disposals one.

Most crimers on the shuttle are usually in the port a brig or being babysat by a security officer anyways, but also yeah I should probably change the chute into the more sleek red version.

virvatuli-gh:
I love the idea just for the funny situations it'll cause.

Jak27:
A few weird issues where you might get ejected in the wall instead of the shuttle based on where the shuttle currently is, I'll probably move around the outlets tomorrow to see if that fixes it.

flappybatpal:
Do pipes even work when the shuttle transitions to another Z? Flushing someone at the same time might be funky.

Exester509:
> Do pipes even work when the shuttle transitions to another Z? Flushing someone at the same time might be funky.

Travelling shuttles are on a different z-level, and pipes work fine in some AZones like the owlery, so it shouldn't cause any problems

Jak27:
> Do pipes even work when the shuttle transitions to another Z? Flushing someone at the same time might be funky.

Yeah I just tested it to be sure, the pipe landing/leaving while someone is still travelling in a pipe will not mess anything up.
Though the outlet chutes have a weird tendency to only shoot people only up regardless of what direction it's pointed in when the shuttle is docked at the station or in transit, but it should be fine.

Sord213:
This is very funny, my one concern is this is a VERY good hiding spot on the shuttle. The indication that someone(or something) is in the disposal chute is just a tiiiiny group of pixels, compared to the sleepers which have a major sprite change.

Jak27:
> This is very funny, my one concern is this is a VERY good hiding spot on the shuttle. The indication that someone(or something) is in the disposal chute is just a tiiiiny group of pixels, compared to the sleepers which have a major sprite change.

I think this is just an issue with disposal chutes in general and how little indication they have to show that someone is inside it besides two pixels. Might be a separate PR that could more evidently show that a whole ass person is inside of a trash chute lol

PR #13892, opened 2023-04-29
Adds a new chess themed 3x5 random room

DanaTarSUS:
[MAPPING]
## About the PR 

Adds a room containing some silly little bots playing chess in a candlelit room

## Why's this needed? 

It just seemed like the right thing to do


DisturbHerb:
Could you post an image of the prefab? MapDiffBot renders certainly exist but they're a tad blurry.

DanaTarSUS:
> Could you post an image of the prefab? MapDiffBot renders certainly exist but they're a tad blurry.

image

Here's an image preview as shown in the map editor

PR #13894, opened 2023-04-29
Phone Resprite

Fungal-Infection:
## About the PR 
Changes The Phone Textures as they are quite old looking

Phones

## Why's this needed? 
The current cellphone textures are like solid black with a white keypad, and don't look that great.


ZeWaka:
I don't think these sprites are a good fit. They're not alternative retro scifi. The phones are literally landlines, not modern touchscreens.

PR #8016, opened 2022-03-29
Donut 3 sec layout modifications

Chatauscours:
## About the PR 
This PR rebuilds a part of the sec on donut 3, especially the foyer and the equipment room.

## Why's this needed? 
The sec on donut 3 was always a bit bland. The foyer/checkpoint is not very useful and more of a liability than an actual front desk. That same foyer is not very prectical to use, as the Secmate computer is on the opposite side of the room, blocked by Hooty, there are no camer, lack of tools for a front desk, making that generally never used in comparison to other maps, notably due to it being separated from main sec. And a front desk not used is a shame, especially on RP.

## Changelog

```changelog
(u)Chatauscours
(*)Modifies the foyer, main sec, south brig, equip and dorms in the security area of donut 3.
```

### Old:
old32

### Original proposition:
New

### First revision:
new2

PR #13898, opened 2023-04-29
Relocation Of The Donut3 Security Foyer

Mister-Moriarty:
[Mapping] [QoL]

## About the PR:
Adoption and rework of #8016.

Moves the Donut3 Security Foyer to be adjacent to the entrance, a position in which it should be more accessible to Security staff, and contracts Brig's south side in order to do so. Additionally adds camera monitors, a wanted poster computer, a phone, recharger, and some office stationary.
The four communicating doors between the entrance and hallway have been linked to two switches situated inside of the office; these doors still remain locked to those without Security access.
Visitation has been moved to where the office originally was, and the south side expanded to connect to it.

image

## Why's this needed?
In it's current iteration, Donut3's Security Foyer is of little practical use to Security, not being connected to the rest of the department, and lacking camera monitors and other front desk related equipment.

## Changelog:
```changelog
(u)Mr. Moriarty
(*)Donut3's Security Foyer has been relocated to be adjacent to the entrance.
```


Mister-Moriarty:
Ah, this is what I get for delaying this PR for too long.

nefarious6th:
looks good

- table on the south side of genpop against the wall has two floating cameras over it (on either tile)

Sord213 (changes requested):
Great idea, I've always disliked how the foyer was basically unused because it's not actually attached to sec. Some things I'd like to see, tho:
Screenshot_289
Hallway facing windoors so people don't have to enter partially into sec where the yellow highlight is.
Push the walls up maybe where I put the red arrows? Idk, a single tile window ala oshan or cog1 isn't bad but a double tile opening would be nice.

Also the hallway facing windoors in the public visitation area seem odd. Would rather have normal windows.

Main thing is the hallway facing windoor in the foyer.

Sord213 (approved):
lgtm (when merge conflict is fixed)

PR #13902, opened 2023-04-29
fixes NTSC radio icon hover

### LINKED ISSUE #13862: Wrong Radio Icon on NTSC
### Labels

[BUG][TRIVIAL]

### Description

Hovering over the NTSC's radio icon tells you they're an NTSO.

### Steps to reproduce

1. Have an NTSC speak
2. Hover over their icon
3. NTSC = NTSO

### Expected Behavior

For them to be an NTSC

Reported by: Goat_Real
Active test merges: #13023 and #13667
Round log date: 2023-04-27-02-40
Reported on: 2023-04-27 03:53:23
Map: OSHAN

nefarious6th:
[Bug][game objects]

## About the PR 

fixes ntsc radio icon hover description

## Why's this needed? 

Fixes #13862

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -2523,7 +2523,7 @@ ABSTRACT_TYPE(/datum/job/special/halloween/critter)
 		..()
 		M?.traitHolder.addTrait("training_medical")
 
-// Use this one for late respawns to dael with existing antags. they are weaker cause they dont get a laser rifle or frags
+// Use this one for late respawns to deal with existing antags. they are weaker cause they dont get a laser rifle or frags
 /datum/job/special/nt_security
 	linkcolor = "#3348ff"
 	name = "Nanotrasen Security Consultant"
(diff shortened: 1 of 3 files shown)


nefarious6th:
okay I realized the problem here maybe, will edit the other file just to keep parity, but otherwise will wait for instruction

- the headset given to NTSCs is the same as the one given to NTSOs, so the hover tooltip when edited will be incorrect one way or the other
- would have to split the headsets into two items to resolve this, which would require duplicating the block of code in radios.dm, maybe not great for two objects that will do the same thing?

Mister-Moriarty:
> ...would have to split the headsets into two items to resolve this, which would require duplicating the block of code in radios.dm, maybe not great for two objects that will do the same thing?

Why not use a subtype of `/obj/item/device/radio/headset/command/nt` for NTSCs?

nefarious6th:
> Why not use a subtype of `/obj/item/device/radio/headset/command/nt` for NTSCs?

overthinking struck again; ty for keeping it real, Moriarty 😊

Tarmunora (approved):

PR #13918, opened 2023-05-01
adds catwalks under grilles at cog1 telecom dish

### LINKED ISSUE #13917: Gaps in the cog1 comms dish antenna
### Labels

[BUG][trivial][mapping]

### Description

There's slight but noticable gaps between the lattice and grilles because the grilles don't extend to the edges of the turfs

### Steps to reproduce

N/A

### Expected Behavior

Something that looks visually connected

These gaps wouldn't have existed before perspective grilles

image

nefarious6th:
[mapping][bugfix]

## About the PR 
throws some catwalks down to connect the lattices on cog1's telecomms dish that go underneath the grilles 

## Why's this needed? 
Fixes #13917

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -50694,6 +50694,14 @@
 	},
 /turf/simulated/floor/grime,
 /area/station/storage/hydroponics)
+"fHV" = (
+/obj/grille/steel,
+/obj/lattice{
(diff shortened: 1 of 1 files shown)


flappybatpal:
Guess which dummy fixed this without checking for a PR first.

PR #13927, opened 2023-05-01
exploration tools random room

CoolKakapo:
## About the PR 
Adds a random room which contains all the tools needed for exploration/adventuring. Contains a full toolset, a space suit/helmet, insulated gloves and magnetic boots, brute/fire/toxin medkits, a toolbelt, a flashlight and a fire extinguisher. The spawn weight is 50.
Schermafbeelding 2023-05-01 180659

## Why's this needed? 
Adds more variation to random rooms, which encourages more players to look for random rooms.


flappybatpal:
Disagree conceptually with "loot rooms", rooms that provide equipment.
Minor items appropriate to a themed room can be OK but there should be a really good reason for anything of value.

A random room should be a point of interest, tell a story, be an interesting or funny scene etc.

TobleroneSwordfish:
Yep, what Katzen said.

PR #13938, opened 2023-05-02
adjust filgrastim recipe

TDHooligan:
[C-QOL][A-Chemistry][A-Medical]

## About the PR 

https://forum.ss13.co/showthread.php?tid=21112 
Suggests changing this, and it makes sense as pathology is kill.  Probably better than having to add more accessible e-coli.

I figure beff's a pretty good substitute for 'some effort required', There's probably some third-party/pharmaceutical strains of bacteria in whatever cheap meats 'beff' is implied to be made from. Beff's recipe is also made easier with hemolymph, another type of blood. Win-win? 

Ultimately not fussed on if this specific patch is merged or another route is picked, but figured the forum thread had merit, so it's worth raising.

## Why's this needed? 

pathology is dead.

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -3430,7 +3430,7 @@ datum
 			name = "Filgrastim"
 			id = "filgrastim"
 			result = "filgrastim"
-			required_reagents = list("blood" = 1, "dna_mutagen" = 1, "e.coli" = 1, "spaceacillin" = 1)
+			required_reagents = list("blood" = 1, "dna_mutagen" = 1, "beff" = 1, "spaceacillin" = 1)
 			result_amount = 2
 
 		ecoli // needed for filgrastim vOv


Tarmunora (approved):

nefarious6th:
consider this the poor man's wiki tag

minor changelog needed for both recipes?

colossusqw:
I'm not sure if this is quite the best chem to replace E. Coli, considering how heparin never saw the light of day having meat slurry as a precursor, and beff seems even harder to acquire. Sure, heparin doesn't quite have as many straight up uses as filgrastim, but even then, i'm not sure you're gonna see filgrastim around 99% of the time.

TobleroneSwordfish:
Beff is a pretty rough pre-requisite for a niche medical chem, ideally it'd be nice to make e. coli accessible in some other way, but I don't know enough (any) microbiology to figure out what would be a sane way of getting it

TDHooligan:
added to wiki.

PR #13946, opened 2023-05-03
Adds Promethazine, a chemical that stops vomiting

colossusqw:
[LABEL][chemistry][feature]

## About the PR 
This PR adds Promethazine, a anti-emetic chemical that can be used to prevent many instances of vomiting. It has some slight OD effects such as a small chance of slowing you down slightly on severe ODs, but besides that it's mainly just a chemical to stop you from puking. Made by combining, oil, ammonia, sulfur and space cleaner.(Recipe idea by TDHooligan)

## Why's this needed? 
Too many things make you puke out your guts and make the station filthy, so i tried to get something to help out the poor janitors :).

```changelog
(u)Colossusqw
(*)Added Promethazine, a anti-emetic agent that stops vomiting.
(+)Slightly refactored many instances of vomiting messages, please do point out any weirdness that occurs.
```


CaioFrancisco:
My only two balancing concerns is how there's barely any punishment for smoking this chem into the halls (including overdosing someone with it.) There's also the small fact that it completely nullifies drinking too much Bo Jack, which is weird.

colossusqw:
> My only two balancing concerns is how there's barely any punishment for smoking this chem into the halls (including overdosing someone with it,) and also the small fact that it completely nullifies drinking too much Bo Jack, which is weird.

I'm not sure it being easy to give out to people is exactly a balancing concern, considering all this does is reduce the amount of puke around. I could change the bojack part easily, maybe just changing it to only stop you puking? Hmm, would feel weird for it to remove some bojack without you puking though...

M-Earthfire:
> My only two balancing concerns is how there's barely any punishment for smoking this chem into the halls (including overdosing someone with it,) and also the small fact that it completely nullifies drinking too much Bo Jack, which is weird.

Tbh, it's good to not have a punishment for smoking it. Smoke can and should have good effects if applied with certain chems. 

Personally, i think the recipe is too complex for what a mediocre effect the chem has. Personally, i would throw out sulfur and ammonia and add something like sugar. Or is the recipe there to emulate a certain reaction?

code/modules/chemistry/Reagents-Medical.dm
```diff
@@ -1083,6 +1083,37 @@ datum
...
+		medical/promethazine // This stops you from vomiting
+			name = "promethazine"
+			id = "promethazine"
+			description = "Promethazine is a anti-emetic agent."
+			reagent_state = LIQUID
+			fluid_r = 180
+			fluid_g = 255
+			fluid_b = 140
+			depletion_rate = 0.1
+			penetrates_skin = TRUE
```
Tarmunora, line 1095:
```suggestion
			depletion_rate = 0.4
```

Tarmunora (changes requested):

TDHooligan:
going to add a suggestion specifically to keep skinpen, following its' initial discussion:
1 - balance wise, the chem benefits only janitors, who are also the only ones who care about vomiting. janitors lack penetrative means to transfer reagents, but could totally splash a bucketful of this stuff onto irritating nerds with drug problems.
2 - its' use as a poison is limited to foods (where it could already be added to the food itself)
3 - it's technically a topical treatment

the standard depletion rate should see to it that smoking it isn't too effective, without skinpen we'll probably see very little usage of it.

pali6:
@Tarmunora

PR #13951, opened 2023-05-03
Reworks Energy Gun Charge Indicators To Use Overlays

Flaborized:
[CLEAN]

## About the PR 

This PR changes many (not all, some are unchanged because they were already using overlays and have special code for it, could also be added but considering my goals with this PR I didn't) energy guns that use charge indicators to use overlays for them instead of bespoke sprites. From a user perspective should have no apparent difference unless there's bugs. 

Also changes a few egun names to be a bit more descriptive while I was editing their sprites anyways

Picture of me making sure this works:

How the dmi looks:

## Why's this needed? 

All energy guns using indicators were using a copy and pasted block of code to do so, this condenses that into one proc which should make it easier to change if we want to change something about how this behaves.

More importantly (to me) this allows for easier sprite-related feature additions to eguns. Originally I wanted to add in a egun variant sprites (one for the captain and one for the HoS, so they looked different) but I realized that would be very annoying to implement since I'd have to add 100 million icon states to match every single possible combination of charge, setting and gun type. Now something like that would entail adding exactly two sprites. Other features (firing animations? glow in the dark lights? rainbow animated taser?) become much easier to implement too, so I think this is a much better way to have this all laid out.


Mister-Moriarty:
Something that you may want to consider is that this implemation allows for charge indicators to use lighting overlays, so that they may glow in the dark, similar to computer screens or cyborg eyes.

Flaborized:
> Something that you may want to consider is that this implementation allows for charge indicators to easily use lighting overlays, so that they may glow in the dark, similar to computer screens or cyborg eyes.

yeah! there's quite a lot of visual improvements I've wanted to do with energy guns that just haven't been possible with the way they're currently implemented, my hope is to add more onto them with this (and hopefully down the line others will find it easier to add things as well)

Gannets:
This seems cool!

TobleroneSwordfish (approved):
lgtm

code/obj/item/gun/energy.dm
```diff
@@ -15,6 +15,9 @@ TYPEINFO(/obj/item/gun/energy)
 	var/custom_cell_max_capacity = null // Is there a limit as to what power cell (in PU) we can use?
 	var/wait_cycle = 0 // Using a self-charging cell should auto-update the gun's sprite.
 	var/can_swap_cell = 1
+	var/uses_charge_overlay = FALSE // Does this gun use charge overlays on the sprite?
```
ZeWaka, line 18:
```suggestion
	var/uses_charge_overlay = FALSE //! Does this gun use charge overlays on the sprite?
```

ZeWaka (approved):

Sovexe:
This PR is causing infinite recursion runtimes

PR #13952, opened 2023-05-03
prevents cyborgs from deconstructng the robotics control console

nefarious6th:
[game objects][balance][controversial]

## About the PR 
this prevents silicons from deconstructing the robotics control console by using the `NO_BORG` deconstruction flag. however, to achieve no-borg interaction, it is also now not possible to deconstruct the console with tools (`perma` value set to `1`). what this means is if someone wanted to remove the robotics control console, they must have a.) a deconstruction tool, and b.) sufficient access to use the console.

i'm not super happy with my implementation, but if anyone has any ideas of how to build a check for the interaction with the screwing tool to discount specifically any use of a silicon omnitool, i'd be happy to hear. 

## Why's this needed? 
idea from: https://forum.ss13.co/showthread.php?tid=21138
can help with unkillable silicons powergaming their rogue-ness.

## Changelog 
```changelog
(u)mona
(+)To remove the robotics control console, you now need a deconstruction tool and sufficient access permissions
(+)Silicons can no longer remove the robotics control console with their tools
```

### DIFF
--- a/code/obj/machinery/computer/robotics.dm
+++ b/code/obj/machinery/computer/robotics.dm
@@ -4,11 +4,12 @@
 	icon_state = "robotics"
 	req_access = list(access_robotics)
 	object_flags = CAN_REPROGRAM_ACCESS | NO_GHOSTCRITTER
+	deconstruct_flags = DECON_NOBORG
 	desc = "A computer that allows an authorized user to have an overview of the cyborgs on the station."
 	power_usage = 500
 	circuit_type = /obj/item/circuitboard/robotics
 	id = 0
-	var/perma = 0
+	var/perma = 1
 
 	light_r =0.85
 	light_g = 0.86


TheColorCyan:
Yeah same as I said on forum post, good change to prevent sillicons from becoming completely immortal with too much ease

Sord213:
Meh, killswitch sucks from a gameplay perspective. Borgs are already incredibly weak unless they take the single upgrade that makes them able to recover from being stunned.

Retrino:
personally think a better route would be putting the circuitboard onto maps that don’t have a spare/make the location of it more consistent (the bridge or away from the killswitch itself). cyborgs can still destroy the console with relative ease, as well as the roguer being able to do it themselves in a few clicks. 

the spare board on most maps already allows for another killswitch to be built out of sight of the silicons, making it very hard for them to counterplay. it also doesnt really address the event of the silicon running back to their roguer to be reactivated immediately.

M-Earthfire:
Personally, it bugs me to no end to have inconsistencies when it comes to deconstructing consoles. So i would keep it being deconstructable with tools.

I agree with retrino that a better approach would be to have spare robotics console curcuitboards be mire avaible. I think adding them to the MD (because they are head of robotics department) or Captains locker would be an idea, similar to having spare law rack manudrives or ai shells

nefarious6th:
closed for stale and i don't like this implementation

PR #13957, opened 2023-05-03
more skeleton mutantrace bugfixes

### LINKED ISSUE #13954: Skeleton heads placed onto other bodies can't speak
### Labels

[BUG][MINOR]

### Description

With the recent skeleton overhead text fix, it seems placing a skele head onto another body makes the game think it's in a bag of sorts, making you inaudible to anyone.

### Steps to reproduce

1. Transplant head onto monkey
2. Other people cannot hear you

### Expected Behavior

Overhead text over the money as if it was my body.

Reported by: CalliopeSoups
Active test merges: nothing
Round log date: 2023-05-03-18-39
Reported on: 2023-05-03 18:58:46
Map: COGMAP

Skeletonman0:
[OBJECTS] [BUG] [QOL]
## About the PR 

this PR makes it so:

- speech code now checks if the head is the right type for skeletons to use it
- speech code also checks if a skeleton skull is attached to a mob, allowing its owner to have speak from that mob's location

- organ code will forcibly unlink a skeleton from their head if their head is attached to a mob and has a brain inserted into it

- skeletons will nolonger get kicked back to their body if a non skeleton head is attached
- skeleton blindness check now uses the linked head's eyes, only uses the body's eyes if no head is linked
- skeleton maptext should nolonger overlap itself when without a linked head

ive probably made some dumb mistakes somewhere, but it appears to work

## Why's this needed? 
fixes #13954

also fixes a bug which hasn't been reported where skeleton mutantrace code would set your view back to your body if a head it refused to use (non skeleton heads) got attached

## Changelog 

```changelog
(u)Skeletonman0
(+)Skeleton heads will now also speak when attached to corpses while their skull has no brain
```


code/obj/item/organ_holder.dm
```diff
@@ -879,6 +886,18 @@
 				newBrain.op_stage = op_stage
 				src.brain = newBrain
 				src.head.brain = newBrain
+
+				// if the head has an skeleton, and we're not taking it, eject the skeleton out of the head
+				if (src.head.head_type == HEAD_SKELETON && !isskeleton(src.donor))
```
TobleroneSwordfish, line 891: shouldn't this be checking if it's the correct skeleton, not just *a* skeleton?

code/obj/item/organ_holder.dm
```diff
@@ -828,10 +828,17 @@
...
+
+					if (isskeleton(H))
 						var/datum/mutantrace/skeleton/S = H.mutantrace
-						S.set_head(newHead)
-					H.set_eye(null)
+						if (newHead.head_type == HEAD_SKELETON) // dont set eye position if we're not linking to it
+							S.set_head(newHead)
+							H.set_eye(null)
+						else if (!S.head_tracker) // set head position if there's no head
+							H.set_eye(null)
```
TobleroneSwordfish, line 838: these comments seem to contradict the code?

PR #13958, opened 2023-05-03
Transception array attempt_transceive fix

Kubius:
[BUG]
## About the PR 

Changes a check for Nadir's transception array to check overall list length. instead of the presence of a specific index that may or may not exist at check time.

## Why's this needed? 

Should prevent rare runtimes in the case of multiple transception pads attempting to retrieve the same pending cargo item at the same time.


Kubius:
Failing check appears to be unrelated to PR contents.

PR #13959, opened 2023-05-03
Refactors Antagonist Overlays To Use Client Image Groups

Mister-Moriarty:
[Gamemodes] [Internal] [Code Quality]

## About the PR:
Antagonist overlays have been refactored so to use client image groups in conjunction with a new `/datum/mind_mob_overlay` system, which permits a mob overlay image to move with a player's mind as it moves between mobs, updating the `loc` of the image as the mind does so.
`antagonist_overlay_refresh()` has been deprecated and removed from the code.
Additionally refactors `traitorsseeeachother` into `antagonists_see_each_other`, which now permits all antagonists to see all other antagonists, in lieu of solely traitors seeing each other.

## Why's this needed?
Antagonist overlays should now be significantly more responsive to new antagonists being added or clients being permitted to see antagonist overlays, alongside the greater versatility granted by client image groups.


flappybatpal:
Skimmed through the code and given some basic testing locally where it performed well.
Should be very useful for revs where we have had some problems.

Will need testmerge to get a good idea for Wraiths, Revs etc in practice.

code/modules/antagonists/antagonist_helpers.dm
```diff
@@ -64,3 +64,17 @@ proc/get_all_gangs()
...
+			"Head of Security",
+			"Head of Personnel",
+			"Chief Engineer",
+			"Research Director",
+			"Medical Director",
+			"Communications Officer",
+			))
+		return TRUE
+
+	return FALSE
```
TobleroneSwordfish, line 80:
```suggestion
/datum/mind/proc/is_head_of_staff()
	return src.assigned_role in list(
			"Captain",
			"Head of Security",
			"Head of Personnel",
			"Chief Engineer",
			"Research Director",
			"Medical Director",
			"Communications Officer",
			)
```

TobleroneSwordfish (changes requested):
Code looks vaguely sane, one runtime found in testing:
image
Seems like something to do with admin removing an antag, although I'm not quite sure of the repro.
Overall I'm very happy to see `antagonist_overlay_refresh` being killed :)

Mister-Moriarty:
> ...one runtime found in testing:

The runtime implies that a mind mob overlay was somehow added to `/datum/mind/var/antagonists`, which should not be possible? I am unable to reproduce the runtime.

My `/datum/mind/var/antagonists` conclusion derives from:
admin.dm, line 2141, `antag.display_name`,
admin.dm, line 2137, `antag = locate(href_list["target_antagonist"])`,
player_options.dm, line 124, `...target_antagonist=\ref[antagonist_role]'>Remove</a><br>"`,
player_options.dm, line 115, `for (var/datum/antagonist/antagonist_role as anything in M.mind.antagonists)`.

flappybatpal:
Admins appear to always have antag tags shown. Previously this only happened when Observer or Adventure View had been toggled.

PR #13960, opened 2023-05-03
fix skeleton skull hearing logic

Skeletonman0:
[OBJECTS] [BUG]
## About the PR 
quick fix that should allow skeleton skulls in someone's bag to hear the person holding them, instead of just skeleton skulls in their pockets

untested

## Why's this needed? 
oversights / bugs bad


Skeletonman0:
fixing so it actually works

PR #13962, opened 2023-05-03
Implements Faction Framework

### LINKED ISSUE #13658: Multitude of wraith portal bugs with mobs
### Labels

[BUG]

### Description

Summons don't get deleted when they die. summons don't get deleted when the portal gets broken. dead summons don't decrement the points(?). portal stops summoning after a while even while making portals. portal summons attack other wraith summons (needs seek_target override)

### Steps to reproduce

1.  summon a wraith harbringer critter portal
2. spawn mob critters
3. bugs AAAAAAAA

### Expected Behavior

Better worky

Due to mobification and it orginally only being built for obj critters, best to just rework it to use mobs as a whole. really I made this for myself for later...

Wisemonster:
[INTERNAL] [QOL] [FEATURE] [BUG] [AI]

## About the PR 
This pr adds a framework that let's mobs be apart of factions.

## Why's this needed? 
Currently, there is no way to prevent AI critters from attacking certain mobs without disabling the AI all together. This implements the framework for a system that let's you assign mobs/critters different factions, and prevents different critters that are a part of the same faction as other mob/critters from attacking each other. This was also talked about in #The-Mobbening on discord

### DIFF
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -137,6 +137,8 @@
 
 	var/job = null
 
+	var/faction = 0 // For assigning mobs various factions
+
 	var/nodamage = 0
 
 	var/spellshield = 0
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -137,7 +137,7 @@
 
 	var/job = null
 
-	/// For assigning mobs various factions
+	/// For assigning mobs various factions, see factions.dm for definitions
 	var/faction = 0
 
 	var/nodamage = 0
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -137,7 +137,8 @@
 
 	var/job = null
 
-	var/faction = 0 // For assigning mobs various factions
+	/// For assigning mobs various factions
+	var/faction = 0
 
 	var/nodamage = 0
 
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -181,8 +181,6 @@
 /mob/living/death(gibbed)
 	#define VALID_MOB(M) (!isVRghost(M) && !isghostcritter(M) && !inafterlife(M))
 	src.remove_ailments()
-	if (src.faction)
-		src.faction = null
 	if (src.ai) src.ai.disable()
 	if (src.key) statlog_death(src, gibbed)
 	if (src.client && ticker.round_elapsed_ticks >= 12000 && VALID_MOB(src))
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -43,6 +43,7 @@
 	var/in_point_mode = 0
 	var/butt_op_stage = 0.0 // sigh
 	var/dna_to_absorb = 10
+	var/faction = null // For assigning mobs various factions
 
 	var/canspeak = 1
 
@@ -181,6 +182,8 @@
 /mob/living/death(gibbed)
 	#define VALID_MOB(M) (!isVRghost(M) && !isghostcritter(M) && !inafterlife(M))
 	src.remove_ailments()
+	if (src.faction)
+		src.faction = null
 	if (src.ai) src.ai.disable()
 	if (src.key) statlog_death(src, gibbed)
 	if (src.client && ticker.round_elapsed_ticks >= 12000 && VALID_MOB(src))
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -43,7 +43,6 @@
 	var/in_point_mode = 0
 	var/butt_op_stage = 0.0 // sigh
 	var/dna_to_absorb = 10
-	var/faction = null // For assigning mobs various factions
 
 	var/canspeak = 1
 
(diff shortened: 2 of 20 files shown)


code/mob/living.dm
```diff
@@ -181,6 +182,8 @@
 /mob/living/death(gibbed)
 	#define VALID_MOB(M) (!isVRghost(M) && !isghostcritter(M) && !inafterlife(M))
 	src.remove_ailments()
+	if (src.faction)
+		src.faction = null
```
Tarmunora, line 186: will cause SRed/otherwise-revived mobs to be factionless. Intentional?

Wisemonster: No, was intended to prevent respawned mobs from keeping factions across mobs, but I believe the mobs faction is set on creation, so I'll remove this.

CheffieGithub:
For mobs which can be spawned outside of the wraith (shades, brullbars, etc) the faction flags should get set when they are spawned by the wraith summon portal, not on that mob

code/mob/living/critter.dm
```diff
@@ -1305,6 +1305,7 @@ ADMIN_INTERACT_PROCS(/mob/living/critter, proc/modify_health)
 			if (isintangible(C)) continue
 			if (isdead(C)) continue
 			if (istype(C, src.type)) continue
+			if (src.faction != null && (C.faction == src.faction)) continue //Checks if they are in the same faction
```
amylizzle, line 1308: If you're doing bitflags for faction alignment, you should do `C.faction & src.faction != 0` instead. That way if they share any factions they are friends, and if not, enemies.

code/mob/living.dm
```diff
@@ -43,6 +43,7 @@
 	var/in_point_mode = 0
 	var/butt_op_stage = 0.0 // sigh
 	var/dna_to_absorb = 10
+	var/faction = null // For assigning mobs various factions
```
amylizzle, line 46: put this on `/mob` instead of `/mob/living` 

critters can target any `/mob`

amylizzle (changes requested):

Wisemonster:
After some discussion on discord, I've opted to avoid messing with the disaster/wraith portal summons till after this is merged, due to it being more complicated than I originally thought.

code/mob.dm
```diff
@@ -137,6 +137,8 @@
 
 	var/job = null
 
+	var/faction = 0 // For assigning mobs various factions
```
TobleroneSwordfish, line 140:
```suggestion
	/// For assigning mobs various factions
	var/faction = 0
```
Doc comments

code/mob/living/critter/adventure.dm
```diff
@@ -77,6 +77,7 @@
 			if (isintangible(C)) continue
 			if (isdead(C)) continue
 			if (istype(C, src.type)) continue
+			if (C.faction & src.faction != 0 && (C.faction == src.faction)) continue //Checks if they are in the same faction
```
TobleroneSwordfish, line 80: Not sure what you think this is doing, but one half of it is definitely redundant. Also somewhat out of scope, but this practice of copy pasting a bunch of checks into `seek_target` for every new mobcritter is bad and should probably be a `valid_target` proc instead.

TobleroneSwordfish (changes requested):

code/mob/living/critter/adventure.dm
Wisemonster: Did this as suggested by amylizzle in a previous comment. I believe it prevents factionless mobs from being friendly to other factionless mobs. I could also remove the copy-pasted code from various critters and save that for a future pr, but that will mean that you cannot implement a faction for these mobs until that is pr-ed.

Wisemonster: And I mean save changing the copy-pasted code into a proc for a future pr, not just copy-pasting.

TobleroneSwordfish: Right, but why are you doing the bitwise check and then checking exact equality as well? I think Amy's suggestion was to do `if (src.faction != null && (C.faction & src.faction != 0))` since that would cause null factions to always be aggressive, and any mobs that share factions to be passive.

Wisemonster: If I only had it check ```if (src.faction != null && (C.faction & src.faction != 0))```, then that would mean that a SYNDICATE faction mob and a NANOTRASEN faction mob would ignore each other, cause they both meet this if check, even if they would not be on the same team, unless I'm misunderstanding how bitmask checks work.

amylizzle: You are misunderstanding how bit checks work. `1 & 2 = 0`, but `3 & 2 = 2` so syndie and NT would be enemies, but a mob could be both syndie and NT and be friendly to both.

Wisemonster: Ok, I believe I've changed them to what ya'll suggested

code/mob.dm
```diff
@@ -137,6 +137,9 @@
 
 	var/job = null
 
+	/// For assigning mobs various factions
```
TobleroneSwordfish, line 140:
```suggestion
	/// For assigning mobs various factions, see factions.dm for definitions
```

PR #13964, opened 2023-05-04
Add artifact bag of holding

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
This PR adds a new handheld artifact, the bag of holding. 

The bag of holding is an artifact that acts as a storage, just like a box, but that has more variety in how it stores things or acts.

It comes in three varieties, depending on origin.

Eldritch-
Eldritch bags of holding are a general variety storage, which can hold anywhere from a few to a good amount of items, from tiny to bulky. Higher size capacities are more rare for higher slots. However, this type has no visible HUD, and items stored in them are stored in a FIFO, LIFO, or random order. FIFO/LIFO/random order idea from Flourish.

Martian-
Randomized, "shape-shifting" storage. Every so often, its storage slots and max holdable size will randomize. They can always be worn on your belt or back.

Wizard-
This is a large storage artifact that can hold from a few to a lot of items from tiny to small sized. However, you can only see a smaller number of the items in the storage at a time, and every time you add an item, take an item out, or look in the storage, you see a random set of items. Idea from zjdtmkhzt.

Larger storages can be worn on belts, for Martian, or your back, for all types. Artifacts transform their icon state when put in a belt or back slot.

Faults can trigger when storing items.

If you manage to put a bag of holding into a bag of holding, something bad happens too which can include
-Explosions
-Black holes
-Getting trapped in a pocket dimension
-Contents teleporting randomly

A side effect of the current rustle effect with how handheld artifacts are coded/sprited right now is that the sprite will flip sideways or upside down for rustles sometimes

Sprites made by DimWhat (Wizard sprites by my self)-
image
image

## Why's this needed? 
It'd be cool to see a handheld/wearable storage artifact, since that would allow for more fun and varying storage items

## Changelog 
```changelog
(u)FlameArrow57 and DimWhat
(*)Added the bag of holding artifact! Appears with three variants. Ones with larger storages can be worn on your back, and the Martian one on your belt too.
```


flrsh:
I definitely like the idea of storage artifacts, though storage in general can be rather annoying to balance and I'm unsure about just making these straight up Better Storage. Personally I'd like these more if there were weirder aspects you had to work around, e.g. FIFO / LIFO, all of one type share the same storage, etc.

FlameArrow57:
Okay yeah makes sense. FIFO/LIFO/random pick is possible through another PR I'm working, which I think would be fun to add, I'll try to work on some more things like that

zjdtmkhzt:
This seems really cool! 

Yeah, storage is a hard thing to balance, in addition to things like LIFO/FIFO/random there could also be something like a type with infinite storage, but which only ever shows a random selection from all stored items as available.

Maybe watch out with the naming on things like "fault" for the burning stuff, because artifact faults are already a specific artifact mechanics, so it could potentially be confusing to talk about in coder talk.

If some of these can be worn, maybe there should be a worn sprite for them? I was under the impressions most larger wearable containers, like backpacks or belts had them.

amylizzle:
Very cool, though personally I think all artifacts ought to be activated before they can be used. I've been meaning to fix artifact beakers not doing that for a while now.

Exester509:
For the wizard one, it would be really cool if on fail it swapped the item for a note saying "This item has been confiscated by the Space Wizard Federation" or something similar. I'd find it a hilarious fail

Glamurio:
Some of these should have a chance to spawn with a random array of items in them, akin to how beakers have random chems. Would add an additional layer of lootbox fun to finding one, rather than just for storage-sake.

FlameArrow57:
> For the wizard one, it would be really cool if on fail it swapped the item for a note saying "This item has been confiscated by the Space Wizard Federation" or something similar. I'd find it a hilarious fail

That'd be fun, though idk if how much Wizards are tied to the artifacts in lore

FlameArrow57:
> Some of these should have a chance to spawn with a random array of items in them, akin to how beakers have random chems. Would add an additional layer of lootbox fun to finding one, rather than just for storage-sake.

Agree that'd be cool, but think that's already a thing for the container artifacts, was trying to not do the same effect

FlameArrow57:
Okay sprites and the martian bag of holding is still WIP, but current code works and is ready for review

_std/macros/atom_properties.dm
```diff
@@ -298,7 +298,7 @@ To remove:
 #define PROP_ATOM_TELEPORT_JAMMER(x) x("teleport_jammer", APPLY_ATOM_PROPERTY_SUM, REMOVE_ATOM_PROPERTY_SUM, PROP_UPDATE_TELEBLOCK_CAT)
 #define PROP_ATOM_FLOCK_THING(x) x("flock_thing", APPLY_ATOM_PROPERTY_SIMPLE, REMOVE_ATOM_PROPERTY_SIMPLE)
 #define PROP_ATOM_FLOATING(x) x("floating", APPLY_ATOM_PROPERTY_SIMPLE, REMOVE_ATOM_PROPERTY_SIMPLE)
-
+#define PROP_ATOM_PRECURSOR_BOH_ENTERED(x) x("precursor_boh_entered", APPLY_ATOM_PROPERTY_SIMPLE, REMOVE_ATOM_PROPERTY_SIMPLE)
```
ZeWaka, line 301: need some code docs on this

code/modules/storage/items/bag_of_holding.dm
```diff
@@ -0,0 +1,212 @@
...
+		var/obj/item/artifact/bag_of_holding/boh = I
+		var/datum/artifact/artifact = boh.artifact
+		if (artifact.activated)
+			combine_bags_of_holding(user, boh, src.linked_item)
+			return FALSE
+	var/obj/item/artifact/bag_of_holding/boh = src.linked_item
+	boh.ArtifactFaultUsed(user, boh)
+	return TRUE
+
+/datum/storage/artifact_bag_of_holding/wizard
```
ZeWaka, line 30: could use comments/spacing separating the different types

code/modules/storage/items/bag_of_holding.dm
```diff
@@ -0,0 +1,212 @@
...
+			explosion_new(boh_1, T, 3) // causes a one tile hull breach
+			T.visible_message("<span class='alert'><B>The artifacts explode! HOLY SHIT!!!")
+			playsound(T, "explosion", 25, TRUE)
+			user?.gib()
+
+			effect = "explosion"
+		// implosion
+		if (2)
+			var/obj/dummy/artifact_boh_singulo_dummy/black_hole = new (T)
+			T.visible_message("<span class='alert'><B>The artifacts shink to nothing! UH OH.")
```
zjdtmkhzt, line 126:
```suggestion
			T.visible_message("<span class='alert'><B>The artifacts shrink to nothing! UH OH.")
```

FlameArrow57:
Maybe will finish later

FlameArrow57:
Okay, added martian storage, everything should work to what I tested, ready for re-review

Things from when this was last worked on
- Might just do away with the precursor storage
- DimWhat might not be available for doing the sprites for this, so by the time this is code reviewed if there's no sprites for them, either I'll just set it so that only the Martian variant can appear, or the Martian one can have any of the possible varieties, until each gets sprited if they ever do

FlameArrow57:
Ready for review again

code/z_adventurezones/void.dm
```diff
@@ -39,6 +39,12 @@ CONTENTS:
 		for(var/mob/living/carbon/human/H in src)
 			H.client?.playAmbience(src, AMBIENCE_FX_2, 50)
 
+/area/crunch/artifact_boh_pocket_dimension
```
ZeWaka, line 42: why crunch

code/obj/artifacts/artifact_items/handheld_storage.dm
```diff
@@ -0,0 +1,180 @@
...
+/obj/item/artifact/bag_of_holding
+	name = "artifact bag of holding"
+	associated_datum = /datum/artifact/bag_of_holding
+	w_class = W_CLASS_NORMAL
+	/// original icon state, without artifact activation fx
+	var/base_icon_state
+	/// original item_state
+	var/base_item_state
+	/// worn belt icon states
+	// TODO: sprite eldrich and wizard belts
```
ZeWaka, line 14: ?

code/obj/artifacts/artifact_items/handheld_storage.dm
```diff
@@ -0,0 +1,180 @@
...
+				if (slots >= 3 && slots <= 5)
+					wclass = pick(prob(10); W_CLASS_TINY, prob(50); W_CLASS_SMALL, prob(100); W_CLASS_NORMAL, prob(40); W_CLASS_BULKY)
+				else if (slots > 5 && slots <= 9)
+					wclass = pick(prob(20); W_CLASS_TINY, prob(100); W_CLASS_SMALL, prob(50); W_CLASS_NORMAL)
+				else if (slots >= 9)
+					wclass = pick(prob(30); W_CLASS_TINY, prob(100); W_CLASS_SMALL, prob(10); W_CLASS_NORMAL)
+
+				// allow it to be worn if it can hold a lot of items
+				//if (slots > 4)
+				//	boh.c_flags |= ONBELT
```
ZeWaka, line 132: ?

code/obj/artifacts/artifact_items/handheld_storage.dm
```diff
@@ -0,0 +1,180 @@
...
+
+			// immediately start the storage effect, can always be worn
+			if ("martian")
+				boh.c_flags |= (ONBELT | ONBACK)
+				boh.martian_change_shape()
+
+			// "large sized" storage that can hold small items
+			if ("wizard")
+				boh.w_class = W_CLASS_BULKY
+				//boh.c_flags |= (ONBELT | ONBACK)
```
ZeWaka, line 152: ?

ZeWaka (changes requested):

code/z_adventurezones/void.dm
FlameArrow57: I wanted to reuse the void ambience mostly

code/obj/artifacts/artifact_items/handheld_storage.dm
FlameArrow57: Planned slot requirement for belt wearability, if it gets sprited, can remove if that'd be better

code/obj/artifacts/artifact_items/handheld_storage.dm
FlameArrow57: Place holder for if belt gets added, I'll just remove for now

code/obj/artifacts/artifact_items/handheld_storage.dm
FlameArrow57: Meant it as a todo for the future, if anyone can sprite it (didn't think I could very well), I'll reclarify

ZeWaka (approved):

PR #13967, opened 2023-05-04
Changes the stats button on 3 huds

DimWhat:
[ui] [sprites]
## About the PR 
Swaps out the "stat" button on old, classic, and quilty huds for somewhat more fitting versions.
image
image

## Why's this needed? 
It looks a bit better with a fitting button instead of a completely unfitting one.


Tarmunora (approved):

BatElite:
Not hugely familiar with HUD code, but since some of these  how do they look on the TG layout?

ZeWaka:
> Not hugely familiar with HUD code, but since some of these clearly go to the edge of the sprite how do they look on the TG layout?

.

PR #13970, opened 2023-05-04
Mobifies Spacewasps

CheffieGithub:
## About the PR 
Adds AI to wasps and replaces object wasps with them.

Makes wasp grenades use angry wasps because I would be angry stuck in that thing...

## Why's this needed? 
bad object critter. Wasp grenades are a traitor item and the wasp bow already used angry wasps.


Tarmunora (approved):

TobleroneSwordfish (changes requested):
Since you're replacing all the obj critter wasps with the mob variant it should probably have its health reduced to avoid doubling the health of every wasp from pest events, prefabs etc.

CheffieGithub:
I feel like reducing their health back to 10 will make them a non threat for the most part, being killed before they can even sting anyone, unless there is a big stack of them like bees (since they inherit no density from small animal). I did reduce how many spawned from pest spawns already increasing their cost to 2.

PR #13988, opened 2023-05-05
Fixes A Parallax Runtime

Mister-Moriarty:
[Internal] [Runtime]

## About the PR:
Fixes a runtime resulting from `/client/proc/respawn_self()` calling `Login()` twice on a mob, resulting in the parallax signals attempting to be registered twice.

### DIFF
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -1504,7 +1504,6 @@
 	var/mob/new_player/M = new()
 
 	M.key = usr.client.key
-	M.Login()
 	return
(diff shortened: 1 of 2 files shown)


ZeWaka (commented):
All tested?

ZeWaka:
> > All tested?
> 
> Indeed; assigning a new value to `/mob/var/key` appears to innately call `Login()`.

Yep, thought so. cargo cult code moment

Issue #12875, opened 2023-01-25
Mop with water doesn't seem to clean up bloody footprints anymore


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Mop with water doesn't seem to clean up bloody footprints anymore

### Steps to reproduce

Wet mop with bucket with a bit of cleaning fluid in it
First click will clean bloody footprints, maybe the second one will, but anything after that will not

### Expected Behavior

Mop with water and no cleaning fluid used to clean bloody footprints

### Additional Information & Screenshots

If this is an intentional change it kind of makes the mop way worse than the sponge in every way other than cleaning up big fluid puddles.

Reported by: Stush
On server: Goonstation 3 Roleplay: Morty
Round log date: 2023-01-25-04-05
Reported on: 2023-01-25 05:12:44
Map: DONUT3

Retrino:
fixed

PR #13993, opened 2023-05-06
Makes the mop clean tracked blood again

### LINKED ISSUE #12597: Mop doesn't clean up bloody footprints
### Labels

[BUG] [MAJOR] [GAME OBJECT]

### Description

As title, the mop no longer cleans up bloody tracks by default.

### Steps to reproduce

1. Wet a mop (WITHOUT CLEANER)
2. Attempt to mop up tracked blood
3. Tracks are not cleaned up

### Expected Behavior

I expected the mop to clean up tracks as it had previously

I double-checked from an older copy of the codebase to confirm if this actually changed or not. I'm assuming that this wasn't an intentional change

RubberRats:
[CODE QUALITY][BUG][GAME OBJECTS]
## About the PR 

Changes all usages of `cleanable_fluid_react()` to use nulls and TRUE/FALSE

This PR makes bloody footprints get cleaned up when a fluid passes over them. They will still remain if another blood decal gets place on their tile. This allows mops to clean them once more.

## Why's this needed? 

e1fee89 changed blood tracks to not get absorbed when somebody bled on them but also made them unmoppable. This change keeps that while making them moppable like before.
Fixes #12597 and the very similar #12875

## Changelog 

```changelog
(u)RubberRats
(+)Bloody footprints are moppable again.
```

### DIFF
--- a/code/modules/fluids/fluid_procs.dm
+++ b/code/modules/fluids/fluid_procs.dm
@@ -131,11 +131,11 @@ turf/simulated/floor/plating/airless/ocean_canpass()
 		if (istype(T) && T.messy > 0)
 			var/found_cleanable = 0
 			for (var/obj/decal/cleanable/C in T)
-				if (istype(T) && !T.cleanable_fluid_react(C, 1)) // Some cleanables need special treatment
+				if (istype(T) && !T.cleanable_fluid_react(C, TRUE)) // Some cleanables need special treatment
 					found_cleanable = 1 //there exists a cleanable without a special case
 					break
 			if (found_cleanable)
-				T.cleanable_fluid_react(0,1)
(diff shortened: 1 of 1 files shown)


flappybatpal (approved):
Functions as expected but needs further review for any potential implications of the changes.

BatElite:
I take it this fixes the floor buffer as well?

RubberRats:
> I take it this fixes the floor buffer as well?

The change specifically affects what happens when a fluid puddle touches a cleanable, so it _should_ also affect the buffer.

TobleroneSwordfish (approved):
makes sense to me, should only affect fluids specifically

BatElite:
Bless this and godspeed, can't wait for janitoring to not suck anymore

PR #14001, opened 2023-05-07
Fixes handcuffed sprites for chimps

DimWhat:
[bug] [mutantrace]
## About the PR 
As the title says it fixes the unique handcuff sprites monkeys have when they are cuffed. Should make detaining all sorts of angry chimps less buggy looking. However slightly.
image

## Why's this needed? 
It was in the wrong perspective. Why not fix it?


flappybatpal:
!merge_upstream

PR #14003, opened 2023-05-07
Some Extra Colorful Tools

Flaborized:
[FEAT]

## About the PR 

Adds some colorful tools which are tied to different sources. Printed tools are green, QM purchased tools are orange. Robotics toolbelt has grey, and the starting doctor crowbar is a bright blue. Includes an orange-stripe multitool from the QM crates. Regular tools that spawn in toolboxes or on tables or what-have-you remain generic (or red). There's no mechanical difference at all between any of the tools, they're all 'just tools'.

Also apparently the yellow welder set its item state to a unique in-hand that, as far as I can tell, doesn't exist, so cleaned that up. Also gave it a description to fit in with all the other colorful tools.

## Why's this needed? 

This is pretty much me expanding on the yellow engineer tools I made a while ago. The intent is that it'd be cool to be able to tell where someone got their tools from if you see them, and gives them a sorta cute personalized 'history'. Not really useful detective work, but still neat. You can look at someone's tools and go "Oh they got this from an engineer" or "They bought this from QM" or "they printed this" and so on. I also think people might develop preferences for some which I think could be fun.

Originally I made the grey tools for fabricators, but they seemed way too robotics-y to me after testing in-game, and they wouldn't feel right being common imo. I picked green since it's a very basic color, but also decently nice looking. I'd be open to changing it to something if the consensus is that it doesn't really fit fabricators (or adding new colors if there's some large source of them that could do with having one)

Also you could theoretically arrange an almost-full rainbow of crowbars which I think is a positive

## Changelog 

```changelog
(u)Flaborized
(+)Tools now come in additional colors, based on where they're sourced from.
```


CaioFrancisco:
I honestly feel like coloring tools if they came from a manufacturer is a bit too much. There should be a source of normal tool sprites, since I like how it already looks like. And I kinda think that the Robotics tools sprites are **too** grey, specially the tools that aren't the multitool nor the welder. It's practically monochrome, and I think it clashes against a bunch of other colored sprites. 

Although, here's an idea, what if the tool color changed on the `color` var? It could mean that someone could reasonably paint their own tools, and gives the players some way to customize their own tools. I'd personally love to see that happen.

ZeWaka:
> Although, here's an idea, what if the tool color changed on the `color` var? It could mean that someone could reasonably paint their own tools, and gives the players some way to customize their own tools. I'd personally love to see that happen.

Not technically possible to do that with that var. It would need specific code for paint cans and such.

Flaborized:
> I honestly feel like coloring tools if they came from a manufacturer is a bit too much. There should be a source of normal tool sprites, since I like how it already looks like. 

I get this perspective, but changing fabricated tools is part of the idea of "visually showing where they came from". Anything that spawns on a table or in a toolbox lying around or whatever will still have default colors, I don't think this would by any means remove the originals, even though technically they'd be in limited supply as this replaces the two renewable sources with colorful variants.

> And I kinda think that the Robotics tools sprites are too grey, specially the tools that aren't the multitool nor the welder. It's practically monochrome, and I think it clashes against a bunch of other colored sprites.

These use the same colors present on the roboticists' jumpsuits, I think they match the belt quite well in particular. This is mostly why I have them like this

BatElite:
So unless someone goes and maps them in, the grey multitool is unobtainable?

Also, grey and green multitools are hard to distinguish at a glance

Flaborized:
> So unless someone goes and maps them in, the grey multitool is unobtainable?
> 
> Also, grey and green multitools are hard to distinguish at a glance

Correct yeah. Also, green multitools aren't a thing in this, I accidentally included them in the screenshot of my aseprite file because I got all the tools by copying them over and swapping the colors, and I forgot to remove them (also why there are two orange wirecutters). Pretend green multitool isn't real :sheltersweat:

DimWhat:
If this PR is merged I'll probably make the wrench that spawns by cryo on most maps have a blue handle like the crowbar.

Sord213 (approved):

PR #14019, opened 2023-05-08
Recharge upgrades can only be used once every 7 minutes as borg.

ThatFiveGuys:
## About the PR 
When you use a recharge pack as a cyborg, you get the quick charged status effect for seven minutes. While you have this status effect you cannot activate another recharge pack. This also refactors some code to make this work

## Why's this needed? 
For powerborging, the downside of recharging can practically be avoided through a teleporter upgrade and a charging dock loaded with recharge packs. This makes charging packs a compliment to charging instead of a replacement

## Changelog 

```changelog
(u)Ikea
(*)There is now a seven minute cooldown for using cyborg recharge upgrades 
```


code/modules/status_system/statusEffects.dm
```diff
@@ -2335,6 +2335,15 @@
 #undef LAUNDERED_COLDPROT_AMOUNT
 #undef LAUNDERED_STAIN_TEXT
 
+/datum/statusEffect/quickcharged
+	id = "quick_charged"
+	name = "Quick charged"
+	maxDuration = 7 MINUTES
+
```
Tarmunora, line 2342: does this have an icon?

ThatFiveGuys: Oh shoot forgot to add that uhh one second

BatElite:
What problem needs solving here? Getting a teleport + recharges sounds like an active effort on the borg's part to gain benefit, so good on them?

Goat-Real:
Are we trying to solve an issue of cyborgs using the recovery upgrade a lot? Maybe a different idea would be adding a delay on the teleporter upgrade. Change it from an instant vanish button to an escape with couple seconds of charge time.

ThatFiveGuys:
> Are we trying to solve an issue of cyborgs using the recovery upgrade a lot? Maybe a different idea would be adding a delay on the teleporter upgrade. Change it from an instant vanish button to an escape with couple seconds of charge time.

This would be a good change but outside the scope of this pr

TobleroneSwordfish (changes requested):

PR #14022, opened 2023-05-08
Mobifies Golems

CheffieGithub:
## About the PR 
Mobifies and reblances golems a bit.

Golems are slow and have 40 health with high resistance.
They have strong arms which have longer delay and transfer reagents like before.
Spell cooldown increased by 25 seconds.

## Why's this needed? 
Objects bad


TDHooligan:
i've not heard any balance concerns around current golems, doubling(brute)/tripling(burn) their health seems unneccessary. this should also be labelled as a balance change and actually given changelog entries.  

i'm a little concerned every critter mobification so far has been stacked with balance changes and not marked or labelled as such. it's not transparent to the folks who aren't devs, i've gone back and looked through and with surprising regularitiy mobifications have changed core critter behaviour. Transposed scientists and shades aren't fireproof any more etc.

surely patches labelled 'mobification' should be about maintaining close to identical critter behaviour.

CheffieGithub:
burn health is slightly reduced and brute health is very slightly increased unless vuln is different between mob and obj
burn health was 100 (25 with 0.25 vuln) now 80 (40 with 0.5 vuln)
brute health was 50 (25 with 0.5 vuln) now 53.3 (40 with 0.75 vuln)

Making changes to health and behaviour in the initial PR is easier then making multiple after and mobification opens a lot of doors to make changes such as with movement speed and abilities not previously possible with objects.

And buffs are in order for many object critters since some before were just complete jokes, such as lions who had 10 health and 5-7 damage. but I suppose it would make sense to label mobifcation PRs with balance as well since just being a mob changes balance.

Thanks for bringing up that scientist goof that was my mistake ill fix it soon.

code/datums/limb.dm
```diff
@@ -1557,3 +1557,28 @@
...
+			logTheThing(LOG_COMBAT, user, "attacks [constructTarget(target,"combat")] with a golem arm at [log_loc(user)].")
+
+		if(target.reagents)
+			if(user.reagents && user.reagents.total_volume)
+				user.reagents.reaction(target, TOUCH)
+				user.reagents.trans_to(target, 5)
+
+		var/datum/attackResults/msgs = user.calculate_melee_attack(target, 6, 9, rand(4, 6), can_punch = FALSE, can_kick = FALSE)
+		user.attack_effects(target, user.zone_sel?.selecting)
+		msgs.base_attack_message = "<b><span class='alert'>[user] punches [target] with [src.holder]!</span></b>"
```
Tarmunora, line 1578: > `can_punch = FALSE`

> `[user] punches [target] with [src.holder]`

Well which is it?

Tarmunora: 🤷 I'd have just let them punch, but this works too

Tarmunora (approved):

PR #14026, opened 2023-05-09
Adds a bunch of blinds to Donut 3

frawhst:
## About the PR 

blinds are cool we should have more of those
adds blinds to some areas including:
- ranch
- toxins
- botany
- QM
- engi break room
- genetics
- medical break room
- mining
- refinery
- and more!

Non-blind changes:
- Changes the layer of windoors and blinds so that blinds layer above windoors but below lights
- Changes some doors to opaque where a transparent door would counteract the blinds

## Why's this needed? 
privacy (crime)


frawhst:
ok i merge

PR #14031, opened 2023-05-09
microwave reagent warming mode / skull microwaving

Skeletonman0:
[OBJECTS] [QOL]
## About the PR 
adds a warming mode to microwaves
- this mode is usable if there is an organ or an item that contains reagents in the microwave
    - it only works if there are no other ingredients
    - the code for heating the reagents is borrowed from igniters, unsure if the values are okay

- also allows organs / skeleton skulls to be microwaved
    - microwaving a skeleton skull uses code borrowed from tanning beds to stun the skeleton

i also replaced the sleep() calls with SPAWN() since that seemed like a good idea

## Why's this needed? 
people keep microwaving skeleton skulls, which is funny

would be nice to have the ability to heat stuff up at the space diner or something

## Changelog 

```changelog
(u)Skeletonman0
(*)Microwaves can now warm chemicals and skeleton skulls / other organs are now microwavable
```


nefarious6th:
would love this for bogus debris field chem

PR #14038, opened 2023-05-10
Adds a bunch of blinds to Oshan

frawhst:
## About the PR 
Same as the last one but it's Oshan

non-blind things:
- adds some windoors to the HoP deck for security (and so I had an excuse to add more blinds)
- replaces transparent doors with opaque ones to preserve blind  integrity
## Why's this needed? 
crime (privacy)

### DIFF
--- a/maps/oshan.dmm
+++ b/maps/oshan.dmm
@@ -1232,8 +1232,8 @@
 /turf/simulated/floor/shuttlebay,
 /area/shuttle/sea_elevator_room)
 "acM" = (
-/obj/machinery/door/airlock/pyro/glass,
 /obj/firedoor_spawn,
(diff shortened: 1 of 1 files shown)


Sord213 (approved):
lgtm (for crime)

Cherman0:
I feel like its not really necessary? Oshan (and Donut3, for that matter) already have pretty low visibility on account of having fairly deep maintenance and regular department rooms stretching past vision range. If blind sprites were cleaner looking (and had full spritesets for insides and outsides of windows, so they wouldnt be inconsistently on the inside or outside depending on what sprites already exist for) and blind switches weren't identical to lightswitches (invisible) then I feel like I'd be less apprehensive about it.

PR #14046, opened 2023-05-11
fix magboots anchoring people to space turfs

### LINKED ISSUE #12890: Magboots can anchor you to space
### Labels

[BUG][minor]

### Description

You won't get space drift with magboots, even though the magnetism would imply it's sticking you to the station structure only.

### Steps to reproduce

1. Get magboots, jetpack and space wear
2. go out into space, observe your space drifting when you let go of movement
3. Activate magboots
4. Stop movement, you can precisely position in the middle of space

### Expected Behavior

space drift continuing

_No response_

Skeletonman0:
[OBJECTS] [BALANCE] [CONTROVERSIAL]
## About the PR 
fixes #12890 by making magboots unanchor / reanchor the user when they move into space / stop space drifting

also makes it update the magboots of people when trying to throw them just to be safe

based on the reactions when i mentioned possibly fixing this in imcoder, this is apparently controversial and has balance implications as it causes magboot users to drift in space

currently unsure if this is even a good idea
this also probably isnt the best way to code this aswell

## Why's this needed? 
magboots probably shouldn't anchor people to space itself

## Changelog 

```changelog
(u)Skeletonman0
(+)Magboots nolonger anchor people to space
```


M-Earthfire:
Do they still stay active when you are adjacent to a surface but on a space tile? Because that would be very important for all of magboots applications within engineering.

Sord213:
This isn't really a bug, it's an intentional feature. Mostly from the days of space wind being a pain in the ass but still. Not opposed, but eh

PR #14049, opened 2023-05-11
Fix richest escapee being incorrect

FlameArrow57:
[PLAYER ACTIONS][BUG]

## About the PR 
This PR fixes richest escapee not counting cash on mobs

## Why's this needed? 
Bug fix

### DIFF
--- a/code/procs/scorestats.dm
+++ b/code/procs/scorestats.dm
@@ -288,7 +288,7 @@ var/datum/score_tracker/score_tracker
 			. += SC.amount
 		else if (istype(A, /obj/item/card/id))
 			var/obj/item/card/id/ID = A
-			. += ID.amount
+			. += ID.money
 		else if (A.storage)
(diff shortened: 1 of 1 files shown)


code/procs/scorestats.dm
```diff
@@ -283,13 +283,18 @@ var/datum/score_tracker/score_tracker
...
-				. += SC.amount
-			else if (istype(I, /obj/item/card/id))
-				var/obj/item/card/id/ID = I
-				. += ID.amount
+		if (istype(A, /obj/item/spacecash))
+			var/obj/item/spacecash/SC = A
+			. += SC.amount
+		else if (istype(A, /obj/item/card/id))
+			var/obj/item/card/id/ID = A
+			. += ID.amount
```
Tarmunora, line 291:
```suggestion
			var/obj/item/card/id/ID = A
			. += ID.money
```
Pretty sure amount is not correct here...

PR #14064, opened 2023-05-12
makes sticky floor slowdown noticeable

colossusqw:
[LABEL][balance][status-effects]

## About the PR 
This PR changes the amount of slowdown caused by sticky floors from the 2 it curently is to the standard 10 that all slowdowns cause.

## Why's this needed? 
There has been discussion about buffing glue slowdowns on the forums, since it's a barely seen chem, as well as it being terminally bad sticky-floor-wise to the point of it being unnoticeable. Even compared to the basically instant to make space lube, it is miles behind.

```changelog
(u)Colossusqw
(+)Sticky floors, such as the ones made by glue, now slow you down to a noticeable amount.
```


colossusqw:
https://forum.ss13.co/showthread.php?tid=21185 here's the forum post

PR #14069, opened 2023-05-12
Replaces The Armoury's 40mm Pod Seekers With A Pod-Targeting Missile Launcher

Mister-Moriarty:
[Game Objects] [Feature] [QoL]

## About The PR:
Replaces the Armoury's 40mm HE pod-seeking shells with a pod-targeting missile launcher and four pod-seeking missiles.

image

The launcher itself is collapsible, and may be carried in a single hand, a backpack, or its box in its collapsed state, but requires to be extended before it is able to be loaded and fired. In its extended state, it may only be carried inhand, and takes two hands to carry.

When extended, the launcher will attempt to target nearby pods with a new smartgun component subtype; only one pod may be targeted at a time, and will be displayed inside an aiming reticle on the player's screen, as seen above.

Unlike the smartgun, the launcher will not fire a missile in the immediate direction of the pod, aiming is relegated to the player; the fired missiles will instead attempt to course-correct towards the targeted pod, much akin to the pod-seeking shells the missiles are meant to replace.

## Why Is This Needed?
My primary motivation behind devising a replacement to the 40mm pod-seeking shells was that it is currently exceedingly difficult to ensure that the shells fire at the correct pod when multiple pods are involved, i.e. in a Security-Salvager dogfight, or firing to space from a pod bay, and it was suggested in Discord after one such round that pod seekers would benefit from possessing a similar tracking HUD to the NukeOp's smartpistols.

## Changelog:
```changelog
(u)Mr. Moriarty
(*)Replaced the Armoury's 40mm pod-seeking shells with a pod-targeting missile launcher. The launcher will display an aiming reticle over the pod being targeted.
```


CheffieGithub:
emag act for people explosions? or bad homing

Mister-Moriarty:
> emag act for people explosions? or bad homing

I don't believe that offering a somewhat inferior RPG-7 to those with EMAGs would be particularly balanced. The inaccurate homing idea has more merit, although I don't see that there would be much incentive to EMAG the launcher.

FlameArrow57:
Hm, I think most use cases of pod seekers are when having a jetpack and holding a riot launcher and a backpack in the other hand, which has extra rounds in it. 

If this needs to be extended to be reloaded, don't you need to click ammo that's on the ground with it or vice versa to reload it?  I think it's one shot to take out a small pod? For large pods though I feel like that would make using it much slower, they could escape at the first sight of using a pod seeker and it would make space fights much more difficult or near impossible to happen, it pretty much requires you to pull a backpack along with you and have precision clicking on your backpack in space to reload, while pod you're fighting could hit you into space or shoot you still

CheffieGithub:
> I don't believe that offering a somewhat inferior RPG-7 to those with EMAGs would be particularly balanced. The inaccurate homing idea has more merit, although I don't see that there would be much incentive to EMAG the launcher.

Then just smack them with more disorientate and a bit more extra damage? or a weaker explosion such as potassium water

Mister-Moriarty:
> Hm, I think most use cases of pod seekers are when having a jetpack and holding a riot launcher and a backpack in the other hand, which has extra rounds in it.
> 
> If this needs to be extended to be reloaded, don't you need to click ammo that's on the ground with it or vice versa to reload it? I think it's one shot to take out a small pod? For large pods though I feel like that would make using it much slower, they could escape at the first sight of using a pod seeker and it would make space fights much more difficult or near impossible to happen, it pretty much requires you to pull a backpack along with you and have precision clicking on your backpack in space to reload, while pod you're fighting could hit you into space or shoot you still

This is a valid concern which I did not consider when developing this PR. Could this potentially be alleviated by permitting the launcher and ammunition to be stored inside of belts? Suggestions on this issue would be appreciated.

Exester509:
> > Hm, I think most use cases of pod seekers are when having a jetpack and holding a riot launcher and a backpack in the other hand, which has extra rounds in it.
> > If this needs to be extended to be reloaded, don't you need to click ammo that's on the ground with it or vice versa to reload it? I think it's one shot to take out a small pod? For large pods though I feel like that would make using it much slower, they could escape at the first sight of using a pod seeker and it would make space fights much more difficult or near impossible to happen, it pretty much requires you to pull a backpack along with you and have precision clicking on your backpack in space to reload, while pod you're fighting could hit you into space or shoot you still
> 
> This is a valid concern which I did not consider when developing this PR. Could this potentially be alleviated by permitting the launcher and ammunition to be stored inside of belts? Suggestions on this issue would be appreciated.

I'd say making both the launcher and the rockets able to be stored in a standard security belt would be enough, allowing you to carry 2-3 rounds of ammo. Alternatively, you could add a specialized belt able to hold the launcher and the rockets

FlameArrow57:
Allowing it to be stored in sec belts would be much nicer, but you still won't be able to hold your backpack very well for space fights, which is an added risk that wasn't there before, would have to pull it still or hide it somewhere, which wouldn't be very good if antagonists or others have sec access. If you want to hold the standard security gear and you start with and have utilities, inventory space could be really tight still

Exester509:
> Allowing it to be stored in sec belts would be much nicer, but you still won't be able to hold your backpack very well for space fights, which is an added risk that wasn't there before, would have to pull it still or hide it somewhere, which wouldn't be very good if antagonists or others have sec access. If you want to hold the standard security gear you start with and have utilities, inventory space could be really tight still

I feel like that isn't too bad of a trade, not to mention you can venture out without a jetpack and shoot from the ship's hull if you value the backpack so much. Besides, if armory is authorized, antagonists aren't gonna go for your backpack (especially since most sec gear goes on the pouch or belt), they're gonna go for armory gear, which is a much bigger concern

Exester509:
> really I don't think this needs to fit into belts considering that grenade launchers never did

Considering grenade launchers could shoot ANY grenade, and pod hunting was secondary at most, they made sense not being able to fit in a belt. Plastic batons and shot flashbangs are particularly devastating, and neither can be used with this.

code/obj/item/gun/kinetic.dm
```diff
@@ -1719,10 +1719,12 @@ ABSTRACT_TYPE(/obj/item/gun/survival_rifle_barrel)
 			if (src.collapsed)
 				src.w_class = W_CLASS_NORMAL
 				src.has_empty_state = FALSE
+				src.cant_be_placed_into_storage = FALSE
 
 			else
 				src.w_class = W_CLASS_BULKY
 				src.has_empty_state = TRUE
+				src.cant_be_placed_into_storage = TRUE
```
FlameArrow57, line 1727: w_class is already W_CLASS_BULKY here, I don't think any storage can hold something of this size atm

code/obj/item/gun/kinetic.dm
```diff
@@ -1643,6 +1643,119 @@ ABSTRACT_TYPE(/obj/item/gun/survival_rifle_barrel)
...
+
+			if (src.collapsed)
+				src.w_class = W_CLASS_NORMAL
+				src.has_empty_state = FALSE
+				src.cant_be_placed_into_storage = FALSE
+
+			else
+				src.w_class = W_CLASS_BULKY
+				src.has_empty_state = TRUE
+				src.cant_be_placed_into_storage = TRUE
```
FlameArrow57, line 1727: Could probably use a new W_CLASS_UNHOLDABLE define, there's a check_wclass variable in storage which works with can_hold

Mister-Moriarty: `check_wclass` does not check the `w_class` of objects in the `can_hold` list; I'll implement the new define, however there will still require to be an explicit check for it in the component.

pali6:
Neat

_std/defines/item.dm
```diff
@@ -194,6 +194,7 @@
 #define W_CLASS_BULKY 4
 #define W_CLASS_HUGE 5
 #define W_CLASS_GIGANTIC 6
+#define W_CLASS_UNSTORABLE 7
```
Tarmunora, line 197: this will have odd side-effects, such as monkeys not picking up anything of this class, and TK pull not working on them either. I don't think that this is a good approach - consider one of the flags vars instead?

Tarmunora (changes requested):
Mechanically, this looks pretty cool and only have some minor nitpicks

Aesthetically though, I'm really not a fan, not for something going in the armory - it gives off much more of a (para)military vibe than anything else in there imo.

code/modules/projectiles/bullet.dm
```diff
@@ -1400,12 +1400,14 @@ datum/projectile/bullet/autocannon
     var/easemult = 0.
 
     var/auto_find_targets = 1
+    var/list/targets = list()
```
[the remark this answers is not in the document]
Mister-Moriarty: `/datum/projectile/bullet/homing/var/targets` is currently only used to pass a list of atom references to `/obj/projectile/var/targets`, which will also prevent targets from GCing.
I could null the list after the references have been passed to the projectile object, but that does not in any way alleviate the issue, given that references to them will still exist in `/obj/projectile/var/targets`.

I was unable to find a method whereby I could pass the list of targets directly to the projectile object.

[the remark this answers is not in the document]
Tarmunora: well, the projectile object will get cleaned up after the projectile dies, so that's fine - the list on the datum will stick around longer

Tarmunora:
Hm. I still like the concept of this and maybe I can get someone else to resprite it to something less... "ntso-ish" down the line 🤷

PR #14079, opened 2023-05-13
Add forcefields to Oshan exterior doors

flappybatpal:
## About the PR 
Adds three door linked forcefields to single door exits on Oshan.

## Why's this needed? 
I noticed a massive flood from a the north east set of these doors being left open and they weren't even hacked.
While flooding is not as dangerous as oxygen loss in space Oshan has something of a trend for massive floods. It feels a bit excessive to me that potentially this can just happen from regular doors.
This doesn't touch the larger sets of doors which have more drainage/visibility.


M-Earthfire:
Oshan normally has sinks for that in most areas. I would prefer these to forcefields, since they can be welded shut by antags

flappybatpal:
Drains are weak against a constant influx of water. These are door linked forcefields so they can be removed by destroying the door. Alternatively there is a heck of a lot of plasma glass windows that can be removed or destroyed.

PR #14081, opened 2023-05-13
Fix queen spider AI breaking on vomit egg

CheffieGithub:
## About the PR 
fix issue where queen clown spiders get stuck on vomiting egg

## Why's this needed? 
bad bug

### DIFF
--- a/code/mob/living/critter/ai/spider.dm
+++ b/code/mob/living/critter/ai/spider.dm
@@ -21,11 +21,41 @@
 	transition_tasks += holder.get_instance(/datum/aiTask/timed/wander/critter/aggressive, list(holder, src))
 	transition_tasks += holder.get_instance(/datum/aiTask/sequence/goalbased/critter/attack, list(holder, src))
 	transition_tasks += holder.get_instance(/datum/aiTask/sequence/goalbased/critter/scavenge, list(holder, src))
-	transition_tasks += holder.get_instance(/datum/aiTask/succeedable/critter/vomit_egg, list(holder, src))
+	transition_tasks += holder.get_instance(/datum/aiTask/sequence/goalbased/critter/vomit_egg, list(holder, src))
+
+/datum/aiTask/sequence/goalbased/critter/vomit_egg
+	name = "clown spider vomit egg"
+	weight = 1
+	max_dist = 5
+	can_be_adjacent_to_target = FALSE
+
+/datum/aiTask/sequence/goalbased/critter/vomit_egg/New(parentHolder, transTask)
+	..(parentHolder, transTask)
+	add_task(holder.get_instance(/datum/aiTask/succeedable/critter/vomit_egg, list(holder)))
+
+/datum/aiTask/sequence/goalbased/critter/vomit_egg/precondition()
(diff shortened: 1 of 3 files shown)


code/mob/living/critter/ai/spider.dm
```diff
@@ -21,11 +21,42 @@
...
+/datum/aiTask/sequence/goalbased/critter/vomit_egg/New(parentHolder, transTask)
+	..(parentHolder, transTask)
+	add_task(holder.get_instance(/datum/aiTask/succeedable/critter/vomit_egg, list(holder)))
+
+/datum/aiTask/sequence/goalbased/critter/vomit_egg/precondition()
+	var/mob/living/critter/spider/clownqueen/C = holder.owner
+	var/datum/targetable/critter/vomitegg/egg = C.abilityHolder.getAbility(/datum/targetable/critter/vomitegg)
+	if(!egg)
+		egg = C.abilityHolder.getAbility(/datum/targetable/critter/vomitegg/cluwne)
+	return !egg.disabled && (length(C.babies) < max_spiders)
```
TobleroneSwordfish, line 42: Clownspiders don't remove themselves from this list on death, which is a GC issue and will mean this makes them only ever spawn 4 total.

TobleroneSwordfish (changes requested):

code/mob/living/critter/ai/spider.dm
CheffieGithub: yeah I found this out while I was testing it but I dunno how to fix

TobleroneSwordfish (changes requested):
If you're going to make them weakrefs then you'll need to update the other code that accesses that list (it's just the spider `was_harmed` proc I think) to deref them properly, also probably best stick a comment on it to say it's full of weakrefs.

CheffieGithub:
forgot about that part

code/mob/living/critter/spider.dm
```diff
@@ -495,7 +495,10 @@
 			return
 		var/defenders = 0		//this is the amount of babies that will defend you
 		var/count = 0
-		for (var/mob/living/critter/spider/clown/CS in babies)
+		for (var/datum/weakref/ref as anything in babies)
+			var/mob/living/critter/spider/clown/CS = ref.deref()
+			if (CS == null)
+				babies.Remove(ref)
```
TobleroneSwordfish, line 501: you'll want to `continue` here or it's probably still going to runtime below (might get caught by `GET_DIST`, but I wouldn't rely on it)

code/mob/living/critter/spider.dm
```diff
@@ -495,7 +495,10 @@
 			return
 		var/defenders = 0		//this is the amount of babies that will defend you
 		var/count = 0
-		for (var/mob/living/critter/spider/clown/CS in babies)
+		for (var/datum/weakref/ref as anything in babies)
+			var/mob/living/critter/spider/clown/CS = ref.deref()
+			if (CS == null)
```
TobleroneSwordfish, line 500:
```suggestion
			if (!istype(CS))
```

TobleroneSwordfish (changes requested):

PR #14086, opened 2023-05-13
Dispensary interlink

TobleroneSwordfish:
## About the PR 
Adds a dispensary interlink to every medbay and chemlab. (except cog 1, where there's just one by the little shared desk bit)
image

- Contents are synched, allowing chemistry to conveniently supply chemicals to medbay.
- They accept bottles, pills, pill bottles, patches, ampoules and syringes.
- They will loudly announce when they receive new supplies and occasionally advertise their contents.
- Each interlink has replaced one out of three Nanomeds.
- They are locked to `medical_locker` or `chemistry` access.

Also if someone can come with a less painfully generic name for them that would be cool, I have no imagination.
## Why's this needed? 
Easier and more organised way for chemistry to supply medbay with useful medical supplies.

## Changelog 

```changelog
(u)LeahTheTech (name courtesy of Kubius)
(*)Added dispensary interlinks to medbay and chemistry lab for easier and more organised supply of medical chemicals.
(+)The interlinks replace one existing nanomed, synchronize their contents and accept bottles, pills, pill bottles, patches, ampoules and syringes.
```


ThatFiveGuys:
Make this thing cost telecrystals?

TobleroneSwordfish:
Cost telecrystals to use or to build?

Sbmhawk:
Having it replace nanomeds is pretty major imo.

ThatFiveGuys:
> Cost telecrystals to use or to build? @ThatFiveGuys

Build

TobleroneSwordfish:
> Having it replace nanomeds is pretty major imo.

It is, but medbay already gets a ludicrous amount of supplies at rounstart and I bet 90% of players have never found the third nanomed on most maps.

nefarious6th:
i like this. i just spent a chemistry round running barrels of stuff to medbay, so a fast way to stock immediately to them is really nice. i assume the 3 medvends doesnt necessarily include the ones in the medbay booths? if it does, then i think maaaaaybe there's a bit of a problem, but if it's replacing one of three within the medbay itself, i think it's fine.

also bad at names but maybe pharmacy cabinet? it would indicate that science and/or med could contribute to it then?

Kubius:
Name ideas:

- RX Distributor
- Pharm-o-Matic
- dispensary endpoint
- dispensary interlink

On the map side:

- I would prefer that Nadir's treatment center not be left without a vendor, and this vendor would be a good fit for it. The new one that was added is also in a good spot (I think you can have 3 total?) 
- Similar deal for Kondaru - main room keeps its nanomed (as it's the main triage area) and treatment center gets the view vendor.
- As a general thing, it might be good to have one of these in the medical booth of maps which have a medical booth (in Kondaru's case, it could perhaps replace the medicine storage locker there?)

TobleroneSwordfish:
I like the name ideas, as for the map issues: the problem with the treatment center areas is that no one ever realistically goes there, and this machine relies on being seen to be utilized (the overhead text announcements etc.). Also part of this change is reducing the amount of supplies medbay has access to at the start of the round to a reasonable amount (3 nanomeds down to 2), although I'm not 100% set on that if it's unpopular.

codeJester27:
Can anyone insert medicine into the machine regardless of access? If so, could these be added into/near hydroponics?

Exester509:
> Can anyone insert medicine into the machine regardless of access? If so, could these also be added into/near hydroponics?

I think the chemistry-hydroponics relation suffices with mail chutes, while medbay does need these supplies stored in a specific, accessible place rather than a crate or a locker. This way, all doctors, regardless of their communication with chemistry that round, know where to find the supplies, which is something pretty important for medbay.
Hydroponics having less personnel and duties allows them to just ask for chemicals when needed, and mail back what chemistry asks for without much hassle.

TobleroneSwordfish:
I kinda like the idea of adding them to hydroponics too, but that can happen after seeing how they work in game.

PR #14089, opened 2023-05-13
Adds A Proc For Mob Critters Validating Targets

Wisemonster:
[CODE QUALITY] [INTERNAL] [QOL]

## About the PR 
This adds a new proc, valid_target, that helps tell mob critters what to attack and what to ignore. This also replaces a fair amount of copy-pasted code to use this proc instead.

## Why's this needed? 
Currenty, a lot of mobs critters that have unique interactions with other mobs need to copy-paste the same for loop. This helps reduce that copy-pasted code, and makes it easier to add unique interactions for mob critters.

### DIFF
--- a/code/mob/living/critter/adventure.dm
+++ b/code/mob/living/critter/adventure.dm
@@ -368,7 +368,6 @@
 		return ..()
 
 	seek_target(var/range = 5)
-		. = ..()
 
 	critter_basic_attack(var/mob/target)
 		if(prob(30))
--- a/code/mob/living/critter/adventure.dm
+++ b/code/mob/living/critter/adventure.dm
@@ -73,13 +73,7 @@
 				src.say( pick("Please! Help! I need help!", "Please...help me!", "Are you real? You're real! YOU'RE REAL", "Everything hurts! Everything hurts!", "Please, make the pain stop! MAKE IT STOP!") )
 
 	seek_target(var/range = 5)
-		. = list()
-		for (var/mob/living/C in hearers(range, src))
-			if (isintangible(C)) continue
-			if (isdead(C)) continue
-			if (istype(C, src.type)) continue
-			if (src.faction != null && (C.faction & src.faction != 0)) continue //Checks if they are in the same faction
-			. += C
+		.=..()
 
 		if (length(.) && prob(5))
 			src.say(pick("Please...help...it hurts...please", "I'm...sick...help","It went wrong.  It all went wrong.","I didn't mean for this to happen!", "I see everything twice!") )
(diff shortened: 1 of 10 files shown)


code/mob/living/critter/adventure.dm
```diff
@@ -73,13 +73,7 @@
 				src.say( pick("Please! Help! I need help!", "Please...help me!", "Are you real? You're real! YOU'RE REAL", "Everything hurts! Everything hurts!", "Please, make the pain stop! MAKE IT STOP!") )
 
 	seek_target(var/range = 5)
```
TobleroneSwordfish, line 75: May as well remove this definition as it's just calling parent now. (same for all the other definitions that are just doing `. = ..()`)

code/mob/living/critter/mimic.dm
```diff
@@ -112,13 +112,14 @@
 			pounce.handleCast(target)
 			return TRUE
 
+	valid_target(mob/living/C) //apparently they don't have the check for isdead, so gotta do this
```
TobleroneSwordfish, line 115: since dead is a subset of incapacitated you could just check `is_incapacitated` and then return parent

code/mob/living/critter/small_animal.dm
```diff
@@ -1848,17 +1846,18 @@ var/list/mob_bird_species = list("smallowl" = /mob/living/critter/small_animal/b
...
 			return .
 		for (var/mob/living/C in hearers(range, src))
-			if (isdead(C)) continue //don't attack the dead
-			if (isintangible(C)) continue //don't attack the AI eye
-			if (istype(C, src.type)) continue //don't attack other scorpions
-			if (istype(C, /mob/living/critter/small_animal/rattlesnake)) continue //don't attack space rattlesnakes(the snake would lose)
-			if (C in src.friends) continue //don't attack frens :)
-			. += C
+			if (src.valid_target(C))
+				. += C
```
TobleroneSwordfish, line 1860: shouldn't this be a parent call?

code/mob/living/critter/small_animal.dm
```diff
@@ -1975,29 +1974,29 @@ var/list/mob_bird_species = list("smallowl" = /mob/living/critter/small_animal/b
...
-						SPAWN(1 SECONDS)
-							icon_state = "rattlesnake"
-						src.visible_message("<span class='combat'><B>[src]</B> rattles, better not get much closer!</span>")
-					continue
-				else if(GET_DIST(src, C) > 3) //humans and silicons that are farther than 3 tiles do not interest the snake
-					continue
-
-			. += C
+			if (src.valid_target(C))
+				. += C
```
TobleroneSwordfish, line 1999: ^

TobleroneSwordfish (changes requested):

code/mob/living/critter/adventure.dm
Wisemonster: Ok, although some of these (for example, the seek_target at this line of code) have if checks after the initial parent call, so I don't think those can be removed.

code/mob/living/critter/mimic.dm
```diff
@@ -112,13 +112,11 @@
...
```
TobleroneSwordfish, line 115:
```suggestion
	valid_target(mob/living/C)
```

PR #860, opened 2020-05-26
Add storage components

flrsh:
[CLEAN] [REWORK] 

## About the PR 

This PR adds storage components, which can be added to any objects. This means you don't have to type your cool new thing as `/obj/item/storage/very_cool` if you don't want to, and can then make use of other types' functionalities.  

I've added a bunch of signals as part of this rework. Some of the `COMSIG_STORAGE` ones can probably be extended to past storage components. All storage functionality should be handled through signals and the storage component. One thing to note, though, is that we still use the object's `contents` to keep track of stored contents; I will refactor that in a later PR (which should then allow all movable atoms, or atoms in general if we want). Additionally, I chose to keep `spawn_contents` on `/obj/item/storage`, to make mapping more convenient.

I've also changed bibles from `/obj/item/storage/bible` to `/obj/item/bible` and added an incomplete lanyard clothing item to serve as an examples.

```
/obj/item/bible
	name = "bible"

	New()
		..()
		START_TRACKING
		#ifdef SECRETS_ENABLED
		ritualComponent = new/datum/ritualComponent/sanctus(src)
		ritualComponent.autoActive = 1
		#endif
		BLOCK_BOOK
		AddComponent(/datum/component/storage/bible)
```

```
/obj/item/clothing/lanyard
	name = "lanyard"

	New()
		..()
		AddComponent(/datum/component/storage, max_wclass = 1, slots = 3)
```

I've tested all of this fairly extensively and everything seems to work as usual. 

## Why's this needed? 

It's a cool thing you can (should) use.

PR #14093, opened 2023-05-14
Add lanyards

FlameArrow57:
[GAME OBJECTS][FEATURE][CLEANLINESS]

## About the PR 
This PR adds a new item, the lanyard. Lanyards fit in your ID slot and are a three item storage that can hold tiny items. They also accept up to one ID card, and if you hold an ID in your lanyard, using the lanyard will work for any ID checks, such as doors and vending machines. They are a random spawn in lootcrates.

Sprites are by Flaborized, from #860. Original idea came from there too.

image

image

Also cleans up a lot of istype() checks around ID cards and PDAs that get the stored ID card into a proc that will do this instead, in order to make this work.

## Why's this needed? 
Thought it would be fun, adds an alternative ID item that can be used and a small storage item

## Changelog 
```changelog
(u)FlameArrow57 and Flaborized
(+)Added lanyards as a random lootcrate item! These are three item storages that can hold tiny items and fit in your ID slot. They accept IDs and can be used for ID checks if holding one.
```


FlameArrow57:
Where these are found isn't put in yet, was looking for feedback. Was thinking loot crates, a new cargo crate, or uniform manufacturers maybe?

nefarious6th:
uniform manufacturers seems okay? not sure how i feel about having something compete with pdas

FlameArrow57:
Added to loot crates, from Flourish's suggestion in Discord

flrsh (approved):

PR #14096, opened 2023-05-14
Fixes double stack doors on cog2

### LINKED ISSUE #14094: Cogmap 2 Pod Bay / PTL has double stacked doors
### Labels

[BUG]

### Description

On Cogmap 2 there are a set of doors between the two pod bays and the PTL. There are multiple doors there.

### Steps to reproduce

1. Load up Cogmap 2 and go to the civilian pod bay.
2. Try to break into the security pod bay via the PTL doors.
3. Notice you hack one set of doors, to reveal: ***another set of doors!***

### Expected Behavior

I expect that there not be multiple doors stacked on the same tile.

image

DanaTarSUS:
## About the PR 
Removes the beamline interlock on that were previously stacked ontop of the airlock doors connecting the regular and security podbay. This does leave room for someone to walk into the ptl while it's on but if cog1 can get away with its' ptl placement then so can cog2.

## Why's this needed? 
Fixes #14094

### DIFF
--- a/maps/cogmap2.dmm
+++ b/maps/cogmap2.dmm
@@ -74221,14 +74221,6 @@
 	layer = 2.5;
 	name = "Beamline"
 	},
-/obj/machinery/door/airlock/pyro/glass{
-	autoclose = 0;
-	density = 0;
-	icon_state = "glass_open";
-	id = "pt_laser";
-	name = "Beamline Interlock";
-	req_access_txt = "54"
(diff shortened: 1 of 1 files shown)


DanaTarSUS:
Just to clarify, it only removes the beamline interlocks that were in this image. I have not removed the ones found in the hallway for obvious reasons.
image

PR #14102, opened 2023-05-14
Changes a "2053" to "[CURRENT_SPACE_YEAR]" in rubber stamps.

DisturbHerb:
[CLEANLINESS][OBJECTS]

## About the PR 
Replaces an instance of the year 2053 in rubber stamp code to instead use the global `CURRENT_SPACE_YEAR` define, which at present is a macro for 2053.

## Why's this needed? 
Hear me out, as silly as it might sound. What if it *wasn't* 2053? Our stamps would be hilaaaariously out of date!


TobleroneSwordfish:
I wonder what fork would have the year *not* be 2053 🤔

DisturbHerb:
> I wonder what fork would have the year _not_ be 2053 🤔

No such thing. Literally impossible.

PR #14103, opened 2023-05-14
Makes Signifer II stun mode fully automatic.

TheColorCyan:
[A-Game-Objects] [C-Balance] [C-QoL] [E-Input-Wanted]

## About the PR 
Makes Signifer II handgun mode full auto and makes in less pain to use in general. Gives it a bit more spread angle as well.

## Why's this needed? 
Signifer II stun mode was in general very painful to use because of the spam click it required. Some HoS players say it's very uncomfortable to use and that it would be better as a full auto gun.
Signifier II stun mode on full auto drains energy pretty quickly, has decent spread angle and doesn't stun for that long so it seems good in terms of balance.

## Changelog 

```changelog
(u)Sian
(+)Makes Signifer II stun mode fully automatic.
```

### DIFF
--- a/code/obj/item/gun/energy.dm
+++ b/code/obj/item/gun/energy.dm
@@ -1669,7 +1669,6 @@ TYPEINFO(/obj/item/gun/energy/wasp)
 	New()
 		set_current_projectile(new/datum/projectile/energy_bolt/signifer_tase)
 		projectiles = list(current_projectile,new/datum/projectile/laser/signifer_lethal)
-		..()
 		AddComponent(/datum/component/holdertargeting/fullauto, 1.5, 1.5, 1)
 		..()
 
--- a/code/obj/item/gun/energy.dm
+++ b/code/obj/item/gun/energy.dm
@@ -1670,15 +1670,18 @@ TYPEINFO(/obj/item/gun/energy/wasp)
 		set_current_projectile(new/datum/projectile/energy_bolt/signifer_tase)
 		projectiles = list(current_projectile,new/datum/projectile/laser/signifer_lethal)
 		..()
+		AddComponent(/datum/component/holdertargeting/fullauto, 1.5, 1.5, 1)
+		..()
 
 	update_icon()
 		..()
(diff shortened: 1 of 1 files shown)


CheffieGithub:
maybe reduce the firerate a bit

TheColorCyan:
yeah I will reduce the stun it does and it's firerate when I will be able to

Enakai:
I would prefer to avoid this as I had designed (sprited) the Signifer to be semi-automatic. Kyle, who did the code, I believe may also share the same sentiment, but I'll wait for his input.

TheColorCyan:
Should now need less PU for less stun power, reduces the spread and makes shots more delayed.

Kyle2143:
Yeah, like Enakai said, we didn't have full-auto in mind for this weapon design. We've spoken a bit about this in admin chat and I admit, the design choice here that led to the gameplay reality of "mash click to shoot fast enough to be good" is not great.

The original design of this fire setting for the weapon was something like, "a difficult to use weapon that required a lot of hits to connect in order to incapacitate someone". Sort of an energy weapon take on the original Clock-188 that the NTSecO had, but with less bleeding and damage (which admittedly wasn't a great design choice either).

I'd prefer changing the projectile to function some other way instead of making it a full-auto. We were discussing some ideas, such as some simple changes to make click-mashing not required or desirable like changing the click-shoot delay and power or such. But that could be temporary before a rework or not if it just works well...

TheColorCyan:
> I'd prefer changing the projectile to function some other way instead of making it a full-auto. We were discussing some ideas, such as some simple changes to make click-mashing not required or desirable like changing the click-shoot delay and power or such. But that could be temporary before a rework or not if it just works well...

Would it be better if I made it shoot 3 projectiles in one click sum of which would roughly equal to a taser shot?

TheColorCyan:
Since devs said that they dont really want it to be fullauto/burst and will be looking for other ways to rework the signifer i will be closing this PR.

PR #14113, opened 2023-05-14
VR Antagonist Fixes

Mister-Moriarty:
[Internal] [Bug]

## About The PR:
Fixes VR antagonists removing non-VR antagonist datums should a mind possess more than one datum of the same type, such as roundstart wizards using VR, opting to become a VR wizard, then being deprived of their real wizard antagonist status when they log out. This has been fixed by permitting `remove_antagonist()` to accept a reference to an antagonist datum as its `role` argument.
Additionally fixes pseudo antagonists attempting to remove their owner from client image groups that they are not registered to.

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -258,19 +258,30 @@ datum/mind
 				return TRUE
 		return FALSE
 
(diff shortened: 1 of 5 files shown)


TobleroneSwordfish (approved):
lgtm

PR #14114, opened 2023-05-15
Nadir "Unwrap" Fixbatch

Kubius:
## About the PR 

Mid-size tuneup to Nadir, primarily focused on getting rid of excessive reinforcement that keeps hazards a little too partitioned and can hinder construction or crime.

- Most internal windows that were previously crystal glass have been converted to simple reinforced glass. In some cases where the windows are protecting sensitive areas (such as the escape arm and department head offices), electrification has been added to maintain security.
- Internal R-walls protecting areas that are both less sensitive and unlikely to be breached in the normal course of events have largely been changed to standard walls.
- The Extraction Nexus' control room has been considerably overhauled, now being contained within the siphon chamber. This setup is less safe, but only for the operators (which isn't necessarily a bad thing); in exchange, it makes entering the Siphon vicinity from the control area quick and easy, and frees the entire Siphon operating radius for users without requiring any disassembly. A manufacturer is also included for convenience.

image

- MULE beacons have been introduced to the map for feature-completeness. While the map still has none at round start (and deliberately so), MULEs ordered by QM should now be able to operate and reach a few basic destinations.
- Replaced all varedited catwalks with the base variant, since they're now auto-tiling.
- Added a table between the chairs in the treatment room, in the space previously occupied by a NanoMed before the dispensary interlink update.

## Why's this needed? 

Improves Nadir's usability and feature completeness. Wall and window downgrades in particular should improve the viability of break-ins and destructive strategies to be more similar to other maps.

## Changelog 

```changelog
(u)Kubius
(+)Reduced Nadir's internal reinforcement, improved its extraction nexus, and gave it MULE beacons to accommodate purchased MULEs.
```


flappybatpal (approved):
Slightly concerned about removing medbay reinforcement considering how big a bomb target it is.

Kubius:
Robotics and Cloning have a significant gap between them, the department as a whole is highly partitioned, and the arguably most effective bombing target (medbay lobby) is about as effective as before. Re-reinforcement is definitely worth keeping in mind, but I think it's worth a shot to be more broad with the de-reinforcement and then selectively reintroduce it as needed.

PR #14121, opened 2023-05-16
auto cable and pipe repath

TheTyrant25:
[INTERNAL] [MAPPING] [QOL]
## About the PR
Okay so cable spawners was my first proper coding project back when i was little baby coder. It wasn't... great. And now that i'm getting into mapping I'm realising I should have just put it under auto. So I'm doing that now. This change applies to disposal pipe spawners too.
Changes include:
- `/obj/cablespawner` has been turned into `/obj/cable/auto`
- `/obj/disposalpipespawner` similarly is now `/obj/disposalpipe/auto`
- auto cables will now force override and make a node when they're under terminals or smes units.

P.S. i got hit with a case of the scope creep, so i'm also making it so the pipes don't delete themselves.
## Why's this needed?
Matches up with other auto connecting subtypes, such as walls. Much easier for mappers to find and therefore use. And is more consistent with other auto thingys.


TheTyrant25:
has been tested
image
Repathing has broken nothing that wasn't already broken

ZeWaka:
I'm going to (personally) say this is a non-starter unless you make a plug-in/render pass for spacemandmm that renders them correctly.

TheTyrant25:
okay that's fair

ZeWaka:
> 2. Auto cables quite often spawn multiple cables instead of just one,

Render pass can handle this

> 3. This PR is just repathing them, not uh, making them.

With the intent for people to use them, right? Otherwise there's no point having them.

TheTyrant25:
okay having gone through spacemandmm and strongdmm
it is tentatively doable
but i'd need this merged first so that i can mess with their repos and pull request there, which might take
a while

TheTyrant25:
it keeps staling so im just gonna request some reviews

PR #14123, opened 2023-05-16
Auto lattices

TheTyrant25:
[MAPPING] [INTERNAL] [QOL]
## About the PR
Now introducing auto lattices. Should make mapping these things out a lot nicer. One day palettes will be obsolete...
They auto connect to other auto lattices, and connect to regular lattices which "point" at them.

Other changes:
- All lattices have been moved to a new file, `code/obj/lattice.dm`
- Made a reference to lattice and catwalk spawners in the mapping guidelines.
- Regular lattices now have a dirmask variable, a bitmask of the dirs it points in. It'll be set up automatically most of the time. The `New()` can be called with either you setting the dirmask and it picking the correct icon, or picking the icon and it setting up the mask. It's flexible that way. The default, in order to accomodate existing lattices, is that the var edited icon take priority.
- Regular lattices now have `/obj/lattice/auto/directional-icons` subtypes, to make selecting those dirs easier. Palettes are probably favoured but i thought it'd be nice. Automatically sets the icon state and does nothing else.
- By using the `/obj/lattice/auto/wall_attaching` subtype, they'll connect to simulated and unsimulated walls too.
- The `/obj/lattice/auto/turf_attaching` subtype will similarly connect to anything that's not a space tile.

## Why's this needed?
It's nice to have. Consistency with other things like auto catwalks, and cables, and disposal pipes, and walls, and etc.
Unlike auto catwalks however you can still place them manually.

## Postscript
Please excuse the horrible, **horrible** code I had to add to the `/obj/lattice/New()`. This is because the sprites for lattices are stored in three separate dmi iconstates, which are:
"lattice", the general connecting in all four directions variant.
image
"lattice_dir", which contains all of the three way connections, but in an awful way. Plus other stuff.
image
and "lattice_dir_b", which contains all the corners and some other random stuff. They almost match the dirs too, but they don't.
image
If i fucked with the dmi file, every single map using lattices will have their icons break. So i will simply    not do that.


TheTyrant25:
Everything is now working except for the connecting to walls bit. Honestly can be left out if I can't fix it, but would be really nice to have.

TheTyrant25:
the walls now connect 😎

TheTyrant25:
image
all variants now work fully as intended.

flappybatpal (approved):
Comparing station and secret Z and I can't see any errors.
No real opinion on the code but I am side eyeing that giant New()

TheTyrant25:
well, that'd what happens when the dmi is arranged by ~~something very hurtful~~, can't do much about the random arrangement of the different icons.

TheTyrant25:
shouldnt affect any existing lattices, just adds a new subtype

Fungal-Infection:
badass

pali6:
> Please excuse the horrible, horrible code I had to add to the /obj/lattice/New(). This is because the sprites for lattices are stored in three separate dmi iconstates, which are:

I feel your pain. The April fools map upscaler code for this was also a fucking mess.

PR #14125, opened 2023-05-16
Updates and moves map files and data around

TheTyrant25:
[INTERNAL] [QOL]
## About the PR
Not really a refactor per se, but just puts the maps in rotation into the spotlight a bit better. Moves the non rotation maps into unused, properly. Also adds some more comments to the build.dm defines.

The arbitrary order is:
1. Special modes (construction, pod wars, event)
2. regular ship maps (cogmaps, donuts, kondaru)
3. ship maps (atlas and clarion)
4. ocean maps (oshan and nadir)
5. out of rotation maps

## Why's this needed?
It'd be nice. Cleaner code is a cleaner mind. Or something


TheTyrant25:
note that the reshuffling on things makes horrible diffs.

Issue #14068, opened 2023-05-12
Critter Gaunlet Blindness


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Critter gaunlet is now almost impossible due to the mimics giving you perma blindness in two hits.

### Steps to reproduce

Go to critter gaunlet
Get hit by two mimics
Go blind forever

### Expected Behavior

For critter gaunlet to be fun, and not make you go blind on the first round.

### Additional Information & Screenshots

Reported by: Sbmhawk
On server: Goonstation 1 Classic: Heisenbee
Active test merges: #13959 and #14038
Round log date: 2023-05-12-21-49
Reported on: 2023-05-12 22:40:12
Map: COGMAP

Sbmhawk:
Please give us occuline or VISORS or *something*

Sbmhawk:
Or even better, make the mimics in there not make you go blind

flappybatpal:
Needs a virtual type of critter without the mimic sting.

Sbmhawk:
Needs closing

PR #14127, opened 2023-05-16
Replaces critter gaunlet mimics with a subtype that does not sting

Sbmhawk:
## About the PR 

This adds a subtype of mimics, /mob/living/critter/mimic/virtual, that is identical to the parent, but cannot sting, and replaces the critter gauntlet mimics with it.

## Why's this needed? 

Going permablind in two stings in the critter gauntlet, and having 0 way to fix/mitigate this, is kinda dumb.

Fixes(?) #14068

### DIFF
--- a/code/datums/gauntlet/crittergauntlet.dm
+++ b/code/datums/gauntlet/crittergauntlet.dm
@@ -1041,7 +1041,7 @@ var/global/datum/arena/gauntletController/gauntlet_controller = new()
 		name = "Mimic"
 		point_cost = 1
 		count = 6
-		types = list(/mob/living/critter/mimic)
+		types = list(/mob/living/critter/mimic/virtual)
 
 	meaty
 		name = "Meat Thing"
--- a/code/mob/living/critter/mimic.dm
+++ b/code/mob/living/critter/mimic.dm
@@ -174,3 +174,12 @@
 		HH.can_range_attack = FALSE
 		var/datum/limb/small_critter/L = HH.limb
 		L.max_wclass = W_CLASS_SMALL
+
+/mob/living/critter/mimic/virtual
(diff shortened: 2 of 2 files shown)


amylizzle (changes requested):
A better way to do this would be to change `critter_ability_attack()` to do a null check on the abilities like `if(sting && !sting disabled... `, then just remove the ability from the subclass.

Airport code review go!

Github sucks on mobile

amylizzle:
Also maybe just add occuline autoinjectors to the round loot.

Sbmhawk:
I considered the occuline injectors, but mimics are so common in the gauntlet, I feel like it would not help much, especially with multiple people doing the gauntlet.

Sbmhawk:
I was under the impression that add_abilities was the thing that gives them the ability, and removing it from that list would get rid of the ability.

I assume you mean something like:

if(sting != NULL && !sting.disabled && sting.cooldowncheck())
			sting.handleCast(target)
			return TRUE
			
I'm not sure what you mean by removing the ability after the null check, as my understanding is that it is null because it was already removed.

amylizzle:
So there are a couple of important parts to this:
1. Yes, the ability list does tell the critter what abilities it gets, so removing the sting is what you want - that bit is correct
2. The AI for the mimic however assumes that the mimic has a sting ability always, so if you just remove it, the AI will crash. You don't notice it because the AI processing loop has a bunch of protections against that, and so those errors are only visible in the debug log.

So you've done it right in the latest commit. The only thing I'd change is the `if(string != null` to just if `if(!sting` because `null` is falsey. Also do the same to the other ability check.

Sbmhawk:
Wouldn't it be if(sting  
As we would only want the statement to pass if sting exists?

Sbmhawk:
Did testing with both, if(!sting does not work

amylizzle:
yeah you right, my bad

PR #14139, opened 2023-05-17
allows bread to roll a chance after enough teleporting to turn into a mimic

Skeletonman0:
[OBJECTS] [FEATURE]

## About the PR 

- bread loaves now track how many times they've been teleported and now have a variable to determine the threshold when they start rolling to become a mimic
    - as of writing this, the threshold is 100 teleports

- adds a signal (COMSIG_MOVABLE_TELEPORTED) for when a movable atom is teleported through do_teleport
adjusts some things to use do_teleport

- makes teleporting bread after hitting the threshold roll to become a mimic
    - prob( 0.01 )

mimics are annoying sometimes so hopefully i made this annoying enough to achieve

## Why's this needed? 
funny reference

## Changelog 

```changelog
(u)Skeletonman0
(*)Teleporting bread excessively now rarely can cause dangerous results

### DIFF
--- a/_std/defines/component_defines/component_defines_atom.dm
+++ b/_std/defines/component_defines/component_defines_atom.dm
@@ -71,6 +71,8 @@
 	#define COMSIG_MOVABLE_POST_RADIO_PACKET "mov_post_radio_packet"
 	/// when an atom hits something when being thrown (thrown_atom, hit_target, /datum/thrown_thing)
 	#define COMSIG_MOVABLE_HIT_THROWN "mov_hit_thrown"
+	/// when an AM is teleported by do_teleport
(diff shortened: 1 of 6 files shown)


TobleroneSwordfish:
I'm going to say no to the chance scaling over time, as being able to guaranteed spawn 100s of mimics is probably a very bad idea.

Fungal-Infection:
the madman actually did it

ThatFiveGuys:
> I'm going to say no to the chance scaling over time, as being able to guaranteed spawn 100s of mimics is probably a very bad idea.

Maybe just put a cap on it instead?

TobleroneSwordfish:
this counts as a speedmerge by goonstation standards

PR #14142, opened 2023-05-18
Fish

Fungal-Infection:
## About the PR 

Literally just allows fishing in drains because i think its silly you cant

## Why's this needed? 

i like fishe


Fungal-Infection:
god i love fish

pali6 (approved):
🎣

PR #14146, opened 2023-05-18
Resprites Sponge Capsules

TheColorCyan:
[GAME OBJECTS]

## About the PR
Adds new sprites to Syndie sponge capsules and normal ones
S on syndie one clearly stands for sponge capsules
capsules

## Why's this needed?
They needed a resprite


CheffieGithub:
can you show examples of the filled sprites with the overlay?

TheColorCyan:
they look like  this in game
sprites
this

flrsh (changes requested):
These look very flat to me. Usually we try to have items in a quasi-3/4 perspective. If you're going more for a bag look, peach ring bags may be a good example to refer to. If you're going more for a box look, try cereal boxes or flour bags.

TheColorCyan:
> These look very flat to me. Usually we try to have items in a quasi-3/4 perspective. If you're going more for a bag look, peach ring bags may be a good example to refer to. If you're going more for a box look, try cereal boxes or flour bags.

The old sprites and how they actually look IRL are actually flat and not a bag or a box so it barely makes any sense to me but I can try making it a box if you want
images (2)
Без названия (3)

TheColorCyan:
oh and on another note, when you try to put capsule back it doesn't let you with a message 'you try to put capsule back but it won't stay' which makes no sense if item was a bag/box

flrsh:
The issue is that the sprites look inconsistent with our style and giving them extra dimension would help. They're also too large. If you can think of a way to retain their blister pack aesthetic while making them fit our art style, that's fine. But otherwise, making them boxes might be easier.

TheColorCyan:
Yeah I can try
The issue with making them a box, as I already said before, would be it's interaction with a capsule after you remove it and try to put it back so it will probably have to be changed

ZeWaka (approved):

PR #14147, opened 2023-05-18
allows ice cubes to stop decomposition of corpses

Skeletonman0:
[OBJECTS] [BALANCE] [FEATURE]

## About the PR 
- allows ice cubes to halt decomposition of a corpse

- ice cubes now also take damage from the air if its warmer than their melting temperature variable and its not melting due to body heat
(melting temperature is set to 0 C for regular ice cubes instead of being 0 K like before)
  - the temperature used to cool the occupant of the cube is a seperate variable incase that should be left at 0 K
  - damage from air temp seemed to be too fast so i slapped a 50% chance on there

## Why's this needed? 
i remember one time i tried freezing a corpse until i could revive them
i realized: they still melted the ice cube even with no body heat
they also still rotted

## Changelog 

```changelog
(u)Skeletonman0
(*)Freezing people in ice and keeping the ice cold is now a valid way to delay corpses rotting.

### DIFF
--- a/code/mob/living/critter/bee.dm
+++ b/code/mob/living/critter/bee.dm
@@ -547,6 +547,8 @@
 			SPAWN(2 SECONDS)
 				var/obj/icecube/honeycube = new /obj/icecube(src)
 				MT.set_loc(honeycube)
+				honeycube.melttemp = T20C
+				honeycube.cooltemp = T20C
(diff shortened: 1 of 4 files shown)


Glamurio:
Very neat idea.

PR #14150, opened 2023-05-19
Aetherion Station

TheTyrant25:
[MAPPING]
## About the PR
Adds a new lowpop station designed to fill the same bracket as Atlas: below 30 players.
Aetherion station is a station designed to be small and compact, while still having maintenance shafts and therefore more criming opportunities, which Atlas somewhat lacks. So far its main features are:
- Centralised bridge
- Fairly high usage of the new glass floorings
- Chapel and Medbay both being connected to the morgue, and the morgue connected to the crematorium (no crematorium chute needed)
- Dual access chemlab, doubling as a pharmacy.
- Courtroom (specifically, the jury room) doubling as the northwest corridor.
- And more

## Why's this needed?
When population is below about 20, there's not many real options. Of course, you can select donut 2, clarion, oshan, kondaru or nadir too, but they're not designed for lowpop in the way atlas is. This is.

## Notes
This map at the moment is still under development. Things like the layout and even the name are still able to be changed. There is a forum post describing the process in greater detail [here](https://forum.ss13.co/showthread.php?tid=21191). The main reason I opened a pull request this early was to fix this bloody merge conflict in the only way I know how, which is through the merge conflict thing in a PR. Sorry.

## Changelog 
```changelog
(u)Tyrant
(*)Added a new lowpop map: Aetherion.
```

## Picture of Aetherion (as of 22nd May 2023, may be out of date)
image


Fungal-Infection:
i've not checked it out but from the map the armory door seems to be blocked mayhaps by a shelf, though that may just be a silly image thing

TheTyrant25:
> i've not checked it out but from the map the armory door seems to be blocked mayhaps by a shelf, though that may just be a silly image thing

Ah, yeah
originally it was in a different spot and i moved the entrance
will fix that, no worries 👍 thanks for pointing it out, I hadn't noticed it

ThatFiveGuys:
I'll give my thoughts on specific things before I give my thoughts on the map as a whole. I will try to be thorough here in order to give you a sense of where to improve. To my knowledge this is your first mapping project and thus naturally you will have lots of mistakes. Please do not take this with discouragement, you have shown a lot of eagerness with this project in the first place and I'm eager to see what you will do in the future. That being said, apologies for the wall of text that follows.

The workflow of engineering seems bad. For starters, to get to the engine, you either have to cross 12 tiles of space, or sneak around to take catwalks (see picture below). The former requires you to have a spacesuit on you (as there are no spacesuits nearby the former passage), or take space damage, and the latter route is unintuitive for new players as the map is done in such a way to communicate that going across 12 tiles of space is the intended way of traversal (the doors are right across from each other). Not to mention, moving canisters across space is a pain.

I am overall meh about excessive airgaps for unrelated reasons, but if one should be had 1. Make the gap as short as possible (ie the distance from the exit of engineering to the engine shouldn't be 12 tiles long) and 2. Have an airbridge, this way its convenient by default but retractable when needed.  This would make working on the engine much more convenient. You will definitely need to rearrange engineering for this, but make the "hallway" from engineering core to the engine way shorter.
image

Ranching has a few flaws. The ranches take up 3 tiles max, this means all your chickens will be in 3 tiles of space. This introduces lots of issues with management, merging the ranches would help loads here. Also why is there a UV lamp inside of a ranch with no hydroponic tiles nearby? UV lamps are something that's nice to reposition so putting it inside of fences seems like a bad idea.

This map seems like it uses way too many catwalks, this may seem like a nitpick (and it is), but with each of these points I've been trying to make wider points that I will go into more detail on at the next section. A good mapping principle is to try to make sure things work both in the context of gameplay, and in the context of setting. I think this fails in both regards. For gameplay, what purpose does being able to go to routing depots on? Why should there be catwalks there. It's important to note too, catwalks circumvent space damage. This means that you can leave the station from anywhere, and as long as you stay within one tile of the station, you're safe. This has massive implications for antags who decide to focus there efforts on space, as a secoff can go after them with far less risk then normal. 

From an in universe perspective, the catwalks don't make much sense either. Nanotrasen generally speaking, does not like to waste money. Would the routing depots and hull need to be serviced enough to justify dedicated catwalks just from getting to them? Why would you use catwalks for structural units when lattices suffice.

Here is where I go into the more wider general issues of this map. There are two main issues with this map, the first is more concrete, poor workflow, and the second is more abstract, messiness. 

When you map, you have to constantly think about how people will play in this map. Stuff like having to carry a canister across 11 tiles of space or putting a UV lamp within railings is not good for gameplay. This is a skill that takes time to develop, and something that needs to be actively nurtured. You have to constantly think about from where people will be going to and from. 

The second issue is a lot less constrained to specific areas and more of an issue with the map as a whole. The map feels messy and undefined. When I first saw this map, something rubbed me the wrong way about it and I think I finally understand why. A big part of a map is getting across the idea that you are in a space station made and lived in by people. This map does not feel like a map made by Nanotrasen, it feels like a map made by Tyrant25. 

For starters, why would a map have so many glass floors. Presumably glass floors are more expensive then regular floors, so why would they be there? Is it for morale, is it because nanotrasen really decided to be extravagant with this one? Why is there so much maintenance as well? The in universe purpose for maintenance is in the name, its a nice place to do repairs and run pipes through and do all sorts of backend stuff. Why would such a small station have a massive maintenance. If there are intended answers to these questions, the map does not convey them. Figuring out how your map will exist in universe will lead to a much more organic and full feeling to it. 

For example, what if the reason why there's so much maintenance and so much fancy glass is because this station was intended to be a station for more people. However after the initial structure got finished, they realized they wanted to have much less people on it, so they channeled all the remaining budget into the rest of the station and left the rest neglected. This is not the route you have to take with the map of course, but this is to give an example of a route. Having a clear picture of what your map is and why it exists would help a lot with the general feel of the map and lead to lots of interesting design emerging.

TheTyrant25:
I should have appended that this is an old image shouldn't i

ThatFiveGuys:
> Note on catwalks: I liked the idea of there being a walkable surface on the outer perimeter, but if it prevents space damage (which i didn't know about) then i'll definitely remove it, or create a subtype which does cause space damage (or just use lattices everywhere). 

Rest of this sounds good, though please dont create a subtype as that seems very unintuitive. Id reccomend just lattices if you want to go with a walkable surface on outer perimeter.

nefarious6th:
i'll leave the general map-building comments to others, and i'd love to come back and take a look at this/playtest once it's a little further along. i just wanted to say that this map reminds me of trunkmap in a good way, i love the observatory, and i like the pop size you're angling for here.

ZeWaka:
You do know we're not really looking for new maps at this time right?

TheTyrant25:
had an informal greenlight from some admins...

Glamurio:
Big fan of the map, think there's a few kinks to be ironed out about some small sizing issues, I think some of the maintenance tunnels are too wide, though that's not necessarily a bad thing and will allow for changes later. Additionally, it could maybe use a couple more random room placements.

Big fan of the pool area. Excited to see where it goes. And it doesn't get added, I think it can never harm to have a few maps on the backburner foor later.

flrsh (changes requested):
We're not really looking for new maps. An admin's offhanded comment should not be taken as actual support of your project.

TheTyrant25:
anyhoo the pr did what it was meant to: gather opinion and let me fix merge conflict easily. So i'll close it now, possibly reopen it later when it's finished

TheTyrant25:
#16571 has the new and updated pr

PR #14152, opened 2023-05-19
Quick Little Poster

Fungal-Infection:
## About the PR 

Adds a funky little poster about a space Opera
image

## Why's this needed? 

The funny jestr hat !!!


code/obj/decal/poster.dm
```diff
@@ -678,7 +678,8 @@
 					5 ;"contest-other4",
 					5 ;"contest-other5",
 					5 ;"contest-other6",
-					5 ;"contest-other7"
+					5 ;"contest-other7",
+					5 ;"contest-other8"
```
FlameArrow57, line 682: This is for contest posters I think. poster_rand may be the place?

Fungal-Infection: good point, ill change that a tad bit later

Fungal-Infection:
i made a mistake, and then immediately realized and had to fix it

Fungal-Infection:
how does the QUICKFIX have a fail oh my goodness hold on

Fungal-Infection:
it is now a bad icon operation aegh

Fungal-Infection:
i will never touch postercode again 😭

FlameArrow57:
wdym? Code seems fine, just adding to random_icon_states

Fungal-Infection:
ive got nO idea

FlameArrow57:
If it's because of the error that's showing up, that's something outside of this PR, not caused by it

PR #14156, opened 2023-05-19
Singularity Gravitational Lensing

amylizzle:
## About the PR 
Makes space bendy.
image

Essentially adds a displacement filter to most (not all) of the plane masters, and replaces the singulo's black orb with a small singularity and a large gravitational distortion.

Performance wise: I tested this against no filter, `COOL_PLANE_STUFF`, and the current approach, and none were obviously slower than the others, which is weird. Obviously the filter must necessarily be slower, but it doesn't seem to be a significant enough effect that I could measure it easily. I suggest we test merge and check for complaints.

## Why's this needed? 
It looks cool. Also the same effect could (and should) be applied to white holes, gravity arts, whatever else.

## Changelog 

```changelog
(u)Amylizzle
(*)The singularity has received a visual overhaul. Tell someone in #imcoder if it makes your framerate suck. Bendy space.
```


MitchTheUser:
👀

pali6:
I personally think that a more symmetric distortion based on the matric of a non-rotating black hole would look better. But very cool. I hope it doesn't destroy performance too much.

nefarious6th:
i mean, visually cool idea but seems like it makes finding the spot to hit with the sing buster even more complicated

pali6:
Maybe the size of the distortion should be exactly the size of the singularity itself. That way nothing is lost visually compared to the current visuals. Only pixels currently covered by the singularity sprite would instead reflect some other pixels in the world.

amylizzle:
Right now it's set up so the distortion starts a little beyond the danger as a warning, but the actual edge is when the distortion gets real intense. The singularity is a little slow to affect things as is though, so it's not a super hard edge.

Regarding the singulo buster though, I actually think it's easier with this. The center of the black hole is much more obvious, and thus easier to target.

amylizzle:
Reception seems positive - main feedback was that it shouldn't warp ghosts and that the distortion can make it tricky to work in, which seems more like a feature than a bug. It could also be less warpy on small small singulo and more warpy on big ones.

No mentions of reduced performance yet, continuing to monitor.

amylizzle:
I spent all day trying to get this to look better than it currently does, and was unsuccessful. Gonna make it not warp ghosts and call it a day.

flappybatpal:
It can be difficult to tell where the deadly area is in a huge singularity, it will also eventually start to cause the edge of the screen to bend into view and leave a black area but it didn't look too bad.

flappybatpal (approved):

amylizzle:
[Screencast from 22-05-23 22:09:57.webm](https://github.com/goonstation/goonstation/assets/3855802/f227b261-2a52-4eb3-a931-7995e78a6d1e)

pali6:
Is there any chance we could maybe get a toggle for these effects like these?
image
Could be in a subsequent PR if you don't wanna code that now.

pali6 (approved):

PR #14159, opened 2023-05-19
Changed "Gender" to "Sex" in Character Setup and a couple other places

gedrocht:
[ui] [qol]

## About the PR 
Grammatical Fix

## Why's this needed? 
The term "gender" is often used interchangeably with "sex," but they are not the same thing. Gender is a social construct, while sex is a biological reality. This change is to be more accurate and inclusive.

## Changelog 

```changelog
(u)gabringis
(+)changed "gender" to "sex" in a few places
```


gedrocht:
I'm not certain how to successfully run build_tgui.sh or if that's even required in this case. Can someone please fill me in?

pali6:
An easy way is to press Ctrl+Shift+\` and type in `tgui/bin/tgui`. That should do the trick. But also you will first need to pull changes from master.

flrsh (changes requested):
Imo if anything we should change all the instances of "sex" to "gender" or just remove it. "What sex your character was assigned at birth based on a cursory examination of their genitals area" is irrelevant to the game.

flrsh:
Oh, though, it seems like right now one of these instances also controls body type? In which case we should just rename that to "body type" with body types A and B (which also allows more body types in the future) when they exist (some mutants only have one).

MarkNais:
One-line summary of my thoughts: Close because the game has a young audience and also a free audience and swapping gender for sex, and keeping it a binary, might be more of an overall-annoyance than a benefit.

Similar to Flourish's latest comment: I think the game's gender representation could use an overhaul. As-is it only supports a binary male/female option which is "more like" a biological-sex representation (though it's not a true binary in reality). I can see why you were compelled to make this PR.

While players having greater accuracy of representation of characters can be fun and fulfilling: I don't think biological-sex as a fluff-attribute is something I want added to the game; I strongly do not want biological-sex to become a crunchy-mechanic of the game.

IIRC the character option only affects pronouns for humans (and probably other mobs). An idea I pitched long long ago was to make the PDA's display in-round customizable preferred pronouns, that could also be preset in character customization. Thematically other characters would see the PDA's screen while it's attached to your overall's ID slot. This could potentially remove the need for a sex option, and could maybe remove the need for a gender option. (Doesn't provide such flexibility to mobs / aliens, a short-coming)

Keeping in mind the wide audience that reaches our servers: having a sex option could lead to a lot of moderation scenarios that could be avoided by keeping it worded as gender. I don't feel the need to elaborate or exemplify this point.

gedrocht:
I can't say I'm not disappointed

PR #14163, opened 2023-05-19
Gives robotic critters the same electric immunity as silicons

CheffieGithub:
## About the PR 
Gives robotic critters the same immunity to electricity that borgs have.

## Why's this needed? 
Silicon flavour critters deserve parity with normal silicons.

## Changelog 

```changelog
(u)Cheffie
(+)Robotic critters are now immune to electric shocks
```


TobleroneSwordfish:
I think I'd probably prefer adding a flash effect or something to one or more arcfiend abilities? Being immune to electricity is a pretty core part of borgs.

CheffieGithub:
Ok I can do that ill make another PR, is it still fine to keep the damage for when they have exposed wires / brain?

PR #14175, opened 2023-05-20
Prefab correctness and modernising

CheffieGithub:
## About the PR 
Changes all Z5 prefabs to use variable turfs instead of space / normal asteroid walls floors.
Updates prefabs with non perspective walls/lights/doors with perspective stuff.
Changes thindow and grilles to wingrille and grilles.
Adds catwalk supports to catwalks that were missing them.
Replaces fake object broken lights with broken real lights (adds a few broken subtypes for this)

Makes a few minor changes to some prefabs so they look more natural on the mining map. (No more sharp edges on the medical shuttle and sleeper shuttle)

## Why's this needed? 
Prefab correctness good old sprites and such not good.


CheffieGithub:
Note for assets/maps/prefabs/prefab_grill.dmm a dev will need to re-add the grillnasium area to it since its in the secret submodule it was removed.

pali6 (approved):

PR #14176, opened 2023-05-20
fix broken zoldorf appearance stealing

Skeletonman0:
[OBJECTS] [BUG]
## About the PR 
at some point, the feature where zoldorfs copied the appearance of the person who turned into them broke

this ended up being due to the alpha filter being passed an image which somehow caused it to turn the entire overlay invisible

also fixes some code which is shared by both wanted posters and zoldorf.
the images will now respect the dir you want them to have

## Why's this needed? 
bug bad, visible zoldorf heads good
image


code/obj/posters.dm
```diff
@@ -216,8 +216,9 @@ var/global/icon/wanted_poster_unknown = icon('icons/obj/decals/posters.dmi', "wa
 		undies.Blend(src.bioHolder.mobAppearance.u_color ? src.bioHolder.mobAppearance.u_color : "#FFFFFF", ICON_MULTIPLY)
 		return_icon.Blend(undies, ICON_OVERLAY)
 		undies = null
-
-	var/icon/comp = getFlatIcon(src, direction ? direction : null)
+	var/image/I = src // force the image to have the direction we want
```
TobleroneSwordfish, line 219: ???

PR #14177, opened 2023-05-20
Gives viper venom a stupid and funny recipe

colossusqw:
[chemistry][feature]
## About the PR 
This PR adds a recipe for viper venom, it includes 'thematically sound' chemicals, causes a rattling sound, and produces a very low amount of viper venom.

## Why's this needed? 
Since viper venom got it's stun and histamine production removed, it's nowhere near as debilitating and deadly as it used to be, and being a close to 0.4 depletion chem with no skin pen that doesn't exactly stack with any of the other chems, this feels like a feasable addition, expecially considering the recipe itself is not an easy one.
And i also just find it funny. Feel free to make this a secret recipe too, since it pretty much fits that whole category of dumb recipe.

```changelog
(u)Colossusqw
(+)Added a very dumb recipe for viper venom.
```


Tarmunora:
This might have been interesting as a secret recipe, but as a normal reaction, I don't think we're interested.
I might think about adding a recipe in secrets myself at some point though.

PR #14179, opened 2023-05-20
Add ability to attach items to legs

Glamurio:
[Objects] [Feature]

## About the PR 
This is a preliminary pull request to add the fundamental ability to attach (almost) any item to legs, similar to how they are able to be attached to arms. Unlike arm items, however, leg items are currently unable to be used, and simply provide a new leg sprite, paired with a slight movement penalty per leg.

image image image

The default sprite draws from a Crowbar, as it seemed like the best overall item choice for a generic sprite, making it a semi-peg-leg. Later on I'd like to implement more unique sprites for various items, as well as some fun bonus interactions of mechanics, though I wanted to break this up into multiple PRs for digestibility.

I've added a new NO_LEG_ATTACH variable that currently acts synonymous to NO_ARM_ATTACH, but will lay the groundwork later to specify which items should maybe be allowed on arms, but not legs, for whatever reasons.

## Why's this needed? 
To have more consistent mechanics present in the game. There's no logical reason for people to be able to attach items to arms but not legs. Furthermore, it will add another way for characters to express themselves, and may even lay the foundations for new fun mechanics and interactions.

I've also made a thread on the forums discussing this, and the overall reception was positive.
https://forum.ss13.co/showthread.php?tid=21168

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Added the ability to attach items to legs.
```


nefarious6th:
love the idea of a crowbar being a sort of in-a-fix knock-off cyborg leg. cool idea

M-Earthfire:
Kinda cannot wait to replace the clowns arm and legs with bike horns or pizza slices

TobleroneSwordfish:
This is a cool idea, but comes with some issues:
- You'd either need to make a unique "attached as leg" sprite for each item or have every item look the same (item arms get around this by re-using in-hand sprites very jankily).
- Presumably this is intended to allow for the item legs to have effects in the future, but since there's no way to decide how an item leg should work (item arms can just call attackby) you'd need to decide how item legs work on a case by case basis, leading to most items not working the way players think they will.

An alternative might be to designate certain items as being able to be attached as legs and provide unique sprites/effects for each of them? That way you can't walk around on a bullet casing leg that looks like a crowbar and functions identically to a normal leg, but you can replace your leg with a mop which splashes water when you walk and looks like a mop.

Glamurio:
> An alternative might be to designate certain items as being able to be attached as legs and provide unique sprites/effects for each of them? That way you can't walk around on a bullet casing leg that looks like a crowbar and functions identically to a normal leg, but you can replace your leg with a mop which splashes water when you walk and looks like a mop.

This is something I realized fairly early on as well, the way that item arms jankily re-use in-hand sprites can't easily be replicated. I've been playing around with the idea of simply using the in-hand sprite and rotating them to make them work as legs, but that still makes a good 80% of all items look weird or jank, with the few exceptions being swords, mops, etc.

Making certain items attachable with unique sprites/effects was planned, and I wanted to simply have a generic system accepted first before I start sketching out more content, but if the idea of simply having a placeholder crowbar isn't good enough and ideally we'd only like to use unique items, that's certainly also doable. Realistically, I'd imagine, people would only want to attach very specific thematic items to their legs anyway, such as crowbars, swords, mops, chainsaws or perhaps arms (which is a whole other can of worms).

I'll sketch up and code some unique variants.

PR #14183, opened 2023-05-21
Categorize lollipops as candy

Glamurio:
[Objects] [QoL]
## About the PR 
This recategorizes lollipops from 
`/obj/item/reagent_containers/food/snacks/lollipop` to `/obj/item/reagent_containers/food/snacks/candy/lollipop`.

It also changes all references from the old to the new definition.

## Why's this needed? 
Lollipops are candy, they should be classified as such. This also adds sugar to lollipops, which they should also have.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Lollipops are now candy, and thus contain sugar.
```


TheTyrant25 (approved):
looks good to me, as long as you didn't miss any spots (which, if you did find-replace like I think you did, shouldn't be a problem) then it's good to go.

CaioFrancisco:
by the way, it's not as much of a "categorization" when it comes to types, than "inheriting what i like about my parent type". but the type change doesn't seem to affect much of the item, so i'm fine with it.

PR #14188, opened 2023-05-21
minor turf.dm refactor

TheTyrant25:
[CODE-QUALITY]
## About the PR
Was going to work on some turf stuff and noticed it was a little messy. It is not in my nature to ignore such things so i atomised out a tiny refactor pr for you lot.
## Why's this needed?
Good Code Is Good. Make Code Gooder wherever you go, and eventually one day it will all be Good.

### DIFF
--- a/code/turf/turf.dm
+++ b/code/turf/turf.dm
@@ -22,7 +22,7 @@
 	proc/burn_down()
 		return
 
-	/// Properties for open tiles (/floor)
+	// Properties for open tiles (/floor)
 	#define _UNSIM_TURF_GAS_DEF(GAS, ...) var/GAS = 0;
 	APPLY_TO_GASES(_UNSIM_TURF_GAS_DEF)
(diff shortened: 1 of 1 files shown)


code/turf/turf.dm
```diff
@@ -2,57 +2,57 @@
...
+		can_replace_with_stuff = TRUE  //Shitty dumb hack bullshit
 #endif
-		allows_vehicles = 0
+		allows_vehicles = FALSE
 
 	proc/burn_down()
 		return
 
-		//Properties for open tiles (/floor)
+	/// Properties for open tiles (/floor)
```
TobleroneSwordfish, line 25:
```suggestion
	// Properties for open tiles (/floor)
```
I think this expands into multiple properties, and the above doc comment would only apply to the first one

TobleroneSwordfish (changes requested):

PR #14197, opened 2023-05-21
Adds lots of factions and stuff

CheffieGithub:
## About the PR 
Adds 9 new factions and assigns them to various things.

Notable changes:
Jobs and antag datums can be assigned a faction.
All botany related critters, botanists and the rancher now share a BOTANY faction.
Clowns and clown spiders share a CLOWN faction.
Drones and gunborgs are SYNDICATE though this won't do anything until they get AI.
All syndicate related jobs are SYNDICATE and therefore not attacked by drones. (once they are mobs and not traitors ect)
Podwars roles are NANOTRASEN and SYNDICATE respectively.
Things spawned from void portal events are part of the DERELICT faction as are abandoned bots.
Things spawned from sponge capules are part of the SPONGE faction.
Things spawned from wraith portals, wraiths, and wraith critters are part of the WRAITH faction.

Fixes sponge capsule, wraith and disaster critters killing eachother and not the crew.

## Why's this needed? 
Factions are good


nefarious6th:
potentially out of scope and depending on mobification of transposed particle fields if possible, but this would hypothetically allow for fixing disaster round transposed fields killing other disaster round mobs? Or is that trickier since particle fields can appear in other circumstances and their fa tion has to be set when they're created based on round type etc?

CheffieGithub:
yes the vortexes that spawn those mobs can set the faction on creation

TobleroneSwordfish (changes requested):

code/mob/living/critter/small_animal.dm
```diff
@@ -236,7 +236,6 @@ ABSTRACT_TYPE(/mob/living/critter/small_animal)
 
 /mob/living/critter/small_animal/mouse/mad
 	ai_type = /datum/aiHolder/mouse/mad
-	faction = MOB_AI_FACTION_WRAITH
```
TobleroneSwordfish, line 239: Won't this make the mad mice from rat dens attack plague rats?

code/datums/jobs.dm
```diff
@@ -904,6 +910,8 @@ ABSTRACT_TYPE(/datum/job/civilian)
 	slot_ears = list(/obj/item/device/radio/headset/civilian)
 	items_in_backpack = list(/obj/item/device/camera_viewer/ranch,/obj/item/storage/box/knitting)
 
+	faction = FACTION_BOTANY
```
TobleroneSwordfish, line 913: Is rancher botany? Currently wasps and such only care about specifically botanists
image

TobleroneSwordfish (changes requested):

code/datums/jobs.dm
CheffieGithub: Ok removed the faction from rancher and added to the beekeeper instead since they get wasp grenades

PR #14199, opened 2023-05-22
Allow shuriken pouches to be held in pocket

FlameArrow57:
[GAME OBJECTS][QoL]

## About the PR 
This PR gives shuriken pouches correct pouch functionality, and also removes incorrect type they're on

Slight balance change with slots they can hold

## Why's this needed? 
Allows them to be held in pocket


TobleroneSwordfish (changes requested):
may as well just make them `/obj/item/storage/pouch` subtypes

PR #14200, opened 2023-05-22
Fancy ladder stuff

Skeletonman0:
[SPRITES] [FEATURE] [QOL]
## About the PR 
- adds some new ladder sprites i made using the palette from the space diner walls
- shifts some stuff around to allow ladders to be extradimensional, using a subtype of the extradimensional_storage component
- as a result of the ladders being slightly larger than the old sprite, i had to move some around to make them still about as hidden as they were
  - one of the ladders on cog2 has been replaced with an "embed" subtype ladder which acts as a enterable/exitable gap in the wall
  - the closet infront of said ladder has been pixel shifted up to hide it as well as it hid the ladder that existed before

image

image

- extra dimensional ladders exist for potential admin gimmicks, but can be created the same way the lockers can (if you can manage it)
  - a void overlay is added onto the ladder to visually show that the ladder is spooky now

## Why's this needed? 
new sprites + extradimensional ladders good


TobleroneSwordfish:
Neat! What happens if you destroy a wall with a secret entrance in it? Not sure if negative matter ladders make as much sense as lockers but 🤷

code/datums/components/extradimensional_storage.dm
```diff
@@ -56,13 +78,40 @@ TYPEINFO(/datum/component/extradimensional_storage)
...
-	UnregisterSignal(parent, COMSIG_ATOM_ENTERED)
+// subtype of the component that instead handles ladders
+/datum/component/extradimensional_storage/ladder
+
+/datum/component/extradimensional_storage/ladder/Initialize(width=9, height=9, region_init_proc=null)
+	. = ..()
+	if(!istype(parent, /obj/ladder))
+		return COMPONENT_INCOMPATIBLE
+	exit = get_turf(src.parent)
+	. = ..()
```
pali6, line 94: you definitely don't want to be calling parent twice 🐸

code/datums/components/extradimensional_storage.dm
```diff
@@ -56,13 +78,40 @@ TYPEINFO(/datum/component/extradimensional_storage)
...
+	UnregisterSignal(src.parent, COMSIG_PARENT_PRE_DISPOSING)
+	. = ..()
 
-/datum/component/extradimensional_storage/UnregisterFromParent()
-	var/obj/storage/parent = src.parent
-	src.region.move_movables_to(parent.loc)
-	region.clean_up(/turf/space, /turf/space)
-	qdel(region)
-	UnregisterSignal(parent, COMSIG_ATOM_ENTERED)
+// subtype of the component that instead handles ladders
```
pali6, line 86:
```suggestion
/// subtype of the component that instead handles ladders
```

code/datums/components/extradimensional_storage.dm
```diff
@@ -56,13 +78,40 @@ TYPEINFO(/datum/component/extradimensional_storage)
...
+	var/obj/ladder/ladder = src.parent
+	var/image/I = image(ladder.icon,ladder,"[ladder.icon_state]-extra")
+	ladder.UpdateOverlays(I,"extradim")
+
+	RegisterSignal(src.parent, COMSIG_ATTACKHAND, PROC_REF(on_entered))
+	RegisterSignal(src.parent, COMSIG_PARENT_PRE_DISPOSING, PROC_REF(on_disposing))
+
+/datum/component/extradimensional_storage/ladder/on_entered(atom/movable/thing,mob/user)
+	var/obj/ladder/ladder = src.parent
+	if (istype(ladder,/obj/ladder/embed))
```
pali6, line 105:
```suggestion
	if (istype(ladder, /obj/ladder/embed))
```

code/datums/components/extradimensional_storage.dm
```diff
@@ -10,43 +11,64 @@ TYPEINFO(/datum/component/extradimensional_storage)
...
+	qdel(src)
+
+/datum/component/extradimensional_storage/UnregisterFromParent()
+	var/obj/parent = src.parent
+	src.region.move_movables_to((exit || parent))
+	region.clean_up(/turf/space, /turf/space)
+	qdel(region)
+	. = ..()
+
+// subtype of the component that handles storage
```
pali6, line 59:
```suggestion
/// subtype of the component that handles storage
```

code/datums/components/extradimensional_storage.dm
```diff
@@ -1,6 +1,7 @@
 /datum/component/extradimensional_storage
```
pali6, line 1: Should probably be marked as abstract, right? Or am I missing something? At a glance it feels like no one should be instantiating this type.

code/datums/components/extradimensional_storage.dm
```diff
@@ -56,13 +78,40 @@ TYPEINFO(/datum/component/extradimensional_storage)
...
+	if (istype(ladder,/obj/ladder/embed))
+		boutput(user, "You enter the gap in the wall.")
+	else
+		boutput(user, "You climb [ladder.icon_state == "ladder" ? "down" : "up"] the ladder.")
+	user.set_loc(region.turf_at(rand(3, region.width - 2), rand(3, region.height - 2)))
+
+/datum/component/extradimensional_storage/ladder/UnregisterFromParent()
+	var/obj/ladder/ladder = src.parent
+	ladder.UpdateOverlays(null,"extradim")
+	ladder.broken = FALSE
```
pali6, line 114: `broken = FALSE`? Why here? What does it do?

code/turf/turf.dm
```diff
@@ -851,6 +856,8 @@ proc/generate_space_color()
 		if (map_settings.auto_windows)
 			for (var/obj/window/auto/W in orange(1))
 				W.UpdateIcon()
+	for (var/obj/ladder/embed/L in orange(1))
+		L.UpdateIcon()
```
pali6, line 860: Why are these updates only in these specific rarely used procs and not in base `ReplaceWith`? Shouldn't the ladder update when the turf is updated in any way?

(also as a sidenote, we really should do this stuff in some better way, signals sent on turf updates; but out of the scope of this PR)

code/z_adventurezones/lavamoon.dm
```diff
@@ -1211,52 +1211,132 @@ var/global/iomoon_blowout_state = 0 //0: Hasn't occurred, 1: Moon is irradiated
...
-		if (user.stat || user.getStatusDuration("weakened") || BOUNDS_DIST(user, src) > 0)
-			return
-		src.climb(user)
+/obj/ladder/embed/update_icon()
+	. = ..()
+	if (!src.hidden)
+		var/turf/T = get_step(src,NORTH)
+		if (!istype(T,/turf/simulated/wall) && !istype(T,/turf/unsimulated/wall))
+			src.alpha = 0 // if there's no wall above us, hide in a way that the turf can update us
+			src.mouse_opacity = FALSE
```
pali6, line 1249: if you want a thing to be both unclickable and invisible the easier way to do this is to set invisibility
`APPLY_ATOM_PROPERTY(src, PROP_MOB_INVISIBILITY, "hidden", INVIS_ALWAYS)`

code/z_adventurezones/lavamoon.dm
```diff
@@ -1211,52 +1211,132 @@ var/global/iomoon_blowout_state = 0 //0: Hasn't occurred, 1: Moon is irradiated
...
+/obj/ladder/embed/ex_act(severity,last_touched)
+	if (src.hidden)
+		src.hidden = FALSE
+		src.UpdateIcon()
+	. = ..(severity, last_touched)
+
+/obj/ladder/embed/extradimensional/New()
+	..()
+	src.setMaterial(getMaterial("negativematter"))
+/obj/ladder/extradimensional/New()
```
pali6, line 1267:
```suggestion

/obj/ladder/extradimensional/New()
```

code/datums/components/extradimensional_storage.dm
pali6: Ah. It disables normal ladder behaviour. I feel like the var name could be better in that case.

pali6 (changes requested):

code/datums/components/extradimensional_storage.dm
Skeletonman0: renamed to unclimbable, hopefully that's a good enough name

code/z_adventurezones/lavamoon.dm
Skeletonman0: unfortunately INVIS_ALWAYS and INVIS_ALWAYS_ISH ended up being TOO invisible and the ladder never ended up in orange(1)

pali6: :melterfrog:

pali6 (approved):
lgtm I think

PR #14207, opened 2023-05-22
Mobifies Townguards

CheffieGithub:
## About the PR 
Mobifies townguards, hopefully makes them immune to arrows because of that fancy armour they have.

## Why's this needed? 
HALT! object critters need to go


code/obj/item/sponge_capsule.dm
```diff
@@ -31,7 +31,7 @@
 					/mob/living/critter/brullbar,
 					/obj/critter/bat/buff,
 					/mob/living/critter/spider/ice,
-					/obj/critter/townguard/passive,
+					/mob/living/critter/townguard,
```
pali6, line 34: Is the change to non-passive intentional?

code/mob/living/critter/adventure.dm
```diff
@@ -425,6 +427,117 @@
...
+		src.set_a_intent(INTENT_HARM)
+		src.active_hand = 2
+		return ..() // Stab them
+
+	death(var/gibbed)
+		src.can_lie = FALSE
+		..()
+
+	proc/HALT()
+		if(!ON_COOLDOWN(src, "say_HALT!", 3 SECONDS))
```
pali6, line 527: Could the cooldown length be a var on the townguard? For evil admin reasons.
HAL-HAL-HAL-HALT!

code/obj/item/sponge_capsule.dm
CheffieGithub: Yeah I changed this one to agressive since the other spawns are.

PR #14217, opened 2023-05-23
Adds a missing security scanner to Atlas bridge

DanaTarSUS:
[QOL] [MAPPING]
## About the PR 
Title, one of the two doors leading to the bridge from HoP desk had a security scanner while the other did not. This adds a scanner to the other door so that no evildoer can hide from the law or whatever.

## Why's this needed? 
Symmetry Good!!!!!!! :D 
Oversight Bad!!!!!!! >:(


Gannets (approved):
Sure

PR #14219, opened 2023-05-24
Adds Wanderer AI to most bird critters 

DanaTarSUS:
modified:   code/mob/living/critter/small_animal.dm
	modified:   goonstation.dme

## About the PR 
Adds basic AI to most birds mobs that fall under the parent /mob/living/critter/small_animal/bird, in addition, penguins will now retaliate when hit in the same way that most mice currently do (will hit their attacker once before returning to normal.)

## Why's this needed? 

Some of bird the mobs categorized under the parent /mob/living/critter/small_animal/bird had no existing AI and would simply stand still and not retaliate when hit. This pr allows some birds such as cassowaries and penguins to now wander around instead of simply standing still and doing nothing.


DanaTarSUS:
Apologies for the borderline insane rant in the original pr comment, I just have zero confidence that this was remotely coded correctly. On the slim chance it is coded correctly, that's cool. If not, then please understand that this was not done incorrectly out of malice.

code/mob/living/critter/small_animal.dm
```diff
@@ -1240,8 +1240,16 @@ var/list/mob_bird_species = list("smallowl" = /mob/living/critter/small_animal/b
...
+	health_burn = 30
 	species = "penguin"
-
+	ai_type = /datum/aiHolder/penguin
+	is_npc = TRUE
+	ai_retaliates = TRUE
+	ai_retaliate_patience = 0 //retaliate when hit immediately
+	ai_retaliate_persistence = RETALIATE_ONCE //but just hit back once
+	var/attack_damage = 5
+	var/use_custom_color = FALSE
```
frawhst, line 1252: These vars aren't actually used so you don't need them here. Looks like they do things in mouse code.
```suggestion
```

frawhst:
> I'm never touching mob code ever again. I just wanted to replace the cog2 owlery with a room full of snow and penguins in a seperate pr and instead I'm now down a rabbit hole of code that I barely understand and want to be free from.

This is just the consistent experience of working on this codebase. Your only mistake is thinking it's a you problem.
Code looks good.

code/mob/living/critter/small_animal.dm
frawhst: Wait `attack_damage` doesn't do anything on mice either. That var does nothing on like 3 different critter types 😢

amylizzle:
Yeah that's about right. You don't need to create an AI holder for just wandering though, there's a few premade ones for just this kinda thing. Just look for `/datum/aiholder/wander ` (not exact name, I'm afk)

amylizzle:
Oh! That's a bug on the aiholder code. I'll PR a fix for it. Maybe leave yours as it is for now then, and I'll update it once I fix the wanderer holder.

CheffieGithub:
yeah the wandering thing not retaliating is probably because its old, when making default aiHolders I just moved it

CheffieGithub:
id say it would be better to just add the Ai holder on the small animal parent instead

CheffieGithub:
wanderer should work now

DanaTarSUS:
Finished up the changes just now. I ended up moving ai_type and is_npc to /mob/living/critter/small_animal/bird instead of /mob/living/critter/small_animal/bird/penguin. This has also added wanderer AI to birds that didn't really have any sort of AI before such as the cassowary.

I also deleted the separate penguin AI file in favor of using /datum/aiHolder/wanderer.

ZeWaka:
i don't see how this PR is related to penguins? probably need to update the title and description

DanaTarSUS:
> i don't see how this PR is related to penguins? probably need to update the title and description

Woops, I've now updated the title and description to better match the contents of the pr.

ZeWaka (approved):

PR #14238, opened 2023-05-25
Martian fixes and updates

CheffieGithub:
## About the PR 
Replaces broken martian.dmi.
Replaces martian walls in Z5 and Oshan with simulated turfs, which recreate orginal martian wall behaviour, since the tiles around them are simmed.
Updates oshan martian ship to be the same as debris.
Replaces martian computer icon varedits with real subtypes.
Makes oldwalls autoconnect to martian walls and vice versa for martain ship.

## Why's this needed? 
Bug bad. varedits bad. consistency good.


CheffieGithub:
not sure why the diff is freaking out with missing tiles but the sprites are there

PR #14243, opened 2023-05-26
Redoes the minigun from the ground up

Sir-Mora:
[C-FEATURE] [C-SPRITES] [A-GAME-OBJECTS]

## About the PR 
The minigun has been retooled and recreated
-New sprites and inhands + firing animation
-renamed to "Alpha Hydrae minigun"
-provides a bigger slowdown but no slowdown after firing
-ONLY full auto + a spool up time
-slightly less spread
-more ammo (200 from 100)
-two handed

Demonstration video

https://github.com/goonstation/goonstation/assets/84998542/77676b17-c6a7-4c4a-8ca1-4e0b872427e0

## Why's this needed? 
The current minigun sprites and code are very old, using the gun in game doesn't feel as powerful as a minigun should rightfully feel. These changes do make the minigun much more powerful however, and some balance changes could likely be needed should the minigun ever show up in game under normal circumstances.


flappybatpal:
I don't think this should eject casings, the light machine gun doesn't.

In my opinion ideally the solution would be to have a reduced casing capability for these weapons that only fires create half as many casings.

Sir-Mora:
> I don't think this should eject casings, the light machine gun doesn't.
> 
> In my opinion ideally the solution would be to have a reduced casing capability for these weapons that only fires create half as many casings.

Could have it eject casings then the casings delete themselves soon after, seems like it would create a good balance

Sord213:
Self deleting casings just seems....meh. Doesn't make much sense and since it's an admin only weapon it's not like you need the casings for forensics. I'd rather have it not drop 200+ tiny useless objects on the ground and handwave the excuse away as "they're caseless rounds"

otherwise sprite real cool and fullauto cool

Sir-Mora:
> Self deleting casings just seems....meh. Doesn't make much sense and since it's an admin only weapon it's not like you need the casings for forensics. I'd rather have it not drop 200+ tiny useless objects on the ground and handwave the excuse away as "they're caseless rounds"
> 
> otherwise sprite real cool and fullauto cool

sounds fair, no more casings are a go 👍

PR #14245, opened 2023-05-26
An implementation of pet carriers.

DisturbHerb:
[OBJECTS][FEAT]

## About the PR 

This implementation was made without any access provided to myself to the rancher chicken carriers, so the pet carrier's behaviour may differ from the chicken carriers. Such differences are intentional to attempt to balance them against each other.

This PR adds pet carriers to the game. Icons by Sunkiisu/Azzy/Nico.

image

Pet carriers are a bulky one-handed item that can hold (by default) a single instance of any `mob/living/citter/small_animal` inside. A small animal can be clicked with the carrier in the active hand to attempt to trap it inside, while using it in-hand or using the "Release occupant" context action will attempt to remove an animal. The first animal out of the list (if the capacity is greater than 1) will be removed upon doing so. Coaxing animals in and out of the carrier has a 2 second actionbar.

The following demo uses the old version of the icons, but functionally remains identical.

https://github.com/goonstation/goonstation/assets/66253095/4e147d7e-bec5-4b80-9e63-4ccce13c38b8

Any trapped mob can attempt to escape with the `*flip` emote or by resisting. Doing so will cause 6 damage to the carrier out of a maximum of 30 health by default. All occupants are released when the health reaches zero, and then the health of the door resets.

Pet carriers currently spawn in the Captain, HoP, Research Director, and Medical Director lockers. These are the only Heads whose pets are mob critters. Apologies to all Rocko and Sylvester fans.

### Technical details

The given mobs that a carrier is allowed to hold is determined by `var/mob/allowed_mobs` which should be overridden by any child types of the carrier. The maximum capacity is determined by `var/carrier_max_capacity` which can also be overridden. Whether a mob can break out or not is determined by `var/can_break_out`, and if `FALSE`, any occupant cannot escape the carrier on their own.

The carrier has support for spawning with an animal already inside by overriding `var/mob/default_mob` with a valid path.

The icon for the carrier is constructed using a base and two `/obj/dummy`s are layered on top. These dummy objects are held in the carrier's `vis_contents` and each of them hold the mobs and the door of the carrier. The reason why the `vis_contents` of `vis_contents_proxy` hold the mobs rather than the carrier doing so directly is so that an alpha mask can prevent parts of the occupants from rendering outside of the inside of the carrier.

For any mobs inside, the atmosphere they'll attempt to breathe from is taken from the current `turf` that the carrier occupies, even if the carrier is being held in someone's hand.

There's also an admin crimes version that can hold anyone. Have fun with it.

## Why's this needed? 

Greater convenience for transporting command pets and the like, without having to corral them into crates. Easier to take them on walks. I think the idea's just cute.

## Changelog 

```changelog
(u)DisturbHerb + Sunkiisu
(*)Adds pet carriers to the game. These bulky carriers can hold a single small animal.
```


code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,319 @@
...
+ *
+ * There is support for spawning with an animal already inside by overriding var/mob/default_mob with a valid path.
+ *
+ * The icon for the carrier is constructed using a base where an overlay of the front of the carrier and two /obj/dummys are layered on top.
+ * These dummy objects are held in the carrier's vis_contents and each of them hold the mobs and the door of the carrier. The reason why the
+ * vis_contents of vis_contents_proxy hold the mobs rather than the carrier doing so directly is so that an alpha mask can prevent parts of the
+ * occupants from rendering outside of the inside of the carrier.
+ *
+ * For any mobs inside, the atmosphere they'll attempt to breathe from is taken from the current turf that the carrier occupies, even if the carrier
+ * is being held in someone's hand.
```
FlameArrow57, line 32: Might suggest removing most of these comments, a lot of it is implementation details which would need to be rewritten when changing code so, and someone can read the doc comments anyways for understanding usage

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,319 @@
...
+			if (!ispath(src.default_mob, src.allowed_mobs))
+				return
+			var/mob/spawned_mob = new src.default_mob
+			if (spawned_mob)
+				src.add_mob(spawned_mob)
+
+	disposing()
+		if (length(src.carrier_occupants))
+			for (var/occupant in src.carrier_occupants)
+				src.eject_mob(occupant)
```
FlameArrow57, line 111:
```suggestion
		for (var/occupant in src.carrier_occupants)
			src.eject_mob(occupant)
```

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,319 @@
...
+			return TRUE
+		else if (!mob_to_check.l_hand)
+			return TRUE
+		else return FALSE
+
+	/// Calls src.AttackSelf(user) with a context action. Yeah, I know.
+	verb/release_occupant_verb(mob/user)
+		set name = "Release occupant"
+		set category = "Local"
+		set src in oview(1)
```
FlameArrow57, line 247: Feel like oview might confuse some people who are standing right over it, it also makes the verb not able to be used when holding it in hand

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,319 @@
...
+	var/grate_closed_icon_state = "grate-closed"
+
+	// Carrier item state names.
+	var/carrier_open_item_state = "carrier-open"
+	var/carrier_closed_item_state = "carrier-closed"
+
+	/// Carrier-related (grate_proxy, vis_contents_proxy) vis_flags.
+	var/carrier_vis_flags = VIS_INHERIT_ID | VIS_INHERIT_LAYER | VIS_INHERIT_PLANE
+	/// Mob-specific vis_flags.
+	var/mob_vis_flags = VIS_INHERIT_LAYER | VIS_INHERIT_PLANE
```
FlameArrow57, line 72: Lot of vars here that aren't ever changed or are only referenced in one spot, I would think they could just be removed, or made constants

FlameArrow57:
Oh also, for every case where you're checking for if the user has a free hand and giving a message if they don't, that message will always be given if the person isn't a human, which may be misleading

code/obj/item/pet_carrier.dm
DisturbHerb: This is all DMDOC, which is documentation.

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,319 @@
...
+			src.eject_mob(user)
+			src.visible_message("<span class='alert'>[user] kicks the door of [src] open and crawls right out!</span>")
+			return
+		boutput(user, "<span class='alert'>Maybe this door could give out if you put up some more effort!</span>")
+
+	throw_impact(atom/hit_atom, datum/thrown_thing/thr)
+		..()
+		if (src.carrier_occupants)
+			for (var/mob/occupant in src.carrier_occupants)
+				occupant.throw_impact(hit_atom, thr)
```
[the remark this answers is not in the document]
DisturbHerb: Uh, I don't think so. This is so that stuns and window damage are accurately included upon imapct.

code/obj/item/pet_carrier.dm
DisturbHerb: That last part's somewhat intentional, the verb's available for people who want to put the carrier down and subsequently open it up. If it's in your hand, you can already just attack_self(). As for oview, that's what a lot of other local verbs on larger non-item objects use, so that's what I've got.

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,319 @@
...
+		var/turf/current_turf = get_turf(src)
+		. = current_turf.return_air()
+
+	mob_flip_inside(mob/user)
+		..(user)
+
+		if (!src.can_break_out)
+			boutput(src, "<span class='alert'>It's no use! You can't leave [src]!</span>")
+			return
+		if (prob(BREAK_OUT_PROB))
```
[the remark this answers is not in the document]
DisturbHerb: I still think it should be probabilistic. You're not destroying the door, you're just kicking it open.

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,319 @@
...
+
+	// There has to be a better way of checking that the other hand is empty. Please.
+	proc/return_other_hand_empty(mob/living/carbon/human/mob_to_check)
+		if (!ishuman(mob_to_check))
+			return FALSE
+		if (mob_to_check.hand && !mob_to_check.r_hand)
+			return TRUE
+		else if (!mob_to_check.l_hand)
+			return TRUE
+		else return FALSE
```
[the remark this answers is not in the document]
DisturbHerb: It can't swap hands, that's not intended behaviour. It's just to check that the off-hand is empty and return TRUE or FALSE.

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,321 @@
...
+		src.icon = icon
+		src.icon_state = icon_state
+		src.action = carrier_action
+
+	onStart()
+		if (!ismob(owner))
+			interrupt(INTERRUPT_ALWAYS)
+			return
+		src.mob_owner = owner
+		if (BOUNDS_DIST(mob_owner, target) > 0 || !target || !mob_owner || !src.carrier || (src.action == TRAP_MOB && mob_owner.equipped() != carrier) || !carrier.find_empty_hand(mob_owner))
```
TobleroneSwordfish, line 292: I know action bars tend to be copy pasted to hell, but please just make this a proc

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,321 @@
...
+/obj/item/pet_carrier
+	name = "pet carrier"
+	desc = "A surprisingly roomy carrier for transporting small animals."
+	icon = 'icons/obj/items/pet_carrier.dmi'
+	icon_state = "carrier-full"
+	item_state = "carrier-open"
+	w_class = W_CLASS_BULKY
+
+	/// Please override this in child types to specify what can actually fit in.
+	var/mob/allowed_mob_type = /mob/living/critter/small_animal
```
TobleroneSwordfish, line 25:
```suggestion
	var/allowed_mob_type = /mob/living/critter/small_animal
```
I think this is less confusing?

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,319 @@
...
+	w_class = W_CLASS_BULKY
+
+	/// Please override this in child types to specify what can actually fit in.
+	var/mob/allowed_mobs = /mob/living/critter/small_animal
+	/// If FALSE, an occupant cannot escape the carrier on their own.
+	var/can_break_out = TRUE
+	/// How many mobs can fit inside the crate. Usually not overridden by anything, this is to let the system be permissive for var-editing.
+	var/carrier_max_capacity = DEFAULT_MAX_CAPACITY
+	/// If not null, the pet carrier will spawn with one of this mob on New().
+	var/mob/default_mob = null
```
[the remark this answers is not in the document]
TobleroneSwordfish:
```suggestion
	/// Type path, If not null, the pet carrier will spawn with one of this mob on New().
	var/default_mob_type = null
```

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,321 @@
...
+	name = "pet carrier (ADMIN CRIMES EDITION)"
+	desc = "A surprisingly roomy carrier for transporting living things. All of them."
+	allowed_mob_type = /mob
+	can_break_out = FALSE
+	carrier_max_capacity = INFINITY
+	door_health_max = INFINITY
+
+/// Pertains to actions executed by the pet carrier.
+/datum/action/bar/icon/pet_carrier
+	duration = 2 SECONDS
```
TobleroneSwordfish, line 269: would be nice if this was a var on the pet carrier, for advanced criming :)

code/obj/item/pet_carrier.dm
TobleroneSwordfish: This verb doesn't seem to work? I'm guessing `AttackSelf` requires you to actually be holding it or something

TobleroneSwordfish (changes requested):
Overall the code looks good, would be nice if there was some way to tell what's inside the carrier while someone is holding it, sprites would probably be a mess so maybe examine text at least?

TobleroneSwordfish (approved):

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,331 @@
...
+			return
+		boutput(user, "<span class='alert'>You try to bust open the door of [src]!</span>")
+		src.take_door_damage(src.damage_per_resist)
+
+	throw_impact(atom/hit_atom, datum/thrown_thing/thr)
+		..()
+		if (src.carrier_occupants)
+			animate_storage_thump(src)
+		for (var/mob/occupant in src.carrier_occupants)
+			occupant.throw_impact(hit_atom, thr)
```
frawhst, line 178: This janks out with disposal chutes, since those move you inside them when you impact them- phases the animals out of the carrier into the chute and doesn't update the carrier properly.
image

Cleanest way to fix (along with other critter-teleported-out bugs) is to add an `Exited()` override that checks if the exiting thing is in `carrier_occupants`, and then calls `eject_mob` on it if it is. (Won't cause any `loc` fuckery since you used `MOVE_OUT_TO_TURF_SAFE` >:) )

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,331 @@
...
+	/// Called when a given mob/user releases an mob after an actionbar.
+	proc/release_mob(mob/mob_to_release, mob/user)
+		if (mob_to_release)
+			src.eject_mob(mob_to_release)
+			user.update_inhands()
+			return
+		boutput(user, "<span class='alert'>Unable to release anyone from [src]!</span>")
+
+	proc/attempt_removal(mob/user)
+		if (!src.find_empty_hand(user))
```
frawhst, line 200: Does this mean that one armed people can never remove anything from the carrier?
If so, I'm fine removing it. I don't think people needing a free hand to do that is very relevant balance-wise.

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,331 @@
...
+		if (!can_act(user))
+			return
+
+		src.attempt_removal(user)
+
+/obj/item/pet_carrier/admin_crimes
+	name = "pet carrier (ADMIN CRIMES EDITION)"
+	desc = "A surprisingly roomy carrier for transporting living things. All of them."
+	allowed_mob_type = /mob
+	actionbar_duration = 1 SECOND
```
frawhst, line 267:
```suggestion
	actionbar_duration = 0
```
cmon now

frawhst (commented):
Cool feature, I enjoy.
Feels a bit odd adding these to command lockers, since those are already pretty bloated with stuff, but considering command pets it makes sense (and I cant think of a better place to put them).
Maybe make them orderable through cargo as well? Shouldn't be too hard to make a simply supply pack for them.

code/obj/item/pet_carrier.dm
```diff
@@ -0,0 +1,320 @@
...
+			occupant.throw_impact(hit_atom, thr)
+		if (length(src.carrier_occupants))
+			src.take_door_damage(src.damage_per_resist * length(src.carrier_occupants))
+
+	Exited(Obj, newloc)
+		if (Obj in src.carrier_occupants)
+			src.eject_mob(Obj)
+			src.visible_message("<span class='alert'>[Obj] bursts out of [src]!</span>")
+			return
+		..()
```
frawhst, line 184:
```suggestion
		. = ..()
		if (Obj in src.carrier_occupants)
			src.eject_mob(Obj)
			src.visible_message("<span class='alert'>[Obj] bursts out of [src]!</span>")
```
Should pretty much always call the parent, and I believe the return value of `Exited` holds some meaning (even if we rarely use it)

frawhst (approved):
Otherwise lgtm

Kyle2143:
I don't know about this, it looks alright, but I don't like how it means you can invisibly transport a pet. And so easily too. It makes it convinient to remove the pets from sight, which is bad IMO.

DisturbHerb:
> I don't know about this, it looks alright, but I don't like how it means you can invisibly transport a pet. And so easily too. It makes it convinient to remove the pets from sight, which is bad IMO.

Increasing the availability of it by putting it in the supply pack list does make this a more significant problem. This was initially restricted to a few head of staff lockers, and I wouldn't consider it a great loss if it was removed from Cargo.

PR #14264, opened 2023-05-27
Add ProDoc icon indicator for those with cyber organs

FlameArrow57:
[MEDICAL][BALANCE][QoL]

## About the PR 
This PR adds a purple ProDoc icon for those with cyber organs in them.

With everything:
image

Also makes it so that scanners without an organ upgrade can still detect if someone has a cyber organ for parity.

## Why's this needed? 
Right now it's nearly impossible to tell if someone has cyber organs in them, the only way is by clicking them with a health analyzer or they are jittery from certain emagged cyber organs, which there are other sources of being jittery too. It makes it really hard for Security to deal with those using emagged cyber organs.

This makes it more clear when someone has an augmented organ, and also makes it easier for Medical staff to assist Security when an antagonist is using cyber organs.

## Changelog 

```changelog
(u)FlameArrow57
(*)ProDoc goggles now show a purple indicator next to health icons for those with cyber organs of any form.
(+)Health analyzers without an organ scan upgrade now show if patients have a cyber organ, but not which ones.
```


CheffieGithub:
Health icons getting crowded

pali6 (approved):

ZeWaka (approved):

PR #14270, opened 2023-05-27
Partial materials refactor

CheffieGithub:
## About the PR 
I needed material stuff for a prefab, ended up making furniture use the same material set call instead of 5 different implentations across all the things. Now for objects its set for everything on New() on the base object type.

Fixes wooden tables and the like giving you steel sheets probably.

Test merge this since it touches a lot of material stuff, I tested it myself but can't be sure nothing broke.

## Why's this needed? 
Consistency is good. Easier to make new material stuff.

### DIFF
--- a/code/obj/stool.dm
+++ b/code/obj/stool.dm
@@ -26,6 +26,8 @@
 	throwforce = 10
 	pressure_resistance = 3*ONE_ATMOSPHERE
 	layer = STORAGE_LAYER //dumb
+	default_material = "steel"
+	uses_material_appearance = FALSE
 	var/allow_unbuckle = 1
 	var/mob/living/buckled_guy = null
 	var/deconstructable = 1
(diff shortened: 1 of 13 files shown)


pali6:
Good PR at a glance. Will review once it passes checks.

code/obj/stool.dm
```diff
@@ -874,6 +876,23 @@
...
+/* -------------------- Material Chairs ------------------ */
+/* ======================================================= */
+
+/obj/stool/chair/material
+	name = "material chair"
+	desc = "A chair made from a material"
+	uses_material_appearance = TRUE
+	comfort_value = 4
+
+	mauxite
```
ZeWaka, line 889: Can you absolutely path this?

ZeWaka (approved):

code/obj/stool.dm
```diff
@@ -874,6 +876,23 @@
...
+
+/obj/stool/chair/material
+	name = "material chair"
+	desc = "A chair made from a material"
+	uses_material_appearance = TRUE
+	mat_changename = TRUE
+
+/obj/stool/chair/material/mauxite
+	name = "chair"
+	icon_state = "chair$$mauxite"
```
pali6, line 891: hmm Is this necessary, I reckon the setMaterial code will replace it with this icon state automatically

code/obj/table.dm
```diff
@@ -532,6 +535,22 @@ TYPEINFO_NEW(/obj/table/scrap)
...
 		auto = 1
 
+TYPEINFO(/obj/table/mauxite)
+TYPEINFO_NEW(/obj/table/mauxite)
+	. = ..()
+	smooth_list = typecacheof(/obj/table/mauxite/auto)
+/obj/table/mauxite
+	name = "table"
+	icon = 'icons/obj/furniture/table.dmi'
+	icon_state = "0$$mauxite"
```
pali6, line 545: ditto

code/obj/table.dm
```diff
@@ -532,6 +535,22 @@ TYPEINFO_NEW(/obj/table/scrap)
...
+	desc = "A table made of dense mauxite"
```
pali6, line 546: Won't the description already get changed by the mat to add the mauxite bit? (i.e. isn't this kinda redunant and possibly false if the table gets transmuted later?)

pali6 (approved):

code/obj/stool.dm
CheffieGithub: for map editor sprites

PR #14283, opened 2023-05-29
Mobifies blobmen

CheffieGithub:
## About the PR 
Changes the blobman critter to a mob.

## Why's this needed? 
less object critter = good


CheffieGithub:
will need secret changes

PR #14291, opened 2023-05-30
makes solipsizine not perma deaf you

colossusqw:
[lavel][science][bug]

## About the PR 
Solipsizine started somehow making people just deaf forever, this fixes it(hopefully)[actually not just hopefully yeah this is what fixes it i've done more testing].

## Why's this needed? 
Because a chem shouldn't make you deaf for longer than the max round time with no cure whatsoever.


frawhst (approved):
lgtm

PR #14299, opened 2023-05-31
Adds printing press to chapel on all rotation maps

TobleroneSwordfish:
## About the PR 
Also adds a nearby paper bin where there wasn't one already.

## Why's this needed? 
Printing presses are cool but currently only exist in cargo crates, they also seem like something the chaplain would have.

## Changelog 

```changelog
(u)LeahTheTech
(*)Added a printing press in or near the chapel on all rotation maps.
```

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -17011,6 +17011,10 @@
 	},
 /turf/simulated/floor/plating/random,
 /area/station/maintenance/southwest)
+"chL" = (
+/obj/item/paper_bin,
+/turf/simulated/floor/plating/random,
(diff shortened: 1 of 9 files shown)


CaioFrancisco:
Not sure about the Chapel. I feel like it would fit more in the book room that most maps on-rotation already have. If it doesn't have one, it's probably too small for the printing press anyway.

TobleroneSwordfish:
Some of them are actually in the book room because it made more sense for them to fit there and it's right next to the chapel anyway. Also I think only a couple of maps have a book room?

PR #14307, opened 2023-05-31
Mobifies Zombies

CheffieGithub:
## About the PR 
Converts object zombies into mob critters.
Seperates the zombie bite ability from the critter ability.
Critters have the same stamina and movement slowdown as human zombies and have slightly increased health over objects.
Stopped zombie arms from infecting when not an infectious zombie trace and stops the disorientate if the holder is not human.

Zombies follow pretty much the same behaviour as before:
If they eat brains they attack you a bit and then try eat your brain.
If they don't eat brains they just attack you.
The wrestler zombie tries to grab and piledriver / throw you sometimes.
The rad zombie retains its ability to give you rads changing its affect to its ability attack.

## Why's this needed? 
Object critter bad. Human Zombies should have a distinict ability.


code/datums/abilities/critter/zombify.dm
```diff
@@ -1,68 +1,12 @@
...
 	name = "Zombify"
 	desc = "After a short delay, instantly convert a human into a zombie."
 	icon_state = "critter_bite"
 	cooldown = 0
 	var/actual_cooldown = 200
-	targeted = 1
-	target_anything = 1
+	disabled = FALSE
+	targeted = TRUE
+	target_anything = TRUE
```
frawhst, line 9:
```suggestion
```
Don't see why this should be able to target anything if it only works on humans

code/mob/living/critter/zombie.dm
```diff
@@ -0,0 +1,243 @@
...
+		HH.icon_state = "handr"
+		HH.limb_name = "right zombie arm"
+
+	setup_healths()
+		add_hh_flesh(src.health_brute, src.health_brute_vuln)
+		add_hh_flesh_burn(src.health_burn, src.health_brute_vuln)
+
+	valid_target(mob/living/C)
+		if (istype(C, /mob/living/critter/zombie)) return FALSE
+		if (iszombie(C)) return FALSE
```
frawhst, line 66: imo the critter check should just be rolled into the `iszombie` macro

code/mob/living/critter/zombie.dm
```diff
@@ -0,0 +1,243 @@
...
+			if (EATS_BRAINS)
+				infect = src.abilityHolder.getAbility(/datum/targetable/critter/zombify)
+			if (HUMAN_INFECTION)
+				infect = src.abilityHolder.getAbility(/datum/targetable/zombie/infect)
+		if (!infect?.disabled && infect.cooldowncheck())
+			infect.handleCast(target)
+			return TRUE
+
+	death(var/gibbed)
+		if (istype(src, /mob/living/critter/zombie/biosuit)) return //special death
```
frawhst, line 86: This typecheck is a bit gross- maybe put the biosuit-ghost-spawn stuff in `gib()` (and just don't call the parent). Since zombies always gib when they die anyways.

code/mob/living/critter/zombie.dm
```diff
@@ -0,0 +1,243 @@
...
+	name = "Lost Cosmonaut"
+	desc = "Soviet presence near NT stations is rarely overt. For good reasons, as this fellow probably learned too late.  Seriously, where is his face? Grody."
+	icon = 'icons/misc/meatland.dmi'
+	icon_state = "sovmeat"
+	health_brute = 15
+	health_brute_vuln = 0.6
+	health_burn = 15
+	health_burn_vuln = 1
+
+	//playsound(src.loc, 'sound/misc/meatmonaut1.ogg', 50, 0)
```
frawhst, line 239:
```suggestion
```
this isn't even in a proc!!

code/modules/antagonists/zombie/abilities/infect.dm
```diff
@@ -0,0 +1,89 @@
+/datum/targetable/zombie/infect
+	name = "Infect"
+	desc = "After a short delay, infect a human. If they are damaged enough or dead this will convert them instantly."
+	icon = 'icons/mob/critter_ui.dmi'
+	icon_state = "critter_bite"
+	cooldown = 0
+	var/actual_cooldown = 200
```
frawhst, line 7: Should document why this exists (i assume so the action doesn't go on cooldown if the action doesn't complete?)

code/modules/antagonists/zombie/abilities/infect.dm
```diff
@@ -0,0 +1,89 @@
...
+			return 1
+		var/mob/living/carbon/human/H = target
+		if (istype(H.mutantrace, /datum/mutantrace/zombie))
+			boutput(holder.owner, "<span class='alert'>You can't infect another zombie!</span>")
+			return 1
+		actions.start(new/datum/action/bar/icon/infect_ability(target, src), holder.owner)
+		return 0
+
+/datum/action/bar/icon/infect_ability
+	duration = 40
```
frawhst, line 43:
```suggestion
	duration = 4 SECONDS
```

code/modules/antagonists/zombie/abilities/infect.dm
frawhst, line 7:
```suggestion
	var/actual_cooldown = 20 SECONDS
```

code/modules/antagonists/zombie/abilities/infect.dm
```diff
@@ -0,0 +1,89 @@
...
+			if (target.ghost?.mind && !target.mind.get_player()?.dnr) // if they have dnr set don't bother shoving them back in their body (Shamelessly ripped from SR code. Fight me.)
+				target.ghost.show_text("<span class='alert'><B>You feel yourself being dragged out of the afterlife!</B></span>")
+				target.ghost.mind.transfer_to(target)
+		logTheThing(LOG_COMBAT, ownerMob, "zombifies [constructTarget(target,"combat")].")
+		playsound(ownerMob, 'sound/impact_sounds/Flesh_Crush_1.ogg', 50, 0)
+		ownerMob.visible_message("<span class='alert'><B>[ownerMob ] successfully infected [target]!</B></span>", 1)
+		ownerMob.health = ownerMob.max_health
+		target.TakeDamageAccountArmor("head", 30, 0, 0, DAMAGE_CRUSH)
+		target.changeStatus("stunned", 4 SECONDS)
+		target.contract_disease(/datum/ailment/disease/necrotic_degeneration/can_infect_more, null, null, 1) // path, name, strain, bypass resist
```
frawhst, line 87: Doesn't this just make them into human zombies? Seems like it would make sense to have critter zombies make more critter zombies. Also concerns about people finding zombies just to get killed by them so they can infect the station- should make the non-infectious at least.

maps/z2.dmm
```diff
@@ -46085,20 +46014,6 @@
...
-	atkcarbon = 0;
-	atksilicon = 0;
-	butcherable = 0;
-	can_revive = 0;
-	density = 0;
-	health = 0;
-	icon_state = "seczombie-dead";
-	layer = 2;
-	name = "Gutted Corpse"
-	},
```
frawhst, line 46101: The existing code is gross as hell but these should be replaced with something, rather than straight removed- maybe just make a subtype of `/obj/decal/fakeobjects/skeleton` which has the dead zombie icon?

CheffieGithub: Was planning to replace them with corpse spawners at a later time

code/modules/antagonists/zombie/abilities/infect.dm
CheffieGithub: this infect is the human zombie verison the critter one creates critter zombies

CheffieGithub: /datum/targetable/critter/zombify vs /datum/targetable/zombie/infect

code/modules/antagonists/zombie/abilities/infect.dm
CheffieGithub: This was just copied over from before but yeah thats the use since the ability uses an action bar, theres probably a better way and this isn't the only place var is used its also used on most actions with action bars

code/modules/antagonists/zombie/abilities/infect.dm
frawhst: Ahh my mistake.

CheffieGithub:
Checks seem unrelated to this PR

CheffieGithub:
well what I said about the checks was true then

CheffieGithub:
doesn't seem to be skipping checks now either

TobleroneSwordfish (changes requested):
again, as with skeletons there's stuff that relies on the `friends` var, might be worth just moving it to `critter` and integrating it into the base critter targeting

PR #14309, opened 2023-05-31
Sound Synthesizer - Control Unit Text Alignment

Coll6:
Fixes text alignment
Removes need for user variable

## About the PR 

Removed the user check to to determine if a component mainly the sound synthesizer and device frame was inside the Mech Cabinet or held by a mob namely the user calling the signal, However this affects all components that calls the componentSay proc. Changed how the text location is determined and attached. These changes fixed a misalignment of text when the control unit was used with the sound synthesizer. These changes also consequently added further functionality to the device frame allowing text to be placed even if the device is in some kind of storage container an AI mob and even inside a person.

## Why's this needed? 
When the control unit was used in any part of a mechanic chain there would be null user data. This breaks a check and consequently caused the text to display over the owner of whoever placed the component. This removed the need to check for a user as this should not be necessary it is mech after all. This issue could potentially show up in any future components that may not send user information so this brings more consistency to mechcomps.


code/modules/mechanics/MechanicMadness.dm
```diff
@@ -502,11 +501,24 @@
...
-			maptext = make_chat_maptext(src.owner, "[string]", "color: #FFBF00;", alpha = 255)
+		var/loc = null //Location used for center of all_hearers scan "Probably where you want your text attached to."
+
+		if(src.stored?.linked_item && !src.storage) //If the sound synth is attached inside something.
+			if(istype_exact(src.stored?.linked_item, /obj/item/storage/mechanics/housing_large)) //Mech Cabinet check
+				loc = src.loc
+
+			else if(istype_exact(src.stored?.linked_item, /obj/item/storage/mechanics/housing_handheld)) //Device Frame check
+				if(!obj_loc_chain(src)) //If not stored anywhere usually in your hands or on the ground
+					loc = src.loc
```
pali6, line 512: This condition will never be true because obj_loc_chain returns a list with is "truthy" in byond even if empty.

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -502,11 +501,24 @@
...
+
+				else //Handles rest of cases where it is stored somewhere whether on person in a container or some combination
+					var/list/atom/movable/loc_chain = obj_loc_chain(src)
+					loc = loc_chain[length(loc_chain)] //location of stop most container or possibly a mob.
```
pali6, line 516: I feel like you couldd remove the other cases for the housings and always just call this, it should handle everything correctly unless I'm missing something. Could you try if that works?

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -502,11 +501,24 @@
 	proc/componentSay(var/string)
 		string = trim(sanitize(html_encode(string)))
 		var/maptext = null
-		var/mob/user = usr
-		if (src_exists_inside_user_or_user_storage && !src.storage)
-			maptext = make_chat_maptext(src.owner, "[string]", "color: #FFBF00;", alpha = 255)
+		var/loc = null //Location used for center of all_hearers scan "Probably where you want your text attached to."
```
pali6, line 504: Naming the var `loc` isn't necessarily the best idea as it shaows the eixsting var `loc` on the type. Probably better call it something like `maptext_loc`.

code/modules/mechanics/MechanicMadness.dm
Coll6: I'll have to check for sure on the cabinet but it does not work in some cases as it will report null so I forced a check just in case.

code/modules/mechanics/MechanicMadness.dm
Coll6: Was there initially I believe I will go ahead and fix that.

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -502,16 +501,22 @@
...
+
+		maptext = make_chat_maptext(maptext_loc, "[string]", "color: #FFBF00;", alpha = 255)
+
+		for(var/mob/O in all_hearers(7, maptext_loc))
 			O.show_message("<span class='game radio' style='color: #FFBF00;'><span class='name'>[src]</span><b> [bicon(src)] [pick("squawks",  \
-			"beeps", "boops", "says", "screeches")], </b> <span class='message'>\"[string]\"</span></span>",1)
-			O.show_message(assoc_maptext = maptext)
-		playsound(src.loc, 'sound/machines/reprog.ogg', 45, 2, pitch = 1.4)
+			"beeps", "boops", "says", "screeches")], </b> <span class='message'>\"[string]\"</span></span>",1) //Places text in the radio
+			O.show_message(assoc_maptext = maptext) //Places text in world
```
pali6, line 518:
```suggestion
			"beeps", "boops", "says", "screeches")], </b> <span class='message'>\"[string]\"</span></span>",1,//Places text in the radio
				assoc_maptext = maptext) //Places text in world
```

PR #14310, opened 2023-05-31
Resprites Reagent Heater/Cooler

TheColorCyan:
[GAME OBJECTS]
## About the PR 
Title. Attempts a reagent heater/cooler resprites and makes it look more like an oven and a fridge merged together instead of oven and urinal.
chem grill
^fixed version
directionals were specifically requested in resprite request.
heatergif
real and true
If you have any suggestions i can listen.

## Why's this needed? 
It looked a lot like urinal and generally confused new players. This does resemble a heater/cooler more.


pali6:
I was about to suggest blue / red light when the device is hearing cooling but then I checked the sprites and you already did that. Good job!

Vaskritaya:
The open doors on the back (north-facing) sprite look pretty weird to me. I feel like they should be higher up. Currently, it looks like there's a strange amount of space between the doors and the top of the heater that doesn't exist with the other directional sprites.

TheColorCyan:
changed some sprites, should be okay now

TobleroneSwordfish (changes requested):
This looks great for science, but it could do with a generic version for the bar too as a purple striped object would stick out there.

Gannets:
Agree with the changes and otherwise approve of this.

TobleroneSwordfish (approved):
Looks good, but you missed Atlas

TheColorCyan:
ough

PR #14322, opened 2023-06-01
Lighting Overlay Fixes And Organisation

### LINKED ISSUE #6409: Autowall lighting overlays don't match up with sprites
### Description

Walls are supposed to have lighting overlays that match the shape of the wall. But with the hole-filling wall sprites this is often no longer the case which leads to annoying visual glitches with darker / brighter walls.

### Steps to reproduce

1. Build a wall which uses the autowall sprites, only some connection shapes reproduce this bug (basically you need wall connections which render differently after whichever PR added the hole-filling autowalls, for example an isolated wall will do).
2. Grab a Robust Light source and move around.
3. See visual glitches.

### Expected Behavior

Lighting overlay matches the wall sprite.

image
Here it can be seen on the left side of the sprite. A sharp divide between the brighter and the purplier part.
Here's how the overlay itself looks for this wall (clearly it doesn't match the shape):
image
Another example:
image

### LINKED ISSUE #10130: light_overlay.dmi missing states to support 8-way tiled walls properly
### Labels

[BUG] (well, oversight but that's not a tag)

### Description

icons/effects/light_overlay.dmi only has the icon states to support 4-way autowalls, leading any of the more complex wall junctions 

### Steps to reproduce

N/A - you can see it most easily for yourself by replacing some wall graphics with a flat colour though.

### Expected Behavior

A full set of lighting overlays

image
Here's an example of the cog1 computer core where I replaced the graphics for supernorn walls with light grey. None of the turfs in that 2 wide section (including walls under the grilles) or  those diagonally connecting to the thindows up top have functioning overlays, so the lighting gets applied uniformly on them.

Mister-Moriarty:
[Internal] [Bug] [Code Quality]

## About The PR:
Touches up the existing lighting overlays and adds additional overlays to accommodate for eight-directional walls.
Separates the lighting overlays into distinct files, alongside codeside changes to permit this.

The lighting overlays, with the alpha values clamped to 255 for legibility. The RGBA channels of each pixel represent the influence that the lights on the parent tile, the east tile, the north tile, and the northeast tile have on that pixel on parent tile respectively, somewhat akin to a normal map.
image

Fixes #6409 and fixes #10130.

<details>
<summary> Python Code</summary>

```
import matplotlib.pyplot as plt
import numpy as np

array_width = 32
array_height = 32

def create_array(upperleft, upperright, lowerleft, lowerright):
    array = np.linspace(
        np.linspace(lowerleft, lowerright, array_width), 
        np.linspace(upperleft, upperright, array_width),
        array_height, dtype = int)
    
    return array[:, :, None]

r = create_array(0, 0, 255, 0)
g = create_array(0, 0, 0, 255)
b = create_array(255, 0, 0, 0)
a = create_array(1, 255, 1, 1)

image = np.concatenate([r, g, b, a], axis=2)

plt.imshow(image, origin = "lower")
plt.axis("off")
plt.savefig("output.png", transparent = True)
plt.show()
```
Note that `output.png` will require to be cropped due to transparent margins.

</details>


ZeWaka (changes requested):
Could you include the python code in the initial post in a dropdown? A la:

\<details>
\<summary>Code\</summary>
fsdfdsfsdf
\</details>

Kapu1178:
i love you

pali6 (approved):
Excellent thank you <3

PR #14328, opened 2023-06-01
Unsimulated wooden wall fix

### LINKED ISSUE #14233: Wizard den walls look like steel
### Labels

[BUG]

### Description

They do

### Steps to reproduce

1. Go to the wizard den
2. Observe walls

### Expected Behavior

I expected wood to be wood!

image

CheffieGithub:
## About the PR 
Actually fixes #14233.

## Why's this needed? 
I forgot those walls were unsimulated.


CheffieGithub:
before they used sprites which had specific overlays and now they don't have them so it uses the default ones. But it shouldn't be connecting with an overlay at all :screm:

PR #14333, opened 2023-06-01
Add plasmatoid trait

Glamurio:
[secret] [traits] [feature]
## About the PR 
Adds "Plasmatoid" starting trait, alongside some refactoring of plasma tanks. 

"Plasmatoid" changes both lungs to plasmatoid lungs at round start, and additionally spawns the player with a new pocket plasma tank which is active by default, as well as a plasma emergency box instead of an oxygen emergency box.

Changes `/obj/item/tank/plasma` to `/obj/item/tank/plasma/container`, in order to inherit some properties to the new pocket plasma tank  `/obj/item/tank/plasma/pocket`. Also renames "gas tank (BIOHAZARD)" to "gas tank (plasma)" for consistency.

## Why's this needed? 
Plasmatoid go brrrr

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Added a new starting trait.
```


CaioFrancisco:
Can't say I'm a fan of the feature. I would elaborate, but I'm too tired to do so. Other than that, the changlogs are too vague to be major like that.

flrsh:
Not a fan either. I don't see a good reason to add this.

pali6:
As discussed in Discord before this is not something we want but don't get discouraged from making more PRs!

PR #14339, opened 2023-06-02
Add storage datums to tables with desk drawers

FlameArrow57:
[GAME OBJECTS][REWORK][CLEANLINESS][BUG]

## About the PR 
This PR replaces the /obj/item/storage/desk_drawer that tables are using right now for storage with a storage datum

Something that's slightly different now though is that for adding an item to a storage by clicking its HUD, it relays the click to the item. Now that tables use a storage datum on them directly, this means that in order for storage to work on them, clicking the table while the HUD is shown will now place the item in the storage rather than on the table. Table storages now show at the location of the table always though, making it hard to click the table with the HUD up anyways, so it might be fine? Not sure if that should be considered a bug or not

They visually rustle now too

Also fixes a bug where chemistry tables that were supposed to be spawning with beakers weren't spawning with them

## Why's this needed? 
Cleanliness

### DIFF
--- a/code/WorkInProgress/AzrunStuff.dm
+++ b/code/WorkInProgress/AzrunStuff.dm
@@ -28,21 +28,13 @@
 	item_state = "sheet";
 	name = "Strange Blueprint"
 	interesting = "There is additional detail regarding the creation of flora and fauna."
-
-/obj/item/storage/desk_drawer/azrun/
-	spawn_contents = list(	/obj/item/raw_material/molitz_beta,\
-	/obj/item/raw_material/molitz_beta,\
-	/obj/item/raw_material/plasmastone,\
-	/obj/item/organ/lung/plasmatoid/left,\
-	/obj/item/pen/crayon/red,\
-
-)
 /obj/table/wood/auto/desk/azrun
-	New()
-		..()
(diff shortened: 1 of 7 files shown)


FlameArrow57:
Requested if you want to look

TobleroneSwordfish (changes requested):
Thanks for fixing this, code looks good at a glance, couple of issues I noticed:
- Being locked doesn't appear to stop you opening the drawer by clicking on it or taking stuff out of it
- Not sure if this is just a case of someone putting a weird item in a desk, but if you spawn `/obj/table/wood/auto/desk/lily` you won't be able to remove the large cake inside.

Also this will require secret changes so `TODO` for me or whoever merges this

FlameArrow57:
Fixed first one

For the cake, yeah, when clicking it it's trying to show the context action menu, which has an option for clicking the cake, but I guess those don't show while inside a storage. 
Could make it so that
-Clicking cake while stored, will automatically put it in your hand
-Click dragging cake onto yourself puts it in hand
whichever sounds best?

TobleroneSwordfish:
Probably the first one? I never really liked the weird cake context menu anyway

PR #14340, opened 2023-06-02
Fix storage on disposed atoms not dropping contents

FlameArrow57:
[GAME OBJECTS][BUG]

## About the PR 
This PR fixes a bug where items with storages on them weren't dropping their contents when the item was being qdel()'d

This will only work for /atom/movable right now, which can be added to any /atom when that's done for the first time

## Why's this needed? 
Bug fix

### DIFF
--- a/code/atom.dm
+++ b/code/atom.dm
@@ -197,8 +197,6 @@ TYPEINFO(/atom)
 		atom_properties = null
 		if(!ismob(src)) // I want centcom cloner to look good, sue me
 			ClearAllOverlays()
-
-		src.remove_storage()
 		..()
 
 	proc/Turn(var/rot)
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
Do we want them to? If I `qdel` a backpack full of crime I'm not sure I want the crime to spill everywhere.

FlameArrow57:
This is just for default. I think whether contents are deleted or not depends on the source of the qdel(). In the end though, the source of the qdel() will drop all contents or delete items in the storage, and then the final atom.remove_storage() will do the default action to make sure contents are cleaned up, just happened to go with dropping items

TobleroneSwordfish:
Right but this is reversing the default behaviour, and that's going to cause issues especially with admin tools.

PR #14341, opened 2023-06-02
Touches up on Heater, Shitty Grill and Reclaimer sprites

MitchTheUser:
## About the PR 

Redoes the color schemes of the Heater, Shitty Grill, and Portable Reclaimer, as well as doing some adjustments like making the wheels of the Reclaimer stick out, giving the Shitty Grill an actual grill and coals, and making parts of the sprites contain modern versions of sprites they borrow like the fuel tank, battery, O2 tank and plasma can.

## Why's this needed? 

The current colors are quite dull and look drab in contrast to surrounding machinery. This should make them look nicer and more in line with the shinier colors of more modern sprites.

Heaters Grills and Reclaimers oh my

## Changelog 

```changelog
(u)MitchSmith
(+)The Heater, Shitty Grill, and Portable Reclaimer have been given a new coat of paint.
```


Gannets:
I do think these look better than what we have, thought this was maybe an opportunity for a bigger redesign of these items?

PR #14346, opened 2023-06-02
Adds missing plating tile in Clarion maints

Chatauscours:
[mapping] [bug]

## About the PR 
Adds a missing plating tile, that I forgot to add.

## Why's this needed? 
bug bad

## Changelog 

N/A


pali6 (approved):
How was this not failing map checks hm

PR #14351, opened 2023-06-02
Nuclear Reactor TGUI Performance - Tooltips

Garash2k:
[A-UI][C-Performance]

## About the PR 
Avoiding rendering up to 49 tooltips when they're not needed significantly improves performance on lower end PCs.
I have an i7-3770K, not great, not terrible, and after a bit of debugging, I've noticed that performance gets way better when the control rod tooltips were omitted. Tooltips, unlike most DOM elements, are appended directly to the root/the body which might be why adding dozens of them could cause performance issues unlike other components.
image

Before:
https://github.com/goonstation/goonstation/assets/6396368/de8f4119-4651-4f43-88b3-53a46a6fe49f

After:
https://github.com/goonstation/goonstation/assets/6396368/1a7935d7-d821-4af1-b05f-9ad59081e9d0
(I still have *some* slowdown while hovering, but that's way less of a big deal than the constant slowdown from having the window open)

## Why's this needed? 
Performance improvement good


pali6 (approved):

TobleroneSwordfish (changes requested):
Moving my mouse in a circle over the reactor UI with this PR is enough to hang my client for several seconds, this seems worse.

Garash2k:
> Moving my mouse in a circle over the reactor UI with this PR is enough to hang my client for several seconds, this seems worse.

Dang I assumed that was just me, alright that solution is a no go then.

Garash2k:
Good news, said slowdown got fixed by 5d4b53842ddd574e273848df1bdbe974c9945110!

PR #14355, opened 2023-06-03
Adds a perspective map editor walls define to _build.dm

DisturbHerb:
[MAPPING][INPUT]

## About the PR 

It is now possible to define which wall icons are used in a map editor by uncommenting the `USE_PERSPECTIVE_EDITOR_WALLS` define in `_std/_build.dm`. By default, flat wall and wingrille spawner icons are used.

## Why's this needed? 

As someone who occasionally uses a map editor, I've found the recent default wall icon changes to, instead of improving the map editing experience, actively hinders it. The icons have significantly more going on in the perspective versions, and the introduction of so many lines and shapes makes reading maps in the editor significantly more difficult than flat icons. This matters as, if you're making large scale changes and need to use SDMM for extended periods, the new icons can cause some eye strain; at least for me. 

I think the most convincing way of demonstrating my frustration would be a visual. From looking at the following screenshots, I think that the top one is a little harder to look at, it looks cluttered and the lines constantly break up my focus rather than letting me to simply gloss over them like I could with flat walls.

A section of Cogmap's security foyer with perspective wall icons.
A section of Cogmap's security foyer with flat wall icons.

~~I think a case can be made for the perspective wingrille spawners, they're a little less intrusive than the flat wingrille spawner icons we had previously.~~

~~Of course, my experience is not analagous to everyone's, and I think I'd need more opinions to make sure that this isn't just a me thing. The original PRs were out to solve a problem that does exist -- that of aligning wall-mounted objects properly -- but it hasn't come up regularly enough for me to see the benefits. I myself was personally happy with modifying the base icon state as necessary, or drawing from a palette of modified walls that I already had.~~

~~I think a solution that would satisfy everyone would be a native SDMM solution for switching between perspective and flat icons, but there seem to be very difficult barriers to overcome to render perspective wall outside of runtime.~~

These points have been since superseded by the recent changes made as of 10 June 2023. Now that any given contributor has personal agency over how their editor looks, I believe we have come to a more satisfactory solution.


DisturbHerb:
n.b. Looking at it again, reinforced and regular perspective wingrille spawners aren't very well distinguished from each other in my opinion.

Laboredih123:
perhaps a good idea would be to make custom simplistic icons to show where the lines are on the walls and windows?

CheffieGithub:
I personally prefer the perspective icons. Though im bias since I was the one who changed them. I do think the best solution to leave the icons as is or create simpler icons in perspective as Laboredih says.
Ill PR a fix for the wingrille spawner colours, I neglected to add in the colour difference between reinforced and not.

CheffieGithub:
I think most people would find perspective icons more convenient but more voices on this matter are needed.

I made the changes not because I prefer them myself but because perspective is the standard for sprites and it made sense for the map editor to mirror how it would look in-game. The old sprites were just bugs from having no real "map" iconstate and it making a default one.

I also asked for opinions before I made the PR to get an idea of how it would be percieved, but there wasn't any strong opinions eitherway aside from your own. Personally I don't really mind eitherway but I do think perspective makes more sense.

CheffieGithub:
I also think if you do go down this route you should make every wall not perspective to avoid inconsistencies

Laboredih123:
> What worries me most is the inability to pick what works best for you when working in the editor.

Hm not exactly. Could have a define in build.dm letting us switch between the two tilesets for the map editor. Wouldn't be so hard too.

DisturbHerb:
> Hm not exactly. Could have a define in build.dm letting us switch between the two tilesets for the map editor. Wouldn't be so hard too since Byond lets us use booleans in macro ifs. We can't do it on the go as you would need to reopen the environment every time you switch, but this is next best no?

This is a better idea, I'll give it a try.

BatElite:
To me it comes down to noise, and the perspective graphics are considerably noisier and have more internal contrast than the flat icons so I do prefer the old sprites.

Sord213:
Yeah I'm not sure if it's just "new thing bad" but man I don't like the perspective sprites in editor.

DisturbHerb:
i ate @Laboredih123's idea yummy yummy

Laboredih123:
darn

ZeWaka (approved):

PR #14358, opened 2023-06-03
Mining PDA apps and module

Cherman0:
## About the PR 

Adds a few programs and a module to mining PDAs to allow them to use a lot of those tools that normally just get instant trashed for inventory space, like the geological scanner and GPS. Also makes the mining map a miner-specific program (and the CE, and also engineers on Oshan because its used for hotspots) instead of being baked into the PDA operating system.

- Adds Space GPS Pro as a function of the mining module accessible from the PDA home screen. This is identical to a real GPS, and so is functionally an upgrade over the normal Space GPS app that everyone has.
- Adds the geological scanner AoE scan to the module ability button, for portable scanning.
- Adds the geological scanner targeted scan as a PDA program, for scanning specific rocks.
- Adds the mining map as an app, removes it from the universal operating system. Mining level powergamers can suffer, lol.
- Maybe will add a geiger counter if i feel like it.

## Why's this needed? 

Mining currently has no module or cartridge at all, just as generic a PDA as possible. This adds mostly existing scanner functionality to their PDA, since pda scanners are an established thing and miners could use the inventory space. Also spites nerds who powergame the mining level prefabs by memorizing their shapes.

## Changelog 

```changelog
(u)Cherman0
(*)Miners now have a PDA module and cartridge with some mining related apps. The mining map has been moved to miner PDAs, so knowledge of the prefabs in the mining level is less publicly available.
```


Sbmhawk:
I do not agree with getting rid of the mining map, or at the very least the trench map.
It is the only way of gett8ng your bearings when spaced/trenches.

On space maps at least, you can go to an edge and go to debris, however on trench maps it is not possible.

Sbmhawk:
If powergaming is an issue, perhaps only showing the asteroid/trench maps before the prefab are generated for the general use, if that is possible.

Cherman0:
> I do not agree with getting rid of the mining map, or at the very least the trench map. It is the only way of getting your bearings when spaced/trenched. On space maps at least, you can go to an edge and go to debris, however on trench maps it is not possible.

Aside from using a GPS or the Space GPS app that comes in every PDA, that is. I feel like making it harder for someone to rescue themselves from space without calling for help is a good thing.

Sbmhawk:
> > I do not agree with getting rid of the mining map, or at the very least the trench map. It is the only way of getting your bearings when spaced/trenched. On space maps at least, you can go to an edge and go to debris, however on trench maps it is not possible.
> 
> Aside from using a GPS or the Space GPS app that comes in every PDA, that is. I feel like making it harder for someone to rescue themselves from space without calling for help is a good thing.

The GPS app is almost useless in the trench, I do not believe it tells you the location of other GPS, and by extension the elevator.

I'm not nearly as concerned with space as I am with the trench.

Cherman0:
It tells you your location. If you ended up in the trench with no idea where the elevator is and also dont have a trench map (which are common items on oshan) then congrats, you're lost. Radio a miner to pick you up or something. Seems fine to me.

Sbmhawk:
On space maps, assuming you have EVA equipment, you can get to the station yourself by finding a location with a pod, whether that's in z5 with the map, or going to z3 which has known locations, whatever.

On trench maps, your only option to save yourself is with a map, or a GPS.

Again, if powergaming is an issue, just have a seperate map that is generated between the z5 rock generation, and the z5 prefab generation, and give that to the general population.

Sbmhawk:
On RP, people will probably help you out.

On classic? Better hope you are sec or command in that case.

Cherman0:
"Impossible" is a strong word, but "just having a seperate map without the prefabs" is dramatically more complicated than it sounds and well beyond the scope of this PR. There are lots of publicly available trench maps on oshan, if you are concerned about being lost and alone in the trench and are unwilling to rely on your fellow crew then just pick one of those up. Making it easier for the average crew to get lost beyond the likely ability for them to resolve alone is an intentional reason behind making this change in the first place.

PR #14371, opened 2023-06-03
Modernizes Clown/Cluwn spider sprites

MitchTheUser:
## About the PR 

The masks and shoes of all clown spider variants have been replaced with their modern, nicer counterparts. In addition, the queens had their color schemes altered to be more pleasant, though their overall shape has stayed the same to be in line with all other large spiders.

BIGDISPLAY

## Why's this needed? 

Clown spiders were quite flat and looked weird, in addition to their masks and shoes being outdated sprites (The small clown spider having notably old clown shoes). Now they should look more perspectivized and nice to look at with the replacement of their outdated parts.


CheffieGithub:
Not sure if the moving sprites are needed now since the spiders have animated waddling

MitchTheUser:
> Not sure if the moving sprites are needed now since the spiders have animated waddling

I tested in a private server with and without the jumping animations and found that they are overridden by the waddle and don't do anything. Am going to remove them since they are obsolete.

Issue #14359, opened 2023-06-03
Vending Machine payout broken


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

Player made vending machines do not deposit money into owners bank account.

### Steps to reproduce

1. Make a player vending machine.
2. check your balance at an ATM.
3. Sell something for a noticable amount.
4. Re-Check your ATM and find nothing.

### Expected Behavior

Vending machine sales should go to owners bank account.

### Additional Information & Screenshots

It's my money and I want it NOW!!!

Reported by: Vorlice
On server: Goonstation 3 Roleplay: Morty
Active test merges: nothing
Round log date: 2023-06-03-05-54
Reported on: 2023-06-03 07:03:15
Map: COGMAP

PR #14372, opened 2023-06-03
Stop vending machines from eating money

monkeymilesmoo:
[bug]
## About the PR 
Makes vending machines actually pay either the owner if it is a player machine or the station if not.
I have no idea when this broke since it used to work

Fixes: https://github.com/goonstation/goonstation/issues/14359

## Why's this needed? 
The code for payout was already there, but it was under Topic() instead of ui_act() so it never got called

## Changelog 

```changelog
(u)monkeymilesmoo
(+)Fixed vending machines not paying the owner.
```

### DIFF
--- a/code/modules/vending/vending.dm
+++ b/code/modules/vending/vending.dm
@@ -1127,6 +1127,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item)
 	New(vending_machine, var/Owner)
 		src.vendor = vending_machine
 		src.owner = Owner
+		src.owneraccount = FindBankAccountByName(Owner.scan.registered)
 		..()
 
 	onUpdate()
--- a/code/modules/vending/vending.dm
+++ b/code/modules/vending/vending.dm
@@ -618,6 +618,14 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item)
 							account["current_money"] -= product.product_cost
 						else
 							src.credit -= product.product_cost
+					if (!player_list)
+						wagesystem.shipping_budget += round(product.product_cost * profit) // cogwerks - maybe money shouldn't just vanish into the aether idk						logTheThing(LOG_DEBUG, src, T.owneraccount["current_money"])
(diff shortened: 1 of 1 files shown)


code/modules/vending/vending.dm
```diff
@@ -617,6 +618,13 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item)
 							account["current_money"] -= product.product_cost
 						else
 							src.credit -= product.product_cost
+					if (!player_list)
+						wagesystem.shipping_budget += round(product.product_cost * profit) // cogwerks - maybe money shouldn't just vanish into the aether idk
+					else
+						//Players get 90% of profit from player vending machines QMs get 10%
+						var/obj/machinery/vending/player/T = src
```
TobleroneSwordfish, line 625: `T` as a var name is only acceptable for turfs`

code/modules/vending/vending.dm
```diff
@@ -617,6 +618,13 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item)
 							account["current_money"] -= product.product_cost
 						else
 							src.credit -= product.product_cost
+					if (!player_list)
```
TobleroneSwordfish, line 621: This should all probably be in the above `if(src.pay)` block

TobleroneSwordfish (changes requested):

code/modules/vending/vending.dm
monkeymilesmoo: Above? Wouldn't it make more sense inside since it would only pay out if the money is actually transferred?

code/modules/vending/vending.dm
monkeymilesmoo: Oh sorry, I just moved the Topic code over and didn't think about variable names. Fixed now

code/modules/vending/vending.dm
TobleroneSwordfish: Yep, that's what I meant

PR #14379, opened 2023-06-04
Adds Corpse Spawners

CheffieGithub:
## About the PR 
Adds a new mob spawner object which can spawn a mob and a container. Creates a corpse spawner for critters and for humans from this. They are mainly intended for mapping but could be placed as item drops from things.

Adds some variables to decomposition to allow for humans to not emit miasma or decay since having every corpse rot and fill where it spawns with miasma wouldn't be great.

Adds concrete types for webbing icecubes and adds a variable to stop them melting. (for the webbed corpse spawner)

Adds a human varible silent_death which prevents the death message and gasp for spawners.

The Human spawner has variables for controlling decomposition, being husked, removing random organs, dealing damage to the body, removing items (bag contents, ID, breaking headset, hand items, pockets) to allow for customisation and removal of certain high risk items.

The human spawner also has a random type which will randomly spawn a mob from a weighted list. This has variants to spawn the body inside a bodybag or spiders web.

This PR does not add any of these spawners into maps or loot tables.

image

## Why's this needed? 
Allows for minor random loot to be brought in to prefabs, or adventure areas. Allows corpses to be added to these areas as real mobs and not just fake objects which can better show a story and makes areas feel more real.


code/mob/living/carbon/human.dm
```diff
@@ -83,7 +83,10 @@
 	var/yeet_chance = 0.1 //yeet
 
 	var/decomp_stage = DECOMP_STAGE_NO_ROT
+	var/no_miasma = FALSE
+	var/no_decomp = FALSE
 	var/time_until_decomposition = 0
+	var/silent_death = FALSE // Stops deathgasping and clown confetti
```
pali6, line 89: I feel like these being simple mob properties would be better as they are gonna be FALSE 99% of the time.
See for example the `PROP_MOB_SUPPRESS_LAYDOWN_SOUND` for something similar I did not too long ago.

pali6 (changes requested):

pali6 (approved):

PR #14397, opened 2023-06-05
Add sugar cubes and respective packaging

Glamurio:
[Objects][Feature][QoL]
## About the PR 
Adds sugar cubes as a new snack item, which contains 10u of sugar. They can be found in the new "sugar cubes" container item, with 12 in each package.

image

Sugar cubes can either be eaten, or be added to any `/obj/item/reagent_containers/food/drinks` to add 10u of liquid sugar.

Additionally, this adds a sugar package to most coffee machines on  Cogmap, Cogmap 2, Clarion, Oshan, Nadir, Donut 2, Donut 3 and Kondaru, and thusly also rearranges some of the tables / chairs / mugs to make room. It also adds some creamer to some coffee machines that were missing it, though this wasn't a thorough endeavor.

## Why's this needed? 
This was sparked by a conversation I've had in-game where people talked about how they liked their coffee, where it was established that "it would be nice if we could put sugar in our coffee". Outside of having to walk to a chem dispenser.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Added sugar cubes next to most coffee machines.
```


DisturbHerb:
Mild disagree with the mass placement of sugar on-station by virtue of it currently only being given to station and Central commands. I think the class divide (in something mildly inconsequential like this) is kinda funny. Good sprites!

Glamurio:
> Mild disagree with the mass placement of sugar on-station by virtue of it currently only being given to station and Central commands. I think the class divide (in something mildly inconsequential like this) is kinda funny. Good sprites!

Sugar is available from any chem dispenser (science, bartender), so I'm not sure I agree on the class divide aspect (though I find the idea humorous). Not particularly bothered about the placement, however, just merely treated sugar the same as creamer, and tried to mimic the distribution.

DisturbHerb:
> Sugar is available from any chem dispenser (science, bartender), so I'm not sure I agree on the class divide aspect (though I find the idea humorous).

The only coffee dispensers (right now) with sugar available right next to them exist on the Nadir Bridge and in various locations at Central Command.

flrsh:
Very cute idea! One minor thing is that 12 cubes seems like a lot - maybe reduce the amount to like 5 or 6?

Glamurio:
> Very cute idea! One minor thing is that 12 cubes seems like a lot - maybe reduce the amount to like 5 or 6?

Done!

wolfolotl:
the box seems a little bulky of a sprite for something minor like this, would you be able to make a little sugar bowl or something for it? it just strikes me as visually strange that it's the same size as a donut box

code/modules/chemistry/tools/food_and_drink.dm
```diff
@@ -575,6 +575,18 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks)
 			src.splash_all_contents = 1
 		return
 
+	attackby(obj/item/I, mob/user)
```
frawhst, line 578: Would rather put this in the `afterattack` proc of sugar cubes than here. It's a pretty specific interaction so I don't think `drinks` items need to know about it, just let the sugar cube handle itself.

Also, putting it here means this won't work with beakers/water tanks/whatever. A slightly cleaner way of doing it would be to check `if (target.is_open_container() && target.reagents)` (where `target` is the thing you're clicking with the cube) in the cube's `afterattack`, which lets you dissolve the cubes in anything open that has a reagent holder.

Glamurio: Totally valid criticism, in fact I wanted to do something similar with `attack`, but that only has a `mob/user`, and I wasn't sure if `afterattack` was sufficient. Also the reagents check is smarter, thanks!

frawhst (approved):

PR #14398, opened 2023-06-05
Kills off trace gases

Laboredih123:
[ATMOSPHERICS][REWORK]

## About the PR 
Removes trace gases as much as possible, moving them to the APPLY_TO_GASES system.

## Why's this needed? 
The original purpose of trace gases of being modular has been rendered useless by the APPLY_TO_GASES system, now only really slowing down the game and ruining readability.


Laboredih123:
Hm, that should be pretty easy. If it equals zero, just don't show it.

Laboredih123:
but feature creep...

Laboredih123:
never mind adding that var is not so easy due to shifting the arguments in every single macro. ill just switch the analyser to simple output.

Laboredih123:
this seem good?
image

Laboredih123:
Same as always, if it has zero gas it says there is no gas. I didn't modify that part, I just switched it into simple output so it only shows gases that actually are there.

Laboredih123:
find any bugs please

pali6 (approved):

PR #14401, opened 2023-06-05
Glass Recycler QoL update

Mirodir:
[Objects][QoL][Sounds]

## About the PR 
If successful, putting objects into the Glass Recycler(GR) plays one out of three glass-breaking sounds at random; If unsuccessful, no sound is played.
Creating an object in the GR makes an output sound and if the creation is rejected for not having enough glass, an error sound is played. (As far as I could tell, not having enough glass shouldn't normally happen, because then the button to create in the UI shouldn't be enabled in that case. However, I'm sure there are weird moments when two players try to interact with it at the same time.)

Additionally, creating glassware in the GR will put that item directly into one of the player's free hands if possible.

## Why's this needed? 

This aligns the functionality of the GR with most other machines that create objects.

## Changelog 

```changelog
(u)Nelamil
(+)Glass Recyclers play sounds when used and put created glassware in user's hands if possible.
```

### DIFF
--- a/code/obj/machinery/glass_recycler.dm
+++ b/code/obj/machinery/glass_recycler.dm
@@ -165,7 +165,7 @@ TYPEINFO(/obj/machinery/glass_recycler)
 		product_list += new /datum/glass_product("flute", /obj/item/reagent_containers/food/drinks/drinkingglass/flute, 1)
 		product_list += new /datum/glass_product("pitcher", /obj/item/reagent_containers/food/drinks/drinkingglass/pitcher, 2)
 
-	proc/create(var/type)
+	proc/create(var/type, mob/user)
 		var/datum/glass_product/target_product = null
 		for (var/datum/glass_product/product in product_list)
 			if(product.product_type == type)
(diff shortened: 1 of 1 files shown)


code/obj/machinery/glass_recycler.dm
```diff
@@ -175,12 +177,15 @@ TYPEINFO(/obj/machinery/glass_recycler)
...
 			src.visible_message("<span class='alert'>[src] doesn't have enough glass to make that!</span>")
+			playsound(src.loc, 'sound/machines/buzz-sigh.ogg', 40, 0, 0.1)
 			return
 
 		var/obj/item/G = new target_product.product_path(get_turf(src))
 		src.glass_amt -= target_product.product_cost
 
 		src.visible_message("<span class='notice'>[src] manufactures \a [G]!</span>")
+		playsound(src.loc, 'sound/machines/vending_dispense_small.ogg', 40, 0, 0.1)
+		usr.put_in_hand_or_eject(G)
```
pali6, line 188: Ideally add a new `mob/user` argument to this proc and pass the user explicitly, then use that. The `usr` var can backfire in various fun ways.

Also use the `?.` operator instead of `.` to handle the case when `usr`/`user` is null.

Mirodir:
Assuming I understood your concerns correctly: Done

pali6 (approved):

PR #14402, opened 2023-06-05
get_one_match now sorts results

Sovexe:
## About the PR 
* get_one_match now sorts results. alphabetically by default but if desired proc can be overridden to use a different or no sorting proc. 

## Why's this needed? 
* There get_one_match is great in that it searches for the desired type and has built in prefix trimming, but the results are unsorted by default

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -4483,7 +4483,7 @@ var/global/noir = 0
 
 	. = matches
 
-/proc/get_one_match(var/object, var/base = /atom, use_concrete_types=TRUE, only_admin_spawnable=TRUE)
+/proc/get_one_match(var/object, var/base = /atom, use_concrete_types=TRUE, only_admin_spawnable=TRUE, cmp_proc=/proc/cmp_text_asc)
 	var/list/matches = get_matches(object, base, use_concrete_types, only_admin_spawnable)
 
 	if(!length(matches))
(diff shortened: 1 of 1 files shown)


code/modules/admin/admin.dm
```diff
@@ -4483,7 +4483,7 @@ var/global/noir = 0
 
 	. = matches
 
-/proc/get_one_match(var/object, var/base = /atom, use_concrete_types=TRUE, only_admin_spawnable=TRUE)
+/proc/get_one_match(var/object, var/base = /atom, use_concrete_types=TRUE, only_admin_spawnable=TRUE, cmp_proc=/proc/cmp_text_asc)
```
pali6, line 4486: Imo default should be `null` to:
- keep the current behaviour
- give the caller the option not to sort. Sorting can be slow and sometimes thes get_on_match calls show an absolute crapton of stuff when you for example accidentally or intentionally enter something like "obj"

ZeWaka: if there was some way to filter this based off of the base type length or something that would be cool

Sovexe: it's null for now, I guess we could add another arg, `sort_limit` that when set will sort any list below the length specificed with some sane default value, and otherwise won't bother

code/modules/admin/admin.dm
```diff
@@ -4483,7 +4483,7 @@ var/global/noir = 0
 
 	. = matches
 
-/proc/get_one_match(var/object, var/base = /atom, use_concrete_types=TRUE, only_admin_spawnable=TRUE)
+/proc/get_one_match(var/object, var/base = /atom, use_concrete_types=TRUE, only_admin_spawnable=TRUE, cmp_proc)
```
pali6, line 4486:
```suggestion
/proc/get_one_match(var/object, var/base = /atom, use_concrete_types=TRUE, only_admin_spawnable=TRUE, cmp_proc=null)
```

code/modules/admin/admin.dm
ZeWaka: i like that tbh

PR #14404, opened 2023-06-05
Alters path for Donut3 Morgue Chute from Med Booth

DanaTarSUS:
## About the PR 
Slightly changes the path of the  morgue chute from the med booth so that it's no longer under as many walls nor in range of mining magnet explosions.

## Why's this needed? 
Donut3 pipes suck, this pr makes them suck slightly less


frawhst (approved):
;MINING?

PR #14405, opened 2023-06-05
apply materials to deployed frames if they dont have one

Skeletonman0:
[OBJECTS] [FEATURE]
## About the PR 
allows soulsteel frames to create soulsteel objects if the object doesnt have a material

add some safety checks to prevent soulsteel mechcomp cabinets from ceasing to exist when attacking themselves

## Why's this needed? 
soulsteel mechcomp cabinet go brrr

## Changelog 

```changelog
(u)Skeletonman0
(+)Arcplated electronics frames will now create arcplated devices if they could be arcplated


pali6 (approved):
I feel like this is a mistake. But a fun one!

PR #14415, opened 2023-06-06
Gene Booth Speedup Research

Nexusuxen:
## About the PR 

Adds 3 new researches to the GeneTek console, which each add +25% to the Gene Booth's working speed. Each one subsequently reduces the waiting time by approximately 5 seconds (usually a bit lower due to it relying on Process()), down to ~5 seconds. This is done by giving the gene booth a new static var, `process_speedup`, and using that in the Process() check.

Below is a screenshot of the new researches.

image

## Why's this needed? 

The gene booth takes too damn long to work - 20 seconds per gene! And with dozens of people in a server and potentially half a dozen or more genes in the booth, that'll take a WHILE. This makes it at least bearable.

## Changelog 

```changelog
(u)Nexusuxen
(*)Gene Booths can now be upgraded with GeneTek research to work faster by up to 75% (~5 seconds per gene).
```

### DIFF
--- a/code/modules/medical/genetics/geneticsResearch.dm
+++ b/code/modules/medical/genetics/geneticsResearch.dm
@@ -321,9 +321,9 @@ var/datum/geneticsResearchManager/genResearch = new()
 	requiredResearch = list(/datum/geneticsResearchEntry/genebooth)
 
 	onFinish()
-		var/obj/machinery/genetics_booth/_type = /obj/machinery/genetics_booth
-		_type.process_speedup += 0.25
-		_type = "byond shut the hell up" // we get warned about _type being defined but unused otherwise
+		var/obj/machinery/genetics_booth/type = /obj/machinery/genetics_booth
+		type.process_speedup += 0.25
+		. = type // we get warned about type being defined but unused and idk how else to make byond shut up
 		..()
 
 // TIER TWO
@@ -374,9 +374,9 @@ var/datum/geneticsResearchManager/genResearch = new()
 	requiredResearch = list(/datum/geneticsResearchEntry/genebooth)
 
 	onFinish()
-		var/obj/machinery/genetics_booth/_type = /obj/machinery/genetics_booth
-		_type.process_speedup += 0.25
-		_type = "byond shut the hell up"
+		var/obj/machinery/genetics_booth/type = /obj/machinery/genetics_booth
+		type.process_speedup += 0.25
+		. = type
 		..()
 
 // TIER THREE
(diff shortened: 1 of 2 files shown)


Nexusuxen:
Ideally with this speedup the booth shaking/clonging should happen more frequently too, but it seems like Process() goes so slowly that it often finishes before it even gets a chance to clonk. Could do something with SPAWN() but honestly I don't feel like reworking gene booths to solve a non-existent problem.

TobleroneSwordfish (approved):

code/modules/medical/genetics/geneticsResearch.dm
```diff
@@ -311,6 +311,21 @@ var/datum/geneticsResearchManager/genResearch = new()
...
+	researchTime = 1200
+	researchCost = 150
+	tier = 2
+	requiredMutRes = list("early_secret_access")
+	requiredResearch = list(/datum/geneticsResearchEntry/genebooth)
+
+	onFinish()
+		var/obj/machinery/genetics_booth/type = /obj/machinery/genetics_booth
+		type.process_speedup += 0.25
+		. = type // we get warned about type being defined but unused and idk how else to make byond shut up
```
frawhst, line 326:
```suggestion
		type = type // we get warned about type being defined but unused and idk how else to make byond shut up
```
We'll be able to suppress warnings in 515. Until then this is a bit cleaner since it removes the warning while accomplishing absolutely nothing.

code/modules/medical/genetics/geneticsResearch.dm
```diff
@@ -350,6 +365,20 @@ var/datum/geneticsResearchManager/genResearch = new()
...
+	desc = "Increases the working speed of the Gene Booth by 25%."
+	researchTime = 1200
+	researchCost = 75
+	tier = 2
+	requiredResearch = list(/datum/geneticsResearchEntry/genebooth)
+
+	onFinish()
+		var/obj/machinery/genetics_booth/type = /obj/machinery/genetics_booth
+		type.process_speedup += 0.25
+		. = type
```
frawhst, line 379:
```suggestion
		type = type
```

code/modules/medical/genetics/geneticsResearch.dm
```diff
@@ -398,6 +427,21 @@ var/datum/geneticsResearchManager/genResearch = new()
...
+	desc = "Increases the working speed of the Gene Booth by an additional +25%."
+	researchTime = 1800
+	researchCost = 100
+	tier = 3
+	requiredResearch = list(/datum/geneticsResearchEntry/genebooth, /datum/geneticsResearchEntry/gene_booth_speedup)
+
+	onFinish()
+		var/obj/machinery/genetics_booth/type = /obj/machinery/genetics_booth
+		type.process_speedup += 0.25
+		. = type
```
frawhst, line 442:
```suggestion
		type = type
```

frawhst (commented):
Code/feature is good, small nitpick.

PR #14416, opened 2023-06-06
Completely resprites the Buttcrab

MitchTheUser:
## About the PR 

The Buttcrab has been redesigned from the ground up, with a slightly smaller body based on the worn sprite of the butt, allowing it to have a more vibrant look compared to the comparatively less bright color scheme of the original. It's legs are more spindly and distinct, and it's walk cycle has been completely remade, now easier to see as it's legs are no longer squashed together and crossing over each other, leading to a much more clear and clean animation. The Buttcrab is also now in perspective, and has 4 unique sprites for each direction it looks in, creating a much more three dimensional look.

CrabDisplay

## Why's this needed? 

The current Buttcrab is not in perspective and does not have unique directional sprites. These new sprites and animations give a lot of life to the critter which should make it a joy to interact with in rounds.


ZeWaka (approved):
Significantly better than prior!

PR #14436, opened 2023-06-06
Prevents wholetuna cordata fish to be able to be turned in  the aquatic research upload terminal

M-Earthfire:
[BUG][Catering]

## About the PR 

This PR makes it so that fish from botany cannot be turned into fishing credits and angler exp. Instead, the fish will be just ejected.
Also, botany fish have an addition in their description that points out that they are non-compareable to fish caught in the ~~kitchen sink~~ wild.

## Why's this needed? 

The wholetuna cordata plant enables anglers to very easily gather angler exp without much well... actuall using their fishing rod. Thus, this PR fixes what could be considered an exploit.


code/modules/fishing/fishing_gear.dm
```diff
@@ -233,27 +233,31 @@ TYPEINFO(/obj/item/fish_portal)
...
-					new/obj/item/requisition_token/fishing/epic(src.loc, src.layer + 0.1)
-					JOB_XP(user, "Angler", 4)
-					qdel( P )
-				if (FISH_RARITY_LEGENDARY)
-					new/obj/item/requisition_token/fishing/legendary(src.loc, src.layer + 0.1)
-					JOB_XP(user, "Angler", 5)
-					qdel( P )
+			//No botany fish. Be a real angler and use that damn fishing rod
+			if (P.fishing_upload_blacklisted)
+				P.set_loc(get_turf(src))
```
pali6, line 238: I feel like this should maybe give some message?

pali6: Or perhaps track if any fish get rejected and display one message at the end if yes. To prevent it from being spammy.

M-Earthfire: That makes certainly sense, will add a little message right now.

PR #14474, opened 2023-06-08
Deprecates /datum/mind/var/master

Mister-Moriarty:
[Internal] [Code Quality]

## About The PR:
Deprecates `/datum/mind/var/master` and `/proc/is_mentally_dominated_by()`, replacing them in most cases with the new `/datum/mind/proc/get_master()`.
Removes superfluous vampire thrall checks in the vampire bite ability code.

## Why Is This needed?
Removal of remnants from the old antagonist system that has been superceded by antagonist datums.

### DIFF
--- a/code/datums/gamemodes/spy.dm
+++ b/code/datums/gamemodes/spy.dm
@@ -111,14 +111,14 @@
 	src.spies.Add(spymind)
 	src.spies[spymind] = leadermind
 	spymind.special_role = "spyminion"
-	spymind.master = leader.ckey
+	//spymind.master = leader.ckey
 
 	return 1
 
 /datum/game_mode/spy/proc/remove_spy(mob/living/spy)
 	src.spies.Remove(spy)
 	spy.mind.special_role = null
-	spy.mind.master = null
+	//spy.mind.master = null
 	return 1
 
 /datum/game_mode/spy/declare_completion()
--- a/code/datums/gamemodes/spy.dm
+++ b/code/datums/gamemodes/spy.dm
@@ -111,14 +111,12 @@
 	src.spies.Add(spymind)
 	src.spies[spymind] = leadermind
 	spymind.special_role = "spyminion"
-	//spymind.master = leader.ckey
 
 	return 1
 
 /datum/game_mode/spy/proc/remove_spy(mob/living/spy)
 	src.spies.Remove(spy)
 	spy.mind.special_role = null
-	//spy.mind.master = null
 	return 1
 
 /datum/game_mode/spy/declare_completion()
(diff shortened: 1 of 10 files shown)


code/datums/gamemodes/spy.dm
```diff
@@ -111,14 +111,14 @@
 	src.spies.Add(spymind)
 	src.spies[spymind] = leadermind
 	spymind.special_role = "spyminion"
-	spymind.master = leader.ckey
+	//spymind.master = leader.ckey
```
ZeWaka, line 114: Much rather this gets removed than commented out and forgotten about

ZeWaka (changes requested):

code/modules/antagonists/antagonist_helpers.dm
```diff
@@ -17,7 +17,19 @@ proc/get_all_gangs()
...
+ */
+/datum/mind/proc/get_master(antagonist_role_id)
+	if (antagonist_role_id)
+		var/datum/antagonist/subordinate/antagonist_role = src.get_antagonist(antagonist_role_id)
+		if (antagonist_role?.master)
+			. = antagonist_role.master
+		return
+
+	for (var/datum/antagonist/subordinate/antagonist_role in src.antagonists)
+		. = antagonist_role.master // Return the master of the newest antagonist datum in `antagonists`.
```
ZeWaka, line 32: is the newest at the end or something?

code/modules/antagonists/antagonist_helpers.dm
```diff
@@ -17,7 +17,19 @@ proc/get_all_gangs()
 	for (var/datum/antagonist/gang_leader/antag_datum in get_all_antagonists(ROLE_GANG_LEADER))
 		. += antag_datum.gang
 
-	return .
+/** Returns the mind of this mind's master.
+ * If a specific antagonist role ID is provided, then the master of the antagonist datum of that ID belonging to this mind will be returned.
+ * If no antagonist role ID is provided, then the master of the newest antagonist datum will be returned.
+ */
+/datum/mind/proc/get_master(antagonist_role_id)
+	if (antagonist_role_id)
```
ZeWaka, line 25: what if this is called with a no id and there is no antagonists? you should set a default return value

ZeWaka (changes requested):

code/modules/antagonists/antagonist_helpers.dm
Mister-Moriarty: That is how lists work, to my knowledge. The most recently instantiated antagonist datum _should_ append itself to the end of the `antagonists` list on the owner mind.

code/modules/antagonists/antagonist_helpers.dm
Mister-Moriarty: Is the default return value not implicitly `null`? If so, to me that seems a sensible value to return if no master could be found.

ZeWaka: it is, just wanted to make sure you thought about it

PR #14492, opened 2023-06-09
TGUI Ability manager

Sovexe:
## About the PR 
* Recreation of the ability manager in TGUI
* Point cost input added directly to panel for ease of modification
* rename button added per request
image

## Why's this needed? 
* modern admin tools


code/modules/admin/abilitymanager.dm
```diff
@@ -0,0 +1,80 @@
...
+	return tgui_admin_state.can_use_topic(src, user)
+
+/datum/abilitymanager/ui_interact(mob/user, datum/tgui/ui)
+	ui = tgui_process.try_update_ui(user, src, ui)
+	if (!ui)
+		ui = new(user, src, "AbilityManager")
+	ui.open()
+
+/datum/abilitymanager/ui_data(mob/user)
+	var/list/ability_props
```
ZeWaka, line 17: You probably want to initialize this empty?

Sovexe: The JS is able to handle receiving it as either null or empty, I can initialize it as empty to be clearer though

ZeWaka (approved):

PR #14497, opened 2023-06-09
Reworks /proc/get_all_antagonists()

Mister-Moriarty:
[Internal] [Code Quality]

## About The PR:
Reworks `/proc/get_all_antagonists()` to iterate through a new `antagonists` global list, as opposed to iterating through every mind datum.
The `antagonists` global list will contain all antagonist datums sorted by role ID; i.e. all traitor datums will be filed under `antagonists["[ROLE_TRAITOR]"]`.

`/datum/game_mode/var/traitors` and `/datum/game_mode/var/Agimmicks` will eventually be superceded by `get_all_antagonists()`.
Deprecates `/proc/get_all_enemies()`.

## Why Is This needed?
Optimisation of the new system and removal of remnants from the old antagonist system that has been superceded by antagonist datums.

### DIFF
--- a/code/modules/antagonists/__antagonist.dm
+++ b/code/modules/antagonists/__antagonist.dm
@@ -72,9 +72,11 @@ ABSTRACT_TYPE(/datum/antagonist)
 			return FALSE
 		src.owner.antagonists.Add(src)
 
-	Del()
+	disposing()
 		if (owner && !src.pseudo)
-			antagonists["[src.id]"] -= src
+			LAZYLISTREMOVE(antagonists["[src.id]"], src)
+			if (isnull(antagonists["[src.id]"]))
+				antagonists -= "[src.id]"
 
 			owner.former_antagonist_roles.Add(owner.special_role)
 			owner.special_role = null // this isn't ideal, since the system should support multiple antagonists. once special_role is worked around, this won't be an issue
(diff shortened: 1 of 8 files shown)


Mister-Moriarty:
I do suppose `antagonists` could contain a list of antagonist datums, then antagonist minds can be accessed through the `owner` variable.

code/modules/antagonists/__antagonist.dm
```diff
@@ -51,6 +51,10 @@ ABSTRACT_TYPE(/datum/antagonist)
 		src.pseudo = do_pseudo
 		src.vr = do_vr
 		if (!do_pseudo && !do_vr) // there is a special place in code hell for mind.special_role
+			if (!islist(antagonists["[src.id]"]))
+				antagonists["[src.id]"] = list()
+			antagonists["[src.id]"] += src
```
pali6, line 56:
```suggestion
			LAZYLISTADD(antagonists["[src.id]"], src)
```
Seldom used but useful macro for this.

code/modules/antagonists/antagonist_helpers.dm
```diff
@@ -1,25 +1,33 @@
...
-				break
+
+	if (antagonist_role_id)
+		return antagonists["[antagonist_role_id]"]
+
+	for (var/antagonist_type in concrete_typesof(/datum/antagonist))
+		var/datum/antagonist/A = antagonist_type
+		var/list/datum/antagonist/antagonist_roles = antagonists["[initial(A.id)]"]
+		if (length(antagonist_roles))
+			. += antagonist_roles
```
pali6, line 18: Am I missing something or could we just iterate over `antagonists` and join the values into one list?

Mister-Moriarty: Yes, that would be a far better method than the above.
Thank you!

ZeWaka (approved):

Issue #14385, opened 2023-06-05
Fake wall disassembly material mismatch.


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

Disassembling a soulsteel fake wall (or presumed any kind) yeilds normal steel girders and plate.

### Steps to reproduce

1) Make soul steel wall from soul steel girder and soulsteel plates.
2) Make fake wall. 
3) Disassemble with screwdriver
4) Observe steel girder and steel plate.

### Expected Behavior

soulsteel girder and soulsteel plate after disassembling.

### Additional Information & Screenshots

Reported by: Vorlice
On server: Goonstation 3 Roleplay: Morty
Active test merges: #14354
Round log date: 2023-06-05-02-19
Reported on: 2023-06-05 03:06:01
Map: KONDARU

Issue #13870, opened 2023-04-27
You can plate girders with glass


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

You can complete unfinished walls with glass sheets, but they don't take on the glass material, they just turn into normal walls.

### Steps to reproduce

1. Make girder
2. Apply glass
3. Wall

### Expected Behavior

Can't use glass/makes a glass wall

### Additional Information & Screenshots

Reported by: ATDoop
On server: Goonstation 1 Classic: Heisenbee
Active test merges: #13023 and #13667
Round log date: 2023-04-27-18-06
Reported on: 2023-04-27 18:45:07
Map: COGMAP

PR #14501, opened 2023-06-09
Walls now save their girder material when built

monkeymilesmoo:
[Bug][Materials]
## About the PR 
Untitled-1
Untitled-2

Previously walls would just take on the material of whatever girder they were built on. This ended up in the girder basically eating whatever sheet was put onto it regardless of the material. Now the girder's material is stored in the wall until the wall is broken in which case a girder of the correct material spawns. Also, in order to place a sheet on a girder, the sheet has to be a metal sheet, so glass walls won't be able to be built.

Should fix:
https://github.com/goonstation/goonstation/issues/14385
https://github.com/goonstation/goonstation/issues/13870

## Why's this needed? 

Girders shouldn't determine a wall's material since the plating is what is actually visible from the outside.

## Changelog 

```changelog
(u)monkeymilesmoo
(*)Plating on a wall can now be of a different material from the girder it was built on without being absorbed by the girder on placement.
```

### DIFF
--- a/code/obj/false_wall.dm
+++ b/code/obj/false_wall.dm
@@ -123,7 +123,7 @@ ADMIN_INTERACT_PROCS(/turf/simulated/wall/false_wall, proc/open, proc/close)
 				var/datum/material/M = getMaterial("steel")
 				A.setMaterial(src.material ? src.material : M, copy = FALSE)
 				B.setMaterial(src.girdermaterial ? src.girdermaterial : M, copy = FALSE)
-				
+
 				var/floorname1	= src.floorname
 				var/floorintact1	= src.floorintact
 				var/floorburnt1	= src.floorburnt
@@ -159,7 +159,7 @@ ADMIN_INTERACT_PROCS(/turf/simulated/wall/false_wall, proc/open, proc/close)
 		if (src.operating)
 			return 0
 		src.operating = 1
-		src.name = "false [src.material.matname] wall"
+		src.name = "false [src.material.name] wall"
 		animate(src, time = delay, pixel_x = 25, easing = BACK_EASING)
 		SPAWN(delay)
 			//we want to return 1 without waiting for the animation to finish - the textual cue seems sloppy if it waits
@@ -181,7 +181,7 @@ ADMIN_INTERACT_PROCS(/turf/simulated/wall/false_wall, proc/open, proc/close)
 		if (src.operating)
 			return 0
 		src.operating = 1
-		src.name = "[src.material.matname] wall"
+		src.name = "[src.material.name] wall"
 		animate(src, time = delay, pixel_x = 0, easing = BACK_EASING)
 		src.set_density(1)
 		src.gas_impermeable = 1
--- a/code/obj/false_wall.dm
+++ b/code/obj/false_wall.dm
@@ -120,9 +120,9 @@ ADMIN_INTERACT_PROCS(/turf/simulated/wall/false_wall, proc/open, proc/close)
 				//a false wall turns into a sheet of metal and displaced girders
 				var/atom/A = new /obj/item/sheet(src)
 				var/atom/B = new /obj/structure/girder/displaced(src)
-				var/datum/material/M = getMaterial("steel")
-				A.setMaterial(src.material ? src.material : M, copy = FALSE)
-				B.setMaterial(src.girdermaterial ? src.girdermaterial : M, copy = FALSE)
+				var/datum/material/defaultMaterial = getMaterial("steel")
+				A.setMaterial(src.material ? src.material : defaultMaterial, copy = FALSE)
+				B.setMaterial(src.girdermaterial ? src.girdermaterial : defaultMaterial, copy = FALSE)
 
 				var/floorname1	= src.floorname
 				var/floorintact1	= src.floorintact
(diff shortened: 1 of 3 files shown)


code/obj/false_wall.dm
```diff
@@ -160,7 +159,7 @@ ADMIN_INTERACT_PROCS(/turf/simulated/wall/false_wall, proc/open, proc/close)
 		if (src.operating)
 			return 0
 		src.operating = 1
-		src.name = "false wall"
+		src.name = "false [src.material.name] wall"
```
TobleroneSwordfish, line 162: Can you rely on all false walls having a material? I'm guessing there might be some in azones and such that don't

code/obj/false_wall.dm
```diff
@@ -182,7 +181,7 @@ ADMIN_INTERACT_PROCS(/turf/simulated/wall/false_wall, proc/open, proc/close)
 		if (src.operating)
 			return 0
 		src.operating = 1
-		src.name = "wall"
+		src.name = "[src.material.name] wall"
```
TobleroneSwordfish, line 184: ditto

code/obj/structure.dm
```diff
@@ -121,7 +121,9 @@ obj/structure/ex_act(severity)
 			actions.start(new /datum/action/bar/icon/girder_tool_interact(src, W, GIRDER_REINFORCE, null, user), user)
 
 		else
-			actions.start(new /datum/action/bar/icon/girder_tool_interact(src, W, GIRDER_PLATE, null, user), user)
+			if (S.material.material_flags & MATERIAL_METAL)
+				actions.start(new /datum/action/bar/icon/girder_tool_interact(src, W, GIRDER_PLATE, null, user), user)
+			else(boutput(user, "You cannot plate [src] with [S]!"))
```
TobleroneSwordfish, line 126:
```suggestion
			else
				boutput(user, "You cannot plate [src] with [S]!")
```

code/obj/structure.dm
```diff
@@ -121,7 +121,9 @@ obj/structure/ex_act(severity)
 			actions.start(new /datum/action/bar/icon/girder_tool_interact(src, W, GIRDER_REINFORCE, null, user), user)
```
TobleroneSwordfish, line 121: Might want to prevent reinforcing with glass too?

code/obj/structure.dm
```diff
@@ -250,15 +252,15 @@ obj/structure/ex_act(severity)
 				var/turf/Tsrc = get_turf(the_girder)
 				var/turf/simulated/wall/WALL
 				var/obj/item/sheet/S = the_tool
+				var/datum/material/M = getMaterial("steel")
```
TobleroneSwordfish, line 255:
```suggestion
				var/datum/material/defaultMaterial = getMaterial("steel")
```

code/obj/structure.dm
```diff
@@ -299,19 +301,16 @@ obj/structure/ex_act(severity)
...
 
 		T.ReplaceWith(target_type, FALSE, FALSE, FALSE)
 		var/atom/A = src.loc
-		if(oldmat)
-			A.setMaterial(oldmat)
-		else
-			var/datum/material/M = getMaterial("steel")
-			A.setMaterial(M)
-
+		var/datum/material/M = getMaterial("steel")
```
TobleroneSwordfish, line 309: ditto

code/turf/walls.dm
```diff
@@ -131,86 +132,65 @@
 	qdel(parts)
 
 /turf/simulated/wall/proc/dismantle_wall(devastated=0, keep_material = 1)
+	var/datum/material/M = getMaterial("steel")
```
TobleroneSwordfish, line 135: ditto

TobleroneSwordfish (changes requested):

code/turf/walls.dm
```diff
@@ -131,86 +132,65 @@
...
-				C.setMaterial(src.material)
-			else
-				var/datum/material/M = getMaterial("steel")
-				A.setMaterial(M, copy = FALSE)
-				B.setMaterial(M, copy = FALSE)
-				C.setMaterial(M, copy = FALSE)
+
+			A.setMaterial(src.girdermaterial ? src.girdermaterial : defaultMaterial, copy = FALSE)
+			B.setMaterial(src.material ? src.material : defaultMaterial, copy = FALSE)
+			C.setMaterial(src.material ? src.material : defaultMaterial, copy = FALSE)
```
pali6, line 167: Also should copy.

code/obj/structure.dm
```diff
@@ -299,19 +301,16 @@ obj/structure/ex_act(severity)
...
 		var/turf/simulated/wall/false_wall/FW = A
+
+		FW.setMaterial(S.material ? S.material : M, copy = FALSE)
```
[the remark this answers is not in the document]
monkeymilesmoo: oh yeah that, I think that was just from me moving code from what it was before into ? : , It should be S.material or M, copy = FALSE so that it doesn't run the copymaterial proc im pretty sure, but I'm not actually sure the priority of ? : now that I think about it. does the M, copy = FALSE need to be inside parenthesis
i.e. FW.setMaterial(S.material ? S.material : (M, copy = FALSE))

PR #14509, opened 2023-06-10
Mobifies Martians

CheffieGithub:
## About the PR 
Adds AI to martains and replaces the objects with them.

Critter changes for convience:
Makes retaliation use its own proc for children overrides. (martians use this)

Removes the cold instant killing martian critters.

## Why's this needed? 
more object critter removal


ZeWaka:
ur fault hehe

CheffieGithub:
I am the bringer of conflicts

flappybatpal:
The main thing seems to be them rotating 90 degrees when dead which doesn't match their animation and is really awkward with the map dead bodies as you will walk in and they will suddenly rotate 90 degrees.

PR #14511, opened 2023-06-10
Visible reactor neutrons for infrared vision

Garash2k:
[Objects][Feature]

## About the PR 
Makes it so that the neutrons emitted by the nuclear reactor are visible if the player has infrared vision, which is granted by mesons.

https://github.com/goonstation/goonstation/assets/6396368/193e3444-6789-4637-98ea-1170d57ec8de

I picked an icon named trace and used a green tint, if anyone has better sprite ideas I'm all eyes!

## Why's this needed? 
Looks nice, is fun, lets engineers see how harmful the reactor actually is, and how neutrons get stopped by objects/walls.
Amy gave the idea a thumbs up

## Changelog 

```changelog
(u)Garash
(+)Neutrons emitted by the nuclear reactor are now visible for people with infrared vision, (which is granted by mesons amongst other means.)
```

### DIFF
--- a/code/modules/projectiles/projectile_parent.dm
+++ b/code/modules/projectiles/projectile_parent.dm
@@ -278,11 +278,13 @@
 		if(istype(proj_data))
 			src.icon = proj_data.icon
 			src.icon_state = proj_data.icon_state
+			src.invisibility = proj_data.invisibility
 			if (!proj_data.override_color)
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish (approved):

amylizzle (approved):
lgtm. Elegant solution.

PR #6651, opened 2021-11-12
Adds the ability to combine a breathing mask with a clown/mime mask to create a mask with internals

Momoberry:
## About the PR 
Adds functionality for putting a breathing mask into a mime/clown mask to act as a mask with internals. They do not start with them. Ideally will make it less restrictive to play the jobs. They still have their thematic handicaps, such as the mime cannot speak, and the clown is clumsy, etc. Honestly, this wouldn't even really affect much balance wise, would just let the players not have to spend a portion of their round just getting countermeasures to not being able to have internals. Not exactly many clowns and mimes, after all.

## Why's this needed? 
While this is pretty hotly debated, I think generally people would like to be able to play the roles they most enjoy without having to be so handicapped they have to spend a significant amount of time just making counter measures to survive depressurization or gas leaks/floods. I know I'm tired of having to do a niche mechanic of suiciding to get through a gas flood as a Mime.

## Changelog

```changelog
(u)MomoBerry
(+)Clowns and Mimes can now combine a breathing mask with their own to create a mask with internals.
```

PR #14513, opened 2023-06-10
Breath masks can be attached to clown masks

monkeymilesmoo:
[clothing]
## About the PR 
Lets people shove a breath mask underneath their clown mask in order to breathe while performing funny business.
All it does is add the flag MASKINTERNALS to the clown mask so it can connect to an oxygen tank.

## Why's this needed? 
Even clowns deserve oxygen.

## Changelog 

```changelog
(u)monkeymilesmoo
(+)Clicking on a clown mask with a breath mask now shoves the breath mask inside, allowing for clowns to breathe without having to break character.
```

### DIFF
--- a/code/obj/item/clothing/masks.dm
+++ b/code/obj/item/clothing/masks.dm
@@ -379,6 +379,15 @@ TYPEINFO(/obj/item/clothing/mask/monkey_translator)
 			src.item_state = "clown_hat"
 			user.show_text("You untuck the wig from the [src].")
 
+	attackby(obj/item/W, mob/user)
+		if(istype(W, /obj/item/clothing/mask/breath) && !(src.c_flags & MASKINTERNALS))
+			src.c_flags |= MASKINTERNALS
+			boutput(user, "You shove [W] under the [src]")
+			src.desc += " There seems to be a breathing mask stuck under the nose."
+			qdel(W)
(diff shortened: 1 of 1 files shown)


DisturbHerb:
I project that this PR would be quite controversial, but you never know.

monkeymilesmoo:
well that was a quick reply

monkeymilesmoo:
You are probably right, there was a PR from a few years ago that got shut down, but I'm wondering what people think now. That PR had a similar ability with the mime mask, but the clown mask seems more reasonable to hide a breath mask inside of.

CheffieGithub:
I like this, having to take your mask off / run though / carry oxygen meds for depressurised areas as a clown just gets annoying fast and for me a clown without a mask isn't the same. Want me to slap feedback wanted on this?

monkeymilesmoo:
Sure, I'd like to know what people think about this since I'm sure some won't like it.

pali6:
For reference: https://github.com/goonstation/goonstation/pull/6651
https://forum.ss13.co/showthread.php?tid=17509

Gannets:
I don't think clowns should be able to wear oxygen gear and their mask simultaneously.

Tarmunora:
superseded by a4648b1

Issue #14515, opened 2023-06-10
Can only assemble wooden floor tiles one at a time


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Unlike crafting with other materials, you can't bulk craft wooden floor tiles and can only do it four at a time.

### Steps to reproduce

1. Turn wooden planks into sheets
2. Craft wooden floor tiles

### Expected Behavior

To be able to bulk craft wood flooring

### Additional Information & Screenshots

Reported by: Goat_Real
On server: Goonstation 4 Roleplay: Sylvester
Active test merges: nothing
Round log date: 2023-06-10-15-04
Reported on: 2023-06-10 16:48:12
Map: DONUT2

PR #14518, opened 2023-06-10
Wood tiles can be built more than once at a time

monkeymilesmoo:
[BUG] [TRIVIAL]
## About the PR 
For some reason wood tiles were unable to be crafted more than once. This was because the crafting id of them had _wood at the end for some reason, even though it had no special crafting exception. Removing _wood from the crafting id makes it able to be crafted more than once at a time.
Fixes https://github.com/goonstation/goonstation/issues/14515

## Why's this needed? 
This shouldn't have ever even a thing.

### DIFF
--- a/code/obj/item/building_materials.dm
+++ b/code/obj/item/building_materials.dm
@@ -1254,7 +1254,7 @@ ABSTRACT_TYPE(/datum/sheet_crafting_recipe/wood)
 
 	wood
 		fl_tiles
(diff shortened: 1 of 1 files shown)


M-Earthfire:
does it apply to bamboo as well? that material has the same problem

monkeymilesmoo:
Since they both have MATERIAL_WOOD tags and that is what determines which recipes to check for, yes.

PR #14520, opened 2023-06-10
Adds a stardew valley-esque minigame to fishing!

Merlin1230:
## About the PR 
Giving fishing a minigame instead of just some bar filling up!

Currently a work in progress, with a list here of what I've done (and plan to do) so far:

- [x] Have a block move down, and go up when the mouse is held
- [ ] Have an actual fish move up and down, and progress a bar when the block covers it
- [ ] (potentially) have some different fish behaviors/movement patterns

## Why's this needed? 

Fishing as it is doesnt really have any skill other then click and get fish. This adds something fun to the act of fishing, as well as making it need SOME skill

## Changelog 

```changelog
(u)Merlin1230
(+)Fishing now has a minigame, similar to stardew valley!
```


Gannets:
Thank you for your work on this!

Merlin1230:
Of course! Sidenote if anyone happens to notice a way to improve the code by all means point it out, im SURE this code sucks/has some oversight

wolfolotl:
potentially controversial take: I don't think having the fishing minigame resemble stardew valley's so closely is a great idea. it can make fishing feel more like a feature Lifted from stardew than one made with the rest of goon's nature in mind - nothing else in gooncode has this method of engagement, which makes this feel more incongruous. additionally, I can see this being pretty frustrating for someone who deals with a lot of lag, or even just someone who HATES the stardew valley fishing minigame (there are more of these kinds of people out there than you would expect)

monkeymilesmoo:
> potentially controversial take: I don't think having the fishing minigame resemble stardew valley's so closely is a great idea. it can make fishing feel more like a feature Lifted from stardew than one made with the rest of goon's nature in mind - nothing else in gooncode has this method of engagement, which makes this feel more incongruous. additionally, I can see this being pretty frustrating for someone who deals with a lot of lag, or even just someone who HATES the stardew valley fishing minigame (there are more of these kinds of people out there than you would expect)

Maybe it could be optional for a reward such as being quicker or something?

Merlin1230:
> potentially controversial take: I don't think having the fishing minigame resemble stardew valley's so closely is a great idea. it can make fishing feel more like a feature Lifted from stardew than one made with the rest of goon's nature in mind - nothing else in gooncode has this method of engagement, which makes this feel more incongruous. additionally, I can see this being pretty frustrating for someone who deals with a lot of lag, or even just someone who HATES the stardew valley fishing minigame (there are more of these kinds of people out there than you would expect)

fair enough, I dont really have any alternatives in mind, however, stardew's is really the only that comes to mind thats simple yet more then just clicking when some bubble shows up or whatever. I do think its fair to think just copying it exactly, so perhaps some modifcations could be made to it? I dont have any off the top of my head but feel free to suggest something!

wolfolotl:
maybe something utilizing the intent system in some way? it could help people train their ability to use the intent hotkeys as well as being not as strenuous as a mouse-based thing

monkeymilesmoo:
> maybe something utilizing the intent system in some way? it could help people train their ability to use the intent hotkeys as well as being not as strenuous as a mouse-based thing

This seems like a good idea but it also seems like it would just be something like this:
image

Merlin1230:
I did wanna try to avoid TGUI since I did feel having it be ingame would be cleaner (since I feel like having it show up rather then popping a window up is nicer), butt tgui is also really the only way to.. make that client side (at least easily)

I will say I do like the sound of the one ikea purposed, its kinda like stardew but with an extra dimension! I a little TGUI experience but I have never used JS to make something like that, though, so someone other then me would probably be better for doing that if thats what people wanted

RSG250:
as long as it doesnt FEEL slower than the current way of fishing, i wouldnt mind a mini game, also as long as it does NOT mean that failing it means no fish

Azrun:
> as long as it doesnt FEEL slower than the current way of fishing, i wouldnt mind a mini game, also as long as it does NOT mean that failing it means no fish

As someone who investigated trying to minigame this some time ago, these are good points.  The increased timing demands are going to really be problematic if the client has any tires is connection issues, as was had already been mentioned. 

Additionally part of the gamification is a bit antithetical to the original chill vibes intent of the feature, from my recollection in my conversations in this topic. So this should be something that is opted into in some way preferably.

Merlin1230:
> Additionally part of the gamification is a bit antithetical to the original chill vibes intent of the feature, from my recollection in my conversations in this topic. So this should be something that is opted into in some way preferably.

I dont think a minigame really subtracts from something being chill, honestly. Something can still be relaxing even if you have to do some input, and hell stardew's is still something relaxing, thats for sure

Azrun:
> > Additionally part of the gamification is a bit antithetical to the original chill vibes intent of the feature, from my recollection in my conversations in this topic. So this should be something that is opted into in some way preferably.
> 
> I dont think a minigame really subtracts from something being chill, honestly. Something can still be relaxing even if you have to do some input, and hell stardew's is still something relaxing, thats for sure

I agree with that, typically I need a certain amount of stimulus to really get something to click for me. What works for each person is going to be subjective. I was just trying to give some second hand historical context regarding my understanding of the original feature.

Merlin1230:
closing this for now, I haven't been too motivated recently and I get the feeling I wont be for a while as I take a bit of a hiatus

PR #14530, opened 2023-06-11
New syndicate item: Agent radio

Valtsu0:
[OBJECTS][FEATURE]
## About the PR 
Agent radio is a 1 cost syndicate and a 3 cost revolutionary item that is loosely a radio counterpart to the agent card.

When activating the agent radio, you choose the appearance of the radio itself and the icon of the radio on air.

It starts with access to the syndicate channel, but more secure channels can be scanned in similarly to the agent card.

## Why's this needed? 
This is an alternative for going to the listening post for secure communication with other antags or for revealing yourself as a syndicate agent.

It is a slight buff to revolutionaries, allowing the heads to coordinate with each other at a price, hopefully avoiding the situation where the last rev head plays recklessly and dies while 90% of the crew are revs, not realizing all other rev heads are dead.

Adds some gimmick possibilities and ways to confuse the crew when used with an agent card and a voice changer. For example, you can pretend to be an AI:
image

## Changelog

```changelog
(u)Valtsu0
(+)Added new traitor and revolutionary item: agent radio
```


Valtsu0:
Silicon have their own say verbs (E. g. states), so any eagle-eyed players will instantly know that someone is not an AI. There does not seem to be any easy way for a radio to change that and i do not think it is the agent radios job to change that. Instead, the voice changer should be able to do this. That should probably be it's own PR

PR #14537, opened 2023-06-11
Mechcomp Grabber Component

Skeletonman0:
[OBJECTS] [FEATURE]
## About the PR 
adds a mechcomp component that has three inputs:
* change direction (only accepts cardinal directions)
* pickup/drop item (admins can varedit what it considers pickupable)
* use item (only works with /obj/item)
  * this will smack an object infront of it with the item
  * emagging it disables the safety check which prevents it from attacking mobs if the force is more than 5 on the item
  * the grabber arm itself holds a mob inside used as the user for attacks made by the arm
* **cabinet banned is set to true on this, as moving death machines seemed a bit too wacky**

so far, ive seen that it cant use guns. it just beats you to death with them if its emagged

the code for this is slightly cursed

## Why's this needed? 
image
allows people to make assembly lines and play discount factorio

## Changelog 

```changelog
(u)Skeletonman0
(*)Added the Mechcomp Grabber Component, which allows engineers to pick up and poke things on an assembly line
```


code/modules/mechanics/MechanicMadness.dm
```diff
@@ -4129,6 +4129,199 @@ ADMIN_INTERACT_PROCS(/obj/item/mechanics/trigger/button, proc/press)
...
+	var/emagged = FALSE
+
+	New()
+		. = ..()
+		SEND_SIGNAL(src, COMSIG_MECHCOMP_ADD_INPUT, "Pickup/Drop held item", PROC_REF(do_pickup))
+		SEND_SIGNAL(src, COMSIG_MECHCOMP_ADD_INPUT, "Use held item", PROC_REF(do_use))
+		SEND_SIGNAL(src, COMSIG_MECHCOMP_ADD_INPUT, "Change Direction", PROC_REF(do_rotate))
+		src.UpdateIcon()
+
+		dummy_mob = new(src)
```
pali6, line 4157: Make an actual /mob/living subtype for this. Don't instantiate /mob

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -4129,6 +4129,207 @@ ADMIN_INTERACT_PROCS(/obj/item/mechanics/trigger/button, proc/press)
...
 		logTheThing(LOG_STATION, null, "draws on [src]: [log_loc(src)]: canvas{\ref[src], [x], [y], [dot_color], [x2], [y2]}")
 
+// exists just to allow the grabber to attack
+/mob/living/grabber_goblin
+	desc = "you shouldn't see this!"
+
+	New()
+		. = ..()
+		if (src.loc)
+			src.name = src.loc.name
```
pali6, line 4139:
```suggestion
	New(loc)
		. = ..()
		if (loc)
			name = loc.name
			real_name = loc.name
```

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -4129,6 +4129,207 @@ ADMIN_INTERACT_PROCS(/obj/item/mechanics/trigger/button, proc/press)
 		// so you can tell if scrimblo made a cool scene and then dogshit2000 put obscenities on top or whatever.
 		logTheThing(LOG_STATION, null, "draws on [src]: [log_loc(src)]: canvas{\ref[src], [x], [y], [dot_color], [x2], [y2]}")
 
+// exists just to allow the grabber to attack
```
pali6, line 4132:
```suggestion
/// exists just to allow the grabber to attack
```

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -4129,6 +4129,207 @@ ADMIN_INTERACT_PROCS(/obj/item/mechanics/trigger/button, proc/press)
...
+/mob/living/grabber_goblin
```
pali6, line 4133:
```suggestion
/mob/living/grabber_goblin
	name = "\improper Grabber Component"
```

pali6 (commented):
Are you ***absolutely*** sure the mob needs literally nothing modified from base /living?

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -4129,6 +4129,240 @@ ADMIN_INTERACT_PROCS(/obj/item/mechanics/trigger/button, proc/press)
...
+
+
+/obj/item/mechanics/grabber
+	// picks up items, can spin, and can smack something with it
+	name = "\improper Grabber Component"
+	desc = "Usable to grab items and create a mini assembly line"
+	icon_state = "grabber"
+	cabinet_banned = TRUE
+	plane = PLANE_DEFAULT
+	density = TRUE
```
TobleroneSwordfish, line 4176: A dense item you can pick up seems dangerous, I can imagine people using this to block stuff and cause jank. Either make it only dense when anchored or just not dense at all since I don't think any other mechcomp component is?

TobleroneSwordfish (changes requested):
the factory must grow

Bartimeus973:
Reopening with pali's approval.

deathrobotpunch:
add changes from leah and this is pretty much testmergeable imo

hobartbartholomewpepperidgejr:
the robot arms yearn for the mines

Metekillot:
Instead of having it only be able to target the last object on a tile, perhaps this can be changed to do that when a 'Target' signal is empty? If it attempts an attack or pickup with a target signal but no target item named so, it won't do anything. An empty target signal would behave as it does as currently written.

deathrobotpunch:
damned merge conflicts are delaying Factorio: SS13 EDITION

PR #14538, opened 2023-06-11
Spin emote now closes the bartender's sawn off

blackrep:
[Game Objects][Feature]

## About the PR 
This PR allows a user to use `*spin` or `*twirl` emotes to snap just the sawn off shotgun if the action is open. This also makes it so that if a user snaps the sawn off shotgun closed using these emotes, there is a chance they accidentally shoot themselves (existing gun code).

## Why's this needed? 
Silly flare, and increase the chance the ~bartender~ user accidentally shoots themselves trying to be cool.

## Changelog 

```changelog
(u)Blackrep
(+) The bartender's sawn off can now be spun closed using twirl emote
```

### DIFF
--- a/code/obj/item/gun/kinetic.dm
+++ b/code/obj/item/gun/kinetic.dm
@@ -2459,5 +2459,5 @@ ABSTRACT_TYPE(/obj/item/gun/survival_rifle_barrel)
 
 		playsound(user.loc, 'sound/weapons/gunload_click.ogg', 15, TRUE)
 
-		update_icon()
+		UpdateIcon()
 
--- a/code/obj/item/gun/kinetic.dm
+++ b/code/obj/item/gun/kinetic.dm
@@ -2453,3 +2453,9 @@ ABSTRACT_TYPE(/obj/item/gun/survival_rifle_barrel)
 	alter_projectile(obj/projectile/P)
 		. = ..()
 		P.proj_data.shot_sound = 'sound/weapons/sawnoff.ogg'
+
+	on_spin_emote(mob/living/carbon/human/user)
+		if(src.broke_open)
+			src.attack_self(user)
+			user.visible_message("<span class='alert'><b>[user]</b> snaps shut [src] with a [pick("spin", "twirl")]!</span>")
+		..()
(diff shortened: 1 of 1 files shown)


virvatuli-gh (approved):
:pray:

code/obj/item/gun/kinetic.dm
```diff
@@ -2453,3 +2453,9 @@ ABSTRACT_TYPE(/obj/item/gun/survival_rifle_barrel)
 	alter_projectile(obj/projectile/P)
 		. = ..()
 		P.proj_data.shot_sound = 'sound/weapons/sawnoff.ogg'
+
+	on_spin_emote(mob/living/carbon/human/user)
+		if(src.broke_open)
+			src.attack_self(user)
```
pali6, line 2459: doc comment on attack_self
image
```suggestion
			src.AttackSelf(user)
```
but also a sensible option would be to just put the code for doing this into a separate proc that's called from on_spin_emote and attack_self both.

pali6, line 2459: doc comment on attack_self
image
```suggestion
			src.AttackSelf(user)
```
but also a sensible option would be to just put the code for doing this into a separate proc that's called from on_spin_emote and attack_self both.

pali6 (changes requested):

code/obj/item/gun/kinetic.dm
blackrep: Moved the action toggling code into a shared proc that's called by both `on_spin_emote` and `attack_self`.

code/obj/item/gun/kinetic.dm
```diff
@@ -2453,3 +2441,23 @@ ABSTRACT_TYPE(/obj/item/gun/survival_rifle_barrel)
...
+			src.casings_to_eject = src.shells_to_eject
+
+			if (src.casings_to_eject > 0) //this code exists because without it the gun ejects double the amount of shells
+				src.ejectcasings()
+				src.shells_to_eject = 0
+		src.broke_open = !src.broke_open
+
+		playsound(user.loc, 'sound/weapons/gunload_click.ogg', 15, TRUE)
+
+		update_icon()
```
pali6, line 2462:
```suggestion
		UpdateIcon()
```

PR #14539, opened 2023-06-11
Make gas sharing transfer enough gas in order to attempt to equalize in one step

pali6:
## About the PR 
Before this PR gas amount transferred was the difference between the two gas mixtures divided by 5.
This PR makes it so enough gas is transferred that (ignoring other effects and gas transfers happening on the same tick) the two gas mixtures become equalized immediately. In practice this means that repressurization is faster but might also have various other side effects.

Making this PR largely to testmerge it and for people to post feedback on the change here.

(With the addition of this PR we *might* need to re-enable gas archiving because the directional artifacts can get a bit silly now. Shouldn't be a large performance impact.)

## Why's this needed? 
Repressurization being slow kinda sucks and the /5 constant looks pretty pulled out of someone's ass so changing it doesn't seem horrible. This should give feel of faster atmos.

## Changelog 

```changelog
(u)pali
(*)The ratio at which gas gets transferred in atmos got increased. This should make repressurization faster. Let me know if you see any positive or negative impacts of this experiment while it's being tested.
```

### DIFF
--- a/code/modules/atmospherics/FEA_gas_mixture.dm
+++ b/code/modules/atmospherics/FEA_gas_mixture.dm
@@ -271,7 +271,7 @@ What are the archived variables for?
 /datum/gas_mixture/proc/check_gas_mixture(datum/gas_mixture/sharer)
 	if (!sharer)
 		return SELF_CHECK_FAIL
-	#define _DELTA_GAS(GAS, ...) var/delta_##GAS = (GAS - sharer.GAS)/5;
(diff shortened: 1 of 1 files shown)


pali6:
@Azrun @Laboredih123 give opinions, alternatives and/or thoughts if you want

Laboredih123:
I'll see how it feels during testing.

Laboredih123:
oh yeah, it felt okay enough

Laboredih123:
oh

PR #14546, opened 2023-06-12
Halves strange green goops depletion rate and amount generated by ecto cooler decay

M-Earthfire:
[Chemistry][Balance]

## About the PR 

This PR sets the decay rate of strange green goop to the standard chem depletion rate of 0,4, down from 0,8 . To compensate and to keep the same effect, the generation of strange green goop for each unit of ecto cooler was halved as well.

## Why's this needed? 

Strange green goop has a recipe that is hard to mass produce and has the problem of decaying too fast to properly pull off gimmicks with it. The change of depletion rate brings it in line with other gimmick/utility chems. 

Because this affects the visual effect of ecto cooler, the amount created was adjusted as well to keep the effect equal.

This means this change is mostly targeted towards folks who try to create the chem directly.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Strange green goop decay rate was halved. The amount of strange green goop created by ecto cooler was halved as well.
```


Tarmunora:
Ghostvision is a fairly powerful and unique effect - going to want some reason a bit more convincing, or at least less vague, than "[to more easily] pull off gimmicks with it."

M-Earthfire:
If there are balance concerns behind that this is absoluely fair. I haven't considered these, really.

I rather considered it a gimmick than sonething that has a tangible impact of the round, which makes the chem depletion rate look out of place in comparison to other minor gimmick chems (e.g. simethicone). Thinking about it from a balance angle in hindsight, that is certainly not true.

What i hadn't considered, being able to see dead people that are on the manifest can spark investigations, especially when i take the knowledge and vision ghosts naturally have into account. Using them as scouts or for valid hunting is a thing that can happen, even though the rules prevent parts of it (not sure if that applies to classic?)

The motivation i had for the PR was my perceived lack of actual usage by players and. Given it is a very unique effect, the thought was to gice it a nudge so it eill be trued out more often.

I still think it could work with a 0.4 depletion rare, but i am seriously looking from an RP-google on that. So if you think it will get abused too strongly with a lower depletion rate, feel free to close the PR

PR #14555, opened 2023-06-13
New 3x5 Random Room: Penguinroom_90

DanaTarSUS:
## About the PR 

Contest entry, adds a new random room featuring a snowy room with a few shrubs, a snowy tree, and two penguins named Earl and Steven!

## Why's this needed? 

Penguins good, random room also good, satisfactory combination methinks


amylizzle:
It'd be good to add a picture of the room, either in map editor, game, or (ideally) both.

DanaTarSUS:
> It'd be good to add a picture of the room, either in map editor, game, or (ideally) both.

You make an excellent point, here is the map as previewed in StrongDMM as well as ingame 
image
image

PR #14556, opened 2023-06-13
Constructable conveyor belts

CaioFrancisco:
## About the PR 

It is what it says on the tin. Also, the design docs are in https://hackmd.io/@ASO4nZ-8S9eoQdhojH1tYg/ByhMVBZP2.

## Why's this needed? 

To keep the game's design of being almost a sandbox: everything is constructible, modifiable or deconstructable.

## Changelog 

```changelog
(u)Caio029
(*)Conveyor belts are now buildable! Use a device analyzer on a conveyor belt and a conveyor switch, then go ham. See the help message on a conveyor belt for more information.
```


M-Earthfire:
Is there a reason for the welding step with a 5/2 second action bar? The decon tool is already a real pain to deconstruct stuff with and a very long sction bar on top if it is kinda excessive. Especially for an object that would, for rather common changes, require the removal amd construction of belts in the total number of 20+

CheffieGithub:
I still think it would make more sense to use a screwdriver for protection, then wrench for unanchoring or crowbar for turning in 4 directions. This should ease editing of conveyors after placement. 

If you have concerns over the sabotage of belt hell then I would make a subtype/variable for whether it needs welding. (In this case I would have the 'panel' or such you unscrew be welded into place.)

I also think a way of telling the direction of a conveyor when off would be nice.

Glamurio:
> I still think it would make more sense to use a screwdriver for protection, then wrench for unanchoring or crowbar for turning in 4 directions. This should ease editing of conveyors after placement.

I agree with this. It's also a unanimous design decision that opening panels is done with a screwdriver, and wrenching is for anchoring. I'd be okay with the idea of welding the belt _after_ it has been secured. That way, people can choose to make the belt more secure, but they don't have to.

CaioFrancisco:
> Is there a reason for the welding step with a 5/2 second action bar? The decon tool is already a real pain to deconstruct stuff with and a very long sction bar on top if it is kinda excessive. Especially for an object that would, for rather common changes, require the removal amd construction of belts in the total number of 20+

I don't think it's as bad as you may think. If an engineer were to build a conveyor system of 50 belts, they would take only 100 seconds (less than 2 minutes) to lock the panel of all of them. And, by design, I'm making it so Engineers have more access into building conveyors than normal crew.

> I still think it would make more sense to use a screwdriver for protection, then wrench for unanchoring or crowbar for turning in 4 directions. This should ease editing of conveyors after placement.

That's fair. I'll commit that in once I get some time to code.

M-Earthfire:
>  And, by design, I'm making it so Engineers have more access into building conveyors than normal crew.

That is already the case because of the restriction to the mechscanner and the decon tool.

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -402,6 +409,85 @@
...
+		if(!istype(over_object, /obj/machinery/conveyor_switch)) return ..()
+		var/obj/machinery/conveyor_switch/new_switch = over_object
+		src.id = new_switch.id
+		src.linked_switches += new_switch
+		new_switch.conveyors += src
+		usr.show_text("You connect \the [new_switch] to the [src].", "blue")
+
+	return ..()
+
+/obj/machinery/conveyor/MouseDrop_T(dropped, user, src_location, over_location) // Pretty much a copy-paste. Both procs are almost identical.
```
CaioFrancisco, line 455: Is there a way to avoid copy-pasting? I *thought* about making a separate proc that handles both, but that just sounds like bad practice.

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -763,6 +903,42 @@ ADMIN_INTERACT_PROCS(/obj/machinery/conveyor_switch, proc/trigger)
...
+		while (new_id in switch_ids) // If we somehow reach a state where there's already a conveyor switch with the same ID, we loop through 0 to BYOND's max integer value and check if it is unique.
+			src.id = new_id + "[id_disambiguation_number]"
+			id_disambiguation_number++
+
+		src.switch_ids += src.id
+
+	attackby(var/obj/item/I, mob/user)
+		if (ispulsingtool(I))
+			var/datum/component/mechanics_connector/connector = I.GetComponent(/datum/component/mechanics_connector)
+			if (!connector) return ..()
```
CaioFrancisco, line 935: `if (!istype(connector)) return ..()`

CaioFrancisco (commented):
self code-review yaaaaaaay (please, any dev, do a proper code review.)

CaioFrancisco:
Looks good enough for a review.

CaioFrancisco:
Since the PR is getting close to the second page, and you were the dev who most gave attention to it, I'm requesting you for review.

CaioFrancisco:
Alright, that's it, it actually looks presentable now.

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -18,12 +18,16 @@
 	anchored = ANCHORED
 	power_usage = 0
 	layer = 2
+	mouse_drag_pointer = MOUSE_ACTIVE_POINTER
 	machine_registry_idx = MACHINES_CONVEYORS
-	var/operating = OP_OFF	// 1 if running forward, -1 if backwards, 0 if off
+	mechanics_type_override = /obj/machinery/conveyor/built
+	var/operating = CONVEYOR_STOPPED	// 1 if running forward, -1 if backwards, 0 if off
 	var/operable = TRUE	// true if can operate (no broken segments in this belt run)
```
ZeWaka, line 25: code docs

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -33,7 +37,7 @@
 	var/obj/machinery/conveyor/next_conveyor = null
 	event_handler_flags = USE_FLUID_ENTER
 	/// list of conveyor_switches that have us in their conveyors list
-	var/list/linked_switches
+	var/list/linked_switches = list()
```
ZeWaka, line 40: initialize this in new

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -764,6 +925,61 @@ ADMIN_INTERACT_PROCS(/obj/machinery/conveyor_switch, proc/trigger)
 				C.setdir()
 		SEND_SIGNAL(src,COMSIG_MECHCOMP_TRANSMIT_SIGNAL,"switchTriggered")
 
+/obj/machinery/conveyor_switch/built/
+	desc = "A conveyor control switch. This one looks like it was built recently."
+	var/static/list/switch_ids
```
ZeWaka, line 930: i don't see where this list is used really
does it only exist because if switches are built on the same tick the TIME would be the same?

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -764,6 +925,61 @@ ADMIN_INTERACT_PROCS(/obj/machinery/conveyor_switch, proc/trigger)
...
+		var/id_disambiguation_number = 0
+		while (new_id in switch_ids) // If we somehow reach a state where there's already a conveyor switch with the same ID, we loop through 0 to BYOND's max integer value and check if it is unique.
+			src.id = new_id + "[id_disambiguation_number]"
+			id_disambiguation_number++
+
+		src.switch_ids += src.id
+
+	MouseDrop_T(dropped, mob/user)
+		if (ispulsingtool(dropped))
+			if (!istype(dropped, /obj/machinery/conveyor_switch)) return ..()
```
ZeWaka, line 956: indent pls

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -670,20 +826,25 @@
...
+	mouse_drag_pointer = MOUSE_ACTIVE_POINTER
 	/// current direction setting
 	var/position = CONVEYOR_STOPPED
 	/// last direction setting
 	var/last_pos = CONVEYOR_REVERSE
 	// Checked against conveyor ID on link attempt
 	var/id = ""
 	/// the list of converyors that are controlled by this switch
-	var/list/conveyors
+	var/list/conveyors = list()
```
ZeWaka, line 847: init in new pls

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -764,6 +925,61 @@ ADMIN_INTERACT_PROCS(/obj/machinery/conveyor_switch, proc/trigger)
...
+			connected_switch.conveyors = src.conveyors
+			user.show_text("You connect \the [src] to \the [dropped]. Both share the same ID now.", "blue")
+
+		return ..()
+
+	attackby(var/obj/item/I, mob/user)
+		if (ispulsingtool(I))
+			var/datum/component/mechanics_connector/connector = I.GetComponent(/datum/component/mechanics_connector)
+			if (!connector) return ..()
+			if (!istype(connector.connectee, /obj/machinery/conveyor_switch)) return ..()
```
ZeWaka, line 972: indent

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -495,6 +626,31 @@
...
+				if (direction == turn(neighbor_conveyor.dir_in, 180)) // Is it contrary to a dir_in of another conveyor belt?
+					favorable_dir_outs += direction // Great! A favorable dir_out.
+				if (direction == turn(neighbor_conveyor.dir_out, 180)) // Is it contrary to the dir_out of another conveyour belt?
+					favorable_dir_ins += direction // Great! A favorable dir_in.
+
+		if (length(favorable_dir_ins) > 0) src.dir_in = favorable_dir_ins[1] // If there are any favorable dir ins, pick the first index.
+		if (length(favorable_dir_outs) > 0) src.dir_out = favorable_dir_outs[1]
+
+		if (!src.dir_in) src.dir_in = NORTH // In the case there's no favorable dir_ins, pick north.
+		if (!src.dir_out) src.dir_out = turn(src.dir_in, 180) // In the case there's no favorable dir_outs, pick the contrary of dir_in.
```
ZeWaka, line 650: indent

code/WorkInProgress/recycling/conveyor.dm
ZeWaka: copy paste from where

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -402,6 +408,87 @@
...
+
+		if (dir_target_atom in ordinal)
+			dir_target_atom &= (WEST | EAST)
+
+		PLAYER_SET_DIR_IN(dir_target_atom, src)
+		src.update()
+		return
+
+	if (ispulsingtool(equipped_object))
+		if(!istype(dropped, /obj/machinery/conveyor_switch)) return ..()
```
ZeWaka, line 481: i am going to stop commenting indent, pls just do it everywhere

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -402,6 +408,87 @@
...
 		walk(AM, 0)
 
+/obj/machinery/conveyor/get_desc()
+	if (src.deconstructable)
+		. += " <span class='notice'>It's cover seems to be open.</span>"
+
+
+// Swap directions if player is trying to set them to the same direction
+#define PLAYER_SET_DIR_IN(dir, conveyor) if (dir == conveyor.dir_out) { conveyor.dir_out = conveyor.dir_in}; conveyor.dir_in = dir
+#define PLAYER_SET_DIR_OUT(dir, conveyor) if (dir == conveyor.dir_in) { conveyor.dir_in = conveyor.dir_out}; conveyor.dir_out = dir
```
ZeWaka, line 418: these are only used once why are they defines

ZeWaka (changes requested):

code/WorkInProgress/recycling/conveyor.dm
CaioFrancisco: `mouse_drop` and `MouseDrop_T` are almost the same. I thought about one calling the other, but that makes the current implementation of changing conveyor dirs almost impossible without jank.

code/WorkInProgress/recycling/conveyor.dm
CaioFrancisco: Was there before I touched the code, but alright.

code/WorkInProgress/recycling/conveyor.dm
CaioFrancisco: Yeah. On hindsight, that's almost impossible from happening. I'll delete the code, tell me if it should be put back again;

CaioFrancisco: Yeah. On hindsight, that's almost impossible from happening. I'll delete the code, tell me if it should be put back again.

CaioFrancisco:
Did indents, did DMDocs, and deleted not very useful code. Tell me if it needs any more changes.

code/WorkInProgress/recycling/conveyor.dm
ZeWaka: I mean, it could still happen. I don't see how it could happen though with this current implementation. It could be solved cleaner if it's an issue.

CaioFrancisco: Probably won't happen, and it is easily fixable if it does happen (two switches with the same ID can, and is supposed to happen in the PR).

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -18,18 +18,27 @@
...
+	/// The direction the conveyor is going to. 1 if running forward, -1 if backwards, 0 if off
+	var/operating = CONVEYOR_STOPPED
+	/// true if can operate (no broken segments in this belt run)
+	var/operable = TRUE
+	var/dir_in = NORTH
+	var/dir_out = SOUTH
 	var/currentdir = SOUTH
-
-	var/id = ""			// the control ID	- must match controller ID
+	var/deconstructable = FALSE
```
TobleroneSwordfish, line 31: this is confusingly named, doc comments would help

TobleroneSwordfish (changes requested):
- Attempting to set the input and output to the same direction results in the cool and hip `Fuck` sprite:
image
Other than that, I really like the rotation mechanics, takes a bit to get used to but very intuitive once you've grasped it.
- There's no visual indication of the panel being open or shut, I know the panel was an anti-griefing measure but it would be much harder to fix a line of conveyors where one of them isn't working for invisible reasons vs fixing a line of conveyors where one is obviously pointing the wrong way.
- The action bar for unscrewing the panel is excessively long, it could be less than a second and still serve the same purpose (ie preventing people just running down a line of conveyors whacking them with a screwdriver)
- The changelog doesn't mention that you need a multitool to link conveyors to switches.
- The instructions on how to modify conveyor belts would be good for the new `get_help_message` system, since they're pretty arcane (by necessity)

sry 4 wall of text, good PR thank u 👍

CaioFrancisco:
> - Attempting to set the input and output to the same direction results in the cool and hip `Fuck` sprite:
> image

I thought I had fixed that, but oh well. I'll debug some more and see what is going on

> - There's no visual indication of the panel being open or shut, I know the panel was an anti-griefing measure but it would be much harder to fix a line of conveyors where one of them isn't working for invisible reasons vs fixing a line of conveyors where one is obviously pointing the wrong way.

I **knew** there was something missing on the PR that I was planning on doing. I hope there's a sprite that fits the criteria already.

> - The action bar for unscrewing the panel is excessively long, it could be less than a second and still serve the same purpose (ie preventing people just running down a line of conveyors whacking them with a screwdriver)

Fair, half of a second for engineers it will be. And probably 2 seconds for non-engineers.

> - The instructions on how to modify conveyor belts would be good for the new `get_help_message` system, since they're pretty arcane (by necessity)

I had thought about that, but got too lazy. Will do.

> sry 4 wall of text, good PR thank u 👍 

It's great to get some input on my PRs, dw about that.

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -170,6 +183,17 @@
...
+/obj/machinery/conveyor/get_help_message(dist, mob/user)
+	if (src.deconstructable)
+		return {"To change the conveyor belt directions, you must use a <b>crowbar</b> or any other prying item.
+		Click-drag anything into the conveyor belt, and the direction into it will change to that direction.
+		Click-drag the conveyor belt to anything, and the direction out of it will change to that direction.
+		Click-drag a conveyor belt to another to automatically assign each one's direction.
+		To change the conveyor belt's direction, link it to a conveyor belt switch through a <b>mulitool</b> by click-dragging one to another.
+		To close the conveyor belt's cover and make it operational again, simply use a <b>screwdriver</b> on it."}
+	else
+		return "To open the conveyor belt's cover and make any changes to it, use a <b>screwdriver</b> on it. For more instructions on it, see the help message with it's cover open."
```
TobleroneSwordfish, line 195:
```suggestion
		return "To open the conveyor belt's cover and make any changes to it, use a <b>screwdriver</b> on it."
```

TobleroneSwordfish (changes requested):
Looks good, just a couple of minor things. Belts keep their ID when deconstructed (which is probably good?) but this means that you can have switches on the other side of the map or across z-levels which is kind of jank. Ideally you'd solve this by checking if belts are in a chain which is connected to a switch, but a decently long range check would probably be fine.

CaioFrancisco:
I just nulled all the important variables upon being built from a frame. Should fix weirdness.

TobleroneSwordfish (approved):
I've pushed some slightly better conveyor arrow sprites to master so we're not using the crusty MSpaint ones at least

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -495,6 +660,40 @@
...
+			src.dir_in = favorable_dir_ins[1] // If there are any favorable dir ins, pick the first index.
+		if (length(favorable_dir_outs))
+			src.dir_out = favorable_dir_outs[1]
+
+		if (!src.dir_in)
+			src.dir_in = NORTH // In the case there's no favorable dir_ins, pick north.
+		if (!src.dir_out)
+			src.dir_out = turn(src.dir_in, 180) // In the case there's no favorable dir_outs, pick the contrary of dir_in.
+
+		. = ..()
```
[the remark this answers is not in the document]
zjdtmkhzt: example: https://streamable.com/on1645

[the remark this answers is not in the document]
CaioFrancisco: gonna debug at home.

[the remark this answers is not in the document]
zjdtmkhzt: The conveyors themselves seem very cool and chill to build, even if the time to screw them in is a tad long for non-engineers, but I guess that's the cost of not being an engineer. I am not sure if I agree with that trait design, but that's not really to do with this PR.^^

What seemed kind of annoying was connecting the switches to the belts, having to clickdrag for every single one seems pretty annoying? Clickdragging things with small sprites (like switches) is a relatively intensive thing to do, especially repeatedly. Also it's kind of confusing because it uses the exact same message as linking via something like mechanics connecting? This still happens if you use the Connect Component function of the multitool, even if the panel is closed.

Also, my conveyor switches seem to not be the /built subtype you created, so I am not sure if the above paragraph is a bug? Ideally I think it would be nice if the conveyor switches could be connected to the belts via the Connect Component function of the multitool as well. I am not entirely sure if this was already the intent and it's just bugged?
image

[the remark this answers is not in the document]
zjdtmkhzt: Also, at some point it could also be cool to allow for scanning and making the oshan carousel conveyors and carousel unit, but that can certainly be handled in a later PR. :D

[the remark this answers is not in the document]
CaioFrancisco: Sorry for the enormous delay, @zjdtmkhzt, but I did the few fixes. The conveyor switch thing is really a bug, Connect Component was supposed to work with conveyors.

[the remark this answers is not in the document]
zjdtmkhzt: It's no problem to take a few days to get to making review changes!^^
I'll take a look at it later :3

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -764,6 +970,56 @@ ADMIN_INTERACT_PROCS(/obj/machinery/conveyor_switch, proc/trigger)
...
+		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_INPUT,"trigger", PROC_REF(trigger))
+		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_CONFIG,"Refresh ID", PROC_REF(get_new_id))
+		src.get_new_id()
+
+	proc/get_new_id()
+		src.id = TIME // Create a new ID on the fly. It doesn't need to be readable, it just needs to be unique.
+		src.conveyors = list()
+
+	MouseDrop_T(dropped, mob/user)
+		if (ispulsingtool(dropped))
```
zjdtmkhzt, line 992: This checks if the item being dropped is a multitool, perhaps you wanted something like this instead?
```suggestion
		if (user.find_tool_in_hand(TOOL_PULSING))
```

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -764,6 +970,56 @@ ADMIN_INTERACT_PROCS(/obj/machinery/conveyor_switch, proc/trigger)
...
+
+			var/obj/machinery/conveyor_switch/connected_switch = connector.connectee
+			src.id = connected_switch.id
+			connected_switch.conveyors = src.conveyors
+			for (var/obj/machinery/conveyor/conveyor in src.conveyors)
+				conveyor.id = connected_switch.id
+				conveyor.linked_switches += connected_switch
+			user.show_text("You connect \the [src] to \the [connector.connectee]. Both share the same ID now.", "blue")
+
+		return ..()
```
zjdtmkhzt, line 1022: This does not seem to be usable for me, since the mechanics-component-connecting-component triggers before it and then cancels the thing?

It just opens the normal "link devices" popup for mechanics components. Not sure what the best way to solve this would be...

zjdtmkhzt (changes requested):

code/WorkInProgress/recycling/conveyor.dm
CaioFrancisco: The best solution to this is just not doing it at all. I'll include the change in the changelogs.

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -764,6 +970,38 @@ ADMIN_INTERACT_PROCS(/obj/machinery/conveyor_switch, proc/trigger)
...
+		UnsubscribeProcess()
+		START_TRACKING
+		UpdateIcon()
+		AddComponent(/datum/component/mechanics_holder)
+		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_INPUT,"trigger", PROC_REF(trigger))
+		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_CONFIG,"Refresh ID", PROC_REF(get_new_id))
+		src.get_new_id()
+
+	proc/get_new_id()
+		src.id = TIME // Create a new ID on the fly. It doesn't need to be readable, it just needs to be unique.
```
TobleroneSwordfish, line 988: this is not a good way to get a unique ID, I'd advise using ref instread

code/WorkInProgress/recycling/conveyor.dm
```diff
@@ -764,6 +970,38 @@ ADMIN_INTERACT_PROCS(/obj/machinery/conveyor_switch, proc/trigger)
 				C.setdir()
 		SEND_SIGNAL(src,COMSIG_MECHCOMP_TRANSMIT_SIGNAL,"switchTriggered")
 
+/obj/machinery/conveyor_switch/built/
```
TobleroneSwordfish, line 973: Why is this a separate type? Also the clicking on one and then the other mechcomp behaviour still works and overrides the panel openness.

TobleroneSwordfish (changes requested):

code/WorkInProgress/recycling/conveyor.dm
CaioFrancisco: So it can spawn with it's cover open. Will try to fix that as soon as I get home.

code/WorkInProgress/recycling/conveyor.dm
CaioFrancisco: Not sure what you mean by "ref" @TobleroneSwordfish

CaioFrancisco: Oh, wait, you mean the memory address for the atom? That makes sense.

ZeWaka: `ckey("\ref[src]")` or so - computers do this

CaioFrancisco:
Fixed conveyors being able to be linked with it's cover closed, because I accidentaly made it so only conveyors with closed cover could be connected by the mechcomp way. Also changed how IDs are done, by the way ZeWaka recommended to do.

PR #14574, opened 2023-06-13
Add escape pods to Donut2

Glamurio:
[Feature] [QoL]
## About the PR 
Reworks escape to have a more open feel to it, I tried to keep the "airport bistro-esque" feel to it by adding a coffee machine and some mugs for people to enjoy as they wait for escape. To the West, the science lobby has been expanded with a similar feel and to seem a little less crowded.

image

Moved the monkey wrench to a different part (I don't know if it's relevant), also added a broken escape pod room with a MiniPutt instead to the West of crew quarters (formerly an airlock exit)
Additionally, because of the restructuring of the maintenance around it, I squeezed in a new 3x3 random room.

image

## Why's this needed? 
Donut2 is an older map, this change will bring it more in-line with other modern maps. Having access to escape pods is important to incentivize heroic rescue missions and reward the risk of staying behind.

Also, random maps gud.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Donut 2 Escape has been refurbished to include a larger area and escape pods.
(+)Donut 2 Robotics and adjacent maintenance has received an overhaul.
```


Sord213:
Not a fan of this implementation. The main escape pod area feels just really slapped together. It also makes a weird voidspace that will get blocked in when the shuttle docks between the docking point where you board the shuttle and the escape pod dock.

Screenshot_377
I also don't like narrowing down maintenance to have a single tiles width (red highlighted area). There's also the misclick tiles in space (yellow)

Glamurio:
> I also don't like narrowing down maintenance to have a single tiles width (red highlighted area). There's also the misclick tiles in space (yellow)

Very valid criticism (oops misclick)

Anyhow, here's a new updated version. I've changed the layout of robotics to accomodate the wider hallway and still squeeze in the random room. Whether or not we'll keep the pod at the top left is optional (I'm fond of it)

image

I'll work on a more elaborate restructuring of the escape area which will now aim North (kudos to CalliopeSoups for the mockup), if that's desirable.

image

Glamurio:
Alright, so this is the new and revised escape.

image

It has a more open feel to it, I tried to keep the "airport bistro-esque" feel to it by adding a coffee machine and some mugs for people to enjoy as they wait for escape. To the West, the science lobby has been expanded with a similar feel and to seem a little less crowded.

The escape pods have been placed to the side, and at the end of each hallway toward escape.

I forgot to mention it earlier, but I moved the monkey wrench to a different part (I don't know if it's relevant), also, because I now have 4 pods at escape, I just replaced the former escape pod with a MiniPutt and a broken down bay. It's just flavor, though if it bothers anyone I'll remove it.

image

Sord213:
I'll review this when I can but don't forget to change the escape shuttle dir in code/map.dm

Sord213 (changes requested):
Screenshot_382
This APC/Light won't work, the light will be covering the apc when the offsets are applied automatically.
Screenshot_383
These airbridge controllers are A) Not configured correctly, one half is labeled "escape" and the other is "mining" and B) wholly unnecessary. You don't need airbridge controllers if there's no gap to the shuttle. In fact configured as is they might runtime when the shuttle arrives or if someone tries to extend them manually lol. Just remove em
Screenshot_384
Please change this door out to an external airlock so it's obvious from maint it's an airlock.
Screenshot_385
Errant runway light
Screenshot_386
Could you just.... shift the whole thing down a bit so there's less dead space below the shuttle?

Glamurio:
Implemented all the changes. On the note of airbridge controllers, it didn't occur to me at first they had a function, by the time I realized, it was too late. I mostly wanted to use them for aesthetic purposes, but oops!

I also added a bunch more lighting because some spots felt a bit glum. Also blinds around the science shuttle lobby for when you want to have your lunch in peace.

Sord213 (changes requested):
Screenshot_388
This fire alarm needs to be moved to not be floating. Moved, not deleted as its the only fire alarm in robotics.
Screenshot_389
Misclick item on the wall/rack
Screenshot_390
This poster should be on the tile below and offset to be positioned on the wall.

PR #14584, opened 2023-06-14
Cleans up pipe_network.dm and pipeline.dm

Laboredih123:
[ATMOSPHERICS][INTERNAL][CLEANLINESS]
## About the PR 
Moves the pipe files to the atmospherics module, dmdocs them, optimises them, and in general just cleans them up.

## Why's this needed? 
Slowness bad, ugliness bad, speed good, cleanliness good.


code/modules/atmospherics/pipe_network.dm
```diff
@@ -1,49 +1,48 @@
...
-		for(var/obj/machinery/atmospherics/machine in normal_members)
-			machine.network_disposing(src)
-		normal_members.len = 0
-	normal_members = 0
-	if (line_members)
-		for(var/datum/pipeline/member in line_members)
-			member.network = null
-		line_members.len = 0
+
+	for(var/obj/machinery/atmospherics/machine as anything in normal_members)
```
ZeWaka: what happens when one of them blows up? you'll get a null in this list

Laboredih123: Tested, the exploding normal_member calls dispose on the pipe_network while still existing during disposing(), meaning there will never be a null in this list, hopefully.

ZeWaka: What about any of the other lists you are skipping the typechecking for?

Laboredih123: air_disposing_hook() is used by normal members for removing gases from /datum/pipe_network/var/list/gases, pipeline disposing removes its gases and itself from the pipe_network, and pipelines are disposed whenever a member disposes.

PR #14586, opened 2023-06-14
Update Centcom Viewer to allow searching by key

Sovexe:
## About the PR 
* Updates the centcom viewer tool to allow admins to input any key for lookup

image

## Why's this needed? 
* The current version is limited to only looking up a player directly through their player options panel


tgui/packages/tgui/interfaces/CentComViewer.js
```diff
@@ -121,7 +121,16 @@ export const CentComViewer = (props, context) => {
...
           <Stack.Item>
             <Section
-              title={`Ban Data for ${key}`}
+              title={
+                <>
+                  {`Ban Data for `}
+                  <Input
+                    width="50vw"
+                    value={key}
+                    onChange={(e, value) => act('updateKey', { value })}
```
ZeWaka, line 130: This will result in a change every time a character is typed, right? Could be a bit funky with the API calls. Consider a submit button?

Sovexe: No that's the behavior of OnInput, OnChange is only on submission
```markdown
- `onChange: (e, value) => void` - An event, which fires when you commit
the text by either unfocusing the input box, or by pressing the Enter key.
- `onInput: (e, value) => void` - An event, which fires on every keypress.
 ```

ZeWaka (approved):

PR #14596, opened 2023-06-15
Diversify valid reactor coolants

amylizzle:
## About the PR 
This does a couple major things:
1 - It rebalances the heat capacity of several of the gasses. Notably it reduces plasma's heat capacity by 25% and basically doubles a few of the other gasses. In terms of gameplay, this will have almost no effect outside of the reactor and TEG.
2 - It adds some more effects to gasses in the reactor to make the choice of coolant dependant on the operating conditions of the reactor, so plasma is not always the best choice for all situations. See changelog for details.

Minor thing: changed the available gas cans on Nadir and Clarion. I had to slightly tweak Nadir's engineering room to fit this.

## Why's this needed? 
Goon's atmos tech stuff is kinda weak, and always having plasma be the best coolant is boring.

## Changelog 

```changelog
(u)Amylizzle
(*)NT Research has revealed that Plasma can become unstable in the nuclear reactor. Please consider using other coolants. (See the minor changelog for details)
(+)Plasma's heat capacity has been reduced by 25%
(+)Oxygen, Nitrogen, N2O and CO2's heat capacities have been doubled (or more)
(+)In the nuclear reactor, Plasma will now produce lots of neutrons and decay more quickly into Fallout. 
(+)In the nuclear reactor, CO2 will act like a control rod, absorbing neutrons and slightly increasing in temperature.
(+)In the nuclear reactor, Fallout will occasionally decompose into a random gas.
```

### DIFF
--- a/_std/macros/atmos.dm
+++ b/_std/macros/atmos.dm
@@ -221,13 +221,13 @@ What can break when adding new gases:
 	TEG stats computer will ignore your new gas. Feel free to add it to reactor_stats.dm manually but good luck.
 */
 
-#define SPECIFIC_HEAT_PLASMA	200
-#define SPECIFIC_HEAT_O2		20
-#define SPECIFIC_HEAT_N2		20
-#define SPECIFIC_HEAT_CO2		30
+#define SPECIFIC_HEAT_PLASMA	150
+#define SPECIFIC_HEAT_O2		40
+#define SPECIFIC_HEAT_N2		50
+#define SPECIFIC_HEAT_CO2		40
 #define SPECIFIC_HEAT_FARTS 	69
-#define SPECIFIC_HEAT_RADGAS 	20
-#define SPECIFIC_HEAT_N2O		40
+#define SPECIFIC_HEAT_RADGAS 	5
+#define SPECIFIC_HEAT_N2O		60
 #define SPECIFIC_HEAT_AGENTB	300
 
 #define _APPLY_TO_GASES(PREF, SUFF, MACRO, ARGS...) \
--- a/code/obj/nuclearreactor/nuclearreactor.dm
+++ b/code/obj/nuclearreactor/nuclearreactor.dm
@@ -60,7 +60,7 @@
 	/// INTERNAL DEBUG: tracks total stored thermal energy in the coolant
 	VAR_PRIVATE/_last_total_coolant_e = 0
 	/// INTERNAL DEBUG: set to true to output debug messages
-	VAR_PRIVATE/_debug_mode = FALSE
+	VAR_PRIVATE/_debug_mode = TRUE
 
 	New()
 		. = ..()
--- a/code/obj/nuclearreactor/reactorcomponents.dm
+++ b/code/obj/nuclearreactor/reactorcomponents.dm
@@ -287,6 +287,8 @@ ABSTRACT_TYPE(/obj/item/reactor_component)
 	var/datum/gas_mixture/air_contents
 	gas_volume = 100
 	thermal_mass = 420*50//specific heat capacity of steel (420 J/KgK) * mass of component (Kg)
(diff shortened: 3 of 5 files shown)


Laboredih123:
the previous values for oxy, nitrogen, and co2 were based on real life values but eh, we live in spess so it doesnt matter.

amylizzle:
yeah I looked them up, and tried to preserve the general ratios but plasma's ludicrous capacity just ruins everything.

ZeWaka:
How does this affect the teg?

amylizzle:
From what I've tested, with the caveat that I'm not super great at setting up the TEG
- Standard furnace and chamber burns seem pretty much unaffected, though I'd expect a more experienced TEG engineer would notice the difference 
- I'd expect you'd see a 25% drop on max energy, in accordance with the 25% drop in heat capacity. A slightly slower ramp up as well. 
- Pipe burns I can't even guess at, since they generally end up in super weird states anyway. 

This is probably a test-merge candidate, but I wouldn't expect significant differences in the function of the TEG

Sbmhawk:
What about TTV's, single tank bombs, and canbombs?

amylizzle:
> What about TTV's, single tank bombs, and canbombs? (Mainly concerned with STB's, they are already pipebomb sized)

Those all work on gas temperature, not heat capacity. They should be unaffected.

Glamurio:
Big fan of this. When I started familiarizing myself with atmospherics, and the general answer was "always use plasma", I kind of quickly lost interest in delving deeper into the system. My only concern is about the power decrease of the TEG, but it's probably fine.

M-Earthfire:
There are 2 things this will affect as well:

Firstly, this will make plasma heat up faster. Noticeable for the burn chamber.

Secondly, this majorly screws up TTV's. I would need to make some calculations and tests, but TTV's are based of the mixing of plasma and oxygen with very high temperature differential and the subsequently starting reaction.

While i think the decreased energy capacity of plasma would only cause slight changes, the doubling of oxygen heat capacity is a major problem. It will require far more energy to heat up the oxygen to get a sufficient reaction rate and reach proper pressures.

I will whip up my calculation sheet once i come hone and throw up some results.

amylizzle:
https://cdn.discordapp.com/attachments/1023694631118057472/1119213887477272707/image.png

TTVs work fine. The math for maxcap is likely subtly different, and oxygen pressure seems to matter more, but otherwise fine.

M-Earthfire:
From my tests and calculations the ranges for a maxcap TTV got  more precise, but that is certainly nothing too bad. 

Test-TTV

_std/macros/atmos.dm
```diff
@@ -221,13 +221,13 @@ What can break when adding new gases:
...
-#define SPECIFIC_HEAT_CO2		30
+#define SPECIFIC_HEAT_PLASMA	150
+#define SPECIFIC_HEAT_O2		40
+#define SPECIFIC_HEAT_N2		50
+#define SPECIFIC_HEAT_CO2		40
 #define SPECIFIC_HEAT_FARTS 	69
-#define SPECIFIC_HEAT_RADGAS 	20
-#define SPECIFIC_HEAT_N2O		40
+#define SPECIFIC_HEAT_RADGAS 	5
+#define SPECIFIC_HEAT_N2O		60
```
Azrun, line 230: What was the basis for the new numbers? I think I've seen some take since showing similar specific heart values for some of the gases.

amylizzle: the basis was the isochoric heat capacity of the various gasses doubled and rounded to the nearest multiple of 10, with a bit of fudging to add some variety. It was not very scientific.

I've been kinda thinking maybe it would've been better to just nerf plasma all the way down to 100, instead of doubling the others, but that's a pretty huge nerf.

PR #14612, opened 2023-06-17
Adds a clipboard to detective lockers

RubberRats:
[GAME OBJECTS] [QOL]
## About the PR 

This PR adds a clipboard to the contents of all detective lockers.

## Why's this needed? 

Having a tool to organize notes and forensic printouts would be useful for detectives, especially for RP. Having them all in one place and easily editable would be nice for putting together a case.

## Changelog 

```changelog
(u)RubberRats
(+)The detective now has a clipboard in their locker.
```


frawhst (approved):
Sure

PR #14633, opened 2023-06-18
Surplus crates no longer contain extra large shot glasses

Valtsu0:
[OBJECTS] [BALANCE]

## Why's this needed? 
Extra large shot glass is a very niche traitor item that is practically useless on everything except bartender. Surplus crates should not contain them.

## Changelog 

```changelog
(u)Valtsu0
(*)Surplus crates no longer contain extra large shot glasses
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -844,6 +844,7 @@ This is basically useless for anyone but miners.
 	item = /obj/item/storage/box/glassbox/syndie
 	cost = 2
 	desc = "A box of shot glasses that hold WAAAY more that normal. Cheat at drinking games!"
+	not_in_crates = TRUE
 	job = list("Bartender")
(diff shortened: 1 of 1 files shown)


BatElite:
So I feel like there's some conflicting opinions on what the surplus crate should be, but my understanding of it is it's a gamble without consideration of how useful the contents are going to be (in exchange for higher TC value). Getting niche or awkward shit is part of the charm.

Even with that in mind though, whipping up (or stealing) some poisons and leaving them in glasses in the bar should be well in the remit of the average traitor. Hardly useless at all.

Sord213:
Nah, the traitor shot glass does have some uses over a regular shot glass including the old glass behavior of spilling reagents when thrown at someone (go crazy, chemists). Plus even if it's not very useful to you that's the whole point of the surplus crate, you don't know what you're going to get and you might not get a winner every time.

BUT if someone would want to maybe make a bundle just for surplus crates of something like traitor shotglass + moonshine or a single random poison bottle that would be cool.

PR #14267, opened 2023-05-27
Say Rework

amylizzle:
## About the PR 
Trying to make `say()` something vaguely maintainable and sensible.

https://hackmd.io/@amylizzle/sayrewrite

https://discord.com/channels/182249960895545344/1088956667250610176

## Why's this needed? 
spaget.jpg

Fixes #14599

## todo list

- [x] Basic core
- [x] Module sorting
- [x] Radios
- [x] Languages
- [x] Some accents (thanks herb)
- [x] Singing
- [x] Basic deadchat (including spooky overhearing of ghosts)
- [x] Identity system (voice/id/face/realname)
- [x] Formatting (prefix/suffix/span class/free form?)
- [x] AI internal radios
- [x] intercoms
- [ ] mechcomp microphones and speakers (radios also? think they just use packets)
- [ ] megaphones
- [ ] speech bubbles
- [x] lawbringer
- [ ] phones
- [x] blob chat
- [x] hivemind chat
- [ ] thrall chat
- [ ] kudzu chat
- [ ] flock radio
- [x] ooc/looc
- [ ] bunch of accents still need finishing (herb on that?)
- [x] logging
- [ ] uncool filter
- [x] whispering
- [ ] silicon speech
- [ ] cutting out all the deadcode

PR #14639, opened 2023-06-19
Remove Werewolves Scooby Doo Speech

Retrino:
## About the PR 

Allows werewolves to speak coherantly without having to shift out of their wolf form. Accents can be very hard to read through for some people and can often result in attempts to communicate being brushed off. Players can do their own funny little voice modifications if they want, and not worry about actual speaking attempts being ignored.

## Why's this needed? 

Werewolves were given the Scooby Doo accent a while ago, which was a funny gag for them for a while. However, the accent makes communication incredibly difficult, and seeing as they were recently given some weight to spawn on mixed rp rounds, that.. Isn't the best. QoL should come before a gimmick. The accent is still accessible via genetics, so I don't see much harm in its removal from wolves.

(i also have no idea if i coded this correctly, but it worked fine on local)
Screenshot (147)

## Changelog 

```changelog
(u)Retrino
(+)Werewolves will now speak normally again.
```

### DIFF
--- a/code/datums/mutantraces.dm
+++ b/code/datums/mutantraces.dm
@@ -641,7 +641,7 @@ ABSTRACT_TYPE(/datum/mutantrace)
 	mutant_appearance_flags = HUMAN_APPEARANCE_FLAGS
 	dna_mutagen_banned = FALSE
 	race_mutation = /datum/bioEffect/mutantrace/human
-	
+
 /datum/mutantrace/blob // podrick's july assjam submission, it's pretty cute
 	name = "blob"
 	icon = 'icons/mob/blob_ambassador.dmi'
@@ -1343,7 +1343,6 @@ ABSTRACT_TYPE(/datum/mutantrace)
 			src.mob.UpdateName()
 
 			src.mob.bioHolder.AddEffect("protanopia", null, null, 0, 1)
-			src.mob.bioHolder.AddEffect("accent_scoob_nerf", null, null, 0, 1)
 			src.mob.bioHolder.AddEffect("regenerator_wolf", null, null, 0, 1)
 
 	disposing()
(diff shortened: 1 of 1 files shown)


M-Earthfire:
In my opinion, i don't think werewolfs should be comprehensible while in wolf form. 

There should be a distinct difference in communication between human and wolf form and i would rather see their speech being entirely replaced with growls amd howls instead of be given clear communication.

If what they say is hardly understood and mostly brushed iof, i think the mutation is doing exactly what i think tt should.

Retrino:
> In my opinion, i don't think werewolfs should be comprehensible while in wolf form...

my thoughts can be summarized as follows: communication hinderance shouldnt be a balance feature of an antag. the only other antags that do this are flock or blobs, which are completely nonhuman. werewolves are humanoid antags. they can growl and howl all they want, if thats your playstyle. it should be up to the person playing the antag how much they want to lean into animal-ish speech.

flappybatpal:
I believe the intention of the accent is to reduce the number of "friendly werewolves" by making communication less practical.

Glamurio:
I think this is a good change. As mentioned, antags shouldnt be balanced around their ability to talk (if anything, we already have blob for that). Changing this will allow people to talk more freely, which is always a good thing. Additionally, if you're worried about people not seeing if they're werewolf or not when hearing conversations, an indicative speech verb such as "growls" or "howls" would give it away.

That being said, I think this PR should wait until the Say Rework (#14267) is done, to also allow the Werewolf to have, for instance, an icon when talking in werewolf form on radio.

CalliopeSoups:
I agree that antags shouldn't be balanced on removing communication, and that the accent paradoxically causes more 'friendly' werewolves. Still... I really like it. I dunno, it's funny, cute. Maybe someone can put together a milder version of the accent some day. I'd suggest removing it only on RP but server differences should be kept low.

Retrino:
felt like it was worth mentioning Glamurio, werewolves do have a special speaking verb, "snarls". They also show up over radio and in chat as "werewolf", so clarity of them being a wolf seems pretty solid. also may as well throw controversial label on this mf

flappybatpal:
> I believe the intention of the accent is to reduce the number of "friendly werewolves" by making communication less practical.

Regarding this, Aloe has changed the hunger motive on RP werewolf to be Ravenous hunger that requires mauling a human but decreases at half the rate to discourage friendly antags. Looks like it takes about 35 minutes to empty.

PR #14642, opened 2023-06-19
Adds wall trophy for fish as a purchase for fishing tickets.

TheColorCyan:
[GAME OBJECTS] [CATERING]

## About the PR 
Adds wall trophy for fish as a purchase for fishing tickets. You can attach it to a wall so everyone can see what kind of fish you caught.
I think its mostly done with the code part so draft status can be removed
nofish
trophy_screen
It has 2 vars that anchor stuff when set to true/false in mapping so it can be used and not stolen off the walls

## Why's this needed? 
Would be a nice addition to fishing so you could actually show off your fishing skills (and not just put all the fish on the floor). Make a fish museum today!!

## Changelog 
```changelog
(u)Sian
(*)Added Wall Trophy for fish available as a ticket purchase in Fishing Supplies Vendor.

```


Gannets:
I like this idea.

TheColorCyan:
this code scares me, i need someone to point where it is wrong

code/obj/item/wall_trophy.dm
```diff
@@ -0,0 +1,114 @@
...
+		..()
+		/*for mapping
+			checks for item in initial items and adds them to the trophy
+
+		*/
+		BLOCK_SETUP(BLOCK_BOOK)
+		for (var/type in src.initial_item)
+			. = src.add_item(new type(src.loc))
+			if (!.)
+				stack_trace("Failed to add item to trophy with initial items. [identify_object(src)]- likely can not accept this kind of object")
```
frawhst, line 42: This should be made not a loop if `initial_item` is a single type instead of a list

code/obj/item/wall_trophy.dm
```diff
@@ -0,0 +1,114 @@
...
+	//fish wall mount
+	fish_trophy
+		name = "'Biggest catch' Fish Wall Mount"
+		desc = "This Wall Mount can be hung on a wall and display fish for everyone to see. Show off your catch!"
+		allowed_item = /obj/item/fish/
+
+		//allows player to regulate where to put it, some fish sprites look janky otherwise
+		attackby(var/obj/item/W, var/mob/user, click_params)
+			add_item(W, user, click_params)
+			return
```
frawhst, line 114: I think this should just be on the parent type, makes sense that regardless of what we're mounting people should be able to mount things where they want on it

frawhst (commented):
I enjoy this

PR #14670, opened 2023-06-22
Deprecates /datum/admins/proc/evilize()

Mister-Moriarty:
[Internal] [Code Quality]

## About The PR:
Deprecates `/datum/admins/proc/evilize()` and `/client/proc/gearspawn_traitor()`.
Additionally renovates the "Make everyone an antagonist" admin command.

## Why Is This needed?
Removal of remnants from the old antagonist system that has been superceded by antagonist datums.

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -2420,31 +2420,49 @@ var/global/noir = 0
 								tgui_alert(usr,"The game hasn't started yet!")
 								return
 
-							var/which_traitor = input("What kind of traitor?","Everyone's a Traitor") as null|anything in list(ROLE_TRAITOR,ROLE_WIZARD,ROLE_CHANGELING,ROLE_WEREWOLF,ROLE_VAMPIRE,ROLE_ARCFIEND,ROLE_HUNTER,ROLE_WRESTLER,ROLE_GRINCH,ROLE_OMNITRAITOR)
-							if(!which_traitor)
+							var/list/antag_options = list()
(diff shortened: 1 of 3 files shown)


ZeWaka (approved):

code/procs/antagonist_procs.dm
```diff
@@ -1,30 +1,3 @@
-/client/proc/gearspawn_traitor()
```
TobleroneSwordfish, line 1: Don't we need this for omnitraitor and weird edge cases?

code/modules/admin/admin.dm
```diff
@@ -2420,31 +2420,49 @@ var/global/noir = 0
...
 
 							if (escape_objective == "None")
 								escape_objective = null
 
-							for(var/mob/living/carbon/human/H in mobs)
-								if(isdead(H) || !(H.client)) continue
-								if(checktraitor(H)) continue
-								evilize(H, which_traitor, hardmode, custom_objective, escape_objective)
+							for (var/mob/living/carbon/human/H in mobs)
+								if (isdead(H) || !H.mind || !H.client)
```
TobleroneSwordfish, line 2451:
```suggestion
								if (isdead(H) || !H.mind || !H.key)
```
I think this is more accurate, since we want to hit logged out players but not NPCs?

code/procs/antagonist_procs.dm
Mister-Moriarty: Not to my knowledge; all instances of `gearspawn_traitor()` were removed as I came across them while refactoring.

PR #14673, opened 2023-06-22
Moves Clarion Chemistry Request Console in QM so it no longer blocks a mulebot beacon 

DanaTarSUS:
## About the PR 
Moves the Chemistry Request Console in Clarion QM to no longer be on top of a Mulebot beacon.

## Why's this needed? 
Previously, the console was sitting directly on top of a mulebot beacon. This caused the Mulebot that spawns at the start of the round to be incapable of navigating to its own home beacon. As a Mulebot enjoyer, I cannot allow this to stand.


flappybatpal (approved):

flappybatpal:
Shame to lose a table but there doesn't look to be many options.

PR #14676, opened 2023-06-23
Runtime checks on random rooms

TheTyrant25:
[TOOLING]
## About the PR
Prefab checks but for random rooms

## Why's this needed?
What if there's runtimes in random rooms?


ZeWaka:
knew i was forgetting something

pali6 (approved):

PR #14684, opened 2023-06-24
oshan medical tweaks

TheTyrant25:
[MAPPING] [QOL] [INPUT]
## About the PR
- Adds a bin (or disposal chute, if you'd rather call it that) to genetics.
- Adds a door connecting robotics (the storage room, specifically) and the morgue, for easier access.

## Why's this needed?
Geneticists didn't deserve having to walk all the way down to medbay proper just to throw away things. Now they can remain in their room forever.
The morgue enjoys a special relationship with robotics, and having to go through the main medbay is also a bit annoying.

## Changelog
```changelog
(u)Tyrant
(+)Genetics in Oshan now has a disposal chute.
(+)Robotics Storage in Oshan now has a door connecting to the morgue directly.
```


TheTyrant25:
here's the MapDiffBot2 render for those who are lazy
image

Sord213 (changes requested):
I don't know why two tiny changes needed a forum thread.

Change the access spawner on the morgue -> robotics storage door to robotics access so medical doctors can't loot the robotics storage room

TheTyrant25:
people get sensitive over oshan changes so i wanted to be safe. I'll switch the access in a bit

PR #14685, opened 2023-06-24
Makes space GPS be able to fit into mechcomp cabinets

M-Earthfire:
[FEATURE][OBJECTS]

## About the PR 

Enables the space GPS to be able to be wrenched into the mechcomp cabinets (device frame/MechComp Cabinet).

## Why's this needed? 

With the movement components, people are enabled to make little bots that can move arround and do some mischief. What is currently missing is a convenient way to make them navigate themselves. While a space GPS can be attached to a frame via space glue, this is very temporary and can be easily destroyed by just removing the glued in component from the frame This makes a reliable, less sabotage-able way to make component bots able to navigate from A to B.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Enables the space GPS to be able to be wrenched into the device frame and mechcomp cabinet.
```

### DIFF
--- a/code/modules/mechanics/MechanicMadness.dm
+++ b/code/modules/mechanics/MechanicMadness.dm
@@ -19,7 +19,7 @@
 	name="Generic MechComp Housing"
 	desc="You should not bee seeing this! Call 1-800-CODER or just crusher it"
 	icon='icons/misc/mechanicsExpansion.dmi'
-	can_hold=list(/obj/item/mechanics)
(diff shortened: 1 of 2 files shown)


M-Earthfire:
I was either thinking about adding this functionality to the space gps or making a dedicated gps-component. But a gps component doesnt really make sense to be wrenched onto the ground, since it would just return a fixed position you could just check once and hack it in.

TobleroneSwordfish (approved):
seems cool 👍

PR #14686, opened 2023-06-24
Fixes every doubled turf in maps and most missing types in some old maps

CheffieGithub:
## About the PR 
Fixes every map key that has a doubled turf. 
Fixes old types on oldmap, trunkmap, chiron, concept, density, mushroom, samedi, and the silo setpiece (most should compile now)
Fixes non perspective walls on kondaru as well as the doubled turfs.
Fixes invisible plating using icon_states for damage instead of the subtype with overlays in these maps.

## Why's this needed? 
Updatepaths kept screaming about them and it was annoying. Oldmaps updated for the possible gimmick and ease of viewing.


Laboredih123:
this is so epic, i was planning to do this but im too lazy

Laboredih123:
oldmap seems a bit broken however still, lot missing

Laboredih123:
still, nice work

CheffieGithub:
oldmap had some things without mirrors but it also won't ever work as it is with current code I don't think

CheffieGithub:
but also most of the missing stuff is just bad icon edits

flappybatpal:
How the heck to review a 400k line change PR 💀

CheffieGithub:
smallest mapping PR

CheffieGithub:
I would say just review the mdb diff but the diff is broken

CheffieGithub:
the diff is back

CheffieGithub:
Im going to atomise this out

PR #14694, opened 2023-06-25
Choose your own ~~adventure~~ engine!

amylizzle:
## About the PR 
Adds a random engine to Cogmap 1, with a small chance to get the "Choose your own engine" prefab which comes with a console that lets you select the others.

[Screencast from 2023-06-25 16-55-53.webm](https://github.com/goonstation/goonstation/assets/3855802/ef659e7b-1ae3-49c1-b757-1ea97e507f14)

## Why's this needed? 
It's the most popular map by far, but that restricts 90% of rounds to the TEG. I would like a bit of variety.

## Concerns about randomness ##
It'd be trivial to make this random rather than a choice. We could try and then swap it if it doesn't work out.

## Changelog 

```changelog
(u)Amylizzle & Garash
(*)Cogmap now comes with a random engine type, or a rare chance to use the brand new NT Engine Teleportation System, which lets engineers select the engine they want to use.
```


BatElite:
~~geothermal when~~

amylizzle:
> ~geothermal when~

geothermal is a lie. We're in space, there is no geo to be thermal.

Gannets:
this is a cool idea

wolfolotl:
can we get a hamster wheel layout for if the console gets emagged please please pretty please

Tarmunora:
Would prefer to see this as a (weighted?) random selection during setup - having a "select your engine" console feels quite out-of-place

amylizzle:
> Would prefer to see this as a (weighted?) random selection during setup - having a "select your engine" console feels quite out-of-place

Easy enough to make the switch. Could even leave "choose your own engine" in as a rare pick. People do seem to be enjoying picking their favourite engine though.

Wisemonster:
Instead of teleporting in, could have it so the console "assembles" the desired engine on-site, and rename the console to reflect that.

ZeWaka:
> Instead of teleporting in, could have it so the console "assembles" the desired engine on-site, and rename the console to reflect that.

Even still. I would much prefer if maps have a single/random engine, and you could swap it out by disassembling the current one and rebuilding the new one in its place.

amylizzle:
I could route it through the windows instead? I can't have a bend on that section because the pipe unbends when it is severed, which breaks the replacement connection.

CheffieGithub:
If it needs to be straight then I would remove the bottom table and move its contents to the one on the other side of the room. Could also replace that window with a wall and shuffle the lights but it would break the nice symmetry.

deathrobotpunch:
> geothermal when

no *astrothermal*

PR #14696, opened 2023-06-25
Resprites Fire Alarms

MitchTheUser:
## About the PR 

Fire alarms have been resprited, keeping their general design but having their size reduced to match other buttons and equipment while still being large enough to be easily found when going off. They have also gotten a new color scheme to match similar equipment, and are overall more vibrant and colorful.

FireAlarm

## Why's this needed? 

alarm

The current Fire alarms are pretty outdated, taking up more space than is needed for visibility, and being dull in color which makes them stand out poorly compared to the comparatively more colorful equipment. Fire Alarms were also on the wanted resprites list, and the same points were made there about size and look.

## Changelog 

```changelog
(u)MitchSmith
(+)Resprited Fire Alarms
```


flappybatpal (approved):

Gannets (approved):
Yeah, these are nice.

PR #14706, opened 2023-06-26
Repaths Access, Mob, Firedoor and Wingrille spawns and adds airlock bolters/welders

CheffieGithub:
## About the PR 
Repaths Access, Mob, Firedoor, Window grille spawns to /obj/mapping_helper/ due to their shared behaviour, organised into a mapping folder in objects. Adds a mapping helper for bolting and welding airlocks.

Behaviour of these is to run setup() and delete themselves.

Cleans up the wingrille spawners with absolute paths and fixes a broken spawner.

## Why's this needed? 
No need to duplicate the same code. 
Bolting and welders are to help mapping and the new object parent as a whole is to allow for more convience objects to be added in future.


ZeWaka:
more like MDB suck it

Laboredih123:
absolute pathing my beloved

PR #14714, opened 2023-06-26
Expands Clarion QM and reorganizes QM Storage

DanaTarSUS:
## About the PR 
Expands Clarion QM into part of QM storage and reorganizes many of the crates and fabricators littered throughout the storage room.

## Why's this needed? 
Currently, Clarion QM itself is horribly cramped while QM storage is far too large. In addition, every other department that falls under Engineering has access to the storage space in one way or another. With that in mind, I believe that QM should have sufficient space to operate without needing to make use of a shared space to prevent their own department from becoming cluttered by crates.

## Changelog 

```changelog
(u)DeeDotVeeA
(+) Clarion QM is now slightly more spacious.
```


DanaTarSUS:
Ignore this most recent commit I goofed up trying to fix the merge conflict

PR #14725, opened 2023-06-26
Makes galoshes grant immunity against lube-slips

M-Earthfire:
[Chemistry][Balance][Objects]

## About the PR 

This PR adds a new clothing flag, `NO_LUBE_SLIP`. This one can be added to shoes to prevent the shoes from slipping on tiles with lube/harmlube. walking on lube/harmlube checks `mob/proc/canslip` now. This proc got a variable called lube_slip, which checks for the case someone is walking on lube.

Galoshes gain the `NO_LUBE_SLIP` flag. This means galoshes are now immune to lube-slips. Also, they are avaible at the Janitorial Supplies kit at the QM now, with the corresponding crate getting a price increase from 500 credits to 1500.

## Why's this needed? 

This change adds counterplay to people spamming space lube while also giving antag janitors the ability to far better and easier deploy and use space lube.

Locking it behind janitors prevents people from rushing them and requiring interdepartmental cooperation if security needs to stop an antag spamming lube all over the station. And honestly, seeing security chasing someone in galoshes just sounds funny.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Galoshes grant immunity to lube and harmlube slips now.
(+)The janitorial supplies crate at the QM contains galoshes now and got its price increased to 1500 credits
```

### DIFF
--- a/_std/defines/clothing.dm
+++ b/_std/defines/clothing.dm
@@ -64,7 +64,8 @@
 #define ONBACK						(1<<17)
 /// can be work on the belt
 #define ONBELT						(1<<18)
-
+/// for galoshes that are immune to lube + harmlube
+#define NO_LUBE_SLIP				(1<<19)
 
 //Suit blood flags
 #define SUITBLOOD_ARMOR 1
--- a/code/mob/living/carbon.dm
+++ b/code/mob/living/carbon.dm
@@ -44,20 +44,22 @@
 							src.inertia_dir = 0
 							return
 					if (2) //lube
-						src.remove_pulling()
-						src.changeStatus("weakened", 3.5 SECONDS)
-						boutput(src, "<span class='notice'>You slipped on the floor!</span>")
-						playsound(T, 'sound/misc/slip.ogg', 50, 1, -3)
-						var/atom/target = get_edge_target_turf(src, src.dir)
-						src.throw_at(target, 12, 1, throw_type = THROW_SLIP)
+						if (src.can_slip(TRUE))
+							src.remove_pulling()
+							src.changeStatus("weakened", 3.5 SECONDS)
+							boutput(src, "<span class='notice'>You slipped on the floor!</span>")
+							playsound(T, 'sound/misc/slip.ogg', 50, 1, -3)
+							var/atom/target = get_edge_target_turf(src, src.dir)
+							src.throw_at(target, 12, 1, throw_type = THROW_SLIP)
 					if (3) // superlube
-						src.remove_pulling()
-						src.changeStatus("weakened", 6 SECONDS)
-						playsound(T, 'sound/misc/slip.ogg', 50, 1, -3)
-						boutput(src, "<span class='notice'>You slipped on the floor!</span>")
-						var/atom/target = get_edge_target_turf(src, src.dir)
-						src.throw_at(target, 30, 1, throw_type = THROW_SLIP)
-						random_brute_damage(src, 10)
+						if (src.can_slip(TRUE))
+							src.remove_pulling()
+							src.changeStatus("weakened", 6 SECONDS)
+							playsound(T, 'sound/misc/slip.ogg', 50, 1, -3)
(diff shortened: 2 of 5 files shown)


TheColorCyan:
thank god

M-Earthfire:
> I think the price would need to be increased more. I also think that galoshes should give you slowdown with this change to balance out complte immunity to slips.

I like the idea of galoshes adding a slowdown. I am not sure how our fellow janitors would like it, though.

Is there something like janitorial training? Maybe make just non-janitors have a slowdown.

Sord213:
Yeah I'd rather not give powergamers who rush galoshes another advantage. There's plenty of ways to obtain galoshes outside of QM as well that would need to be rebalanced.

CheffieGithub:
Magboots aren't immune they just don't move they still get stunned

M-Earthfire:
> Part of the game design of lube is that no one (including the person spreading the lube) is immune from its effects. This feels like it ruins a lot of the fun and risk of lube.

We sadly live in times where prople are overusing lube and we got people crying sbout it instead of adapting (use flashbangs/CC grenades and don't engage in chases with a luber, dammnit). 
And because i think special effects should be and stay impactful, i rather add counters to stuff than water it down by nerfs.

With the changes, at best, it should be still highly discouraged to chase a lube-user.

> Yeah I'd rather not give powergamers who rush galoshes another advantage. There's plenty of ways to obtain galoshes outside of QM as well that would need to be rebalanced.

Agreed. I test around with movement süeed penalties for non-janitors and look whst feels right. If you have ways of obtaining galoshes i should look upon, you can tell me them and i replace them with the torn variant that does not grant slipping protection at all.

TobleroneSwordfish:
This would almost certainly lead to far more lube spam. Imagine being able to run around spraying lube everywhere while being completely immune to its effects.

M-Earthfire:
> This would almost certainly lead to far more lube spam. Imagine being able to run around spraying lube everywhere while being completely immune to its effects.

Fully agree there and i am aware of this. It will make lube use more common, but would make it less problematic over the span of a whole round, since the lube user would see the effectiveness quickly decline as people give out galoshes. It forces a transition to different tools/weapons.

The complains over the forums were over people being unable to fight it, meaning stopping lube rampages. This PR gives a convenient way to stop them, while also enabling them if you invest more time in it (by aquiring galoshes and deal with the ms penalty).

Overall, it's a shift of the role of lube in general.

M-Earthfire:
Thats why it will get, once i come home to code it, a significant speed penalty for non-janitors.

That way it has consequences for users that are not meant to wear it all the time.

For janitors it will be a straight up buff. They honestly need it.

TheColorCyan:
Honestly a good alternative may be just making lube not slip you if you are walking

M-Earthfire:
> Honestly a good alternative may be just making lube not slip you if you are walking

That would give potential everyone, even normal crew, lube-protection without any work. Thats a very clear no from my side, at least.

TheColorCyan:
> > Honestly a good alternative may be just making lube not slip you if you are walking
> 
> That would give potential everyone, even normal crew, lube-protection without any work. Thats a very clear no from my side, at least.
> 
> Lube-rampages are a problem. In any other situation lube is, in my opinion, fine.

walk makes you slow as hell and it still makes people that sprint and run slip which is what usually happens in combat.... banana peels and other stuff work the same, lube is just a bit more dangerous since it can slip people with both noslips and without. I REALLY doubt people will ALWAYS switch to walk  because it not only sucks for chases but also makes you slow (and people hate it). Making it not slip you when walking seems like a solution that wont benefit the powergamers devs are so scared of while making lube less painful to deal with....

Tarmunora:
Hard no on this one. As others devs said earlier:

- Not a fan of making galoshes cargo items
- Galoshes don't need a buff
- Being able to spam lube while immune sounds like a bad idea (disregard cyborgs)

PR #14726, opened 2023-06-26
Delete pills which are used to transfer reagents

### LINKED ISSUE #14636: Making unlimited pipe bomb fuel
### Labels

[BUG][MINOR]

### Description

You can make black powder(unsure if this works with anything else) into a pill and use it as infinte fuel for pipebombs. Unsure if the pipe bombs do anything when exploding.

### Steps to reproduce

1.Make gunpowder
2.Turn to pill
3.Use pill on hollowed pipe
4.Do it again and again and again and again 
5.Blow up clown

### Expected Behavior

Make it so that the pill is used up once it runs out of black powder

Hi :3

Reported by: Captain-Caffeine-Fiend
Active test merges: nothing
Round log date: 2023-06-19-00-38
Reported on: 2023-06-19 01:20:12
Map: COGMAP

MeggalBozale:
## About the PR 
Pills now will delete themselves if they are used to transfer reagents to closed containers like pipebomb frames

## Why's this needed? 
Fixes #14636
It doesn't make sense to be able to use anything but the whole pill, because as far as I know, they are capsules with what is usually liquid inside them, which would spill all over the place if you tried to take an amount out of it. That, and it's already the behavior for open containers like beakers.


MeggalBozale:
Actually I might want to figure out how to put this into a proc for pills since this can happen with things that aren't pipebomb frames. I have no idea how, though.

code/modules/chemistry/tools/pills.dm
```diff
@@ -91,6 +91,8 @@
 			qdel(src)
 			return
 		else
+			if (src.reagents.total_volume <= 0)
+				qdel(src)
```
Tarmunora, line 95: Wouldn't this check be more appropriate in `on_reagent_change()`?

MeggalBozale: I had no idea that existed, yeah that makes a lot more sense. 

I'm also thinking that maybe pills should delete themselves on reagent change to any closed containers too, since it was pointed it that you can't exactly split what is probably wrapping around chems. Not sure how I would do that, though.

MeggalBozale: to make more sense of that second paragraph, is there an `on_transfer()`? That might be more useful so pills don't qdel themselves on chem reactions if it makes sense to change it that pills qdel on being used even if not all reagents are used.

MeggalBozale:
I changed pretty much everything uhh let me know what you think? And if pills should be allowed to transfer part of their reagents after all it can just be reverted back to the previous commit

code/atom.dm
```diff
@@ -942,6 +942,10 @@ TYPEINFO(/atom)
 	SEND_SIGNAL(src, COMSIG_ATOM_REAGENT_CHANGE)
 	return
 
+/atom/proc/on_reagent_transfer()
+	SEND_SIGNAL(src, COMSIG_ATOM_REAGENT_TRANSFER)
+	return
```
Tarmunora, line 947: If you want to have a signal on this, add the `SHOULD_CALL_PARENT(TRUE)` linter directive here - you want to ensure that any overrides to this proc will call back up to the parent here so the signal gets sent

code/modules/chemistry/tools/pills.dm
```diff
@@ -93,6 +93,9 @@
 		else
 			return ..()
 
+	on_reagent_transfer()
+		qdel(src)
```
Tarmunora, line 97: and similarly, this should call parent

Tarmunora (changes requested):

code/modules/chemistry/tools/pills.dm
MeggalBozale: added `..()` before `qdel(src)`

code/atom.dm
MeggalBozale: Added `SHOULD_CALL_PARENT(TRUE)` as it was done in the other proc

Tarmunora (approved):

ZeWaka (changes requested):
why are you making a signal and then not using it? makes no sense.

MeggalBozale:
Not sure, was following the footsteps of the proc above it so it would work. The change works fine without anything in the proc at all, I don't know if I need on_reagent_transfer in `atom.dm` at all at that point, though.

Tarmunora (approved):

code/atom.dm
```diff
@@ -942,6 +942,8 @@ TYPEINFO(/atom)
 	SEND_SIGNAL(src, COMSIG_ATOM_REAGENT_CHANGE)
 	return
 
+/atom/proc/on_reagent_transfer()
+
 /atom/proc/Bumped(AM as mob|obj)
 	SHOULD_NOT_SLEEP(TRUE)
```
ZeWaka, line 948:
```suggestion
/atom/proc/on_reagent_transfer()
	return

 /atom/proc/Bumped(AM as mob|obj)
 	SHOULD_NOT_SLEEP(TRUE)
```

code/modules/chemistry/Chemistry-Holder.dm
```diff
@@ -889,6 +889,11 @@ proc/chem_helmet_check(mob/living/carbon/human/H, var/what_liquid="hot")
 			my_atom.on_reagent_change(add)
 		return
 
+	proc/reagents_transferred()
+		if (my_atom)
+			my_atom.on_reagent_transfer()
```
ZeWaka, line 894:
```suggestion
		my_atom?.on_reagent_transfer()
```

ZeWaka (approved):

PR #14741, opened 2023-06-27
This is a nuclear reactor. It is radiating blindingly with a worrying blue light.

amylizzle:
## About the PR 
Increases the range of radiation the reactor can be at for the radioactivity component, and makes it more responsive to changes.
At above 50 rads, the reactor goes to neutron radiation. Neutron rads are 4x stronger than normal rads, so the strength is divided by 4 to keep it consistent. 
Also it increases the dose from the reactor when you're stood next to it by 33% because it made it go blue at 50 rads which matches the danger area on the dial.

## Why's this needed? 
It indicates high radiation levels visually without needing the UI open.


amylizzle:
I have decided I don't like this.

PR #14743, opened 2023-06-27
Adds spacecillin cure to food poisoning description

Wisemonster:
[QOL] [MEDICAL]

## About the PR 
Adds spacecillin to the list of cures on the medical scanner for food poisoning.

## Why's this needed? 
Considering it is a cure, might as well let people know through the scanner.

### DIFF
--- a/code/modules/medical/diseases/food_poisoning.dm
+++ b/code/modules/medical/diseases/food_poisoning.dm
@@ -2,7 +2,7 @@
 	name = "Food Poisoning"
 	max_stages = 3
 	spread = "Non-Contagious"
-	cure = "Sleep"
+	cure = "Sleep, Spaceacillin"
 	associated_reagent = "salmonella"
 	reagentcure = list("spaceacillin")
 	affected_species = list("Human")


Tarmunora (changes requested):
That's not how this works - that var isn't for description, that's what actually checked by code to see if sleeping etc. should cure the disease [`code/datums/disease.dm:187`](https://github.com/Tarmunora/goonstation/blob/7d59a73235c36a5497d635fc8af07f53eafdcd6e/code/datums/disease.dm#L187)

Wisemonster:
Odd, that appears to be how it works for code\modules\medical\diseases\avian_flu.dm

Tarmunora:
> Odd, that appears to be how it works for `code\modules\medical\diseases\avian_flu.dm`

That one isn't curable by sleep though - current malady behaviour for sleep curing is an exact string comparison on the `cure` var for the string `"Sleep"`, if you have a cure of `"Sleep, foo"`, it will no longer be curable by sleep without also fixing that lmao

Wisemonster:
Buh

Wisemonster:
> That's not how this works - that var isn't for description, that's what actually checked by code to see if sleeping etc. should cure the disease [`code/datums/disease.dm:187`](https://github.com/Tarmunora/goonstation/blob/7d59a73235c36a5497d635fc8af07f53eafdcd6e/code/datums/disease.dm#L187)

What I could do is add a check for "Sleep, foo" on that line, although that seems to be more of a quick fix rather than a proper solution.

frawhst:
(This isn't really your fault, disease cures are awful)

Wisemonster:
Ok, this seems more like something that a refactor is needed for, rather than a simple fix

PR #14747, opened 2023-06-27
Resprites Breath Masks and Medical Masks

MitchTheUser:
## About the PR 

Breath and Medical Masks have been resprited, retaining a similar look but having multiple changes. The shading has been improved to bring it up to current standards and look like a bright/reflective surface. The straps have been connected and go behind the mask instead of sitting broken apart, and are now the same color between the two. The masks are now distinguished by unique mask color. The masks are also less circular and have a pointed nose arch, and the tube on the bottom has also been changed to look more like a tube and come out of the front like a real oxygen mask.

MaskComparison

## Why's this needed? 

NewMaskDisplay
The current sprites are quite old, and have fallen behind a bit in terms of art style. Their new colors and shading should now be more consistent with the modern art style and should hopefully be an overall improvement.

## Changelog 

```changelog
(u)MitchSmith
(+)Resprited Breath/Medical Masks
```


DisturbHerb:
I'll be honest, this is - to me - one of those cases where the current icons look and read better than the resprite attempts. I'm just really not a fan. Maybe it's S O U L related.

flappybatpal:
Unfortunately I don't think it looks like a mask.

Terms-On-Dragon:
After seeing your message asking for feedback, I went ahead and compared your sprite to oxygen masks.  These are the three features I noticed that may require some adjustments.

These masks are too rounded and do not look like they would cover the nose all that well, I suggest adding a ridge at the top of the masks.

In oxygen masks, there is a slot where the tube would connect; extending the tube to be slightly in front of the mask instead of extruding from the bottom may make it more clear.  The tube itself would also be transparent; lighter in color.

If possible, the straps should be connected as they look cut apart.

Hopefully this criticism will help improve the design you currently have, as I would like to see how it turns out in the end!

MitchTheUser:
Updated the sprites pieces of feedback, the images and description have been changed to reflect this.

PR #14757, opened 2023-06-28
Adds better surgery slip-up indicators

Bartimeus973:
## About the PR 
Adds very obvious surgery slipup indicators when you mess up a step of a surgery.

Importantly, this icon is ONLY visible by the surgeon. No one else can see it. They can still see the text in the chat and the small blood gush.

Demonstration (ingame icon appears twice as fast compared to the gif):

slipup2

Clumsy surgeons get unique slipup messages. Spawns the "Oops" message the very vast majority of the time, "Sorry" 10% of the time and "Malpractice'd" 1% of the time.

slipup3

## Why's this needed? 

Has been discussed here: https://forum.ss13.co/showthread.php?tid=21437

Newer surgeons often do not see that they messed up a surgery step and get confused as to why they have to redo it. This should make it a lot more obvious when something happens wrong without resorting to comical amounts of blood flying all over the place.

## Changelog 

```changelog
(u)Bartimeus973
(+)Surgery slipups are a lot more noticeable.
```


M-Earthfire:
Personally, i would make them slightly smaller and make them start slightly above the lying sprite of the person. Right now they completely cover the patient at the start.

Beyond that, i absolutely love them!

Bartimeus973:
Making them start higher makes sense, i'll do that

Bartimeus973:
It now spawns above the patient and is 10% smaller. If i wanna make it even smaller i'll need to redo the sprites entirely.

Kyle2143:
I was thinking of a more modest text/button like the "GRAB"/"BLOCK" button that gets slapped onto the patient then fades away.

The graphics look more like speech coming from someone based on the way they float away like speech map-text does.

Silly can be fun in many cases, and here too, but this feels a bit much for an action that can possibly happen very frequently during surgery.

Bartimeus973:
Alright, i'll rework it to be inside of a frame and i'll rework the movement so it fits a bit better!

Kyle2143:
Though, don't throw them away! Some people mentioned that these graphics could work well for clown surgery as is or maybe with a colour scheme tweak and that sounds like a fantastic idea to me.

flappybatpal:
The text would be a good fit for clown yes but I think it's too much for regular surgery. Something way more subtle (just not as subtle as the existing failure) would be a better fit for normal surgery.

amylizzle:
Very cool, but why not just use maptext?

Bartimeus973:
Regular slipups made by people are a lot more subtle compared to before, but still pop out a fair bit.
Clowns keep the big bulky text recolored to pink to better fit them.
These messages are still only seen by the surgeon.

NotDucc:
OOooo, I like this! Definitely helps convey a slip up was made a lot better

code/modules/medical/surgery_procs.dm
```diff
@@ -369,6 +370,7 @@ var/global/list/chestitem_whitelist = list(/obj/item/gnomechompski, /obj/item/gn
 					patient.TakeDamage("head", damage_high, 0)
 					take_bleeding_damage(patient, surgeon, damage_high, surgery_bleed = 1)
 					CREATE_BLOOD_SPLOOSH(patient)
+					display_slipup_image(surgeon, patient.loc)
```
ZeWaka, line 373: This should all be split into a proc, this exact 3-5 phrases are repeated like 12 times

ZeWaka (changes requested):

code/modules/medical/surgery_procs.dm
Bartimeus973: I put it all into a "do_slipup" proc.

code/obj/decal.dm
```diff
@@ -809,3 +809,73 @@ obj/decal/fakeobjects/teleport_pad
...
+		var/matrix/original = matrix()
+		original.Scale(0.05)
+		src.transform = original
+
+		animate(src,transform = matrix(1, MATRIX_SCALE), time = 1 SECONDS, alpha = 255, pixel_y = 16, pixel_x = rand(-32, 32), easing = ELASTIC_EASING)
+		animate(time = 1 SECOND, alpha = 0, pixel_y = 8, easing = CIRCULAR_EASING)
+		SPAWN(2 SECONDS)
+			qdel(src)
+
+/obj/decal/slipup_clumsy
```
ZeWaka, line 850: Why doesn't this inherit?

code/modules/medical/surgery_procs.dm
```diff
@@ -74,8 +74,10 @@ var/global/list/chestitem_whitelist = list(/obj/item/gnomechompski, /obj/item/gn
 	else // if all else fails?
 		return 1 // head surgery is okay
 
-#define CREATE_BLOOD_SPLOOSH(T) var/obj/itemspecialeffect/impact/E = new /obj/itemspecialeffect/impact/blood;\
-			if (E){E.setup(get_turf(T));}
+/proc/create_blood_sploosh(var/T)
+	var/obj/itemspecialeffect/impact/E = new /obj/itemspecialeffect/impact/blood;\
+	if (E)
+		{E.setup(get_turf(T));}
```
ZeWaka, line 80: If you're going to make it not a macro make it normal

code/modules/medical/surgery_procs.dm
```diff
@@ -2083,4 +1970,25 @@ var/global/list/chestitem_whitelist = list(/obj/item/gnomechompski, /obj/item/gn
 			logTheThing(LOG_COMBAT, surgeon, "removed [constructTarget(patient,"combat")]'s [limb] with [src].")
 			return TRUE
 
-#undef CREATE_BLOOD_SPLOOSH
+//You messed up. Cause damage and spawn some indicators.
```
ZeWaka, line 1973:
```suggestion
/// You messed up. Cause damage and spawn some indicators.
```

code/modules/medical/surgery_procs.dm
```diff
@@ -2083,4 +1970,25 @@ var/global/list/chestitem_whitelist = list(/obj/item/gnomechompski, /obj/item/gn
...
+/proc/do_slipup(var/mob/surgeon, var/mob/patient, var/damage_target, var/damage_value, var/fluff_text)
+	surgeon.visible_message("<span class='alert'><b>[surgeon][fluff_text]!</b></span>")
+	patient.TakeDamage(damage_target, damage_value, 0)
+	take_bleeding_damage(patient, surgeon, damage_value, surgery_bleed = 1)
+	create_blood_sploosh(patient)
+	display_slipup_image(surgeon, patient.loc)
+
+/proc/display_slipup_image(var/mob/surgeon, var/loc)
+	//Spawns an image above a patient when you slip up. Only the surgeon sees it.
```
ZeWaka, line 1982: Move to code doc

ZeWaka (changes requested):

code/obj/decal.dm
```diff
@@ -809,3 +809,65 @@ obj/decal/fakeobjects/teleport_pad
...
+	icon = null
+	icon_state = null
+	alpha = 0
+	opacity = 0
+	pixel_x = 0
+	pixel_y = 8
+	plane = PLANE_NOSHADOW_ABOVE
+
+	New(var/location = null, var/state = null, var/mob/target = null)
+		if (!istype(src, /obj/decal/slipup/clumsy))
```
ZeWaka, line 827: ??? just clean up and fix the code instead of this weird junk

code/modules/medical/surgery_procs.dm
```diff
@@ -74,8 +74,10 @@ var/global/list/chestitem_whitelist = list(/obj/item/gnomechompski, /obj/item/gn
 	else // if all else fails?
 		return 1 // head surgery is okay
 
-#define CREATE_BLOOD_SPLOOSH(T) var/obj/itemspecialeffect/impact/E = new /obj/itemspecialeffect/impact/blood;\
-			if (E){E.setup(get_turf(T));}
+/proc/create_blood_sploosh(var/T)
+	var/obj/itemspecialeffect/impact/E = new /obj/itemspecialeffect/impact/blood;\
```
ZeWaka, line 78: still need to cleanup, also `E` is a bad name. T should be typed

ZeWaka (changes requested):

code/obj/decal.dm
Bartimeus973: I'm not sure what kind of shitcode demon possessed me there. Sorry about that

ZeWaka (approved):

ZeWaka:
Thanks for the changes!

Bartimeus973:
Setting ready for final review since code was apparently fine. If i shouldnt set this label, please tell me so and remove the label.

imawxome:
I think there ought to be a (very) small chance to do a massive slipup if you are a clown, or have some huge disadvantage, like incredibly drunk and jittery. with an extra large "OOPS!!" and severe damage to the one being treated, like losing a random limb.

PR #14759, opened 2023-06-28
Mobifies Gunbots

CheffieGithub:
## About the PR 
Converts obj/critter/gunbot and /heavy to mob critters.
Adds a new smg limb for /gunbot/light (previously obj/critter/gunbot) since 3 burst .38 would be brutal.

Fixes up gun limb cooldowns since they weren't working correctly.
Makes the gun limb have shot spread since it was a bit of a laser on the AI.

Obj versions retained in code for now since the drones rely on them, will be removed once they are mobs.
Fixes some owlery areas and old walls while I was replacing decorative critters with decals.

## Why's this needed? 
Cleans up some code and removes obj critters from gameplay.

## Changelog 

```changelog
(u)Cheffie
(+)Robots in space are now mobs and are resistant to unarmed strikes. Bring something heavy!
```


flappybatpal:
While the old gunbots were frankly overpowered in weaponry these seem to be heavily nerfed. The lower rate of fire and reload period where they do nothing at all makes them comparatively very easy to bait their shots and melee them to death while they reload forever.

They also seem to have lost their speech and sound effects.

CheffieGithub:
Yeah the reload stuff was just me reusing the 38 limb gunbots already have, Ill create a faster subtype.
I forgot about the speech stuff ill re-add that.

The AI stuff, currently it maintains a distance while trying to shoot you but doesn't change this when on cooldowns and just sits there. I want to further improve the ranged AI to change hands and attack according to the situation (hence the get hand procs), move when reloading, etc. But those changes are out of scope for this PR.

CheffieGithub:
It seems that being on cooldowns stops them from moving which ill look into

CheffieGithub:
Gave the npc gunbot a faster .38 limb with 2 seconds shot delay and 8 seconds reload.
Ported the glitch visible message because I couldn't get it to work with say at all... will come back and change it once the speech rework is in.

flappybatpal (approved):
Secrets will need checking for paths too

PR #11970, opened 2022-11-15
tgui id computer

Garash2k:
[A-UI][C-Feature]

## About the PR
Reworks the ID Computer to be TGUI, also affects the foldable portalbe ID computer.
Little functionality changes have been made besides removing a useless "Log in" since you'd get logged out upon removing any card, (you now get logged in automatically upon entering both cards assuming you have access)
image
image
image

Draft for now, todos:
- make more compact
- compare custom card style old behavior
- compare card title update old behaviour
- compare old unauthorized message
- make sure foldable works as well as it did before

## Why's this needed?
Color coding and groupings should make it easier to use. Also a useless Login feature has been removed, making it faster to use. Plus, TGUI looks more modern than raw html

## Changelog 
```changelog
(u)Garash
(+)New snappier UI for the Identification Computer, go commit ID fraud today!
```

PR #14766, opened 2023-06-29
TGUI ID Computer

Garash2k:
## About the PR 
Rising  #11970 from it's grave from when I worked on it 6 months ago. Adressed the "loading from game data point," made it compatible with dept. ID computers, and added a few goodies such as a selection on the card look button, faster updating card name, and some put_in_hand_or_eject.
I know that compactness is a concern when going from raw html to tgui, but I think this result is ok, with every job being visible, and the area accesses peeking out of the fold with this new default window size.

image

older pics:

image
image
image
image
image

## Why's this needed? 

Color coding and groupings should make it easier to use. Also a useless Login feature has been removed, making it faster to use. Plus, TGUI looks more modern than raw html

## Changelog 

```changelog
(u)Garash
(+)New snappier UI for the Identification Computers, go commit ID fraud today!
```

[UI][Feature]


Luxizzle:
A little nitpick, but the background colors make the interface look extremely busy and the options hard to read.

Glamurio:
Love this, great job, fantastic work, this makes me excited to work on my HoP rework again.

> A little nitpick, but the background colors make the interface look extremely busy and the options hard to read.

Suggestion: Only have the headers colored, but the main content be black. Here's a mockup:

image

Luxizzle:
> Love this, great job, fantastic work, this makes me excited to work on my HoP rework again.
> 
> > A little nitpick, but the background colors make the interface look extremely busy and the options hard to read.
> 
> Suggestion: Only have the headers colored, but the main content be black. Here's a mockup:
> 
> image

Yeah, something like that, I'd suggest making them collapsible, see the WeaponVendor as an example. And don't darken the colors by 20% as they ruin the contrast and makes the colors hideous in my opinion

Garash2k:
> Yeah, something like that, I'd suggest making them collapsible, see the WeaponVendor as an example. And don't darken the colors by 20% as they ruin the contrast and makes the colors hideous in my opinion

The reason the colors are darkened is for accessibility concerns. Because the text is white, it has poor readability unless the background color is darkened
image
image

As for collapsible, I'm not seeing the use for it, each section would only collapse 1 to 3 lines of buttons, and it's not like a seed vendor where you can be interested in a specific category that would otherwise be a few pages down.

Glamurio:
> The reason the colors are darkened is for accessibility concerns. Because the text is white, it has poor readability unless the background color is darkened.

Totally valid, however they have a point about the colors. Whenever it comes to accessibility, I like to do a luminance comparison of the colors and decide if using black font color might be better. I'm not sure however, how good it would look to have one or two departments with black headers.

Perhaps we could achieve a happy medium? How does 10% look, does that still pass or at least barely fail?

Additionally, since we are using HTML and presumably CSS, you could try the text-shadow property:
`text-shadow: 0 0 1px black, 0 0 1px black, 0 0 1px black, 0 0 1px black; `
With this, white text could be outlined black. That way, the text will stand out no matter the background.

Glamurio:
image

Small typo, you misspelled "maintenance"

Garash2k:
!merge_upstream

tgui/packages/tgui/styles/interfaces/IDComputer.scss
```diff
@@ -0,0 +1,31 @@
+/**
+ * @file
+ * @copyright 2022
```
ZeWaka, line 3:
```suggestion
 * @copyright 2023
```

tgui/packages/tgui/interfaces/IDComputer.js
```diff
@@ -0,0 +1,203 @@
+/**
+ * @file
+ * @copyright 2022
```
ZeWaka, line 3:
```suggestion
 * @copyright 2023
```

ZeWaka (changes requested):

ZeWaka (approved):

ZeWaka (changes requested):
actually one small change i see that can help with the verticality
do this with the top - probably a 2/3 1/3 split because long names
image

Garash2k:
Horizontal!
image

Also while testing it, I noticed that very long buttons would grow out of the window/their parent. With a whitespace property we can make them wrap, which is better than then ending up on top of the Pronouns/PIN section.
image

tgui/packages/tgui/interfaces/IDComputer.js
```diff
@@ -0,0 +1,218 @@
...
+                        <LabeledList.Item label="Pronouns">
+                          <Button
+                            onClick={() => act("pronouns", { pronouns: "next" })}>
+                            {pronouns || "None"}
+                          </Button>
+                          {pronouns
+                      && (
+                        <Button onClick={() => act("pronouns", { pronouns: "remove" })}
+                          icon="trash" tooltip="Remove" />
+                      )}
```
ZeWaka, line 158:
```suggestion
                        && (
                          <Button onClick={() => act("pronouns", { pronouns: "remove" })}
                            icon="trash" tooltip="Remove" />
                        )}
```

ZeWaka (approved):

Garash2k:
!merge_upstream

PR #14773, opened 2023-06-29
Lube foam!

M-Earthfire:
[Chemistry][Feature]

## About the PR 

This PR enables that, if 25% of a chemical mixture inside a foam is lube/harmlube, the foam takes on the slippery properties of that particular type of lube.

## Why's this needed? 

I had a talk ago with yass, like 3 months ago, who told me that fake cleaner grenades actual use was to make people slip and dissolve their items. Since the mechanics surrounding slipping and foam had changed, it kinda lost this use.

Since fake cleaner grenades are considered fairly weak currently, this PR adds the slippery properties back to it. 

Also, this PR makes the behaviour apply to foam in general, since it is considered weak in comparison to smoke powder. One of the main reason behind this is that foam applies chemicals only if people slip in it or if the foam dissolves and only on touch. With slipping happening very rarely with foam in its current state and dissolving happening very delayed, foam left much to be desired in comparison to smoke powder, which applies its chems instantly in the area on creation with ingest and/or touch.

With enabling slipping more often in lube-foam (although with less splashed chemicals because of the dilution of the chems with lube), this PR tries to decrease the gap somewhat.

For application of lube, smoke poweder still leaves superior because of the lesser visibility of it in comparison to foam. For fighting or chases, both methods are inferior in comparison to manual application of lube on the ground.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Having at least 25% lube in a foam mixture, water and fluorosurfactant excluded, applies the lube's slippery properties to the foam
```


flappybatpal:
I don't think mass lube is a good idea at all outside of the antag grenade.

We already have issues with spray bottles and flamethrowers.

M-Earthfire:
I can understand that thought. The thing is foam already spreads the lube on the groumd, just after it dissipates (this means with a 2-3 second delay)

And lube smoke already mass lubes floors directly in a wide area.

flappybatpal:
> I can understand that thought. The thing is foam already spreads the lube on the groumd, just after it dissipates (this means with a ~4 second delay). So mas lubing with foam is already a thing.
> 
> This PR more or less applies the lube effect in the time the foam lingers and thus bridges the gap.
> 
> Restricting it to the fake cleaner grenade would mean making a new form of foam chemical. I could do that, but would redesign foam from the ground up, since in its current form, without slips, its mostly useless in delivering chems.
> 
> Besides, lube smoke already mass lubes floors directly in a wide area.

Sorry, my misunderstanding of what this does!

deathrobotpunch:
i never knew the fake cleaner grenades were supposed to melt your items

PR #14777, opened 2023-06-29
Fixes for mob spawners

CheffieGithub:
## About the PR 
Fixes death alert spam and laysound sounds. Adds puritan to stop revival. Adds a variable for PDAs which can be set to stop the device showing on scans to prevent the list getting clogged by corpses.

## Why's this needed? 
10x owlery death alerts not good nor is the pda list getting filled with dead PDAs.


code/obj/mob_spawn.dm
```diff
@@ -95,8 +95,14 @@
 		if (H.r_hand)
 			qdel(H.r_hand)
 
+		SPAWN(1)
+			for (var/obj/item/implant/health/implant as anything in H.implant)
+				qdel(implant)
```
TobleroneSwordfish, line 100:
```suggestion
				qdel(implant)
			H.implant = list()
```
best not to leave references to qdel'd things lying around

TobleroneSwordfish (changes requested):

Tarmunora:
I think the corpses' PDAs might be showing up when scanning for PDAs - probably want to fix that one way or another

CheffieGithub:
I have an idea for how to fix the scanning which could also be extended to salvagers and the ops pda, probably

CheffieGithub:
Added a variable for PDAs to indicate scannablity if this is a good solution it should be added to pda2/salvager and pda2/syndicate I think

PR #14784, opened 2023-06-30
New Book: Where am I?

TheTyrant25:
## About the PR
Adds a new book called "Where am I?" describing the Frontier system.
## Why's this needed?
The lore is established but it is not in the game. Put it in the game.
## Changelog
```changelog
(u)Tyrant
(+)There is a new book: "Where am I? A guide to the Frontier System"
```


TheTyrant25:
notes:
the sprite for the book is coder art (bad)

UrsulaMejor:
I personally don't like the level of detail. Broad strokes are fine, and already established things (such as Abzu and the fact that Clarion went there)  are fine, but discord in jokes (like Shidd and Fugg) live their best life outside the game itself, and naming every planet and moon is too much.

I feel that we shouldnt "lock in" lore unless that lore has a material presence in the game and gameplay. This way, we keep ourselves free to pivot on any non established piece in order to make the content we want to make without having to worry about finding a place for it in an already existing map. 

The 12 moons at will exist and have names when there are twelve azones for them, and until then any fact about them (including the idea that there are 12) should stay out of the game. Ditto the info on the red sun and its planets, etc

TobleroneSwordfish:
Not sure the casual and self-referrential style fits for a sort of encyclopedia
image
image

Tarmunora:
> I personally don't like the level of detail. Broad strokes are fine, and already established things (such as Abzu and the fact that Clarion went there) are fine, but discord in jokes (like Shidd and Fugg) live their best life outside the game itself, and naming every planet and moon is too much.
> 
> I feel that we shouldnt "lock in" lore unless that lore has a material presence in the game and gameplay. This way, we keep ourselves free to pivot on any non established piece in order to make the content we want to make without having to worry about finding a place for it in an already existing map.
> 
> The 12 moons at will exist and have names when there are twelve azones for them, and until then any fact about them (including the idea that there are 12) should stay out of the game. Ditto the info on the red sun and its planets, etc

Agreed. I'm a fan of the "show, don't tell" approach. generally.

TheTyrant25:
so, less detail?

ZeWaka (commented):
yeah, it seems like this information might be better split into multiple places.
also makes no sense to add a book if it's not used anywhere

TheTyrant25:
oh right silly me, forgot to put it in vendors and stuff
anyway, i'll close this now

PR #14794, opened 2023-07-01
Reworks brew_result into associated list for brew amounts & adds random fruit spawner

Glamurio:
[Feature] 
## About the PR 
The PR adds two new random spawners under `/obj/random_item_spawner/fruits` and `/obj/random_item_spawner/fruits/veggies` that can be used to spawn a random assortment of each.

Additionally, it adds a new alcoholic beverage "Schnapps", which is made from fermented fruit and contains a lot of alcohol. While I was at it, I have added several new brewable recipes to fruits in particular that had none, such as the aforementioned schnapps, wine to berries, fresh coffee to coffee berries and Piña Colada to pineapples and coconuts.
It also adds banana juice to bananas.

This is paired alongside a change to stills that allows them to be supplemented with a `brew_amount` to fine-tune the amount of reagents created, rather than a blank 20 units every time. This is done via associated lists.

Finally, it adds a room to showcase the new features.

image

The room contains two fruit spawners, and one veggie spawner in the fridge.

## Why's this needed? 
The fruit & veggie spawners will make it easier to make more dynamic rooms in the future.

Many fruits were not usable in the still, despite being very prominently brewable. Schnapps was added because it's a very common fruity alcohol that served as a good addition to the existing line up, and "Peach Schnapps" already existed.

I'd also like to do a follow-up PR at some point to overhaul the brewing system in general and allow the still to produce beverages that contain more than one simple final product.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Adds a new strong alcoholic beverage, Schnapps.
(+)More fruits are now brewable in the still.
(+)New scientific breakthroughs have led to the discovery of banana juice.
```


Tarmunora (changes requested):
I think I'm going to have to say no to the banana-juice part here - a chemical that decays over time into potassium in the bloodstream would be... not a good idea, considering potassium-water reactions etc.

Glamurio:
> I think I'm going to have to say no to the banana-juice part here - a chemical that decays over time into potassium in the bloodstream would be... not a good idea, considering potassium-water reactions etc.

Valid criticism, I'd have thought the interaction is cool and it keeps the "bananas contain potassium". I could change it instead to making it so that bananas contain banana juice and potassium, and have banana juice simply act like other juices, I just simply didn't want to replace potassium as a whole. But I have no preference regarding this.

code/modules/chemistry/Reagents-FoodDrink.dm
```diff
@@ -303,6 +303,17 @@ datum
 			taste = "fruity"
 			reagent_state = LIQUID
 
+		fooddrink/alcoholic/schnapps
+			name = "schnapps"
+			id = "schnapps"
+			fluid_r = 240
+			fluid_g = 240
+			fluid_b = 240
+			alch_strength = 1
```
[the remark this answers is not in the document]
Azrun: Still seems like a rather HIGH ABV for schnapps based on a quick google.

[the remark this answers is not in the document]
TDHooligan: yeah, but this requires effort to make (vs bashing out 3 clicks of vodka) 
i'd happy to pass it off as 'space produce has some rockin microbes' or something

Azrun (approved):

code/modules/food_and_drink/plants.dm
```diff
@@ -1080,6 +1098,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 	slice_product = /obj/item/reagent_containers/food/snacks/plant/lemon/wedge
 	slice_amount = 6
 	slice_suffix = "wedge"
+	brew_result = "lemonade"
```
Azrun, line 1101: One lemon = 20u of lemonade
One sliced lemon = 120u of lemonade?

code/modules/food_and_drink/plants.dm
```diff
@@ -1050,6 +1067,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 	slice_product = /obj/item/reagent_containers/food/snacks/plant/lime/wedge
 	slice_amount = 6
 	slice_suffix = "wedge"
+	brew_result = "limeade"
```
Azrun, line 1070: One lime = 20u of limeade
One sliced lime = 120u of limeade?

code/modules/food_and_drink/plants.dm
```diff
@@ -872,6 +884,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 	icon_state = "poison"
 	food_color = "#AC1515"
 	initial_volume = 100
+	brew_result = "capulettium"
```
Azrun, line 887: This ensures 20 units instead of engaging with potency system, which seems like a cleaner means of developing the poison.

code/modules/food_and_drink/plants.dm
```diff
@@ -634,6 +636,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 	heal_amt = 2
 	initial_volume = 100
 	food_effects = list("food_refreshed")
+	brew_result = "capsaicin"
```
Azrun, line 639: This ensures 20 units instead of engaging with potency system. Seems like this would make generating smokes a lot easier.

Azrun (commented):
I think I'd have preferred the routes for harmful reagents not been part of this PR. Cool conceptually though.

code/modules/food_and_drink/plants.dm
Glamurio: This is definitely more of a statement on outdated brewery code rather than an issue with the brew result. Will fix.

Glamurio:
@Azrun I've reworked the way stills produce reagents now. The whole process now expects a `brew_amount`, which defaults to the pre-existing 20 units. Supplementary, any snack can supply a custom amount of the reagents, and additionally, you can supply a list of amounts alongside a list of reagents, and it will set an amount for each.

Coinciding with that, I have drastically lowered the amount of capulettium obtained per poison apple, and have replaced the majority of its content with cider. I have considered tying the amount of content per reagent to its potency, but was worried that it could be too easily abused, but it's an interesting thought.

I've also removed capsaicin from brew results, as I could not find a good enough brew result to alongside it to justify it remaining.

Let me know if these changes are better, and if anything else needs changing.

code/obj/item/plants.dm
```diff
@@ -10,6 +10,7 @@ ABSTRACT_TYPE(/obj/item/plant)
 	desc = "You shouldn't be able to see this item ingame!"
 	icon = 'icons/obj/hydroponics/items_hydroponics.dmi'
 	var/brew_result = null // what will it make if it's brewable?
+	var/brew_amount = 20 // how much brew could a brewstill brew if a brewstill still brewed brew?
```
Azrun, line 13: Why not an associated list so each reagent also specifies the amount?
`list("capulettium"=5, "cider"=15)`
`list("luminol"=20)`

code/modules/food_and_drink/plants.dm
Azrun: How so?  As implemented that is the behavior, otherwise I would have possibly just expected the whole fruit to brew the result and then the children to clear the brew_result.

code/modules/food_and_drink/plants.dm
```diff
@@ -1050,6 +1069,8 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 	slice_product = /obj/item/reagent_containers/food/snacks/plant/lime/wedge
 	slice_amount = 6
 	slice_suffix = "wedge"
+	brew_result = "limeade"
+	brew_amount = 30
```
Azrun, line 1073: Why does this result in a higher yield than other products?

code/modules/food_and_drink/plants.dm
```diff
@@ -1080,6 +1102,8 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 	slice_product = /obj/item/reagent_containers/food/snacks/plant/lemon/wedge
 	slice_amount = 6
 	slice_suffix = "wedge"
+	brew_result = "lemonade"
+	brew_amount = 30
```
Azrun, line 1106: Why does this result in a higher yield than other products?

code/obj/item/plants.dm
Azrun, line 13: I'm not sure `brew_amount` is necessary? What is the benefit of having multiple ways of defining things.
```
brew_result = "water"
brew_amount = 5
```
```
brew_result = list("water"=5)
```

code/modules/chemistry/tools/food_and_drink.dm
```diff
@@ -12,6 +12,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food)
 	var/custom_food = TRUE 						//! Can it be used to make custom food like for pizzas
 	var/festivity = 0 							//! Amount of cheer this food adds/subtracts when eaten
 	var/brew_result = null 						//! What reagent will it make if it's brewable?
+	var/brew_amount = 20 						//! How much units of the brewed reagent?
```
Azrun, line 15: Is nessy? 🦕

code/obj/item/plants.dm
Glamurio: Primarily, it's so that older stuff uniformly uses the newer system and I don't have to manually rewrite every single brew_result into a list, since the PR was already pushing itself a bit out of scope. But I'll just do that then.

code/modules/food_and_drink/plants.dm
Glamurio: Because wedges were brewable, and I wanted it to make sense that 6 wedges give 5u each to a combined 30u.

Azrun (approved):

PR #14797, opened 2023-07-01
more atmos balloon options

hobartbartholomewpepperidgejr:
## About the PR 
you can fill balloons from most gas containers now as opposed to just large gas canisters
## Why's this needed? 
being restricted to only large gas canisters for balloon shenanigans is really boring, considering what kinds of wizardry you can get up to inside air scrubbers


code/modules/atmospherics/portable/portable_atmospherics.dm
```diff
@@ -142,5 +142,33 @@
...
+				user.update_inhands()
+				return
+			else
+		var/datum/gas_mixture/removed = src.air_contents.remove(amount)
+		balloon.air.merge(removed)
+		balloon.UpdateIcon()
+		playsound(get_turf(src), 'sound/machines/hiss.ogg', 50, 1)
+		user.visible_message("<span class='notice'>[user] fills [balloon] from src.</span>", "<span class='notice'>You fill [balloon] from [src].</span>")
+		return
 	return
```
frawhst, line 174:
```suggestion
```
Empty `return` at the end of a proc doesn't do anything (since the default behavior is to return anyways)

code/modules/chemistry/tools/balloons.dm
```diff
@@ -183,6 +183,10 @@
 					qdel(src)
 
 			if ("Inhale")
+				if (src.tied == TRUE)
```
frawhst, line 186:
```suggestion
				if (src.tied)
```
Slightly prettier, if something is only ever `TRUE/FALSE` we prefer to just check `if (thing)`
(or `if (!thing)` if it should be FALSE)

code/modules/atmospherics/portable/portable_atmospherics.dm
```diff
@@ -142,5 +142,33 @@
...
+			boutput(user, "<span class='alert'>[balloon] is already full.</span>")
+			return
+		else if (istype(src, /obj/machinery/portable_atmospherics/scrubber))
+			var/fillfrom = tgui_alert(user, "Fill balloon from liquid tank or air tank?", "Fill balloon", list("Air tank", "Liquid tank", "Cancel"))
+			if (fillfrom == "Cancel")
+				return
+			else if (fillfrom == "Liquid tank")
+				if (!src.reagents.total_volume && src.reagents)
+					user.show_text("[src] is empty.", "red")
+					return
```
frawhst, line 158:
```suggestion
				if (src.reagents && !src.reagents.total_volume)
					user.show_text("[src] is empty.", "red")
					return
```
This doesn't work as written, if `reagents` is null then the first statement will just cause an error since it's trying to access a var of `null`

hobartbartholomewpepperidgejr: thanks for these!!! i'd copied this bit from balloons.dm and only swapped a few values around, so something like this would've never occurred to me

frawhst (approved):

ZeWaka:
please label your PRs

flappybatpal:
!merge_upstream

PR #14807, opened 2023-07-01
Add lanyards to tourist (also adds blob and cow mutantrace)

Glamurio:
[Feature] [Respawning]
## About the PR 
This PR makes all tourists start with a lanyard in their ID slot, which contains their ID.

It also adds `/datum/mutantrace/blob` and `/datum/mutantrace/cow` to its possible mutant races, putting it in line with the Ambassador job.

## Why's this needed? 
Lanyards are cool, tourists seem like the perfect job for it because they don't start with a PDA.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Tourists now spawn with a lanyard containing their ID.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -1323,6 +1323,8 @@ ABSTRACT_TYPE(/datum/job/civilian)
 		M.equip_if_possible(L, M.slot_wear_id, FALSE)
 		M.spawnId(src)
 		var/obj/item/card/id = locate() in M
+		if (!id)
+			return
 		L.storage.add_contents(id, M, FALSE)
 
 /datum/job/special/space_cowboy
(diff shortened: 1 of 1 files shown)


CaioFrancisco:
> It also adds `/datum/mutantrace/blob` and `/datum/mutantrace/cow` to its possible mutant races, putting it in line with the Ambassador job.

Devilish sneaky change. I've got no problems with it though, and I'm sure it's a welcome change.

Glamurio:
> Devilish sneaky change. I've got no problems with it though, and I'm sure it's a welcome change.

I didn't mean to be devilish or sneaky, my original intent was to add the lanyards, then I saw someone asking if mutant races might also be added to tourist, and since it's basically the same lines of code I figured I can group it. I'll atomize if necessary, but it just seems really minor that I didn't bother.

frawhst (approved):
Seems good

Roninnozlo:
i hope someone turns blob into an actual mutrace at some point and not just a human with less equip slots, a bad language flag, and a different sprite (that doesn't even fit with equipment overlays)

same with martian... my boys need some love

anyway, a welcome change

PR #14813, opened 2023-07-02
Adjustable IV drips + Helptext

TDHooligan:
[MEDICAL][QOL][BALANCE]
## About the PR 
By using the grab intent on IV drips, you can now adjust the drip rate in increments of 1, from 1u/tick to 5u/tick.
This feature **requires the IV drip to be on a drip stand.** 

Also adds helptext to describe this functionality.
Grab was picked because it's a niche intent for surgery, and technically you're 'grabbing' an anchor clamp or whatever.

## Why's this needed? 
Some players have voiced that the IV tends to overfill patients very quickly. 
This lets you fine tune drip-rates to only give a patient the medicine they need, rather than having them walk out of medbay dosed up on 100+ units of saline for a single operation.

## Changelog 

```changelog
(u)TDHooligan
(+)You can now adjust the drip rate of IV bags with the grab intent. This only works when the bag is properly hung from a stand!
```

### DIFF
--- a/code/modules/chemistry/tools/iv_drips.dm
+++ b/code/modules/chemistry/tools/iv_drips.dm
@@ -26,7 +26,10 @@
 	var/mode = IV_DRAW
 	var/in_use = 0
 	var/slashed = 0
+	var/hung_transfer_rate = 5 //how much this injects/pulls when on an IV stand
+	var/standard_transfer_rate = 5 // how much this injects/pulls when not on a stand
 	var/max_transfer_rate = 5
+
 	HELP_MESSAGE_OVERRIDE({"
 		Use this item in-hand with the <b>grab</b> intent to adjust the drip rate,
 		use any other intent to switch between drawing blood or injecting."})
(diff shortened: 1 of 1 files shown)


DisturbHerb:
this would make IV drips very, very powerful. i would recommend adding this functionality to the poles, if only to prevent people from running around being drip-fed with all the meth they could possibly want below overdose levels

TDHooligan:
> this would make IV drips very, very powerful. i would recommend adding this functionality to the poles, if only to prevent people from running around being drip-fed with all the meth they could possibly want below overdose levels

totally forgot about overdose levels. will add that now

TobleroneSwordfish:
So you have to use it in-hand while on grab intent to adjust the rate at which it drips only when on a stand? That seems incredibly confusing, would be much better as an interface on the stand or something.

TDHooligan:
Could be, but I'm not one for UI work. I can match the behaviour when you hit the IV stand with 'grab' intent. 
I don't know if the value is going to be changed frequently or precisely enough to require a whole window dedicated to it.

PR #14816, opened 2023-07-02
Changes damaged/scorched turfs in maps to use a mapping helper object instead

CheffieGithub:
## About the PR 
Changes floors and the damaged1/scorched2/etc tiles to use a burner / damager object instead.
Changes mirrored on unsimulated turfs by adding a mirror of the `burn_tile()` and `break_tile()` procs.

These procs select a random scorched or burnt tile and sets the relevant variables
Doesn't remove the old subtypes from code due to them being used by map generation.

## Why's this needed? 
image

From left to right:

Its also a pain to navigate the subtype trees of turfs.


code/turf/floors_unsimulated.dm
```diff
@@ -30,6 +34,62 @@
 			return
 	..()
 
+/turf/unsimulated/floor/proc/break_tile(var/force_break)
```
frawhst, line 37: Since all these procs are basically copy pasted from `/turf/simulated/floor`, I'd rather just put them on the parent `/turf` and use vars to track whether something can be burnt/broke. 
ie add `can_burn` and `can_break` as vars on `/turf` and then move these procs there too, with
```
if (!can_break) // (or burn)
  return
```
at the top.

frawhst: (yes turf inheritance is a pain)

frawhst (commented):
Overall seems helpful.
If you're replacing all instances of `/turf/simulated/floor/damaged#`, should probably delete those types as well.

code/turf/floors_unsimulated.dm
CheffieGithub: yeah I thought that a base turf definition would be better, will do

ZeWaka:
Probably wait to do this until all the random room stuff is merged and the rest of the mapping PRS since this is very conflicty

CheffieGithub:
Thats understandable, Ill close this and update when needed. Reopen when its safe to do so.

PR #14823, opened 2023-07-03
Expands Clarion QM and Reorganizes QM Storage

DanaTarSUS:
## About the PR 
Expands Clarion QM into part of QM storage and reorganizes many of the crates and fabricators littered throughout the storage room.

## Why's this needed? 
Currently, Clarion QM itself is horribly cramped while QM storage is far too large. In addition, every other department that falls under Engineering has access to the storage space in one way or another. With that in mind, I believe that QM should have sufficient space to operate without needing to make use of a shared space to prevent their own department from becoming cluttered by crates.

## Changelog 

```changelog
(u)DeeDotVeeA
(+) Clarion QM is now slightly more spacious.
```

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -16061,6 +16061,14 @@
 	},
 /turf/simulated/floor,
 /area/station/science/lab)
+"beH" = (
+/obj/disposalpipe/segment,
+/obj/machinery/light{
+	dir = 4;
+	pixel_x = 10
+	},
+/turf/simulated/floor/grime,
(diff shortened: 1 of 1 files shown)


DanaTarSUS:
This PR is largely the same as my first attempt at this found in #14714. Seeing as I made the questionable decision to remake this pr from scratch, there are some smaller additions but the layout and contents of the rooms affected are largely the same.

frawhst (approved):
Looks good. that QM really is cramped, sheesh.

PR #14850, opened 2023-07-05
Adds the Rancher's closet

TheTyrant25:
[MAPPING] [QOL]
## About the PR
Ranching closets were manually applied across maps by var editing a closet and giving them items. This makes a secure locker variant with all the necessary items already included.
This PR DOES NOT add them to maps. That is being atomised out for merge conflict reasons.
## Why's this needed?
More convenient for mappers, and makes things more consistent across multiple maps. For instance, clarion and atlas lack a ranching locker altogether. And the contents, while the core ones stay the same, vary across maps (e.g. some maps lack syringe boxes and produce bags, while donut2 has a watering can and sponge in addition, etc.)
Once this is merged I'll start swapping them out on various maps.

### DIFF
--- a/code/obj/item/kitchen.dm
+++ b/code/obj/item/kitchen.dm
@@ -454,6 +454,9 @@ TRAYS
 		allowed_food = /obj/item/reagent_containers/food/snacks/ingredient/egg
 		contained_food_name = "egg"
 
+		rancher
+			count = 4
+
 	lollipop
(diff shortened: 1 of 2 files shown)


M-Earthfire:
If it is a secure locker, i would think about throwing a box with eggs in (like 4 eggs). That would give late-joining ranchers a security for vandals/roaches running into the ranch and eating the initial eggs away

Issue #7466, opened 2022-02-11
Windoors don't have a north facing sprite


TobleroneSwordfish:
### Labels

[BUG] [SPRITES]

### Description

Windoors' north facing sprite is a copy of the south facing one, leading to jank.

### Steps to reproduce

1. Use reverse engineering fabricator to make a windoor frame
2. Place it facing north
3. Observe that the sprite is just the south facing one and doesn't match the bounding box at all

### Expected Behavior

I expected it to have sprites for all directions

### Additional Information & Screenshots

the path is /obj/machinery/door/airlock/pyro/glass/windoor, found this while fixing another bug with them

pali6:
Ideally would be done via the south sprite shifted north but in order to layer properly that would need the SIDE_MAP map mode, except that currently disables blackness plane which makes some stuff look extremely jank. Byond feature request to change this behaviour has been made but received no reply.

PR #14853, opened 2023-07-05
Fix glassdoors having no north facing sprite and using the wrong welded icons

### LINKED ISSUE #13753: Windoors use full door welded overlays
### Labels

[BUG]

### Description

`/obj/machinery/door/airlock/pyro/glass/windoor`s don't have their own welded overlays apparently, leading to this

### Steps to reproduce

1. Weld a windoor shut
2. Observe

### Expected Behavior

I expected the welding overlay to only be along one side of the door

ugly!!!1
image

CheffieGithub:
## About the PR 
Adds a proper north facing sprite for pryo glass doors since currently they currently just use the south facing sprite.
Not perfect but its better than south facing icons.

image

Open panel overlay adjusted to fit though its not perfect.
Adds welded overlays for this door to match up.

Fixes #13753

I did see the other issue ( #7466) this addresses but it seems this is not the ideal solution. 
Maybe it should be left open until what pali suggests can be achieved?

## Why's this needed? 
the difference between south and north should be clear since they have different collision. Its not good to have the incorrect overlays.


Tarmunora (approved):
A bit hard to read the difference between open/closed, but better than just using the south sprite imo

PR #14854, opened 2023-07-05
General trinary types cleanup, and some more

Laboredih123:
[ATMOSPHERICS][CLEANLINESS][INTERNAL][BUG]
## About the PR 
Creates the trinary subtype, moves trinary things to the subtype, cleans up and adds to their files, and perhaps fixes some icons.

## Why's this needed? 
I prefer my clean and performant code and correct icons.


Laboredih123:
i forgot the last thing i was gonna do so im just gonna open

code/modules/atmospherics/machinery/trinary/manifold_valve.dm
```diff
@@ -0,0 +1,140 @@
+
+/obj/machinery/atmospherics/trinary/manifold_valve
+	icon = 'icons/obj/atmospherics/manifold_valve.dmi'
+	icon_state = "manifold_valve0"
+
+	name = "manifold valve"
+	desc = "A pipe valve"
+
+	var/divert = FALSE
```
TheTyrant25, line 9: might be worth doc commenting what divert does

code/modules/atmospherics/machinery/trinary/manifold_valve.dm
```diff
@@ -0,0 +1,140 @@
...
+	var/id = null
+
+/obj/machinery/atmospherics/trinary/manifold_valve/New()
+	..()
+	qdel(src.air1)
+	src.air1 = null
+	qdel(src.air2)
+	src.air2 = null
+	qdel(src.air3)
+	src.air3 = null
```
TheTyrant25, line 21: not sure why this nullifying is necessary?

code/modules/atmospherics/machinery/trinary/mixer.dm
```diff
@@ -0,0 +1,187 @@
...
+
+/obj/machinery/atmospherics/trinary/mixer
+	name = "Gas mixer"
+	icon = 'icons/obj/atmospherics/mixer.dmi'
+	icon_state = "intact_off"
+	layer = PIPE_MACHINE_LAYER
+	plane = PLANE_NOSHADOW_BELOW
+
+	var/id_tag
+	var/master_id
```
TheTyrant25, line 15: what are these id tags for btw

code/modules/atmospherics/machinery/trinary/mixer.dm
```diff
@@ -0,0 +1,187 @@
...
+
+	var/target_pressure = ONE_ATMOSPHERE
+	var/node1_ratio = 0.5
+	var/node2_ratio = 0.5
+
+	var/frequency
+
+/obj/machinery/atmospherics/trinary/mixer/New()
+		..()
+		air3.volume = 300
```
TheTyrant25, line 26: why is this not just the default value declared higher up? Does something in a parent override it?

TheTyrant25 (commented):
Didn't read retrofilter or trinary base, but most of it looks okay. I do have some questions on how it all works, but at a surface level glance over by somebody not familiar with the code, all I can really say is "use src.variablename more please" and "doc comment more vars and procs".

code/modules/atmospherics/machinery/trinary/manifold_valve.dm
Laboredih123: we don't use these gasmixes, so i disposed of them to save memory. It's explicitly set to null to clear our reference to it.

code/modules/atmospherics/machinery/trinary/mixer.dm
Laboredih123: Asking stuff and modifying what the mixer mixes.

code/modules/atmospherics/machinery/trinary/mixer.dm
Laboredih123: The parent usually does 200L, the mixer uses 300L

code/modules/atmospherics/machinery/trinary/manifold_valve.dm
Laboredih123: ight

code/modules/atmospherics/machinery/trinary/manifold_valve.dm
TheTyrant25: fair enough

code/modules/atmospherics/machinery/trinary/mixer.dm
TheTyrant25: does `air3.volume = 300` not work in the `/obj/machinery/atmospherics/trinary/mixer` part then

Laboredih123: gotta do it in New, cause otherwise the gas doesn't exist yet.

TheTyrant25: i see

PR #14855, opened 2023-07-05
Arithmatic Component and Association component extended functionality

Coll6:
## About the PR 

Adds the function to pick which input causes the arithmetic component to activate auto evaluate.
Adds the function to send associated values using index position. Now values can be picked by key and index.

## Why's this needed? 
The arithmetic component has two modes. Either auto evaluate when A or B is set and don't evaluate and send a specific signal to do so. This allows the component to set which input triggers auto evaluate as well as retaining triggering on Both.

The association component is much like a database to grab and store values and data. You can search by a key value and even send a random one but you cannot send based on index or position in that list of associations. This allows you to send a value based on index. Sending keys will not work as it only accepts numbers.


ZeWaka:
please label your PRs

ZeWaka:
.

PR #14856, opened 2023-07-05
Remove maintenance cameras for cogmap1

flappybatpal:
## About the PR 
Removes nearly all cameras from cogmap1 maintenance.

A camera in disposals was moved to reduce it's vision into disposals belt hell.

Cameras I could not decide if should be removed:
Air hookup
Router
Electrical Substation
North Solar Maintenance
Exterior cameras

## Why's this needed? 
Cogmap1 has a very high level of AI camera coverage with almost all of maintenance covered.
In comparison, Donut3 has almost no camera coverage of maintenance and Cogmap2 only has a few areas covered.

## Changelog 

```changelog
(u)Flappybat
(*)AI camera coverage of Cogmap1 maintenance has been significantly reduced.
```

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -9770,6 +9770,11 @@
 /area/station/crew_quarters/catering)
 "aDp" = (
 /obj/machinery/portable_atmospherics/canister/air/large,
+/obj/machinery/camera{
+	c_tag = "autotag";
+	dir = 1;
+	name = "autoname - SS13"
+	},
 /turf/simulated/floor/plating,
 /area/station/maintenance/solar/west)
 "aDq" = (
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Keeping the cameras covering the solar maint rooms is probably a good idea since that's a standard AI thing, air hookup might be nice too to remind AIs they can use it, but it's also an oft crimed area so 🤷

Sord213 (commented):
Yeah I'd probably put the cameras back in the solar substations. A lot of AIs will set solars and it's technically an engineering area, makes sense to be visible. Otherwise everything else seems good to me.

flappybatpal:
Merging my own change based on positive reception from devs here!

PR #14881, opened 2023-07-08
Reworks Interdictor Mob Protection

Wisemonster:
[BALANCE] [STATUS EFFECTS] [INPUT WANTED]

## About the PR 
This pr refactors the interdictor's mob protection effects for wormholes/radstorms/magnetic biofields to a status effect. This also makes it so the interdictor's power drain will be affected by the amount of players in it's AoE, and not be drained whenever someone within the AoE negates a portal/magnetic biofield.

## Why's this needed? 
Currently, not many people know what interdictors do, and not many people realize that being within an interdictor during a radstorm protects you. This pr aims to make it so that people are more aware that the interdictor protects them from wormholes/radstorms/magnetic biofields, and make it so that the amount of people protected will influence power drain to a degree.

## Changelog 

```changelog
(u)Wisemonster
(*)Interdictors will now drain 6u of power per player in their influence
(*)Removed power drain on interdictors from negating worm holes and magnetic biofields
(+)Changed the interdictor protection to be tied to a status effect
(+)Added a particle effect for the interdictor status effect
```

### DIFF
--- a/code/mob/living/life/Life.dm
+++ b/code/mob/living/life/Life.dm
@@ -277,7 +277,7 @@
 			//Interdictor's protections for mobs
 			if (isliving(src) && !isintangible(src))
 				for_by_tcl(IX, /obj/machinery/interdictor)
-					var/area/A = src.loc.loc
+					var/area/area = get_area(src)
 					if (IX.expend_interdict(6,src,TRUE)) //This protects mobs from radstorms/wormholes/magnetic biofields
 						src.changeStatus("spatial_protection", 3 SECONDS)
 					if ((src.loc && isarea(A)) && A.irradiated)
--- a/code/mob/living/life/Life.dm
+++ b/code/mob/living/life/Life.dm
@@ -277,8 +277,11 @@
 			//Interdictor's protections for mobs
 			if (isliving(src) && !isintangible(src))
 				for_by_tcl(IX, /obj/machinery/interdictor)
+					var/area/A = src.loc.loc
 					if (IX.expend_interdict(6,src,TRUE)) //This protects mobs from radstorms/wormholes/magnetic biofields
 						src.changeStatus("spatial_protection", 3 SECONDS)
+					if ((src.loc && isarea(A)) && A.irradiated)
+						IX.resisted = TRUE
 					if (!iscarbon(src)) //Prevents non-carbons from getting the Zephyr stam boost, but still protects other mobs
 						break
 					if (IX.expend_interdict(4,src,TRUE,ITDR_ZEPHYR)) // Zephyr-class interdictor: carbon mobs in range gain a buff to stamina recovery, which can accumulate to linger briefly
--- a/code/mob/living/life/Life.dm
+++ b/code/mob/living/life/Life.dm
@@ -280,7 +280,7 @@
 					var/area/area = get_area(src)
 					if (IX.expend_interdict(6,src,TRUE)) //This protects mobs from radstorms/wormholes/magnetic biofields
 						src.changeStatus("spatial_protection", 3 SECONDS)
-					if ((src.loc && isarea(A)) && A.irradiated)
+					if (istype(area) && area.irradiated)
 						IX.resisted = TRUE
 					if (!iscarbon(src)) //Prevents non-carbons from getting the Zephyr stam boost, but still protects other mobs
 						break
--- a/code/mob/living/life/Life.dm
+++ b/code/mob/living/life/Life.dm
@@ -274,10 +274,14 @@
 					animate(src, transform = matrix(), time = 1)
 				last_no_gravity = src.no_gravity
 
-			// Zephyr-class interdictor: carbon mobs in range gain a buff to stamina recovery, which can accumulate to linger briefly
-			if (iscarbon(src))
+			//Interdictor's protections for mobs
+			if (isliving(src) && !isintangible(src))
 				for_by_tcl(IX, /obj/machinery/interdictor)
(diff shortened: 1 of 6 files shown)


code/mob/living/life/disability.dm
```diff
@@ -37,12 +37,9 @@
 			if (A.irradiated)
 				//spatial interdictor: mitigate effect of radiation
 				//power expenditure is managed centrally by the interdictor
-				var/interdictor_influence = 0
-				for_by_tcl(IX, /obj/machinery/interdictor)
-					if (IX.radstorm_interdict(owner))
-						interdictor_influence = 1
-						break
-				if(!interdictor_influence)
+				if (owner.hasStatus("spatial_protection"))
```
Kubius, line 40: Removing the radstorm_interdict call here means the baseline cost of interdicting a radstorm (independent of number of protected individuals) doesn't apply. My recommended change to preserve the baseline cost would probably be some functionality inside the main process() to apply that baseline cost if there's a currently active mass radiation event and the interdictor applied at least one protection buff on the last tick; to compensate for the per-user cost, it could probably be reduced slightly as well.

code/mob/living/life/disability.dm
```diff
@@ -37,12 +37,9 @@
...
+					return
```
Kubius, line 41: Returning here prevents the checks further below from happening. Could invert the hasStatus check with ! instead.

code/mob/living/life/Life.dm
```diff
@@ -274,10 +274,14 @@
...
-			if (iscarbon(src))
+			//Interdictor's protections for mobs
+			if (isliving(src) && !isintangible(src))
 				for_by_tcl(IX, /obj/machinery/interdictor)
-					if (IX.expend_interdict(4,src,TRUE,ITDR_ZEPHYR))
+					if (IX.expend_interdict(6,src,TRUE)) //This protects mobs from radstorms/wormholes/magnetic biofields
+						src.changeStatus("spatial_protection", 3 SECONDS)
+					if (!iscarbon(src)) //Prevents non-carbons from getting the Zephyr stam boost, but still protects other mobs
+						break
+					if (IX.expend_interdict(4,src,TRUE,ITDR_ZEPHYR)) // Zephyr-class interdictor: carbon mobs in range gain a buff to stamina recovery, which can accumulate to linger briefly
```
Kubius, line 284: As there's now an ongoing cost from having people in range at all, the Zephyr cost could probably be reduced, or skipped entirely (making it simply check for mob's carbon status and interdictor's board type inside of the main expend_interdict). Not vital, but worth considering.

Kubius (commented):
I think this is a very solid idea overall for intuitive awareness of interdictors' protection functionality - exact power draw values might need tuning, but that's generally the case no matter what you change.

It's been mentioned that the buff could have a small visual effect, and if that's something you're inclined to do I think it might be interesting to have that as a faint golden "aura" underlaid around protected individuals, somewhat sporadic and sparkling in appearance.

Wisemonster:
> It's been mentioned that the buff could have a small visual effect, and if that's something you're inclined to do I think it might be interesting to have that as a faint golden "aura" underlaid around protected individuals, somewhat sporadic and sparkling in appearance.

I don't really have enough knowledge with making visual effects to make something more advanced than simply the magnetized aura, but yellow.

Kubius:
That'd probably be good enough, yeah.

code/mob/living/life/disability.dm
Wisemonster: Alright, done

code/mob/living/life/Life.dm
Wisemonster: I could do that, although I'm unsure how costly the interdictors are atm, and having one status effect from interdictors drain battery, and not the other could be a bit confusing

code/mob/living/life/disability.dm
Wisemonster: By the main_process(), do you mean for the radstorm, or interdictor?

Wisemonster: Ok, so I simply added the radstorm_interdict proc to the end of the interdict proc, and it seems to work. It seems that something is already checking whether or not a radiation storm is going on for the proc. Although as a side effect, it will start draining the increased power when the event is announced, rather than when the radiation hits the station. If this is a problem, I'll try looking for another solution.

Kubius (approved):
LGTM. Haven't locally compiled it on my end but appears structurally correct; passive protection draw might need tuning, but that can be done easily down the line, and these values are a solid starting point.

TobleroneSwordfish (changes requested):
making interdictors less opaque is good 👍

code/modules/status_system/statusEffects.dm
```diff
@@ -2404,3 +2404,19 @@
...
+/datum/statusEffect/interdictor //Status effect for letting people know they are protected from some spatial anomalies
+	id = "spatial_protection"
+	name = "Spatial Protection"
+	desc = "You are being protected from wormholes, radiation storms, and magnetic biofields."
+	icon_state = "blocking" //This gives the general idea that they are being protected, but could use a better icon
+	maxDuration = 4 SECONDS
+	effect_quality = STATUS_QUALITY_POSITIVE
+
+	onAdd(optional=null)
+		owner.add_filter("protection", 1, outline_filter(color="#e6ec21"))
```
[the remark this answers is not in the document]
Wisemonster: NCS Atlas 7_28_2023 6_38_04 AM

The biofield outline appears to be closer to the mob than the interdictor outline, but both are visible

PR #14892, opened 2023-07-10
Big landmark visual update and pathing update

CheffieGithub:
## About the PR 
Most landmarks now have unique icons and concrete paths instead of varedits.

Icons for antagonists and jobs are taken from spawned normal NPC humans and are 32x48.

Antagonist spawns:
image

Job spawns:
image
image

Misc:
image

## Why's this needed? 
Its more clear to not have 50 of the same blue X / player start which are varedited with different purposes.


ZeWaka (approved):
Gaming

Sord213 (approved):
oh hell yeah. I've wanted to do this for ages but was super lazy

Laboredih123:
wow very cool

Issue #14891, opened 2023-07-09
Unscrewing and re-screwing a roundstart computer duplicates some of their pheripherals


Garash2k:
### Labels

[BUG]

### Description

Unscrewing and re-screwing a roundstart computer duplicates some of their pheripherals

### Steps to reproduce

1. Find a computer or terminal such as the ones in tech storage
2. Unscrew and rescrew it
3. Notice that there's an extra Card input that popped up
4. Also: deconstruct to the point of prying out the pheripherals and notice that there are extras
![image](https://github.com/goonstation/goonstation/assets/6396368/ba15c71c-4d66-4a17-8924-e95f13eb8eb7)

### Expected Behavior

No duped pheripherals

### Additional Information & Screenshots

MMMM
![image](https://github.com/goonstation/goonstation/assets/6396368/03bcf4d4-4167-427b-b252-d4d15a868c8d)

Some passerby told me this was a long standing bug but searching for Pheripherals I didn't see it already reported

PR #8518, opened 2022-05-08
TGUI computer3 interface (Dwaine/Thinkdos computers)

Skeletonman0:
## About the PR 

image

This PR is my first attempt at making a TGUI interface.

the buttons on the bottom are the peripherals (disk drive, id scanner, etc.)

if you click on them while holding an id or a disk, and the item is correct for the slot, it will be inserted into the computer.
this is intended to be similar to how the old GUI handled peripherals.

if the peripheral has a disk / id inside, the button will turn green.

also the TGUI interface is resizable

## Why's this needed? 

this PR's TGUI version is probably better than what we have currently:

image

PR #14895, opened 2023-07-10
TGUI Advanced Computer

Garash2k:
## About the PR 
Adopting #8518!
image
I've fixed a handful of things to have feature parity with the old html UI, plus some minor goodies such as being able to click a card into a computer and ejecting stuff in hands, see the commit titles for details.

Here's a list of concerns I had while adopting/testing and why they do not apply/are out of scope:
- outstanding original pr points -- all fixed before being adopted
- Topic had fingerprints, does ui_act need em? -- finger prints are added upon opening the ui, no need for extra code
- What happened to var/font_size -- TGUIs don't care about the "Popup Font Size" setting so no need to care
- What happened to user.literate -- ghost critters are blocked from using most TGUI and that's what I assume it was for so we good
- bug: unscrewing and rescrewing causes an extra Card button to appear -- Outsanding bug #14891
  - Make terminals be able to survive being unscrewed rescrewed -- Caused by #14891
- Print module says "Command unsuccessful." upon working, normal? -- yup, tiny bug TKTK bug report
- terminal magically works with only a terminal frame, id scan, network card and a blank drive, normal? -- I guess it's fine, same behaviour as pre TGUI

## Why's this needed? 
Nicer UI that's more responsive and is re-sizable.

## Changelog 

```changelog
(u)Skeletonman0 and Garash
(+)New UI for the station's advanced computers, go bother cargo for a Home Networking Kit and build your own dream computer today!
```

[Feature][UI]


code/modules/networks/computer3/peripherals.dm
```diff
@@ -942,12 +942,22 @@ TYPEINFO(/obj/item/peripheral)
 		desc = "A peripheral board for editing ID cards."
 		can_manage_access = 1
 
+		return_badge()
+			// label text, icon, contents
+			var/dat = list("label" = "Card","icon" = "edit","contents" = src.authid)
+			return dat
```
ZeWaka, line 948:
```suggestion
			. = list("label" = "Card","icon" = "edit","contents" = src.authid)
```
ditto for the rest

ZeWaka (changes requested):

PR #14896, opened 2023-07-10
Fix balloons not bursting

### LINKED ISSUE #14884: Filling balloons in contaienrs can fail to burst them
### Labels

[BUG][MAJOR]

### Description

filling balloons with chames whilst they are in a container can display the message "The balloon bursts in your hands! <curse>!" whilst the balloon remains intact

### Steps to reproduce

Get box of balloons
open box of balloons
pick up chem container
fill balloons 
observe chat

### Expected Behavior

constsiant between chat and reality

Reported by: Balarak
Active test merges: #14694
Round log date: 2023-07-08-13-07
Reported on: 2023-07-08 13:35:21
Map: COGMAP

TheTyrant25:
[MAJOR]
## About the PR
Changes `smash()` to accept an atom instead of a turf, and then does get_turf on that atom. Then it turned out that this was not the problem whatsoever and it was an early return in `smash()` that was doing it. So fixed that, plus did some minor cleanup on balloon.dm. Bigger refactor coming soon ofc.

## Why's this needed?
Fixes #14884

### DIFF
--- a/code/modules/chemistry/tools/balloons.dm
+++ b/code/modules/chemistry/tools/balloons.dm
@@ -67,12 +67,12 @@
 			return
 		if (!user && usr)
 			user = usr
-		else if (!user && !user && ismob(src.loc))
+		else if (!user && ismob(src.loc))
 			user = src.loc
 		if (!ohshit)
 			ohshit = (src.reagents.total_volume /  (src.reagents.maximum_volume - 10)) * 33
 		if (prob(ohshit))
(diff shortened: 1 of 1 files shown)


TheTyrant25:
found the actual issue, it was an early return if there was less than 10u of reagent. Which should go outside of the bursting proc ideally.

ZeWaka (approved):

PR #14910, opened 2023-07-11
Adds Mapo Tofu

zelda4040404:
[GAME OBJECTS][CATERING]
## About the PR 
Adds two new dishes for the chef (or anyone else) to make. One is made with meat and one without for those that prefer it.

both_tofu_V4

Each dish contains 25u of capsaicin due to the spicy nature of the dish.

## Why's this needed? 
More variety for spicy dishes + more use for soybeans in the kitchen beyond them being processed into milk.


frawhst (approved):
looks good (will wait for sprite review though)

Azrun:
My first read of the sprite is that of yogurt with veg on top.
Knowing what it is supposed to be it looks someone drizzled hot oil on some silken tofu.

I'd expect the tofu to be more colored like the sauce and either broken into pieces or textured so it seems like fluid like.  Wish I had a more simple suggestion given the size and medium in question. 😦 
🪙 🪙

zelda4040404:
Updated sprites and edited PR to show new ones. Tried to make it look more broken up, with a slightly adjusted color somewhere on the gradient between the original white color and the color at the edge of the bowl.

Azrun:
On the fence regarding all the buffs, but we can always something away.

PR #14916, opened 2023-07-12
Makes fluids and vapors unclickable, adapts most interactions to account for this

frawhst:
## TODO
- [x] Spectro goggles (going to be a bit painful)
- [ ] Whatever else i think of

## About the PR 
*\*Nothing in this PR which refers to 'gas' refers to atmos gas. This is all chems.\**

The Big Change: fluids can no longer be clicked, and neither can gases. They won't show up in the right-click menu either.

This also changes fishing rods, beakers, glasses, sponges, mops, the BloodTrak, etc so that instead of clicking on a fluid to make them work, you just click a turf, and if the turf has a fluid it'll relay the action over to the fluid. Notably, if a turf is covered in fluid, you can just click the fluid like normal and it'll be handled for you.

The reagent scanner has an edge case in that turfs can have both a fluid and a gas, both of which are relevant, and can ALSO have reagents inside itself. Thus, reagent scanners now display the readouts of the turf, the turf's gas, and the turf's liquid, if they exist.

While sometimes fluids don't create issues, this is in part because there's a fair amount of jank fixes in place already which expect people to click on fluids accidentally- for example, clicking a gas with a grab redirects you to pin the grabbed person to the floor, and a few random-ish items (rods, floor tiles, RCDs, and a couple more) completely ignore fluids because they expect that if you click a fluid with them, you're really trying to click the turf. This PR allows us to delete all those hacky workarounds.

## Why's this needed? 
Currently clicking on things while there's a lot of gas or very high liquid (most notable on Oshan, but also sometimes blood or tomato juice get there) is a pain, and you usually have to use the right click menu to be able to actually interact with anything. Now you'll be able to pick things up and smack them like usual even if there's a gas and a fluid in the way.

This also has a slight self-indulgent balance change, that you're no longer able to just eyeball what the most significant fluid in a puddle is. You'll just have to grab a reagent scanner or guess based off the color. Or taste, if you're bold. 

## Changelog 

```changelog
(u)aloe
(*)Fluids are no longer clickable. Anything you used to click fluids with now works based on clicking the turf below the fluid. This also means that you can no longer see the most dominant reagent in a fluid by the name of the fluid.
```


TheTyrant25:
so right clicking won't tell you what fluid is there? Seems like a good change

frawhst:
> so right clicking won't tell you what fluid is there?

Yep!

frawhst:
(Should actually work now. There was an issue with how testmerges and the secret repo interact).

frawhst:
im on vacation so won't do any work on this for the next weekish. Almost finished though, just need to make spectro goggles scan prettily (harder than it sounds) and mop (heh) up some assorted bugs

Issue #14921, opened 2023-07-12
Deluxe Mindhack gets destroyed when self-mindhacking


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Deluxe Mindhack gets destroyed when using it on yourself, it doesn't get implanted.

### Steps to reproduce

1. Accidentally deluxe mindhack yourself because you're not very robust.
2. Try to remove the deluxe and realise it's been destroyed.
3. Be sad you've wasted 6tc.

### Expected Behavior

Deluxe mindhack is removable if you mindhack other people, it should be the same when you accidentally mindhack yourself

### Additional Information & Screenshots

Reported by: ArlianDeBias
On server: Goonstation 1 Classic: Heisenbee
Active test merges: #14916
Round log date: 2023-07-12-14-48
Reported on: 2023-07-12 15:10:56
Map: COGMAP

PR #14922, opened 2023-07-12
Fixes mindhackers unable to extract their own implant from themselves

PrimeNumb:
## About the PR 
Mindhackers that mindhack themselves can now extract that implant

## Why's this needed? 
fixes https://github.com/goonstation/goonstation/issues/14921


flappybatpal:
TBH debatable if this needs "fixing".

deathrobotpunch:
this has no use as the mindhacker can say don’t remove mindhack

PR #14928, opened 2023-07-13
Adds Newspapers

TheTyrant25:
[FEATURE] [OBJ]
## About the PR
Adds newspapers that can be found lying around, or custom printed by players. They hide your face when you flick them open.
Features:
- You can roll up newspapers to hold them in one hand, or open them up to hold them two handed and hide your face. The opened up newspapers have an inhand sprite that visually blocks the front of your face and the face_visible gets set to false. Of course, you're still recognisable by your hair and clothes, but it's funny and blocks the AI's facial recognition algorithm.
- Newspaper headlines, contents and publishers are by default randomly generated. Currently done by picking from a list in `strings/newspapers.txt`, with the most focus on headlines at the moment. It can be expanded later to use the phrase log and possibly even use `pick_smart_string` to construct new random ones based on partial headlines.
- Newspapers can be printed with the printing press (if it has the newspaper upgrade, which it currently starts with). You can write custom headlines, publishers, and contents, containing whatever you like. Or leave it blank to have it be random. Unlike books, which cost 2 ink and 2 paper, newspapers only cost 1 ink.

## Image
image

## Why's this needed?
Excellent opportunities for roleplay. Sitting on a bench hiding behind a newspaper, selling them while wearing a flat cap... can you imagine? Also: a nefarious way for gangs to spread propaganda.

## Changelog
```changelog
(u)Tyrant, Flaborized
(+)Added newspapers! You can open them up to hide your face behind them, and read the headlines. Make some at the printing press!
```


TheTyrant25:
todo:
- [ ] See if the face covering works
- [ ] Get item sprites
- [ ] Get held sprites. Ideally those that would actually physically block the view of the player's face.
- [ ] Randomly generate headlines

zelda4040404:
Would it be at all possible for the newspapers to contain written content as well? I would love to be able to do an actual newspaper gimmick without having to rely on booklets when writing articles.

TheTyrant25:
Possible, yes. Could make them writable like paper is. I was just keeping them headline only to make them simpler, for the random generation aspect. Generating an article that relates to a random headline... is hard. Then again, could have it so that the article itself is optional when making a custom newspaper.

Glamurio:
Suggestion: Get the printing press involved. They've recently been spread on all stations and would work perfectly for this. Put written paper in press to copy text from, fill it up with empty paper, create newspaper that cannot be edited anymore, but can be mass produced and spread on the station. This opens a cool new journalism branch that the journalist could get involved in.

TheTyrant25:
Excellent idea.

TheTyrant25:
hopefully this fixes things

code/mob/living/carbon/human.dm
```diff
@@ -1176,6 +1176,13 @@
 		var/obj/item/clothing/hider = src.back
 		if (!hider.see_face)
 			. = FALSE
+	for (var/obj/item/hand in list(src.l_hand, src.r_hand))
+		if (istype(hand, /obj/item/paper/newspaper))
+			var/obj/item/paper/newspaper/np = hand
```
frawhst, line 1181: I don't think you need this cast, `two_handed` is defined on `item` which is what `hand` is typed to anyways

frawhst (commented):
Once you make it so people can actually write things in newspapers, would also appreciate phrase logging them (`phrase_log.log_phrase("category name", text)`) both for Rude Article Checking and also so we can toss random newspapers out of white holes and stuff

code/mob/living/carbon/human.dm
TheTyrant25: yeah im having a lot of bugs with the opening and closing. This part is purely meant to hide the face when two_handed is open. I don't know if it works or not at present.

TheTyrant25: i was playing it safe. Removed the typecast now.

TheTyrant25:
should be ready for review now i believe

code/obj/machinery/printing_press.dm
```diff
@@ -531,13 +629,46 @@
...
+		var/obj/item/paper/newspaper/rolled/NP = new
+		NP.desc = ""
+		if (src.newspaper_publisher)
+			NP.publisher = src.newspaper_publisher
+			NP.name = "[NP.publisher] newspaper"
+		if (src.newspaper_headline)
+			NP.headline = src.newspaper_headline
+		// it can auto generate headlines and publisher if left alone.
+		NP.desc = "Its from [NP.publisher]. Its headline reads: [NP.headline]"
+		NP.info = pick_smart_string("newspaper.txt", "article")
```
TobleroneSwordfish, line 661: So you can't actually set the contents? That seems weird

strings/newspaper.txt
```diff
@@ -0,0 +1,39 @@
...
+NT Space Station 17 voted 'worst station' for 9th year running.@,
+Calls for more railings around trench holes.@,
+Oshan tourism at all time highs.@,
+Highest peak on Mundus climbed for first time. Climber says 'hardest part was actually getting there'.@,
+Pug detective sniffs out crimes.
+
+article
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Mollis aliquam ut porttitor leo a diam sollicitudin tempor. Libero justo laoreet sit amet cursus. Nam at lectus urna duis convallis convallis tellus id interdum. Donec pretium vulputate sapien nec sagittis. Duis tristique sollicitudin nibh sit amet commodo. Ut sem nulla pharetra diam sit amet nisl suscipit adipiscing. Sed odio morbi quis commodo. Nulla aliquet porttitor lacus luctus accumsan tortor posuere ac ut. In egestas erat imperdiet sed euismod nisi porta lorem.@,
+Introduction. Todo: actually write this damn article.@,
+Once upon a time, there was a very nosy person. They opened a newspaper and decided to read the news. That person is you. Mind your own business.
```
TobleroneSwordfish, line 39: Feels like there should be more of these? Maybe use the CFG stuff to generate them rather than writing out each one.

TobleroneSwordfish (commented):
The unfurled item sprite is veeeerry large, maybe consider just using the in-hand one since that's already pretty big? Also this does make the printing press menu tree even worse, it could *really* do with having an actual dedicated menu.

code/obj/machinery/printing_press.dm
TheTyrant25: i figured having yet another field would be a hassle... but I can add it in.

strings/newspaper.txt
TheTyrant25: damn i completely forgot about these. Yeah I'll add some more. What's CFG?

TobleroneSwordfish: Context free grammar, basically the system used to build up new categories from previous ones like `[pronoun] [verb]ed a [noun]`

TheTyrant25:
okay random articles have been implemented

TobleroneSwordfish (changes requested):
Sprites are still way too big, items should not cover two tiles and clip into the wall next to them

TheTyrant25:
ah i have some new sprites which were made but haven't included them yet

TobleroneSwordfish (changes requested):
The face blocking doesn't seem to work?
image
also nitpick but it'd be neat if the headline was added to the content of the newspaper with header styling, since hiding it in the examine text is a bit obscure

TheTyrant25:
well, that's an issue.

TheTyrant25:
okay so i've fixed it and it now works as long as the person isn't wearing their ID. Which is how stuff like welding masks works too. very close to completion now. Just trying to fix one last bug where the article is always a handwritten one.

DisturbHerb:
everythimg is brokegn

PR #14938, opened 2023-07-13
Mobifies Lavacrabs

CheffieGithub:
## About the PR 
Changes object lavacrabs into a mob subtype of crab. 
As before, they do absolutely nothing on their own using an empty aiHolder. (maybe slightly cursed?) They will only move when provoked.

## Why's this needed? 
Object critter bad.


flappybatpal:
!merge_upstream

flappybatpal:
Technically they used to chase you after being attacked but I think the "reactive only" suits an otherwise inactive lava crab better,

PR #14945, opened 2023-07-14
fix lollipops always being white

TheTyrant25:
[MINOR]
## About the PR
Because the main reagent in lollipops is sugar (5u), it would always override the colour of the actual sweet (which is supposed to be randomly assigned, or based on the main reagent). This PR fixes that.
## Why's this needed?
They're supposed to be colourful damn it.
## Changelog
```changelog
(u)Tyrant
(+)Lollipops have colours once again.
```

### DIFF
--- a/code/modules/food_and_drink/candy.dm
+++ b/code/modules/food_and_drink/candy.dm
@@ -297,7 +297,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/candy/jellybean)
 	icon = 'icons/obj/foodNdrink/food_candy.dmi'
 	icon_state = "lpop-0"
 	sugar_content = 5
-	var/icon_random = FALSE // does it just choose from the existing random colors?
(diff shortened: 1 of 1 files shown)


TheTyrant25:
yes i like absolute pathing, so what
at least i didn't redo the whole file like i was tempted to.

code/modules/food_and_drink/candy.dm
```diff
@@ -297,41 +297,41 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/candy/jellybean)
 	icon = 'icons/obj/foodNdrink/food_candy.dmi'
 	icon_state = "lpop-0"
 	sugar_content = 5
-	var/icon_random = 0 // does it just choose from the existing random colors?
+	var/icon_random = FALSE // does it just choose from the existing random colors?
```
ZeWaka, line 300:
```suggestion
	var/icon_random = FALSE //! does it just choose from the existing random colors?
```

ZeWaka (approved):

PR #14956, opened 2023-07-15
Printing press cleanup refactor thing

TheTyrant25:
## About the PR
I was gonna mess with the printing press but noticed most of it was untouched since the 2020 epoch. Lots of it was missing doc comments, `TRUE FALSE` defines, and etc. So i cleaned that up a little. Also, did a bit of absolute pathing for some of the longer procs, for reducing the amount of indents reasons.

## Why's this needed?
Clean code is good.


TobleroneSwordfish (approved):
lgtm, git has exploded trying to comprehend indent changes so I'll trust you that the proc pathing stuff is right

PR #14959, opened 2023-07-15
fix chairs becoming steelgrey when unfolded

### LINKED ISSUE #14941: Steel Chairs Changing Color
### Labels

[BUG][MINOR]

### Description

When folding and unfolding a preset steel folding chair the sprite color changes from normal to a dark gray color.

### Steps to reproduce

1. Fold preset chair
2. Unfold chair
3. ???

### Expected Behavior

It shouldn't change color at all

Reported by: Thrillhaus
Active test merges: #14916
Round log date: 2023-07-14-04-53
Reported on: 2023-07-14 05:45:10
Map: NADIR

TheTyrant25:
[MINOR]
## About the PR
When a chair is unfolded, it checks whether it has a material, and if it does, it sets that material to the unfolded version. Sadly, this means that unfolding a steel chair always makes it darker, as it becomes a steel steel chair. If that makes sense.

This PR makes it so that if chairs are the same as the default materials, the material is not manually set. Testing seems to indicate that this has worked.

## Why's this needed?
Fixes  #14941

### DIFF
--- a/code/obj/stool.dm
+++ b/code/obj/stool.dm
@@ -553,6 +553,7 @@ TYPEINFO(/obj/stool/wooden)
 	scoot_sounds = list( 'sound/misc/chair/normal/scoot1.ogg', 'sound/misc/chair/normal/scoot2.ogg', 'sound/misc/chair/normal/scoot3.ogg', 'sound/misc/chair/normal/scoot4.ogg', 'sound/misc/chair/normal/scoot5.ogg' )
 	parts_type = null
 	material_amt = 0.1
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Probably makes more sense to use the "material appearances to ignore" flag? Since this is just a problem with steel.

TheTyrant25:
i uh, don't know how that works (nor that it existed). Guess I'll do that then

TheTyrant25:
tbh its `uses_material_appearance = FALSE` so im not sure why it would do it anyway

PR #14983, opened 2023-07-17
Fix various Mulebot things

TheTyrant25:
[MAPPING] [MINOR] [QOL]
## About the PR
- Unswaps the Donut2 Medbay and Hydroponics beacons.
- Makes the mulebot tooltip holder "set destination" use a dropdown menu like in the pda
- Makes it so that destinations are scanned automatically when you open the PDA app (you can still rescan them ofc)
- removed a load of empty lines that were there for no reason

## Why's this needed? 
Mulebots are somewhat neglected. What would wisemonster say? We should give them a bit more care.

## Changelog
```changelog
(u)Tyrant
(+)The mulebot's destinations are now scanned automatically when opening the menu and displayed as a list.
```


ZeWaka:
Please condense into a single changelog entry

TheTyrant25:
> Please condense into a single changelog entry

done

ZeWaka:
More like
> (+)The mulebot's destinations are now scanned automatically when opening the menu and displayed as a list.
:feelsgood:

PR #14989, opened 2023-07-17
Material mutability

amylizzle:
## About the PR 
Cleans up material code so that we're no longer making thousands of copies for no reason. 
Major changes:

- If you want a material instance that you can edit, you must call `getMutable()` on it or create a new one entirely. 
- Using `getMaterial()` will return an *immutable* instance of the material, which will throw a runtime if you try to edit it

## Why's this needed? 
Material datums have been stuck halfway between singletons and not for a while now, and it's messy and prone to errors. This should hopefully prevent further errors.

TODO: triggers


code/modules/materials/Mat_Materials.dm
```diff
@@ -257,13 +446,104 @@ ABSTRACT_TYPE(/datum/material)
...
+		handleTriggerGenerations(src.triggersTemp)
+		handleTriggerGenerations(src.triggersChem)
+		handleTriggerGenerations(src.triggersPickup)
+		handleTriggerGenerations(src.triggersDrop)
+		handleTriggerGenerations(src.triggersExp)
+		handleTriggerGenerations(src.triggersOnAdd)
+		handleTriggerGenerations(src.triggersOnLife)
+		handleTriggerGenerations(src.triggersOnAttack)
+		handleTriggerGenerations(src.triggersOnAttacked)
+		handleTriggerGenerations(src.triggersOnEntered)
```
TobleroneSwordfish, line 510: would also make this look less like an unrolled loop

code/modules/materials/Mat_Materials.dm
```diff
@@ -257,13 +446,104 @@ ABSTRACT_TYPE(/datum/material)
...
+ABSTRACT_TYPE(/datum/material/interpolated)
 /datum/material/interpolated
 	mat_id = "imcoderium"
 	name = "imcoderium"
 	desc = "You should not be seeing this"
 	color = "#6f00ff"
 
+	New(var/datum/material/mat1,var/datum/material/mat2,var/t)
+		..()
+		var/ot = 1 - t
```
TobleroneSwordfish, line 460: t? ot?

code/modules/materials/Mat_Materials.dm
```diff
@@ -14,67 +14,264 @@
...
 
-	var/owner_hasentered_added = FALSE
 
 	New()
 		. = ..()
 		for(var/datum/material_property/propPath as anything in concrete_typesof(/datum/material_property))
 			if(initial(propPath.default_value) > 0)
 				src.setProperty(initial(propPath.id), initial(propPath.default_value))
 
+	//getters for all the protected vars
```
TobleroneSwordfish, line 108: something something proc call overhead (probably fine?)

TobleroneSwordfish (commented):
This will probably break some things but in a good way that leads to better code 👍

code/modules/materials/Mat_Materials.dm
```diff
@@ -126,61 +331,45 @@ ABSTRACT_TYPE(/datum/material)
 				return 1
 		return 0
 
-	proc/addTrigger(var/list/L, var/datum/materialProc/D)
+	///Triggers is specified using one of the TRIGGER_ON_ defines
+	proc/addTrigger(var/triggerListName as text, var/datum/materialProc/D)
+		if(!src.mutable)
+			CRASH("Attempted to mutate an immutatble material!")
+		var/list/L = src.vars[triggerListName]
```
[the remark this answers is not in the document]
amylizzle: that is a bit cleaner, but also I'm using defines to keep that cleaner.

code/modules/materials/Mat_Materials.dm
amylizzle: no idea, copy pasted old code and just moved it around

code/modules/materials/Mat_Materials.dm
amylizzle: I made a change, now it looks less like an unrolled loop. I'm not sure it's better though.

code/modules/materials/Mat_Materials.dm
amylizzle: anything querying material properties fast enough for that to be a problem should probably be rewritten anyway.

TobleroneSwordfish:
note to self: this requires secret changes stupid!!!1

amylizzle:
I see there were a couple of bugs - was that just from needing secrets updating?

frawhst:
Why am i a code own-
image
carry on

code/mob/living/critter/flockcritter_parent.dm
```diff
@@ -62,7 +62,6 @@ TYPEINFO(/mob/living/critter/flock)
 /mob/living/critter/flock/New(var/atom/L, var/datum/flock/F=null)
 	..()
 	remove_lifeprocess(/datum/lifeprocess/radiation)
-	src.material.setProperty("reflective", 5)
```
frawhst, line 65: Does this get done somewhere else? because this seems like a (very minor) balance change if the stats are actually changing

code/mob/living/intangible/blob_overmind.dm
```diff
@@ -353,9 +354,10 @@
 		src.upgrade_id = 1
 		src.lipids = new()
 		src.nuclei = new()
-		src.my_material = copyMaterial(getMaterial("blob"))
-		src.my_material.color = "#ffffff"
-		src.initial_material = copyMaterial(getMaterial("blob"))
+		src.my_material = getMaterial("blob")
+		src.my_material = src.my_material.getMutable()
```
frawhst, line 358: Early in review, but this pattern seems to happen often enough that it might be worth a `getMutableMaterial` proc, for when you're starting from scratch and don't already have a material instance to work off of. Any reason not to do that?

code/modules/power/cable.dm
```diff
@@ -167,9 +167,6 @@
 		if(netnum && powernets && length(powernets) >= netnum) //NEED FOR CLEAN GC IN EXPLOSIONS
 			powernets[netnum].cables -= src
 
-	insulator.owner = null
-	conductor.owner = null
-
```
frawhst, line 172: Is this handled somewhere else? or just not needed now

frawhst (commented):
I'll get to the rest (including the actual important changes) sometime soon, just lots of files to look through

code/modules/power/cable.dm
frawhst: oh nevermind, I misunderstood what `insulator` and `conductor` were

code/mob/living/critter/flockcritter_parent.dm
amylizzle: yeah, this was discussed in #flockmind, and we essentially came to the conclusion "why do flock drones even have material". Leah was gonna remove it.

code/mob/living/intangible/blob_overmind.dm
amylizzle: mostly because I don't want people to copy paste the pattern unless they **really** need a mutable material. If `getMutableMaterial()` exists, people will just cargo-cult it, and then we'll have to try and catch it in review instead. Explicitly making a material mutable mollifies that somewhat, and at least requires an extra proc call to be intentionally added. I also considered `getMaterial(material, mutable=FALSE)`  and discarded it for the same reason.

frawhst: Seems reasonable. Would appreciate a comment on getMutable explaining why it is that way (if there already is something like that, just ignore)

TobleroneSwordfish (changes requested):
Okay no setting fire to the world just yet, looks like alloying doesn't work because `/datum/material/interpolated` is an abstract type?

amylizzle:
huh, I tested it and it worked. I'll check again.

amylizzle:
image
seems to work fine, produces interpolated material with no runtimes.

amylizzle:
meh, just removed the abstract, can't remember why I set it in the first place

amylizzle:
oooh that's why

code/modules/materials/Mat_Materials.dm
```diff
@@ -257,12 +451,85 @@ ABSTRACT_TYPE(/datum/material)
...
+			var/list/colA = normalize_color_to_matrix(mat1.color)
+			var/list/colB = normalize_color_to_matrix(mat2.color)
+			src.color = list()
+			for(var/i in 1 to length(colA))
+				src.color += colA[i] *left_bias+ colB[i] * bias
+		else
+			src.color = rgb(round(GetRedPart(mat1.color) *left_bias+ GetRedPart(mat2.color) * bias), round(GetGreenPart(mat1.color) *left_bias+ GetGreenPart(mat2.color) * bias), round(GetBluePart(mat1.color) *left_bias+ GetBluePart(mat2.color) * bias))
+		src.properties = mergeProperties(mat1.properties, mat2.properties, bias)
+
+		src.edible_exact = round(mat1.edible_exact *left_bias+ mat2.edible_exact * bias)
```
frawhst, line 488: Won't this just always equal `0` or `1` since 1-argument `round` just floors the number?
Should probably look something like
```
src.edible_exact = [calculation]
src.edible = round(src.edible_exact, 1)
```
(yes, `round(n)` and `round(n, 1)` behave differently for some reason. why)

code/modules/materials/Mat_Materials.dm
```diff
@@ -14,67 +14,269 @@
...
+		if(src == M2) //since we're actually doing mutable/immutable now, we can frequently shortcut this with an actual equal check
+			return TRUE
+
+		if(isnull(M2))
+			return FALSE
+
+		if(src.properties.len != M2.properties.len || src.getID() != M2.getID())
+			return FALSE
+
+		if(src.value != M2.value || src.name != M2.name  || src.color ~! M2.color ||src.alpha != M2.alpha || src.getMaterialFlags() != M2.getMaterialFlags() || src.texture != M2.texture)
```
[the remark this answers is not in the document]
frawhst: OH completely blanked on the fact that `color` can be a matrix

code/modules/materials/Mat_Materials.dm
amylizzle: I'm not entirely sure why the round is there anyway. I wonder if this has just not been working since alloying was put in.

amylizzle, line 488:
```suggestion
		src.edible_exact = mat1.edible_exact * left_bias + mat2.edible_exact * bias
```

frawhst (approved):
My stuff is all addressed

PR #14991, opened 2023-07-17
ABCU Marker usability improvements

kbsmilk:
## About the PR 
ABCU marker now uses pixelaction instead of afterattack, and its range was uncapped. It now makes 2 different noises when selecting/unselecting a tile.

## Why's this needed? 
Marker is less annoying to use now. Previously, it couldn't target wall corners from inside

## Changelog 

```changelog
(u)Smilg
(+)Increased ABCU marker selection range
```

### DIFF
--- a/code/WorkInProgress/blueprints.dm
+++ b/code/WorkInProgress/blueprints.dm
@@ -436,8 +436,8 @@
 
 	var/static/savefile/save = new/savefile("data/blueprints.dat")
 
-	afterattack(atom/target as mob|obj|turf, mob/user as mob)
-		if(GET_DIST(src,target) > 2) return
+	pixelaction(atom/target as mob|obj|turf, params, mob/user as mob)
+	//	if(GET_DIST(src,target) > 2) return
 
 		if(!isturf(target)) target = get_turf(target)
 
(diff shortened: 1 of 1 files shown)


code/WorkInProgress/blueprints.dm
```diff
@@ -436,8 +436,8 @@
 
 	var/static/savefile/save = new/savefile("data/blueprints.dat")
 
-	afterattack(atom/target as mob|obj|turf, mob/user as mob)
-		if(GET_DIST(src,target) > 2) return
+	pixelaction(atom/target as mob|obj|turf, params, mob/user as mob)
+	//	if(GET_DIST(src,target) > 2) return
 
 		if(!isturf(target)) target = get_turf(target)
```
Mister-Moriarty, line 442:
```suggestion
	pixelaction(atom/target, params, mob/user)
		if(!isturf(target)) target = get_turf(target)
```
Remove the commented out code; Git history will remember.
I don't believe that there is any need for `as mob|obj|turf` here.

ZeWaka (approved):

code/WorkInProgress/blueprints.dm
```diff
@@ -436,9 +436,7 @@
 
 	var/static/savefile/save = new/savefile("data/blueprints.dat")
 
-	afterattack(atom/target as mob|obj|turf, mob/user as mob)
-		if(GET_DIST(src,target) > 2) return
```
TheTyrant25, line 440: seems you removed the distance check, is that intended?

TheTyrant25 (approved):
looks good.

code/WorkInProgress/blueprints.dm
Tarmunora: should probably retain *some* distance check (7 or 10), to lock it to stuff on-screen (camera shenanigans etc.)

kbsmilk: I just copied Floor and Wall Designer's design (obj/item/room_planner), which also has no distance check, and seemingly no problems

Tarmunora: those should be limited to on-screen too 😱

PR #15014, opened 2023-07-19
No dumping reagents in sink from closed containers

flappybatpal:
## About the PR 
Adds an open container check to the chef_sink when cleaning items to avoid dumping reagents from food etc.

## Why's this needed? 
You could wash all the reagents out of a MONSTER burger and other food items.


DisturbHerb:
sœul

ZeWaka (approved):

frawhst:
I feel like we should have some way of doing this. Imo syringes should be able to remove/insert reagents from/into food regardless of open-ness, but that's out of scope.

Azrun (approved):

PR #15025, opened 2023-07-20
Refactors butcherable

CheffieGithub:
## About the PR 
Its not a boolean but its used as one, now it has defines:
`BUTCHER_NOT_ALLOWED`
`BUTCHER_ALLOWED`
`BUTCHER_YOU_MONSTER`

Adjusted some butcherable values where it makes sense:
- Can't butcher robotic
- `BUTCHER_YOU_MONSTER` for seals and walruses
- Allows butchering of more small animals

Fixes some stuff in the owlery file while I was there it was annoying.

## Why's this needed? 
Clarity of what butcherable is.


DimWhat:
I suggest adding bees to the "BUTCHER_YOU_MONSTER" section

code/WorkInProgress/Hydrothings.dm
```diff
@@ -1,12 +1,8 @@
 //Content related to Operation Fuckable Owls
```
TheTyrant25, line 1: seems like most of these refactors here are unrelated to butchering

TheTyrant25 (commented):
Seems mostly good except for these owlrey changes which seem to have bled over from another project. The actual find replace of butcherable looks to have worked fine

code/WorkInProgress/Hydrothings.dm
CheffieGithub: `Fixes some stuff in the owlery file while I was there it was annoying.` Its just removing whitespace and adding missing slashes

DisturbHerb: i would atomize those changes out regardless

CheffieGithub: seems like too small of a change to me to warrant its own PR

DisturbHerb: no change is ever too small :beesleep:

frawhst (approved):
lgtm. Whitespace stuff is ok even if it produces a fat diff.

CheffieGithub:
will fix the conflict once I have some working cleared

PR #15033, opened 2023-07-21
Minor weapon rack UI fix

Nitre88:
## About the PR 
Makes the ui for weapon stands update when hacked.  Now able to use hacking tools to open the ui, if the wires are open. Doesn't change the actual ui at all. _Reading bug reports this is more of a temporary fix due to someone working on tgui for all of the outdated ui's._
My first pr, I probably did stuff wrong

## Why's this needed? 
Makes the interface less clunky and brings weapon racks more into shape with the rest of the hacking in the game.


TnTmage:
I feel like my code is terrible...

code/obj/machinery/weapon_racks.dm
```diff
@@ -223,6 +227,10 @@
 		if (src.malfunction)
 			user.shock(src, 7500, user.hand == LEFT_HAND ? "l_arm" : "r_arm", 1, 0)
```
ZeWaka, line 229: don't reference line numbers, they WILL change

ZeWaka (changes requested):

code/obj/machinery/weapon_racks.dm
TnTmage: thanks

PR #15054, opened 2023-07-23
Light Space Suit Resprite

Flaborized:
## About the PR 

Resprites light engineering space suits. Also adds monkey variations. Both items and worn states changed as pictured:

## Why's this needed? 

Current version is a recolor of the Mars suits, which look pretty dated and generally I think this is an improvement which brings them closer to the current on-station set of suits while still trying to look unique and 'lightweight'. I think it looks better. Also good to have these monkey versions. Made to conform to the wearer's body more closely than regular suits since they're lightweight and maneuverable and stuff. They're like a silly space pajama suit.

## Changelog 

```changelog
(u)Flaborized
(+)Resprited light engineering spacesuits.
```


Flaborized:
added tails and a bigger helmet for lizards so their snout fits in. ty to dimwhat for reminding me suit tails exist

BatElite:
I liked the lighter grey personally, have you tried to see what that looks like?

Gannets (approved):
looks good to me

PR #15055, opened 2023-07-23
Do the thing where I clean up var/level

Laboredih123:
[INTERNAL][CLEAN]
## About the PR 
DMDocs var/level and macros its values.

## Why's this needed? 
I like to know what the numbers do.

### DIFF
--- a/code/modules/mechanics/MechanicMadness.dm
+++ b/code/modules/mechanics/MechanicMadness.dm
@@ -311,7 +311,7 @@
 		if (!istype(src.stored?.linked_item,/obj/item/storage/mechanics/housing_handheld))
 			qdel(src) //if outside the gun, delet
 			return
-		if(level == 1)
+		if(level == UNDERFLOOR)
 			src.icon_state=icon_down
 			SPAWN(1 SECOND)
 				src.UpdateIcon()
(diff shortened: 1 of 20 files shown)


code/modules/mechanics/MechanicMadness.dm
```diff
@@ -482,11 +482,11 @@
 		return ..()
 
 	pick_up_by(var/mob/M)
-		if(level != 1) return ..()
+		if(level == UNDERFLOOR) return ..()
```
TobleroneSwordfish, line 485: this is backwards I think

code/obj.dm
```diff
@@ -356,6 +356,7 @@
 /obj/proc/alter_health()
 	return 1
 
+/// Whether or not to hide something based on the value of h.
```
TobleroneSwordfish, line 359: h

TobleroneSwordfish (changes requested):

BatElite:
The connector port has a level of 0 and it seems to be the only thing that does so. Might as well change that to overfloor.

(god, imagine underfloor connector ports. evil.)

code/modules/mechanics/MechanicMadness.dm
Laboredih123: feck

code/obj.dm
Laboredih123: h

PR #15057, opened 2023-07-23
Create atom/item specific procs to handle most in-combat material procs and make most of these material procs compatible with atoms

### LINKED ISSUE #12093: Flock walls don't reflect projectiles
### Labels

[BUG]

### Description

Flock converted walls don't reflect energy and laser projectiles

### Steps to reproduce

1. Fire taser or laser gun at Flock converted wall

### Expected Behavior

For the projectiles to be reflected

First reported by someone else

M-Earthfire:
[Material][Quality][Refactor]

## About the PR 

This PR moves most direct calls of material procs into item- or atom specific procs. To accomodate it, most material procs that are affected by this are refactored to be passed the broadest class possible, which is for many of these procs atoms.

This fixes #12093 

Thus, multiple procs are introduced,

``atom/proc/material_on_attack_use`` is used when the material is used for an attack

``atom/proc/material_trigger_on_mob_attacked`` is called on every atom in a mob when that mob that contains something out of the material is attacked. "situation modifier" gets passed on this proc the attack zone that the attack tries to hit, if there was any. The respectable material proc was only called on items in the mob, e.g. implants, now it is called on equipped items as well. With some caveat.

``atom/proc/material_trigger_on_bullet`` has the same behaviour, just for bullet hits.

``atom/proc/material_trigger_when_attacked`` is called when an item out of the material is affected. This now gets passed mobs as attacking atoms for cases like wall shamshes or laywer hands on tables

``/atom/proc/material_trigger_on_temp``, ``atom/proc/material_trigger_on_explosion``, ``atom/proc/material_trigger_on_blob_attacked``, ``atom/proc/material_trigger_on_chems``, ``/obj/item/proc/material_on_drop``, ``/obj/item/proc/material_on_pickup`` were added for their respectable triggers.

This two material procs that are not routed to atom-calls are ``triggersOnLife`` and ``triggersOnEntered``. These require some untangling and more severe modification, to some degree in their design, because of which i decided to move these into their own PR's

To show off these what these procs enable, the behaviour on how equipped items on mobs call their material effects when the mob is hit was added.

Before, equipped items didn't reacted at all if you attacked someone. Now, they get affected, based on hit location and object size. If you target a location that is covered by a clothing, that clothing's material effect gets triggered.

For items that are not clothing, they are only affected guaranteed when you are blocking with them. Else, there is a chance based of the item size to hit it.  

In general, that means someone with an erebite PDA might find it exploding with a small chance while they are wearing it.

Beyond that, multiple occassions were fixed where the wrong proc was called, e.g. when destroying plated handcuffs

## Why's this needed? 

In general, routing material procs of the respectable atoms enables to more finetune when and how the effects of materials are called. This is especially important since matsci items (spacesuits, spears, arrows) are very janky to make their material effects work properly. This PR will very easy enable to e.g. make the shaft affect the on-entered or on-life effect and the tip of the spear to affect the on-attack-effect.

Also, the cleanup has some interesting side effects. Like reflective walls being able to reflect bullets again. This will also apply to mobs made out of the material now (e.g. syreline-ore-chickens).

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Material effects on-attacks are now happening more consistent, e.g. erebite hats weared on you will explode if you get bonked on the head.
```


code/modules/materials/Mat_ProcsDefines.dm
```diff
@@ -239,6 +239,72 @@ var/global/list/material_cache = buildMaterialCache()
...
 	return new /datum/material/interpolated(mat1, mat2, 0.5)
 
+//custom matsci event procs
+//Use these if you want the stom in general to interact in a special way with the items procs e.g. spears on attack triggering the tip, but on pickup the shafts material
+//situation_modifier is for when you want something like specifying "chest" or "L_hand" for clothes
+
+/// Called when a mob holding this atom is attacked for mat effects
+/atom/proc/material_trigger_on_mob_attacked(var/mob/attacker, var/mob/attacked, var/atom/weapon, var/situation_modifier)
+	if (src.material)
+		src.material?.triggerOnAttacked(src, attacker, attacked, weapon)
```
ZeWaka, line 249: this is redundant, what do you think `src.material?.` does

ZeWaka (changes requested):

ZeWaka (approved):
ideally this would be done using signals but this is fine for now i think

PR #15061, opened 2023-07-23
Adds 2 monkey glasses.

DimWhat:
[clothing]
## About the PR 
Adds a bespoke sprite for nightvision sec huds and flashy goggles for monkeys, very minor differences but it looks a little nicer.
Also removes the moustache sprite from the file and puts it in the mask file

## Why's this needed? 
Makes the game look slightly better.


frawhst (approved):
I thought it was called `nightvisionchud` at first

PR #15062, opened 2023-07-23
Moves monkey moustache sprite to the correct file.

DimWhat:
[clothing]
## About the PR 
Readds the monkey moustache sprite to the mask file. Apparently this was left in the glasses folder during the big monkey sprite split. I moved it over to the mask file where it belongs.

## Why's this needed? 
It was an error for it not to be here in the first place.


ZeWaka (commented):
if it's being moved then why is it not deleted from a file

DimWhat:
> if it's being moved then why is it not deleted from a file

Yeah fair point, merged this PR with #15061

PR #15065, opened 2023-07-23
Ethanol strength nerf for non-evil alcoholic drinks

MintyPhresh:
[Chemistry] [Catering] [Balance] [Bug] [QoL] [add-to-wiki]
## About the PR 
Alcoholic drinks have a variable called "alch_strength" that determines how much ethanol is put into the drinker's bloodstream per tick. These values are often based on real-life alcohol by volume (ABV) values: drinks like beer, wine, and bourbon have alch_strength values that closely follow conventional ABVs.

However, because alch_strength does not account for the depletion rate of the drink (often 0.4), these values do not represent ABV. Bourbon, with an ABV of 0.45, adds 0.45 units of ethanol every time 0.40 units of it decay; Hilariously, this makes it _stronger than pure ethanol,_ and easier to get into a drink too. More extreme cases of easy-to-mix drinks with high ABVs include Tequila, with an ABV of 150%, and Rum and Coke with an ABV of 175%. Rather than change the alch_strength of all 112 drinks in the game to account for this, this PR multiplies the depletion rate of the reagent to the amount of ethanol added per life tick. Reagents that rely on their alch_strength to function as poison have had their alch_strength increased to avoid nerfing them.

Gin Fizz also got its recipe changed to finally use tonic water instead of using normal water and relying on magic, and has a more realistic ABV because it was looking at me funny. Gin Fizz's description and mix phrase have had minor changes so that the drink doesn't start to bubble "on it's own," but the observer is still seemingly incredulous to the mechanics of the eponymous fizz. The spaceman fears the Gin Fizz, for he does not understand it.

## Why's this needed? 

Low-effort mixed drinks having a higher potency than pure ethanol is unintuitive for new players and discourages people from having drinks at the bar, for fear of taking in debilitating levels of alcohol. Rough for RP, funny for classic. Imagine a world where a drinking glass with 30u of Rum and Coke and 20u of ice only gives you 17.7u of ethanol at its peak rather than putting you halfway to the 100 unit overdose threshhold. That could be _this_ world!

This change doesn't block off the "hilariously drunk" aspect from Classic, either: drinks that would be used as poisons are actually at slightly higher ABVs now than they were before the change, and there's still plenty of options for drinks at the power level they used to be (such as three mile island iced tea, Squeeze, and Boorbon). 

```changelog
(u)Mintyphresh
(*)Greatly reduced the ABV of most non-evil drinks. See minor changes for exceptions. Drink up!
(+)Necroni, Grog, Bo Jack Daniels, Squeeze, Suicider, Murdini, Moonshine, BOOrbon, Bloody Scary, and Ling Island Iced Tea have increased in strength.
```

### DIFF
--- a/code/modules/chemistry/Reagents-FoodDrink.dm
+++ b/code/modules/chemistry/Reagents-FoodDrink.dm
@@ -206,7 +206,7 @@ datum
 
 			on_mob_life(var/mob/M, var/mult = 1)
 				if(!M) M = holder.my_atom
-				M.reagents.add_reagent("ethanol", alch_strength * mult)
+				M.reagents.add_reagent("ethanol", alch_strength * mult * depletion_rate) //Multiplying by depletion rate makes alch_strength describe ABV, with 1 being 100% ABV
 				M.reagents.remove_reagent(src, 1 * mult)
 				..()
 				return
@@ -564,7 +564,7 @@ datum
 			fluid_g = 163
 			fluid_b = 30
 			transparency = 190
-			alch_strength = 0.1 //half vodka by content, half vodka strength
+			alch_strength = 0.25 //half vodka by content, half vodka strength
 
 		fooddrink/alcoholic/cocktail_bloodymary
 			name = "Bloody Mary"
@@ -576,7 +576,7 @@ datum
 			fluid_g = 53
 			fluid_b = 0
 			transparency = 190
-			alch_strength = 0.1
+			alch_strength = 0.25
 
 		fooddrink/alcoholic/cocktail_bloodyscary
 			name = "Bloody Scary"
@@ -980,7 +980,7 @@ datum
 			fluid_r = 248
 			fluid_g = 255
 			fluid_b = 206
-			alch_strength = 0.25
+			alch_strength = 0.15
 			description = "Don't question how it's fizzing without seltzer."
 			reagent_state = LIQUID
 			taste = "fizzy"
--- a/code/modules/chemistry/Reagents-FoodDrink.dm
+++ b/code/modules/chemistry/Reagents-FoodDrink.dm
@@ -209,7 +209,6 @@ datum
 				M.reagents.add_reagent("ethanol", alch_strength * mult * depletion_rate)
 				//Multiplying by depletion rate makes alch_strength describe ABV, with 1 being 100% ABV
 				//This means that drinks ~15% ABV need a higher depletion rate so that the ethanol can accumulate
-				M.reagents.remove_reagent(src, 1 * mult)
 				..()
 				return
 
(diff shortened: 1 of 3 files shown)


MintyPhresh:
I had to push a quick change increasing the depletion rate of chems with ABV <15% so that the alcohol could still accumulate in your body. Should be here shortly..

code/modules/chemistry/Reagents-FoodDrink.dm
```diff
@@ -206,7 +206,9 @@ datum
 
 			on_mob_life(var/mob/M, var/mult = 1)
 				if(!M) M = holder.my_atom
-				M.reagents.add_reagent("ethanol", alch_strength * mult)
+				M.reagents.add_reagent("ethanol", alch_strength * mult * depletion_rate)
+				//Multiplying by depletion rate makes alch_strength describe ABV, with 1 being 100% ABV
+				//This means that drinks ~15% ABV need a higher depletion rate so that the ethanol can accumulate
 				M.reagents.remove_reagent(src, 1 * mult)
```
TobleroneSwordfish, line 212: So it looks like this line was an attempt to solve something like this problem in the past, except it doesn't work because `remove_reagent` takes a reagent ID and not the reagent itself.
Since you've now solved it in a better way, you should probably remove this line!

TobleroneSwordfish (changes requested):
Generally looks good, is there a reason you're setting the depletion rate of a bunch of drinks to 0.7?

code/modules/chemistry/Reagents-FoodDrink.dm
MintyPhresh: Yes, there is a reason for the depletion rate change! With the new alcohol strength change making everything 2.5x weaker, many of the drinks now add ethanol at a slower rate than Spacemen bodies purge it. To make it possible to get drunk off of these drinks, I upped the depletion rate.

MintyPhresh:
Yes, there is a reason for the depletion rate change! With the new alcohol strength change making everything 2.5x weaker, many of the drinks now add ethanol at a slower rate than Spacemen bodies purge it. To make it possible to get drunk off of these drinks, I upped the depletion rate.

MintyPhresh:
To clarify, I added a depletion rate of 0.7 for drinks below .16 alch_strength, and a depletion rate of 1 for drinks below 0.9 alch_strength. If the drinks were to stay at 0.4 depletion rate, drinks with a strength below .13 are impossible to get drunk from without slow metabolism; for 0.7 depletion rate, drinks below .08 strength are impossible to get drunk on without slow metabolism. My choice of depletion rates was mostly arbitrary, though.

code/modules/chemistry/Reagents-FoodDrink.dm
MintyPhresh: the line's been deleted, thanks for the tip!

Tarmunora (approved):

PR #14808, opened 2023-07-01
Adds pre-filled chemlab counter subtypes

MintyPhresh:
[Chemistry] [Mapping] [Science] [QoL]

## About the PR 

This PR is a set of subtypes for the chemlab counters that come filled with useful chemlab equipment. There's 7 new labcounter subtypes in this PR:

**Basic Supply**
_Everything an aspiring chemist needs to start making chemicals!_
Pharmacopia
Beaker Box x2
Large Beaker x2
Spectroscopic Goggles
Reagent Scanner
Mechanical Dropper
Dropper x2

**Auxiliary Supply**
_Extra supplies for the discerning chemist._
Beaker Box
Patch Box
Syringe Box
400u Anti-Static Fuel Tank
Large Beaker
Spectroscopic Goggles x2
Reagent Scanner x2
Mechanical Dropper
Dropper x2

**Clerical Supply**
_It's only science if you write it down! Or blow yourself up._
Paper Bin x2
Hand Labeler
Pen
Stamp
Audio Log
Audio Tape x2

**Toxin Care**
_These drawers have been labeled EMERGENCY TOXIN CARE, which means they're probably already close to empty._
Toxin First Aid Kit
Calomel Autoinjector

**Chemical Storage** (Intended to replace the locker in Atlas)
_A set of basic precursor chemicals to expedite order fulfillment, increase efficiency, and synergize your workflow. Whatever the fuck that means._
50u Phenol bottle
50u Sulfuric Acid bottle
50u Acetone bottle
50u Oil bottle
50u Diethylamine bottle
50u Ammonia bottle

**Seedy-Looking**
_A stash of drugs provided by the RD in an attempt to placate their underlings._ 
Cannabis Leaf x2
Zippo Lighter
Syringe (Krokodil)
Syringe (Morphine)
Pill Bottle (???)

**Jam-Packed** (Intended for use in Atlas)
_The drawers on these barely close, and rattle loudly when moved. Guess they tried to put too much crap in it._
Beaker Box x2
Syringe Box
Paper Bin
Hand Labeler
Mechanical Dropper
Dropper
Toxin First Aid Kit
Spectroscopic Goggles
Reagent Scanner

In addition, there's also a Paper subtype added that tells unfamiliar players how to access these troves of equipment:
**Stern Lab Safety Warning**

_I've had it with you nincompoops taking shortcuts. For the last
	time, **when you open the drawers under the lab counter,
	USE AN EMPTY HAND!** There's no excuse for you to be
[...]

PR #15067, opened 2023-07-23
Adds Welding Fuel cans to Chemlab secure lockers

MintyPhresh:
[QoL] [Chemistry]

## About the PR 
This PR adds a 400u anti-static welding fuel container to the secure locker in Chemistry, as well as to its (currently unused) counterpart lab table subtype.

## Why's this needed? 
Welding Fuel is an essential precursor for the creation of tons of reagents in the game, including chems that people order often like graphene hardening compound and pentetic acid. New players looking to learn Chemlab may struggle to find easy access to a bottle of the stuff despite it being such an important part of gameplay (Donut2 only has a large tank, Oshan has two anti-statics, Kondaru's is hidden in the science podbay, etc, etc.) Instead of forcing new players to learn how to scavenger hunt on each map, they get easy access to the reagents they need to do their job.


CheffieGithub:
Theres already typically a crate with 2 and bottles in most chemlabs, it would make sense to standardise that as well additionally or instead to normalise the amount of fuel tanks.

MintyPhresh:
That's fair but it would mean reworking every map. Which I plan to do when I implement the chemlab table subtypes that have equipment in them #14808

MintyPhresh:
I don't understand...!!!! Why is this size M now..??? WHY IS IT MERGE CONFLICTING THIS IS SOMETHING THAT'S SUPPOSED TO CHANGE!!!!

PR #15082, opened 2023-07-25
Adds Censorium to replace char+rock

MRLinew:
## About the PR 

This PR adds a material called censorium, and its recipe is char + rock. Char + rock previously made a not so community friendly name, and this is the workaround Tarm suggested in the discord.

## Why's this needed? 

Suggested fix for char + rocks old name in the nanocrucible.

## Changelog 

```changelog
(u)Froggit Dogget
(+)Adds the censorium alloy. It does nothing.
```


MRLinew:
yeah my workspace is all messed up. I don't know how to fix it yet.

TheTyrant25 (commented):
Seems like a fun way to remove alloys named Cock. Unfortunately, coral rock and cobryl rock probably also generate the same name, or very similar names.
Maybe make it so rock alloys have a naming different system (charrock or charstone maybe) or implement an explicit language check in the autogenerated name maker thing.

code/modules/materials/Mat_Materials.dm
```diff
@@ -1440,6 +1440,18 @@ ABSTRACT_TYPE(/datum/material/fabric)
 		setProperty("thermal", 9)
 		setProperty("electrical", 7)
 
+/datum/material/censorium
```
Tarmunora, line 1443: should probably be a child of one of the abstract parents like `/datum/material/metal` or /datum/material/organic`

MRLinew: got it

Vaskritaya:
Feel like it'd make sense as a very dark colored material. Don't have to worry about wacky effects getting run into the ground due to how easy it is to make, but would still have a niche use for aesthetic purposes and is thematic with the name.

CaioFrancisco:
I'd make it a transparent material, with a blurred effect. That would allow censodium to have an use beyond (haha byond) being a simple joke, that is, being able to be infused into objects to make funny objects.

Tarmunora:
thermal prop is a bit extreme and might get pushed up a little but that's for future me

MRLinew:
Yeah i tried averaging the stats and i dont know what average is for thermal so i took half of chars base thermal.

PR #15087, opened 2023-07-25
phrase log Books from the Printing Press

TheTyrant25:
[INTERNAL]
## About the PR
Adds phrase logging to books custom made in the printing press. Tracks authors, titles, and contents. No duplicates.

## Why's this needed?
It's good to log custom text so that we can make random versions later.


Tarmunora (approved):
oh god this will spam us so hard with false positives

PR #15090, opened 2023-07-25
Mentormouse ability incapacitation fix

Bartimeus973:
## About the PR 
Mentormouse abilities can now be cast while stunned, and while inside a container or not on a turf.

Also minorly changes mentormouse disappear ability so that the mouse visually disappears during it.

## Why's this needed? 
As funny as it is that you can kidnap a mentormouse in a pet carrier, not being able to escape is a problem. Also being able to stunlock a mentormouse and preventing it from disappearing without a suicide should probably be fixed too.

### DIFF
--- a/code/datums/abilities/critter.dm
+++ b/code/datums/abilities/critter.dm
@@ -4,7 +4,8 @@
 			return
 		if (!owner.holder)
 			return
-		if (!isturf(usr.loc))
+		var/datum/targetable/critter/ability = owner
+		if (!isturf(usr.loc) && ability.needs_turf)
 			return
 		..()
 
@@ -28,6 +29,7 @@
 	disabled = 0
 	var/toggled = 0
 	var/is_on = 0   // used if a toggle ability
+	var/needs_turf = TRUE	//set to false if it can be cast while not on a turf. E.g. inside a locker
 	preferred_holder_type = /datum/abilityHolder/critter
 
 	New()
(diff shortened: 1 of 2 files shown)


Bartimeus973:
I was told by Jeb they werent able to leave but perhaps they didnt try enough.

DisturbHerb:
> I was told by Jeb they werent able to leave but perhaps they didnt try enough.

it does take five attempts of flipping/resisting

code/mob/living/critter/small_animal.dm
```diff
@@ -3903,7 +3903,8 @@ var/list/mob_bird_species = list("smallowl" = /mob/living/critter/small_animal/b
 	name = "Vanish"
 	desc = "Leave your body and return to ghost form"
 	icon_state = "mentordisappear"
-
+	needs_turf = FALSE //always castable
+	var/disappearance_time = 0.5 SECONDS
```
ZeWaka, line 3907:
```suggestion
	var/const/disappearance_time = 0.5 SECONDS
```

ZeWaka (approved):

code/datums/abilities/critter.dm
```diff
@@ -28,6 +29,7 @@
 	disabled = 0
 	var/toggled = 0
 	var/is_on = 0   // used if a toggle ability
+	var/needs_turf = TRUE	//set to false if it can be cast while not on a turf. E.g. inside a locker
```
ZeWaka, line 32:
```suggestion
	var/needs_turf = TRUE	//! set to false if it can be cast while not on a turf. E.g. inside a locker
```

PR #15096, opened 2023-07-25
A Better Observer Menu

amylizzle:
## About the PR 

image

## Why's this needed? 
Current one is lammeee

## Changelog 

```changelog
(u)Amylizzle
(*)The observer menu has been replaced with an upgrade. 
```


CheffieGithub:
Is numbering for duplicates possible? Like the spiders in the provided screenshot

amylizzle:
Oh dammit, I knew I forgot something. The information's there, I just forgot to display it!

amylizzle:
Done, I also added the observer count since I forgot it

CaioFrancisco:
My biggest worry about the new UI is the fact you can see who's an antagonist as a DNR ghost, which may make meta-comms incredibly easy.

amylizzle:
TBF meta comming is already bannable, and even without DNR it's totally possible to figure out who the antags are as a ghost.

blackrep:
We've had this discussion in Discord, but also just posting here for posterity purposes, I am not a huge fan of "wall of buttons". I feel like it's much better to use a List format instead, and leverage Tabs to reduce the length of your UI. Something like this mock-up I sent in the Discord: 
image

CheffieGithub:
cross posting also
Idea from Ormakent `I suggested it last time, but what if there is a separation of the list by job like in manifest, that way even if it's buttons and not a list there are many categories so it's not a big clump`

My idea in the same sort of design is icons for jobs of players but that seems hard to get in the UI nicely.

And a question, do mobs in cryo stop being observable?

amylizzle:
Bugs:
-Mimics don't show up at all
-AI eye appears as "Ai name #2" 

Features to do:
- Observer random player button
- Start NPCs and Objects as collapsed, storing as local state so it's preserved

amylizzle:
mimics whyyyy 
image

flappybatpal:
Breaking a bunch recently for all players but with no clear cause
**TypeError: Unable to get property 'toLowerCase' of undefined or null reference**

amylizzle:
> Breaking a bunch recently for all players but with no clear cause **TypeError: Unable to get property 'toLowerCase' of undefined or null reference**

Should be fixed now, plz testmerge update

TobleroneSwordfish:
One minor issue seems to be that players are only searchable by their alias, so if someone puts on a mask they will show up as "Unknown" rather than their name.

PR #15100, opened 2023-07-26
Fix Doubled Turfs in in rotation maps and Z files

CheffieGithub:
## About the PR 
Fixes every map key with a doubled turf of rotation maps and of z2, z3 and z5

atomised from #14686

Creates diff on lavamoon due to the lava under catwalks getting removed, will PR a fix for those catwalks after.

## Why's this needed? 
Turfs are meant to be alone :(. Update paths keeps screaming.


ZeWaka (changes requested):
fix lavamoon in this pr?

ZeWaka:
image
 ?

CheffieGithub:
looks correct in game, issue with them not having a map icon which they haven't for ages. Ill fix that.
image

PR #15118, opened 2023-07-26
Atlas Lab Drawer Conversion

MintyPhresh:
[Chemistry] [Mapping] [Science] [QoL]
## About the PR
This PR replaces the empty drawer in Atlas' chemlab with a filled version that now contain its chemistry equipment, using the counters added in #14808. Most handheld equipment lying on the counter has been removed, and a note detailing how to open drawers has been left in their stead. Additionally, this PR replaces the chemical storage locker in Atlas with a filled chemlab counter containing the same contents, as well as putting equipment previously in a crate outside chemistry into the chemlab counter that the glass recycler rests on top of.

## Why's this needed?
(Copied from #14808)
These components are part of a larger goal to eliminate lab counter clutter in chemlabs while simultaneously standardizing the equipment that a player could expect to find in a typical chemlab. Switching to this method will provide a more consistent chemlab experience between maps; it also means that if these loadouts need to be changed in the future, they can be changed within their subtypes rather than having to change each map individually.

Additionally, while Atlas specifically may not have that much counter clutter, the addition of the extra table more than doubles usable counter space and edges the lab into being just a little more bearable to dual-chemist.

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -4785,13 +4785,14 @@
 /turf/simulated/wall/auto/reinforced/supernorn,
 /area/station/engine/combustion_chamber)
 "aqh" = (
-/obj/storage/secure/closet/research/chemical,
 /obj/machinery/light{
 	dir = 1;
 	layer = 9.1;
 	light_type = /obj/item/light/tube/light_purpleish;
 	pixel_y = 21
 	},
+/obj/table/reinforced/chemistry/chemstorage,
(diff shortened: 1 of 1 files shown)


MintyPhresh:
All Chemlab Conversion PRs

Atlas - #15118
Clarion - #15120
Cog1 - #15117
Cog2 - #15125
Donut2 - #15124
Donut3 - #15123
Kondaru - #15122
Nadir - #15121
Oshan - #15119

PR #15127, opened 2023-07-26
Nadir security pod bay + associated adjustments

Kubius:
[BALANCE]
## About the PR 

Replaces Nadir's security bathroom with a secure internal pod bay linked to the exterior through a set of blast doors. This is stocked with two immersion suit vaults and a set of hardware for equipping and maintaining pods. Other associated changes include:

- The armory has received a pod weapons crate and an immersion suit vault.
- The section of breach hull around the new pod bay has been rearranged and is once again camera-monitored, and has an infrared beam monitor outside in the acid for additional advance warning.
- Security's main room has had one immersion suit vault replaced with a table; the TicketWriter and whistle have been moved to this table. (This doesn't reduce the complement of immersion suits available to Security, since the hangar has two entirely new suit vaults.)

## Why's this needed? 

The steady threat from Syndicate nuclear attacks and the new skirmishes by Salvagers are posing a threat that's difficult to respond to without defensive vehicles. While I do want to keep sub access lower on Nadir versus Oshan to avoid halls getting clogged by ill-advised drivers, a pair of defensive minisubs in a secured hangar shouldn't upset things too significantly, while making it much more viable to fend off vehicular threats.

## Changelog 

```changelog
(u)Kubius
(*)Nadir is now equipped with a Security pod bay.
```

### DIFF
--- a/maps/nadir.dmm
+++ b/maps/nadir.dmm
@@ -128,7 +128,7 @@
 	pixel_x = -12
 	},
 /obj/cable{
-	icon_state = "1-4"
+	icon_state = "2-4"
 	},
 /turf/simulated/floor/industrial,
 /area/station/maintenance/outer/sw{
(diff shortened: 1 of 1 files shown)


deathrobotpunch:
where will sec poop?

Kubius:
...In the same bathrooms everyone else does? Not a downside, necessarily. More opportunities for interaction en route.

PR #15128, opened 2023-07-27
change oxygen grenade name to the correct colour

deathrobotpunch:
## About the PR 
changes the oxygen grenade name from red oxygen grenade to blue oxygen grenade

## Why's this needed? 
because it has been bugging me how the red oxygen grenade is actually blue

## Changelog 

```changelog
(u)deathrobotpunch1
(*) the oxygen grenades name is now true to it’s appearance!
```


deathrobotpunch:
> The oxygen inside of it is red https://en.m.wikipedia.org/wiki/Octaoxygen

if i don’t see red air flying out of the grenade when it explodes then i beg to differ

Tarmunora:
> The oxygen inside of it is red https://en.m.wikipedia.org/wiki/Octaoxygen

the name is correct

PR #15138, opened 2023-07-27
makes the new lab table subtypes auto-connect

MintyPhresh:
[Mapping] [Chemistry] [Bug] 
## About the PR 
This PR repaths the tables added in #14808 to be subtypes of the "auto" table, which allows their textures to connect. Maps that have already been merged to include these were also changed.

## Why's this needed? 
Connected textures look way better and provide more counter space.


MintyPhresh:
This bad boy is a dependency for #15124, #15123, #15122, #15121, and #15117.

Issue #14804, opened 2023-07-01
Watering can says "...pours 60 units..." despite the can not even having 60


goonstation-issuebot:
### Labels

[BUG][TRIVIAL]

### Description

same as title

### Steps to reproduce

1.Empty a watering can from botany onto the floor
2.Scoop some water from the floor back into the watering can
3.Water a hydroponics tray

### Expected Behavior

To not say that you poured 60 units of the contents when there's less than 60 units

### Additional Information & Screenshots

https://cdn.discordapp.com/attachments/713838565200756866/1124797950913159199/Screenshot_158.png

Reported by: The_Real_Mr_N
On server: Goonstation 4 Roleplay: Sylvester
Active test merges: #12845 and #14694
Round log date: 2023-07-01-19-06
Reported on: 2023-07-01 20:27:56
Map: NADIR

TealSeer:
Fixed in #15149

PR #15145, opened 2023-07-28
Watering can reagent amount

Nitre88:
[Hydroponics] [QOL]
## About the PR 
Fixes issue #14804
Adds correct amount poured from watering cans(and other containers) when used on hydroponic trays

## Why's this needed? 
The game currently gives the amount per transfer which is the maximum instead of what the reagents are actually giving. 
This would improve the correctness of ingame feedback your given. When interacting with hydroponics


Nitre88:
I did an oopsie

PR #15152, opened 2023-07-28
TGUI Cryo Pod

Garash2k:
## About the PR 
Updates the Cryo Pod HTML UI into TGUI

There's one functional change that could be atomized or undone if it's controversial: being unable to press buttons while inside the pod (How can the spaceman hit the zap button or remove the beaker while inside the pod!)

Without upgrades
image

With upgrades (Health analyzer and Defib) (note: radiation, brain damage, and implants still show up without upgrades, the user just didn't have any in the previous screenshot)
image

While outside
image

Old UI
image

## Why's this needed? 
Forums wanted it! (It was on my shortlist too though!)
TGUI good

## Changelog 

```changelog
(u)Garash
(+)New UI for the Cryo Pods
```
[UI][Feature]


tgui/packages/tgui/interfaces/CryoCell.js
```diff
@@ -0,0 +1,224 @@
...
+ * @license ISC
+ */
+
+import { useBackend } from "../backend";
+import { Button, Box, Icon, AnimatedNumber, Section, LabeledList, ProgressBar, Dimmer, Table } from "../components";
+import { Window } from '../layouts';
+import { getTemperatureColor, getTemperatureIcon } from './common/temperatureUtils';
+import { ReagentGraph, ReagentList } from './common/ReagentInfo';
+import { HealthStat } from './common/HealthStat';
+import { DisplayBloodPressure, DisplayTempImplantRow, DisplayRads, DisplayBrain, DisplayEmbeddedObjects } from '../interfaces/OperatingComputer/index';
```
ZeWaka, line 14: @mordent-goonstation should these be in common?

tgui/packages/tgui/interfaces/CryoCell.js
```diff
@@ -0,0 +1,224 @@
...
+
+const damageNum = num => !num || num <= 0 ? '0' : num.toFixed(1);
+
+const OccupantStatus = {
+  Conscious: 0,
+  Unconscious: 1,
+  Dead: 2,
+};
+
+const occupantStatuses = {
```
ZeWaka, line 64: these should probably be in common

code/modules/atmospherics/machinery/unary/cryo_cell.dm
```diff
@@ -255,7 +304,22 @@
 			else
 				I.attack(src.occupant, user)
 
-		src.updateUsrDialog()
+		tgui_process.update_uis(src)
+
+	proc/insert_beaker(var/obj/item/reagent_containers/glass/I, var/mob/user)
+		if (I.cant_drop)
```
ZeWaka, line 310: need to check `can_act`

Garash2k:
Cryo pods and health scan (without the organ uprade) don't tell you about missing or hurt organs, but they do warn you about cyberorgans "Unknown augmented organs detected." New UI should try to do the same

Garash2k:
"Unknown augmented organs" indicator that was previously in health_scan added
image

tgui/packages/tgui/interfaces/CryoCell.js
Garash2k: Refactored to common/KeyHealthIndicators

Tarmunora (approved):

TobleroneSwordfish:
checks are for cowards

PR #15153, opened 2023-07-28
Fix bug that makes PDA UI open when you receive a message

TealSeer:
[BUG] [GAME OBJECTS]
## About the PR 

This fixes the bug that happens after you use your PDA to message another PDA (specifically clicking the reply hyperlink in chat) that makes your PDA open every time you receive a message after that point, or even randomly (its actually cause someone on station is scanning for PDAs). This is accomplished by removing two calls to the updateSelfDialog proc that don't really have a reason to be there (when you receive a message and when your PDA responds to a "scan" ping). 

## Why's this needed? 

This bug seems to be really old and also the bane of a lot of peoples existence including my own.

### DIFF
--- a/code/obj/item/device/pda2/base_os.dm
+++ b/code/obj/item/device/pda2/base_os.dm
@@ -886,8 +886,6 @@
 					if(length(src.hosted_files) >= 1)
 						src.CheckForPasskey(signal.data["message"], signal.data["sender"])
 
-					src.master.updateSelfDialog()
-
 				if("file_send_req")
 					if(!message_on)
 						return
--- a/code/obj/item/device/pda2/base_os.dm
+++ b/code/obj/item/device/pda2/base_os.dm
@@ -772,14 +772,6 @@
 				newsignal.data["owner"] = src.master.owner
 				src.post_signal(newsignal)
 
-				if(!ON_COOLDOWN(src.master, "report_pda_refresh", 1 SECOND))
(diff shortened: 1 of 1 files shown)


ZeWaka (commented):
not really a bug, pretty sure this was added intentionally because people wouldn't know they got a message

deathrobotpunch:
instead change it so when it opens it’s not on the screen you where last on to the one that is relevant

TealSeer:
Whatever the original intention was, you get both a chat message with the message and an audible beep when you get a message. I've never seen this behavior regarded as anything but an incredibly obnoxious bug because the window suddenly pulls focus and makes you stop moving if you're doing that. I see no reason to keep it in, there's no benefit and it can easily get you killed if you're in the middle of something involving moving when the window pops up

MintyPhresh:
> not really a bug, pretty sure this was added intentionally because people wouldn't know they got a message

This "feature" is intrusive to an insane degree, and it doesn't even function unless you've been sending PDA messages. Maybe if it was less inconsistent and didn't self-focus I'd be OK with it. But this PR is by far my preference, get that nasty popup outta my face

Laboredih123 (commented):
uh, lgtm?

Laboredih123:
those comments never showed up before huh

ZeWaka:
if people complain they can open up a revert PR

PR #15155, opened 2023-07-28
Resprites the JC Denton outfit.

DimWhat:
[clothing]
## About the PR 
Resprites the JC Denton outfit to be a bit more in line with current sprite standards
image

## Why's this needed? 
It looks better I think and I want the monkey version to look good too when I sprite that later.


Gannets:
Sprite looks better 👍

PR #15157, opened 2023-07-29
Adds a station wide annoucement when the entire security team dies.

unfunnyperson:
[BALANCE] [PLAYER ACTIONS]

## About the PR 
https://forum.ss13.co/showthread.php?tid=21581

When the entire sec team dies an annoucement is sent warning the crew to get improvised weapons and stay in pairs. 

The way this works is through the anti-mindhack implants. Each time someone dies with an anti-mindhack or gets it removed the game checks if anyone else has a functional anti-mindhack implant. If no implants are found in living/non-croyed officers the alert is sent.

Removing a implant and placing it in a random monkey to avoid the alert wont work because the implant only works if its inside its original owner.

The alert doesnt care about detectives or security assistants because detectives arent very trustworthy and sec assistants cant do much to stop antagonists.

## Why's this needed? 

If a rampager kills all of sec the crew should have some amount of warning before they start murdering the crew.

## Changelog 

```changelog
(u)UnfunnyPerson
(*)When the entire security team is killed(Not including detectives and sec assistants) an annoucement is sent to the station warning the crew.
```

### DIFF
--- a/code/obj/item/implant.dm
+++ b/code/obj/item/implant.dm
@@ -365,14 +365,14 @@ var/global/sec_wipe_alert = FALSE
 	/// Security member this implant originally belonged to
 	var/mob/living/original_owner
 
-	/// Check if anyone with a sec implant is alive, if sec is dead send the annoucement that sec has been killed
+	/// Check if anyone with a sec implant is alive and not in cryo, if sec is dead send the annoucement that sec has been killed
 	proc/send_annoucement()
 		var/sec_alive = FALSE
 		for_by_tcl(I, /obj/item/implant/health/security/anti_mindhack)
 			var/possible_implantee = I.loc
 			if (istype(possible_implantee, /mob/living))
 				var/mob/living/implantee = possible_implantee
-				if (I.functional && isalive(implantee))
+				if (I.functional && isalive(implantee) && !istype(implantee.loc, /obj/cryotron))
 					sec_alive = TRUE
 
 		if (!sec_alive && !sec_wipe_alert)
--- a/code/obj/item/implant.dm
+++ b/code/obj/item/implant.dm
@@ -401,7 +401,6 @@ var/global/sec_wipe_alert = FALSE
 
 	death_alert()
 		. = ..()
-		src.send_annoucement()
 		src.on_remove(src.owner)
 		qdel(src)
 
--- a/code/obj/item/implant.dm
+++ b/code/obj/item/implant.dm
@@ -368,7 +368,7 @@ var/global/sec_wipe_alert = FALSE
 	/// Check if anyone with a sec implant is alive, if sec is dead send the annoucement that sec has been killed
 	proc/send_annoucement()
 		var/sec_alive = FALSE
-		for (var/obj/item/implant/health/security/anti_mindhack/I in by_type[/obj/item/implant/health/security/anti_mindhack])
+		for_by_tcl(I, /obj/item/implant/health/security/anti_mindhack)
 			var/possible_implantee = I.loc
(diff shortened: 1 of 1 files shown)


flappybatpal:
This works when compiled in VSC but on a local server to try with two clients the global sec_wipe_alert is somehow set to 1 before I even have a client connected.

unfunnyperson:
> Cloning secoffs would break this, since they get cloned without implants (by design choice)

This is sort of intentional, Cloned secoffs can be mindhacked so the crew should still be aware there are no 100% trustworthy security officers left

code/obj/item/implant.dm
```diff
@@ -354,10 +354,50 @@ THROWING DARTS
...
 	impcolor = "b"
+	/// Set to FALSE on remove, this is here so that nerds cant remove the implant and put it in a random monkey to avoid triggering the alert.
+	var/functional = TRUE
+	/// Security member this implant originally belonged to
+	var/mob/living/original_owner
+
+	/// Check if anyone with a sec implant is alive, if sec is dead send the annoucement that sec has been killed
+	proc/send_annoucement()
+		var/sec_alive = FALSE
+		for (var/obj/item/implant/health/security/anti_mindhack/I in by_type[/obj/item/implant/health/security/anti_mindhack])
```
CaioFrancisco, line 371:
```suggestion
		for_by_tcl(I, /obj/item/implant/health/security/anti_mindhack)
```

Pretty much the same thing, but quicker.

Flaborized:
It took me like three days to remember I wanted to post on this but here we are :sheltersweat: . One issue I have with implementing this in such a 'foolproof' way is that if someone otherwise goes through the effort to impersonate a sec officer (going as far as to use their implant and change their DNA with a scrambler or changeling ability), this would just like, magically give them away. I think it's a little silly that you could get caught out over something automatic that *seems* like it wouldn't be able to tell the difference, in-world. I understand that it's the *point* to make the crew sure that shit's going down when it is, but I still don't like it! (This is just my personal view and not general dev opinion though so it's not set in stone or anything)

pgmzeta:
I like this idea, but I think a different target of the message would create some dynamic gameplay that would keep the harder detection code while putting a cap on its power. my suggestion is to PDA this alert to the HoP instead of a direct station announcement. stay with me here :)

The HoP comes to mind first because it's a 'staffing issue' in the setting. It gives the HoP something important to do later in the shift which helps their somewhat stale mid/endgame. Because the HoP could be dead, PDA-less, non-loyal, distracted, etc., it keeps a station-wide announcement decently probable but with zero guarantee it will happen every time. By telling the HoP and having them do the announcement, it gives no way for the crew to *know* it was automated detection, which puts an upper limit on how much this can influence a round.

This of course could be used by nefarious HoPs, but hey, that's ss13 :^)

deathrobotpunch:
if this duty was given to the Hop then the chances are that they never tell anyone

Azrun:
I agree with Flaborized's assessment as well.  This kind of automation undermines the ability for antagonists to engage with various forms of subterfuge.  In my head cannon it seems that thematically AI is the only thing I can think of that might be readily able to make this kind of determination as well as be responsible for it.  Additionally that could add another level of treachery in requiring dealing with the AI, but I don't think this level of automation is necessary.  If the AI (or brocops) wants to keep their eye on sec they can, keep a head count,

unfunnyperson:
closing for now, going to work on this later and ill revamp how I do the alert

unfunnyperson:
actually I might just leave this closed, implementing this in a way that allows for disgusing as a secoff would defeat the purpose of the alert imo

PR #15161, opened 2023-07-29
Adds the hitman jacket to the clothing booth and makes it no longer a labcoat subtype

DimWhat:
[clothing] [balance]
## About the PR 
Adds the "hitman" jacket to the clothing booth and changes it from a subtype of labcoat for balance reasons. This should allow players to properly make suits without being too fancy.

## Why's this needed? 
We have a LOT of shirts with ties and stuff but not many coats, even less of our coats are short coats! It feels like a missed opportunity.

## Changelog 

```changelog
(u)DimWhat
(+)Added a simple black jacket to the clothing booth.
```


deathrobotpunch:
make it so you can hide an orion silenced pistol inside of it

PR #15172, opened 2023-07-29
Merges the two dispensers in the bar into one

MintyPhresh:
[QoL] [Catering] [Game Objects] 
## About the PR
This PR adds a new type of dispenser, the bar dispenser, that can dispense all of the reagents that the alcohol and soda dispensers can, plus chocolate, mint, strawberry milk, and chocolate milk.

image

Note that the chem dispenser in Chemistry has 30 reagents. This dispenser has 31. I'm really not sure why they were separate in the first place.

This PR also cleans up a little of the code for all of the subtypes of /obj/machinery/chem_dispenser/alcohol/, since there were some repeated lines of code that didn't need to be there.

This PR does not add the dispenser to the maps in rotation.

## Why's this needed?
Like the chemlab, the bar would benefit appreciably from having a little big more counter space. And since the two separate ones have the same sprites, it's impossible to tell which is which without an alt-click, which is quite irksome.

### DIFF
--- a/code/modules/chemistry/Chemistry-Dispenser.dm
+++ b/code/modules/chemistry/Chemistry-Dispenser.dm
@@ -436,13 +436,13 @@ TYPEINFO(/obj/machinery/chem_dispenser)
 
 	dispense_sound = 'sound/misc/pourdrink2.ogg'
 
-//Combines alcohol and soda dispenser, plus mint, chocolate, and normal/strawb milk. Wow!
+//Combines alcohol and soda dispenser
 /obj/machinery/chem_dispenser/alcohol/bar
 	name = "bar dispenser"
 	desc = "You see a small, fading warning label on the side of the machine:<br>WARNING: Contents pending approval for human consumption. User assumes all risks.</br>"
-	dispensable_reagents = list("beer", "bitters", "bourbon", "champagne", "juice_cherry", "chocolate", "cider", \
+	dispensable_reagents = list("beer", "bitters", "bourbon", "champagne", "juice_cherry", "cider", \
 								"coconut_milk", "cola", "juice_cran", "gin", "ginger_ale", "grenadine", "juice_lemon", \
-								"juice_lime", "milk", "mint", "juice_orange", "juice_pineapple",  "rum", "strawberry_milk", "sugar", \
+								"juice_lime", "juice_orange", "juice_pineapple",  "rum", "sugar", \
 								"tea", "tequila", "juice_tomato", "tonic", "vanilla", "vermouth", "vodka", "water", "wine")
 
 // Dispenses any drink you want. Designed for the afterlife bar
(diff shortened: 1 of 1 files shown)


flappybatpal:
I think the description is a bit much considering it's a specialist bar equipment not made by Discount Dan. Instead of "not intended for human consumption" you could make it something like "pending approval for human consumption".

TobleroneSwordfish:
Eh, I always struggle to find specific elements in the chem dispenser so not a huge fan of combining these two, maybe better to just give them actually distinct sprites?

MintyPhresh:
> I think the description is a bit much considering it's a specialist bar equipment not made by Discount Dan. Instead of "not intended for human consumption" you could make it something like "pending approval for human consumption".
It's copied from the alcohol dispenser. Could be worth an update, fair.

> Eh, I always struggle to find specific elements in the chem dispenser so not a huge fan of combining these two, maybe better to just give them actually distinct sprites?
I like the solution in this PR just because it lowers the amount of inputs you need to do for something like a mimosa. 

Without the change, it's 7 inputs that are all over the center of the screen:
(insert bottle into alcohol disp -> dispense champagne -> eject bottle -> close window -> insert bottle into soda disp -> dispense OJ -> eject bottle)
But with, it's only 4: 
(insert bottle into bar disp -> dispense champagne -> dispense OJ -> eject bottle)

This is assuming you do everything perfectly. If you're making a three-component drink and forget to get two in one trip, you've just racked up another 3 inputs. Doesn't seem like much but it really adds up.

also, ehe... Nothing I sprite would get accepted unless it was like, a simple color band on the bottom of the dispenser.

Glamurio:
Big big big fan of this, also allowing some more niche drinks like chocolate milk on the standard menu is great.
Totally could use a new sprite, maybe ask if someone wants to supply one (or perhaps just a simple artistic change, different colors used on the sprite to indicate it will be enough for now)

Glcbitj:
Why add strawberry milk plus chocolate milk but not regular milk?

MintyPhresh:
> Why add strawberry milk plus chocolate milk but not regular milk?

I'm only adding reagents that are already available to bartenders on some maps (like Nadir and Kondaru) via the HAPPY CHEF dispenser, typically found in the shared space between the kitchen and bar. I didn't want to make this PR into a balance discussion, and I'm pretty sure Skeletons only get that ingest effect from unadulterated milk (which is my assumption as to why it's not in the dispenser to begin with).

pgmzeta:
good affordance (for a future PR): different icon for alcoholic drinks (e.g. a beer mug) when icon mode is on; maybe default-on for this sub-type. Will make sifting through this list visually a lot easier.

Tarmunora:
Mixed feelings on combining the dispensers, but going to say no to adding milk/mint/etc

MintyPhresh:
Bummer. But I guess it makes sense considering that the Chef still needs their own dispenser purely for access reasons. Will make those changes now

ZeWaka (approved):

PR #15175, opened 2023-07-30
Adds ONMASK flag to scarves / HoS medal

Glamurio:
[Clothing] [QoL]

## About the PR 
This PR adds a ONMASK flag similar to ONBACK which serves as a way to allow some items to be worn in the mask slot. It also moves some of the variables from `/obj/item/clothing/mask` to `/obj/item/clothing` to avoid conflicts with mask specific checks.

Items with the ONMASK are `/obj/item/clothing/suit/scarf`, `/obj/item/clothing/head/bandana`, `/obj/item/clothing/suit/scarfcape`, `/obj/item/clothing/head/goggles` and `/obj/item/clothing/suit/hosmedal`

## Why's this needed? 
There's a couple neck items that would be nice to wear for style purposes, but do not necessarily stand out enough to warrant giving up your suit or back slot for. Allowing them to be worn on the mask slot will allow for more customization.

## Changelog 

```changelog
(u)Ryou (Glamurio)
(+)Items that are worn on the neck, such as scarves and medals, can now be worn in the mask slot.
```


deathrobotpunch:
i also would really like to see the sec badges to have the ability to be attached to the sec uniform, since the exosuit slot is usually taken up by the armour vest that you don’t really want to replace for a badge

DisturbHerb:
There might be a conflict of interest here for me to say this, but I don't think jamming these into the mask slot is the best solution. Then again, I don't know what is.

Glamurio:
> There might be a conflict of interest here for me to say this, but I don't think jamming these into the mask slot is the best solution. Then again, I don't know what is.

Yeah, I've thought a long time about this. Currently, we're at this awkward place where we don't have enough items to justify a new slot being made (and all the headache it involves), but shoving them in the suit slot / head slot as we have been is not a good solution either. 

There is the option of maybe implementing more of a "clothing craft system", where you could stick any kind of accessory to any kind of hat, and neck items to the suit slot. To do that dynamically though would require a pretty massive PR, and requires to first separate all the "true hats" from the "accessory hats".



> i also would really like to see the sec badges to have the ability to be attached to the sec uniform, since the exosuit slot is usually taken up by the armour vest that you don’t really want to replace for a badge

This is a cool idea, I think I'll implement that in a separate PR. Maybe just to the vest, so it's actually visible. Will think about it.

Tarmunora:
Oh right forgot a comment: 

- Pushes a bunch of variables up to `.../clothing` without handling any of the functionality
- Jamming stuff into the mask slot like this is pretty meh
- Form competing with function is by design

PR #15178, opened 2023-07-30
Adds pronouns to human examine text

CaioFrancisco:
## About the PR 

Simple change, right? One line, just a quick browse around the code and you find all the important bits to make this feature work. Well, because I'm a **great** coder, it took me 40 minutes to do because I had to browse Youtube for something that works as background noise, and I kept getting interrupted every 5 or 10 minutes. Anyway,

this PR adds pronouns to the examine text. This will make pronouns easier to see, as you don't have to squint on the examine mouse-over text, nor you have to discover the examinee's pronouns by bits through the "person seems to have something on **their** head."

## Why's this needed? 

Pronouns should be easier to notice and memorize, and keeps getting expanded with the [cyborg pronouns](https://forum.ss13.co/showthread.php?tid=21592) suggestion thread on the forums, which got a pretty big traction after only having just a day of existing.

## Examples

Screenshot_770
Screenshot_769

## Changelog 

```changelog
(u)Caio029
(+)Pronouns will now be shown on the examine text of humans.
```

### DIFF
--- a/code/mob/living/carbon/human/procs/get_desc.dm
+++ b/code/mob/living/carbon/human/procs/get_desc.dm
@@ -29,12 +29,14 @@
 			. = "<br>[src.bioHolder.mobAppearance.flavor_text]"
 		catch
 			//nop
-
-	. +=  "<br><span class='notice'>*---------*</span>"
-
 	// crappy hack because you can't do \his[src] etc
 	var/t_his = his_or_her(src)
 	var/t_him = him_or_her(src)
+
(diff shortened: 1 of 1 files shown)


CaioFrancisco:
Linked to some examples of the examine text.

ZeWaka:
Why not just do `you look closely at XYZ XYZ (they/them)`

CaioFrancisco:
> Why not just do `you look closely at XYZ XYZ (they/them)`

I'd like to make it so the pronouns and the name are in separate lines, to avoid possible confusion, but I can do that.

ZeWaka (approved):

PR #15182, opened 2023-07-30
Fix medal frames always returning standard glass sheets

Glamurio:
[Internal] [QoL]

## About the PR 
This adds a new `var/glass_type` to `/obj/decal/poster/framed_award`, and subsequently uses this to determine what kind of glass to return. It defaults to `/obj/item/sheet/glass`.

## Why's this needed? 
-Take out glass from frame
-Put plasma glass in
-Plasma glass turns into regular glass
:o


flappybatpal:
Most niche PR goes to...

flappybatpal (approved):

PR #15187, opened 2023-07-30
Pull borg pronouns from preferences when latejoining and being constructed

TealSeer:
[SILICONS] [BUG]
## About the PR 

Copies a client's pronoun preference in character setup to 1. SICC cyborgs they're latejoining into, and 2. A freshly built cyborg their brain is inside on activation

## Why's this needed? 

Cyborgs only got pronouns if they were one of the roundstart ones, this makes the behavior consistent in (I believe) all the ways someone can become a cyborg


CaioFrancisco:
[This PR has originated from this forum's thread, by the way.](https://forum.ss13.co/showthread.php?tid=21592) I'd just want to have that documented.

PR #15194, opened 2023-07-31
Fix Plasmastone & Molitz not working after one use

amylizzle:
## About the PR 
Adds a bit of logic to these triggers so that the parent material becomes mutable when they fire.

## Why's this needed? 
Plasmastone and molitz currently only trigger a couple times, because they're operating on the immutable one in the cache which is bad.

### DIFF
--- a/code/modules/materials/Mat_MaterialProcs.dm
+++ b/code/modules/materials/Mat_MaterialProcs.dm
@@ -449,7 +449,7 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
 			return
 		if(!owner.material.isMutable()) //this is a little hacky, but basically ensure it's mutable and then do the trigger
 			owner.material = owner.material.getMutable()
-			return owner.material.triggerExp(owner)
+			return owner.material.triggerExp(owner, sev)
 		var/turf/target = get_turf(owner)
 		if(sev > 0 && sev < 4) // Use pipebombs not canbombs!
 			if(molitz.iterations >= 1)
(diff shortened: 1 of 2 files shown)


code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -346,6 +346,9 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
 		var/turf/T = get_turf(location)
 		if(!T || T.density || !istype(location))
 			return
+		if(!location.material.isMutable()) //this is a little hacky, but basically ensure it's mutable and then do the trigger
+			location.material = location.material.getMutable()
+			return location.material.triggerTemp(location, 0)
```
TobleroneSwordfish, line 351: won't this just fail if you pass it 0 temp?

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -441,6 +447,9 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
 		var/datum/material/crystal/molitz/molitz = owner.material
 		if(molitz.unexploded <= 0)
 			return
+		if(!owner.material.isMutable()) //this is a little hacky, but basically ensure it's mutable and then do the trigger
+			owner.material = owner.material.getMutable()
+			return owner.material.triggerExp(owner)
```
TobleroneSwordfish, line 452: same here, no `sev`

code/modules/materials/Mat_MaterialProcs.dm
amylizzle: for this material the same proc is used for temp and exp triggers, and in both the second argument is ignored. It is very hacky, and this continues that tradition.

code/modules/materials/Mat_MaterialProcs.dm
amylizzle: this one is a mistake, fixed

PR #15196, opened 2023-07-31
Preassembled Pod Kits

Garash2k:
## About the PR 
Adds 3 manufacturing recipes: `/datum/manufacture/pod/preassembeled_parts`, `/datum/manufacture/sub/preassembeled_parts` and `/datum/manufacture/putt/preassembeled_parts`.
This PR does not implement them in the game, for atomization's sake, this can be done in another PR for either the regular maps' Ship Component Fabricator and/or the pod wars one.
Preassembled kits result in the same ships as their current counterpart, with the difference that they have:
1. Action bars for building steps
2. "Built-in" Engine manifolds, Circuit boards, Control interfaces, metal, glass and wires.
3. Different building steps that end up in roughly the same amount a time pod used to take to build.
    - The steps are Wrench, Weld, Screw, Put armor, Wrench, Weld, Screw, each with a 10s action bar.  
    **[Recording of a pod construction](https://files.catbox.moe/7a36ut.mp4)**
5. Their cost and build times are upped the sum of the built in parts (roughly)  
    image

## Why's this needed? 
Follow up of an idea I had a year ago: https://forum.ss13.co/showthread.php?tid=17100&pid=183818#pid183818
The pod building experience isn't great. Every time I go back to building a pod for the first time in months, I tend to confuse what's a (Pod) Engine Manifold vs a MiniPutt Engine Manifold vs an actual engine. Every icon being the down arrow mechanic box doesn't help. Having to take a break from pod construction and going out to scavenge for wires or reinforced glass isn't much fun either.
With this new system, all you need is the pod kit, some armor, a wrench, a wielder and a screwdriver. To improve the pod wars experience, we could also have "Building Pods for Dummies" image guides scattered around, listing what you need and what the steps are.
Action bars and help text are nice.

[Vehicles][Rework]

### DIFF
--- a/code/modules/transport/pods/preassembled_frame_kit.dm
+++ b/code/modules/transport/pods/preassembled_frame_kit.dm
@@ -106,77 +106,98 @@ ABSTRACT_TYPE(/obj/structure/preassembeled_vehicleframe)
 /* Construction                */
 /*-----------------------------*/
 
-/obj/structure/preassembeled_vehicleframe/attackby(obj/item/W, mob/living/user)
+/obj/structure/preassembeled_vehicleframe/attackby(obj/item/I, mob/living/user)
+	var/datum/action/bar/icon/callback/action_bar
+
+	if (I)
+		action_bar = new /datum/action/bar/icon/callback(user, src, src.step_build_time, \
+			/obj/structure/preassembeled_vehicleframe/proc/step_wrench_1, \
+			list(user), I.icon, I.icon_state, null, null)
+		action_bar.maximum_range = 2
+
 	switch(src.stage)
 		if(BUILD_STEP_PLACED)
-			if (iswrenchingtool(W))
+			if (iswrenchingtool(I))
 				user.visible_message("[user] begins securing the frame...")
 				playsound(src.loc, 'sound/items/Ratchet.ogg', 50, 1)
-				SETUP_GENERIC_ACTIONBAR(user, src, src.step_build_time, /obj/structure/preassembeled_vehicleframe/proc/step_wrench_1, list(user), \
-			 		W.icon, W.icon_state, "[user] finishes wrenching the frame parts together.", null)
+				action_bar.proc_path = /obj/structure/preassembeled_vehicleframe/proc/step_wrench_1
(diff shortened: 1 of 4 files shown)


flappybatpal (changes requested):
Big pod is difficult to construct because you are too far away except if standing near the bottom left, causing your action bars to silently fail.

Step advice should be included in the interaction text/description so it's clear to players what the next construction step should be.

code/modules/transport/pods/preassembled_frame_kit.dm
```diff
@@ -0,0 +1,270 @@
...
+	bound_width = 64
+	bound_height = 64
+	box_type = /obj/item/preassembled_frame_box/pod
+	vehicle_name = "Pod"
+
+/*-----------------------------*/
+/* Construction                */
+/*-----------------------------*/
+
+/obj/structure/preassembeled_vehicleframe/attackby(obj/item/W, mob/living/user)
```
flappybatpal, line 109: Please use I for item vars - `obj/item/I`

code/modules/transport/pods/preassembled_frame_kit.dm
```diff
@@ -0,0 +1,270 @@
...
+/*-----------------------------*/
+
+/obj/structure/preassembeled_vehicleframe/verb/deconstruct()
+	set src in oview(1)
+	set category = "Local"
+
+	if (usr.stat)
+		return
+
+	boutput(usr, "Deconstructing frame...")
```
flappybatpal, line 240: I'd prefer a visible_message informing others of the action being performed than boutput.

Garash2k:
> Big pod is difficult to construct because you are too far away except if standing near the bottom left, causing your action bars to silently fail.
Fixed by using `maximum_range`

> Step advice should be included in the interaction text with empty hand/description on examine so it's clear to players what the next construction step should be.
Step advice text added to empty hands hits, but as far as examine text is concerned, there's a help_message that shows up on alt+double click. I could replace it with a dynamic description, but it seems like a pretty good use for help_messages.

flappybatpal (approved):

ZeWaka (approved):

PR #15203, opened 2023-08-01
add mister to Atlas and the afterlife bar garden

soapcyst:
## About the PR 
This PR simply adds mister machines to hydroponics in atlas and the afterlife bar garden.
atlas:
image
afterlife bar:
image

## Why's this needed? 
Every other map currently in rotation has one by default, why shouldn't these ones have it as well

## Changelog 

```changelog
(u)soapcyst
(+)Added misters to atlas and afterlife bar
```

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -5875,6 +5875,7 @@
 /obj/cable{
 	icon_state = "2-4"
 	},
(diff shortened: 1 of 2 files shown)


soapcyst:
[QoL]

PR #15204, opened 2023-08-01
ABCU marker rectangle select, overwrite notice prompts, better usability

kbsmilk:
## About the PR 
ABCU marker becomes a lot friendlier to use. Rectangle select for less clicks. Audio feedback added to some actions.
Range limit increased to 10 tiles, and cooldown removed (via changing to pixelaction).
Help/info messages were rewritten to be more clear about the ABCU system as a whole.
User now gets confirmation prompts when overwriting or deleting a saved blueprint.

## Why's this needed? 
The marker tool is way less tedious to use now.

**Video clip of rectangle select :)**
https://github.com/goonstation/goonstation/assets/40084056/cb787093-861e-4ec6-ab72-6eb0644b2c41

dreamseeker_pZgM63KfM6
dreamseeker_Cg1NGjQwcH

## Changelog 

```changelog
(u)Smilg
(+)ABCU marker can now rectangle select, and has longer range.
```

### DIFF
--- a/code/WorkInProgress/blueprints.dm
+++ b/code/WorkInProgress/blueprints.dm
@@ -365,6 +365,12 @@
 			//rebuild powernets etc.
 */
 
+#define SELECT_SKIP 0
+#define SELECT_FIRST_CORNER 1
+#define DESELECT_FIRST_CORNER 2
+#define SELECT_SECOND_CORNER 3
+#define DESELECT_SECOND_CORNER 4
+
 /obj/item/blueprint_marker
 	name = "Blueprint Marker"
 	desc = "A tool used to map rooms for the creation of blueprints."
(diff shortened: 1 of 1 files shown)


code/WorkInProgress/blueprints.dm
```diff
@@ -477,16 +479,80 @@
...
+			if (1,2) // set to 1 or 2 by use-in-hand option list
+				qdel(corner1img)
+				selectcorner1 = target
+				selecting += 2 // if 3 then select, if 4 then deselect
+				corner1img = image('icons/misc/old_or_unused.dmi', selectcorner1, "marker", layer = HUD_LAYER)
+				user << corner1img
+				playsound(src.loc, 'sound/machines/tone_beep.ogg', 15)
+				return
+
+			if (3,4)
```
Tarmunora, line 497: values of `selecting` should probably be `#define`d for clarity

Tarmunora (changes requested):

Tarmunora (approved):

PR #15207, opened 2023-08-01
Readds the wannabe pirate costume to the clothing booth

Mantweerp:
[Game Objects]

## About the PR 
Readds the wannabe pirate costume to the clothing booth, which was (in my opinion) wrongly removed from the clothing booth for being bloat last year.

## Why's this needed? 
The only way of obtaining it now is from the novelty clothing crate from cargo, which costs 7500 credits, and defeats the 
 purpose of the clothing booth (providing a reliable and Relatively cheap way of obtaining clothing).
Also it makes sense for there to be a pirate outfit for the two pirate hats that're already in the costumes section

### DIFF
--- a/code/modules/vending/clothingbooth_items.dm
+++ b/code/modules/vending/clothingbooth_items.dm
@@ -1131,6 +1131,10 @@ ABSTRACT_TYPE(/datum/clothingbooth_item/costume)
 	slot = SLOT_WEAR_SUIT
 	cost = PAY_TRADESMAN/2
 
+	/datum/clothingbooth_item/costume/guybrush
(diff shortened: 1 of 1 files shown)


pgmzeta:
This change makes running Pirate RP events slightly more difficult, but that constraint should perhaps be handled alongside pirate event code/map tweaks. Not a blocker IMO but surfacing here as I vaguely recall some pirate event resources being non-public or local.

flrsh:
I don't think having a pirate costume in the clothing booth makes sense, especially when, as mentioned, there can be real actual pirates. If anything we should remove the hats.

PR #15208, opened 2023-08-01
Tanning beds burn vampires alive

Bartimeus973:
## About the PR 
Suggested here : https://forum.ss13.co/showthread.php?tid=21618
I found the idea hilarious so here it is.

Vampires take 15 burn damage and have a 15% chance to scream per tick while inside a sunbed. They will also get custom chat messages and audibly sizzle.

Sunbeds are now also an appropriate way of disposing of vampires corpses, leaving behind all the items they had on them.

## Why's this needed? 
It's pretty funny.

## Changelog 

```changelog
(u)Bartimeus973
(*)Vampires should now avoid using tanning beds.
```


Roninnozlo:
the best part is that it's funny AND makes sense

code/obj/machinery/traymachines.dm
```diff
@@ -526,6 +526,28 @@ ABSTRACT_TYPE(/obj/machine_tray)
...
+								src.visible_message("<span class='alert'>A puff of smoke erupts from the machine as it grinds to a halt! It smells like a graveyard caught fire!</span>")
+								var/turf/T = get_turf(src)
+								if (istype(T))
+									var/datum/effects/system/bad_smoke_spread/smoke_effect = new /datum/effects/system/bad_smoke_spread/(T)
+									smoke_effect.set_up(15, 0, T, null, "#000000")
+									smoke_effect.start()
+								src.locked = FALSE
+								power_usage = initial(power_usage)
+								playsound(src.loc, 'sound/machines/ding.ogg', 50, 1)
+								update()
```
ZeWaka, line 549: refactor this into a end_tanning proc or something

ZeWaka (changes requested):

code/obj/machinery/traymachines.dm
Bartimeus973: Donezo

PR #15213, opened 2023-08-02
Build.dm reshuffle + new testmaps

TheTyrant25:
[INTERNAL]
## About the PR
- Adds two new maps to the blank_maps folder, `unsimmed_test.dmm` and `simmed_test.dmm`, for coders to edit and use. 50x50 empty maps which should load as quick as a bullet, while also skipping all other Z levels. Comes with an area that doesn't require power and has fullbright ambient lighting for extra convenience. One has all unsimmed turfs and one has all simmed turfs.
- Cleans up the commenting of build.dm a little bit too. Just thought I'd clean the tabbing of comments while i was there. (Some of it looks weird in github but i assure you it looks great in vsc)
- rearranged some stuff in build.dm to make them clearer or stand out better
- removed the gehenna and icarus build defines since they don't work (the map files are gone)
- reshuffled some devtest files to make it a bit more compact. Devtest loading is now handled in `testmap.dm` alongside the other two dev maps
- moves map editor defines up to the top (just the one currently) for easier finding.

## Why's this needed? 
Makes the build.dm file clearer and cleaner. Makes devtest and the blank maps more blatantly obvious to bug testers.
Makes it so that `STOP_DISTRACTING_ME` doesn't appear to include `Z_LOG_ENABLE` which it of course doesn't but it wasn't fully clear before.
More tools for coders to quickly test features is nice. It means you don't have to load all of atlas in its 300x300 glory, which should speed up some things a tiny bit.


TheTyrant25:
who broke all this rust g then eh
me checks are failing and my crops are unwatered

ZeWaka:
i don't really see a use for this, loading a 20x20 map vs loading atlas isn't really going to have significant impact on loading times i don't think (unless you can show otherwise)
a more relevant map would be a dedicated dev testing one that's more like a small station map, but honestly i think atlas serves it well currently

Tarmunora (changes requested):
Actually could you bump it up to like, 50x50 or so? On second consideration, just having a real nice open space would be good

Tarmunora (approved):

TheTyrant25:
on second thought, im putting this onto draft until #15227 is merged so that i can incorporate that properly and not clash

Tarmunora:
To be frank, with devmap I no longer really feel the need for this

PR #15216, opened 2023-08-02
Changes damaged/scorched turfs in maps to use a mapping helper object instead 

CheffieGithub:
## About the PR 
Remake of #14816, conflicts were not kind.
Changes floors and the damaged1/scorched2/etc tiles to use a burner / damager object instead.
`break_tile()` `burn_tile()` `restore_tile()` moved up to /turf along with related variables.

Keeps the old subtypes but changes them to just /scorched and /damaged since they can be used in map generation.

## Why's this needed? 
image

From left to right:

The damaged subtypes only look like default floors / plating.


BatElite:
So do these helpers have sprites?

CheffieGithub:
not unique ones they reuse a damaged tile and scorched tile overlay sprite

ZeWaka:
How much time does this add to map loading cog1?

CheffieGithub:
not sure of a better way of getting times but just loading the map and timing from the main screen to 'ready' cog 1 took 2.25 seconds with and 2.30 seconds without. Im guessing very little impact on loads with some expected variance

CheffieGithub:
The mapping gods have truly blessed us with 770,000 lines of changes

CheffieGithub:
going to close to split this up

PR #15220, opened 2023-08-02
Light space suit change (atmos_movespeed)

Garash2k:
[Controversial][Input Wanted]

## About the PR 
Implements a new property, atmos_movespeed, which doesn't slow you down unless the pressure on a turf is above half an atmosphere (50kPa)
Adds a atmos_movespeed property to the light engi suits. The property is speedier than the standard space_movespeed on eva suits.

## Why's this needed? 
There seem to be a fair amount of dissatisfaction towards the light engi suits. I asked #spacestation-chat what they though of giving them a small amount of slowdown and 4 people instantly replied with "remove them."
Personally I don't feel as strongly against them, but I'm still not a fan how they're strictly better than not wearing anything at all.  It makes me a bit unhappy when I roll CE and I have to choose between looking like a powergaming nerd and putting the suit on at roundstart or "playing wrong" by leaving power on the table.

After getting input, I've learned that the purpose of the suits is for engineers to fix breaches without being slowed down, atmos_movespeed lets them serve their purpose still.

## Changelog 

```changelog
(u)Garash
(+)New type of slowdown implemented, "atmos_movespeed," which doesn't slow you down in low pressure environments or in space.
(+)Light space suits now have a small mount of atmos_movespeed
```
 [Clothing][Balance]


Garash2k:
Crud, I coded an atmos_movespeed property, then I #imcoder made me remember the "puncturable spacesuit" solution which sounded really neat

Penalty starts applying when the air is at more than half an atmosphere.

Video of atmos_movespeed in action with a debug log for `special_movedelay_mod()`: https://files.catbox.moe/jd2oub.mp4

Garash2k:
Checked with sord in #imcoder, this is cool for now, and later on in the medium term some general outer suit damage system would be neat.

Sord213 (commented):
code looks good to me but I'll let someone more competent review. otherwise I approve

deathrobotpunch:
what about the ntso pressure suit, are you going to change that too?

Garash2k:
> what about the ntso pressure suit, are you going to change that too?

Not planning to? This change doesn't affect NT pressure suits (`/obj/item/clothing/suit/space/ntso`), it only affects light space suit (`/obj/item/clothing/suit/space/light`) and their helmets.

Garash2k:
!merge_upstream

PR #15240, opened 2023-08-04
Latejoining Stowaways, Pilots, and Heavy Sleepers no longer get announced

MintyPhresh:
## About the PR 
Makes the announcement computer not announce characters who have ostensibly been on the station for the whole shift, or who aren't on the station.

## Why's this needed? 
Doesn't make sense for the announcement computer to be all-knowing about stowaways & pilots, and heavy sleepers have been on station all shift. Makes things like undercover secasses possible for non-roundstart players.

```changelog
(u)Mintyphresh
(+)Latejoining players with the Stowaway, Pilot, or Heavy Sleeper trait will no longer be announced.
```

### DIFF
--- a/code/obj/machinery/computer/announcement.dm
+++ b/code/obj/machinery/computer/announcement.dm
@@ -206,8 +206,8 @@
 	proc/announce_arrival(var/mob/living/person)
 		if (!src.announces_arrivals)
 			return 1
-		if ((person.traitHolder.hasTrait("stowaway")) || (person.traitHolder.hasTrait("sleepy")))
-			return 1 //people who have been on the ship the whole time won't be announced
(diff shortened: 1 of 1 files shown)


MintyPhresh:
> It feels like there should be a points re-arrangement for the traits (except maybe stowaway), then. Coming in dark on the station has it's enormous plus, specially if you are an antagonist.

It's definitely a massive boost, and something to seriously consider... But you can "join dark" by just getting in at the start of the round, maybe finding a hiding spot and going AFK. I'm not sure this is necessary

TobleroneSwordfish (approved):

Sbmhawk:
I don't see why heavy sleepers should be excluded from the manifest. Stowaways and pilots sure, as I don't believe they are technically NT employees, but that does not apply to heavy sleepers.

MintyPhresh:
Oh, yeah, no, they're on the manifest after they join, there's just no way for the manifest to know that beforehand

PR #15243, opened 2023-08-04
- antag status effects

UrsulaMejor:
## About the PR 

Status effect effect for when you're antagonist

## Why's this needed? 

Sometimes people crash at round start or dismiss the popup by reflex and then forget. Shrug.

## Changelog 

```changelog
(u)Emily
(*)Antagonist Status Effect so you can tell you're an antagonist. Click the effect to quick-open your notes.
```


flappybatpal:
This is a good idea, people do miss it.

Tarmunora (approved):

MintyPhresh:
Do note that this can make it easier for observers to find antags in the new observer menu without setting DNR. Personally I don't agree with the restriction being there in the first place, but implementing this will make the restriction weaker.

Sbmhawk:
> Do note that this can make it easier for observers to find antags in the new observer menu without setting DNR. Personally I don't agree with the restriction being there in the first place, but implementing this will make the restriction weaker.

9/10 times if you are observing the antag they have an ability or an item that already outs them.

You just cant instantly tell who the antags are from the observe menu without DNR.

TobleroneSwordfish (changes requested):
Doesn't handle cloning, mindswaps, headspiders, borging etc. Might be possible to hijack client image groups for this somehow or just handle mind transfer signals.

Mister-Moriarty:
> Doesn't handle cloning, mindswaps, headspiders, borging etc. Might be possible to hijack client image groups for this somehow or just handle mind transfer signals.

I believe it would be best to create a status effect subtype that tracks when a mind moves between mobs, similar to how `/datum/mind_mob_overlay` functions.

PR #15256, opened 2023-08-05
Parallax on Generated Planets

Azrun:
## About the PR 
Adds handling needed for parallax to be applied to an arbitrary zlevel.
Support for dynamically generated layers to be applied with same conditions/options
% chance for Desert and Snow to use randomized occluded foreground parallax layers similar to existing azones.

Added foreground occlusion object to allow for exclusion of turfs/regions that may not be visible to players.

https://streamable.com/cojptj

## Why's this needed? 
Improve immersion and dynamic experience of procedural generated content.

### DIFF
--- a/_std/map.dm
+++ b/_std/map.dm
@@ -24,7 +24,6 @@ var/list/default_z_level_parallax_settings = list(
 		/atom/movable/screen/parallax_layer/asteroids_far,
 		/atom/movable/screen/parallax_layer/asteroids_near,
 		),
-	"[Z_LEVEL_SECRET]" = list(),
 	)
 
 /// A list of each z-level define and it's current associated parallax layer types.
--- a/_std/map.dm
+++ b/_std/map.dm
@@ -24,6 +24,7 @@ var/list/default_z_level_parallax_settings = list(
 		/atom/movable/screen/parallax_layer/asteroids_far,
 		/atom/movable/screen/parallax_layer/asteroids_near,
 		),
+	"[Z_LEVEL_SECRET]" = list(),
 	)
 
 /// A list of each z-level define and it's current associated parallax layer types.
--- a/code/WorkInProgress/artemis/GeneratePlanets.dm
+++ b/code/WorkInProgress/artemis/GeneratePlanets.dm
@@ -273,9 +273,11 @@ var/global/datum/planetManager/PLANET_LOCATIONS = new /datum/planetManager()
 			/atom/movable/screen/parallax_layer/foreground/dust/sparse=list(color=color_matrix, scroll_speed=scroll_speed*1.5, scroll_angle=angle)
 		)
 
+	// Occlude overlays on edges
 	if(planet_area.area_parallax_layers)
 		for(var/turf/cordon/CT in planet_area)
-			CT.update_parallax_occlusion_overlay()
+			new/obj/foreground_parallax_occlusion(CT)
+
 
 	//Populate with Biome!
 	var/turfs = block(locate(region.bottom_left.x+1, region.bottom_left.y+1, region.bottom_left.z), locate(region.bottom_left.x+region.width-2, region.bottom_left.y+region.height-2, region.bottom_left.z) )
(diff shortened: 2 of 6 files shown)


_std/map.dm
```diff
@@ -24,6 +24,7 @@ var/list/default_z_level_parallax_settings = list(
 		/atom/movable/screen/parallax_layer/asteroids_far,
 		/atom/movable/screen/parallax_layer/asteroids_near,
 		),
+	"[Z_LEVEL_SECRET]" = list(),
```
ZeWaka, line 27: This is already above

code/WorkInProgress/artemis/GeneratePlanets.dm
```diff
@@ -243,6 +243,42 @@ var/global/datum/planetManager/PLANET_LOCATIONS = new /datum/planetManager()
...
+								0, 0, 0, 1,
+								0, 0, 0, -1)
+		planet_area.area_parallax_layers = list(
+			/atom/movable/screen/parallax_layer/foreground/dust=list(color=color_matrix, scroll_speed=scroll_speed, scroll_angle=angle),
+			/atom/movable/screen/parallax_layer/foreground/dust/sparse=list(color=color_matrix, scroll_speed=scroll_speed*1.5, scroll_angle=angle)
+		)
+
+	// Occlude overlays on edges
+	if(planet_area.area_parallax_layers)
+		for(var/turf/cordon/CT in planet_area)
```
ZeWaka, line 278: Aren't cordons in their own area?

ZeWaka (changes requested):

code/WorkInProgress/artemis/GeneratePlanets.dm
Azrun: I didn't want to add additional processing where it wasn't necessary, but that does seem like a much better solution.

Mister-Moriarty: They are, but such a snowflake check is required to permit full-tile occlusion overlays, so to prevent visual artefacts.

image

code/WorkInProgress/artemis/GeneratePlanets.dm
```diff
@@ -243,6 +243,42 @@ var/global/datum/planetManager/PLANET_LOCATIONS = new /datum/planetManager()
...
+			new/obj/foreground_parallax_occlusion(CT)
```
Mister-Moriarty, line 279: I would recommend replacing `/obj/foreground_parallax_occlusion` with an `UpdateOverlays()` call, so to remain consistant with existing occlusion overlays and prevent edge case bugs with using object overlays.

Using a key other than `"foreground_parallax_occlusion_overlay"` would be advisable, so to prevent undesirable interactions with the existing occlusion overlays.

Mister-Moriarty (changes requested):

code/WorkInProgress/artemis/GeneratePlanets.dm
Azrun: This is specifically to do deal with the the fact that cordon turfs are invisible.

Mister-Moriarty: The `RESET_ALPHA` or `KEEP_APART` appearance flags may also be used to resolve that.

code/WorkInProgress/artemis/GeneratePlanets.dm
Azrun: Okay, that solution doesn't really work as I kind of like the cordon's being part of the allocated area (at least at the moment).  Reverted to be snowflake. 😞

code/WorkInProgress/artemis/GeneratePlanets.dm
Azrun: Ooooo

Azrun: Seems to still be an issue due to invisibility. 🙁

Azrun: Though having the overlay present should help standard turfs expanding their overlay to the edge and not just having a strip down the center.

PR #15265, opened 2023-08-06
Adds a new 3 tc revhead item that lets revheads see how difficult it would be to convert someone

unfunnyperson:
[FEATURE] [PLAYER ACTIONS]
## About the PR 
Adds a 3 tc rev uplink item.

The item is a pair of mesons that allow the user to see flash protection, loyalty implants, and if the person is an unconvertable job on examine.

If a non-headrev wears the goggles all they can see is flash protection to prevent nerds from easily finding out who revs are.

## Why's this needed?
Allows for more stealth as a revhead while keeping counterplay. 
Also we dont have many unique revhead items.

## Changelog 

```changelog
(u)UnfunnyPerson
(+)New 3 tc revhead item. Revolutionary Goggles. They allow the revhead to see flash protection, loyalty implants, and unconvertable roles when examining a player.
```


unfunnyperson:
> Seems rather "cheap" to give it as a magical ability to regular humans, vamps also struggle with it. Maybe if it was a purchasable eyewear that 1. gave that power 2. had no eye protection itself as a tradeoff 3. looked as a regular eyewear on inspect then I could see it be more "realistic" for the gamemode. Exposing heads of staff / security in disguise also seems rather strong to get for free in current implementation.

good idea

deathrobotpunch:
👍

unfunnyperson:
unsure about the pricing on this, lmk if you think it needs changes

deathrobotpunch:
why close?

unfunnyperson:
I feel the coding could be better and I didnt feel like redoing the entire item

PR #15266, opened 2023-08-06
All gloves now leave partial prints + partial prints changes

Bartimeus973:
## About the PR 
All gloves now leave partial prints, however the readability of the prints will greatly vary depending on the gloves.

When a glove leaves a muddled print, the "fake" characters are now highlighted in red for readability.

Gloves now always leave a material on prints, so for example even latex gloves now leave "latex fibers".

Gloves which were already easily readable (latex for example) now have a 40% chance (was 20%) to replace a character in a fingerprint string to a random corrupted character.
Example of new latex prints: 

latexprints

Gloves which were unreadable now leave partial prints which may be up to 85% scrambled.
Example of new insulative glove prints with 85% chance for each character to be muddled:

insulprints

Syndicate gloves like swat gloves, bladed gloves or power-gloves have a 95% scramble chance.
Example of bladed gloves prints: 

bladedprints

Chameleon clothing gloves have higher than normal scrambling chance compared to their regular counterpart.

## Why's this needed? 
Forensics is a bit weak, especially on classic. It is absolutely trivial to pick up some insuls or black gloves, do crime, then swap gloves and throw the old pair in the grinder, leaving forensic gathering with very little purpose. By leaving behind heavily obfuscated prints it is now possible if enough traces are found to make up a whole print and get an investigation going.

In order to not be a straight nerf, "weaker" gloves like latex are now a bit more scrambled by default so that some research work might be required.

## Changelog 

```changelog
(u)Bartimeus973
(*)All gloves now leave partial prints. Some gloves (like insulated gloves) have harder to read prints!
```

### DIFF
--- a/code/obj/item/clothing/chameleon_js.dm
+++ b/code/obj/item/clothing/chameleon_js.dm
@@ -1273,7 +1273,7 @@
 	var/sprite_hand = 'icons/mob/inhand/hand_feethand.dmi'
 	var/print_type = "black leather fibers"
 	var/hide_prints = TRUE
-	var/print_difficulty = 0
+	var/print_difficulty = 60
 
 	insulated
 		desc = "Tough rubber work gloves styled in a high-visibility yellow color. They are electrically insulated, and provide full protection against most shocks."
--- a/code/obj/item/clothing/chameleon_js.dm
+++ b/code/obj/item/clothing/chameleon_js.dm
@@ -1187,7 +1187,6 @@
 	var/current_choice = new/datum/chameleon_gloves_pattern
 	material_prints = "black leather fibers"
 	hide_prints = TRUE
-	scramble_prints = FALSE
 
 	New()
 		..()
@@ -1261,6 +1260,7 @@
 			src.wear_image = image(wear_image_icon)
 			src.inhand_image = image(inhand_image_icon)
 			src.material_prints = T.print_type
+			src.print_scramble_difficulty = T.print_difficulty
 			usr.set_clothing_icon_dirty()
 
 /datum/chameleon_gloves_pattern
@@ -1273,7 +1273,7 @@
 	var/sprite_hand = 'icons/mob/inhand/hand_feethand.dmi'
 	var/print_type = "black leather fibers"
 	var/hide_prints = TRUE
-	var/scramble_prints = FALSE
+	var/print_difficulty = 0
 
 	insulated
 		desc = "Tough rubber work gloves styled in a high-visibility yellow color. They are electrically insulated, and provide full protection against most shocks."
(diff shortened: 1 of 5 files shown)


MintyPhresh:
Seeing as insuls have much higher functionality than latexes as is, I'd personally like to see latexes be better at hiding fingerprints than insuls... Not sure how realistic that is, but I like there being a meaningful choice between the two

Bartimeus973:
Considering that latex gloves are purchasable at every single medvend, are plentiful in medbay, and can be found in many other places, i'd rather leave them as poor criming tools.

CheffieGithub:
theres an issue with partial prints where you can find people incredibly easily with just a few characters searched

Bartimeus973:
I am aware, however the gloves with the highest scrambling should require a few different prints to obtain a string complete enough to have secmate find the prints.

Bartimeus973:
Partial prints are only created once when the item is touched. Rescanning an item only gives the same print again. This is as it already is ingame.

Glamurio:
> Partial prints are only created once when the item is touched. Rescanning an item only gives the same print again. This is as it already is ingame.

This is likely outside the scope of this PR, but it would be nice for the sake of forensics if Forensic Scanners (or PDAs) were to keep track of scanned IDs. Ideally they'd sort it systematically if they find a semi-matching print on two gloves, but that may truly be too difficult to implement. The other thing, at least, saves around 10 seconds for every print we scan, where we have to copy paste the thing into a PDA or textfile, which can get very tedious.

Bartimeus973:
It seems that it is several people's shared feeling that forensics as a whole could use some more love. At this point i am considering reworking the entire thing, depending on what i can figure out and dev input.

The new implementation i am considering, without gutting the entire thing, would be the following:

All gloves would muddles prints by 90%. This is to make it very difficult to immediatly find out who left a print, but also to leave the option of brute forcing it for the people who like it, as well as to add an element of ineluctability since enough prints will eventually reconstitute a full print.

Additionally, instead of prints there could occasionally be "additional genetic material". These wouldn't give you the subject's DNA but instead give you hints towards who they could be. For instance lizards could leave scales or skeletons could leave bone dust so you could get a hint towards the person's mutantrace. Alternatively that genetic material could contain pigments that could hint towards a person's hair color. The idea being to give hints towards the culprit, but not give a straight answer.

Flaborized:
>All gloves would muddles prints by 90%. This is to make it very difficult to immediatly find out who left a print, but also to leave the option of brute forcing it for the people who like it, as well as to add an element of ineluctability since enough prints will eventually reconstitute a full print.

Without trying to sound too nitpicky about this, I think even four characters matching from a found print is like, very sufficient evidence for a match, since there's so many randomized characters in unique positions that the odds of getting even that few to match up incidentally is vanishingly rare. (In theory you could write external software to check copy/pasted print codes against each other, though I'm not really sure anybody would do this). My point being that you wouldn't actually have to use lots of prints to figure this stuff out usually.

But it still just like, in my opinion, adds a layer of 'sometimes I get to just do a boring easy puzzle and then I know who did the crime', just less common than this PR is proposing. It's basically the same concept so I feel the same critique applies. Other parts of the post are fine though and I don't really have much issue with them, little vague hints I think are very In The Spirit of fictional forensic studies, and would be Cool.

Bartimeus973:
I am intrigued to know what other opinions i could gather, considering that this PR got way more responses than i thought it would. I will make a forum post, gather information and ideas, and make a followup PR implementing those solutions.

PR #15273, opened 2023-08-06
Allow vending machines to be destroyed by force (& slightly more)

MeggalBozale:
[A-Game-Objects] [C-Balance] 
## About the PR 
Vending machines are currently immune to being completely destroyed, and immune to gunshots. This PR allows guns and weapons to destroy vendors. By default, all vendors have 400 health, which is the same as doors. This means you need to unload 7 shots from a predator revolver to do the deed. Pretty tanky!

Blobs also can damage vendors now, outside of the random chance shenanagains. This makes destroying vendors less RNG.
Lastly, hacking the AI wire on a vendor allows you to deconstruct it, even if you don't have access. It just makes sense.

## Why's this needed?
Nuke Ops don't have the time or the deconstruction devices to do things like this, and so breaking them down by brute force only makes sense. The current behavior is that vendors are outright immune to projectiles and such, the most that can be done is to tip them or explode them. This change gives Nuke Ops more control over their environment that they otherwise don't have.

My main concern is people destroying medical vendors, but they're easier to deconstruct if you can get your hands on the right tools (along with deconstruction being _slightly_ more discrete. Feedback on the difficulty of destroying certain vendors is wanted and welcome.

## Examples
Shooting with Predator Revolver
dreamseeker_v3LR6YD9rA

Manually deconstructing the vendor ~23s
ezgif-2-e4406cacb7

Destroying with a weldingtool ~30s
dreamseeker_1loQ20dmnV

## Changelog

```changelog
(u)MeggalBozale
(+)Vending machines can now be destroyed, with enough persistence.
```


MeggalBozale:
OK whoops different approach now after conversation in imcoder

MeggalBozale:
Outside of the name of this PR, but I think there's some other things that need adjusting too. Namely:

### Things I think should be shootable over
KitchenHelper
Microwave
Processor
Oven
Deep Fryer
Stove
Espresso Machine
Vending Machines (PR'd already)
Freezer
TEG
Furnace
Portable Air Scrubber
Air Pump
Portable Reclaimer
Artifact lab machines
Hydrophonics trays
Aquatic Research Pool
Material Processor
Nanofabs
Electroplater
Ruckingeur kit
monkeyplant
piano
loudspeaker
crematorium
coffins (act like lockers RN, but IMO too short for that to be reasonable)
morgue slabs
tray

### Things I think should be destroyable via shooting
Consoles (Gene, QM, etc.)
Cloning (iffy there ey)
SMES
Foamtank/Fueltank/etc (all fluid tanks)
Scanner (same method as printer)
Chem dispenser (REALLY iffy, could shoot over too but feels too tall)
Reagent heater
Fishing Supplies Vendor (tip over like vendor?)
Aquatic Research Upload Terminal (tip over like vendor?)
Arcade machine
Cryogenic Healing pod
Compact Sleeper
Cyborg Docking Station
Gene booth (tip over like vendor?)
GeneTek Scanner
Tank Storage Unit
Disposals Outlet
Records cabinets (or shoot past, but mmmm)
Water cooler (or shootable over, with first few shots knocking off the jug and then shootable over)

I got these objects by running around as a machine gunner on Atlas and shooting everything in sight. Foolproof.
Objects I think should be destroyable would need varying amounts of health, which would probably require discussion

code/modules/vending/vending.dm
```diff
@@ -277,18 +278,23 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item)
...
 				src.malfunction()
 			else
 				src.fall()
 		return
 
 	return
 
+/obj/machinery/vending/bullet_act(var/obj/projectile/P)
+	if((src.can_fall) && prob(P.power))
+		src.fall()
```
Tarmunora, line 295: should probably do checks on `ks_ratio` and/or `damage_type` - knocking over vending machines with a phaser would feel odd, and with a taser even odder

Tarmunora (changes requested):

code/modules/vending/vending.dm
MeggalBozale: Ah, gocha. Not too familar with how guns work - no problem restricting it to kinetics, nukies don't have energy weapons last I checked.

MeggalBozale: Made a change so that guns that deal kinetic damage count (such as macrophasers) since those make sense

Tarmunora: If filtering by damage_type, should probably include `D_PIERCING` and `D_SLASHING`, (though the latter is... uncommon) to cover the full gamut

MeggalBozale: Huh, didn't know they did those types of damage. Looking at the defines for projectile damage types I see:
#define D_KINETIC				(1<<0)
#define D_PIERCING				(1<<1)
#define D_SLASHING				(1<<2)
#define D_ENERGY				(1<<3)
#define D_BURNING				(1<<4)
#define D_RADIOACTIVE			(1<<5)
#define D_TOXIC					(1<<6)
#define D_SPECIAL				(1<<7)
D_KINETIC, D_PIERCING and D_SLASHING are the only ones of those that make sense yeah

I can't see a case where a weapon with that damage type stuns, even det revolver is D_ENERGY, so I'll add checks for only those two

PR #15282, opened 2023-08-07
Fix atlas toxins gas mixer

MeggalBozale:
[BUG] [MAPPING]

## About the PR
There's a rancher-arrivals frankenstein door that mysteriously appeared in place of a connector port and it really isn't supposed to be there

## Why's this needed?
Players should be allowed to use toxins on Atlas


ZeWaka:
you really need to clean up your master branch

PR #15296, opened 2023-08-08
remove duped camera on Cogmap1

### LINKED ISSUE #14801: Double camera outside mining
### Labels

[BUG] [map]

### Description

There should only be one camera outside mining

### Steps to reproduce

1. boot up cogmap
2. go to mining's catwalks
3. guh??? wuh??? huh???

### Expected Behavior

should only be one

image

TheTyrant25:
[MAPPING] [MINOR]
## About the PR
Two cameras existed in the same spot on the same turf on Cogmap 1 with different offsets. This removes the one floating in space.

## Why's this needed?
Fixes #14801

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -56263,12 +56263,6 @@
 	tag = ""
 	},
 /obj/grille/catwalk,
-/obj/machinery/camera{
-	c_tag = "autotag";
-	dir = 4;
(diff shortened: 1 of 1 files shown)


TheTyrant25:
image
oh and i rotated a plastic flap to actually match the direction of the conveyors
forgot about that

PR #15303, opened 2023-08-08
Makes banana peels fit in produce satchels

M-Earthfire:
[game-objects][feature][hydroponics]

## About the PR 

This adds banana peels into the list of items being able to be stuffed into produce satchels

## Why's this needed? 

Mass banana peels are a headache to clean for non-janitors and become annoying if they are not cleaned up. This places a tool to optimally clean them near the department that is most prone to create the mess in the first place. While this makes transporting big amounts of peels easier, cleaning them up with satchels is more effective since pulling into a satchels cleans a 3x3 area while emptying makes them concentrated in a single tile.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Banana peels fit in produce satchels now.
```


deathrobotpunch:
then we would have clowns dumping them in the middle of the hallway

deathrobotpunch:
they could also use it as an ammo pouch of mass slipping

Glamurio:
I'm not sure if I like produce satchels carrying anything other than produce. I'd rather see new QoL for the trash cart introduced that mimics the produce behavior, because that would allow general clean-up of a larger area and is more thematically fitting.

deathrobotpunch:
that’s a good idea

deathrobotpunch:
> Bananas are produce, though? So the peels are more ir less as well.
> 
> It's like feathers being part of chickens and thus fitting in produce satchels as well.
> 
> I like the idea to add the stuffibg functionality to chests/lockers/trash carts as well (they all share that behaviour iirc)

that’s a good idea

PR #15304, opened 2023-08-08
Adds basic critter humans

CheffieGithub:
## About the PR 
Adds mob critter basic humans. These have variables for a mob spawner and a human. The mob appearance is stolen from the human. If a mobspawner is set that is spawned on death, else the mob is gibbed.

Adds a syndicate parent mob and two example syndicate mobs: 
One with a knife
image

One with a Sirius rifle.
image

Adds a mob and corpse spawner for these.

Adds a variable for the hand holder datum for a dummy object which can be used to create inhand sprites for that hand holder, for example making a sword limb have a sword as an inhand.

Also organises the humanoid mobs into their own folder where they have their own files.

Known issue is that the two mobs (The critter and the corpse) will have differing appearances, I tried to find and elegant solution but copying the appearance holder didn't work and the bioholder copy needs the owner alive.

## Why's this needed? 
Sometimes you just need things simple and want a human without any of the extra faff. Currently critter AI is much easier to work with than human AI and trying to make it handle its inventory as a human would be very complex.


code/datums/critter_mobs/hands.dm
```diff
@@ -21,6 +21,9 @@
 	var/cooldown_overlay = 0
 	var/mob/holder = null
 
+	/// Object for inhand sets the inhands of that item, for limbs with weapons etc
+	var/object_for_inhand = null
```
TobleroneSwordfish, line 25: best make it clear this is a type not an instance (because we don't have a type for types)

code/mob/living/critter.dm
```diff
@@ -1033,8 +1034,16 @@ ADMIN_INTERACT_PROCS(/mob/living/critter, proc/modify_health)
 
 	update_inhands()
 		var/handcount = 0
-		for (var/datum/handHolder/HH in hands)
+		for (var/datum/handHolder/HH as anything in src.hands)
 			handcount++
+			if (HH.object_for_inhand)
+				var/obj/item/I = new HH.object_for_inhand
```
TobleroneSwordfish, line 1040: does this have to instantiate or can it use `initial`?

code/mob/living/critter/humanoid/human.dm
```diff
@@ -0,0 +1,127 @@
...
+		if (isnull(H))
+			return
+		var/mob/living/carbon/human/target = new H
+		if (target.l_hand) // don't want artifacts of papers / etc.
+			qdel(target.l_hand)
+		if (target.r_hand)
+			qdel(target.r_hand)
+		src.appearance = target
+		src.overlay_refs = target.overlay_refs?.Copy()
+		qdel(target)
```
TobleroneSwordfish, line 63: bear in mind this won't `qdel` because you're storing a ref here (also humans generally don't `qdel` 💀), might be fine for small amounts of these but you might need to consider using a typed cache of human instances if you plan to spawn a lot of these

code/mob/living/critter/humanoid/human.dm
```diff
@@ -0,0 +1,127 @@
...
+	health_brute = 50
+	health_brute_vuln = 1
+	health_burn = 50
+	health_burn_vuln = 1
+	is_npc = TRUE
+	ai_retaliates = TRUE
+	ai_retaliate_patience = 3
+	ai_retaliate_persistence = RETALIATE_UNTIL_INCAP
+	ai_type = /datum/aiHolder/wanderer
+	/// What do we spawn when we die should be a human corpse spawner leave null for gibs
```
TobleroneSwordfish, line 25: again, make it clear this is a type

TobleroneSwordfish (changes requested):

code/mob/living/critter.dm
CheffieGithub: seems like its always null on initial

PR #15308, opened 2023-08-09
Adds bottles of white wine and other juices to the bartender vendor

MintyPhresh:
[QoL] [Catering]
## About the PR
This PR adds bottles of white wine to the hidden section of the bartender vendor, as well as 2 boxes with a random assortment of fruit juices (apple, blackberry, blue raspberry, raspberry, capsaicin, grapefruit, peach, and strawberry juice) to the normal selection.

## Why's this needed?
Currently these juices are only available through botany, and only service between 1-3 possible mixes. This change allows a bartender to mix these drinks regardless of whether botany is around. It's hard to justify asking botany for 7 trays (3 of them mutations) for drink ingredients that might not even be needed, so this change allows bartenders to have a little more freedom in the drinks they offer. Notably, however, the selection of juices available is entirely random. It's not guaranteed that any juice will be there, and it might not be in the quantity wanted (each of the 7 bottles in a box has 30u of juice), so botany may still be required.

## Changelog
```changelog
(u)Mintyphresh
(+)Boxes of juices are now available in the bartending vendor. Order a Prairie Fire today! :)
```

### DIFF
--- a/code/modules/chemistry/tools/bottles.dm
+++ b/code/modules/chemistry/tools/bottles.dm
@@ -399,6 +399,23 @@
 	amount_per_transfer_from_this = 5
 	initial_reagents = "formaldehyde"
 
+/* ========================================================= */
+/* -------------------- Bartender Juice -------------------- */
+/* ========================================================= */
+/obj/item/reagent_containers/glass/bottle/juice
+	name = "juice bottle"
+	desc = "A small bottle for juice. There's a warning sticker about preservatives that's completely covering the label."
+	bottle_style = "2"
+	initial_volume = 35
+	amount_per_transfer_from_this = 5
+	initial_reagents = null
+	New()
+		src.initial_reagents = list(pick("juice_apple", "juice_blackberry", "juice_blueberry", "juice_blueraspberry", "capsaicin", "juice_grapefruit", "juice_peach", "juice_strawberry") = 30)
+		if (prob(50))
+			src.initial_reagents["formaldehyde"] = 5
+		else
+			src.initial_reagents["water"] = 5
+		..()
 /* ============================================== */
 /* -------------------- Misc -------------------- */
 /* ============================================== */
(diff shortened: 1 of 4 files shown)


deathrobotpunch:
good idea, alcohol vendor doesn’t have much purpose RN

Tarmunora:
Eh, not huge on cutting hydroponics out of the loop like this

MintyPhresh:
Would reducing the number of bottles available/increasing the number of juices possible help alleviate that? I did leave out juices that aren't parts of cocktail reagent recipes like carrot and watermelon

MintyPhresh:
I ran some numbers; On average, with the current setup of 14 30-unit bottles, a bartender will get around 45 units of a specific juice on average, which although not much for drinking on its own, is far too much when you consider that that would yield 135u of a 3-component mixed drink. I'm gonna preemptively nerf the capacity of the bottles down to 15, as well as add carrot, pumpkin, and watermelon juice which would make that average go down to 18u of juice. This does mean that there's only a 70% chance that a bartender will get ANY of a specific juice they're hoping for, but if they're truly desperate they can raid the space diner and increase their chance to 91%.

Tarmunora:
Going to go with a no on juice boxes in the vendor, but the white wine can stay. Could see adding juices to a "Mixed Mixers Crate" in QM perhaps though

TDHooligan:
seconding a mixed fruit juice crate from the QM, as it's great for chef as well as bartender. though i think youd be expecting 50u bottles at least, though the formaldehyde is a funny touch.
while on paper it's 'avoiding botany', fruit juice is just not going to happen more than 95% of the time. especially when the output is something as mundane as cocktails. 

getting fruit from botany requires botanists to plant fruits with a splash of potash, then give the results to chef (and is already given frustratingly poor odds)
getting juice from botany requires not only that, but they must also understand infusing & using their blessed saltpetre, then potash, for every fruit juice you want.

PR #15317, opened 2023-08-09
General Nettle Update and Overhaul

colossusqw:
[LABEL][hydroponics][feature]

## About the PR 
This PR does a few things:
- Changes nettles from firing off at random(fequency based on endurance) to only firing when someone moves to within their range(on a cooldown, reversely linear to endurance)
- Makes it so there is a Smooth Nettles mutation that occurs naturally, and stops nettle leaves from stinging on touch, or the standard production of histamine on nettles(useful for medicinal nettles or any nettle that doesn't intend on hurting you)
- Makes it so Botany jumpsuits counter nettle stings, to ensure botanists can work on them without getting pumped full of drugs/poison/medicine.

## Why's this needed? 

I've always wanted to give nettles this form of trigger, but couldn't make it work quite right until now, so i had borrowed rafflesia's method. It works well for raffles, but is somewhat inconsistent on nettles, and this way works much better.

The jumpsuit immunity was supposed to be for their aprons, but since that is locked behind botany xp, i've switched to the suits. This was done to give botanists a good way of actually planting and tending the nettles, since even for more dangerous plants such as rafflesias, they can easily get gas masks.

## Changelog 

```changelog
(u)Colossusw
(*)Overhauled Stinging Nettles slightly(minor changes for details).
(+)Nettles now won't pass reagents to people wearing botany jumpsuits.
(+)Nettles now have a 'smooth nettle' mutation that does not produce any histamine.
(+)Nettles now are triggered by movement near them, on a cooldown inversely scaled to their endurance, as opposed to firing off randomly.
```


colossusqw:
> I think the smooth nettles are missing a sprite. Maybe make them bloom with little flowers ehich are often used to differentiate between dead nettles and nettles?

I can definitely look into that, maybe even make the flowers harvestable in some other PR too..

CheffieGithub:
I think making the effect stopped by gloves (or specifically black ones) would make more sense also the mutation would be better as a chemical infusion one than a random one since having plants randomly mutate can get annoying.

colossusqw:
image
image
image

colossusqw:
!merge_upstream

PR #15319, opened 2023-08-09
Fixes people tied to constructed conveyor belts moving in wrong directions

TealSeer:
[BUG] [GAME OBJECTS]
## About the PR 

Fixes a bug wherein someone tied to a player constructed conveyor belt would move in the direction the conveyor belt was originally placed and not the one that was chosen by click-dragging. 

## Why's this needed? 

moustache-twirling villain antics are down 200%


deathrobotpunch:
i didn’t even know you could do that

PR #15324, opened 2023-08-10
Gives the HoP command teleporter access

### LINKED ISSUE #15165: Nadir: HOP doesn't appear to have access to command teleporter
### Labels

[BUG][MINOR]

### Description

HOP cannot enter command teleporter door. 

### Steps to reproduce

1. Begin round as Head of Personnel
2. Try walking into command teleporter room.
3. Door says no. 

### Expected Behavior

Given the HOPs command rank, I'd expect them to be able to enter the command teleporter.

Reported by: Zcrow
Active test merges: #15096
Round log date: 2023-07-29-15-16
Reported on: 2023-07-29 15:31:50
Map: NADIR

TheTyrant25:
[BUG]
## About the PR
Gives the HoP command teleporter access. Since they're part of the command structure.

## Why's this needed?
Seems like an oversight. Fixes #15165
If this isn't a bug, then, well, better close the issue as intended.

## Changelog
```changelog
(u)Tyrant
(+)The HoP now has access to the command teleporter by default.
```

### DIFF
--- a/code/procs/access.dm
+++ b/code/procs/access.dm
@@ -207,7 +207,7 @@
 						access_all_personal_lockers, access_tech_storage, access_maint_tunnels, access_bar, access_janitor,
 						access_crematorium, access_kitchen, access_robotics, access_cargo, access_supply_console,
 						access_research, access_hydro, access_ranch, access_mail, access_ai_upload, access_pathology, access_researchfoyer,
-						access_telesci)
+						access_telesci, access_teleporter)
 		if("Head of Security")
 #ifdef RP_MODE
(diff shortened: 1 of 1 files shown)


frawhst:
Imo it’s silly that teleporter access even exists when most maps seem to operate under the assumption that all heads can access it (which is basically the norm since oshan is the only map I can think of that actually puts teleporter access on the doors)

TheTyrant25:
nadir does as well i think. i can whip up a mapping pr to fix those command accesses later if need be but it probably should just match head access

PR #15342, opened 2023-08-11
Add wall busting shrapnel to the nuclear meltdown

amylizzle:
## About the PR 
Adds a blast of 5 to 20 radioactive shrapnel projectiles which can bust walls and piles of thindows.

## Why's this needed? 
The explosion of the reactor is turf safe so the fallout gas doesn't leak into space, but that also means that walls and windows aren't destroyed either so it doesn't spread. 
Now when the reactor blows, it will destroy a handful of immediately close walls and piles of thindows (which some engineers are using to deal with radiation). Additionally if a mob is hit with one of these projectiles instead, they get implanted with radioactive shrapnel, though I don't expect this to happen often since being in the room when the meltdown goes off is basically an instagib.

## Changelog 

```changelog
(u)Amylizzle
(+)The catastrophic nuclear meltdown is now slightly more catastrophic. Enjoy.
```

### DIFF
--- a/code/modules/projectiles/bullet.dm
+++ b/code/modules/projectiles/bullet.dm
@@ -1731,3 +1731,26 @@ datum/projectile/bullet/autocannon
 	shot_sound = null
 	projectile_speed = 12
 	implanted = null
+
+/datum/projectile/bullet/wall_buster_shrapnel // for nuclear meltdowns
+	name = "shrapnel"
+	damage = 70
+	damage_type = D_PIERCING
+	armor_ignored = 0.66
+	hit_type = DAMAGE_CUT
+	window_pass = 0
+	icon = 'icons/obj/scrap.dmi'
+	icon_state = "2metal0"
+	casing = null
+	impact_image_state = "bhole-staple"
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/bullet.dm
```diff
@@ -1731,3 +1731,26 @@ datum/projectile/bullet/autocannon
...
+	icon = 'icons/obj/scrap.dmi'
+	icon_state = "2metal0"
+	casing = null
+	impact_image_state = "bhole-staple"
+	implanted = /obj/item/implant/projectile/shrapnel/radioactive
+
+	on_hit(atom/hit, angle, obj/projectile/O)
+		if(istype(hit,  /turf/simulated/wall))
+			//let's pretend these walls were destroyed in the explosion
+			hit.ex_act(2)
```
TobleroneSwordfish, line 1751: maybe just call `ex_act` on whatever gets hit so they don't get blocked by random station objects?

TobleroneSwordfish (approved):

PR #15373, opened 2023-08-14
adds bacon grease & lard that can be ignited/eaten

klushy225:
[Catering]
[Chemistry]
[game objects]
[feature]
## About the PR 

This PR adds two chems, bacon grease and bacon lard. bacon grease and lard both metabolize as cholesterol, but the lard metabolizes into more cholesterol than the grease. You get bacon grease by putting bacon in a reagent extractor, and you get lard by cooling down the lard. The bacon grease can be ignited at 200C, being a little less powerful than welding fuel.

## Why's this needed? 

I think people should be able to drink bacon grease in space station 13, and I think that the clown should be able to fill balloons with boiling bacon grease that lights things on fire.

## Changelog 

```changelog
(u)Klushy225
(+)Added bacon lard and bacon grease found in bacon. Can be ignited
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -598,6 +598,37 @@
 				boutput(M, "<span class='notice'>A faint cheesy smell drifts through the air...</span>")
 			return
 
+
+	baconGrease
+		name = "Bacon Grease"
+		id = "lbacongrease"
+		result = "lbacongrease"
+		required_reagents = list("cbacongrease" = 1)
+		min_temperature = T0C + 35
+		result_amount = 1
+		mix_phrase = "The mass melts into a yellow oil."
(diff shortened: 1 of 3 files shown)


TheTyrant25:
probably reduce that changelog entry to one minor change

DisturbHerb:
why not add these behaviours to porktonium instead?

klushy225:
> why not add these behaviours to porktonium instead?

Because porktonium metabolizes into radium and cyanide. I don't want the grease to be a poison

Tarmunora:
eh, between being... not terribly enthused by the premise and this having a lot of copy-pasted code, going to go with no

klushy225:
> eh, between being... not terribly enthused by the premise and this having a lot of copy-pasted code, going to go with no

yeah that's fair. Was my first pr, and I didn't really expect it to get greenlighted. Still just trying to figure out how dm works

PR #15380, opened 2023-08-14
Kinetics improvements & magazines

TDHooligan:
[INPUT][OBJECTS][FEATURE]
## About the PR 

This draft PR was raised to discuss the core technical changes while spriting & testing is done, and is very WIP (to save there being dozens of projectiles in the review). If there's no major issues with the changes underlined below I'll move the rest of the PR up.

Guns have a new ammo storage solution, magazine objects:
- External magazines are built around the idea of swappable magazines.
- Current ammo objects are now scoped to being 'loose bullets', they work identically for guns that don't use external magazines (Shotguns, zipgun, revolver would remain unchanged)
- Ammo objects can be loaded into compatible magazine objects (while the mag isn't loaded into a gun)
- **Magazines actually swap properly, rather than magically refilling** 
- You can now have 'one in the chamber' in closed bolt kinetic weapons (so a magless gun can still fire bullets you find)

Guns can now override their projectiles' default firemode:
- Firemodes have been datumised, and are still set per-projectile (to remain compatible with existing non-firearm projectiles)
- **An optional override_firemode proc may be passed to the 'shoot' methods**
- Kinetic firearms have a default firemode override that determines burst size using bullet_number & cost (so burst fire works with 1-2 bullets in your mag)
- If your internal & external magazine don't match, kinetics now have a framework for handling more than 1 type of ammo in a burst 

## Why's this needed? 
Gannets thought itd be neat if mags properly swapped. 
Then I thought that having a bunch of magazines having 1~2 rounds in would be annoying, so fixed it.

It's also technically a lot closer to having Cooler ammo, like mixed ammo belts.

### DIFF
--- a/code/modules/projectiles/firemode.dm
+++ b/code/modules/projectiles/firemode.dm
@@ -1,0 +1,54 @@
+datum/firemode
+	var/shot_number = 0          // How many projectiles should be fired, each will cost the full cost
+	var/shoot_delay = 0.1 SECONDS          // Time between shots in a burst.
+	single
+		shot_number = 1
+
+	two_burst
+		shot_number = 2
+
+	three_burst
+		shot_number = 3
+
+	four_burst
+		shot_number = 4
+
+	five_burst
+		shot_number = 4
+
+	plasma_burst
+		shot_number = 4
+		shoot_delay = 1
+
+	akm
+		burst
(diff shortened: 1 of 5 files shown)


TDHooligan:
shelving this branch

PR #15381, opened 2023-08-14
Make click dragging an item onto a stored HUD equip it

FlameArrow57:
[GAME OBJECTS][QoL]

## About the PR 
This PR makes it so that click dragging an item stored in a storage HUD onto the HUD location where it's stored will equip it

## Why's this needed? 
When you are trying really fast to equip an item, it's common that instead of just clicking the item, you will click the item, hold left mouse button, and release when it's on the HUD where it's stored which won't equip it, making you think you never clicked it or some bug happened. This makes it so that doing so will actually equip it instead of doing nothing.

## Changelog 
```changelog
(u)FlameArrow57
(+)Improved item equip consistency when equipping an item from your inventory with fast mouse flicks.
```

### DIFF
--- a/code/datums/hud/storage.dm
+++ b/code/datums/hud/storage.dm
@@ -31,61 +31,16 @@
 	relay_click(id, mob/user, params)
 		switch (id)
 			if ("boxes")
-				if (params && islist(src.obj_locs))
-					var/list/prams = params
-					if (!islist(prams))
-						prams = params2list(prams)
-					if (islist(prams))
-						var/clicked_loc = prams["screen-loc"] // should be in the format 1:16,1:16 (tile x : pixel offset x, tile y : pixel offset y)
-						//DEBUG_MESSAGE(clicked_loc)
-						//var/regex/loc_regex = regex("(\\d*):\[^,]*,(\\d*):\[^\n]*")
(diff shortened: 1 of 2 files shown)


code/datums/hud/storage.dm
```diff
@@ -105,6 +60,70 @@
...
+			return
+		if (!(O in src.master.get_contents()))
+			return
+		var/clicked_position = src.get_clicked_position(user, params)
+		if (!clicked_position)
+			return
+		if (src.obj_locs[clicked_position] == O)
+			user.click(O, params2list(params))
+
+	proc/get_clicked_position(mob/user, list/params)
```
TobleroneSwordfish, line 74: doc comment pls

TobleroneSwordfish (changes requested):
may as well make it work for unequipping too for consistency

FlameArrow57:
Can you clarify? Unequipping is done with hotkey, or just click drag to a single inventory spot?

TobleroneSwordfish:
Click dragging to hand didn't seem to work for unequipping? I know it's specifically coded for backpacks but it makes sense to be general if we're doing this.

FlameArrow57:
I see what you're saying, but point of the PR was to just fix the one issue that's in the description, that's a different thing

PR #15391, opened 2023-08-15
Make guns that fit on your back store to open inventory first

FlameArrow57:
[GAME OBJECTS][QoL]

## About the PR 
This PR makes it so that if your backpack open, pressing the equip hotkey to equip weapons that fit on your back will attempt to add it to your inventory first, instead of swapping with whatever's being held on your back

Improvement over #13888

## Why's this needed? 
When you press the equip hotkey it makes more sense that something's added to open inventory

## Changelog 
```changelog
(u)FlameArrow57
(+)Pressing the equip hotkey while holding a gun that fits on your back will now attempt to add it to your backpack first if it's open.
```


NightmareChamillian:
Big fan of this.

PR #15395, opened 2023-08-15
Add slash wounds

FlameArrow57:

## About the PR 
This PR adds slash wound effects for certain weapons that appear when attacking with them. Mostly for high damage ones like swords, c-sabers, and chainsaws. The effect can be removed by brute healing, or instantly with synthflesh.

Ex.
image

## Why's this needed? 
Thought slash wounds would be cool, adds more variety in wound types

Affects forensics

## Changelog 

```changelog
(u)FlameArrow57
(+)High damage slashing/cutting weapons will now leave cut wounds.
```


Glamurio:
If it's not included, maybe werewolf claws could also cause wounds like that.

PR #15397, opened 2023-08-15
New sound for default laser.

Vispectral:
[Game Objects]

## About the PR 
Changes the default sounds of laser guns. Affects pods (Mk.II Scout), E-gun Lethal Mode, and Laser Guns found in safes.

## Why's this needed? 
Gives it a distinct sound instead of using the default taser sound. Helps differentiate between the two!

https://github.com/goonstation/goonstation/assets/137826296/4bd88cbc-fb62-4e40-b4a6-ef8f55e4ff8f


Tarmunora:
Some feedback from adminchat: sounds too big/punchy for handheld lasers

and speaking personally: I'd prefer to keep the abiguousness of not being able to tell by sound alone if an egun is shooting lethals or not

pali6:
To me this sound feels like it carries a lot of punch. I feel like this would fit for something more powerful / destructive.

Vispectral:
That's understandable :) I appreciate the feedback ! Mmm I don't know If I can apply it elsewhere. If I can I'll try.

PR #15401, opened 2023-08-16
Graphene reacts to completion

MintyPhresh:
[chemistry][bug]
## About the PR 
This PR changes the mechanics of the recipe for graphene and one of the e.coli recipes so that they react to completion. Reaction speed in-game is unchanged.

## Why's this needed? 
Due to the way the code works, if the reaction_speed and the result_amount of a chemical aren't equal to one another, it can cause the ratios to react in such a way that there's some leftover chemicals. (anyone who's tried to make graphene knows the pain of that last 0.5u of SiO2...)

This fixes that the easy way by changing the recipes to fit the code's quirks.

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -2101,10 +2101,10 @@
 		name = "Graphene"
 		id = "graphene"
 		result = "graphene"
-		required_reagents = list("fuel" = 4, "iron" = 1, "silicon_dioxide" = 1)
+		required_reagents = list("fuel" = 2, "iron" = 0.5, "silicon_dioxide" = 0.5)
 		min_temperature = T0C + 150
 		result_amount = 2
-		reaction_speed = 1
+		reaction_speed = 2
 		instant = 0
 		mix_phrase = "A small particulate forms into a tiny lattice."
 		on_reaction(var/datum/reagents/holder, var/created_volume)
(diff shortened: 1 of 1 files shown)


code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -2101,10 +2101,10 @@
 		name = "Graphene"
 		id = "graphene"
 		result = "graphene"
-		required_reagents = list("fuel" = 4, "iron" = 1, "silicon_dioxide" = 1)
+		required_reagents = list("fuel" = 2, "iron" = 0.5, "silicon_dioxide" = 0.5)
 		min_temperature = T0C + 150
 		result_amount = 2
-		reaction_speed = 1
+		reaction_speed = 2
```
Tarmunora, line 2107: will end up doubling the overall yield. Revise.

code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -3501,10 +3501,10 @@
 		name = "contaminated food E.Coli"
 		id = "contaminated food E.Coli"
 		result = "e.coli"
-		required_reagents = list("sewage" = 5, "cholesterol" = 4)
+		required_reagents = list("sewage" = 5, "cholesterol" = 1)
 		result_amount = 1
 		instant = 0 // bacteria needs some time to grow in this medium
-		reaction_speed = 0.25
+		reaction_speed = 1
```
Tarmunora, line 3507: messes with ratios, and if changing one recipe why would you not change the others

Tarmunora (changes requested):

MintyPhresh:
Ah, you're right-- so then this problem is a quirk exclusive to Graphene (I'll revert the changes to the e.coli recipe). Is it possible to actually do what you described? Even with a 40-unit cryostylane heat sink at 473K (any higher and the welding fuel explodes), the reaction temperature didn't stay high enough.

code/modules/chemistry/Chemistry-Recipes.dm
MintyPhresh: Reverted this change.

Issue #12927, opened 2023-01-29
Orange mop


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

When using the orange mop given to you by reaching level 15, the icon that appears above your head when you use it is the blue mop.

### Steps to reproduce

1. Get orange mop
2. Use it

### Expected Behavior

Image should be that of the orange mopwd

### Additional Information & Screenshots

https://imgur.com/a/DCiw5u5

Reported by: Straws
On server: Goonstation 1 Classic: Heisenbee
Round log date: 2023-01-29-19-12
Reported on: 2023-01-29 19:36:19
Map: OSHAN

PR #15402, opened 2023-08-16
Use orange mop icon overhead while mopping

gear-space:
[GAME OBJECTS][BUG]
## About the PR 
When using an orange mop from Janitor job rewards level 15, the mopping icon above the player's head is now orange. 
image

## Why's this needed? 
Experienced janitors should be able to flaunt their sanitation superiority more easily. Fixes [#12927](https://github.com/goonstation/goonstation/issues/12927)

### DIFF
--- a/code/datums/controllers/action_controls.dm
+++ b/code/datums/controllers/action_controls.dm
@@ -1732,6 +1732,8 @@ var/datum/action_controller/actions
 		mop = Mop
 		target = Target
 		duration = istype(target,/obj/fluid) ? 0 : 10
+		if (istype(mop, /obj/item/mop/orange))
+			icon_state = "mop_orange"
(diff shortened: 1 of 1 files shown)


code/datums/controllers/action_controls.dm
```diff
@@ -1732,6 +1732,8 @@ var/datum/action_controller/actions
 		mop = Mop
 		target = Target
 		duration = istype(target,/obj/fluid) ? 0 : 10
+		if (istype(mop, /obj/item/mop/orange))
+			icon_state = "mop_orange"
```
Tarmunora, line 1736: any reason not just use the mop's iconstate directly?

Tarmunora (changes requested):

PR #15406, opened 2023-08-16
Unary Atmos refactor

Laboredih123:
[ATMOS][INTERNAL][BUG]
## About the PR 
Cleans up code with absolute pathing, light DMDoc, some macros, and general cleanliness.
Repaths pressure tanks and the connector port to unary.
Side effect of the above means pressure tanks no longer leak when disconnected, but that was rarely noticed anyway.
Removes the tiny orange thing on freezers that is an ancient relic from the beginnings of cryo and aligns their sprites.
Fixes unary vents not showing them being on and siphoning.
Removes useless -f sprites from connector ports.
Removes Generator Input, which does absolutely nothing.
Injectors now properly update sprite layering.
Freezer no longer have hardcoded clamps on temperature, these have instead been moved into a var.
Heat exchangers now properly dispose of their buddy.
And i think that is all.

## Why's this needed? 
Better atmos code is better, from what I remember.


TobleroneSwordfish:
Okay I did my best to try to get this to merge but it's throwing all kinds of abstract type violation errors which means that somewhere in those 30k mapping changes a bunch of stuff has been reverted. Please remake this with the repathing separated from the bugfixes and cleanup so it stands a chance of being mergeable.

PR #15415, opened 2023-08-16
Normalize all inventory slots

Glamurio:
[Internal] [Code Quality] [Secret]

## About the PR 
Changes any and all hardcoded mentions of inventory slots such as `H.slot_head` to their equivalent `SLOT_HEAD` from the `clothing.dm` global define. Also removes the outdated local defines in `/mob/living/carbon/humans.dm`.

## Why's this needed? 
According to pali, it was a legacy thing, and it's much cleaner this way.


pali6 (approved):
I skimmed it at fast speed and saw nothing wrong

PR #15417, opened 2023-08-16
Makes a SICC core if no AI joins at roundstart

Paaiii:
## About the PR 
Replaces that horrible, mocking cardboard box that spawns in the core if there's no AI at roundstart with a empty latejoin AI core.

## Why's this needed? 
Because I think it's silly to completely remove AIs from the round if one doesn't roundstart and the MDir doesn't feel like giving out their manudrive.

## Changelog 

```changelog
(u)Paaiii
(+)Replaces that cardboard box that spawns in the AI core if there's no AI with an empty SICC AI core.
```

### DIFF
--- a/code/datums/gameticker.dm
+++ b/code/datums/gameticker.dm
@@ -275,7 +275,7 @@ var/global/current_state = GAME_STATE_WORLD_INIT
 
 	for(var/turf/T in job_start_locations["AI"])
 		if(isnull(locate(/mob/living/silicon/ai) in T))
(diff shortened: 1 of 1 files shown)


DisturbHerb:
Personally think we should come to a latejoin AI solution that doesn't kill the cardboard box. Its comedy value cannot be understated.

frawhst:
Going to make a thread because this might have Implications and I don’t want to miss anything

frawhst (approved):

PR #15422, opened 2023-08-17
Fix crystal wingrille spawns not being purple in map editor

TheTyrant25:
[MAPPING]
## About the PR
They were showing up as blue previously, due to erroneous installation of perspective windows in the map editors. This should allow crystal and reinforced crystal windows to appear purple again, as well as letting bohrum glass appear correctly (aka hardened windows).

## Why's this needed?
Bugs bad tm


TheTyrant25:
image
this fix definitely works btw

PR #15429, opened 2023-08-17
Adds clothing booth support to robots and allows robots to dress themselves

Glamurio:
[Feature] [QoL] [Clothing]

## About the PR 
This PR adds a new proc `/mob/living/silicon/proc/handle_clothing` which is used to dress up robots. It supports both the original method via recharging stations, as well as a method that's more similar to other humanoids, by simply dropping the clothing on the ground.

Robots can now dress themselves via `/datum/action/bar/icon/otherItem`, this will start a dress action that can be interrupted. This is done by dragging a clothing item onto the robot. Both the robot as well as other people can start this action.

Additionally, this PR adds support for robots to the clothing booth. This will display a generic robot that will display the clothing. If clothing items don't fit, they won't be displayed on the robot. Robots can wear heads, masks, uniforms and suits. They can still purchase other items, but they won't be able to wear them.

To facilitate silicon usage, money can now also be click and dragged onto the clothing booth to slot it in.

## Why's this needed? 
The idea that clothing booths can only be used by humans is probably antiquated by now, given that its display message reads "Human clothes don't fit you!". Many robots have been using regular human clothing for a while now, and usually have to ask a human to buy items from the clothing booth for them. This process is tedious and tricky, especially if you want to browse for the selection and do not have it memorized.

Additionally, it's a quality of life feature for robots to allow dressing themselves anywhere, even without a recharging station. As clothing is primarily used cosmetically, I don't see this as an issue, but just to alleviate any potential problems this requires an interruptible action that takes anywhere from 2-5 seconds depending on the item.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Robots can now use the clothing booth. Turns out human clothes DO fit!
(*)Robots can now dress and be dressed by dragging a piece of clothing onto them.
```

### DIFF
--- a/code/datums/character_preview.dm
+++ b/code/datums/character_preview.dm
@@ -83,7 +83,7 @@
 	. = ..()
 
 /// See: [/datum/movable_preview/var/custom_setup]
-/datum/movable_preview/proc/custom_setup(client/viewer, window_id, control_id)
+/datum/movable_preview/proc/custom_setup(client/viewer, window_id, control_id, is_robot = FALSE)
 	return
 
 /**
@@ -113,6 +113,11 @@
 	real_name = "character preview"
 	unobservable = TRUE
 
+/mob/living/silicon/robot/preview
+	name = "character preview"
+	real_name = "character preview"
+	unobservable = TRUE
+
 /**
  * # Character Preview
  *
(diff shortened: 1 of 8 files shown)


trollbreeder:
Does the clickdrag functionality for borg dressing work from inside peeked inventories? I.E. can you put on the janitor jumpsuit from inside a janitor equipment box, or a spacesuit from inside someone's backpack?

Tarmunora:
1. Clothing booth is not intended to be used by silicons
2. Cyborg customization being limited to docking stations is by design

PR #15430, opened 2023-08-17
Organise robotic critters into their own DMI files

CheffieGithub:
## About the PR 
Moves robotic critters out of mixed places in `icons/misc` and `icons/obj` into `icons/mob/critter/robotic`
Updates some icon_state names to be more clear, especially the drones.
Icons are divided into single DMIs where applicable to keep them small as possible. 
With folders setup for organisation and so you have descriptive icon paths like `robotic/drone/glitch.dmi` you know this will be the glitch drone and is robotic.

Though its not necessarily correct to have all this pathed under mob right now, the hope is that soon they all will be mobs.

## Why's this needed? 
Large DMIs are bad, smaller ones are better for performance and easier to find things in. icons should be in one place ideally.


code/obj/critter/automaton.dm
```diff
@@ -41,11 +42,6 @@ var/global/the_automaton = null
 			the_automaton = null
 		..()
 
-	disposing()
-		if (the_automaton == src)
-			the_automaton = null
-		..()
-
```
TobleroneSwordfish, line 48: ?

code/obj/critter/drone.dm
```diff
@@ -431,50 +431,56 @@ TYPEINFO(/obj/critter/gunbot/drone/helldrone)
...
-			return
+
+		applyDeathState()
+			overlays += image('icons/mob/critter/robotic/drone/overlays.dmi', "dying-overlay")
 
 	heavydrone
 		name = "Syndicate Hunter-Killer Drone"
 		desc = "A heavily-armed Syndicate hunter-killer drone."
-		icon_state = "drone2"
+		icon = 'icons/mob/critter/robotic/drone/disruptor.dmi'
```
TobleroneSwordfish, line 456: This seems excessive, some of these files have a single icon state in them.

TobleroneSwordfish (changes requested):
better late than never :)))

code/obj/critter/automaton.dm
CheffieGithub: it was duplicated

PR #15433, opened 2023-08-17
Fixes helmets that hide hair fucking up hair layering after unequip

TealSeer:
[BUG] [CLOTHING]
## About the PR 

Fixes a bug wherein equipping headgear that hid your hair (like a biohood) would screw up the layering on your sprite and make your hair appear underneath your backpack. update_hair_layer() checks src.head for a helmet to determine layering but on unequip src.head was nulled after the call to it instead of before.

Proper hair layering:
image

Bugged hair layering after equipping and removing bio hood:
image

## Why's this needed? 

Looks dumb and is wrong

### DIFF
--- a/code/mob/living/carbon/human.dm
+++ b/code/mob/living/carbon/human.dm
@@ -1703,8 +1703,8 @@
 		src.update_clothing()
 	else if (W == src.head)
 		W.unequipped(src)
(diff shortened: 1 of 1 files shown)


ZeWaka (approved):
and this is why functional programing is good

PR #15437, opened 2023-08-18
Auto cameras with offsets

TheTyrant25:
[MAPPING]
## About the PR
Introducing two new subtypes of camera: the auto camera and the alternate auto camera.
There are two subtypes because there are two different ways for the camera sprite to attach to a certain direction. The correct sprites should show up as 4 way sprites in the map editors.

These cameras, like fluorescent lights, automatically offset themselves to match with the perspective walling. This works for Jen walls too. If you have no var edited the dir and there is no wall to the south, it will attempt to find a wall in the north east and west to connect to.

_This does not replace or offset existing cameras within maps, this is a new subtype with new behaviour._

Actually fixing all cameras in all maps to use this is being atomised out to another PR.

## Why's this needed? 
Makes mapping more convenient and prevents more of the dreaded floor cameras.


CheffieGithub:
is the one deci second spawn needed? the helpers don't wait to spawn now its spawn(0) (which I don't even know if is needed)

CheffieGithub:
also autotag / auto name?

TheTyrant25:
> is the one deci second spawn needed? the helpers don't wait to spawn now its spawn(0) (which I don't even know if is needed)

i have no idea whether it is needed but i don't think there's any harm in playing it safe.

> also autotag / auto name?

good point

TheTyrant25:
had some broken behaviours in it but I have now tested it and confirmed it to work.
image

flappybatpal:
I'm confused, what are the purposes of the direction subtypes? I thought they would be to set a direction when you are surrounded by multiple walls but they all point north in testing.

TheTyrant25:
> I'm confused, what are the purposes of the direction subtypes? I thought they would be to set a direction when you are surrounded by multiple walls but they all point north in testing.

image
I can't seem to reproduce your issue?

CheffieGithub:
I still think it would be better to have just one auto camera that handles offsets like lights do, more convenient for mappers even if it looks worse.

TheTyrant25:
the parent does that.

CheffieGithub:
Then I don't really see a need for the directional subtypes in that case. 
I think there are accidental reverts on paper camera interactions?

code/obj/machinery/camera.dm
```diff
@@ -298,14 +408,14 @@
 		for(var/mob/O in mobs)
 			if (isAI(O))
 				boutput(O, "[user] holds a paper up to one of your cameras ...")
-				X.ui_interact(O)
+				O.Browse(text("<HTML><HEAD><TITLE>[]</TITLE></HEAD><BODY><TT>[]</TT></BODY></HTML>", X.name, X.info), text("window=[]", X.name))
```
CheffieGithub, line 411: Accident?

code/obj/machinery/camera.dm
```diff
@@ -298,14 +408,14 @@
...
 				logTheThing(LOG_STATION, user, "holds up a paper to a camera at [log_loc(src)], forcing [constructTarget(O,"station")] to read it. <b>Title:</b> [X.name]. <b>Text:</b> [adminscrub(X.info)]")
 			else
 				var/obj/machinery/computer/security/S = O.using_dialog_of_type(/obj/machinery/computer/security)
 				if (S)
 					if (S.current == src)
 						boutput(O, "[user] holds a paper up to one of the cameras ...")
-						X.ui_interact(O)
+						O.Browse(text("<HTML><HEAD><TITLE>[]</TITLE></HEAD><BODY><TT>[]</TT></BODY></HTML>", X.name, X.info), text("window=[]", X.name))
```
CheffieGithub, line 418: Same

TheTyrant25: looks like the result of a botched merge conflict fixing

TheTyrant25:
i prefer to give people options to have them placed correctly in the map editor, rather than having them all be auto.

CheffieGithub:
Im currently making a PR for camera network QoL including more networks and such so just one would be ideal for cases like auto/ (mining outpost) (science outpost) (ranch) (AI) etc

flappybatpal:
I think directions setting an alternative dir is a bit weird but I'm guessing this is to make sense with the map editor etc because they are on the inside of walls? Either way, seems fine atm.

PR #15448, opened 2023-08-20
Give Head Revolutionary headsets a syndicate channel

FlameArrow57:
[GAME OBJECTS][GAME MODES][FEATURE][BALANCE][QoL][EXPERIMENTAL]

## About the PR 
This PR gives head rev headsets a wiretap radio upgrade that lets themselves communicate with each other.

This can alternatively be used by the crew to listen in on what's going on once a headrev is killed, which may be bad, would need more feedback on that part. Maybe just destroying the wiretap when they die?

## Why's this needed? 
Helps for better planning with each other and coordination, rather than each head rev doing their own thing.

## Changelog 

```changelog
(u)FlameArrow57
(*)Head Rev radio headsets now start with a private comms channel for communication with each other.
```

### DIFF
--- a/code/modules/antagonists/revolutionary/head_revolutionary.dm
+++ b/code/modules/antagonists/revolutionary/head_revolutionary.dm
@@ -4,6 +4,10 @@
 	antagonist_icon = "rev_head"
 
 	var/static/list/datum/mind/heads_of_staff
+	/// headrev headset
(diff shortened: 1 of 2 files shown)


FlameArrow57:
Communication implant instead maybe, that's destroyed on removal?

PR #15458, opened 2023-08-20
Random gang member selection at roundstart

pali6:
## About the PR 
Turns on the experimental toggle that makes gang member selection random and disables recruitment. Essentially this acts as if every gang leader immediately recruited random people to their gang. Gang member names and jobs will be in chat once you spawn. Gang name is shown to you when the leader chooses it, similarly the locker location is shown to you when the leader places it.

For now this PR exists just as a temporary test merge to test the waters on the idea. I don't plan on merging it unless there's positive feedback on this.

## Why's this needed? 
Recruitment leads to weird meta strategies of picking only the most "robust" people on classic and to awkwardness of having to justify why your character would join a gang on RP. I'd like to try out this alternative idea and see how it holds up.

## Changelog 

```changelog
(u)pali
(*)Experiment: Gang members are recruited randomly at roundstart instead of being recruited manually.<br>Give feedback <a href="https://forum.ss13.co/showthread.php?tid=21701">here</a>.
```

### DIFF
--- a/code/datums/gamemodes/gangwar.dm
+++ b/code/datums/gamemodes/gangwar.dm
@@ -5,7 +5,7 @@
 	regular = FALSE
 
 	/// Makes it so gang members are chosen randomly at roundstart instead of being recruited.
-	var/random_gangs = FALSE
+	var/random_gangs = TRUE
 
 	antag_token_support = TRUE
 	var/list/gangs = list()


pali6:
Reminder to self: For the duration of this experiment gang mode weight in being picked has been increased in https://github.com/goonstation/goonstation-secret/commit/1b1631f55824f360e2ac60e331227026598ee352, revert that after this experiment ends in one direction or the other.

FlameArrow57:
Hard to tell for sure since end of round popups and gang reports in chat don't seem to work properly, but there was a gang round just a few minutes ago that I think only had one gang in it?

DisturbHerb:
lol. wut

PR #15469, opened 2023-08-21
Replace air injectors in corridors with vents.

TheTyrant25:
[MAPPING] [ATMOS]
## About the PR
Replaces the air injectors in the hallways of cogmap 1, kondaru, and wrestlemap to be unary air vents, like they are on cogmap 2. Other maps simply don't have hallway air vents.

Another change that i felt was too minor for a full pr makes underfloor manifolds appear visibly partially transparent in the map editor, like how underfloor simple pipes do. It's a tad more consistent and visually matches the pipes.

## Why's this needed?
According to @Laboredih123 on discord, the air injectors will always inject air, while the vents will stop injecting at 120kPa, which is better for not overpressurising. Plus, they're nicer looking.


TheTyrant25:
seeing as normal atmospheric pressure is 101kpa and they can go up to 120kpa, this won't hinder co2/plasma/n2o flooding that much, although it will get capped. Still yet to see how much of an impact this makes but I feel it's one of those changes that Should Have Been That Way Already, as it were.

TobleroneSwordfish:
Not sure this is actually a good thing, since it'll just make repressurizing using the vents even slower and leaving the canisters set up with the vents open still won't be a good idea because all the air will be lost trying to repressurize a breached hallway.

TheTyrant25:
@Laboredih123 suggested the change I just mapped it. @Laboredih123 ?

TheTyrant25:
im just gonna stop fixing these merge conflicts
in theory merge upstream when it's ready should work just fine

TobleroneSwordfish:
Closing as afaik the above issue still stands, ping me if I'm wrong.

Laboredih123:
the speed of these things have been doubled, hopefully that fixes the issue?

TheTyrant25:
should i reopen then?

PR #15475, opened 2023-08-21
Bundled all calls for `assoc_reagents` in plants into `proc/HYPget_assoc_reagents`

M-Earthfire:
[Hydroponics][Code-Quality]

## About the PR 

This PR adds `proc/HYPget_assoc_reagents`, which returns a list of all chems a plant can potentionally produce for use in produce/procs. This bundles the effect of all plant genes that change the list of chems a produce can receive, included the inert gene strain.

This proc is now called in any instance produce/plant effects now need a list of chems the plant can produce.

Above this, this PR adds a new gene strains, `/datum/plant_gene_strain/reagent_blacklist`. This gene strain removes 3 chems, which can be very dangerous when applied in combat and fast. These are ghost chilli juice and the alkali metals lithium and potassium.

Also, this PR fixes a bug where the inert-plant gene strain was not appearing on the wholetuna cordata plant on mutation.

## Why's this needed? 

To get the chems a plant can produce, there were multiple instances used which did not take plant genes in account or were kinda convulent. This cleans up the code and makes the behaviour consistent all around

The blacklist gene strain enables the addition of certain plants which would be very problematic when spliced in certain composition, like the maneater rework with ghost chilli juice. Also, it enables the addition of certain gene strains like a "cleaning" one, which removes chems like sewage or miasma out of an assoc reagent list. But i am spitballing ideas at that point.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Rafflesias and nettles work, as expected, with the "inert" plant gene strain now
```

### DIFF
--- a/code/WorkInProgress/AzrunStuff.dm
+++ b/code/WorkInProgress/AzrunStuff.dm
@@ -110,9 +110,9 @@
 			var/datum/plantgenes/DNA = POT.plantgenes
 			var/datum/reagents/reagents_temp = new/datum/reagents(max(1,(50 + DNA.cropsize))) // Creating a temporary chem holder
 			reagents_temp.my_atom = POT
-
-			for (var/plantReagent in assoc_reagents)
-				reagents_temp.add_reagent(plantReagent, 2 * round(max(1,(1 + DNA?.get_effective_value("potency") / (10 * length(assoc_reagents))))))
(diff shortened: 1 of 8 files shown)


Tarmunora (approved):

ZeWaka (approved):
this is good

PR #15476, opened 2023-08-21
Makes botanical mister be able to be emagged

M-Earthfire:
[Game-objects][Hydroponics][Machinery]

## About the PR 

Adds an effect for emagged botanical misters:

If emagged, the range of the botanical mister increases from 2 to 3 tiles. Beyond that, instead of only filling hydroponic trays, it begins filling every open container-object in it's area.

On examination, the mister give a hint on it being emagged.

## Why's this needed? 

The effect of the emagged mister is intended to enable some sabotage of departments like chemistry. Also, it is pretty unique and a natural expansion of its normal capabilities. 

Uses include:
Add this to the chemistry department active and see the scientist or bartender fuming as all their containers get filled with water or other funny chems you filled in the mister. Make a bar where each glass automatically gets refilled with beer. Drag it through medbay and top up each slightly used flask with ricin. Drag it through security and fill each coffee mug with tea.... you monster

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Emagging a botanical mister makes it fill beakers, glasses or other open containers in its vicinity
```


code/obj/machinery/plantpot.dm
```diff
@@ -1922,24 +1924,47 @@ TYPEINFO(/obj/machinery/hydro_mister)
...
-		return "<br>It's [!src.active ? "off" : (!src.mode ? "on low" : "on high")]. It's about [round(src.reagents.total_volume / src.reagents.maximum_volume * 100, 1)]% full. It seems to contain [reag_list]."
+		complete_description += " It seems to contain [reag_list]."
+		return complete_description
 
 	process()
 		..()
 		if(src.active)
-			for (var/obj/machinery/plantpot/P in view(2,src))
-				if(P.reagents.get_reagent_amount("water") >= 195)
+			for (var/obj/potentional_target in view(mist_range,src))
```
Tarmunora, line 1940: typo lol

Tarmunora (approved):
Code seems fine, concept is a bit meh

PR #15478, opened 2023-08-21
Refactor trashbags

flappybatpal:
## About the PR 
Removes `/obj/item/clothing/under/trash_bag` and replaces with `/obj/item/trash_bag/biohazard`.
Changes wearing the trash_bag into creating a trashsinglet with snipping tools.
Indirectly fixes the broken trash bag clothing icons.
The trashsinglet now functions as storage thanks to storage datums, converting a trashbag into this type will move over all it's contents.

## Why's this needed? 
We have two types of trash bag clothing.

## Changelog 

```changelog
(u)Flappybat
(+)Trash bags now must be converted into clothing using snipping tools instead of just wearing them directly. They will still function as clothing with storage afterwards.
```


flappybatpal:
There is at least one map that needs repathing in secret.

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,99 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+	New()
+		..()
+		src.create_storage(/datum/storage/no_hud, prevent_holding = list(/obj/item/trash_bag), max_wclass = W_CLASS_NORMAL, slots = 20,
+			params = list("use_inventory_counter" = TRUE, "variable_weight" = TRUE, "max_weight" = 20))
+
+	equipped(mob/user)
+		..()
+		for (var/i = 1 to round(length(src.storage.get_contents()) / 3))
+			src.remove_random_item(user)
+
```
FlameArrow57, line 1222:
```suggestion
```
Could probably remove this code? Was used for when equipping the trashbag on your back, but seems it can't be anymore

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,99 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+	attackby(obj/item/I, mob/user)
+		if (issnippingtool(I))
+			var/list/actions = list("Cut into an outfit", "Put it into the bag")
+			var/action = input(user, "What do you want to do with [src]?") as null|anything in actions
```
FlameArrow57, line 1226: tgui_input_list? Which also covers any no user cases, dunno if input() does too

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,99 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+						qdel(src)
+						return
+		..()
+
+	attack_hand(mob/user)
+		..()
+		var/mob/living/carbon/human/H = src.loc
+		if (istype(H) )
+			src.remove_random_item(H)
+
```
FlameArrow57, line 1256:
```suggestion
```
Used for when it was worn, otherwise this will make two items come out every time you click it with empty hand

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,99 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+	proc/remove_random_item(mob/user)
+		if (!length(src.storage.get_contents()))
+			return
+		var/obj/item/I = pick(src.storage.get_contents())
+		src.storage.transfer_stored_item(I, get_turf(src))
+		if (user)
+			user.visible_message("\An [I] falls out of [user]'s [src.name]!", "<span class='alert'>\An [I] falls out of your [src.name]!</span>")
+		else
+			src.loc.visible_message("\An [I] falls out of [src]!")
+
```
FlameArrow57, line 1287:
```suggestion
```
Doesn't seem it's used anymore

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,99 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+						boutput(user, "<span class='alert'>You were interrupted!</span>")
+						return
+					else
+						var/obj/item/clothing/under/shorts/trashsinglet/trash_outfit = new /obj/item/clothing/under/shorts/trashsinglet(get_turf(src))
+						playsound(src.loc, 'sound/items/Scissor.ogg', 100, 1)
+
+						user.u_equip(src)
+						if (length(src.storage.get_contents()))
+							for(var/obj/item/contents in src.storage.get_contents())
+								src.storage.transfer_stored_item(contents, get_turf(src))
```
FlameArrow57, line 1242:
```suggestion
							for(var/obj/item/I as anything in src.storage.get_contents())
								src.storage.transfer_stored_item(I, get_turf(src))
```

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,99 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+
+
+/obj/item/trash_bag
+	name = "trash bag"
+	desc = "A flimsy bag for filling with things that are no longer wanted."
+	icon = 'icons/obj/janitor.dmi'
+	inhand_image_icon = 'icons/mob/inhand/hand_tools.dmi'
+	icon_state = "trashbag-f"
+	item_state = "trashbag-f"
+	uses_multiple_icon_states = 1
```
FlameArrow57, line 1206:
```suggestion
	uses_multiple_icon_states = TRUE
```

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,99 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+	w_class = W_CLASS_TINY
+	rand_pos = 1
```
FlameArrow57, line 1208:
```suggestion
	rand_pos = TRUE
```

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,99 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+		for (var/i = 1 to round(length(src.storage.get_contents()) / 3))
+			src.remove_random_item(user)
+
+	attackby(obj/item/I, mob/user)
+		if (issnippingtool(I))
+			var/list/actions = list("Cut into an outfit", "Put it into the bag")
+			var/action = input(user, "What do you want to do with [src]?") as null|anything in actions
+			if (!action)
+				return
+			switch (action)
```
FlameArrow57, line 1229: Could probably do without the switch?

code/obj/item/mops_cleaners.dm
flappybatpal: I was 100% sure I removed and committed this as you say it's no longer relevant.

code/obj/item/mops_cleaners.dm
flappybatpal: Not sure what is the best approach.

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,72 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+	var/clothing_type = /obj/item/clothing/under/gimmick/trashsinglet
+
+	New()
+		..()
+		src.create_storage(/datum/storage/no_hud, prevent_holding = list(/obj/item/trash_bag), max_wclass = W_CLASS_NORMAL, slots = 20,
+			params = list("use_inventory_counter" = TRUE, "variable_weight" = TRUE, "max_weight" = 20))
+
+	attackby(obj/item/I, mob/user)
+		if (issnippingtool(I))
+			var/action = tgui_input_list(usr, "What do you want to do with [src]?", "Trash Bag", list("Cut into an outfit", "Add to contents"))
```
FlameArrow57, line 1221:
```suggestion
			var/action = tgui_input_list(user, "What do you want to do with [src]?", "Trash Bag", list("Cut into an outfit", "Add to contents"))
```

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,72 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+			if (action == "Cut into an outfit")
```
FlameArrow57, line 1222:
```suggestion
			if (!action)
				return
			if (action == "Cut into an outfit")
```
From cancel option

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,72 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+				if (!do_after(user, 3 SECONDS))
+					boutput(user, "<span class='alert'>You were interrupted!</span>")
+					return
+				else
+					var/obj/item/clothing/under/gimmick/trashsinglet/trash_outfit = new clothing_type(get_turf(src))
+					playsound(src.loc, 'sound/items/Scissor.ogg', 100, 1)
+					user.u_equip(src)
+					if (length(src.storage.get_contents()))
+						for(var/obj/item/contents as anything in src.storage.get_contents())
+							src.storage.transfer_stored_item(contents, trash_outfit, TRUE)
```
FlameArrow57, line 1233:
```suggestion
					for(var/obj/item/contents as anything in src.storage.get_contents())
						src.storage.transfer_stored_item(contents, trash_outfit, TRUE)
```

code/obj/item/mops_cleaners.dm
```diff
@@ -1192,3 +1194,72 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+					usr.put_in_hand_or_drop(trash_outfit)
+					qdel(src)
+					return
+		..()
+
+	update_icon(mob/user)
+		if (!src.storage || !length(src.storage.get_contents()))
+			src.icon_state = initial(src.icon_state)
+			src.item_state = ""
```
FlameArrow57, line 1242: "" or initial(src.item_state)? Currently initialized as something other than ""

code/obj/item/mops_cleaners.dm
flappybatpal: Whoops, I forgot cancel!

code/obj/item/mops_cleaners.dm
flappybatpal: TBH I have no idea what this is for and why you'd want the item state to be blank, it was in the original code.

PR #15482, opened 2023-08-22
Give bruise packs and ointment a 2 second action bar and removes a certain dusty first aid kit from debris

unfunnyperson:
## About the PR 
Bruise packs and ointment will have a 2 second action bar when using them.

Also removes a certain dusty first aid kit from debris (Was located in the derelict ai satellite)
image

## Why's this needed? 
Both of these items are stupidly overpowered in terms of healing(60 brute/40 burn a second). Especially during combat encounters when most other healing items dont work. For example someone can take 3 csaber hits and heal to full health in 3 seconds.

This keeps them as a really good healing item outside of combat but makes them useless in combat.

I removed the debris kit because its laughably easy to rush and other easily rushable dusty kits got removed(recent oshan/nadir change)

## Changelog

```changelog
(u)UnfunnyPerson
(+)Bruise packs and Ointment take 2 seconds to use instead of being instant.
(+)Removed a certain dusty first aid kit from the debris field.
```
[BALANCE]

### DIFF
--- a/code/obj/item/bruisepack_ointment.dm
+++ b/code/obj/item/bruisepack_ointment.dm
@@ -85,7 +85,6 @@
 			qdel(src)
 		src.inventory_counter?.update_number(src.amount)
 
-
 /obj/item/medical/bruise_pack
 	name = "bruise pack"
 	desc = "A pack designed to treat blunt-force trauma."
--- a/code/obj/item/bruisepack_ointment.dm
+++ b/code/obj/item/bruisepack_ointment.dm
@@ -67,7 +67,7 @@
 			var/self = FALSE
 			if (M != user)
 				self = TRUE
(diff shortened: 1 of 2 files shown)


Sbmhawk:
They are already quite rare, and 3 seconds seem a little overkill. All nerfs recently have been "action bar added" 💀

MeggalBozale:
You can get them fairly trivially from static locations in the debris field. Loot like this should not be so easy to obtain if it's to be so powerful. Currently, it beats any other healing method, and that's just ridiculous. I don't think anyone would use this seriously besides powergamers.

If it's not to have an actionbar, they should be loot in the same way laser guns are.

Azrun (approved):
Action bars! I love action bars!

PR #15487, opened 2023-08-22
adds a spawnable light borg and screenhead borg

Retrino:
## About the PR 

adds a spawnable borg subtype with full light parts, and another with a screenhead, treads, and standard chest/arms

## Why's this needed? 

requested by angel. no longer do you have to build an entire light borg or screenhead borg just to debug

### DIFF
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -3225,6 +3225,20 @@
 			src.shell = 1
 			..(loc, frame, starter, syndie, frame_emagged)
 
+/mob/living/silicon/robot/spawnable/light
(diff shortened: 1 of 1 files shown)


TheTyrant25:
while you're there, could you also possibly make a heavy borg subtype? just cover all our bases

Retrino:
> while you're there, could you also possibly make a heavy borg subtype? just cover all our bases

there is a heavy subtype, do you mean one without treads?

TheTyrant25:
oops didnt know there was one just from checking the diff mb

TheTyrant25 (approved):
lgtm

TobleroneSwordfish (changes requested):
Code looks good, but you're using double space indentation instead of tabs, pls fix

Retrino:
so long gay bowser

PR #15490, opened 2023-08-22
Refactors Parallax To Use Render Sources

Mister-Moriarty:
[UI] [Experimental] [Rework]

## About The PR:
As it stands currently, for a parallax layer type in a reference frame, a z-level or an area, each client possesses their own parallax layer created from that type, with each layer instance being responsible for its own appearance and parallax attributes. This PR defers a large portion of the attributes stored by parallax layers to parallax render source screens: for a parallax layer in a given reference frame a single render source will handle all appearance and parallax data for every parallax layer created from it and displayed to a client.

In practice, this means that a single change to the parallax render source will take effect across every parallax layer created using that render source, without the necessity to loop through each subscribed client. An immediate benefit of this new implementation is the expedition of the process of recolouring parallax layers and animating transitions.

The primary benefit of such a refactor is the ability for parallax render sources to maintain a constant mutable state, in which changes to the render source will persist across clients leaving and reentering the reference frame. This is not possible with the current parallax system, as parallax layers are referred by type to the parallax controller to instantiate; it is only achieved on a small scale by use of list associations on the area parallax layers list.

Implementing complete support for persistent mutability of render sources has necessitated a refactor of the current rather disparate way of referring layer types to the parallax controller, with z-levels and areas using different methods to pass layer types to the controller. Z-levels, areas, and planet areas now utilise parallax layer render source groups, which govern which parallax render sources should be shown to a client, and have inherited the formerly global procs to add, remove, reset, and recolour parallax layers, permitting a more granular control over which parallax layers should be displayed where. In addition to this, default z-level parallax layers, which were handled by a global list, are now dictated by lists on map settings datums, which are passed to parallax render source groups corresponding to each z-level.

The code is heavily documented, and further information can be found on the new systems and types expressed above in their respective files.

This PR is likely to affect parallax layers on secret adventure zones. Such conflicts will be relatively easy to spot, as affected variables will have been renamed or outright removed.

There should not be any performance cost to this refactor, however I would recommend a test merge prior to a full merge to remediate any bugs created as a result of the refactor.

## Why Is This Needed?
I've made several points above explaining why this is needed, however it will also allow for the creation of a centralised admin panel for parallax settings across all z-levels, areas, and Artemis project planets.


code/modules/parallax/parallax_render_sources/parallax_render_source_group_parent.dm
```diff
@@ -0,0 +1,147 @@
...
+/**
+ *	The parent type for parallax render source groups, which govern which parallax render sources should be shown to a client
+ *	in a given area or z-level. They also allow for the efficient manipulation of parallax render sources and layers within a
+ *	specified z-level or area, permitting more granular control over the appearance of parallax render sources.
+ */
+/datum/parallax_render_source_group
+	/// An associative list of parallax render source types and the instance of that type belonging to this render source group.
+	var/list/atom/movable/screen/parallax_render_source/parallax_render_source_types_and_sources
+	/// A list of default parallax layer render source types that should be instantiated on New().
+	var/list/parallax_render_source_types = list()
```
ZeWaka, line 11: do this in New also

code/modules/parallax/parallax_controller.dm
```diff
@@ -187,8 +159,8 @@ var/global/parallax_enabled = TRUE
 /mob/proc/update_parallax(datum/component/component, turf/old_turf, turf/new_turf)
 	src.client?.parallax_controller?.update_parallax_layers(old_turf, new_turf)
 
-/mob/proc/update_parallax_z()
-	src.client?.parallax_controller?.update_parallax_z()
+/mob/proc/update_parallax_z(datum/component/component, old_z_level, new_z_level)
```
ZeWaka, line 162: any arbitrary component?

pali6:
Core idea sounds good, in fact I was wondering why it was not coded this way thinking I missed something. I applied a test merge to catch possible bugs. Will look through code itself later (unless someone else does, feel free to merge without me if confident in the PR).

code/modules/parallax/parallax_render_sources/parallax_render_source_group_parent.dm
Mister-Moriarty: `parallax_render_source_types` is the list populated with, as one could guess, default parallax render sources in the various subtypes of `/datum/parallax_render_source_group` - setting it to `list()` in `New()` would erase those default render sources.

code/modules/parallax/parallax_controller.dm
Mister-Moriarty: No, specifically `/datum/component/complexsignal/outermost_movable`, however the `component` argument here will never be used be parallax code, however it will still be passed by the complex signal, therefore we must accommodate for it.

Tarmunora:
areas in secrets will need compatibility fixes

PR #15492, opened 2023-08-23
Adds the missing clown hologram module to Nadir

cheekybrdy:
## About the PR 
Adds a clown hologram module to nadir, which is currently missing.

## Why's this needed? 
Standardised map items good


maps/nadir.dmm
```diff
@@ -19360,6 +19364,11 @@
 	dir = 4
 	},
 /area/station/security/main)
+"iqn" = (
+/turf/space,
+/area/station/hallway/secondary/construction{
+	name = "The Warrens"
+	})
```
TobleroneSwordfish, line 19371: this is a hole into space, I feel like it shouldn't be here

TobleroneSwordfish (changes requested):

maps/nadir.dmm
Azrun: image
Leaky station.

cheekybrdy:
lgtm

Azrun (approved):

PR #15493, opened 2023-08-23
Adds the ability to repair damaged windows with sheets of the same material

unfunnyperson:
[GAME OBJECTS] [PLAYER ACTIONS]

## About the PR 
Adds the ability to repair damaged windows by clicking on them with sheets of the same material this takes 4 seconds but can be reduced to 2 if the user has carpenter or engi training

## Why's this needed? 
The current ways to repair a damaged window are either silicate(which ruins the color of the window) or tearing down and replacing the window(griefs the atmosphere).

Adding a non-intrusive way to repair station windows would make repairs a lot less tedious to do.

## Changelog

```changelog
(u)UnfunnyPerson
(+)Clicking a damaged window with a sheets of the same material will repair the window
```

### DIFF
--- a/code/obj/window.dm
+++ b/code/obj/window.dm
@@ -464,6 +464,8 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
 
 		else if (iswrenchingtool(W) && src.state == 0 && !src.anchored)
 			actions.start(new /datum/action/bar/icon/deconstruct_window(src, W), user)
+		else if (src.health < src.health_max && istype(W, /obj/item/sheet) && W.material == src.material)
+			SETUP_GENERIC_ACTIONBAR(user, src, 1 SECOND, /obj/window/proc/fix_window, list(W), null, null, "<span class='notice'> [user] repairs \the [src] with \the [W] </span>", null)
 		else
 			attack_particle(user,src)
 			playsound(src.loc, src.hitsound , 75, 1)
@@ -567,6 +569,12 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
 			T.selftilenotify() //for fluids
 
 		return 1
+
+	proc/fix_window(var/obj/item/sheet/S)
+		health = health_max
+		UpdateIcon(src)
+		if (S)
+			S.change_stack_amount(-1)
 
 
 /datum/action/bar/icon/deconstruct_window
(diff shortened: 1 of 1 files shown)


Tarmunora (approved):
ah yes let me just glue another pane of glass on the window. That ought to fix it

code/obj/window.dm
```diff
@@ -464,6 +464,8 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
 
 		else if (iswrenchingtool(W) && src.state == 0 && !src.anchored)
 			actions.start(new /datum/action/bar/icon/deconstruct_window(src, W), user)
+		else if (src.health < src.health_max && istype(W, /obj/item/sheet) && W.material == src.material)
```
TobleroneSwordfish, line 467: this should use `isSameMaterial` instead of direct reference equality

TobleroneSwordfish (changes requested):

Azrun:
I think this really interferes with the function of `silicate` for something that seems be described as a vanity QoL.  I'd prefer the actionbar duration was longer but also factors in `hasTrait("carpenter")` and `hasTrait("training_engineer")`.

MeggalBozale:
I spun up a bit of discussion in imcoder, perhaps a weldingtool would make more sense for this situation.
[Link](https://discord.com/channels/182249960895545344/890226559691157524/1144659754547937291) to the discussion so you can take a look

Azrun:
I think I prefer the idea of the small silicate dispenser that can be refilled and improve connection to chemistry.  Use of welder to reduce or minimize visual effect of silicate might be cool as well.

NightmareChamillian:
Big disagree on @Azrun 's silicate idea. Engineers already have far, far too many things they have to carry, and adding another is just going to result in it going unused.

A welding tool or, by going the C:DDA route, soldering iron could work, but I also think the use of glass sheets is fine. It fits in with some of the more cartoony things that can happen, and makes sense, given the fact that engineers are likely to already be carrying them in the event that windows are broken/damaged.

Azrun:
I meant that I preferred the idea discussed in imcoder, wasn't my idea.  I don't think engineers *have* to carry around everything.  They could be a tool that is available. That is what makes those carts handy.

I like the idea of slapping the glass on top better than this, I feel like glass is a more limited resource.  My understanding is that's isn't even a very feasible thing to do and could damage the glass further.

Azrun (approved):

PR #15495, opened 2023-08-23
Tweak AI shell naming behavior slightly

Retrino:
## About the PR 

Basically, adds the variable "chosename" to determine if an AI has changed the name of a shell within a charging dock, and uses said variable to keep that name consistent when deploying and undeploying. Also HUGE thanks to callow for helping me figure out how to code this

## Why's this needed? 

Allows for easier organization of AI shells. Trying to figure out whether shell 89796 or shell 12415 is the one you want to deploy to is kind of annoying, be it because of specific modules in one vs the other, or in a case where 2 or more AIs exist and they want to keep track of which shell is theirs.

Screenshot (263)

## Changelog 

```changelog
(u)Retrino
(+)AIs can now name their shells to better keep track of them.
```

### DIFF
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -71,6 +71,7 @@
 	var/alohamaton_skin = 0 // for the bank purchase
 	var/metalman_skin = 0	//mbc : i'm getting tired of copypasting this, i promise to fix this somehow next time i add a cyborg skin ok
 	var/glitchy_speak = 0
+	var/chosename = 0
 
 	sound_fart = 'sound/voice/farts/poo2_robot.ogg'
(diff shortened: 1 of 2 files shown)


Retrino:
> This removes the ability to tell at a glance (or possibly at all?) whether a borg is an AI shell…

i dont entirely disagree with this point, but i feel like i should note that ais can already rename their shells, it just goes away on logout. personally, id rather go the route of adding something to the examine text of an ai shell that it *is* one, other than the fact that ai shells examine as “completely dormant” when inhabited which is more of a bug than a feature of spotting them

TobleroneSwordfish:
I'm going to say them being able to rename temporarily is a bug, so I've fixed it for now along with the description issue.

Retrino:
going to workshop this idea more and potentially come back to it in the future

PR #15496, opened 2023-08-23
TGUI tg Controls Info Window

Garash2k:
## About the PR 
Converts the tg Controls Info popup into TGUI
image

Under the hood, adds a `content_window` parameter to `tgui_alert` which, if set, will lookup for a part of the same name in the `AlertContentWindows` folder. This should be useful for antag popups! When making new parts, besides the part itself, all that needs to be done is adding an `import` and a `switch-case` in `AlertContentWindows/index.ts`.

## Why's this needed? 
Having one of the first popup someone sees upon connecting look modern = good

[UI][Rework]


tgui/packages/tgui/interfaces/AlertContentWindows/tgControls.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * @copyright 2023
+ * @author garash2k
+ * @license ISC
+ */
+import { AlertContentWindow } from './types';
+import { Box, Image } from '../../components';
+
+export const tgControls: AlertContentWindow = {
+  width: 470,
+  height: 317,
```
ZeWaka, line 12:
```suggestion
  width: 470,
  height: 320,
```
cleaner

code/modules/tgui/tgui_alert.dm
```diff
@@ -33,8 +35,10 @@
 	if(length(items) > 3)
 		log_tgui(user, "Error: TGUI Alert initiated with too many items. Use a list.", "TguiAlert")
 		return tgui_input_list(user, message, title, items, timeout, autofocus)
-	var/datum/tgui_modal/alert = new(user, message, title, items, timeout, autofocus)
+	var/datum/tgui_modal/alert = new(user, message, title, items, timeout, autofocus, content_window)
 	alert.ui_interact(user)
+	if (!do_wait)
+		return
```
ZeWaka, line 41: cleanup?

code/client.dm
```diff
@@ -402,7 +402,7 @@
 #ifndef IM_TESTING_SHIT_STOP_BARFING_CHANGELOGS_AT_ME
 				//preferences.randomizeLook()
 				preferences.ShowChoices(src.mob)
-				src.mob.Browse(grabResource("html/tgControls.html"),"window=tgcontrolsinfo;size=600x400;title=TG Controls Help")
+				tgui_alert(src, content_window = "tgControls", title = "Use /tg/ style interface?", do_wait = FALSE)
```
ZeWaka, line 405: why is title not in the tgControls also

code/modules/tgui/tgui_alert.dm
Garash2k: Ref stops existing after closing the popup
image

Garash2k:
Recent thought: if we think that piggy backing off `AlertModal.tsx`+`tgui_alert.dm` to make this sort of AlertContentWindow system is less ideal than having it be a seperate tgui export from a separate DM file, I could make that change. At first I just liked getting to re-use stuff like the ok buttons and the onKeyDown event, but I can see the upside of not messing with the original alert too.

PR #14516, opened 2023-06-10
Mobification and rework of the man-eating plant

M-Earthfire:
[AI][Critters][Hydroponics][Balance][Rework][Input-Wanted]

## About the PR 

This PR fundamentally reworks the maneater in a way that stays true to its former role and theme but removes out a bunch of jank and bullshit out of its behaviour and mechanics.

This fixes #13001 

Beyond that, this PR also makes botany produce now support mob critters. They will get handled slightly different than items, but name-changes and quality status will be reflected properly. A new parent for botany-critter was created, `/mob/living/critter/plant`. A mob critter from this parent that is harvested calls `proc/HYPsetup_DNA`, enabling easy modifications from within the object. Also, it carries the contributors to that originating plant within `growers` for easy access within the AI-behaviour. The implementation and passing of the respectable plants gene also enables any critter of this parent to be scanned by a plantscanner and give out the respectable stats of the plant they originated from. 

The maneater was changed drastically in it the way how it attacks and eats people:

https://github.com/goonstation/goonstation/assets/109365375/2fa0fd80-2523-4a3f-93b5-536be9c78905

Instead of insta-stunning their victim and having an invisible action that gibs the person, the maneater now need to drain their victims stamina and eat them with a visible action bar while their victim is in an aggressive grab.

The maneater's attack deals 8 - 12 brute damage to the target and drains 50 stamina while imposing a 2 second desorient. If the target gets their stamina drained completely, the target gets stunned for 5 seconds. The stun cannot be refreshed or prolonged by the maneaters attack and borgs are immune to the desorient.

If stunned and the devour ability is avaible, the maneater grabs the target and upgrades the grab. Once that is completed, it uses its ability to devour the target. **This is an 8 second action bar that cannot be interrupted by moving the maneater.** The victim needs to struggle out of the grab or needs to be grabbed out by another person to cancel the ablity. The devour ability got a 20 second cooldown.

The maneater won't try to eat  while another maneater is in direct contact to the target.

Once the target is devoured, all of their equipable items are moved into the maneater. If there are more than 6 of them in the maneater, it will try to vomit out an item each minute. If the maneater is gibbed or butchered, it will eject the items as well.

Also, the maneater, if no mobs, corpses or humans are avaible, will try to eat raw slabs of meat it can find. This can lead to some interesting use-cases:

https://github.com/goonstation/goonstation/assets/109365375/570e1616-afef-4732-9f11-feddb4762fde

[...]

PR #15503, opened 2023-08-23
Moves ricin and ghost chilli juice to be brewing results instead of inherent reagents

TobleroneSwordfish:
[BALANCE] [HYDROPONICS]
## About the PR 
Ricin rice and ghost chilli peppers no longer contain their poisons directly, instead needing to be refined in a still to get it.

## Why's this needed? 
The existence of extremely deadly and stealthy spliceable poisons severely limits botany design space with regards to things that inject reagents. I think it'd be cool if we could have more interesting ways to apply reagents without giving rise to infinite ricin death traps.
See #15317 and #14516 for examples of features that are hampered by the existence of these spliceable poisons.
## Changelog 

```changelog
(u)LeahTheTech
(*)Ricin and ghost chilli juice reagents are no longer present directly in their respective crops, instead requiring them to be brewed in a still.
```


Glamurio:
Good idea, might I suggest cross-referencing it with #14794 because it contains a still rework that includes `brew_amount` which could be used to further refine the amount of reagents produced via still.

TDHooligan:
i think a visual effect for these reagents would solve the issue nicely, and make uber-spliced plants even more  awful looking

colossusqw:
> i think a visual effect for these reagents would solve the issue nicely, and make uber-spliced plants even more awful looking. but this is also a good bandaid solution

A visual indicator seems like it would be a cool way to make them still possible whilst a very obvious threat

TobleroneSwordfish:
I've given my reasoning for this, I think it's the best way to allow for more creative and consistent botany features, and losing ricin rafflesias for that is frankly an upside. This is also the only way I can see the nettle and maneater updates being mergeable.

PR #15506, opened 2023-08-23
Add TGUI color picker to character creation

Glamurio:
[QoL] [UI] [Rework]
## About the PR 
Converts the color picker in character creation to TGUI, using a heavily modified version of 

https://github.com/omgovich/react-colorful/ (MIT License, copyright and liability statements included in code)

This was made largely made possible by utilizing a lot of existing color picker code from this PR: 

https://github.com/BeeStation/BeeStation-Hornet/pull/9417

## Why's this needed? 
The current color picker is, in my knowledge, the default Windows color picker. This works most of the time, but is clunky, and prone to errors for some people. The new one is not only sleek and easy to use, but also modifiable and much more versatile, and will allow us to limit color selections by palettes or other restrictions in the future.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Added TGUI to the character creation color pickers.
```


Azrun:
Was permission granted to re-license the non-MIT aspects?

Glamurio:
> Was permission granted to re-license the non-MIT aspects?

According to @ZeWaka we have permission.

ZeWaka:
> > Was permission granted to re-license the non-MIT aspects?
> 
> According to @ZeWaka we have permission.

To clarify:
They were not granted a relicense under our license, but under MIT. So, license headers will need to be added.

code/modules/tgui/tgui_color_picker.dm
```diff
@@ -0,0 +1,222 @@
+/**
```
Azrun, line 1: Please add MIT header with appropriate attribution.

https://github.com/itsmeowForks/BeeStation-Hornet/blame/cb28c720d98d18dde0fc55c0ad0cb425680c5961/code/modules/tgui_input/color.dm

tgui/packages/tgui/styles/themes/generic.scss
```diff
@@ -0,0 +1,37 @@
+$generic: #484455;
```
Azrun, line 1: Please add MIT header with appropriate attribution.

https://github.com/itsmeowForks/BeeStation-Hornet/commits/cb28c720d98d18dde0fc55c0ad0cb425680c5961/tgui/packages/tgui/styles/themes/generic.scss

Azrun (changes requested):

Azrun (approved):

Sovexe:
Seeing a few bugs since the merging of PR

* there is now a color.js and color.ts, and both export a Color class 
* some uses of colors such as the interaction eye and font using style such as `good` `bad` `average` are no longer working on local servers
* interface not found errors occurring on live when trying to use the color selector modal

PR #15510, opened 2023-08-23
AIs speaking to radio channels no longer also speaks locally from their core.

Paaiii:
## About the PR 
Stops AIs talking into their radio comms from broadcasting as normal speech to players in hearing range of their core.
(Basically, AIs speaking to radio channels works like... every other player speaking to radio channels now. )

## Why's this needed? 
If you're talking to the AI in person, rather than over comms, it can be really difficult to tell if what the AI is saying is directed at you, or responding to something said on comms.
Noticing that their speech is in italics in the sidebar is the only real way to tell at current if you don't have access to the comms frequency the AI is talking on.

This is kind of dumb and can be confusing and I don't really see any reason for AI radio speech to be broadcast from their core. So.

## Changelog 

```changelog
(u)Paaiii
(+)AIs speaking to radio channels will no longer appear as regular speech for players standing in proximity to their core. Should make in-person conversations with your local AI less of a headache. 
```

### DIFF
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -1024,6 +1024,7 @@
 					if (R1 && !(A.stat || A.hasStatus(list("stunned", "weakened")))) // Mainframe may be stunned when the shell isn't.
 						R1.talk_into(src, messages, null, A.name, lang_id)
 						italics = 1
+						message_range = 1
 						skip_open_mics_in_range = 1 // First AI intercom broadcasts everything by default.
 						//DEBUG_MESSAGE("AI radio #1 triggered. Message: [message]")
 					else
@@ -1032,6 +1033,7 @@
 					if (R2 && !(A.stat || A.hasStatus(list("stunned", "weakened"))))
 						R2.talk_into(src, messages, null, A.name, lang_id)
 						italics = 1
+						message_range = 1
 						skip_open_mics_in_range = 1
 						//DEBUG_MESSAGE("AI radio #2 triggered. Message: [message]")
 					else
@@ -1040,6 +1042,7 @@
 					if (R3 && !(A.stat || A.hasStatus(list("stunned", "weakened"))))
 						R3.talk_into(src, messages, secure_headset_mode, A.name, lang_id)
 						italics = 1
+						message_range = 1
 						skip_open_mics_in_range = 1
 						//DEBUG_MESSAGE("AI radio #3 triggered. Message: [message]")
 					else


Glamurio:
Thank you, it just caused needless amounts of confusion and didn't really seem realistic. If anything, silicons should be the only ones capable of talking in a radio without it being audible, and humans should always be audible, but for obvious reasons that's not the case.

TheTyrant25:
I believe people always are actually audible, they just 'whisper' when talking into a radio, meaning that if you stand right next to them you can overhear the radio.

Paaiii:
> I believe people always are actually audible, they just 'whisper' when talking into a radio, meaning that if you stand right next to them you can overhear the radio.

Yes, they are! It-- functionally, at least, makes them whisper, so they don't get maptext and what they say shows in italics in the sidebar.
That's what this does-- it doesn't make the AI completely SILENT when using comms, it just... treats it like with any other player. Makes them whisper so they don't get maptext and can't be heard from more than a tile away. Sorry if it doesn't clarify that perfectly well!

Glamurio:
> I believe people always are actually audible, they just 'whisper' when talking into a radio, meaning that if you stand right next to them you can overhear the radio.

Well, yeah. I'm aware, I'm just pointing out that people speaking into a phone are audible to everyone in the room (always leads to confusion as well), but radios don't. The logistics behind the process make sense.

TobleroneSwordfish:
I dunno, I quite like being able to snoop on what the AI's saying if you're standing next to it. It also gives an in-game reason to keep it somewhere secure so people can't overhear comms.

Paaiii:
> I dunno, I quite like being able to snoop on what the AI's saying if you're standing next to it. It also gives an in-game reason to keep it somewhere secure so people can't overhear comms.

You still can snoop on what it's saying this way, same as you can with anyone using radio comms! It's just a bit harder. Besides, I've never seen anyone apply that 'snoopability' for actual use ingame- being able to hear ONE SIDE of a comms conversation with presumably several important participants really isn't very useful.
For things like, for example, people breaking into the AI core and wanting to know if the AI is calling security on them - A lot of AIs are very concious of how audible their comms are, and will just use department PDA to alert security instead. 
With that in mind it flags as more 'bug' than 'feature' to me.
Whereas the maptext for AI comms is a consistent source of confusion when trying to have a conversation.

444Portal:
Makes me wonder why Silicons can't whisper in the first place. As funny as the AI Radio is, this is probably for the best.

kbsmilk (approved):

Glamurio (approved):

TobleroneSwordfish:
I'm going to close this for the reasons listed above, I don't think removing the ability to snoop is a good thing, and talking to the AI in person while it's maintaining conversations on several radio channels being difficult seems in theme.

PR #15511, opened 2023-08-24
directional fire alarm presets

TheTyrant25:
[MAPPING] [QOL] [INTERNAL]
## About the PR
Adds some fire alarm presets with built in pixel offsets, for ease of mapping. This now matches similar convention with light switches and blind switches and APC units. Doesn't affect existing fire alarms.

## Why's this needed?
Makes mapping easier and offsets more consistent for future mapping if needed. Mapping QoL is good.

### DIFF
--- a/code/obj/machinery/firealarm.dm
+++ b/code/obj/machinery/firealarm.dm
@@ -256,7 +256,6 @@ ADMIN_INTERACT_PROCS(/obj/machinery/firealarm, proc/alarm, proc/reset)
 	dir = NORTH
 	pixel_y = -22
 
-
 /obj/machinery/firealarm/east
 	dir = WEST
 	pixel_x = 24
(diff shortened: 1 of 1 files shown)


code/obj/machinery/firealarm.dm
```diff
@@ -247,3 +247,20 @@ ADMIN_INTERACT_PROCS(/obj/machinery/firealarm, proc/alarm, proc/reset)
...
+
+// these seem kind of inverted but it's because an alarm on a wall to the north faces south and etc
+/obj/machinery/firealarm/north
+	pixel_y = 30
+
+/obj/machinery/firealarm/south
+	dir = NORTH
+	pixel_y = -22
+
+
```
Azrun, line 259:
```suggestion
```

Azrun (approved):

PR #15519, opened 2023-08-25
Add grenade storage capability to several Nuke Ops armor suits

FlameArrow57:
[GAME MODES][FEATURE][QoL][BALANCE]

## About the PR 
This PR adds a grenade storage to the nuke ops armor suits for classes that come with a grenade pouch. Specifically, Assault, Heavy, Firebrand, Marksman, and Grenadier. Commander's coat gets one too. The storage can also hold firebot deployers.

Maybe replacing with ammo storage instead, or allowing ammo and grenades would be better? Wanted to propose an idea at least

At best, this just frees up one slot in your inventory that would otherwise be used for holding the grenade pouch, so I don't think it should negatively impact balance too much?

## Why's this needed? 
Some of the nuke ops classes are in an awkward position where they get two pouches, one for ammo and one for grenades, but have one pocket reserved for oxygen. You just have to hold the other pouch in your hand and take out items with your free hand to use it.

One way to make it more QoL would be to add storage to the armor suit. Thought it might be a fun change to add to the tactical/specialist/operative vibe of nuke ops. Not sure if good or bad change.

## Changelog 

```changelog
(u)FlameArrow57
(+)Nuke Ops classes Assault, Heavy, Firebrand, Marksman, and Grenadier, now have storage on their armor suit, as well as for the Commander's coat, that can hold the grenades that their grenade pouches come with.
```

### DIFF
--- a/code/obj/item/clothing/suits.dm
+++ b/code/obj/item/clothing/suits.dm
@@ -1255,7 +1255,7 @@
 		assault
 			New()
 				..()
-				src.create_storage(/datum/storage, can_hold = list(/obj/item/chem_grenade/flashbang, /obj/item/chem_grenede/stinger), slots = 6, opens_if_worn = TRUE)
+				src.create_storage(/datum/storage, can_hold = list(/obj/item/chem_grenade/flashbang, /obj/item/old_grenade/stinger), slots = 6, opens_if_worn = TRUE)
 
 		medic
 			name = "specialist operative medic uniform"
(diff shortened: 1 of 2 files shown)


FlameArrow57:
For the armors listed above, I would think probably not

For all armors, there's some grenades in the starting room so might affect balance some more, since it's in a way adding an extra grenade pouch for classes that didn't have it before

Probably need some more input on this

Glamurio:
I agree with the notion of letting them carry any and all types. Feels weird and arbitrary to limit them, and will cause confusion in players as to why some grenades fit and others don't. Grenades are pretty uniformly the same size.

PR #15521, opened 2023-08-25
Removes spawning with lighter on Smoker trait

### LINKED ISSUE #15389: Smoker trait overrides Trinket trait (Lunchbox)
### Labels

[BUG][MINOR]

### Description

Took smoker trait and lunchbox trait and the lighter it gave me overrode my lunchbox trait

### Steps to reproduce

-  Take smoker trait
-  Take Lunchbox trait
-  Lighter but no lunchbox!

### Expected Behavior

Either you get both a lunchbox and a lighter or the smoker trait should be mutually exclusive with trinket traits like they are with each other

Reported by: Joeled
Active test merges: nothing
Round log date: 2023-08-15-01-59
Reported on: 2023-08-15 02:05:52
Map: NADIR

TealSeer:
[BUG] [REMOVAL] [TRAITS]
## About the PR 

Fixes #15389 by removing the zippo trinket from the Smoker trait. 

## Why's this needed? 

Currently the Smoker trait sets your trinket to be a lighter as well as stopping you from absorbing chemicals from smoking, the problem with this is it's not categorized as a trinket trait (which are all set to be mutually exclusive with each other) so if you take it and a trinket trait it can overwrite whatever other trait you took. My decision to just remove the lighter is what I see as the best of three possible options:

1. Remove the lighter from Smoker.
2. Keep the lighter on Smoker and move it to the trinket category, thus fixing the overwriting behavior. I don't like this option because Smoker as a trait isn't exactly billed on giving you a lighter; it's for stopping you from absorbing chems from smoking, and making you choose between also getting a lighter or getting to customize your trinket in another way kind of sucks in my eyes.
3. Rewrite the way trinkets are handled so you can get both. I really really really don't want to do this one trait code and trinket code already has a million weird interactions and special cases that have cropped up please don't make me rewrite it

## Changelog 

```changelog
(u) TealSeer
(+) The smoker trait no longer gives you with a lighter.
```

### DIFF
--- a/code/procs/jobprocs.dm
+++ b/code/procs/jobprocs.dm
@@ -691,8 +691,6 @@ var/global/totally_random_jobs = FALSE
 			trinket = new/obj/item/reagent_containers/food/snacks/ingredient/egg/bee/buddy(src)
 		else
 			trinket = new/obj/item/reagent_containers/food/snacks/ingredient/egg/bee(src)
-	else if (src.traitHolder && src.traitHolder.hasTrait("smoker"))
-		trinket = new/obj/item/device/light/zippo(src)
 	else if (src.traitHolder && src.traitHolder.hasTrait("lunchbox"))
 		var/random_lunchbox_path = pick(childrentypesof(/obj/item/storage/lunchbox))
 		trinket = new random_lunchbox_path(src)


TealSeer:
Alternatively, if there's enough interest, I can add a trinket trait to give you a lighter alongside this change.

DisturbHerb:
honestly option 3 is best imo

Glamurio:
Hate to say it but I agree, option 3 is the correct option. This is one of those situations where people knew about the interaction with the lighter, but they knew that the proper fix is far too much of a hassle to implement, so it was just left this way. So, in my mind, option 1 is a band-aid fix to circumvent having to fix the actual problem, which is trait and trinket code.

TealSeer:
Closing cuz it sounds like I'll need to come up with a more graceful way to handle this

PR #15522, opened 2023-08-26
tgui Oxygen Vending Machine

Garash2k:
## About the PR 
Refactors the oxgyen vending machine UI in tgui.
image

## Why's this needed? 
Breathing is good

## Changelog 

```changelog
(u)Garsh
(+)New UI for the Oxygen Vending Machine
```
[Refactor][UI]


Garash2k:
I asked about the ellipsis issues on coderbus because I'm pretty sure it's an upstream issue that's yet to be fixed.

Glamurio:
I know it's a minor issue that other TGUI's suffer from as well, but could we start doing a linebreak for name and occupation on ID cards? The whole being cut off and it saying "Staff..." is not only ugly, but some jobs aren't apparent at first glance anyway, such as Security Officers or Assistants, Head of Personnel or Security, or Medical Director and Doctor.

If it's too hard to line break, I'd personally rather see it removed entirely, considering it doesn't add any valuable information.

TobleroneSwordfish:
Some other alternatives include: cutting the `'s ID card` part (although I think that's how most machines display it) or just making the window wider to accommodate.

Garash2k:
The problem with a line break inside buttons is that it wraps the text round the icon in an ugly fashion
kllmf2

I don't think the ellipsis is a big deal for the oxygen vendor ui, especially since the full card name is not important info, and there's a title tooltip too.
image

The title that's plugged there is the card's name. It's what's currently done everywhere, and I like how simple it is, instead of having to build it, or hack a line break in it with a string replace
image

As for making the window bigger, there's always people that are gonna be of the opinion that it should be more compact, it's hard to balance.

For now I'll increase it from 280px to 350px, so that medium sized names with typical jobs of medium length fit.
image

code/modules/vending/vending.dm
```diff
@@ -652,6 +655,37 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item, proc/admin_command
...
+				if(credit >= cost)
+					src.credit -= cost
+					air_machine.fill()
+					return
+				else if(src.scan)
+					var/datum/db_record/account = FindBankAccountByName(src.scan.registered)
+					if (account && account["current_money"] >= cost)
+						account["current_money"] -= cost
+						air_machine.fill()
+						return
```
TobleroneSwordfish, line 688: Why not just put this on the subtype instead of casting `src`?

TobleroneSwordfish (changes requested):
UI looks good, code looks okay at a glance apart from this

code/modules/vending/vending.dm
Garash2k: At first I tried doing that, then I gave up pretty fast when I noticed hiccups with the parent's interactivity check and when I noticed that "player vending machine exclusives" were also in that same `ui_act`.

Looking into it more, the problem is that the parent `ui_act` returns true when the user is disallowed from interacting with the ui, and the children `ui_act`s return true when an interface refresh is needed. So if I were to add a second `ui_act` on the subtype, I wouldn't be able to know if the TRUE I'm getting from calling the parent is because UI interaction isn't allowed or if it is allowed, and it's that the children wants the UI refreshed.

Should I do some sort of hack where it returns something like -1 when the UI isn't interactive, or make some `ui_act_o2` method on the subtype and call it from `ui_act`? These are the alternatives I'm seeing

TobleroneSwordfish: Ah, well the issue is that `ui_act` is meant to return true when it "handles" a request and false otherwise, but this one is set up to return false when it fails to handle a request and true otherwise. Basically you can add `return TRUE` to all the cases on the parent where it's meant to return true and remove the `. = TRUE` to then be able to do `if (..()) return` in the child `ui_act` to check if either the UI is not interactive *or* the parent has already handled the request.

PR #15527, opened 2023-08-26
Give bullets a low chance to ricochet

FlameArrow57:

## About the PR 
This PR gives bullets a 10% chance to ricochet off walls. Bullets that ricochet have reduced much reduced power, range, and armor pierce.

Bullets that can ricochet are single round, lethal bullets that just deal damage. Doesn't include ones like rubber slugs, pipeshot, or buckshot.

Sound from https://freesound.org/people/aust_paul/sounds/30932/, used under [CC0](https://creativecommons.org/publicdomain/zero/1.0/)

Ex.
https://streamable.com/croqdn

## Why's this needed? 
Thought it would be cool. More for a fun effect, not gameplay changes.

## Changelog 

```changelog
(u)FlameArrow57
(+)Common lethal bullets and grenade shrapnel now have a low chance to ricochet off walls. Ricocheted projectiles have reduced damage and don't travel as far.
```


ZeWaka (commented):
Why don't rubber slugs ricochet... they're rubber?

FlameArrow57:
Yeah I can add it. Rubber slugs have knockback on hit, just wasn't sure on those types of projectiles.

Gannets:
oh sweet, I can finally add a revolver ocelot nuke ops class

code/turf/walls.dm
```diff
@@ -340,6 +340,32 @@ TYPEINFO(/turf/simulated/wall)
...
+	if (!istype(P.proj_data, /datum/projectile/bullet))
+		return
+	var/datum/projectile/bullet/bullet = P.proj_data
+	if (!bullet.ricochets)
+		return
+	if (prob(90))
+		return
+
+	var/proj_name = P.name
+	var/obj/projectile/p_copy = SEMI_DEEP_COPY(P)
```
Tarmunora, line 354: this feels cursed. Is there really not a better way?

FlameArrow57: Not sure really. I was trying it without the copy, modifying the variables on the reflected projectile, and it was compounding for bullets in the rest of the magazine - first shot 1/2 the distance, second shot 1/4 distance, and so on

code/turf/walls.dm
```diff
@@ -340,6 +340,32 @@ TYPEINFO(/turf/simulated/wall)
...
+	p_copy.proj_data.shot_sound = "sound/weapons/ricochet/ricochet-[rand(1, 4)].ogg"
+	p_copy.proj_data.power = sqrt(p_copy.proj_data.power)
+	p_copy.proj_data.dissipation_delay *= 0.5
```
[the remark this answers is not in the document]
FlameArrow57: Initial thought was to make it so that bullets reflected will always travel up to half their max distance on reflection, rather than taking into account current distance travelled, if that's what you're meaning

PR #15536, opened 2023-08-27
Adds industrial chair & parts

Glamurio:
[Objects] [Feature]

## About the PR 
Adds an industrial chair in the same visual style as the existing industrial table.
image

Industrial chair parts are crafted with a sheet of reinforced metal, and constructed the same as any other chair out of parts.

Also changes `/obj/stool/chair/wooden/` to `/obj/stool/chair/dining/` and instead makes it inherit from that. The same is done for `/obj/item/furniture_parts/wood_chair` to `/obj/item/furniture_parts/dining_chair`.

## Why's this needed? 
It's always bugged me that there's no fitting chairs for industrial tables, because I love how they look.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Added industrial chairs. You can craft them out of reinforced metal.
```


code/obj/item/table_rack_parts.dm
```diff
@@ -471,6 +471,13 @@ TYPEINFO(/obj/item/furniture_parts/table/glass)
 	icon_state = "scrapchair_parts"
 	furniture_type = /obj/stool/chair/wooden/scrap
 
+/obj/item/furniture_parts/wood_chair/industrial
+	name = "industrial chair parts"
+	desc = "An collection of rods and scaffolding that can be used to make an industrial chair."
+	icon_state = "ichair_parts"
+	furniture_type = /obj/stool/chair/wooden/industrial
```
TobleroneSwordfish, line 478: this doesn't look very wooden

TobleroneSwordfish (changes requested):

code/obj/item/table_rack_parts.dm
Glamurio: Neither does the scrap table, but it just so happens that "dining chairs" are all wooden. I was genuinely considering changing the name of the path, but figured that's excessive. But if this is asked for, I'll do it

Glamurio:
@TobleroneSwordfish I hope you're satisfied now

PR #15542, opened 2023-08-27
Allow some older antagonists to use usual hotkeys for their abilities

FlameArrow57:
[PLAYER ACTIONS][GAME MODES][BALANCE][QoL]

## About the PR 
This PR allows binding of Shift, Alt, and Ctrl for abilities for vampires, werewolves, and changelings.

## Why's this needed? 
Conventional hotkey binding, faster gameplay

## Changelog 

```changelog
(u)FlameArrow57
(*)Vampires, werewolves, and changelings can now bind abilities to Shift, Alt, and Ctrl.
```

### DIFF
--- a/code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
+++ b/code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
@@ -160,6 +160,9 @@
 		if (incapacitationCheck())
 			boutput(holder.owner, "<span class='alert'>We cannot use our abilities while incapacitated.</span>")
 			return 0
+		if (!isturf(src.holder.owner.loc) && !src.can_use_in_container)
+			boutput(src.holder.owner, "<span class='alert'>You can't use this ability here.</span>")
+			return FALSE
 		if (!human_only && !abomination_only)
 			return 1
 		var/mob/living/carbon/human/H = holder.owner
@@ -171,12 +174,6 @@
 			else
 				return 1//what could possibly go wrong
 		return 0
-
-	cast(atom/target)
-		. = ..()
-		if (!isturf(src.holder.owner.loc) && !src.can_use_in_container)
-			boutput(src.holder.owner, "<span class='alert'>You can't use this ability here.</span>")
-			return TRUE
 
 	Stat()
 		if (!human_only && !abomination_only)
--- a/code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
+++ b/code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
@@ -172,6 +172,12 @@
 				return 1//what could possibly go wrong
 		return 0
 
+	cast(atom/target)
+		. = ..()
+		if (!isturf(src.holder.owner.loc) && !src.can_use_in_container)
+			boutput(src.holder.owner, "<span class='alert'>You can't use this ability here.</span>")
+			return TRUE
+
 	Stat()
 		if (!human_only && !abomination_only)
 			..()
--- a/code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
+++ b/code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
@@ -1,40 +1,3 @@
-/atom/movable/screen/ability/topBar/changeling
-	clicked(params)
-		var/datum/targetable/changeling/spell = owner
-		var/datum/abilityHolder/holder = owner.holder
-
-		if (!istype(spell))
-			return
-		if (!spell.holder)
-			return
-
-		if(params["shift"] && params["ctrl"])
-			if(owner.waiting_for_hotkey)
-				holder.cancel_action_binding()
-				return
-			else
-				owner.waiting_for_hotkey = 1
-				src.UpdateIcon()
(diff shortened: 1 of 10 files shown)


Tarmunora (approved):

code/modules/antagonists/wrestler/abilities/kick.dm
```diff
@@ -2,10 +2,7 @@
 	name = "Kick"
 	desc = "A powerful kick, sends people flying away from you. Also useful for escaping from bad situations."
 	icon_state = "Kick"
-	targeted = 1
```
TobleroneSwordfish, line 5: This is a massive buff to wrestling belts, it'll let them fight multiple people without having to deal with popup menus. I know them being balanced by ancient ability jank is bad but it's better than nothing.

code/modules/antagonists/wrestler/abilities/strike.dm
```diff
@@ -2,10 +2,7 @@
 	name = "Strike"
 	desc = "Hit a neaby opponent with a quick attack."
 	icon_state = "Strike"
-	targeted = 1
```
TobleroneSwordfish, line 5: also here

TobleroneSwordfish (changes requested):

code/modules/antagonists/wrestler/abilities/kick.dm
Tarmunora: imo better to unjank it then balance as needed

FlameArrow57: It says it's targeted, but its actually not. Current wrestlebelt abilities already activate on the nearest mob. Old code was overriding it, so changes here don't affect anything

TobleroneSwordfish: Okay I tested it and it actually uses a mouse target selector now which is less jank. Overall though if you're want to make them auto-target you're going to have to nerf wrestlebelts quite a bit since the only real counter to them is multiple people.

FlameArrow57: I'm sorry, but I'm confused why it's being suggested that something be balanced by an anti-QoL feature??? Allowing hotkeys to be used bypasses targeting anyways, and the only time you are really going to be surrounded by multiple people is likely when you are being arrested by multiple security officers or are being beaten to death. Fighting multiple people is not a good idea.

Agree that targeting is lost now when surrounded by multiple mobs, but allowing manual targeting is not supported that well... Can look into it some more if needed but for the moment I just made it pick a random nearby mob

TobleroneSwordfish: Right but again in practical terms this removes the item's only counterplay. If you want to do that then you're going to have to balance it some other way.

FlameArrow57: Okay. I'm going to undo the wrestler code for the moment, was mostly just looking to convert to hotkeys. Not sure how to best balance wrestling belt

FlameArrow57:
Was originally like that

code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
```diff
@@ -172,6 +172,12 @@
 				return 1//what could possibly go wrong
 		return 0
 
+	cast(atom/target)
+		. = ..()
+		if (!isturf(src.holder.owner.loc) && !src.can_use_in_container)
+			boutput(src.holder.owner, "<span class='alert'>You can't use this ability here.</span>")
+			return TRUE
+
```
TobleroneSwordfish, line 180: this should be in `castcheck` I think

code/modules/antagonists/werewolf/abilities/_werewolf_ability_holder.dm
```diff
@@ -400,6 +400,10 @@
 			boutput(M, "<span class='alert'>You can't use this ability when restrained!</span>")
 			return 0
 
+		if (!isturf(src.holder.owner.loc))
```
TobleroneSwordfish, line 403: Aren't there some werewolf abilities you'd want to be used in a container? (Like transform)

TobleroneSwordfish (changes requested):

code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
FlameArrow57: Works here, which is why I left it, does it matter?

code/modules/antagonists/werewolf/abilities/_werewolf_ability_holder.dm
FlameArrow57: IDK, I'm just going with how it was originally

code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
TobleroneSwordfish: Right but the checks should be in the check proc, otherwise you'll get issues with code assuming the ability has successfully cast when it hasn't.

TobleroneSwordfish (changes requested):

PR #15546, opened 2023-08-27
Update floating tile sprite, dirs in void and some simmed walls in void

CheffieGithub:
## About the PR 
Modernises the floating tiles sprite and makes the directions correct, updated the void dirs to be the same as before.
Replaced the simulated void walls in the vault area with unsimulated versions.

image

## Why's this needed? 
The sprites are really old, they now use the current plating sprite as reference.


flappybatpal:
image

I wouldn't say the originals look flawless but the loss of the edge shadowing seems like a downgrade.

CheffieGithub:
I didn't add them because I have no experience with shadowing but I can try

PR #15547, opened 2023-08-27
Adds a random wig crate

unfunnyperson:
[CLOTHING] [FEATURE]

## About the PR 
Adds a wig crate for 2000 credits that contains 7 randomized wigs. 

## Why's this needed? 
More things for cargo to buy are always cool. Also allows for bald people, borgs, and mutantraces to get hair more easily.

## Changelog

```changelog
(u)UnfunnyPerson
(+)Added a randomized wig crate. Cargo can order this crate for 2000 credits.
```

### DIFF
--- a/code/modules/barber/barber_shop.dm
+++ b/code/modules/barber/barber_shop.dm
@@ -58,17 +58,21 @@
 /obj/item/clothing/head/wig/spawnable/random
 
 	New()
-		var/list/possible_hairstyles = null
+		src.name = "wig"
+		var/list/possible_hairstyles
+
 		if (prob(50))
 			possible_hairstyles = filtered_concrete_typesof(/datum/customization_style, /proc/ismasc)
 		else
 			possible_hairstyles = filtered_concrete_typesof(/datum/customization_style, /proc/isfem)
-		var/datum/customization_style/hair_type = null
+
+		var/datum/customization_style/hair_type
 		var/picked_color = rgb(rand(0,255),rand(0,255),rand(0,255))
 		hair_type = pick(possible_hairstyles)
 		first_id = initial(hair_type.id)
 		first_color = picked_color
 		if (prob(33))
+			hair_type = pick(possible_hairstyles)
 			second_id = initial(hair_type.id)
 			second_color = picked_color
 		..()
--- a/code/modules/barber/barber_shop.dm
+++ b/code/modules/barber/barber_shop.dm
@@ -54,6 +54,24 @@
 		hair_list[third_id] = third_color
 		src.setup_wig(hair_list)
 
+///Randomized wig for the cargo crate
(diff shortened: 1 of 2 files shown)


TealSeer:
I love the idea, but the way you have wigs being generated currently leads to some really bizarre ones that I don't know if anybody would want to wear (multiple base hairstyles overlapping, ones that are just the detail modifier of a hairstyle and nothing else, etc.) It might be worth checking out the code used to generate a random character because it has some checks to make sure you only get base hair styles first, then roll a chance to add the details for that hair style. Relevant stuff starts here:

https://github.com/goonstation/goonstation/blob/e39db65de4806ecde98c9ed0ae75ac74344abbba/code/datums/preferences.dm#L2004-L2059

code/modules/barber/barber_shop.dm
```diff
@@ -54,6 +54,25 @@
 		hair_list[third_id] = third_color
 		src.setup_wig(hair_list)
 
+///Randomized wig for the cargo crate
+/obj/item/clothing/head/wig/spawnable/random
+
+	New()
+		var/list/possible_hairstyles = null
```
TobleroneSwordfish, line 61:
```suggestion
		src.name = "wig"
		var/list/possible_hairstyles = null
```
they're called toupées by default for some reason

code/modules/economy/supply_packs.dm
```diff
@@ -1187,6 +1187,15 @@ ABSTRACT_TYPE(/datum/supply_packs)
 	containertype = /obj/storage/crate/packing
 	containername = "Haberdasher's Crate"
 
+/datum/supply_packs/random_wigs
+	name = "Wigs"
```
TobleroneSwordfish, line 1191:
```suggestion
	name = "Spare wigs crate"
```
To match up with how other crates are named.

TobleroneSwordfish (changes requested):

code/modules/barber/barber_shop.dm
```diff
@@ -54,6 +54,25 @@
...
+		else
+			possible_hairstyles = filtered_concrete_typesof(/datum/customization_style, /proc/isfem)
+		var/datum/customization_style/hair_type = null
+		var/picked_color = rgb(rand(0,255),rand(0,255),rand(0,255))
+		hair_type = pick(possible_hairstyles)
+		first_id = initial(hair_type.id)
+		first_color = picked_color
+		if (prob(33))
+			second_id = initial(hair_type.id)
+			second_color = picked_color
```
Tarmunora, line 73: wouldn't you want a *different* second style here

unfunnyperson: I... how did I miss that :sheltersweat:

Tarmunora (approved):

PR #15548, opened 2023-08-28
Tries out doubling the speed of Atmospheric machines and pipe_networks.

Laboredih123:
[ATMOSPHERICS][EXPERIMENTAL]
## About the PR 
Tests out doubling processing speeds for atmos machines and pipe_networks.

## Why's this needed? 
I guess because it would be nice not to wait a century for a room to fill from a vent?

## Changelog 

```changelog
(u)cringe
(*)The rate at which pipe_networks and atmospheric machines do stuff has been doubled. Experience things such as vents, injectors, pumps, and probably more are no longer painfully slow. Also observe as nothing is done to preserve balance.
```

### DIFF
--- a/code/datums/controllers/process/machines.dm
+++ b/code/datums/controllers/process/machines.dm
@@ -30,7 +30,7 @@
 	doWork()
 		var/c = 0
 
-		if (ticker % 8 == 0)
+		if (ticker % 4 == 0)
 			src.atmos_machines = by_cat[TR_CAT_ATMOS_MACHINES]
 			for (var/obj/machinery/machine as anything in atmos_machines)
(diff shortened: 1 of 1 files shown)


Sbmhawk:
+1, I hooked up Co2 to cog1 medbay air hookups, and it took a full 10 minutes before it took effect.

Wisemonster:
If this isn't too intensive on the server, I'd like this. Though, I'm not sure how intensive this would be, especially on higher pop counts.

Azrun (commented):
Consider using a define so that they are coupled.

pali6:
> Consider using a define so that they are coupled.

Frankly they should just use machine processing tiers. Those already do exactly this same functionality.

PR #15549, opened 2023-08-28
Donut2 AIsat tweaks

TheTyrant25:
[MAPPING]
## About the PR
- adds a secoff spawn point in the little checkpoint in the ai satellite. For fun. (they're able to use the teleporter to get back)
- Changes the interior ai satellite power grid to blue cabling to distinguish it from the out solar connections (red and yellow). The blue network is controlled by the SMES.

## Why's this needed?
Sec spawn? It's funny. And there's actually equipment there that they can use to suit up with first anyway.
Blue cables? It visually distinguishes the inner and outer grids. And looks nice.

### DIFF
--- a/maps/donut2.dmm
+++ b/maps/donut2.dmm
@@ -18174,10 +18174,9 @@
 /turf/simulated/floor/plating/airless,
 /area/station/turret_protected/AIbaseoutside)
 "btm" = (
-/obj/cable/yellow,
 /obj/grille/steel,
-/obj/cable{
-	icon_state = "0-2"
+/obj/cable/yellow{
+	icon_state = "1-2"
 	},
 /turf/simulated/floor/plating/airless,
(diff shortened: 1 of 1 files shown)


Sord213 (changes requested):
The sec spawn is funny but if a newer player spawns there that doesn't know how to use teleporters it could cause some confusion.

Glamurio:
> The sec spawn is funny but if a newer player spawns there that doesn't know how to use teleporters it could cause some confusion.

Don't Security Officers have like a minimum of 30 or 50 rounds played before you can pick them? If it's SecOff only and not SecAss, it should be okay.

TheTyrant25:
I thought new players had to have a certain amount of hours before they could play secoff?

TheTyrant25:
that was my thought too

TheTyrant25:
i've ditched the spawn anyway

PR #15550, opened 2023-08-28
makes lockers with leg holes faster

ZWRh3:
## About the PR 
makes lockers with leg holes move 1 tile per second (used to be per 5)

## Why's this needed? 
Leg holes are something you have to go out of your way to add to a locker and 1 tile per 5 seconds is just painful to use even for gimmicks. I tested some things and it seems it won't have balance issues.

You cannot even open doors in a locker or interact with the world in most ways without opening it thanks to that you have to choose between being welded and somewhat safe but nearly powerless or 1 left click away from being defeated

The moving of the floor closet is also quite blatant and has the downsides above as well so it should not be an issue and if it is its movement time can be increased to a higher time.

## Changelog 

```changelog
(u)ZWRh3
(+)Lockers with leg holes are now 5x faster.
```

### DIFF
--- a/code/obj/storage/large_storage_parent.dm
+++ b/code/obj/storage/large_storage_parent.dm
@@ -6,7 +6,7 @@
 // PLEASE JUST MAKE A MESS OF make_my_stuff() INSTEAD
 // CALL YOUR PARENTS
 
-#define RELAYMOVE_DELAY 50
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
I'm sure this won't go horribly wrong

PR #15554, opened 2023-08-28
twirl knife to flick off the blood

TheTyrant25:
[FEATURE]
## About the PR
Adds the ability to flick blood off knives and onto the floor. Performed by using the twirl emote. Forensic traces still remain though.

Currently added to kitchen knives, all swords, and syndicate butcher knives.

## Why's this needed?
It's a fun little thing that could be used in rp for extra drama. For classic, you can think of it as wiping it off on a shirt.

## Changelog
```changelog
(u)Tyrant
(+)Twirling a bloody knife will flick the blood off and onto the floor. Doesn't work if the blood is dried.
```

### DIFF
--- a/code/datums/components/bloodflick.dm
+++ b/code/datums/components/bloodflick.dm
@@ -41,7 +41,7 @@ TYPEINFO(/datum/component/bloodflick)
 		make_cleanable(/obj/decal/cleanable/blood, get_turf(src.parent))
 		playsound(src.weapon.loc, 'sound/impact_sounds/Slimy_Splat_1.ogg', 40, 1)
 		SPAWN(1 DECI SECOND) // so that the twirl emote message appears first (in theory)
-			boutput("Blood splatters onto the floor!") // i will be accepting suggestions on this line btw, reviewers
+			boutput("<span class='notice'>Blood splatters onto the floor!</span>") 
 
 /// applies wet blood to the knife and starts the blood drying countdown
 /datum/component/bloodflick/proc/wetten()
--- a/code/datums/components/bloodflick.dm
+++ b/code/datums/components/bloodflick.dm
@@ -2,22 +2,20 @@ TYPEINFO(/datum/component/bloodflick)
 	initialization_args = list()
 
 /datum/component/bloodflick
+	/// typecasted thing that we're flicking blood off.
+	var/obj/item/blade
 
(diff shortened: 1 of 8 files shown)


Glamurio:
Please also add this to katanas.

TheTyrant25:
i forgot about them, good point. Anything else that should have this? and should i componetize it?

Glamurio:
> i forgot about them, good point. Anything else that should have this? and should i componetize it?

Honestly, if you're willing to make this a feature, you might as well add this to any kind of bladed weapon. Ideally it would be cool if you could only do this if the blood is fresh and not dried, but I don't know if "blood freshness" gets tracked on a blade.

TheTyrant25:
it doesn't seem like it does, no. I can add it though.

code/datums/components/bloodflick.dm
```diff
@@ -0,0 +1,77 @@
...
+	// if all the blood is wet, flicking it off cleans it.
+	if (!src.hasdry)
+		src.weapon.clean_forensic()
+	// if there's wet blood on it, flick it off
+	if (src.haswet)
+		src.haswet = FALSE
+		make_cleanable(/obj/decal/cleanable/blood, get_turf(src.parent))
+		playsound(src.weapon.loc, 'sound/impact_sounds/Slimy_Splat_1.ogg', 40, 1)
+		SPAWN(1 DECI SECOND) // so that the twirl emote message appears first (in theory)
+			boutput("Blood splatters onto the floor!") // i will be accepting suggestions on this line btw, reviewers
```
frawhst, line 44:
```suggestion
			boutput("<span class='notice'>Blood splatters onto the floor!</span>") 
```
Line is fine but gotta style it

code/datums/components/bloodflick.dm
```diff
@@ -0,0 +1,77 @@
...
+	var/obj/item/dummy = src.parent
+	if (!dummy.blood_DNA) // not all attacks leave blood on the parent
+		return
+	if (!src.haswet)
+		src.haswet = TRUE
+		src.iswet += 1
+	SPAWN(drytime)
+		// in case
+		if (!src)
+			return
```
frawhst, line 57: Fairly sure this will never happen unless someone sets `src` to null manually in which case god help us
```suggestion
```

PR #15558, opened 2023-08-29
Abstracts under/suit parent. makes some under/suits absolute. add Mortician suit.

CheffieGithub:
## About the PR 
Seperates the black suit from `/obj/item/clothing/under/suit` and marks the parent abstract because later it could use its own file or something, for now its just empty. 

Absolutely paths under/suit subtypes that weren't for ease of searching.

Adds a mortician suit as a subtype to replace varedits.

Add name and desc to black dress to replace varedits

## Why's this needed? 
Parents should generally be abstract. Absolute paths for subtypes are easier to search for. Having a morticans concrete type with a fitting description is nice.


Glamurio:
> For some reason this is not updating with new commits?

Perhaps try resetting your `goonstation.dme` as well as your `__build.dm`. I find that often solves weird commit issues for me.

Tarmunora (approved):

CheffieGithub:
Ready now

PR #15569, opened 2023-08-29
Nuclear glowsticks

amylizzle:
## About the PR 
Adds glowsticks as an alternative fuel source for the nuclear reactor. They're really bad.

## Why's this needed? 
Sometimes there's nobody in mining.

## Changelog 

```changelog
(u)Amylizzle
(*)You can now make shoddy fuel rods using a glowstick in the nuclear fabricator.
```

### DIFF
--- a/code/modules/materials/Mat_FabRecipes.dm
+++ b/code/modules/materials/Mat_FabRecipes.dm
@@ -23,6 +23,24 @@
 	materials = list("!any"=2)
 	result = /obj/item/reactor_component/fuel_rod
 
+/datum/matfab_recipe/makeshift_fuel_rod
+	name = "Makeshift Nuclear Fuel Rod"
+	desc = "A fuel rod for a nuclear reactor, made out of glowsticks"
+	category = "Nuclear"
+
+	New()
+		..()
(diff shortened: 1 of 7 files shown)


code/modules/materials/Mat_FabRecipes.dm
```diff
@@ -23,6 +23,24 @@
...
 	result = /obj/item/reactor_component/fuel_rod
 
+/datum/matfab_recipe/makeshift_fuel_rod
+	name = "Makeshift Nuclear Fuel Rod"
+	desc = "A fuel rod for a nuclear reactor, made out of glowsticks"
+	category = "Nuclear"
+
+	New()
+		..()
+		required_parts.Add(new/datum/matfab_part/glowstick {part_name = "Glowstick"; required_amount = 1} ())
```
TobleroneSwordfish, line 33: Says it's made of multiple glowsticks, but it's only made of one! Lies!
(if you want to make it multiple then you could either just pick one colour to make the rod or steal the rainbow light from glowstick trees)

TobleroneSwordfish (approved):

code/modules/materials/Mat_FabRecipes.dm
amylizzle: ah it was going to be but they don't stack in the fabricator interface :skull:

PR #15575, opened 2023-08-30
Redo and improve Mushroom toxins

MeggalBozale:
[ENHANCEMENT] [MAPPING] [SCIENCE] 
## About the PR
Completely transforms toxins on Mushroom to contain all the features necessary for.. toxins reasearch (TTV Making, heating up/cooling down gases)
This is my first mapping change to more significantly affect a station, even if one that isn't often seen in rotation.

Old:
dreamseeker_SiVrQEDkWB

New:
dreamseeker_oZetgrtGSC

## Why's this needed?
Currently, toxins in Mushroom is so old that it does not have the usual features seen in other toxins maps. The cooling and heating loops are completely absent from the map, and toxins storage spawns with air canisters, for some reason. Remapping toxins lets it be usuable, and if I've done it right, it'll look better too.

### DIFF
--- a/maps/unused/mushroom_new_walls.dmm
+++ b/maps/unused/mushroom_new_walls.dmm
@@ -3003,6 +3003,9 @@
 	},
 /turf/simulated/floor/plating,
 /area/station/maintenance/solar/west)
+"ans" = (
+/turf/simulated/floor,
+/area/station/science/lab)
 "anv" = (
 /obj/machinery/power/smes{
 	chargelevel = 5000;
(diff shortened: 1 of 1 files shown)


Sord213 (changes requested):
Couple small changes I'd like to see.

Screenshot_465
Move the tox storage area up so it covers the north walls + the airlock into tox storage so the airlock names properly
Screenshot_466
Use the toxins access spawners on the doors in toxins, please. I know the rest of sci doesn't use the newer ones but if you change an area like this you should bring it up to standard. On the yellow highlighted gas purge could you put a valve on it. If you have to move the connect up one tile that's fine. Maybe on the cold loop bit put some blue floors under it to show it's the cooling loop not the vent.

MeggalBozale:
image
I moved both up, and added a valve to the purge line. I moved the cold loop up one just so you can see the blue pipe (There's one in the glass, but it's very hard to see)
I would rather not change the flooring, I don't know how I could do it well without wrapping the purple trim around blue flooring. If you have an idea of how to do so well, do let me know, I am open to it but I believe I would not make it look good.

I put the proper door access on, apologies for that. I've just learned I can have several maps up at once in StrongDMM, so I'm taking a quick look-over with other toxins maps to see if there's anything else I could do.

Sord213:
god why are those gas sensors INVISIBLE in sdmm. lmao who did that

PR #15231, opened 2023-08-04
Add listening post access to sleepers

cheekybrdy:
## About the PR 
This is a rework of https://github.com/goonstation/goonstation/pull/15222 which is a rework of https://github.com/goonstation/goonstation/pull/14992 because I messed something up again. This PR adds handscanners to the listening post which allow sleeper agents to leave and enter it.

## Why's this needed? 
This gives sleepers access to items in the listening post giving them traitor items without the full power of the traitor uplink. This makes them feel like more then a late HMT.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Sleeper agents can now access the listening post through scanners at the entrance.
```

PR #15578, opened 2023-08-30
Donut External Door Fix

cheekybrdy:
## About the PR 
Fixes some doors that broke with #15231.

## Why's this needed? 
fix good


TealSeer:
Think your branch was outdated, afaik this was already fixed in https://github.com/goonstation/goonstation/commit/c4f01b2d9e4427ee1111ae7b8b8120e8485155e8

TealSeer:
Okay whoops no this was actually getting two doors that were missed in that fix apologies

PR #15583, opened 2023-08-30
remove "Assless Chaps" from gang jumpsuits

keiyakins:
## About the PR 
Removes the assless chaps from the list of valid items for gang uniforms.

## Why's this needed? 
Now that gangs are preassigned, you don't have the choice to reject on uniform. Forcing people to wear leather fetish outfits is incredibly fucked up.

## Changelog 

```changelog
(u)Keiya
(+)Removed Assless Chaps from the available gang uniforms.
```


Tarmunora (approved):

keiyakins:
I don't disagree that they could just be removed, but that would involve modifying all the maps and generally just be more work, and I don't really mind other people wearing them.

PR #15589, opened 2023-08-30
Fix cloner defects lingering references

CheffieGithub:
## About the PR 
Disposes the cloner_defect_holder when human is disposed, disposing for that which nulls owner and disposes the active defects which also null owner.

## Why's this needed? 
Fixes them preventing human GC


code/WorkInProgress/cloner_defects.dm
```diff
@@ -64,6 +64,12 @@
 					// lazylistassoc when (wici)
 					sublist[defect_type] = initial(defect_type:weight) // we can use initial to get initial values of a type without an instance
 
+	disposing()
+		src.owner = null
+		for (var/datum/cloner_defect/defect as anything in src.active_cloner_defects)
+			defect.dispose()
```
pali6, line 70: Any particular reason why dispose instead of qdel?

Also you should null the active_cloner_defects list here too.

Does owner have no reference to src here? Like onwer.cloner_defects or something?

CheffieGithub: I disposed because when I was checking human disposing it called dispose on things, not sure which was best so I just went with dispose().

CheffieGithub: active_cloner_defects should get nulled at the end of the loop?

TobleroneSwordfish: `qdel` is almost always preferred over manually calling dispose unless you have a reason to do otherwise

CheffieGithub:
owner has no refernce to src as far as I know, I tested log harddel refs with dispose() = qdel and there was no GC fail.

PR #15595, opened 2023-08-31
Gives AIs a command to unlock their core cover.

Paaiii:
## About the PR 
Adds a verb to AI commands that allows the AI player to unlock or relock their own core cover, similar to cyborgs.

## Why's this needed? 
Cyborgs already have a command to do this with their cover, and the fact that the AI can't unlock themselves can be a consistent pain, especially if there aren't any command.

## Changelog 

```changelog
(u)Paaiii
(+)Gives AIs a command to unlock their own core cover.
```


code/mob/living/silicon/ai.dm
```diff
@@ -126,6 +126,7 @@ var/global/list/ai_emotions = list("Annoyed" = "ai_annoyed-dol", \
 	var/moustache_mode = 0
 	var/status_message = null
 	var/mob/living/silicon/deployed_shell = null
+	var/locking = 0
```
Tarmunora, line 129: If you're going to mirror this var onto AI frames you should also mirror how it's used - i.e.: resetting to `0` when manually (un)locking, or popping open with tools/etc.

Paaiii: Good catch- should be sorted now!

code/mob/living/intangible/ai-camera.dm
```diff
@@ -453,6 +453,29 @@
 		if(mainframe)
 			mainframe.deploy_to()
 
+	verb/toggle_lock()
+		set category = "AI Commands"
+		set name = "Toggle Cover Lock"
+		if(mainframe)
```
TobleroneSwordfish, line 459: Why not just call the proc on the mainframe?

TobleroneSwordfish (changes requested):

code/mob/living/intangible/ai-camera.dm
Paaiii: .....very good question actually

code/mob/living/silicon/ai.dm
```diff
@@ -505,6 +508,8 @@ or don't if it uses a custom topopen overlay
 					src.dismantle_stage = 0
 				else
 					src.dismantle_stage = 1
+				if (src.locking)
+					src.locking = 0
```
frawhst, line 512: You don't need this check, `locking` is going to be `0` after either way
```suggestion
				src.locking = 0
```

code/mob/living/silicon/ai.dm
```diff
@@ -448,6 +449,8 @@ or don't if it uses a custom topopen overlay
 		if (src.dismantle_stage == 1)
 			playsound(src.loc, 'sound/items/Crowbar.ogg', 50, 1)
 			src.visible_message("<span class='alert'><b>[user.name]</b> opens [src.name]'s chassis cover.</span>")
+			if (src.locking)
+				src.locking = 0
```
frawhst, line 453: Same thing
```suggestion
			src.locking = 0
```

frawhst (commented):
Yeah I got annoyed that I couldn't do this recently.
Couple small things.

PR #15598, opened 2023-08-31
Increase blob respawn time period

FlameArrow57:
[GAME MODES][BALANCE][QoL]

## About the PR 
This PR increases the blob respawn time. Proposed time is 15 minutes, which may be a bit high, but just a number for the moment. Tried to set it at a good number where it's a bit higher than the time most blobs die, but high enough that a good blob player won't respawn.

Random event blobs have their time set just to 7 minutes.

## Why's this needed? 
Couple reasons
-Currently it's randomized between 6 and 8 minutes. Don't think randomization is good for respawn time.
-Getting blob is very rare, and often blobs will die somewhere close to 10 minutes. Think it would be nice to have a higher respawn time for how often you get it, especially with the difficulty of it
-In the case blobs die earlier, helps the round keep going a little longer instead of ending really early
-An argument could be made that respawn time shouldn't be increased, because it's a blob issue, like them not making enough special tiles. Regardless, increased respawn timer helps newer blobs and those who aren't much used to blob

## Changelog 
```changelog
(u)FlameArrow57
(*)Roundstart blob respawn time increased to 15 minutes. Random event blobs will now have a 7 minute respawn time.
```

### DIFF
--- a/code/mob/living/intangible/blob_overmind.dm
+++ b/code/mob/living/intangible/blob_overmind.dm
@@ -71,6 +71,8 @@
 	var/spawn_time = 0
 	var/respawned = FALSE
 
+	var/random_event_spawn = FALSE
+
 	var/last_blob_life_tick = 0 //needed for mult to properly work for blob abilities
 
 	var/admin_override = FALSE //for sudo blobs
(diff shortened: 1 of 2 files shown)


CheffieGithub:
I think it would be a good idea to follow flock and do it off X amount of blob tiles (total not spreads) instead of time

PR #15599, opened 2023-08-31
Adds a ranch to the afterlife bar.

cheekybrdy:
[FEATURE]

## About the PR 
Adds a ranch to the afterlife bar.
This PR also moves gemnorium V's location on Z2 to make room for it, it still functions as intended.

## Why's this needed? 
Afterlife bar has botany in it as a pastime, being a long job to use as a pastime. Adding the ranch to the list improves the activity availability there

## Changelog 

```changelog
(u)Cheekybrdy
(*)Added a ranch to the afterlife bar.
```


cheekybrdy:
!merge_upstream

UrsulaMejor:
Hey! I appreciate the mapping work, it looks good. However, chickens' exclusion from the afterlife bar (where they can be experimented on freely without intervention from extant round factors) is deliberate. Chickens have their own heaven (chicken heaven) that humans are not privy to.

If you have more ideas for chickens and the ranch, I don't want to discourage you. However, please consider pinging me in the discord @ursulamajor before starting any significant work, so that you can be sure it won't be rejected before you put in all this effort!

Thank you.

PR #15605, opened 2023-08-31
Fixes revflashes

### LINKED ISSUE #13978: Revflash conversion is... weird
### Labels

[BUG]

### Description

Converting someone with a revflash results in a series of confusing messages and the noise being played twice.

### Steps to reproduce

1. Be a head rev
2. Flash someone with the special revflash

### Expected Behavior

Probably just one of these messages

This is from one click:
image

Found while testing #13959 but I don't think it's related?

itchybones:
[Game Objects]  [Bug]
## About the PR 
Fixes  #13978 by removing the `flash_mob` override from the `flash/revolution` subtype and
moving the loyalty implant override code to the new `on_counterrev()` proc.
Thanks to @pali6 for helping me with some (90%) of the code!
## Why's this needed? 
Bugfixes are good!


CaioFrancisco:
Checking if you're a sub-type in the parent itself is usually bad practice. You could probably move the whole `convert` proc to the revflash sub-type.

code/obj/item/device/flash.dm
```diff
@@ -316,6 +316,8 @@ TYPEINFO(/obj/item/device/flash)
 			if (!nostun)
 				M.apply_flash(1, 2, 0, 0, 0, 0, 0, burning, 100, stamina_damage = 210, disorient_time = 40)
 
+obj/item/device/flash/proc/on_counterrev(mob/living/M, mob/user)
```
frawhst, line 319:
```suggestion
/obj/item/device/flash/proc/on_counterrev(mob/living/M, mob/user)
```
tiny style thing

frawhst (approved):

PR #15612, opened 2023-08-31
Organise humanoid critters into their own DMI files

CheffieGithub:
## About the PR 
Organises humanoid critters into `icons/mob/critter/humanoid` though this is very broad so I would like feedback on whether the folder name should be changed or subdivided etc.
Currently this encompasses everything with human shape so most bipedial critters. But some of these are strange cases like shades and bears.

## Why's this needed? 
Easier to locate specific icons with everything in one place. Better for performance.


frawhst:
imo `humanoid` should specifically be for things that look basically like humans- zombies/fire elementals/skeletons/etc all make sense, but bears and brullbars aren't very human-like other than matching the silhouette. Micromen and mini-syndies are also iffy, since they look vaguely like a human, but not an in-game human.

ZeWaka (approved):

DisturbHerb (approved):

ZeWaka:
this is better than existing so yeah

CheffieGithub:
Ill probably move around the brullbars and bears later when I decide where they should be

PR #15616, opened 2023-09-01
Removes the old tracking implant kits

Garash2k:
## About the PR 
Replaces `trackimp_kit`s content with `trackimp_kit2`s. Deletes the old version.

## Why's this needed? 
`trackimp_kit2` is great, it's the tracking box we all know and love what with GPSes in them.
`trackimp_kit` is some ancient tech that was left over on some maps, notably oshan. I don't think is worth keeping around, so let's replace them and delete the code!

`trackimp_kit` UIs:
261920295-34604899-3873-463f-bdde-4f4290a9715a
261920297-6f5a0f2e-a8d8-4687-98ec-d0f873743d22

As a refresher: `trackimp_kit2` box contents:
7srg58

[Game Objects][Mapping][Code Quality]


flappybatpal:
If `trackimp_kit` is being removed why not take it's name as well? It's awkward to have a 2 version of something with no other version.

PR #12728, opened 2023-01-13
Cable Network Colour Coding

TheTyrant25:
[MAPPING] [SYSTEMS] [QOL]
## About the PR 
So the cables that connect the TEG to the SMES units on cogmap 1 are brown, while the rest are red, and having some snazzy yellow solar cables. Donut2 does a similar thing with the brown, plus black cables leading to the PTL.

This PR applies this colour scheme to almost every other (in rotation) map, for consistency and style reasons:
- Brown for engine connections
- Black for PTL cabling
- yellow on solars
- orange on nadir's nuclear engine because kubius said so
- red everywhere else

Oshan is not included, though, as they have their reinforced cables for undersea hotspot thingys.
## Why's this needed? 
Lets you see at a glance which cable network is which. Plus, it's weird that only two maps have it that way, might as well.

## IN ADDITION
I have removed every instance of d1 and d2 usage in every cable across the rotation maps. The regex used is in the conversation, and it's quite handy.
## Changelog 
```changelog
(u)Tyrant
(+)Engine room cables are now brown, solar cabling is now all in a consistent yellow, and PTL cables are black across all maps.
```

PR #15619, opened 2023-09-01
Power Network Colour Coding

TheTyrant25:
[MAPPING] [QOL]
## About the PR
Recolours several wires based on power net connections:
- red is for station grid
- orange is for main power source
- yellow is for solars (or catalytic generators)
- blue is for the cable connecting to the PTL
The only exception to this is the AI satellite on donut2, which has blue wiring instead of red.
This is also applied to various syndicate listening posts and offstation things.
## Why's this needed?
Did this once before in #12728 , but for some reason all the cables on kondaru are back to being red. After gathering some opinions, brown for engine grid and black for ptl works really poorly with the T ray scanner, as the cables are dark and hard to spot. That is why brighter colours, orange and blue, have been chosen.

Colour coding will make hotwiring the engine a bit more obvious and intuitive (you can visually see that it's two separate networks connected by only SMES), and also easier to avoid doing by accident.
## Changelog 
```changelog
(u)Tyrant
(+)Cables are colour coded in the following way: Red=Station, Orange=Engine, Blue=PTL, Yellow=Solars
```


TheTyrant25:
as a big ol mapping pr, this will merge conflict a lot.

TobleroneSwordfish:
This looks great, from scanning through the diffs it all looks good, although did you mean to remove this cable from the cog 1 engine choice prefab?
image

TheTyrant25:
yeah, i changed the way that the ptl line connects to the engine grid, so it didn't make sense to still have the dangling cable there. It's hard to see on the other engine prefabs because of the glass, but i removed the ptl cables there too

TheTyrant25:
this might merge conflict badly but hopefully it's solvable. gonna run the driver on it

PR #15622, opened 2023-09-01
Give mining more room + oxygen on Clarion

MeggalBozale:
[QOL] [MAPPING] 

## About the PR
Changes Clarion mining to be more roomy, and adds some objects commonly found on other maps.
- Expands the department up two tiles to make room for more things.
- Removes a gas fabricator from QM and moves around a bunch of crates to make room for mining.
- Removes the nanofab (WHY do they have this) from QM as well to make room for the water cooler, moved to refinery instead.
- Adds an oxygen canister & tank dispenser to mining, gives them a breakfast table, and overall rearranges their room to look and feel nicer.

## Why's this needed?
Cramping miners and making them find a different source of oxygen isn't fun for them, especially on a map which always has a reactor, which needs mining's cooperation for good output.

## Changelog
```changelog
(u)MeggalBozale
(+)Miners on Clarion now have more space in their lobby.
```


Sord213 (changes requested):
PR looks good, but I got a couple requests.
Screenshot_469
Cyan highlight: add a forcefield to this airlock. Yellow highlight: add a mining access spawner on top of this door so mining or QM can use it. Red highlight: this wiring...bad. no wires in walls wherever possible. Please route it through the warehouse
Screenshot_472
Plating under windows, weird gray corner tile
Screenshot_473
Could you move the ship component fab into the hangar somewhere? It makes more sense being in there.

CheffieGithub:
After moving the ship part fabricatior I would move the ore magnet to its place or move the reclaimer over and have it there, just so its not blocking the telepad

MeggalBozale:
Alright gocha, a lot of other mining departments have tank storage units so the reasoning was merely 'they do it, why not here'

PR #15623, opened 2023-09-01
Unify Fire alarm offsets

TheTyrant25:
[MAPPING]
## About the PR
Replaces nearly every fire alarm in every rotation map with the new subtypes with built in pixel offsets. Doesn't change any that have both pixel x and pixel y, unless one is unneeded.

## Why's this needed?
A lot of southward alarms had the south facing sprite instead of the north facing. Similarly a lot (if not all) of east facing alarms used the west dir instead, as they sat on a west wall. Easy mistake to make, given that the "right way up" isn't terribly obvious. Those should be fixed now.
Also, there are multiple different offsets for the same direction, so the var edited pixel offsets could have multiple different ones. Now they are more uniform. Which I feel is better.
Also, by using subtypes instead of var editing, we have saved 3,283 lines of dmm files. Minor but still valid.

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -242,9 +242,7 @@
 "abd" = (
 /obj/stool/chair/comfy,
 /obj/landmark/start/job/detective,
-/obj/machinery/firealarm{
-	pixel_y = 25
-	},
+/obj/machinery/firealarm/north,
 /turf/simulated/floor/carpet/grime,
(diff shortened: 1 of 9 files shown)


flappybatpal (approved):
After inspecting all the changes this looks correct. The two discrepancies in cogmap2 and donut2 look to just be the map groups changing

PR #15625, opened 2023-09-01
remove multitool duplicated file

TheTyrant25:
## About the PR
The code for the multitool was for some reason smeared across two different dm files in two different folders. This removes one and moves its code to the other

## Why's this needed?
Because having the code being duplicated while also having part of the code in one and not the other is horrendous and should be made illegal immediately.


code/obj/item/device/multitool.dm
```diff
@@ -1,30 +1,96 @@
 TYPEINFO(/obj/item/device/multitool)
 	mats = list("CRY-1", "CON-2")
 
+// this was in code/obj/item/tool/multitool. i don't know which to keep so i'm leaving it in.
+TYPEINFO(/obj/item/device/multitool)
+	mats = 6
+
```
TobleroneSwordfish, line 7: the other one is the definition that's used I think
```suggestion
```

TobleroneSwordfish (changes requested):

PR #15628, opened 2023-09-01
Add more ashtrays and cigarette vendors to bar.

TheTyrant25:
[MAPPING] [QOL]
## About the PR
It was noticed that the donut2 bar doesn't have an ashtray nor a nearby cigarette vendor. I decided to make sure every bar had an ashtray, and a cigarette machine reasonably close by.

## Why's this needed?
Let smokers rp. Also map consistency is nice.

## Changelog
```changelog
(u)Tyrant
(+)All bars now have an ashtray and a nearby cigarette vendor.
```

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -7170,7 +7170,6 @@
 	dir = 4
 	},
 /obj/decal/stage_edge,
-/obj/machinery/vending/cigarette,
 /turf/simulated/floor/wood,
 /area/station/crew_quarters/bar)
 "aCv" = (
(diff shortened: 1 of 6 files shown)


flappybatpal:
I'm not super against it

flappybatpal (changes requested):
- I don't think the one on Oshan is required. Oshan has vending machine clusters in corridors and there is one very close right now.
- Don't think the one on Clarion is required for similar reasons.
- Nadir is OK if the existing machine is relocated up near medical.

Clarion has one outside the bar pretty close:
image

Nadir has one 21 tiles away from the new machine, against fairly close:
image

Oshan is very close
image

TheTyrant25:
my eyes apparently glossed over those machines, whoops

flappybatpal (approved):

PR #15630, opened 2023-09-01
Reworks chest organ surgery

Bartimeus973:
[INPUT]
## About the PR 
Reworks organ surgery in it's entirety and makes it behave in the following way. This does not affect head surgery or limb surgery in any way.

When operating on a subject, begin with a cutting tool, then a snipping tool. This will open the chest cavity. Clicking on a patient with an open chest cavity while holding a cutting tool, snipping tool, sawing tool or an empty hand will display the following interaction wheel:

smallwheel

Clicking on one of the organ icons will then prompt a list of required steps to proceed with the surgery, similarly to deconstructing a device, depicted here:

step_surgery

To remove a heart or lungs, the surgeon must first use a saw on an opened up patient which prompts a 4 second action bar (2 second with medical training). They may then access the heart and lungs.

bigwheel

Organs may also be ripped out with an open hand. This begins a 10 second action bar which does continuous moderate damage to the patient and the organ. On completion, this does a chunk more damage and puts the organ in the surgeon's hand.

Implant and parasite removal surgery is now redone and requires opening up a patient. If an implant or parasite is inside someone, a context action bubble will appear alongside the other organ choices.

Chest item surgery is redone. To put an item inside someone, open them up with the usual knife -> scissors steps, then click drag any item to the patient. If the surgeon is holding a suture at the moment of the click drag, the item is automatically sewn in. It may also be sewn in afterwards with a suture on grab intent.
To remove a chest item, you must open up the patient and a context action bubble will appear alongside the other organ choices.

chest_item

Operating on disarm intent will now always cause an intentional slipup.

Operating on harm intent will now always attack the patient.

Butt/Tail surgery is now done with scalpel -> saw -> scalpel -> saw on grab intent on the patient. The surgeon may pick to remove either the tail or the butt.

Hemostats have been reworked. You may now use an hemostat on someone in surgery conditions. If they are bleeding, the hemostat cures it. If their chest is open, the patient's blood vessels are clamped. A patient with an open chest wound and un-campled blood vessels will bleed heavily forever until the use of a hemostat or until the wound is sutured.

Todo:
- [x] Rewrite relevant parts in the surgery handbook
- [x] Finish up suturing
- [x] Finish up cauterizing
- [x] Cleanliness

As an alternative to the multiple steps surgery, action bars may instead possibly be used as depicted here:

actionbarsurgery

Feedback would be greatly valued.

## Why's this needed? 
This PR aims to provide better visual feedback, streamline the surgery process and fix the following "issues":

- In order to remove an organ, a series of complex tool usage was needed in a particular order. A single misstep meant suturing up the patient and starting over. By streamlining the process of opening up a patient and providing clear feedback of what organ is targeted, surgery shouldn't require the wiki open to remove specific organs.

- Once an organ was removed, the surgeon had to suture up the patient before removing another which made little sense. This PR allows multiple organs to be removed in succession without having to re-suture between every step.

- Bundles up tail and butt surgery together as neither should require an open chest to be removed and provides better visual feedback

- If a patient needed any surgery and they had an implant, the surgeon had to remove every single implant in order to access the organs, then put them back in. This PR offers the option to remove the implants at will and remove the hurdle.

- Offering a way to mess up the surgery on purpose and rip out an organ by hand provides antagonists with a few fun ways to cause more chaos.
 
- Patients can currently walk around with a gaping chest wound with no consequences. Slowly bleeding out while their wound is un-tended or un-clamped should provide for an interesting alternative

## Changelog 

```changelog
(u)Bartimeus973
(*)Organ surgery is now redone! Instructions are found in the surgery handbook. For more details click on the PR number in the top right!
```


DisturbHerb:
should probably edit the surgical textbook accordingly

Kyle2143 (changes requested):
Mirroring my feedback from the forum here:

This is something I considered in the past as well. Having to memorize which tool to use to get to the right organ has always been a subpar system. But it was the solution to path routing without trivializing how you perform surgery by just making a popup box that you select which organ you wanna remove and then holds your hand down the path.

I'd want to meet in the middle. Remove the initial radial menu on surgery start that shows every organ that can be selected. Instead, have it let you target regions of the body:
- Ribs(and organs inside them),
- Subcostal(below ribs, right or left for liver, spleen, pancreas),
- Abdomen/navel(intestines/stomach),
- Flanks, left or right(kidneys)

Then take it from there with zero-ing in on an organ with the "ideal" surgery tool.

If the popup part of this was implemented as-is, I'd see it as an attachment to the surgery computer or medical training instead of everywhere. Since it's more like a tool to teach you than a new system.

The other parts of this do sound like they will solve a lot of issues like the implants, suturing and pathing. Which is sorely needed and am glad to see addressed.

Bartimeus973:
Taking into account the feedback, i went and implemented the following:

Upon opening up a patient, you are now prompted with four "regions". To open up those regions, you must use different tools. Once a region is opened up, you may freely examine and take out the organs in that specific region. Suturing up a patient closes that region, which means that to get access to the patient's organs again you have to re-open that specific region.

Here is an example of the current way things work (not fully finished and icons are placeholders): 

reworked

Bartimeus973:
Here is the latest implementation, showcasing the current region and organ icons and the way their color updates to reflect how advanced a surgery is.

newest

code/obj/item/organ_holder.dm
```diff
@@ -47,6 +47,33 @@
...
+	///How cut up is our back for surgery purposes
+	var/back_op_stage = 0
+	///How cut up are our ribs? 0 = Not. 1 = Halfway. 2 = Open.
+	var/ribs_stage = 0
+	///How cut up is our subcostal region? 0 = Not. 1 = Halfway. 2 = Open.
+	var/subcostal_stage = 0
+	///How cut up is our abdominal region? 0 = Not. 1 = Halfway. 2 = Open.
+	var/abdominal_stage = 0
+	///How cut up are our flanks? 0 = Not. 1 = Halfway. 2 = Open.
+	var/flanks_stage = 0
```
TobleroneSwordfish, line 76: these should be defines

code/obj/item/organ_holder.dm
```diff
@@ -57,6 +84,258 @@
...
+
+		//begin by adding regions
+		var/datum/contextAction/surgery_region/ribs/ribs_action = new /datum/contextAction/surgery_region/ribs()
+		switch (src.ribs_stage)
+			if (0)
+				ribs_action.icon_background = "bg"
+			if (1)
+				ribs_action.icon_background = "yellowbg"
+			if (2)
+				ribs_action.icon_background = "greenbg"
```
TobleroneSwordfish, line 101: this should be abstracted, probably into the surgery_region contextAction parent

TobleroneSwordfish (changes requested):
Had a brief look at the code, will require more thorough review probably though.
Also the hemostat clamping should have an action bar.
The butt surgery feels really janky, like it's split between two different systems.

Bartimeus973:
I implemented most of the review changes.
May you please elaborate a little bit on the "butt surgery being janky"? I'll try my best to change the system and make it more consistent.

TDHooligan:
looks good but strongly pushing for the hemostat automation returning. removing it isnt adding anything besides a chore that can fail due to lag. 
i dont think it requires being 'more conscious' and the monotony of doing the exact same action every time you do anything will only rarely trip anyone up. It's more likely to catch people misclicking or lagging. which isn't what you want to punish.

Bartimeus973:
The idea was that you only needed to clamp the bleeders once to stop the perpetual bleeding but since doing surgery will cause some amounts of regular bleeding i'll change it to meet halfway. Hemostats will now behave in the following way:

Doing surgery with a hemostat in your off hand will stop all bleeding as it did before. Which means doing a surgery step and succeeding it will cure current bleeding.
Additionally, at any point during the surgery, the surgeon can click the patient with the hemostat to truly "clamp the wound" which will stop spontaneous bleeding. This only needs to be done once. The idea being that if the surgeon doesn't use the clamps once and lets the person go without suturing up their chest wound they will slowly bleed to death. This has practical uses for antagonists as well.

PR #15631, opened 2023-09-01
Burner / damager helpers with just the helper this time

CheffieGithub:
## About the PR 
Adds two new helper objects a turf burner and a turf damager these use `break_tile()` `burn_tile()` and work on a tile as long as its a floor and the `can_break` / `can_burn` variables are `TRUE`.
These have logging for when placed on a tile they can't act on.

 `break_tile()` `burn_tile()` `restore_tile()` moved up to /turf along with related variables.

## Why's this needed? 
image

Map editor view from left to right:

The damaged subtypes only look like default floors / plating.


MeggalBozale:
image
Abstract types are showing up in strongDMM

MeggalBozale (approved):
LGTM hopefully future PR where the ugly damaged turf subtypes get axed

frawhst (approved):
Seems good- one concern, does removing the 'broken' and 'burnt' vars mess up the behavior of crowbarring up broken tiles, or is that tracked in some other way?

MeggalBozale:
To address the concern - No, prying up broken tiles works perfectly fine.

CheffieGithub:
Yeah they aren't removed they are just up on turf now

PR #15632, opened 2023-09-01
Fix tiny bug

amylizzle:
## About the PR 
This should be 0, -1 is beyond the end of the string and not supported by the proc.


amylizzle:
wait, this is wrong

PR #15637, opened 2023-09-02
Remove some unused options from security weapons vendor

FlameArrow57:
[GAME OBJECTS][REMOVAL]

## About the PR 
This PR just removes some options that are very rarely bought from the security weapons vendor.

## Why's this needed? 
They are almost never bought, it just tends to be that the other utlities are more worth spending the points on.

Also,
Forensics scanner - Security PDAs already have enough forensics capability for nearly all forensics needs
Paint marker rounds - Need to land a direct hit to help track a target when they're escaping. Almost all the time, the target escapes and you can't track them already, or the AI can track them for you
RecordTrak - These already lie around security, there's never a need to buy a new one

Not really a situation warranting getting a security token from the armory just to get one of these things

### DIFF
--- a/code/obj/submachine/weapon_vendor.dm
+++ b/code/obj/submachine/weapon_vendor.dm
@@ -124,10 +124,7 @@
 		materiel_stock += new/datum/materiel/utility/morphineinjectors
 		materiel_stock += new/datum/materiel/utility/donuts
 		materiel_stock += new/datum/materiel/utility/flashbangs
-		materiel_stock += new/datum/materiel/utility/detscanner
 		materiel_stock += new/datum/materiel/utility/nightvisionsechudgoggles
-		materiel_stock += new/datum/materiel/utility/markerrounds
-		materiel_stock += new/datum/materiel/utility/prisonerscanner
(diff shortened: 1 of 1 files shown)


Piesuu:
I don't really see a reason for this, especially because I use the forensic scanner myself. there really isn't a visible need for less security options

DisturbHerb:
I use all of these. I don't believe they are worth removing from the vendors simply because they are under-utilised.

Sord213:
Yeah some of these do still get used. While they're not super popular, I wouldn't want to see them gone.

MeggalBozale:
I buy forensics scanner / recordtrak from time to time. I don't play security much, but I do. I appreciate having the options around on occasion for easily setting people off arrest, and in case I don't wanna rely on someone else to run fingerprints.

Cherman0:
While I feel that removing the forensics scanners from security would be an overall positive for game balance and help to define the detective as something other than gunsec, it doesnt seem proper to push it under the guise of culling bloat. At least on RP I'd wager that the forensics scanner is the most picked option.

FlameArrow57:
Sorry, maybe a bad take PR >.<

Agree not the best

PR #15644, opened 2023-09-02
Flipping as a borg with an open head will launch your brain

unfunnyperson:
[SILICONS]

## About the PR 
Using the flip emote as a borg with an exposed brain will give the "Loose Brain" status effect for 2 minutes. Flipping again will fling the user's brain.

image
image

## Why's this needed? 
image

I personally think this is funny and its a cool way to suicide

## Changelog 
```changelog
(u)UnfunnyPerson
(+)Flipping with an exposed brain as a cyborg will give the "Loose Brain" status effect. Flip with the status effect to launch your brain.
```


TheColorCyan:
funny

wolfolotl:
too easy to do accidentally. to me flipping is on the same level as jumping near someone in a mmo where it's a form of communication; I can easily imagine many borgs ejecting their brains when they don't mean to because someone flipped at them at a critical moment and on instinct they just flipped back

unfunnyperson:
> too easy to do accidentally. to me flipping is on the same level as jumping near someone in a mmo where it's a form of communication; I can easily imagine many borgs ejecting their brains when they don't mean to because someone flipped at them at a critical moment and on instinct they just flipped back

I dont think this is an issue because If your brain is exposed whoever you flipped at can very easily plop it back in on an accidental flip

Connie-Dun:
I enjoy this change and am glad it has been made (potentially) real! When I called my idea "bad" in the screenshot it was pretty much a shorthand for "silly idea I'm not certain people would like" more so than an actual value judgement. I understand the concern, but borg heads rarely ever end up unlocked unless it's explicitly about to be de-brained, so I think it's moreso a "funny thing you can intentionally do as a gag" than "sudden lethal thing to happen out of the blue" - plus, as mentioned, it's very easy for someone to just pop it back in if it was an accident.

Glamurio:
Eh, I'm against this one as well. Flipping is a way to communicate and show appreciation. I do it all the time subconsciously, to the point where I consistently trip as a clown or attack people when drunk, especially since flipping is somehow more responsive than emoting (different issue).

I realize the situation where your head is open might not occur easily, but depending on your outfit and headgear, you might not even REALIZE your head is open because of rendering. And then you flip and die. Possibly next to a roboticist who has no idea what they're doing. And then you're dead for the round as this character.

TobleroneSwordfish:
Yeah as funny as this is, the only thing I can see it causing is occasional accidental round-removal

444Portal:
Very funny, but I tend to flip without thinking about it. Maybe after multiple flips, with a easy to see (as in, red) message each time?

unfunnyperson:
adjusted to take 4 flips so this will be a purely intentional thing

code/mob/living/silicon/robot.dm
```diff
@@ -70,6 +70,8 @@
 	var/alohamaton_skin = 0 // for the bank purchase
 	var/metalman_skin = 0	//mbc : i'm getting tired of copypasting this, i promise to fix this somehow next time i add a cyborg skin ok
 	var/glitchy_speak = 0
+	///how many flips with our head exposed have we done?
```
ZeWaka, line 73: I don't really like this solution. If I flip 4 times in a round, blammo.
I'd prefer a status effect or something that is like "loose brain".

frawhst:
Let's try it and see what happens

unfunnyperson:
updated to use a 2 minute long status effect, considering the status effect is pretty noticeable you only need to flip once with the status effect to fling your brain

PR #15662, opened 2023-09-03
allow matsci procs to use both typepaths and matid

TheTyrant25:
[INTERNAL] [QUALITY] [MATERIAL]
## About the PR
set up a second material cache, indexed by type instead of by `getid()`. They're essentially the same list though. Renamed the original `material_cache` to `material_cache_matid`
let `get_material()` use typepaths.
let `set_material()` use matids in a horrible, no good way.

## Why's this needed?
More uniform code is better imo. It's weird that you have to use get_material inside of set_material. Second step is to switch everything over to using typepaths instead.

### DIFF
--- a/code/modules/admin/atom_verbs.dm
+++ b/code/modules/admin/atom_verbs.dm
@@ -45,11 +45,11 @@ var/global/atom_emergency_stop = 0
 		if (!sleep_time)
 			return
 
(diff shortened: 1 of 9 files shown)


TheTyrant25:
actually this is really really far from finishing what am i doing

PR #15664, opened 2023-09-03
makes materials use paths instead of strings

TheTyrant25:
[INTERNAL] [QUALITY]
## About the PR
Makes `getMaterial()` use both type paths instead of mat_id strings. Changes the `material_cache` to be indexed by typepath instead of matid.

An attempt was made to remove most instances of `.getID() == "string"`. Some uses of `.getID()` may still linger, but that code should still function in theory.

All instances of Get_material() and atom.default_material have been swapped out to use typepaths instead. By hand. I hope you appreciate my efforts. At the very least, give me an `E-FUCK` label. I feel I've earned it.

## Why's this needed? 
Using strings in getMaterial() always felt really arbitrary to me, similar to magic numbers. Typepaths are better, and the typepath autofills as you type and can be double checked by ctrl clicking in vsc to find the material itself.
The alternative idea proposed was material id defines. This is better.


code/modules/materials/Mat_ProcsDefines.dm
```diff
@@ -1,5 +1,8 @@
-
-var/global/list/material_cache
+// THESE TWO ARE THE SAME LIST
+/// the material cache, indexed by material id
+var/global/list/material_cache_matid
```
amylizzle, line 3: why both? It seems like it'd be better to just have it be by type

code/world/initialization/0_preMapLoad.dm
```diff
@@ -217,12 +217,14 @@
...
-	material_cache = list()
+	material_cache_matid = list()
+	material_cache_type = list()
 	var/materialList = concrete_typesof(/datum/material)
 	for(var/datum/material/mat as anything in materialList)
 		if(initial(mat.cached))
 			var/datum/material/M = new mat()
-			material_cache[M.getID()] = M.getImmutable()
+			material_cache_matid[M.getID()] = M.getImmutable()
+			material_cache_type[M.type] = M.getImmutable()
```
amylizzle, line 227: If you are doing it by both, the second `getImmutable()` will create a second copy. Better to do this instead.
```suggestion
			material_cache_type[M.type] = material_cache_matid[M.getID()]
```

code/modules/materials/Mat_ProcsDefines.dm
TheTyrant25: there's cases where i need it to still use id (for now)
just being safe. Ideally i'll remove it later

TheTyrant25:
this has very quickly exploded in number of files changed. Who knew so many things used materials huh?

code/modules/materials/Mat_ProcsDefines.dm
TheTyrant25: okay i've decided to rip out the string matids at the cost of my sanity. I hope you are happy.

code/world/initialization/0_preMapLoad.dm
TheTyrant25: I did this and then it turned out to no longer be needed since i ripped out the string matids.

TheTyrant25:
that feeling when after a whole 6 hours of coding the checks pass

TheTyrant25:
FUCK i need to do this to chems too

TheTyrant25:
ah whatever

TheTyrant25:
Oh yeah this will break secret submodule stuff. Hm

code/obj/item/rcd.dm
```diff
@@ -133,7 +133,7 @@ TYPEINFO(/obj/item/rcd)
 	var/list/datum/contextAction/contexts = list()
 
 	get_desc()
-		. += "<br>It holds [matter]/[max_matter] [istype(src, /obj/item/rcd/material) ? material_name : "matter"]  units. It is currently set to "
+		. += "<br>It holds [matter]/[max_matter] [istype(src, /obj/item/rcd/material) ? material_type : "matter"]  units. It is currently set to "
```
amylizzle, line 136: this one should probably be material name

code/obj/item/rcd.dm
```diff
@@ -932,20 +932,20 @@ TYPEINFO(/obj/item/rcd/material/cardboard)
 			boutput(user, "Recycling [W] just doesn't work.")
 		else if (istype(W, /obj/item/paper/book))
 			matter += 5
-			boutput(user, "\The [src] recycles [W], and now holds [src.matter]/[src.max_matter] [material_name]-units.")
+			boutput(user, "\The [src] recycles [W], and now holds [src.matter]/[src.max_matter] [material_type]-units.")
```
amylizzle, line 935: ditto for these

code/obj/machinery/manufacturer.dm
```diff
@@ -624,11 +624,11 @@ TYPEINFO(/obj/machinery/manufacturer)
 				if (src.mode != "ready")
 					boutput(usr, "<span class='alert'>You cannot eject materials while the unit is working.</span>")
 				else
-					var/mat_id = href_list["eject"]
+					var/mat_type = href_list["eject"]
```
amylizzle, line 627: have you tested this? might need text2path or whatever it's called

amylizzle (changes requested):
looking pretty good

code/obj/item/rcd.dm
TheTyrant25: image
you have a point

TheTyrant25:
i have now found that this has broken a lot of the material loading system of manufacturers
and im not sure how to fix it

TheTyrant25:
giving up on this for now. Too lazy to fix merge conflict plus other irl things. Free for adoption, but if no one picks it up, i'll reopen it one day.

PR #15678, opened 2023-09-04
Fixes changelings being able to sting themselves

TealSeer:
[PLAYER ACTIONS] [BUG]
## About the PR 

Goddamn it I missed a check when I was rewriting sting code

## Why's this needed? 

Bug

### DIFF
--- a/code/modules/antagonists/changeling/abilities/stings.dm
+++ b/code/modules/antagonists/changeling/abilities/stings.dm
@@ -53,6 +53,8 @@
 			if (!MT.reagents)
 				boutput(holder.owner, "<span class='alert'>That does not hold reagents, apparently.</span>")
 				return 1
+			if (target == holder.owner)
+				return 1
 			// make some room in the target
 			if (MT.reagents.total_volume + toxin_holder.total_volume > MT.reagents.maximum_volume)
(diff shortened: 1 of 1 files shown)


444Portal:
Wait, this isn't intentional? Personally don't see why it should be removed :p

Glamurio:
In all seriousness, if it wasn't intentional, it absolutely SHOULD be. The ability to sting yourself is a fantastic way to pretend that someone else is a changeling in an RP way and get the heat off you.

Piesuu:
I'd say a popup: "We are going to sting ourselves." "Are We sure?" like some of the other changeling abilities would work or  stinging yourself (at least on RP) could inject you with capulletium plus however since there isn't a safe neurotoxin equivalent for classic one would have to be made. I have stung myself in a combat situation while trying to use the sting offensively and it was not a fun experience.

Connie-Dun:
Popup sounds good. The main reason for this fix is because there has been some buggy behavior with self-stinging, where clicking besides directly clicking yourself could end up making you sting yourself. Obviously, stinging yourself when you don't want to is extremely terrible.

TealSeer:
I'd consider making self stings a feature but definitely after this PR cause this is just to fix an inconsistent buggy behavior not remove a feature

PR #15687, opened 2023-09-04
Introduction of the assembly component and refactoring of pipeframe assemblies

### LINKED ISSUE #14525: Igniter/signaller assemblies do not combine with pipebombs
### Labels

[BUG][MAJOR]

### Description

The stage of crafting where you add the igniter/signaller to the pipebomb if you want a ranged pipebomb seems to not work.

### Steps to reproduce

1. Start making a signaller pipebomb
2. Finish steps up to 7 from the construction wiki
3. Cannot do step 8

### Expected Behavior

Assembly should add to the pipebomb for a signaller pipebomb

Reported by: CalliopeSoups
Active test merges: nothing
Round log date: 2023-06-10-23-32
Reported on: 2023-06-11 00:17:55
Map: COGMAP

M-Earthfire:
[Game-Objects][Code-Quality][QoL]

## About the PR 

This PR introduces `/datum/component/assembly`. It can be added to an atom to call a proc on the atom whenever another atom that is specified in the component is used on it. Optionally, the same proc can be called when the atom the component is added on is used on that specified component.

The component takes the following arguments in the order given here:

- `to_combine_item`: this specifies which items will trigger the interaction defined in `proc_to_call`. This field works with 3 different input types: if a path is given, it checks for this atom path. If a list is given, the component checks against each path within the list. If a tool flag, e.g. `TOOL_CUTTING`, is given, it checks for the corresponding tool type.
- `proc_to_call`: This takes a `PROC_REF` to call when the interaction goes off. Make sure this proc returns TRUE if the assembly was successfull and you don't want the attack to go through
- `on_tool_attack`: If set to TRUE, the interaction will work if either the component-atom is used on a compatible atom or the compatible atom is used on the atom with the component. Set this to FALSE if you want interactions like tools being used on the atom in question.

To demonstrate the use of this component, all assembly steps of the pipeframe were refactored to use this component. This includes many of the pipeshot steps and the pipebomb assemblies. 

Since this PR makes assemblies by hitting cabled frames with igniter-assemblies and vice-versa, the standard pipebomb now requires a timer instead of a timer/igniter-assembly.

This PR fixes #14525 

Also, this PR fixes a bug where pipebomb assemblies made out of wired frames didn't carry special items in them, but pipebomb assemblies made out of standard pipebombs did.

## Why's this needed? 

having to juggle multiple states on assembly process is very cumbersome and requires escalating if-trees. Splitting the steps off by a component enables to just add the new steps and remove the old ones from the item once it has reached a new state. This clear up the code. Also the two-side compability of the component enables to add the component once to the parent that should carry the logic instead of having to copy the assembly step on multiple items.

The refactor of the pipebomb assembly steps serves as an example of it's use. I tried to hold myself back and not make the PR explode too much in scope, since working on pipebomb-, mousetrap- or igniter-assemblies made me scream on multiple occassions. E.g. complete refactor of igniter-assemblies is out of scope of this PR, although the integration of the component helped clean it up.

On the player side, not having to specifically hit the pipebomb with the assemblies helped clear up some confusions and makes construction less of a hassle. The drawback was the necessity to move the standard pipebomb to use a timer.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Pipebomb- and beaker-assemblies should be more intutive to create now.
(+)Standard pipebombs are now created by using a timer instead of a timer/igniter-assembly with a pipebomb frame.
```


code/obj/item/assembly/misc_assemblies.dm
```diff
@@ -653,6 +702,94 @@ Contains:
 		src.name = "Radio/Igniter/Beaker Assembly"
 	return
 
+// proximity-ignite-assemblies
+//Oh for fucks sake, why the fuck aren't these under a single parent? I have added these assembly procs three times already.
```
amylizzle, line 706: if you're already rewriting them, why not make this change?

code/obj/item/assembly/misc_assemblies.dm
```diff
@@ -229,8 +180,102 @@ Contains:
...
+	manipulated_beaker.layer = initial(manipulated_beaker.layer)
+	user.u_equip(manipulated_beaker)
+	manipulated_beaker.set_loc(src)
+	src.c_state(0)
+	boutput(user, "You attach the timer/igniter assembly to the beaker.")
+
+	//Since we completed the assembly, remove all assembly components
+	var/list/assembly_component_list = src.GetComponents(/datum/component/assembly)
+	for (var/datum/component/component_to_remove in assembly_component_list)
+		component_to_remove.RemoveComponent()
```
Tarmunora, line 211: Feel like there should be some more elegant way of handling this than this, considering how it seems to be being universally applied 🤷

Tarmunora (approved):

code/datums/components/_component.dm
```diff
@@ -433,6 +433,18 @@ var/datum/signal_holder/global_signal_holder
...
+/**
+  * Calls RemoveComponent on all components of a given type that are attached to this datum
+  *
+  * Arguments:
+  * * c_type The component type path
+  */
+
+/datum/proc/RemoveComponentsOfType(c_type)
+	var/list/component_to_remove_list = src.GetComponents(c_type)
+	for (var/datum/component/component_to_remove in component_to_remove_list)
```
ZeWaka, line 445:
```suggestion
	var/list/datum/component/component_to_remove_list = src.GetComponents(c_type)
	for (var/datum/component/component_to_remove as anything in component_to_remove_list)
```
i assume this is true

code/datums/components/assembly_comp.dm
```diff
@@ -0,0 +1,62 @@
...
+
+///This component calls a procref with a assembly_information string on the atom it was added to when it gets attacked with an object specified in the to_combine_item
+///This component can be used to get the convulent if-then-else trees of assemblies under control
+///Make the proc you call on sucessfull assembly return TRUE. Else the attack will go through!
+
+/datum/component/assembly
+	dupe_mode = COMPONENT_DUPE_ALLOWED //We want to have multiple differnt assembly types on an item for different assembly steps
+	///The item(s) that are used to combine with the atom this component is attached to
+	///use a tool-bitflags like TOOL_CUTTING, if you want to go with groups of tools
+	var/to_combine_item = null
```
ZeWaka, line 16:
```suggestion
	var/atom/to_combine_item = null
```

code/datums/components/assembly_comp.dm
```diff
@@ -0,0 +1,62 @@
...
+	if(on_tool_attack)
+		RegisterSignal(src.parent, COMSIG_ITEM_ATTACKBY_PRE, PROC_REF(on_pre_attack))
+
+/datum/component/assembly/UnregisterFromParent()
+	. = ..()
+	UnregisterSignal(src.parent, COMSIG_ITEM_ATTACKBY_PRE)
+	UnregisterSignal(src.parent, COMSIG_ATTACKBY)
+
+
+/datum/component/assembly/proc/on_pre_attack(var/atom/affected_parent ,var/atom/to_combine_atom, var/mob/user, var/damage)
```
ZeWaka, line 38:
```suggestion
/datum/component/assembly/proc/on_pre_attack(var/atom/affected_parent, var/atom/to_combine_atom, var/mob/user, var/damage)
```

code/datums/components/assembly_comp.dm
```diff
@@ -0,0 +1,62 @@
...
+			if(istype(checked_atom, to_checked_type))
+				is_combineable = TRUE
+	//if it is a number, it most likely is a bitflag we passed
+	else if (isnum_safe(src.to_combine_item))
+		if (istool(checked_atom, src.to_combine_item))
+			is_combineable = TRUE
+	else if (istype(checked_atom, src.to_combine_item))
+		is_combineable = TRUE
+
+	if(is_combineable)
```
ZeWaka, line 59:
```suggestion
	var/is_combinable = FALSE
	//if to_combine_item is a list, we look if we find the item in there
	if (islist(src.to_combine_item))
		var/list/combinable_items = src.to_combine_item
		for(var/to_checked_type in combinable_items)
			if(istype(checked_atom, to_checked_type))
				is_combinable = TRUE
	//if it is a number, it most likely is a bitflag we passed
	else if (isnum_safe(src.to_combine_item))
		if (istool(checked_atom, src.to_combine_item))
			is_combinable = TRUE
	else if (istype(checked_atom, src.to_combine_item))
		is_combinable = TRUE

	if(is_combinable)
```

ZeWaka (approved):

code/datums/components/assembly_comp.dm
ZeWaka: yes it would

PR #15708, opened 2023-09-05
Update the donut 3 armoury

CheffieGithub:
## About the PR 
Expands the donut 3 armoury by a few tiles and shuffles the contents around on the west side. Adds a mini bar to fill in the new empty space. Removes the air tank since I have never seen it used and it becomes ineffective if the riot gas is used.

Slightly old in-game image (glasses are not prefilled)
image

Current map editor image
image

All the "loose" items are now under protection of the turret with the east raised section being turned into a mini bar, moving the alcohol dispenser into there and adding glasses, bottles and a new armoury secure crate just for alcohol. 
Crate specifications below:
1 box of cocktail umbrellas,
1 box assorted cocktail doodads,
1 box of fruit wedges,
1 box of straws,
1 box assorted beers,
1 box drinking glasses,
4 random styled drinking glasses,
3 bottles of vodka,
1 bottle of curacao.

Moved the anti-bio crate into the armoury.

Fixes the missing access spawners on the turret control and removes the one from the turret itself.

## Why's this needed? 
I thought it made sense for the loose items to be protected by the turret but just moving them as is made it really cramped with lots of left over space. 
Since there was already the emergency alcohol dispenser I added a bar to fill the space as a fun thing to make donut 3 have a more unique armoury since its the largest.
The turret also has less blindspots in this configuration and the most commonly stolen items are better protected by it.

## Changelog 

```changelog
(u)Cheffie
(+)The armoury on donut 3 has been renovated 
```


Tarmunora (approved):

CheffieGithub:
I put the armoury locker there because the west feels cramped with everything there and all the secure storage is on the east side, the postions of the vendor and the locker are pretty interchangable though or the table can get moved to fit it next to it if wanted. 
Also the most common means to access it I see is with a shotgun with buckshot meaning needing to go in line of fire anyway. But from my experience its also rare for it to get broken into by any means and the most common targets are protected by the turret.

I think there would be an increase of people going into the armoury for a few days after merge just because people tend to want to see new thing. But I think extra cases after wouldn't have significant increase over before and if they are in there just for the bar theres not much harm done, just like the occasional shifts where someone breaks in to make a bar or kitchen.

flappybatpal:
I am overall indifferent.

CheffieGithub:
So the layout is fine as it is?

Sord213 (changes requested):
I know the two weapon racks were close together before but I'd prefer to see them split apart if you're modifying the armory so someone can't gain access to both weapon racks just by breaking one wall above them. Also move the special equipment locker into the turret protected area.

CheffieGithub:
Updated layout
image

Removed the muzzles to move the tranqs and syringes down. Moved the antibio crate into the armoury where the special equipment was.

Sord213 (changes requested):
Looks great. One more super minor change, could you extend the armory area down onto the outer walls of the hos office area. That way if someone breaks into the armory from hos office (I dunno how likely that is to happen but still) the turrets will potentially activate.
Screenshot_507

PR #15709, opened 2023-09-06
Smoke-powder-on-closed-containers

colossusqw:
[LABEL][chemistry][balance]
## About the PR 
This PR makes it so smoke powder no longer disappears if it reacts in a closed space, instead, remaining dormant until it is able to expand.

## Why's this needed? 
It seems rather weird to have the reagent simply disappear when it has nowhere to go, this version makes it react more like steam, that leaves a container as soon as the lid is opened.

```changelog
(u)Colossusqw
(+)Smoke powder no longer fizzles into nothing when in a closed container, instead, reacting only when it is opened.
```


imawxome:
this seems like a nice way to add chem accidents, and its not super weponisable because you are forced to be in the reaction

Tarmunora (changes requested):
not mirroring behaviour to aerosol?

colossusqw:
> not mirroring behaviour to aerosol?

I could do that, didn't think about it since aerosol is pretty much rarely seen lately, but makes sense. Gonna change that as soon as i can.

deathrobotpunch:
the cool thing about this is since drinking bottles smash and release their reagents which allows the creation of molotov cocktails or ghetto smoke bombs

colossusqw:
> the cool thing about this is since drinking bottles smash and release their reagents which allows the creation of molotov cocktails or ghetto smoke bombs

I'm not sure drinking bottles are closed containers now though, pretty sure that won't happen

Tarmunora (approved):

PR #15712, opened 2023-09-06
Rearrange the Oshan Bar

TheTyrant25:
[MAPPING] [QOL] [CATERING]
## About the PR
Moves the oshan bar south a bit so that it hugs the wall, and moves the kitchen entrance up to the north side. The north side of the bar now is much more spacious and is nicer to sit in.
Adds a chemaster and a reagent extractor to the bar against the south wall.
Changed the floor tiling slightly to match the new position.
Made the bar less round and gave it a normal 90 degree corner, for more space/mobility.

## Why's this needed?
QOL for bartenders. More equipment, and more space. Plus people can now sit comfortably against the top edge of the bar and reach stuff like the claw machine without have to squeeze through a 1 tile wide gap.

## Changelog
```changelog
(u)Tyrant, GreenJelly
(+)The Oshan bar has been rearranged to be more spacious.
```

### DIFF
--- a/maps/oshan.dmm
+++ b/maps/oshan.dmm
@@ -16793,12 +16793,10 @@
 /obj/machinery/vending/fortune,
 /obj/machinery/light/incandescent/warm,
 /obj/decal/tile_edge/line/black{
-	dir = 4;
+	dir = 8;
 	icon_state = "line1"
 	},
-/turf/simulated/floor/darkblue{
-	dir = 10
-	},
+/turf/simulated/floor/specialroom/bar,
 /area/station/crew_quarters/bar)
 "bhS" = (
 /obj/table/wood/round/auto,
@@ -16930,6 +16928,10 @@
 	dir = 9;
 	icon_state = "line1"
 	},
+/obj/decal/tile_edge/line/black{
+	dir = 4;
+	icon_state = "line1"
+	},
 /turf/simulated/floor/darkblue{
 	dir = 10
 	},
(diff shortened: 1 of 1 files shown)


DisturbHerb:
I'm not gung-ho about these changes, the bar loses all of its aesthetic appeal to me in favour of homogenising it with literally every other map's bars; especially with squaring off the bar counters and moving it away from its centralised location.

flrsh (changes requested):
I like the current bar layout and would like it to stay as it is. I don't think reagent extractors are necessary, but if you'd like to add a chemaster, I think you could add it to the kitchen nearish the bar entrance (e.g. replace food processor with chemaster, replace crate with peach rings with food processor, put peach rings + misc in one of the crates in the carpeted backroom).

CheffieGithub:
The chemmaster and reagent extractor seems to be for parity with other bars and makes sense since oshan is a high pop map.
I like the layout changes myself, the top section is currently a squeeze and these changes look like it will have better flow. but it does lose uniqueness without the central bar.

There is an accidently removed hazard stripe below the north bar door into the hallway and the glass recycler placement is a downgrade since its no longer public.

Another plus is no disposal pipes under walls.

TheTyrant25:
> I like the current bar layout and would like it to stay as it is. I don't think chemasters are necessary, but if you'd like to add a reagent extractor, I think you could add it to the kitchen nearish the bar entrance (e.g. replace food processor with reagent extractor, replace crate with peach rings with food processor, put peach rings + misc in one of the crates in the carpeted backroom).

might as well start a whole new branch then

TheTyrant25:
sequel in the form of  #15720 has been posted

PR #15713, opened 2023-09-06
Orbital shaker

DisturbHerb:
[CHEMISTRY][OBJECTS][FEATURE]

## About the PR 

This PR introduces orbital shakers to the code base; in two different sizes, and two different colour schemes for departmental flair.

orbital_shaker

The orbital shaker applies physical shock `force = 5` to all `/obj/item/reagent_containers/glass` sitting utop its platform. By plagiarizing plate code, one can easily remove and add containers to the machine at their leisure. It can be (de)activated by interacting with an empty hand.

The larger orbital shaker can carry four vessels, while the smallest can carry a single vessel.

This PR does not introduce any mapping changes, and these machines are not presently player-obtainable.

## Why's this needed? 

In light of the recent chemistry rework, some form of continuously applying physical shock to reaction vessels could be desired above going for a jog every time you needed some styptic.

Also, look at him go.

ezgif-5-65cba691ee


TobleroneSwordfish (changes requested):
Looks cool, neat effect, have some nitpicks

code/modules/chemistry/Chemistry-Machinery.dm
```diff
@@ -291,6 +291,183 @@ TYPEINFO(/obj/machinery/chem_heater)
...
+	proc/arrange_containers()
+		if (!src.count_held_containers()) return
+		for (var/i in 1 to length(src.held_containers))
+			if (!src.held_containers[i]) continue
+			var/current_y = ceil(i / src.container_row_length)
+			var/current_x = i - (src.container_row_length * (current_y - 1))
+			src.held_containers[i].pixel_x = src.first_container_offsets["X"] + ((current_x - 1) * src.container_offsets["X"])
+			src.held_containers[i].pixel_y = src.first_container_offsets["Y"] + ((current_y - 1) *src.container_offsets["Y"])
+
+	proc/count_held_containers()
```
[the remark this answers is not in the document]
DisturbHerb: > Do I dare ask why this isn't just `length()`? Should probably be managing null refs in some other way.

Hooh boy, I was afraid of this conversation.

The reason why beaker indexes get nulled instead of removing them directly from the list was due to the fact that the pixel offset of each container on the platform is calculated with that beaker's index in the `held_containers` list. If newly added beakers were removed from the list and changing the indexes of each one still remaining, it'd start reshuffling containers around when they shouldn't reasonably be. Unless there are better ways of tracking beaker positions properly, this was about the only method that I reasoned would work for the effect I wanted.

DisturbHerb:
> ... could maybe allow glasses as well as beakers?

Specifying `glass` containers was a deliberate choice to the exclusion of drinking glasses. This was made without the bar in mind, if non-researchers need their chems shaken, they still have the option of going for a run.

code/modules/chemistry/Chemistry-Machinery.dm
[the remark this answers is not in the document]
TobleroneSwordfish: That makes sense actually, it's just rare that treating a byond list like a regular array is the right thing to do.

TobleroneSwordfish (approved):
Looks cool, neat effect, have some nitpicks

code/modules/chemistry/Chemistry-Machinery.dm
```diff
@@ -304,7 +304,6 @@ TYPEINFO(/obj/machinery/chem_shaker)
 #else
 	icon_state = "orbital_shaker"
 #endif
-	density = 1
```
TobleroneSwordfish: ?

code/modules/chemistry/Chemistry-Machinery.dm
```diff
@@ -397,10 +412,14 @@ TYPEINFO(/obj/machinery/chem_shaker)
...
-		src.audible_message("<span class='notice'>[src] whirs to life, rotating its platform!</span>")
-		processing_items.Add(src)
+		src.power_usage = src.emagged ? 1000 : 200
+		animate_orbit(src.platform_holder, radius = src.radius, time = src.emagged ? src.orbital_period / 5 : src.orbital_period, loops = -1)
+		if (src.emagged)
+			src.audible_message("<span class='alert'>[src] is rotating a bit too fast!</span>")
+		else
+			src.audible_message("<span class='notice'>[src] whirs to life, rotating its platform!</span>")
+		if (!(src in processing_items))
+			processing_items.Add(src)
```
TobleroneSwordfish, line 424: Shouldn't this use the machine process loop?

TobleroneSwordfish (changes requested):

code/modules/chemistry/Chemistry-Machinery.dm
DisturbHerb: glass recycler is undense, it's also a tabletop device so i reasoned that it shouldn't be dense either

code/modules/chemistry/Chemistry-Machinery.dm
```diff
@@ -420,15 +420,15 @@ TYPEINFO(/obj/machinery/chem_shaker)
 			src.audible_message("<span class='alert'>[src] is rotating a bit too fast!</span>")
 		else
 			src.audible_message("<span class='notice'>[src] whirs to life, rotating its platform!</span>")
-		if (!(src in processing_items))
-			processing_items.Add(src)
+		if (!(src in processing_machines))
+			processing_machines.Add(src)
```
TobleroneSwordfish, line 424: not quite that simple, sorry I should have specified that you need to use `SubscribeToProcess` and `UnsubscribeProcess` as `processing_machines` is a multidimensional list full of buckets

DisturbHerb: waa

PR #15715, opened 2023-09-06
changes ether's behaviour and recipe

colossusqw:
[LABEL][chemistry][feature][balance]

## About the PR 
This PR changes renames the ether chemical to Diethyl Ether, and restructures some of it's behaviour and the way in which it is made.

### Behavior changes:
- Diethyl Ether is extremely flammable, and besides combusting into a small fireflash at relatively low temperatures, also does it when in contact with any open flame.

- Diethyl Ether now produces vapor when above 10c, making it continuously offgass at room temperature, which can be dangerous considering the vapor easily catches fire.

- Diethyl Ether works slightly better as an anesthetic, having a higher overdose threshold(there is a big increase in difference between therapeutic and toxic doses), reducing the recent trauma effect, and having a 0.2 depletion unless currently KO-ing or above 10 units.

- Diethyl Ether's KO moved to 40 cycles, but this can be reduced to 20 if more than 10 units present.

### Recipe changes:
- Base reagents(ethanol, hidrocloric acid and oxygen) remain the same, though the reaction only starts when above 30c.

- Reaction now takes time, yielding Diethyl Ether at a base speed of 0.5 units per cycle, but that can increase with temperature. The reaction offgasses Diethyl Ether as well, but keeping the container closed causes it to cool the reaction, and requiring one to pay attention on heating to upkeep the reaction speed.

- The reaction only happens while the amount of ether is inferior to the amount of ethanol present.

## Why's this needed? 
This gives the currently overlooked reagent Ether some more interesting effects, both in it's use in moderation as a easier to acquire anesthetic, and in it's possible use as a KO chemical. While it still is very much worse than capulettium in most regards, it now bridges the gap between it and morphine.

As per the reaction, Diethyl Ether's temperature based creation makes a compromise between how safe one wants to be and how expedient(and possibly effective) it's creation will be. It is easier to make than most other generic precursors, and if one keeps an eye on the reaction, it tends to go smoothly as well.

```changelog
(u)Colossusqw
(*)Changed Diethyl Ether's base behavior and recipe(more info in minor changes).
(+)Diethyl Ether is now a time reaction that scales in speed with temperature, up until 80c, in which it becomes dangerously flammable. The reaction gives off fumes and only lasts as long as there is more ethanol than ether.
(+)Diethyl Ether now has a smaller depletion when below 10us and not currently KOing, if above 10 units, KOs in 20 cycles, else KOs in 40. It also helps reduce the recent trauma status effect.
(+)Diethyl Ether now gives off Ether vapor when above 10c.
```


Flaborized (commented):
The design ideas/concepts look good to me, though I'll defer to another more experienced dev's review for the code. I like the changes though!

colossusqw:
!merge_upstream

TobleroneSwordfish (changes requested):
The reaction appears to duplicate ethanol?
image
It can also spontaneously detonate due to offgassing heating the mixture for some reason.
Aside from that I doubt it will see much use due to having a relatively annoying recipe and mediocre effects, but that was already the case so I don't see harm in trying to make it more interesting.

code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -2651,14 +2651,74 @@
...
+		mix_sound = null
+		instant = FALSE
+		stateful = TRUE
+		min_temperature = T0C + 10 // Generates vapor in room temperature
+		reaction_icon_color = "#c5d1d3"
+		var/count = 0
+		var/amount_to_smoke = 1
+
+		does_react(var/datum/reagents/holder)
+			if (holder.my_atom && holder.my_atom.is_open_container() || istype(holder,/datum/reagents/fluid_group))
```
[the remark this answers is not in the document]
colossusqw: Added the check like the one from cyanide.

code/modules/chemistry/Reagents-Medical.dm
```diff
@@ -146,20 +149,60 @@ datum
...
 			on_mob_life(var/mob/M, var/mult = 1)
 				if(!M) M = holder.my_atom
 				if(!counter) counter = 1
 				M.jitteriness = max(M.jitteriness-25,0)
 				if(M.hasStatus("stimulants"))
-					M.changeStatus("stimulants", -7.5 SECONDS * mult)
+					M.changeStatus("stimulants", -4 SECONDS * mult)
+				if(M.hasStatus("recent_trauma")) // can be used to help fix recent trauma
+					M.changeStatus("recent_trauma", -2 SECONDS * mult)
+				if(holder.has_reagent(src.id,10)) // large doses progress somewhat faster than small ones
```
TobleroneSwordfish, line 192: The PR description says that it has increased depletion rate while the mob is KOd, that isn't here. Not sure we want a 0.2 depletion rate knockout chemical, especially with the offgassing behaviour potentially leading to easy accidental perma-stuns.

TobleroneSwordfish (changes requested):

code/modules/chemistry/Reagents-Medical.dm
colossusqw: Oh, it used to, but since the amount needed for a KO is 8, i wasn't sure if the on KO change was necessary, i can definitely re-add the on KO change if continuous new doses are a problem. I personally don't think they would be considering the offgass happens in very minute amounts, though.

PR #15718, opened 2023-09-06
Gas powered chemical mixer

TobleroneSwordfish:
[CHEMISTRY] [BALANCE] [FEATURE]
## About the PR 
Adds a new atmos/chemistry machine: a chemical mixer that runs on thermal energy from hot gasses to mix chemicals in the barrel placed on top of it, increasing the reaction speed.

image
(clown not included)

Mechanically it acts as a freezer, cooling gasses down while operating, meaning it requires a steady supply of hot gas to keep operating. The speedup ranges from 1-3x, with 2x being easily possible with a basic 2/3s oxygen to plasma chamber burn, but higher values getting exponentially harder.

image

How does it work? Gas make thing go spinny, every barrel canonically has a stir bar or something.

## Why's this needed? 
Chemicals are slow to mix in large containers, this should help mitigate that somewhat. Also finally a use for toxins that isn't just bombs and mayhem.

TODO:
- [ ] Uncoder the sprite a bit
- [ ] Tweak the maths (please help the asymptotes are all wacky)
- [ ] Define `mats` cost
- [ ] Map changes

Oh and also map changes.
## Changelog 

```changelog
(u)LeahTheTech
(*)Added a chemical mixer that increases the reaction speed of barrels using hot gasses.
```


TDHooligan:
mixed feelings. this feels like it might become what is weighed in as 'the norm' for chemistry, and toxins is already a lab that you can't use if someone else is using it (and is really hard to reset for bomb making, with lingering gas in pipes & half-used cans & superheating...). 

if it becomes normal to use it, toxins becomes 'the chemical mixer room', and bomb makers can get stuffed. if people don't regularly use it, i'd be concerned we get another 'ask botany' when people have issues with how slow certain recipes are. 

For the numbers, tripling speed seems like such a weak outcome for an entire department's co-operation, when you could just use 3 beakers. If it's going to put toxins out of commission, i'd kick it up closer to 10x. Then it's also overkill unless you're truly doing a Chemistry Operation worthy of using 2 parts of science.

Laboredih123:
interesting concept

TobleroneSwordfish:
Most toxins labs have multiple burn chambers, and being unable to properly clear atmos pipes sounds like something that should be fixed. In terms of not being worth it, it takes maybe a minute or two to set up a burn that can run this at 2-2.5x speed, which I'm hoping would be worth it if you're producing any mass amount of chemicals? The numbers can easily be adjusted if not though.

TDHooligan:
if this comes with a rework of toxins to have multiple independent functioning burn chambers, sure. most have only 1 means to heat gas without burning it atm. 
agree that numbers can be tweaked after implementation (i still think 3x is a little low)
still uncertain on how the speeds impact reaction speed balancing going forward, but i guess that really is just a thinly veiled 'slippery slope' argument

Azrun:
But... how do we put the clown in it?

Wisemonster:
I do like this, considering it finally gives toxins a use that isn't just "Make bombs for antaging", though as was stated, this may require the two burn chambers in toxin to be more independent. Right now, most toxins have a single pipe going through the burn chambers to heat the gasses inside said pipes (hell, Donut 3's doesn't even go to the second burn chamber), which would make using both burn chambers for different purposes (ex. One for the machine is this pr, another for agent B farming) rather difficult.

PR #15731, opened 2023-09-07
Adds banana juice to Honky Tonic recipe

itchybones:
[Chemistry][Catering][QoL]
## About the PR 
This PR adds the new juice_banana reagent to the Honky Tonic recipe and
removes colored reagent from it.

## Why's this needed? 
Honky Tonic is rather difficult to mix compared to other gimmick
drinks, this PR aims to make it a little easier by adding a botany exclusive
chem in the place of a (usually) chemistry exclusive one.
Also, it makes sense for a Clown-themed drink to include banana juice.

## Changelog 

```changelog
(u)Chasu
(+)Banana juice has replaced colorful reagent in the Honky Tonic recipe.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -1883,7 +1883,7 @@
 		name = "Honky Tonic"
 		id = "honky_tonic"
 		result = "honky_tonic"
-		required_reagents = list("tonic" = 1, "lube" = 1, "colors" = 1, "neurotoxin" = 1)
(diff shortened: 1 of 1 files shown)


itchybones:
> Honky tonic is far too complicated already... adding another chem, especially a botany-exclusive one, sounds like a really bad idea.

Yeah, I agree that it's overly complicated at the moment, maybe the juice could replace neurotoxin?

itchybones:
Good idea, colored reagent is definitely a too complex chem for this drink, I'll change that once I've got time. Thanks for the feedback!

TobleroneSwordfish (approved):

PR #15732, opened 2023-09-07
change amanitin to on_remove instead of moblifecomplete

MRLinew:
## About the PR 
This PR changes amanitin to use on_remove instead of onmoblifecomplete. Right now, if anything flushes amanitin out, it does zero damage. using on_remove still has amanitin do its damage after its been flushed out. One thing to keep in mind is that this includes adminhealing, which purges the amanitin causing it to do damage after the players been healed.

## Why's this needed? 
Amanitin is kind of useless because when discovered it can be entirely negated. 

## Changelog 

```changelog
(u)Froggit Dogget
(+)Amanitin now does its damage when flushed out of the system.
```

### DIFF
--- a/code/modules/chemistry/Reagents-PoisonEtc.dm
+++ b/code/modules/chemistry/Reagents-PoisonEtc.dm
@@ -213,6 +213,7 @@ datum
 				if (!istype(M)) return
 				M.take_toxin_damage(damage_counter)
 				logTheThing(LOG_COMBAT, M, "took [damage_counter] TOX damage from amanitin.")
+				damage_counter = 0
 
 
 		harmful/chemilin
--- a/code/modules/chemistry/Reagents-PoisonEtc.dm
+++ b/code/modules/chemistry/Reagents-PoisonEtc.dm
@@ -207,11 +207,12 @@ datum
 					M = holder.my_atom
 				damage_counter += rand(2, 4) * mult * min(1, volume)
 				..()
-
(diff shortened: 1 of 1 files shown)


code/modules/chemistry/Reagents-PoisonEtc.dm
```diff
@@ -207,11 +207,12 @@ datum
...
-			on_mob_life_complete(mob/living/M)
-				if(M)
-					M.take_toxin_damage(damage_counter)
-					logTheThing(LOG_COMBAT, M, "took [damage_counter] TOX damage from amanitin.")
+			on_remove()
+				..()
+				var/mob/living/carbon/human/M = holder.my_atom
+				if (!istype(M)) return
+				M.take_toxin_damage(damage_counter)
+				logTheThing(LOG_COMBAT, M, "took [damage_counter] TOX damage from amanitin.")
```
Tarmunora, line 216: should 0 out `damage_counter` here - in case on_remove starts acting weird again, better to just prevent double-ticking the damage

MRLinew: done!

Retrino:
kind of seems like too big of a buff, considering amanitin has no indication of being in the bloodstream aside from randomly deciding to scan to check reagents? for something that can be grown and mass produced by botany it Should be purgable as a counter

PR #15734, opened 2023-09-07
Use Standard tgui UI for Sillicon Paper

Garash2k:
## About the PR 
When an AI gets shown a paper through a camera, or when sillicons left click papers, it used to be shown by an html popup with the paper's contents plugged inside. This change makes it so that we just open the paper's tgui window for them.
Other small change: fixes a bug where silicons could edit notes from afar.

## Why's this needed? 
Makes sense to only have 1 paper viewing technique, probably fixes a font bug with the AI camera, might fix href exploits if there are any, and is also what https://github.com/tgstation/tgstation/pull/68612 did anyways (no need to port the whole refactor for this, just pointing out it's how they solved a "showing the paper to AI cameras" problem)

Paper to camera AI PoV:
image
Borg left clicking paper:
image

How they would see it before:
266410142-277b2244-e0ba-4178-8fae-15939d0b6137

Edit while in range, with a pen selected, yes:
image

Edit while out of range, with a pen selected, no:
image

[UI][Bug]

### DIFF
--- a/code/modules/writing/paper.dm
+++ b/code/modules/writing/paper.dm
@@ -123,19 +123,9 @@
 		qdel(src)
 
 /obj/item/paper/attack_ai(var/mob/AI as mob)
-	var/mob/living/silicon/ai/user
-	if (isAIeye(AI))
-		var/mob/living/intangible/aieye/E = AI
-		user = E.mainframe
-	else
-		user = AI
-	if (!isAI(user) || (user.current && GET_DIST(src, user.current) < 2)) //Wire: fix for undefined variable /mob/living/silicon/robot/var/current
-		var/font_junk = ""
-		for (var/i in src.fonts)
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
Does this let them edit the paper remotely? I'd assume not, but silicons do weird things with `tgui_status` checks

Garash2k:
> Does this let them edit the paper remotely? I'd assume not, but silicons do weird things with `tgui_status` checks

It does let them edit remotely! But, that's an outstanding bug that would be made more visible by this PR, they can already do so by alt+clicking the paper.

There's probably an simple fix for it by adding a distance check in `ui_data()` to end up with `"editMode" = PAPER_MODE_READING`, I'll look into it soon

Garash2k:
`in_interact_range()` check added to non reading edit modes, works well, PR description updated

PR #15735, opened 2023-09-07
Fix DWAINE Scanner not scanning if item is inserted through UI

TealSeer:
[BUG] [GAME OBJECTS]
## About the PR 

The code for the scanner only called the scan_document() function if you hit the scanner with a paper/photo to insert it, you could also insert something by clicking on the ------- slot in the UI, but this did not call scan_document().

## Why's this needed? 

Bugg

### DIFF
--- a/code/modules/networks/computer3/mainframe2/misc_terms.dm
+++ b/code/modules/networks/computer3/mainframe2/misc_terms.dm
@@ -2257,6 +2257,8 @@ TYPEINFO(/obj/machinery/networked/printer)
 					if (istype(mag.holding, /obj/item/paper) || istype(mag.holding, /obj/item/photo))
 						I = mag.holding
 						mag.dropItem(0)
+					else
+						return
 				else if (istype(I, /obj/item/paper) || istype(I, /obj/item/photo))
 					usr.drop_item()
 				else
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (approved):

code/modules/networks/computer3/mainframe2/misc_terms.dm
```diff
@@ -2252,19 +2252,24 @@ TYPEINFO(/obj/machinery/networked/printer)
...
+				else if (istype(I, /obj/item/paper) || istype(I, /obj/item/photo))
+					usr.drop_item()
+				else
+					return
+				I.set_loc(src)
+				src.scanned_thing = I
+				boutput(usr, "You insert [I].")
+				SPAWN(0)
+					if(!scan_document())
+						use_power(200)
```
ZeWaka, line 2271: Why would it only use power if it fails

TealSeer: @ZeWaka scan_document() returns false if it succeeds and true if it fails for some reason

ZeWaka: it's because aibm wrote it

PR #15737, opened 2023-09-07
Fixes being able to scoot lockers through space

MRLinew:
## About the PR 

This PR removes the ability to scoot lockers through space after welding holes in them.

## Why's this needed? 

squashing bugs..


flappybatpal:
Can you PR the same for chairs (assuming this method is approved)?

TobleroneSwordfish (approved):
Cursed but probably right

PR #15738, opened 2023-09-07
Bunsen Burners

Flaborized:
## About the PR 

Adds bunsen burner items. Still a draft and doesn't add them specifically anywhere but I imagine I'll put them in lab counters somewhere and also add an option to print them. You can put items on them when they're out in the world (not in your inventory) and they'll "connect" to them, and when on it'll heat the container. But it doesn't put the reagent container *inside* the burner, it puts it *on* it, so you can still hook up condensers and stuff while having a heating element.

It uses context menus to select between three heat levels, which do as follows:

- Low: Heats to 110K, 20K every processing tick
- Medium: Heats to 400K, 40K every processing tick
- High: Heats to 900K, 65K every processing tick

This is much less precise than chem heater/coolers, but gives you the advantage of being able to manipulate the beaker *while* putting the heat on it, so is hopefully distinct. Clicking on the bunsen burner while it's "in-world" and has a container attached to it will open the context menu instead of picking it up (using it in-hand does the same). Moving the burner or container at all when they're connected will disconnect them. There's also a  green light on the bottom left corner of the burner that lights up when there's a connected container.

## Why's this needed? 

With the more complicated chemistry of the rework, I think a device like this fills a cool niche for chemists and feels very lab equipment-y. It's distinct in a few ways from current methods of heating (dropped welders, chem heaters, artifact heat plates) so I'm hoping it'll fit in nicely with other fancy equipment for chem.

TODO:
- Currently if you open the context menu while it's on the ground, moving won't close the menu, need to fix this
- Need to fine-tune heating temps, current are picked arbitrarily and should maybe? work more like heaters/coolers instead of linear heat, or less precisely stop at the 'max' temperature
- Currently they don't heat the environment around them since they'd also heat the beakers attached to them twice over, it'd be cool if it could start plasma fires or heat stuff up in general without affecting the beaker heating
- I'm also drafting this to gather feedback to improve QoL and stuff.

## Changelog 

```changelog
(u)Flaborized
(*)Added bunsen burners (add more detail to changelog entry when draft is done ty future flabo)
```


code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -996,3 +996,122 @@ proc/ui_describe_reagents(atom/A)
...
+			src.UpdateOverlays(light_image, "light")
+		else
+			src.UpdateOverlays(null, "light")
+
+	proc/change_status(var/status)
+		src.remove_simple_light("bunsen_flame")
+		if(status != "off")
+			switch(status)
+				if("low")
+					add_simple_light("bunsen_flame", list(245, 181, 61, 0.7 * 255))
```
TobleroneSwordfish, line 1055: if you change this to use `/datum/component/loctargeting/simple_light` it'll work for people holding it too

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -996,3 +996,122 @@ proc/ui_describe_reagents(atom/A)
...
+					current_temp_to_heat = 20
+					current_temp_cap = 110
```
TobleroneSwordfish, line 1057: these are in kelvin, so this is -163C

TobleroneSwordfish (changes requested):
These look so preeeettyy!
You might be able to use `hotspot_expose` if you `set_loc` the beaker inside the burner, but I'm not sure what would do to its pickuppableness. Maybe look at how plates/serving tables work?
Also might be nice to see the flame level on examine.

code/modules/chemistry/Chemistry-Tools.dm
Flaborized: lol oop!

Flaborized:
> You might be able to use hotspot_expose if you set_loc the beaker inside the burner, but I'm not sure what would do to its pickuppableness. Maybe look at how https://github.com/goonstation/goonstation/pull/15713 does it, since they're very similar machines and I think that implementation looks good?

Yeah the issue I have with set_loc in this instance is I want beakers to still be clickable (like you could fill them and stuff) and condensers specifically to be able to be hooked up. I sorta want it to be able to heat *certain* things around it but not the beaker it has on it but I also want the beaker to still be on the same turf which is a dilemma. One thing I'm considering is it only using hotspot_expose() when it has no attached container, like the container is "covering" the flame... though idk if that actually makes any sense.

Also on temperature_reagents() I added a new arg, "cannot_be_cooled" which, when true, prevents the proc from cooling a mixture. This was discussed on discord but I'm also posting here so others can see and stuff. I plan on later putting this on other things that *can* cool containers but shouldn't be able to (welding tools over 700K) later after this if it turns out this is a good idea. Otherwise addressed the code review stuff

Flaborized:
undrafted since it seems my issue with the context menu not vanishing when moving went away by just setting close_moved = TRUE, as for ambient heat I feel pretty much fine with it not making hotspots atm :beesleep:

TobleroneSwordfish (approved):

TobleroneSwordfish:
!merge_upstream

PR #15741, opened 2023-09-07
Turns caffeine into it's own chemical

colossusqw:
[LABEL][chemistry][feature]
## About the PR 

This PR adds the chemical caffeine, which functions much like the ethanol chemical already ingame does. It removes the stamina regen buffs that most caffeinated drinks directly add, and instead causes them to deplete into caffeine.
Caffeine has a depletion of 0.1 when below 30 units, and 0.2 when above it, causes messages about feeling sharp and motivated.

Like Ethanol, Caffeine has most of it's effects as volume dependant variables, making it so a small amount will not have many noticeable effects, but a large amount will have much stronger effects.

- Reduces blurry vision

### 0 to 5 units:
- +1 stamina regen & 3% stun resist

### 5 to 20 units:
- +2 stamina regen & 7% stun resist 
- reduction of drowsiness

### 20 to 40 units:
- +3 stamina regen & 12% stun resist
- reduction of drowsiness

### 40 to 60 units:
- +4 stamina regen & 20% stun resist
- messages about feeling VERY motivated and possibly a bit of heartburn
- big reduction of drowsiness
- causes heart failure in 150 cycles 

### above 60 units:
- +5 stamina regen & 25% stun resist
- lots of messages about heartburn and way too much energy
- random chance of walking in a random direction 
- causes heart failure in 30 cycles

## Why's this needed? 
The way alcohol works in this game is quite interesting and fun, and making caffeine itself work like it would give it some more interesting interactions and in general just more flavor. It's not stronger than most stimulants and definitely not as instantly powerful as mixing 5 different versions of coffee, but it makes up for it by lasting longer.

It also has some interesting adverse effects in the form of the heart damage, being something that you definitely don't want too much of in you. Another effect of this is not having to have yet another coffee every 2 minutes due to it's very fast depletion, it's very possible to enjoy the small upsides of 10 units of coffee for quite a while.

```changelog
(u)Colossusqw
(*)Caffeine is now it's own standalone chem.
(+)Caffeine is formed by caffeinated drinks as they deplete, and it's effects can vary wildly based on volume, much like current ethanol.
```


CheffieGithub:
Considering the low depletion, and scaling nature which will lead to high and long doses. I think the addiction chance is a bit high and having 10 units is probably always going to give you one since it will roll the 10% chance 90 times roughly.

colossusqw:
I see, I'm going to reduce it to a 1, it should still be addictive in that amount.

A thing I'd like to point out is that keeping some of the stun resistance from the stronger chems like energy drink and expresso is intentional, as to not keep them useless.

CheffieGithub:
I would increase the minimum cycles for addiction a little bit as well, people tend to avoid coffee, ethanol and nicotine due to the high chance for addiction, which cripples you for an incredibly long time.
Consider that higher doses will last longer and therefore be more addictive as well which is a good thing but a low dose shouldn't have a massive chance.

Glamurio:
> Personally, i am absolutely not a fan about adding a damaging overdose effect on caffeine. Coffee was always a safe stam chem to have avaible for fun recreational use or to top off a bit of stam reg, and making it more potent but also more dangerous both goes against this. So personally, i would be fine with it if both tresholds over 40u would be simply removed.

Eh, I think it's fine to have damaging side-effects with positive buffs. Something that makes you more robust, with drawback that you need to play around. That's how most drugs function, and it's more fun than to just have it be fairly safe with no drawbacks.

CheffieGithub:
It having damage above 40 is good, its unlikely you will get that high during normal consumption and chemcials just being buffs with no downsides besides addictions _maybe_ isn't good.

CheffieGithub:
below 30 units depletion rate is doubled, so their actual dose will be much lower. People can adjust how much they drink, no-one is being forced to drink 100 units of coffee. Heart damage is only an issue if you end up with organ failure with it, its more of a flavour thing tbh.

CheffieGithub:
Testing and 133 units freshly brewed coffee is ~~48 units of caffine, which caused me 15.8 toxin damage and 18.6 heart damage, a bit high for such a short time above 40 units something to adjust? but still easily treatable. 
Although this stops coffee being completely safe, they don't need to drink that much in the first place.

The problem with alcohol contents was that drinks were more alcoholic than pure ethanol.

Returning to the actual PR are there plans to change drinks that use coffee to decay into caffeine? 
I think the depletion rate would be better as 0.2 on the chemical and reduced to 0.1 over 30 units. Instead of being 0.1 and increased to 0.2 below 30. 
The same effect but more sane.

Glamurio:
> Returning to the actual PR are there plans to change drinks that use coffee to decay into caffeine? I think the depletion rate would be better as 0.2 on the chemical and reduced to 0.1 over 30 units. Instead of being 0.1 and increased to 0.2 below 30. The same effect but more sane.

I spoke with colossus and have already said I've more or less primed a follow-up PR, since we'd both been working on a caffeine rework. The idea is to introduce something like `fooddrink/caffeinated` which inherits a similar logic to ethanol. Meaning any drink that's `caffeinated`, such as cocktails containing coffee, or even cola and tea, could have some form of minor caffeine amount, with coffee being the mid-point, teas being on the low-end and energy drinks perhaps being the highest extreme.

colossusqw:
> Testing and 133 units freshly brewed coffee is ~~48 units of caffine, which caused me 15.8 toxin damage and 18.6 heart damage, a bit high for such a short time above 40 units something to adjust? but still easily treatable.
> 
> Although this stops coffee being completely safe, they don't need to drink that much in the first place. if it really is an issue the depletion into caffine can simply be adjusted without removing the negative effects.
> 100 units of fresh coffee was too little to cause adverse effects.
> 
> The problem with alcohol contents was that drinks were more alcoholic than pure ethanol.
> 
> Returning to the actual PR are there plans to change drinks that use coffee to decay into caffeine? 
> I think the depletion rate would be better as 0.2 on the chemical and reduced to 0.1 over 30 units. Instead of being 0.1 and increased to 0.2 below 30. 
> The same effect but more sane.

I did increase the depletion of lots of caffeine depleting chems, so for instance freshly brewed coffee depletes into 0.5 units of caffeine while removing 1 u of itself.

With those parameters, it's really hard to unintentionally give yourself a heart attack.

colossusqw:
> > Testing and 133 units freshly brewed coffee is ~~48 units of caffine, which caused me 15.8 toxin damage and 18.6 heart damage, a bit high for such a short time above 40 units something to adjust? but still easily treatable.
> > 
> > Although this stops coffee being completely safe, they don't need to drink that much in the first place. if it really is an issue the depletion into caffine can simply be adjusted without removing the negative effects.
> > 100 units of fresh coffee was too little to cause adverse effects.
> > 
> > The problem with alcohol contents was that drinks were more alcoholic than pure ethanol.
> > 
> > Returning to the actual PR are there plans to change drinks that use coffee to decay into caffeine? 
> > I think the depletion rate would be better as 0.2 on the chemical and reduced to 0.1 over 30 units. Instead of being 0.1 and increased to 0.2 below 30. 
> > The same effect but more sane.
> 
> I did increase the depletion of lots of caffeine depleting chems, so for instance freshly brewed coffee depletes into 0.5 units of caffeine while removing 1 u of itself(that along with a 0.1 or 0.2 depletion means way less than half caffeine yield as well).
> 
> With those parameters, it's really hard to unintentionally give yourself a heart attack.

Tarmunora:
conceptually, having a unified caffeine like this is interesting, but I'm not a fan of this implementation.
Over-complicated and randomly poisonous, and tbh message spam is always annoying

colossusqw:
> conceptually, having a unified caffeine like this is interesting, but I'm not a fan of this implementation. Over-complicated and randomly poisonous, and tbh message spam is always annoying

I can definitely simplify it.

colossusqw:
Alright, I've removed a lot of the messages, the tox and heart damage damage, added a cooldown to the remaining messages, and changed the previous heart damage to a counter that increases when over OD thresholds and can cause heart failure after a while(similar to energy drink's behavior), and normalized some behavior across all volumes
This should make it even less dangerous than before, and a lot simpler overall

code/modules/chemistry/Reagents-Drugs.dm
```diff
@@ -492,6 +492,138 @@ datum
...
+			overdose = 60
+			threshold = THRESHOLD_INIT
+			var/heart_failure_counter = 0
+
+			proc/caffeine_message(var/mob/M, var/power) // condensing all the messages in the same place
+				if (ON_COOLDOWN(M, "Caffeine Message", 10 SECONDS)) //Reduce chat spam
+					return
+				switch(power)
+					if(1)
+						boutput(M, pick("<span class='notice'>You feel refreshingly energised.</span>",\
```
TDHooligan, line 517: the player probably only needs to know about the caffeine they've drank if it's killing them

colossusqw: Makes sense, done

code/modules/chemistry/Reagents-Drugs.dm
```diff
@@ -492,6 +492,127 @@ datum
...
+					depletion_rate = 0.1
+
+				switch(caffeine_amt)
+					if(0 to 5)   //This is your trace amount of caffeine, doesn't do much
+						stun_resist = 3
+
+					if(5 to 20)  //A regular coffe mug's worth
+						if (M.get_eye_blurry() && prob(75))
+							M.change_eye_blurry(-1 * mult)
+						stun_resist = 6
```
Tarmunora, line 549: setting stun resist like this does not work

colossusqw: Oh, i see, alrighty, gonna figure out a way to make it work then.

colossusqw: This seems to work pretty well, but if it's still too complex, i can keep it at a set value?

colossusqw:
!merge_upstream

PR #15752, opened 2023-09-08
make unloading a gun make a sound

### LINKED ISSUE #7837: Unloading guns is silent
### Labels

[trivial]

### Description

Unloading a gun makes no noise

### Steps to reproduce

1. unload a gun
2. it makes no noise

### Expected Behavior

The gun makes a noise!!!

_No response_

TheTyrant25:
[SOUND] [BUG]
## About the PR
Makes unloading a gun play the same sound as loading it.

## Why's this needed? 
It did not have a sound
Fixes  #7837

### DIFF
--- a/code/obj/item/gun/kinetic.dm
+++ b/code/obj/item/gun/kinetic.dm
@@ -162,7 +162,6 @@ ABSTRACT_TYPE(/obj/item/gun/kinetic)
 				user.show_text("You can't unload this gun.", "red")
 				return
 			src.eject_magazine(user)
-			playsound(src, src.ammo.sound_load, 50, 1)
 		return ..()
 
 	attack(mob/M, mob/user)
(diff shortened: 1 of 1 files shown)


CheffieGithub:
The same sound as loading could cause confusion

TheTyrant25:
what sound do you suggest then? different guns have different loading sounds so it could get weird if they all were the same

CheffieGithub:
Im not sure but the two should be distinct in some way. you can use the same sound with sound settings to make it a bitdifferent I suppose

TheTyrant25:
that's a good idea, i'll do that

frawhst (approved):
Slightly varied sound would be nice but yeah overall I want this (i made the issue report, so slightly biased).

TheTyrant25:
it turns out i could not figure out how to do that

CheffieGithub:
proc/playsound(atom/source, soundin, vol, vary, extrarange, pitch, ignore_flag, channel, flags) 
I would mess with vol, vary, and pitch

TheTyrant25:
turns out that vary is a bool which automatically messes with the pitch for you, who knew. I didn't.

PR #15760, opened 2023-09-08
Resprite sawfly remote and fix range

CheffieGithub:
## About the PR 
Resprites the sawfly remote and adds a clicking animation of 0.7 seconds with a cooldown of 1.5 seconds to allow the sawflies to do their thing and protections for chatspam.
image

Fixes the range not being a uniform 5 tiles as the help message would suggest.

## Why's this needed? 
The sprite doesn't fit guidlines.


NightmareChamillian:
If you're altering the sawfly remote, do you think you could make it so that sawflies made via mechscanning don't also spawn with the remote?

CheffieGithub:
will do in a different PR

NightmareChamillian:
youuuu're the best, Cheffie!

Tarmunora (approved):

PR #15762, opened 2023-09-08
Remove 'murder' from the list of possible eldritch artifact faults

444Portal:
[BALANCE] [REMOVAL] [SCIENCE]
## About the PR 
Comments out /datum/artifact_fault/murder = 2, from the list of faults the Eldritch Artifacts can roll. That's it. That's the PR.

## Why's this needed? 
Instantly gibbing with no real warning, even if they are funny on paper, aren't fun to have happen to you.


444Portal:
I forgot to merge master so it has the commits from my old PR, sorry, I'm pretty dense. shouldn't change anything but artifact_controls.dm however.

Laboredih123:
counterpoint: funny when i die. i meddled with eldritch forces beyond my understanding, funny eldritch force says no

Glamurio:
> counterpoint: funny when i die. i meddled with eldritch forces beyond my understanding, funny eldritch force says no. could lower the chance of no being said.

Counter counterpoint: Funnier if it's not instant. Injects you with a bomb that will gib you. Rush to doctor to be saved or die.

MeggalBozale:
Should definitely go two ways, if artlab can get instant kill weapons they should get weapons that instantly kill themselves and vice versa. I think a cap should be put on both ways for artifacts to keep instant death out of the equation.

code/datums/controllers/artifact_controls.dm
```diff
@@ -476,7 +476,7 @@ var/datum/artifact_controller/artifact_controls
 		'sound/musical_instruments/artifact/Artifact_Eldritch_3.ogg',
 		'sound/musical_instruments/artifact/Artifact_Eldritch_4.ogg')
 	fault_types = list(
-		/datum/artifact_fault/murder = 2,
+//		/datum/artifact_fault/murder = 2,
```
frawhst, line 479:
```suggestion
```
Just delete, no need to have outdated stuff lying around in comments

frawhst (commented):
I'm fine with this as long as you increase the weight of one of the other murdery faults to compensate- add a few points to `poison` or `explode` or something.

Tarmunora (changes requested):
Id rather just change the fault to kill in a way that is borg and/or cloneable, rather than just delete outright

MeggalBozale:
I don't see why making a clone scanner a necessity for artifact lab without being disintegrated should be considered a good idea, let alone any department.

444Portal:
Going to close this and make a new PR at some point - a bit too tired/busy/whatever and it'll probably go stale. Until then I guess.

PR #15767, opened 2023-09-09
Reworks plastic flaps, makes them constructable & adds material support

Glamurio:
[Feature] [Objects] [Balance]

## About the PR 
Changes `/obj/plasticflaps` to `/obj/strip_door`.

Adds `/obj/strip_door/attackby` that react to snipping and screwing, allowing plastic flaps to be snipped with wirecutters. They can also be rotated via crowbars.

Flaps

Adds `/datum/material/rubber/plastic` and `/obj/item/material_piece/rubber/plastic`, as well as an entry in `matsci_guide_new.txt`. It also adds `/obj/item/sheet/plastic` and basic floor tiling support.

Plastic has a density of 3, hardness of 1, elec. conductivity of 2, thermal conductivity of 3 and chemical resistance of 5.

Snipping plastic flaps makes them passable and yields its `flap_material`, defaulting to `/datum/material/rubber/plastic`. Materials can be placed inside the plastic flaps, inheriting their color and density. These materials will need to be `/obj/item/sheet`.

Density of the flap material now determines the slowdown of the flaps. Low density materials cotton are barely noticeable, while high density like Uqill will cause some noticeable slowdown. The slowdown is cleared when leaving the flap tile.

image

## Why's this needed? 
Plastic flaps are actually a really cool item with a unique interaction of acting as a doorway that's only accessible via crouching. Though they were already deconstructable, deconstruction devices have limited access. Additionally, this will now allow cyborgs to get past previously impassable flaps by cutting them.

This also allows people to get creative with flaps. Once merged, perhaps more unique properties could be added to flaps, such as radioactivity or electrical conductivity for death traps.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Plastic flaps are now called strip doors and are constructable! Simply use some steel sheets to create a frame.
They can be snipped to retrieve the flaps and turned via crowbar.
Their flaps are now separate, meaning you can use any material sheet as a flap.
The density of the material determines the flap slowdown.
(+)Plastic is a new material and can be obtained from snipping plastic flaps.
```


Glamurio:
Actually, hang on a minute. I'm going to add constructible flaps, so I'll convert it to a draft real quick.

BatElite:
I feel like there's a missed opportunity atm to set opacity based on material alpha (if you're fine with plastic material being transparent anyway)

code/modules/materials/Mat_RawMaterials.dm
```diff
@@ -248,16 +248,24 @@
 	default_material = "slag"
 	mat_changename = FALSE
 
+ABSTRACT_TYPE(/obj/item/material_piece/rubber)
+/obj/item/material_piece/rubber
+	setup_material()
+		src.create_reagents(10)
+		reagents.add_reagent("rubber", 10)
```
BatElite, line 255: So putting plastic under the rubber material group is probably fine (we're not trying to make a comprehensive taxonomy here after all) but I don't think it's correct to say that plastic _contains_ rubber

Glamurio: That's very fair. I did a lot of googling to find out that rubber and plastic often contain the same polymers, but I'll remove the reagent bit.

Glamurio:
> I feel like there's a missed opportunity atm to set opacity based on material alpha (if you're fine with plastic material being transparent anyway)

Very good idea. I'll add it.

code/modules/economy/supply_misc.dm
```diff
@@ -116,6 +172,44 @@ ABSTRACT_TYPE(/area/supply)
 			if (prob(5))
 				qdel(src)
 
+/obj/strip_door/attackby(obj/item/I, mob/user)
+	if (ispryingtool(I))
+		playsound(src.loc, 'sound/items/Crowbar.ogg', 30, 1, -2)
+		src.change_direction()
+	if (isscrewingtool(I))
+		playsound(src.loc, 'sound/items/Screwdriver.ogg', 30, 1, -2)
+		SETUP_GENERIC_ACTIONBAR(user, src, 1 SECOND, /obj/strip_door/proc/screw_flaps, null, I.icon, I.icon_state, null, INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION | INTERRUPT_MOVE)
```
Tarmunora, line 181: Not in favour of making these unanchorable.

Tarmunora (changes requested):

code/modules/economy/supply_misc.dm
```diff
@@ -105,7 +143,25 @@ ABSTRACT_TYPE(/area/supply)
...
+// Slow down of flaps dependant on density of material
+// The more dense, the more slowdown
+/obj/strip_door/Crossed(atom/A)
+	..()
+	if (!src.flap_material)
+		return
+	if (isliving(A))
+		var/mob/living/M = A
+		var/density = src.flap_material.hasProperty("density") ? src.flap_material.getProperty("density") : 3
+		M.changeStatus("slowed", 2 SECONDS, density * 5)
```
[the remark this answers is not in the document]
Glamurio: Is that a gut feeling, or have you tried it in practice? Because of how `/Crossed` gets applied, probably due to tick shenanigans, you actually only end up being slowed for a very brief period of time. In practice, it's not noticeable, especially not on lower density materials. I'd be happy to lower it.

The idea was just to incentivise experimenting with different materials for different effects.

code/modules/economy/supply_misc.dm
```diff
@@ -116,6 +172,44 @@ ABSTRACT_TYPE(/area/supply)
...
+	if (isscrewingtool(I))
+		playsound(src.loc, 'sound/items/Screwdriver.ogg', 30, 1, -2)
+		SETUP_GENERIC_ACTIONBAR(user, src, 1 SECOND, /obj/strip_door/proc/screw_flaps, null, I.icon, I.icon_state, null, INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION | INTERRUPT_MOVE)
+	if(src.flap_material)
+		if (issnippingtool(I))
+			SETUP_GENERIC_ACTIONBAR(user, src, 1 SECOND, /obj/strip_door/proc/snip_flaps, user, I.icon, I.icon_state, null, INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION | INTERRUPT_MOVE)
+	else
+		if (istype(I, /obj/item/material_piece))
+			SETUP_GENERIC_ACTIONBAR(user, src, 1 SECOND, /obj/strip_door/proc/insert_flaps, I, I.icon, I.icon_state, null, INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION | INTERRUPT_MOVE)
+	..()
```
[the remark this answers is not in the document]
Glamurio: Specifically left it open, because I implemented it to work with any kind of material. Since in some industrial cases, things other than plastic would realistically see usage. I could limit it, but feel like it's just more fun to let people apply anything.

code/modules/economy/supply_misc.dm
Glamurio: I'd really love a balance discussion on this, because when you construct and build these in practice, I could see it getting annoying really quickly to have to move them via deconstruction. I'd happily make the process a bit more involved to find a middle ground, even if it's just turning up the length of the animation, but for convenience I'd really like them unanchorable, like any other construction item like pipes or conveyor belts.

code/modules/economy/supply_misc.dm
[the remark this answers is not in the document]
Tarmunora: Gut feeling - I think that the lying-down slow is sufficient here

code/modules/economy/supply_misc.dm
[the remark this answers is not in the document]
Tarmunora: Still seems very weird to just slap it with a cube of steel/etc. and have that just... work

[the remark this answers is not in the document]
Glamurio: Suppose you're right, I just find the materials able to be turned into sheets a shame, guess people will just have to use MatSci to get sheets of items. Plastic might be an exception, or get actual sheets, I'll work something out.

code/modules/economy/supply_misc.dm
Tarmunora: Imo having these able to slide around would open up weirdness for little gain - like doors, I don't see much need to be able to drag these around at will

Glamurio: Since you can place them now, being able to unachor them is just convenience. It's a simpler solution than having to grab a deconstruction device, grab a solderin iron, deconstruct them, scoot them over one tile, and then construct them. I'll remove it if it's really necessary, but you can already "slide them around", it just takes a lot more steps.

Tarmunora (changes requested):
please cut the unanchoring code

PR #15783, opened 2023-09-10
tgui Elevators

Garash2k:
## About the PR 
New UI for Oshan and Nadir's sea elevators, 2 azone elevators, and an admin only toilet paper roll elevator(?!)
image
image

Fixes the bug where 2 people on different floors could make the elevator move at the same time.
Fixes a bug where one azone elevator would actually make the other azone elevator play a ringing sound. Bug fixed by using the right type in a for in loop.

## Why's this needed? 
Looks better and more consistent with the AI turret panel which I sorta based it on
Old UI:
261918623-a774e870-0068-4991-8242-bf33e15b6e6a

## Changelog 

```changelog
(u)Garash
(+)New UI for the elevators, descend to the trench in style!
```

[UI][Rework]


TobleroneSwordfish (changes requested):
I hate to do this, but these really should be repathed to `/obj/machinery/computer/elevator` or similar. I know they're copy-pasted now, but that's not a reason to keep doing it.

Aside from that, this looks sweet!

Garash2k:
Repath is go!
Done, here are the repaths:
`/obj/machinery/computer/icebase_elevator` ->
`/obj/machinery/computer/elevator/icebase`
`/obj/machinery/computer/biodome_elevator` ->
`/obj/machinery/computer/elevator/biodome`
`/obj/machinery/computer/sea_elevator` ->
`/obj/machinery/computer/elevator/sea`
`/obj/submachine/centcom_elevator` ->
`/obj/machinery/computer/elevator/centcomm`

I'm not gonna be able to repath the other side of the centcomm elevator nor be able to test it very well (it goes to the void then stops working without secrets)

ZeWaka:
> I'm not gonna be able to repath the other side of the centcomm elevator nor be able to test it very well (it goes to the void then stops working without secrets)

Could do an UpdatePaths but not really necessary for only one thing

PR #15784, opened 2023-09-10
Context menu fluid canisters

Garash2k:
## About the PR 
Replaces the fluid canister UI with Context menu buttons
image image image

They're tooltipped which helps differentiate IN/OUT, but if someone has better icon ideas I'm all ears. I took fontawesome icons and resized them to 14px +2px for padding. My initial idea was to use small text and it [didn't look great](https://cdn.discordapp.com/attachments/890226559691157524/1150332323125018634/image.png) even without realizing I needed padding in the already small 16px icon.

The description is updated to indicate what state it's on and how much it contains
image

Note: the state, and fluid levels, can also be determined by the already existing blinking light animation, and fluid meter on the sprite.

## Why's this needed? 
Prunes a raw HTML UI. Context menus are nice.

Old UI: 
261897129-649b9263-41d3-47ce-b396-b0d0c59ed853

## Changelog 

```changelog
(u)Garash
(+)Fluid canisters now use a Context Actions menu.
```
[UI][Rework]


pgmzeta:
i think rotating the icons you have so the arrows point up for in and down for out and will be a little easier to read

TobleroneSwordfish:
Maybe also make the power icon change colour based on state?

Garash2k:
Ok, IN/OUT buttons have been rotated, and the power icon turns green when IN and blue when OUT.
image
image

TobleroneSwordfish:
Err no, I meant the power button should change colour based on power state, but having the active flow button light up would also be cool.

Cherman0:
While im not strongly opposed, i feel like context menus arent the way to go here. Using context menus creates the problem of conveying information about states, and having it in inspect feels like a bandaid solution to a caused problem. Using a tgui would allow you to preserve the existing information presentation.

And in my opinion, radial context menus are just a bit grody in general.

Garash2k:
> Using context menus creates the problem of conveying information about states

It's true that popup menus are better at conveying information, but they're not exclusively upsides. They take focus away from the main game screen, both figuratively and literally. In this case, I don't think that the fluid canister's state is important enough. On top of the inspect text and the icon change, you can just see what it's doing after 1 tick. I think the immersion of using a context menu is worth more.

As for their look, we can always improve their look in another PR :)

> Err no, I meant the power button should change colour based on power state, but having the active flow button light up would also be cool.

Done!
image image image

Garash2k:
!merge_upstream

ZeWaka (approved):

PR #15786, opened 2023-09-10
Resprites hoodie hoods

Glamurio:
[Clothing]

## About the PR 
Resprites `/obj/item/clothing/suit/hoodie` hoods.

image

Also moves icons to `'icons/obj/clothing/overcoats/hoods/hoodies.dmi'` and their wear icon to `'icons/mob/clothing/overcoats/hoods/worn_hoodies.dmi'` for convenience.

There's a plan to add alpha masks for hairs to the hoods, but that will have to be in a follow-up PR, if at all.

## Why's this needed? 
Old sprite always looked a bit dorky to me.


444Portal:
These look really nice! Excited to see how the follow-up goes as well.

TheColorCyan:
i really hope alpha masks will be added as well

ZeWaka:
looks good

PR #15792, opened 2023-09-10
No Cyborg Beaker Shattering

### LINKED ISSUE #15777: Photophosphide (and other explosive chems) Permanantly Break Silicon Breaker Tools
### Labels

[BUG][MINOR]

### Description

per title. Probably intended, But also not good if both your breakers break.

### Steps to reproduce

1. make Photophosphide 
2. read too late about it exploding on contact with light
3. explosion!
4. Breaker breaks (naturally)
5. Realise you just lost one of your only breakers...

### Expected Behavior

honestly, I can see it break, but maybe a way to get it back after a bit would be nice.

I don't know if this is technically a "bug", considering the breakers SHOULD break, but things get weird once you add borg tools in the mix. No idea on how to make sure they still break without losing the breaker permanantly (until you get a new module)

Reported by: Musrom
Active test merges: #15458 and #14916
Round log date: 2023-09-09-20-13
Reported on: 2023-09-09 21:00:22
Map: DONUT2

Flaborized:
[BUG]

## About the PR 

This PR gives cyborgs a new variant of beaker that cannot shatter. Both mediborgs and chemborgs. It's pathed at /obj/item/reagent_containers/glass/beaker/large/cyborg and is literally the only distinction between them and regular large beakers.

## Why's this needed? 

Fixes #15777 , cyborg stuff isn't meant to break and I already implemented this for extinguishers, just forgot beakers. They can't get new ones, so they get shatter-proof stuff instead. Could possibly make it pour reagents on the ground or something if this is too lenient but most reactions that shatter beakers explode anyways so it's not like you *want* to blow yourself up. Shouldn't have to get a new module to replace this stuff imo.

### DIFF
--- a/code/modules/chemistry/tools/beakers.dm
+++ b/code/modules/chemistry/tools/beakers.dm
@@ -182,8 +182,7 @@
 	initial_reagents = "silver_sulfadiazine"
 
 /obj/item/reagent_containers/glass/beaker/large/cyborg
-	shatter_chemically(var/projectiles = FALSE)
-		return FALSE
+	shatter_immune = TRUE
 
 /*  Now found in hydroponics.dm!
 
--- a/code/modules/chemistry/tools/beakers.dm
+++ b/code/modules/chemistry/tools/beakers.dm
@@ -181,6 +181,10 @@
 	icon_state = "largebottle-burn"
 	initial_reagents = "silver_sulfadiazine"
 
+/obj/item/reagent_containers/glass/beaker/large/cyborg
+	shatter_chemically(var/projectiles = FALSE)
+		return FALSE
+
 /*  Now found in hydroponics.dm!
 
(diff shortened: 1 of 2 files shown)


code/modules/chemistry/tools/beakers.dm
```diff
@@ -181,6 +181,10 @@
 	icon_state = "largebottle-burn"
 	initial_reagents = "silver_sulfadiazine"
 
+/obj/item/reagent_containers/glass/beaker/large/cyborg
+	shatter_chemically(var/projectiles = FALSE)
+		return FALSE
```
TobleroneSwordfish, line 186: this works, but I added a `shatter_immune` var to `/obj/item/reagent_containers/glass` to fix this for the chemicompiler ones

Flaborized: makes sense, fixed!

PR #15796, opened 2023-09-11
Nuclear Meltdowns Scale By Radioactive Material

amylizzle:
## About the PR 
Instead of a flat explosion + dump of fallout gas, meltdowns now scale from "little explosion that destroys the engine room" to "the station is utterly destroyed and everyone is dead" depending on the amount of nuclear fuel in the reactor. Gone are the days of pipeburns outclassing TTVs with zero effort.

## Why's this needed? 
Currently you can blow up a sizable chunk of the station by just starting a pipeburn in the reactor. That's not really in the spirit of what the meltdown mechanic was about.

Meltdown badness is now calculated as a weighted sum of radioactivity, neutron radioactivity, and spent fuel in each component. Melted component count for twice the badness.

To see some typical values, call `catastrophicOverload()` on 
- `prefilled/normal` 
- `prefilled/meltdown` 
- `prefilled/insane` (this one is the absolute  upper limit and is not achievable through normal gameplay)

## Changelog 

```changelog
(u)Amylizzle
(*)Nuclear meltdowns now scale with the amount of radioactive fuel in the reactor.
```

### DIFF
--- a/code/obj/nuclearreactor/nuclearreactor.dm
+++ b/code/obj/nuclearreactor/nuclearreactor.dm
@@ -374,20 +374,50 @@
 		world << alarm //ew
 		command_alert("A nuclear reactor aboard the station has catastrophically overloaded. Radioactive debris, nuclear fallout, and coolant fires are likely. Immediate evacuation of the surrounding area is strongly advised.", "NUCLEAR MELTDOWN")
 
-		logTheThing(LOG_STATION, src, "[src] CATASTROPHICALLY OVERLOADS (this is bad)")
+
 		//explode, throw radioactive components everywhere, dump rad gas, throw radioactive debris everywhere
 		src.melted = TRUE
+		var/meltdown_badness = 0
+		var/turf/epicentre = get_turf(src)
 		if(!src.current_gas)
 			src.current_gas = new/datum/gas_mixture()
-		src.current_gas.radgas += 6000
-		src.current_gas.temperature = src.temperature
+
+		//determine how bad this meltdown should be
+		//basically this is a points system, more points = worser
+		//at 49 fresh melted cerenkite rods, this is 490 (feasible, but difficult). at 49 fresh melted plutonium rods this is 1274 (basically upper limit - it would insane to reach this without admemes)
+		//at 3 fresh melted cerenkite rods this is 30 (most common meltdown due to pipeburn)
+		//total sum of radioactive junk
+		for(var/x=1 to REACTOR_GRID_WIDTH)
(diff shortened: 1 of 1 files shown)


Sbmhawk:
I'm curious, how much radioactive material would be needed to get the same/similar explosion we get now?

code/obj/nuclearreactor/nuclearreactor.dm
```diff
@@ -374,20 +374,50 @@
...
+						throwcomp.throw_at(get_ranged_target_turf(epicentre,pick(alldirs),rand(1,20)),rand(1,20),rand(1,20))
+					else
+						qdel(src.component_grid[x][y])
+						var/obj/decal/cleanable/debris = make_cleanable(/obj/decal/cleanable/machine_debris/radioactive, epicentre)
+						debris.streak_cleanable(dist_upper=20)
+					src.component_grid[x][y] = null //get rid of the internal ref once we've thrown it out
+
+		src.current_gas.radgas += meltdown_badness*15
+		src.current_gas.temperature = max(src.temperature, src.current_gas.temperature)
 		var/turf/current_loc = get_turf(src)
```
pgmzeta, line 411: should epicentre on L381 replace current_loc here on L411?

amylizzle:
> I'm curious, how much radioactive material would be needed to get the same/similar explosion we get now?

To get the same as now, you'd need a total badness of 360, which is about equivelant to 36 melted cerenkite rods.

pgmzeta:
I think this does the math right https://www.desmos.com/calculator/syts3qtxij

amylizzle:
comment your math D:

pgmzeta:
Commented - https://www.desmos.com/calculator/ibckfvb42r

TobleroneSwordfish:
I think it could be a little bigger at lower fuel amounts, 36 cerenkite rods is a LOT more than the average reactor setup, and the default config barely explodes. Otherwise good change, every day we get more realistic nuclear accidents :)

amylizzle:
Alright, to get the same as before you need about 25 cerenkite rods now, which is still a lot but also not that hard to do. The upper limit on the explosion is higher too, although at those numbers it hardly makes a difference.

pgmzeta:
updated calculator for number tweaks https://www.desmos.com/calculator/fpat8m7tqe

Tarmunora (commented):
only concern here is that this doesn't really address being able to detonate the reactor by pumping in hot gas, just requires you to add more boom juice first

amylizzle:
Yeah Im fine with that tbh. Its still fun and you can still blow it up, but you have to put some effort in to take the rest of the station with you.

PR #15797, opened 2023-09-11
Add emergency flares

flappybatpal:
## About the PR 
Adds an emergency flare item, the more serious alternative to the glowstick which is also an ignition source!
Implements a loctargeting medium light component
3-6 minute duration with particle fx.
Won't fit in a pocket or storage after ignition.
Sprites by Flaborized.

## Why's this needed? 
I think in power outages or exploration these will be quite fun compared to glowsticks or flashlights.

## Changelog 

```changelog
(u)Flappybat & Flaborized
(*)Adds emergency flares to the game, when a glowstick just won't cut it! These come as standard issue in emergency toolboxes. Just be careful around flammable gas...
```


CheffieGithub:
Are there plans to increase their range under the `Fix light, which is barely visible except in the dark.` task?

image

While brighter I feel it should be larger than the PDA light is

flappybatpal:
image

Meidum light component go!

code/obj/item/device/lights.dm
```diff
@@ -735,3 +735,155 @@ TYPEINFO(/obj/item/device/light/floodlight)
...
+		playsound(user, 'sound/items/matchstick_light.ogg', 80, FALSE)
+		light_c.enable()
+
+		src.life_time = (world.time + rand(1800,3600))
+		processing_items |= src
+		if (istype(user))
+			user.update_inhands()
+		src.UpdateParticles(new/particles/roadflare_smoke,"roadflare_smoke")
+
+	proc/put_out(var/mob/user as mob)
```
frawhst, line 807:
```suggestion
	proc/put_out(mob/user)
```
Don't need the `as mob` stuff

code/obj/item/device/lights.dm
```diff
@@ -735,3 +735,155 @@ TYPEINFO(/obj/item/device/light/floodlight)
...
+		processing_items.Remove(src)
+		src.ClearSpecificParticles("roadflare_smoke")
+
+	temperature_expose(datum/gas_mixture/air, temperature, volume)
+		if (src.on == FLARE_UNLIT)
+			if (temperature > T0C+200)
+				src.visible_message("<span class='alert'>[src] ignites!</span>")
+				src.light()
+
+	ex_act(severity)
```
frawhst, line 828: This should call the parent, otherwise these will be indestructible to explosions (lighting off small explosions makes sense but it seems odd that a nuke wouldn't destroy them lol)

frawhst (commented):
Maybe also replace the glow sticks in the emergency crate from cargo with these.

code/obj/item/device/lights.dm
flappybatpal: Forgot plus this is based on a match so lol, matches are nuke proof

code/datums/components/simplelight.dm
```diff
@@ -167,6 +167,88 @@ TYPEINFO(/datum/component/loctargeting/sm_light)
 
 
 
+/datum/component/loctargeting/medium_light
```
TobleroneSwordfish, line 170: Pretty sure `/datum/component/loctargeting/sm_light` adds a medium light?
image
confusingly named though

code/obj/item/cigarette.dm
```diff
@@ -996,6 +996,9 @@
 				src.light()
 
 	ex_act(severity)
+		..()
+		if (!istype(src))
```
TobleroneSwordfish, line 1000: What's this supposed to do? `src` should always be of its given type by definition, if you're trying to check for it being deleted then do `QDELETED(src)` instead

code/obj/item/device/lights.dm
```diff
@@ -735,3 +735,158 @@ TYPEINFO(/obj/item/device/light/floodlight)
...
+	var/on = FLARE_UNLIT
+
+	var/life_time = 0
+	rand_pos = 1
+
+	var/col_r = 0.95
+	var/col_g = 0.7
+	var/col_b = 0.25
+	var/brightness = 0.4
+	var/datum/component/loctargeting/medium_light/light_c
```
TobleroneSwordfish, line 769: Ideally you shouldn't be storing references to components, if you want it to start off then there's a bool argument for that, and it can be turned on and off by sending `COMSIG_LIGHT_ENABLE` and `COMSIG_LIGHT_DISABLE` respectively

code/obj/item/device/lights.dm
```diff
@@ -735,3 +735,158 @@ TYPEINFO(/obj/item/device/light/floodlight)
...
+	proc/light(var/mob/user as mob)
+		src.on = FLARE_LIT
+		w_class = W_CLASS_BULKY
+		src.firesource = FIRESOURCE_OPEN_FLAME
+		src.icon_state = "roadflare-lit"
+
+		playsound(user, 'sound/items/matchstick_light.ogg', 80, FALSE)
+		light_c.enable()
+
+		src.life_time = (world.time + rand(1800,3600))
```
TobleroneSwordfish, line 801: use `SECONDS` or `MINUTES` defines here

code/obj/item/device/lights.dm
```diff
@@ -735,3 +735,158 @@ TYPEINFO(/obj/item/device/light/floodlight)
...
+
+	temperature_expose(datum/gas_mixture/air, temperature, volume)
+		if (src.on == FLARE_UNLIT)
+			if (temperature > T0C+200)
+				src.visible_message("<span class='alert'>[src] ignites!</span>")
+				src.light()
+
+	ex_act(severity)
+		..()
+		if (!istype(src))
```
TobleroneSwordfish, line 830: same here

TobleroneSwordfish (changes requested):

PR #15800, opened 2023-09-12
Update factions with new neutral faction and updates

CheffieGithub:
## About the PR 
Adds a new faction faction` FACTION_NEUTRAL` mobs with this faction aren't attacked by default.
Applied to cockroaches, non angry mice, firelies and float eyes for now. Since these are common "decorative" mobs.

Adds a new generic faction `FACTION_GENERIC` for generic groups that don't need a unique one.

Update abberation with a faction and update it to use `valid_target()` so it doesn't kill void mobs.

Prevents targeting of ghost critters and ghost drones by mob AI, since these roles should have minimal round impact they shouldn't be targeted by mobs.

Adds a global `proc/faction_check()` which takes mob A, mob B and whether A attacks neutral mobs. Returns` TRUE` if A attacks B `FALSE` otherwise.

## Why's this needed? 
While its funny for places to turn into blood baths before you even get there, mobs that are intended for flavour should stay as decoration and not get killed by hostile mobs. Likewise mobs in the same areas should ideally not attack eachother.

### DIFF
--- a/_std/macros/factions.dm
+++ b/_std/macros/factions.dm
@@ -1,22 +1,34 @@
 //Faction bitmasks, for setting what teams mobs/critters are on. Barebones atm.
 
+/// Faction which is not targeted by default
+#define FACTION_NEUTRAL			(1<<0)
+/// Generic faction for groups you don't want killing eachother
+#define FACTION_GENERIC			(1<<1)
 /// Wraith critters and summons
-#define FACTION_WRAITH			(1<<0)
-/// Maneaters, Tomatoes, Wasps and Plasmaspores
-#define FACTION_BOTANY			(1<<1)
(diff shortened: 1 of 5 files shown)


_std/macros/factions.dm
```diff
@@ -1,22 +1,34 @@
 //Faction bitmasks, for setting what teams mobs/critters are on. Barebones atm.
 
+/// Faction which is not targeted by default
+#define FACTION_NEUTRAL			(1<<0)
+/// Generic faction for groups you don't want killing eachother
+#define FACTION_GENERIC			(1<<1)
```
ZeWaka, line 6: that doesn't make sense. this would make more sense as a 'default faction' or similar.

ZeWaka (changes requested):

PR #15808, opened 2023-09-12
Make borgs climb railings

### LINKED ISSUE #15807: borgs cannot climb rails
### Labels

[BUG]

### Description

title. you can no longer climb railing as a borg by clicking or walking into it

### Steps to reproduce

1. be a cyborg
2. encounter a railing
3. cope because clicking it and walking into it dont work

### Expected Behavior

be able to climb rails

_No response_

TealSeer:
[BUG] [SILICONS]
## About the PR 

Fixes #15807 by implementing attack_ai for railings and changing Bumped() to work for borgs too.

## Why's this needed? 

Probably an important thing for borgs to be able to do

### DIFF
--- a/code/obj/railing.dm
+++ b/code/obj/railing.dm
@@ -151,6 +151,8 @@
 		src.try_vault(user)
 
 	attack_ai(mob/user)
+		if(!can_reach(user, src) || isAI(user) || isAIeye(user))
+			return
 		return src.attack_hand(user)
 
 	Bumped(var/mob/AM as mob)
--- a/code/obj/railing.dm
+++ b/code/obj/railing.dm
@@ -150,10 +150,13 @@
 	attack_hand(mob/user)
 		src.try_vault(user)
 
+	attack_ai(mob/user)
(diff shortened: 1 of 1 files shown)


code/obj/railing.dm
```diff
@@ -150,10 +150,13 @@
 	attack_hand(mob/user)
 		src.try_vault(user)
 
+	attack_ai(mob/user)
+		return src.attack_hand(user)
+
```
Tarmunora, line 155: Needs either appropriate range checks, or just drop this in favor of just the `Bumped()` support

Tarmunora (changes requested):

code/obj/railing.dm
TealSeer: I thought it was okay to do because the action bar has range checks but I can add them there too if you'd like

TealSeer: ok nvm it is not okay to do I see what you mean lol

Tarmunora (approved):

PR #15812, opened 2023-09-12
Chemistry Cryoxadone + Pyrosium Area Heat Effect + Slight Recipe Tweaks

Flaborized:
[CHEMISTRY]

## About the PR 

This PR changes a few chemistry things with cryoxadone, cryostylane and pyrosium, nothing super major:

- Cryoxadone now non-instant, speed 6u.
- Mixing cryoxadone now creates a cold area effect around it, cooling people and beakers and stuff.
- You can also mix cryoxadone with oxygen in a new reaction that creates the effect, using up the cryoxadone. (Speed 1u)
- Pyrosium can be mixed with magnesium (instead of oxygen) to create a similar heat effect, with little flame animations (more dangerous with flammables nearby for sure)
- All of these reactions and pyrosium + oxygen and cryostylane + oxygen now don't consume the oxygen/magnesium *when* reacting, but keep count of how many cycles have passed and removes the oxy/mag right at the end in equal amounts. This means that if you mix 50u cryo and 50u oxy, the end of the reaction will deplete everything, but you can also just use 1u or 10u oxy if you prefer. Meant to be QoL and let you keep these reactions running longer if you want, vaguely like oil.

https://github.com/goonstation/goonstation/assets/53125172/0175fb88-7110-4323-8caf-4d2aef017048

## Why's this needed? 

Something I wanted to add for a bit, I think I also saw someone in discord suggest something similar, I think the utility of being able to heat/cool multiple beakers with one reaction based on positioning is really cool, and I like adding more dramatic visuals to chemistry. I also think the "only technically requiring 1u of oxy" buffs to the cryo and pyro reactions should make them much more fun to use and cut down on some clicking.

Feedback on the changelog entries very welcome, not my necessarily strong suit :sheltersweat:

## Changelog 

```changelog
(u)Flaborized
(*)Some small chemistry changes to pyrosium, cryoxadone, and cryostylane, including an area of effect heat/cool effect. More detailed description in minor logs.
(+)Pyrosium can now be mixed with magnesium to create a 3x3 area of effect heating effect.
(+)Cryoxadone now non-instant (speed 6u), mixing creates a 3x3 area of effect cooling effect, can also be mixed with oxygen to get the same effect.
(+)Pyrosium, cryostylane and cryoxadone reactions now longer use up the oxygen/magnesium until the end of the reaction. This means it'll still deplete the oxygen, but you only technically need 1u to carry out the full reaction.
```


code/modules/animation/AnimationLibrary.dm
```diff
@@ -617,6 +617,24 @@ proc/muzzle_flash_any(var/atom/movable/A, var/firing_angle, var/muzzle_anim, var
 	icon_state = "shine"
 	plane = PLANE_OVERLAY_EFFECTS
 
+/obj/particle/cryo_sparkle
+	icon = 'icons/effects/chemistry_effects.dmi'
+	icon_state = "cryo-1"
+	plane = PLANE_OVERLAY_EFFECTS
+
+	New()
+		icon_state = pick("cryo-1", "cryo-2", "cryo-3", "cryo-4") //slightly different timings on these to give a less static look
```
ZeWaka, line 626: consider just having the same sprite (less work) but offsetting horizontally for some of these, so it's less 'tiled'

wolfolotl:
very neat, though I worry that the visual effects for the heat aura make it look equivalent to an area of open flame in the sense that it communicates a level of potential destruction that isn't there. it's a little more Actual Fire than a tile of an inferno, currently

Tarmunora (changes requested):
Using croxadone as a cooling agent is really weird, considering when acting as a medicine it heats the patient up (and requires cooling the patient to actually do anything etc.). I don't think this is a good direction to take it, personally

Flaborized:
> Using croxadone as a cooling agent is really weird, considering when acting as a medicine it heats the patient up (and requires cooling the patient to actually do anything etc.). I don't think this is a good direction to take it, personally

forgot about that honestly :sheltersweat: will change it to be a variant of crostylane then (like how the other one is pyrosium + magnesium instead of oxygen

> very neat, though I worry that the visual effects for the heat aura make it look equivalent to an area of open flame in the sense that it communicates a level of potential destruction that isn't there. it's a little more Actual Fire than a tile of an inferno, currently

sure, I see what you mean, though I still want it to look like a heat-related hazard since it can light up fuel tanks and stuff. will try and draw something up that communicates this and then probs use this flame effect for something more fire-y later since I do think it looks pretty cool

Flaborized:
Switched the animation for the heat effect to this more abstract warm swirlies, hopefully looks less like it should burn you to death. Kept the fire in the file though since it looks nice and would fit well on something that actually does burn you to death later. It still looks a bit like a hazard but it's difficult to draw something that looks hot but not dangerous and I think this is Better.

Changed the cryoxadone reactions back to regular, instead moved the cold aoe to cryostylane + iodine. No particular reason I used iodine, magnesium and it were picked mostly because they're seldom used in recipes so you won't get a lot of conflicts or accidental triggers.

Tarmunora (approved):

PR #15817, opened 2023-09-13
add door linked atmos shields to donut 2

Retrino:
## About the PR 
adds door linked atmospheric shields to external airlocks on donut 2. space facing doors on inner and outer maintenance tunnels, engineering, solars, an external airlock in sci, etc have received them

## Why's this needed? 

keep air from leaking out of random doors on donut 2. other maps have them, may as well have them on this map too


Retrino:
why the frick is clarion here. go away

PR #15819, opened 2023-09-13
Cyborg cell cables usable on other cyborgs

kbsmilk:
[SILICONS] [FEATURE]
## About the PR 
* Cyborg cell cables can now donate charge to other /mob/living/silicon.
* You cannot siphon power from silicons.
* It's a copy of the original's action bar with adjusted messages, and removed interrupt_flag INTERRUPT_ATTACKED to allow for silly business.

## Why's this needed? 
Has some niche usage in off-station ops, like azones. Also, sensible that jumper cables would work on anything with a cell.

Silly Business:
dreamseeker_SFqIFGTCvG

## Changelog 

```changelog
(u)Smilg
(+)Cyborg cell cables can now be used on other cyborgs.
```

### DIFF
--- a/code/obj/submachine/robotics.dm
+++ b/code/obj/submachine/robotics.dm
@@ -200,9 +200,8 @@
 			if (jumper.positive)
 				donor_cell = silicon_user.cell
 				recipient_cell = silicon_target.cell
-			else if (!silicon_user.emagged) // don't let us drain other cyborgs unless we're emagged
+			else if (!silicon_user.emagged)
 				boutput(user, "<span class='alert'>A core law submodule limits you from draining the power of other cyborgs!</span>")
-				// also hint to the user that being emagged removes this restriction
 				..()
 				interrupt(INTERRUPT_ALWAYS)
 				return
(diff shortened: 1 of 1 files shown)


Tarmunora:
Not a fan of having an EMAG interaction in here. Emagged cyborgs are just cyborgs with no laws, without changing anything in the modules themselves, and I'd rather not open up that can of worms

kbsmilk:
> Not a fan of having an EMAG interaction in here. Emagged cyborgs are just cyborgs with no laws, without changing anything in the modules themselves, and I'd rather not open up that can of worms

Ya, I wasn't sure whether to permit/block draining other cyborgs, so, I settled on that. Thoughts? I lean towards block entirely, but no strong feelings, here.

Tarmunora:
> > Not a fan of having an EMAG interaction in here. Emagged cyborgs are just cyborgs with no laws, without changing anything in the modules themselves, and I'd rather not open up that can of worms
> 
> Ya, I wasn't sure whether to permit/block draining other cyborgs, so, I settled on that. Thoughts? I lean towards block entirely, but no strong feelings, here.

I'd lean towards block. Alternatively, could rework the jumpers in such a way that they can only reach an equilibrium of charge (i.e.: can't easily drain to 0)

kbsmilk:
Block it is. Noting that idea down though, might fold it in for future ideas.

Tarmunora (approved):

pgmzeta (commented):
pronoun function suggestions

PR #15820, opened 2023-09-13
Assless chaps, uniform -> Chaps, outer suit

kbsmilk:
[GAME OBJECTS] [FEATURE]
## About the PR 
* Assless chaps are now just called chaps.
* Moved to outer suit slot due to its sprite. Repathed from /obj/item/clothing/under/gimmick/chaps to /obj/item/clothing/suit/chaps, within code\obj\item\clothing\gimmick.dm.
* Sprites are now moved, to the 3 gimmick overcoat .dmis.

## Why's this needed? 
This "leather fetish item" is not good anymore

## Changelog 

```changelog
(u)Smilg
(+)Assless chaps replaced with regular chaps.
```


flappybatpal:
Between the inclusion of the strapped top and the name (there is no such thing as assless chaps, it's just chaps) it does seem like a poorly aged item.

kbsmilk:
Sprites are now moved, also the obj def is moved to the gimmick clothes super-file.

Tarmunora (approved):

PR #15824, opened 2023-09-13
Add NT commander to all rotation maps

Valtsu0:
## About the PR 
Adds NT Commander to the space diner and nadir. Makes all syndicate drones drop boards, boss drones dropping 10. 

kuva
kuva

## Why's this needed? 
Having a reason to hunt drones on other maps is nice

## Changelog 

```changelog
(u)Valtsu0
(*)NT commander can be found eating spaghetti on the space diner and spectating boxing matches on Nadir.
```


TobleroneSwordfish:
Seems odd for an NT official to be on the space diner instead of the station.

DisturbHerb:
Given how originally ocean-centric (hahah. oshan-centric) the NT commander was, I don't think the commander's rewards are the most thematically relevant outside of that at the moment. Right?

Valtsu0:
> Seems odd for an NT official to be on the space diner instead of the station.

The main reason i put in the diner was because i didn't want it being in the way of any construction projects, and because it takes more work. If the syndicate assest he is investigating is the syndicate shuttle on the debris field, it kinda makes sense for him the be on the diner. But i can change it.



> Given how originally ocean-centric (hahah. oshan-centric) the NT commander was, I don't think the commander's rewards are the most thematically relevant outside of that at the moment. Right?

Cairngorm is classified as a battlecruiser, the station is led by a captain. I think it fits enough, well maybe with the exception of the seaman uniform.

PR #15833, opened 2023-09-14
Steals sec's crematorium access on RP

frawhst:
## About the PR 
yoink
Also takes away the HoP's access on both servers because cmon they don't need that

There's already a thread so no feedback tag:
https://forum.ss13.co/showthread.php?tid=21785
## Why's this needed? 
i am game balancer

## Changelog 
```changelog
(u)aloe
(+)Removed security's access to the crematorium on the RP servers.
```


cheekybrdy:
In my personal opinion as a classic only, I'd rather the HoP had crematorium access as its part of their department through chaplain. Imho I'd rather HoP had chap office too for full AA in their own department by default. A big thing to consider is that at least on classic most HoPs will just give AA to themselves so giving the ones that don't a lack of access to something in their department may feel unrewarding for doing "the right thing"

PR #15836, opened 2023-09-14
Dialysis machine

DisturbHerb:
[MEDICAL][OBJECTS][FEATURE]

## About the PR 
This PR does not introduce any mapping changes.

This PR introduces the dialysis machine to the game's code, as described in [this forum thread.](https://forum.ss13.co/showthread.php?tid=21781) It operates in much the same way as an IV drip; clickdragging patients onto it will (dis)connect them from the machine, and can be (un)anchored with an empty hand.

dialysis horror blod

It takes 10u of blood in on each machine process tick from a patient, scrubs it of anything outside of the global chem whitelist, and attempts to put it back in if the patient's blood volume is less than 500. The machine can store around 500u of blood which it will continuously attempt to re-introduce. Stopping the machine will purge its internal container of reagents, and they cannot be accessed via any means.

image

Slapping it with an EMAG reverses the chemical filter; it will destroy anything on the whitelist and pass non-whitelist chemicals back into the body.

## Why's this needed? 

This would introduce another method of purging chemicals from the body with fewer safety risks than heart squeezing with the trade-off of it being comparatively slower. While the machine works, one could still apply most medical chems so as to not run into the problems with other purgatives like charcoal or pent. In spite of this, some whitelist chems like meth will not be scrubbed, introducing some potential counterplay (even if meth is pretty much exclusively good?).


DisturbHerb:
I thought it could just start injecting you with heparin, but eh.

amylizzle:
Any way a visual indicator could be added to show it being connected to the patient?

Also proposed emag effect: instead of filtering chems from blood, filter blood from chems.

DisturbHerb:
> Any way a visual indicator could be added to show it being connected to the patient?

At the moment, the screen just turns off when it's not plugged in; there's also visual feedback from the screen flashing. Suggestions are appreciated.

Garash2k:
Has ricin been considered? Since ricin messes with purgative speeds, maybe it should mess with this mahine too, unless the counterplay is intended.

DisturbHerb:
> Has ricin been considered?...

This machine doesn't work like a purgative, it physically draws blood from the patient like a syringe or IV with the `transfer_blood` proc. It does not account for depletion rates.

amylizzle:
I wonder what it'd do to someone infected with gnesis. I expect they'd probably just end up with severe hypotension.

Valtsu0:
I don't think Hemolymph is on the whitelist. Does this just filter out all of roachperson's blood?

DisturbHerb:
> I don't think Hemolymph is on the whitelist. Does this filter out roachperson's blood?

Only one way to find out.

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,169 @@
+TYPEINFO(/obj/machinery/dialysis)
+	mats = list("MET-1" = 20, "CRY-1" = 5, "CON-2" = 5)
+
+/obj/machinery/dialysis
+	name = "dialysis machine"
```
Kyle2143, line 5: "Space"

DisturbHerb: No.

444Portal: This and the response shouldn't have made me giggle. Heheh.

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,169 @@
...
+					"<span class='notice'><b>[user]</b> removes [src]'s cannulae from [new_patient == user ? "[his_or_her(new_patient)]" : "[new_patient]'s"] arm.</span>",\
+					"<span class='notice'>You remove [src]'s cannulae from [new_patient == user ? "your" : "[new_patient]'s"] arm.</span>",\
+					"<span class='notice'>[new_patient == user ? "You remove" : "<b>[user]</b> removes"] [src]'s cannulae from your arm.</span>")
+					return src.stop_dialysis()
+				else return boutput(user, "<span class='alert'>[src] already has a patient attached!</span>")
+			new_patient.tri_message(user,\
+			"<span class='notice'><b>[user]</b> begins inserting [src]'s cannulae into [new_patient == user ? "[his_or_her(new_patient)]" : "[new_patient]'s"] arm.</span>",\
+			"<span class='notice'>[new_patient == user ? "You begin" : "<b>[user]</b> begins"] inserting [src]'s cannulae into your arm.</span>",\
+			"<span class='notice'>You begin inserting [src]'s cannulae into [new_patient == user ? "your" : "[new_patient]'s"] arm.</span>")
+			logTheThing(LOG_COMBAT, user, "tries to hook up a dialysis machine [log_reagents(src)] to [constructTarget(new_patient,"combat")] at [log_loc(user)].")
```
TobleroneSwordfish, line 71: log whether it's emagged too

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,169 @@
...
+		if (src.reagents.is_full())
+			src.reagents.clear_reagents()
+			src.audible_message("<span class='game say'><span class='name'>[src]</span> beeps, \"Blood buffer has reached maximum capacity. Purging internal reservoir.\"")
+
+		transfer_blood(src.patient, src, src.draw_amount)
+
+		// Re-implemented here due to all the got dang boutputs.
+		var/list/whitelist_buffer = src.whitelist + src.patient_blood_id
+		for (var/reagent_id in src.reagents.reagent_list)
+			if ((!src.hacked && !whitelist_buffer.Find(reagent_id)) || (src.hacked && whitelist_buffer.Find(reagent_id)))
```
TobleroneSwordfish, line 105: use `y in x` instead of `.Find`

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,169 @@
...
+				src.reagents.del_reagent(reagent_id)
+
+		src.output_blood_colour = src.reagents.total_volume ? src.reagents.get_average_color().to_rgba() : null
```
TobleroneSwordfish, line 108:
```suggestion
		src.output_blood_colour = src.reagents.total_volume ? src.reagents.get_average_rgb() : null
```

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,169 @@
...
+
+		// Infuse blood back in if possible. Don't wanna stuff too much blood back in.
+		// The blood that's not actually in the bloodstream yet, know what I mean?
+		var/non_blood_volume_blood = src.patient.reagents.reagent_list["blood"]?.volume || 0
+		if (!src.patient.reagents.is_full() && (src.patient.blood_volume + non_blood_volume_blood) <= 500)
+			src.reagents.trans_to(src.patient, src.draw_amount)
+			src.patient.reagents.reaction(src.patient, INGEST, src.draw_amount)
```
TobleroneSwordfish, line 115: doesn't an ingest reaction do the transfer anyway?

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,169 @@
...
+		src.UpdateOverlays(image(src.icon, "screen-on"), "screen")
+		src.UpdateIcon()
+		SubscribeToProcess()
+
+	proc/stop_dialysis()
+		UnsubscribeProcess()
+		src.patient = null
+		src.patient_blood_id = null
+		src.output_blood_colour = null
+		if (!!src.reagents.total_volume)
```
TobleroneSwordfish, line 165:
```suggestion
		if (src.reagents.total_volume)
```

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,169 @@
...
+	var/last_in = 0
+	var/last_out = 0
+	var/output_blood_colour
+	/// Reagent ID of the current patient's blood.
+	var/patient_blood_id
+
+	New()
+		..()
+		src.create_reagents(100)
+		if (islist(chem_whitelist) && length(chem_whitelist))
```
TobleroneSwordfish, line 28: pointless check as it defaults to null anyway, if you want to check for this failure state then CRASH or stack trace instead

TobleroneSwordfish (changes requested):
Very cool, more interesting medical equipment is good and I think this fills a niche? Might need numbers balancing, hard to test.

code/obj/machinery/dialysis.dm
DisturbHerb: Surprisingly, no. The `trans_to()` line is essential. 😔

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,172 @@
...
+
+	New()
+		..()
+		src.create_reagents(100)
+		if (!length(chem_whitelist))
+			CRASH("[src] tried to fetch the global chem whitelist but it has a length of 0!")
+		src.whitelist = chem_whitelist
+		src.UpdateOverlays(image(src.icon, "pump-off"), "pump")
+		src.UpdateOverlays(image(src.icon, "screen-off"), "screen")
+		src.UpdateOverlays(image(src.icon, "tubing"), "tubing")
```
pali6, line 33: Wouldn't this make more sense as a part of `update_icon`?

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,172 @@
...
+			"<span class='notice'><b>[user]</b> begins inserting [src]'s cannulae into [new_patient == user ? "[his_or_her(new_patient)]" : "[new_patient]'s"] arm.</span>",\
+			"<span class='notice'>[new_patient == user ? "You begin" : "<b>[user]</b> begins"] inserting [src]'s cannulae into your arm.</span>",\
+			"<span class='notice'>You begin inserting [src]'s cannulae into [new_patient == user ? "your" : "[new_patient]'s"] arm.</span>")
+			logTheThing(LOG_COMBAT, user, "tries to hook up a dialysis machine [log_reagents(src)] to [constructTarget(new_patient,"combat")] at [log_loc(user)].")
+			SETUP_GENERIC_ACTIONBAR(user, src, 3 SECONDS, PROC_REF(cannulate), list(new_patient, user), src.icon, "dialysis-map", null, null)
+		..()
+
+	process(mult)
+		..()
+		if (!src.patient || !ishuman(src.patient))
```
pali6, line 80:
```suggestion
		if (!src.patient || !ishuman(src.patient) || QDELETED(src.patient))
```

pali6: Though that's probably also caught in in_interact_range below. But idk makes more sense here.

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,172 @@
...
+			"<span class='notice'>You insert [src]'s cannulae into [new_patient == user ? "your" : "[new_patient]'s"] arm.</span>")
+		logTheThing(LOG_COMBAT, user, "connects a dialysis machine [log_reagents(src)] to [constructTarget(new_patient,"combat")] at [log_loc(user)].")
+		src.start_dialysis()
+
+	proc/start_dialysis()
+		if (!src.patient) return
+		src.power_usage = 500
+		src.patient_blood_id = src.patient.blood_id
+		src.UpdateOverlays(image(src.icon, "pump-on"), "pump")
+		src.UpdateOverlays(image(src.icon, "screen-on"), "screen")
```
pali6, line 159: Again, wouldn't this make sense in update_icon instead?

code/obj/machinery/dialysis.dm
```diff
@@ -0,0 +1,172 @@
...
+	proc/cannulate(mob/living/carbon/new_patient, mob/user)
+		src.patient = new_patient
+		new_patient.tri_message(user,\
+			"<span class='notice'><b>[user]</b> inserts [src]'s cannulae into [new_patient == user ? "[his_or_her(new_patient)]" : "[new_patient]'s"] arm.</span>",\
+			"<span class='notice'>[new_patient == user ? "You insert" : "<b>[user]</b> inserts"] [src]'s cannulae into your arm.</span>",\
+			"<span class='notice'>You insert [src]'s cannulae into [new_patient == user ? "your" : "[new_patient]'s"] arm.</span>")
+		logTheThing(LOG_COMBAT, user, "connects a dialysis machine [log_reagents(src)] to [constructTarget(new_patient,"combat")] at [log_loc(user)].")
+		src.start_dialysis()
+
+	proc/start_dialysis()
```
pali6, line 154: I think it'd be a tiny bit cleaner if this proc got a `new_patient` argument and set `src.patient` to that. Also I don't see any handling of what happens when old patient is overwritten with a new patient. Shouldn't that at least have its own message or something?

DisturbHerb: Patients can't (or shouldn't) be overridden without removing the current patients first, with how this has been implemented.

TobleroneSwordfish (approved):
It's a bit janky that using this will eventually end up making you hypotensive with a bloodstream full of the "blood" chem, but also dialysis can make you hypotensive IRL so... feature??

DisturbHerb:
> It's a bit janky that using this will eventually end up making you hypotensive with a bloodstream full of the "blood" chem, but also dialysis can make you hypotensive IRL so... feature??

I've never actually seen that happen in my testing under normal conditions, as you'll reach an equilibrium where more blood is pulled from your reagent container than your blood volume.

PR #15839, opened 2023-09-14
Adds sounds to the Jester's Hat and Jester's Mask

CrystalClover:
[CLOTHING] [FEATURE] [SOUND]
## About the PR 
Adds a jingling bell to the Jester's Hat as the players walks
Adds a laughing effect to the Jester's Mask when interacted with (similar to the clown mask)
also changes jester_laugh.ogg to laugh_evil.ogg to seperate it from the new jester_laugh_1.ogg &jester_laugh_2.ogg

All sounds added were recorded and edited by me

https://github.com/goonstation/goonstation/assets/94804240/e134218e-3b99-4559-bc1a-38c937cef971

## Why's this needed? 
Makes the Jester outfit have more similarities with the base clown, giving it more flavour.

```changelog
(u)CrystalClover
(+)Added jingling bell sounds to the Jester's Hat
and laughing sounds to the Jester's Mask when used

```


code/obj/item/clothing/masks.dm
```diff
@@ -641,6 +641,25 @@ TYPEINFO(/obj/item/clothing/mask/wrestling)
 	item_state = "jester"
 	see_face = 0
 
+	var/spam_flag = 0
+	var/spam_timer = 100
+	var/list/sounds_instrument = list('sound/misc/jester_laugh_1.ogg', 'sound/misc/jester_laugh_2.ogg')
```
Tarmunora, line 646:
```suggestion
	var/list/sounds_instrument = list('sound/misc/jester_laugh_1.ogg', 'sound/misc/jester_laugh_2.ogg')
```
...instrument?

code/obj/item/clothing/masks.dm
```diff
@@ -641,6 +641,25 @@ TYPEINFO(/obj/item/clothing/mask/wrestling)
...
+	var/volume = 100
+	var/randomized_pitch = 1
```
Tarmunora, line 648: do these need to be vars?

code/obj/item/clothing/masks.dm
```diff
@@ -641,6 +641,25 @@ TYPEINFO(/obj/item/clothing/mask/wrestling)
...
+
+
+	proc/jester_giggle(mob/user as mob)
+		if (!spam_flag)
+			spam_flag = 1
```
Tarmunora, line 653: use the `ON_COOLDOWN` macro instead

Tarmunora (changes requested):

code/obj/item/clothing/masks.dm
CrystalClover: added all the changes you suggested

TobleroneSwordfish (changes requested):
The jingling is cool, but I'm not a fan of the laugh sound, it's pretty weird and in general we don't have voice acting in this game aside from some very cursed cluwne laughter.

CrystalClover:
> The jingling is cool, but I'm not a fan of the laugh sound, it's pretty weird and in general we don't have voice acting in this game aside from some very cursed cluwne laughter.

Do you have any suggestions of a replacement for the laugh as the use noise for the mask?

flappybatpal:
Unless anyone can think of anything the answer is probably nothing. It doesn't need to try and replicate the clown mask.

PR #15842, opened 2023-09-15
Adds a large storage that can only be pried open for QM critter imports

CheffieGithub:
## About the PR 
Adds a type of large storage which can only be pried open with a crowbar, cannot be broken by flipping and cannot be closed or fixed once opened. Meant for import kits or crates that mappers use that are filled with mobs they don't want to get free early.
Adds a subtype of this for an animal crate with its own sprite based on the wooden crates and uses it for critter imports.
image

## Why's this needed? 
Cases of lots of birds in space have been on a rise and people want their money back. Mappers don't like mobs getting out of their boxes.

## Changelog 

```changelog
(u)Cheffie
(+)Crates ordered from cargo containing critters now need to be pried open upon arrival.
```


pgmzeta:
if there is nothing you can do with the crate after, why not have it turn into a sheet on opening?

CheffieGithub:
If they want it to be sheets they can wrench it, it makes no sense for it to fall apart because it was pried open

PR #15844, opened 2023-09-15
Make playsound's vary var use TRUE/FALSE defines

TheTyrant25:
[INTERNAL] [SOUND]
## About the PR
proc/playsound has several arguments. One of them is `vary`. Looking into the generatesound proc, it turns out that this is used as a boolean to do pitch shifting, but many many uses of playsound use `1` or `0` instead of the defines.
Thousands of playsound proc calls were edited using the following two regexes:
`(playsound\(\w*,\s?(\w|('.*'))*,\s?\d*,\s?)(1)([,)])`
into `$1TRUE$5`
and
`(playsound\(\w*,\s?(\w|('.*'))*,\s?\d*,\s?)(0)([,)])`
into `$1FALSE$5`

## Why's this needed?
Cleaner code is better. It should be more obvious what that var actually does.


TheTyrant25:
this pr contains only the find replace. No other changes were made.

frawhst:
“Hey can you make unloading guns play a slightly different sound from loading guns”
2 days pass
359 files changed 
truly this is the Goonstation Coding Experience

TheTyrant25:
it needed to be done.

PR #15853, opened 2023-09-16
Add ability to wield shinai in one hand & wear on belt

Glamurio:
[Objects] [Balance]
## About the PR 
Adds the ability to wield shinai in a single hand by default. Doing so will not allow any of its unique stance-specific combat methods, but will default to its default help values and make it a bit less cumbersome to use.

Using a shinai in your hand will dual-wield it, at which point it functions exactly the same as it did before.

Additionally, it allows shinai to be worn on a belt, and they continue to only be stored in shinai bags.

image

Lastly, it adds a unique suicide animation for shinai, in which the user gently places the sword on the ground, recites one of several appropriate Japanese proverbs (inspired by the katana), while cherry blossoms briefly fall to the ground as they perish.

List of proverbs:
"Take wa yowaku magaru ga ki wa tsuyoshi gouku tatsu." - "The bamboo that bends is stronger than the oak that resists."
"Kyoufu wa kokoro ga yurusu kagiri fukai monode shika nai." - "Fear is only as deep as the mind allows."
"Mizu ni nagasu." - "The water flows."
"Anshin shite shineru." - "I die in peace."
"Shiranu ga hotoke." - "Ignorance is bliss."
"Makeru ga kachi." - "To lose is to win."
"Ashita wa ashita no kaze ga fuku." - "Tomorrow’s winds will blow tomorrow."

## Why's this needed? 
Shinai are cool and deserve some love. Allowing them to be weakly held in one hand shouldn't affect anything too much, but will allow people to sort their inventory or do other things, and items being wearable on belts is always cool, especially since carrying a shinai bag is cumbersome.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Shinai can now be held in one hand and worn on a belt. They also have a new tranquil suicide animation.
```


DisturbHerb:
Don't think these should be wearable on the belt, it doesn't quite make sense with the other swords in the game.

DisturbHerb:
Moreover, shinai are supposed to be two-handed implements anyway. I'd suggest looking into nito-ryu if you want some one-handed shinai.

Glamurio:
> Don't think these should be wearable on the belt, it doesn't quite make sense with the other swords in the game.
Shinai aren't swords, like, very categorically they are `obj/item`. They have their own separate system of mechanics that doesn't deal with any delimb or parry systems with other swords. By that sense, they behave more like a crowbar than a sword.
I'd have added a sheath akin to a smaller shinai bag, but found that to be an odd solution. I also tried shinai bags worn on the belt but that looked visually just not right because it's basically a duffle bag, and those don't look good on the belt.

> Moreover, shinai are supposed to be two-handed implements anyway. I'd suggest looking into nito-ryu if you want some one-handed shinai.
I'm aware, and like I said in the PR, they only still work as a shinai when manually held in two-hands. When holding them in one hand, it's really more of a convenience for taking them out of a bag and giving them to someone else, without freeing your other hand. Functionally they only turn into a "shinai" when held in two hands, when held in one hand they're basically just a stick.

DisturbHerb:
> Shinai aren't swords, like, very categorically they are `obj/item`. They have their own separate system of mechanics that doesn't deal with any delimb or parry systems with other swords. By that sense, they behave more like a crowbar than a sword.

I don't think their arbitrarily set type path should determine whether or not they fit on the belt. Even allowing them to be held in one hand should be ample enough of a QoL improvement to me.

Shinai are big and cumbersome. I don't think I could reasonably shove one through the belt of my hakama, but that could just be on account of me being short. I'll have to test that.

flappybatpal (changes requested):
I don't think it makes sense for a weapon of this size to be belt rather than back.

PR #15854, opened 2023-09-16
Allows You To Pour Cut Open IV Bags on Pustules

Flaborized:
[CHEMISTRY]

## About the PR 

This PR lets you pour cut open IV bags (including blood bags) on synthflesh pustules. It pours the whole thing in one go. This is equivalent to pouring a beaker, so if there's things that won't go into pustules in the bag, it won't go into the pustule.

## Why's this needed? 

Qol! It's useful to use blood bags on pustules but clicking a beaker 10 times to fill it up and then splashing on the pustule is annoying, and there's already a built-in delay to how much synthflesh you can make per second anyways so it's not like there's a real balance reason you should have to click that much. 

## Changelog 

```changelog
(u)Flaborized
(+)You can now directly pour into synthflesh pustules from cut open blood bags.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Tools.dm
+++ b/code/modules/chemistry/Chemistry-Tools.dm
@@ -894,6 +894,22 @@ proc/ui_describe_reagents(atom/A)
 		..()
 
 	attackby(var/obj/item/W, mob/user)
+		if(istype(W, /obj/item/reagent_containers/iv_drip))
+			var/obj/item/reagent_containers/iv_drip/iv = W
+			if(!iv.slashed)
+				boutput(user, "<span class='alert'>The [iv.name] needs to be cut open first!</span>")
+				return
+			else if (reagents.total_volume >= reagents.maximum_volume)
+				boutput(user, "<span class='alert'>The [src] is too full!</span>")
+				return
+			else if (!iv.reagents.total_volume)
+				boutput(user, "<span class='alert'>The [iv.name] is empty!</span>")
+				return
+			else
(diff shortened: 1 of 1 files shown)


code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -894,6 +894,22 @@ proc/ui_describe_reagents(atom/A)
 		..()
 
 	attackby(var/obj/item/W, mob/user)
+		if(istype(W, /obj/item/reagent_containers/iv_drip))
+			var/obj/item/reagent_containers/iv_drip/iv = W
+			if(!iv.slashed)
+				boutput(user, "<span class='alert'>The [iv.name] needs to be cut open first!</span>")
+				return
+			else if (reagents.total_volume >= reagents.maximum_volume)
```
pgmzeta, line 902: would using [trans_to](https://github.com/goonstation/goonstation/blob/a6950c7e5705a4d4973a3262a27cbd59fde6a41a/code/modules/chemistry/Chemistry-Holder.dm#L295) on the IV's reagent holder cause any trouble? It does some of these checks too

Flaborized: The reason I made it add chems this way instead of doing it directly is because currently all of the checks pustules do to play sounds/reject chems is on reagent_act(), which is triggered by pouring stuff on it. It's probably be better to rewrite that code to work with trans_to and then just call that on reagent_act though in hindsight I guess? Just swapping it to trans_to as is though would let you pour stuff into pustules you shouldn't be able to though

Tarmunora (approved):

frawhst:
The fact that all the chem-pouring code is so hard coded is unfortunate but this is the best way to do it for the time being.

PR #15855, opened 2023-09-16
Take the brunt of grenade and pipebomb explosions by laying on top and blocking

pgmzeta:
## About the PR 

When grenades and pipe-bombs detonate, add the following:
1. If the explosive is directly on a tile (e.g. not in a closet)
2. Get a list of human mobs on the same tile and filter them
  a. is directly laying down on the same tile
  b. has 'blocking' status effect from resisting (press `z`)
3. Pick one of the matching mobs

The hero receives a more direct personal effect from the explosive, but the world impact is limited in some way. Each explosive has a different effect depending on what it does. Note that for grenades not nullified, the remainder of the detonation will still occur directly under the hero.

##### old_grenade/spawner/wasp
Hero: Gibbed (by mass wasp implantation)
Grenade: Nullified (but some will explode out of you)

##### old_grenade/graviton
Hero: Takes massive (suicide-level) chest crunch damage
Grenade: Half range

##### old_grenade/stinger & old_grenade/stinger/frag
Hero: Half of the projectiles (10) hit you directly
Grenade: Quarter of the projectiles (5) are still flung out

##### old_grenade/high_explosive
Hero: Take direct explosion
Grenade: Half power

##### old_grenade/sonic
Hero: Take direct ear damage
Grenade: Half power

##### old_grenade/emp
Hero: All organs are EMPed
Grenade: Half range

##### gimmickbomb/owlclothes & gimmickbomb/hotdog
Hero: Dressed up in *completely* unremovable clothes
Grenade: Nullified

##### pipebomb/bomb

Strength
Hero: Throw chest organs out. Over 9 power guarantee gibs. Still at center of reduced explosion.
Grenade: reduce power by one quarter, max reduction of three. This means after 12 initial power, the max reduction is 3. 
Power reduction calculator: https://www.desmos.com/calculator/hqdqnv5zcy

Various pipe-bomb customizations e.g. power cells are also applied to the hero before lowering their overall effectiveness.

## Why's this needed? 
* rule of cool
* existing discussion: https://forum.ss13.co/showthread.php?tid=21772

## TODOs
- [x] find a better way to convey 'diving' onto an explosive
- [x] implement on more explosives because it *really* sucks to try and fail
- [x] feedback
- [x] tinker with balancing / tuning

## Changelog

```changelog
(u)glowbold
(*)Play hero by laying down and blocking on top of live hand grenades and pipebombs.
```


Sbmhawk:
I would say not gibbing outright, but ruining organs in the chest area, and a lot of damage.

(At the very least, for stinger nades)

Sbmhawk:
Also, I haven't looked at the code, but I would make sure that throwing a nade on a stunned person (or vice versa) does NOT trigger this

pgmzeta:
> Also, I haven't looked at the code, but I would make sure that throwing a nade on a stunned person (or vice versa) does NOT trigger this

yep, considered this - you have to be pulling the active grenade and be lying down on the same tile.

Absolute-Baller:
I'm sorta just spitballing and imnotcoder so this might be completely off-base or unfeasible, but I imagine a more robust system would be a general explosion-power reduction feature. It could be something like: when laying down on top of a tile where an explosion is occurring, you eat up to [x] explosion power, and it gets applied to you in the form of brute damage or whatever. Some damage still always occurs, but just enough to show that an explosion happened. Then, if the power of the explosion exceeds [x], you eat [x] explosion power, gib, and the rest of the explosion power gets applied to the explosion.

Glamurio:
Didn't check all of the code, but how does this interact with gimmick grenades? I don't want people gibbed for banana or owlbombs. Perhaps in the case of owlbombs they can turn into an actual owl critter (concentrated effect).

pgmzeta:
> Didn't check all of the code, but how does this interact with gimmick grenades? I don't want people gibbed for banana or owlbombs. Perhaps in the case of owlbombs they can turn into an actual owl critter (concentrated effect).

owlbomb/hotdogbomb only force-clothe the 'sacrifice', and my plan for the future is to make the clothing completely irremovable if you dive on those two grenades specifically.

the other gimmick grenades do not call this and have no effect.

TDHooligan:
could also have slide (resist + throw) be a shorthand for this that doesnt require the aiming of grab + move + lie.

pgmzeta:
> for an action that's not likely to be done except explicitly to suicide.
in retrospect making these custom suicides feels like an obvious solution

pgmzeta:
Rather major change in approach; please see the updated PR description.

code/mob/living/carbon/human/procs/damage.dm
```diff
@@ -641,3 +641,40 @@
 /mob/living/carbon/human/UpdateDamage()
 	..()
 	src.hud?.update_health_indicator()
+
+///The deliberate act of using one's body to cover a live time-fused hand grenade.
+/mob/living/carbon/human/proc/cover_explosion_with_body(power, brisance=1)
```
Tarmunora, line 646: a decent amount of this proc appears to be copy-paste from `ex_act()` - any reason not to just call that to handle non-organ stuff?

code/datums/controllers/explosion_controls.dm
```diff
@@ -216,11 +217,22 @@ var/datum/explosion_controller/explosions
...
+			if(istype(source, /obj/item))
+				var/list/covering_mobs = list()
+				for (var/mob/living/carbon/human/H in epicenter)
+					if(H.lying && H.hasStatus("blocking")) //trying to cover it
+						covering_mobs.Add(H)
+				var/mob/living/carbon/human/hero = PICK_OR_NULL(covering_mobs)
+				if(hero) //there goes my hero
+					hero.visible_message("<span class='alert'>[hero] wraps [his_or_her(hero)] body tightly around [src]!!</span>")
+					hero.cover_explosion_with_body(power, brisance)
+					power *= ITEM_EXPLOSION_COVERED_REDUCTION_MULT
```
Tarmunora, line 232: would make more sense for `cover_explosion_with_body()` to return a modified `power` imo, and/or have a maximum amount that power can be lowered by - e.g.: a maxcap TTV should probably be relatively unaffected

Tarmunora (changes requested):
Interesting idea but I'm feeling like it (further) complicates explosion code while also not affecting the majority of actual grenades (stinger/frag grenade projectiles come to mind).

pgmzeta:
> Interesting idea but I'm feeling like it (further) complicates explosion code while also not affecting the majority of actual grenades (stinger/frag grenade projectiles come to mind).

Thank you for the feedback - I'll work on integrating this and the previous eat-shrapnel implementation. Agreed on adding explosion complexity, but I am worried that it will be frustrating to dive on an explosive and have nothing come of it just depending on the bomb type. Don't mind cutting it if others don't think that's a valid concern.

Tarmunora:
> ... I am worried that it will be frustrating to dive on an explosive and have nothing come of it just depending on the bomb type...

I think having a system that explicitly covers only pipebombs and some grenades would cover most people's expectations fairly well, for what it's worth

Tarmunora (changes requested):
oops i forgot about this PR

code/obj/item/grenades.dm
```diff
@@ -1897,37 +1975,44 @@ ADMIN_INTERACT_PROCS(/obj/item/pipebomb/bomb, proc/arm)
...
-				for (var/turf/T in oview(1+ round(src.reagents.total_volume * 0.12),src.loc) )
+				if (istype(hero))
+					src.reagents.trans_to_direct(hero, src.reagents.total_volume / 2)
+				for (var/turf/T in oview(1+ round(src.reagents.total_volume * 0.12), src.loc))
 					src.reagents.reaction(T,1,5)
 
+			if (istype(hero))
+				if(src.strength >= 9)
+					hero.gib()
+				src.strength = max((src.strength * 0.75), (src.strength - 3))
```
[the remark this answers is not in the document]
pgmzeta: Switched to applying the explosion act using its power, but doing that will result in them getting double-exploded once the bomb itself detonates. Could convert this to crush damage with a suitable strength multiplier instead, or another route if you think there's a better way.

Tarmunora (approved):

PR #15857, opened 2023-09-16
Use dir defines

TheTyrant25:
[INTERNAL]
## About the PR
Replaces several hundred instances of dir being set to `1, 2, 4, 5, 6, 8, 9, 10` instead of `NORTH, SOUTH, EAST, NORTHEAST, SOUTHEAST, WEST, NORTHWEST, SOUTHWEST`. Ignores dir = 0.

Doesn't fully use a find replace, because some things that are called "dir" or end with "dir" don't use the system and instead use something else. Some things may be missed.

## Why's this needed?
[CODE QUALITY]
No need to confuse people with `dir = 9`. If it's a direction, we should use the defines.


code/modules/networks/computer3/mainframe2/misc_terms.dm
```diff
@@ -2749,7 +2749,6 @@ TYPEINFO(/obj/machinery/networked/printer)
 	var/obj/beam/next
 	var/limit = 48
 
-	dir = 2
 	layer = NOLIGHT_EFFECTS_LAYER_BASE
```
Tarmunora, line 2752: ?

Tarmunora: Also the other instances of this change in the file

code/turf/floors_carpet.dm
```diff
@@ -26,1046 +26,1042 @@
...
-		dir = 1
-	south
-		icon_state = "red4"
-		dir = 2
-	east
-		icon_state = "red4"
-		dir = 4
-	west
-		icon_state = "red4"
-		dir = 8
```
Tarmunora, line 124: icon_state change appears to have been lost

Tarmunora (changes requested):

code/modules/networks/computer3/mainframe2/misc_terms.dm
TheTyrant25: ah right
so the pr started out as something being done by hand
i ditched the dir = 2 in multiple objects because the default dir for `obj` is south anyway and it was an unneeded line.
I can put it back if you like but it works fine without.

Tarmunora (approved):

PR #15858, opened 2023-09-16
Adds Glaucus Fishing Rod, 6tc traitor item

MylieDaniels:
[A-Game-Objects] [C-Feature] [C-Sprites] [E-Add-to-Wiki]
## About the PR 
This adds the Glaucus Fishing Rod, a 6 tc item for Rancher and Angler traitors.
image

Traps can be set up with the lure, which can be disguised (kinda like a chameleon bomb) by smacking the Glaucus with any item. Click to throw the lure out, click to reel in. Use in-hand to reel in one tile, for the classic dollar bill trick or to adjust your trap. If someone picks it up, they'll suffer a 4 second stun, 15 damage, and get embedded with the hook. If they are hit with the hook while it's being thrown out or reeled in, they'll also be embedded, but will suffer 5 damage and no stun.

While hooked, you are quite slowed. The hook can be removed through surgery (this is more of a technicality, but if you have a morphine auto-injector and a knife, go wild), by getting 9 tiles away from the rod, or by the wielder using the rod in hand.

Reeling in while someone is hooked will start a looping 0.7 second action bar- on each completion, the hooked mob is pulled 1 tile closer, and if they are in melee range they take 7 damage. If they aren't in melee range, 3 damage is added to the "pending" damage- this is applied to the next successful melee reel in the same set of reels, and if a melee reel does 25 or more damage, it stuns for SQRT(damage) / 3 seconds. After a successful melee reel or if the line is broken/lure removed, damage is reset to 7 again.

Clicking while a mob is on the line will throw them about 3-4 tiles on a 6 second cooldown. Best used as an opener, so place lures next to glass tables, vending machines, or open airlocks!

https://github.com/goonstation/goonstation/assets/25803648/44fd25b8-234f-44b7-ba9d-a0384e454e5d

TODO: Completed
- [x] Unique inhand sprites - now bulky red fishing rod
- [x] Non-beartrap sprite for default lure - now uses minnow
- [x] Fix casting from inside lockers - now only allows reeling, casting, and actionbar while on turf

## Why's this needed? 
I like fishing, I like traitor weapons, and I think you should too. Also, rancher has 1 traitor item and angler has 0.

## Changelog 

```changelog
(u)MylieDaniels, AveryQuill
(*)Adds the Glaucus Fishing Rod, a 6 tc tacticool fishing rod for catching crew with disguised hooks. Available to Rancher and Angler traitors, as well as through surplus crates.
```


Glamurio:
What a sick item! I love it! One small caveat I have, and realize might be more problematic to implement, but for visual clarity, would it be possible to draw a line from the rod user to the target?

Maybe also some interaction about people running into the line, like, tripping or something. Would make it even more amusing. Just ideas, ultimately I wouldn't be upset if the item got in the way it is. Excited for it!

444Portal:
> What a sick item! I love it! One small caveat I have, and realize might be more problematic to implement, but for visual clarity, would it be possible to draw a line from the rod user to the target?
> 
> Maybe also some interaction about people running into the line, like, tripping or something. Would make it even more amusing. Just ideas, ultimately I wouldn't be upset if the item got in the way it is. Excited for it!

Would probably counteract some of the utility of being able to scan in an item as bait, and hiding in like a bush or something. Just my personal opinion.

Vispectral:
Very Interesting Item I'd say, shame it's locked behind Rancher.  Unique, flavorful and fits a niche theme. 
The design seems alright. I'm sure Gannets would be a massive fan of this item. 

My downside would be cost. I do not think personally this is on par with 7 TC items. Compared to Cyalume Saber which does 60 and reflects, fits in your pockets.
Wrestlebelt which is probably the ultimate 1v1 item and 
Katana which can totally be used to wipe out stations ..

This by comparison is rather tame. I'd advocate for 6 or 5 TC perhaps allowing it to used in conjunction with other items to promote interesting gameplay. Unless if as some people suggested here, you could "scan" items to use as bait and even then, I still feel 7 is expensive. 

Excellent work Mylie.

MylieDaniels:
Ah, I don't know if I was clear in my explanation, you can in fact scan items to act as bait- this disguises the bait as the item.
However I do see that it's definitely not as strong as a c-saber or katana and doesn't have anywhere near the power of a wrestlebelt or red chainsaw. I will lower the price to 6 tc, mostly because I worry about the combo potential with 7 tc items if it were 5- though if a developer has opinions on it being higher or lower, I'd love to hear from them.

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+		..()
+		src.lure = new (src)
+		src.lure.rod = src
+
+	get_desc(dist)
+		..()
+		if (dist < 1) // on our tile or our person
+			if (.) // we're returning something
+				. += " " // add a space
+			if (!src.lure)
```
TobleroneSwordfish, line 398: seems odd to create the lure in `get_desc`, usually you don't want side effects here

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+				else
+					step_towards(src.lure, src)
+			else
+				user.visible_message("<span class='alert'><b>[user] yanks the lure out of [src.lure.owner]!</b></span>")
+				src.lure.set_loc(get_turf(src.lure.loc))
+				src.lure.on_remove(src.lure.owner)
+
+	// here so that afterattack is called at range, which is the least bad way to throw the lure i think maybe
+	pixelaction(atom/target, params, mob/user, reach)
+		..()
```
TobleroneSwordfish, line 448: How does a default implementation of `pixelaction` cause this? If it's the lack of a return value then you should be explicitly returning `null` or `FALSE` or w/e

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+
+	afterattack(atom/target, mob/user)
+		..()
+		if (!isturf(user.loc))
+			return
+		if (!src.lure)
+			src.lure = new (src)
+			src.lure.rod = src
+			tooltip_rebuild = 1
```
TobleroneSwordfish, line 457: other versions of this snippet don't set `tooltip_rebuild`, maybe consider making this a proc so it isn't inconsistent?

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+			src.icon_state = "syndie_fishing_rod-active"
+			src.item_state = "syndie_fishing_rod-active"
+		//state for not fishing
+		else
+			src.icon_state = "syndie_fishing_rod-inactive"
+			src.item_state = "syndie_fishing_rod-inactive"
+
+	disposing()
+		. = ..()
+		qdel(src.lure)
```
TobleroneSwordfish, line 502: swap these round, always clean up before calling parent in disposing

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+	New(var/user, var/target, var/rod, var/lure)
+		..()
+		src.user = user
+		src.target = target
+		src.rod = rod
+		src.lure = lure
+
+	onStart()
+		..()
+		if (!src.user || !src.target || !src.rod || !src.lure || (src.target == src.user) || !(src.lure.owner == src.target) || !(src.user.equipped() == src.rod) || !isturf(src.user.loc) || GET_DIST(src.user, src.target) > src.rod.line_length)
```
TobleroneSwordfish, line 650: please make this a proc or a local macro or something, that's a long if chain to be duplicating three times

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+			src.lure.clean_forensic()
+		else
+			if(!src.lure.owner)
+				if(BOUNDS_DIST(src.lure,src) == 0)
+					src.is_fishing = FALSE
+					src.UpdateIcon()
+					user.update_inhands()
+					src.lure.set_loc(src)
+				else
+					step_towards(src.lure, src)
```
TobleroneSwordfish, line 440: This should check that the lure is on a turf, currently it runtimes if you try to reel it in while it's in a disposal pipe etc.

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+		if (!(W == src.rod) && isliving(user) && !user.nodamage && GET_DIST(src, src.rod) < src.rod.line_length)
+			user.changeStatus("weakened", 5 SECONDS)
+			user.TakeDamage(user.hand == LEFT_HAND ? "l_arm": "r_arm", 15, 0, 0, DAMAGE_STAB)
+			user.force_laydown_standup()
+			implanted(user, null)
+		else
+			return ..()
+
+	attack_hand(mob/user)
+		if (isliving(user) && !user.nodamage && GET_DIST(src, src.rod) < src.rod.line_length)
```
TobleroneSwordfish, line 552: Kind of weird that you can safely pick up the lure when it's out of range. Wouldn't it make more sense if the lure was returned when the rod moved too far away from it? See outermost movable signals.

TobleroneSwordfish (changes requested):
Horrific, I love it.

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+	New()
+		..()
+		src.lure = new (src)
+		src.lure.rod = src
+
+	get_desc(dist)
+		..()
+		if (dist < 1) // on our tile or our person
+			if (.) // we're returning something
+				. += " " // add a space
```
Tarmunora, line 397: unreachable code

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+
+	attackby(obj/item/I, mob/user)
+		if (!src.lure)
+			src.lure = new (src)
+			src.lure.rod = src
+		if (src.lure.loc == src)
+			boutput(user, "You scan \the [I.name] onto \the [src.name]'s holographic bait projector.")
+			src.lure.name = I.name
+			src.lure.real_name = I.name
+			src.lure.desc = I.desc
```
Tarmunora, line 411: name and desc are getting set by `src.lure.appearance = I` further down

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+	HELP_MESSAGE_OVERRIDE({"The Glaucus starts with 7 damage on a melee reel, but builds up 3 additional damage on each ranged reel. If this reaches <b>25 damage</b>, or 6 ranged reels before a melee reel, the target will be stunned when damaged."})
+
+	New()
+		..()
+		src.lure = new (src)
+		src.lure.rod = src
+
+	get_desc(dist)
+		..()
+		if (dist < 1) // on our tile or our person
```
Tarmunora, line 395: Need to add `REBUILD_DIST` to tooltip flags

code/modules/fishing/fishing_gear.dm
```diff
@@ -350,3 +350,341 @@ TYPEINFO(/obj/item/fish_portal)
...
+				if (target.bioHolder && target.bioHolder.Uid && target.bioHolder.bloodType)
+					gibs(target.loc, blood_DNA=target.bioHolder.Uid, blood_type=target.bioHolder.bloodType, headbits=FALSE, source=target)
+				else
+					gibs(target.loc, headbits=FALSE, source=target)
+			return TRUE
+		else
+			step_towards(target, user)
+			return FALSE
+
+/obj/item/implant/syndie_lure
```
Tarmunora, line 529: So. What happens when you load this into an implant gun.

code/modules/fishing/fishing_gear.dm
Tarmunora: Or refactor to use `canRunCheck()`

Tarmunora (changes requested):

MylieDaniels:
Thank you for the reviews, I'll get to work fixing the issues in the morning.

code/modules/fishing/fishing_gear.dm
MylieDaniels: Thank you, I wasn't sure how to handle that.

code/modules/fishing/fishing_gear.dm
MylieDaniels: Oh, of course.  Thank you, I'll fix that!

MylieDaniels:
> It'd be neat if it could act as a high tier fishing rod too.
I'm unsure how to best implement this, as clicking on a fishing spot while fighting would be frustrating I think.

code/modules/fishing/fishing_gear.dm
MylieDaniels: Yes, it was the lack of a return value, I now return null explicitly.

code/modules/fishing/fishing_gear.dm
MylieDaniels: That makes sense, I've changed it to instead only add the bonus section of the tooltip if a lure exists.

code/modules/fishing/fishing_gear.dm
MylieDaniels: I've changed the lure to a generic item and stopped piggybacking on implant code.  If it's desired that a knife or something can remove the lure, I'll find a way to do that!

444Portal:
> > It'd be neat if it could act as a high tier fishing rod too.
> 
> I'm unsure how to best implement this, as clicking on a fishing spot while fighting would be frustrating I think.

How would using it on help intent sound?

MylieDaniels:
> > > It'd be neat if it could act as a high tier fishing rod too.
> > 
> > 
> > I'm unsure how to best implement this, as clicking on a fishing spot while fighting would be frustrating I think.
> 
> How would using it on help intent sound?

I don't know about others, but I'm the type to use weaponry on help intent at times...

PR #15864, opened 2023-09-16
Handcuffs: small usage nerf

kbsmilk:
[GAME OBJECTS] [BALANCE]
## About the PR 
The handcuff action bar now **requires one of** these conditions, both on start and on end, or else it fails:
* Target is incapacitated (stunned, knockdown, unconscious, or pinned).
* Target is on help intent, and has turned their back to the handcuffer.

Don't worry, this doesn't change much for Security - the baton+cuff combo works just the same.

## Why's this needed? 
Handcuffing doesn't have an interrupt flag for being attacked. So the only way to escape cuffing is getting distance - which isn't possible sometimes, e.g. very crowded or cramped areas. 

This PR stops such scenarios, because it feels really bad to get cuffed with no recourse. It also fights off getting cuffed while typing.

(I didn't simply add the flag for other reasons: e.g. someone automending you while cuffing would've canceled it)

## Changelog 

```changelog
(u)Smilg
(*)Handcuffing now requires the target to be incapacitated (pinned, stunned, asleep, etc) or cooperating (facing away from you on Help intent).
```

### DIFF
--- a/code/datums/actions/actions.dm
+++ b/code/datums/actions/actions.dm
@@ -727,6 +727,27 @@
 
 		logTheThing(LOG_COMBAT, owner, "attempts to handcuff [constructTarget(target,"combat")] with [cuffs] at [log_loc(owner)].")
 
+		if (!is_incapacitated(target)) // same check in both onStart and onEnd.
+			if (target.a_intent == INTENT_HELP)
+				if (owner.dir != target.dir)
+					owner.tri_message(target,
+						"<span class='alert'><B>[owner] attempts to handcuff [target],</B> \
+							but [he_or_she(owner)] can't reach [target]'s back.</span>",
+						"<span class='alert'>[target]'s back isn't turned! Cuff them from behind.</span>",
+						"<span class='alert'><B>[owner] can't cuff you from this angle!</B> Turn your back to cooperate.</span>"
+					)
+					interrupt(INTERRUPT_ALWAYS)
+					return
+			else
+				owner.tri_message(target,
+					"<span class='alert'><B>[owner] attempts to handcuff [target]. \
+						[he_or_she(target)] dodge[target.get_pronouns().pluralize ? "s" : ""] away!</B></span>",
+					"<span class='alert'>[target] slips from your grasp! Pin or incapacitate them first.</span>",
+					"<span class='alert'><B>You evade [owner]'s handcuff attempt!</B> Use help intent and turn your back to cooperate.</span>"
+				)
+				interrupt(INTERRUPT_ALWAYS)
+				return
+
 		duration *= cuffs.apply_multiplier
 
 		if(ishuman(owner))
@@ -742,6 +763,27 @@
 		var/mob/ownerMob = owner
 		if(!istype(ownerMob) || !target || !cuffs || target.hasStatus("handcuffed") || cuffs != ownerMob.equipped() || BOUNDS_DIST(owner, target) != 0)
 			return
+
+		if (!is_incapacitated(target))
+			if (target.a_intent == INTENT_HELP)
+				if (owner.dir != target.dir)
+					owner.tri_message(target,
+						"<span class='alert'><B>[owner] attempts to handcuff [target],</B> \
+							but [he_or_she(owner)] can't reach [target]'s back.</span>",
+						"<span class='alert'>[target]'s back isn't turned! Cuff them from behind.</span>",
+						"<span class='alert'><B>[owner] can't cuff you from this angle!</B> Turn your back to cooperate.</span>"
+					)
+					interrupt(INTERRUPT_ALWAYS)
+					return
+			else
+				owner.tri_message(target,
+					"<span class='alert'><B>[owner] attempts to handcuff [target]. \
+						[he_or_she(target)] dodge[target.get_pronouns().pluralize ? "s" : ""] away!</B></span>",
+					"<span class='alert'>[target] slips from your grasp! Pin or incapacitate them first.</span>",
+					"<span class='alert'><B>You evade [owner]'s handcuff attempt!</B> Use help intent and turn your back to cooperate.</span>"
+				)
+				interrupt(INTERRUPT_ALWAYS)
+				return
 
 		if (initial(cuffs.amount) > 1)
 			if (cuffs.amount < 1)


Piesuu:
This is going to remove all viability for security to peacefully cuff someone without having to baton/flash them beforehand, not good for RP.

colossusqw:
I like this, it seems like it will mainly make type baiting harder, but not affect much else, cool.

frawhst:
I like the sound of this honestly. Let's give it a run

frawhst:
Un-testmerging and re-testmerging to update the changelog to be more descriptive

DisturbHerb:
Having their back turned is a weird requirement considering that the worn handcuff icons display them in front of a person. Intent based resisting is neat though.

TobleroneSwordfish:
Not a fan of this, feels like it just further encourages batoning people first and asking questions never, plus direction based stuff tends to feel really bad and janky in this game, see fibre wire.

kbsmilk:
> Not a fan of this, feels like it just further encourages batoning people first and asking questions never, plus direction based stuff tends to feel really bad and janky in this game, see fibre wire.

How so? The way I see it, the only way a standing, capable person is getting cuffed, is by force or by cooperation. Or, they get cuffed the bad, unfun way: a momentary distraction IRL, abusive trickery in game (like type baiting), or other such nonsense.

The dir thing is deliberately janky too. It takes two people working together to overcome. Easy if they cooperate (turning yourself in). Otherwise, it's another protection in place for people on help intent, who don't actually want to be cuffed.

444Portal:
Interesting. I'll need to see this in place first before knowing how I feel about it past that, though. Guess I'll be playing some Security tomorrow.

DisturbHerb:
I would prefer to see this go through without the direction requirement.

Tarmunora:
Ah yes, the two things that always result in fun and intuitive gameplay: intent-based and direction-based action.
If the issue is that handcuffing doesn't get interrupted when attacked, why not just... do that instead.

TheColorCyan:
I dont like this
Not only direction thing is very janky and weird (fibre wire) the fact that you have to stun a person to cuff ruins a lot of interactions as well
Vampires will be more of a pain to deal with
If a person you gotta arrest gives up and lets themselves get cuffed the fact that you NEED to stun them is just a weird interaction in general.
The intent thing is just bad
But its a good way to encourage newbies to just beat shit out of people because cuffs require that many steps 👍
I agree with Tarm, if the issue is it not being interrupted by attacks why not just make it do that

kbsmilk:
> I dont like this Not only direction thing is very janky and weird (fibre wire) the fact that you have to stun a person to cuff ruins a lot of interactions as well Vampires will be more of a pain to deal with If a person you gotta arrest gives up and lets themselves get cuffed the fact that you NEED to stun them is just a weird interaction in general. The intent thing is just bad But its a good way to encourage newbies to just beat shit out of people because cuffs require that many steps 👍 I agree with Tarm, if the issue is it not being interrupted by attacks why not just make it do that

I believe you missed the part where cuffing requires either of those conditions, not both.

The INTERRUPT_ATTACKED flag has some unwanted side effects. It's a reasonable alternative to this, but not as good (imo), and thus wasn't chosen.
Intent-based passive cuff resistance gives more protection than needing to reactively punch your assailant. Which, if they're security, you have 2 seconds to deal with. Not very long, with this game's controls + latency.

TheColorCyan:
What side effects? If its some stopping the action that's exactly what is supposed to happen
If the issue is getting stuck in a crowded place i feel like it will be much better to shove an officer than have the jankiest ever dir and intent based stuff
Also i have not seen a single time someone fail to dodge getting handcuffed without being stunned: even when typing its easily interrupted 
Im up for handcuffing being stopped on attack but not a weird dir intent stun system

Flaborized:
This change doesn't make much sense to me at all. Intent is not something you generally set to affect things that others do to you, it near-universally is used to augment how you interact with hand clicks, and pretty much every example of intent being used in other ways (other than the passive pass-through-ness effect) is unintuitive and unfitting of what the system is meant to do. I also don't much like the direction idea, tends to be janky and fiddly to work with in my experience.

If the goal is to make it so you cannot be cuffed while not stunned without you giving explicit consent, why are these two measures even necessary? You could do something like have a yes/no prompt pop up or slow down cuff times but only when they're not stunned, so people could always react. You could also interrupt the cuffing if the victim moves or even only if the victim attacks the cuffer (and nobody else). I can't see how these are the two most ideal criteria for this. 

Though that said I don't even really think this is a real issue being solved;  baiting people into typing so you can cuff them as sec specifically (where you have the shortest timer therefore the least time to react) is already less applicable than just stunning them when they're typing anyways, and is difficult to actually pull off if you don't have fast cuffing speed. I think cornering someone so you can cuff them is interesting and difficult enough to pull off in a real situation that it's alright if it happens. And yeah this makes arresting vampires incredibly difficult in situations where you could likely kill the vampire anyways, and it should be easier to restrain them for sec than it should be to kill them dead!

kbsmilk:
Lotsa feedback saying intent and dir-based mechanics shouldn't go in. I have 3 alternatives in mind.

### Option A: Block cuffs entirely unless target lays down/is incapacitated
* Good visual equivalence - if someone is on the ground for any reason, you can cuff them.
* Still requires force or willing cooperation to cuff someone.
* Passive protection from cuffs, if you're standing and healthy.

### Option B: Interrupt if cuffer gets attacked by anyone
* This is the simple implementation, with the INTERRUPT_ATTACKED flag being added to cuffs, and no other change.
* Anyone can guaranteed interrupt the cuffer, with any attack, such as a punch, shove, thrown object, medical patch, hypospray, bullet, etc.
* This'd be cool for teams fighting together - stop your buddy getting cuffed by throwing at, or shooting, the cuffer.
* It's also a larger nerf to handcuffs than I'd first intended.

### Option C: Both of the above.
* Nothing to add here.

Glamurio:
> Lotsa feedback saying intent and dir-based mechanics shouldn't go in. I have 3 alternatives in mind.
> 
> ### Option A: Block cuffs entirely unless target lays down/is incapacitated
> * Good visual equivalence - if someone is on the ground for any reason, you can cuff them.
> * Still requires force or willing cooperation to cuff someone.
> * Passive protection from cuffs, if you're standing and healthy.
> 
> ### Option B: Interrupt if cuffer gets attacked by anyone
> * This is the simple implementation, with the INTERRUPT_ATTACKED flag being added to cuffs, and no other change.
> * Anyone can guaranteed interrupt the cuffer, with any attack, such as a punch, shove, thrown object, medical patch, hypospray, bullet, etc.
> * This'd be cool for teams fighting together - stop your buddy getting cuffed by throwing at, or shooting, the cuffer.
> * It's also a larger nerf to handcuffs than I'd first intended.
> 
> ### Option C: Both of the above.
> * Nothing to add here.

### Option D: Pop-up prompt to accept being cuffed.
* Very intuitive and hard to miss.
* Literally cannot have any weird behavior quirks.
* The cuffed always has the choice in the matter.
* Stunned people simply skip the prompt and get cuffed instead.

This is what Flaborized suggested and it's honestly a really good suggestion. It's extremely straight-forward, does not allow for any jank and still lets people with a baton or stun cuff as per usual.

kbsmilk:
> Option D: Pop-up prompt to accept being cuffed.

I considered this, but as an HTML link in the chat, similar to a pda message reply. I'm iffy on full-on popups that the user doesn't initiate (and thus expect).

DisturbHerb:
Popups are generally less responsive in normal gameplay and are more intrusive in my opinion. The attack interruption flag is, I believe, the best option.

Valtsu0:
I believe the best option would be having resisting/blocking cancel cuffing

TobleroneSwordfish:
I'm going to close this as it doesn't seem workable currently, feel free to re-open with a different approach

PR #15871, opened 2023-09-17
Add new traitor item "Reflective Glasses"

FlameArrow57:
[GAME OBJECTS][GAME MODES][FEATURE]

## About the PR 
This PR adds a new 1 TC traitor item, Reflective Glasses.

Reflective glasses act just like sunglasses, initially appearing as prescription glasses. They can have their appearance set once by using them in-hand, and they redirect flashes and flash cell assembly burning effects back at the flasher. Doesn't count area flashes, and it doesn't recurse if the attacker has reflective glasses too.

They can also be bought by Revheads. If someone tries to flash a Revhead wearing them, and they don't have eye protection themselves, they will be converted to a Revolutionary.

## Why's this needed? 
This gives a defensive traitor option against a common stun option the crew has, while also being offensive and having some gimmicky-ness. Helps traitors and Revheads survive some more and gives them some resistance against being arrested.

While you can find sunglasses around the station, or use welding masks, this would be an available option that isn't hindering like welding masks and provides more utility than either.

Not sure if this is too similar to anything currently available for antagonists, defensive options like this aren't a traitor thing, or if this isn't a good solution for stun protection

## Changelog 
```changelog
(u)FlameArrow57
(*)Added a new 1 TC traitor item - Reflective Glasses. They reflect incoming flashes at attackers, and their appearance can be set once by using them in-hand. Also available for Revheads, where reflected flashes will convert attackers to Revolutionaries if they aren't wearing eye protection.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -1011,7 +1011,7 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 
 /datum/syndicate_buylist/traitor/sunglasses
 	name = "Reflective Glasses"
-	desc = "A randomly selected pair of glasses, with built-in sun and flash protection, covered by their 100%, one-way reflective shine!"
+	desc = "A customizable pair of glasses, with built-in sun and flash protection, covered by their 100%, one-way reflective shine!"
 	item = /obj/item/clothing/glasses/sunglasses/reflective
 	cost = 1
 	can_buy = UPLINK_TRAITOR | UPLINK_SPY_THIEF | UPLINK_NUKE_OP | UPLINK_HEAD_REV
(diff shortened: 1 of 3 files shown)


Wisemonster:
I do like this, although it taking on a random appearance can make it rather difficult to use, considering if it rolls the appearance of glasses your job doesn't have access, it'll be a give away, especially if everyone is aware it's revs. This is especially bad with the thermals appearance, considering only sec/det has access to it, and people usually will not be running around with thermals on during revs.

444Portal:
An inhand interaction to change between a few generic glasses types could fix that. I think it's fine given how often people wear gas masks though personally

frawhst:
Agree with the above, random glasses appearance could be really convenient or really inconvenient depending on your job. Would either make it so you can set the appearance once, or make it so you can reset it an unlimited amount of times and bump the cost to 2TC.

PR #15873, opened 2023-09-17
Move various things out of lavamoon.dm

TheTyrant25:
[INTERNAL] [CODE QUALITY]
## About the PR
Moves ladders, golden gannets delivery takeout, yo yos and dead shrubs out of lavamoon.dm and into appropriate files. Adds a new file, ladder.dm, since that code is really massive.

## Why's this needed?
I was looking at the takeout code and was like wtf why is it in there. It's not even in io. It's only orderable by cargo. What the.
(edit: it is in fact in iomoon, i had dmm file searching off)
So i made a pull request.


ZeWaka:
Takeout is on the lava moon though...

TheTyrant25:
oh whoops

PR #15900, opened 2023-09-19
Resprites EOD Gear

MitchTheUser:
[Clothing]

## About the PR 

Resprites both the Bomb Defusal Suit and the Blast Helmet found in the armory. The outfit has been changed to match it's worn sprites, which now causes the items sprites to be of higher quality than the current sprites. The inhand sprites for both the suit and the helmet have also been completely remade to match the new sprites.

EODDisplay 

## Why's this needed? 

The current sprites are old, and they are drastically different than their worn sprites. Consistency between the item sprites and worn sprites are good, and the new sprites can now benefit from having more detail and will fit in visually with other Security gear.


Glamurio (approved):

Valtsu0:
IIRC the blast helmet shares a sprite with another helmet (i think donk co helmet, maybe elite helmet?). I can't test at the moment if it is affected, but you should make sure it is how it is supposed to br

MitchTheUser:
> IIRC the blast helmet shares a sprite with another helmet (i think donk co helmet, maybe elite helmet?). I can't test at the moment if it is affected, but you should make sure it is how it is supposed to be

You're right, the Donk space helmet shares the same sprite as the Blast Helmet, so it was affected by this PR. However, when testing it, I found that the Donk helmet and Blast helmet share the same worn sprite, which is what I based the new helmet sprite off of, so it actually kind of works in its favor, as the Donk space helmet now matches its worn sprite, though unintentionally.

frawhst:
!merge_upstream

PR #15903, opened 2023-09-20
fix emergency shuttle takeoff sounds

pgmzeta:
## About the PR 
Replace area-based sound with landmarks, and updates/adds landmark to all in-rotation maps.

## Why's this needed? 
Better cues for emergency shuttle takeoff

### DIFF
--- a/code/modules/transport/shuttle/shuttle_controller.dm
+++ b/code/modules/transport/shuttle/shuttle_controller.dm
@@ -201,16 +201,19 @@ datum/shuttle_controller
 						announcement_done = 1
 
 					else if (announcement_done < 2 && timeleft < 30)
-						if (istype(src.sound_turf)) playsound(src.sound_turf, 'sound/effects/ship_charge.ogg', 100)
+						if (istype(src.sound_turf))
+							playsound(src.sound_turf, 'sound/effects/ship_charge.ogg', 100)
 						announcement_done = 2
 
 					else if (announcement_done < 3 && timeleft < 4)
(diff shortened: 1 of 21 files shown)


CheffieGithub:
Maybe it would be better if it just picked the closest door / engine instead. Just so you don't need a landmark on every map

pgmzeta:
> Maybe it would be better if it just picked the closest door / engine instead. Just so you don't need a landmark on every map

Some of the shuttles use real engines, and some use decals - and I imagine admins will end up editing shuttles in-situ as well that might throw a wrench in that. Is there a different way you had in mind?

Tarmunora (approved):

code/modules/transport/shuttle/shuttle_controller.dm
```diff
@@ -194,17 +194,17 @@ datum/shuttle_controller
...
-						var/area/sound_location = locate(/area/shuttle_sound_spawn)
+						var/turf/sound_location = pick_landmark(LANDMARK_SHUTTLE_SOUND)
 						playsound(sound_location, 'sound/effects/ship_engage.ogg', 100)
 						announcement_done = 3
 
 					else if (announcement_done < 4 && timeleft < 1)
-						var/area/sound_location = locate(/area/shuttle_sound_spawn)
+						var/turf/sound_location = pick_landmark(LANDMARK_SHUTTLE_SOUND)
 						playsound(sound_location, 'sound/effects/explosion_new4.ogg', 75)
 						playsound(sound_location, 'sound/effects/flameswoosh.ogg', 75)
```
ZeWaka, line 209: should have a fallback and log if there's no landmark, also repeated code

ZeWaka (changes requested):

PR #15908, opened 2023-09-20
Introduction of lungrot as a complication of salbutamol usage in areas with miasma

M-Earthfire:
[CHEMISTRY][FEATURE]

## About the PR 

This PR adds a new chemical: lungrot
lungrot bloom is a toxic fungus created by the following recipe:

(1) miasma + (0.1) salbutamol -> (1) lungrot bloom

This recipe has a 0.5 reaction time, no reaction sound or message and can only take place in carbon-based mobs, e.g. humans or monkeys. This reaction cannot take place if robustissin is in the person. Also, it will only create lungrot bloom until 15u have accumulated in a person.

lungrot bloom, when hitting the treshold of 7.5u (should be around 15u miasma) will make the person contract lungrot. Also, 1u of lungrot bloom decays into 1u miasma.

Lungrot is a disease with 5 stages. It has the following effect:

- make the person suffer 0.2/0.4/0.5/0.5/0.5 tox damage each cycle
- make the person generate 0.8/1/1.2/1.5 miasma in their bloodstream each cycle
- adds a 0/5/8/12/17% chance each cycle to cough out some chems in your bloodstream and additional miasma. This effect has a 15 second cooldown. This generates a cloud of  0/3-8/8-12/12-18/18-28u of miasma. Also, it transfers chemicals out of your bloodstream into the cloud equal to 35% of the amount of miasma generated.

The cure for this disease is robustissin. The cure chance is a measily 1% if any salbutamol is in the person left (any amount of robustissin turns the disease immediatly remissive). Once all salbutamol is out of the person, the chance jumps out to 20%, making it an effective cure. Also, while remissive, the disease is treated as 2 stages lower than it actually is.

## Why's this needed? 

This recipe adds a complication to the medical system. If medbay does not take care of decomposing corpses in medbay, treating people with salbutamol can accidently create lungrot bloom in people.

The inspiration for this is the risk factor of salbutamol inhalers to promote mouth trush. Since this is a silly spess game, this risk factor for a mundane issue is jacked up to 11.

In general, though, the only effect this has is making the problem with miasma in people worse. This gives doctors plenty of time to treat the disease. Also, mixing salbutamol with robustissin supresses the reaction in general, which gives witty doctors an additional angle to show their skill.

Since in general salbutamol is rarely used in medbay, this is not something that is encountered too often in medbay. I assume this is much more encountered when people try to uses buffmixes with salbutamol.

For antags, this creates an interesting mechanic, especially for doctors. Since salbutamol is on the medbay whitelist, antag doctors can inject people in miasma-loaded areas with salbutamol and, due to the stealthy reaction, can effectively poison them and further spread miasma.

On the other hand, antags can mix pure lungrot bloom with toxins to  cause people to cough out the chemicals and spread parts of them to other victims. Pure lungrot is hard to isolate, though (e.g. it smokes the area while you try to isolate it out of monkeys)

To summarize, this chem adds a further incentive to take care of corpses in medbay while increasing the skill ceiling of treatment. Also, this features a problem for buffmix abusers and enables antags to create interesting effects by smoking chemicals in people. 

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Miasma-exposure to people with salbutamol in them causes the creation of lungrot bloom. This toxic fungus causes lungrot, causing people to cough out more miasma!
```


TheTyrant25:
pathology returns...
Just kidding
I like the coughing miasma/lungrot idea.

Kyle2143 (changes requested):
> Completely honest, i was thinking about turning this into a disease chemical. But personally i found that to be far too punishing for how hard diseases are to treat/mitigate.

This really should be a disease/malady instead of a chemical. It doesn't need to be a pathology disease, you can make them easy to treat. The system is a bit more obtuse than reagent, but it really doesn't make sense to make a "disease" into a reagent just because it should be simple to cure. There are simple to cure diseases like the Cold or Shock, you can look to them for ideas if that helps.

TobleroneSwordfish:
I agree with Kyle, it seems like a waste for the cure to just be pumping purgatives in.

code/modules/medical/diseases/lungrot.dm
```diff
@@ -0,0 +1,79 @@
+/datum/ailment/disease/lungrot
+	name = "Lungrot"
+	max_stages = 5
+	stage_prob = 4
+	spread = "Non-Contagious"
+	cure = "Robustissin application after ablation of salbutamol."
```
TobleroneSwordfish, line 6: Not sure `ablation` is the right word here? It's a bit confusing anyway, maybe just put `removal` or something.

code/modules/chemistry/Reagents-Diseases.dm
```diff
@@ -90,6 +90,20 @@ datum
 			transparency = 235
 			disease = /datum/ailment/disease/cold
 
+
+		disease/lungrot
+			name = "lungrot bloom"
+			id = "lungrot_bloom"
+			description = "highly toxic fungal colonies created in the enviroment of a weakened lung."
+			reagent_state = SOLID
+			minimum_to_infect = 2.5
```
TobleroneSwordfish, line 99: Considering the likelyhood of walking into a random miasma cloud while using salbutamol, either this number should be much higher or the amount produced should be much lower.

code/modules/medical/diseases/lungrot.dm
```diff
@@ -0,0 +1,79 @@
...
+	associated_reagent = "lungrot_bloom"
+	reagentcure = list("cold_medicine")
+	//for as long as salbutamol is in the patient, robustissin is extremly ineffective
+	recureprob = 1
+	affected_species = list("Human")
+
+
+
+/datum/ailment/disease/lungrot/stage_act(var/mob/living/affected_mob, var/datum/ailment_data/affecting_ailment, mult)
```
TobleroneSwordfish, line 15: As written this will still do all its effects while remissive (ie cured)

TobleroneSwordfish (changes requested):

code/modules/medical/diseases/lungrot.dm
TheTyrant25: Ablation in english generally refers to gradual destruction by chipping away at the edges of thing, which is sort of correct? But I think it's meant for solid objects and not chemicals.
The word is notably used in heat shields for spacecraft, they're usually made of ablative material and disintegrate to get rid of heat more effectively.

TobleroneSwordfish (approved):
Cautious yes, may need further balancing.

PR #15910, opened 2023-09-20
Add tacticool NT id to eligible jobs

Glamurio:
[Clothing]

## About the PR 
Adds a new, more appropriately colored ID to the special NT jobs, namely Nanotrasen Special Operative, Nanotrasen Emergency Repair Technician, Nanotrasen Emergency Medic and Nanotrasen Security Consultant.

image

This only adds it to the jobs, the identification console does not have access to this color scheme.

## Why's this needed? 
Since these are uncommon jobs and have a generally "tacticool blue" outfit going on, it makes sense to have them be unique. 

Forum suggestion thread:
https://forum.ss13.co/showthread.php?tid=21804

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)NT Specialists now have their own unique ID.
```


TobleroneSwordfish (approved):

Valtsu0:
While this isn't exactly what the linked thread suggest, i like it for the most part. I still think NTSC and inspector should have the existing NT ID. I also object naming this id NT (maybe tactical)

code/obj/item/card.dm
```diff
@@ -163,6 +163,9 @@ TYPEINFO(/obj/item/card/emag)
 		access = get_access("Captain")
 		src.AddComponent(/datum/component/log_item_pickup, "Captain")
 
+/obj/item/card/id/nt
+	icon_state = "id_nt"
+
```
DimWhat, line 168:
```suggestion
/obj/item/card/id/nt
	icon_state = "id_nt"
	keep_icon = TRUE
	
```
Would be a good idea to prevent it from changing from the ID computer. A unique description would be cool as well.

DisturbHerb:
Not a particularly big fan of differentiating whitelisted roles from the regular crew any more than they already are, including the visual appearance of their IDs.

Glamurio:
!merge_upstream

TobleroneSwordfish:
Okay sorry this has been sitting around for ages, I've gone ahead and just changed these jobs to use the existing centcom ID card sprite instead since it may as well be consistent.

PR #15911, opened 2023-09-20
refactor light fixture placing

zjdtmkhzt:
## About the PR 
Refactors light fixture placement code a bit to move the stuff out of the walls and floors and into the light fixture object.
## Why's this needed? 
* The initial impetus for this was to make it possible to simply place light fixtures onto windows.
* Light fixtures should now also be offset correctly depending on the wall sprite, like those that are defined in maps already are.
* Removes a lot of sort of duplicate code?
* Ideally it will also prevent the thing where you smack walls with the fixtures while placing them.
## Changelog 

```changelog
(u) zjdtmkhzt
(+) Light fixtures can now be placed directly on windows.
```


zjdtmkhzt:
I still have some minor merge conflicts to resolve and want to test it a bit before then, after that it should be good to merge hopefully.

Also, the RCD will still smack things with this, but I'd rather change that in a subsequent PR since this is already sizable and I might have to shuffle some RCD code around.

ZeWaka:
finally, nice

code/modules/power/lighting.dm
```diff
@@ -52,6 +55,70 @@ TYPEINFO(/obj/item/light_parts)
...
+			return FALSE
+		dir = get_dir(get_turf(target), user)
+		if (!is_cardinal(dir))
+			boutput(user, "You can't seem to reach that part of \the [target]. Try standing right up against it.")
+			return FALSE
+	if (locate(/obj/machinery/light) in get_turf(target))
+		boutput(user, "There's already a lamp there!")
+		return FALSE
+	return TRUE
+s
```
ZeWaka, line 78: s

zjdtmkhzt: s

ZeWaka: pls no s

zjdtmkhzt:
Also added functionality now for stacking several lights on one tile, if it makes sense with the types and directions.
image

code/modules/power/lighting.dm
zjdtmkhzt: ok

flappybatpal (approved):
Tested ingame and seems good.

PR #15912, opened 2023-09-20
Use IMMUNE_TRENCH_WARP for atoms that shouldn't fall to z5 via seafloor hole

pgmzeta:
## About the PR 

adds two new exceptions to the list of things that get warped to the trench on a surface hole: forcefields and lasers

maybe a "floating" object property makes sense at this point idk?

## Why's this needed? 

* oshan 'meteors' can sink forcefield segments :|||||
* it's a little silly that energy manifestations have physical gravity 

## TODO
- [x] Double-check laser interactions
- [x] make it a property

### DIFF
--- a/_std/defines/atom.dm
+++ b/_std/defines/atom.dm
@@ -63,6 +63,8 @@
 /// overrides the click drag mousedrop pickup QOL kinda stuff
 #define NO_MOUSEDROP_QOL			(1 << 10)
 #define MOVE_NOCLIP 				(1 << 11)
+/// Atom won't get warped to z5 via floor holes on underwater maps
+#define IMMUNE_TRENCH_WARP			(1 << 12)
 
 
 //THROW flags (what kind of throw, we can have ddifferent kinds of throws ok)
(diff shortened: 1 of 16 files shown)


imawxome:
Im pretty sure neutrons can fall into trench holes as well, i remember seeing "the neutron falls into the deep hole!"

Glamurio:
> maybe a "floating" object property makes sense at this point idk?

An object property would make a lot of sense, given the list of exceptions, however I don't think "floating" is a consistent name. Perhaps something more along the lines of "can_fall"? That covers most of these and sounds less odd than saying reinforced cables float.

frawhst (approved):
Yeah we really need a better system for generally untouchable/intangible things. This is much cleaner than the current gross check.

amylizzle (approved):
Oh that's nice, maybe chuck it on `/datum/projectile` as well?

pgmzeta:
> Oh that's nice, maybe chuck it on `/datum/projectile` as well?

didn't find a read the flag off the datum when set directly. added a blanket exception instead, which is probably fine? Not sure if any slow-moving ones would want to sink

pgmzeta:
> Oh that's nice, maybe chuck it on `/datum/projectile` as well?
had to do it on /obj/projectile for the effect

PR #15914, opened 2023-09-21
Fix fried foods permanently being placed inside mob

Glamurio:
[Bug] [Catering] [Balance]

## About the PR 
Fried foods no longer do `set_loc()` on the mob, but instead just get digested normally like any other item would.

## Why's this needed? 
Currently, when any item is fried and eaten, it will place itself inside a mob via `set_loc`. There are no known methods (to me) to remove this item again, meaning that you end up with strange results like people eating telecrystals permanently teleporting with no cure, or endlessly radiating erebite that cannot be removed even with ipecac.


Tarmunora:
what about changing it from using `stomach_contents` to `stomach_process`? `The former appears mostly deprecated and seems intended to only contain uh... mobs.

Glamurio:
> stomach_process

Sort of irrelevant now that #15919 got merged which overhauls the entire system, so I'm just going to use `C.organHolder.stomach.consume` instead.

PR #15916, opened 2023-09-21
Fix fuel tanks saying "Nothing in it." when there is in fact something in it

TealSeer:
[BUG] [GAME OBJECTS]
## About the PR 

Fixes a bug wherein fuel tanks would say nothing in it on examine because they only had the flag to make reagents viewable with spectroscopic goggles, and none like RC_VISIBLE, RC_SCALE, RC_FULLNESS to give basic information on contents.

## Why's this needed? 

I dumped 800u of welding fuel on the floor of engineering last round because I thought they were empty and now I'm embarrassed


frawhst:
really need to add a lint for same-type var overrides. grumble

PR #13538, opened 2023-04-01
Implements a limit to how much can be eaten at once.

Wisemonster:
[BALANCE] [CATERING] [INPUT WANTED] [STATUS EFFECT] [EXPERIMENTAL]

## About the PR 
This pr adds in a system for tracking how much mobs have eaten. in this pr, if a mob takes 11 bites (about the equivalent of 2 hamburgers) of food, they'll become "full", meaning eating any more while full will result in damage to the stomach, eventually killing the stomach. This damage is only applied to the stomach if they aren't a mob critter, so the mobs themselves won't take any direct damage from over-eating, except mob critters, which will take direct damage due to not having a stomach.

They are 2 status effects tied in with this system, a "recently eaten" status effect, and a "over-eaten" status effect. Upon the status effects expiring, the mob's bite count is reset to 0, allowing them to eat more without taking stomach damage. I've made the over-eaten status effect take longer to expire to disincentive over-eating even more. If this is deemed more appropriate for rp, I can change some things to restrict this to rp.

## Why's this needed? 
Currently, it is possible for a staffie to come to the kitchen and gobble up all of the chef's food in a minute. This will prevent the chef's hard work from being devoured by a single person, and also give more use to the stomach cyber organ, which currently does very little. This also allows for ways to alter the mobs stomach capacity in future prs, such as some antags having higher/lower capacities, or organ types changing capacity

## Changelog 

```changelog
(u)Wisemonster
(*)There is now a limit on how much a mob can eat at a time, check minor changes for details
(+)You can only eat around 11 bites of food before the next bite will make you "over-eaten", if you eat more while in this state, you will damage your stomach.
(+)Waiting out the "recently eaten" and "over-eaten" status effects will reset you eat count.
```

PR #15919, opened 2023-09-21
Gluttony rework

TobleroneSwordfish:
[BALANCE] [CATERING] [MEDICAL]
## About the PR 
This PR is a spiritual successor to #13538, based on some of the feedback it got.
Stomachs now have a capacity (default 7, cyberstomach 12) and can fill up if you eat too much. Each food item contributes an amount determined by its `fill_amt` var which defaults to 1 but varies between ~0.5-4 depending on the item.
Also adds a status effect to indicate being full.
image
Undigestible items consumed by either deepfrying or matter eater will stay in the stomach, contribute to the fullness amount and can be cut out by using a scalpel on the extracted stomach.
Stomach transplants now work as you'd expect, transferring the contents properly.

TODO:

- [x] Deal with edible deepfried items not being digested maybe?

## Why's this needed? 
Same reasoning as #13538, one determined staffie should not be able to demolish 40 minutes of cooking on their own, and a lot of stomach code was old and weird and bad and now it's better.
Also since #15914 was made while I was writing this, provides an alternative solution that keeps things able to have effects after being eaten, but with a way to remove them.

The numbers are pretty arbitrary and based on "this seems about right" so feel free to suggest changes.

## Changelog 

```changelog
(u)LeahTheTech
(*)Stomachs now have a capacity and can fill up if you eat too much, cyberstomachs have more. No longer can a single determined staffie consume the entire kitchen.
(+)Undigestible items now remain in the stomach after being eaten by frying or matter eater and can be retrieved by cutting them out of the extracted stomach with a scalpel.
```


amylizzle:
*Consequences* for eating deep fried rocks? What is this, 1984?

code/mob/living.dm
```diff
@@ -2260,3 +2254,11 @@
 		else
 			src.say(message)
 		src.stat = old_stat // back to being dead 😌
+
```
amylizzle, line 2257: commented?

code/modules/chemistry/tools/food_and_drink.dm
```diff
@@ -439,19 +450,17 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks)
...
-					L.stomach_process += B
-
+			if(ethereal_eater)//ghost critters can get a little ingest reaction and a tiny amount of reagent, but won't actually take reagents
+				src.reagents.reaction(eater, INGEST, 3)
+				if(!ON_COOLDOWN(src, "critter_reagent_copy_\ref[eater]", 15 SECONDS))
+					src.reagents.copy_to(eater.reagents, 3/max(src.reagents.total_volume, 3)) //copy up to 3u total, once per food per 15 seconds
+			else
+				var/obj/item/reagent_containers/food/snacks/bite/B = new /obj/item/reagent_containers/food/snacks/bite
+				B.fill_amt = src.fill_amt/initial(src.bites_left) //so all the bites add up to the full item fillness
+				B.reagents.maximum_volume = reagents.total_volume/(src.bites_left+1 || 1) //MBC : I copied this from the Eat proc. It doesn't really handle the reagent transfer evenly??
```
amylizzle, line 460: ambiguous? 
```suggestion
				B.reagents.maximum_volume = reagents.total_volume/((src.bites_left+1) || 1) //MBC : I copied this from the Eat proc. It doesn't really handle the reagent transfer evenly??
```

code/obj/item/organs/stomach.dm
```diff
@@ -7,42 +7,86 @@
...
+		if (istype(AM, /obj/item/reagent_containers/food))
+			var/obj/item/reagent_containers/food/food = AM
+			return food.fill_amt
+		else
+			return 1 //other stuff can clog your stomach
+
+	proc/consume(atom/movable/AM)
+		AM.set_loc(src.donor)
+		src.stomach_contents |= AM
+		src.food_amount += src.food_value(AM)
```
amylizzle, line 55: probably not necessary, but you could maybe do a length check before & after to see if `food_amount` needs incrementing. This would only be useful if it was in some way possible to eat things that are already in your stomach, but gooncode does like to surprise.

code/obj/item/organs/stomach.dm
```diff
@@ -7,42 +7,86 @@
...
+		if (src.is_full())
+			src.donor.setStatus("full")
+
+	proc/eject(atom/movable/AM)
+		AM.set_loc(src.donor.loc)
+		src.stomach_contents -= AM
+		src.food_amount -= src.food_value(AM)
```
amylizzle, line 62: ditto

code/obj/item/organs/stomach.dm
```diff
@@ -67,6 +111,53 @@
...
+				logTheThing(LOG_COMBAT, user, "gibs [constructTarget(src.donor,"combat")] breaking out of their stomach at [log_loc(src.donor)].")
+				src.donor.gib()
+
+	proc/handle_digestion(mult = 1)
+		if (!length(src.stomach_contents))
+			return
+		src.digest_food(mult)
+		src.digest_mobs(mult)
+
+#define DIGESTION_PER_LIFE_TICK 3 //Total amount of reagents we can digest each Life tick
```
amylizzle, line 135: seems like this should probably just be a var on the stomach, then you can have cyberstomachs process faster or whatever

amylizzle (commented):
vorestation 2024

Glamurio:
This is a really cool addition and the implementation seems neat. My primary reason for making #15914 though, was that we had an antag that had eaten telecrystals and was nigh impossible to catch. I'm not sure that, even with these additions, that it'll be possible to put them down long enough to cut out the telecrystals.

code/modules/chemistry/tools/food_and_drink.dm
TobleroneSwordfish: sus

Glamurio:
> I'm pretty sure you could do the same thing with chest implanted items or telecrystal item arms so shrug. Personally I recommend hitting them until they stop moving.

Totally, but the time it takes to eat a telecrystal vs. do surgery on someone is different. Nevertheless, that's a unique issue and not related to this PR.

PR #15920, opened 2023-09-21
Goonhub API: deleting/updating player notes

TheTyrant25:
[INTERNAL] [TOOLING]
## About the PR
Adds api support for the remaining two PlayerNotes Endpoints: Update and Delete. Also adds a new apiBody for updating, since that apparently doesn't need round_id and doesn't use the same one as Add.

This is my first Goonhub API PR, please lemme know if i did anything tremendously terribly wrong.

## Why's this needed?
Wire asked really nicely


code/modules/goonhub/api/requests/players/notes/UpdatePlayerNotesBody.dm
```diff
@@ -0,0 +1,35 @@
+
+/datum/apiBody/players/notes/update
+	var/game_admin_ckey	= "string"
+	var/server_id			= "string"
```
ZeWaka, line 4:
```suggestion
	var/game_admin_ckey		= "string"
	var/server_id			= "string"
```

code/modules/goonhub/api/routes/players/notes/GetPlayerNotes.dm
```diff
@@ -1,8 +1,7 @@
 
 /// GET /players/notes
-/// Get list
+/// List paginated and filtered player notes
 /datum/apiRoute/players/notes/get
 	method = RUSTG_HTTP_METHOD_GET
 	path = "/players/notes"
 	queryParams = list("filters", "sort_by", "descending", "per_page") // string, string, string, string
-	correct_response = "string"
```
ZeWaka, line 8: bug wire this shouldn't be null on the swagger

ZeWaka (changes requested):

PR #15925, opened 2023-09-21
Gun holster buff

DimWhat:
[game objects] [balance]
## About the PR 
Increases the amount of slots all gun belts have to 6 and allows more than just the gun and ammo to fit in there. This should make nukies want to use them more often.

## Why's this needed? 
When you play nuke ops you tend to see these belts all over the floor as they are just kinda worse than a fanny pack or a utility belt thanks to their limited selection of things it can hold as well as having a low amount of slots as well. This should make them more of a sidegrade as they can hold a pistol but also hold one less item than a normal belt.

## Changelog 

```changelog
(u)DimWhat
(+)Made nukie pistol holsters hold more stuff, and more types of stuff.
```


Gannets (approved):
Would be nice to see these used more

PR #15929, opened 2023-09-21
Adds more TTV tank sprites.

DimWhat:
[game objects] [sprites]
## About the PR 
Adds a a pair TTV versions of mini oxygen tanks and extended oxygen tanks, renames a few sprites to actuallly work with the current pocket oxygen tank, as well as monkey versions of the empty tank by TealSeer. Also fixes a couple name bugs for chimps that made certain worn air tanks not show up.
image

## Why's this needed? 
It let's cursed TTV set ups be visible to the world also buggy bugs bug me.


TealSeer (approved):

MeggalBozale (changes requested):
Backpack sprites are null when only one tank is added. Works fine when two tanks are on though. I LOVE THIS, just please fix the one-tank backpack TTVs.

DimWhat:
> Backpack sprites are null when only one tank is added. Works fine when two tanks are on though. I LOVE THIS, just please fix the one-tank backpack TTVs.
> 
> EDIT: Also if you feel like it, you could resprite the inhand... :O

One-tank TTVs going null seemed to not actually be caused by this PR but a preexisting behaviour.

MeggalBozale:
Should probably still fix that? Either way that's on my todo if you don't end up doing it (that's perfectly fine) which can be made as a separate PR. I think the sprites look great.

MeggalBozale (approved):

MeggalBozale:
Ok the issue is just that it doesn't update the wear image when removing a tank so

MeggalBozale:
Alright a PR is opened to fix the issue with TTV overlays

TobleroneSwordfish (changes requested):
The back sprites are really indistinct, which is very bad for something like a worn TTV which should stand out like a sore thumb: 
image
Regular TTV for reference: 
image
I think if you want to make the back sprites accurately reflect the tanks then the base sprite needs to be more identifiable, ideally with the big red valve in the middle as shown on the item sprite:
image

DimWhat:
> The back sprites are really indistinct, which is very bad for something like a worn TTV which should stand out like a sore thumb: image Regular TTV for reference: image I think if you want to make the back sprites accurately reflect the tanks then the base sprite needs to be more identifiable, ideally with the big red valve in the middle as shown on the item sprite: image

Yeah the base TTV worn sprite is not ideal. There does seem to be a valve but its at the top of the little black box and also hidden by your hair. (even less ideal) I can try to make the tanks a bit larger, I just took the belt tank's worn sprite and just plopped it onto the TTV when I initially added all those sprites a while back.

TobleroneSwordfish:
Okay for now I've gone ahead and just made the smaller tanks incompatible with TTVs.

MeggalBozale:
Unfortunate. Would this be undone if the resprite is completed?

TobleroneSwordfish:
Maybe if someone can come up with a way of making them easily readable while on your back 👍

PR #15933, opened 2023-09-21
Used pressure crystals: credit value scaling with explosion power

kbsmilk:
[GAME OBJECTS] [SCIENCE] [FEATURE]
## About the PR 
* Pressure crystals, an item used to measure explosion power, can be sold via cargo crate. Their value is based on last measured explosion power, with this (subject to change) formula: credits = explosion_power ^ 1.1 * 100
* A bounty system for pressure crystals. Try to match the requested measurement! Get closer, get richer.
* A PDA program is added to Science PDAs. It tells the value of crystals previously sold, and also lists the bonus bounties.
* **Credit value diminishing returns. A sale impacts the market value of similar crystals. An identical crystal to a previous sale bottoms out at 0 credit value; however, less similar crystals smoothly scale to the normal value.**
* **You can't stack multiple crystals on an epicenter to duplicate results.**
* **You can stack simultaneous explosions to gain more power, but diminishingly. Each successive explosion is compared to the current measurement, and the lowest of the pair is square rooted, then summed.**
* Crystal resupply crates added to the QM console. You can already get more crystals with vendomat resupply cartridges, but a special crate is +convenience.
* +Some code readability and grammar and kiloblast

## Why's this needed? 
Fun and purpose for toxin lab. They can contribute to the budget, but they have to go do field work and expend lab equipment. Also pressure crystals report real explosion power now, to compare results and do Science™
Can also deliberately test detonate artifact bombs, for reliably more credit value than selling it.

### Developers please tweak and change all the balance values you want.
There are some defines at the top of shippingmarket.dm for balance tweaks.

Here's some reference credit values with the formula: explosion_power ^ 1.1 * 100 = credits
* 5 (welding fuel pipe bomb): 587 credits
* 12.5 (black powder pipe bomb assembly): 1609 credits
* 87.89 (low end artifact bomb): 13,750 credits
* 192.516 (350u black powder in a barrel): 32,577 credits
* 194.45 (maxcap ttv estimate): 32,937 credits
* 337.64 (low end devastating artifact bomb): 60,436 credits
* 7000 (mega canister bomb estimate): 1,696,710 credits

## Changelog 

```changelog
(u)Smilg
(*)Pressure crystals from Vendomats, when used, can now be sold for credits via Cargo. Work in Toxins today!
(+)New program for Scientist PDAs that shows you previously sold pressure crystals, and their bounties.
```


CheffieGithub:
Damn a real use for toxins, maybe we need a proper explosives testing platform now

kbsmilk:
> Damn a real use for toxins, maybe we need a proper explosives testing platform now. Is there anything stopping someone stacking 10 of these on a tile for one blast?

Yes. Stacking doesn't work (i hope). You CAN harvest the same explosion with multiple crystals by doing this: line them up one per turf, in a line leading away from the epicenter. But by nature, each further crystal receives much less explosive power. (fixing this is beyond my ability, but it's also not that bad)

code/obj/item/device/transfer_valve.dm
```diff
@@ -529,9 +551,10 @@ TYPEINFO(/obj/item/device/transfer_valve/briefcase)
 			boutput( user, "<b>There's no crystal in this here device!</b>")
 		else
 			if(crystal.pressure)
-				boutput( user, "The reader reads <b>[crystal.pressure/25]</b> kilojoules." )
+				boutput( user, "The reader reads <b>[crystal.pressure]</b> kiloblast." )
 			else
-				boutput( user, "The reader reads a firm 0. It guilts you into trying to read an unexploded pressure crystal, and seems to have succeeded. You feel ashamed for being so compelled by a device that has nothing more than a slot and a number display.")
+				boutput( user, "The reader reads a firm 0. It guilts you into trying to read an unexploded pressure crystal, and seems to have \
+					succeeded. You feel ashamed for being so compelled by a device that has nothing more than a slot and a number display.")
```
CheffieGithub, line 557: I know these didn't have them before but they probably should have style classes.

kbsmilk: planning to atomize the bulk of the pressure sensor stuff into another pr

code/obj/item/device/transfer_valve.dm
```diff
@@ -509,17 +509,39 @@ TYPEINFO(/obj/item/device/transfer_valve/briefcase)
...
-	var/total_pressure = 0
-	desc = "A pressure crystal. We're not really sure how it works, but it does. Place this near where the epicenter of a bomb would be, then detonate the bomb. Afterwards, place the crystal in a tester to determine the strength."
-	name = "Pressure Crystal"
+	w_class = W_CLASS_SMALL
+	var/pressure = 0 // used to calculate credit value, in shippingmarket.dm proc/appraise_value
+	var/last_explode_time = 0
+	var/static/explosion_id = 0
+	desc = "A pressure crystal. We're not really sure how it works, but it does. Place this near where the epicenter of a bomb would be, \
+		then detonate the bomb. Afterwards, place the crystal in a tester to determine the strength.<br>\
+		Spent pressure crystals can be sold to researchers on the shipping market, for a credit sum depending on the measured power."
```
Azrun, line 518: This seems... really verbose.  Should some of this be moved over to a help message?

code/obj/item/device/transfer_valve.dm
```diff
@@ -529,9 +551,10 @@ TYPEINFO(/obj/item/device/transfer_valve/briefcase)
 			boutput( user, "<b>There's no crystal in this here device!</b>")
 		else
 			if(crystal.pressure)
-				boutput( user, "The reader reads <b>[crystal.pressure/25]</b> kilojoules." )
+				boutput( user, "The reader reads <b>[crystal.pressure]</b> kiloblast." )
```
Azrun, line 554: What is a kiloblast?

kbsmilk: A nonexistent unit (just like joules in _units.dm) :)
Anyway I looked into explosion power unit -> Joule conversion, but it is not worth it. So kiloblast is just 1:1 with an explosion's power value instead

code/obj/item/device/transfer_valve.dm
kbsmilk: Ahh forgot about those, good point. It is done

TobleroneSwordfish:
This seems really neat, but the problem especially with TTVs is that if you can make one then you can make 20. A good solution to this might be giving diminishing returns for values close to previously shipped crystals, ie on selling a crystal, go through each previously shipped crystal and scale down the new profit value by the inverse of the distance between them.
Also in general I think the values are a little high, but that would be less of an issue with diminishing returns.

TobleroneSwordfish:
Alternative idea from dev chat: make it like requisitions where you get an order for a particular value of pressure crystal and are paid based on how close you get to it (and maybe also how high the value is?)

kbsmilk:
> This seems really neat, but the problem especially with TTVs is that if you can make one then you can make 20. A good solution to this might be giving diminishing returns for values close to previously shipped crystals, ie on selling a crystal, go through each previously shipped crystal and scale down the new profit value by the inverse of the distance between them. Also in general I think the values are a little high, but that would be less of an issue with diminishing returns.

Only after making this PR did I learn that the valve is mechscannable. I'd been guessing the time economy on manufacturing, setting up the test area, etc. would be comparable to what Cargo can do in the same time. Then Emily's cargo update went in today, lol

I agree with diminishing returns on similar values. It's just gonna take me a while to puzzle through the math. Maybe I'll try something simpler like pregenerating a list of ranges w/ scaling multipliers, and fitting sold crystals into those

kbsmilk:
Scope creep has arrived on scene. Please check out the newly updated PR description

code/modules/economy/shippingmarket.dm
```diff
@@ -1,6 +1,13 @@
 #define SUPPLY_OPEN_TIME 1 SECOND //Time it takes to open supply door in seconds.
 #define SUPPLY_CLOSE_TIME 15 SECONDS //Time it takes to close supply door in seconds.
 
+#define PRESSURE_CRYSTAL_VALUATION_FORMULA ** 1.1 * 400 /// The full explosion-power-to-credits conversion formula
+#define PRESSURE_CRYSTAL_SALES_RANGE_LENGTH 5 /// The span size of each range for the pressure crystal sales list. Measured in explosion power
+#define PC_VALUE_DIMINISH_MULTIPLIER (rand(30, 60) / 100) // Every sale will multiply the current multiplier for its range by this value
+#define PC_HIGH_VALUE_RANGE_QUANTITY 7 // How many ranges will be given credit multipliers in New()?
```
TobleroneSwordfish, line 7: this is way too many, make it more like 1 or 2

code/modules/economy/shippingmarket.dm
```diff
@@ -1,6 +1,13 @@
 #define SUPPLY_OPEN_TIME 1 SECOND //Time it takes to open supply door in seconds.
 #define SUPPLY_CLOSE_TIME 15 SECONDS //Time it takes to close supply door in seconds.
 
+#define PRESSURE_CRYSTAL_VALUATION_FORMULA ** 1.1 * 400 /// The full explosion-power-to-credits conversion formula
+#define PRESSURE_CRYSTAL_SALES_RANGE_LENGTH 5 /// The span size of each range for the pressure crystal sales list. Measured in explosion power
```
TobleroneSwordfish, line 5: 5 power either side is very hard to get right, it'd make more sense to scale it by proximity rather than have a sharp cutoff

code/modules/economy/shippingmarket.dm
```diff
@@ -392,6 +414,20 @@
 
 		return duckets
 
+	proc/appraise_pressure_crystal(var/obj/item/pressure_crystal/pc, var/sell = 0)
+		if (pc.pressure <= 0)
+			return
+		var/index = "[ceil(pc.pressure / PRESSURE_CRYSTAL_SALES_RANGE_LENGTH)]"
```
TobleroneSwordfish, line 420: Not really sure what this index system is supposed to be doing, but it's it doesn't seem like it's going to prevent slightly diluting the mix a few times to get different enough values to trick the system. I suggest making it gradually scale by proximity instead, so a lot of similar values will all dilute any more sales in the approximate area.

TobleroneSwordfish (changes requested):

code/modules/economy/shippingmarket.dm
kbsmilk: The index takes a measured pressure and converts it into an integer, which represents a range. e.g. pressure of 27.4 gets fit into the range of 25-30. Then if it gets sold, the value of the 25-30 range gets an entry in the sales list as a key, with its value as a multiplier (rand 0.3 to 0.6)
After that, credit value for pressures in that range receives that multiplier. Which stacks multiplicatively with further sales.

It's not perfect: has a hard cutoff, like you say. But I don't have the knowledge to make the system you're describing.

PR #15938, opened 2023-09-21
APIV2 split parameters into queryParams and routeParams

TobleroneSwordfish:
## About the PR 
Splits the api route parameters into query and route params and adds handling for them in the route generation.

## Why's this needed? 
Route params do exist

### DIFF
--- a/code/modules/goonhub/api/routes/apiRoute.dm
+++ b/code/modules/goonhub/api/routes/apiRoute.dm
@@ -8,8 +8,10 @@
 	var/method = null
 	/// Actual path of the api query, for example `/rounds`
 	var/path = null
-	/// Parameters for the call
-	var/list/parameters = null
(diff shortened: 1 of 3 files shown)


code/modules/goonhub/api/routes/apiRoute.dm
```diff
@@ -8,8 +8,10 @@
 	var/method = null
 	/// Actual path of the api query, for example `/rounds`
 	var/path = null
-	/// Parameters for the call
-	var/list/parameters = null
+	/// Route parameters for the call, ie /rounds/{round_id}/
```
ZeWaka, line 11:
```suggestion
	/// Route parameters for the call, ie /rounds/{round_id}/ - Must be in order
```

ZeWaka (approved):

PR #15949, opened 2023-09-22
ERROR: don't directly set plane to a number

ZeWaka:
## About the PR 

grep lint for direct plane setting

## Why's this needed? 

94f1c07e112417d0338301ed9fc5f1402138200d

### DIFF
--- a/_std/__std.dme
+++ b/_std/__std.dme
@@ -14,6 +14,7 @@
 #include "defines\lag.dm"			// the all important LAGCHECK()
 #include "types.dm"					// stuff like childrentypesof()
 #include "datum.dm"					// hasvar()
+#include "plane.dm"                 // debug icons
 #include "defines\abilities.dm"
 #include "defines\access.dm"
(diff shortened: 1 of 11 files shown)


pali6:
This by itself sorta creates a false sense of security as it still allows the more common antipattern of `plane = PLANE_FOOBAR - 1`.

ZeWaka:
(resolved)

_std/__std.dme
```diff
@@ -14,6 +14,8 @@
 #include "defines\lag.dm"			// the all important LAGCHECK()
 #include "types.dm"					// stuff like childrentypesof()
 #include "datum.dm"					// hasvar()
+#include "macros\spawn.dm"			// SPAWN
+#include "plane.dm"                 // debug icons
```
ZeWaka, line 18:
```suggestion
#include "macros\spawn.dm"			// SPAWN
#include "plane.dm"					// debug icons
```

Issue #8252, opened 2022-04-17
Dummies' Guide to Material Science has outdated material values


zjdtmkhzt:
### Labels

[BUG]

### Description

Dummies' Guide to Material Science has outdated material values

### Steps to reproduce

1. Read Guide
2. It's wrong

### Expected Behavior

the guide to be correct

### Additional Information & Screenshots

For instance, the density of viscerite is now 45 when it used to be 65.
Some other values that were changed in #7443 are presumably not updated either, and maybe things that were changed earlier as well.

In an ideal world this would probably be automatically generated with the actual material stats?

zjdtmkhzt:
Adding "add to wiki" because logically the mirror of the book on the wiki is outdated too. 
(https://wiki.ss13.co/Dummies%27_Guide_to_Material_Science)

pgmzeta:
In-game fixed in #10109 but it still needs a wiki update

Chatauscours:
Fixed by #10109

PR #15957, opened 2023-09-23
Dummies' Guide to Material Science, 8th Ed.

MeggalBozale:
## About the PR 
dreamseeker_GGiJTyn3UV
Update's the "Dummies' Guide to Material Science, 7th Ed." book with minor tweaks to updated material values, and resolves some minor spelling mistakes.

Feel free to make suggestions to rewording and whatnot.

## Why's this needed?
In-Game reference material should be accurate and up to date. Also spelt correctly.
#8252 should be fixed by this but will still need a wiki update (It was already, but needed another once-over)

## Changelog
```changelog
(u)MeggalBozale
(+)NT has spent a few million credits stocking its stations with newly printed guides to material science, featuring minor spelling corrections and updated material values.
```

### DIFF
--- a/code/obj/item/book.dm
+++ b/code/obj/item/book.dm
@@ -107,7 +107,7 @@ Custom Books
 	file_path = "strings/books/matsci_guide_old.txt"
 
 /obj/item/paper/book/from_file/matsci_guide
-	name = "Dummies' Guide to Material Science, 7th Ed."
+	name = "Dummies' Guide to Material Science, 8th Ed."
 	desc = "An explanation of how to work materials and their properties. Nanotrasen missed buying a few editions between the old one and this..."
 	icon_state = "matscibook"
 	file_path = "strings/books/matsci_guide_new.txt"
--- a/strings/books/matsci_guide_new.txt
+++ b/strings/books/matsci_guide_new.txt
@@ -27,7 +27,7 @@ table td.unavaib {
 <body>
 <h3>Foreword:</h3>
 <hr>
-This text serves as a comprehensive overview to Materials Science. It is detailed but by no means exhaustive of all potential methods and projects. At the time of writing, all information in this book is correct. However, over time, the information may become out of date. Please see the footnote for additional reading.
+This text serves as a comprehensive overview of Material Science. It is detailed but by no means exhaustive of all potential methods and projects. At the time of writing, all information in this book is correct. However, over time, the information may become out of date. Please see the footnote for additional reading.
 <hr>
 <h2>
   Contents</h2>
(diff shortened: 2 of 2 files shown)


MeggalBozale:
Oh hey, just to mention: I wanted to make the values dynamic, or maybe even have it go material by material. But I have NO idea how to make a book dynamic that loads itself from a text file. If anyone does know, I'd be happy to have this book keep itself up to date once and for all.
(Though, I think it would be nice to have hand-typed descriptions for each ore, and if new mats are added, "tell a coder!" pops up for no desc.)

Glamurio:
> Oh hey, just to mention: I wanted to make the values dynamic, or maybe even have it go material by material. But I have NO idea how to make a book dynamic that loads itself from a text file. If anyone does know, I'd be happy to have this book keep itself up to date once and for all. (Though, I think it would be nice to have hand-typed descriptions for each ore, and if new mats are added, "tell a coder!" pops up for no desc.)

I don't know if it's possible in DM, but we do use TypeScript for TGUI stuff, and I'm wondering if we can run a script at the beginning of a round that reads the values (from an XML, or whatever) and then generates the resulting text file, which in turn gets loaded by the game. How we would generate the XML without manually creating it when all values exist in .DM files is beyond me, however.

MeggalBozale:
First step would prolly be autogenerating as much as possible, then getting the descriptions for unique materials and filling them in.
E.G.
[material_name]
[custom_description]
[material_properties]

Bohrum
A green, mildly conductive metal which closely resembles steel, but with higher hardness and density. It's useful for tools and equipment which need something a bit extra. (idk i made it up)
Electrical Conductivity: 5
Thermal Conductivity: 6
Hardness: 5
Density: 6
Flammability: 1
Chemical Resistance: 7

TobleroneSwordfish:
The smart string picker system lets you embed DM proc calls into text files, you could probably hack that into generating a matsci book with stuff like `[get_material_property("bohrum", "density")]`. I defer to pali on whether that's a good idea though :)

Glamurio:
> The smart string picker system lets you embed DM proc calls into text files, you could probably hack that into generating a matsci book with stuff like `[get_material_property("bohrum", "density")]`. I defer to pali on whether that's a good idea though 

@pali6 Is that a good idea? Might be better than to hardcode values at multiple locations, less likely to mess things up

pali6:
Sounds like a good idea yes! Note that you will need to pass the proc ref with its name as one of the arguments. Look at Father Grife's lines for an example of ping me on discord.

pgmzeta:
updated wiki

PR #15962, opened 2023-09-23
Revolutionary sign damage scales with number of nearby revs

unfunnyperson:
[BALANCE] [GAME OBJECTS]
## About the PR 

Revsign damage scales with nearby revs in a 5 tile radius, Each rev is 5 extra damage up to a max of 25 damage(The user will count toward this)
Lowers base damage of the sign to 3
Non-revs do not gain the damage buff if they attack using the sign

## Why's this needed? 
The revolutionary sign isn't a very good item in its current state. Its extremely loud, deals little damage, and the buff it gives isnt very good. Compared to the syndie dagger which has a very powerful stun ability and a cheaper cost, the revsign is a poor choice. This change is intended to make the revsign worth the 4 tc.

## Changelog

```changelog
(u)UnfunnyPerson
(+)Revolutionary sign damage scales depending on how many revs are nearby the user
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -1319,7 +1319,7 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/generic/head_rev)
 	name = "Revolutionary Sign"
 	item = /obj/item/revolutionary_sign
 	cost = 4
-	desc = "This large revolutionary sign will inspire all nearby revolutionaries and grant them small combat buffs. Additionally the sign will channel the fury of nearby revs to deal increased damage! Best used in conjunction with a horde of angry revolutionaries."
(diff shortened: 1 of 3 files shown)


code/datums/syndicate_buylist.dm
MeggalBozale, line 1: Looks good but maybe a rewording on 'increased damage' to be more IC

code/obj/item/misc_weapons.dm
MeggalBozale, line 1: You should probably just tally up the amount of revolutionaries in range, and then do the damage math at the end of it

MeggalBozale (changes requested):

Tarmunora (approved):

MeggalBozale (approved):
Good work

PR #15963, opened 2023-09-23
Makes syndicate thermals require a syndicate scanner

unfunnyperson:
[BUG]

## About the PR 
Adds is_syndicate = TRUE to traitor thermals, this makes it so you need a syndie scanner to scan them

## Why's this needed? 
bugs bad


MeggalBozale (approved):
Are people getting these from somewhere or somethin? I dont understand why its a bug or where this change came from is all

PR #15971, opened 2023-09-23
drop items on cig smoke cough

FlyzerMizer:
About the PR  makes you drop whatever your holding on your targeted hand upon coughing on a cigarettes puff of smoke

Why's this needed?  just some silly interaction i thought of, this is my first pull request and i did this to prove that i can learn DM and contributing to goonstation


TheTyrant25:
okay so first things first is that you should edit the pr description to be more like this:

## About the PR:
Makes you drop whatever you're holding on your targeted hand when you cough on a cigarette.

TheTyrant25:
Also the code looks like it'd work but it seems a bit too harsh on smoker roleplayers, at least in my opinion. Possibly make it a lower chance, so like only once in every four or five coughs makes you drop your held item.

FlyzerMizer:
> Also the code looks like it'd work but it seems a bit too harsh on smoker roleplayers, at least in my opinion. Possibly make it a lower chance, so like only once in every four or five coughs makes you drop your held item.

FlyzerMizer:
my bad, you can tell im a complete newbie at this

FlyzerMizer:
> Also the code looks like it'd work but it seems a bit too harsh on smoker roleplayers, at least in my opinion. Possibly make it a lower chance, so like only once in every four or five coughs makes you drop your held item.

ill try that, thanks for the feedback!

unfunnyperson:
> my bad, you can tell im a complete newbie at this

we all started somewhere, dont worry about it

code/obj/item/cigarette.dm
```diff
@@ -258,7 +258,10 @@
 
 			var/mob/living/carbon/human/human_target = target
 			if (human_target && rand(1,5) == 1)
-				SPAWN(0) target.emote("cough")
+				SPAWN(0)
+				target.emote("cough")
+				if (rand(1, 5) == 1)
```
TealSeer, line 263: BYOND has a quicker wait to do this kind of thing with the prob proc, i.e. prob(20) for a 20% chance

TealSeer (approved):

code/obj/item/cigarette.dm
```diff
@@ -258,7 +258,10 @@
 
 			var/mob/living/carbon/human/human_target = target
 			if (human_target && rand(1,5) == 1)
-				SPAWN(0) target.emote("cough")
+				SPAWN(0)
```
CheffieGithub, line 261:
```suggestion

```
This spawn is doing nothing here its best to just remove it. change the rand() stuff above to prob(20) as well.

CheffieGithub (approved):
code is good

code/obj/item/cigarette.dm
```diff
@@ -257,8 +257,10 @@
 			user.visible_message("<span class='alert'><B>[user]</B> blows smoke right into <B>[target]</B>'s face![message_append]</span>", group = "[user]_blow_smoke_at_[target]")
 
 			var/mob/living/carbon/human/human_target = target
-			if (human_target && rand(1,5) == 1)
-				SPAWN(0) target.emote("cough")
+			if (human_target && prob(20))
+				target.emote("cough")
```
Tarmunora, line 261: should check the `issmokeimmune()` macro here (which covers checks for gas masks, internals, anaerobic, etc.)

Tarmunora (changes requested):

code/obj/item/cigarette.dm
FlyzerMizer: im new to this so how would i go about doing that?

CheffieGithub: `if (human_target && !issmokeimmune(human_target) && prob(20))`

PR #10697, opened 2022-09-06
Adds dwarves as a trait

mantwerp:
## About the PR 
Adds dwarves as a trait, available to be chosen from the character setup menu

## Why's this needed? 
Popular consensus on the forum thread https://forum.ss13.co/showthread.php?tid=19629

## Changelog

```changelog
(u)Mantwerp
(+)Adds a dwarf trait, available to be chosen from the character setup menu
```

PR #15978, opened 2023-09-23
Adds dwarves as a trait

Mantweerp:
[FEATURE] [TRAITS]

## About the PR 
Adds a dwarf trait, which is a body trait. It costs one trait point, and has a custom icon (made by mcdangus 🐀).

## Why's this needed? 
I've said my main reasons in this forum post: https://forum.ss13.co/showthread.php?tid=21822 (yell at me here)
The short of it is that genetics takes far too long for something that is sometimes a major part of a character's gimmick, which wastes a lot of time that you can spend actually doing your gimmick.

## Changelog 

```changelog
(u)Mantwerp
(*)Added a dwarf trait. Rock and Stone! icon sprite by mcdangus
```


TealSeer (approved):

_std/macros/genetics_traits.dm
```diff
@@ -1,2 +1,3 @@
 //for trait / genetics checks
 #define isalcoholresistant(x) ((x.bioHolder && x.bioHolder.HasEffect("resist_alcohol")) || (x.traitHolder && x.traitHolder.hasTrait("training_drinker")))
+#define isdwarf(x) ((x.bioHolder && x.bioHolder.HasEffect("dwarf")) || (x.traitHolder && x.traitHolder.hasTrait("dwarf")))
```
kbsmilk, line 3:
```suggestion

```
If you're not using this, you don't need to add it.
Dwarfism also doesn't have mechanical effects (afaik), unlike the alcoholic trait.

Mantweerp: Yea I mostly just copied whatever the career alcoholic trait did seeing as that was functionally the same (and im not exactly the greatest coder). Its removed now though

DimWhat:
#10697 is relevant. None of my sprite issues have been addressed since last year so I'm still against it.

_std/macros/genetics_traits.dm
Valtsu0: > If you're not using this, you don't need to add it. Dwarfism also doesn't have mechanical effects (afaik), unlike the alcoholic trait.

Dwarfism gives cold res right?

TobleroneSwordfish:
Please don't open a copy of a rejected PR with no substantive changes, none of the concerns on the original PR or in the linked forum threads have been addressed.

PR #15980, opened 2023-09-23
Fix backpack TTV sprite overlay code

MeggalBozale:
## About the PR 
Fixes putting one/two tanks + removing tanks from back-worn TTVs

## Why's this needed? 
Makes #15929 work properly. And bugfixes are good!!!

Thank u so much to pali for the help with the code, I never touched overlays before


MeggalBozale:
ah heck i did it wrong

PR #15982, opened 2023-09-23
Wasp related items in surplus crates come with a wasp implant

unfunnyperson:
## About the PR 
If wasp grenades are rolled in a surplus crate they come with a wasp implant. Wasp implants give the user the botany faction which prevents wasps from attacking them. Wasp implants also cause the user to explode into 8 wasps on death.
Additionally enables the wasp crossbow in surplus crates and adds a wasp implant when that is rolled.

## Why's this needed? 
Wasp grenades are significantly less useful when rolled in a surplus crate because the user lacks botany's wasp immunity. The implant allows this item to be used to its full potential. 
I added the crossbow to crates cause I noticed a comment about it being disabled due to issues with wasps killing their summoners but that shouldn't an issue because of the implant.

The user exploding into wasps on death is meant to be a fun little bonus when you get one of these items in a surplus.

## Changelog 

```changelog
(u)UnfunnyPerson
(+)Added the wasp crossbow to surplus crates.
(+)Wasp related items in surplus crates come with a wasp implant. The wasp implant will cause wasps to ignore the user and for the user to explode into wasps on death.
```

### DIFF
--- a/code/obj/item/implant.dm
+++ b/code/obj/item/implant.dm
@@ -706,6 +706,35 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
 			src.owner?.elecgib()
 		. = ..()
 
+/obj/item/implant/revenge/wasp
+	name = "wasp implant"
+	big_message = " buzzes, what?"
+	small_message = "buzzes loudly, uh oh!"
+	power = 4
+	/// did we change our user's faction? this is in case a botanist gets this implant then gets it removed
+	var/changed_faction = FALSE
+
+	implanted(var/mob/M, mob/I)
+		..()
+		if (istype(M) && M.faction != FACTION_BOTANY)
+			M.faction = FACTION_BOTANY
+			changed_faction = TRUE
+
+	on_remove(var/mob/M)
+		..()
+		if (istype(M) && changed_faction == TRUE)
+			M.faction = 0
+			changed_faction = FALSE
+
+	do_effect(power)
+		// enjoy your wasps
+		for (var/i in 1 to power)
+			new /mob/living/critter/small_animal/wasp/angry(get_turf(src))
+
+		SPAWN(1)
+			src.owner?.gib()
+		. = ..()
+
 
 /obj/item/implant/robotalk
 	name = "machine translator implant"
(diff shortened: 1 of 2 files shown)


MeggalBozale:
Personally I think the amount of wasps you burst into on death should be a bit more, especially since you gib for it. 8 pops into my head, but just IMO

code/obj/item/implant.dm
```diff
@@ -706,6 +706,39 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+	name = "wasp implant"
+	big_message = " buzzes, what?"
+	small_message = "buzzes loudly, uh oh!"
+	power = 8
+	/// did we change our user's faction? this is in case a botanist gets this implant then gets it removed
+	var/changed_faction = FALSE
+
+	implanted(var/mob/M, mob/I)
+		..()
+		if (istype(M) && M.faction != FACTION_BOTANY)
```
Tarmunora, line 719: `faction` is a bitfield - use bitwise operators when dealing with it when appropriate

Tarmunora (changes requested):
Faction implant idea is cool, not too sure on having it also be a revenge implant - probably fine though

Tarmunora (approved):

code/obj/item/implant.dm
```diff
@@ -706,6 +706,35 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+
+	on_remove(var/mob/M)
+		..()
+		if (istype(M))
+			M.faction &= ~FACTION_BOTANY
+
+	do_effect(power)
+		// enjoy your wasps
+		for (var/i in 1 to power)
+			var/mob/living/critter/small_animal/wasp/W = new /mob/living/critter/small_animal/wasp/angry(get_turf(src))
```
pgmzeta, line 728: if you spawn them inside the owning mob, they'll all burst out on gib :)

CheffieGithub: This will also throw them outwards and can damage / them

unfunnyperson: not a fan of this because like cheffie said, the wasps will spawn stunned/damaged

PR #16013, opened 2023-09-25
Refactor armored suit variants

Glamurio:
[Clothing] [Internal]

## About the PR 
Refactors the assembly of armored suit variants with the help of  the assembly component `/datum/component/assembly`.

All suits now have an `armor_variant` which is a path to the new armor that it spawns after combination.
Additionally, suits now have a `default_variant`, this gets set to the `src.type` after applying armor and is used to restore the original suit after removing the armor.

Armored suits can now be interacted with a cutting or snipping tool to remove the armor from the suit, which yields both armor and suit again.

## Why's this needed? 
Components good. Also being able to restore bullet vests after someone slapped them on somewhere is also good.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Armored suits can now be separated with a cutting or snipping tool.
```

### DIFF
--- a/code/obj/item/clothing/suits.dm
+++ b/code/obj/item/clothing/suits.dm
@@ -19,7 +19,14 @@ ABSTRACT_TYPE(/obj/item/clothing/suit)
 	var/team_num
 	/// Used for the toggle_hood component, should be the same as the default icon_state so it can get updated with medal rewards.
 	var/coat_style = null
+	// Used for the armor_suit proc to check if the suit has an armored variant
+	var/armor_variant = null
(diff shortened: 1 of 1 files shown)


Tarmunora:
Eh, if we're doing a refactor of armored suits, what I'd really prefer to see would be refactoring them to have the original clothing persist and be modified, rather than deleting the components and spawning a new item etc.

Glamurio:
> Eh, if we're doing a refactor of armored suits, what I'd really prefer to see would be refactoring them to have the original clothing persist and be modified, rather than deleting the components and spawning a new item etc.

So just replace the sprite but apply the stats?

PR #16016, opened 2023-09-25
Fix jean plated clothing losing special naming on UpdateName()

### LINKED ISSUE #15959: washing a jean-plated item removes the jonsonant jeplacement from its name
### Labels

[BUG][TRIVIAL]

### Description

if you wash a jean-plated item in the sink (presumably in a washing machine too) you lose the silly consonant replacement in its name. presumably it gets washed off like the "blood-stained" label does

### Steps to reproduce

1. put together a joutfit
2. get blood on it
3. go to rinse it in the sink
4. see that your jechanical joots have reverted back to simple mechanical boots

### Expected Behavior

for the name to remain the same

Reported by: Waffleloffle
Round log date: 2023-09-23-01-56
Reported on: 2023-09-23 02:01:10
Map: OSHAN

TealSeer:
[BUG] [CLOTHING] [MATERIALS]
## About the PR 

Fixes #15959 by making UpdateName() on clothing apply any material-based special naming the item has.

## Why's this needed? 

Currently materials that do special naming (which I think rn is only jean) only apply that naming to src.name when SetMaterial is called, as soon as something else updates the name with real_name, it's lost

### DIFF
--- a/code/obj/item/clothing.dm
+++ b/code/obj/item/clothing.dm
@@ -60,7 +60,10 @@ ABSTRACT_TYPE(/obj/item/clothing)
 
 
 	UpdateName()
-		src.name = "[name_prefix(null, 1)][src.get_stains()][src.real_name ? src.real_name : initial(src.name)][name_suffix(null, 1)]"
+		src.name = src.real_name ? src.real_name : initial(src.name)
+		if(src.material?.usesSpecialNaming())
+			src.name = src.material.specialNaming(src)
+		src.name = "[name_prefix(null, 1)][src.get_stains()][src.name][name_suffix(null, 1)]"
 
 	proc/add_stain(var/stn)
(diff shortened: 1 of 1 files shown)


code/obj/item/clothing.dm
```diff
@@ -60,7 +60,10 @@ ABSTRACT_TYPE(/obj/item/clothing)
 
 
 	UpdateName()
-		src.name = "[name_prefix(null, 1)][src.get_stains()][src.real_name ? src.real_name : initial(src.name)][name_suffix(null, 1)]"
+		src.name = src.real_name ? src.real_name : initial(src.name)
```
TobleroneSwordfish, line 63:
```suggestion
		src.name = src.real_name || initial(src.name)
```

TobleroneSwordfish (approved):

PR #16017, opened 2023-09-25
Makes continous reactions not leave behind fractional amounts of chems

### LINKED ISSUE #15691: new Sulfuric Acid recipe does not react to completion
### Labels

[BUG][MINOR]

### Description

Certain amounts of sulfur, oxygen, and water will leave remnants, others will not

### Steps to reproduce

1. Add 2u oxygen, sulfur, water (recipe added by #15634
2. wait for rxn to complete
3. it never does. Leaves 0.5u of each of the precursors leftover.
4. add 2u of each again
5. it reacts to completion

### Expected Behavior

Adding whole numbers of these reagents will not leave a remnant due to the reaction being non-instant

Reported by: Mintyphresh
Active test merges: #15458 and #15634
Round log date: 2023-09-04-22-32
Reported on: 2023-09-04 22:57:42
Map: KONDARU

M-Earthfire:
[chemistry][bug]

## About the PR 

This PR makes `reagents/proc/handle_reactions` skip the check `required_volume` on reactions of the reaction in question is already processing since last tick and non-instant. It will still check if a `CHEM_EPSILON` amount of the chem is avaible.

This fixes instances where a continous reaction is taking place and leaves on the last tick fractional amounts of chems, making it stop reacting.

## Why's this needed? 

There exist a check for `required_volume` in `reagents/proc/process_reactions` which handles this exact behaviour and considers reaction rate. This most often would not come in use since `reagents/proc/handle_reactions` would terminate the reaction beforehand. Thus, this PR changes that.

This will cause reactions to start if the full required amount of chems is in the container (so 0.05u of oil in 50u welding and 5u chlorine won't start any reaction, it will need the required 1u of oil, like intended). The change makes the reaction work until one or all of the reactants runs out.

This fixes #15691 

This fixes the issue #15401 tried to solve, but instead of adjusting the recipes, this should fix the issues in general.
This touches kinda the core of chemical reactions. I tested many of the non-instant recipes, and all worked as intended, so i hope this doesn't break anything.


Tarmunora (approved):

Flaborized:
Good idea! I like this implementation and this fixes the issue pretty much outright, though I didn't look at the code to review that part the idea is good 👍 :sheltersweat:

TobleroneSwordfish:
I believe in this not breaking anything!

PR #16019, opened 2023-09-25
Introduces the COMSIG_MOB_EX_ACT signal and makes syringe implants remove on explosion of the mob

M-Earthfire:
[Internal][Balance][Game Objects]

## About the PR 

Introduces the `COMSIG_MOB_EX_ACT` signal. It is send on `ex_act` on a mob.

Also, this PR uses this signal to make syringe implants caused by the syringe gun to be removed when the mob carrying them suffers from an explosion.

## Why's this needed? 

Pottassium+Ice in syringe guns proven to be far too effective for how little effort they need to set up. While it is nice to explode people with this combination, a single syringe could cause 4-5 explosions with this method.

By changing syringes to be removed on explosion of the mob, a single syringe can only explode the person at two times at maximum (due to delays) until the syringe+the chemicals are removed. 

In general, this makes getting hit with a PotIce syringe gun much more surviveable and severly diminishes their utility for crowd control while still keeping this mixture viable for the syringe gun. Get some better chem mixes, nerds.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Syringes, fired from the syringe gun, stuck in a human desintegrate when that human is affected by an explosion.
```

### DIFF
--- a/_std/defines/component_defines/component_defines_atom.dm
+++ b/_std/defines/component_defines/component_defines_atom.dm
@@ -219,6 +219,8 @@
 	#define COMSIG_MOB_UPDATE_DAMAGE "mob_update_damage"
 	/// Sent when a mob resists, return TRUE to prevent other resist code from running
 	#define COMSIG_MOB_RESIST "mob_resist"
+	/// Sent when the mob is affected by an explosion
+	#define COMSIG_MOB_EXPLODES "mob_explosion"
 
 	// ---- cloaking device signal ----
 
--- a/_std/defines/component_defines/component_defines_atom.dm
+++ b/_std/defines/component_defines/component_defines_atom.dm
@@ -220,7 +220,7 @@
 	/// Sent when a mob resists, return TRUE to prevent other resist code from running
 	#define COMSIG_MOB_RESIST "mob_resist"
 	/// Sent when the mob is affected by an explosion
-	#define COMSIG_MOB_EXPLODES "mob_explosion"
+	#define COMSIG_MOB_EX_ACT "mob_explosion_act"
 
 	// ---- cloaking device signal ----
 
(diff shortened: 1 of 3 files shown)


_std/defines/component_defines/component_defines_atom.dm
```diff
@@ -219,6 +219,8 @@
 	#define COMSIG_MOB_UPDATE_DAMAGE "mob_update_damage"
 	/// Sent when a mob resists, return TRUE to prevent other resist code from running
 	#define COMSIG_MOB_RESIST "mob_resist"
+	/// Sent when the mob is affected by an explosion
+	#define COMSIG_MOB_EXPLODES "mob_explosion"
```
Tarmunora, line 223:
```suggestion
	/// Sent when the mob is affected by an explosion
	#define COMSIG_MOB_EX_ACT "mob_explosion_act"
```
and rename throughout

code/obj/item/implant.dm
```diff
@@ -1020,6 +1020,22 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
 				..()
 				implant_overlay = image(icon = 'icons/mob/human.dmi', icon_state = "syringe_stick_[rand(0, 4)]", layer = MOB_EFFECT_LAYER)
 
+			implanted(var/mob/receiving_mob, var/mob/implanting_mob)
+				..()
+				RegisterSignal(receiving_mob, COMSIG_MOB_EXPLODES, PROC_REF(on_explosion_reaction))
+
+			on_remove(var/mob/loosing_mob)
+				..()
+				UnregisterSignal(loosing_mob, COMSIG_MOB_EXPLODES)
```
Tarmunora, line 1029:
```suggestion
			on_remove(var/mob/losing_mob)
				..()
				UnregisterSignal(losing_mob, COMSIG_MOB_EXPLODES)
```

Tarmunora (changes requested):
should call it `...EX_ACT` imo for clarity

TobleroneSwordfish:
Good solution 👍

PR #16028, opened 2023-09-26
Limit AI cross Z level interactions

Azrun:
## About the PR 

The following AI & AIeye interactions are limited to the AI core being on the same Z level:
* Deployment to Shells
* Opening nearest Door

The following AI & AIeye interactions are limited to the AI core being on the same Z level or from the station to an unrestricted Z level:
* Interactions with tgui.
* Hotkey Interactions (Open, Bolt, Shock)
* Attack AI

The following AI & AIeye interactions are limited to the AI core being on being on the station Z level:
* Holograms
* Calling Emergency Shuttle
* Viewing the Crew Manifest
* Resetting APCs
* Removes electrification from all airlocks on the station
* Unbolts all airlocks on the station
* AI Station Announcement

Killswitching the AI deals damage to the brain AND the core.  This will require repair of the core prior to replacement of the brain (wires and welding).

## Why's this needed? 
The creation of an AI off station has problematic gameplay impacts. There is reduced ability for counterplay by the crew as they are forced to find the AI which can be difficult or impossible.  This limits the impact a remote AI has to the station while still allowing for general observation and communication.  

Assuming the killswitch is/was available organ damage is done to the brain to ensure that has lasting effect.

## Changelog 

```changelog
(u)Azrun
(*)Reduced ability for AI and AIeye to interact across levels.
```


Glamurio:
How would this affect Salvagers, who routinely have the ability to build an AI off-station? I'm worried that with the inability to operate doors, deploy to shell or TGUI there's hardly any reason to go through the efforts of assembling an AI and likely sacrificing a player in the process.

Azrun:
> How would this affect Salvagers, who routinely have the ability to build an AI off-station? I'm worried that with the inability to operate doors, deploy to shell or TGUI there's hardly any reason to go through the efforts of assembling an AI and likely sacrificing a player in the process.

Directly.  
This PR came about to address recent situations in which an AI **is** built off station.  
* Assuming there is someone to slap the brain back in the core, killswitching has negligible impact (assuming the console wasn't removed/destroyed/disabled).
* Areas like the M4GP13 are inaccessible to the crew so no direct action can be taken against the AI.

The crew is left with little to no recourse outside of removing AI control throughout the station which then severely hampers the Station AI and Cyborgs. 

Also I really don't think the salvagers need more ways to operate doors, they can make ~~doors~~ holes.  They also have a tool that tells them which wires do what to doors.

Keiiro:
> > Shell deployment is a big point of contention in my view, as it would be more difficult for AI shells to assist in off-station and off-Z level situations like azoning or rescues if they can't (un)deploy from them seamlessly.
> 
> 8/10 successful generated planet adventures I've seen involve the AI or other silicon saving the crew.
> 
> Interesting point. Do you think it should list all shells in the same Z level and ALL shells if on the station level?

That sounds like a good solution. Another way it could be done is to have all shells displayed, but lose access to ones outside of your current Z level if they're low on of energy.

Retrino:
a big concern about this imo is how easy it could be to soft lock the ai by throwing them out the export belt, or sending them to an azone and disabling telesci. the ai could call for help, sure, but theres no guarantee someone would care enough to hunt down the core in either situation. 

my suggestion would be possibly making it work via lawrack instead. an ai that’s lawd up on the salvager base wouldnt have control over the standard station stuff or the ai shells that were already made over there, but still keeps the ability to spy and give information back to salvagers (and still adjust records if they build a console on the magpie).

Azrun:
> a big concern about this imo is how easy it could be to soft lock the ai by throwing them out the export belt, or sending them to an azone and disabling telesci. the ai could call for help, sure, but theres no guarantee someone would care enough to hunt down the core in either situation.
> 
> my suggestion would be possibly making it work via lawrack instead. an ai that’s lawd up on the salvager base wouldnt have control over the standard station stuff or the ai shells that were already made over there, but still keeps the ability to spy and give information back to salvagers (and still adjust records if they build a console on the magpie). this could also lightly prevent cases of someone building a lawrack off station to perma rogue an ai on station

Of the instances that I am aware of the lawrack has not been placed on the Magpie. 🤔

In the azone situation there isn't much difference compared to a human, you are limited to others assisting you.  Though export belt is a little more rough, but presumably the AI had some way of engaging with the situation and it seems like they failed.

Glamurio:
> In the azone situation there isn't much difference compared to a human, you are limited to others assisting you. Though export belt is a little more rough, but presumably the AI had some way of engaging with the situation and it seems like they failed.

The main difference between a human and an AI core is that humans can be infinitely cloned on death, but AI core brains cannot be. Not to mention a new core will need to be constructed, of which there are limited available.

Retrino:
> > a big concern about this imo is how easy it could be to soft lock the ai by throwing them out the export belt, or sending them to an azone and disabling telesci. the ai could call for help, sure, but theres no guarantee someone would care enough to hunt down the core in either situation.
> > my suggestion would be possibly making it work via lawrack instead. an ai that’s lawd up on the salvager base wouldnt have control over the standard station stuff or the ai shells that were already made over there, but still keeps the ability to spy and give information back to salvagers (and still adjust records if they build a console on the magpie). this could also lightly prevent cases of someone building a lawrack off station to perma rogue an ai on station
> 
> Of the instances that I am aware of the lawrack has not been placed on the Magpie. 🤔
> 
> In the azone situation there isn't much difference compared to a human, you are limited to others assisting you. Though export belt is a little more rough, but presumably the AI had some way of engaging with the situation and it seems like they failed.

a law rack *could* be placed on the magpie for the ai to be connected to, like the syndicate having its own law rack. a pre existing one, not one that gets built by the salvagers. if they have an ai core, itd make sense for them to have one tucked away, at least. 

i heavily disagree that the ai core is in the same boat as a human that’s been trapped. along with the point glamurio mentioned, an ai core has no way of moving itself, while a human that was trapped does. some azones have built in methods of return to station that the ai core simply couldnt reach. if they threw in a signal jammer as well, the ai core would be permanently stuck in that location with no means to move away and contact anyone. 

the pr does address issues of off z ais causing chaos, but i do feel the method could be refined in a way that doesn’t potentially cause a soft lock on normal ais.

PR #16031, opened 2023-09-26
Yeti resprite

flappybatpal:
## About the PR 

- [ ] Dead sprite

Updates the yeti sprites with a new sprite by disgustingfreakfingers
image

## Why's this needed? 
I believe we do not have permission to use the current sprites.

## Changelog 

```changelog
(u)disgustingfreakfingers
(+)Updated yeti sprite. Disregard for perspective and realism remains intact.
```


Glamurio:
If we don't have permission to use the current sprites, and the new sprites are _very_ similar to the old ones, aren't we still somehow infringing on rights?

ZeWaka:
> If we don't have permission to use the current sprites, and the new sprites are _very_ similar to the old ones, aren't we still somehow infringing on rights?

Not really, there's a big difference between `literally the exact pixels` and *resembling* another image. I can paint a shitty andy warhol painting and sell it all I want.

PR #16032, opened 2023-09-26
Reduce blob spawn minimum, increase players per blob

MeggalBozale:
[Balance]
## About the PR 
Reduces the blob minimum to 1 and the players per blob (roughly) from /18 to /25. Also changes rand(-5,0) to rand(0,5) since I feel like that evens it out a bit. The scaling change is so that 30 players don't get rounded up to 2 blobs. Maybe need a flooring proc if round acts like ceil, unsure.

## Why's this needed?
image
It is certainly a bad idea with how strong blobs can be to put 20-30 people against two blobs at once. One blob can be hard enough.

## Changelog
```changelog
(u)MeggalBozale
(+)Blob Overminds feel they can do more with less and now send less blobs to smaller stations.
```

### DIFF
--- a/code/datums/gamemodes/blob.dm
+++ b/code/datums/gamemodes/blob.dm
@@ -5,7 +5,7 @@
 	shuttle_available_threshold = 20 MINUTES
 
 	antag_token_support = TRUE
-	var/const/blobs_minimum = 2
+	var/const/blobs_minimum = 1
 	var/const/blobs_possible = 4
 	var/const/waittime_l = 600 //lower bound on time before intercept arrives (in tenths of seconds)
 	var/const/waittime_h = 1800 //upper bound on time before intercept arrives (in tenths of seconds)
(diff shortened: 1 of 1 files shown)


MeggalBozale:
There's a good chance the scaling may not be 'balanced' as-is(?) Perhaps a feedback thread may be a good idea. But LMK what you think first

Glamurio:
Personally, I think Blob spawn rates are fine, but that's probably based on the fact that in my over 600 hours of playtime have never even seen a Blob, let alone played as one, so I might be biased. 🙃

AXLFAN2000:
I like this change, rounds with 2 blobs and 30 or less players feel absolutely brutal

TheColorCyan:
> How in the world have you never seen a blob in 600h :O they arent THAT rare

RP has blobs fully turned off and i will just assume they play rp

Good change tho, i love it. Even one blob is barely possible to outdamage (pls buff blob counters) on lowpop and it usually spawns two so i support this.

Tarmunora (approved):

MeggalBozale:
For the sake of it, here's the change graphically.

Current worst case (Lowest blobs, rand = -5)
qLhSFOCQKU

Current best case (Most blobs, rand = 0)
Y0n6hiO4gG

Proposed worst case (Lowest blobs, rand = 0)
dyrKXmaM9R

Proposed best case (Most blobs, rand = 5)
QVlWCraY3Q

I anticipate that some blob rounds on lower population could end early due to the only chosen blob being new. However, the blob round could be a sweep should a more robust blob player get chosen. The skill floor of blob combined with the drastic change of adding or removing a blob makes this antagonist gamemode difficult to balance. But in its current state, two blobs on a very low pop is a near guaranteed loss, and this should not be the case.

ZeWaka (approved):

PR #16040, opened 2023-09-26
Pressure sensor streamlining

kbsmilk:
[GAME OBJECTS] [QOL]
## About the PR 
* Makes the pressure sensor easier to use in general. Mousedrop support + crystal extraction with bare hand (or crowbar like it was before)
* pressure crystal survives the sensor getting exploded.
* visible messages instead of just boutput.
* +examine text. +code quality. +overlay bug fixed.

## Why's this needed? 
Less clunk. Something of a companion PR to #15933 

## Changelog 

```changelog
(u)Smilg
(+)Pressure sensor can now use empty hand, and mouse drop.
```


code/obj/item/device/transfer_valve.dm
```diff
@@ -518,45 +518,103 @@ TYPEINFO(/obj/item/device/transfer_valve/briefcase)
...
+		if (!istype(thing, /obj/item/pressure_crystal))
+			return
+		if (src.insert_crystal(user, thing))
+			user.visible_message("[user] drops [thing] into [src].",
+								"<span class='notice'>You drop the crystal into [src].</span>")
+
+	attack_hand(mob/user)
+		if (src.loc != user || !user.find_in_hand(src))
+			..()
+		else if (src.crystal && src.remove_crystal(user))
```
TobleroneSwordfish, line 586: This removes the crowbar requirement, either make it require a crowbar to extract or remove the crowbar interaction entirely.

code/obj/item/device/transfer_valve.dm
```diff
@@ -518,45 +518,103 @@ TYPEINFO(/obj/item/device/transfer_valve/briefcase)
...
-		else return ..()
+		if (src.remove_crystal(usr, over_location))
+			usr.visible_message("[usr] removes the crystal from [src].",
+								"<span class='notice'>You pull the crystal out of [src].</span>")
+
+	MouseDrop_T(atom/movable/thing, mob/user)
+		. = ..()
+		if (!istype(thing, /obj/item/pressure_crystal))
+			return
+		if (src.insert_crystal(user, thing))
```
TobleroneSwordfish, line 579: This displays "You contemplate how to place the crystal in an occupied sensor, but can't manage to figure out how." when putting a crystal into an empty sensor.

TobleroneSwordfish (changes requested):

code/obj/item/device/transfer_valve.dm
kbsmilk: Can't reproduce this with mouse drop. It should be fully handled in insert_crystal()

PR #16041, opened 2023-09-26
shittiest elevator animation code known to man

Skeletonman0:
[OBJECTS] [FEATURE]
## About the PR 
makes elevators do a little animation when going up / down

is the code good? hell no
does it look nice? when it works

does it even work?   sometimes????

## Why's this needed? 

it looks cool when it works


ZeWaka:
video?

PR #16045, opened 2023-09-26
Add `IMMUNE_TRENCH_WARP` to more things

pgmzeta:
## About the PR 

Searched through `event_handler_flags` in repo to find places where IMMUNE_TRENCH_WARP should be redeclared.

## Why's this needed? 
Fix stuff falling into the trench that shouldn't

### DIFF
--- a/code/datums/controllers/process/forcedMovement.dm
+++ b/code/datums/controllers/process/forcedMovement.dm
@@ -1,5 +1,3 @@
-/obj/effect/event_handler_flags = IMMUNE_MANTA_PUSH //mbc dont know where to put this because we have no master effects file lol lets leave it here, maybe someone wil clean up later >:)
-
 proc/BeginSpacePush(var/atom/movable/A)
 	if (!(A.temp_flags & SPACE_PUSHING))
 		spacePushList += A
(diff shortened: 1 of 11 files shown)


code/obj/shieldgen.dm
```diff
@@ -156,16 +156,6 @@ Shield and graivty well generators
 
 		return 1
 
-/obj/shieldwall
```
Glamurio, line 159: You removed shieldwalls, either by accident or on purpose.

pgmzeta: this is the shorter of two duplicate definitions, the other still in [code/obj/item/device/energy_shield_gen.dm](https://github.com/goonstation/goonstation/pull/16045/files/9fa500f956fc07fc939ce8696d972ae6d5ba1749#diff-457e91bffd5a4a80de7d782f30cfe2ffa995677e7ed406df48f974c069e81919)

code/datums/controllers/process/forcedMovement.dm
```diff
@@ -1,5 +1,3 @@
-/obj/effect/event_handler_flags = IMMUNE_MANTA_PUSH //mbc dont know where to put this because we have no master effects file lol lets leave it here, maybe someone wil clean up later >:)
```
pgmzeta, line 1: moved to `/code/obj/effect.dm`

PR #16051, opened 2023-09-27
Don't trigger action bar for Pod Eject Occupants if no occupants in pod cabin

### LINKED ISSUE #7425: Corpses wont eject from pods
### Labels

[MINOR]

### Description

Corpses dont eject from pods

### Steps to reproduce

Stuff a corpse in a pod, attempt to eject occupents later, the action bar will fill up but when it completes the corpse is still in the pod and I still can eject occupents from the pod, futily. 

### Expected Behavior

Corpses to fall out of the pod when ejected

Reported by: Subject_Name
Round log date: 2022-02-07-02-53
Reported on: 2022-02-07 03:56:48

pgmzeta:
## About the PR 

when completing the eject occupants action bar, in addition to the driver/passenger ejection, check if the secondary system of the pod is storage, and if so unload any mobs in the direct load list.

## Why's this needed? 
Fix #7425
it's a pretty long action bar for it to do nothing if the intent is to empty people

### DIFF
--- a/code/modules/transport/pods/vehicle.dm
+++ b/code/modules/transport/pods/vehicle.dm
@@ -1077,7 +1077,7 @@
 	actions.start(new/datum/action/bar/icon/eject_pod(src,usr), usr)
 	return
 
-/obj/machinery/vehicle/proc/eject_pod(mob/user, dead_only=FALSE)
+/obj/machinery/vehicle/proc/eject_pod(var/mob/user, var/dead_only = 0)
 	for(var/mob/M in src) // nobody likes losing a pod to a dead pilot
 		if (!dead_only)
 			leave_pod(M)
@@ -1086,7 +1086,6 @@
 			if(M.stat || !M.client)
 				leave_pod(M)
 				boutput(user, "<span class='alert'>You pull [M] out of [src].</span>")
-
 			else if(!isliving(M))
 				leave_pod(M)
 				boutput(user, "<span class='alert'>You scrape [M] out of [src].</span>")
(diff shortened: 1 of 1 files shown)


Glamurio (approved):

TobleroneSwordfish (approved):
I guess?

UrsulaMejor (commented):
I don't think this is needed. You load corpses in by click dragging them to the pod, and you unload them by clickdragging the pod to the floor. There's really no need for this.

Instead, ejecting occupants should just not start the bar if there's nobody in it to eject. People getting forgotten in the trunk of the pod is a feature, not a bug.

code/modules/transport/pods/vehicle.dm
```diff
@@ -1074,8 +1074,11 @@
 		boutput(usr, "<span class='alert'>[src] is locked!</span>")
 		return
 
-	actions.start(new/datum/action/bar/icon/eject_pod(src,usr), usr)
-	return
+	for(var/mob/M in src)
+		actions.start(new/datum/action/bar/icon/eject_pod(src,usr), usr)
+		return
```
UrsulaMejor, line 1079: Should instead use locate(Type) in Container 
e.g.,
```

if(locate(/mob) in src.contents)
    actions.start(new/datum/action/bar/icon/eject_pod(src,usr), usr)
    return
```

UrsulaMejor (changes requested):

UrsulaMejor (commented):
ye

PR #16060, opened 2023-09-27
Minor update to donut 2 toxins and bomb area.

CheffieGithub:
## About the PR 
Moves the chamber access to the bottom so the north area isn't so cramped.
Adds an emergency space set and jetpack for bomb retrieval.
Adds lattices with landing lights to bomb launch start and end.
Moves around the chamber heating pipes.
Adds an electrical toolbox since wires are needed for certain bomb assemblies.
image

image

## Why's this needed? 
The top room is incredibly cramped at the minute and scientists should have some fast access to their testing area.


CheffieGithub:
Now that should be everything

Sord213 (changes requested):
Couple minor things. I'm fine with the airlock but I'd rather them use a pod if needed rather than giving them a free jetpack. The glass airlock to the launcher is intended to have no access on it for crime purposes. And finally as a personal minor gripe, leave the lights directionless. They set their dir automatically and it makes me think they're old lights, when I look at it, that never got replaced.

CheffieGithub:
image
Lights are all south now. 
I think a micro jetpack has much less value than a pod but I made it into a gas tank regardless.
Access requirements removed from the thindoor. 
Doors changed to the toxins variant unless you object?

PR #16069, opened 2023-09-27
Gasmasks protect eyes from capcaisin

CrystalClover:
[CLOTHING] [CHEMISTRY] [QoL] [STATUS EFFECTS]

## About the PR 
Allows gasmasks to protect the user from the blinding/stunning effects of capcaisin gas (e.g. crowd dispersal grenades)
(you will still need an oxygen supply to prevent capcaisin inhilation, which still does stun the user)

## Why's this needed? 
The effects are described as burning the players eyes, so it makes sense if a gasmask fitted to a player's face would protect their eyes from chemicals.


Glamurio (approved):

Wisemonster:
For reference, the reason I didn't let gas masks completely block capsaicin in the pr that changed it's effects is cause at the time, gas masks were rather easy to get, even being printable from the science fabricator, so I excluded it for balance purposes rather than logical one. Though considering gas masks are somewhat difficult to get in mass quantities nowadays, this change seems fine.

UrsulaMejor (approved):

UrsulaMejor:
This also makes other masks work if they cover the eyes. I don't think that's necessarily bad to do.

PR #16074, opened 2023-09-28
fix bald trait not updating hair properly

### LINKED ISSUE #15988: Bald trait not working as intended
### Labels

[BUG][TRIVIAL]

### Description

I have hair and a wig on spawn

### Steps to reproduce

Have bald trait and hair
Spawn
Observe as you have hair that can't be cut in the barbers chair but is visually present on the model while you still have a wig in the backpack

### Expected Behavior

Being bald

Reported by: ZombieApocalypseDK
Round log date: 2023-09-24-09-53
Reported on: 2023-09-24 10:54:56
Map: OSHAN

zjdtmkhzt:
## About the PR 
fixes #15988 
this was caused by #15780

I am not 100% sure if update_colorful_parts() is the best thing to use here or if there is a lesser proc just for hair that should be used instead?
Also, the barber code seems to also call set_clothing_icon_dirty() after changing hair, but I am not sure if this is actually necessary?

Anyway, this seems to fix the bug with the hair still appearing.
## Why's this needed? 
the people demand baldness

### DIFF
--- a/code/procs/jobprocs.dm
+++ b/code/procs/jobprocs.dm
@@ -680,6 +680,7 @@ var/global/totally_random_jobs = FALSE
 		src.bioHolder.mobAppearance.customization_first = new /datum/customization_style/none
 		src.bioHolder.mobAppearance.customization_second = new /datum/customization_style/none
 		src.bioHolder.mobAppearance.customization_third = new /datum/customization_style/none
+		src.update_colorful_parts()
 	else if (src.traitHolder && src.traitHolder.hasTrait("loyalist"))
 		trinket = new/obj/item/clothing/head/NTberet(src)
 	else if (src.traitHolder && src.traitHolder.hasTrait("petasusaphilic"))


code/procs/jobprocs.dm
```diff
@@ -680,6 +680,7 @@ var/global/totally_random_jobs = FALSE
 		src.bioHolder.mobAppearance.customization_first = new /datum/customization_style/none
 		src.bioHolder.mobAppearance.customization_second = new /datum/customization_style/none
 		src.bioHolder.mobAppearance.customization_third = new /datum/customization_style/none
+		src.update_colorful_parts()
```
Glamurio, line 683: Might be enough to simply call `src.update_face()` instead?

zjdtmkhzt: Doesn't seem to suffice in my testing, though i dare not speculate why at this point.

zjdtmkhzt: Seems the thing that actually updates it is `H.organHolder.head.UpdateIcon()`, but I feel like maybe I shouldn't be directly calling that?

shrug, devs let me know pls

TobleroneSwordfish (approved):
I think this is the correct proc

ZeWaka (approved):
can always do different if this is a problem

PR #16076, opened 2023-09-28
reagent container openness rework

TheTyrant25:
[CHEMISTRY] [CODE QUALITY] [EXPERIMENTAL] [INTERNAL]
## About the PR
`is_open_container()` previously used a horrendous mess of `TRUE`, `FALSE`, `1`, `0`, `-1`, and `2` to control behaviour. This PR changes that.

Summary of changes:
- redid `is_open_container`'s code to be consistent across different uses. Some overrides of it still exist.
- Gave `is_open_container()` an argument: `inward`. It being set to _TRUE_ means only check if `src` is open for chemicals to be **poured into**. It being set to _FALSE_ means check if it's open to be **poured out** of. This var MUST be used whenever `is_open_container` is used, and for clarity is explicitly used (e.g., `if (src.is_open_container(inward = TRUE))` is how it should be used)
- Redid `OPEN_CONTAINER` flag to be an `rc_flags` and to also dictate which directions of chemical pouring the atom is open in. It's now replaced with mainly `ISOPEN_BOTH`.
- Added `ISOPEN_INWARD` and `ISOPEN_OUTWARD` flags as well, for things that only allow chems to flow one way. `ISOPEN_BOTH` is just the same as `ISOPEN_INWARD | ISOPEN_OUTWARD`
- redid `rc_flags` to be on atom, not /obj/item, so that they can be used in more places. `rc` stands for reagent container, apparently.
- Moved `NOSPLASH` and `ACCEPTS_MOUSEDROP_REAGENTS` to `rc_flags`.
- `set_open_container()` now has one additional arg, which is for closing it in only a certain way (inward or outward). By default, if `which_ways_to_change` is `FALSE` it changes both settings.

## Why's this needed?
There was no consistency in having magic numbers behave sometimes in one way and sometimes in another. Cleaner code is better to work with.

## Changelog
I have included a changelog so that if there are a tonne of bugs that appear during testmerge/merge, people can pinpoint why more easily. I wouldn't have, but as of 15th of December 2023 this pr is `size/XL`, `+451 -333`, and `110 files changed`, so... yeah.
```changelog
(u)Tyrant
(*)Massive internal code change. Report any bugs that you find related to pouring chemicals!
```


TheTyrant25:
cannot find any places where `is_open_container() == 2` was used, which makes it essentially useless if my searching skills aren't rubbish. And the `-1` does seem to have been done haphazardly and with entirely different meanings, but again, i'm not sure. This code is a tad bit confusing. Double checking would be very much appreciated.

TobleroneSwordfish (changes requested):
If you're going to standardize it this way you'll have to remove every call that's just `if(thing.is_open_container())` because you can't guarantee that it'll be open in the right way, also should revert my hacky conditional `open_container` stuff from 3724ac5 and replace it with `ISOPEN_IN_ONLY`

TheTyrant25:
me when the project turns out to be bigger than initially expected 😔 
still, worth doing

TheTyrant25:
right, gonna change the scope

_std/defines/item.dm
```diff
@@ -49,6 +47,20 @@
 /// If an item cannot be crushed by the crusher
 #define UNCRUSHABLE              (1<<23)
 
+// Defines for is_open_container returns and checking. Don't check them directly, use the procs.
+/// stuff can go in
+#define ISOPEN_INWARD			(1<<24)
+//// stuff can come out
+#define ISOPEN_OUTWARD			(1<<25)
```
TobleroneSwordfish, line 54: Yeah no you can't do this, byond only gives us 24 bits to mess with. If you want to do bitflags here you'll have to move it to a separate var.

code/atom.dm
```diff
@@ -238,19 +238,28 @@ TYPEINFO(/atom)
...
+	  * - Takes arguments of whether this openness is for the purpose of pouring something in or out.
+	  * - `TRUE` means you're checking chems go in. `FALSE` means pouring outward. Leaving it null means checking either way.
+	  * - returns `TRUE` if you can access insides, `FALSE` if closed.
+	  * - child procs may override this for special behaviours, such as checking size.
 	  */
-	proc/is_open_container(input = FALSE)
-		return flags & OPENCONTAINER
+	proc/is_open_container(inward)
+		. = FALSE
+		if (isnull(inward) && (src.flags & ISOPEN_BOTH))
```
TobleroneSwordfish, line 249: Could maybe do away with this default case? It's kind of confusing and should be mostly unused

TobleroneSwordfish (changes requested):
Apologies for late review, didn't realise it was ready!

_std/defines/item.dm
TheTyrant25: ah i didn't know this

TheTyrant25: i've dropped it to 23 and 24... hopefully that still works

code/atom.dm
TheTyrant25: i've made it just crash if used without overriding

TobleroneSwordfish (changes requested):
L + 15 runtimes generated (the checks are failing)

TheTyrant25:
GAH

TheTyrant25:
okay so bit of a problem, bottles are being handled as open at the moment, since you can pour them out. however
this means they offgass cyanide. Two solutions:
1. we let the cyanide just offgas off
2. make bottles have lids

TheTyrant25:
my god that's a lot of conflicts

TheTyrant25:
ready

pali6:
Instead of introducing a new bitfield why not just move `var/rc_flags` up to `/atom` and add these there too? That already stands for Reagent Container flags (which tbh took me embarassingly long to realize).

code/WorkInProgress/AzrunStuff.dm
```diff
@@ -922,7 +922,7 @@ ADMIN_INTERACT_PROCS(/turf/unsimulated/floor, proc/sunset, proc/sunrise, proc/se
 	check_match(obj/item/craftingitem)
 		if(..() && length(reagents_req))
 			var/obj/item/reagent_containers/RC = craftingitem
-			if(RC.is_open_container())
+			if(RC.is_open_container(FALSE))
```
pali6, line 925: It's not something I really insist on to merge this but I think that for boolean arguments using named arguments helps with readability a lot. This way when I just read this line I don't know what the argument means. And even if I vaguely do remember its semantics... was it `inward` or `outward` that's true?

code/WorkInProgress/AbilityItem.dm
```diff
@@ -854,7 +854,8 @@
 	layer = HUD_LAYER
 	plane = PLANE_HUD
 	anchored = ANCHORED
-	flags = NOSPLASH
+	flags = 0
```
pali6, line 857:
```suggestion
```
I'd just remove the `flags` overrides from where you remove the NOSPLASH. I know that this is technically changing the behaviour as default flags are FPRINT (a mistake imo, we should have had a flag for NO_FPRINT instead) but I guarantee you that 99% of cases of removing the FPRINT flag from things are accidental.

code/atom.dm
```diff
@@ -241,21 +242,25 @@ TYPEINFO(/atom)
...
+	proc/set_open_container(open, which_ways_to_change = ISOPEN_BOTH)
+		CRASH("You're meant to override set_open_container()! The culprit: [src]")
+		/*if (open)
+			ADD_FLAG(src.flags, which_ways_to_change)
 		else
-			REMOVE_FLAG(src.flags, OPENCONTAINER)
+			REMOVE_FLAG(src.flags, which_ways_to_change)
 		src.reagents?.handle_reactions()
-
+		*/
```
pali6, line 263: Just delete this.

code/WorkInProgress/AbilityItem.dm
TheTyrant25: gotcha

code/atom.dm
```diff
@@ -241,21 +242,25 @@ TYPEINFO(/atom)
...
 	  */
-	proc/is_open_container(input = FALSE)
-		return flags & OPENCONTAINER
-
-	/// Set a container to be open or closed and handle chemistry reactions that might happen as a result
-	proc/set_open_container(value)
-		if (value)
-			ADD_FLAG(src.flags, OPENCONTAINER)
+	proc/is_open_container(inward)
+		CRASH("You're meant to override is_open_container()! The culprit: [src]")
```
pali6, line 252: Ok I don't quite get why this CRASH-es by default. I would expect the default behaviour to use `chem_flags`? What's the point of having `chem_flags` if they do not actually affect this proc? Is it overriden to use `chem_flags` somewhere? If so I don't see where but also there are 130 files to go through so

_std/defines/item.dm
TheTyrant25: moved it to a new var

code/atom.dm
TheTyrant25: holdover from a previous comment from @TobleroneSwordfish 
image
I could make the default case work using those flags now.

code/WorkInProgress/AzrunStuff.dm
TheTyrant25: all instances of inward being used now declare the argument for readability

code/atom.dm
TheTyrant25: I've made the default case check flags now.

TheTyrant25:
fuck
i fixed the merge wrong

.vscode/settings.json
```diff
@@ -10,7 +10,7 @@
 		"tgui/.pnp.*": true
 	},
 	"editor.codeActionsOnSave": {
-		"source.fixAll.eslint": true
+		"source.fixAll.eslint": "explicit"
 	},
```
TobleroneSwordfish, line 14: this doesn't look intentional

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -720,7 +721,7 @@ proc/ui_describe_reagents(atom/A)
 	incompatible_with_chem_dispensers = TRUE //could maybe be ok? idk
 	can_recycle = FALSE //made of glass, but would be a waste and almost certainly accidental so no
 	splash_all_contents = FALSE
-	object_flags = FPRINT | OPENCONTAINER | SUPPRESSATTACK
+	flags = FPRINT | SUPPRESSATTACK
 	initial_volume = 100
```
TobleroneSwordfish, line 725: seem to be missing the `rc_flags` conversion?

code/modules/chemistry/tools/iv_drips.dm
```diff
@@ -67,8 +67,8 @@
 			src.UpdateOverlays(null, "inj_dr")
 		signal_event("icon_updated")
 
-	is_open_container()
-		return 1
+	is_open_container(inward)
+		return TRUE
```
TobleroneSwordfish, line 71: these should be replaced with the inward flag

TobleroneSwordfish (changes requested):

.vscode/settings.json
TheTyrant25: ughh vscode keeps doing this automatically and i have to keep uncommitting it.

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -517,9 +518,9 @@ proc/ui_describe_reagents(atom/A)
 		qdel(src)
 		return TRUE
 
-	is_open_container()
-		if(..() && !GET_ATOM_PROPERTY(src, PROP_ITEM_IN_CHEM_DISPENSER))
-			return 1
+	is_open_container(inward)
+		if(!GET_ATOM_PROPERTY(src, PROP_ITEM_IN_CHEM_DISPENSER))
+			return TRUE
```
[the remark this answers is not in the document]
TheTyrant25: made it so that it returns false if in chem dispenser, and otherwise just calls parent

code/modules/chemistry/Chemistry-Tools.dm
TheTyrant25: whoops

code/modules/chemistry/tools/iv_drips.dm
TheTyrant25: i've changed its function slightly, since the original code was just it always returning true. Now it checks `src.slashed`. Stuff can go in when poured in, but can only pour out if slashed.

TheTyrant25:
we fucked up

PR #16082, opened 2023-09-28
Adds pressure visualization goggles

TobleroneSwordfish:
[FEATURE] [QOL]
## About the PR 
When equipped, these goggles periodically scan the surrounding air pressure and display it as an overlay on each turf. The lower the pressure the more transparent the overlay, with orange colours representing areas of >1 atmospheric pressure.
They can be toggled on and off like mesons and spawn in engineering lockers.
image

## Why's this needed? 
Repressurizing things is annoying, often it's hard to tell where the air is going or what areas have pressure.
This is part 3/4ish of me making repressurizing more bearable.

## Changelog 

```changelog
(u)LeahTheTech
(*)Added pressure visualization goggles which show surrounding atmospheric pressure when worn, found in engineering lockers.
```


TobleroneSwordfish:
oh I forgot to find-replace maps, will do that after dinner

pgmzeta:
Maybe scope-creep / for a follow-up PR, but: Would it be possible to add this to the engineering specialist fab? There are only 3-4 engineering lockers for 7-8 engineer slots. Might make sense in the breach response cart, as well?

PR #16096, opened 2023-09-29
Adds Pressure Goggles to atmos tech and science fabricator

cheekybrdy:
## About the PR 
Adds atmos goggles to the atmos technician locker, their job spawn items and the science fabricator

## Why's this needed? 
While a gimmick job, the atmospheric technician's role would be greatly enchanced by pressure goggles, which would be officially integrated into their kit. Also adds the to the science fabricator for usage in toxins.


TobleroneSwordfish:
sure!

PR #16117, opened 2023-09-29
hopefully fixed clothing booth previews from using the movement state

DisturbHerb:
[QOL]

## About the PR 

title. i noticed a few lag spikes upon selecting items, but i don't know if this was an issue caused by these changes and i'm too lazy to pull up tracy.

## Why's this needed? 

broken interface window bad

look at this shit


TobleroneSwordfish (approved):
`get_icon` already called `getFlatIcon` so there shouldn't be any extra load there (hopefully)

PR #16125, opened 2023-09-30
Fixes ai being able to unglue things from people

### LINKED ISSUE #16011: AI Eye Can Unglue Things From People?
### Labels

[BUG][MINOR]

### Description

I clicked on something glued to someone and "Cool AI manages to unglue the message sign component from Barrel Punchbeef."

### Steps to reproduce

1.Find someone with a glue item
2.Click on it as an AI Eye
3.Become confused as an action bar appears
4.Unglue the item when it finishes

### Expected Behavior

AI eyes shouldn't be able to do this

Reported by: Monoblaster
Round log date: 2023-09-25-06-01
Reported on: 2023-09-25 06:25:42
Map: OSHAN

unfunnyperson:
## About the PR 
Stops the ai from being able to unglue items from things

## Why's this needed? 
Fixes #16011

### DIFF
--- a/code/datums/components/glued.dm
+++ b/code/datums/components/glued.dm
@@ -76,7 +76,7 @@ TYPEINFO(/datum/component/glued)
 	qdel(src)
 
 /datum/component/glued/proc/on_attackhand(atom/movable/parent, mob/user)
-	if(user?.a_intent == INTENT_HELP && !isintangible(user))
+	if(user?.a_intent == INTENT_HELP && !isintangible(user) && can_act(user) && can_reach(user, parent))
 		src.start_ungluing(parent, user)
 	else
 		src.glued_to.Attackhand(user)
--- a/code/datums/components/glued.dm
+++ b/code/datums/components/glued.dm
@@ -76,7 +76,7 @@ TYPEINFO(/datum/component/glued)
 	qdel(src)
 
 /datum/component/glued/proc/on_attackhand(atom/movable/parent, mob/user)
-	if(user?.a_intent == INTENT_HELP)
+	if(user?.a_intent == INTENT_HELP && !isintangible(user))
 		src.start_ungluing(parent, user)
 	else
 		src.glued_to.Attackhand(user)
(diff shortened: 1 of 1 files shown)


code/datums/components/glued.dm
```diff
@@ -76,7 +76,7 @@ TYPEINFO(/datum/component/glued)
 	qdel(src)
 
 /datum/component/glued/proc/on_attackhand(atom/movable/parent, mob/user)
-	if(user?.a_intent == INTENT_HELP)
+	if(user?.a_intent == INTENT_HELP && !isintangible(user))
```
ZeWaka, line 79: This should be can_act / the other physical interaction check

ZeWaka (changes requested):

ZeWaka (approved):

ZeWaka:
> can_act and can_reach are unnecessary here, attackhand should not happen in the first place if that is the case

why are intangibles even getting through then

pali6:
Because some stuff implements `ai_attack` as just a call to `AttackHand` (when the correct thing would be for both to call a third, new proc instead).

code/datums/components/glued.dm
```diff
@@ -76,7 +76,7 @@ TYPEINFO(/datum/component/glued)
 	qdel(src)
 
 /datum/component/glued/proc/on_attackhand(atom/movable/parent, mob/user)
-	if(user?.a_intent == INTENT_HELP)
+	if(user?.a_intent == INTENT_HELP && !isintangible(user) && can_act(user) && can_reach(user, parent))
```
ZeWaka, line 79:
```suggestion
	if(user?.a_intent == INTENT_HELP && !isintangible(user))
```
okay for now

ZeWaka (approved):

ZeWaka:
so what would you recommend here @pali6

pali6:
I don't really care either way. It's not like more checks hurt anything (most likely), they're just redundant

PR #16128, opened 2023-09-30
crematoriums open body bags on ignition

### LINKED ISSUE #15806: being in a body bag lets you survive cremation
### Labels

[BUG][MINOR]

### Description

being cremated while inside of a closed body bag doesn't cremate you

### Steps to reproduce

1. be in a closed body bag
2. get cremated
3. live unharmed

### Expected Behavior

cremation

probably applies to all other things like lockers

pgmzeta:
## About the PR 

Add special body bag handling to crematoriums 
Alters body-bags to eject to their container instead of the turf itself on cut_open

## Why's this needed? 
Fixes #15806

### DIFF
--- a/code/modules/medical/surgery_tools.dm
+++ b/code/modules/medical/surgery_tools.dm
@@ -1050,10 +1050,10 @@ TYPEINFO(/obj/machinery/defib_mount)
 
 	proc/cut_open()
 		for (var/obj/O in src)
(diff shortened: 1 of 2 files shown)


UrsulaMejor:
Are there other instances where setting location to src.loc might not be ideal?

Should the crematorium instead dump the contents into itself manually rather than handling it on the bag?

PR #16162, opened 2023-10-01
Supply console can print trader/requisition barcodes directly

pgmzeta:
[feature][game objects]
## About the PR 

In the cargo supply console, add links to easily print a single barcode for traders and requisitions. These new print options share a 2 second cooldown on the machine, which also includes the existing requisition print list option.

The barcode machine has bulk print and wage-split tagging as unique features; this does not intend to replace it.

## Why's this needed? 

Easier for new cargo players to start working on requisitions and trades. Less trouble with duplicate name definitions in requisitions.

### DIFF
--- a/code/obj/machinery/computer/QM_supply.dm
+++ b/code/obj/machinery/computer/QM_supply.dm
@@ -1252,7 +1252,7 @@ var/global/datum/cdc_contact_controller/QM_CDC = new()
 					</li>
 					"}
 
-			bottomText += {"</ul>"}
+			bottomText += {"</ul><br><em>To sell goods to this trader, print a barcode for <strong>[T.name]</strong>, attach it to a crate containing the goods, and send the crate out the 'sell' mass driver.</em>"}
 
 		if ("selling")
 			bottomText += "<h3>Goods For Sale</h3><ul class='shoplist'>"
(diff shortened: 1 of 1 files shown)


frawhst (approved):

code/obj/machinery/computer/QM_supply.dm
```diff
@@ -1229,13 +1252,7 @@ var/global/datum/cdc_contact_controller/QM_CDC = new()
 					</li>
 					"}
 
-			bottomText += {"
-				</ul>
-				<br>
-				<br><em>To sell goods to this trader, print a barcode for <strong>[T.name]</strong> on the barcode computer, attach it to a crate containing the goods, and send the crate out the 'sell' mass driver.
-				<br>
-				<br>Load no more than 50 items into a crate at once, or the trader's cargo computer may not be able to keep up!</em>
```
ZeWaka, line 1237: why rempove?

pgmzeta: the 50-item limit doesn't exist; re-added the note on how to sell

pgmzeta:
I had some other ideas on implementing this, but went with a solution that keeps the change local to this machine to avoid adding too much complexity.

Let me know if one of these other options would be preferred, and I can work on that in a separate PR:
* Same option exists in the UI, but the barcode instead prints at the nearest QM barcode computer
* New 'requisition' paper type that has a sticker you can peel off, which could be used to help usability around special order requisitions as well.

Issue #15960, opened 2023-09-23
Clothing looks wrong on cyborgs


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

I just equiped a purple witch hat, and the hat, instead of being on the head, was on the body. Unequipping and equipping it again didnt fix it

### Steps to reproduce

1-equipt a purple witch hair, unsure if what other clothes makes it looks like this
2- style ruined

### Expected Behavior

The hat should be on the head, not on the body

### Additional Information & Screenshots

Reported by: Bilo216
On server: Goonstation 3 Roleplay: Morty
Active test merges: #15458, #15759, and #15833
Round log date: 2023-09-23-01-55
Reported on: 2023-09-23 02:21:39
Map: COGMAP

Retrino:
was the hat ever equipped by a monkey?

UrsulaMejor:
Verfied: It breaks when the hat was previously worn by a monkey

UrsulaMejor:
also breaks for other hats

PR #16164, opened 2023-10-01
Sets wear_image pixel offsets to initials in case they've changed

### LINKED ISSUE #8583: Clown masks aren't aligned on monkeys
### Labels

[TRIVIAL]

### Description

Trying to wear a clown mask as a monkey ends up looking.. off

### Steps to reproduce

1.Monke
2.Find a clown mask
3.Put it on

### Expected Behavior

It looks normal, I turned it bald and put it on first before toggling and putting it on again. The bald mask looks fine.

Reported by: Danger Noodles
Round log date: 2022-05-14-14-14
Reported on: 2022-05-14 14:42:46

UrsulaMejor:
## About the PR 
- if no_offset = TRUE on update icons, resets possible existing offsets to initials

## Why's this needed? 

- fixes #8583


pgmzeta:
could i use the same approach on cyborg clothing for #15960?

Sovexe:
Could potentially cause minor offset errors if there's clothing out there whose initial offsets are not OK to use due to changes in New or something...but my quick search of the code did not turn up any clothing that does so. If we do discover any though will just need to note it to fix

PR #16172, opened 2023-10-01
(Ocean) Water is now Wet

### LINKED ISSUE #9013: Ocean doesn't wet loofah
### Labels

[MINOR]

### Description

Putting the loofah into the middle of the ocean such as on Oshan doesn't fill it up with water like you'd expect.

### Steps to reproduce

1. Get a Loofah
2. Put it in the ocean
3. Cry because you have a dry loofah in the ocean

### Expected Behavior

The loofah to fill up with water

Reported by: Diamond_Plated
Round log date: 2022-06-07-09-08
Reported on: 2022-06-07 09:21:04

UrsulaMejor:
## About the PR 

- water is now wet

## Why's this needed? 

- fixes #9013

### DIFF
--- a/code/obj/item/mops_cleaners.dm
+++ b/code/obj/item/mops_cleaners.dm
@@ -557,11 +557,14 @@ TRASH BAG
 
 /obj/item/sponge/process()
 	if (!src.reagents) return
-	if (!istype(src.loc,/turf/simulated/floor)) return
+	if (!istype(src.loc,/turf/simulated/floor) && !istype(src.loc,/turf/space/fluid) ) return
 	if (src.reagents && src.reagents.total_volume >= src.reagents.maximum_volume) return
 	var/turf/simulated/floor/T = src.loc
 	if (T.active_liquid && T.active_liquid.group)
 		T.active_liquid.group.drain(T.active_liquid,1,src)
+	var/turf/space/fluid/F = src.loc
+	if(istype(F))
+		src.reagents.add_reagent(ocean_reagent_id,10)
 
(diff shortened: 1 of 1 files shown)


code/obj/item/mops_cleaners.dm
```diff
@@ -557,11 +557,14 @@ TRASH BAG
...
 	if (!src.reagents) return
-	if (!istype(src.loc,/turf/simulated/floor)) return
+	if (!istype(src.loc,/turf/simulated/floor) && !istype(src.loc,/turf/space/fluid) ) return
 	if (src.reagents && src.reagents.total_volume >= src.reagents.maximum_volume) return
 	var/turf/simulated/floor/T = src.loc
 	if (T.active_liquid && T.active_liquid.group)
 		T.active_liquid.group.drain(T.active_liquid,1,src)
+	var/turf/space/fluid/F = src.loc
+	if(istype(F))
+		src.reagents.add_reagent(ocean_reagent_id,10)
```
Sovexe, line 567: * Kind of weird that these didnt allow unsimulated floors so added that...
* no need to store the fluid turf in a var since we aren't getting the ocean reagent from it but instead of going to the global var
* Existing code had a redundant check of reagents
```suggestion
	if (src.reagents.total_volume >= src.reagents.maximum_volume) return
	if (isfloor(src.loc))
		var/turf/T = src.loc
		if (T.active_liquid && T.active_liquid.group)
			T.active_liquid.group.drain(T.active_liquid,1,src)
	else if (istype(src.loc, turf/space/fluid))
		src.reagents.add_reagent(ocean_reagent_id,10)
```

PR #16181, opened 2023-10-01
Makes Rad and Wasp crossbow's cells unswappable

TheColorCyan:
[GAME OBJECTS] [BALANCE] 

## About the PR 
Makes you unable to swap radiation and wasp crossbow's powercells

## Why's this needed? 
You can put extra fast recharging 100pu cell from sec vendor in it (which is actually pretty easy to get) and just spam the crossbows. Not nice considering the damage wasp and rad crossbows can do. 
And a lot of people assumed it was unswappable anyway....

## Changelog 
```changelog
(u)Sian
(+)Radiation and Wasp crossbow's powercells are now unswappable.
```

### DIFF
--- a/code/obj/item/gun/energy.dm
+++ b/code/obj/item/gun/energy.dm
@@ -319,7 +319,7 @@ TYPEINFO(/obj/item/gun/energy/crossbow)
 	is_syndicate = 1
 	silenced = 1 // No conspicuous text messages, please (Convair880).
 	hide_attack = ATTACK_FULLY_HIDDEN
-	custom_cell_max_capacity = 100 // Those self-charging ten-shot radbows were a bit overpowered (Convair880)
+	can_swap_cell = FALSE
 	muzzle_flash = null
 	uses_charge_overlay = TRUE
 	charge_icon_state = "crossbow"
@@ -1589,7 +1589,7 @@ TYPEINFO(/obj/item/gun/energy/wasp)
 	projectiles = null
 	is_syndicate = 1
 	silenced = 1
-	custom_cell_max_capacity = 100
+	can_swap_cell = FALSE
 
 	New()
 		set_current_projectile(new/datum/projectile/special/spreader/quadwasp)


NightmareChamillian:
While I agree with the nerf, I don't necessarily think this is the best way to go about it. If a traitor wants to risk detection, or downs a secoff, I think earning an upgrade is fair, and the whole point of energy weapons is the ability to swap cells. 

I think instead you should up the charge consumption of the weapons and adjust their cells appropriately.

flappybatpal:
Not 100% sure. You are spending 6TC at least on a crossbow and while the upgrade is a pretty large buff (two shots and faster recharge) you do have to manage to get hold of it, which is rare.

TheColorCyan:
> Not 100% sure. You are spending 6TC at least on a crossbow and while the upgrade is a pretty large buff (two shots and faster recharge) you do have to manage to get hold of it, which is rare.

i wouldnt say getting the upgrade is hard unless there is AI with armory viewport (armory literally gives you tools to break the locker thanks armory), and people started abusing the epic wasp spam more. I think it's fine with default powercell it goes with if you buy it like that. And people use it dual wield usually which makes things worse. I'd say it would be fine to let the cells be swappable if wasps were still critters, but they are much tankier and dangerous now.

CheffieGithub:
Why nerf the regular crossbow as well?
A better nerf would be a limit on maxium recharge rate of a swapped cell OR reducing the maximum capacity of cell that you can add.
A specific nerf to the wasp bow is reducing the damage that the spawned eggs do directly as pointblanking can easily put someone into critical.

The bows are already pretty niche and underused and there are much better options, being able to make a bit of an upgrade is nice and not that strong. 
But the main issue is with the disruptor cell being the perfect charge and also having amazing recharge rate so it would be better to nerf that side rather than remove a feature of these weapons entirely.

unfunnyperson:
alternatively you could make the shots take 50 pu each and make the starting cell 50 pu max, 25% less wasps is a pretty impactful change

Tarmunora:
Did some master-pushes that supersede this

PR #16183, opened 2023-10-01
Bit of a camera refactor

CheffieGithub:
## About the PR 
Moves camera related things to the camera module.
Adds monitors for most networks.
Adds subtypes for common network varedits.

Currently cameras are colour coded like the ranch ones are so its easy to tell at a glance what network and monitor you need but this can be changed if its too much:

image

Standardisation of networks:
Television cameras are on the "public" network instead of "Zeta".
VR cameras are on the "VR" network instead of "Zeta".
Telescience has its own "telesci" network for its mobile cameras and camera hats.
The default network for camera monitors is now "public" intead of "SS13" the latter being moved to a security subtype.

These standards have not yet been applied to maps, due to the headache this will probably be I will do it later as this PR is big enough already.

Camera monitors now have a list of networks so you can have a monitor with extra networks.
From this the security monitor how has access to the mining and science outpost camera networks and there is an ADMIN CRIMES MONITOR with access to every network of cameras.
Added documentation to `camera.dm`.
Speeds up camera monitors by removing some unneeded things and using `as anything in`. 
Monitors now force your eyes back to you when the monitor is dropped.

Areas have had their motion camera based code removed for now as they are unused in modern maps and the code could be better.

Setup for cameras has been improved:
autoname can now add a prefix and suffix to the camera in the format of `[prefix] camera - [area.name]` with `var/prefix` and `var/uses_area_name`.
autotag now only sets the tag to the area name, the code for indexing and tallying camera tags is now always run for every camera to prevent issues of inaccessible cameras and removing the need for having cameras edited to be for example `research 1` and `research 2` manually.
autotag is no longer dependant on autoname.

## Why's this needed? 
Currently most camera stuff is varedited which isn't good for maintability. FIxes bugs relating to camera networks and adds documentation so people know what the code does.

## Changelog 

```changelog
(u)Cheffie
(*)Cameras and monitors have had backend changes. Report any camera / monitor issues or oddities! 
(+)Certain roles have had their camera monitors changed to be public network devices only.
```


DisturbHerb:
🍕

code/modules/camera/monitor.dm
```diff
@@ -0,0 +1,99 @@
...
+
+		var/selected_camera = tgui_input_list(user, "Which camera should you change to?", "Camera Selection", sortList(displayed_cameras, /proc/cmp_text_asc))
+
+		if (!selected_camera)
+			user.set_eye(null)
+			return 0
+
+		var/obj/machinery/camera/C = displayed_cameras[selected_camera]
+
+		if ((!user.contents.Find(src) || !can_act(user) || !user.sight_check(1) || !(C.camera_status)) && (!issilicon(user)))
```
ZeWaka, line 39:
```suggestion
		if ((!(src in user.contents) || !can_act(user) || !user.sight_check(1) || !(C.camera_status)) && (!issilicon(user)))
```

code/modules/camera/camera.dm
```diff
@@ -0,0 +1,341 @@
...
+	deconstruct_flags = DECON_SCREWDRIVER | DECON_WELDER | DECON_WIRECUTTERS | DECON_MULTITOOL
+	text = ""
+
+	/// Used by camera monitors to display certain cameras only
+	var/network = "SS13"
+	/// Used by autoname: EX "security camera"
+	var/prefix = "security"
+	/// Used by autoname: EX "camera - west primary hallway"
+	var/uses_area_name = FALSE
+	// These stack: EX "security camera - west primary hallway"
```
ZeWaka, line 15: comment doesn't make sense, is this for the prior thing?

code/modules/camera/camera.dm
```diff
@@ -0,0 +1,341 @@
...
+	var/obj/machinery/camera/c_north = null
+	var/obj/machinery/camera/c_east = null
+	var/obj/machinery/camera/c_west = null
+	var/obj/machinery/camera/c_south = null
+
+	/// Here's a list of cameras pointing to this camera for reprocessing purposes
+	var/list/obj/machinery/camera/referrers = list()
+
+	/// Robust light
+	var/datum/light/point/light
```
ZeWaka, line 43: disposing

code/modules/camera/camera.dm
```diff
@@ -0,0 +1,341 @@
...
+	/// automatically offsets and snaps to perspective walls. Not for televisions or internal cameras.
+	var/sticky = FALSE
+	/// do auto position cameras use the alternate diagonal sprites?
+	var/alternate_sprites = FALSE
+
+	//This camera is a node pointing to the other bunch of cameras nearby for AI movement purposes
+	var/obj/machinery/camera/c_north = null
+	var/obj/machinery/camera/c_east = null
+	var/obj/machinery/camera/c_west = null
+	var/obj/machinery/camera/c_south = null
```
ZeWaka, line 37: disposing

code/modules/camera/camera.dm
```diff
@@ -0,0 +1,341 @@
...
+
+	/// Here's a list of cameras pointing to this camera for reprocessing purposes
+	var/list/obj/machinery/camera/referrers = list()
+
+	/// Robust light
+	var/datum/light/point/light
+
+	/// The cameras viewer, for quickly disconnecting them when needed
+	var/mob/viewer
```
ZeWaka, line 46: disposing

code/modules/camera/camera.dm
```diff
@@ -0,0 +1,341 @@
...
+	Set by having "autotag" anywhere in the c_tag variable.
+	Sets the tag of the camera to the name of the area.
+	All cameras are tallied regardless of this tag to apply a number to them.
+	*/
+
+/obj/machinery/camera/New()
+	..()
+	START_TRACKING
+	var/area/area = get_area(src)
+	//if only these had a common parent...
```
ZeWaka, line 64:
```suggestion
	//TODO: if only these had a common parent...
```

code/modules/camera/camera.dm
```diff
@@ -0,0 +1,341 @@
...
+						boutput(O, "[user] holds a paper up to one of the cameras ...")
+						X.ui_interact(O)
+						logTheThing(LOG_STATION, user, "holds up a paper to a camera at [log_loc(src)], forcing [constructTarget(O,"station")] to read it. <b>Title:</b> [X.name]. <b>Text:</b> [adminscrub(X.info)]")
+
+/obj/machinery/camera/ex_act(severity)
+	if(src.invuln)
+		return
+	else
+		update_coverage() // explosion happened, probably destroyed nearby turfs, better rebuild
+		..(severity)
```
ZeWaka, line 153: why wouldn't this happen in the opposite order, what if it gets destroyed?

code/modules/camera/camera.dm
```diff
@@ -0,0 +1,341 @@
...
+
+/obj/machinery/camera/emp_act()
+	..()
+	if(!src.network)
+		return //avoid stacking emp
+	if(!istype(src, /obj/machinery/camera/television)) //tv cams were getting messed up
+		src.icon_state = "cameraemp"
+	src.network = null                   //Not the best way but it will do. I think.
```
ZeWaka, line 161:
```suggestion
	src.network = null //Not the best way but it will do. I think.
```

code/modules/camera/camera.dm
```diff
@@ -0,0 +1,341 @@
...
+		if(node == vars[N])
+			vars[N] = null
+	node.referrers -= src
+
+/obj/machinery/camera/proc/hasNode(var/obj/machinery/camera/node)
+	if(!istype(node)) return 0
+	. = 0
+	. = (node == c_north) + (node == c_east) + (node == c_south) + (node == c_west)
+
+// here there was an antisemitic joke, commented out, that persisted until february 27 2020. Why the fuck it lasted so many years really puts ones morals into question.
```
ZeWaka, line 219:
```suggestion
```
this can go

code/modules/camera/camera.dm
CheffieGithub: yeah, updated the comment to clarify

ZeWaka (approved):

ZeWaka:
Why not just visible message instead of boutput paper holding

CheffieGithub:
Yeah I thought of changing it to that but would that work for the boutputs to the camera viewer or just the initial one to the user?

pali6 (approved):
Looks good at a glance. I'm ok to merge once emerge conflicts are gone.

PR #16214, opened 2023-10-02
Adds a reliable way to quickly stop a securitron.

MRLinew:
## About the PR 

If you have access, clicking a securitron while on disarm or grab intent will stop it from moving for a few seconds. This could be seen as buff to antagonists with AA, though securitrons rarely go after people with AA.

## Why's this needed? 
Right now, the onclick menu for securitrons is useless, as they just move away before you can use it. The only way to stop them is through the slow PDA app. See this forum thread: https://forum.ss13.co/showthread.php?tid=21854&pid=198161#pid198161

## Changelog 

```changelog
(u)Froggit Dogget
(+) If you have security access, you can click on a securitron using grab or disarm intent to make it temporarily stop moving/chasing its target.
```

### DIFF
--- a/code/modules/robotics/bot/secbot.dm
+++ b/code/modules/robotics/bot/secbot.dm
@@ -288,6 +288,11 @@
 		Status: <A href='?src=\ref[src];power=1'>[src.on ? "On" : "Off"]</A><BR>
 		Behaviour controls are [src.locked ? "locked" : "unlocked"]"}
 
+		if(!src.emagged)
+			if(src.allowed(user))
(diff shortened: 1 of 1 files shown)


Sbmhawk:
Perhaps this could not work for ids set to arrest?

MeggalBozale:
Beepsky only listening to those it deems non-crimers would make sense overall

Tarmunora:
> Oh I didn't see this, uhh probably not needed now that #16260 is merged? Also intent based interaction with machines + further buffing antags with AA isn't great.

PR #16221, opened 2023-10-03
Tiny fix to requisition info text formatting for seeds

Kubius:
[BUG]
## About the PR 

The front-end display of seed requisitions wasn't updated to match the fixed back-end behavior (higher maturation and production stats being better), incorrectly implying that lower is better for those stats. This fixes that.

## Why's this needed? 

Corrects small goof that inadvertently misled people about what some requisitions actually need. ty clover


Kubius:
Checks are flaking out because of rust-g gateway error :screm:

PR #16222, opened 2023-10-03
Ensure recipient of 'Give Item' is alive and not yourself

### LINKED ISSUE #12973: You can accept items while unconscious
### Labels

[BUG][minor]

### Description

The give command apparently works on unconscious people

### Steps to reproduce

1. Have a friend
2. Put one of the two of you under anaesthetic
3. The awake person tries to give something to the other player

### Expected Behavior

For it to fail before the prompt even shows

If it wasn't clear from reproduction steps, I found this bug while using an N2O tank

MeggalBozale:
## About the PR 
Removes the ability to give yourself an item or to try and give someone under anesthetic/who is asleep/who is dead an item.

## Why's this needed? 
Fixes #12973

TODO:

- [x] Remove Right-Click option to give yourself an item as well 
Cant give items to yourself but it still shows up cos verb sadly
- [x] Prevent people lying down/stunned from receiving items as well
(Not planned unless requested, unconscious is fine enough)

### DIFF
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -1390,7 +1390,7 @@
 			L.give_to(src)
 
 /mob/living/proc/give_to(var/mob/living/M)
-	if (!M)
+	if (!M || M == src || M.stat != STAT_ALIVE)
 		return
 
 #ifdef TWITCH_BOT_ALLOWED
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -1384,6 +1384,10 @@
 	set src in view(1)
 	set category = "Local"
 
+	if (usr == src)
+		boutput(usr,"<span class='alert'>You can't give items to yourself!</span>")
+		return
+
 	SPAWN(0.7 SECONDS) //secret spawn delay, so you can't spam this during combat for a free "stun"
 		if (usr && isliving(usr) && !issilicon(usr) && BOUNDS_DIST(src, usr) == 0)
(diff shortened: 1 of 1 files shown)


UrsulaMejor:
Use isalive() instead of checking stat directly

ZeWaka (approved):

UrsulaMejor:
Not seeing ` Prevent people lying down/stunned from receiving items as well`? Am I missing something?

MeggalBozale:
Oh right yeah uhhhh I would have to check for if someone is horizontal

Is there a nice way to do that? Islying or something? Would that count for being stunned?

Currently checking isalive(M) means people under anesthetic won't work, but if you're just lying down you probably can

UrsulaMejor:
isincapacitated(X)

UrsulaMejor:
I think letting lying down (not stunned) people accept items would be fine. It's funny to imagine someone taking a thing and then just dropping it lazily

MeggalBozale:
In that case no changes would be needed as such is the current behavior

ZeWaka:
> In that case no changes would be needed as such is the current behavior

i don't see that being called tho?

MeggalBozale:
> > In that case no changes would be needed as such is the current behavior
> 
> i don't see that being called tho?

Apologies for the vagueness, I meant it in reply to the previous comment that being able to hand items to the incapacitated is the current behavior.

PR #16232, opened 2023-10-03
Flock Absorber Calls .dropped()

### LINKED ISSUE #10190: Items falling out of Flockdrone reclaimer when they die appear animated
### Labels

[BUG]

### Description

When a Flockdrone dies, any item in it being reclaimed will fall out appearing animated

### Steps to reproduce

1. Kill Flockdrone reclaiming item

### Expected Behavior

For the item to not be animated

_No response_

UrsulaMejor:
## About the PR 

- some flock item drop stuff to prevent sprite glitches

## Why's this needed? 

- fixes #10190

### DIFF
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -836,6 +836,9 @@
 	if(core)
 		core.resources = src.resources
 		src.pay_resources(src.resources) // just in case any weirdness happens let's pre-empt the dupe bug
+
+	if(absorber.item)
+		absorber.drop(absorber.item)
 	..()
 	src.icon_state = "drone-dead"
 	src.set_density(FALSE)
@@ -1300,6 +1303,8 @@
 		animate(item)
 		if(item.material)
 			item.setMaterialAppearance(item.material)
+		item.UpdateIcon()
+		item.dropped(holder)
 	..()
 
 /datum/equipmentHolder/flockAbsorption/proc/tick(mult)


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -1300,6 +1303,8 @@
 		animate(item)
 		if(item.material)
 			item.setMaterialAppearance(item.material)
+		item.UpdateIcon()
+		item.dropped(holder)
```
ZeWaka, line 1307: this seems a bit strange to call directly

UrsulaMejor: where would be better to call it?

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -836,6 +836,9 @@
 	if(core)
 		core.resources = src.resources
 		src.pay_resources(src.resources) // just in case any weirdness happens let's pre-empt the dupe bug
+
+	if(absorber.item)
+		absorber.drop(absorber.item)
```
TobleroneSwordfish, line 841: maybe call `src.u_equip` here instead of manually calling `dropped` in the limb code? It should handle critter limbs fine

TobleroneSwordfish (changes requested):

code/mob/living/critter/flock/flockdrone.dm
UrsulaMejor: It's not a limb, is the thing

TobleroneSwordfish: I said this in discord, but it's an equipment slot which should also be handled fine

PR #16233, opened 2023-10-03
Mail chute funnelling

TheTyrant25:
[MAPPING] [QOL]
## About the PR
This PR finds all(?) of the mail chutes that are close to other pipe chutes and either moves the exit chute or places a thindow. This is to prevent mail from immediately going into the disposal chute and being missed by the recipient.

## Why's this needed? 
Missing out on mail just because a mail chute and disposal chute were put next to each other shouldn't really be a thing. It's been reported in discord as an issue as well.

## Changelog
```changelog
(u)Tyrant
(+)New barriers have been placed to stop mail from going straight into disposal chutes.
```

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -437,6 +437,7 @@
 /obj/disposalpipe/trunk/mail{
 	dir = 8
 	},
+/obj/window/reinforced/west,
 /turf/simulated/floor/grass/random/alt,
 /area/station/garden/aviary)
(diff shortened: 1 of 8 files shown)


TheTyrant25:
should I do "input wanted" on this? Some people might consider it part of the fun or intentionally placed that way etc.

TheTyrant25:
A much better solution was presented. Closing this pr now. Thindows bad etc

PR #16236, opened 2023-10-03
Set a cooldown for things launched out of pipes.

TheTyrant25:
[INTERNAL]? [GAME OBJECTS] [QOL] [EXPERIMENTAL]?
## About the PR
In multiple maps, disposal chutes and mail chutes are next to each other, leading to mail getting stuck in adjacent chutes. This PR adds a cooldown called `PipeEject` which lasts two seconds and prevents things falling into disposal chutes.

This means that things shot out of mail chutes (and other outlets) can't immediately enter chutes by being thrown (including, by the outlet itself). Clickdrag and attackby to put things in should still work though.

As a consequence, things will be able to be thrown into mail chutes once again. This was disabled previously in order to prevent mail falling into the chute it just exited. This will no longer happen, and so the behaviour can return.

Edit: Stuff flung out of vending machines will also now not fall into the chutes.

## Why's this needed? 
People losing mail is bad. Given there are so many disposalchute-mailchute spots in current maps, we should do something about this.


TheTyrant25:
wasn't sure if i should add a changelog so i omitted it

TheTyrant25:
todo: use the ON_COOLDOWN macro instead

pgmzeta:
also [vending.dm](https://github.com/goonstation/goonstation/blob/master/code/modules/vending/vending.dm#L2152) maybe

ZeWaka:
why isn't this implemented with a cooldown

TheTyrant25:
Sorry!!!!

ZeWaka (approved):

ZeWaka:
.

TealSeer (approved):

imawxome:
I remember this happening on the Oshun sci transport tubes, i got stuck getting launched from one tube into the other over and over "You fall into the chute . . . Accidently hitting the handle!" "You fall into the chute . . . Accidently hitting the handle!" "You fall into the chute . . . Accidently hitting the handle!" "You fall into the chute . . . Accidently hitting the handle!"

PR #16243, opened 2023-10-03
Removes locker from the middle of Donut 2 hallway

Chatauscours:
## About the PR 

Снимок экрана (159)

This goes bye.

## Why's this needed? 
Bug


CheffieGithub:
theres a missing pipe there too

pgmzeta:
> theres a missing pipe there too, or is that the pipe being perfectly covered?

pipe is there https://github.com/goonstation/goonstation/pull/16243/checks?check_run_id=17365822691

PR #16244, opened 2023-10-03
Use regular speech bubble for whispers

### LINKED ISSUE #16140: Incorrect speech bubbles for whispering
### Labels

[BUG]

### Description

When you whisper, the little speech bubble that comes out when you say something appears incorrectly.

### Steps to reproduce

1. Wait for someone to sing
2. Whisper
3. Your speech bubble comes out as a musical note despite you not singing.

### Expected Behavior

Whispering to use its own bubble

As far as I can tell, it uses the bubble for the last message said on the server. I was able to notice it using both versions of the musical note bubble and the exclamation mark.

pgmzeta:
## About the PR 

Whispering never sets a chat bubble icon state, so it used whatever the last one was. This explicitly sets the "speech" icon state when whispering.

## Why's this needed? 
Fixes #16140

### DIFF
--- a/code/mob/living/carbon/human.dm
+++ b/code/mob/living/carbon/human.dm
@@ -1660,6 +1660,7 @@
 		if (isdead(M) && !(M in heard_a) && !istype(M, /mob/dead/target_observer) && !(M?.client?.preferences?.local_deadchat))
 			M.show_message(rendered, 2)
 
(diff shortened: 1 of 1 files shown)


Tarmunora:
might be worth considering having whispering not show a speech bubble going forwards, for purposes of being sneaky

PR #16262, opened 2023-10-04
Change API format query params to correct format for Laravel

Wirewraith:
## About the PR 

Previous method formatted nested lists to a CSV structure. Laravel instead expects array parameters as foo[bar]=val

### DIFF
--- a/code/modules/goonhub/api/routes/apiRoute.dm
+++ b/code/modules/goonhub/api/routes/apiRoute.dm
@@ -19,14 +19,17 @@
 
 
 /// Formats a given parameter associated list into a urlstring format
-/// E.g. `list("ckey"="zewaka") to `?&ckey=zewaka` and `list("x"=list("a", "b"))` to `?&x=a,b`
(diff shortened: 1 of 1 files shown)


code/modules/goonhub/api/routes/apiRoute.dm
```diff
@@ -19,14 +19,17 @@
...
-/// E.g. `list("ckey"="zewaka") to `?&ckey=zewaka` and `list("x"=list("a", "b"))` to `?&x=a,b`
+/// E.g. `list("ckey"="zewaka") to `?&ckey=zewaka` and `list("x"=list("a" = "foo", "b" = "bar"))` to `?&x[a]=foo&x[b]=bar`
 /datum/apiRoute/proc/formatQueryParams()
 	if (length(src.queryParams))
-		var/paramListClone = src.queryParams.Copy()
-		for (var/key in paramListClone)
-			if (islist(paramListClone[key])) // Do we need to encode the value?
-				paramListClone[key] = jointext(paramListClone[key], ",") // lists become csvs by convention
-		. = list2params(.)
+		var/list/ret = list()
```
ZeWaka, line 25: you could use `.` here but whatever

ZeWaka (approved):
ihatethisihatethisihatethisihatethisihatethisihatethisihatethisihatethisihatethis

PR #16264, opened 2023-10-04
Crate Manufacturer Rebalance

CrystalClover:
[BALANCE] [QoL] [GAME OBJECTS]
## About the PR 
Crate Manufacturer starting materials changed from 5 steel to 2 steel and 2 wood
Crates now only require 0.1 material to make, doubling the amount of crates able to be made
Wooden and Packing (cardboard) Crates now require wood material instead of steel

## Why's this needed? 
Crates needn't require so much material to make, especially when they are deconstructed into only 1 sheet.
Transmuting steel into wood/cardboard doesn't make sense
Wood is now easily obtainable to QMs via ordering, so it can be refilled easily

```changelog
(u)CrystalClover
(+)Crates of all sorts are now cheaper to fabricate
```


mordent-goonstation:
Doesn't this quadruple the amount that can be made, given that you can make 20 steel types and 20 wood types? Drop to 1 steel/1 wood at round start and I think it'll be fine. That's 10 steel-type and 10 wood-type ones, and as you say very cheap to refill.

BatElite:
I don't really see the point of the cost change. crate production being a loss of materials is fine and frankly sensible, it's not like either part of the equation is in short supply.

Also we have cardboard material, why not use that in the packing crates directly rather than substitute one poor fit for another?

CrystalClover:
> I don't really see the point of the cost change. crate production being a loss of materials is fine and frankly sensible, it's not like either part of the equation is in short supply.
> 
> Also we have cardboard material, why not use that in the packing crates directly rather than substitute one poor fit for another?

Packing crates use wood since its easier to refill, cardboard is a difficult material to obtain

CrystalClover:
> Doesn't this quadruple the amount that can be made, given that you can make 20 steel types and 20 wood types? Drop to 1 steel/1 wood at round start and I think it'll be fine. That's 10 steel-type and 10 wood-type ones, and as you say very cheap to refill.

This makes sense, i will reduce it to 1

MeggalBozale (approved):
Code LGTM, it's just number and recipe tweaks after all.

I'm all for this change. QM can buy 20 metal bars whenever they want (Or 5 wood) and I don't see why crates, which turn into 1 sheet when wrenched, should cost more than exactly this to make (1 sheet).

mordent-goonstation (approved):
Needs a minor changelog entry, else LGTM

PR #16266, opened 2023-10-04
Make blasters and incapacitor use charge overlays

CheffieGithub:
## About the PR 
Replaces broken incapacitor icons with charge overlays. Changes blaster_X weapons to use charge overlays instead of icons.

## Why's this needed? 
Fixes incapacitor charge icons. Blasters should use what other things use.


Tarmunora:
The incapacitor sprite seems a little off according to the icondiffbot

CheffieGithub:
Icon diff bot is being weird yeah, its fine. 
image

Not only diff render ive seen recently with issues, maybe someone should make a bug report.

PR #16273, opened 2023-10-05
Adds a couple worn sprites for guns

DimWhat:
[game objects]
## About the PR 
Adds a couple bespoke sprites for when you put a gun in your belt slot. This is a slight nerf to guns as it lets people see when you have one on your belt slot but I don't think that matters as the times your put one there are rare.
image

To-do

- [ ] Monkey versions
- [ ] More guns
- [ ] Maybe move some of these to be back slot guns :shelterhands:
- [ ] Better PR description

## Why's this needed? 
When you put a gun on your belt it just goes invisible which is BORING. I want to see that gun just hanging there.

## Changelog 

```changelog
(u)DimWhat
(+)Added worn sprites to several guns
```


DimWhat:
By the way I'm open to suggestions and requests.

PR #16278, opened 2023-10-05
adds isbool()

TheTyrant25:
[INTERNAL] [QOL]
## About the PR
Adds a definition for `isbool(x)`. Puts it in `_std/macros/vars.dm`.

## Why's this needed? 
We have checks for lists, paths, text, numbers, infinity, obj, pointers, movable, null, loc... all built straight into DM. But not `isbool()`. Apparently they expected you to just use truthiness, but what if you need to check it properly and not truthy?

### DIFF
--- a/_std/__std.dme
+++ b/_std/__std.dme
@@ -148,6 +148,7 @@
 #include "macros\transfers.dm"
 #include "macros\turf.dm"
 #include "macros\twitch_sbill.dm"
+#include "macros\vars.dm"
 #include "macros\vehicles.dm"
 #include "address.dm"
 #include "color.dm"
--- a/_std/macros/vars.dm
+++ b/_std/macros/vars.dm
@@ -1,0 +1,5 @@
+// Macros for checking details of vars
+
+// Some are built into dm: islist(), istext(), isnum(), istext(), ispath(). Some are not.
+
+#define isbool(x)	(x == TRUE || x == FALSE)
--- a/goonstation.dme
+++ b/goonstation.dme
@@ -30,6 +30,7 @@ var/datum/preMapLoad/preMapLoad = new
 #include "code\__secret_public.dme"
 
 // BEGIN_INCLUDE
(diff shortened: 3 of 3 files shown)


TheTyrant25:
also should it be like, unticked
i have never really added a file to `_std` before

MeggalBozale:
So this should work like:
```
var/foo = 0
var/bar = FALSE
isbool(foo) // false
isbool(bar) // true
```
? If so LGTM and might make some things easier

BatElite:
> So this should work like:
> 
> ```
> var/foo = 0
> var/bar = FALSE
> isbool(foo) // false
> isbool(bar) // true
> ```
> 
> ? If so LGTM and might make some things easier

`FALSE` is a macro around `0`, so both of those would evaluate true.
--------

So given the dynamic typing thing, what's the usecase for this? Beyond adhering to convention, you can't guarantee something's a boolean and it's trivial for this macro to give false positives.

MeggalBozale:
That's what I thought. This doesn't give much of a use case if that's not the behavior. I had thought that somehow it might.

TheTyrant25:
The usecase isn't for catching the TRUE FALSE defines. The point is for checks to make sure that no one accidentally stick foo = 2 instead of using the binary values intended for a proc argument or whatever. Future proofing and whatnot.

Say you have a proc that accepts 1 and 0, and you want to make sure no one puts any other values in. Obviously you could just redo the logic in the macro on the spot, but it's just a neat little readability thing.

Azrun (approved):

Azrun:
Is this blocking something?

TheTyrant25:
> Is this blocking something?
no, it isn't
it started because i needed it for something but then i realised there was a better way to do the thing that didn't need isbool
I thought it was still a neat idea to have though, and it doesn't really add anything negative.

ZeWaka:
closing because this will probably just confuse people

PR #16298, opened 2023-10-06
Refactor apiBody to support autocompletion

Wirewraith:
## About the PR 

This format allows for working autocompletion for setting body fields


code/modules/goonhub/api/requests/apiBody.dm
```diff
@@ -3,23 +3,35 @@ ABSTRACT_TYPE(/datum/apiBody)
 /// Represents the body of an API request - ABSTRACT
 /// Everything is generally required unless annotated as such
 /datum/apiBody
+	var/list/fields
```
ZeWaka, line 6:
```suggestion
	var/list/fields = list()
```
?

ZeWaka (approved):

PR #16316, opened 2023-10-07
Swing sign

Ozzzim:
[GAME OBJECTS] [SPRITES]

_This is a repost of my [previous pull request](https://github.com/goonstation/goonstation/pull/16106), without broken +secret commits. Sorry for the mess!_

## About the PR 

This fork adds a swing sign (or more accurately a people stopper, though I find that name stupid), a foldable sign that lets you write and display messages. Messages can be then read by examining the sign.

Here's a presentation video:
[www.youtube.com/watch?v=mUk8okGf_Mo](https://www.youtube.com/watch?v=mUk8okGf_Mo)

Video uses outdated sprites, here are the new ones:
image

And a screenshot of the typing interface, since video capture didn't catch it.
Swingsign ui

For now the only way to get the sign is by crafting it with wood. Ideally they should be available to both catering and chapel, with perhaps a spare one in the storage/civilian quarters.

## Why's this needed? 

I feel like there's a lack of ways to advertise your stuff on the station. Your best shot is either drawing on the floor with crayons or leaving a paper which can easily be mistaken for junk or be buried under the clutter. With the swing sign Chefs will be able to show off their menu, Chaplains could write something about their sermons, Clowns could advertise their brand new show etc.

## Changelog 

```changelog
(u)Ozzzim
(*)Added a swing sign and it's closed counterpart
(+)Added new recipe to wood crafting in building_materials.dm
```


Ozzzim:
I've applied all the suggestions of [TobleroneSwordfish](https://github.com/TobleroneSwordfish), except using the tgui_input_text interface instead of my custom one. Interface I'm using has code needed to properly format the message. 
I'll have to do some research into BYOND's string functions and if I find appropriate counterparts I might move the formatting to the swingsign.dm and then I'll be able to use tgui_input_text.

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,182 @@
...
+TYPEINFO(/obj/swingsign)//No idea what TYPEINFO is, I just know it lets me disable the brown overcoat when crafting signs
+	mat_appearances_to_ignore = list("wood")
+/obj/swingsign
+	name = "swing sign"
+	desc = "A foldable sign for writing annoucements or advertisements."
+	icon = 'icons/obj/furniture/swingsign.dmi'
+	icon_state = "blank"
+	throwforce = 10
+	density = 1
+	anchored = UNANCHORED
```
Sovexe, line 18:
```suggestion
```
unanchored is already the default

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,182 @@
...
+	/// Stored message
+	var/message = ""
+	var/defaultdesc = "A foldable sign for writing annoucements or advertisements."
+	/// There's a /div in SetMessage. If changing the preamble remember to respect it.
+	var/descpreamble = "It says:<br><div style='text-align:center'>"
+	var/secured = FALSE
```
Sovexe, line 24: lacks a descriptive comment

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,182 @@
...
+	var/maxmessagecols = 40
+
+	proc/setmessage(var/newmessage)
+		message = newmessage
+		if(message == "")
+			desc = defaultdesc
+			icon_state = "blank"
+		else
+			desc = descpreamble + message + "</div>"
+			icon_state = "written"
```
Sovexe, line 39: icon changes should be handled within `UpdateIcon()` overrides and called via the `update_icon()` wrapper proc

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,182 @@
...
+
+	proc/fold()
+		var/obj/item/swingsignfolded/C = new/obj/item/swingsignfolded(src.loc)
+
+		if (src.material)
+			C.setMaterial(src.material)
+		if (src.icon_state)
+			C.icon_state = "folded"
```
Sovexe, line 47: icon changes should be handled within `UpdateIcon()` overrides and called via the `update_icon()` wrapper proc

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,182 @@
...
+		message = newmessage
+		if(message == "")
+			desc = defaultdesc
+			icon_state = "blank"
+		else
+			desc = descpreamble + message + "</div>"
+			icon_state = "written"
+
+	proc/fold()
+		var/obj/item/swingsignfolded/C = new/obj/item/swingsignfolded(src.loc)
```
Sovexe, line 42: `C` does not follow our normal naming conventions. If you are going to use a single letter var name for this it should at least be `O` people will read `C` as critter

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,182 @@
...
+
+		if (src.material)
+			C.setMaterial(src.material)
+		if (src.icon_state)
+			C.icon_state = "folded"
+		if (src.message)
+			C.message = src.message
+
+		qdel(src)
```
Sovexe, line 51: not a big fan of deleting and spawning replacement items. Loses important metadata for admins as well as for players. Feel like this could be a single type instead of two, or possibly make use of a component similar to `/datum/component/foldable`

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,182 @@
...
+			random_brute_damage(M, rand(1,maxcrashdamage),1)
+			L.force_laydown_standup()
+			L.changeStatus("stunned", 1 SECONDS)
+			src.visible_message("<b><font color=red>[M] is people-stopped by [src]!</font></b>")
+			playsound(src, 'sound/impact_sounds/Wood_Hit_1.ogg',50,1)
+			fold() //Change to item
+			return
+		..()
+
+	custom_suicide = 1
```
Sovexe, line 94: vars belong at the top

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,182 @@
...
+		"maxRows" = src.maxmessagerows,
+		"maxCols" = src.maxmessagecols
+	)
+
+/obj/swingsign/ui_act(action, params)
+	. = ..()
+	if (.)
+		return
+	if(action == "save_message")
+		var/new_message = params["message"]
```
Sovexe, line 123: unsafe, need to `html_encode` the players input to avoid malicious injections of scripts, images, etc

Sovexe (changes requested):
Did a brief review of the dm code. More changes possibly later on...did not review TGUi

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,182 @@
...
+
+	hitby(atom/movable/M, datum/thrown_thing/thr)
+		if (isliving(M) && M.throwing)
+			var/mob/living/L = M
+			var/area/T = get_area(src)
+			if(T?.sanctuary)
+				return
+			random_brute_damage(M, rand(1,maxcrashdamage),1)
+			L.force_laydown_standup()
+			L.changeStatus("stunned", 1 SECONDS)
```
Sovexe, line 87: should be using `do_disorient`

Ozzzim:
Some of these are a handful, but I think I'll come up with something

Sovexe:
> Author

if attempting to come up with a single type solution is a bit too of a challenge at the moment we can skip it for now

code/obj/item/swingsign.dm
Sovexe:
```suggestion
		var/new_message = copytext(html_encode(params["message"]), 1, (maxmessagerows * maxmessagecols))
```

code/obj/item/swingsign.dm
Sovexe:
```suggestion
			L.do_disorient(25, 1 SECOND, stack_stuns = FALSE)
```
personal opinion. Adds on stamina damage, changes from stunned to weakened (aka knocked down), prevents stun stacking

code/obj/item/swingsign.dm
Ozzzim: I was worried this might be tough to implement without breaking the formatting. Turns out there are already BYOND functions that do exactly what I needed (newline_html_encode() from text.dm)

code/obj/item/swingsign.dm
Ozzzim: Tweaked a couple of parameters and kept force_laydown_standup. For some reason mobs thrown at the sign do not trip without it, even though it's already in the do_disorient code.

Ozzzim:
Crap I've did this again

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,197 @@
+#define CHARACTERLIMIT 350
```
Sovexe, line 1: unused

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,197 @@
...
+		"maxRows" = src.maxmessagerows,
+		"maxCols" = src.maxmessagecols
+	)
+
+/obj/swingsign/ui_act(action, params)
+	. = ..()
+	if (.)
+		return
+	if(action == "save_message")
+		var/new_message = newline_html_encode(params["message"])//Encodes the message for safety and replaces all \n with <br> (text.dm)
```
Sovexe, line 138: you still want a copytext here to enforce character limit. The client side TGUI JS is not trustworthy enough to enforce.

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,197 @@
...
+		setmessage(new_message)
+		. = TRUE
+	update_icon()
```
Sovexe, line 141:
```suggestion
	UpdateIcon()
```

Sovexe (changes requested):

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,197 @@
...
+			return
+		else if (isscrewingtool(W))
+			if(secured)
+				boutput(user, "<span class='notice'>You unsecure the swing sign.</span>")
+				anchored = UNANCHORED
+			else
+				boutput(user, "<span class='notice'>You secure the swing sign.</span>")
+				anchored = ANCHORED
+			playsound(src, 'sound/items/Screwdriver.ogg', 50, 1)
+			secured = !secured
```
Sovexe, line 89: do you really need this var? Seems like you could just rely on anchored directly

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,197 @@
...
+
+TYPEINFO(/obj/swingsign)//No idea what TYPEINFO is, I just know it lets me disable the brown overcoat when crafting signs
+	mat_appearances_to_ignore = list("wood")
+/obj/swingsign
+	name = "swing sign"
+	desc = "A foldable sign for writing annoucements or advertisements."
+	icon = 'icons/obj/furniture/swingsign.dmi'
+	icon_state = "blank"
+	throwforce = 10
+	density = 1
```
amylizzle, line 17:
```suggestion
	density = TRUE
```

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,197 @@
...
+	custom_suicide = 1
```
amylizzle, line 18:
```suggestion
	custom_suicide = TRUE
```

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,197 @@
...
+			newSwingsign.setMaterial(src.material)
+		//if (src.icon_state)
+		//	newSwingsign.icon_state = "folded"
+		if (src.message)
+			newSwingsign.message = src.message
+
+		qdel(src)
+
+	attack_hand(mob/user)
+		if (!isliving(user)) return
```
amylizzle, line 68: probably want more protections than just this. A lot of things are living but shouldn't be able to do this - critters for example.

amylizzle (changes requested):

Ozzzim:
Okay that's the new batch of change requests done I believe. I've also added the updated holding sprites.

> Sanity check on this needing a custom "SwingSign" interface instead of a textarea-like standard input (akin to how we have standard text input ones now)?

I got that suggestion in the previous pull request. I made the Swingsign interface to inject newlines upon row overflow and to enforce rows limit while message is still being typed. I could theoretically move both functionalities to the dm, but then input assist would be limited to only maxlength.
That could result in the text getting trimmed without warning or being displayed differently than user expected.
 
To be fair I have very little experience with React (or whatever similar thing tgui is based on) and I don't have a decent grasp on it's capabilities. If it's possible to extend the standard text input to implement those features, I'd be very much open to that.

Now that I think about it I should copy latter functionality into the dm, or someone could in theory just manually send couple hundred \n's and make the sign a massive pain.

mordent-goonstation:
> Okay that's the new batch of change requests done I believe. I've also added the updated holding sprites.
> 
> > Sanity check on this needing a custom "SwingSign" interface instead of a textarea-like standard input (akin to how we have standard text input ones now)?
> 
> I got that suggestion in the previous pull request. I made the Swingsign interface to inject newlines upon row overflow and to enforce rows limit while message is still being typed. I could theoretically move both functionalities to the dm, but then input assist would be limited to only maxlength. That could result in the text getting trimmed without warning or being displayed differently than user expected.
> 
> To be fair I have very little experience with React (or whatever similar thing tgui is based on) and I don't have a decent grasp on it's capabilities. If it's possible to extend the standard text input to implement those features, I'd be very much open to that.
> 
> Now that I think about it I should copy latter functionality into the dm, or someone could in theory just manually send couple hundred \n's and make the sign a massive pain.
> 
> Perhaps I'm just overthinking the whole thing and text will look just fine if I let people control the newlines and text width

Text-wrapping exists, so I think it would be "fine" if it was done that way; you'd just want to make the editing interface look similar to the display interface, so that people can accurately represent what they want to represent on the sign while editing it. That you'd want to inject newlines accurately when people examine it is certainly a little trickier, but a custom interface here doesn't solve that, really.

Take a look at [TextInputModal](https://github.com/goonstation/goonstation/blob/master/tgui/packages/tgui/interfaces/TextInputModal.tsx). This is the interface that I *think* you could reasonably try to use as a base. You may want to add the functionality to be able to customize the font to it, if monospace font is important for your use-case, as well as functionality to let you lock the width (or a maximum width).

If that doesn't seem like a sensible path for you, for whatever reason, I'm okay with a custom SwingSign interface. I would prefer to explore this option first.

Ozzzim:
Lemme run something with you before I actually commit it. In order to make `/datum/component/foldable` work for me I had to make somewhat radical changes to it. First of all I've added new variable `can_fold = TRUE` so folding process can be stopped if for example swing sign is secured. The other is that I swapped `foldUpIntoBriefcase()` with this:
```
/obj/proc/foldUpIntoBriefcase()
	set category = "Local"
	set name = "Fold up"
	set src in view(1)

	if(foldUp())
		usr.visible_message("<span class='alert'>[usr] folds [object] back up!</span>")

/obj/proc/foldUp()
	if(usr.stat)
		return
	var/atom/movable/object = src
	if(!object)
		return

	var/datum/component/foldable/fold_component = src.GetComponent(/datum/component/foldable)
	if(!fold_component)
		return FALSE
	if(!fold_component.the_briefcase)
		return FALSE//fold_component.the_briefcase = new/obj/item/objBriefcase(get_turf(object), object)
	if(!fold_component.can_fold)
		return FALSE
	var/obj/item/objBriefcase/briefcase = fold_component.the_briefcase

	if(object.loc == briefcase)
		return FALSE

	if(object.loc == usr)
		usr.drop_from_slot(src)

	briefcase.set_loc(get_turf(object))
	src.set_loc(briefcase)
	return TRUE
```
Originally I tried to add a default variable to `foldUpIntoBriefcase(var/mute = TRUE)` so I can just tie visible_message to it, but that broke the right mouse button menu. So now there are two procs with `foldUpIntoBriefcase()` calling `foldUp()` to do most of the work.
All in all, I don't think all those changes should affect already existing foldables (or just a foldable, I think only Portable Identification Computer uses this), but since this is not my code, I thought it might be a good idea to ask first.

Sovexe:
After speaking with other devs don't worry about trying to make use of the folding component. We have to do a larger refactor on it out of scope for this PR for what we are wanting. There are many other objects that we want to tie into a more robust component later down the line

Sovexe:
Seems to have been some confusion here - per my previous comment do not use the foldable component.

It's not in a state where it makes sense for this and is pending refactor. Which makes my next point here moot but, for future reference, you shouldn't be storing references to components on other objects

You can just use the behavior or having two types that spawn each other like you had previously.

Ozzzim:
This is awkward, I did read your comment incorrectly. I'll change back to the original method based on folding chair tomorrow.

Ozzzim:
So there's this problem I've had with the UI ever since I began working on it. In the examine text, I'm having it aligned to the center to make it pop out more, and I'm trying to have the UI editor text do the same thing, but `textAlign: "center"` doesn't work with textareas. People on the internet suggest `text-align-last` property, but that's a somewhat recent CSS feature and it doesn't work with TGUI display system.
Anyone knows any workarounds for this?

Ozzzim:
Where did that review request come from?

Anyway, I wasn't exactly sure what to do, so I did the following. I copied the entire TextInputModal and stuffed in functionality from the SwingSign interface. As a result we now have SwingSignTIM.tsx which looks and works like text input modal, but also does all the formatting I need. I've also changed the row trimming so it doesn't delete excess lines (instead just throws a 'message too long' message). I've also added a trimming functionality to the .dm in case someone tries sending messages manually.

I still got a couple thing I want to do, namely have the text box adjust to the provided message diameters. 
As always please alert me if I messed something up.

Tarmunora (approved):
No issues with light-mode readability

Ozzzim:
I'm scrapping the UI scaling. The "too long" error message gets buried under the buttons when UI gets too narrow.

Aside from cleaning I'm done with coding for now. I'll see if I can add signs to the maps now.

Ozzzim:
Signs can be now found in Chapel lockers as well as strewn throughout catering areas on all major maps

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,220 @@
...
+	var/defaultdesc = "A foldable sign for writing annoucements or advertisements."
+	/// Text inserted before the actual message
+	var/descpreamble = "It says:<br><div style='text-align:center'>"// There's a /div in SetMessage. If changing the preamble remember to respect it.
+	/// Damage when thrown into a swing sign
+	var/maxcrashdamage = 5
+	/// Max length of the message
+	var/maxmessagerows = 10
+	/// Max width of the message
+	var/maxmessagecols = 30
+	//HELP_MESSAGE_OVERRIDE({"Use any writing utensil to write a message. You can also secure a swing sign to the floor with a screwdriver."})
```
TobleroneSwordfish, line 28:
```suggestion
	HELP_MESSAGE_OVERRIDE({"Use any writing utensil to write a message. You can also secure a swing sign to the floor with a <b>screwdriver</b>."})
```

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,220 @@
...
+		qdel(src)
+
+	///Trims the message if it's longer than expected. UI has a similar function so it'll probably never be fully used.
+	proc/overflowControl(var/string)
+		var/stringarr = splittext(string,"\n")
+		if(maxmessagerows>0 && length(stringarr)>maxmessagerows){
+			var/result=stringarr[1]
+			for(var/i in 2 to maxmessagerows){
+				//if(length(stringarr[i])>maxmessagecols)
+				//	stringarr[i] = copytext(stringarr[i],1,maxmessagecols)
```
TobleroneSwordfish, line 66:
```suggestion
```
Don't leave commented code around

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,220 @@
...
+			return result
+		}
+		return string
+
+	attack_hand(mob/user)
+		if (!iscarbon(user)) return
+		if(anchored == UNANCHORED)
+			user.visible_message("<b>[user.name] folds [src].</b>")
+			playsound(src, 'sound/impact_sounds/Clock_slap.ogg',50,1)
+			fold()//
```
TobleroneSwordfish, line 78:
```suggestion
			fold()
```

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,220 @@
...
+	hitby(atom/movable/M, datum/thrown_thing/thr)
+		if (isliving(M) && M.throwing)
+			var/mob/living/L = M
+			var/area/T = get_area(src)
+			if(T?.sanctuary)
+				return
+			random_brute_damage(M, rand(1,maxcrashdamage),1)
+			L.do_disorient(10, 0, 1 SECOND, 0, disorient = 0, stack_stuns = FALSE)
+			L.force_laydown_standup()
+			//L.changeStatus("stunned", 1 SECONDS)
```
TobleroneSwordfish, line 107:
```suggestion
```

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,220 @@
...
+    	rows = src.maxmessagerows,
+    	columns = src.maxmessagecols,
+	)
+/*
+	. = list(
+		"message" = newline_html_decode(src.message),//Decodes the message for the UI component and replaces all <br> with \n (text.dm)
+		"maxRows" = src.maxmessagerows,
+		"maxCols" = src.maxmessagecols
+	)
+*/
```
TobleroneSwordfish, line 147:
```suggestion
```

code/obj/item/swingsign.dm
```diff
@@ -0,0 +1,220 @@
...
+
+	/// Stored message for the deployed object
+	var/message = ""
+
+	New()
+		..()
+		src.setItemSpecial(/datum/item_special/swipe)
+		BLOCK_SETUP(BLOCK_LARGE)
+
+	///*
```
TobleroneSwordfish, line 194:
```suggestion
```

TobleroneSwordfish (changes requested):
This PR has been open for an unacceptable amount of time, sorry, some final nitpicks before I hit merge.

PR #16260, opened 2023-10-04
Securitron whistle commands

TobleroneSwordfish:
[BALANCE] [FEATURE]
## About the PR 
Adds a new type of whistle that all security personnel (assistants included) spawn with. Upon blowing it, all nearby securitrons will stop what they're doing and listen, pointing at a person in the next 3 seconds will command the securitrons to arrest the person, pointing at anything else will command the securitrons to go there. The whistle has a contraband level of 4 so any non-security personnel holding it will be targeted by securitrons.
image

Also fixes securitrons being bricked after summoning and whistles not showing a chat message when used.

Not sure who came up with the original idea, someone mentioned it in admin chat and I thought it would be cool.
## Why's this needed? 
Adds a neat new thing to use to arrest people and helps solve the problem of trying to call off an angry beepsky. Also the whistles can be stolen for crime purposes.

## Changelog 

```changelog
(u)LeahTheTech
(*)Added a security whistle that all security officers and assistants start with. Blowing it will let you command nearby securitrons by pointing.
```

PR #16317, opened 2023-10-07
Prevents active secbots from being pulled and makes them unable to scan while not on turfs

### LINKED ISSUE #9057: Amusing Ducks, Guardbuddies, Securitrons, etc, can be used as dollar store AA
### Labels

[MAJOR]

### Description

Bots with the appropriate access level can be used as door openers so long as they are pulled underneathe the door using right click context menus.

### Steps to reproduce

For Amusing Duck (Chance based to start with AA)
a1. Set their egg laying to not true for lack of movement
a2. Pull them infront of a high access door
a3. Right click context on the floor underneathe the door

For Guardbuddies
b1. Get flash resist eyes
b2. Beat them once
b3. Pull them around as they flash you
b4. Same steps as a2 & a3

For Securitrons (or other bots
c1. Deactivate or find one that starts idle (Non-patrolling)
c2. Same steps as a2 & a3

### Expected Behavior

Preferably either remove the context menu for flooring underneathe glass or doors, or just disallow from pulling these robots

I thought this was reported ages ago, turns out it wasn't?

Reported by: Nihisohel
Round log date: 2022-06-09-00-48
Reported on: 2022-06-09 01:30:01

M-Earthfire:
[player actions][bug][balance][critters]

## About the PR 

This PR introduces a few changes to bots on the station, most notable secbots.

Firstly, this PR changes `mob.pulled_by` to be located on `atom/moveable` instead. This means items and other atoms track which mob is pulling them.

This new behaviour was used in multiple areas.
Firstly, trying to pull a bot into a door it has access to won't open the door anymore. This fixes #9057.

Secondly, active securitrons cannot be pulled anymore while they are active. To pull them around, you need to deactivate them.

Lastly, securitrons won't search for targets to arrest anymore while they are not on a turf, this means securitrons in lockers/crates won't jump out to capture anybody if they were idling there.  Ordering them to move somewhere/patrol will still make them leave lockers/crates.

## Why's this needed? 

Firstly, dragging ducks or securitrons for cheap AA is a bad idea.

Secondly, with the introduction of the securitron whistle to the brobocop module in #16260, we saw a massive uptick in borgs roleplaying as security by dragging around securitrons and using the whistle to actively persecute crimers. 

Since secborgs was a thing that was very intentionally removed from the game because of all the problem it brought with validhunting and imbalance between antags/security, this is a behaviour that should not be incentivized.
This PR prevents many of the means of actively hunting people with securitrons. You still are able to move inactive securitrons, move them in a crate and release them or order securitrons onto an area, but these changes add delays into that behaviour that makes it much more predictable and open a bigger window of opportunity to react.

On the other hand, this changes how to deal with securitrons. They need a crate/locker to be moved out of an area, which makes it a little more harder and obvious for antags to move the securitron somewhere so you can crime in an area. On the other hand, this prevents surprise-securitrons from jumping out of lockers on their own.

On the antag side, emagged securitrons are easier to transport, since they only jump out of lockers if they are already trying to baton someone (most likely you).

In general, these changes should highly diminish the active patrolling of borgs with securitrons and should make securitrons slightly more harder to move away from an area they should guard.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Activated securitrons cannot be pulled anymore.
(+)Securitrons don't search for targets while in a locker. They still move out of they are patrolling/chasing someone.
```

### DIFF
--- a/code/atom.dm
+++ b/code/atom.dm
@@ -64,6 +64,9 @@ TYPEINFO(/atom)
 	/// Storage for items
 	var/datum/storage/storage = null
 
+	/// Which mob is pulling this atom currently
+	var/mob/pulled_by = null
+
 /* -------------------- name stuff -------------------- */
 	/*
 	to change names: either add or remove something with the appropriate proc(s) and then call atom.UpdateName()
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -54,7 +54,6 @@
 	var/other_mobs = null
 	var/memory = ""
 	var/atom/movable/pulling = null
-	var/mob/pulled_by = null
 	var/stat = STAT_ALIVE
 	var/next_click = 0
 	var/transforming = null
@@ -1129,10 +1128,7 @@
 			return
 
 	src.pulling = A
-
-	if(ismob(src.pulling))
-		var/mob/M = src.pulling
-		M.pulled_by = src
+	A.pulled_by = src
 
 	//robust grab : a dirty DIRTY trick on mbc's part. When I am being chokeholded by someone, redirect pulls to the captor.
 	//this is so much simpler than pulling the victim and invoking movment on the captor through that chain of events.
@@ -1147,9 +1143,8 @@
 	pull_particle(src,pulling)
 
 /mob/proc/remove_pulling()
-	if(ismob(pulling))
-		var/mob/M = pulling
-		M.pulled_by = null
+	if(src.pulling)
+		src.pulling.pulled_by = null
 	src.pulling = null
 
 // less icon caching maybe?!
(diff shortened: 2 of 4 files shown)


Chatauscours:
Good and needed nerf to brobocops in my opinion. Even this PR doesn't make it through, I would support the pull and click gameplay of brobocops being limited if not removed since it is getting too powerful for sec to effectively have another secoff ( or in some rounds 2 or even 3 more) at all times.

UrsulaMejor:
Overall support the changes to securitrons and bots. Haven't looked at the code so I'm not marking it approved, but the ideas seem solid

Tarmunora (changes requested):
Also need to break pulls when turning on via `attack_ai()`

PR #16319, opened 2023-10-07
tgui teleconsole

Valtsu0:
## About the PR 
Replaces teleconsole's old ui with a tgui one
As a consequence, cyborgs can no longer interact with it from off-screen
image
image

## Why's this needed? 
tgui good

## Changelog 

```changelog
(u)Valtsu0 and Mordent
(+)Science teleporter console now has a fancy new tgui ui
(+)Cyborgs can no longer use the science teleporter console from more than a screen away
```


DisturbHerb:
Why's the window so big compared to its content?

Valtsu0:
I have made the window slightly smaller by default. Images updated to reflect this

tgui/packages/tgui/interfaces/TeleConsole.tsx
```diff
@@ -0,0 +1,191 @@
...
+            />
+            <Button
+              color={host_id ? "green" : "gray"}
+              icon="sign-in-alt"
+              onClick={() => act("receive")}
+              content="Recieve"
+            />
+            <Button
+              color={host_id ? "green" : "gray"}
+              onClick={() => act("portal")}
```
ZeWaka, line 137: no, trigger the `disabled` prop, not change colors

ZeWaka (changes requested):

Paaiii:
hmm.. not an inherent ISSUE per say? well. depends how you look at it
but with the current telescience console UI, cyborgs are able to interact with it from anywhere.
which can be useful for quick motion and such for a lot of things - i personally use it for that constantly just casually for moving around or gimmicks because it's so handy

making it TGUI would make it only interactable to cyborgs within viewing range. Which, while it seems to be the intention with computer interfaces as a general rule, should be at least noted down, because it was a pretty valuable tool in the silicon arsenal.

Valtsu0:
> hmm.. not really an issue exactly, but with the current telescience console UI, cyborgs are able to interact with it from anywhere. which can be useful for quick motion and such for a lot of things - i personally use it for that constantly just casually for moving around or gimmicks because it's so handy
> 
> making it TGUI would make it only interactable to cyborgs within viewing range. Which, while it seems to be the intention with computer interfaces as a general rule, should be at least noted down, because it was a pretty valuable tool in the silicon arsenal.

I feel like this should be part of sciborgs kit/an upgrade

mordent-goonstation:
> hmm.. not really an issue exactly, but with the current telescience console UI, cyborgs are able to interact with it from anywhere. which can be incredibly useful for quick motion and such for a lot of things, benign or otherwise - i personally use it constantly just casually for moving around or gimmicks because it's so handy.
> 
> making it TGUI would make it only interactable to cyborgs within viewing range. Which, while it seems to be the intention with computer interfaces as a general rule, should be at least noted down, considering how valuable a tool telescience can be in the silicon arsenal.

Standardizing this is worthy of a note, but I would think is worth doing as it reduces surprises. The AI (and I believe AI controlled cyborgs, though would have to double check that) still have access to it from anywhere.

tgui/packages/tgui/interfaces/TeleConsole.tsx
```diff
@@ -0,0 +1,196 @@
...
+            );
+          })}
+        </Section>
+        {panel_open ? (
+          <Section width="80%" mx="auto">
+            <Box>
+              {"Open panel:"}
+            </Box>
+            <Box>
+              {"Linked pad number:"}
```
mordent-goonstation, line 185:
```suggestion
              Linked pad number:
```

tgui/packages/tgui/interfaces/TeleConsole.tsx
```diff
@@ -0,0 +1,196 @@
...
+                <Button
+                  icon="trash"
+                  color="red"
+                  onClick={() => act("deletebookmark", { value: mark["ref"] })}
+                />
+              </Box>
+            );
+          })}
+        </Section>
+        {panel_open ? (
```
mordent-goonstation, line 179:
```suggestion
        {!!panel_open && (
```
Then you can drop the `: null` at the other end of this. `!!` is required because tgui does not pass booleans around as true/false, so this will probably render as `0` or something if you just leave as `panel_open && (`.

tgui/packages/tgui/interfaces/TeleConsole.tsx
```diff
@@ -0,0 +1,196 @@
...
+          />
+          {bookmarks.map(mark => {
+            return (
+              <Box key={mark["ref"]}>
+                <Button
+                  icon="bookmark"
+                  onClick={() => act("restorebookmark", { value: mark["ref"] })}
+                  content={mark["name"]}
+                />
+                {mark["xyz"]}
```
mordent-goonstation, line 169: Given how you're using this data, why not pass up as three fields `x`, `y`, `z` which are just numbers then format however you want here?

code/modules/networks/computer3/mainframe2/telesci.dm
```diff
@@ -1234,6 +969,148 @@ TYPEINFO(/obj/machinery/networked/teleconsole)
...
+				var/datum/teleporter_bookmark/bm = locate(params["value"]) in bookmarks
+				if(!bm) return
+				bookmarks.Remove(bm)
+
+			if ("addbookmark")
+				if(length(bookmarks) >= max_bookmarks)
+					boutput(usr, "<span class='alert'>Maximum number of Bookmarks reached.</span>")
+					return
+				var/datum/teleporter_bookmark/bm = new
+				var/title = tgui_input_text(usr, "Enter name:", "Name", "New Bookmark")
```
mordent-goonstation, line 1077: This is probably "fine", but might be neater to have it as a simple text input as part of the UI rather than a popup?

mordent-goonstation: Bad two minute MS Paint mockup here:
image

tgui/packages/tgui/interfaces/TeleConsole.tsx
```diff
@@ -0,0 +1,196 @@
+import { useBackend } from '../backend';
```
mordent-goonstation, line 1: Add a license header to the top of the file. See other tgui interfaces for an example.

mordent-goonstation (changes requested):
I have some additional design feedback but that can wait for a follow-up PR from myself or someone else, the port as the main chunk of the work seems mostly decent.

code/modules/networks/computer3/mainframe2/telesci.dm
```diff
@@ -1234,6 +969,148 @@ TYPEINFO(/obj/machinery/networked/teleconsole)
...
+				newsignal.data["address_1"] = old
+				newsignal.data["sender"] = src.net_id
+
+				src.link.post_signal(src, newsignal)
+				SPAWN(1 SECOND)
+					if (!old_host_id)
+						old_host_id = old
+
+			if ("setpad")
+				src.padNum = (src.padNum & 3) + 1
```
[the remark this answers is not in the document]
Valtsu0: The telepad number cycles the numbers between 1 and 4 (including), this is to make it wrap back to 1

code/modules/networks/computer3/mainframe2/telesci.dm
```diff
@@ -1234,6 +969,148 @@ TYPEINFO(/obj/machinery/networked/teleconsole)
...
+
+	ui_act(action, params)
+		//. = TRUE means the action was handeled
+		. = ..()
+		if (.)
+			return .
+
+		. = TRUE
+
+		playsound(src.loc, 'sound/machines/keypress.ogg', 50, 1, -15)
```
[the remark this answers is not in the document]
Valtsu0: This is the current behaviour and i don't find it too disrupting

tgui/packages/tgui/interfaces/TeleConsole.tsx
Valtsu0: The reason i did this is because i didn't feel like figuring out template strings. Should do this the proper way

Valtsu0:
If it's not obvious, DNM

some conversations have been resolved without the changes being pushed

mordent-goonstation (approved):
Broadly approving (with changes below), I think there's room for some design enhancements here but the base goal of "get it into tgui" seems to be fine, and I don't want to block that with my (subjective) design thoughts.

mordent-goonstation (changes requested):
Putting a "request changes" block on this until I've got some improvements added to the branch. Working on those actively, please hassle me if this takes >24 hours.

mordent-goonstation:
PRed my changes to your branch, so they can be merged in in one go rather than incrementally.

https://github.com/Valtsu0/goonstation/pull/1

PR #16323, opened 2023-10-07
Makes the stealth storage more stealthy

M-Earthfire:
[internal][game objects][balance]

## About the PR 

This PR introduces the stealthy_storage variable for the storage datum.

Setting this variable to true will hide the tooltip that says how many items a storage can and is containing.

This was used to make stealth storage hide it's contents on the tooltip.

## Why's this needed? 

Stealth storages are far too easy to make out due to their tooltip being a dead giveaway that you are not holding an item but a stealth storage. This change will make it harder to distinguish the storage from normal items and makes it more likely that the storage won't be revealed on a search by security

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Stealth storages have become a little bit more stealthy.
```


Tarmunora (approved):
lgtm
although, any reason not to just tie this to `sneaky`?

PR #16326, opened 2023-10-07
HoS pocket pouch spawn consistency

Chatauscours:
## About the PR 
Makes the HoS spawn with their pouch in their left pocket like all of the other security personnel. Consistency.

## Why's this needed? 
Because everyone else (Secass, Secoff, NTSC, NTSOs, ... ) is spawning with the pouch in their left pocket. The HoS spawning with it in the left as well one is absolutely critical to HoSing. Very critical.


444Portal:
lmao

MeggalBozale:
these are the real changes goon needs

PR #16337, opened 2023-10-08
Resprites Hadar Heavy Power Sword, fixes bug where Hadar had 5 extra max damage

MylieDaniels:
[Sprites] [Game Objects]
## About the PR 
This PR resprites the Hadar, giving it new visuals that reflect the current charge level and indicate what the mode does.
When damage is less than half of maximum damage (50, for the typical 100 damage capped Hadar), there are these unanimated sprites:
hadarlow3x
After reaching half the maximum damage, this loop is animated:
hadarmed3x
And finally, after hitting 90% charge, the loop is made a bit more extra:
hadarhigh3x
Matching inhand sprites:
hadarinhand3x
Matching back sprites:
hadarback3x

While adding code for these animations to be switched to, I found a small bug - the Hadar was able to increase damage if it had exactly it's supposed maximum damage, which allowed for an actual maximum of 105 - I fixed this here, but can atomize that bugfix if necessary.

## Why's this needed? 
The Hadar is a powerful, iconic weapon with a sprite that was, according to Gannets, put together with no mouse on the floor to get it in a playable state. It also did not have a unique inhand icon (it was using d-saber inhands).

## Changelog 

```changelog
(u)MylieDaniels
(+)Nukie Knights have a refreshed visual on their Hadar power sword.
```


ZeWaka:
definitely not a major change.

MylieDaniels:
Sorry, I don't actually know the criteria for either. Thank you for fixing it.

flappybatpal:
I appreciate that this has big, obvious held sprites as right now the held sword is small and indistinct.

Gannets (approved):
Perfect, thank you!

PR #16338, opened 2023-10-08
Reverse PTL

Azrun:
## About the PR 
Adds a new machinery, the Power Reception Laser! Now you can buy power!

Using a conversion kit, you can convert the PTL into a PRL.
Set the power you wish to receive via Laser, and the amount of power to put onto the powernet.
Power is based on a price per MW, with a tiered increasing cost for every 10 MW.

image

Buying and providing power greater than MW is behind emag.

image

## Why's this needed? 
With the continued work on economy, this adds another possible use for those large payroll budgets.  Beyond Hamster-wheels and furnaces this is a possible alternative to provide power to the station upon the loss of stations primary means of providing power.

## Changelog 

```changelog
(u)Azrun
(+)New conversion kit for the PTL. What if you could BUY power?!?
```


wolfolotl:
fun stuff, though I feel like we don't really have anything right now that necessitates buying power short of Massive Station Emergency. could be nice to have an option if there's also a way to change the ptl back

TheTyrant25:
how does this interact with PTL beam mirrors? does the beam come in the other way?

Azrun:
> how does this interact with PTL beam mirrors? does the beam come in the other way?

Correctly? Yes.

Azrun:
> fun stuff, though I feel like we don't really have anything right now that necessitates buying power short of Massive Station Emergency. could be nice to have an option if there's also a way to change the ptl back

Massive Station Emergency is really the only real "need" for one.  Otherwise it might be a goofy way to start the round. 😈

Sbmhawk:
> > how does this interact with PTL beam mirrors? does the beam come in the other way?
> 
> Correctly? Yes.
Does the beam always aim incoming at the original PTL location, or will centcomm ""know"" how to bounce the beam to get to the PTL

Azrun:
> > > how does this interact with PTL beam mirrors? does the beam come in the other way?
> > 
> > Correctly? Yes.
> 
> Does the beam always aim incoming at the original PTL location, or will centcomm ""know"" how to bounce the beam to get to the PTL
> 

It will come from the edge of the map the PTL would have shot to.

Azrun:
!merge_upstream

Azrun:
!merge_upstream

Vincent983:
Perhaps you should add a way to turn the PRL back into a PTL?

PR #16346, opened 2023-10-08
Rework of the  Chemical Grenade Starter Kit, additional chemnade convenience

M-Earthfire:
[game objects][rework][code quality][player actions]

## About the PR 

This PR replaces the chemical grenade starter kit with a chemical grenade pouch with increased grenade count and reworks how chemical grenades are assembled. It also adds a new feature to chemical grenades in general.

The chemical grenades come now in a chemical grenade pouch with 5 disassembled chemical grenades:
chem grenade rework

Instead of having the fuse seperate from the casing, fuse and casing now come in one item and can be assembled together by using the disassembled chemical grenades in hand. 
chem grenade WIP

The rest of the assembly steps are the same like pre-rework. Due to the use of the assembly component, you can hit the grenade with the beakers and vice-versa and it still works. 

Now, you can disassemble chemical grenades by using a wrench on them, returning the used beakers and disconnecting the grenade from the fuse again. They are also now mechscanable via a syndicate mechscanner.

Chem grenades in general got a new change as well. You can use a screwdriver on them to change the detonation time in 3 second steps from 3 seconds to 15 seconds.

Last but not least, this PR makes the whole file (grenade.dm) use absolute pathing.

## Why's this needed? 

chemical grenades are considered a very underwhelming convenience buy because most of their effects can nowadays be achieved by beaker assemblies, save a few exceptions. To make them more appealing to buy and compareable to other grenade packs, a pouch was added and their number increased to 5.

The Chemical Grenade Starter Kit still exists in prefabs, but still only has 3 disassembled chemical grenades.

To enable the increase in number, the fuse and casing were combined. Also, the mess that is their assembly was brought in line via the newly added assembly component.

This enabled some additional convenience changes, like grenade disassembling or adjustment of the explosion time, which is already a thing for other grenades. The time of 15 seconds was used to enable delivery of grenades via mail chutes or delayed traps.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)The Chem Grenade Starter Kit was replaced with a chemical grenade pouch with 5 diassembled grenades. The grenade casing are no longer seperated from their fuse and can be simply added together by using the disassembled grenades in your hand.
(+)Non-armed chemical grenades from the chemical grenade pouch can now be disassembled with a wrench.
(+)Chemical grenades, like incendiary grenades or FF-grenades, can have their timer extended with a screwdriver to up to 15 seconds.
(+)Disassembled chemical grenades can be scanned with a syndicate mechscanner.
```


Tarmunora (changes requested):
Hm. Kinda getting the feeling the custom grenade code should be moved to a child type e.g.: `obj/item/chem_grenade/custom`, and have the parent type only handle arm/detonation/etc. code

PR #16361, opened 2023-10-09
Medal reward to turn back into a green captain

Valtsu0:
## About the PR 

Adds a medal reward to the medal [Tinkerer](https://www.byond.com/games/Slurm/GoonStationMedals?command=view_medals&medal=25498) called `(Skin Set) Captain Uniform`
Works the same as `(Skin Set) CENTCOM Executive Uniform` and `(Skin set) NT-SO Commander Uniform` except it turns your stuff back to the regular green.

I considered making this a reward for [It's not easy being green](https://www.byond.com/games/Slurm/GoonStationMedals?command=view_medals&medal=530) but that is much easier than the medals for the skins and unlocking this first is weird

## Why's this needed? 
You can already switch between red and blue at will, why not green too?

## Changelog 

```changelog
(u)Valtsu0
(+)Added a medal reward to turn your captain items back to the familiar green
```


Sbmhawk:
I think this would be better just as a level 0 job reward, or for something like captains log. This seems more qol than an actual reward.

444Portal:
Could you not make it check for either Brown Pants or Icarus? I don't see why reverting something to it's default state would be locked behind a medal other than ones that let you change them, personally.

DimWhat:
Tbh, i'd rather just have being able to only pick one colour for captain gear. Being able to change back and forth is kinda between red and blue feels more like an oversight than intentional. 

Secondly a reward that simply does nothing doesn't feel good

DimWhat:
Fuck misclick.

Tarmunora:
Yeah I don't see the point of this at all

TobleroneSwordfish:
I'm closing this, see DimWhat's comment for reasoning.

PR #16362, opened 2023-10-09
Make bovine udders smaller so they don't clip

kbsmilk:
[MUTANTRACE]
## About the PR 
Cuts 8 pixels off so the bovine body doesn't clip through/under jumpsuits. An attempt was made to preserve the shape and appearance, but 8 pixels made quite a difference.

## Why's this needed? 
I don't like the clipping pixels.
Some suggested alternatives, including this PR (skinny cow), and the implicit "leave the udder as-is."

#### Displacement map on clothes to cover the udder:
* Requires maps on multiple different slots due to udder size. All of these listed overlap pixels with the clipping udder pixels.
* Jumpsuit, of course.
* And exosuit, for all the jackets in the game.
* Belt slot. No belt sprite is prepared to handle The Udder.
* Weirdly enough, glove slot. Glove sprites render over the udder (and jumpsuits) so these need trimming, too.
* This is so much work and I don't think it'll look good. e.g. monkeys got bespoke clothing sprites instead of displacement maps for a reason.

#### Hidden udder overlay
* Essentially just squishing or hiding the extra udder pixels while a jumpsuit is being worn.
* Not too complex to make, but also how often is anyone really playing without a jumpsuit on?
* Still results in skinny cow the majority of the time.

#### Always skinny cow
* Permanently flatten the udder. Make cows human-sized and thus human clothes-compatible.

#### Udder as an overlay, on top of jumpsuits, but under exosuits
* Yes this means cows are cutting udder-shaped holes in their clothes.
* This likely won't play very well with belts and gloves.
* Pretty funny though.

### Before
image
image

### After
image
image

### In-editor screenshots
image
image
paintdotnet_h1GYXiCMCl
paintdotnet_tfS4377BqN
paintdotnet_bKR7dqwPHP

## Changelog 

```changelog
(u)Smilg
(+)Bovine udders no longer visually clip in the east/west directions.
```


DisturbHerb:
free the udder!! 🐮

PR #16366, opened 2023-10-09
Add Action Bar to Removing Floor Reinforcement

TealSeer:
[GAME OBJECTS] [PLAYER ACTIONS] [QOL]
## About the PR 

Changes un-reinforcement behavior of wrench/welder on reinforced floors to have an action bar.

## Why's this needed? 

do_after is nasty and gives no feedback on when it'll be done and is obnoxious to cancel in case of a misclick


frawhst:
all do_after removals get my speedmerge of approval

PR #16368, opened 2023-10-09
Outfit datums to replace clothing in jobs

CheffieGithub:
## About the PR 
Replaces every set of clothing in jobs with new outfit datums. These datums just hold that same information but can be applied else where so you can equip a mob quickly without needing a job associated with it, allowing for better setup for corpse spawners, critter humans, npcs, chameleon clothing, antagonists, etc. as you only need to define clothing once to use in multiple places instead of needing to keep repeating it.

Everything should be exactly the same as it was before for the player with no visible change.

This will break every custom outfit that exists currently so sorry about that.

- [ ] Test every outfit to make sure they aren't missing anything
- [ ] ID handling somehow without conflicting maybe

## Why's this needed? 
Its better to have an outfit you define once that you can use for jobs, equips, antagonists, etc. instead of needing to keep repeating the same code.

## Changelog 

```changelog
(u)CodeDude
(*)Added soft soft pizza to the game - a tasty drink found in soda machines!
(+)Drinking any kind of soda will cause you to burp violently.
```


ZeWaka:
oh god

CheffieGithub:
Sisyphus-e1557869810488-1652777917

PR #16383, opened 2023-10-10
QoL Changes to Sword Sheaths + Katana Sakuride Coating

KevinfromHP:
## About the PR 
- All sword sheaths can now be interacted with in a more comfortable manner. You can now mouse drag to remove the sheath like any other bag.
- New feature: all katanas can now be coated with sakuride! Clicking on a pure beaker of sakuride will coat the katana in up to 15u of it. When you use your dash, it will now leave a trail of sakura petals behind you. Each tile uses one unit, so you have 5 dashes before it's used up. As of now, you can only coat it in sakuride, and it has no effect beyond this.

## Why's this needed? 
- The katana can be a pain in the ass to work with. The control changes felt like a necessity since you would have to take the sword out, then use you other hand to remove the sheath.
- The sakuride coating feels like a nice touch to the katanas. I just thought it'd be fun.

## Changelog 

```changelog
(u) Kevin From HP
(+) Sword sheaths can now be dragged out of belt slot onto hand/ground.
(+) Using any katana on a container of pure sakuride will coat it in up to 15u. Dashing leaves a trail of petals behind you.
```


DisturbHerb:
Don't think the workshop katana needs a sheath, myself.

CheffieGithub:
Its incredibly weak for the time investment so a sheathe is a nice small buff

CheffieGithub:
Personally intent based interactions feel bad but we don't really have a framework for other shortcuts as alternatives

KevinfromHP:
I can drop the grab intent thing if that's what's preferred, I personally like how the shinai bag's intent changes things but if that's not a popular thing then that's alright

CheffieGithub:
Grab is fine but maybe being able to drag click it into your hands would be good as well

KevinfromHP:
> Grab is fine but maybe being able to drag click it into your hands would be good as well

Already done, same as a backpack/belt

TobleroneSwordfish (changes requested):
- Click dragging is good, but intent based inventory interactions are very unclear.
- The workshop katana not having a sheath is very intentional.
- The sakura petals are cute and cool 👍

zjdtmkhzt:
What's the intent with the workshop not having a sheath? I feel like it would be much cooler if it did, katana without a sheath kind of just sucks from a coolness standpoint.

I agree though that it should probably have a different name and icon to make it so you do not appear to be an antagonist?

KevinfromHP:
I considered a new sheath, but alrady being able to make both a handcrafted and reverse blade (which already has a sheath) there made me think it was best to go with a vanilla sheath.

I can drop grab intents easy enough, but if it were to receive its own unique sheath, could that change be kept?

DisturbHerb:
I think being able to wear and draw a katana at will is something best left to the traitor katana. That's what I assumed the reasoning to be against providing sheaths for the workshop ones.

KevinfromHP:
Though I disagree due to the lack of power of the katana(s) compared to other options, I'll drop the sheath.
However, I will leave my final thoughts on it. Reading over the discussion in the discord over this, I do not think the sheath would make a significant change to the way people play given the time it would take to create, a traitor's instant access to a better katana with a sheath, and the nature of "power gaming".
It surprised me how controversial this is considering many of the other options that are available, like the (storable) predator revolver and rifle on ice moon, a scrap plasma dagger, and the 120 seconds or less to obtain phaser in debris field. If someone were interested in a viable weapon, the 20-30 minute katana from the workshop seems illogical unless it was solely an aesthetic choice. I wouldn't see this as a major change to the game though people might see a katana slightly more often if it had a sheath.

KevinfromHP:
> * Click dragging is good, but intent based inventory interactions are very unclear.
> * The workshop katana not having a sheath is very intentional.

These two things have been revised as requested

code/obj/item/misc_weapons.dm
```diff
@@ -1269,9 +1269,24 @@ TYPEINFO(/obj/item/swords/katana)
...
 
+	afterattack(atom/target, mob/user)
+		..()
+		apply_coating(target, user)
+
+	proc/apply_coating(var/atom/target, var/mob/user)
+		if (target.is_open_container())
+			var/obj/item/reagent_containers/RC = target
+			if(RC.reagents.has_reagent("sakuride", 1))
+				if(RC.reagents.reagent_list.len > 1)
```
TobleroneSwordfish, line 1284:
```suggestion
				if(length(RC.reagents.reagent_list) > 1)
```
always use `length()` over `.len`, byond jank don't worry about it

_std/defines/item.dm
```diff
@@ -212,6 +212,8 @@
 
 // for pen reagent dipping
 #define PEN_REAGENT_CAPACITY 4
+// for katana coating, applied to turf on dash. Only used for sakuride for now
+#define KATANA_REAGENT_CAPACITY 15
```
TobleroneSwordfish, line 216: No particular reason for this to be a global define, may as well just be on the katana type.

code/obj/item/misc_weapons.dm
```diff
@@ -1269,9 +1269,24 @@ TYPEINFO(/obj/item/swords/katana)
...
+					boutput(user, "<span class='alert'>This coating is impure!</span>")
+					return
+				RC.reagents.trans_to(src, min(KATANA_REAGENT_CAPACITY , src.reagents.maximum_volume - src.reagents.total_volume))
```
TobleroneSwordfish, line 1287: the `min` is pointless here, `trans_to` should handle all that itself

code/obj/item/misc_weapons.dm
```diff
@@ -1269,9 +1269,24 @@ TYPEINFO(/obj/item/swords/katana)
...
+				boutput(user, "<span class='hint'>You apply the coating to the blade.</span>")
```
TobleroneSwordfish, line 1288: add a message here saying something like "this blade is already coated" if it's full

TobleroneSwordfish (changes requested):

Azrun:
I think this PR is doing far too many things. Sakura should be atomized to a separate PR.

KevinfromHP:
> I think this PR is doing far too many things. Sakura should be atomized to a separate PR.
If you removed sakuride from this PR, wouldn't it just be making the sword sheath mouse draggable?

code/obj/item/misc_weapons.dm
KevinfromHP: I'll get all this done btw. Most of this I referenced from pen reagents since this is baby's first work in DM. Sakuride was my test project to give it a use besides beaker bombs

code/obj/item/misc_weapons.dm
```diff
@@ -1269,9 +1269,29 @@ TYPEINFO(/obj/item/swords/katana)
...
+		src.create_reagents(src.reagent_capacity)
 		src.setItemSpecial(/datum/item_special/katana_dash)
 		BLOCK_SETUP(BLOCK_SWORD)
 
+	afterattack(atom/target, mob/user)
+		..()
+		apply_coating(target, user)
+
+	proc/apply_coating(var/atom/target, var/mob/user)
+		if (target.is_open_container())
```
TobleroneSwordfish, line 1281: actually thinking about it, this message will show when you attack a headless human (which is pretty common with a katana for obvious reasons), best to check it isn't a mob at least

code/obj/item/misc_weapons.dm
```diff
@@ -1269,9 +1269,29 @@ TYPEINFO(/obj/item/swords/katana)
...
+			var/obj/item/reagent_containers/RC = target
```
TobleroneSwordfish, line 1282: you don't need this, `.reagents` is defined on `atom`

TobleroneSwordfish (changes requested):
This ideally should be atomized, yes, but I'm fine with merging it as the changes are relatively minor. Just in future please don't package different changes into one PR.

KevinfromHP:
That commit says checks if human but it checks if it's a mob, my mistake

PR #16393, opened 2023-10-11
Soda Water now quenches thirst

CrystalClover:
## About the PR 
Changes Soda Water to now quench as much thirst as regular water

## Why's this needed? 
Soda Water currently quenches no thirst, which makes it bad for making drinks
Also it just makes sense to have it quench thirst


ZeWaka:
Thank Goodness This Devilish TV Static Tasting Liquid Doesn't Have A Value Of 1

ZeWaka (approved):

PR #16394, opened 2023-10-11
Refactors the contraband section of the arrest_icon life process into a component

MylieDaniels:
[Station Systems] [Code Quality]
## About the PR 
<b>Contraband Component:</b>
This PR removes the section of the arrest_icon life process that recalculated contraband every life tick.
To replace it, it introduces a component, `datum/component/contraband`, which can be attached to any movable.
It takes two arguments, contraband and firearm contraband, stores them, applies them as atom properties to any mob with a contraband component that carries the item, and has a getter to return the contraband value of the owner movable.
This component is automatically attached to all instances of `mob/living/carbon/human`, with no contraband value set, and all instances of `obj/item` that have a `contraband` variable greater than 0, with that a value equal to that variable - if the item is an `obj/item/gun` then this is set as firearm contraband. Attaching a contraband sticker will send a signal that re-applies contraband from the item upward. Both can be set, allowing a movable to be, for instance, 4 contraband that requires a contraband license and 4 more contraband that requires a carry permit.

<b>Security Scans:</b>
All checks of `obj/item/get_contraband()` have been changed to send a signal to this component instead, and the logic for getting a movable's visible contraband value is now `SEND_SIGNAL(mob, COMSIG_MOVABLE_GET_CONTRABAND, nonfirearms = TRUE, firearms = TRUE)`
The life process for arrest icons is now, to preserve player-facing visuals, only responsible for drawing the contraband icon, not calculating it. No change should be observed in the time it takes for SecHUDs to update, but securitrons may be very slightly more responsive to contraband being briefly pulled out.

<b>Admin Perspective:</b>
To change an item's (or any movable's) contraband value from an admin perspective, you would now apply this component instead of changing `contraband`
Attaching this component to a mob allows it to have innate contraband levels and/or inherit contraband from carried contraband items, which I plan to PR for all sources of non-item contraband (mutantraces like abomination, mobs like gunbots)

<b>Defines:</b>
COMSIG_MOVABLE_CONTRABAND_CHANGED - a signal to a movable when anything directly overrides its contraband level
COMSIG_MOVABLE_GET_CONTRABAND - a signal that returns the summed contraband visible on a movable, taking two boolean arguments (nonfirearm, firearm)
COMSIG_ITEM_EQUIPPED_NOCHECK - a signal sent when an item is equipped before any checks that may cause the rest of the equipped proc to return
COMSIG_ITEM_UNEQUIPPED_NOCHECK - vice versa

PROP_MOVABLE_VISIBLE_CONTRABAND - visible contraband on a movable, summed
PROP_MOVABLE_VISIBLE_GUNS - visible firearm related contraband on a movable, summed

## Why's this needed? 
To support further work (securitron mobification, arrest icon componentization, property based contraband) and remove the inconsistent delay on contraband levels being calculated.

## Changelog 

```changelog
(u)MylieDaniels
(+) The way Contraband is calculated has been changed under the hood, nothing should appear different but please report any bugs.
```


DasBrain:
>  but securitrons may be very slightly more responsive to contraband being briefly pulled out.

This may be bad for emaging them. Not that it matters after they have been emaged, but now they already have a target.

TealSeer (approved):

TealSeer:
not the most well versed on components but this LGTM

code/datums/components/contraband.dm
```diff
@@ -0,0 +1,86 @@
...
+		return_val += GET_ATOM_PROPERTY(owner, PROP_MOVABLE_VISIBLE_GUNS)
+		return return_val
+	if (nonfirearms)
+		return_val += HAS_ATOM_PROPERTY(owner, PROP_MOVABLE_CONTRABAND_OVERRIDE) ? GET_ATOM_PROPERTY(owner, PROP_MOVABLE_CONTRABAND_OVERRIDE) : src.contraband_level
+		return_val += GET_ATOM_PROPERTY(owner, PROP_MOVABLE_VISIBLE_CONTRABAND)
+	if (firearms)
+		if (src.carry_level)
+			return_val += HAS_ATOM_PROPERTY(owner, PROP_MOVABLE_CONTRABAND_OVERRIDE) ? GET_ATOM_PROPERTY(owner, PROP_MOVABLE_CONTRABAND_OVERRIDE) : src.carry_level
+		return_val += GET_ATOM_PROPERTY(owner, PROP_MOVABLE_VISIBLE_GUNS)
+	return return_val
```
Tarmunora, line 45: this is an incorrect use of signal return values - the rule for signal returns is that they get bitwise ORed together, so any values you return from a proc invoked by a signal should be in form of bitflags (incl simple boolean TRUE/FALSE)

Can look at how the power cell and cell holder components handle the get_charge signal as an example of how to work around that restriction

Yes it's awkward for something like this where you're wanting to just return a number, but I think that maintaining consistency is worth the added code hassle

Tarmunora (changes requested):

Tarmunora (approved):

ZeWaka (approved):

PR #16400, opened 2023-10-11
Airlock.dm cleanup and airlock cycling polish

TheTyrant25:
[MAPPING] [FEATURE] [QOL]
## About the PR
Cleans up the airlock.dm file massively.
Particularly, procs are now arranged by what they're inheriting from, and duplicated definitions have been merged.
Airlock cycling has also been changed.

### About Airlock cycling
Airlock cycling has been changed to use a global ID system instead of only being able to be linked to one other door. Airlock cycler mapping helpers have also been added to facilitate the new usage of them. For those asking "What's airlock cycling?", it basically means opening one door attempts to close some others. Closing any doors in a linked system doesn't open any others, though. It also only tries once, so anything holding a door open will prevent the mechanism from working, and the door will just stay open.

Link multiple airlocks by placing mapping helpers with matching `cycleid` vars, and the doors will be linked. 7 subtypes with different `cycleid` vars are provided for your convenience, although you can var edit the doors directly as well.

Fun fact: as I went to code this in, I realised that airlock cycling was actually already a thing in game, and had pre existing vars and code for it. It was just unused in every map because it required var editing and could only have two doors linked at once, and was gradually forgotten about. It may be used in the secret submodule but this is unlikely. They are not being added to maps in this PR, only having the code cleaned up.

## Why's this needed?
The code was quite honestly a mess. Cleaner code is better and makes future work easier.
As for making airlock cycling usable, it seems like a nifty bit of quality of life that any self respecting station would have. It makes the airlocks actually lock air, which is a plus, but it isn't intrusive enough that it requires a separate button, and shouldn't impede that much.

(Plus, Sord wanted them.)
image


TheTyrant25:
All below examples are on cogmap1
Examples of spots they can be used:
image
image
Examples of spots where they can't be used as is and would require extra code:
image
image

CheffieGithub:
It seems counterintuitive for a helper if you need to set an ID for every door, pairing in a line based on helper dir would be better

CheffieGithub:
Its like a case to case thing. Generally most airlocks are in lines so I think a directional approach and an ID approach both as options would be ideal. ID has the benefit of being able to link multiples where directionals are pairs. But directionals are more convienient than needing to worry about IDs and whether or not its been used or not already.

TheTyrant25:
idk man the pre existing system just uses id tags 👻

CheffieGithub:
You can do directional with ID tags you just need to create a unique one and assign it to both

mordent-goonstation:
Seems thematically neat, but isn't the "problem" this solves mostly covered by the atmos shields we have on basically every airlock these days?

CheffieGithub:
Atmos shields are a pretty boring solution for that problem for every airlock imo, but these have more uses than just that.

TheTyrant25:
True, but atmos shields are pretty high tech stuff imo. For older prefabs and some interior station stuff (e.g. decontamination airlocks in research or engineering, and several spots in the outer mainta of nadir) these could work nicely.
Again, this is code that already existed, I'm just making it more accessible to mappers.

TobleroneSwordfish:
Since the point of these is "a thing other servers have", I checked how TG do it and they just scan in a direction. I think the existing ID system works good for complex cases but the mapping helpers should be to solve the common case of two aligned airlocks.

(Sord feel free to contradict me, but I don't think they're reviewing PRs right now)

TheTyrant25:
right, well i'm going to have two different kinds of map helpers then

TheTyrant25:
well the scope has crept. gimme a second

TheTyrant25:
right, ready for reopening now that i've changed the title and description

pali6:
This PR has been reverted as it breaks airlock code in multiple places. For example open / close procs have their checks moved to after the parent call. PLEASE atomize your PRs. Don't do everything in one huge monolithic chunk. If you are reorganaizing stuff do that first. It is VERY hard to distinguish when you just moved something and when you moved it and also changed it.

PR #16404, opened 2023-10-11
Remove non-player Pests event

flappybatpal:
## About the PR 
Removes the minor Pests event that does not involve player characters.

## Why's this needed? 
Feedback on Discord and ingame is consistently negative.
While we have a limited number of ways for critters to come to the station (science teleporter, spooky vortex) we do not appear to be gaining anything positive from this event.
Pests are either totally ignorable with no gameplay effect (cockroach/mouse) or dangerous (wasp/scorpion/rattlesnake) particularly in numbers.

Considerations:
Presence of Wasps will now guarantee a traitor is around?
Limited use of scorpions/rattlesnakes outside of terrainify until Artemis/planets become more of a thing,

## Changelog 

```changelog
(u)Flappybat
(+)The minor pest event (non playable critters) has been removed.
```


CheffieGithub:
Maybe it would be a good idea to look at WHY exactly people don't like the event and make adjustments instead of outright removal? At least if its not going to have a replacement for forseeable future.

blackrep:
Personal opinion, I think the hostile pest events are hilarious, and it adds a bit of variability to the round giving security, and medbay a few more things to deal with.

Chatauscours:
I believe that the pest events are fine overall. However, from what i've seen, there is a lot of frustration caused by the places where they spawn, and in the condition they spawn in (aka any).
For the first part, generally get very unhappy that 3 snakes spawned in the middle of a very used hallway, though in the gym is fine, since it can wait till a secoff comes around.
For the second part, pests have a nasty possibility of spawning right under someone's feet, instantly stinging them multiple times, and causing a rapid death.

PR #16405, opened 2023-10-11
Dispensary Interlink Meter Screen

Flaborized:
[CHEMISTRY] [FEAT]

## About the PR 

This PR makes the dispensary interlink's screen into a meter that shows you how full it is. As a trade off of not wanting to make versions of this for each dir (and also thinking that south facing is the most visually clear anyways so it should always show that) I also cut the directionality of it, so it only has south facing sprites.

There's 10 icon states: it'll show a red bar if it's completely devoid of chemicals and otherwise show a blue wavy animation and fill up, increasing one stage around every 55u of chemicals it has inside its products. It doesn't care if it's *medical* chemicals, just that there's chemicals. A pill bottle full of epi pills will fill the counter, but an empty patch won't!

interlink_screenshot

also it glows in the dark :)

chem_dispenser_animation

Also I added a new file just for the dispenser and its animations, since there's lots of them and I'm aware that very large dmi files have a negative impact on performance. easily fixed if that's not preferred though

## Why's this needed? 

I like the sorta retro light up wave visual! Also I was playing chemist and found myself not wanting to use this thing because I didn't think doctors would even realize I put chemicals inside the thing to use (since people rarely do). This is a feedback loop!! Nobody checks it because it's empty... nobody fills it because nobody checks it... now you can just *see* that it's full of useful stuff. This might not entirely fix the reasons people don't use this so much but I think this is a pretty obvious first-step-ish approach to making it more usable.

## Changelog 

```changelog
(u)Flaborized
(+)The dispensary interlink dispenser in medical will now visually show how many units of chems it has in it, the more blue the screen the more meds it has to vend.
```


code/modules/vending/vending.dm
```diff
@@ -3372,13 +3374,41 @@ ABSTRACT_TYPE(/obj/machinery/vending/jobclothing)
...
+		var/total_reagents = 0
+		var/fluid_state = 0
+		for(var/datum/data/vending_product/player_product/R in src.player_list)
+			for(var/obj/item/product in R.contents)
+				if(istype(product, /obj/item/item_box/medical_patches) || istype(product, /obj/item/chem_pill_bottle))
+					for(var/obj/item/reagent_containers/container in product)
+						total_reagents += container.reagents.total_volume
+				if(istype(product, /obj/item/reagent_containers))
+					var/obj/item/reagent_containers/container = product
+					total_reagents += container.reagents.total_volume
```
TobleroneSwordfish, line 3400: You could probably cut the istypes here since it can be assumed that we want to read the reagents and contents of anything we put in here

code/modules/vending/vending.dm
```diff
@@ -3372,13 +3374,41 @@ ABSTRACT_TYPE(/obj/machinery/vending/jobclothing)
...
+
+		if(!total_reagents) //only show it as empty if it truly has Nothing at All, chemically
+			fluid_state = 0
+		else
+			fluid_state = round(clamp((total_reagents / 500 * 9 + 1), 1, 9)) //it'll jump up a state every ~55 reagents added or so
+		src.fill_image.icon_state = "medchem-[fluid_state]"
+		src.UpdateOverlays(src.fill_image, "fill_image")
```
TobleroneSwordfish, line 3407: Since you're never swapping out the overlay you should just be able to call `UpdateOverlays` once and then change the icon with your image reference?

TobleroneSwordfish (changes requested):
This is great! Should check that tipped over sprites look right with the lack of directionality if you haven't already.

code/modules/vending/vending.dm
Flaborized: unless I'm misunderstanding the suggestion trying this has made the icon not actually swap in-game? I *think* most overlays use UpdateOverlays to update icon changes like this, at least, though again it's entirely possible I just don't Get What You Mean :sheltersweat:

code/modules/vending/vending.dm
Flaborized: yup did this and it's much cleaner and easier to read and works!

Flaborized:
> This is great! Should check that tipped over sprites look right with the lack of directionality if you haven't already.

Yea funnily enough I don't think *any* of the tipped over sprites use directionality, so I kept this one the same. Icondiff is showing Very Wrong sprites though, possibly because of the new file or something ??

PR #16410, opened 2023-10-12
Add Tickets/Fines to end-of-round TGUI

pgmzeta:
## About the PR 

Moves tickets/fines display to the same end-of-round dialog as crew credits

## Why's this needed? 
reduce number of end-game pop-ups

## Screenshots
Screenshot 2024-04-14 161316
Screenshot 2024-04-14 161338

## Changelog
```changelog
(u)glowbold
(*)End-of-round tickets/fines now show up in the crew credits window
```


pgmzeta:
added a major changelog since i think this is the last shared end-of-round dialog?

tgui/packages/tgui/interfaces/CrewCredits/ReportTab.tsx
```diff
@@ -0,0 +1,100 @@
...
+                </Tabs.Tab>
+              );
+
+            })}
+          </Tabs>
+        </Stack.Item >
+        {reports.map((report, index) => {
+          return (
+            <Stack.Item key={index}>
+              { menu === index && <Report multi={index} {...report} />}
```
mordent-goonstation, line 48: This prop name is confusing. Consider something as simple as `id`.

tgui/packages/tgui/interfaces/CrewCredits/ReportTab.tsx
```diff
@@ -0,0 +1,100 @@
...
+  return (
+    <Stack vertical>
+      <Stack.Item>
+        <Section title={issuer}>
+          <Tabs>
+            {pages.map((page, index) => {
+              const multi_index = `${multi}-${index}`;
+              return (
+                <Tabs.Tab
+                  key={multi_index}
```
mordent-goonstation, line 71: You don't need to worry about the `key` of different reports being the same, `key` only has to be unique at the layer it's applying to. That is to say: I think you can get away with stripping out that multi prefix.

tgui/packages/tgui/interfaces/CrewCredits/type.ts
```diff
@@ -82,8 +82,45 @@ export interface ScoreItemProps {
...
+}
+
+interface PaperData {
+  name?: string;
+  sizeX: number;
+  sizeY: number;
+  text: string;
+  paperColor: string;
+  stamps: [string, number, number, number]
+}
```
mordent-goonstation, line 117: Refactoring PaperSheet to a .ts file is outside of the scope of your PR, but this type should definitely live over with it rather than here.

mordent-goonstation (changes requested):

code/datums/gameticker.dm
```diff
@@ -807,11 +807,9 @@ var/global/current_state = GAME_STATE_INVALID
 	SPAWN(0)
 		for(var/mob/E in mobs)
 			if(E.client)
-				E.verbs += /mob/proc/show_credits
-				if (E.client.preferences.view_tickets)
-					E.showtickets()
+				E.verbs += /mob/proc/show_round_statistics
 				if (E.client.preferences.view_score)
-					creds.ui_interact(E)
+					round_stats.ui_interact(E)
```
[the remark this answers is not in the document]
pgmzeta: going to split these into two options:
1. When should we pop up the credits window
2. What tab do you want to start on

tgui/packages/tgui/interfaces/CrewCredits/CitationsTab.tsx
```diff
@@ -0,0 +1,71 @@
...
+const CitationList = (props: CitationTargetData) => {
+  const { name, citations } = props;
+  return (
+    <Section title={name}>
+      {!!citations && citations.length && citations.map((citation, index) => {
+        return (
+          <Box
+            mb={2}
+            key={index}
+            dangerouslySetInnerHTML={{ __html: citation }}
```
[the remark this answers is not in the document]
pgmzeta: Right now the ticket text is generated as it will be read on paper in both places you can issue tickets:

https://github.com/goonstation/goonstation/blob/master/code/obj/item/device/pda2/smallprogs.dm#L1093
https://github.com/goonstation/goonstation/blob/master/code/obj/item/device/scanners.dm#L998

could rework `/datum/ticket` to hold its own ticket generation code and keep some data separate for the UI?

pgmzeta:
!merge_upstream

tgui/packages/tgui/interfaces/CrewCredits/CitationsTab.tsx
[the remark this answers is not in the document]
pgmzeta: sent the fields over and displayed those

code/datums/gameticker.dm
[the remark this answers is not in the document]
pgmzeta: reverted any settings changes. added logic to only open the window if there are tickets/fines to be seen.

code/datums/crew_credits/crew_credits.dm
```diff
@@ -4,6 +4,9 @@
 	var/list/crew_tab_data
 	var/list/antagonist_tab_data
 	var/list/score_tab_data
+	var/list/citation_tab_data
+
+	var/has_citation_data = FALSE
```
ZeWaka, line 9: why not just `if length citation tab data [ tickets] || length citation tab data [ fines]`

pgmzeta: d2fc640

ZeWaka (approved):

TobleroneSwordfish:
One small issue: 
image
encoding broken

PR #16423, opened 2023-10-13
Exptend pickup logging to allow logging of every pickup

flappybatpal:
## About the PR 
Expands pickup logging component to allow for logging of each time a pickup occurs.
Implements this component on all guns and batons.
Pickups by the same mob are filtered to avoid spam.

## Why's this needed? 
Improves admins ability to understand situations, "X stole my Y" is now a lot more clear cut.


pali6 (changes requested):
I think for clarity I'd use named arguments. I.e.
`first_time_only=TRUE, messge_admins_too=FALSE)`
instead of
`TRUE, "", FALSE)`

PR #16429, opened 2023-10-13
Fix respawned players not being clonescannable if old record still exists

### LINKED ISSUE #15400: Respawning players that have been clone scanned in their previous life cannot be scanned again
### Labels

[BUG][MAJOR]

### Description

If a player gets a clone scan then dies and respawns, the scanner will say they are scanned already if their old record is not removed.

### Steps to reproduce

1. clone scan someone
2. said person dies and respawns
3. try to clone scan them again
4. they show up as scanned until they remove the old characters record

### Expected Behavior

for the two characters to be separate after a respawn

Reported by: Piesuu
Active test merges: nothing
Round log date: 2023-08-15-22-38
Reported on: 2023-08-16 00:20:28
Map: COGMAP

TealSeer:
[BUG] [RESPAWNING]
## About the PR 

Fixes #15400 by adding code to delete cloning records of the previous character from consoles if the player is now respawning. The bug happens because clone records are looked up by ckey and thus trying to scan a player who was scanned in a previous life will still yield a record.

## Why's this needed? 

Bug

### DIFF
--- a/code/datums/controllers/respawn_controls.dm
+++ b/code/datums/controllers/respawn_controls.dm
@@ -191,6 +191,13 @@ var/datum/respawn_controls/respawn_controller
 		//try to break all links with the previous body so we don't get pulled back by changeling absorb, cloning etc.
 		usr.mind = null
 
+		// remove old cloning records because they are tied to ckey/mind and can cause problems
+		for_by_tcl(C, /obj/machinery/computer/cloning)
+			var/datum/db_record/old = C.find_record(src.ckey)
+			if(!isnull(old))
+				C.records.Remove(old)
+				qdel(old)
+
 		var/mob/new_player/M = new()
 		M.adminspawned = 1
 		M.is_respawned_player = 1


mordent-goonstation (commented):
I agree this is a bug that could be fixed, but I think this PR will just introduce some more weirdness around "why are random cloning records getting deleted?"

The closest thing I have for a "fix" is flagging an old clone record as "stale" in terms of not conflicting, but persisting it in the data.

You can't use things like name or even DNA as a check for if they're a new character or not, because those could be changed for a given character easily enough.

TealSeer:
your concerns are valid, gonna attack this a different way

pgmzeta:
What about clearing the `ckey` record field held in the scanner on DNR or Respawn instead of wiping the record completely? Could add with a check in cloning.dm to explicitly mark as 'mental interface failure'.

Would need to do the same on any existing cloning disks (yuck) to super-avoid being yanked back from a file.

TealSeer:
@pgmzeta after talking with pali im gonna change it so clone records are looked up by mind instead, should fix this bug and make things better in general

TealSeer:
going to make a new PR because of the scope change

PR #16430, opened 2023-10-14
Lowers the Maximum Players for Atlas to be in Rotation from 30 to 20

wolfolotl:
[player actions] [balance]
## About the PR 
Bumps the pop threshold where the map Atlas can be voted for down to 20 players or less, from 30 players or less.

## Why's this needed? 
Atlas is evidently not built to support 30 players at the same time. I am making this PR right after experiencing an Atlas round with 30+ people on it, and I think player experience would greatly benefit from narrowing the probability of this happening again.

### DIFF
--- a/code/map.dm
+++ b/code/map.dm
@@ -29,7 +29,7 @@ var/global/list/mapNames = list(
 	"Donut 2" =				list("id" = "DONUT2",		"settings" = "donut2",			"playerPickable" = TRUE,	"MaxPlayersAllowed" = 80),
 	"Donut 3" =				list("id" = "DONUT3",		"settings" = "donut3",			"playerPickable" = TRUE, 	"MinPlayersAllowed" = 40),
 	"Kondaru" =				list("id" = "KONDARU",		"settings" = "kondaru",			"playerPickable" = TRUE,	"MaxPlayersAllowed" = 80),
-	"Atlas" =				list("id" = "ATLAS",		"settings" = "atlas",			"playerPickable" = TRUE,	"MaxPlayersAllowed" = 30),
+	"Atlas" =				list("id" = "ATLAS",		"settings" = "atlas",			"playerPickable" = TRUE,	"MaxPlayersAllowed" = 20),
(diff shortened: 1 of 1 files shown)


mordent-goonstation:
I disagree that 20-30 player Atlas is not viable. It is simply a different set of challenges presented compared to other maps. I think the root issue stems from the vote being carried out at roundstart and the *next* round being the ones to deal with it, which means that if a 25-player vote (with a subset of them actually voting) has the result of Atlas that would be okay for them, but then 20 more people join the game between the two and so you end up with *40* player Atlas (or whatever), as well as the round gradually getting "worse" as latejoins swell the numbers.

I don't *mind* this going through, it certainly wouldn't hurt the above problem, but I think it's fighting the symptom rather than the root cause.

wolfolotl:
> I disagree that 20-30 player Atlas is not viable. It is simply a different set of challenges presented compared to other maps. I think the root issue stems from the vote being carried out at roundstart and the _next_ round being the ones to deal with it, which means that if a 25-player vote (with a subset of them actually voting) has the result of Atlas that would be okay for them, but then 20 more people join the game between the two and so you end up with _40_ player Atlas (or whatever), as well as the round gradually getting "worse" as latejoins swell the numbers.

yeah, this is essentially what I meant with this pr. 25-30 pop on atlas is cramped but _manageable_ but even then, certain areas like robotics become agonizing to try doing anything in if there's just 2 or 3 people in it at the same time. I'm quite fond of atlas, but it doesn't have the tightest considerations towards staffing slots or space allocation, and the more people happen to be on, the more you start being able to feel it

I have no way of automatically disabling an upcoming atlas round if the server pop passes the threshold, nor do I think that would really be a good thing to implement. so this pr doesn't really PREVENT Atlas With Too Many People rounds, it just makes them a lot less likely to happen unintentionally

TheTyrant25 (approved):

TealSeer (approved):

TheTyrant25:
yippee

PR #16440, opened 2023-10-14
Usage of API 2 for job bans

Wirewraith:
## About the PR 

Change job ban system to use the new API system


code/modules/admin/admin.dm
```diff
@@ -704,7 +704,7 @@ var/global/noir = 0
 			if (src.level >= LEVEL_SA)
 				var/M = href_list["target"]
 				var/job = href_list["type"]
-				var/list/cache = apiHandler.queryAPI("jobbans/get/player", list("ckey"=M), 1)[M]
+				var/list/cache = jobban_get_for_player(M)
```
ZeWaka, line 707: usage does not match definition, i assume M is a mob not a ckey

ZeWaka (changes requested):

code/modules/admin/admin.dm
Wirewraith: It's not, it's a ckey apparently

ZeWaka: please change it from M then

Wirewraith: No, out of scope

ZeWaka (approved):

PR #16441, opened 2023-10-14
Removes securitron whistle from the brobocop module

Chatauscours:
[objects] [balance]

## About the PR 
Removes securitron whistle from the brobocop module, replaces it with a normal one.

## Why's this needed? 
Not being able to drag securitrons while they are active anymore, borgs just began dragging them around with the securitron off, and quickly turning it on and whistling when an arrest needs to be made. This is annoying to see both as  security and antags, since it shifts the scale unevenly towards security. 

## Changelog 

I am unsure whether one is needed or not.


444Portal:
Really, really sad to see this having to be PR'd, but, if it's gotten to this point where it's been PR'd, it's probably best. A shame as well considering I got some neat use out of it that wasn't doing this.

Issue #16438, opened 2023-10-14
Brobocop whistles trigger the metal detectors


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Brobocop whistles trigger the metal detectors

### Steps to reproduce

1) Look at a brobocop going through a metal detector
2) Hear the detector go off

### Expected Behavior

Whistle not being detected when in a brobocop.

### Additional Information & Screenshots

Reported by: Chatauscours
On server: Goonstation 3 Roleplay: Morty
Active test merges: #14886 and #16296
Round log date: 2023-10-14-11-33
Reported on: 2023-10-14 12:20:16
Map: OSHAN

Chatauscours:
Fixed with #16441

PR #16444, opened 2023-10-14
Makes security scanner check silicon IDs for contraband permits.

urotsuki-ln:
## About the PR

Fix for issue [#16438](https://github.com/goonstation/goonstation/issues/16438)
Makes security scanners check if non-human targets are silicons, and if they are, check their ID for contraband permits.

### DIFF
--- a/code/obj/machinery/secscanner.dm
+++ b/code/obj/machinery/secscanner.dm
@@ -176,8 +176,8 @@ TYPEINFO(/obj/machinery/secscanner)
 		if(src.emagged)
 			return rand(99,2000) //very high-end bad vibes level assessor
 
-		var/has_carry_permit = 0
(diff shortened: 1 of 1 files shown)


code/obj/machinery/secscanner.dm
```diff
@@ -176,12 +176,24 @@ TYPEINFO(/obj/machinery/secscanner)
 		if(src.emagged)
 			return rand(99,2000) //very high-end bad vibes level assessor
 
+		var/has_carry_permit = 0
```
pgmzeta, line 179: TRUE/FALSE defines please

```suggestion
		var/has_carry_permit = FALSE
```

urotsuki-ln: Right! Done.

PR #16449, opened 2023-10-14
Rewrite cloning to lookup by mind instead of ckey

TealSeer:
[INTERNAL] [MEDICAL] [CODE QUALITY]
## About the PR 

Rewrites all cloning related code to use `mind` as the primary index for all lookups and related stuff, cuts out as much usage of ckeys as possible. Summary of changes:

- Makes the `/datum/mind` `transfer_to` proc set mob's `oldmind` var (needed for cloning changes)
- All cloning UI related calls use a unique record ID number as an arg instead of the record ckey
- Change ID number generation to be based off mind ref so it's actually unique (was using real_name before)
- replace `find_record(ckey)` with `find_record_by_id` (only used for UI calls) and `find_record_by_mind` (the primary way to look up a cloning record)
- Tweak an azone to use `find_record_by_mind`
- Makes new proc `eligible_to_clone(mind)` which checks all the conditions that `find_ghost_by_key` checked, because the latter isn't used in this anymore.
- Eliminates a few instances of duplicate code with usage of `eligible_to_clone`

## Why's this needed? 

Should squash some outstanding bugs, makes cloning code less error-prone in general.


mordent-goonstation:
tgui is fine. Not reviewed the rest.

TealSeer:
!merge_upstream

TealSeer:
!merge_upstream

PR #16452, opened 2023-10-15
Tweaking to try to improve icon button margins

mordent-goonstation:
[BUG][UI]

## About the PR 

I'm sure this will probably break a few other places that have specifically overridden it already, but I cannot divine the reason for this margin to exist and it seems to be the cause of some issues I'm seeing with buttons that *only* have an icon being strangely sized.

## Why's this needed? 

Changes buttons that look like this:

image

Into buttons that look like this:

image


DisturbHerb:
have we considered an upstream PR? would be worth asking tg maintainers why da heck it exists (and why there isn't an override prop)-

waka merged this the moment i started typing

PR #16468, opened 2023-10-15
Fix sponges wiping down ability buttons

### LINKED ISSUE #12798: Wipe down UI
### Labels

[BUG][MINOR]

### Description

pressing the toggle flashlight button on the Hud while holding a wet sponge will wipe it down

### Steps to reproduce

1. have a flashlight in your inventory
2. Hold a wet sponge
3. Press the flashlight button in the top left

### Expected Behavior

I expected my flashlight to turn on without being wiped by a sponge

Reported by: YoshiManden
Round log date: 2023-01-17-19-22
Reported on: 2023-01-17 20:05:20
Map: DONUT2

TealSeer:
[BUG] [PLAYER ACTIONS]
## About the PR 

Fixes #12798 with a type check on sponge `afterattack`

## Why's this needed? 

/obj/ability_button my beloathed

### DIFF
--- a/code/obj/item/mops_cleaners.dm
+++ b/code/obj/item/mops_cleaners.dm
@@ -592,6 +592,9 @@ TRASH BAG
 	if (!src.reagents)
 		return ..()
 
+	if (istype(target, /obj/ability_button))
+		return
+
 	var/list/choices = src.get_action_options(target)
 
 	if (!length(choices))


code/obj/item/mops_cleaners.dm
```diff
@@ -592,6 +592,9 @@ TRASH BAG
 	if (!src.reagents)
 		return ..()
 
+	if (istype(target, /obj/ability_button))
```
ZeWaka, line 595: not a correct fix, this won't work for other ui elements

ZeWaka (changes requested):

code/obj/item/mops_cleaners.dm
TealSeer: @ZeWaka the bug doesn't happen on other UI elements, tested it on everything in the bottom bar, health/temp/stamina, and proper ability buttons. This is a special case because /obj/ability_button is what it is and not /atom/moveable/screen

ZeWaka: why isn't it

TealSeer: just never got moved over, I actually found a comment in the code from you saying it should be moved over but it was out of your depth :P I'd like to try refactoring it in the future but in the mean time this was a more immediate fix

ZeWaka: i think it should just be refactored, this is a bit of a bandaid on a festering wound

TealSeer: o7

PR #16473, opened 2023-10-15
Fixes fires not processing on singletons

Laboredih123:
[ATMOSPHERICS][BUG]
## About the PR 
Fixes fires not processing on singletons by adding a proc that handles adding fires and making it so that an active fire causes the tile to be processed.

## Why's this needed? 
It is good that fires can process, not good that they can't.


ZeWaka (approved):
guh

PR #16474, opened 2023-10-15
Usage of api2 for gauntlet matches

Wirewraith:
## About the PR 

Use api2 stuff for the gauntlet thing that retrieves how many gauntlets a player has completed


ZeWaka (approved):
guh

PR #16477, opened 2023-10-15
Caustics Parallax

Mister-Moriarty:
[UI] [Feature]

## About The PR:
Implements an undersea caustics parallax render source for use on Oshan. This layer is not present in the Trench, so to imply that the Trench is at a depth where starlight scarcely reaches it. This is also true for Nadir.

See demonstration here: https://discord.com/channels/182249960895545344/890231630097182720/1158151755306111037

Several new areas have been added to Oshan so to accommodate for parallax occlusion overlays, such as the sunken asteroids and derelict diner.

## Why Is This Needed?
Oshan deserves some parallax layers too.

## Changelog:
```changelog
(u)Mr. Moriarty
(*)Oshan now has it's own parallax layer, depicting an undersea caustics pattern.
```


Laboredih123:
yo, image?

Mister-Moriarty:
> yo, image? i wanna see the funny lights

Restricted once again by the 10MB file size limit.
Have a Discord link instead.

ZeWaka:
i personally think they're too bright/obvious/noticable

TobleroneSwordfish:
Agree, it looks amazing but could do with being just a little bit more subtle.

Mister-Moriarty:
> i personally think they're too bright/obvious/noticable

In the demonstration, the obsever's larger light cone results in the layers appear brighter. Wthout the light code, the effect is much more subtle.

image

flappybatpal:
https://github.com/goonstation/goonstation/assets/70909958/e95456ae-948c-4fea-87e9-5e0423bf6735

Two things
1. It shows over the sea elevator, this area is pathed as shuttle for some reason. If we are making map changes might as well adjust this here too.
2. Shows over trench holes. Not quite sure of this, on one hand it looks quite pretty but on the other it's a hole so nothing for the light to hit?

TobleroneSwordfish:
> In the demonstration, the observer's larger light cone results in the layer appearing brighter. Without the light cone, the effect is much more subtle.

I've tested it on local, yep, and I still think the strongest parts of the animation are just a little too obvious. Remember, Oshan has a daylight cycle (even if it's pitch black for most of UK time) so it will show up a lot more brightly during the day.

flappybatpal:
New changes are good. Uncertain about client performance implications of the large sprites, there seems to be more hitching but this is not a scientific analysis.

AXLFAN2000:
This looks SUPER gorgeous ingame, and I had no performance issues with it in the classic round that just ended

PR #16485, opened 2023-10-16
changes the consume verb to be more noticable

Croissant-Man:
[QOL] [PLAYER ACTIONS]

## About the PR
Changes the hidden "consume" verb to allow it to be noticable beyond it showing two messages in the chat 

## Why's this needed? 
The novelty of knowing this verb exists quickly wears off when you do it in front of several people and get no reaction.

```changelog
(u)Croissant_Man
(+)Consuming a human now makes sounds and is bloodier (on living targets)
(+)Attempting to eat a skeleton/skeletonized corpse will now have negative results for your teeth
(+)Rotting corpses don't taste very good and will now give you food poisoning before ending the consumption early
```


Tarmunora:
We're not entirely sure what we want to do with the `consume` verb, but "increasingly graphic cannibalism" is not on the list

Croissant-Man:
Can’t say I’m surprised

PR #16498, opened 2023-10-16
Use api2 for vpn checking

Wirewraith:
## About the PR 

Use API 2 for vpn checking and whitelist management


code/modules/goonhub/api/routes/vpn/AddVpnWhitelist.dm
```diff
@@ -0,0 +1,13 @@
+/// POST /bans
```
ZeWaka, line 1: bonk

ZeWaka (approved):

PR #16499, opened 2023-10-16
Makes tinfoil hats prevent artifact hallucinations

itchybones:
[SCIENCE][FEATURE][CLOTHING]
## About the PR 
This PR makes #16411's madness artifact's hallucination effect ignore people with tinfoil hats.

## Why's this needed? 
I believe this to be a neat gimmick, and another use for the tinfoil hats, that, as of now, only block weak martian psyblasts.
This PR also adds a **potential** counter to the hallucinations, although I don't think the hats can be aquired without having
the conspirator trait.

## Changelog 

```changelog
(u)Chasu
(+)Tinfoil hats now block artifact hallucinations.
```

### DIFF
--- a/code/obj/artifacts/artifact_machines/madness.dm
+++ b/code/obj/artifacts/artifact_machines/madness.dm
@@ -349,8 +349,8 @@
 			var/turf/T = get_turf(O)
 			T.visible_message("<b>[O]</b> shimmers briefly!")
 
-		for (var/mob/living/L in range(range,O))
-			if (!L.client) //no point hallucinating if there's nobody to see it
+		for (var/mob/living/carbon/human/L in range(range,O))
+			if (!L.client || istype(L.head, /obj/item/clothing/head/tinfoil_hat)) //no point hallucinating if there's nobody to see it
 				continue
 
 			for(var/list/comp_args_tuple in src.madness_effects[src.effect_type])


code/obj/artifacts/artifact_machines/madness.dm
```diff
@@ -349,8 +349,8 @@
 			var/turf/T = get_turf(O)
 			T.visible_message("<b>[O]</b> shimmers briefly!")
 
-		for (var/mob/living/L in range(range,O))
-			if (!L.client) //no point hallucinating if there's nobody to see it
+		for (var/mob/living/carbon/human/L in range(range,O))
+			if (!L.client || istype(L.head, /obj/item/clothing/head/tinfoil_hat)) //no point hallucinating if there's nobody to see it
```
amylizzle, line 353:
```suggestion
		for (var/mob/living/L in range(range,O))
			if (!L.client) //no point hallucinating if there's nobody to see it
				continue
			var/mob/living/carbon/human/H  = L
			if(istype(H) && istype(H.head, /obj/item/clothing/head/tinfoil_hat)) 
```

amylizzle (changes requested):
Love the idea, but recommend a tweak to the implementation

itchybones:
Thanks for the suggestion and making the code easier to understand!

amylizzle (approved):

PR #16530, opened 2023-10-18
Removes additional access from Security Officers and Heads of Security on the RP servers.

DisturbHerb:
[BALANCE][INPUT][ROLEPLAY]

## About the PR 

Removes the additional access levels granted to Security Officers and Heads of Security on the RP servers.

## Why's this needed? 

The discourse surrounding this subject has been ongoing for some time, with the latest round of discussion occuring on [this forum thread.](https://forum.ss13.co/showthread.php?tid=21885)

While my thoughts on the matter are self-evident due to me being the progenitor of the latest forum thread, it is important that ample opportunities for discussion and testing be provided for before considering a change such as this one. The dynamic between Security, the rest of the crew, and the antagonists will be necessarily impacted.

\* Of all the potential options to address this issue, this is the most extreme.

Undoubtedly, this change will be subject to a high degree of scrutiny due to the varying differences in opinion among the playerbase. This solution may not end up being the best way to address the concerns of those who believe that the status quo is unsatisfying; in fact, it represents a sea change in how Security Officers will operate on RP in contravention of the previously established precedent. I cannot come out and determine at this stage if the new state of affairs will be better than what we presently have. However, that there is a perception that there is an issue in the first place warrants some concrete data on how such a change would actually impact the game. I simply ask that any discussion surrounding this matter remain civil and productive.

## Changelog 

```changelog
(u)DisturbHerb
(*)Security Officers on RP now have equivalent access to those on the Classic servers.
```

### DIFF
--- a/code/obj/machinery/computer/card.dm
+++ b/code/obj/machinery/computer/card.dm
@@ -646,18 +646,9 @@
 	department = 4
 	req_access = list(access_maxsec)
 
-	#ifdef RP_MODE // fuckin RP mode giving secoffs more access *grumble grumble*
-	civilian_access_list = list(access_morgue, access_maint_tunnels, access_chapel_office, access_tech_storage, access_bar, access_janitor, access_crematorium, access_kitchen, access_hydro, access_ranch)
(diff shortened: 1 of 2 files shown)


DisturbHerb:
oh i forgot that HoS gets AA on rp, oopsie

PR #16531, opened 2023-10-18
Fixes #16279: Allowing critters to cryo

### LINKED ISSUE #16279: can't cryo as plushie
### Labels

[BUG][MINOR]

### Description

plushies can't enter the cryo sleepers

### Steps to reproduce

1. be plushie
2. click on cryo sleeper
3. no response
4. i just wanna observe again lemme free up the slot

### Expected Behavior

be able to do that

Reported by: RDCB
Active test merges: #16028 and #16185
Round log date: 2023-10-05-14-43
Reported on: 2023-10-05 15:17:20
Map: COGMAP

mloccy:
[Bug] [Critters] [Game Objects] [QOL]
## About the PR 
Allows critters to cryo the same way other players can, and as a result fixes #16279 

## Why's this needed? 
Special event critters occupy a round slot, and players should be encourage to release that slot if for whatever reason they need to leave by using the cryo mechanic, this PR allows critter players such as plush toys to cryo like everyone else. 

Tests identified
- [x] Human self-cryo
- [x] Human other player insertion (grab)
- [x] Human other player insertion (drag)
- [x] Silicon self-cryo
- [x] Silicon other player insertion (drag)
- [x] Critters with hands self insertion
- [x] Critters without hands
- [x] AI cannot cryo
- [x] Ghosts / Observers cannot cryo


code/obj/cryotron.dm
```diff
@@ -318,29 +318,31 @@
...
-		if(isgrab(user.l_hand))
+	/// Helper to check if a player is trying to insert another player
+	proc/cryo_attack_hand(var/mob/user)
+		if (isgrab(user.l_hand))
 			src.Attackby(user.l_hand, user)
-		else if(isgrab(user.r_hand))
+		else if (isgrab(user.r_hand))
 			src.Attackby(user.r_hand, user)
-		else if (!enter_prompt(user))
-			return ..()
```
TheTyrant25, line 327: I don't think you can just replace `attack_hand` like this. Why not just put it in the regular `attack_hand`? Why was a new proc needed?

code/obj/cryotron.dm
```diff
@@ -318,29 +318,31 @@
...
-			return FALSE
-		if (!enter_prompt(user))
-			return ..()
+		else
+			enter_prompt(user)
+			return
 
-	attackby(var/obj/item/W, var/mob/user)
+	/// Helper to check if a player is trying to insert another player
+	proc/cryo_attackby(var/obj/item/W, var/mob/user)
```
TheTyrant25, line 332: you've done the same thing here.

code/obj/cryotron.dm
```diff
@@ -361,9 +363,58 @@
...
+		insert_prompt(target, user)
 		return ..()
 
+	/// Override for handling all interactions with the cryo chamber
+	Click()
+		if (!usr)
+			return
+
+		if (isAIeye(usr))
+			boutput(usr, "<span class='alert'>An incorporeal manifestation of an artificial intelligence's presence can't enter \the [src]!</span>")
```
TheTyrant25, line 389: can probably skip this bouput line. No need for it.

code/obj/cryotron.dm
```diff
@@ -361,9 +363,58 @@
...
+			return
+
+		if (isobserver(usr))
+			return
+
+		if (isintangible(usr))
+			return
+
+		if (issilicon(usr))
+			enter_prompt(usr)
```
TheTyrant25, line 408: instead of using multiple ifs chained together, you could probably combine them into one if statement.
`if (this || that || thisthing || otherthing)`

code/obj/cryotron.dm
```diff
@@ -361,9 +363,58 @@
...
+
+		if (in_hand_item != null)
+			src.cryo_attackby(in_hand_item, usr)
+
+		else if (usr.equipped_limb() != null && in_hand_item == null)
+			src.cryo_attack_hand(usr)
+		else
+			enter_prompt(usr)
+
+		. = ..()
```
TheTyrant25, line 420: kind of strange to put this at the end? `. = ..()` means roughly "return `..()` when the proc ends", so you could just put `..()` instead.

TheTyrant25 (changes requested):
I'm not sure this code will work correctly as is. Best to get a developer to weigh in on it.

code/obj/cryotron.dm
TheTyrant25: I know it existed before but I think just nothing happening when the AI clicks on it is good enough honestly.

code/obj/cryotron.dm
mloccy: The reason I'm removing the attack_hand/attackby/attack_ai is because the yare now handled by Click. The Click implementation was done initially to allow critters without hands to cryo, but that cause the issue that now critters and mobs with hands would get the prompt twice, once in attack_by and once in Click. 

So the solution could be to either 1. Keep both and type filter in Click which is kinda nasty as it requires you to keep a list of types to filter out, or 2. Remove the implementations of attack_hand / attackby / attack_ai and route them through Click. The methods that remain are now local to cryotron (cryo_attack_hand, cryo_attackby). 

I kept them this way so the code would read a bit better, but I could inline the functionality to the caller (Click) if desired.

TheTyrant25: Ah, I see what you mean. Inlining it into the Click proc may make it clearer, yes. The naming of those procs is a bit... ambiguous. At least to me.

code/obj/cryotron.dm
mloccy: Removed

code/obj/cryotron.dm
mloccy: Grouped up some of the contracts

code/obj/cryotron.dm
mloccy: Done

code/obj/cryotron.dm
```diff
@@ -361,9 +342,65 @@
 		if (!exit_prompt(user))
 			return ..()
 
+	/// Handling dragging players in to cryo, mainly for silicon players.
 	MouseDrop_T(atom/target, mob/user as mob)
-		if (ishuman(target) && isrobot(user) && BOUNDS_DIST(src, user) == 0 && BOUNDS_DIST(src, target) == 0 && BOUNDS_DIST(user, target) == 0)
-			insert_prompt(target, user)
+		if (!ishuman(target) && !isrobot(target))
```
ZeWaka, line 347: not the right logic

ZeWaka (changes requested):

code/obj/cryotron.dm
mloccy: I believe I have fixed this now, would appreciate another look when you have the chance, appologies for the delay I haven't been able to get back to this for a while.

Tarmunora (approved):

PR #16535, opened 2023-10-18
Make latejoin antags not target VR people

### LINKED ISSUE #10850: Players in VR can be chosen as latejoin antag targets
### Labels

[BUG]

### Description

^

### Steps to reproduce

1. Have someone in VR
2. Latejoin as antag

### Expected Behavior

I expected all assassination targets to actually exist

image

TheTyrant25:
[BUG] [VR] [GAMEMODE]
## About the PR
Adds a check that makes sure that the "possible targets" internal list doesn't include people who are currently inhabiting vr human bodies.

## Why's this needed? 
Fixes  #10850

### DIFF
--- a/code/datums/objective.dm
+++ b/code/datums/objective.dm
@@ -49,6 +49,8 @@ ABSTRACT_TYPE(/datum/objective)
 					continue
 				if (!possible_target.current.client)
 					continue
+				if (isvirtual(H) || istype(get_area(H),/area/afterlife))
+					continue
 				possible_targets += possible_target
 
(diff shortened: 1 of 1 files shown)


flappybatpal:
It's probably better as
`if (isvirtual(H) || istype(get_area(H),/area/afterlife))`

Taken from spies, from a look we don't actually seem to have a check for `isafterlife()` which we probably should.

TheTyrant25:
we do have `inafterlife`, but i'm not entirely certain on what it does. Could we use that?

PR #16540, opened 2023-10-18
Capybaby

amylizzle:
## About the PR 
Capybaby
image

## Why's this needed? 
Capybaby

## Changelog 

```changelog
(u)Amylizzle & Waffleloffle
(+)Capybaras should not be alone, and so the pool room capybara now has a baby. Sprites by Waffleloffle.
```


wolfolotl:
I could make some bespoke sprites for the baby, if you like. just gotta know if it needs to sit too

amylizzle:
Could potentially but I'm honestly quite pleased with how just scaling it down looks!

pali6:
I don't think the scaled down version looks good personally.

amylizzle:
Oh no, in that case @wolfolotl yeah bonus sprites would be great. Sitting would be required.

pali6:
At the very least a proper outline 🥺

pali6:
!merge_upstream

pali6 (approved):

PR #16542, opened 2023-10-18
Fix temperature math for adding preexisting reagent

TDHooligan:
[BUGFIX] [CHEMISTRY]
## About the PR 

An oversight in the temperature/heat capacity math currently counts the existing chemical twice when calculating the heat capacity and reagents. 

To put it simply - if you add 0.01u of room-temperature coffee to 1000u of hot coffee, the temperature of the resulting mixture is basically halfway between the two.


TobleroneSwordfish (approved):
I am at least 70% confident this makes more sense 👍

PR #16543, opened 2023-10-18
Simplifies neurodepressant's and solipsizine's recipe

colossusqw:
[LABEL][chemistry][balance]

## About the PR 
This PR changes the recipe of Neurodepressant, and the only chem that also requires it as a precursor to be produced, solipsizine. It does it by removing diphenhydramine from solipsizine's recipe, and swapping the acetone in the neurodepressant recipe with another part of sulfuric acid.

## Why's this needed? 
The chemistry recipe update has made a lot of changes, and a few chemicals that were already hard to made were made even more so, particularly neurodepressant and solipsizine(partly on account of needing the former as a precursor).

Neurodepressant currently needs both acetone and sulfuric acid, the first being significantly harder to produce now, and the second being quite volatile to work with. Neurodepressant is still not close to being easy to make, and overall more dangerous considering the sulfuric acid offgass, however it should be closer to how time and resource consuming it was originally to make.

```changelog
(u)Colossusqw
(+)Solipsizine's recipe no longer requires diphenhydramine.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -3167,7 +3167,7 @@
 		name = "neurodepressant"
 		id = "neurodepressant"
 		result = "neurodepressant"
-		required_reagents = list("acid" = 1, "neurotoxin" = 1)
+		required_reagents = list("acid" = 1, "neurotoxin" = 1, "acetone" = 0.1)
 		result_amount = 1
 		min_temperature = T0C + 450
 		mix_phrase = "The neurotoxin breaks down, bubbling violently."
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -3167,8 +3167,8 @@
 		name = "neurodepressant"
 		id = "neurodepressant"
 		result = "neurodepressant"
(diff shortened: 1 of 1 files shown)


colossusqw:
> Changing the antihistamine is probably fine, but I'd keep the acetone

would it make sense to do something similar to how the proposed decaffeination process is? As in the acetone being needed in a smaller amount as a catalyst instead of a big volume ingredient? I understand not wanting the chem to be easily available but it did get a big jump in time investment to make..

code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -3167,8 +3167,8 @@
 		name = "neurodepressant"
 		id = "neurodepressant"
 		result = "neurodepressant"
-		required_reagents = list("acid" = 1, "neurotoxin" = 2, "acetone" = 1)
-		result_amount = 2
+		required_reagents = list("acid" = 1, "neurotoxin" = 1, "acetone" = 0.1)
+		result_amount = 1
```
Tarmunora, line 3171:
```suggestion
		required_reagents = list("acid" = 1, "neurotoxin" = 2, "acetone" = 1)
		result_amount = 2
```

Tarmunora (approved):

PR #16544, opened 2023-10-18
Fixes #16377 Plush Toys Spawning with Nectar Perspiration 

### LINKED ISSUE #16377: Bee gene attacks plusies
### Labels

[BUG][TRIVIAL]

### Description

You can spawn as a plushie with a gene active if you have the random gene trait on

### Steps to reproduce

Have the random gene active trait
Join the game as a plush
get stung by an invisible bee cloud

### Expected Behavior

idk not invisible bees

Reported by: EmeraldCrow
Active test merges: #14886
Round log date: 2023-10-10-01-21
Reported on: 2023-10-10 02:20:22
Map: NADIR

mloccy:
[Bug] [Critters]
## About the PR 
Small fix to remove Nectar Perspiration if they spawn with it, fixes #16377 

## Why's this needed? 
It has little counterplay as plush toys aren't going to find it easy to get mutadone, or even attention from doctors most of the time and the normal bee cloud effect doesn't seem to render on their sprites.

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -2189,7 +2189,8 @@ ABSTRACT_TYPE(/datum/job/special/halloween/critter)
 		..()
 		if (!M)
 			return
-		// Deactivate any gene that was activated by
+
+		// Deactivate any gene that was activated by Mildly mutated trait
 		M.bioHolder.DeactivateAllPoolEffects()
 		M.critterize(/mob/living/critter/small_animal/plush/cryptid)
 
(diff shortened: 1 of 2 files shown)


code/datums/jobs.dm
```diff
@@ -2189,6 +2189,7 @@ ABSTRACT_TYPE(/datum/job/special/halloween/critter)
 		..()
 		if (!M)
 			return
+		M.bioHolder.RemoveEffect("buzz")
```
ZeWaka, line 2192: this really looks like a bandaid and not the actual fix, just make random genes not apply to ghost respawn shit like this

ZeWaka (changes requested):

code/datums/jobs.dm
mloccy: Makes sense, could a solution to this be to do M.bioHolder.RemoveAllPoolEffects()? Looking at how traits work it seems as though they're applied before getting to the job setup code.

mloccy: I think I have the fix now, I'm deactivating anything that was activated from effectsPool prior to spawn, and I extended the fix to the rest of the halloween critters too.

PR #16558, opened 2023-10-20
Use api2 for dectalk

Wirewraith:
## About the PR 

john madden


code/modules/goonhub/api/models/dectalk/DectalkPlayResource.dm
```diff
@@ -0,0 +1,20 @@
+
+
+/// DectalkPlayResource
+/datum/apiModel/DectalkPlayResource
+	var/audio = null // integer
```
ZeWaka, line 5: integer? really?

code/modules/goonhub/api/models/dectalk/DectalkPlayResource.dm
```diff
@@ -0,0 +1,20 @@
+
+
+/// DectalkPlayResource
+/datum/apiModel/DectalkPlayResource
+	var/audio = null // string
```
ZeWaka, line 5: what kind of string? a url? an id? usage is not clear

ZeWaka (approved):

PR #16562, opened 2023-10-20
Make Security RecordTrak search by real_name

### LINKED ISSUE #15727: Placing a 1st Place or some such ribbon on someone makes a duplicate Security Record for them
### Labels

[BUG][TRIVIAL]

### Description

If you have a 1st place or some such ribbon placed on you, it seems to automatically create a second Security record for you. This can cause problems, such as being unable to be taken off an arrest status with a handheld record scanner, necessitating having to manually find the correct record (not the duplicate) on the Sec Computer to change an arrest status.

### Steps to reproduce

1. Place a 1st place ribbon on yourself
2. Somehow get an arrest status put on you (could be from the jailbird trait, reading Beepsky's diary, etc.)
3. Attempt to reset your arrest status with a handheld record scanner. If a SecHUD is worn, your status won't change.

### Expected Behavior

Using a handheld record scanner or a computer to reset an arrest status should reset it rather than doing nothing due to two existing records now existing.

During a round, the MD had a 1st place ribbon on them and also had the Jailbreak trait. When I tried to set their status to none with both the computer and the handheld scanner, it turns out they had two different records, one with their name and one with a 1st place indicator plus their name. Attempting to use the handheld scanner on them did nothing and I had to find the correct record on the computer to change it.

Reported by: ThatNavigator
Active test merges: #15458
Round log date: 2023-09-07-13-43
Reported on: 2023-09-07 13:55:24
Map: KONDARU

TealSeer:
[BUG] [GAME OBJECTS]
## About the PR 

Fixes #15727 by making the RecordTrak search by real_name instead of name

## Why's this needed? 

It should not be using src.name to determine whether to make a new security record or not


FlameArrow57:
Won't using real_name will go through masks? Like you could use an agent card with gas mask and it'll still know your name. Maybe only real_name if face is visible? Otherwise, idk, using name seems fun to scan in someone with an appearance name still

TealSeer:
@FlameArrow57 the problem stemming from this is the fact that it automatically creates new security records for people if it can't match names, when you just use name it leads to weirdness like a new record being created named "Unknown" if someone is wearing a gas mask and no ID (but it'd still magically pull a bunch of other info about them off their body like age and pronouns). 

The way I see it the fix is either to just have it check real_name (my rationale being if someone is using the prison scanner on you to set your arrest status you are probably already in custody, and chasing someone down and using this on them isn't exactly going to give you life changing information that you couldn't get with say a forensics scanner), or to remove the record creation functionality altogether, which may be better but seems to be a unique and potentially desirable behavior for things like stowaways.

TobleroneSwordfish:
I think it'd be better if it just errored if the face isn't visible, "hey you, take off your mask so I can scan you" is a good interaction anyway.

TealSeer:
Gonna go with that then, seems to satisfy both problems

TobleroneSwordfish (approved):
lgtm, should also prevent issues with labellers and such if those were an issue

PR #16570, opened 2023-10-20
fix custom soups runtiming

### LINKED ISSUE #16569: Soup does neither give food effects nor chemicals
### Labels

[BUG][MAJOR]

### Description

Soup does neither give food effects nor chemicals

### Steps to reproduce

1. have a soup like:
This is an ordinary hauntdog, ordinary hauntdog, and ordinary hauntdog soup in ephedrine and methamphetamine broth. 
Ah, the, uh, wonders of the kitchen stove. 
 This food has the following effects: Food (HP+++); Food HoT (All); Food HoT (Brute); 
2. eat it
3. no food effects or chems

### Expected Behavior

to have my soup be as awesome as i designed it

Maybe it is related to the gluttony rework somewhere runtiming.

Reported by: Lord_Earthfire
Active test merges: #14886
Round log date: 2023-10-20-08-36
Reported on: 2023-10-20 10:13:07
Map: COGMAP

M-Earthfire:
[catering][bug][]

## About the PR 

This PR adds an initial bites_left amount to custom soup. This stops custom soup from runtiming and makes custom soup able to fill the stomach of people.

## Why's this needed? 

Runtimes are bad. This fixes #16569

### DIFF
--- a/code/obj/soup_pot.dm
+++ b/code/obj/soup_pot.dm
@@ -12,7 +12,7 @@
 	icon_state = "bowl"
 	name = null
 	desc = "Ah, the, uh, wonders of the kitchen stove."
-	bites_left = null
+	bites_left = 6
 	heal_amt = null
 	initial_volume = null
 	initial_reagents = null


pali6:
Could you explain how it currently stays at null? `New` is supposed to overwrite it, if it is not doing that then that seems to be the root cause.

pali6:
Though frankly the argument order on the `New` is sus. Atoms should always have the first argument be their new `loc`

pali6:
Ah. This ugly thing. Likely the easiest solution is to introduce a proc/initial_bites() that has implementation which defaults to initial(src.bites_left) but can be overridden. That wouldn't require much changing of existing code. Or a var which gets initialized in snack/New to initial(src.bites_left).

PR #16580, opened 2023-10-21
Pod HUD shows fuel tank pressure

pgmzeta:
## About the PR 

* Re-uses the pod healthbar code to create a second fuel bar
* Adjusts health bar placement to fit both bars
* Add icons to more easily tell them apart
* Add checks for updating fuel on use / tank change
* Update pod fuel and health bars during pod spawn
* Change pod bar readability for lower values

## Why's this needed? 
Feedback about pod fuel levels, now that they use them

## Screenshots
Pod bar detail
image

Full UI
Dyna-Co Šid Transport Platform XV 2023-10-20 174610 (1)

## Changelog 

```changelog
(u)glowbold
(*)The Pod HUD shows the pressure of your fuel tank.
```


pali6:
I'm unsure what units are the best to show here. Technically the engine consumes a set amount of mols so those are the ultimate measure of track miles. But also how do you set a maximum value for that correctly. If you cool down the plasma enough then the amount of mols can go arbitrarily high, can't really cram that into a bar.

Pressure is on the other hand a measure that a device in a pod could reasonably measure. But it doesn't necessarily correspond to how much mileage you have left because temperature and the ideal gas law. Also the value you have there as maximum is somewhat arbitrary, with how it is currently set up you could end up with it above 100%. Unsure if the bar code handles that correctly and clamps it.

Perhaps the best idea would be to show mols of plasma as a percentage of what is in a default normal tank. If someone uses a custom made cooled down plasma tank with more mols in it they get display that shows e.g. 420% and the bar displays as full while above 100%, then drops normally.

BatElite:
If you wanted to be really extra about it, have >100% fuel act like a stacked boss health car with the different colours.

pgmzeta:
Currently in this PR, any fuel pressure level above 999 is capped to 999 using the existing health bar behavior. I think it's a reasonable accuracy slip for pressure values between 1014 and 999 to just show "999" on the gauge.

I agree that mols-of-plasma is the most accurate measurement of effective fuel, but using pressure and matching the Plasma Fuel vendor UI will be less confusing for the majority of pod users. For a current-day metaphor, a gas-powered car might run different on 87, 89, or 91 octane but nothing about those fuels would change what level the fuel displays. If the Plasma Fuel vendor switched to using mols, then 100% this should too.

For those who are working with custom fuels, I would hope they have an interest in the mechanics behind fuel and work out that it's a mol conversion. Perhaps what would ultimately be best is switching between alternative UI displays by clicking the icon ala the cyborg health display readout?

If mols are still the preferred measurement, I can implement it but do need to understand the math first. If you already know the math behind it please leave code suggestions; pre-emptive permission for anyone to adopt/take over/merge-and-then-patch.

kbsmilk:
Mileage is already going to vary wildly depending on what systems are installed, and whether the pilot has them enabled. Any way you slice this, it comes down to the pilot "getting a feel" for how much fuel/time they have left. Which any kind of gauge will let them do, as long as they look at it occasionally.
So I think it's best to just have the gauge show tank pressure, matching the value you're told by vendors and the tank's own gauge. Keeps it simple.

pali6:
For mols you'd actually require less math than for pressure. As mols are what the vars on gas datum store directly. But it seems like people want pressure so I'll merge it as is.

PR #16587, opened 2023-10-21
Adds support for syndicate items with different weights and changes not_in_crates/vr_allowed to use TRUE/FALSE

unfunnyperson:
## About the PR 
Makes surplus crates/spy bounties use weighted lists to pick items.
Makes vr_allowed/not_in_crates use TRUE and FALSE defines

Items have 50 weight by default.

Removes the 1/20 chance for syndicate in pipebombs to come with a mininuke or defended mininuke
Adds a surplus micronuke with 5 weight and 5 tc cost, Also adds a defended micronuke with 1 weight and 9 tc cost

Minor but adds vr_blacklist to the spy ohr to prevent 2 ohr's from appearing in the vr uplink

## Why's this needed? 
Allows cool items that shouldn't be an every round occurrence to exist such as the micronukes in surplus crates.  In addition, having an item with a 1/20 chance to come with a different rarer item felt weird to me.

### DIFF
--- a/code/datums/gamemodes/spy_theft.dm
+++ b/code/datums/gamemodes/spy_theft.dm
@@ -105,9 +105,9 @@
 				continue
 
 			if (S.cost <= value_high && S.cost >= value_low)
-				possible_items += S
+				possible_items[S] = S.surplus_weight
 
-		reward = pick(possible_items)
+		reward = weighted_pick(possible_items)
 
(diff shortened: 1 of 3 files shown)


Tarmunora (changes requested):
Weighting is a better solution for the micronuke, yeah.
However, the spythief OHR should retain the increased cost because it serves not as weighting but to filter it out of the reward list for low-tier bounties outright

PR #16588, opened 2023-10-21
Local dev fixes for api2

Wirewraith:
## About the PR 

Avoids runtimes during local development. Cleans up apiHandler a bit


ZeWaka:
linter

ZeWaka (commented):
good other than lint

PR #16591, opened 2023-10-21
clean up RobustLight2.dm

CheffieGithub:
## About the PR 
Pathing made fully absolute for the light datums, partially for atom stuff at the bottom.
`src `and `global` added for scope. (shouldn't effect behaviour just more readable)
moved `/atom/movable` mechanics related things to `atom.dm`.
`TRUE` / `FALSE` conversion.

## Why's this needed? 
Much easier to find things. Mechanics interactions don't belong in the robust lights file. Arguably more readable this way.

### DIFF
--- a/code/_lighting/RobustLight2.dm
+++ b/code/_lighting/RobustLight2.dm
@@ -1,4 +1,30 @@
 var/RL_Generation = 0
+var/RL_Started = 0
+var/RL_Suspended = 0
+
+proc/RL_Start()
+	global.RL_Started = 1
+	for (var/datum/light/light)
+		if (light.enabled)
+			light.apply()
+	for (var/turf/T in world)
+		LAGCHECK(LAG_HIGH)
+		RL_UPDATE_LIGHT(T)
+
+proc/RL_Suspend()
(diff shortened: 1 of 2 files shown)


Azrun:
What does "moved /atom/movable mechanics related things to atom.dm." have to do with robust lights... 🫠

TobleroneSwordfish (changes requested):
cooltext445710790206610

BatElite:
Well would you rather keep those vars in the robustlight code file then?

ZeWaka (approved):
thanks, this still needs to be split but this is a great start

CheffieGithub:
How much should it be split? The diff is massively inflated from what vsc said it was due to indent changes.

ZeWaka:
> How much should it be split? The diff is massively inflated from what vsc said it was due to indent changes.

not this pr, i mean splitting this file into multiple others, out of scope of this PR

TobleroneSwordfish (approved):
Oh sorry I didn't realise the mechanics stuff was actually *in* the robustlight file, that's awful

PR #16599, opened 2023-10-21
Change no damage crew objective description to be accurate

### LINKED ISSUE #13587: medical healself crew objective check doesn't match description
### Labels

[minor]

### Description

The objective says "Make sure you are completely unhurt when the escape shuttle leaves." but it actually checks when arriving at centcomm

### Steps to reproduce

1. Call the shuttle as a doctor with the healself objective
2. Be buckled in and completely healthy when the shuttle departs
3. Take damage on the shuttle
4. Fail the objective when you arrive

### Expected Behavior

 For the description to match the completion check, whichever of the two gets changed

_No response_

TealSeer:
[BUG] [UI]
## About the PR 

Fixes #13587 

## Why's this needed? 

Shouldn't lie

### DIFF
--- a/code/datums/crew_objective.dm
+++ b/code/datums/crew_objective.dm
@@ -804,7 +804,7 @@ ABSTRACT_TYPE(/datum/objective/crew/medicaldirector)
 				check_result = TRUE
 		return check_result
 /datum/objective/crew/medicaldirector/healself
-	explanation_text = "Make sure you are completely unhurt when the escape shuttle leaves."
+	explanation_text = "Make sure you are completely unhurt at the end of the round."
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (changes requested):
I think most objectives try to be at least slightly in character, so it should be something like "when the emergency shuttle docks at centcom"

TealSeer:
@TobleroneSwordfish I was just going off what all the other objectives around it are phrased like, they all say "at the end of the round"

Azrun:
I think "when the emergency shuttle lands at centcom" is preferential normally but rounds can end without a shuttle call.

PR #16602, opened 2023-10-22
Adds a plastic recipe through chemistry

MylieDaniels:
## About the PR 
Adds a chemical recipe for plastic sheets, which always requires at least 10 carbon, 20 hydrogen, and 10 oil to complete.
In addition, to stand in for a Phillips catalyst, we use chromium here. Without silicon dioxide, this chromium will deplete 40 units over the process.
Phillips catalysts sit on silica gel to prevent ashing, and to follow with this, silicon dioxide prevents the consumption of the catalyst by being consumed at a lower rate (2u per sheet)
If there's magnesium chloride present, which was discovered to increase polyethylene yields in the 1970s, it will prevent the carbon and hydrogen from consuming each other (you can also use fuel for this) and will speed the reaction up linearly over time - but 2u will be consumed per sheet.

## Why's this needed? 
Plastic is new, not powerful, and could probably use a source.
This is a way to introduce a material source that can be somewhat industrialized through prepwork.

## Changelog 

```changelog
(u)MylieDaniels
(*)Plastic sheets can now be formed with 10 oil, 10 carbon, 20 hydrogen, and 40 chromium. Takes a while!
(+)Silicon dioxide will prevent chromium from depleting in plastic production, while magnesium dioxide will speed it up and prevent the carbon and hydrogen from fighting. Optimize!
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -2392,12 +2392,12 @@
 		on_reaction(datum/reagents/holder, created_volume)
 			count += created_volume
 			if (holder.has_reagent("silicon_dioxide")) // prevent catalyst ashing with a support
-				holder.remove_reagent("silicon_dioxide", created_volume / 2)
+				holder.remove_reagent("silicon_dioxide", created_volume / 5)
 			else
-				holder.remove_reagent("chromium", 5 * created_volume) // super exaggerated catalytic ashing
+				holder.remove_reagent("chromium", 4 * created_volume) // super exaggerated catalytic ashing
 			if (holder.has_reagent("magnesium_chloride", 10))
 				reaction_speed = 0.25
-				holder.remove_reagent("magnesium_chloride", created_volume / 2)
+				holder.remove_reagent("magnesium_chloride", created_volume / 5)
 			else
 				reaction_speed = 0.05
 			if (count >= 10)
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -2384,25 +2384,25 @@
 		var/count = 0
 
 		does_react(datum/reagents/holder)
-			if (holder.get_reagent_amount("chromium") / holder.total_volume < 0.33) // generously rounded
+			if (holder.get_reagent_amount("chromium") / holder.total_volume < 0.3)
 				return FALSE
 			else
 				return TRUE
 
 		on_reaction(datum/reagents/holder, created_volume)
 			count += created_volume
-			if (holder.get_reagent_amount("silicon_dioxide") > (5 * created_volume)) // prevent catalyst ashing with a support
-				holder.remove_reagent("silicon_dioxide", created_volume / 3)
+			if (holder.has_reagent("silicon_dioxide")) // prevent catalyst ashing with a support
+				holder.remove_reagent("silicon_dioxide", created_volume / 2)
 			else
-				holder.remove_reagent("chromium", 2 * created_volume) // super exaggerated catalytic ashing
-			if (holder.get_reagent_amount("magnesium_chloride") > 20 * created_volume)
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish (changes requested):
This is cool, I like recipes based on real-world processes, but it's a little overcomplicated I think for something that just produces a few low-value plastic sheets?
It also doesn't inhibit the carbon-hydrogen dissolving reaction so is almost impossible to complete right now.
Oh and test what happens when it occurs in smoke clouds.

MylieDaniels:
> This is cool, I like recipes based on real-world processes, but it's a little overcomplicated I think for something that just produces a few low-value plastic sheets?

I was a bit worried about making actual materials with chemistry as a concept, so started on the complex and technical side.
Do you have suggestions of what might make this more appealing or more fun to work with?

> It also doesn't inhibit the carbon-hydrogen dissolving reaction so is almost impossible to complete right now.

For making a low amount of plastic, I was producing it with fuel in the mix to make oil as it processed, and for bulk plasticmaking, this makes magnesium chloride inhibit the carbon-hydrogen reaction when in quantities greater than 5 units.

> Oh and test what happens when it occurs in smoke clouds.

Will do!

TobleroneSwordfish:
Oh I missed the oil recipe thing (sleepy brain), that's neat! For reducing complexity, I'd probably just remove the magnesium chloride interactions entirely, as they don't really add much other than making it (potentially infinitely?) faster and as it stands this is basically three different recipes combined into one. I'll see what other chemistry-interested devs think though.

MylieDaniels:
I've simplified that magnesium chloride logic and made it either 0.25 speed (40 second plastic) with, or 0.05 (200 second plastic) without. It will run at the same rate as unheated oil without, so the least industrial method is a room temperature large beaker of 10 carbon, 20 fuel, 20 hydrogen, and 40 chromium, which will produce a sheet of plastic in 200 seconds.

I've removed the ratio check for chromium, simplifying the recipe and allowing silicon dioxide to prop up the chromium content entirely.

The most industrial way, which requires more upfront prep but less time per sheet, would be to add a few units of chromium in a barrel, then add at least 2u silicon dioxide and 2u magnesium chloride per sheet you want to produce. You can then either produce the oil with hot fuel (it needs to be slightly heated to keep up with the faster pace) or add oil directly, but either way, 10 oil, 10 carbon, and 20 hydrogen will produce a sheet every 40 seconds.

Flaborized:
This is neat yeah, I like the idea of more physical, non-chemistry orientated products that you can make from chemistry. I would've also had the feedback about the complexity:usefulness ratio but the newer updated recipe seems more reasonable, good to have it be more complicated without actually being *that* difficult to make.

It could even be neat if we later expanded on this a bit and made plastic useful to research in some way ie: maybe letting you fabricate lids or something with it, but that's sorta beyond the scope here (maybe). Could do like anything, and would fill a neat niche.

Tarmunora (approved):

Tarmunora:
I still think that using ethanol+sulfuric acid as the ethylene standin would be better but this is fine

code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -2368,6 +2368,39 @@
...
+		reaction_icon_color = "#8c866d"
+		var/count = 0
+
+		on_reaction(datum/reagents/holder, created_volume)
+			count += created_volume
+			if (holder.has_reagent("silicon_dioxide")) // prevent catalyst ashing with a support
+				holder.remove_reagent("silicon_dioxide", created_volume / 5)
+			else
+				holder.remove_reagent("chromium", 4 * created_volume) // super exaggerated catalytic ashing
+			if (holder.has_reagent("magnesium_chloride", 10))
```
TobleroneSwordfish, line 2392:
```suggestion
			if (holder.has_reagent("magnesium_chloride"))
```
this is just going to confuse people otherwise

PR #16605, opened 2023-10-22
Fix shuttle areas requiring power during halloween

### LINKED ISSUE #16341: shuttle areas require power during halloween
### Labels

[BUG][minor]

### Description

Some bad compiler define code on shuttle areas leaves out the override for requires_power when it's halloween

### Steps to reproduce

1. build with the month define set to the 10th
2. open a map in a map editor
3. check the requires_power variable on the escape shuttle area (which is also black now)
4. see it's true

### Expected Behavior

no APC requirement (bonus points if y'all map in ephemeral shuttle APCs for halloween though)

relevant code is at https://github.com/goonstation/goonstation/blob/e7886d245a25b4b03a4666ccec013ea157b54886/code/area.dm#L654

We ran into this on cool where github checks currently run halloween code and it threw APC errors, but I don't think any of goon's checks or tests ever compile with halloween enabled so it's gone unnoticed all this time. Consider adding a check?

TealSeer:
[BUG] [STATION SYSTEMS]
## About the PR 

Fixes #16341 

## Why's this needed? 

This is leading to weird shit like ion storms affecting shuttle areas

### DIFF
--- a/code/area.dm
+++ b/code/area.dm
@@ -661,7 +661,7 @@ ABSTRACT_TYPE(/area/shuttle)
 	luminosity = 1
 	force_fullbright = 0
 #endif
-	requires_power = 0
+	requires_power = FALSE
 	sound_environment = 2
 	expandable = 0
 
--- a/code/area.dm
+++ b/code/area.dm
@@ -655,14 +655,13 @@ ABSTRACT_TYPE(/area/shuttle)
 	alpha = 128
 	icon = 'icons/effects/dark.dmi'
 #elif defined(UNDERWATER_MAP)
-	requires_power = 0
 	force_fullbright = 0
 	luminosity = 0
 #else
-	requires_power = 0
 	luminosity = 1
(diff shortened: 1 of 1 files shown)


code/area.dm
```diff
@@ -655,14 +655,13 @@ ABSTRACT_TYPE(/area/shuttle)
...
 #elif defined(UNDERWATER_MAP)
-	requires_power = 0
 	force_fullbright = 0
 	luminosity = 0
 #else
-	requires_power = 0
 	luminosity = 1
 	force_fullbright = 0
 #endif
+	requires_power = 0
```
Azrun, line 664:
```suggestion
	requires_power = FALSE
```

Azrun (approved):

code/area.dm
TealSeer: fuck i forgot we're doing that now mb

Azrun (approved):

PR #16608, opened 2023-10-22
Resprite jukebox

BatElite:
[sprites][station objects]
## About the PR 

image

image

## Why's this needed? 
I really don't like the old sprite and I think this looks way better.


DisturbHerb:
This might just be me, but I legitimately thought that the old sprite was your resprite when looking at the first preview. I think my thoughts on the icons are diametrically opposed to yours, especially with the yellow palette you went with.

BatElite:
The yellow is partly harkening back to the _old_ sprite, along with the dear green
image

blackrep:
I feel the new sprite is a step backwards. The yellow, green, and brown stripes feel a lot flatter in the new sprite compared to the old one.

wolfolotl:
I agree that this is a downgrade from our current one and would much prefer if you tried applying the details to said current one instead. the run of colors used especially are a major step back, imo

Azrun:
I appreciate the mix of yellow and green neon in the old-old sprite. My brain identities that as a jukebox probably more out of the three.

The lighting for the new spirit confuses me. Why is the top dark like that?

I like the roundness of the new spirit.

ZeWaka:
yeah i like the old sprite WAY more

BatElite:
> The lighting for the new spirit confuses me. Why is the top dark like that?

Cause it's the wooden top of the machine, that's just how jukeboxes in this style are.

Azrun:
What did they do to that wood to ensure that makes the highlighting on the top left side of the wood is non-existent?  That was present in the old-old sprite as well.

Gannets:
Don't like this one, sorry!

Glamurio:
The shape is better, but the palette is worse. Perhaps you could use the current palette and colors / shading, but stick with your new shape.

PR #16620, opened 2023-10-23
Bans List Layout Frame

mordent-goonstation:
## About the PR 

I spent a not-okay amount of time getting this to layout in a way that was sane with the scroll how I want. IE11 is a riot.

Of interest: "Cell" component (will probably need renaming) that is the start of something akin to a config-driven grid. This could have application for similar interfaces, such as the Player Panel, as the point of it is to be feature-rich without requiring people to implement everything.

To be clear, this is very much a WIP. Everything is still "in flight" in that it might change, but this builds so...

## Why's this needed? 

*Plenty* more to do, but this is an incremental step that will let someone with a configured v2 API connection start seeing some stuff.


tgui/packages/tgui/interfaces/BanPanel/BanList/Cell.tsx
```diff
@@ -0,0 +1,56 @@
+import { InfernoNode } from 'inferno';
```
ZeWaka, line 1: smh smh why isn't this in common smh

ZeWaka (approved):

Tarmunora (approved):
no issues with light-mode readability

ZeWaka:
image
current

ZeWaka:
😀

PR #16631, opened 2023-10-25
Replace Martian mutant gibstare ability with radiation blast

flappybatpal:
## About the PR 
Replaces the gibstare on Martian Mutants with the ability to cause radiation blasts (from rad storm).

## Why's this needed? 
I think having standard critters gib (not just kill) is a bit much.
Players learning about these critters have to get used to a very dangerous ability, this ability is also unique in that it gibs at range and all martians are carrying a 7 second stun retaliation ability.
The radiation blasts keep a "zone control" mechanic to the martians without being as lethal. Radiation is still fairly bad for you!

## Changelog 

```changelog
(u)Flappybat
(+)Martian Mutants gib ability has been swapped out for a radiation blast.
```

### DIFF
--- a/code/datums/abilities/critter/radiation_blast.dm
+++ b/code/datums/abilities/critter/radiation_blast.dm
@@ -1,0 +1,17 @@
+/datum/targetable/critter/radiationblast
+	name = "Radiation Blast"
+	desc = "Focus your psychic powers to unleash a concentrated blast of radiation at a spot."
+	cooldown = 100
+	disabled = FALSE
+	targeted = TRUE
+	target_anything = TRUE
+
+/datum/targetable/critter/radiationblast/cast(atom/target)
+	if (..())
+		return 1
+	if (disabled)
+		return
+	var/turf/T = get_turf(target)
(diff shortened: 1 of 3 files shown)


TobleroneSwordfish:
Do martians have rad resist? Otherwise this is probably going to lead to a lot of friendly fire.

UrsulaMejor:
Any way to make this a "psychic blast" using a subtype of the wave spawner so tinfoil hats can still do something?

Tarmunora:
Eh, this feels pretty off-theme

flappybatpal:
I'm going to go with adjusting this to a psychic damage AOE as suggested.

BatElite:
I think the gibstare is just about the only thing martians have going for them that sets them apart from a dozen critter types you just tank the damage while you melee them to death. It's good to have a critter with a real bastard ability just to keep even experienced folks on their toes.

Azrun:
I read this as "radiant burst" and was like "Oh wow, holy martians!"

Turns out I just no read good.

Issue #13585, opened 2023-04-06
Cooked bacon is thought of as raw.


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

When bacon is cooked and eaten, you get text about how you shouldn't eat raw meat.

### Steps to reproduce

1. Cook bacon
2. Eat bacon
3. Get food poisoning

### Expected Behavior

for cooked bacon to not be thought of as raw.

### Additional Information & Screenshots

N/A

Reported by: Sonnzer
On server: Goonstation 3 Roleplay: Morty
Active test merges: nothing
Round log date: 2023-04-05-23-48
Reported on: 2023-04-06 01:14:14
Map: OSHAN

PR #16633, opened 2023-10-25
Bacon Fixes

gear-space:
[Catering][Bug]
## About the PR 
Makes bacon work as you'd expect: cooked bacon doesn't give food poisoning and pigs drop raw bacon instead of cooked bacon. 

Cooked bacon seems to be implemented a bit differently than other meat based food items. When cooked, bacon (/obj/item/reagent_containers/food/snacks/ingredient/meat/bacon) is still a meat ingredient. As a meat ingredient, it is inheriting its heal proc from /obj/item/reagent_containers/food/snacks/ingredient/meat, specifying that there’s a 33% chance (if you’re not a maneater) to get food poisoning. https://github.com/goonstation/goonstation/blob/78ac4e4ef4941059e76910a3433131ac054536d2/code/modules/food_and_drink/ingredients.dm#L20-L24

Note: If there's a better way to wrangle proc inheritance where you can specify the parent to inherit from (aka here going to /obj/item/reagent_containers/food/snacks/ingredient instead of /obj/item/reagent_containers/food/snacks/ingredient/meat) that might be cleaner? Otherwise a bacon refactor seems... intimidating. But if that's the more correct way to go, let me know.

## Why's this needed? 
Fixes https://github.com/goonstation/goonstation/issues/13585


code/modules/food_and_drink/ingredients.dm
```diff
@@ -19,7 +19,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/ingredient)
 
 	heal(var/mob/living/M)
 		..()
-		if (!(istype(M, /mob/living/critter/plant/maneater)) && prob(33))
+		if (!(istype(M, /mob/living/critter/plant/maneater)) && !(src.name == "bacon") && prob(33))
```
TobleroneSwordfish, line 22: Name checks are incredibly brittle, for example labellers can modify the name directly. If you want cooked bacon to not inherit raw meat code then it probably shouldn't inherit from raw meat!

TobleroneSwordfish (changes requested):

gear-space:
Aww beans, didn't realize this would auto-update the PR. Ignore me for a bit while I double check this did what I intended.

code/modules/food_and_drink/ingredients.dm
TDHooligan: for what it's worth i think editing the superclass is the correct approach. making meat ingredients not explicitly poisonous (like giving meat ingredients a 'poisonous' var) makes sense given that both cooked and uncooked meats can be valid in the same cooking recipes. so you can just use a 'snacks/ingredient/bacon' check.

PR #16637, opened 2023-10-25
ABCU: Blueprint rework, QOL, whitelist bugfix

kbsmilk:
[QOL] [CODE QUALITY] [GAME OBJECTS]
## About the PR 
* Reworks blueprints. Old one was an obj that refs up to 1000 datums, gets used once, then discarded. Now, blueprints are a datum created on demand in the ABCU itself, which get left to GC upon replacement.
* Blueprints can be printed to share with other players, temporarily or permanently. This is a new obj that only stores the savefile's path. Clients can use it on the ABCU to load that savefile, or on themselves to copy that savefile to their ckey folder.
* File system-related procs were moved to be global, independent of datums, for organization and to facilitate usage by other code. 

* Whitelist and blacklist now work retroactively.
* Adds a little more logging.
* The terminal-like system used by the Blueprint Marker to select and manage savefiles, is gone. Replaced with searchable TGUI input lists across the board.
* The code quality of the blueprint marker is largely untouched, and still in need of fixing. I left this to atomize, but I'm open to merging it with this PR.

## Why's this needed? 
Less duplicate code. Less memory usage. Blueprint sharing is fun. +QOL. Paves the way for silicons being able to use the ABCU.
Unfortunately this PR turned out larger than expected, but I didn't want to leave anything unfinished, unpolished, or needlessly removed. 

## Changelog 

```changelog
(u)Smilg
(+)ABCU blueprint management streamlined. You can now print blueprints for other players to copy/save directly.
```


code/WorkInProgress/blueprints.dm
```diff
@@ -384,127 +385,338 @@
...
+// blacklist overrules whitelist
+#define BLACKLIST_OBJECTS list( \
+	/obj/disposalpipe/loafer, \
+	/obj/submachine/slot_machine/item, \
+	/obj/machinery/portable_atmospherics/canister, \
+)
+
+#define WHITELIST_TURFS list(/turf/simulated)
+
+/datum/abcu_blueprint
```
ZeWaka, line 521: can you move this to a different file than the abcu and all of this to the `modules/blueprints` folder or something? it's way due for organization

ZeWaka (changes requested):

code/WorkInProgress/blueprints.dm
kbsmilk: Im glad you asked. How much harder will that make the diff to read?

ZeWaka: it shouldn't, really

kbsmilk:
the diff was bad, the merge conflicts were bad, i think it made other devs groan. organizing this file is gonna get its own dedicated PR.

ZeWaka (approved):

pali6 (approved):

PR #16649, opened 2023-10-26
Fixes The TGUI Fatal Error TermOS Logo Art

Mister-Moriarty:
[UI] [Bug]

## About The PR:
Readds the correct spacing and indentation to the TermOS logo art that displays on TGUI error screens.


ZeWaka:
ty

PR #16653, opened 2023-10-26
reduces amount of phenol needed for glue production

colossusqw:
## About the PR 

This PR changes glue's recipe to 1 plasma, 1 oxygen, 1 hydrogen, 1 formaldehyde, 0.25 phenol, reduces total yield from 5 to 4.

## Why's this needed? 

After the chem update, many chems have currently become very resource intensive to produce, this PR focuses on reducing the work and time investment in the production of space glue via reducing the needed amount of formaldehyde and phenol.

```changelog
(u)Colossusqw
(+)Glue's recipe now only requires 0.25 units of phenol.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -2617,8 +2617,8 @@
 		name = "Space Glue"
 		id = "spaceglue"
 		result = "spaceglue"
-		required_reagents = list("plasma" = 1, "phenol" = 1, "oxygen" = 1, "hydrogen" = 1, "formaldehyde" = 1)
-		result_amount = 5
+		required_reagents = list("plasma" = 1, "phenol" = 0.25, "oxygen" = 1, "hydrogen" = 1, "formaldehyde" = 0.75)
+		result_amount = 4
 		mix_phrase = "The substance turns a dull yellow and becomes thick and sticky."
 
 	superlube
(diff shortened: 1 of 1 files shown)


TDHooligan:
3.2x yield for phenol,  same yield for diethylamine. given diethylamine is a lot easier to make it seems fine. might not be a big enough change but it's a step in the right direction

TobleroneSwordfish:
I think I'd rather remove one of the precursors to make it easier to make, the idea is that stuff should require less chemicals to make but those chemicals should be more involved to produce.

TDHooligan:
i think that's a bit difficult as this is based on phenol-formaldehyde glue...

code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -2617,8 +2617,8 @@
 		name = "Space Glue"
 		id = "spaceglue"
 		result = "spaceglue"
-		required_reagents = list("plasma" = 1, "phenol" = 1, "oxygen" = 1, "hydrogen" = 1, "formaldehyde" = 1)
-		result_amount = 5
+		required_reagents = list("plasma" = 1, "phenol" = 0.25, "oxygen" = 1, "hydrogen" = 1, "formaldehyde" = 0.75)
```
TobleroneSwordfish, line 2620:
```suggestion
		required_reagents = list("plasma" = 1, "phenol" = 0.25, "oxygen" = 1, "hydrogen" = 1, "formaldehyde" = 1)
```
maybe change this back to 1 since phenol is the main bottleneck and having only one fractional part is easier to think about?

TobleroneSwordfish (approved):
mmm oka this works too

PR #16654, opened 2023-10-26
Adds fractional condensers

TDHooligan:
[CHEMISTRY] [FEATURE]
## About the PR 
Adds a new kind of condenser - The fractional condenser. 
image
Sprites by Flaborized

Instead of evenly distributing reaction outputs between connected beakers, the fractional condenser has a unique connection for each beaker, up to 4 total.
Yellow -> Red -> Green -> Blue

When a reaction occurs, the reaction results are distributed down the appropriate condenser tube:

Acetone +      Phenol
Yellow    ->     Red

If any reaction outputs are unaccounted for, they are instead put in the last possible condenser tube.
Backflow is done per-tube, so if Red is leading to a full beaker, all products going down red will backflow into the condenser. 

TECHNICAL:
This adds a new line drawing method, due to a bug in filter transforms. It functionally looks the same as before but [avoids an issue (discord)](https://discord.com/channels/182249960895545344/890226559691157524/1166717680888315985) that causes clipping. The downside is that it doesn't work with 'source caps' due to how it scales sprites to make properly sized end-caps.

It also adds an explicit ordering to additional chemicals being added by reactions.

## Why's this needed? 
In the short term: People don't like the acetone/phenol split at the minute.

Long term: I think this is a good step towards incentivising larger setups. We can create reactions where expensive reagent could actually be recovered and re-used. etc.

## Changelog 

```changelog
(u)TDHooligan + Flaborized
(*)Chemistry now has a fractional condenser. Split acetone reactions etc.
```


code/WorkInProgress/actuallyKeelinsStuff.dm
```diff
@@ -94,6 +94,20 @@ Returns:
 			I.filters += filter(type="layer", render_source = (islist(render_source_cap) ? pick(render_source_cap) : render_source_cap), transform=M2)
 		I.transform = UNLINT(matrix().Turn(-angle).Translate((dist),0).Turn(angle))
 		result.lineImage = I
+	else if(mode == LINEMODE_STRETCH_NO_CLIP)
```
Laboredih123, line 97: could this be turned into a switch statement instead? a suggestion

TDHooligan:
> Personally, i think this should not be avaible in chemistry at round start. This is a very clear upgrade over the condenser in most applications.

counterpoint:
as far as i have heard, nobody likes the acetone/phenol reaction. everyone will use the one you start with for this. instead it will incentivise anything requiring acetone/phenol to become part of a reaction chian.

code/WorkInProgress/actuallyKeelinsStuff.dm
TobleroneSwordfish, line 97: document why this exists and is different to `LINEMODE_STRETCH` (since the name implies it should just be the default behaviour of `LINEMODE_STRETCH`)

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -817,11 +825,85 @@ proc/ui_describe_reagents(atom/A)
...
+			src.UpdateOverlays(null, "fluid_image")
+			if (reagents.total_volume)
+				var/fluid_state = round(clamp((src.reagents.total_volume / src.reagents.maximum_volume * 5 + 1), 1, 5))
+				if (!src.fluid_image)
+					src.fluid_image = image(src.icon, "fluid-condenser_fractional[fluid_state]", -1)
+				else
+					src.fluid_image.icon_state = "fluid-condenser_fractional[fluid_state]"
+				var/datum/color/average = reagents.get_average_color()
+				src.fluid_image.color = average.to_rgba()
+				src.UpdateOverlays(src.fluid_image, "fluid_image")
```
TobleroneSwordfish, line 847: this looks like it's a duplicate of the parent proc with different `icon_state`s? If so it should just use a var instead

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -817,11 +825,85 @@ proc/ui_describe_reagents(atom/A)
...
+		try_adding_container(var/obj/container, var/mob/user)
+			if (length(connected_containers) >= container_count)
+				boutput(user, "<span class='alert'>The [src.name] is fully connected!</span>")
+				return
```
TobleroneSwordfish, line 851: this shouldn't be needed if you use the parent vars, see above.

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -817,11 +825,85 @@ proc/ui_describe_reagents(atom/A)
...
 			else
 				container.reagents.add_reagent(reagent, divided_amount, sdata, temp_new, donotreact, donotupdate)
-
 	disposing()
 		src.remove_container()
 		. = ..()
 
+	fractional
+		var/container_count = 4
+		var/container_order[4]
```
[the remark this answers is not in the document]
TDHooligan: missed this value existing entirely, good catch

code/WorkInProgress/actuallyKeelinsStuff.dm
TDHooligan: fair, see if the new docs are satisfactory!

TDHooligan: could be. im not really fussed though as the end result wont be something like a nice, compact C# switch, it's still going to be huge

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -817,11 +826,66 @@ proc/ui_describe_reagents(atom/A)
...
+			for(var/i= 1 to max_amount_of_containers)
+				if (container_order[i] == container)
+					src.connected_containers.Remove(container)
+					src.UpdateOverlays(null, "tube\ref[i]")
+					UnregisterSignal(container, COMSIG_ATTACKHAND)
+					UnregisterSignal(container, XSIG_OUTERMOST_MOVABLE_CHANGED)
+					UnregisterSignal(container, COMSIG_MOVABLE_MOVED)
+					container_order[i] = FALSE
+
+		add_container(var/obj/container)
```
TobleroneSwordfish, line 874: needs to either check for or handle two outputs being connected to the same target, currently just runtimes

TobleroneSwordfish (changes requested):

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -710,15 +710,17 @@ proc/ui_describe_reagents(atom/A)
...
 	incompatible_with_chem_dispensers = TRUE //could maybe be ok? idk
 	can_recycle = FALSE //made of glass, but would be a waste and almost certainly accidental so no
 	splash_all_contents = FALSE
 	object_flags = FPRINT | OPENCONTAINER | SUPPRESSATTACK
 	initial_volume = 100
 	accepts_lid = TRUE
+	//prefix for fluid icon state
+	var/fluid_prefix = "condenser"
 	var/list/connected_containers = list() //! the containers currently connected to the condenser
+	var/list/unique_containers = list() //! connected_containers, but without duplicate entries
```
TobleroneSwordfish, line 723: copying what I said from discord: what's with the unique_containers var? Surely you can just check if it's in connected_containers and achieve the same thing?

TobleroneSwordfish (changes requested):

PR #16657, opened 2023-10-27
Unary Refactor part 2

Laboredih123:
[ATMOSPHERICS][CLEAN]
## About the PR 
Moves the portable connector, pressure tank, and passive vent over to the unary subtype.

## Why's this needed? 
Having only pipes in my pipe subtype is good.


code/modules/atmospherics/machinery/unary/vent.dm
```diff
@@ -1,57 +1,65 @@
...
+/obj/machinery/atmospherics/unary/vent/process()
 	..()
-	if(parent)
-		parent.mingle_with_turf(loc, 250)
 
-/obj/machinery/atmospherics/pipe/vent/disposing()
-	node1?.disconnect(src)
-	parent = null
-	..()
+	//stolen from mingle_with_turf WOOHOOH
```
pali6, line 16: Why are we stealing from mingle_with_turf instead of using it directly? What's the difference of this code compared to mingle_with_turf?

Laboredih123: mingle_with_turf belongs to pipeline, which is a pipe only thing, which means it isn't accessible for nonpipe stuff.

Laboredih123: there is nothing different except adapting the function to our vent.

Laboredih123: unfortunate, but i dont feel like changing the current functionality of the vent so i can remove that code paste.

Laboredih123:
not the conflict!

Laboredih123:
!merge_upstream

Laboredih123:
idk how that works

Laboredih123:
well thats especially not good

Laboredih123:
i think i goofed up more, why do i have more conflicts after merging

Laboredih123:
ill just reset and redo the typepath change

Laboredih123:
a ok now

code/modules/atmospherics/machinery/unary/vent.dm
pali6: Couldn't the relevant parts of the proc be made global and called from both the pipe network's mingle_with_turf and from here?

PR #16660, opened 2023-10-27
BSP Room Generation

Azrun:
## About the PR 
Adds mechanism to fill-out Cell Array with BSP data  
* Allows for room size
* Border size

Manipulate Cell Array to connect BSP leaves via:
* Connecting all siblings
* Connecting to neighbors via Spatial Partitioning

Add ability for Spatial Hashing to be done on arbitrary grid points, independent of being bound to atoms.

## Why's this needed? 
ProcGen ProcGen ProcGen ProcGen


Azrun:
clowns will eat me

PR #16680, opened 2023-10-28
Allows CO2 and fluids to act as neutron shielding

TobleroneSwordfish:
## About the PR 
Neutrons lose power on passing through fluids based on the level and viscosity of the fluid, adding a small amount of radium as they do so to simulate tritium contamination (we don't have tritium)
Neutrons lose power on passing through CO2 gas, adding a small amount of fallout gas as they do to simulate umm idk I like fallout gas.

Numbers are sourced from me recompiling and tweaking them until it feels right, may be subject to change after contact with live server engineering,

## Why's this needed? 
Shielding nuclear reactors is hard, flooding your reactor room with CO2 and liquid dark matter to contain the angry neutrons is cool.

## Changelog 

```changelog
(u)LeahTheTech
(*)CO2 and fluid puddles now work as partial neutron shielding, leaving behind fallout gas and radium contaminants respectively.
```

### DIFF
--- a/code/obj/nuclearreactor/nuclearreactor.dm
+++ b/code/obj/nuclearreactor/nuclearreactor.dm
@@ -840,14 +840,18 @@
 
 	tick(var/obj/projectile/P)
 		var/turf/simulated/T = get_turf(P)
-		if (issimulatedturf(T))
+		if (istype(T, /turf/space/fluid)) //can't really calculate this, just make it do something at least
+			P.power -= 10
+		else if (issimulatedturf(T))
 			var/obj/fluid/fluid = T.active_liquid
 			if (fluid)
-				var/viscosity = max(fluid.group.avg_viscosity, 10) //fluid group viscocity is either 1 or on a scale between 10-20, why
+				var/viscosity = max(fluid.group.avg_viscosity, 10) //fluid group viscocity seems to be either 1 or on a scale between 10-20, why
 				P.power -= viscosity * fluid.my_depth_level / 2
 				fluid.group.reagents.add_reagent("radium", P.power / 10)
 
-			P.power -= min(T.air.carbon_dioxide / 7, 40)
+			if (T.air?.carbon_dioxide)
+				P.power -= min(T.air.carbon_dioxide / 7, 40)
+				T.air.radgas += 1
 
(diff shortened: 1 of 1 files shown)


code/obj/nuclearreactor/nuclearreactor.dm
```diff
@@ -839,6 +839,20 @@
 		return TRUE
 
 	tick(var/obj/projectile/P)
+		var/turf/simulated/T = get_turf(P)
+		if (istype(T, /turf/space/fluid)) //can't really calculate this, just make it do something at least
+			P.power -= 10
```
pali6, line 844: The reagent generated is stored in `global.ocean_reagent_id` if you want to use that.

code/obj/nuclearreactor/nuclearreactor.dm
```diff
@@ -839,6 +839,20 @@
...
+		else if (issimulatedturf(T))
+			var/obj/fluid/fluid = T.active_liquid
+			if (fluid)
+				var/viscosity = max(fluid.group.avg_viscosity, 10) //fluid group viscocity seems to be either 1 or on a scale between 10-20, why
+				P.power -= viscosity * fluid.my_depth_level / 2
+				fluid.group.reagents.add_reagent("radium", P.power / 10)
+
+			if (T.air?.carbon_dioxide)
+				P.power -= min(T.air.carbon_dioxide / 7, 40)
+				T.air.radgas += 1
```
pali6, line 854: I feel like this should also remove part of the CO2.

But more importantly as written this does not work when group processing is on I believe. You will probably want to check `T.parent?.group_processing` and if it is true then do this all for `T.parent.air` except divided by `T.parent.air.group_multiplier`.

TobleroneSwordfish: ohh right, I was reading the reactor component code which doesn't use group processing

amylizzle: If we're doing gas checks anyway, do you maybe want to just take the neutron interaction code from the gas cooling reactor component and generalise it to this? Then we can have neutrons interacting with plasma in the air and making fallout and all that jazz.

PR #16698, opened 2023-10-29
Purgatives and dialysis machine rebalance

M-Earthfire:
[Balance][Chemistry][Game objects][Input wanted]

## About the PR 

This PR reduced the strenght on several purgative chems and increased the numbers on the dialysis machine.

Chemicals that purge all chemicals have their effectiveness reduced by 40%. This applies to the following chemicals:
- calomel
- pentetic acid
- hunchback

To compensate for these changes, the dialysis machine was improved significantly. It drains and cleans blood with an 60% increased rate.

## Why's this needed? 
When it comes to purging chemicals, the dialysis machine is vastly outclassed in effectiveness against purgative chemicals.

Purgative chemicals, while dealing damage and having negative implications like stuns/slowdowns, are easily transportable due to pills and work with a very high efficiency. The dialysis machine, in contrast to this, is bulky and purges chems fairly slowly. the positive is being able to selectively purge non-whitelist chems only.

What this creates is that the dialysis machine is very much unused except for very serious poison cases and on RP. This makes purging chemicals, a task medbay should be the most competent for, to be relegated to people running to the bar or having a few pills prepared in chemistry.

This PR aims to make the dialysis machine, a medbay-exclusive item, able to outcompete general purgatives. From my tests, with the numbers from the PR, purges chems from a victim injected with 100u plasma with almost the same speed calomel does.

Under this point, calomel is faster for the task than the dialysis machine, above that number the dialysis machine is faster, due to chems in blood scaling with the amount of chemicals in someone.

This should be a step to further move doctors away from being glorified chem dispensers. And Purgatives are still effective due to having the ability to be very easily transportable from A to B.

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Calomel, hunchback and pentetic acid were found to be far less effective. To compensate, NT has greatly improved the effectiveness of its dialysis machines.
```

### DIFF
--- a/code/modules/chemistry/Reagents-FoodDrink.dm
+++ b/code/modules/chemistry/Reagents-FoodDrink.dm
@@ -87,7 +87,7 @@ datum
 					M = holder.my_atom
 				if (M.get_toxin_damage() <= 25)
 					M.take_toxin_damage(-1 * mult)
-				flush(holder,4 * mult, flushed_reagents)
+				flush(holder,5 * mult, flushed_reagents)
 				..()
 				return
 
@@ -2552,7 +2552,7 @@ datum
 					description = initial(description)
 
 			on_mob_life(var/mob/M, var/mult = 1)
-				flush(holder, 0.8 * mult, flushed_reagents)//Tea is good for you!
+				flush(holder, 1 * mult, flushed_reagents)//Tea is good for you!
 				..()
 				return
 
@@ -4696,7 +4696,7 @@ datum
 			var/list/flushed_reagents = list("cholesterol")
 
 			on_mob_life(var/mob/M, var/mult = 1)
-				flush(holder, 2.4 * mult, flushed_reagents)
+				flush(holder, 3 * mult, flushed_reagents)
 				..()
 
 		fooddrink/iced/coconutmilkespresso
--- a/code/modules/chemistry/Reagents-FoodDrink.dm
+++ b/code/modules/chemistry/Reagents-FoodDrink.dm
@@ -87,7 +87,7 @@ datum
 					M = holder.my_atom
 				if (M.get_toxin_damage() <= 25)
 					M.take_toxin_damage(-1 * mult)
-				flush(holder,5 * mult, flushed_reagents)
+				flush(holder,4 * mult, flushed_reagents)
 				..()
 				return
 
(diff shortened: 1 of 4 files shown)


TheColorCyan:
Calomel, pent acid and smelling salts are used to purge neurotoxin in case you get stung by a changeling and reducing its effectiveness seems like an a massive eh to me

TDHooligan:
i think this change needs a bit more thought, rather than just flat out weakening even the joke purgatives.
why would anyone think "damn. now my atrazine only purges 0.8 weed"

you could probably scope this to weakening only pent and calomel and it'd have the desired effect. on medbay, pentetic can fall back on being an incredible TOX healing agent, calomel not so much.

MeggalBozale:
Strongly agree with hooligan, I don't believe nerfing purgative chems is an appropriate way to make the dialysis machine more appealing. I've never even seen it used, which is certainly horrible, but it should instead have something actually appealing to it so that it gets used. For instance, allowing players to pre-fill it with whitelisted chems, which get transferred to the patient's bloodstream, would do a great deal of good for making it more appealing. But that's just one idea.

I would love to see this machine have a use, but I don't like the idea of nerfing an existing machine to crown it champion, so to speak.

Tarmunora (changes requested):
Re chemical changes: imo only the general purgatives (charcoal/pent/hunchback/calomel) should be changed, if any - and probably charcoal should be left alone as well.

DisturbHerb:
as the person that made dialysis, making the machine take too much blood is gonna make people feel yucky inside :((

TDHooligan:
i ran the dialysis machine  on myself earlier for purging a MONSTER and it took a really long time. i get it sort of should but it was 'just use calomel or something' speed. i think given the cost/effort involved in using the dialysis machine, and how it'd be cool to get working, it should probably get a bigger speed boost than this.

DisturbHerb:
so long as whatever equilibrium it sits at isn't just constant hypotension, i suppose

Tarmunora (approved):

DisturbHerb:
16u per cycle is a *weird* number if I'm calculating this right. To me, anyway.

DisturbHerb:
I'm a pedant and I like factors of 5 and 10 👽

PR #16703, opened 2023-10-29
Smelling salts behaviour changes + 'Inhaled' trigger

TDHooligan:
[CHEMISTRY][FEATURE]
## About the PR 

When inhaled via an ampoule (using a new 'inhaled' feature specific to them or chemsmoke), Smelling Salts burn out your sinuses for 6 minutes:
 - You are woken up (Removes Drowsy and Passing out, once)
 - If you are weakened, it is removed, burning you for 10 damage **(this requires a second party, or rare item)**
     - NOTE: this does not affect stunned. ie, stun batons. Or paralysis (fully passed out)
 - You gain 30 max health, decaying linearly over 1 minute.
 - You cannot get these benefits again until the 6 minutes are up.
 
status effect icons (health bonus active, then the remaining 5 minutes):
image image

Smelling salts no longer has an as-yet-unexplained healing effect between -5~-35 health.

TECHNICAL: Adds a new parameter to ampoules, signifying the contents were inhaled. This still counts as INGEST, but chemicals that care for the parameter can look for it.

## Why's this needed? 
Smelling salts' main effect right now is as a purgative, I thought it'd be neat to give it some cool unique effects for particularly tense wrestling matches, or fights. Your gang leader KO'ed? crack open an ampoule under their nose. 

I set the cooldown as a 'makes sense between wrestling matches' but I think there could be a rationale for further increasing it. The strongest benefits of this are reserved to it being applied externally, though.

## Changelog 

```changelog
(u)TDHooligan
(+)Smelling salt ampoules now trade your sinuses for a short, risky consciousness boost.
```


colossusqw:
I like this a lot, but it gets me thinking, wouldn't it make sense to have a INHALE parameter that differs from INGEST? Like it could be useful for differentiating behavior in smoke and ampoules from behavior in drinking or taking pills.

TDHooligan:
> I like this a lot, but it gets me thinking, wouldn't it make sense to have a INHALE parameter that differs from INGEST? Like it could be useful for differentiating behavior in smoke and ampoules from behavior in drinking or taking pills.

as mentioned, this is added via param_list. i dont see a need to completely rewrite chemistry to support more than INGEST/TOUCH  considering that INHALE will ultimately be INGEST with extra effects.

code/modules/chemistry/Reagents-Medical.dm
```diff
@@ -666,14 +666,26 @@ datum
...
+				if(method == INGEST && volume_passed >= 3)
+					if(isliving(M) && !M.hasStatus("smelling_salts") && ("inhaled" in paramslist))
+						var/mob/living/H = M
+						H.delStatus("drowsy")
+						H.delStatus("passing_out")
+						if (H.stamina < 0 || H.hasStatus("weakened")) //enhanced effects if you're K/Oed (also implies a second person is applying this)
+							H.TakeDamage("chest", 0, 10, 0, DAMAGE_BURN) // a little damage penalty
+							if (H.use_stamina)
+								H.stamina = max(H.stamina_max*0.2,H.stamina)
+							H.changeStatus("weakened", -20 SECONDS)
```
Tarmunora, line 679: are you sure you want to be checking for `"weakened"`? `"paralysis"` is the "you are knocked out and asleep" status

Tarmunora (changes requested):

code/modules/chemistry/Reagents-Medical.dm
TDHooligan: weakened is predominantly what gets you downed in crit and from punches to the face - paralysis is odd as it covers stuff I wouldn't ordinarily want this to work on, like literal paralysants.

after re-reading this i think that given it's a one-off effect it *could* work on both, as full-on paralysants also tend to leave you with >20s of down time. your th oughts?

Tarmunora: could go either way - just that the comment shouldn't say "K/Oed" if it's not for knockouts lol

code/modules/status_system/statusEffects.dm
```diff
@@ -2451,7 +2451,43 @@
...
+	getTooltip()
+		if (duration > 5 MINUTES)
+			return "Smelling salts have knocked you back into being awake!"
+		else
+			return "Your sinuses are burning! Smelling salts can't perk you up."
+
+	onUpdate(optional=null)
+		var/seconds_left = round(duration/10)
+		var/cooldown = (maxDuration/10 - benefit_duration)
+		var/bonusRemaining = max(0,(seconds_left - cooldown)/benefit_duration)
```
Tarmunora, line 2479: I feel like this math could be simplified quite a bit - `1 + (duration - maxDuration) / (benefit_duration SECONDS)` is (essentially) equivalent

(Also not actually an issue but swapping from snake_case to camelCase over the course of 3 lines is a little silly)

code/modules/status_system/statusEffects.dm
```diff
@@ -2451,7 +2451,43 @@
...
+			icon_state = "smelling_salts"
+		return
+
+	proc/affectHealth(var/newBonus)
+		if (current_bonus != newBonus)
+			var/change = newBonus - current_bonus
+			var/mob/M = owner
+			M.max_health += change
+			current_bonus = newBonus
+			health_update_queue |= M
```
Tarmunora, line 2493: Should either typecheck somewhere so this won't runtime if put on a non-mob, or filter to mobs in `preCheck()`

Tarmunora (changes requested):

code/modules/status_system/statusEffects.dm
TDHooligan: too much codebase swapping :''(

PR #16705, opened 2023-10-29
Clean up fireflash.dm

FlameArrow57:
[INTERNAL][CLEANLINESS][BALANCE]

## About the PR 
This PR does a clean up of fireflash.dm

Currently in this file, there are three fireflash procs that all do different things, but have similar goals. The file is not commented and the code is hard to understand. Some code is very old or doesn't actually work

_**Old procs:**_
**fireflash()**
-Deletes kudzu
-Always creates new hotspots
-Line of sight check only checks turf density
-Burns mobs based on radius and proc temp
-Mob temp is set by proc temp

**fireflash_s()**
-Does not delete kudzu
-Creates new hotspots if not existing, otherwise just modifies existing hotspot temperature in a confusing way
-Line of sight check checks gas crossing, turf density, and blob tile blockers like fire walls. Object density checks are present but don't work (bug)
-Burns mobs based on hotspot temp
-Mob temp is set by proc temp

**fireflash_sm()**
-Same as fireflash_s(), but allows melting of floors

_**New procs:**_
**fireflash()**
-Now a combination of fireflash() and fireflash_s(), using parameters from both
-Deletes kudzu
-Always creates new hotspots
-Line of sight checks gas crossing, turf density, and blob tile blockers
-Burns mobs and sets mob temp based on hotspot temp

**fireflash_melting()**
-Acts the same as fireflash(), but allows for melting of floors
 
Side note of these changes - Allows fire sources to damage blob tiles even if there's already existing fire on them. Which can be changed if wanted, just a result of how it's coded at the moment.

Also removes the /obj/critter fire damage dealing from these procs, since most common critters have been made /mob/living/critter, and I think /obj/critter fire damage needs a more proper rework if it's to be implemented at all. Seems fireflashes are the only source of fire damage an /obj/critter can take

## Why's this needed? 
Large clean up


ZeWaka (changes requested):
> fireflash_melting()
> -Acts the same as fireflash(), but allows for melting of floors

so why not just make it an argument

FlameArrow57:
> > fireflash_melting()
> > -Acts the same as fireflash(), but allows for melting of floors
> 
> so why not just make it an argument

fireflash_melting() uses two arguments for melting stuff, so moving into fireflash() would add three arguments for melting. Thought it would be more concise leaving it separate, but could still merge

ZeWaka (approved):

PR #16708, opened 2023-10-29
Ports and uses frontend TGchat

CheffieGithub:
## About the PR 
Ports https://github.com/tgstation/tgstation/pull/52426
Ports https://github.com/tgstation/tgstation/pull/52947
Ports https://github.com/tgstation/tgstation/pull/54517
Ports https://github.com/tgstation/tgstation/pull/56575
Ports https://github.com/tgstation/tgstation/pull/59843
Ports https://github.com/tgstation/tgstation/pull/69211
Ports https://github.com/tgstation/tgstation/pull/72514
Ports https://github.com/tgstation/tgstation/pull/77228

Going to start with an important note that style made

# Please annotate more messages. Thank you.

For now a message 'annotation' is a type variable sent with boutput and a message having it instead of getting it from the class. This reduces what it needs to do when processing the message for chat, the type should still match the constants set in the `./tgui-panel/chat/constants.js` file though since they are used for filtering.

This could be expanded later though with the idea of handling plaintext messages on the js side by type for styling.
see https://github.com/tgstation/tgstation/pull/52947.

So for example: a message with the class `.alert` would be type `MESSAGE_TYPE_WARNING`

Removes old browserOutput browser assets.

The chat panel is now a TGUI interface with feature parity (almost) with the old chat system see above PRs for details.

TGchat has had variables adjusted for our chat such as the `MESSAGE_TYPES` which are used for the chat filter have had the selector classes changed to match.  (loosely because our CSS is inconsistent)

Context flags are now on client and are always set. If an admin has playermode enabled they will have no flags unless they are an observer ghost like a normal player.

Context flags global vars have been turned into defines.

For regex:

Expressions need to be inside of backslashes (/).
There is a limitation of `allowedRegex = /^[a-z0-9_\-$/^[\s\]\\]+$/gi` you would ned to ask StyleMistake as to why that is.
Its possible the limitation could be reduced as seen in https://github.com/tgstation/tgstation/pull/60146 to `/[-\/\\^$*+?.()|[\]{}]/g`, in the context of `.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&'))` in regards to the highlight string itself.

New filtering:

These settings can be found in the chat tab options and are unique to each tab.

image

Local: IC messages local to the player. (`.say`, `.emote`, `.sing`, etc.)
Broadcasted: IC messages that are broadcast to a group. (`.roboticsay`, `.thrallsay`, etc.) but not `.deadsay`.
Radio: All radio messsages.
Info: Messages from the game or items that _**aren't**_ important. `.notice`, `.success`, `.hint`, etc.)
Warning: Messages from the game or items that _**are**_ important. (`.alert`, `.blobalert`, `.lawupdate`)
Deadchat: Any `.deadsay` messages. Seperated from broadcast due to high message volume.
OOC: Self-explanatory 
LOOC: Self-explanatory
Mentor PMs: Self-explanatory
Combat: Any `.combat` messages. (Not a widely used class but a lot of alerts should probably be this instead)
Unsorted: anything not matching other filters.

image

Admin chat: ASAY messages. Filtered from the `.adminMsgWrap` class.
Admin log: Admin log messages. Filtered from new class `.adminLog`.
Attack log: Attack log messages. Filtered from new class `.attackLog`.
Coder log: Coder log messages. Filtered from new class `.coderLog`.

Darkmode:

image

Lightmode:

image

Both support odd / zebra highlight:

image

image

And with an admin / ghost context menu like before (but with tooltips and a name!):

image

image

A big thanks to 
StyleMistake for the orginal PR
ZeWaka for the initial backend port
and Mordent for fixing the initial port

- [x] Fix / Workaround process issues
- [x] Full port https://github.com/tgstation/tgstation/pull/56575
- [x] Port https://github.com/tgstation/tgstation/pull/72514
- [x] Use TGUI panel for admin audio / dectalk
- [x] Use telemetry
- [ ] Pull out any other old browerpanel stuff thats not needed

## Why's this needed? 
Allows us to port:
https://github.com/tgstation/tgstation/pull/66971
https://github.com/tgstation/tgstation/pull/75431 / oldver

for performance increases + other benefits

GIves us a chat panel which is extensible, maintainable and hooks into the current UI system which is becoming standard.

This chat panel should be a bit faster than the current one.

## Changelog 

```changelog
(u)Cheffie, ZeWaka and Mordent
(*)The chat panel is now a TGUI interface. Please report any chat related issues! Even if its just a styling problem!
(*)Client darkmode preference is now controlled by the chat panel.
(*)If chat fails to load, go to commands and try "fix chat". This hopefully shouldn't be an issue though.
(*)Not all regex will work after import as there is a limitation of allowedRegex = /^[a-z0-9_\-$/^[\s\]\\]+$/gi;
(+)Chat filtering options have been expanded, check PR page for details. Report any unsorted messages you think fit
 in a filter!
```


CheffieGithub:
I literally can't test playing audio with the changes (at least not normally) so a dev will need to do that.

CheffieGithub:
Its ready, Im not sure what exactly is "any other old browerpanel stuff thats not needed". There are two minor issues I know of that I will fix tomorrow being: ctx flags need their setting changed and incorrectly using TGs defines someplace

CheffieGithub:
Its now ready ready

CheffieGithub:
A known issue is that for some messages as an admin in player mode you can still right click to get an empty menu with just the close button, this is due to not every admin hearing boutput checking player mode and just generally being really inconsistent. its something ill fix in a later PR by making a proc to do it with all the relevant checks.

CheffieGithub:
Thanks to herb hosting a round with this PR I can confirm that the chat works fine during normal play, minor issues found in that round have been fixed.

ZeWaka:
!merge_upstream

TobleroneSwordfish (changes requested):
Some of the filtering doesn't seem to have survived the port very well:
- Admin chat is being filtered under admin log, probably because it uses the same output procs
- The round start "Welcome to goonstation" text is under warnings for some reason

Some messages can appear out of order with the way it handles stacking, which is a little confusing? This is the result of me pointing to space, then the floor, then space again:
image
I assume this is so that alternating messages can still stack, which is an issue with our current chat but it is a behaviour change so 🤷
 
There's an option to toggle off admin PMs, but it doesn't do anything because admin PMs are filtered under warnings. Imo admin PMs should just show up on every chat tab regardless of settings and I'm not sure why there's a button to toggle them.

The admin hearing toggle doesn't do anything, unsure what it's meant to be hooked up to? Maybe admin deadchat hearing, but we have settings for that in admin prefs iirc.

The case sensitivity option on the chat highlighting doesn't seem to work.

I don't think the music player is hooked up to anything so it should probably just be hidden until/if it ever is.

This is likely more of a gripe with the original interface than the port, but the delete tab button has no confirmation and there's no way to re-order tabs, meaning any accidental deletion forces you to set everything up again.

pali6:
Some things that need to be checked before merging: (note mostly for my own sake so I don't forget)
- [ ] Do goonchat settings migrate to tgchat correctly?
- [ ] Has the process scheduler been modified on master in a way that allows for this to work correctly?
- [ ] Is the secret module updated to correctly work with this?
- [ ] Is the performance when chat spamming bearable? (Note: This needs the server running from Dream Daemon explicitly to test.)

CheffieGithub:
Yeah filtering will not be consistent without massive changes to classes across the codebase, the best I can do is add / remove classes from the filter to try make it better or reduce the amount of filters.

Admin hearing is meant for any message with .adminHearing .name[data-ctx].

The filter selector has support for making admin PMs unfilterable.

The music player is hooked up to play_music_remote and dectalk. Its only for adjusting sound once something is played.

Case setting is an upstream issue https://github.com/tgstation/tgstation/issues/77720 I attempted some fixes but couldn't get it working, it would be best to remove the button until its fixed.

CheffieGithub:
Changed
No filtering .ahelp or .bigpm. Seperate filtering / css for different log messages. Remove admin hearing filter, it overlaps with chat and its not possible to make it work for now.

pali6:
image
this seems like something to fix

pali6:
> Some messages can appear out of order with the way it handles stacking, which is a little confusing? This is the result of me pointing to space, then the floor, then space again:
> 
image

Honestly I feel like this could be sorta neat if it was ordered according to the LAST message in the collapsed message group and not the first.

pali6:
Hitting File -> Reconnect seems to erase the chat completely for me
image

pali6:
Some inconsistency in newlines. I think maybe goonchat only renders <br> as a newline and tgchat renders both \n and <br> as a newline?
image
image
Unsure if we want to fix or just change the relevant descriptions etc.

flappybatpal:
> Some inconsistency in newlines. I think likely goonchat only renders as a newline and tgchat renders both \n and as a newline. image image Unsure if we want to fix or just change the relevant descriptions etc.

Seems to have issue with <br> too, adding extra lines such as the object build mode hints.
image

ZeWaka:
> Also along with the <br> extra lines it's showing all the spaces if the text contains them and before we removed any after the first space? Not sure which is desired.
> 

Probably should just fix the actual text

pali6:
Agreed with ZeWaka but the amount of text fixed might be large. Though one major source would be stuff (e.g. examine_verb()) doing .Join or jointext with \n (which currently don't dispalay as newlines in goonchat but display as newlines in tgchat).

CheffieGithub:
Im happy to fix these as they come up as long as I get information on them, ill keep an eye out myself though. I imagine some can be fixed with `whitespace: normal;`

CheffieGithub:
The text issues can be fixed by changing chat to use whitespace normal instead of pre-wrap. Work can be put towards fixing the text and changing it back to pre-wrap later if its desired, but this seems like a fine solution.

Tarmunora (approved):
No issues with light-mode readability

TobleroneSwordfish (changes requested):
The regex doesn't seem to work:
image
image
The top one is the auto-migration, but putting them in manually also does nothing.

There's no button to clear the chat history, that's a useful feature we have with the old chat that will get even more necessary now that the chat history saves between connections.

Admin PMs can't be filtered anymore, but now they only show up in the first chat tab, meaning it's possible to miss them by being on a different tab. They should be sent to all tabs regardless of settings.

ZeWaka:
1. chat defaults to a tile too wide it seems, tg has a different wideness than us
2. font is smaller (13 vs 14)
3. the darkmode background has way more contrast and actively hurts my eyes, please don't use tg's shitty color scheme
4. does the music player actually work? i don't think so, and thus the button shouldn't be there until it does

CheffieGithub:
> The regex doesn't seem to work: image image The top one is the auto-migration, but putting them in manually also does nothing.

From the message I added to chat on migration:
`Note you need to encase regex in / / for it to work.`

ll add the clear button but the admin PMs should show on all tabs ill need to look into that.

ZeWaka:
Ah, it doesn't say anywhere you need to enclose regex in slashes from what I saw.

CheffieGithub:
Yeah ill add it to the top of the highlight settings tab

CheffieGithub:
> 1. chat defaults to a tile less wide it seems, tg has a different wideness than us
> 
>     2. font is smaller (13 vs 14)
> 
>     3. the darkmode background has way more contrast and actively hurts my eyes, please don't use tg's shitty color scheme
> 
>     4. does the music player actually work? i don't think so, and thus the button shouldn't be there until it does
> 
> 
> image

I didn't notice it being too wide? It seems like normal widescreen setting width to me.

The fontsize of 13 px is better since the settings scales with em it starts to look worse 14 px and up. But I can change it if it needs to be 14.

Ill change the background to the old colour. I was just using the setting already there, though I quite like the higher contrast. I might make a high contrast version of the chat css later... I will be changing some css to have at least a contrast ratio of 5 though.

As previously states the music button does have a use, once music is played via it a widget shows up on chat with the title and a volume nob. Its not actually for playing music directly as far as I know but that can change.

ZeWaka:
> The fontsize of 13 px is better since the UI scales with em the settings panel starts to look worse 14 px and up. But I can change it if it needs to be 14.

it's personally much harder to see on my 1440p monitor, i think the settings panel looks fine. other devs can weigh in, but... (see next point)

> Ill change the background to the old colour. I was just using the setting already there, though I quite like the higher contrast. I might make a high contrast version of the chat css later... I will be changing some css to have at least a contrast ratio of 5 though for both dark and light since some classes have a really low contrast (less than 3 ratio)

i think keeping this PR as close as 1:1 with existing chat is best for now. we can make changes after rather than having to evaluate multiple new features at once, see: atomization

> As previously states the music button does have a use, once music is played via it a widget shows up on chat with the title and a volume nob. Its not actually for playing music directly as far as I know but that can change.

didn't work for me. my reproduction: upload file and 'play music' on the dj panel

tgui/packages/tgui-panel/settings/migration.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+  if (oldCookies.shighlightTerms) {
+    let savedTerms = JSON.parse(oldCookies.shighlightTerms).filter(entry => {
+      return entry !== null && /\S/.test(entry);
+    });
+    let actualTerms = savedTerms.length !== 0 ? savedTerms.join(', ') : null;
+    if (actualTerms) {
+      highlightSettings.terms = actualTerms;
+      message.push(`Imported highlight terms of "${highlightSettings.terms}".`);
+      message.push(`Note you need to encase regex in / for it to work.`);
```
ZeWaka, line 86:
```suggestion
      message.push(`Note you need to encase regex in forward slashes (/) for it to work.`);
```

tgui/packages/tgui-panel/settings/SettingsPanel.js
```diff
@@ -94,49 +129,159 @@ export const SettingsGeneral = (props, context) => {
...
+  );
+};
+
+const TextHighlightSettings = (props, context) => {
+  const highlightSettings = useSelector(context, selectHighlightSettings);
+  const dispatch = useDispatch(context);
+  return (
+    <Section fill scrollable height="200px">
+      <Box fontSize="1.1em" ml={1}>
+        Regular expressions must be inside of /
```
ZeWaka, line 177:
```suggestion
        Regular expressions must be encased within forward slashes (/)
```

ZeWaka: or similar - could be too long

CheffieGithub:
> > The fontsize of 13 px is better since the UI scales with em the settings panel starts to look worse 14 px and up. But I can change it if it needs to be 14.
> 
> it's personally much harder to see on my 1440p monitor, i think the settings panel looks fine. other devs can weigh in, but... (see next point)
> 
> > Ill change the background to the old colour. I was just using the setting already there, though I quite like the higher contrast. I might make a high contrast version of the chat css later... I will be changing some css to have at least a contrast ratio of 5 though for both dark and light since some classes have a really low contrast (less than 3 ratio)
> 
> i think keeping this PR as close as 1:1 with existing chat is best for now. we can make changes after rather than having to evaluate multiple new features at once, see: atomization
> 
> > As previously states the music button does have a use, once music is played via it a widget shows up on chat with the title and a volume nob. Its not actually for playing music directly as far as I know but that can change.
> 
> didn't work for me. my reproduction: upload file and 'play music' on the dj panel

Yeah the contrast stuff would be a later PR

'play music' doesn't use the panel. it needs a HTTPS url since it uses the HTML audio thing. Only play_music_remote() and dectalk use that.

CheffieGithub:
That should be just about everything.
I looked into making the combined message move to the front of chat but I didn't figure it out just yet, its not really critical so it could be added later.
The only thing I haven't looked into yet is the dropdowns, I also want to look at why the light vs dark positioning of the inside item is different as well as the previously mentioned 'doesn't update' thing when importing settings.

Tarmunora:
formatting things: player LOOC text is no longer blue - I personally preferred it having the OOC formatting so it stood out better
I swear admin log messages used to be bolded but no longer are

CheffieGithub:
> formatting things: player LOOC text is no longer blue - I personally preferred it having the OOC formatting so it stood out better I swear admin log messages used to be bolded but no longer are

Yeah accidental changes, making fixes now

Sovexe:
Noticed that when I initially launched into the game I had a darkmode chatbox but the rest of the client still used a lightmode theme, changing the chat theme from light back to dark resolved this correctly skinning the client in dark mode. Presumably an issue caused by me previously having darkmode set for goonchat which relied on both the chatbox and the client's display menu dark mode options being enabled, the two of which are encompassed into one setting in tgchat

Sovexe:
dectalk doesnt appear to be working. Granted it was already broken but now its no longer saying the incorrect head surgeon line either

CheffieGithub:
Dectalk should be fixed

CheffieGithub:
Theme loading is fixed, It should be ready for testmerge round 2 now

Xkeeper0:
notes (tested on dark mode)

* the text highlight always highlights words in black, even if the highlight color is dark (e.g. black on dark purple)
* the :information_source: next to player names in many admin logs is missing
* mentorhelps are strange (examples with names substituted are below), notice the lack of arrow/symbol and missing names / "not ingame" indicator

if this is like standard tg chat, i would like to see the "you are afk or the connection has closed" message be relocated to the top of the chat log, because at the bottom it can and will obscure the chat when a random ping fails

```
MENTOR PM: frommentor (Discord)  toplayer/ : sample mentorhelp
MENTOR PM: fromplayer   tomentor(Discord): sample reply
```

Sovexe:
Players have reported issues with various chat types being unchecked when joining the server (notably the `LOOC` and `Broadcast` channels). Connecting after clearing IE cache appears to have them all selected so I'm suspecting this may relate to their old goonchat filters

ZeWaka:
I prefer the old chat stacking style to TG's, yeah - it's the inferior copy

flappybatpal:
Died on 4 for some reason

`TypeError: Object expected
   at t.PingIndicator (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:238489)
   at Te (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:155873)
   at Anonymous function (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:156141)
   at Ae (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:156119)
   at Pe (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:156972)
   at Ae (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:155983)
   at Anonymous function (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:156141)
   at Ae (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:156119)
   at Ve (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:157167)
   at Pe (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:157006)
User Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.2; WOW64; Trident/7.0; .NET4.0C; .NET4.0E; .NET CLR 2.0.50727; .NET CLR 3.0.30729; .NET CLR 3.5.30729)
State: {"ckey":"flappybat","window":{"fancy":0,"locked":0}}

UnhandledRejection: TypeError: Object expected
   at t.PingIndicator (https://cdn.goonhub.com/tgui/tgui-panel.bundle.js?v=a79b9b00fcb65b3c930150ed7db84aaf80e7102f:1:238489)
`

CheffieGithub:
> Players have reported issues with various chat types being unchecked when joining the server (notably the `LOOC` and `Broadcast` channels). Connecting after clearing IE cache appears to have them all selected

That will be because they have saved settings from before they were added, which is why clearing cache fixes it as it makes it remake the main page.

CheffieGithub:
As to the termOS error, I haven't encountered it in testing at all. The only thing I can think of is that something only present on live such as the secret module has some message to the browseroutput which causes the first error which somehow also breaks reloading of the panel?

CheffieGithub:
Ive adjusted the css of the notifications and context menu, the menu is smaller and the notifications are at the top of the chat window

image

(bad screenshot)

ZeWaka:
> That will be because they have saved settings from before they were added...

it should probably default to On

CheffieGithub:
> notes (tested on dark mode)
>    * the player options right click menu doesn't close itself when choosing an option
>       * they are also really large with every option having huge vertical padding, the menu takes up 70% of my chat height
>       * it also does not close if you click outside of the menu, so the only way to get rid of it is clicking "close"
>       * it also doesn't seem to work on yourself? weird

I decided to make it not close when selecting an option since you might want to use multiple options at once.
I stopped it working on yourself since some of the options are buggy if you use them on yourself or just won't work as expected.
Actually making the menu css not show up on your own messages is something I plan to do in a different PR, assuming thats wanted and it continues to not work on yourself.
I was planning on making it close on offclick, will do soon.
Size has been reduced.

> * the ℹ️ next to player names in many admin logs, mentorhelps, ahelps etc are missing
> * mentorhelps are strange (examples with names substituted are below), notice the lack of arrow/symbol and missing names / "not ingame" indicator

I would need some actual comparison for this, im not sure what exactly is missing.

> * the text highlight always highlights words in black, even if the highlight color is dark (e.g. black on dark purple)

I can see if I can some up with some test of the highlight colour to change it to white or black for contrast but this contrast issue is present on current chat as well. Maybe the inverse of the highlight colour would work well?

> * multiple messages in the same group will only show the _first_ message that arrives; this is an inversion of current behavior where the line will update to the most recent one

This is somthing im looking into changing but ive been having a hard time.


> * there's additional extra padding on the chat window, so chat rows don't extend to the full width of the panel

Theres always been a gap, Its just bigger by a few pixels. Im not really sure why exactly that is though the css is exactly the same. Its probably to do with the JSX UI rendering and not easily fixable though.

Without PR:
image

 
With PR:
image

Xkeeper0:
image

The (i) links will open that person's player panel. When this was testmerged they were not appearing anywhere.

The actual username links open Admin PM boxes, so without the (i) link there's no simple way of opening the player panel.

CheffieGithub:
Ill see whats causing that

CheffieGithub:
while not exactly the right icon I know whats causing the issue and can fix it, 
image
the bad news is trying to find the same icon in the tgui icon css instead

CheffieGithub:
> > That will be because they have saved settings from before they were added...
> 
> it should probably default to On

They do default to on but its only done when its created as its a list of accepted types thats populated when the tab is created

CheffieGithub:
I found some code that changes the text colour to white or black depending on the highlight colour, it will probably increase the time to set highlights though. Great for accessibility however.

image

tgui/packages/tgui-panel/chat/renderer.js
```diff
@@ -32,10 +32,27 @@ const findNearestScrollableParent = startingNode => {
...
+  // adapted from https://stackoverflow.com/a/35970186
+  // should be in form "#FFFFFF" or "#FFF"
+  let hex = color.slice(1);
+  if (hex.length === 3) {
+    hex = hex[0] + hex[0] + hex[1] + hex[1] + hex[2] + hex[2];
+  }
+  const red = parseInt(hex.slice(0, 2), 16);
+  const green = parseInt(hex.slice(2, 4), 16);
+  const blue = parseInt(hex.slice(4, 6), 16);
+  logger.log(red + ' ' + green + ' ' + blue);
```
ZeWaka, line 45: do we really

tgui/packages/tgui-panel/chat/model.js
```diff
@@ -14,9 +14,7 @@ export const canPageAcceptType = (page, type) => (
 export const createPage = obj => {
   const acceptedTypes = {};
   for (let typeDef of MESSAGE_TYPES) {
-    if (typeDef.important) {
-      acceptedTypes[typeDef.type] = true;
-    }
+    acceptedTypes[typeDef.type] = true;
```
ZeWaka, line 17: huh. it's making them all false for me, can you repro? 
image

Xkeeper0:
> while not exactly the right icon I know whats causing the issue (outdated fontawesome icons) and can fix it image the bad news is trying to find the same icon in the tgui icon css instead (for all html icons(not inside of chui))
> 
> image

i figured it was something like the icon was missing. looks good to me 👍

ZeWaka:
https://github.com/goonstation/goonstation/blob/e2e92814ce9f9bd552a0112a713bfc1bb344c17f/browserassets/js/browserOutput.js#L143-L161
relevant

tgui/packages/tgui-panel/chat/model.js
CheffieGithub: it works fine when creating a new tab, again old ones will not update its set on creation.
image

CheffieGithub:
The one caveat to the off click closing is that the click needs to be on the chat pane, not the settings or anything else.
image

TobleroneSwordfish:
Seems like the `Broadcasted` setting defaults to off? 
image
Might just be because I had chat tabs set up previously, but ideally new settings should default to on, otherwise people will randomly lose bits of their chat whenever we add new settings.

CheffieGithub:
> * Admin chat isn't going to discord, and discord admin chat isn't appearing in-game.

Im not sure how I would've caused this here, unless for some reason it was dependant on the browser output files? but it seems like it just gets a msg from the bot and sends that DM side into boutput?

Xkeeper0:
This makes it impossible to copy-paste in-game logs/messages as an admin with the mouse, because right-clicking _anywhere_ pops open the player menu, _even in places where there is no text_ (e.g. after the message, in the whitespace area). The only place you can right click w/o triggering the pop menu now is in non-player text like examines or other messages.

Previously this would only happen when right clicking on names, which is why they have the dotted underline.

Tarmunora (changes requested):
Regression: Message stacking has lightmode readability issue now - message counter is white text on white background.

Xkeeper0:
the reconnect at the end of a round (and probably when reconnecting in general) reinits the chat and resets the music player, cutting off anything that hasn't finished playing yet

flappybatpal:
Players report seeing a random 2, screenshotted:
image

The telemetry payload kick should get logged.

code/modules/admin/adminhelp.dm
```diff
@@ -223,22 +223,24 @@
...
-				else if(M.client.holder.audible_prayers == 2) // this is a terrible idea
-					if(!audio)
-						audio = dectalk(msg)
-					var/vol = M.client.getVolume(VOLUME_CHANNEL_ADMIN)
-					if(vol)
-						M.client.chatOutput.playDectalk(audio["audio"], "prayer by [src] ([src.ckey]) to [M.ckey]", vol)
+		if (!M.client.holder.hear_prayers || (M.client.player_mode == 1 && M.client.player_mode_ahelp == 0)) //XOR for admin prayer setting and player mode w/ no ahelps
+			continue
+		else
+			boutput(M, "<span class='notice' [in_chapel? "style='font-size:1.1em'":""]><B>PRAYER: [is_atheist ? "(ATHEIST) " : ""]</B><a href='?src=\ref[M.client.holder];action=subtlemsg&targetckey=[client.ckey]'>[client.key]</a> / [client.mob.real_name ? client.mob.real_name : client.mob.name] <A HREF='?src=\ref[M.client.holder];action=adminplayeropts;targetckey=[client.ckey]' class='popt'><i class='fas fa-circle-info' />: <I>[msg]</I></span>")
```
pali6, line 235:
```suggestion
			boutput(M, "<span class='notice' [in_chapel? "style='font-size:1.1em'":""]><B>PRAYER: [is_atheist ? "(ATHEIST) " : ""]</B><a href='?src=\ref[M.client.holder];action=subtlemsg&targetckey=[client.ckey]'>[client.key]</a> / [client.mob.real_name ? client.mob.real_name : client.mob.name] <A HREF='?src=\ref[M.client.holder];action=adminplayeropts;targetckey=[client.ckey]' class='popt'><i class='fas fa-circle-info'></i>: <I>[msg]</I></span>")
```
this doesn't close properly and causes prayers to be ugly and allcaps as they are in Font Awesome 6 font or whatever

TobleroneSwordfish:
From current testmerge: the chat seems to break every time the round restarts and requires fix-chat to reload.

mralexs5:
`TypeError: Object expected
   at t.PingIndicator (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:238489)
   at Te (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:155873)
   at Anonymous function (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:156141)
   at Ae (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:156119)
   at Pe (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:156972)
   at Ae (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:155983)
   at Anonymous function (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:156141)
   at Ae (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:156119)
   at Ve (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:157167)
   at Pe (https://cdn-main1.goonhub.com/tgui/tgui-panel.bundle.js?v=11c6b2c81986388e4be8ccc0a3171f7e74da42ee:1:157006)
User Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.2; WOW64; Trident/7.0; .NET4.0C; .NET4.0E; Tablet PC 2.0; .NET CLR 2.0.50727; .NET CLR 3.0.30729; .NET CLR 3.5.30729; Zoom 3.6.0)
State: {"ckey":"mralexs","window":{"fancy":0,"locked":0}}`
I keep getting this error, seems to have to do with the ping display?

Piesuu:
Getting a similar error as well, being forced into a bluescreen over and over
```
TypeError: Object expected
   at t.PingIndicator (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:238489)
   at Te (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:155873)
   at Anonymous function (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:156141)
   at Ae (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:156119)
   at Pe (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:156972)
   at Ae (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:155983)
   at Anonymous function (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:156141)
   at Ae (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:156119)
   at Ve (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:157167)
   at Pe (https://cdn-main3.goonhub.com/tgui/tgui-panel.bundle.js?v=db4b2f47e7510fc8a111d1b91a10e6e56884fbff:1:157006)
User Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.2; WOW64; Trident/7.0; .NET4.0C; .NET4.0E; .NET CLR 2.0.50727; .NET CLR 3.0.30729; .NET CLR 3.5.30729; Tablet PC 2.0; InfoPath.3)
State: {"ckey":"piesuu","window":{"fancy":0,"locked":0}}
```

Piesuu:
Also probably related: I left the broken tgui chat goon window in the background waiting for a new shift to start and then alt-tabbed back into it after about half an hour, right after which my PC was spammed with dreamseeker processes (task manager was filled with the process before the freeze) until it froze completely and had to be restarted manually.

ZeWaka:
me issue on the builds btw

ZeWaka:
BEGONE BOT

Tarmunora:
note: I did not test any of the ban detail-related fixes other than making sure it compiled

CheffieGithub:
Thanks for the fixes

PR #16713, opened 2023-10-30
Fixes phone renaming not updating label

### LINKED ISSUE #16712: Phone renaming not working
### Labels

[BUG][MINOR]

### Description

Pulsing phone with multitool lets you rename phones, but it doesn't actually change the name on the label

### Steps to reproduce

1. Build a phone (I was using a wall phone)
2. Use multitool to rename
3. Label still reads the area name (built zone in this case)

### Expected Behavior

Phone rename to rename phone

Reported by: Joeled
Active test merges: nothing
Round log date: 2023-10-30-03-17
Reported on: 2023-10-30 03:20:06
Map: DONUT2

TealSeer:
[BUG] [GAME OBJECTS]
## About the PR 

Fixes #16712 by moving the label examine text to `get_desc()` so it reflects the current value of `phone_id`. Also switches the input box to tgui so max_length can be used, adds an early return for null input instead of checking `t` in each if, adds a range check, and adds a message indicating successful renaming. 

## Why's this needed? 

Bug fix for the label description, the other stuff is just general QOL and cleaner code.

### DIFF
--- a/code/obj/machinery/phone/phone.dm
+++ b/code/obj/machinery/phone/phone.dm
@@ -131,7 +131,7 @@ TYPEINFO(/obj/machinery/phone)
 			if(length(t) > 50)
 				return
 			src.phone_id = t
-			boutput(user, "You rename the phone to \"[src.phone_id]\".")
+			boutput(user, "<span class='notice'>You rename the phone to \"[src.phone_id]\".</span>")
 			return
 		..()
 		src._health -= P.force
--- a/code/obj/machinery/phone/phone.dm
+++ b/code/obj/machinery/phone/phone.dm
@@ -128,8 +128,6 @@ TYPEINFO(/obj/machinery/phone)
 				return
 			if(!in_interact_range(src, user))
 				return
-			if(length(t) > 50)
-				return
 			src.phone_id = t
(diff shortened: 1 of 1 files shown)


code/obj/machinery/phone/phone.dm
```diff
@@ -119,13 +121,17 @@ TYPEINFO(/obj/machinery/phone)
...
-			if(t && length(t) > 50)
+			if(!t)
 				return
-			if(t)
-				src.phone_id = t
-			src.labelling = FALSE
+			if(!in_interact_range(src, user))
+				return
+			if(length(t) > 50)
+				return
```
TobleroneSwordfish, line 132:
```suggestion
```
Fun fact actually, you don't need this since `tgui_input_text` validates the length server side already (even though I'm pretty sure most of our code still checks again), and putting the check here might break with html encoding modifying the length.

TobleroneSwordfish (changes requested):

code/obj/machinery/phone/phone.dm
TealSeer: oh cool, I didn't realize that!

PR #16726, opened 2023-10-31
adds purple and black folding chairs

WALPVRGIS:
## About the PR
Adds purple folding chair and black folding chair. This doesn't place it anywhere on any maps, nor does it make it available for purchase (afaik you can't order these specific chairs from cargo?? they just Exist). Feel free to use these in mapping bc im SICK and TIRED of MEDBAY GETTING BLUE CHAIR BUT SCI NOT GETTING PURPLE CHAIR. I just added black for completion :')

image
Here they are next to blue for comparison.

## Why's this needed?
When I'm mapping and research doesn't get its own cool folding chair colour like the other departments smdh and also it took me 5 minutes so why not!

## Changelog 

```changelog
(u)WALPVRGIS
(+)Teeny tiny sprite addition: added black and purple folding chairs for mappers, yippee!
```


pali6:
I feel like this is missing inhands

WALPVRGIS:
> I feel like this is missing inhands

Oh you might be right ty, I was just picking up old sprites id finished and PRing them, lemme check! cause its a quick fix if so B)

pali6 (approved):

PR #16728, opened 2023-10-31
Afterattack changes

pali6:
## About the PR 

Currently afterattack() gets run before attack() before the call to attack() is in a SPAWN. This PR removes that spawn (and adds set waitfor = 0 to mob/attackby to prevent bad sleeps).

This was discovered by looking into why https://github.com/goonstation/goonstation/commit/16bd50cd0b4428a772aa849c92f0dd6fd6bd42b7 broke patches. However, the above change alone does not fix the issue. ~~So this PR also loosens the check on AfterAttack() call to allow for it even if the item changed places. As a tradeoff the "is this item deleted" checks have been fixed.~~ No longer the case. Instead I will modify patches later. Calling afterattack on moved stuff causes more problems than it fixes. See issues below.

The intent of this PR is that it causes not observable changes other than fixing bugs (such as patches not working). The *likely* outcome of this PR is more bugs. I recommend testmerging for some time.

## Why's this needed? 
It's called **after**attack, it running **before** attack() is madness.

### DIFF
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -364,25 +364,23 @@
 	return FALSE
 
 /mob/living/proc/weapon_attack(atom/target, obj/item/W, reach, params)
-	var/usingInner = 0
 	if (W.useInnerItem && length(W.contents) > 0)
 		var/obj/item/held = W.holding
 		if (!held)
 			held = pick(W.contents)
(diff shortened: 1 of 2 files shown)


TheTyrant25:
seems to cause #16728

pali6:
> seems to cause #16734

thanks

PR #16729, opened 2023-10-31
Componetize Pitfalls

TheTyrant25:
[INTERNAL] [CODE QUALITY]
## About the PR
Adds `/datum/component/pitfall`, which teleports a movable atom on `Entered()` signal. There are 3 subtypes which control three different kinds of pitfall, namely Area targeted pitfalls (like falling to the pitfall area in the trench), landmark targeted pitfalls (most azone pitfalls use this, e.g. icemoon), and coordinate targeted pitfalls (e.g. when you use a sea ladder). The parent is abstract.
There are two arguments which apply to all 3 subtypes.
```
ARG_INFO("BruteDamageMax", DATA_INPUT_NUM, "The maximum amount of random brute damage applied by the fall.", 0),
ARG_INFO("FallTime", DATA_INPUT_NUM, "How long it takes for a thing to fall into the pit.", 0.3 SECONDS)
```
For Area Targeted pitfalls, there is another argument:
```
ARG_INFO("TargetArea", DATA_INPUT_TYPE, "The area typepath that the target falls into. If null, then it drops onto the same coordinates.", null)
```
Similarly for Landmark targeted pitfalls, there is one argument:
```
ARG_INFO("TargetLandmark", DATA_INPUT_TEXT, "The landmark that the fall sends you to.", "")
```
Finally, for coordinate targeted pitfalls, the most complex one code wise, there are two arguments. There is also an internal variable called `TargetList` which is a list of valid turfs to fall to, determined by the arguments and the topology.
```
ARG_INFO("TargetZ", DATA_INPUT_NUM, "The z level that the target falls into.", 5),
ARG_INFO("LandingRange", DATA_INPUT_NUM, "If true, try to find a spot around the target to land on in range (x). Only for 'direct drops'.", 0)
```
Using this, it's possible to strongly customise pitfalling.
Most instances use the simple landmark targeting system. The only places that don't are the generic oshan/nadir pitfalls, which either use the `target_area` subtype (to target `/area/trench_landing`) or use the `target_coordinate` direct drop down to a close by spot in `range(src.LandingRange)` on z level `TargetZ` instead depending on the type of pitfall. This component covers these use cases and can be expanded upon later in case other pitfalling behaviour is needed.

The twelve spots they are used at time of writing are:
- Oshan/Nadir: 2 kinds
- NSS Polaris (2 kinds of pit)
- sea elevator shaft
- icemoon (2 kinds)
- Biodome (3 kinds)
- WIP folder content (2 kinds)

No behaviour should be changed by this pr. This is a code reshuffle. All pits should function identically to before, except for perhaps being a little more robust in checking what things can fall in.

## Why's this needed?
Pitfalls are a fairly common thing, so much so that they're used in 12 separate spots. In each of these separate spots, they are defined and used completely separately. This is harder to maintain and overall worse than having the code unified within a component.


code/modules/fluids/fluid_objects.dm
```diff
@@ -467,11 +467,10 @@ TYPEINFO(/obj/item/sea_ladder)
 
 	afterattack(atom/target, mob/user as mob)
 		if (istype(target,/turf/space/fluid/warp_z5))
-			var/turf/space/fluid/warp_z5/hole = target
-			hole.try_build_turf_list() //in case we dont have one yet
-
-			deploy_ladder(hole, pick(hole.L), user)
-
+			SPAWN(0.1 SECONDS)	// give the turf's pitfall component time to work
```
pali6, line 470: Why is this necessary? Work on what?

TheTyrant25: there is a linking mechanism for holes and sea ladders. I moved it to the component seeing as it was related, but it also means that every time attackby is called, the `update_targets()` proc needs to run to update the list.

It's messy though

code/WorkInProgress/ZeWakasStuff.dm
```diff
@@ -210,15 +210,9 @@
...
-			return ..()
-
-		var/turf/T = pick_landmark(LANDMARK_FALL_GREEK)
-		if(T)
-			fall_to(T, A)
-			return
-		else ..()
+	New()
+		. = ..()
+		src.AddComponent(/datum/component/pitfall, 50, LANDMARK_FALL_GREEK, null, null, null, 0 SECONDS)
```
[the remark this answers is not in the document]
TheTyrant25: done

code/modules/fluids/fluid_objects.dm
pali6: Why not just call `update_targets()` here though?

pali6 (changes requested):

TheTyrant25:
ready

code/datums/components/pitfall.dm
```diff
@@ -0,0 +1,199 @@
...
+
+	// --------------- area targeting
+	/// The area path that the target falls into. For area targeting
+	var/TargetArea = null
+
+	// --------------- coordinate targeting
+	/// The z level that the target falls into if not via area or landmark.
+	var/TargetZ = 5
+	/// If truthy, try to find a spot around the target to land on in range(x).
+	var/LandingRange = 8
```
pali6, line 39: What's the point of having these on the parent. Put them on the subtypes and just override some `get_target_turf` proc on those.

code/datums/components/pitfall.dm
```diff
@@ -0,0 +1,199 @@
...
+ * Here is a note preserved from the initial commit, referring to the icemoon abyss pitfall
+ * 	// this is the code for falling from abyss into ice caves
+ *	// could maybe use an animation, or better text. perhaps a slide whistle ogg?
+ **/
+
+ABSTRACT_TYPE(/datum/component/pitfall)
+/// A component for turfs which make movable atoms "fall down a pit"
+/datum/component/pitfall
+	/// a list of targets for the fall to pick from
+	var/list/TargetList = list()
```
pali6, line 21: This too doesn't need to exist on the parent imo. It is not necessary to really have this for the landmark fall and arguably also not for the area fall. (To keep the current behaviour of z level fall it does need to exist I think but also the current behaviour of z level falls is kinda weird. Changing it is out of the scope of this PR though.)

code/datums/components/pitfall.dm
```diff
@@ -0,0 +1,199 @@
...
+	RETURN_TYPE(/turf)
+	. = src.parent
+
+/// updates targets for area/coordinate targeting. is overridden added to in child types.
+/datum/component/pitfall/proc/update_targets()
+	return
+
+/// called when movable atom AM enters a pitfall turf.
+/datum/component/pitfall/proc/start_fall(var/signalsender, var/atom/movable/AM)
+	if (!istype(AM, /atom/movable) || istype(AM, /datum/projectile/))
```
pali6, line 64:
```suggestion
	if (!istype(AM, /atom/movable) || istype(AM, /obj/projectile))
```

Heh, funny how this was incorrect in the original code too. I guess that's where the bug report of neutrons falling down holes comes from.

code/datums/components/pitfall.dm
```diff
@@ -0,0 +1,199 @@
...
+	if (istype(AM, /obj/machinery/vehicle))
+		var/obj/machinery/vehicle/V = AM
+		var/turf/target_turf = V.go_home()
+		if (V.going_home && target_turf)
+			V.going_home = 0
+			AM.set_loc(target_turf)
+			return
+
+	if (!src.TargetList || !length(src.TargetList))
+		src.update_targets()
```
pali6, line 94: Yeah basically here I'd just call some `var/turf/target = src.get_target_turf(AM)` proc and then fall to that target. Those subtypes which need a TargetList can update it if they want. Having a separate proc to get the target turf allows for more flexibility (for example it avoids the current awkward workaround for the landmark stuff).

pali6 (changes requested):

code/datums/components/pitfall.dm
TheTyrant25: i've moved it to the z level fall code section, but i've had to make it a little messy.

TheTyrant25:
can confirm that with in game testing all three kinds of pitfalls work, and bombing the seafloor to make pitfalls also works. Only thing I'm not 100% certain on is sea ladders, but I presume they work.

pali6 (approved):
Looks good at a glance and I'm too tired to do another thorough rereview

PR #16731, opened 2023-10-31
Separates the Singularity from its behavior

CaioFrancisco:
[FEATURE] [INTERNAL]
## About the PR 

This PR separates all the singularity behavior in `/obj/machinery/the_singularity` into a separate datum, `/datum/singularity_behavior`. Most procs that only relate to moving and consuming atoms are considered behavior, for the purposes of this PR.

## Why's this needed? 

This allows for future coders (me) to create singularity sub-types with special behaviors, that may or may not be radically different from the current singularity. And, since I'm removing code from the `/obj/machinery/singularity.dm` file, it gets a little more readable.


CaioFrancisco:
The code looks readable, compiles, and doesn't have any glaring glitches when I let the singularity run for a bit on DevMap. I think it's ready for a review. Oh, and the linter keeps tripping at `goonstation.dme`, although it only has one line addition. Which is weird.

Azrun:
Why not just subtypes of `/obj/machinery/the_singularity`? Why not components to ECS on an arbitrary atom?

CaioFrancisco:
> Why not just subtypes of `/obj/machinery/the_singularity`? Why not components to ECS on an arbitrary atom? Seems like subtypes should achieve that?

I asked on #imcoder about how I should have gone with doing this, and was told that I should bring the behavior to a separate datum. Or maybe I misunderstood that. I'll move back all behavior to the singularity, and keep everything as sub-types

pali6:
> Why not just subtypes of `/obj/machinery/the_singularity`? Why not components to ECS on an arbitrary atom?
> 
> > singularity sub-types with special behaviors, that may or may not be radically different from the current singularity
> 
> Seems like subtypes should achieve that?

You can't switch between subtypes at runtime. My goal for encouraging this approach was that with behaviours decoupled to datums you could do stuff such as the singularity keeping track of its "affiliation" towards possible behaviour changes and then as it exists it would switch its behaviour when some thresholds are crossed. E.g. you toss enough "explosive-aligned" items into the singulo and the singulo starts causing small explosions around it. You could not sensibly do this with subtypes and I object again to the suggestion that subtypes would be a better choice.

As for why not components - I think those would also work but at a glance I see very little benefit compared to using a datum because they would essentially be locked to only work for this one every specific object type (the singularity). If it is assumed that the behaviours would be general enough to apply to other atoms then sure components would be a bit better.

But also doing this using components adds some non-trivial amount of annoying boilerplate that has to be written. My estimate was that the benefits of using components would not outweigh the increased complexity of the code. Though if the PR author wants to use components that's fine by me.

code/datums/singularities.dm
```diff
@@ -0,0 +1,310 @@
...
+
+#define EVENT_GROWTH 3//the rate at which the event proc radius is scaled relative to the radius of the singularity
+#define EVENT_MINIMUM 5//the base value added to the event proc radius, serves as the radius of a 1x1
+
+ABSTRACT_TYPE(/datum/singularity_behavior/)
+/datum/singularity_behavior/
+	/// What singularity object is using me?
+	var/obj/machinery/the_singularity/singularity
+	/// Use the warpy effect from modern singularity. Props to amylizzle!
+	var/use_amylizzle_animation = TRUE
```
pali6, line 19: I feel like a var name like this isn't particularly descriptive. Knowing who made the animation doesn't help the person reading the code. And also it is not animated so calling it an animation is a misnomer. I'd probably name this var something like `use_distortion_effect`

code/datums/singularities.dm
```diff
@@ -0,0 +1,310 @@
...
+
+This file contains the base type and:
+- Normal Singularity
++ Katamari (normal singulo sub-type)
+*/
+
+///
+
+#define EVENT_GROWTH 3//the rate at which the event proc radius is scaled relative to the radius of the singularity
+#define EVENT_MINIMUM 5//the base value added to the event proc radius, serves as the radius of a 1x1
```
pali6, line 12:
```suggestion
#define EVENT_GROWTH 3 //! the rate at which the event proc radius is scaled relative to the radius of the singularity
#define EVENT_MINIMUM 5 //! the base value added to the event proc radius, serves as the radius of a 1x1
```
doc comments to see them on hover in vscode etc

code/datums/singularities.dm
```diff
@@ -0,0 +1,310 @@
...
+
+/// Random events the singularity does occasionaly
+/datum/singularity_behavior/proc/event()
+	return
+
+// qdeling the behavior itself makes it not gc, and to avoid race conditions,
+// i'm not nullifying the behavior on the singularity.
+
+/datum/singularity_behavior/disposing()
+	src.singularity = null
```
pali6, line 66: I feel like you should remove the distortion thing here if you added it in New.

pali6 (commented):
Not a full review, just bits I looked at. However, I wonder if it would be a good idea to let the singularity have multiple behaviours at once.

This obviously doesn't make sense for every possible property. For example whether the singularity has a distortion filter can't sensibly be handled by multiple behaviours at once. On the other hand the callback procs generally can be dispatched to multiple.

So possibly there could be /datum/singularity_behavior/main of which there would be just one on the singularity, and /datum/singularity_behavior/minor of which there would be a list. Or possibly just everything is a list and come up with some way of merging clashing things (e.g. OR of whether we want a distortion filter or not).

Feel free to disregard this if you don't want to code it. It's just an idea.

CaioFrancisco:
> Feel free to disregard this if you don't want to code it. It's just an idea.

I got the time **and** willingness to learn proper coding. I'll code that in when I touch VSCode again.

code/datums/singularities.dm
CaioFrancisco: oh, right, this reminds me to think up of a clean way to GC the behavior without deleting the singularity, which is one of the main points behind the design of the datum behavior.

pali6:
Apologies for the conflict. Ping me if there is any trouble with it.

CaioFrancisco:
just after i typed that, school gave me 5 school works at once, and scheduled a bunch of exams. i'll have to pause this project.

CaioFrancisco:
Laziness got the better of me. I'll re-open this when I get the drive to code this again. I'd set this to Adopt Me but I dunno if any of my code is useable.

PR #16732, opened 2023-10-31
Minimum distance option for sound hallucinations and option to have image override hallucinations appear only out of view

MylieDaniels:
[A-Internal]{C-Feature]
## About the PR 
`random_sound` hallucinations now have an argument called min_distance, default -1 (no minimum distance, old behavior), that provides a limit on how close the sound can originate from.

`random_image_override` hallucinations now have an argument called visible_creation, default TRUE, which if FALSE causes images to only appear in spots within range that can NOT currently be viewed by the affected mob.
I did not add this to `random_image` hallucinations, as those appear to be used as a lighter version of random_image_override when only the barebones features are needed.

As a minor demonstration, the screaming madness artifact now has a minimum distance of 5, and the flock and ghost madness artifacts now only override human images outside view.

## Why's this needed? 
Honestly, I'm working on hallucination things for my own adventure/prefab stuff and am atomizing this out of that.  I suppose it just helps make more convincing hallucinations?

### DIFF
--- a/code/datums/components/hallucinations.dm
+++ b/code/datums/components/hallucinations.dm
@@ -138,7 +138,7 @@ ABSTRACT_TYPE(/datum/component/hallucination)
 				parent_mob.playsound_local(origin, chosen, 100, 1)
 		. = ..()
 
-	CheckDupeComponent(timeout, sound_list, sound_prob)
+	CheckDupeComponent(timeout, sound_list, sound_prob, min_distance)
 		if(sound_list ~= src.sound_list) //this is the same hallucination, just update timeout and prob
 			if(timeout == -1)
 				src.ttl = timeout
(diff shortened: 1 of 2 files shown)


code/datums/components/hallucinations.dm
```diff
@@ -111,21 +113,26 @@ ABSTRACT_TYPE(/datum/component/hallucination)
...
 	do_mob_tick(mob, mult)
 		if(probmult(src.sound_prob))
 			var/atom/origin = parent_mob.loc
 			var/turf/mob_turf = get_turf(parent_mob)
 			if (mob_turf)
-				origin = locate(mob_turf.x + rand(-10,10), mob_turf.y + rand(-10,10), mob_turf.z)
+				if(src.min_distance > 0)
+					origin = locate(mob_turf.x + pick(rand(-10,-src.min_distance),rand(src.min_distance,10)), mob_turf.y + pick(rand(-10,-src.min_distance),rand(src.min_distance,10)), mob_turf.z)
+				else
+					origin = locate(mob_turf.x + rand(-10,10), mob_turf.y + rand(-10,10), mob_turf.z)
```
Tarmunora, line 135: Does this need to be split like this, as opposed to just defaulting `min_distance` to 0, which should match current behaviour?

Tarmunora (changes requested):

code/datums/components/hallucinations.dm
MylieDaniels: I'll change that since it would indeed match current behavior, I was overthinking it.

Tarmunora: still need to set the default to `0` instead of `-1`

Tarmunora (approved):

PR #16733, opened 2023-10-31
Borg-Fishing-Addendums-2. More Betterer.

Francisco-Franks:
[Silicons] [Fishing] [Fish]

## About the PR 
Attempt 2: Borg fishing rod is upgraded by eating the fishing rod from the vendor. Only can go up, not down. Borgs now have a smaller portable aquarium so they can deposit fish in the research machine. Borgs/Silicons can now interact with the vending machine. The research tickets can be click-dragged on the ground to stack them, and click-dragged into the vending machine. Fishing rod upgrades affect the icon correctly.  Eventually, the cyborg rod will have different sprites, or perhaps be swappable with ingame (custom / plated) fishing rods.

## Why's this needed? 
Silicons were incapable of doing any real deal fishing. They could fish in the most basic places, couldn't do research on their catches, or interact with the vending machine AT ALL. This fixes that. It is a directly parallel upgrade path for the fishing rods compared to normal human fishing as well. Nothing bespoke here.

## Changelog 

```changelog
(u)Metareferencer
(+)Silicon fishing improvements. Go fish!
```


code/modules/fishing/fishing_gear.dm
```diff
@@ -179,6 +193,32 @@
...
+			src.item_state = "fishing_rod-active"
+		else if ((!src.is_fishing) && (src.tier == 1))
+			src.icon_state = "fishing_rod-inactive"
+			src.item_state = "fishing_rod-inactive"
+		else if ((!src.is_fishing) && (src.tier == 2))
+			src.icon_state = "fishing_rod_2-inactive"
+			src.item_state = "fishing_rod-inactive"
+		else
+			src.icon_state = "fishing_rod_3-inactive"
+			src.item_state = "fishing_rod-inactive"
```
Tarmunora, line 221:
```suggestion
		var/isactive = src.is_fishing ? "active" : "inactive"
		src.icon_state = "fishing_rod[src.tier > 1 ? "_[src.tier]" : ""]-[isactive]"
		src.item_state = "fishing_rod-[isactive]"
```

Giant elif block can be simplified a bit with some code

code/modules/fishing/fishing_gear.dm
```diff
@@ -33,6 +33,20 @@
...
+				user.visible_message("<span class='notice'>You upgrade your [source] with the Upgraded Fishing Rod</span>")
+				qdel(target)
+				return
+			if ((((istype(target, /obj/item/fishing_rod/master)) && src.tier < 3))&& istype(source, /obj/item/fishing_rod/cybernetic))
+				src.tier = 3
+				src.icon = target.icon
+				src.icon_state = target.icon_state
+				user.visible_message("<span class='notice'>You upgrade your [source] with the Master Fishing Rod</span>")
+				qdel(target)
+				return
```
Tarmunora, line 49: would it not make more sense to define this behaviour on the cybernetic rod itself?

Tarmunora (changes requested):

code/modules/fishing/fishing_gear.dm
Francisco-Franks: Given the structure of the rest of the fishing module it looked like "Keep the sub-items cleanish, put the logic in the root item" meant sense. I can make a proc in the root item that calls to the cybernetic fishing rod to do that logic before continuing on with the rest of the attackby_pre proc but would we really gain all that much?

code/modules/fishing/fishing_gear.dm
Francisco-Franks: Well, that makes things simpler. Thanks!

Tarmunora (approved):

PR #16742, opened 2023-11-01
Medbay Chemical Byproducts

Flaborized:
[CHEMISTRY] [WIP]

## About the PR 

This draft PR is an idea I had quite a bit ago but started making now. I figure I might change some stuff that's in it and adjust it much more so I'll also describe the main idea:

Core concept is that the cloner's reclaimer and cyro tubes produce chemicals as byproducts when used, which are specifically useful to chemists. Doctors don't actually *have* to remove the byproduct, it doesn't impact the usage of the reclaimer/cryo. This is gathered in a new "byproduct can" item that you can smack onto them (they'll both start with one), these are removed with a medical locker access having ID. The recipes should be hard for doctors to utilize on their own, so they're incentivized to give it to chemists when the byproduct cans fill up.

Currently the only I've made is the reclaimer (need to figure out a way to make cryo tubes generate their byproduct when in use healing that doesn't encourage you to just put a monkey in there to farm it), grinding up anything will fill an attached byproduct can with a "genetic sludge" reagent.

Genetic sludge reacts with carbon to make oil and epinephrine, currently 1:1:1 (you don't get tons and tons of sludge though), but you need at least 10u sludge to start the reaction, and sludge reacts with oil to Explode. It also has a reaction speed of 1u, so you basically *need* a condenser to use any of the stuff, to encourage chemistry to use this instead of medical themselves using it.

byproduct_can

## Why's this needed? 

I think it'd be good for incentivizing the departments to work together if medical had something to give to chemistry, basically. It being a two-way process leads to more interactions. Also the chemical rework wants you to make byproducts like oil in the background so you have it for later, instead of just making it when you need it. With this semi-random burst of oil coming in, that pushes you towards that behavior a bit. Hopefully it won't overshadow the existing recipe, and will instead be a nice supplement to it when medbay is cooperating with you (and when the cloner works).

Also working together can be fun and satisfying, wouldn't it be cool if chemists and medbay were good friends

## TODO:
- Make the cryo tube byproduct work, add recipes to whatever reagent goes in it
- Add sounds to the process in general, adding or removing byproduct can should make a Clunk sound
- Tweak the actual reagent amounts and production volume so it doesn't overshadow regular oil mixing, but is still useful
- Read feedback (that's why I posted this early into making this) and adjust the design a bit if issues are pointed out
- Add working overlays for chemical amounts (and when attached to machines)
- Probably some other stuff I'm forgetting
- Also add a way to print these cans in medical fabricators
- To reiterate the chemical amounts produced in the recipe and stuff are WIP, haven't really tuned it at all and am more just Making The Thing firstly and fixing it up later
- thank you

## Changelog 

```changelog
(u)Flaborized
(*)Will write this later when PR is done
(+)
```


TDHooligan:
honest thoughts:
doctors will do this themselves. they will ask for a condenser.
this explosion is a trap for people without condensers and also doesn't really make sense. 
backflow will explode everything. this is annoying. don't make chemist's lives harder than they already are. make it react back into itself as losing carbon is fine.
it just feels really forced??? like it's 'interdepartmental cooperation juice', i get it for the enzymatic reclaimer but it's functionally (and literally) nailed onto cryo too

Flaborized:
> honest thoughts: doctors will do this themselves. they will ask for a condenser. this explosion is a trap for people without condensers and also doesn't really make sense. backflow will explode everything. this is annoying. don't make chemist's lives harder than they already are. make it react back into itself as losing carbon is fine and still a punishment in terms of wasted time. it just feels really forced??? like it's 'interdepartmental cooperation juice', i get it for the enzymatic reclaimer but it's functionally (and literally) nailed onto cryo too

Doctors using it for themselves *sometimes* is fine, though oil is much more useful for chemists than it is doctors and doctors already have quite a bit on their plate. This produces an even quantity of oil to sludge, so backflowing in a condenser is very unlikely if you're running an already empty beaker anyways, this doesn't overproduce oil and I don't think will work well in more dedicated setups because of how inconsistent getting the sludge is. 

I understand that it feels a bit meta but I think both the idea tone-wise of "taking chemicals technically from crewmember's bodies" is good and they're fine sources to generate extra stuff. Hopefully something extra doctors can keep track of and utilize if they want. I'm open to it being something other than an explosion but I'd like misusing it to have some Bad Effect on the world, instead of just wasting your product. Could be something less severe though

TDHooligan:
i dunno. we have a lot of bad effects in chemistry right now and it feels like they're all beating on people sussing out early mechanics; before, it was always for obviously stupid things like spamming chems aimlessly, or deliberately heating stupid things up (because everything was 20c). 

this is just "you explode if you dont realize this needs a condenser" which sucks. condensers aren't the most obvious thing to a new player. and this could just be a simple recipe that helps players learn that condensers exist and is useful when they ask 'why is this reaction looping back into byproducts'

change what comes out to be something medbay actually runs out of, make the recipe super simple but need something like a bunsen and a condenser. if it's easy enough, medbay staff will probably teach the scientists how to use condensers rather than futz around with getting in.

i would note that a lot of our regular doctors are experienced chemists as well from my experience. they are just as capable of using oil as a scientists.

Flaborized:
on second thought I do think removing the explosive aspect is probably a good idea considering lots of stuff explodes already, but I'd like for it to at least do *something* when messed up, maybe not so severe (or maybe a neutral-ish thing you could take advantage of). I know that medical often understands chemistry but I'm more speaking to oil being less mandatory feeling for medical doctors than chemists. Tbh exploding was more of a filler effect since I made the recipe pretty quickly just to get something down. I'll iterate on this more and make the cryo pod stuff and try and cook up some more fitting recipes for the two outputs.

Flaborized:
Ok pushed a few changes, the numbers and stuff are still WIP for the cryogenic sludge recipes since it needs some fine tuning to work. 

- Changed the genetic sludge recipe to not blow up with oil, instead sludge and oil make welding fuel together. This isn't as *good* as fast oil but it's not necessarily terrible and could be useful too. I think the recipes having multiple uses that doctors could theoretically make use of themselves too makes them feel less "meta", like they're *just* for cooperation.
- Also added cryogenic sludge, produced if you use cryoxadone when healing someone in a cryo tube. The person has to be injured and the beaker attached to the cryo tube has to be draining cryoxadone for it to be made.
- Mixing cryogenic sludge with nitrogen gives you cryoxadone back, the nitrogen isn't consumed
- Mixing cryogenic sludge with nitrogen *and oil* makes omnizine (and cryoxadone), but the conditions are a bit harsh. The reaction is quite slow, and you need to keep the oil levels between 1u or 15u. If it goes over or under that more cryogenic sludge gets used up and it mixes cryostylane smoke if it's an open container. 
- Also the omnizine recipe creates the aoe cold effect that cryostylane and iodine make every 3 cycles, which means your oil condenser (which must be in range) will cool off over time, so you need to heat it somehow gradually. Also it only consumes oil every third cycle so it's easier to keep a consistent amount between the high/low threshold.
- Also you have to make sure your oil heating setup doesn't keep running after the reaction ends cuz it'll overheat the fuel and explode

In general I wanna tweak this so setting up two condensers and a bunsen burner will be enough to run this, though with some hazards if you leave it running too long. The omnizine output is deliberately pretty low atm since it's such a potent chem even in small amounts, and nothing particularly about the recipe is set in stone, I've just adjusted it a lot in-game and wanted to push this for more feedback before going back in and tweaking it more. The theme is a bit more towards "useful chemicals that chemists can fully 'unlock' the potential of" and less "Thing that's designed for chemists to use" hopefully

PR #16744, opened 2023-11-02
Take extra death rolls when you're being beaten on

TDHooligan:
## About the PR 
Players in death saves take an extra death roll when they take 5 or more BRUTE, the damage dealt provides a logarithmic penalty to their success rate.

- This only rolls when you're taking damage and would already be rolling for death
- The roll can only be triggered once every 0.4 seconds (to limit rolls from rapid-fire weapons)
- It ignores synatropine 
- It probably needs adjusting, depending on the results of a test merge (if any)
- The penalty is 5% at 5 damage, requiring 2x additional damage to raise another 5% 
(5 = 5%, 15 = 10%, 35 = 15%, 75 = 20%)

image

## Why's this needed? 
Death rolls in crit are balanced around a few things:
 - They occur every life tick (so need to be a low % to not just kill you instantly)
 - People should have time to be treated, given they aren't very unlucky
 - Brain damage and OXY are introduced in huge quantities to adjust survival odds after 30 seconds or so

This causes some weird behaviour:
 - If you lack a crazy weapon, beating people down in deep crit provides little to no benefit (1 punch = +0.05% death odds)
 - Repeatedly point-blanking people with powerful kinetics can waste a LOT of ammo (ie if life ticks are slow)
 - Smacking people with strong weapons, when they are most certainly are dead, can feel frustrating if they have good RNG
 - With synatropine, you can tank obscene damage well past the 'you lose' threshold, RNG permitting. 

I do, however, know that we like a good comeback - I am open to lowering the odds.

## Changelog 

```changelog
(u)TDHooligan
(*)Brute damage now causes people in deep critical to roll extra death saves.
```


TDHooligan:
i forgot to write it only rolls on 5+ brute

Tarmunora (approved):

TDHooligan:
i don't see many negative comments about this - from my perspective the numbers could be good as-is.

PR #16767, opened 2023-11-03
Changes escaped double quotes to single quotes in html strings

CheffieGithub:
## About the PR 
Changes html classes to use single quotes within strings, and other tags that I saw while I was doing classes in those files.

## Why's this needed? 
Massively inconsistent and it all being one type of thing makes searching / regexing it for span macros easier later.

### DIFF
--- a/code/modules/materials/Mat_Processing.dm
+++ b/code/modules/materials/Mat_Processing.dm
@@ -357,7 +357,7 @@ TYPEINFO(/obj/machinery/processor)
 
 	attack_hand(mob/user)
 		var/list/html = list("")
-		html += "<div style='margin: auto;text-align:center'>[first_part ? "<a href='?src=\ref[src];remove=\ref[first_part]'>[first_part.name]</a>" : "EMPTY"] <i class='icon-plus'[]></i> [second_part ? "<a href='?src=\ref[src];remove=\ref[second_part]'>[second_part.name]</a>" : "EMPTY"]   <i class='icon-double-angle-right'></i> [resultName]</div><br>"
+		html += "<div style='margin: auto;text-align:center'>[first_part ? "<a href='?src=\ref[src];remove=\ref[first_part]'>[first_part.name]</a>" : "EMPTY"] <i class='icon-plus'></i> [second_part ? "<a href='?src=\ref[src];remove=\ref[second_part]'>[second_part.name]</a>" : "EMPTY"]   <i class='icon-double-angle-right'></i> [resultName]</div><br>"
 		html += "<div style='margin: auto;text-align:center'><a href='?src=\ref[src];activate=1'><i class='icon-check-sign icon-large'></i></a></div><br><br>"
 
 		for(var/obj/item/I in src)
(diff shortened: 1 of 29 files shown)


CheffieGithub:
I don't really know what the check fail is all about, I tested the thing its complaining about (nano crucible start button) and it works the same so a linter issue?
If it was an issue with single quote usage I would expect the above line and more to also have a warning.

code/modules/materials/Mat_Processing.dm
```diff
@@ -357,8 +357,8 @@ TYPEINFO(/obj/machinery/processor)
 
 	attack_hand(mob/user)
 		var/list/html = list("")
-		html += "<div style=\"margin: auto;text-align:center\">[first_part ? "<a href='?src=\ref[src];remove=\ref[first_part]'>[first_part.name]</a>" : "EMPTY"] <i class=\"icon-plus\"></i> [second_part ? "<a href='?src=\ref[src];remove=\ref[second_part]'>[second_part.name]</a>" : "EMPTY"]   <i class=\"icon-double-angle-right\"></i> [resultName]</div><br>"
-		html += "<div style=\"margin: auto;text-align:center\"><a href='?src=\ref[src];activate=1'><i class=\"icon-check-sign icon-large\"></i></a></div><br><br>"
+		html += "<div style='margin: auto;text-align:center'>[first_part ? "<a href='?src=\ref[src];remove=\ref[first_part]'>[first_part.name]</a>" : "EMPTY"] <i class='icon-plus'[]></i> [second_part ? "<a href='?src=\ref[src];remove=\ref[second_part]'>[second_part.name]</a>" : "EMPTY"]   <i class='icon-double-angle-right'></i> [resultName]</div><br>"
```
pali6, line 360:
```suggestion
		html += "<div style='margin: auto;text-align:center'>[first_part ? "<a href='?src=\ref[src];remove=\ref[first_part]'>[first_part.name]</a>" : "EMPTY"] <i class='icon-plus'></i> [second_part ? "<a href='?src=\ref[src];remove=\ref[second_part]'>[second_part.name]</a>" : "EMPTY"]   <i class='icon-double-angle-right'></i> [resultName]</div><br>"
```
This is the cause for the linter failure. It was just off by one line.

pali6 (changes requested):

code/procs/logged_phrases.dm
```diff
@@ -156,19 +156,19 @@ var/global/datum/phrase_log/phrase_log = new
...
 		#endif
 		if(is_uncool(phrase))
 			var/ircmsg[] = new()
 			ircmsg["key"] = user.key
 			ircmsg["name"] = (user?.real_name) ? stripTextMacros(user.real_name) : "NULL"
-			ircmsg["msg"] = "triggered the uncool word detection: [category]: \"[phrase]\""
+			ircmsg["msg"] = "triggered the uncool word detection: [category]: '[phrase]'"
 			SPAWN(0)
 				ircbot.export("admin", ircmsg)
-			SEND_GLOBAL_SIGNAL(COMSIG_GLOBAL_UNCOOL_PHRASE, "<span class=\"admin\">Uncool word - [key_name(user)] [category]: \"[phrase]\"</span>")
```
pali6, line 171: These re actual quotation marks displayed in a message. Are we sure we want to change that?

CheffieGithub: oh yeah these ones are a mistake

pali6:
Fwiw I'm pretty sure you'd also find old code that uses no quotes at all. E.g. `<FONT SIZE=+1>`

PR #16770, opened 2023-11-04
Remove achievement requirement for Flock structures

FlameArrow57:
[GAME MODES][GAME OBJECTS][REMOVAL][BALANCE]

## About the PR 
This PR removes the achievement requirement from Flock structures, meaning you can unlock all buildable ones other than the Relay immediately.

Not sure regarding the theme of Flock, but I think it would still make sense thematically, that they would be aware of what harms them and what they can utilize. Thinking of Interceptor and Sapper.

## Why's this needed? 
Couple of reasons-
-May have been an interesting way to unlock structures, but in practice with the fast-paced gameplay of Flock and the time to build structures, they just never get used. For any reason, maybe players forget with how much stuff is going on, or just don't see the chat messages for unlocks, they aren't really used
-Assuming Flock is winning, I feel like waiting on specific tasks doesn't really help them much as they are going to be winning without these structures anyways. This would help reduce Flock needing to survive longer than necessary to unlock structures
-Hopefully allows Flock's structures to be utilized more
-Allows Flock to play some more into the building side of RTS, allowing which structures they want immediately, location, and combinations

## Changelog 
```changelog
(u)FlameArrow57
(*)Flock structures no longer require specific tasks to be performed and start off available immediately.
```

### DIFF
--- a/_std/macros/flock.dm
+++ b/_std/macros/flock.dm
@@ -36,8 +36,6 @@
 // achievements
 #define FLOCK_ACHIEVEMENT_CHEAT_STRUCTURES "all_structures"
 #define FLOCK_ACHIEVEMENT_CHEAT_COMPUTE "infinite_compute"
-#define FLOCK_ACHIEVEMENT_CAGE_HUMAN "human_dissection"
-#define FLOCK_ACHIEVEMENT_BULLETS_HIT "bullets_hit"
 
 #define FLOCK_BULLETS_HIT_THRESHOLD 15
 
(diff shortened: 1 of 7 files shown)


TobleroneSwordfish (approved):
Yeah I think this is reasonable, we never did a very good job of communicating what the unlock requirements are and it does seem like a system built for a much slower style of gameplay 👍

amylizzle (approved):
yeah pretty much what Leah said. Seems like a good change.

PR #16772, opened 2023-11-04
Move Counter-Revolutionary Implants from QM to Security

FlameArrow57:
[GAME MODES][GAME OBJECTS][BALANCE][INPUT WANTED][QoL]

## About the PR 
This PR removes counter-revolutionary implants as a QM purchase. 

A counter-revolutionary implant dispenser is now found in the Security department on each map. Each can hold a max of 7 implants and will auto-fabricate implants every 12 seconds up to its max.

Also removes a single loyalty implant box from each map to account for the dispenser.

Maybe it goes against the theme of the station to buy new implants, not sure.

## Why's this needed? 
This is mainly meant to counter revs going for QM deconstruction. If QM computers are deconstructed, and the spare board in tech storage is destroyed, the only other way to get new implants is to mechscan SecTechs, but that's rarely done/thought of too.

This helps add a way of preventing revs from completely shutting down obtaining implants.

Also, probably a lot of ways you can think about how this would change things, but I would imagine it would also help reinforce Security as a place to be wary of, and that revs really need to group up on to attack. Maybe it will allow some more interesting fights when attacking Security?

## Changelog 
```changelog
(u)FlameArrow57
(*)Counter-revolutionary implants are no longer a QM purchase. Instead, a new counter-revolutionary implant dispenser is found in each Security department, which holds up to 7 implants and auto-fabricates implants every 12 seconds.
```


DisturbHerb:
There would be minimal counterplay against Security's ability to keep order. While QM destruction can be very cheesey as a tactic, I don't think this is the solution myself.

Boredone:
If nothing else, only 12 seconds per implant is kinda absurdly fast? That's more then the crate every minute if folks empty it fast enough.

FlameArrow57:
> If nothing else, only 12 seconds per implant is kinda absurdly fast? That's more then the crate every minute if folks empty it fast enough.

Maybe, dunno. Average of 5 per minute, each crate comes with 4 but sometimes they'll be bought in batches.

PR #16779, opened 2023-11-04
Simplify Flock Gnesis Turret compute cost

FlameArrow57:
[GAME MODES][GAME OBJECTS][BALANCE][QoL]

## About the PR 
This PR removes the changing compute cost difference for gnesis turrets when they switch between making darts and being idle

## Why's this needed? 
-Cool idea in theory, but with how rapidly Flock can switch compute I think it just makes it a bit more confusing to manage knowing the compute costs of your structures.
-Reduces complexity, I think just a single cost would make it simpler
-Makes it easier to know which structures cost how many compute


amylizzle (approved):
lgtm

TobleroneSwordfish (approved):

PR #16789, opened 2023-11-04
Move medal procs to /datum/player

pali6:
## About the PR 
For organization purposes it is a good idea to have these unified and on the datum which represents potentially logged out clients. This will also help with potential API 2 stuff.


pali6 (commented):
Tested on Dev, seems to work fine.

PR #16797, opened 2023-11-05
Add help message for wrapping paper

gear-space:
[Game-Objects][QoL]
## About the PR 
Adds a help message to wrapping paper explaining how to use it, with a little extra help if user is an antag.

Antag message
image

## Why's this needed? 
Without a tooltip, users unfamiliar with the process are more likely just to hit things on the floor with wrapping paper and give up. Also more gift wrapped bodies should be sent to security. More people spreading dubious amounts of cheer is better!

### DIFF
--- a/code/obj/item/gift.dm
+++ b/code/obj/item/gift.dm
@@ -14,6 +14,11 @@
 	tooltip_flags = REBUILD_DIST
 	var/style = "r"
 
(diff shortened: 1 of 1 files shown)


pali6:
I feel like the extra bit could be shown to non-antags too.

PR #16798, opened 2023-11-05
Implement API2 for player medals

Wirewraith:
## About the PR 

Use API2 for player medals so we have more control over them than Byond provides.


Wirewraith:
Marked as draft until https://github.com/goonstation/goonstation/pull/16789 is merged to apiv2 branch

code/modules/goonhub/api/models/medals/MedalResource.dm
```diff
@@ -0,0 +1,25 @@
+
+/// MedalResource
+/datum/apiModel/Tracked/MedalResource
+	var/title				= null // string
+	var/description	= null // string
```
ZeWaka, line 5: What about the images?

Wirewraith: Image isn't a stored column. It's a URL derived from the medal ID. e.g. `/storage/medals/1.png`

ZeWaka (approved):

PR #16817, opened 2023-11-06
Fix lamp manufactuerer context menu

### LINKED ISSUE #16808: miniature lamp manufacturer context menu not responding
### Labels

[BUG][MINOR]

### Description

cannot select colors or type (bars, bulb, remove)

### Steps to reproduce

1. press C with manufacturer in hand
2. click on radial menu buttons (colors, types, all buttons dont work)

### Expected Behavior

should be able to select new color or type of light

Reported by: OvidNaso
Active test merges: #16407
Round log date: 2023-11-06-03-33
Reported on: 2023-11-06 04:38:11
Map: COGMAP

pgmzeta:
## About the PR 

checkRequirements has to return TRUE if it's valid, so change the check to return the result of checks.

For the vast majority of other checks, no change is needed as there are other criteria that can evaluate to true, but in this instance it is just the can_use and range check.

## Why's this needed? 

Fix #16808

### DIFF
--- a/code/modules/fluids/fluid_objects.dm
+++ b/code/modules/fluids/fluid_objects.dm
@@ -375,10 +375,8 @@ TYPEINFO(/obj/machinery/fluid_canister)
 			return
 		fluid_canister.change_mode(src.mode)
 
-	checkRequirements(var/obj/machinery/fluid_canister/fluid_canister, var/mob/user)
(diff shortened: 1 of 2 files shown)


pali6:
I knew writing this at 3 AM was a bad idea. Mucha gracias.

PR #16820, opened 2023-11-06
Span macros to replace alot of spans

CheffieGithub:
## About the PR 
Replaces most `<span ='class'>msg</span>` with `SPAN_CLASS(msg)`. There are limitations of this such as not being able to be used in descriptions unless the entire description is in the macro for example (since the macro evaulates to just a string).
Some spans I intentionally left out due to only using local css for one file.

I used these regexes in this order:
```
"<span class='([a-zA-Z_]+)'>(.*?)</span>" // Spans encasing whole strings
SPAN_\U$1("$2")
```

```
<span class='([a-zA-Z_]+)'>(.*?)</span> // Spans in strings
[SPAN_\U$1("$2")]
```
```
<span class='([a-zA-Z_]+)'>(.*?)"\) // Unclosed span tags
```
Any errors were manually resolved (this includes 100+ missing / malformed tags and most instances of `text()` usage)

## Why's this needed? 
Mistakes while writing spans such as missing or malformed end tags are very common.
It should make some areas of code easier to read and shorter depending what classes are used.
Its somewhat easier to maintain this way.

## Changelog 

```changelog
(u)Cheffie
(*)Big internal styling refactor, please report any styling errors you encounter in the chat!
```


_std/macros/chat_output.dm
```diff
@@ -0,0 +1,51 @@
+/* ====== SPAN MACROS ====== */
```
pali6, line 1: Unless there is something preventing that I think these should use string interpolation instead of +.

PR #16821, opened 2023-11-07
Light Fixture Resprite

Flaborized:
## About the PR 

Resprites all three lighting fixtures, adds an overlay that shows a light tube inside the fixture regardless of what material the light is made of

light_fixtures_png

## Why's this needed? 

Pali added bulb fixtures to the wanted resprite list, and I drew these versions which I think look better. And the bulbs showing up reflect the actual usage of the item better since it *comes* with a light inside

## Changelog 

```changelog
(u)Flaborized
(+)Lighting fixtures now properly show a bulb in them.
```


code/modules/power/lighting.dm
```diff
@@ -23,6 +23,19 @@ TYPEINFO(/obj/item/light_parts)
 	var/light_type = /obj/item/light/tube
 	var/fitting = "tube"
 	var/install_type = INSTALL_WALL
+	var/image/light_image = null
+
+	New()
+		..()
+		update_light_tube_overlay()
+
+	proc/update_light_tube_overlay()
```
pali6, line 32: Could likely be override of `update_icon` (which is called by the wrapper `UpdateIcon`). This is useful in case something external modifies us etc.

Also `light_image` does not need to be a var on the object. It's ok to make it a local variable here or even better use `src.SafeGetOverlayImage("light")`. That proc will reuse the last image used for this overlay or create a new one if there is no such overlay.

WALPVRGIS (approved):
Just a sprite review: These look SO much better, great artwork 😸

code/modules/power/lighting.dm
Flaborized: Yeah this makes much more sense, ty pushed these changes

PR #16844, opened 2023-11-08
More span macros and updates

CheffieGithub:
## About the PR 
uses the same regex from https://github.com/goonstation/goonstation/pull/16820

Organise `chat_output.dm`
Adds doc comments for the macros to give a preview of the css styling.
Removes the 'game' class reference from everything that still used it (seemingly empty class with no use or meaning) so that those messages can be macro'd.
Removes non chat compatible span macros that I made accidently so `chat_output.dm` actually only applies to chat output.
`SPAN_NOTIFICATION` & `SPAN_INFO` --> `SPAN_NOTICE`
`SPAN_WARNING` --> `SPAN_ALERT`

## Why's this needed? 
Previews without needing to go to the css file is good. Having unused classes is bad. Having chat output actually be just chat output is good.


ZeWaka:
get merge conflicted lmao

PR #16846, opened 2023-11-09
Allow mouse wheel scrolling to set target body attack zone

FlameArrow57:
[PLAYER ACTIONS][QoL]

## About the PR 
This PR makes it so that scrolling the mouse wheel will set the target body zone that you attack

I saw that mouse wheel implementation in the code is hardly used, which would make this simpler to implement, and I think mouse wheel scrolling is pretty fast, compared to an action like hitting multiple hotkeys or a single hotkey to cycle between zones.

Performance when scrolling the mouse wheel to set the zone:
image
Performance when scrolling over at atom that implements MouseWheel is roughly the same.

Things to note that I found-
-client/MouseWheel() is not called at all when in a scrollable interface, like the chat log, or a TGUI interface
-It is called most everywhere else (I think)

## Why's this needed? 
Currently there's 6 zones you can select, including chest, arms, legs, and head. This makes it so that in order to set your target attack zone faster than clicking the HUD zones, you either have to override more important hotkeys, or angle your hand in a painful way to reach the hotkeys for the zones you want to attack which is even more difficult when moving around

## Changelog 
```changelog
(u)FlameArrow57
(*)Scrolling the mouse wheel will now set the target body zone you are attacking. Scrolling downwards is in the order head -> chest -> left arm -> right arm -> left leg -> right leg, without wrap around. 
```

### DIFF
--- a/_std/plane.dm
+++ b/_std/plane.dm
@@ -87,6 +87,10 @@ client
 	var/list/plane_parents = list()
 	var/list/plane_displays = list()
 	var/atom/movable/screen/plane_display/master/game_display
+	// list of current zone sel to the next zone sel if you scroll up
+	var/static/list/zone_sels_positive_delta = list("chest" = "l_arm", "l_arm" = "r_arm", "r_arm" = "l_leg", "l_leg" = "r_leg", "r_leg" = "head", "head" = "chest")
+	// list of current zone sel to the next zone sel if you scroll down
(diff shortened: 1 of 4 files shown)


kbsmilk:
Golly, the greatest katana buff there ever was...
Nah really though, this is great. There are probably already folks using outside software to "rebind" this, so this levels the playing field.

Sidenote though, I would make it NOT loop. It helps muscle memory if you can jam the wheel all the way up/down to "reset" your position to an end-point, and work from there. Like having a home position, you know?

ZeWaka:
This order is weird. Why not top down or bottom up?

FlameArrow57:
Sure

pali6 (approved):
I feel like chest should be inbetween the arms but that might just be me.

PR #16862, opened 2023-11-10
Fix some items not being addable to artifact bags of holding

### LINKED ISSUE #16834: Artifact Bag of Holding and multitools
### Labels

[BUG][MINOR]

### Description

Multitools won't go into activated bags of holding, even if they can fit. (doesn't give cannot fit message)

### Steps to reproduce

1. Acquire a bag of holding
2. try putting a multitool in it

### Expected Behavior

Well it should work but doesn't! I expected as much though.

Reported by: Poglodyte
Active test merges: #16744 and #16370
Round log date: 2023-11-08-02-56
Reported on: 2023-11-08 03:42:43
Map: COGMAP

FlameArrow57:
[GAME OBJECTS][PLAYER ACTIONS][BUG]

## About the PR 
This PR fixes some items, like multitools, not being able to be added to artifact bags of holding.

Also, activator keys will now have the same effect when added to a bag of holding with an origin they can activate as adding another bag of holding.

## Why's this needed? 
Bug fix, fixes #16834

## Changelog 

```changelog
(u)FlameArrow57
(+)It's no longer advised to put an activator key into a bag of holding it can activate!
```

### DIFF
--- a/code/obj/artifacts/artifactprocs.dm
+++ b/code/obj/artifacts/artifactprocs.dm
@@ -262,12 +262,14 @@
 					if(K.corrupting && length(A.faults) < 10) // there's only so much corrupting you can do ok
 						for(var/i=1,i<rand(1,3),i++)
 							src.ArtifactDevelopFault(100)
-				else if (A.activated)
+					// prevent instantly adding to contents, since a bad effect happens
 					if (istype(src, /obj/item/artifact/bag_of_holding))
-						if (!user || user.a_intent == INTENT_HARM)
-							src.ArtifactDeactivated()
-					else
-						src.ArtifactDeactivated()
+						return
+				else if (A.activated)
+					if (istype(src, /obj/item/artifact/bag_of_holding) && src.storage.check_can_hold(W) == STORAGE_CAN_HOLD)
+						src.storage.add_contents(W, user)
+						return
+					src.ArtifactDeactivated()
 
 	if (isweldingtool(W))
 		if (W:try_weld(user,0,-1,0,1))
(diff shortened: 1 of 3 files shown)


code/obj/artifacts/artifactprocs.dm
```diff
@@ -263,7 +263,11 @@
 						for(var/i=1,i<rand(1,3),i++)
 							src.ArtifactDevelopFault(100)
 				else if (A.activated)
-					src.ArtifactDeactivated()
+					if (istype(src, /obj/item/artifact/bag_of_holding))
+						if (!user || user.a_intent == INTENT_HARM)
+							src.ArtifactDeactivated()
```
Tarmunora, line 268: Intent-based interactions kinda suck. Would rather just have deactivators not be storable at all in bags of holding that they'd deactivate

Tarmunora (changes requested):

code/obj/artifacts/artifactprocs.dm
FlameArrow57: Sure. Alternate idea, maybe allow them to be added by clicking the activator key with the bag of holding?

Tarmunora: comedy option: let it go in but then detonate the bag a la putting a wand of cancellation in a bag of holding in nethack 😈

FlameArrow57: I went with tying it into the existing bag of holding negative effects, so adding a bag of holding or an activator key will do a bad effect

PR #16867, opened 2023-11-10
Jumpseat? Life Support-Supported Cargo Hold

Azrun:
## About the PR 
Adds a pod subsystem: Living Cargo hold
A modified cargo hold intended to support the life of one individual.  Pulls air, at the cost of power, from the Pod's Life-Support system.

## Why's this needed? 
Intended to provide some flexibility in transporting individuals for Salvagers.  Transporting the team or possibly bringing stuff back?

Changelog omitted as this is not added to the game directly yet.


code/modules/transport/pods/secondary_system.dm
```diff
@@ -325,6 +328,56 @@
 			break
 	..()
 
+/obj/item/shipcomponent/secondary_system/cargo/jumpseat
+	name = "Living Hold"
+	desc = "Allows the ship to load people as cargo. It is designed to integrate with the Life-Support System so it is probably safe?."
```
Tarmunora, line 333: Name feels a little weird, can't put my finger on it
also you have a `?.` at the end of the desc

code/modules/transport/pods/secondary_system.dm
```diff
@@ -325,6 +328,56 @@
...
+	switch(loadmode)
+		if("Load")
+			var/atom/movable/AM = null
+			for(var/atom/movable/A in get_step(ship.loc, turn(ship.dir,180) ))
+				if(!A.anchored)
+					AM = A
+					break
+			if(AM)
+				actions.start(new /datum/action/bar/icon/callback(AM, src.ship, 2 SECONDS, /obj/item/shipcomponent/secondary_system/cargo/proc/load, list(AM, usr), \
+																src.icon, src.icon_state, "[AM] successfully enters [his_or_her(AM)] \the [src.ship]'s!", call_proc_on=src), AM)
```
Tarmunora, line 352: Any reason not to use the `SETUP_GENERIC_ACTIONBAR` helper?

Tarmunora (changes requested):

code/modules/transport/pods/secondary_system.dm
Azrun: I thought so, but I'm pretty sure you are right.

code/modules/transport/pods/secondary_system.dm
Azrun, line 333:
```suggestion
	name = "personal containment unit"
	desc = "Allows the ship to load people as cargo. It is designed to integrate with the Life-Support System so it is probably safe?"
```

Tarmunora (approved):

code/modules/transport/pods/secondary_system.dm
```diff
@@ -325,6 +328,56 @@
...
+			if (length(load) == 1)
+				unload(load[1])
+		else
+			return
+	return
+
+/obj/item/shipcomponent/secondary_system/cargo/jumpseat/relaymove(mob/user as mob)
+	if (is_incapacitated(user))
+		return
+	if( ON_COOLDOWN(src, "relaymove", 1.5 SECOND))
```
ZeWaka, line 364:
```suggestion
	if(ON_COOLDOWN(src, "relaymove", 1.5 SECOND))
```

Azrun: Gross, inconsistent spacing between the two ifs.

PR #16884, opened 2023-11-11
Fixes blobs being able to spawn within view of people (among other things)

### LINKED ISSUE #16881: blob can spawn while in vision by holding the popup
### Labels

[BUG][MAJOR]

### Description

blob can plant while in sightlines by holding the deploy window open and then moving to the place they want to plant

### Steps to reproduce

1. be blob
2. press deploy somewhere you can't be seen
3. don't deploy, instead move somewhere you can be seen
4. press deploy
5. it goes through

### Expected Behavior

it should not

theoretically i could be wrong on this one but im pretty sure iw as being watched and this should probably be checked for safety anyways

Reported by: Kotomata
Active test merges: #16744 and #16370
Round log date: 2023-11-11-06-19
Reported on: 2023-11-11 06:21:52
Map: COGMAP

TealSeer:
[BUG] [PLAYER ACTIONS]
## About the PR 

Moves the checks for blob deployment suitability to a proc and calls it both before and after the "Are you sure" prompt. Also fixes a random typo found in the process.

## Why's this needed? 

Fixes #16881

### DIFF
--- a/code/modules/antagonists/blob/abilities/blob_abilities.dm
+++ b/code/modules/antagonists/blob/abilities/blob_abilities.dm
@@ -210,7 +210,7 @@
 				return
 
 			if(IS_ARRIVALS(T.loc))
-				boutput(owner, "<spawn class='alert'>You can't start inside arrivals!</span>")
+				boutput(owner, SPAN_ALERT("You can't start inside arrivals!"))
 				return
 
 			if (istype(T,/turf/unsimulated/))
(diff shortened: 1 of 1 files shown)


pali6:
I kinda feel like these checks should be done before *and* after the popup. So you don't get a popup pointlessly.

TealSeer:
Yeah I was considering that but wasn't sure, will do then

pali6 (approved):

PR #16899, opened 2023-11-12
Add PTL to Nadir

MarkNais:
## About the PR 
Adds a PTL to Nadir, replacing the SE random-spawn room. It is connected to the turbine and southern catalytic engines.
The high-voltage was stylistically laid with large radius bends. The turbine-wire travels through the false walls of SE maint as a shortcut. Notably false-walls must be deconstructed in order to manipulate wires underneath.

**Side-effect**: The southern catalytic engine grid is connected to the turbine grid; either engine can charge the other's intended SMES units; muddies the clarity of the `power monitor` computer's `Engine Output` graph (it is a sum of all engines). The connection can be easily undone (revert a commit).

I also intentionally engine-electrified this wingrille. 
image

Changes have been tested:
- Connections confirmed to both engine types,
- PTL profit generated, and
- All APCs charging.

The windows have become properly transparent since taking the screenshot below.
I used the stronger windows (like those found near Escape) as it's an "important" station area.
image

## Why's this needed? 
💲🤑💸💰💵💴💶💷💱💹💳
Incentivizes aggressive nuclear-engine & catalytic-engine use. Side effects may include:
- Increased cooperation with Siphon personnel.
- Increased usage of the catalytic engines.
- Increased traitor activity w/ reactor.
- Increased traitor activity w/ hot-wiring.
- An decrease in "Days since last meltdown". image

## Changelog 

```changelog
(u)MarkNstein
(*)Nadir: Added a PTL located at the South East hull connected to the turbine and southern catalytic engines.
```

### DIFF
--- a/maps/nadir.dmm
+++ b/maps/nadir.dmm
@@ -1,20 +1,21 @@
 //MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
-"aaq" = (
-/obj/machinery/door/airlock/pyro/external{
-	dir = 4
-	},
-/obj/cable{
-	icon_state = "4-8"
-	},
-/turf/simulated/floor/industrial,
-/area/station/maintenance/outer/se{
-	name = "Southeast Breach Hull"
-	})
 "aas" = (
 /obj/table/wood/round/auto,
 /obj/item/storage/box/nerd_kit,
 /turf/simulated/floor/carpet/grime,
 /area/station/crew_quarters/arcade/dungeon)
+"aat" = (
+/obj/machinery/power/terminal{
+	dir = 4
+	},
+/obj/cable/orange,
+/obj/cable/orange{
+	icon_state = "1-10"
+	},
+/turf/simulated/floor/yellowblack{
+	dir = 4
+	},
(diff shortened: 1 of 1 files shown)


Kubius (approved):
Seems reasonable on the whole. I'd say probably skip connecting up the catalytic substations, as they're the map's equivalent to solars, but I don't have an explicit objection to laying a cable for engineers to hook them up with.

Cherman0:
I'd agree that the cat stations shouldnt be connected to it. Aside from being inconsistent with how passive generators are normally wired, having them on separate grids is helpful because it means that one hotwired grid doesnt compromise multiple generators. Otherwise they sort of lose their utility as a backup.

TheTyrant25 (approved):
looks fantastic!

It's been stale for a while though, so I have to ask; is there a good reason why it hasn't been merged yet that I don't know about? Or was it just forgotten a little (totally understandable)

P.S. typical colour schema for the cable line leading off the reactor network is to use a blue cable to indicate "this bit goes to the ptl". Totally optional of course, but i'm partial to it personally. Maybe worth considering?

TheTyrant25:
Will this be merged?

Glamurio (approved):

PR #16920, opened 2023-11-14
Saxitoxin recipe change and un-secreting

MylieDaniels:
[Chemistry][Player Actions][Add to Wiki]
## About the PR 
This PR does a few things to saxitoxin:
- Saxitoxin is no longer a hidden recipe nor considered secret.
- Saxitoxin is now saxitoxin red instead of sarin white and foams instead of smoking when created.
- Saxitoxin has a new recipe that references the formation conditions of red tides, the algae blooms that produce saxitoxin.

This new recipe is:
Compost + Ammonia + Oil + Iron Oxide + Mercury + E. Coli + Unstable Mutagen + Sulfuric Acid @ 454 K = 3 Saxitoxin

This recipe is easiest if you raid a few departments, and requires leaving the lab at least a tiny bit (my personal method is to make the E. Coli and Compost from cholesterol extracted from burgers and the sewage derived from coffee concentration).

## Why's this needed? 
Saxitoxin still has the old recipe from when it was sarin, and this recipe is also quite easy once you know it.
The concept of gating a powerful poison behind a recipe that you can whip up at a chem dispenser with a bit of fuel, and just have to secretly know is not a good method of balance, in my eyes.

## Changelog 

```changelog
(u)Mylie Daniels
(*)Saxitoxin has a new recipe that references red tides, and is no longer secret. It's still just as lethal, so make sure to never mix Compost, Ammonia, Oil, Iron Oxide, Mercury, E. Coli, Unstable Mutagen, and Sulfuric Acid at 454 Kelvin.
(+)Saxitoxin is now red and foams instead of smoking on creation.

### DIFF
--- a/_std/__build.dm
+++ b/_std/__build.dm
@@ -30,9 +30,9 @@ o+`        `-` ``..-:yooos-..----------..`
 
 //------------ OPTIONS TO GO FAST ------------//
 
-//#define GOTTA_GO_FAST_BUT_ZLEVELS_TOO_SLOW 1  // Only include the tiny map Devtest, no other zlevels. Boots way faster
+#define GOTTA_GO_FAST_BUT_ZLEVELS_TOO_SLOW 1  // Only include the tiny map Devtest, no other zlevels. Boots way faster
 
-//#define IM_REALLY_IN_A_FUCKING_HURRY_HERE 1  // All of the below
+#define IM_REALLY_IN_A_FUCKING_HURRY_HERE 1  // All of the below
 
 //#define SKIP_FEA_SETUP // Skip setting up atmospheric system
 //#define SKIP_Z5_SETUP // Skip generation of mining level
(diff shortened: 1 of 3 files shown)


pali6:
I kinda liked the algae thing being a part of the recipe.

colossusqw:
> I kinda liked the algae thing being a part of the recipe.

It's a different project than the algae, Mylie and I had a different ideas on how saxi production should work out.

TDHooligan:
i dont know how i feel about raising the bar so high to get what is probably the most boring (though admittedly most effective) poison. i think this is WAY out of scope for most scientists, bar the absolute nerdiest, to achieve.

e.coli and oil are already sluggish reactions; compost is interdepartmental-only, and already a blocker for a much more interesting chemical (BP), iron oxide is a huge pain in the ass to top it off.

at the very least, i would remove iron oxide and compost. saxi's strength is really only in how concentrated it is, and it's so easily treatable that these hurdles are crazy.

PR #16925, opened 2023-11-15
Add a ticking sfx for players who haven't readied up before 30s

MeggalBozale:
[QoL]
## About the PR 
Adds a sound effect to play for those connected (That haven't readied up) when a round is 30 seconds away from starting.
I'm not strongly attached to the random royalty-free SFX I chose, I encourage you to suggest an alternative if you dislike it.
(It's from [here](https://pixabay.com/sound-effects/clock-tick-101150/)

## Why's this needed? 
Sometimes you're tabbed out chatting with other people on the discord or whatever else you may be doing, and you forget to ready up. An audible reminder is a good way to get the message through when visual reminders won't.

## Changelog

```changelog
(u)MeggalBozale
(*)Players not readied up when the round is soon to start will hear a mysterious ticking noise. Kind of catchy.
```


code/datums/gameticker.dm
```diff
@@ -69,7 +71,14 @@ var/global/current_state = GAME_STATE_INVALID
...
 				// so this makes it show up a minute before the game starts
 				handle_mapvote()
-				did_mapvote = 1
+				did_mapvote = TRUE
+
+			if (pregame_timeleft <= 30 && !did_reminder)
+				// hey boo the rounds starting and you didnt ready up
+				for_by_tcl(P, /mob/new_player)
+					if (!P.ready)
+						playsound(P.loc, 'sound/misc/clock_tick.ogg', 50, flags = SOUND_IGNORE_SPACE)
```
pali6, line 80: This will be incredibly loud as the sound will be played N times where N is the number of players not readied up. Please don't do this and instead use `playsound_global` to which you pass a list of mobs or clients as I described in #imcoder.

pali6 (changes requested):

code/datums/gameticker.dm
MeggalBozale: oh god what its gonna play for everyone on there i did not know that
okie on it

pali6 (approved):
Looks good. Maybe there could be a preferences toggle but that can be added later. Volume 50 might be too loud, idk.

MeggalBozale:
With all volume settings set to 100 and my 'normal' volume it doesn't break my ears. I think it could stand to be 25, I would rather it be too quiet than too loud.

Azrun (approved):

Gannets (approved):
Tick tick tick tick

AXLFAN2000:
I really really like this!

pali6:
I'm gonna merge this and when people inevitably complain I'll add a preferences toggle or something.

PR #16935, opened 2023-11-15
Dropped atoms in ladders fall to lower turf

### LINKED ISSUE #16401: Items disappear when you drop them whilst climbing down ladder
### Labels

[BUG][MINOR]

### Description

When you drop an item at the same time you are crawling down a ladder it disappears

### Steps to reproduce

1. find a ladder
2. begin to climb down
3. drop your item before the animation is over

### Expected Behavior

To drop my items on the bottom of the ladder

Reported by: Zwucky
Active test merges: #14886
Round log date: 2023-10-11-10-44
Reported on: 2023-10-11 11:11:34
Map: ATLAS

pgmzeta:
## About the PR 

When users climb (or are assisted up the ladder; both use the same code path), add the user to a list of 'climbers'. When any movable enters a ladder, check it against the list of climbers and visual contents. If it's neither of those things, drop it to the bottom level of the ladder.

When any movable enters a ladder, if it's not a 
## Why's this needed? 
Fixes #16401


code/obj/ladder.dm
```diff
@@ -202,3 +202,12 @@ ADMIN_INTERACT_PROCS(/obj/ladder/embed, proc/toggle_hidden)
 
 		src.vis_contents -= proxy
 		qdel(proxy)
+
+/obj/ladder/Entered(atom/movable/AM, atom/OldLoc)
+	. = ..()
+	if(isobj(AM))
```
ZeWaka, line 208: why not any movable

ZeWaka: ex. what if i vomit up a mob i ate while climbing

ZeWaka: (yes, this does pose difficulties with the set loc in climb() if you change it to movables, care)

pgmzeta: aaaaaaaaaaaaaaaaaaaaaaaaaaaa
`var/atom/movable/proxy = new(src)`

code/obj/ladder.dm
```diff
@@ -13,6 +13,8 @@ ADMIN_INTERACT_PROCS(/obj/ladder/embed, proc/toggle_hidden)
 	var/id = null
 	/// if true, disables ladder climbing behavior
 	var/unclimbable = FALSE
+	/// list of active ladder climbers, used for loose item/mob expulsion
+	var/list/climbers = list()
```
ZeWaka, line 17: List of what? 🤓

ZeWaka (changes requested):

Tarmunora (approved):

PR #16941, opened 2023-11-16
Adds a preference for scrollwheel limb targeting

TealSeer:
[QOL] [PLAYER ACTIONS] 
## About the PR 

Adds a new setting in the game options tab of character setup that allows you to change between always scrolling to target limbs, only targeting when hovering over the target doll, or never targeting. Also fixes a small bug where `if(!M?.zone_sel)` being checked before calling `A?.MouseWheel` led to broken functionality for mobs that don't have a zone_sel e.g. admin ghosts using build mode.

## Why's this needed? 

Scrolling the mouse wheel is also used to refocus the game window when you say, open a chui window, tying a new feature to the mouse wheel without an option to change it is annoying for people who make frequent use of the scroll to refocus feature.

## Changelog 

```changelog
(u) TealSeer
(*) The scroll to target limbs feature can now be configured in the game options tab of character preferences (between always, only on hover, or never).
```

PR #16945, opened 2023-11-16
Scroll targeted limb only when hovering with mouse over the zone selector

pali6:
## About the PR 
I'm dumb and when I merged #16846 I thought this is what I was merging. This PR makes it so you only scroll the targeted zone if your mouse is hovering over the zone selector.

## Why's this needed? 
I think targeting zones like this is mainly really useful for katana. Imo we should apply a QoL katana improvement instead. One idea proposed in dev chat was to make it automatically change your limb selection to the next non-missing limb if the targeted limb is missing. If this is something people like as a QoL for katana I'll implement it, I just didn't want to put two different things in one PR.

I would say that using up an interaction like mouse wheel scroll for this is not a great idea. For example there has been an interesting discussion about using the scroll wheel to scroll through piles of items. I feel like that would be a better use.

Also when *any* scroll triggers this then it is easy to accidentally do this when you didn't mean to or when refocusing the window as documented by #16941.

## Changelog 

```changelog
(u)pali
(*)To change targeted zone by scrolling mouse wheel you now need to hover over the target body part UI thing.
```


DisturbHerb:
I would rather be able to quickly swap limbs while my cursor is on the fellow that I'm trying to beat up for whatever reason, rather than being made to flick my mouse to the target doll every single time. Alternatively, if there were buttons that could cycle between the target zones that were closer to the left side of the keyboard, that would be lovely.

pali6:
Maybe some default keybind on the left side of the keyboard to go "next zone" and "previous zone" could be doable. Unsure what, likely something with key modifiers.

Tarmunora:
Would it be reasonable to smash this together with #16941, (i.e.: have 3 options - scroll anywhere, scroll on doll, no scrollwheel anywhere)? @TealSeer

TealSeer:
I think that would be a good way to handle it, yeah!

Tarmunora:
Merged into #16941

PR #16955, opened 2023-11-16
Using a wire cutter on an assembly prevents its inhand interaction

Valtsu0:
[FEATURE][GAME OBJECTS]
## About the PR 
Using a wirecutters on timer, proximity or radio/igniter and radio/horn assemblies prevents its inhand interaction. Changes its examine text to reflect that.  Using a wirecutters again allows you to interact with it again. Trying to use it tells you so.

## Why's this needed? 
Currently, there is little to no point in surgically implanting remote activated bombs into people, since they can just fart and turn them off. Preventing people from turning off the detonator would allow more antag gimmicks

## Changelog 

```changelog
(u)Valtsu0
(+)Using wirecutters on relevant assemblies prevents their inhand interaction.
```


Valtsu0:
Will return to this later

Valtsu0:
Later is now

Valtsu0:
Idk if i should make a new pr. Well this being near the end of the list is probably fine

ZeWaka (commented):
as much as i hate this duplicate code, it doesn't make sense to bring it up to obj/item/assembly i don't think

PR #16977, opened 2023-11-17
Fix a bug with the bug report form

ZeWaka:
## About the PR 

ircbot was showing as loaded despite not being configured whatsoever
due to 8yr old wire code
well, moreso zam making config.sample.txt being a thing

## Why's this needed? 

fix bug with bug report form


pali6 (commented):
What even is the point of the load tries. It's not like it actually pings the bot and verifies that it's running.

ZeWaka:
idk I didn't want to change purported functionality

PR #16992, opened 2023-11-17
Arc-Electroplating-Support-For-Borgs

Francisco-Franks:
[Silicon][Materials]

## About the PR 
This enables Borgs, and well, anyone tangible, to put the correct things into the arc electroplater. Click drag things into the plater.

## Why's this needed? 
Currently the electroplater seems rather seldom used. This will let borgs join in on the fun, maybe we will see more soulsteel orchestras or something.

## Changelog 

```changelog
(u)Metareferencer
(+)Click-dragging now works for the electroplater. Get electroplating, borgos!
```

### DIFF
--- a/code/obj/machinery/arc_electroplater.dm
+++ b/code/obj/machinery/arc_electroplater.dm
@@ -114,7 +114,7 @@ TYPEINFO(/obj/machinery/arc_electroplater)
 			boutput(user, SPAN_ALERT("You can't plate something this tiny!"))
 			return
 
-		if (W.w_class > src.max_wclass || istype(W, /obj/item/storage/secure) || W.anchored >= ANCHORED)
+		if (W.w_class > src.max_wclass || istype(W, /obj/item/storage/secure) || W.anchored)
 			boutput(user, SPAN_ALERT("There is no way that could fit!"))
 			return
 
--- a/code/obj/machinery/arc_electroplater.dm
+++ b/code/obj/machinery/arc_electroplater.dm
@@ -70,6 +70,9 @@ TYPEINFO(/obj/machinery/arc_electroplater)
 		if (W.cant_drop) //For borg held items
 			boutput(user, SPAN_ALERT("You can't put that in [src] when it's attached to you!"))
 			return
+		if (iswraith(user) || isintangible(user) || is_incapacitated(user)|| isghostdrone(user) || isAI(user))
+			boutput(user, SPAN_ALERT("Electroplating is for the Tangible."))
+			return
 		else
 			src.Attackby(W, user)
 
--- a/code/obj/machinery/arc_electroplater.dm
+++ b/code/obj/machinery/arc_electroplater.dm
@@ -45,6 +45,34 @@ TYPEINFO(/obj/machinery/arc_electroplater)
 		SubscribeToProcess()
 		return 1
 
+	// It is time for borgs to get in on this hot electroplating action.
+	MouseDrop_T(var/obj/item/W, var/mob/user)
+		// You know, if all the rest of this code already does the heavy lifting of checking validity and all that...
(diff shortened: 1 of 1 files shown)


code/obj/machinery/arc_electroplater.dm
```diff
@@ -45,6 +45,37 @@ TYPEINFO(/obj/machinery/arc_electroplater)
...
+			boutput(user, SPAN_ALERT("There is no way that could fit!"))
+			return
+		if (W.cant_drop) //For borg held items
+			boutput(user, SPAN_ALERT("You can't put that in [src] when it's attached to you!"))
+			return
+		if (iswraith(user) || isintangible(user) || is_incapacitated(user)|| isghostdrone(user) || isAI(user))
+			boutput(user, SPAN_ALERT("Electroplating is for the Tangible."))
+			return
+		else
+			src.Attackby(W, user)
```
pgmzeta, line 77: calling attackby results in some duplicate checks; I would pull the common stuff into item level procs and call those from both to ensure the logic is mirrored between clickdrag and attackby now and in the future.

Francisco-Franks: Ok, attackby and mousedrop_t now just route directly to a proc which does it all.

code/obj/machinery/arc_electroplater.dm
```diff
@@ -45,8 +45,31 @@ TYPEINFO(/obj/machinery/arc_electroplater)
...
+	// See if I can piece together how to make this fly.
+	proc/PlaterInteract(var/obj/item/W, var/mob/user)
+		// Please don't drag the nano-crucible into the plater. Or any other machine or mob for that matter.
+		if (!istype(W, /obj/item))
+			boutput(user, SPAN_ALERT("Please do not attempt to plate machinery or entities."))
+			return
+		// Theres a suicide button for this.
+		if (W == user)
+			boutput(user, SPAN_ALERT("Please do not drag yourself into the Plater."))
+			return
```
Tarmunora, line 64: These could honestly just fail silently

code/obj/machinery/arc_electroplater.dm
```diff
@@ -45,8 +45,31 @@ TYPEINFO(/obj/machinery/arc_electroplater)
...
+		// Do not attempt to plate objects from a distance.
+		if (BOUNDS_DIST(user, src) > 0)
+			return
+		// Do not attempt to plate objects at a distance.
+		if (BOUNDS_DIST(W, src) > 0)
+			return
```
Tarmunora, line 70: Might also want to check distance between `user` and `W`, so you can't pull something into the plater from the far side of it. Minor thing but good practice

code/obj/machinery/arc_electroplater.dm
```diff
@@ -91,7 +114,7 @@ TYPEINFO(/obj/machinery/arc_electroplater)
 			boutput(user, SPAN_ALERT("You can't plate something this tiny!"))
 			return
 
-		if (W.w_class > src.max_wclass || istype(W, /obj/item/storage/secure))
+		if (W.w_class > src.max_wclass || istype(W, /obj/item/storage/secure) || W.anchored >= ANCHORED)
```
Tarmunora, line 117:
```suggestion
		if (W.w_class > src.max_wclass || istype(W, /obj/item/storage/secure) || W.anchored)
```
Unless something has a negative (or fractional, I suppose) value for `anchored`, the check can be simplified (and usually is)

Tarmunora (changes requested):

Francisco-Franks (commented):
Agreedo

Tarmunora (approved):

pgmzeta (approved):

PR #17010, opened 2023-11-18
Allow synthflesh removal from pustule with syringe

Saicchi:
[ENHANCEMENT][CHEMISTRY]

## About the PR 

Allows removal of synthflesh from pustules using syringes using a 1 second interruptible action. Process is only one way.

A few seconds after poking, the pustule will attempt to heal it using 15 units of synthflesh and reducing efficiency by 20% to prevent it from being abused. If it doesn't have enough to fix the hole it gets destroyed instead.

## Why's this needed? 

Cyborgs can't make synthflesh. It requires picking it up and plopping them on a beaker or barrel, which they can't do.

## Changelog

```changelog
(u)DrWolfy
(+)Synthflesh in pustules can now be recovered with a syringe, reducing the pustule efficiency.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Tools.dm
+++ b/code/modules/chemistry/Chemistry-Tools.dm
@@ -911,6 +911,9 @@ proc/ui_describe_reagents(atom/A)
 	var/angry = FALSE
 	var/angry_timer = 15
 	var/list/convertible_reagents = list("blood","bloodc","bloody_mary","bloody_scary","hemolymph")
+	var/poked = FALSE // poked by a syringe, will try to heal using synthflesh
+	var/pokes = 0 // each poke reduces efficiency
+	var/poke_timer = 2
 
 	New()
 		START_TRACKING
@@ -922,6 +925,15 @@ proc/ui_describe_reagents(atom/A)
 	disposing()
 		STOP_TRACKING
 		..()
+
+	examine(mob/user)
+		. = ..()
+		if (pokes > 5)
+			. += "Its full of patched up holes!"
+		else if (pokes > 3)
+			. += "It looks like it's been poked quite a lot."
+		else if (pokes > 1)
+			. += "It looks like it's been poked a few times."
 
 	proc/convert_reagent(reagent_id, volume)
 		var/reagent_to_add
@@ -980,6 +992,25 @@ proc/ui_describe_reagents(atom/A)
 					convert_reagent(reagent_id, reagent_present)
 				else
 					convert_reagent(reagent_id, amount_of_reagent_to_use)
+
+		if( poked )
+			if ( poke_timer < 0 )
+				if ( src.reagents.get_reagent_amount("synthflesh") < 16)
+					playsound(src.loc, 'sound/effects/espoon_suicide.ogg', 50, 1)
+					src.audible_message(SPAN_ALERT("<B>The [src.name] cannot repair itself and collapses!</B>"))
+					qdel(src)
+				else
+					src.reagents.remove_reagent("synthflesh", 15)
+					// each poke reduces efficiency by 20%
+					src.chemical_efficiency *= 0.80
+					src.organ_efficiency *= 0.80
+					src.pokes += 1
+					src.poked = FALSE
+					src.poke_timer = initial(src.poke_timer)
+					playsound(src.loc, 'sound/misc/cling_flesh.ogg', 50, 1)
+					src.audible_message(SPAN_ALERT("<B>The [src.name] repairs itself from the poked hole.</B>"))
+			else
+				poke_timer -= 1
 		..()
 
 	attackby(var/obj/item/W, mob/user)
--- a/code/modules/chemistry/tools/syringes.dm
+++ b/code/modules/chemistry/tools/syringes.dm
@@ -116,12 +116,16 @@
 					boutput(user, SPAN_ALERT("The [src.name] is full."))
 					return
 
+				if (istype(target,/obj/item/reagent_containers/synthflesh_pustule))
+					actions.start(new/datum/action/bar/icon/syringe(target, src, src.icon, src.icon_state), user)
+					return
+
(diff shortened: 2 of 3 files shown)


code/modules/chemistry/tools/syringes.dm
```diff
@@ -116,6 +116,16 @@
 					boutput(user, SPAN_ALERT("The [src.name] is full."))
 					return
 
+				if (istype(target,/obj/item/reagent_containers/synthflesh_pustule))
+					var/obj/item/reagent_containers/synthflesh_pustule/P = target
+					P.reagents.trans_to(src, src.amount_per_transfer_from_this)
+					logTheThing(LOG_CHEMISTRY, user, "draws 5 units of reagents from [constructTarget(target,"combat")] [log_reagents(target)] with a syringe [log_reagents(src)] at [log_loc(user)].")
```
Azrun, line 122: 5 ->                             src.amount_per_transfer_from_this?

code/modules/chemistry/tools/syringes.dm
```diff
@@ -116,6 +116,16 @@
...
+					if (!P.poked)
+						P.poked = TRUE
+						SPAWN(3 SECONDS)
+							P.try_healpoke()
```
Azrun, line 126: Is the intent you have 3 seconds to use the hole?

code/modules/chemistry/tools/syringes.dm
```diff
@@ -116,6 +116,16 @@
 					boutput(user, SPAN_ALERT("The [src.name] is full."))
 					return
 
+				if (istype(target,/obj/item/reagent_containers/synthflesh_pustule))
+					var/obj/item/reagent_containers/synthflesh_pustule/P = target
+					P.reagents.trans_to(src, src.amount_per_transfer_from_this)
+					logTheThing(LOG_CHEMISTRY, user, "draws 5 units of reagents from [constructTarget(target,"combat")] [log_reagents(target)] with a syringe [log_reagents(src)] at [log_loc(user)].")
+					if (!P.poked)
+						P.poked = TRUE
+						SPAWN(3 SECONDS)
```
Azrun, line 125: Why not use the `process` loop?

code/modules/chemistry/tools/syringes.dm
Saicchi: Yes. Giving the player 3 seconds until it closes and they need to poke a new one.

code/modules/chemistry/tools/syringes.dm
Saicchi: Copied from a few lines below, will fix that one too.

code/modules/chemistry/tools/syringes.dm
Saicchi: Removed the SPAWN and moved it to a process loop.

Saicchi:
Updated the PR.
Drawing from the pustule using a syringe is now a 1 second interruptible action like a player instead of being instant.
Moved the loop to check if it was poked to a `process` loop.

Flaborized:
As the person who made the pustules, I'm not against borgs being able to utilize them and make synthflesh, but I think syringes are a very tedious way to do that and I wanted the 'despositing synthflesh from the pustule' part of the process to be quick and painless. I'd much rather it be doable in some comparably instant way to how humans can do it

Saicchi:
> As the person who made the pustules, I'm not against borgs being able to utilize them and make synthflesh, but I think syringes are a very tedious way to do that and I wanted the 'despositing synthflesh from the pustule' part of the process to be quick and painless. I'd much rather it be doable in some comparably instant way to how humans can do it

I did think about letting borgs just click and drag the pustule to the barrel, and I much prefer that option, but midway through it I gave up since it seemed a bit imbalanced as they don't need to pick up the pustule and risk their arm getting ripped without calming it down first.

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -911,6 +911,9 @@ proc/ui_describe_reagents(atom/A)
 	var/angry = FALSE
 	var/angry_timer = 15
 	var/list/convertible_reagents = list("blood","bloodc","bloody_mary","bloody_scary","hemolymph")
+	var/poked = FALSE // poked by a syringe, will try to heal using synthflesh
+	var/pokes = 0 // each poke reduces efficiency
+	var/poke_timer = 2
```
frawhst, line 916: Would appreciate some comments on how `poke_timer` works, I think I get it, but it's done in imprecise Per Process Tick units so it's hard to understand from the variable definition alone

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -911,6 +911,9 @@ proc/ui_describe_reagents(atom/A)
 	var/angry = FALSE
 	var/angry_timer = 15
 	var/list/convertible_reagents = list("blood","bloodc","bloody_mary","bloody_scary","hemolymph")
+	var/poked = FALSE // poked by a syringe, will try to heal using synthflesh
+	var/pokes = 0 // each poke reduces efficiency
```
frawhst, line 915:
```suggestion
	var/poked = FALSE //! poked by a syringe, will try to heal using synthflesh
	var/pokes = 0 //! each poke reduces efficiency
```
code doc comments, can be read on hover and are generally good

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -923,6 +926,15 @@ proc/ui_describe_reagents(atom/A)
 		STOP_TRACKING
 		..()
 
+	examine(mob/user)
+		. = ..()
+		if (pokes > 5)
+			. += "Its full of patched up holes!"
```
frawhst, line 932:
```suggestion
			. += "It's full of patched up holes!"
```
grammar

frawhst (commented):
few minor things

PR #17011, opened 2023-11-19
Add MouseDrag_T to artlab pitching machine and impact pad

Saicchi:
[QOL][SCIENCE]

## About the PR 

Adds 'MouseDrag_T' to the ArtLab pitching machine and impact pad.

## Why's this needed? 

Silicons can't refill the pitching machine or put an artifact on the impact pad when its raised.

## Changelog 

```changelog
(u)DrWolfy
(+)Add drag and drop to artlab pitching machine and impact pad
```

### DIFF
--- a/code/modules/networks/computer3/mainframe2/misc_terms.dm
+++ b/code/modules/networks/computer3/mainframe2/misc_terms.dm
@@ -3838,6 +3838,12 @@ TYPEINFO(/obj/machinery/networked/test_apparatus)
 			boutput(user, "That is far too big to fit!")
 			return
 
+	MouseDrop_T(atom/movable/O as mob|obj, mob/user as mob)
+		if (!istype(O,/obj/) || O.anchored) return
+		if (BOUNDS_DIST(src, O) > 0 || !isturf(O.loc)) return
+		if (!in_interact_range(user, O) || !in_interact_range(user, src) || !isalive(user)) return
+		src.attackby(O, user)
+
 /obj/machinery/networked/test_apparatus/impact_pad
 	name = "Impact Sensor Pad"
 	desc = "A floor pad that detects the physical reactions of objects placed on it."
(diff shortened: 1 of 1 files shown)


Tarmunora (changes requested):
From a quick read of the code, it looks like if you're a human and you clickdrag something onto these, it'll also make you drop whatever you're holding in your hand. Not ideal behaviour if that is the case

Saicchi:
Added a check to see if the item on the active hand is the same as the one being drag dropped.

Tarmunora (approved):

code/modules/networks/computer3/mainframe2/misc_terms.dm
```diff
@@ -3838,6 +3838,12 @@ TYPEINFO(/obj/machinery/networked/test_apparatus)
 			boutput(user, "That is far too big to fit!")
 			return
 
+	MouseDrop_T(atom/movable/O as mob|obj, mob/user as mob)
+		if (!istype(O,/obj/) || O.anchored) return
+		if (BOUNDS_DIST(src, O) > 0 || !isturf(O.loc)) return
+		if (!in_interact_range(user, O) || !in_interact_range(user, src) || !isalive(user)) return
+		src.attackby(O, user)
```
frawhst, line 3845:
```suggestion
		src.Attackby(O, user)
```
Always override `attackby` and call `Attackby`, the latter has some signal-sending stuff that we pretty much always want to happen

code/modules/networks/computer3/mainframe2/misc_terms.dm
```diff
@@ -3946,6 +3952,12 @@ TYPEINFO(/obj/machinery/networked/test_apparatus)
 
 		return
 
+	MouseDrop_T(atom/movable/O as mob|obj, mob/user as mob)
+		if (!istype(O,/obj/) || O.anchored) return
+		if (BOUNDS_DIST(src, O) > 0 || !isturf(O.loc)) return
+		if (!in_interact_range(user, O) || !in_interact_range(user, src) || !isalive(user)) return
+		src.attackby(O, user)
```
frawhst, line 3959:
```suggestion
		src.Attackby(O, user)
```
ditto

PR #17018, opened 2023-11-19
Don't allow unpowered bolted airlocks to be unbolted

DasBrain:
[Game-Objects] [Bug] [Balance]

## About the PR 

Currently you can raise the bolts of an unpowered door if the door is unpowered because the area does not have power.
This PR fixes that by also checking if the door is actually powered.

Because this change makes it impossible to raise the bolts until the area is powered too, this may have an impact on how the AI may bolt their upload.

## Why's this needed? 

Bugs bad, mkay?

## Changelog 

```changelog
(u)DasBrain
(+)Airlocks now also require the area to have power to raise the bolts.
```

### DIFF
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -981,7 +981,7 @@ About the new airlock wires panel:
 				src.set_locked()
 				tgui_process.update_uis(src)
 			else
-				if(src.arePowerSystemsOn()) //only can raise bolts if power's on
+				if(src.arePowerSystemsOn() && !(src.status & NOPOWER)) //only can raise bolts if power's on
 					boutput(usr, "You hear a clunk from inside the door.")
 					src.set_unlocked()
 			SPAWN(1 DECI SECOND)


444Portal:
AI already couldn't mess with bolts on unpowered doors (from what I remember). Having door wires do the same is good imo

Retrino:
wouldnt this make it impossible to progress in biodome

Xkeeper0:
You can still bash the firelocks down, and the map can (and should) be altered to move the doors into the biodome entry area (right now they're in the inner area)

CaioFrancisco:
This isn't **really** a bug, is it? Or if it was, it has been accepted by the community as a feature. I'm not against the PR at all, just the wording of it being a bug.

BatElite:
This would also affect the space owlery btw

How hard is it to bash airlocks down? I still never see folks doing it even though their health got a major nerf a good while back. My concern is whether or not it'd be hard to just completely trap folks with this change.

Valtsu0:
I do think being able to unbolt doors that don't have power is a good thing.

Scope creep: maybe you could hotwire a door with a cell and some wires

Azrun:
*I* think having some kind of fail-safe seems appropriate and thematic. I'd hope in the event of some kind of failure that people aren't just trapped someplace.

I could see an argument for *some* doors not having a fail-safe to minimize the ability to bypass access restriction in the event of a failure.

*Just musing on is hope a complex/station be built.*

imawxome:
in my headcanon you just power the bolts with the multitool, just enough to make them slide out of the way

PR #17034, opened 2023-11-21
Fix photos showing up in inventory

### LINKED ISSUE #9120: held photos dont show anything
### Labels

[BUG]

### Description

title

### Steps to reproduce

pick up a photo
look at it
see problem

### Expected Behavior

probably would be nice for photos to show stuff when held

Reported by: Skeletonman0
Round log date: 2022-06-13-03-28
Reported on: 2022-06-13 03:46:33

gear-space:
[Bug][Game-Objects]

## About the PR 
First attempt at solving the issue of polaroid photos showing contents when on the ground, but not when in hand/belt/bag/photobook. Do note, this doesn't get other overlays (stickers, blood, signatures, writing) to show up on the photos in hand, but I couldn't find a straightforward way to update that without running a ton of .UpdateOverlays which I understand are kind of resource intensive...  They still show up correctly when the photo is on the floor. 

image

Best I can tell, the `src.UpdateOverlays(IM, "photo")` overlay doesn't get updated when the photo changes layers between floor and hands. Is there a better way to force this overlay to update?

I know upfront this probably isn't the most efficient code or clean enough to push to prod, so I'd appreciate some guidance in getting it cleaned up as I've hit my level of understanding😸😅 

## Why's this needed? 
Fixes #9120 
First attempt at least... Tourists should be able to show off their photobook with ease! Stare at your picture of Jones in your hand forever!

### DIFF
--- a/code/obj/item/cameras.dm
+++ b/code/obj/item/cameras.dm
@@ -215,32 +215,31 @@ TYPEINFO(/obj/item/camera_film/large)
 		..(location)
 		if (istype(IM))
 			fullImage = IM
-			IM.transform = matrix(24/32, 22/32, MATRIX_SCALE)
-			IM.pixel_y = 1
-			src.UpdateOverlays(IM, "photo")
+			render_photo_image(src.layer)
 		if (istype(IC))
 			fullIcon = IC
 		if (nname)
 			src.name = nname
 		if (ndesc)
 			src.desc = ndesc
+
+	/// Resize and update photo overlay (layer)
+	proc/render_photo_image(var/layer)
+		var/image/IM = src.fullImage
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
You could make it into a proc to avoid duplicating the code 3 times, but I don't think you're going to be able to get away without the `UpdateOverlays` call. It's only on pickup and drop though, so it's not like this is anywhere near hot code.

gear-space:
Endlessly amusing
juggle-photo

TobleroneSwordfish (approved):

DisturbHerb:
bogos binted? 👽

PR #17043, opened 2023-11-22
Tubby Cow: better clothing coverage for bovines, via filters

kbsmilk:
[MUTANTRACE] [CLOTHING]
## About the PR 
* Adds filters for 6 different clothing slots, that generally improve clothes' fitting and appearance on the unique bovine shape.
* Affects: jumpsuits, oversuits, belts, satchels, backpacks, and gloves. Only for bovines. **Only in the east/west directions.**
* Adds a simple base mutantrace proc that is called by update_icon()'s sub-procs. It's helpful for applying filters to clothes based on the mob's mutantrace. The proc does nothing by default.
* There's a custom mutant emote (*udder), for users to toggle the filters on/off.
* Renames a few instances of /obj/effect/distort to /obj/effect/rt; removes its default icon.

## Why's this needed? 
Cows have clipping issues with nearly every clothing in the game. See this comparison image.
I might suggest testmerging before a full merge, given the wide variety of clothes in game. It's hard testing all these!

Example 3

## Changelog 

```changelog
(u)Smilg
(*)Bovine clothing now fits better, using distortion maps. Please report any clothes combos that look really bad.
```


code/datums/mutantraces.dm
```diff
@@ -2131,14 +2164,61 @@ ABSTRACT_TYPE(/datum/mutantrace)
 			src.mob.kickMessage = "stomps"
 			src.mob.traitHolder?.addTrait("hemophilia")
 
+			src.mob.vis_contents += src.distort_under
+			src.mob.vis_contents += src.distort_suit
+			src.mob.vis_contents += src.distort_belt
+			src.mob.vis_contents += src.distort_satchel
+			src.mob.vis_contents += src.mask_gloves
+			src.mob.vis_contents += src.mask_backpack
```
pali6, line 2172: This and the subtraction can also just be edone eas `+=`ing a list instead of doing it one by one

code/obj/effect.dm
```diff
@@ -12,5 +12,12 @@
 		..()
 		src.render_target = "*\ref[src]"
 
+/obj/effect/rt
```
pali6, line 15: What's the benefit of this over `/obj/effect/distort`? It seems to serve the same purpose but is named more confusingly and is lacking crucial appearance_flags that should be there.

code/mob/living/carbon/human/procs/update_icon.dm
```diff
@@ -109,7 +109,9 @@
 		var/image/suit_image
 		wear_sanity_check(src.w_uniform)
 		suit_image = src.w_uniform.wear_image
-		suit_image.filters = src.w_uniform.filters.Copy()
+		suit_image.filters = src.w_uniform.filters.Copy() + src.mutantrace?.apply_clothing_filters(src.w_uniform)
+		//suit_image.filters += src.mutantrace?.apply_clothing_filters()
```
pali6, line 113:
```suggestion
```

pali6:
With this many filters and things in vis_contents I do worry how this will affect client performance.

imawxome:
These look really good!

code/obj/effect.dm
kbsmilk: Didn't know why the flags weren't in /obj/effect, and the masks worked without the flags, so they never got put in.
Renaming /distort sounds fine though, it looks like its default icon isn't actually used anywhere.

pali6:
What we could maybe do at some point later (if we care about client performance) is to make a proc that applies a distortion map onto an icon producing a modified icon. This would be a somewhat expensive operation but could be cached and it would remove the potential clientside impact of the many vis_contents things and filters.

kbsmilk:
oooh that does sound fun. For now though, cows seem to be a really rare mutantrace, so while the filter count is high even for me, it shouldn't be a big impact. All this stuff adds up over time though, so if we get a proc like that, I'd make use of it.

AXLFAN2000:
absolutely LOVING how this is looking ingame!

PR #17046, opened 2023-11-22
puffers to the people (monkeys)

Retrino:
## About the PR 

adds monkey puffer sprites where there previously werent

## Why's this needed? 

clothe those beasts


TemThrush:
image
I'll miss him

Roninnozlo:
thats a soda can with legs

PR #17062, opened 2023-11-23
Unify kitchen dispensers across all maps

TheTyrant25:
[MAPPING] [CATERING]
## About the PR
- Adds a new subtype of Kitchen Dispensers to maps, known as `chem_dispenser/chef/happychef`, which uses the reduced ingredients list found on nadir and donut2.
- Adds these "happy chef dispense-o-matic" dispensers to maps that did not have them. Every map that previously used a regular kitchen dispenser now uses the `happychef` variant instead.
- Surrounded every kitchen dispenser with caution stripes like on cogmap1, for easier finding.

## Why's this needed?
Map unity is important. Several chefs rely on these dispensers for their personalised menus. This is not intended to reduce catering-botany interactions, but instead to complement them.

## Changelog
```changelog
(u)Tyrant
(*)Every map now has at least one Kitchen Dispenser. Big day for chefs.
```


DisturbHerb (changes requested):
I would make a type for this rather than copying the var-edited machine across all the maps. Moreover, I remember there being some opposition to adding more of these, but I'm sure any concerns would make themselves known as this gets reviewed.

TheTyrant25:
I don't think it makes sense to have it on some maps and not others. But yeah I'll look into making a subtype.

TDHooligan:
reliably being able to flavor cakes with real ingredients is a Good Thing but yea use a subtype

i will also note that, even since the last kitchen dispenser PR, botany has not once provided enough fruit juice/food to ice a cake. (even the good, saintly botanists who give me everything i want)

TheTyrant25:
scope has been slightly changed. Description updated to match.

DisturbHerb (commented):
Code looks fine.

flrsh:
I think it's fine if this exists on a couple maps but doesn't exist on the others, since I don't want it on the others. I think this makes it trivial to get many "special" food reagents and too easily bypasses botany / QM.

Creathian:
> I think it's fine if this exists on a couple maps but doesn't exist on the others, since I don't want it on the others. I think this makes it trivial to get many "special" food reagents and too easily bypasses botany / QM.

I heavily disagree with this. Looking at what these dispensers include, the only reagents that would otherwise have to be gotten from Botany are strawberries for strawberry milk and mustard from mustard. If this gave the actual mustard seed and strawberries I might agree, but I see this change as basically just giving the chef a way to refill the mustard bottle without asking botany to help. Otherwise, it'd just give chefs more options for things like soups and cake frosting. For any actual dishes, the chef still needs QM and Botany.

PR #17066, opened 2023-11-23
Winter Clown Outfit

Flaborized:
## About the PR 

Adds a winter clown outfit that spawns in the clown crate during Spacemas. Meant to replace the Autumn set when Autumn is done. Also includes a monkey jumpsuit sprite. Comes with jumpsuit, shoes, mask and hat.

winterclown

## Why's this needed? 

The Autumn clown suit is well liked and I wanted to make sure the clown always has some seasonal garb since we obviously cannot stock the autumn kit forever (because wearing the wrong suit for the wrong season would be illegal). Note this only is active during December and the new defines for winter have it go from December to like, February I think, so I might need to add some kinda separate winter but not necessarily Spacemas kinda outfit by then, since this thing is obviously *festive* and not like, embodying winter itself.

Also I'd love to add a seasonally appropriate sleigh-bell footstep sound for the shoes (instead of classic clown shoe squeaking) but I don't have what I'd consider Great sounds for it. No changelog since if it's merged before Spacemas it won't even be visible until then!


Gannets (approved):
Heck yeah this is great!

Flaborized:
Yup it was the wrong way around, good eye, fixed

Azrun (approved):

PR #17081, opened 2023-11-25
prevent catwalks connecting to grilles

### LINKED ISSUE #10733: grilles auto-connect to catwalks
### Labels

[minor]

### Description

As title. Presumably because catwalks are weird grilles internally.

### Steps to reproduce

1. Go on donut3
2. See all the under-window grilles connecting to catwalks nearby

### Expected Behavior

No autoconnect

image
Easiest to see with the south-facing connection

TheTyrant25:
[GAME OBJECT]
## About the PR 
Prevents grilles and catwalks auto connecting to each other. This is done by primarily moving the code for catwalks and grilles apart under a new abstract parent: `/obj/mesh`. `grille` and `catwalk` are now both subtypes of `mesh`.

This PR also replaces every use instance of `/obj/grille/catwalk` and `/obj/grille` to be correctly pathed.

## Why's this needed?
Fixes #10733


TheTyrant25:
!merge upstream

Tarmunora:
Hm. Kinda feel like this would be best handled by refactoring out an abstract parent, so that catwalks aren't a child of grilles and `connect_to` can be used as expected without having to is istype for carveouts

TheTyrant25:
fair enough. I'll do that instead then

TheTyrant25:
massive find replace moment
also, i do not know how to use updatepaths, so any advice in setting that up for this pr would be appreciated

code/obj/grille.dm
```diff
@@ -504,7 +379,7 @@
 				icon_state = "grille[builtdir]" + "-0"
 
 	proc/update_neighbors()
-		for (var/obj/grille/G in orange(1,src))
+		for (var/obj/mesh/grille/G in orange(1,src))
 			G.UpdateIcon()
```
Tarmunora, line 383:
```suggestion
		for (var/obj/mesh/O in orange(1,src))
			O.UpdateIcon()
```

code/obj/grille.dm
```diff
@@ -589,3 +464,137 @@
...
+	icon_state = "C15-0"
+	density = FALSE
+	desc = "This doesn't look very safe at all!"
+	layer = CATWALK_LAYER
+	shock_when_entered = FALSE
+	auto = TRUE
+	plane = PLANE_FLOOR
+	amount_of_rods_when_destroyed = 1
+	var/catwalk_type = "C" // Short for "Catwalk"
+	var/connects_to = list(/obj/mesh/catwalk, /obj/machinery/door) // We're working differently from grilles. We don't check a list and then another, we check all possible atoms to connect to.
```
Tarmunora, line 502: Unused var

Tarmunora (changes requested):
`obj/mesh/update_icon()` implementation should either be written in a general way that works for both grilles and catwalks, or the current implementation should be pulled down to `obj/mesh/grille/update_icon()`

Catwalks do not work, probably due to marked problems

May be more issue yet to reveal themselves

also maps in secrets will need path updates

Tarmunora:
Catwalks still do not work

TheTyrant25:
damn

TheTyrant25:
this has gotten really bad really fast. I can't be bothered to fix this merge conflict. I'll revisit it later

PR #17084, opened 2023-11-25
make cogmap2 genetics/pathology green

TheTyrant25:
[MAPPING]
## About the PR
Makes the edges of the genetics and pathology rooms green instead of red.

## Why's this needed?
Other maps typically use green in their genetics departments. Map parity is good for navigation and clarity.

## Changelog
```changelog
(u)Tyrant
(+)Cogmap2 genetics is now repainted to be green instead of red.
```


Tarmunora (approved):

TheTyrant25:
in theory the tgui bundle is fixed now and shouldnt annoy the linter

PR #17092, opened 2023-11-25
Makes the progression of blood-loss more granular

colossusqw:
[LABEL][Medical][Balance]

## About the PR 
This PR adds a chance of drowsiness, stuttering, blurry vision and shiver slowdowns when the player is between 300 and 415 blood.

## Why's this needed? 

As it stands, the divide between more than 300 blood and less than 300 blood is very binary, and you go pretty fast from very negligible effects to extremely severe ones. Hopefully this PR would make the stage just before the dangerous part of low blood pressure more significant, and make the progression of blood loss more linear and less binary.

## Changelog

```changelog
(u)Colossusqw
(+)Non-lethal bleeding now has a bit more imposing effects.
```


frawhst (approved):
lgtm

colossusqw:
Last commit changed 7 speed slowdowns to 2 speed shivers, timings remain the same as regular shiver times
Should be more on par with what's expected from bleeding, if a tad tamer

PR #17109, opened 2023-11-26
Revise some artifact implant effects

FlameArrow57:
[GAME OBJECTS][REWORK][BALANCE][ADD TO WIKI]
## About the PR 
This PR just revises some of the artifact implant effects to make them a bit less situational, effects a bit more common, and effects a bit less deadly

Good eldritch implant
-In addition to its current effects, it now provides a very slight, passive organ heal effect for all organs

Bad eldritch implant
-Changed so that at 75 brute damage, you start accumulating brute and organ damage. At 175 brute damage, original effects happen.

Bad silicon implant
-Changed so that at 75 oxy damage, you start accumulating oxygen damage. At 175 oxy damage, original effects happen.

Bad wizard implant
-Changed so that at 75 burn damage, you either continually light on fire, or take cold damage and shiver. At 175 burn damage, original effects happen

## Why's this needed? 
Implant revisions

## Changelog 

```changelog
(u)FlameArrow57
(+)Revised some artifact implant effects.
```


Tarmunora (approved):

FlameArrow57:
> In general, this will severly diminish the offensive potential of the bad implanters for little gain, imo. And this was extremly niche to begin with.
> 
> Instant death does not necessarily need to be a thing (but is hella fun/flashy when it happens), but i think once they trigger they should guarantee the death of the implanted if the implant does not get removed. So these effects could be more severe (e.g. brute one continously make organs fail and oxy straight up disabling breathing. The wizard one is ok, if the cold effect deals continous damage)

True yeah. Intent was just to make it instead of instant killing, to change it to something that makes it much harder to surive, or kill you, if not removed. I'll change it up again, to add back in the flashy effects once you reach the point where you'd die (~175 or 200 dmg?) if that sounds better

PR #17127, opened 2023-11-27
Implement preassembled pod frames

flappybatpal:
## About the PR 
Implements https://github.com/goonstation/goonstation/pull/15196
Removes the now unnecessary pod/submarine frame and Engine manifolds, Circuit boards, Control interface.

Needs Magpie manufacturer to be edited in secrets.

## Why's this needed? 
Actually makes Preassembled Pod Kits do something.
Simplifies pod/sub construction.

## Changelog 

```changelog
(u)Garash2k & Flappybat
(*)Pod/sub construction has been replaced with Preassembled Pod Kits. No longer do you need half a dozen components, just assemble and add pod armor!
```


DisturbHerb:
🥲

PR #17138, opened 2023-11-28
Fixes the "You rip up the bedsheet" bug

### LINKED ISSUE #17112: Others ripping up bedsheets displays 'you ripped up the bedsheet'
### Labels

[BUG][TRIVIAL]

### Description

As title says

### Steps to reproduce

1. Get someone to rip up a bedsheet
2. Observe the message in chat box

### Expected Behavior

To see 'John Staffman rips up the bedsheet.'

Reported by: 444explorer
Active test merges: #16370
Round log date: 2023-11-26-22-23
Reported on: 2023-11-26 23:52:44
Map: NADIR

itchybones:
## About the PR 
Fixes #17112 by changing the `end_message` of the action bar called in `clothing.dm`'s `try_rip_up` proc from `"You rip up the [src]"` to `"[user] rips up the [src]"`. The real issue is that the `end_message` of `SETUP_GENERIC_PRIVATE_ACTIONBAR` is sent through the `atom/proc/visible_message` proc, which doesn't allow for seperate messages to the user and mobs around the user, like the `mob/proc/visible_message` proc does. I'm thinking about changing the private actionbar to using the latter proc, as it's more flexible in terms of who recieves what message and because private actionbars are (afaik) only called by mobs, but I'm still unsure about that since I don't wanna risk messing with internal stuff yet.

## Why's this needed? 
Bugfixes are good, everybody loves bugfixes.

### DIFF
--- a/code/obj/item/clothing.dm
+++ b/code/obj/item/clothing.dm
@@ -88,7 +88,7 @@ ABSTRACT_TYPE(/obj/item/clothing)
 	// here for consistency; not all clothing can be ripped up
 	proc/try_rip_up(mob/user)
 		boutput(user, "You begin ripping up [src].")
-		SETUP_GENERIC_PRIVATE_ACTIONBAR(user, src, 3 SECONDS, PROC_REF(finish_rip_up), list(user), null, null, "You rip up [src].", null)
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
This is fine, you'd likely end up wanting some kind of `/datum/tri_message` type or just passing a list to fully support `tri_message` here and actionbar code is messy enough that that would non-trivial.

TobleroneSwordfish (approved):

PR #17151, opened 2023-11-29
Give ghosts a spectroscopic analysis toggle

amylizzle:
## About the PR 
Basically just gives ghosts a toggle to apply `PROP_MOB_SPECTRO` to themselves so you can click on a reagent holder and see what it is.

## Why's this needed? 
Observer main buff, and pali said I could 
image

## Changelog 

```changelog
(u)Amylizzle
(+)Ghosts now have a toggle that allows them to see a spectroscopic breakdown of reagents
```

### DIFF
--- a/code/mob/dead/observer.dm
+++ b/code/mob/dead/observer.dm
@@ -229,6 +229,7 @@
 	src.see_in_dark = SEE_DARK_FULL
 	animate_bumble(src) // floaty ghosts  c:
 	src.verbs += /mob/dead/observer/proc/toggle_tgui_auto_open
+	src.verbs += /mob/dead/observer/proc/toggle_ghost_chem_vision
 	if (ismob(corpse))
 		src.corpse = corpse
 		src.set_loc(get_turf(corpse))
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (changes requested):
Should probably either be a button or just on by default, no-one looks at verbs

TobleroneSwordfish (changes requested):
Oh actually this should probably copy over to target observers too (not hivemind members though)

amylizzle:
I seem to have broken ghost huds

Sbmhawk:
I've experienced observers trying to snoop on chem mixes or when looking for a new secret chem, I think it would be good to have a living player toggle this off for ghosts observing them if possible.

Having some bozo make a video detailing how to make a secret chem after observer snooping on one of the few people who know how to make it does not feel good.

TobleroneSwordfish:
I think we're a little past the stage of only a few people knowing how to make certain chems, we have a spoiler channel in the discord so realistically if you want to spoil yourself or upload a guide on how to make saxitoxin then there easier ways than stalking chemnerds.

TobleroneSwordfish (approved):

PR #17152, opened 2023-11-29
cleanup of shuttle_controller.dm

TheTyrant25:
[CODE QUALITY] [INTERNAL]
## About the PR
Cleans up the code in shuttle_controller.dm to match modern standards.

## Why's this needed? 
I'm about to work on something in this file and i'd rather the file be clean and neat.


code/modules/transport/shuttle/shuttle_controller.dm
```diff
@@ -1,75 +1,75 @@
-// Controls the emergency shuttle
+/// Controls the emergency shuttledisabled
```
DisturbHerb, line 1: find-replace gaffe?

TheTyrant25: huh, that's odd

PR #17153, opened 2023-11-29
Generalise gas/neutron interactions

amylizzle:
## About the PR 
The gas/neutron interactions that previously only occurred in the nuclear reactor can now occur everywhere. When a neutron crosses a turf with gas or when a neutron strikes an object containing gas, nuclear reactions can occur including:
-plasma becoming fallout and making more neutrons
-fallout becoming a random gas
-CO2 absorbing neutrons and getting warmer

There's a 3 second cooldown on any one object/turf interacting to prevent this from happening:
image

## Why's this needed? 
I just think it's neat.

Also it adds new dangers & strategies to the game. You could totally flood the reactor room with CO2 for extra shielding, or plasma for extra radiation. You can put plasma cans next to the reactor to farm fallout-in-a-can. 

## Changelog 

```changelog
(u)Amylizzle
(*)Neutrons can now interact with gasses outside of the reactor. Why not try keeping your plasma cans next to your reactor for an extra exciting shift?
```


TobleroneSwordfish:
Remind me to testmerge this at some point and engineer a meltdown to watch the profiler. I'm not entirely convinced the cooldown is an effective way to limit load since the neutrons are shot in completely random directions.

amylizzle:
The cooldown is not so much to limit processing as it is to stop plasma cans repeatedly triggering eachother while in a small group. Mini meltdown in a can.

PR #17154, opened 2023-11-29
Cleanup Reagents-Base.dm

TheTyrant25:
[CODE-QUALITY] [INTERNAL] [CHEMISTRY]
## About the PR
Unindents the entire file twice, and prefixes each reagent with `/datum/reagent/`. This is the only change done to the file.

## Why's this needed? 
this way of doing things
```
datum
  reagent
    aluminum
      name = "aluminium"
```
should be a crime

### DIFF
--- a/code/modules/chemistry/Reagents-Base.dm
+++ b/code/modules/chemistry/Reagents-Base.dm
@@ -1,963 +1,961 @@
 //Contains base elements / reagents.
-datum
-	reagent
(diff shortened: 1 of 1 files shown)


TheTyrant25:
we're gaming

DisturbHerb:
the death of relative pathing

TheTyrant25:
good.

PR #17160, opened 2023-11-29
AIs talking to comms no longer create maptext.

Paaiii:
## About the PR 
AIs speaking to supplementary radio channels no longer create maptext from their core.
The message still broadcasts locally and is visible in the sidebar like it would be for whispering.

(similar to 15510, but it doesnt remove the ability to snoop the AI's comms, which was the issue people had with that one.)

## Why's this needed? 
When AIs talk to comms, the message is also audible like normal speech to someone by their core-- for snooping purposes.
Unfortunately, this also creates maptext, which can be really confusing for people if you're trying to multitask comms with talking to someone in person.

## Changelog 

```changelog
(u)Paaiii
(+)AIs speaking to supplementary radio channels no longer create maptext from their core.
```


code/mob/living.dm
```diff
@@ -769,6 +769,7 @@
 	var/italics = 0
 	var/forced_language = null
 	var/message_range = null
+	var/skip_maptext = null
```
ZeWaka, line 772: why null, why not 0 like the other one, or even better - true and false

ZeWaka (changes requested):

code/mob/living.dm
Paaiii: was attempting to be consistent across with other ones- which was silly hdsjksd
fixed now!

TobleroneSwordfish:
I still just think this is removing a feature for a very niche complaint.

PR #17162, opened 2023-11-30
Make human equipment slot defines expand to var names

TealSeer:
[QOL] [INTERNAL]
## About the PR 

Changes the value of all the SLOT defines to be the name of the variable they correspond to on human instead of arbitrary numbers. `SLOT_IN_BACKPACK` and `SLOT_IN_BELT` weren't touched because they're stupid fake slots.

## Why's this needed? 

To avoid needing to do a giant switch statement every time you need to access the human var given the slot


TobleroneSwordfish:
Oh this is going to merge conflict

PR #17163, opened 2023-11-30
Fix being unable to shove bodies into furnace

### LINKED ISSUE #17100: Unable to shove bodies into the Kondaru Sauna furnace.
### Labels

[BUG][MINOR]

### Description

Chatbox shows action of "[PLAYER] finishes shoving [CLOWN BODY] into the furnace!" but nothing actually happens.

### Steps to reproduce

1. Kill clown.
2. Click + Drag lifeless corpse into sauna furnace, you monster.
3. Recieve text in chat saying you have shoved them in the furnace.
4. Realize that nothing has actually happened.

### Expected Behavior

Body should be put in the furnace and be burned or incinerated. Getting rid of evidence, etc.

The body was a husk, if that matters at all. I specified the Kondaru sauna furnace because it's where I ran into the bug, but I'm not sure if corpses are allowed into furnaces in general. Hope this isn't a redundant report!

Reported by: Kateaclysm
Active test merges: #16370
Round log date: 2023-11-26-06-32
Reported on: 2023-11-26 06:44:28
Map: COGMAP2

TealSeer:
[BUG] [GAME OBJECTS]
## About the PR 

Adds an `ismob` check to `/obj/machinery/power/furnace/MouseDrop_T` because it was not written to be able to handle mobs, you have to grab someone and click to do that. 

## Why's this needed? 

Fix #17100

### DIFF
--- a/code/modules/power/furnace.dm
+++ b/code/modules/power/furnace.dm
@@ -151,6 +151,9 @@ TYPEINFO(/obj/machinery/power/furnace)
 		if (!in_interact_range(src, user)  || BOUNDS_DIST(O, user) > 0 || !can_act(user))
 			return
 		else
+			if (ismob(O))
+				boutput(user, SPAN_ALERT("You have to grab [O] to stuff [him_or_her(O)] in!"))
+				return
 			if (src.fuel >= src.maxfuel)
 				boutput(user, SPAN_ALERT("The furnace is already full!"))
 				return


TealSeer:
That was my original plan but I was worried about balance

ZeWaka:
you can already do it via grab this is the same thing

TealSeer:
Will do then

code/modules/power/furnace.dm
```diff
@@ -310,3 +295,51 @@ TYPEINFO(/obj/machinery/power/furnace)
...
+
+/datum/action/bar/icon/stuff_mob_into_furnace
+	id = "stuff_into_furnace"
+	duration = 5 SECONDS
+	interrupt_flags = INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_MOVE | INTERRUPT_ATTACKED
+	var/obj/machinery/power/furnace/furnace = null
+	var/mob/victim = null
+	var/mob/user = null
+
+	New(mob/U, obj/machinery/power/furnace/F, mob/V)
```
TobleroneSwordfish, line 307: Please don't use single letter variable names like this, the argument names don't have to be different to the var names in DM, you can just do `src.user = user` and it'll figure it out.

code/modules/power/furnace.dm
```diff
@@ -310,3 +295,51 @@ TYPEINFO(/obj/machinery/power/furnace)
...
+		..()
+		src.user = U
+		src.furnace = F
+		src.icon = F.icon
```
TobleroneSwordfish, line 311: The furnace icon looks really weird here, probably better to use a generic icon like the pocket searching action bar.

code/modules/power/furnace.dm
```diff
@@ -222,6 +194,19 @@ TYPEINFO(/obj/machinery/power/furnace)
...
+			boutput(user, SPAN_ALERT("It'd probably be easier to dispose of [him_or_her(victim)] while the furnace is active..."))
+			return
+		if (!isdead(victim))
+			boutput(user, SPAN_ALERT("[victim.name] needs to be dead first!"))
+			return
+		if(victim?.buckled || victim?.anchored)
+			user.visible_message(SPAN_ALERT("[victim] is stuck to something and can't be shoved into the furnace!"))
+			return
+		actions.start(new /datum/action/bar/icon/stuff_mob_into_furnace(user, src, victim), user)
+		return
```
TobleroneSwordfish, line 208:
```suggestion
```

TobleroneSwordfish (changes requested):

PR #17171, opened 2023-11-30
Spacemas tree 2023: fish edition 🐟 

Gannets:
## About the PR 
New tree for new year. Kinda lazy, sorry. 🐟 

## Why's this needed? 
If I stop now the crushing weight of the years i've spent doing this will destroy me.


pali6:
it's December here, gotta merge

PR #17174, opened 2023-11-30
cleanup of sleeperagent.dm

TheTyrant25:
[CODE QUALITY] [INTERNAL]
## About the PR 
I was gonna do a bugfix in this file and pre emptively cleaned it. As it turns out, the bug no longer exists, but i'd already done the work cleaning.

## Why's this needed?
personally i like clean code

### DIFF
--- a/code/modules/events/sleeperagent.dm
+++ b/code/modules/events/sleeperagent.dm
@@ -1,19 +1,19 @@
 /datum/random_event/major/antag/sleeper_agent
 	name = "Awaken Sleeper Agents"
 	required_elapsed_round_time = 26.6 MINUTES
-	customization_available = 1
-	announce_to_admins = 0 // Doing it manually.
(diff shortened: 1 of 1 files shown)


code/modules/events/sleeperagent.dm
```diff
@@ -35,67 +35,67 @@
...
-			do_event(source=="spawn_antag", source)
+			src.lock = TRUE
+			do_event(source == "spawn_antag", source)
 
-	proc/do_event(var/force_antags = 0, var/source)
+	proc/do_event(var/force_antags = FALSE, var/source)
 		gen_numbers()
 		gather_listeners()
-		if (!listeners.len)
+		if (!src.listeners.len)
```
TobleroneSwordfish, line 70:
```suggestion
		if (!length(src.listeners))
```

PR #17189, opened 2023-12-01
adds spaces to the ends of 3 mutant plant prefixes

NibChocolateeny:
[Bug][Catering]

## About the PR 
Adds spaces to the end of 3 mutant plant name prefixes, so that the name+prefix are separate (like all the other prefxises)

## Why's this needed? 
fixes "Mocha Coffee" and "Latte Coffee" plants coming out as "MochaCoffee" and "LatteCoffee" when looked at/scanned
Also jelley bean, but you could debate about if "jellybean" vs "jelly bean" is better? code calls it "Jelly Bean" so i changed it too

i don't think this needs a changelog? (This is my first pr, pls be nice! :>

### DIFF
--- a/code/modules/hydroponics/plant_mutations.dm
+++ b/code/modules/hydroponics/plant_mutations.dm
@@ -223,7 +223,7 @@
 
 /datum/plantmutation/beans/jelly // hehehe
 	name = "Jelly Bean"
-	name_prefix = "Jelly"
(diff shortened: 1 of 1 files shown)


pali6 (approved):
Looks good. Thanks for fixing. Such a small fix doesn't need a changelog.

PR #17190, opened 2023-12-01
ABCU Blueprint Allowance Revamp - No Syndidoor Plz

Francisco-Franks:
[Construction][ABCU]

## About the PR 
Prep work for changing how the ABCU handles what is and isn't allowed to be scanned/deployed.

## Why's this needed? 
I deployed like 100 Listening Post doors and Pali was not amused.
Otherwise, using a list of what is allowed and not allowed is kinda clunky. We want to make this use a variable in the future, this will start us down that path probably.

## Changelog 

```changelog
(u)Metareferencer
(+)The ABCU will no longer deploy Listening Post (or any other Hardened) doors. New blueprints will not include them. Hardened doors from existing blueprints will not be created.
```

### DIFF
--- a/code/WorkInProgress/blueprints.dm
+++ b/code/WorkInProgress/blueprints.dm
@@ -262,7 +262,9 @@
 					"dir" = O.direction)
 				if (!isnull(O.icon_state)) properties["icon_state"] = O.icon_state // required for old blueprint support
 				new/dmm_suite/preloader(pos, properties) // this doesn't spawn the objects, only presets their properties
-				new O.objecttype(pos) // need this part to also spawn the objects
+				var/obj/spawned_object = new O.objecttype(pos)
+				if(!is_valid_abcu_object(spawned_object))
+					qdel(spawned_object)
 
 	proc/prepare_build(mob/user)
 		if(src.invalid_count)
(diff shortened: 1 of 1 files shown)


code/WorkInProgress/blueprints.dm
```diff
@@ -692,6 +697,13 @@ proc/load_abcu_blueprint(mob/user, var/savepath = "", var/use_whitelist = TRUE)
 	boutput(user, SPAN_NOTICE("Loaded blueprint [bp.room_name], with [turf_count] tile\s, and [obj_count] object\s."))
 	return bp
 
+// I regret everything. Proc so we can see if a door is hardened and henceallowed to be ABCUd (if it isnt hardened).
```
pali6, line 700: Remove the unnecessary bit and make it a doc comment
```suggestion
/// Checks if a thing should be allowed to be saved / loaded by the ABCU. Currently does not do the whitelist / blacklist filtering.
```

code/WorkInProgress/blueprints.dm
```diff
@@ -692,6 +697,13 @@ proc/load_abcu_blueprint(mob/user, var/savepath = "", var/use_whitelist = TRUE)
...
+proc/is_valid_abcu_object(obj/O)
+	if(istype(O, /obj/machinery/door))
+		var/obj/machinery/door/door = O
+		if(door.hardened)
+			return FALSE
+
```
pali6, line 706:
```suggestion
	return TRUE

```
I question how much you tested this code. Not returning true means that this would always fail.

pali6 (changes requested):

Francisco-Franks:
Builds what it is supposed to, does not build what it is not supposed to.

pali6 (approved):

PR #17196, opened 2023-12-02
Add max_seen argument to JPS

TobleroneSwordfish:
[INTERNAL]
## About the PR 
Adds a `max_seen` argument to JPS pathfinding as an alternative way to limit pathfinding range and implements it on guardbuddies and bots with a value of 1000 (no idea if this number is reasonable)

@pali6 let me know if I did this wrong
## Why's this needed? 
Might be faster/better than simple range, worth trying out.

### DIFF
--- a/_std/pathfinding.dm
+++ b/_std/pathfinding.dm
@@ -279,7 +279,6 @@
 		if(path)
 			for(var/i = 1 to round(0.5 * length(path)))
 				path.Swap(i, length(path) - i + 1)
-	// boutput(world, "search ended with total_seen [src.total_seen]")
 	sources = null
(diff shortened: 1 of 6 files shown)


_std/pathfinding.dm
```diff
@@ -268,7 +279,7 @@
 		if(path)
 			for(var/i = 1 to round(0.5 * length(path)))
 				path.Swap(i, length(path) - i + 1)
-
+	// boutput(world, "search ended with total_seen [src.total_seen]")
```
pali6, line 282:
```suggestion
```

pali6 (approved):

PR #17200, opened 2023-12-02
Less annoying fake attackers from hallucinating 

amylizzle:
## About the PR 
Fake attackers now fade away when you hit them, giving a brief respite from constant punching sounds and making hallucinating people look crazy as they swing wildly at the air. 
They also attack slightly slower. 

## Why's this needed? 

It's pretty annoying getting constantly inundated with punching sounds. 

## Changelog 

```changelog
(u)Amylizzle 
(*)Hallucinated attackers now dissappear when you hit them 
```

### DIFF
--- a/code/datums/components/hallucinations.dm
+++ b/code/datums/components/hallucinations.dm
@@ -247,6 +247,7 @@ ABSTRACT_TYPE(/datum/component/hallucination)
 					F.name = clone.name
 					F.weapon_name = clone_weapon
 					halluc = image(clone,F)
+					F.client_image = halluc
 					parent_mob.client?.images += halluc
 				else //try for a predefined critter fake attacker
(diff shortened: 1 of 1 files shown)


pali6 (commented):
I feel like the cleint_image should get reemoved from client.images *somewhere*

amylizzle:
Pretty sure it does get deleted when the fake_attacker object is, but I can make it explicit

pali6:
I am not aware of any code that removes an image from clients when it is disposed.

pali6:
We don't hard delete image objects because obvious performance reasons. `qdel` has nothing to do with byond's hard deletions, it just runs disposing essentially.

amylizzle:
Oh, it was working though, so I guess it was just getting gc'd when the fake attacker dropped the reference. Anyway, it is now explicitly removed in the fake attackers disposing, so should be all good.

PR #17202, opened 2023-12-02
Spawn rules

TobleroneSwordfish:
[ADMIN]
## About the PR 
Adds a "spawn rules" system that lets admins set global rules that will be applied to every player from then on when they latejoin or are cloned. Rules can also be set up before the round starts to apply to all players when they spawn in.

To set a rule, spawn `/mob/living/carbon/human/normal/assistant/the_template` (ie type `spawn template`), any varedits or proc calls done to this mob or its equipment will be recorded.
Some examples:
- Call `equip_new_if_possible(/obj/item/clothing/head/frog_hat, "slot_head") to give everyone a frog hat
- Call `safe_delete()` on the template's shoes to make everyone spawn shoeless.
- Set `color = "#0000FF"` on the template's shoes and then call `update_clothing` on the template to make everyone's shoes blue

Rules can be removed by clicking on the template mob.
## Why's this needed? 
Requested by multiple admins, hopefully this is intuitive enough to be usable while still being powerful.


ZeWaka (changes requested):
needs comments and documentation, there's like none

TobleroneSwordfish:
it's WIP!!! (okay I didn't label it so I deserve this)

PR #17205, opened 2023-12-02
add solars to the ABCU whitelist

Skeletonman0:
[OBJECTS] [QOL]
## About the PR 

image

completely untested, just added the panels and trackers to the whitelist

**this does not add the console to the whitelist, people still have to go scan one of those**

## Why's this needed? 
solar panels are a pain to mass produce in the amounts needed for solar arrays, people suggested this now that the ABCU actually works


ZeWaka:
> completely untested

no

PR #17206, opened 2023-12-03
Emagged borgs electrify doors with no confirmation prompt

kbsmilk:
[SILICONS] [QOL]
## About the PR 
Emagged borgs now skip the "Are you sure? Human harm!!" popup when spacebar+click electrifying doors.
Adds boutput messages to (de)electrifying doors because they previously had no feedback.

## Why's this needed? 
Kill humans faster.
It's a full-on antagonist. Why should it get an intrusive popup warning it about hurting humans? Nobody else gets that.

## Changelog 

```changelog
(u)Smilg
(+)Emagged silicons now skip the confirmation prompt when electrifying airlocks with spacebar+click.
```

### DIFF
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -2024,16 +2024,24 @@ TYPEINFO(/obj/machinery/door/airlock)
 		//electrify door for 30 seconds
 		if (src.secondsElectrified!=0)
 			src.shock_restore(user)
+			boutput(user, SPAN_NOTICE("De-electrified [src]."))
 		else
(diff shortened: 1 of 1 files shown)


kbsmilk:
More justification then:
It's a full-on antagonist. Why should it get an intrusive popup warning it about hurting humans? Nobody else gets that.

If you want parity for all borgs, I propose removing the popup entirely. It's clunky to use (the popups don't appear until you release spacebar, which is *weird* when electrifying multiple doors), and removal would also help out syndicate borgs too.

TobleroneSwordfish:
Mmmm no, borgs are just borgs and frankly it's probably a good thing that mass electrifying doors is clunky.

Issue #13569, opened 2023-04-04
a non emaged cyborg with emag particles


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

same as title, i had emaged particles without being emaged

### Steps to reproduce

1. be assembled as a cyborg
2. somehow have emaged particles as a non emaged cyborg

### Expected Behavior

not having them

### Additional Information & Screenshots

no it was not it being caused by low HP i had it since my assembly at full HP

there was a wraith so it is possible the frame did actually get emaged but i somehow only received the cosmetic part of the emaging

Reported by: ZWRh3
On server: Goonstation 1 Classic: Heisenbee
Active test merges: nothing
Round log date: 2023-04-04-17-51
Reported on: 2023-04-04 18:17:23
Map: NADIR

JORJ949:
Should have been closed with #17215

PR #17215, opened 2023-12-04
Remove cyborg antagonist roles on death

### LINKED ISSUE #17170: Cyborg Emag Issue
### Labels

[BUG][CRITICAL]

### Description

Laws shown as if connected to law rack, but apparently emagged.

### Steps to reproduce

1. Get emagged as cyborg.
2. Die (I died by being remotely detonated by a traitor I believe?)
3. Get rebuilt.
4. Roboticist attempts to emag again.
5. It shows as emagged for them, but for me I show as connected to my laws.
6. Roboticist is unable to emag at all.
7. End of round shows me as emagged; I still see myself connected to the law rack.

### Expected Behavior

It should be consistently shown whether I'm emagged or not rather than different people seeing different things.

Reported by: Mistysprings
Active test merges: #16370 and #17092
Round log date: 2023-11-30-09-41
Reported on: 2023-11-30 09:54:46
Map: KONDARU

DasBrain:
[silicons] [bug]

Fixes #17170, ##16113, #13569

## About the PR 

Cyborg antagonist roles (emagged/syndicate cyborg) are now (hopefully) reliably removed from cyborg antagonists.

## Why's this needed? 

Currently the roles are not removed if the head part is removed.
This leads to a bunch of bugs.

## Changelog 

```changelog
(u)DasBrain
(+)Cyborg antagonist roles are now successfully removed when the cyborg is turned into a ghost.
```


TobleroneSwordfish (changes requested):
If only it was that simple, this will lead to any syndie/emagged borg resurrected as a human through cloning or SR being classed as an antagonist. Roles being removed on death isn't really an issue for crew credits since they should use the historic antagonist list anyway (not sure why they don't)
I think the ideal handling of this is that emagged and syndie antag statuses should always be removed on death and re-added on assembly, since the antagonist status is inherently tied to the frame itself.

DasBrain:
> If only it was that simple

It is.

>this will lead to any syndie/emagged borg resurrected as a human through cloning or SR being classed as an antagonist.

No.

> Roles being removed on death isn't really an issue for crew credits since they should use the historic antagonist list anyway (not sure why they don't)

Because we do not keep historic records.

> I think the ideal handling of this is that emagged and syndie antag statuses should always be removed on death and re-added on assembly, since the antagonist status is inherently tied to the frame itself.

If only it was that simple.

TobleroneSwordfish (changes requested):
You're going to need to explain beyond "No." and "It is.", you haven't actually answered the review.

DasBrain:
So, I... removing the antag status of cyborgs when they are destroyed/brain pulled...
Makes my head hurt.

In theory... an emagged cyborg should have their antag status removed - because `remove_on_death = TRUE`.
But it is not. Emagged cyborgs keep their antag status.

So, the next best thing is: remove the antag status when they are put back into a body.
Cloning should not be a problem - `remove_on_clone = TRUE`.

That works - but... some people prefer antag removal on death.

Ok, let's try that again:
* There are two ways a cyborg can "die": `death()`, and `eject_brain()`.
* `eject_brain` does use `ghostize`. 
* `eject_brain` Currently it also removes the syndicate role - making the body non-syndicate. ARGH.
* `eject_brain` manually calls `on_death` for all antag roles - so emagged (which has `remove_on_death = true`) should be removed. It is not.
* `death` calls `eject_brain`. So - ghosts with emagged cyborg antag role should not be possible. Yet here we are.

Of course, it is bad if the antagonist role is not removed - as they do not get the "equipment" a second time.
For an emagged cyborg that was moved to a fresh frame this means:
* They still have the emagged cyborg role.
* They are connected to the law rack.
And if they are then emagged again, we have the following situation:
* They are not notified that they have been emagged (again).
* They are still connected to the law rack.
* They have the emagged cyborg antagonist role.
* And they spark.
Which is the worst thing that can happen. Also see #17170.

Cloning is not an issue - the antag role is set to be removed on cloning.
It is also very common to have cyborgs with unrelated antagonist roles - when a humanoid antagonist is borged.

TobleroneSwordfish:
I've managed to replicate and fix the syndicate cyborg brain removal issue, it's now handled the same as emag status (which yes may have bugs but at least it doesn't just delete itself on brain removal now)

I can't replicate any of the emagged bugs locally, the antag status has been removed and re-added fine every way I can test. This likely means either there's a major replication step left out of all the bug reports or it's somehow a race condition causing procs to be called in the wrong order due to server lag.

The paranoia `borged` check in this PR is a good workaround in the meantime, but please keep the status set to `remove_on_death = TRUE`  for both emagged and syndicate statuses since that's what should be happening (and I think is, most of the time)

Also apology accepted, code can be frustrating :)

DasBrain:
> I've managed to replicate and fix the syndicate cyborg brain removal issue, it's now handled the same as emag status (which yes may have bugs but at least it doesn't just delete itself on brain removal now)

Yes, I have seen b56dfdc7a7ab2bcd280e929be28655cef1337aa1.
Unfortunately, I suspect destroyed syndicate borgs will drop normal frames now. (The `emagged`, `freemodule and syndicate variables need to be transfered over to the frame.)

> I can't replicate any of the emagged bugs locally, the antag status has been removed and re-added fine every way I can test. This likely means either there's a major replication step left out of all the bug reports or it's somehow a race condition causing procs to be called in the wrong order due to server lag.

It is a very common thing, at least on RP.  
I do not know what causes this - which is frustrating.
I can send out an ahelp next time it happens - maybe it helps to reproduce it.

> The paranoia `borged` check in this PR is a good workaround in the meantime, but please keep the status set to `remove_on_death = TRUE` for both emagged and syndicate statuses since that's what should be happening (and I think is, most of the time)

I think this is a good compromise. I will update this PR to do that and synchronize the `syndicate` flag again.

Thank you for accepting my apology.

DasBrain:
Heureka! I found it!


Reproducer:
* Be an emagged/syndicate cyborg
* Have your head removed (you can do it yourself)

This will remove the head, call `update_apperance`, which will call `death`, which in turn calls `eject_brain`, but `eject_brain` checks at the beginning if there is even a head unit and returns.

In that cases the antag status is not lost.

PR #17216, opened 2023-12-04
Gives cyborgs different radio icons depending on the module

Paaiii:
## About the PR 
The PR is about exactly what it says on the tin - it gives borgs different radio icons depending on the module they're using at the time.

It also updates the old AI core icon to be in line with the new borg ones.

PRimage

## Why's this needed? 
You can tell what job a human player has by their radio icon, so being able to do so with a borg too makes sense!

## Changelog 

```changelog
(u)Paaiii
(+)Gave cyborgs module-specific radio icons.
```


TobleroneSwordfish (changes requested):
I really like this, the icons look great! The only issue I found is that borgs who haven't picked a module yet show up as civilian module instead of no-module.

flrsh (changes requested):
I don't like having images for radio icons instead of words. It takes significantly more cognitive load to figure out what is what, when chat is already very fast paced. For instance, without them being labeled, I would not be able to tell what the third and fourth icons are. 

Additionally, I think it's more important that players know who's a borg than what specific module the borg has. A lot of players aren't familiar with all the modules anyways and this would add confusion in terms of what's a job and what's a borg module (e.g. the icon for brobocop could easily be read as "food" or "chef").

Paaiii:
Hm... I can see that concern, re: symbols. I tried a few other versions when making the icons- my concerns were that blank coloured squares makes it difficult to figure out which job was which, whereas adding text like the human radio icons makes them too similar to, well, the human radio icons. 

Blank icons
PRimage - Copy (2)

Text icons
PRimage - Copy

I could do either of those, or mess around with another variation some.

Valtsu0:
I prefer the symbols. Engie should probably have a wrench tho

PR #17219, opened 2023-12-04
Taser sniper rifle

MylieDaniels:
[Game Objects] [Feature]
## About the PR 
Adds a gun, which was the original impetus for refactoring and cleaning up sniper and movement controller code.
This is a taser sniper, intended for admin spawn at first and possibly being added to the armory afterwards.
It has a windup before firing a taser bolt that becomes more powerful as it travels, a scope, and the projectile passes through windows and grilles.  This also adds code for projectiles to pass through grilles without interaction, when window_pass is set to 2.

image

Todo:
- [ ] Inhand sprite
- [ ] Back sprite

## Why's this needed? 
Hostage situations! Long range takedowns in space! Sneak attacks! Someone proposed it and it made me laugh! It's wildly impractical on most maps!

No changelog due to not being available to players yet.


MylieDaniels:
If there is interest in the concept here, let me know, but the sprite has been used for the Albatross after talking to cogwerks and others!
[#18272](https://github.com/goonstation/goonstation/pull/18272)

PR #17224, opened 2023-12-05
Fix plasmaglass shard/plant pot interaction

NibChocolateeny:
[Bug][Catering]
## About the PR 
This changes the type-check for putting a plasmaglass piece into a plant tray so that it uses all shards made of plasmaclass (instead of just a specific type of shard)
also planting a stack of plasmaglass will only use one shard and not all of them. (Like scrap weapon fixes, right?)

## Why's this needed? 
you couldn't use plasmaglass shards from broken windows for planting before, but now you can. consistancy is good! (and pali said i was allowed to - thanks pali!)
and the other bug bugged me, so i fixed it. :>

### DIFF
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -284,6 +284,7 @@
 
 			var/obj/item/parts/robot_parts/robot_frame/frame =  new(T)
 			frame.emagged = src.emagged
+			frame.syndicate = src.syndicate
 			frame.freemodule = src.freemodule
 
 			src.ghostize()


code/obj/machinery/plantpot.dm
```diff
@@ -718,14 +718,14 @@ TYPEINFO(/obj/machinery/plantpot)
 				return
 
 
-		else if(istype(W, /obj/item/raw_material/shard/plasmacrystal) && !current)
-			// Planting a crystal shard simply puts a crystal seed inside the plant pot for
-			// a moment, spawns a new plant from it, then deletes both the seed and the shard.
+		else if((istype(W, /obj/item/raw_material/shard) && istype(W.material, /datum/material/crystal/plasmaglass)) && !current)
```
TealSeer, line 721:
```suggestion
		else if((istype(W, /obj/item/raw_material/shard) && W.material?.getID() == "plasmaglass") && !current)
```

Tarmunora (approved):

NibChocolateeny:
close to fix conflict - will make a new one!

PR #17225, opened 2023-12-05
Cleanup of burn_possible

TheTyrant25:
[INTERNAL] [CODE QUALITY]
## About the PR 
In the code for `/obj/item` there exists a var called `burn_possible`. It behaved almost like a boolean but could also equal 2 instead, for memory pooling reasons.

This PR replaces all instances of `burn_possible` with `TRUE` and `FALSE` defines, turning instances of `0` into `FALSE` and values of `1` or `2` into `TRUE`. In addition, the check for `if (burn_possible == 2)` has been removed and the `else` block made the standard.

## Why's this needed? 
`pool(src)` has been replaced with `qdel(src)` in the single spot where `burn_possible == 2` was checked, meaning that the var having a value of `2` is now useless. There is no longer any need for it as of about 2 years ago, so it's best to remove the usage entirely to avoid confusion.

### DIFF
--- a/_std/defines/item.dm
+++ b/_std/defines/item.dm
@@ -223,9 +223,3 @@
 #define MATCH_UNLIT 0
 #define MATCH_LIT 1
 #define MATCH_INERT 2 /// broken or burn out
-
-
-// for the burn_possible var in /obj/item
-#define BURN_IMPOSSIBLE 0
-#define BURN_POSSIBLE 1
-#define BURN
--- a/_std/defines/item.dm
+++ b/_std/defines/item.dm
@@ -223,3 +223,9 @@
 #define MATCH_UNLIT 0
 #define MATCH_LIT 1
 #define MATCH_INERT 2 /// broken or burn out
+
+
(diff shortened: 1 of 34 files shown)


_std/defines/item.dm
```diff
@@ -223,3 +223,9 @@
 #define MATCH_UNLIT 0
 #define MATCH_LIT 1
 #define MATCH_INERT 2 /// broken or burn out
+
+
+// for the burn_possible var in /obj/item
+#define BURN_IMPOSSIBLE 0
+#define BURN_POSSIBLE 1
+#define BURN
```
TobleroneSwordfish, line 231:
```suggestion
```
I don't think this is meant to be here?

TobleroneSwordfish (changes requested):

_std/defines/item.dm
TheTyrant25: ah shit yeah that's leftover

TobleroneSwordfish (approved):

PR #17229, opened 2023-12-06
Refactor test_disposal_system

TealSeer:
[ADMIN] [STATION SYSTEMS] [REWORK]
## About the PR 

- Pull `/proc/test_disposal_system` and related objects out of WorkInProgress folder and into its own file `disposal_test.dm`
- Change to use `inonstationz` instead of checking `z` var
- `for_by_tcl` loop for disposal chutes instead of `in world` loop
- Remove unneeded call to rescan for destinations on mail chute
- Change dummy object icon from owl in a bikini to a generic piece of mail because I hate looking at it
- Use `dummy_list` for both mail test and disposal test as its cleaner and won't try to qdel shit twice
- Switch proc to report number of failed tests instead of number of successful tests
- Move `disposalHolder` creation and init to after the flushing animation on mail chutes because this caused hundreds of runtimes when `flush()` is rapidly called
- Add name of source chute to failed routing report because things that use the disposal chute type for non-disposals things (loafer, morgue chute, brig chute, etc.) will report as failures and being able to see that the source was a morgue chute helps discern false positives
- Adds admin verb in the debug tab to use the proc
- Adds debug log entry to usage of the proc

## Why's this needed? 

While testing this I caught multiple instances of broken routing/disposal networks that exist on live, it's a very useful utility for both map developers and coders alike.


ZeWaka (approved):
Ideally this would be run on all in-rotation maps as a part of unit testing, but this is good for now

code/modules/disposals/disposal_test.dm
```diff
@@ -0,0 +1,91 @@
...
+			return
+
+	var/list/dummy_list = list()
+	for_by_tcl(D, /obj/machinery/disposal)
+		if (!inonstationz(D))
+			break
+		/*
+		if (D.type != text2path(test_path))
+			continue
+		*/
```
frawhst, line 23: Probably just axe this if it's going to be commented

code/modules/disposals/disposal_test.dm
```diff
@@ -0,0 +1,91 @@
...
+				var/obj/machinery/disposal/mail/mail_chute = D
+
+				SPAWN(2 SECONDS)
+					for(var/dest in mail_chute.destinations)
+						var/obj/item/disposal_test_dummy/mail_test/MD = new /obj/item/disposal_test_dummy/mail_test(mail_chute, sleep_time)
+						MD.source_disposal = mail_chute
+						MD.destination_tag = dest
+						mail_chute.destination_tag = dest
+						dummy_list.Add(MD)
+						mail_chute.flush()
```
frawhst, line 37: I'm confused why this sleeps for 2 seconds? Seems like it'll just wait 2 seconds and then rapid fire spawn/flush all the items right after each other, so it's not reducing lag or anything.

frawhst (commented):
Seems good for testing. Maybe turn into a map correctness check at some point (with provided values, use landmarks? idk)

code/modules/disposals/disposal_test.dm
TealSeer: without it the process gets all fucked up and broken cuz it waits for all that stuff to be done before continuining, it could be a SPAWN(0) but it was 2 seconds in the original code

PR #17235, opened 2023-12-06
Allow cloning rotten people, add additional cloner defects depending on how rotten they are

frawhst:
[FEEDBACK] [MEDICAL]
## About the PR 
- Removes the NecroScan
- Allows the scan of rotted people (not skeletonized, though).
- Gives people additional cloner defects on clone based on how rotted they were when they were scanned (1 defect per decomposition stage).

## Why's this needed? 
Might be fun?? Currently decomposition varies between 4 and 10 minutes per stage, and so the difference between 'can be cloned' and 'can't be cloned' can vary in timing pretty wildly. This eases that a bit by making you harder to clone (in that you need to be patched up after, most likely), but still cloneable so you can be Alive and stuff.

Concerns:
- Steps on Robotics' toes by making rotten people less of 'immediately borg' status?
- Makes things harder for antagonists since their victims can be cloned more often?

## Changelog 
```changelog
(u)aloe
(*)Rotten people can now be cloned, but they'll come out with more cloner defects the more rotten they are. Skeletonized people still can't be cloned.
```


DisturbHerb:
It warrants a TM to see how this plays out in game, especially with antag toes. Personally, it gets rid of one more thing folks would just rush for, so I'm leaning positive for now.

Valtsu0:
I disagree with the removal of NecroScan. It could just not give additional defects (or give less if we want to nerf)

frawhst:
Closing for now due to feedback.

PR #17239, opened 2023-12-06
Fix disease bypassing 100% resistance

### LINKED ISSUE #10638: Colds bypassing disease resistance?
### Labels

[BUG][MINOR]

### Description

Catches cold even with 100% disease resistance.

### Steps to reproduce

Wear full industrial armour
Be a miner in the vacuum of space
Keep catching colds

### Expected Behavior

No germs finding their way into my lungs with the full getup, in a vacuum.

Reported by: GreenJelly
Round log date: 2022-09-04-15-25
Reported on: 2022-09-04 15:30:59

gear-space:
[Bug][Medical]
## About the PR 
Adds a check so if a mob has 100% viral resistance to disease, they should not get that disease. Should stop the cold/flu from being transmitted when user is on an air tank or has a full biohazard suit on. 

Just as a sanity check, most antag abilities that relate to diseases enable resistance bypassing on the effect or injects a reagent that imparts the disease. I also confirmed you can still get the cold/flu if given the corresponding associated_reagent since that always bypasses resistance here.

https://github.com/goonstation/goonstation/blob/990997a045c4a222d69ffd2573a29538420d806f/code/modules/chemistry/Reagents-Diseases.dm#L34 

Do note that it looks like this ability below does check if a mob has viral resistance, but it is not given to a werewolf what what I can see. This non-bypass of disease resistance looks to be intended, but if we don't want someone wearing a biohazard suit immune to getting turned to a werewolf via this power, we'd have to make an additional edit to what I've got in this PR. 
https://github.com/goonstation/goonstation/blob/990997a045c4a222d69ffd2573a29538420d806f/code/modules/antagonists/werewolf/abilities/werewolf_spread_affliction.dm#L125

## Why's this needed? 
Fixes #10638
Additional detail in bug ticket.

### DIFF
--- a/code/datums/disease.dm
+++ b/code/datums/disease.dm
@@ -323,6 +323,9 @@
 			return 0
 
 		if (istype(A,/datum/ailment/disease/))
+			// If we have 100% immunity at this point to a disease, don't get sick
+			if (resist_prob == 100)
+				return 0
 			var/datum/ailment/disease/D = A
 			resist_prob -= D.virulence
 


Tarmunora (changes requested):
I think it would be more reasonable to either change how virulence works, or just default it to `0`, rather than have a hardcoded check on 100% resist_prob like that

PR #17243, opened 2023-12-07
Minor grammar/spelling tweaks for drugs/coffee

NibChocolateeny:
[Chemistry][Bug]
## About the PR
Fixes a few small grammar and spelling issues in "Reagents-Drugs.dm." Also makes the messages for different amounts of coffee more different so it is easier to tell howmuch you have drank. But i didn't add any new ones, only changed the old ones a little bit

## Why's this needed? 
Fixing tyopos and grammar. issues is good!

@colossusqw, let me know if you don't like the new text. I tried to have it be like it was before, with the same ideas, but with fewer repeat meassages (so in different stages, it is clearer that you have too much coffee to drink!)

### DIFF
--- a/code/modules/chemistry/Reagents-Drugs.dm
+++ b/code/modules/chemistry/Reagents-Drugs.dm
@@ -520,7 +520,7 @@ datum
 						depletion_rate = 0.5
 
 				switch(caffeine_amt)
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (approved):
sure!

PR #9626, opened 2022-07-13
Cloning without records admin toggle

Luxizzle:
## About the PR 

This PR adds a toggle in the **Server Toggles** tab, allowing an admin to switch off cloning records completely. After this, the crew is left with a cloning terminal that has no more "Records" tab and when the scanner has an occupant, the "Scan" button is replaced with a "Scan & Clone" button, which will immediately clone a dead crewmate and refusing alive occupants.

Cloning from disks for security members is still available, except instead of loading the record, the option will now be to clone straight from the disk.

## Why's this needed? 

Cloning is very integral to the stations' lifecycle, and it has come up quite a few times that maybe cloning is too easy and death has no meaning due to how easy it is to clone another crew member by the click of a button. It has always been this way, so at least in my opinion, it is also hard to judge a station with something different.

Whenever a cloner gets missing, we immediately feel this whenever we want to go and clone somebody, so removing the cloner completely feels like a very drastic choice, and one that I personally don't feel would be positive. So instead of removing the cloner, this gives us the option to experiment with a different style of cloning.

With disabling records, there is more importance in your body being found, preserved and brought to medbay to be cloned, health implants and Rotbusttec implants become more important. And hopefully the feeling that death has more meaning is strengthened.

I hope this will lead to more experimentation with the cloner and its surrounding methods of revival.

PR #17262, opened 2023-12-10
Sets cloning with records to FALSE

ThatFiveGuys:
## About the PR 
This pr is a step sister pr to #17235, this sets the global/cloning_with_records established in #9626 to FALSE . This means that you cannot scan someone to have a cloning record in the computer, you can still clone directly from disks. 

## Why's this needed? 
#17235 had a lot of feedback about how that change would be nice with prescanning but not on its own. The intent of this pr is primarily to be testmerged at the devs expression that way they can easily test out this change for up to several days in a row with/without the linked pr, and observe if what is deserved is wanted, merged.

## Changelog 

```changelog
(u)Ikea and Luxizzle
(*)Scanning people to get clone records has been disabled. You can still clone directly from security disks with cloning records on them. 
```

### DIFF
--- a/code/obj/machinery/computer/cloning.dm
+++ b/code/obj/machinery/computer/cloning.dm
@@ -4,7 +4,7 @@
 // time to show the message for before removing it
 #define MESSAGE_SHOW_TIME 	5 SECONDS
 
-var/global/cloning_with_records = TRUE
+var/global/cloning_with_records = FALSE
 ADMIN_INTERACT_PROCS(/obj/machinery/computer/cloning, proc/scan_someone, proc/clone_someone)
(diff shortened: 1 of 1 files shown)


Mister-Moriarty:
> The intent of this pr is primarily to be testmerged at the devs expression that way they can easily test out this change for up to several days in a row with/without the linked pr...

Setting aside the debate of whether prescanning should be disabled by default, which is being discussed presently in the forum thread, the functionality of this PR already exists as an admin server toggle: see `/client/proc/toggle_cloning_with_records()`.
I will say that as a result of the testmerge changelog, prescans being disabled would be better articulated to players than if the admin proc were used in lieu.

PR #17266, opened 2023-12-10
portable reclaimer no longer clears leftovers

Valtsu0:
[Game Objects][Materials][QoL]
## About the PR 
Portable reclaimer no longer clears leftovers. As in: if you put 5 sheets, reclaim, put 5 sheets and then finally reclaim again, it will yield the 1 bar as if you reclaimed the 10 sheets at once. Currently all the materials would be wasted.

## Why's this needed? 
Is nice. Also it makes it so that adding the ability to reclaim trash items (cyborg legs, empty implant cases) makes sense.

There really isn't a reason to clear it

## Changelog 

```changelog
(u)Valtsu0
(+)Portable reclaimers have been upgraded to no longer disintegrate leftovers
```

### DIFF
--- a/code/obj/item/material.dm
+++ b/code/obj/item/material.dm
@@ -726,7 +726,7 @@
 	var/sound/sound_process = sound('sound/effects/pop.ogg')
 	var/sound/sound_grump = sound('sound/machines/buzz-two.ogg')
 	var/atom/output_location = null
-	var/list/atom/leftovers = null
+	var/list/atom/leftovers = list()
 
 	attack_hand(var/mob/user)
 		if (active)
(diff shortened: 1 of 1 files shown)


DisturbHerb:
I would pain-stakingly split my stacks up to magically generate more mats, myself.

Valtsu0:
> I would pain-stakingly split my stacks up to magically generate more mats, myself.

I don't see how this could be used to generate more mats

TealSeer:
Might be a good idea to add a way to eject the contents?

Xkeeper0:
> Might be a good idea to add a way to eject the contents?

the problem is that the leftovers are sub-one-unit materials. same thing as like, being able to withdraw 0.47 of a credit or whatever.

this change makes sense to me. having some way to _list_ what's left over might be nice, but if you put in 7 sheets and 13 sheets and you get 2 full units out of it that makes sense to me.

TobleroneSwordfish (approved):

PR #17271, opened 2023-12-10
Add MechComp integration to the TEG blowers

TealSeer:
[FEATURE] [STATION SYSTEMS]
## About the PR 

Adds a "toggle active" and "set target pressure" input to the TEG blowers. Switches opening the blower UI from click with multitool to click with empty hand because of the overlap with the mechcomp connection popup.

## Why's this needed? 

Opens the door for more automation setups with the TEG in the same vein as the nuclear reactor, which I think is a cool direction for engineering and gives more reason to integrate MechComp into stuff.

## Changelog 

```changelog
(u) TealSeer
(*) The blowers on the TEG can now be controlled by MechComp. Accessing the blower UI has been changed to clicking with an empty hand.
```

### DIFF
--- a/code/modules/power/TEG.dm
+++ b/code/modules/power/TEG.dm
@@ -91,6 +91,10 @@
 		reagents.add_reagent("oil", reagents.maximum_volume*0.5)
 		target_pressure = min_circ_pressure
 		target_pressure_enabled = FALSE
+		AddComponent(/datum/component/mechanics_holder)
+		SEND_SIGNAL(src, COMSIG_MECHCOMP_ADD_INPUT, "Toggle Active", PROC_REF(toggle_active))
+		SEND_SIGNAL(src, COMSIG_MECHCOMP_ADD_INPUT, "Set Target Pressure", PROC_REF(set_target_pressure))
+
 
 	proc/assign_variant(partial_serial_num, variant_a, variant_b=null)
 		src.serial_num = "CIRC-[partial_serial_num][variant_a][rand(100,999)]"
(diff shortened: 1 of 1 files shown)


TealSeer:
> "empty hand" - were silicons addressed in this?

Functions as expected, silicons can access menu with a click, including from a range

code/modules/power/TEG.dm
```diff
@@ -99,6 +103,17 @@
...
 
+	proc/toggle_active()
+		src.target_pressure_enabled = !src.target_pressure_enabled
+		logTheThing(LOG_STATION, src, "toggled blower power [src.target_pressure_enabled ? "on" : "off"] using mechcomp.")
+
+	proc/set_target_pressure(datum/mechanicsMessage/input)
+		if(!length(input.signal)) return
+		var/newpressure = text2num(input.signal)
+		if(!isnum_safe(newpressure) || newpressure == src.target_pressure) return
+		src.target_pressure = clamp(newpressure, 0, 1e5)
```
TobleroneSwordfish, line 114: this 1e5 feels like it should be a var or a define?

TobleroneSwordfish (approved):

code/modules/power/TEG.dm
```diff
@@ -99,6 +103,17 @@
 			src.serial_num += "-[variant_b]"
 			variant_b_active = TRUE
 
+	proc/toggle_active()
+		src.target_pressure_enabled = !src.target_pressure_enabled
+		logTheThing(LOG_STATION, src, "toggled blower power [src.target_pressure_enabled ? "on" : "off"] using mechcomp.")
```
TheTyrant25, line 108: Maybe the proc name should be more specific if it's exclusively for mechcomp? Like `mechcomp_toggle_active()` or something

Just future-proofing so no one accidentally uses it for something else (which is extraordinarily unlikely but still)

TheTyrant25 (approved):
lgtm

ZeWaka (approved):

PR #17277, opened 2023-12-11
resprites indoor snow tiles and snowballs

WALPVRGIS:
## About the PR 
Does what it says, I would just push this but I want to use icondiffbot first and also to leave a paper trail so ppl know the source of these easily 😅 
image

## Why's this needed? 
- Sprites were pretty old
- Requested by Pali :)
- Visual consistency
- I should finish those railings omfg


DisturbHerb:
🫡

PR #17282, opened 2023-12-11
Adds unique QM mailchutes to maps

TheTyrant25:
[MAPPING] [FEATURE]
## About the PR
Adds the sprites from #17280 as a subtype of mail chutes. Places them in all maps with mail systems (everywhere except atlas). Also makes them repressurise at speed `0.5` instead of `0.1`, buffing them slightly in favour of regular ones, since with the new mail update from zamu, they be using them a lot. Repressurisation speed may possibly need to be tweaked to be higher or lower for balance.

## Why's this needed?
They are COOL okay they need to be USED

## Changelog
```changelog
(u)Gannets, & Tyrant
(*)Cargo now has a new mail chute, which repressurises faster!
```


Gannets:
bless you, will fix sprites 🙏

Gannets (approved):
lgtm

PR #17307, opened 2023-12-12
Makes Piña Colada comply with the Law of Conservation of Mass

### LINKED ISSUE #17291: Pina colada breaks the laws of physics
### Labels

[BUG][MINOR]

### Description

30u of ingredients create 40u of product

### Steps to reproduce

1. Mix together:10u rum, 10u coconut milk, 10u pinapple juice
2. Receive 40u of pina colada. 
3. Profit???

### Expected Behavior

30u of ingredients to make 30u of product

Reported by: Asterion0
Round log date: 2023-12-11-22-06
Reported on: 2023-12-11 22:51:44
Map: PAMGOC

itchybones:
[CATERING][CHEMISTRY]

## About the PR 
Adjusts the Piña Colada recipe to result in three parts of Piña Colada instead of four parts of Piña Colada, to match the Piña Colada's recipe, which only uses three ingredients in equal parts. Fixes #17291, sort of.

## Why's this needed? 
It'd make sense for Piña Colada to adhere to the Law of Conservation of Mass, as the majority of other, non-Piña-Colada drinks already does. Since there are three reactants in the reaction, and the mass of the product is equal to the mass of the reactant(s), the amount of Piña Colada should be equal to three.

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -1687,7 +1687,7 @@
 		id = "pinacolada"
 		result = "pinacolada"
 		required_reagents = list("juice_pineapple" = 1, "rum" = 1, "coconut_milk" = 1)
-		result_amount = 4
+		result_amount = 3
 		mix_phrase = "The drink gives off the smell of a rainy beach."
 		mix_sound = 'sound/misc/drinkfizz.ogg'
 


Valtsu0:
Units are volume not mass right?
Not saysing this isn't a good change, and it makes sense to have the same volume here, but i do not want all reactions to have the same amount of products as ingridients

itchybones:
> Units are volume not mass right? Not saysing this isn't a good change, and it makes sense to have the same volume here, but i do not want all reactions to have the same amount of products as ingridients

Heya, don't worry about it, I only made this PR as the 4 parts Piña Colada seemed out of place, I'll leave the other recipes alone. Tbh, I wrote the PR description when I was tired and didn't actually mean for my over-the-top explanation to be taken super seriously.

Also, I think units would measure mass, since they don't increase with higher temperatures like volume would.

PR #17317, opened 2023-12-13
Pirate Event

flappybatpal:
## About the PR 
Adds a pirate event.

- [x] Ship
- [x] Antagonist
- [x] Event
- [ ] Objective
- [ ] Lock pirate pods

## Why's this needed? 
We have pirates but can only be used for admin gimmicks.

## Changelog 

```changelog
(u)Flappybat
(*)Adds a mid-round pirate ghost spawn event.
```


TheTyrant25:
seems like something that'd be certified organic

flappybatpal:
After feedback from the original content creators and others there is a fair point that pirates, without significant gameplay additions, would largely just overlap with Salvagers.

Closing for now as I feel it should be done as a larger project with a lot more thought about objectives and I dislike the reliance on pods,

PR #17325, opened 2023-12-14
More mail sprites

Gannets:
## About the PR 
- Adds a new sprite for mail crates.
- Better "full" light for the QM mail-chute. Full doesn't mean "there's too much stuff in here, you can't add any more" it actually means "not empty."

image

## Changelog 

```changelog
(u)Gannets
(+)Added a new sprite for mail crates.
```


code/obj/storage/crates.dm
```diff
@@ -493,6 +493,13 @@ TYPEINFO(/obj/storage/crate/chest)
 			bux.pixel_x = rand(-9, 9)
 			bux.pixel_y = rand(0, 6)
 
+/obj/storage/crate/mail
+	name = "internals crate"
+	desc = "An internals crate."
```
TheTyrant25, line 498: is this the correct name and desc? doesn't seem right.

TheTyrant25 (commented):
looks really good

code/obj/storage/crates.dm
Gannets: lol whoops it's not, but it gets set in shippingmarket.dm anyway

Gannets: will fix

PR #17337, opened 2023-12-15
use patch name in generating patch name

### LINKED ISSUE #17260: chemmaster no longer produces minipatches
### Labels

[BUG][MINOR]

### Description

when making patches with small amounts of chems, they show as the standard large patch sprite instead of mini-patches as they used to

### Steps to reproduce

1. make silver sulf
2. put beaker in chemmaster
3. make a 5u patch, or other similarly underfilled patch
4. large patch sprite

### Expected Behavior

mini patches

Reported by: Kotomata
Round log date: 2023-12-09-23-13
Reported on: 2023-12-09 23:48:54
Map: DONUT3

gear-space:
[Bug][Chemistry]
## About the PR 
Names individually generated patches from the chemaster according to patch.name instead of a static string "patch".
 patch.name is either "patch" or "mini-patch" as set by the user in the UI.

## Why's this needed? 
Fixes #17260
Chemaster totally does work! In investigating this bug report, I found when making a individual mini-patch it was being called "<item_name> patch" instead of "<item_name> mini-patch"

### DIFF
--- a/code/modules/chemistry/Chemistry-Machinery.dm
+++ b/code/modules/chemistry/Chemistry-Machinery.dm
@@ -1018,7 +1018,7 @@ TYPEINFO(/obj/machinery/chem_master)
 				// unused by log_phrase?
 				//global.phrase_log.log_phrase("patch", src.item_name, no_duplicates=TRUE)
 
-				patch.name = "[item_name] patch"
+				patch.name = "[item_name] [patch.name]"
 				patch.medical = src.check_patch_whitelist()
 				src.beaker.reagents.trans_to(patch, reagent_amount)
 


ZeWaka (commented):
i figure the complaint is also about the sprite not being the minipatch one?

gear-space:
Before TGUI, if I'm understanding the old code correctly, the chemaster would automatically make any patch under 15u a mini patch with no input needed from the user. Best I can tell as part of the TGUI push, the chemaster will only make mini patches if you select the mini patches as the output option. If you make a regular sized patch at 5u, it doesn't automatically mini it like it would have before. So this reads to me as a change in how the chemaster works, but it seems intended from the PR notes?

From testing in devetst, the chemaster makes mini patches when you request it to, they were just not named mini (but did come out as object type mini)
```
The water patch ([0x20016d1]) 
= /obj/item/reagent_containers/patch/mini 
Name: water patch
```
Happy to detach this PR from the original bug report if that'll make things cleaner 👍

PR #17338, opened 2023-12-15
Robot parts now have a material (so you can reclaim them)

Valtsu0:
[Materials][Feature]
## About the PR 
robot parts will now set the material to the one used in fabricating it (default being steel). This means that the robot parts can be reclaimed to get the materials back.

Since the material used doesn't matter, and the sprite of a borg isn't changed, changing appearance for different materials is disabled for robot parts. (Radioactive parts will still glow, and erebite has the orange pulsing effect)

Due to a limitation in material code allowing only one material being set (wires have an hard-coded exception), threads and screen heads will only give the sturdy metal back. Conveniently, these are the parts that basically every borg wants so there isn't much of a reason to reclaim them.

## Why's this needed? 
Roboticists usually end up with a bunch of extra legs and heads from roundstart borgs/part crates, so who not throw them into the reclaimer

## Changelog 

```changelog
(u)Valtsu0
(+)Robot parts can now be reclaimed.
```


Vincent983:
can heads with brains inside be reclaimed?

Valtsu0:
> can heads with brains inside be reclaimed?

Yes. Note that you can reclaim brains by just putting them into the reclaimer.

code/obj/item/mob_parts/robot_parts.dm
```diff
@@ -474,6 +482,7 @@ ABSTRACT_TYPE(/obj/item/parts/robot_parts/chest)
 /obj/item/parts/robot_parts/chest/standard
 	name = "standard cyborg chest"
 	desc = "The centerpiece of any cyborg. It wouldn't get very far without it."
+	material_amt = 1.2
```
ZeWaka, line 485: these magic numbers are gross, they should be based off of the other values

unfortunately, the only really good way to do that at compiletime isn't until 515

you could do it on the parent i guess, it's not like setting the var is /that/ expensive

ZeWaka (changes requested):

code/obj/item/mob_parts/robot_parts.dm
Valtsu0: I am not quite sure what you are suggesting me to do. I could use defines and do `material_amt = ROBOT_PART_LIMB_COST * ROBOT_PART_LIGHT_COST_MODIFIER` that way they could also be used on the fabricator recipes

code/obj/item/mob_parts/robot_parts.dm
```diff
@@ -169,13 +172,46 @@ ABSTRACT_TYPE(/obj/item/parts/robot_parts)
...
+		var/obj/item/parts/robot_parts/newitem = new result(get_turf(src))
+		newitem.setMaterial(src.material)
+
+		boutput(user, SPAN_NOTICE("You reinforce [src.name] with the metal."))
+		M.change_stack_amount(-2)
+		if (M.amount < 1)
+			user.drop_item()
+			qdel(M)
+
+		if (istype(src,/obj/item/parts/robot_parts/head) && istype(newitem,/obj/item/parts/robot_parts/head))
```
TobleroneSwordfish, line 195: `istype(src, ...` is an anti-pattern, consider just using inheritance instead (ie defining this behaviour in an override on the type you want)

code/datums/manufacturing.dm
```diff
@@ -1352,159 +1352,176 @@ ABSTRACT_TYPE(/datum/manufacture)
 /datum/manufacture/robo_frame
 	name = "Cyborg Frame"
 	item_paths = list("MET-2")
-	item_amounts = list(18)
+	item_amounts = list(ROBOT_FRAME_COST * 10)
```
TobleroneSwordfish, line 1355: Why are these all `* 10`? Why not just have the defines be 10x larger?

TobleroneSwordfish (changes requested):

code/datums/manufacturing.dm
Valtsu0: Manufacturing for some reason has the material values as 10 times larger as the material value used in the item. 

For exapmle in order for sheet to cost 0.1 metal it has to have 1 in the recipe and in order to be reclaimed for the same amount it has to have a material amount of 0.1

TobleroneSwordfish (approved):

PR #17347, opened 2023-12-16
Metal foam triggers projectile collision before self-disposal

### LINKED ISSUE #17319: Metal Foam eats (explosive) projectiles (and cause no explosion)
### Labels

[BUG] [MAJOR]

### Description

Metal Foam seems to have some special logic where it gets hit with something and deletes that and itself, I'm not sure if this is actually what's going on though, but, explosive projectiles do not cause an explosion and just vanish when hitting the foam

### Steps to reproduce

1. Find a spare RPG 
2. ..and find some metal foam
3. Aim & Shoot!
4.  No explosion

### Expected Behavior

for the explosion to happen like normal

Before: 
image

After: 
image

pgmzeta:
## About the PR 
Manually trigger the collision of projectiles with metal foam, ensuring `on_hit` is called for the projectile before the foam is disposed.

## Why's this needed? 
Fixes #17319

### DIFF
--- a/code/obj/foamedmetal.dm
+++ b/code/obj/foamedmetal.dm
@@ -43,7 +43,8 @@
 	blob_act(var/power)
 		dispose()
 
-	bullet_act()
+	bullet_act(obj/projectile/P)
+		P.collide(src.loc)
 		if(metal==1 || prob(50))
 			dispose()
 
--- a/code/obj/foamedmetal.dm
+++ b/code/obj/foamedmetal.dm
@@ -44,9 +44,9 @@
 		dispose()
 
 	bullet_act(obj/projectile/P)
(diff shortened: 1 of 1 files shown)


code/obj/foamedmetal.dm
```diff
@@ -43,7 +43,8 @@
 	blob_act(var/power)
 		dispose()
 
-	bullet_act()
+	bullet_act(obj/projectile/P)
+		P.collide(src.loc)
 		if(metal==1 || prob(50))
 			dispose()
```
Tarmunora, line 49: I think that wrapping the dispose call in `SPAWN(0)` would be preferable to recursively calling `collide()` here

PR #17354, opened 2023-12-17
Disables antag captains on classic

TobleroneSwordfish:
[BALANCE]
## About the PR 
Disables captains rolling antag on classic, bringing it back in line with RP.

## Why's this needed? 
After joining yet another round with a rampaging antag captain spawn-killing any secoffs who join, I'm finally convinced a role that spawns with AA, armour and a taser should not be able to roll antag. Also it will hopefully lead to captains being more trusted and therefore responsible (maybe?)

## Changelog 

```changelog
(u)LeahTheTech
(*)Captains can no longer spawn as antagonist roles.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -152,9 +152,7 @@ ABSTRACT_TYPE(/datum/job/command)
 	wages = PAY_EXECUTIVE
 	high_priority_job = 1
 	receives_miranda = 1
-#ifdef RP_MODE
 	allow_traitors = 0
-#endif
 	cant_spawn_as_rev = 1
 	announce_on_join = 1
 	allow_spy_theft = 0


DisturbHerb:
what if they could roll conspiracy

frawhst:
My main concern with this is that it may lead to captains just becoming part of security, if they can never roll antag at all.

TobleroneSwordfish:
I think that's okay honestly? If there's no HoS then the captain is functionally the head of the security department anyway, and defending the station should be a decent part of their duties.

Xkeeper0:
honestly, fair.

one incredibly minor wording note: it's less that captains can no longer be antagonists and more that antagonists cannot be captains.

ZeWaka:
what happen 2 captain traitor item then

TobleroneSwordfish:
Oh umm, make it a trader item and give it to the nukie commander instead? It's just a very loud megaphone. Can also technically be bought by VIPs, regional directors and inspectors, but they're very rare traitor rolls. Oh and I guess admin-created traitor captains too.

pali6:
This is not in any way gonna reduce the number of classic rampages.

Sbmhawk:
The final evolution of secHOPs, hosCAPs.

TobleroneSwordfish:
I'm just going to merge this now, classic has been noticeably better during the testmerge and I haven't seen any more sec-captains than usual.

PR #17361, opened 2023-12-17
Make Air Alarms Good

amylizzle:
## About the PR 
TGUI's air alarms and adds better logic to the atmospheric warning system.

image
image
image

## Why's this needed? 
How can I tell if a room is full of CO2 except for all the unconscious engineers?

## Changelog 

```changelog
(u)Amylizzle
(*)Air alarms have received a rework and new UI. See minor changes for details.
(+)Air alarms now have an additional "caution" state - when it's orange, you can breathe in the room but it won't be good for you. Green indicates optimal status, and flashing red means very dangerous - you will die breathing this.
```


Tarmunora (approved):
no crimes-against-light-theme
However, I'm personally of the opinion that air alarms should be more binary (i.e.: alert on problems, but don't expose the exact details of air composition), leaving that data for engineers with atmos analyzers to collect 🤷

Valtsu0:
Temperature should probably be a gauge too

amylizzle:
> no crimes-against-light-theme However, I'm personally of the opinion that air alarms should be more binary (i.e.: alert on problems, but don't expose the exact details of air composition), leaving that data for engineers with atmos analyzers to collect 🤷

The reason for the high fidelity breakdown and increased error states was because people were saying air alarms were entirely useless. The problem was that there was no differentiation between "the air is literally poison" and "someone farted near the sensor". The main intent of the PR is to fix that issue.

amylizzle:
also I tried temperature as a gauge and I couldn't make it look good :(

code/obj/machinery/alarm.dm
```diff
@@ -5,22 +5,35 @@
 /obj/machinery/alarm
 	name = "Air Monitor"
```
pali6, line 6:
```suggestion
	name = "air monitor"
```

pali6:
The air alarms should also likely have some get_desc implementation.

amylizzle:
!merge_upstream

code/obj/machinery/alarm.dm
```diff
@@ -3,24 +3,37 @@
...
-	var/datum/gas_mixture/environment
-	var/safe
-	/*var/panic_mode = 0 */
-	var/e_gas = 0
+	/// keeps track of last alarm status
 	var/last_safe = 2
+	var/datum/gas_mixture/environment
+
+	//this is a list of safe & good partial pressures of each gas. If all gasses are in the good range, the alarm will show green. If any gas is outside the safe range, the alarm will show alert. Otherwise caution.
+	//most of these values are taken from lung.dm
```
ZeWaka, line 22:
```suggestion
	/// this is a list of safe & good partial pressures of each gas. If all gasses are in the good range, the alarm will show green. If any gas is outside the safe range, the alarm will show alert. Otherwise caution.
	/// most of these values are taken from lung.dm
```

PR #17365, opened 2023-12-17
Generated TGUI Menu support for Componenets and Events

Azrun:
## About the PR 
image
image
image
image

## Why's this needed? 
image
Improve usability of adding components.
Improves usability of adding more complicated events.


code/modules/admin/debug.dm
```diff
@@ -296,7 +296,129 @@ var/global/debug_messages = 0
 	boutput(usr, SPAN_NOTICE("Proc returned: [pretty_returnval]"))
 	return
 
+/datum/proccall_editor
+	var/atom/movable/target
+	var/list/listargs
+	var/list/initialization_args
+	/// Boolean field describing if the tgui_color_picker was closed by the user.
+	var/closed
```
ZeWaka, line 304:
```suggestion
	var/closed = FALSE
```
clearer that it starts falsey

ZeWaka (approved):
could use more documentation and cleaning up, but i'm sure you'll get to that

PR #17369, opened 2023-12-17
Resprites Carved Pumpkins and Pumpkin Lanterns

MitchTheUser:
[OBJECTS] [CATERING]

## About the PR 

Resprites the Carved Pumpkin and Pumpkin Lantern, as well as adding unique inhands for each.

PumpkinsDisplay

## Why's this needed? 

Old sprites are scary, but not spooky. This brings the sprites up to par with the normal pumpkin's sprite, and who doesn't like a fashionably late Halloween pull request?


Gannets (approved):
Oh yeah that looks way better

PR #17372, opened 2023-12-18
Makes ice no longer conductive 

ThatFiveGuys:
## About the PR 

Ice now has conductivity of 2 

## Why's this needed? 
Having a material that you can get in unlimited quantities incredibly easily that's usable for fabs is just bad as it vastly devalues other materials.

## Changelog 

```changelog
(u)Ikea
(+)Ice is now an insulator
```


Valtsu0:
Hard disagree on the uncrystallinezation
Moliz is already is to get in large quantities. Ice can't be picked up with scoops, so it requires a little bit of extra effort. Also Ice is crystalline irl, would be weird making it not

TobleroneSwordfish (commented):
Keep it crystalline, remove the conductivity imo. Ice is a crystal but try making wires out of it.

TobleroneSwordfish (approved):

PR #17376, opened 2023-12-18
Add more oxy tanks to clarion toxins

TheTyrant25:
[MAPPING]
## About the PR
Increases the number of oxygen tanks from 2 to 6. Moves the empty tanks to different locations to compensate.

## Why's this needed?
Having only 2 oxy tanks is simply not enough. If cogmap has 4 and kondaru has 6, then we should make an effort to match that.

## Changelog
```changelog
(u)Tyrant
(+)Clarion's Toxins room now has more than 2 oxygen tanks.
```

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -636,6 +636,7 @@
 	name = "autoname  - SS13";
 	pixel_x = -10
 	},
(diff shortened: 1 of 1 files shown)


TheTyrant25:
image

PR #17377, opened 2023-12-18
add sec showers to every map

TheTyrant25:
[MAPPING] [RP]
## About the PR
Add small dedicated security shower cubicles to every map. Except atlas, where a shower is added to the brig toilet. I tried to place them in spots that aren't visible in the public, and in a way that doesn't impede the shape and structure of the security department. In some places the positioning is a little awkward, but I did as best I could.

Most of them use thindoors and thindows in order to stay more compact.

## Why's this needed?
Typical RP security main behaviour is to shower in the brig to raise hygiene, which can be strange if there are... actual inmates. And showering in the public washrooms is basically begging to get robbed.

## Changelog
```changelog
(u)Tyrant
(*)All Security departments now have dedicated showers for their staff that aren't just the brig showers. Except for Atlas, which has had a brig shower added.
```

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -18098,6 +18098,10 @@
 	},
 /obj/machinery/drainage,
 /mob/living/carbon/human/npc/monkey/stirstir,
+/obj/machinery/shower{
+	dir = 1;
+	pixel_y = 12
+	},
 /turf/simulated/floor/sanitary,
 /area/station/security/brig/genpop)
 "ocq" = (
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -20643,6 +20643,7 @@
 /obj/cable{
 	icon_state = "4-8"
 	},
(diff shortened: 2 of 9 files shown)


DisturbHerb:
i've never been robbed honestly, people should try it more

wolfolotl:
I want security scared of being mugged in the bathroom like the rest of us

TobleroneSwordfish:
These are also weirdly public and glass sided showers just in main sec rooms, doesn't make much sense.

Fuppy1337:
nice and convenient change, don't see a downside here. lets you get that satisfying green RP stat quicker and easier.

TheTyrant25:
i made the pr because a lot of people asked for it. Is it worth an `input-wanted` tag?

pali6:
Being robbed is a feature. The game is not supposed to be designed in a way where everything is designed to be 100% safe

PR #17382, opened 2023-12-18
Add air alarms to Cogmap

amylizzle:
## About the PR 
Easier to add these ones, lots of space. Most major rooms have coverage.


TheTyrant25 (approved):
Looks great, although this one air alarm might be behind the recharger that sticks to the north wall automatically
image

PR #17399, opened 2023-12-20
Override `/mob/proc/get_id` on `/mob/living/silicon`.

DasBrain:
[internal] [silicons] [bug]

## About the PR 

Override `/mob/proc/get_id` on `/mob/living/silicon` to return the internal botcard.

## Why's this needed? 

This will make N.A.R.C.S. not longer shoot at silicons. (They were Bad Boys.)

It also seems the right thing to do - as asking if someone has access can then be done by just asking for the mobs ID, and check if the ID has the desired access.

Currently, most other useses of `get_id(` does not apply to silicons (it is only checked for `/mob/living/carbon`).

## Changelog 

```changelog
(u)DasBrain
(+)N.A.R.C.S. should no longer shoot at silicons.
```

### DIFF
--- a/code/mob/living/silicon.dm
+++ b/code/mob/living/silicon.dm
@@ -591,6 +591,12 @@ var/global/list/module_editors = list()
 
 	. = 'sound/impact_sounds/Metal_Clang_3.ogg'
 
+/mob/living/silicon/get_id(not_worn = FALSE)
+	. = ..()
+	if(. || not_worn)
+		return
+	return src.botcard
+
 /mob/living/silicon/proc/singify_text(var/text)
 	var/adverb = pick("robotically", "synthetically", "electronically")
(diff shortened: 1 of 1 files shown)


DasBrain:
> Does that PR mean that borgs can just go into the armory without problems?

Yes.
On some maps they already can (because they do not have a N.A.R.C.S.), while on others... they can not.
Which is strange in my opinion.

DasBrain:
This can be done in a different PR.

N.A.R.C.S. are not supposed to shoot at things with security or head access, e.g. borgs.
They do.

And while they do minimal damage to people, they just stunlock borgs.
Which they also not supposed to do (they should not shoot at stunned targets, but that is only checked if the target is human)

Borg armory access is out of scope for this PR.

PR #17402, opened 2023-12-20
remove santa's aa and tweak teleport times

Retrino:
## About the PR 

- takes away santa's captain spare
- ups the teleport time to be 80 seconds, like the rest of the shorter ability times
 
## Why's this needed? 

santa is a magical being who spawns on the wizard shuttle, why should he have a golden id and not the rest of the wizards?
seriously though, santa does not need aa. it makes him 1) a lootbox, and 2) interfere with jobs that normally have access locked equipment (ie, ordering things in cargo, making ai laws, dispensing medical equipment). he has the power to teleport, spawn a flurry of items, and heal. aa on top of that just isn't necessary, and often makes him a nuisance that can just teleport away to avoid consequences. 

tldr, it makes thematic sense, and the dude just needs a nerf


RSG250:
will santa still be able to make announcement at the announcement computer

code/modules/holiday/spacemas.dm
```diff
@@ -763,11 +763,6 @@ proc/compare_ornament_score(list/a, list/b)
 
 // Santa Stuff
 
-/obj/item/card/id/captains_spare/santa
```
ZeWaka, line 766: if you do this, please make some alternative all-access card that can be invoked with `spawn re/sa` or similar for adminning/devving purposes

ZeWaka (changes requested):

ZeWaka (approved):

PR #17405, opened 2023-12-20
Adds some sweet optimisations, plus some minor ones.

Laboredih123:
[INTERNAL][PERFORMANCE]

## About the PR 
Replaces various lagchecks with sleep(-1) in init, massively speeding up init since it maximises tick usage without being too hard on entering the game. Also adds other minor optimisations that help out init such as giving asteroids an icon cache and speeding up space/New().

## Why's this needed? 
Entering the game faster is always welcome.

### DIFF
--- a/code/obj/mining.dm
+++ b/code/obj/mining.dm
@@ -897,6 +897,7 @@ TYPEINFO_NEW(/turf/simulated/wall/auto/asteroid)
 	var/mining_toughness = 1 //Incoming damage divided by this unless tool has power enough to overcome.
 	var/topnumber = 1
 	var/orenumber = 1
+	var/list/topoverlaycache = list()
 
 	dark
 		fullbright = 0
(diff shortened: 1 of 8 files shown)


code/obj/mining.dm
```diff
@@ -897,6 +897,7 @@ TYPEINFO_NEW(/turf/simulated/wall/auto/asteroid)
 	var/mining_toughness = 1 //Incoming damage divided by this unless tool has power enough to overcome.
 	var/topnumber = 1
 	var/orenumber = 1
+	var/list/topoverlaycache = list()
```
FlameArrow57, line 900: Maybe better in New() or somewhere else, thinking about hidden init proc?

code/turf/turf.dm
```diff
@@ -332,12 +332,14 @@
 /turf/space/New()
 	..()
 	if(global.dont_init_space) return
-	if (icon_state == "placeholder") icon_state = "[rand(1,25)]"
-	if (icon_state == "aplaceholder") icon_state = "a[rand(1,10)]"
-	if (icon_state == "dplaceholder") icon_state = "[rand(1,25)]"
-	if (icon_state == "d2placeholder") icon_state = "near_blank"
-	if (blowout == 1)
-		icon_state = "blowout[rand(1,5)]"
```
ZeWaka, line 340: deleetd_logik.tcl

Laboredih123: As far as i looked, these are unused. I can still add them back if you want.

code/obj/mining.dm
Laboredih123: Ight

TobleroneSwordfish (approved):
Lgtm, possible race conditions somewhere but that's a skill issue for other code.

PR #17412, opened 2023-12-22
drastically reduce asteroid field generation time

BatElite:
[performance][code quality]
## About the PR 
Asteroid turfs were constantly doing icon updates on their neighbours as they generated, which is a large waste of effort when you're spawning big balls of them. This fixes the neighbour update behaviour for new autowalls including asteroids.

The `worldgenCandidates[src] = 1` -> `worldgenCandidates += src` change is done because it's faster according to a comment elsewhere in the code. I haven't checked this, but this also makes autowalls consistent with how stuff is getting added to worldgenCandidates elsewhere.

## Why's this needed? 
I don't know how this will translate to the live servers, but some data from my machine (3x compiling with no speed-up flags on cog2, with and without the patch) dropped the time it took to generate the asteroid field from ~67s to ~20s

### DIFF
--- a/code/turf/turf_autoalign.dm
+++ b/code/turf/turf_autoalign.dm
@@ -36,14 +36,12 @@ TYPEINFO_NEW(/turf/simulated/wall/auto)
 
 	New()
 		..()
-		if (map_setting && ticker)
-			src.update_neighbors() // lessen these calls
-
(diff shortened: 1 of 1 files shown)


Laboredih123:
another big find, nice

PR #17413, opened 2023-12-22
Makes game ticker re-roll mode immediately on failure

TealSeer:
[GAMEMODES] [REWORK]
## About the PR 

- Reworks game ticker to attempt random mode selection 10 times before giving up and reverting to pre-game lobby (if mode is not a random meta mode like secret only one attempt is made)
- Adds new list argument `exclusions` to `pick_random_mode` containing modes to skip
- Changes game ticker to keep a list of failed modes to skip in subsequent rolls (not implemented for action and intrigue because they just use `pick`)
- Removes extraneous `message_admins` calls in favor of a single one after a mode is successfully chosen and set up
- Comments out world `boutput` messages related to mode setup failure because consecutive failures are now instant and can be spammy (and players don't really need to know which mode failed to roll and why)

## Why's this needed? 

Having to wait an additional 180 seconds because not enough people have an antag enabled is annoying and pointless


ZeWaka:
image

> end result was 2 players tagged as gang leaders at character creation, untagged in the normal game but the tag persisted into the player panel

TealSeer:
will look into it ty

PR #17416, opened 2023-12-22
Adds player interaction logs to pump_ui

vortex1942:
[Admin] [QoL]
## About the PR 

Adds station logs to pump_ui interactions

## Why's this needed? 

A lot of pump_ui objects have high grief potential and can be difficult for inexperienced engineers to notice they are turned off/have had lowered settings.


TobleroneSwordfish (approved):
The bump logging might get a bit spammy, but I don't imagine people use it all that much anyway.

vortex1942:
> The bump logging might get a bit spammy, but I don't imagine people use it all that much anyway.

Yeah I wasn't sure if there is a good way to limit log spam, Personally I never use the bumps because of how slow they are.

Tarmunora (approved):

PR #17419, opened 2023-12-22
Increase minimum number of nukies to two

flappybatpal:
## About the PR 
Increases minimum antag count for nuclear to two.

## Why's this needed? 
Single operatives increase the chance for the only antag to be AFK/have to leave, can result in extremely quick nukie losses, may leave players unhappy that they have no teammates.

## Changelog 

```changelog
(u)Flappybat
(*)Minimum number of nuclear operatives has increased to two.
```


Gannets (approved):
Sensible

PR #17423, opened 2023-12-22
Fixes random room runtime checker

TealSeer:
[TOOLING] [BUG]
## About the PR 

Changes random room checker code to actually load the prefabs properly, I didn't get any runtimes on my local server which is suspicious but it definitely loads all the prefabs. Also adds stack trace for catching invalid types that exist in prefabs

## Why's this needed? 

The random room checker does nothing right now due to a bug


amylizzle:
Make a random room that you know should throw a runtime and test it with that.

TealSeer:
oh good idea

TealSeer:
got it to properly runtime so looks like that works

TealSeer:
lets hear it for failing checks 🎉

ZeWaka (approved):

PR #17436, opened 2023-12-23
Declare /obj/random_item_spawner abstract

TealSeer:
[BUG] [INTERNAL]
## About the PR 

Mark `/obj/random_item_spawner` abstract as it is the base type and contains no list of items to spawn

## Why's this needed? 

I suspect this will catch out the random rooms that have occasionally been runtiming


TealSeer:
this might not be the way i think the varedited ones still work

PR #17453, opened 2023-12-25
Garbage collection? Where we're going, we don't need garbage collection.

amylizzle:
## About the PR 
OD doesn't have softdels yet, and harddels are instant so why not

## Why's this needed? 
OD compat


Laboredih123:
we still gonna work on gc problems tho right

amylizzle:
yeah, and we gonna implement ref counting in OD eventually, but for right now this is needed if we gonna get a goon OD server up

frawhst (approved):

PR #17459, opened 2023-12-27
Airlock Cycling framework.

TheTyrant25:
[FEATURE] [QOL]
## About the PR 
the sequel to #16400 !
Airlocks can now be linked with mapping helpers (similar to the welding and bolting helpers). By var editing the `cycle_id` to match on the mapping helpers, the doors will be linked together. Opening one in a system of airlocks will try to close the others. If you want double doors to function (e.g. two space facing doors in a system, opening one shouldn't open the other there, but opening a space door should close the stationside doors), you can link together two or more doors on a "side" of the airlock by using `enter_id` and setting it to match.

This PR does not add them to maps. It also doesn't refactor any files (my bad last time, sorry). It also shouldn't break all of door code like last time, but just in case, maybe it should be testmerged first >_>

## Why's this needed? 
It adds functionality that a reasonable space station really ought to have. It also gives mappers more options when constructing space facing doorways. And it's just a cool thing to have, really.


goonstation.dme
```diff
@@ -33,6 +33,7 @@ var/datum/preMapLoad/preMapLoad = new
 #include "code\__secret_public.dme"
 
 // BEGIN_INCLUDE
+#include "_std\__odlint.dm"
```
pgmzeta, line 36:
```suggestion
```

not needed, already on L19

TobleroneSwordfish (approved):
Looks good at a glance, no iterating every door in the universe 👍

PR #17471, opened 2023-12-28
Fix intermittent runtime about invalid random item spawner

TealSeer:
[BUG] [MAPPING]
## About the PR 

Change the obj critter snail to a mob critter and just stick it in the locker instead of being in the random item spawner pool cause its not an /obj anymore

## Why's this needed? 

Fixes this fucking thing
image


TealSeer:
hol up

PR #17469, opened 2023-12-28
Swap light manufactrurer and bulb/tube manufacturer availability

pgmzeta:
[game objects][balance]
## About the PR 

Move Lamp Manufacturer to General Manufacturer

Move various Light Bulbs/Tubes to Engi Specialist Manufacturer

Adds blacklight bulbs/tubes to the General Manufacturer as a secret option to ensure quartermasters have a path to completing existing light bulb/tube requisitions without requiring cross-departmental dependence.

Further shifts expectations on basic repair duties to be handled by civilian crew e.g. janitors ressponding to station repair requests. Requires less access/interaction to make more drastic changes to the station. 

## Why's this needed? 
Make it easier for janitors/staff assistants/etc. to handle 'basic' repairs on station.  

## Changelog 

```changelog
(u)glowbold
(*)The lamp manufacturer is now available in general manufacturers.
(+)Light bulb and tube boxes can still be printed from the Engineering Specialist Manufacturer. And in the General Manufacturer with a little effort.
```

PR #17473, opened 2023-12-28
Rework Lamp Manufacturer - Basic/Advanced versions

pgmzeta:
## About the PR 

Refactors the lamp manufacturer.

The base lamp manufacturer can only add / replace lights on existing fixtures.

Engineering now gets the advanced lamp manufactuer by default,

The advanced lamp manufactuers get a total of 40 metal sheet ammo stored.

Adding fixtures costs 2 sheets of metal. Removing fixtures costs 4 sheets of metal. These match the same scaling as silicons.

Both the civilian and engineering modules as well as ghostdrones get the silicon equivilant of the advanced lamp manufacturer.

Consider alongside #17469 with this new base version of the lamp manufacturer in the general manufacturer.

## Why's this needed? 
Require engineering tools to add/remove fittings, separate from light replacement/repair. Maintains an edge in engineering tools for station modification/repair.

## Changelog 

```changelog
(u)glowbold
(+)Engineering has been provided extended-capacity lamp manufacturers.
```


TobleroneSwordfish (changes requested):
Not a huge fan of splitting stuff up like this, "basic vs advanced" is just adding another thing to remember vs one tool which does one thing.

pgmzeta:
could remove the fittings functionality entirely as it's already on RCDs, and they're not terribly complex to create by hand?

pgmzeta:
on second thought, if this becomes a problem it's better to solve it then

PR #17475, opened 2023-12-28
Adds net_id to some atmospheric devices, and turns on packet control for vents, pumps, and injectors!

Laboredih123:
[ATMOS][FEATURE][WIKI]

## About the PR 
Adds net_id to any atmospheric device that uses packets.
Gives public frequency access to vents and injectors for happy nerd fun.
Gives pumps the frequency of 1225, now known as FREQ_PUMP_CONTROL, so people can use them. Frequency has chosen as this was already used by the pump control computer.

## Why's this needed? 
Net_id is good for multitools and for addressing a single device. Being able to control these devices will lead to very interesting contraptions.

## The real deal
image

## Changelog 

```changelog
(u)cringe
(*)Vents, pumps, and injectors can now be controlled through packets! Send them a command with the help value with a packet sniffer to see what they can do. Have fun!
```


Laboredih123:
dont merge this right now lol, just testmerge. there will definitely be balance kinks.

Azrun:
> You probably will need to look at the code to see the available commands. Have fun!

Code diving should not be an expectation.  The devices should support the `help` command to surface the functionality.

Laboredih123:
> > You probably will need to look at the code to see the available commands. Have fun!
> 
> Code diving should not be an expectation. The devices should support the `help` command to surface the functionality.

please wait while i figure this out

Laboredih123:
btw i did it

PR #17480, opened 2023-12-28
Fixes HE pipes showing error sprite when pulled by blackhole.

Laboredih123:
[BUG][MINOR]

## About the PR 
HE pipes now properly show either an intact or exposed sprite, fixing a bug where they would show an error sprite when getting pulled by a blackhole.

## Why's this needed? 
bug bad

### DIFF
--- a/code/modules/atmospherics/machinery/pipe.dm
+++ b/code/modules/atmospherics/machinery/pipe.dm
@@ -524,13 +524,7 @@
 	fatigue_pressure = INFINITY
 
 /obj/machinery/atmospherics/pipe/simple/heat_exchanging/update_icon()
-	if(node1 && node2)
-		icon_state = "intact"
-
-		var/node1_direction = get_dir(src, node1)
-		var/node2_direction = get_dir(src, node2)
-
-		icon_state = "[node1_direction|node2_direction]"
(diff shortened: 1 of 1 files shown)


code/modules/atmospherics/machinery/pipe.dm
```diff
@@ -524,13 +524,7 @@
...
 
 /obj/machinery/atmospherics/pipe/simple/heat_exchanging/update_icon()
-	if(node1 && node2)
-		icon_state = "intact"
-
-		var/node1_direction = get_dir(src, node1)
-		var/node2_direction = get_dir(src, node2)
-
-		icon_state = "[node1_direction|node2_direction]"
+	icon_state = (node1 && node2) ? "intact" : "exposed"
```
TobleroneSwordfish, line 527: this doesn't seem to take into account the direction of the pipe nodes?

TobleroneSwordfish (changes requested):

code/modules/atmospherics/machinery/pipe.dm
Laboredih123: thats set in map :)

Laboredih123: the intact state has dirs

PR #17509, opened 2023-12-30
Nanite Click-ability

Azrun:
## About the PR 

Adjusts clickable area for nanites to the following areas per frame.  This causes the clickable area to be consistant and easier to click.
image

## Why's this needed? 
Nanites are a foe that is typically difficult for newer players due to the difficulty in noticing and clicking on the sprite.  This improves click-ability by making the area consistent.  

## Changelog 

```changelog
(u)Azrun
(+)Nanites should be much easier to target.
```


Azrun:
.... This isn't necessary.  I didn't look at the DMI...

Azrun:
On second thought maybe it is better?

Azrun:
Resolves 2 frames with random empty pixel.

PR #17522, opened 2024-01-01
Expand cog1 public market, shrink stocks room

Sovexe:
## About the PR 
* Reduces the size of the cogmap1 stock exchange room
* Increases the size of the public market stall and the info stall
* Small location shuffle for bushes, cameras, lights, and ATMs to account for new layout
* Removes stock exchange nuke target
* Adds public market as a nuke target

## Why's this needed? 
* Reallocate premium map real estate from one of the least utilized rooms into one of the most used for gimmicks and fun rooms

## Changelog 

```changelog
(u)Sovexe
(+)The Cogmap1 public market has been expanded some, the stock exchange has been shrunk.
```


pgmzeta:
Any concerns about it still being a nukie target at that size?

Sbmhawk:
image
This seems way too small to still be a nukie target.

PR #17548, opened 2024-01-03
Re-path Zoldorf mob to Intangible

pgmzeta:
[INTERNAL][GAME OBJECTS][ZOLDORF]
## About the PR 

Re-paths `/mob/zoldorf` to `/mob/living/intangible/zoldorf`.

Recommend testmerge; don't have tons of Zoldorf hours down, personally.

## Why's this needed? 
Reduce odd special-case pathing.

## Critical TODOs
- [ ] Hearing radio
- [ ] Player store


pgmzeta:
> Only bug found so far is this #17742 I think

can you reproduce locally? Having trouble doing so

TobleroneSwordfish:
Not sure, but it definitely happened multiple times on live

PR #17555, opened 2024-01-03
Swap tread and standard leg health

ThatFiveGuys:
## About the PR 
Treads now have 60 health each, and standard legs now have 115 health each.

## Why's this needed? 
This aims to accomplish two things. The first is to make treads less of a straight upgrade compared to legs. The second thing is to make leg targetting viable against tread borgs to reduce their speed. 

## Changelog 

```changelog
(u)Ikea
(*) Treads now have 60 health for cyborgs, and standard cyborg legs now have 115 health.
```


TDHooligan:
seems a little antithetical in robot game lore terms, but then again, treads being faster than legs is also wrong.

pgmzeta (approved):
Works as expected. Might want to add an in-file define section to make sure health for left/right variations of the same arm type synced in the future, but not required.

PR #17563, opened 2024-01-04
Rework wizard polymorph to timed status effect

flappybatpal:
## About the PR 
Changes wizard polymorph to use a status effect that will revert you to your human form after eight minutes.
Stores the original mob in null with hibernating set so they can return with the same limbs, organs etc.

## Why's this needed? 
Being permanently stuck as an animal critter was kind of lame.

## Changelog 

```changelog
(u)Flappybat
(*)Wizard polymorph now changes you into an animal for eight minutes. When the spell expires your original human form and status will be restored. Status icon by CalliopeSoups.
```


flappybatpal:
Removed the assigned_role changes and check because frankly I couldn't get it to work and it seemed a bit unnecessary now you transform back.

The original mob is qdel on the assumption we _probably_ don't want a build up of humans at null if the critters die.

code/modules/status_system/statusEffects.dm
```diff
@@ -2562,3 +2562,27 @@
...
+		..()
+		var/mob/M = owner
+		if (ismobcritter(M) && isalive(M))
+			var/turf/T = get_turf(M)
+			original.set_loc(T)
+			original.hibernating = FALSE
+			M.mind?.transfer_to(original)
+			qdel(M)
+		else
+			qdel(original)
```
FlameArrow57, line 2588:
```suggestion
			qdel(original)
			
		src.original = null
```
For proper GC

code/modules/antagonists/wizard/abilities/animal.dm
```diff
@@ -115,14 +115,18 @@ var/list/animal_spell_critter_paths = list(/mob/living/critter/small_animal/cat,
 		smoke.attach(target)
 		smoke.start()
 
-		if (target.mind && (target.mind.assigned_role != "Animal") || (!target.mind || !target.client))
+		if (target.mind || (!target.mind || !target.client))
```
FlameArrow57, line 118:
```suggestion
		if (!target.mind || !target.client)
```
Currently checking mind || !mind which always returns TRUE, think the original mind check was just to check for animal?

FlameArrow57, line 118:
```suggestion
		if (!target.mind || !target.client)
```
Currently checking mind || !mind which always returns TRUE

Also lets you polymorph ghost critters, etc., if changes are kept

code/modules/status_system/statusEffects.dm
```diff
@@ -2562,3 +2562,27 @@
 			M.max_health += change
 			current_bonus = newBonus
 			health_update_queue |= M
+
+/datum/statusEffect/wiz_polymorph
+	id = "wiz_polymorph"
+	name = "Polymorphed"
+	desc = "You've been polymorphed by a wizard! It will take a few minutes for the spell to wear off."
+	icon_state = "mindhack"
```
TobleroneSwordfish, line 2570: best not to re-use this one, it has a pretty specific meaning

TobleroneSwordfish (changes requested):

code/modules/antagonists/wizard/abilities/animal.dm
flappybatpal: onStart() and onUpdate() for the action bar contain ishuman() checks. 
cast() for the targetable spell contain ishuman() checks.

Perhaps we should have another check onEnd() but no non human should even be making it to this point so the assigned_role check would be meaningless.

pgmzeta:
not seeing the new sprite when testing locally

pgmzeta (approved):

code/modules/status_system/statusEffects.dm
```diff
@@ -2562,3 +2562,29 @@
...
+	onAdd(mob/living/carbon/human/H)
+		. = ..()
+		original = H
+
+	onRemove()
+		..()
+		var/mob/M = owner
+		if (ismobcritter(M) && isalive(M))
+			var/turf/T = get_turf(M)
+			original.set_loc(T)
```
TobleroneSwordfish, line 2583:
```suggestion
			original.set_loc(M.loc)
```
This is one of the few times I think `M.loc` would make more sense than `get_turf` so that people don't get ejected from lockers etc.

code/modules/antagonists/wizard/abilities/animal.dm
```diff
@@ -115,14 +118,18 @@ var/list/animal_spell_critter_paths = list(/mob/living/critter/small_animal/cat,
...
 
-		if (target.mind && (target.mind.assigned_role != "Animal") || (!target.mind || !target.client))
+		if (!target.mind || !target.client)
 			boutput(target, SPAN_ALERT("<B>You feel your flesh painfully ripped apart and reformed into something else!</B>"))
-			if (target.mind)
-				target.mind.assigned_role = "Animal"
 			target.emote("scream", 0)
 
 			target.unequip_all()
-			var/mob/living/critter/C = target.make_critter(pick(animal_spell_critter_paths))
```
TobleroneSwordfish, line 125: Maybe better to add an argument to `make_critter` that causes it not to delete the original, rather than recreate it here? You're missing things like gender and bioeffect copying.

TobleroneSwordfish (changes requested):

flrsh:
Imo four minutes is too short especially when the cooldown is already over two minutes. I think something like seven or eight minutes would be more impactful while not being horrible.

TobleroneSwordfish (approved):
lgtm

PR #17577, opened 2024-01-06
Cyborg movement speed rebalance

ThatFiveGuys:
## About the PR 
I'd like to preface this by saying that I think this should be testmerged before merging, I cannot do sufficient testing besides making sure my code actually does what I want it to do and the speed feel right on a local. I think my values are good, but it's hard to say without seeing them in practice.

This pr changes the movement modifier of nearly every single borg part in the game. Below is a chart of all the previous and new values, alongside a few presets (the rp meta is light borg w/ screenhead + treads). The human speed sus. is the speed of a base full health human after 10 tiles (the speed below that is 1.7).
image

The key playerside changes are as follows, each heavier part now has an increasing effect on movement speed. Light borg's movement speed is unaffected, but light borg with treads is ~30% slower. Every borg above light borg with the exception of heavy borg w/ treads is faster, but treads have relatively less impact on speed. Thrusters are now slightly better speed wise then treads in every scenario.

For code side changes, weight no longer effects movement speed as it just makes balancing these values harder to minimal benefit. The weight variable is left in code in case anyone wants to use them for future borg expansions, however that variable is not used for anything. Any dev tell me if you want it removed and I will do so.

## Why's this needed? 
In short, light borgs w/ treads are too fast, and heavier borgs are too slow. This should make heavier borgs more playable, and light borgs less desirable. Also its kinda weird that standard chests are objectively better then light chests in every way, like that isnt right. 

## Changelog 

```changelog
(u)Ikea
(*)Borg speed values have been changed all across the board. Treadless light borgs are the same speed. Light and heavy borgs with treads are now slower compared to before. Every other borg body type is faster, see pr for specific values. 
```


mordent-goonstation (changes requested):
https://forum.ss13.co/showthread.php?tid=22156&pid=200165#pid200165

Hold on the no-arms case being considered. Broadly fine with the changes as described, may need some more tweaking and fine-tuning after live play. I do not think this needs test merging, we can adjust iteratively.

PR #17578, opened 2024-01-06
Hub Text shows Shuttle, Next Map, no longer shows "Mode: secret"

13spacemen:
## About PR

Hub Text shows Shuttle, Next Map, no longer shows "Mode: secret"
Also changed the map name to be cleaner in hub, i.e., COGMAP is Cogmap 1, DONUT2 is Donut 2
And Time: shows Time To Start: if the lobby is showing that ingame

## Why's this needed? 
More info is good, you can already see the shuttle ETA when you are in the lobby anyway

Mode: secret was redundant. It's always secret if admins hide the mode. If they dont hide mode it'll show it

## Changelog 

```changelog
(u)13spacemen
(+)The hub entry now shows the shuttle ETA and the next map
```


ZeWaka (changes requested):
Unfortunately, the way the hub is set up means we basically have no more room to work with. We can't add additional characters. Therefore, we need to scrub mode entirely if we're adding shuttle position. No clue what RCL means either.

13spacemen:
@ZeWaka shuttle and mode are conditional
mode wont show unless admins dont hide mode

and shuttle wont show unless theres actually a shuttle

when hub text hits the limit it doesnt corrupt it, just cuts off a bit. so if the shuttle was called and mode was not hidden, it might say "Mode: changl"

its kinda hard for me to test this exactly tho as i cant port forward, but i remember yogstation also had problem with text cutting off until they removed some <a> links

RCL means Recall (you can recall the shuttle on Goon right? i have never actually called/recalled on Goon)

I could remove the link to goonhub.com which is embedded, since its just a statistics site, so generally not useful to those in hub [unlike Discord which banned players can use]

if you want tho i can scrap Mode entirely no prob

awaiting your response

ZeWaka:
> when hub text hits the limit it doesnt corrupt it

Yes I know, I wrote and optimized the current hub text, hence the bit.ly link. The actual limit on the hub is 255 characters.

> RCL means Recall

Got it. Makes sense. The 'ESC' confuses me still though. I don't get why a shuttle that is in transit to centcom would use that. Even ETA (C) would be better imo.

> I could remove the link to goonhub.com

Neither of the links are really up for removal

> shuttle and mode are conditional mode wont show unless admins dont hide mode
> 
> and shuttle wont show unless theres actually a shuttle

The problem is /when/ all of these are true. It'll cause it to be cut off, which isn't really good. Looks kinda grody, unprofessional, and confusing.

We currently tend to be around ~230 characters. I just did 1ce55e6 which bought us another 19 or so, so we should have at least 25 more chars to safely work with now

13spacemen:
> > when hub text hits the limit it doesnt corrupt it
> 
> Yes I know, I wrote and optimized the current hub text, hence the bit.ly link. The actual limit on the hub is 255 characters.
> 
> > RCL means Recall
> 
> Got it. Makes sense. The 'ESC' confuses me still though. I don't get why a shuttle that is in transit to centcom would use that. Even ETA (C) would be better imo.
> 
> > I could remove the link to goonhub.com
> 
> Neither of the links are really up for removal
> 
> > shuttle and mode are conditional mode wont show unless admins dont hide mode
> > and shuttle wont show unless theres actually a shuttle
> 
> The problem is /when/ all of these are true. It'll cause it to be cut off, which isn't really good. Looks kinda grody, unprofessional, and confusing.
> 
> We currently tend to be around ~230 characters. I just did [1ce55e6](https://github.com/goonstation/goonstation/commit/1ce55e69112fb0d81006503e7aa630cd12612cab) which bought us another 19 or so, so we should have at least 25 more chars to safely work with now

@ZeWaka 

ESC means Escaping, i.e., escaping the station. I think in some maps Departures is called Escape or Escape Shuttle Hallway

I trimmed Time To Start (13) -> Starting (8)
Also commented out Mode:

If you would prefer ESC being ETC I don't mind, up to you

ZeWaka:
Yeah, mode wouldn't've fit. Here's an example with it included:
```
<b><a href="https://goonhub.com">Goonstation 1 Classic: Heisenbee</a></b> — The classic SS13 experience. — (<a href="http://bit.ly/gndscd">Discord</a>)<br>
Time: <b>00:46</b> | Shuttle: ETA 5:32<br>
Map: <b>COGMAP 2</b> | Next: DONUT 3</b><br>
Mode: <b>Changeling</b>
```
That's 270 characters, way over. With the `Mode:` bit removed it goes down to 241, which is much more accepting of long map names.

code/world/world.dm
```diff
@@ -41,24 +41,43 @@
...
 	else if(!ticker)
-		statsus += "Time: <b>STARTING</b><br>"
+		statsus += "<b>STARTING</b>"
+	else if(ticker?.pregame_timeleft && current_state <= GAME_STATE_PREGAME)
+		statsus += "Starting: <b>[ticker.pregame_timeleft]</b>"
+
+	if(emergency_shuttle?.online && emergency_shuttle.location < SHUTTLE_LOC_RETURNED)
+		var/timeleft = emergency_shuttle.timeleft()
+		if(timeleft)
+			var/locstr = ""
```
ZeWaka, line 55:
```suggestion
			var/locstr
```
might as well

ZeWaka (approved):

PR #17579, opened 2024-01-06
Urine removal

flappybatpal:
## About the PR 
Removes the urine reagent and all associated code.
Removes Jenkem as a side effect.
Removes the urine based saltpetre recipe.
Replaces triplepiss with triplepissed, which will set your intent to harm randomly when ingested.

Secret chem hint needs replacing.

## Why's this needed? 
Urine only really exists for people to be weird with drinks at the bar.
Jenkem is a fake drug with at the very least a lot of racist messaging online.

## Changelog 

```changelog
(u)Flappybat
(*)Removed the urine reagent and associated actions.
(*)The urine based recipe for Saltpetre has been removed.
(*)Triplepiss is replaced by triplepissed. Get angry!
```


wolfolotl:
since the forum thread got deleted, my one (1) piece of feedback: removing jenkem leaves an empty niche in the Absolutely Godawful Drugs Ecosystem that yearns to be filled. is that within the scope of this pr? probably not, but it's just something I want to bring up

flappybatpal:
> since the forum thread got deleted, my one (1) piece of feedback: removing jenkem leaves an empty niche in the Absolutely Godawful Drugs Ecosystem that yearns to be filled. is that within the scope of this pr? probably not, but it's just something I want to bring up

I don't really want to do it here as new drugs have all the balance etc implications and I already need to do a triplepiss replacement.

Sord213:
yikes.

TDHooligan:
piss is so niche i don't even see it being used for bits since the nerf. it's been discussed to death already and a lot of people have strong opinions on it.
imho removing triplepiss + associated initropidril lore sucks really REALLY hard.

flappybatpal:
> Personally, i don't care about pissing, but tripplepiss biting the dust would be worth the sacrifice.
> 
> Possible balance concern: replacing urine with ammonia in the saltpeter recipe would enable botany to make bulk amounts of saltpeter without science. I thought the slow recipe involving nitrogen was for botany.

Happy for recommendations on this.

TDHooligan:
i'll also raise that it's frustrating that the response to such negative feedback the last time this was raised, was to delete the thread about this PR instead. especially given that a lot of folk dont have github, nor eagle-eye the fact this PR was even raised in the first place. 
yes the last forum thread was awful, but deliberately reducing the audience for this PR makes it look like the decision's already been made, if it hasn't already. i'm not aware of any other controversial PRs that reduced the scope of discussion around it.

Ley-bunned:
The pull request was based on an admin vote, so there was internal discussion and ultimately a forum thread isn't the best indicator of overall feelings but uh, how willing some folks are to argue forever. We'll have a better polling system that gets more of the community sometime, the forum is a pretty small and passionate part of the community.

pali6:
> makes it look like the decision's already been made

Because it has.

PR #17580, opened 2024-01-06
Declare Ready -> Join Game [when game has already started]

13spacemen:
## About the PR 

The "Declare Ready" button shows "Join Game" instead if the game has started
Font is Verdana, bold, size 11, modern sharp rendering, done in Paint.NET

Also hides the "Ready with Antag Token" button when game starts, since you cant even redeem it anyway
This only applies if you're in lobby, with an antag token, and the game starts while you are NOT ready

## Why's this needed? 
Looks cleaner and better, makes more sense, I always thought it was strange that Goon showed "Declare Ready" when game has already started and it opens up the latejoin menu

## Changelog 

```changelog
(u)13spacemen
(+)Declare Ready button becomes Join Game when the game has started
```


pgmzeta (approved):
Works as expected, code seems reasonable.

image

image

TobleroneSwordfish (approved):
lgtm

PR #17583, opened 2024-01-06
Change bathsalts recipe

flappybatpal:
## About the PR 
Simplifies the bathsalts recipe to use catdrugs and saltpetre.

## Why's this needed? 
Existing recipe took a lot of effort to make requiring yuck, denatured_enzyme and mugwort.
After discussion Cogs suggested plant nutrients and catdrugs.

## Changelog 

```changelog
(u)Flappybat & Dr Cogwerks
(*)Bathsalts recipe has changed to cat drugs and saltpetre.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -4203,9 +4203,9 @@
 		name = "Bath Salts"
 		id= "bathsalts"
 		result = "bathsalts"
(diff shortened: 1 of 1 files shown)


Tarmunora (changes requested):
I think this recipe proposal is way too simple, especially considering how wacky the effects of bath salts are (and that you can just buy cat drugs at the diner)

TDHooligan:
yeah i have made bath salts 0 times over my entire chem career. maybe add oil in there for the holy 'botany + chem' tuple.

flappybatpal:
> I think this recipe proposal is way too simple, especially considering how wacky the effects of bath salts are (and that you can just buy cat drugs at the diner)

TBH when discussing this I forgot you could buy drugs at the diner.

PR #17590, opened 2024-01-07
Add Air Alarms to Nadir

amylizzle:
## About the PR 
More air alarms

## Why's this needed? 
Can't breathe, air is acid.


TobleroneSwordfish (changes requested):
Afaik air alarms don't interact with acid in any way, the only place that might make sense to have them is in the reactor room since engineering often floods that with funny gases.

amylizzle:
Don't rooms become vacuum when the floor is breached, even on ocean maps? I was pretty sure they do.

Also it is possible for air to be poisoned through other means.

TobleroneSwordfish:
Nope, the only way rooms can become vacuum on underwater maps is to suck up all the air with a pump. Personally I don't think it makes sense to put air alarms all over Nadir and Oshan unless they're going to send alerts about flooding, but mappers feel free to contradict me 🤷

Kubius:
Air alarms make the most sense for around the nuclear engine, but I don't particularly object to having them around and about. Some thoughts on positions:

- Please don't overlap air alarms with cameras (notable locations, not exhaustive: Arrivals halls, mechlab, outside of robot depot). Opposing walls or putting them behind closets could help.
- Sauna alarm is overlapping a light. No sauna alarm, IMO.
- There should probably be no alarms in the Warrens, even the crew hall.
- Ranch lacks an alarm (wall behind the closet seems like a good placement candidate).
- Air alarms on the west face of a northeast corner, or east face of a northwest corner, look really wonky (examples: extraction nexus, northwest crew quarters by cryo, escape overlapping a camera). This isn't 100% a dealbreaker, but is not preferred - for the extraction nexus in particular, you could put it northwest corner opposing the northeast intercom for symmetry.

amylizzle:
Okay I was gonna look into making fluid displace atmosphere or something to make this be sensible but that code gives me psychic damage so I'mma just do what Leah said and put the air alarms just in engineering.

PR #17593, opened 2024-01-07
Fix bad indent in merge_driver.py

amylizzle:
## About the PR 
merge_driver.py had an extra space on an if statement, cause fail, is fix.


TobleroneSwordfish (approved):
Interpreted language moment (how was this not found sooner??)

PR #17179, opened 2023-11-30
Toysword to deflect foam darts

MRLinew:
## About the PR 
This PR adds the ability to reflect foam projectiles with the toy sword. It impacts playfights, where someone has a toysword and someone else has a foam gun.

## Why's this needed? 
because its funny. and so toy sword wielders can practice and aspire to be as cool as the csaber against the toy guns.

## Changelog 

```changelog
(u)Froggit Dogget
(+)Blocking with a toy sword reflects foam projectiles.
```

PR #17596, opened 2024-01-08
Toy Sabres Reflect Foam Projectiles (Adopt)

pgmzeta:
## About the PR 
Adopt of #17179 - provides a common abstract `itemblock/reflect` component that is then used for `sabreblock` and `toysabreblock`

## Why's this needed? 

> because its funny. and so toy sword wielders can practice and aspire to be as cool as the csaber against the toy guns.

## Test Plan
- [ ] C-Sabre
    - [ ] Inactive held/blocking: Does not reflect
    - [ ] Active held: Does not reflect
    - [ ] Active blocking: Reflects projectile
- [ ] Hadar
    - [ ] Active held: Does not reflect
    - [ ] Active blocking: Reflects projectile
- [ ] Toy Sabre
    - [ ] Active held: Does not reflect
    - [ ] Active blocking: Does not reflect laser/ballistic projectiles
    - [ ] Active blocking: Reflects foam projectile

## Changelog 

```changelog
(u)Froggit Dogget
(+)Blocking with a toy sword reflects foam projectiles.
```


code/obj/item/toys.dm
```diff
@@ -38,6 +39,13 @@
 			else
 				playsound(U, pick(src.sound_attackF1, src.sound_attackF2), 100, 0, 0, U.get_age_pitch())
 
+
+/obj/item/toy/sword/proc/can_reflect_toy()
+	return TRUE
```
Tarmunora, line 44: nitpick but might as well just use `TRUE` as the component arg instead of using a proc ref, since it's always going to be true

PR #17597, opened 2024-01-09
Adds chaplain mechanic 'faith'. Uses it for healing

Valtsu0:
[game objects][feature][wiki][input]
## About the PR 
[Old thread](https://forum.ss13.co/showthread.php?tid=21554) (not that relevant anymore)
Adds a hidden variable faith to chaplain training
All chaplains will receive all of the produced faith. Faith is used individually.

### Faith production
Faith can only be produced by humans, critters, borgs and AI that are a player, conscious and a non-ghost role
chaplains will produce a little bit of faith based on the current faith just by meeting the previous criteria (only to themselves, lower the faith the higher the gain)
everyone else (with a few exceptions) who is in the chapel will produce faith at a much larger speed

### Healing
Healing will use 2.5% of your faith to increase the healing done up to a maximum of 30 (at 5000 faith). Will also apply to attacking vampires.
Faith will also increase the chance of successful healing
Starting amount of faith makes the bible the same as currently. Being near 0 faith makes it really bad for healing.

## Why's this needed? 
It gives a mechanical benefit for chaplains holding sermons and other events.

It also makes it possible for other faith interactions to be added, but they are outside the scope of this pr (maybe chaplain's gun, station grade?)

## Changelog 

```changelog
(u)Valtsu0
(*)Chaplain will now gather faith from people in the chapel. Faith will increase the healing done by the bible
```


code/obj/item/storage/bible.dm
```diff
@@ -107,8 +115,12 @@
 
 			else
 				var/damage = 10 - clamp(target.get_melee_protection("head", DAMAGE_BLUNT) - 1, 0, 10)
+				var/faith_gain = damage
 				if (is_atheist)
 					damage /= 2
+					faith_gain *= 2 // hitting atheists gives more
+
+				add_faith(faith_gain * FAITH_HEAL_FAIL_GAIN)
```
TDHooligan, line 123: might encourage a bit more violence than expected?

TDHooligan: nvm i see it's only if the heal fails

Valtsu0: Healing atheists will always fail so your point is valid. The number is supposed to be small enough that it's more of a joke than a legitimate thing but it still might

TDHooligan:
note that this PR was previously some questionable changes for the Faith gun, which is where the negative initial feedback came from

code/obj/item/storage/bible.dm
Tarmunora, line 123: this will give all chaplains faith

Tarmunora (changes requested):
better than the gun idea

The faith handling feels a little odd to me wrt storing it on the trait, but I suppose it works out

Not sure I'm sold on gaining faith from failed biblehealing, especially what with the interaction with atheist trait. Imo should probably be dropped

Additionally, passive chaplain faith gain could possibly stand to be nonlinear (e.g.: `faith += k / faith**0.3`), so they can return to a "baseline" level faster

TobleroneSwordfish (commented):
Might also be nice to have some occasional vague feedback messages for the chaplain when they heal someone, indicating *veeery roughly* how strong the heal was. Faith as a mechanic needs to be somewhat opaque, but if it's completely impossible to read then it's going to feel unsatisfying.

Valtsu0:
> Might also be nice to have some occasional vague feedback messages for the chaplain when they heal someone, indicating _veeery roughly_ how strong the heal was. Faith as a mechanic needs to be somewhat opaque, but if it's completely impossible to read then it's going to feel unsatisfying.

Probably should to change the healing message based on the amount, just need to come up with a few messages.

TDHooligan:
> I could change the healing message based on the amount, but can't come up with good messages.

i @ed you in discord (i think) with a backing .ogg you can play over the bible thump. just scale it in volume based on faith maybe

code/obj/item/storage/bible.dm
```diff
@@ -62,8 +68,10 @@
 							if (S)
 								S.set_up(5, 0, T, null, "#000000")
 								S.start()
-			M.HealDamage("All", heal_amt, heal_amt)
-			if(prob(40))
+			var/heal = do_heal_amt(user)
+			M.HealDamage("All", heal, heal)
+			playsound(src.loc, 'sound/effects/faithbiblewhack.ogg', heal^0.6*3+10, 1, -1)
```
TobleroneSwordfish, line 73: Not a fan of this, tying a potentially very loud and supernatural sound effect to a common action like this is going to get spammy fast. Not to mention using audio volume as the sole feedback is bad accessibility.
Please just replace this with some vague chat messages or other less obtrusive feedback.

(also `^` is bitwise AND, you mean `**` for power)

TobleroneSwordfish (changes requested):

code/obj/item/storage/bible.dm
TDHooligan: admittedly i'd built the sound to be really quiet, like a very gentle backdrop to the sound of the bible whack. though i'm not sure if the impact sound is loud in implementation ATM

TDHooligan: testing it locally yea it can get a bit spammy. i think having a single volume of like, 10, toggling above a certain faith threshold is sane-r. could also put a cooldown on the sfx. 
i swapped it out for this:
```
				if (!ON_COOLDOWN(src, "faith_sound", 1.5 SECONDS))
					SPAWN(1 DECI SECOND)
						playsound(src.loc, 'sound/effects/faithbiblewhack.ogg', 10, FALSE, -1, (rand(94,108)/100))
```

https://github.com/goonstation/goonstation/assets/6061314/9fdc4271-6fcb-4f7f-a123-811e3ef46a5b


do agree using text output is also good for accessibility.

Tarmunora (approved):

PR #14878, opened 2023-07-07
Player Piano Automatic Linking

Facsimil:
[FEATURE] [GAME OBJECTS] [QOL]

## About the PR 

Adds a new multitool function (based on #7167) that allows players to link a set of player pianos together.

## Why's this needed? 

It gets tedious linking pianos the larger the set is.

## Changelog 

```changelog
(u)Facsimil
(+)Player pianos now can be linked together automatically.
```

PR #9027, opened 2022-06-07
increase player piano maximum note cap

nefarious6th:
[qol]

## About the PR 
does the thing.
selected notecap is kinda arbitrary (again). it could also just be removed or changed to something even more rounded like 20,000. 

15,360 notes is enough for just about 120 measures of a song written in sixteenth divisions. for normal people, maybe 4 and a half minutes of music.

## Why's this needed? 
currently you can shove about 16 measures of a song in the player piano. i don't know of a song that's only 16 measures except maybe like twinkle, twinkle little star. also i really want to be able to transcribe and play an entire song. i think longer songs transcribed to player piano could be more fun, it might encourage more people to interact with it and try their favorite song versus just repetitive little loops that can get old to some people.

PR #17603, opened 2024-01-10
Player Piano Notes Rework

Facsimil:
[GAME OBJECTS] [QOL] [REWORK] [BUG] [MINOR]

## About the PR 

- Adds the Compact Note Format (CF)
    - Notes go from ```F,N,N,2``` to ```f2N```
        - First character is the note name and accidental
            - Uppercase means a sharp note, no (explicit) flats
            - Lowercase means natural
        - Second character is the octave number
        - Third character is the dynamic
            - N/n are natural
            - P/F are piano/forte
            - p/f are mezzo piano/mezzo forte
    - Rests are a single letter, ```R```/```r```
        - May be followed by a number indicating multiple rests, going from R1 to R9999
- Adds the Very Compact Note Format (VCF)
    - Is not meant to be human-readable (like ChemiCompiler)
    - Notes are a single character
        - There are 88 keys on a piano
        - Characters go from 33 (!) to 120 (x), with each character representing a key
    - Rests have four variants
        - Numbered variants indicate multiple rests
        - `{` is one rest
        - `|#` is a rest with one digit
        - `}##` is a rest with two digits
        - `~###` is a rest with three digits
    - Does not support dynamics
    - There are no delimiters
- Adds a format validator that's used for choosing which `build_notes` proc is called
- Changes how notes are counted
    - The length of list holding the ```given_notes``` split by ```|``` is used instead of comparing it to the max note amount times eight
        - The VCF builds notes from the input alone, since there's no delimiters
    - ```MAX_NOTE_INPUT``` was reduced from 15360 to 1920 to account for the note counting change
        - 1920 was found by multiplying 120 (measures) by 16 (16th note), from [PR #9027](https://github.com/goonstation/goonstation/pull/9027)
- Fixed bug where the first given-input's length wasn't checked
    - Steps:
        - Checks if the length of ```note_input``` (not yet set) is below limit
        - Sets ```note_input``` to  ```given_notes```'s value
    - This allowed a user to add up to 3750 notes (text box character limit), though it would error when you tried to set the notes again since ```note_input``` had the previous input's value that was over the limit
- Updated ```player_piano.txt```
    - Includes details about the new formats
    - Changed the structure from using ```<br>``` to using ```<p>```, ```<ul>```, ```<li>```, ```<hr>```, ```<code>```, and ```<h#>```
    - Added information about PRs [#14878](https://github.com/goonstation/goonstation/pull/14878) and [#16647](https://github.com/goonstation/goonstation/pull/16647)

## Why's this needed? 

The current format uses more characters than is necessary. Also fixes a minor bug.

## Changelog 

```changelog
(u)Facsimil
(*)The Player Piano has two new note formats, see the revised Player Piano book for more details
```


DisturbHerb (changes requested):
The piano ought to be backwards compatible with scores that have been authored prior to the notation change. 

Changelog should link to the PR or any relevant documentation for the notation change, as it is presently not sufficient in informing users of the potential obsolescence of their score back catalogue.

Facsimil:
- Backwards Compatibility
    - I'll make a ```Choose Notes``` and a ```Choose Notes (Compact)``` option when interacting with the player piano
    - I'll have to keep the original ```build_notes``` function
- Documentation
    - I'll update the Player Piano book and have the changelog ask players to see it for relevant documentation

Facsimil:
Should the backwards compatibility have a set time limit or last indefinitely?

DisturbHerb:
> I'll make a Choose Notes and a Choose Notes (Compact Format) option when interacting with the player piano

You would ideally not need to prompt the user about these at all, since their meaning would not be particularly obvious for first time users. The piano should be able to identify which notation system has been used if the input is valid.

> Should the backwards compatibility have a set time limit or last indefinitely?

Be like Microsoft. There's no reason to rip out the old system and make old sheets obsolete at any point in the future, we're not hurting for individual bytes of data (when it comes to the player piano).

Facsimil:
Since it's a new format, a header (like ```<!DOCTYPE html>```) probably would work.

Facsimil:
There, it'll now check the notes. Other than the book, is there anything else?

Facsimil:
The book is updated to include the new format, reworked the structure, and updated a few other things.

DisturbHerb:
Re-opened upon request.

Facsimil:
Closing for now as another project has taken priority. I intend to re-open at a later date.

PR #17606, opened 2024-01-10
Improve module tool reorder UI

mordent-goonstation:
## About the PR 

Generally improves scrolling/reordering behavior of Cyborg Module Rewriter, making it easier to move tools up/down. I am very aware that this is not a "best" solution for the function of ordering module tools, but this should hopefully give a better UI to use as a framework going forwards.

image

## Why's this needed? 

- Chasing a tool up/down the list to press the up/down arrows on it to reorder it is awkward.
- Better scrollbar behavior allows for tools list and modules list to grow independently.

## Changelog 

```changelog
(u)Mordent
(+)Reordering tools within the Cyborg Module Rewriter made easier.
```


ZeWaka:
Why not drag and drop?

mordent-goonstation:
> Why not drag and drop?

Have at it, dude. This was a between-rounds-of-Darktide improvement.

PR #17610, opened 2024-01-10
Add +x to pre-commit.hook

amylizzle:
## About the PR 
makes pre-commit.hook executable so I can commit things on Linux. Should have no effect on Windows dev.


ZeWaka (approved):
my bad

PR #17617, opened 2024-01-10
Fix stamina bar not updating to max if you regenerate a tiny amount of stamina

### LINKED ISSUE #17422: Case where you can't reach full stamina
### Labels

[BUG][MINOR]

### Description

There's a small case where you can't reach max stamina if the remaining stamina you have left is less than your regen rate

### Steps to reproduce

1. Reach 195 stamina as a human
2. Have 10 stamina regen rate
3. Hover over stamina bar at the top right and see it stay at 195 stamina

### Expected Behavior

For stamina to increase to full stamina

_No response_

MeggalBozale:
[Bug] [Code-Quality] [UI]
## About the PR 
Title explains this, if you regenerate only ~3% of your stamina the stamina bar icon won't update. This is fine for performance until you are regenerating to max from 195/200 stam, in which case it appears you don't have max stamina when you do.

This change improves the check to self-document a bit better, and to update when the stamina bar just reaches its max value. Also changes the bar update to when you regenerate <5% of your stamina

https://github.com/goonstation/goonstation/assets/27376947/ab9327c8-69a1-45f0-a5d6-f742a171de3d

## Why's this needed? 
Seems pretty annoying that it won't update all the way.
Fixes #17422

### DIFF
--- a/code/obj/screen.dm
+++ b/code/obj/screen.dm
@@ -78,7 +78,13 @@
 
 	proc/update_value(var/mob/living/C)
 		last_update = TIME
-		if(C.stamina_max <= 0 || abs(C.stamina - last_val) * 32 / C.stamina_max <= 1) return //No need to change anything
+
+		if (C.stamina_max <= 0)
+			return
+		// Don't change if we have barely changed stamina to begin with, but change if we just hit the max value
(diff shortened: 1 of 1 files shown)


MeggalBozale:
image
literally the only relevant part of the clip since it cuts out

PR #17620, opened 2024-01-11
cleanup of food_and_drink/plants.dm

NibChocolateeny:
[C-Code-Quality][A-Hydroponics]

## About the PR
cleans up modules/food_and_drink/plants.dm and makes code/descriptions nicer. this means:
- make "edible" var all true/false instead of ture/false and 1/0
- break up some long comments and lines for read ability (not all - not sure if it will break things)
- gets rid of broken description on bean plantes (I can fix in later pr!)
- minor grammar/wording fixes to descriptions aned other messages (onions red, not yellow; coconut not cut in slice; etc)

## Why's this needed? 
better code! Maybe? reviewers tell me, maybe it is not so good :<

### DIFF
--- a/code/modules/food_and_drink/plants.dm
+++ b/code/modules/food_and_drink/plants.dm
@@ -491,8 +491,10 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 /obj/item/reagent_containers/food/snacks/plant/melon/george
 	name = "rainbow melon"
 	crop_prefix = "rainbow "
-	desc = "Sometime in the year 2472 these melons were required to have their name legally changed to protect the not-so-innocent. \
+	desc = "Back in 2013, these melons were required to have their name legally changed to protect the not-so-innocent. \
 	Also for tax evasion reasons."
+	// Dr. George was likely dead by 2013 (lotta radiation), if the lawsuit was from his estate regarding the use of his name by gardengear.
+	// This is just a headcanon for what happened - I don't know the goon lore, ok?
 	icon_state = "george-melon"
 	throwforce = 0
 	w_class = W_CLASS_NORMAL
@@ -783,7 +785,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 
 /obj/item/reagent_containers/food/snacks/plant/pear
 	name = "pear"
-	desc = "Whether or not you like the taste, its freshness is appearant." // boo!
+	desc = "Whether or not you like the taste, its freshness is appearant."
 	icon_state = "pear"
 	planttype = /datum/plant/fruit/pear
 	bites_left = 1
--- a/code/modules/food_and_drink/plants.dm
+++ b/code/modules/food_and_drink/plants.dm
@@ -73,7 +73,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 		made_reagents = TRUE
 
 	attack(mob/target, mob/user, def_zone, is_special = FALSE, params = null)
-		if (src.edible == 0)
+		if (!src.edible)
 			if (user == target)
 				boutput(user, SPAN_ALERT("You can't just cram that in your mouth, you greedy beast!"))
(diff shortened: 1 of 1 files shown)


NibChocolateeny:
...why are the tags not there?

pgmzeta:
> ...why are the tags not there?

it doesn't expect you to put the A/C as a prefix on your tags (e.g. `[Hydroponics]` would work).

code/modules/food_and_drink/plants.dm
```diff
@@ -488,11 +491,12 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 /obj/item/reagent_containers/food/snacks/plant/melon/george
 	name = "rainbow melon"
 	crop_prefix = "rainbow "
-	desc = "Sometime in the year 2472 these melons were required to have their name legally changed to protect the not-so-innocent. Also for tax evasion reasons."
+	desc = "Sometime in the year 2472 these melons were required to have their name legally changed to protect the not-so-innocent. \
```
TheTyrant25, line 494: i know it wasnt you who wrote this but the game is set in the year 2053. 2472 is... quite a lot after that. perhaps change the year.

code/modules/food_and_drink/plants.dm
```diff
@@ -777,7 +783,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 
 /obj/item/reagent_containers/food/snacks/plant/pear
 	name = "pear"
-	desc = "Whether or not you like the taste, its freshness is appearant."
+	desc = "Whether or not you like the taste, its freshness is appearant." // boo!
```
TheTyrant25, line 786: ?

TheTyrant25 (approved):
Looks mostly alright. You definitely double checked to make sure there's no weird instances of `edible == 2`, right?

code/modules/food_and_drink/plants.dm
NibChocolateeny: ok!

NibChocolateeny:
@TheTyrant25 I lookd at rest of code with "edible," saw only mix of 1/0/TRUE/FALSE, no 2 or -1. maybe next fix is to standardize eddible in other files?

code/modules/food_and_drink/plants.dm
```diff
@@ -488,11 +491,14 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 /obj/item/reagent_containers/food/snacks/plant/melon/george
 	name = "rainbow melon"
 	crop_prefix = "rainbow "
-	desc = "Sometime in the year 2472 these melons were required to have their name legally changed to protect the not-so-innocent. Also for tax evasion reasons."
+	desc = "Back in 2013, these melons were required to have their name legally changed to protect the not-so-innocent. \
```
ZeWaka, line 494:
```suggestion
	desc = "Back in 2013, these melons were required to have their name legally changed to protect the not-so-innocent. \
```
we don't reference real life in the game generally, please change this

code/modules/food_and_drink/plants.dm
```diff
@@ -1522,7 +1533,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 
 /obj/item/reagent_containers/food/snacks/plant/purplegoop
 	name = "purple goop"
-	desc = "Um... okay then."
+	desc = "Um...okay then."
```
ZeWaka, line 1536: why

code/modules/food_and_drink/plants.dm
```diff
@@ -632,7 +639,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/plant)
 /obj/item/reagent_containers/food/snacks/plant/chili
 	name = "chili pepper"
 	crop_suffix = " pepper"
-	desc = "Caution: May or may not be red hot."
```
ZeWaka, line 635: i disagree

ZeWaka (approved):

PR #17627, opened 2024-01-11
Changes infusion chem for mutating synthkidneys from ammonia to water

NibChocolateeny:
[Hydroponics][QoL][Add to Wiki]

## About the PR 
Changes the infusion chemical needed to mutate synthmeat plant into synthkidney plant from Ammonia to Water

## Why's this needed?
infusion for sythekidneys used to be pee! but there is no more pee in the game, so someone changed it to ammonia
Problem: ammonia is a very good insfuser chemical. it makes plants grow faster and is used often on all plant kinds. If medical wants synthmeat fast, it is good to put ammonia into synthmeat plant.
having ammonia turn synthmeat into synthkidney is not good if you only want fast-growing synth plant but not synthkindey.

change to water because water is relatyed to kidneys, and also not generrally used for infusing. 

## Changelog

```changelog
(u)NibChocolateeny
(*)Synthkidneys are now mutated from Synthmeat plants by infusing with Water.
```


TDHooligan:
my take is add minor changelog and add to wiki tag. 

this looks good

TobleroneSwordfish (approved):

PR #17636, opened 2024-01-12
Calrion chemlab ergonomics

MintyPhresh:
[mapping] [rework]
## About the PR
This PR changes the layout of Clarion's chemlab so that almost all pathways are two tiles wide, as well as adding more useable table space.
image

## Why's this needed? 
With Barrels and Condensers being added to Science, chemlabs need more space to do stuff in. The one-tile-wide spaces in Clarion's old chemlab left only one spot for a barrel that didn't block a walkway somehow, and the placement of phones, glass recyclers, and even traditionally floor-mounted equipment like heaters on lab tables left little room to place condensers properly.

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -9725,13 +9725,29 @@
 	},
 /area/station/hallway/secondary/central)
 "aRt" = (
+/obj/mapping_helper/wingrille_spawn/auto/reinforced,
+/obj/cable,
+/obj/cable{
+	icon_state = "0-2"
+	},
+/obj/window_blinds/cog2/middle{
+	dir = 4
+	},
+/obj/machinery/door/poddoor/blast{
+	density = 0;
+	dir = 1;
+	icon_state = "bdoormid0";
+	id = "chemblast";
+	layer = 5;
+	name = "Chemistry Blast Shield";
+	opacity = 0;
+	tag = "icon-bdoormidc1"
+	},
 /obj/disposalpipe/segment/mail{
 	dir = 4
 	},
-/turf/simulated/floor/purpleblack{
-	dir = 4
-	},
-/area/station/hallway/primary/east)
+/turf/simulated/floor/plating,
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Oh this looks so much better! The only thing I would say is maybe keep the request console next to the window so you can check it while sitting at the hatch, and move the chair back in front of the chemistry dispenser as that's usually where chairs are in chemlabs.

TobleroneSwordfish:
Oh and actually looking at it, all the south facing sprites facing the wall like that looks really strange 🤔

MintyPhresh:
> Oh this looks so much better! The only thing I would say is maybe keep the request console next to the window so you can check it while sitting at the hatch...

Thanks! 😄  I chose to move the dispenser there so that those loathsome scientists would be closer to our glorious society as they use the dispenser. Easy change, but I'll wait for input from others before changing it because I kind of like it like this.

> and move the chair back in front of the chemistry dispenser as that's usually where chairs are in chemlabs.

Expect a fix for this within the next few minutes

> Oh and actually looking at it, all the south facing sprites facing the wall like that looks really strange 🤔

Necessary evil that also happens on Kondaru and similar. Even if we had north-facing sprites, the backs of machinery wouldn't exactly make it clear what they do. I tried it with just the heaters, and...
image
...Didn't like it.

MintyPhresh:
New north wall looks like this. Moving the chair in front of the dispenser with the old configuration had the front desk chair next to the dispenser chair, which unsettled me for some reason.
image

TealSeer:
Big fan of this, clarions chem lab is insanely cramped, especially if there's more than one person working there

frawhst (approved):

TobleroneSwordfish (changes requested):
Looks like this is now reverting the air alarms PR, make sure you have the map merge driver installed so you're not accidentally reverting changes.

MintyPhresh:
Hopefully that fixes it.

PR #17640, opened 2024-01-13
Adds coffee to bar & soda dispenser

MintyPhresh:
[Chemistry] [QoL]
## About the PR
Adds Coffee (just Coffee, not the freshly-brewed variety) to the Soda dispenser used in the bar, as well as the not-yet-fully-implemented bar dispenser.

## Why's this needed?
8 recipes for drinks do not accept espresso, and must use either coffee or freshly-brewed coffee. This change makes it easier to make these drinks-- no more buying a cuppa from a vending machine, or running out to a carafe every time you need a cuppa for a mixed drink!


Roninnozlo:
why not just make the recipes accept espresso

MintyPhresh:
This way only changed two lines of code

TDHooligan:
make the 'americano' recipe in the espresso machine turn into 'coffee' with a chemistry recipe (2 espresso + 3 water = 5 coffee). they're close enough. also adds the opposite of the coffee -> espresso reaction

MintyPhresh:
Going to implement feedback from Roninnozlo and TD-- which means different PRs than this.

PR #17657, opened 2024-01-13
Resprites THE MONSTER and MacGuffin

tanker4390:
[CATERING]
## About the PR 
THE MONSTER, now with an equally humorous but moreso in-line with the other food sprite, and MacGuffin as well.

ezgif-2-d720c3404b
ezgif-3-c558edf86d

## Why's this needed? 
Updated sprites, yeah.


Azrun:
New monster sprite reads more as a sandwich abomination than a burger abomination.  Seems less round and more baked rectangle on top.

Dim had a great comment: 
> My take would be to focus more on the bun and less on the meat for an implication of size
https://discord.com/channels/182249960895545344/890231630097182720/1196228311957917826

TDHooligan:
why would you remove the gag of 'burger so big it cant fit in an iconstate' ):

Absolute-Baller:
> why would you remove the gag of 'burger so big it cant fit in an iconstate' ):

I'm just a peanut-poster but yes that part of it is too iconic to remove. it also allows you to imply more depth, which it's currently lacking

AXLFAN2000:
That maguffin looks great but don't feel like the new monster sprite looks as burgery or as messy as the old one does, it just looks like a huge rectangle instead of a cartoonishly tall tower of Burger

flappybatpal:
Looks more like a burger now!

TobleroneSwordfish:
Current sprites suck, I'm going to merge this if no-one stops me 😈

TobleroneSwordfish:
Okay github tried to stop me by removing my ability to fix merge conflicts but I foiled it by just direct committing the PR here: bce36f9 so it IS merged despite what this page says. Sorry it stale-closed in the first place, this was a good PR.

PR #17659, opened 2024-01-14
Moves the QM locker from the refinery to the inside of cargo (Cog1).

cheekybrdy:
[MAPPING] [BALANCE]
## About the PR 
Moves the QM locker from the public refinery to a room above the output chute of cargo.

## Why's this needed? 
Even if the QM's locker isn't a highly tided locker, its still a concerning anomaly to have it out in the open. Input is wanted on if the resecuring into a seperate room may be too secure.

## Changelog 

```changelog
(u)Cheekybrdy
(+)The cargo locker in Cogmap1 is no longer in the refinery and now in the cargo bay office above cargo's output chute.
```


cheekybrdy:
image
A picture of the area added

TealSeer (commented):
I think you broke something here, the map diff is huge and seems to touch most of the map

cheekybrdy:
gonna close and rebranch this, should be the easiest way to fix

PR #17674, opened 2024-01-15
Moves the QM locker from the refinery into cargo

cheekybrdy:
## About the PR 
A redo of #17659 due to git demons. Moves the QM locker from the public refinery to a room inside cargo above the output chute.

## Why's this needed? 
While QM locker is barely tided, it being in a public area is a anomaly which probably is best not left ignored.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Moved the QM locker to a new room inside cargo from refinery on Cogmap1.
```


TobleroneSwordfish (changes requested):
You appear to have unintentionally changed some consoles in the stock trading room by public market, and there's a locker in a wall at cargo

TobleroneSwordfish (approved):
Looks good to me, adds another more secluded area for cargo to put shady things in too

Tarmunora:
is mdb lying or did you accidentally make the entrance airlock to QM sideways

cheekybrdy:
Yep I did, beeg oppsies

PR #17686, opened 2024-01-16
adds the ability to trap pizza boxes with mousetraps/landmines

itsloley:
[game objects][feature]

## About the PR 

simply adds the ability to rig pizza boxes with mousetraps or landmines, identically to the way you currently can with other storage containers!
_(you just place an armed mousetrap/rigged mousetrap/landmine inside of a pizza box and close it, and open it for a surprise!)_

## Why's this needed? 

would be pretty funny to hand out pizzas to crewmembers only to be met with a buttbomb to the face (or, if you're eviler, bombs!)

## Changelog 

```changelog
(u)Lyy
(+)Pizza boxes can now be rigged with mousetraps or landmines.
```

### DIFF
--- a/code/obj/item/kitchen.dm
+++ b/code/obj/item/kitchen.dm
@@ -917,6 +917,22 @@ TRAYS
 
 		. = ..()
 
+	proc/mousetrap_check(mob/user)
+		if (!ishuman(user) || is_incapacitated(user))
+			return FALSE
+		for(var/obj/item/mousetrap/MT in src.food_inside)
+			if (MT.armed)
+				user.visible_message(SPAN_ALERT("<B>[user] opens the pizza box and sets off a mousetrap!</B>"),\
+					SPAN_ALERT("<B>You open the pizza box, but there was a live mousetrap in there!</B>"))
+				MT.triggered(user, user.hand ? "l_hand" : "r_hand")
+				return TRUE
+		for(var/obj/item/mine/M in src.food_inside)
+			if (M.armed && M.used_up != TRUE)
+				user.visible_message(SPAN_ALERT("<B>[user] opens the pizza box and sets off a [M.name]!</B>"),\
+					SPAN_ALERT("<B>You open the pizza box, but there was a live [M.name] in there!</B>"))
+				M.triggered(user)
+				return TRUE
+
 	proc/toggle_box(mob/user)
 		if (length(src.contents - src.food_inside) > 0)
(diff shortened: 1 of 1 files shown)


code/obj/item/kitchen.dm
```diff
@@ -917,6 +917,22 @@ TRAYS
 
 		. = ..()
 
+	proc/mousetrap_check(mob/user)
+		if (!ishuman(user) || is_incapacitated(user))
```
TobleroneSwordfish, line 921: What's this check for? Surely we want it to go off when any mob opens it, and if we've gotten here then we can assume whoever opened it was able to open it.

You should probably handle a null user though, since the calling method does. Just make sure the messages don't play if the user is null, the `triggered` procs should handle it fine.

TobleroneSwordfish (changes requested):

code/obj/item/kitchen.dm
itsloley: i didn't really think about it much, truth be told most of this i just nabbed from the storage.dm method of mousetrap trapping and repurposed it for the pizza box :p

code/obj/item/kitchen.dm
```diff
@@ -917,6 +917,20 @@ TRAYS
 
 		. = ..()
 
+	proc/mousetrap_check(mob/user)
+		for(var/obj/item/mousetrap/MT in src.food_inside)
+			if (MT.armed)
+				user?.visible_message(SPAN_ALERT("<B>[user] opens the pizza box and sets off a mousetrap!</B>"),\
+					SPAN_ALERT("<B>You open the pizza box, but there was a live mousetrap in there!</B>"))
+				MT.triggered(user, user.hand ? "l_hand" : "r_hand")
```
TobleroneSwordfish, line 925: This will still runtime with a null user, you can't access a var of `null`

TobleroneSwordfish (changes requested):

itsloley:
third time's the charm

code/obj/item/kitchen.dm
```diff
@@ -942,6 +956,9 @@ TRAYS
 			src.UpdateIcon()
 
 		else
+			if (src.mousetrap_check(user))
+				return FALSE
```
Tarmunora, line 960: Hm. Feels a little odd to leave the box closed

Tarmunora (approved):

PR #17692, opened 2024-01-17
Makes tank explosions scale with volume

TobleroneSwordfish:
## About the PR 
Scales down the power of tank explosions by a factor of their volume relative to the default tank volume, so we get:
- Standard 70l tank = 192 power (maxcap)
- Mini oxygen tank = 108
- Pocket oxygen tank = 50

Assuming a theoretical near-perfect mix of cold oxygen and hot plasma.
This also means that tanks of >70l volume will produce larger explosions (still bounded by maxcap) but afaik there aren't any currently.

## Why's this needed? 
You do not need 5 boxes of pocket sized maxcap hand grenades! Bad clown!

## Changelog 

```changelog
(u)LeahTheTech
(*)Tank explosions now scale with the volume of the tank, smaller tank = less boom :(
```

### DIFF
--- a/code/obj/item/tank.dm
+++ b/code/obj/item/tank.dm
@@ -155,7 +155,8 @@ Contains:
 			air_contents.react()
 			pressure = MIXTURE_PRESSURE(air_contents)
 
-			var/volume_scale = air_contents.volume / 70 //wooo magic number (70 is the default volume of an air tank)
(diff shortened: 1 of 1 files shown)


imawxome:
having some way to make small air tanks into modified bombs would be pretty cool

PR #17703, opened 2024-01-17
sponges wring out when used in-hand

### LINKED ISSUE #17689: Infinitely wring-able sponge
### Labels

[BUG][MINOR]

### Description

Wringing out a sponge doesn't actually wring out the fluid, you can keep wringing it out to create huge floods of water. I don't know why this happened, I was a rancher and used the sponge that spawned in the cog 1 ranch.

### Steps to reproduce

1. Grab a sponge (in this case, the cog 1 ranch sponge)
2. Fill it at a sink
3. Wring it out
4. Keep wringing it out and create a massive puddle

### Expected Behavior

Can only wring it out once before needing to refill it

Sponging the floor like usual still depleted the water from the sponge

Reported by: ATDoop
Active test merges: #17354, #17548, and #17413
Round log date: 2024-01-16-21-52
Reported on: 2024-01-16 22:15:02
Map: COGMAP

pgmzeta:
[game objects][bug]
## About the PR 
Transfers the sponge's contents to the floor instead of doing a touch reaction when using in-hand.

## Why's this needed? 
Fixes #17689 - sponges should be empty after wringing out

## Test Plan
Holding a wet sponge,
- [x] Clicking on self applies touch 
- [x] Pressing the use in-hand key dumps the reagents onto the floor

### DIFF
--- a/code/obj/item/mops_cleaners.dm
+++ b/code/obj/item/mops_cleaners.dm
@@ -514,7 +514,7 @@ TRASH BAG
 	. = ..()
 	var/turf/location = get_turf(user)
 	if (location)
-		src.reagents.reaction(location, TOUCH, src.reagents.total_volume)
+		src.reagents.trans_to(location, src.reagents.total_volume)
 
 /obj/item/sponge/attackby(obj/item/W, mob/user)
 	if (istool(W, TOOL_CUTTING | TOOL_SNIPPING))


NibChocolateeny:
not sure if needed after 25d2e73 ?

TobleroneSwordfish:
Looks like `trans_to` skips the touch reaction that's supposed to happen when you pour out certain reagents (phlog fireflash etc.) so I think we'll stick with the current fix for now.

PR #17726, opened 2024-01-19
Meteor parallax fix

### LINKED ISSUE #17684: Meteor parallax doesn't match direction of meteor approach
### Labels

[BUG][MINOR]

### Description

It goes the wrong way

### Steps to reproduce

1. meteor approaching the east in 300 seconds
2. wait 300 seconds
3. parallax shows meteors heading north.

### Expected Behavior

meteors heading east

Reported by: ChaoticTyrant
Active test merges: #17548
Round log date: 2024-01-16-14-50
Reported on: 2024-01-16 16:03:50
Map: KONDARU

TheTyrant25:
[QOL]
## About the PR
So the meteor parallax had a bug previously. What would happen is that every time a meteor event happened, the background parallax would scroll upward only. If it was approaching from the south, it wouldn't appear at all.

This PR fixes that so that meteor approach direction on the parallax and in the game matches. This also adds support for meteors approaching from all directions by adding 4 layers at once (although it could probably be made into one animated layer if necessary).

## Why's this needed?
Fixes #17684


TobleroneSwordfish (approved):
I think the init assumption is wrong, but this works fine and seems to fix the "from all angles" option that apparently didn't work before?

TheTyrant25:
init assumption?

PR #17727, opened 2024-01-19
make wingrilles layer lower

TheTyrant25:
[MAPPING] [INTERNAL] [QOL]
## About the PR 
Lowers the layer of auto wingrille spawners, so that things like crates and fire alarms layer above them in the map editor. This shouldn't affect the game at all, because they get removed on round start and replaced.

## Why's this needed? 
It's now easier to see in the map editor if something was accidentally placed on a window's tile, since it won't be under the wingrille spawner. It will also presumably mean that changes appear in mapdiffbot better.

### DIFF
--- a/code/modules/mapping/helpers/wingrille_spawner.dm
+++ b/code/modules/mapping/helpers/wingrille_spawner.dm
@@ -92,6 +92,7 @@
 	win_path = /obj/window/auto
 	full_win = TRUE
 	no_dirs = TRUE
+	layer = LATTICE_LAYER	// to stop it rendering over things in map editors
 #ifdef PERSPECTIVE_EDITOR_WALL
 	icon_state = "wingrille_new"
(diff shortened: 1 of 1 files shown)


code/modules/mapping/helpers/wingrille_spawner.dm
```diff
@@ -92,6 +92,7 @@
 	win_path = /obj/window/auto
 	full_win = TRUE
 	no_dirs = TRUE
+	layer = LATTICE_LAYER	// to stop it rendering over things in map editors
```
TobleroneSwordfish, line 95: Won't this render below turfs? Or is that not how lattices work?

TheTyrant25: the lattice layer is actually TURF_LAYER+0.1 so that they can appear on plating, so it should be okay.

PR #17736, opened 2024-01-20
Colorful Roachperson Carapaces!

MintyPhresh:
[mutantraces][feature]
## About the PR
Allows for roachpeople to choose from the full range of skin colors, the same way contributors can.

## Why's this needed?
(copied from the forum thread [here](https://forum.ss13.co/showthread.php?tid=22161))

As of right now, the only ways to get custom-colored skin for a character is either to have the much-coveted Contributor medal, or to go to Genetics in-round (if it's staffed) and ask them to change the color (it has to be simple to do on the annoying RGB knobs since you can't input a hex code, unlike the hair dye machine).

It's frustrating to me that a powerful character customization was put behind a medal so rarely given out to players, and under such arbitrary methods as "bother an admin after you've done some indeterminate (but significant) amount of work for the codebase." With every other character customization option that I know of either being unlocked at the start or unlocked by playing the game (the contest medals do not customize your character), it really just... doesn't seem fair.

I was recently gifted the medal arbitrarily after doing pretty much nothing for the codebase (just made some minor changes to chemlab organization and changed alcoholic drinks), and I've really been enjoying having it, especially for a Blattodean that I roleplay as a mantid. I think being able to change exoskeleton color fits the mutrace well-- and it allows for a variety of insect species to be played using the mutrace, e.g. wasps.

## Changelog 
```changelog
(u)Minty
(*)Roachpeople just got a whole lot more colorful!
```

### DIFF
--- a/_std/macros/ismob.dm
+++ b/_std/macros/ismob.dm
@@ -19,6 +19,7 @@
 #define ishuman(x) istype(x, /mob/living/carbon/human)
 #define iscow(x) (istype(x, /mob/living/carbon/human) && istype(x:mutantrace, /datum/mutantrace/cow))
 #define ispug(x) (istype(x, /mob/living/carbon/human) && istype(x:mutantrace, /datum/mutantrace/pug))
+#define isroach(x) (istype(x, /mob/living/carbon/human) && istype(x:mutantrace, /datum/mutantrace/roach))
 #define isfrog(x) (istype(x, /mob/living/carbon/human) && istype(x:mutantrace, /datum/mutantrace/amphibian))
 #define isskeleton(x) (istype(x, /mob/living/carbon/human) && istype(x:mutantrace, /datum/mutantrace/skeleton))
 #define iscritter(x) istype(x, /obj/critter)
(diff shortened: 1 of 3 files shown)


flrsh:
I think this would interfere with the visual design space for other bugs, which we do have plans for as a separate project. I generally also like having different mutants with different customizations, so that each brings something unique and potentially compelling in terms of "why play this". I know we're at a frustrating point right now where a lot of potential alternative customizations aren't viable pending a rework, but I would rather things be done cleanly.

Glamurio:
This is in line with pug fur colors really, but I'd love if instead we could simply have a limited palette of options available to choose from. Not necessarily all colors, but sensible colors ranging from brown, beige to gray, green or even black.

MintyPhresh:
That could probably be done using the same framework that humans have for skin colors, where there's a few starting hues and you can change how dark or light it is. Might rework this PR to do that when I have the time, but frankly if the colors you listed are all people can get, then it's pretty much just the list as we have it, plus green... not much of a change at all.

Besides, it doesn't change what Flourish said about not wanting to have to revert this change when the planned change to bugs' visual design comes to fruition eventually. 

Frankly-- I don't know what to make of this PR not getting closed yet-- if there's already something else in the works with bug customization, and it's big enough that it takes priority over this PR, then I don't really see what's keeping this PR afloat. Are we just waiting for other devs to weigh in?

PR #17741, opened 2024-01-21
Adds a microbomb cloning module as a 6 tc traitor roboticist item

unfunnyperson:
## About the PR 
Adds a new 6 tc traitor item for roboticists, the microbomb cloning module.
image

The module can be removed with a screwdriver.

When installed on a cloner this will replace the supply of cloner implants with a special brand of microbomb. The module can dispense up to 25 micros.

These microbombs 
- have 1/2 the power of a regular micro
- always drop the brain of target upon explosion (allows for traitor roboticists to create a lot of borgs and also prevents permadeath if a nerd decided to install the module then delete every record.)
- will always detonate regardless of suicide
- break upon surgical removal
- wont appear on medical scans

In addition cloner microbombs report health values to the cloning console like a regular implant but after 10 minutes or so the reported health values start to become extremely weird and might tip off players to the module's existence.

## Why's this needed? 
Traitor roboticists only have 2 unique items. Another item would provide more choice to roboticists while also tying into the gimmick of borg army. 

## Changelog 

```changelog
(u)UnfunnyPerson
(*)Added a new 6 tc traitor item for roboticists, the Microbomb Cloning Module.
```


code/obj/item/implant.dm
```diff
@@ -618,7 +619,7 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
 
 			// If you're suiciding and unlucky, all the power just goes out the window and we don't trigger
 			var/mob/living/source = owner
-			if(source.suiciding && prob(60)) //Probably won't trigger on suicide though
+			if(!src.functional || source.suiciding && prob(60) && allow_suicide == FALSE) //Probably won't trigger on suicide though
```
Valtsu0, line 622:
```suggestion
			if(!src.functional || source.suiciding && prob(60) && !src.allow_suicide) //Probably won't trigger on suicide though
```

unfunnyperson:
> I absolutely love this. This will make a few people explode on death, but their cloning records will still be avaible, so they are not round removed. And the way this can be found out is interesting as well: no cloning implants means when scanned you are not able to view their health state on the cloner, which is quite a giveaway (extra points if you remove the module yourself to reuse it later ;) ).
> 
> Overall, i really like this and look forward to see a few people go pop. Only question would be why only robotocist and not geneticist as well? The cloner was genetics domain at some point, after all.

Not too sure if genetics should get it as they get a lot less leverage out of the item.

Valtsu0:
I heavily dislike how the cloner modules are coded
you should have a single variable like `var/obj/item/cloneModule/upgrade` and not repeat the same code 3 times

code/obj/machinery/computer/cloning.dm
```diff
@@ -183,14 +187,22 @@ ADMIN_INTERACT_PROCS(/obj/machinery/computer/cloning, proc/scan_someone, proc/cl
 		qdel(W)
 
 	else if (istype(W, /obj/item/cloneModule/minderaser))
-		if(allow_mind_erasure || allow_dead_scanning)
+		if(allow_mind_erasure || allow_dead_scanning || microbomb_upgrade)
```
ZeWaka, line 190: yeah this is pretty cringe, this should be refactored to just be handled on the module side

PR #17745, opened 2024-01-22
Rename the paint crate, and add the artistic toolbox to it.

TheTyrant25:
[CARGO]
## About the PR
Adds the artistic toolbox to the paint can crate. In addition, to name it more accurately, renames the paint supply crate to "artistic supplies crate". Also puts it in a packing crate instead of a normal metal crate.

## Why's this needed? 
It's good to be able to get an artistic toolbox if you need one.

## Changelog
```changelog
(u)Tyrant
(+)The Paint Cans crate is now called the Artistic Supplies crate, and comes with an artist's toolbox.
```


UrsulaMejor:
Artistic toolbox already comes with 6 random paintcans and a crayon. Is it intended for this to increase the total number of paintcans to 10? don't actually care, just wanted to mention it in case this was unintended

TheTyrant25:
it wasn't intended but I don't see it as too much of a problem

PR #17746, opened 2024-01-22
Makes screenheads smashable

TobleroneSwordfish:
## About the PR 
Screenheads can now be smashed when the borg is hit by an explosion or whacked in the head by a heavy object. They can be repaired with a short actionbar by clicking either the borg or head with a glass sheet. There's no mechanical effect to this, just a smashed overlay with occasional sparks.

## Why's this needed? 
It's quite a big fragile looking screen and feedback when things get hit is neat. Also adds to "maintenance" a cyborg might need to ask a roboticist to do.

imageimageimage

(I'm too lazy to make it a gif)

## Changelog 

```changelog
(u)LeahTheTech
(*)Cyborg screen heads can now be smashed by explosions and blunt instruments, repaired with glass sheets.
```


TheTyrant25:
Does this interact with the "repair damage" function of cyborg chargers in any way?

TobleroneSwordfish:
Not currently, I think it's fine as a separate thing though since the docking stations repair structural and wiring damage.

PR #17753, opened 2024-01-23
gives borgs the borg power hammer

Cherman0:
## About the PR 

Gives mining module borgs the borg mining hammer that's already in the game.

## Why's this needed? 

Since borgs only get the mining drill and cannot use explosive mining satchels, its quite common for some of the harder ores to just be impossible to break.

Potential balance concern is that the mining hammer is pretty good as a weapon. I dont feel that this is an issue as it is offset by borgs being generally quite weak in open combat. Any stuns pretty much completely nullifies them as a threat.

Could potentially have its damage or energy costs tweaked for balance reasons.

## Changelog 

```changelog
(u)Cherman0
(+)Mining module cyborgs now have the mining power hammer.
```

### DIFF
--- a/code/modules/robotics/robot/module_tool_creator/modules.dm
+++ b/code/modules/robotics/robot/module_tool_creator/modules.dm
@@ -174,6 +174,7 @@
 		/obj/item/device/gps,
 		/obj/item/extinguisher/large/cyborg,
 		/obj/item/mining_tool/drill,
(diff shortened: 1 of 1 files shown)


mordent-goonstation:
Declining as this makes the cyborg mining module too good from the get go, as well as giving them a force 20 melee weapon.

PR #17761, opened 2024-01-23
Lower genomes for high-genome plants

TDHooligan:
[HYDROPONICS] [QOL] [BALANCE]
## About the PR 
Moves genomes for the top 3 plants down:
Slurrypod 45 -> 40
Radweed 40 -> 34
Fungus 30 -> 31

## Why's this needed? 

The hops from citrus -> fungus and fungus -> radweed are relics from way, way back. Back when botany was RNG hell and you *liked* it. This makes the botany RNG moderately less miserable without entirely removing the moderately juicy genome hops you need to do for the radweed/slurrypod splice

## Changelog 

```changelog
(u)TDHooligan
(+)The RNG gods see botanys' plight. The top 3 genomes have, on average, dropped in value. 
```


NibChocolateeny:
any reason to making slurreypod 40? maybe 39, so sepration between radweed and slurreypod is still 5, like before

TDHooligan:
figure its probably safer to keep slurrypod a lil higher. its still a fairly safe splice

TDHooligan:
i thought the same thing but also thought that it requiring a dominant species of a plant, with splice enabler, purely for 10% chance at some progress towards this, sucked

Tarmunora:
> Not sure how i feel about making the jump from "normal" plants to slurrypod/radweed possible without splice enabler.
> 
> Then again, there are not many plants that would make sense to put into their own genome-category. And if there is one coming up, the genomes can always be increased again.

Agreed - I think there's value in having them be a "wait, is this even possible? Oh shit you *can* do that!" moment

Tarmunora (approved):

PR #17764, opened 2024-01-24
Announcement computer tgui

Valtsu0:
[UI]
## About the PR 
image
image
## Why's this needed? 
tgui good

## Changelog 

```changelog
(u)Valtsu0
(+)Announcement computers now use tgui
```


TobleroneSwordfish (changes requested):
The transmit button and message only update when you click away, meaning you can click a visually disabled button. Best to set the status of the button and the "input message" text on the frontend based on whether the textbox contains a message. See how `TextInputModal` does this for details (`useSharedState`)

Valtsu0:
Actually, i see of no reason why the message box should be shared between users. Made it local

Tarmunora (approved):
light theme checks out

Cherman0:
There is a maximum character limit to an announcement that is currently not visible in the UI, leading people to sometimes transmit cut-off messages. Would probably be good to add a "message length / max length" display so people can know about that.

Valtsu0:
> There is a maximum character limit to an announcement that is currently not visible in the UI, leading people to sometimes transmit cut-off messages. Would probably be good to add a "message length / max length" display so people can know about that.

image
well it was kinda in the pr, will fix that

TobleroneSwordfish (changes requested):
The previous interface let you input a message and then hit transmit later, this one clears the message if you close the interface and sends if you hit enter. I think being able to input a message to transmit later is useful QoL and hitting enter to send is mostly just going to result in accidental transmissions as people try to insert newlines.
Other than that, this looks great!

tgui/packages/tgui/interfaces/AnnouncementComputer/automatic.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Input, Section, Stack } from '../../components';
+import { AnnouncementCompData } from './data';
+
+export interface AutomaticAnnouncementData {
+  arrivalalert: string;
+}
+
+export const AutomaticAnnouncement = (_props, context) => {
+  const { act, data } = useBackend<AnnouncementCompData>(context);
+  // Extract `health` and `color` variables from the `data` object.
```
TobleroneSwordfish, line 18:
```suggestion
```

tgui/packages/tgui/interfaces/AnnouncementComputer/manual.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+type Status = {
+  canTransmit: boolean;
+  text: string;
+  color: string;
+};
+
+export const ManualAnnouncement = (_props, context) => {
+  const { act, data } = useBackend<AnnouncementCompData>(context);
+  // Extract `health` and `color` variables from the `data` object.
```
TobleroneSwordfish, line 21:
```suggestion
```

tgui/packages/tgui/interfaces/AnnouncementComputer/data.tsx
```diff
@@ -0,0 +1,16 @@
+/**
```
TobleroneSwordfish, line 1: This file should be called `type.ts` to be consistent with other interfaces.

TobleroneSwordfish (changes requested):
The UI won't register a card if it's inserted when the UI is closed, and re-opening the UI sets the caret position to the start of the text.

Valtsu0:
> re-opening the UI sets the caret position to the start of the text.

That seems to be an issue with tgui/inferno. Changing your character's name has the same issue.

clearing the message after transmitting should help a bit

TobleroneSwordfish (approved):

PR #17765, opened 2024-01-24
Kitchen Helper / Mixer TGUI Conversion

mloccy:
[CATERING] [UI]
## About the PR 
Converts the mixer UI into a TGUI UI which is an ongoing effort to modernize goonstation's UI technology. Also allows players to eject specific items from the mixer, rather than all or nothing.

## Why's this needed? 
TGUI allows for better user experience and is far more flexible and maintainable than the old system.

## Screenshots
image
image
image

## Changelog 

```changelog
(u)mloccy
(+)Kitchen Mixer now uses a TGUI interface
```


mloccy:
I'm not super familiar with React so I'm pretty eager to get some feedback on ways this could be improved.

code/obj/submachine/cooking.dm
```diff
@@ -1573,53 +1576,63 @@ TYPEINFO(/obj/submachine/mixer)
...
 
-	attack_ai(var/mob/user as mob)
-		return attack_hand(user)
+		switch (action)
+			if ("eject")
+				var/index = params["index"]
+				var/obj/item/target = src.contents[index]
+				src.ejectItemFromMixer(target)
+
+				usr.show_text(SPAN_NOTICE("You ejected the [target.name] from the [src]."))
```
TobleroneSwordfish, line 1618: These messages usually use present-tense, ie "You eject" rather than "You ejected" but also they're proobably not needed here since you know what you just clicked on.

code/obj/submachine/cooking.dm
```diff
@@ -1573,53 +1576,63 @@ TYPEINFO(/obj/submachine/mixer)
...
+		if(!ui)
+			ui = new(user, src, "MixerMachine")
+			ui.open()
+	ui_static_data(mob/user)
+		. = list("maxItems" = MIXER_MAX_CONTENTS)
+
+	ui_data(mob/user)
+		var/mixerContents = list()
+		var/index = 1
+		for(var/obj/item/_Item in src.contents)
```
TobleroneSwordfish, line 1592: What's with the `_Uppercase` naming convention? Normally we don't care too much about naming conventions, but `_` usually indicates something is unused so it's odd.

tgui/packages/tgui/interfaces/MixerMachine.js
```diff
@@ -0,0 +1,87 @@
...
+                {
+                  (items.length > 0)
+                    ? items.map(item => (<MixerItem key={item.index} mixerItem={item} working={data.working} />))
+                    :"No contents in mixer"
+                }
+              </Section>
+            }
+          </Box>
+          <Button
+            minWidth="48%"
```
TobleroneSwordfish, line 64: This is not a good way of aligning buttons, consider using a `Stack` instead (will probably also help fix them not being the same height)

tgui/packages/tgui/interfaces/MixerMachine.js
```diff
@@ -0,0 +1,87 @@
...
+import { Window } from '../layouts';
+
+export const MixerItem = (props, context) => {
+  const { act } = useBackend(context);
+
+  const { mixerItem, working } = props;
+
+  return (
+    <Flex>
+      <Flex.Item nowrap key={mixerItem.name}>
```
TobleroneSwordfish, line 19: Not sure what the purpose of this `Flex` with a single `Flex.Item` is? Flex is a layout component so is used to arrange multiple `Flex.Item` children.

TobleroneSwordfish (changes requested):
It's probably arguable whether this really needs an interface, but if you want to be fancy you could look into putting the icons of the content items on the UI. See `icon2base64`.

mloccy:
Updated screenshots after review
image
image
image

code/obj/submachine/cooking.dm
mloccy: Fixed

code/obj/submachine/cooking.dm
mloccy: Yeah, not sure what I was thinking, renamed the variable

tgui/packages/tgui/interfaces/MixerMachine.js
mloccy: Changed to stack, and left the buttons to not fill like they were before, just stack normally.

tgui/packages/tgui/interfaces/MixerMachine.js
mloccy: I was trying things, this was a remenent of something that didn't work, removed.

Tarmunora (approved):
no crimes against lightmode users here 👍

Valtsu0:
Is ntos the correct theme here? I'd use ntos on things that are actually in character guis. I'd use the default

mloccy:
> Is ntos the correct theme here? I use ntos on things that are actually in character guis. But then again the id computer uses default so it seems completely arbitrary
> 
> Anything with dials at least seems to always use the default

Happy to change it to default, I was following the Espresso Machine as an example, so I didn't think there was a choice to be made there, are there any other themes?

Valtsu0:
> Happy to change it to default, I was following the Espresso Machine as an example, so I didn't think there was a choice to be made there, are there any other themes?

All of the themes can be found at `\goonstation\tgui\packages\tgui\styles\themes`. note that some of them, like abductor, do not seem to work

ZeWaka:
Just use default

code/obj/submachine/cooking.dm
```diff
@@ -1573,53 +1576,82 @@ TYPEINFO(/obj/submachine/mixer)
...
+				src.ejectItemFromMixer(target)
+
+				usr.show_text(SPAN_NOTICE("You eject the [target.name] from the [src]."))
+				. = TRUE
+
+			if ("mix")
+				src.mix()
+				. = TRUE
+
+			if ("ejectAll")
```
TobleroneSwordfish, line 1639: This should also return true to update the UI

code/obj/submachine/cooking.dm
```diff
@@ -1573,53 +1576,82 @@ TYPEINFO(/obj/submachine/mixer)
...
+				for (var/obj/item/target in src.contents)
+					src.ejectItemFromMixer(target)
+
+				usr.show_text(SPAN_NOTICE("You eject all contents from the [src]."))
+		if (.)
+			tgui_process.update_uis(src)
```
TobleroneSwordfish, line 1645:
```suggestion
```
Returning true from this function already updates the UI, you don't need to do it explicitly

TobleroneSwordfish (changes requested):
The font sizes and overall UI scale are just a little bit big, see here it compared to the chem dispenser UI: 
image

mloccy:
image

Re-scaled UI with other currently implemented UIs to compare (chem dispensor, expresso machine)

code/obj/submachine/cooking.dm
mloccy: Removed

PR #17767, opened 2024-01-25
Adds viscorgans to the medical fabricator

ThatFiveGuys:
## About the PR 
This pr replaces the cyberorgans in the medical fabricator with viscorgans, organs made out of viscerite. These are functionally similar to normal organs except for the fact they have 20 less health. This also gives the medical fabricator 3 viscerite (and in order to do so ups every other mat to 3) so they can print out roughly 6 roundstart.
image

## Why's this needed? 
This makes cyberorgans unique to robotics manufacturers, which should hopefully help with giving roboticists stuff to do as you have to go to them for organs that enchance. Oh also gives a use to viscerite and just seems thematically neater.

## Changelog 

```changelog
(u)Ikea
(*)The cyberorgans in the medical fabricator have been replaced with cheaper but worse viscorgans. 
```


Tarmunora:
if we want to make cyberorgans a robotics-fab-only thing, then we'll just say "medbay can butcher monkeys for parts", not add another set of functionally identical organs to the code

hard pass from me on this one

BatElite:
rapid-fire monkey liver theft sounds like a grand time actually 

(if only the bastards could be subdued quickly)

I like the sprites, but yeah the meat of the PR isn't much

PR #17774, opened 2024-01-26
Map helpers for technical storage cards

pgmzeta:
[qol][internal][mapping]
## About the PR 

Adds a new type of rack meant to assist mappers: `/obj/rack/organized` that will spawn a list of items onto the rack in an aesthetically pleasing way. Creates two children of this new rack type to spawn the technical storage circuit boards on racks.

Built with some variables to ensure they can be controlled by mappers. Could be extended to e.g. space suit or other common tool racks with pixel-editing.

This does not add the helper to maps; to be done on a follow-up PR to minimize drift/size.

## Why's this needed? 
* easier to maintain map consistency (once implemented to maps)
* more easily see there are multiple cards if they're spread out

## How does it look?
Looking at the racks of circuit boards in tech storage across a few maps to compare.
Cogmap1
image
new map helpers on right, cogmap1 defaults bottom.

Donut3
Screenshot 2024-01-25 150513
New map helpers on right, donut3 defaults left.

Nadir
Screenshot 2024-01-25 144346
Map helpers left, nadir defaults right.

Kondaru
Screenshot 2024-01-25 144003
Map helpers top, kondaru defaults bottom.


ZeWaka:
ok

PR #17782, opened 2024-01-26
Atlas drinkslide fix

MintyPhresh:
[catering][player-actions][bug][
## About the PR
Adds a new variable to tables called "drinkslideable" (default true) that, if false, will not allow drinkslides to start from that table. Implements this on Atlas, switching the back two tables to be non-drinkslideable.

## Why's this needed?
Drinksliding from the east side of atlas bar to the west would cause the player to begin drinksliding from the back table, meaning that the glass would attempt to cross the gap between the back and front of the bar (and consequently shatter on the ground). This mapping tool should let mappers make more compact bars without worrying about this silly bug.

### DIFF
--- a/code/modules/chemistry/tools/food_and_drink.dm
+++ b/code/modules/chemistry/tools/food_and_drink.dm
@@ -1440,7 +1440,7 @@ ADMIN_INTERACT_PROCS(/obj/item/reagent_containers/food/drinks/drinkingglass, pro
 		if(isnull(source_table))
 			for(var/dir in cardinal)
 				source_table = locate() in get_step(user, dir)
-				if(!isnull(source_table))
+				if(!isnull(source_table) && source_table.drinkslideable)
 					user.set_dir(dir)
 					break
 		if(isnull(source_table))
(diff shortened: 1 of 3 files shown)


TobleroneSwordfish (changes requested):
This is a bad way of fixing this, it won't actually make it work and adds needless mapping effort instead of fixing the underlying issue. A better way would be to rearrange the call order so the proc is called before the auto-rotate portion of the click stack, or pass the original direction down the stack as a parameter.

MintyPhresh:
I think this style of fix has merit even if the rotation issue is fixed (working on that now); I can't think of any other interactions in the game where the direction you're facing before you click on something changes the outcome of the action. Bartenders on Atlas face north automatically when they use the espresso and soda machines; unless they're familiar with exactly how this code works, they'll still run into the issue of the drinkslide starting on an unintended table.

Tarmunora:
superseded by https://github.com/goonstation/goonstation/commit/fee309fea2ab09bc038de9ea422e1db0914e972f

PR #17791, opened 2024-01-27
Add some special procs for bulk artifact spawning and (de)/activation

MeggalBozale:
## About the PR 
Adds two admin-only procs, Toggle-All-Artifacts and Bulk-Spawn-Artifacts.
Toggle-All-Artifacts allows you to either turn on all existing artifacts, or turn them all off. (Some may instantly re-activate anyway, though.)
Bulk-Spawn-Artifacts lets you spawn artifacts similarly to a radiation blowout, but with control over the amount of artifacts spawned. Also lets you choose to allow artifacts to spawn in space.

## Why's this needed? 
May be helpful for testing, and hopefully can be fun to use on live if someone wants to. Mostly, I was surprised that these procs didn't exist already when I went to look for them in a local.

Feel free to close if this is simply not wanted, the primary reason I'm creating a PR is because I made this for fun and want to see if it's worth keeping around.

https://github.com/goonstation/goonstation/assets/27376947/b82556c1-26af-4cb6-adc6-46c400c48815

## Changelog 
Doesn't really need one, since it doesn't impact regular players directly

### DIFF
--- a/code/modules/admin/randomverbs.dm
+++ b/code/modules/admin/randomverbs.dm
@@ -3014,6 +3014,10 @@ var/global/force_radio_maptext = FALSE
 				return
 			if("Yes")
 				var/amt_artifacts = min(input(usr,"How many artifacts do we want to spawn?\n(Default: 50)",,50) as num, ARTIFACT_HARD_LIMIT)
+				var/response = input(src, "High number of types: [length(spawn_matches)] - Type YES to continue", "Caution!") as null|text
+				if (response != "YES")
+					message_admins("[key_name(usr)] aborted spawning [amt_artifacts] due to failing to type 'YES'") // im telling on u!!!
+					return
 				var/floors_only = (alert(src, "Only spawn artifacts on floors?",, "Yes", "No") == "Yes")
 				logTheThing(LOG_ADMIN, src, "started spawning [amt_artifacts] artifacts.")
 				message_admins("[key_name(usr)] started spawning [amt_artifacts] artifacts.")
--- a/code/modules/admin/randomverbs.dm
+++ b/code/modules/admin/randomverbs.dm
@@ -2968,3 +2968,60 @@ var/global/force_radio_maptext = FALSE
 				return
 	else
 		boutput(src, "You must be at least an Administrator to use this command.")
+/client/proc/toggle_all_artifacts()
+	// Housekeeping
+	SET_ADMIN_CAT(ADMIN_CAT_FUN)
+	set name = "Toggle All Artifacts"
+	if (holder && src.holder.level >= LEVEL_ADMIN)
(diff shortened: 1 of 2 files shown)


Tarmunora:
superseded by  https://github.com/goonstation/goonstation/commit/fee309fea2ab09bc038de9ea422e1db0914e972f

Tarmunora:
wait wrong PR fuck

code/modules/admin/randomverbs.dm
```diff
@@ -2968,3 +2968,60 @@ var/global/force_radio_maptext = FALSE
...
+				var/artifacts_spawned = 0
+				var/amt_artifacts = input(usr,"How many artifacts do we want to spawn?\n(Default: 50)",,50) as num
+				amt_artifacts = min(amt_artifacts, artifact_hard_limit)
+				var/floors_only = (alert(src, "Only spawn artifacts on floors?",, "Yes", "No") == "Yes")
+				logTheThing(LOG_ADMIN, src, "started spawning [amt_artifacts] artifacts.")
+				message_admins("[key_name(usr)] started spawning [amt_artifacts] artifacts.")
+				while (artifacts_spawned < amt_artifacts)
+					var/turf/T = locate(rand(1, world.maxx), rand(1, world.maxy), Z_LEVEL_STATION)
+					if (floors_only && !istype(T, /turf/simulated/floor))
+						continue
```
Tarmunora, line 3020: would still spawn stuff inside windows/doors/etc.
not necessarily an issue, just something to consider

MeggalBozale: yeah thats fair and something observed in the clip, its worth changing if its going to see use. fwiw radiation blowouts also do that (and operate off a 0.5% chance per tile to spawn an artifact, going through every turf on station ever) and might be worth changing as well? though it mostly works as is and the probability based approach scales amounts of artifacts to map size which is cool

yeah fixing that though

MeggalBozale: pushed commit to resolve

MeggalBozale:
artifacts no longer spawn on dense turfs, added some defines for scope of proc to make more readable, and some additional logging if any artifacts managed to fail to spawn

Tarmunora (approved):

flappybatpal:
Nitpick but I prefer forcing admins to confirm rather than just click yes for anything that could crash the server (after all this did bluerscreen your computer in testing...)
https://github.com/goonstation/goonstation/blob/35833c9222ded73608a3264a52c73821a93f6f0e/code/modules/admin/admin_verbs.dm#L2553

MeggalBozale:
fair enough ill see what I can do for more verification, also TBF I introduced instability in my system trying to overclock my CPU and it was generally unstable until I reinstalled windows. So more of a fault with my computer than byond

PR #17795, opened 2024-01-28
Nadir Cafeteria Changes

MintyPhresh:
### Sets the tables!
image
(ignore the missing knife, that was Mr. Muggles)
Chairs on north and south ends of tables have been placed in a nearby crate. And yes, that's fresh ice water in those glasses! Yum!

### Remy-proofing and tons of counter space!
image
Clever table placement maximizes tabletop space and prevents that DAMN RAT from blocking the door with the beer keg!

### Fancy new Bar Dispenser!
image
Tending bar has never been easier with this combined alcohol + soda dispenser!

## Why's this needed?
Nadir's cafeteria is uniquely refined and clean-feeling compared to others. The changes proposed here expand on that theme and make the map cooler.


MintyPhresh:
I can partially see the issue, there-- taking into account RNG from random spawners, this does almost double the amount of items on the table... What's the issue you have with the cafeteria clutter that's currently there? Is it just that it's disorganized, the visual clutter, or the fact that it eats up useable space on the table? Could your concerns be solved by reducing the number of set seats?

MintyPhresh:
Don't think I can remove those condiment things in good faith, since every map has them, but I was thinking the same thing about the knives. Will be gone once I fix this merge conflict.

MintyPhresh:
Here's how it looks now.
image

Cherman0:
I'd probably unreinforce the tables too. Makes it a bit easier for the cafeteria staff to rearrange them if they want without having to drag an engineer over. And maybe use wooden chairs instead of steel folding chairs. Overall looks good though.

MintyPhresh:
Reinforcing the table has merit, but it's part of the area's aesthetic, in my opinion-- same with the metal chairs. Those two features make Nadir really unique, IMO.

MintyPhresh:
Southeast spoon on the two tables has been fixed, should be no asymmetry issues now

code/obj/item/kitchen.dm
```diff
@@ -54,8 +54,8 @@ TRAYS
 
 	New()
 		..()
-		if(prob(60))
-			src.pixel_y = rand(0, 4)
+//		if(prob(60)) // This was causing issues with placing stuff for setting tables.
+//			src.pixel_y = rand(0, 4)
```
pgmzeta, line 58:
```suggestion
		if(src.pixel_y == 0 && prob(60)) // Don't adjust map-set pixel adjustments
			src.pixel_y = rand(0, 4)
```

code/modules/chemistry/tools/food_and_drink.dm
```diff
@@ -1754,6 +1754,14 @@ ADMIN_INTERACT_PROCS(/obj/item/reagent_containers/food/drinks/drinkingglass, pro
 					 "mojito", "cremedementhe", "grasshopper", "freeze", "limeade", "juice_peach",
 					 "juice_banana")
 
+/obj/item/reagent_containers/food/drinks/drinkingglass/icewater
+	New()
+		..()
+		SPAWN(0)
+			if (src.reagents)
+				src.reagents.add_reagent("ice", 15, null, T0C)
+				src.reagents.add_reagent("water", 35, null, T0C)
```
Kubius, line 1763: Super minor nitpick, but the water being at 0 celsius is weird.

Kubius (approved):

code/modules/chemistry/tools/food_and_drink.dm
MintyPhresh: Ice got changed recently, and ice and water can now coexist with a little cajoling (have to both be near freezing before mixing)
Testing showed that it remained as both ice and water after spawning in. Liquid in the screenshots above would be a lot whiter if it was all ice

PR #17798, opened 2024-01-28
Enable logging runtimes during global init

amylizzle:
## About the PR 
It turns out `/var/global/list/A = list()` is **not equivalent** to `/var/global/list/A = new()`
AS a result, errors occurring during global init were being sent to `/world/Error()` and then failing to log because `runtimeDetails` was in a not-quite-initialised and not-actually-null state (thanks BYOND!)

## Why's this needed? 
1) Makes it so we can log runtimes during init
2) Makes it so OD can get through init without throwing runtimes because things weren't initialised properly before they were used.

### DIFF
--- a/code/datum.dm
+++ b/code/datum.dm
@@ -2,8 +2,8 @@ TYPEINFO(/datum)
 	var/admin_spawnable = TRUE
 
 #ifdef IMAGE_DEL_DEBUG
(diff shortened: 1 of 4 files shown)


TobleroneSwordfish (approved):
⚠️ Cognitohazardous byond behaviour⚠️

PR #17800, opened 2024-01-28
Add knitting supplies to ranching lockers and wool production supplies

TheTyrant25:
[QOL]
## About the PR
See title

## Why's this needed?
They're difficult to re-obtain if you lose the original set ~~to a clown~~

## Changelog
```changelog
(u)Tyrant
(+)Knitting supply boxes can now be found within ranching secure lockers and inside a wool production kit, in case you lose yours.
```


UrsulaMejor (approved):
Yeah

PR #17809, opened 2024-01-29
Change the way tanks explode to be more intuitive

MeggalBozale:
## About the PR 
This change is part of a larger rework focused at making the Toxins department more accessible, more challenging at higher levels, and about adding more interesting and fun mechanics in Toxins for scientists to play with. For more information, see:
[Player Projects - Tank Explosion Rework](https://discord.com/channels/182249960895545344/1199023492503445504)
[HackMD Document - Tank Explosion Rework](https://hackmd.io/@Romayne/S1tTjW2ta)

The current handling of tank explosions is closer to being an ancient bug than a feature -- allow me to explain as I have in the HackMD:
A tank will immediately explode as soon as it surpasses a certain pressure, ~5066.25kPa.
When a tank explodes, it calls react() 3 more times to build up pressure, simulating the uncontrolled reaction of a real explosion.
The final explosion power is determined by the final pressure.

This gives rise to an issue where tanks which approach this threshold from either end will have drastically different explosion ranges. For example, consider a TTV with two tanks. The first tank is full of oxygen (1013.25 kPa) which is at a temperature of 200 degrees Kelvin. The second tank is full of plasma (1013.25 kPa) at a temperature of 599 degrees Kelvin. As it reacts, it eventually reaches a pressure of ~5060 kPa. It reacts again, surpassing the treshold, and then 3 more times. By reaching a pressure just below the explosion treshold of the container, it 'sneaks in an extra react'.

Consider the same TTV, but with the tank of plasma at 600 degrees Kelvin. As it reacts, it eventually reaches a pressure of ~5067 kPa. It reacts only 3 more times, and explodes. Comparatively, these near-identical mixes have drastically different explosion powers, because one tank is given the opportunity to build pressure an additional time.

To fix this, reactions are given a multiplier variable, which may be used to react a fractional amount of times. This is done by assuming reactions are linear (they mostly are, with minor exceptions) and providing a multiplier which acts as if such is true. It is determined on explosion of a tank as to when it exceeded the tank's maximum pressure, and the total amount of additional reacts before exploding is adjusted appropriately. This results in a more reasonable trend of explosion ranges, which is less likely to confuse players.

To show this graphically, here is the current range of explosion values. Blue is the least explosive (~0), while red is the most explosive (~12, 'maxcap'):
temperatures

And here is the same graph after the change:
temperatures

Note that after this change, tanks get four additional reacts, not 3 -- this is to keep roughly the same range of mixes possible, as this change is not intended to affect balance significantly.

## Why's this needed? 

It is important that game systems behave as expected, and I would argue the current explosion behavior of tanks does not act in an intuitive manner. It forces code-diving to understand the pattern, and is likely unpleasant for new scientists to work with. 
It should also be noted that this only could change how difficult it is to 'discover' a mix that creates a maximum range explosion for a TTV, experienced players have no problem memorizing and sticking to a mix of temperatures which will reliably produce that large explosion.

Lastly, this helps with the newly(?) added bazaar, which demands precise explosion values for maximum profit. With the current lack of predictability, it becomes hard for players without direct access to every possible value beforehand to make use of this mechanic, due to the sudden changes in explosion range and odd trends.

## Changelog 

```changelog
(u)MeggalBozale
(*)Tanks will now explode in a more predictable manner. For an in-depth explanation, see the PR.
```


MeggalBozale:
made some oopsies still, hold on

MeggalBozale:
Okay it's ready, sorry for the extensive unrelated commits it was a part of another branch that I cleaned up to include only this change. Gonna see about cleaning that up for future PRs related to this

ZeWaka:
yes

TobleroneSwordfish (commented):
This looks great, the only thing I will say is that it seems like it's widening the maxcap range a bit in addition to making it more consistent. Possibly worth a numbers tweak to make it a little harder?

TDHooligan:
could always set the extra reacts from 4 to 3.5 to get the average of the old math

MeggalBozale:
Easily can set to 3.5, yes. Will post with updated graph when done

MeggalBozale:
temperatures
New temps

MeggalBozale:
Changing scaling is simple enough so LMK if thats too small, new graph is 3.5 reacts minimum rather than 4

TDHooligan:
thats probably ok. could maybe be a little bigger but its not my decision 🐸

MeggalBozale:
I think it's a good trade-off for being more predictable. Having the same range of oxygen values would be the four reacts in the original PR. I'll check out 0.75 in a moment, profiling radgas for now

PR #17813, opened 2024-01-30
Some radioactive materials decay plasma into fallout

MeggalBozale:
## About the PR
Allows materials, namely Cerenkite, to decay fallout on their same tile into fallout gas above a certain temperature threshold.
The radioactive material will begin to lose its radioactive properties as it is used for this purpose

dreamseeker_6q3Lge6a4W
*an example of an experimental setup I created in the development of this feature*

## Why's this needed?

Allows Toxins to create fallout gas for their own purposes, which may not be possible to obtain on some maps. It holds some value as having an abnormally low SHC, and thus may have a use in the environment. I plan to also give this gas an interesting use, also located in toxins, for scientists to desire to use this for.

Numbers are roughly an attempt at balance and allows a handful of cerenkite ores to convert a canister of plasma. Conversion is largely limited by how quickly the conversion cools existing gas, and such the bottleneck in production becomes heat transfer with the plasma gas.

Scientists are given the choice between creating a sealed chamberburn for an output mix containing low percentages of fallout at a high conversion rate, or using heat transfer between chambers to produce a more """pure""" final product. 

Overall allows a lot of experimentation and choice into the methods of production, which I think is sorely needed for Toxins.

## Changelog

```changelog
(u)MeggalBozale
(*)Radioactive material at sufficient temperature will now decay some plasma into fallout gas.
```


MeggalBozale:
Worth noting that this would allow "Cold TTVs" or so I believe to be made in toxins.
AFAIK this refers to combining a low SHC, cold gas (fallout, namely, being a prime candidate here) with a high SHC, hot gas (agent b, plasma, or whatnot being candiate here) together to potentially create a maxcap in an alternate way. I would be for this since it's interesting and more difficult than current TTVs, so.

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -738,3 +738,54 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
...
+				bulkness_multiplier = I.amount
+				if (I.amount <= 0)
+					return // I just have a feeling this will manage to happen and fuck things up
+			if (!O.material.hasProperty("radioactive"))
+				return
+			var/radioactivity = O.material.getProperty("radioactive")
+			var/n_radioactivity = O.material.getProperty("n_radioactive")
+			if (O.material.hasProperty("n_radioactive"))
+				n_radioactivity = O.material.getProperty("n_radioactive")
+			var/turf/T = locate(O.x, O.y, O.z)
```
TobleroneSwordfish, line 761:
```suggestion
			var/turf/T = get_turf(O)
```
Unless you're meaning this to fail inside containers, in which case `O.loc` would better I think.

code/modules/materials/Mat_Materials.dm
```diff
@@ -669,6 +669,7 @@ ABSTRACT_TYPE(/datum/material/metal)
 		setProperty("electrical", 6)
 		setProperty("radioactive", 5)
 		setProperty("hard", 2)
+		addTrigger(TRIGGERS_ON_TEMP, new /datum/materialProc/radioactive_temp())
```
TobleroneSwordfish, line 672: Maybe consider adding this hackily to `setProperty` so any future radioactive materials get it automatically?

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -738,3 +738,54 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
...
+			if (!issimulatedturf(T))
+				return
+			var/turf/simulated/simTurf = T
+			if (!simTurf.air)
+				return
+			if (simTurf.air.toxins > MINIMUM_REACT_QUANTITY)
+				/// Mostly bullshit magic because I don't know how radiation works and plasma isn't real, but is how many moles to convert of existing plasma
+				var/moles_to_convert = (bulkness_multiplier * (1 + radioactivity) * (1 + n_radioactivity) * sqrt(temp)) / 1000
+				if (moles_to_convert > simTurf.air.toxins)
+					simTurf.air.radgas += simTurf.air.toxins
```
TobleroneSwordfish, line 771: This needs to handle group processing somehow or it will do Weird Things, see how plasmastone works.

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -738,3 +738,54 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
 		var/list/color = rgb2num(owner.material.getColor())
 		light_c = owner.AddComponent(/datum/component/loctargeting/sm_light, color[1], color[2], color[3], 255 * 0.33)
 		light_c.update(1)
+
+/datum/materialProc/radioactive_temp
+	max_generations = -1
+
+	execute(var/atom/owner, var/temp)
+		if (isobj(owner) && temp > 500 KELVIN)
```
TobleroneSwordfish, line 746: Nitpick but these kind of conditions are better as guards, ie `if (!condition) return`, leads to less indent chaos

TobleroneSwordfish (changes requested):
Neat! Make sure it doesn't happen in the reactor itself, since that has its own way of handling this and would the two would probably not mix well.

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -738,3 +738,54 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
...
+			if (isitem(O))
+				var/obj/item/I = O
+				bulkness_multiplier = I.amount
+				if (I.amount <= 0)
+					return // I just have a feeling this will manage to happen and fuck things up
+			if (!O.material.hasProperty("radioactive"))
+				return
+			var/radioactivity = O.material.getProperty("radioactive")
+			var/n_radioactivity = O.material.getProperty("n_radioactive")
+			if (O.material.hasProperty("n_radioactive"))
```
amylizzle, line 759: redundant check, you already set the value, and it's 0 if it doesn't have the property.

code/modules/materials/Mat_MaterialProcs.dm
amylizzle: You probably shouldn't be doing it like this at all. See the molitz temp react for an example. Basically you want `owner.return_air()` instead of this.

amylizzle (changes requested):
This definitely needs special handling for the reactor. Personally I'm not sure this is a good thing to add at all - fallout is hard to get on purpose because it's dangerous and expensive cpu-wise. 

An alternate approach would be a way of generating neutrons outside of the reactor, then you can rely on existing gas reaction code, but again, I'm not sure fallout ought to be accessible.

MeggalBozale:
Thank you for all the feedback!! I will work on this asap.
> Personally I'm not sure this is a good thing to add at all - fallout is hard to get on purpose because it's dangerous and expensive cpu-wise.
This should be somewhat difficult to get in large quantities from toxins, it's meant to be another option if there's no reactor or barrels around. As it is, it's possible to get 100 moles of radgas from any barrel of radioactive waste, which is not too bad to get on maps containing reactors. (And w/ a random room containing a *lot* of barrels). Radgas being dangerous is part of why I think it would be nice to make in toxins, as it could be interesting to use in explosives. (I'd like to iron out an idea like that and its practicality first, though.)

Not sure what I can do for it being CPU-intensive except for noting that this is, ideally, difficult to obtain in large quantities quickly using toxins. I can see if I can make it a little better, but I imagine it's just because radgas does so much already.

code/modules/materials/Mat_Materials.dm
Tarmunora: `onAdded()`

MeggalBozale:
This will proc in reactors due to the trigger, but it will return for owner.loc not being a simulated turf. Reactor should be fine

MeggalBozale:
A word on the balance, currently - the "radioactive" property has a 1:1 potential as mols of fallout. So one piece of cerenkite only would decay into 5 mols of fallout, though I can easily change this to be higher/lower depending on perceived need for balance. I've not noticed too much harm from releasing a can of fallout, and doing so should take significantly more effort than plasma -- which is *much* more CPU-intensive than fallout is right now. And just to be sure, all materials now have a 5 second cooldown before decaying fallout (it was getting called a ***lot*** for a stack of 100)

amylizzle (approved):
I still slightly question the balance of making fallout easily available, but code looks good, and you can always revert it if people are shitters with it.

ZeWaka:
x

ZeWaka:
i haven't heard of any problems with this

MeggalBozale:
Neither have I - but then again I haven't seen much feedback on it at all. (It's a pretty niche and presently pointless change, given the lack of use for the produced fallout given the required effort.)

MeggalBozale:
with the buffed numbers it seems we've got one player on the discord messing around with the feature and doing good, considering the infrequency of players messing with toxins thats good to me

PR #17815, opened 2024-01-30
Mining tool refactor

Cherman0:
## About the PR 

Internally refactors mining tools, adding /obj/item/mining_tool/powered as a parent object to powered mining tools. Does some code cleanup and makes some minor noticeable changes.

Some balance tweaks were made in the process of the refactor, listed below.

- The laser drill is now a proper powered mining tool with a battery and stuff. In keeping with its upside of not having power issues, it uses 1.2 power per hit compared to the pickaxe's 2 and the hammer's 3. 
- The effective multiplier for tool power usage when used by a borg was reduced from 13.2x to 5x. Previously this was never relevant as the only mining tool they could use (laser drill) required no power.
- The mining hammer was made more expensive to craft and the laser drill was made less expensive to craft. Previously, the hammer required at least steel and copper (ie immediately craftable) and the drill required at least bohrum and claretine. The hammer now requires at least mauxite and the drill no longer requires at least bohrum.
- The concussive mining gauntlets also had their mining power reduced from 4 to 3. It seemed a bit silly for them to be the strongest mining tool (the hammer has 3) while also not using power. I'd like to look into having them require power in the future but that seemed a bit beyond the scope of a refactor (and also complicated).

## Why's this needed? 

The current implementation of powered mining tools is just a normal mining tool that changes its own stats in attack_self(). I was looking to create power mining tool upgrade modules in a future PR and the lack of a powered mining tool parent made that more difficult than it needs to be.

This PR does not touch matsci mining tools at all. Those are an entirely separate beast, and I'm of the opinion that they should be removed entirely (in favor of progressing tiers of fabricable mining tools), but that's well outside of the scope here.

## Changelog 

```
(u)Cherman0
(+)Laser mining drill now uses power and can be turned on and off.
(+)The power mining hammer is harder to craft and the laser mining drill is easier to craft.
(+)Concussive mining gauntlets had their mining power reduced from 4 to 3.
```


ZeWaka:
no changelog?

Cherman0:
It didnt really seem pertinent since most of the changes are internal. I suppose I can add one though

Tarmunora (changes requested):
you've changed the power costs on mining tools to be absurdly high for actually being useful for mining - previously, using mining tools as a weapon (hitting non-turfs) had a large power cost, but this PR puts that power cost onto mining as well

Cherman0:
I found my mistake. I assumed that the process_charges() in afterattack() was always being used to handle discharging the battery after hitting something, but actually in the original version it specifically checked to make sure you werent hitting a turf because asteroids separately implemented the exact same code. And we wouldn't want to run it twice, now would we.

(Would definitely like to do asteroid rework and other mining reworks at some point too since theres more weirdness there)

Fixed by making hitting asteroids handle discharging the battery the same way as everything else, and multiplying by a variable (defined on the tool) for anything that isn't a turf. Default value for that variable is 5 and power costs were reduced by 5x to compensate. In the process, also fixed that power was being used twice when you mined the nadir decorative greebles, as they incorrectly also separately implemented discharging the battery.

Other fixes: Held icon now properly depowers when you run out of charge, and the hedron beam examine text and power usage is now correct.

code/obj/mining.dm
```diff
@@ -1804,7 +1808,7 @@ TYPEINFO(/turf/simulated/floor/plating/airless/asteroid)
 	powered_force = 14
 	dig_strength = 2
 	powered_dig_strength = 3
-	power_usage = 6
+	power_usage = 1.2
```
Tarmunora, line 1811: eh, might as well round down to 1 imo

Tarmunora (approved):

code/obj/mining.dm
Cherman0: sure

PR #17816, opened 2024-01-30
API 2

Wirewraith:
## About the PR 

This replaces the existing external API communication system with a new version, built from the ground up to be faster and more scale-able than the old one.


TheTyrant25:
awesome

ZeWaka:
bonus points: this one is also open source

PR #17827, opened 2024-02-01
Rename central maint APC in clarion

TheTyrant25:
[MAPPING] [BUG]
## About the PR
the central maintenance APC was falsely var edited to be called "armoury APC", when it in fact controlled central maintenance. This removes the var editing of the name.

## Why's this needed?
Bugs bad. There shouldn't be two armoury APCs, that's misleading

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -20736,12 +20736,6 @@
 /turf/simulated/floor/white,
 /area/station/medical/medbay/lobby)
 "eJJ" = (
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (changes requested):
I think it'd make more sense to put it actually in maint rather than this weird bridge corridor locked behind command access

TheTyrant25:
i'll make a new PR doing that then

PR #17828, opened 2024-02-01
add external cycling airlocks to cogmap 1

TheTyrant25:
[MAPPING] [QOL]
## About the PR
Adds the new cycling airlock mechanism from  #17459 to cogmap1. Spots where they are used are available in mapdiffbot.
To those unfamiliar, all it means is that attempting to open an "outer" door will try to close the "inner door". It only tries it once, however, and is overridden by safety checks. It won't lock you inside, just gently tries to close the other door behind you. These are best applied to setups where two doors separate space and interior.

These were added by using airlock cycle map helpers. I also var edited their names, not just their IDs, so that the strongdmm prefabs tab is more legible.

They have not been added to every single possible place that could use them, notably shuttle docking bays, because they would simply impede foot traffic (and aren't supposed to open out into space anyway). Similarly they're not applied to "departmental airlocks" like in engineering, because it's not... locking air.

This also only adds it to cogmap1 for now. If it's not strongly hated then I'll PR it to other maps.

P.S. This also fixes a pretty critical bug with how cycling airlocks works, ported from coolstation commit d195ae114bca61ec373923b137a323545f929c06

## Why's this needed?
It's nice to use them, and should stop air leaking from careless door opening. Seems like something that should have already existed.

## Changelog
```changelog
(u)Tyrant
(+)Some airlocks are now linked in a way that opening one door tries to close the other doors. "Keep the Air Inside!" ~John Nanotrasen probably.
```

### DIFF
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -1666,6 +1666,8 @@ About the new airlock wires panel:
 /// adds the airlock in question to the global list.
 /obj/machinery/door/airlock/proc/attempt_cycle_link()
 	if (src.cycle_id)
+		if(!cycling_airlocks[src.cycle_id])	// add a list to the list of lists
+			cycling_airlocks[src.cycle_id] = list()
 		if (!(src in cycling_airlocks[src.cycle_id]))
 			cycling_airlocks[src.cycle_id] += src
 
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish (approved):
Very cool 👍

TheTyrant25:
there's also been talk of adding them to interior airlocks but that is out of scope for now. example places would be entrances to research and engineering.

PR #17833, opened 2024-02-02
Individual TGUI Windows for Tickets/Fines & Inspector Report

pgmzeta:
[UI]
## About the PR 
This atomizes the TGUI additions out of #16410 into their own windows, and polishes the associated feature.

The open notes on #16410 have been addressed:
* Tickets/Fines are kept as a separate window, so there are no user preference changes
* Send the individual reason/issuer/etc. components of the ticket/fine to TGUI so user inputted text is not dangerously rendered as HTML

In addition, new buttons have been added to the UI to allow users to open the tickets/fines and inspector's report again ala the Crew Credits button. These use the yellowed ticket paper and inspector's clipboard sprite respectively.

Screenshot 2024-02-01 163435

## Why's this needed? 
Allow more end-game dialogs to be reopened.

## Test Plan
- [ ] End game with no inspector, or nothing in the insepctor's clipboard, should not show inspector button or pop up inspector's report
- [ ] End game with one inspector's clipboard with paper shows end-game button & popup
- [ ] End game with two inspector's clipboards with paper show end-game button, pop-up, and the clipboards are treated distinctly
- [ ] Filing multiple tickets/fines for the same person should group them in the end-screen page
- [ ] Track & indicate when no tickets / no fines are recorded in a round

## Changelog 

```changelog
(u)glowbold
(*)The end-game tickets/fines and the inspector's report use a new dialog and can be re-opened if closed. Please report any bugs.
```


pgmzeta:
Needs a bit more local testing before I feel confident running this on live, but the core here should be solid.

TobleroneSwordfish:
We're aiming to reduce the amount of round-end window clutter, so these would be better as tabs on the main window I think.

pgmzeta:
> We're aiming to reduce the amount of round-end window clutter, so these would be better as tabs on the main window I think.

That is how they were submitted in https://github.com/goonstation/goonstation/pull/16410 but I could not find a solution for the related complications of honoring user preferences.

What's your recommendation for the work here?

flappybatpal:
> We're aiming to reduce the amount of round-end window clutter, so these would be better as tabs on the main window I think.

Agree, clicking multiple windows closed at round end is never good.

PR #17834, opened 2024-02-02
Add external cycling airlocks to clarion

TheTyrant25:
[MAPPING] [QOL]
## About the PR 
Similar to #17828. Requires #17828 to be merged first, as there is a bugfix in that which needs to be applied for cycling airlocks to work.

Adds airlock cycling mechanisms to external, space facing airlocks on clarion. Doesn't apply it to every set of doors (airbridge connections were excluded) for ease of player foot transit.

Also removes some empty access spawners which were on external airlocks. They weren't var edited and weren't doing anything.

## Why's this needed?
It's nice and thematic. Same reasoning as #17828.


TobleroneSwordfish (changes requested):
Actually looking at it, this bit in particular is going to be complete hell on high-pop with airlock cyclers.
image
If they could be disabled while the shuttle is docked then that would work a lot better.

TheTyrant25:
The current way that those are linked is that the east and west are connected, so as to not throttle throughput. opening the north or south ones attempts to shut off the connection. Definitely considering having it detect shuttles, as multiple spots have this "sometimes air facing, sometimes space facing" behaviour

PR #17839, opened 2024-02-02
Add end of round gang portraits

TobleroneSwordfish:
## About the PR 
Adds cool portraits of all the members of the winning gang at the end of the round.
Looks something like this:
image
Only the gang leader's name is gold, dead gang members will show as ghosts.

## Why's this needed? 
More uhhh Pride and Clout for the winning gang members, hopefully making people more likely to care about winning the round.
This might be a little too in your face, I'm open to suggestions to change it.

## Changelog 

```changelog
(u)LeahTheTech
(+)Members of the winning gang will now be displayed at round end.
```

### DIFF
--- a/code/datums/hud/gang_victory.dm
+++ b/code/datums/hud/gang_victory.dm
@@ -3,6 +3,7 @@
 	var/atom/movable/screen/text
 
 	New(datum/gang/winning_gang)
+		//big gang name victory text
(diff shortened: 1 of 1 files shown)


trollbreeder:
Maybe better to show the portraits in the end-of-round summary window, instead of as a HUD overlay

TobleroneSwordfish:
Actually I think TDH wants to do some kind of round-end animation based around this, so I'm just going to merge it as-is and let them cook.

PR #17845, opened 2024-02-02
Makes it possible to change the arithmetic comp's mode using another comp

SirNooben:
## About the PR 

Makes the arithmetic comp's mode able to be modified using other mechcomps.

## Why's this needed? 

It's a nice quality of life change which allows people to create more intricate circuits without having to use an arithmetic comp for each type of calculation, instead they can use one and just change the mode on that.

## Changelog 

```changelog
(u)SirNooben
(+)Makes it possible for the arithmetic comp's mode to be changed using an input coming from another comp.
```


Tarmunora (approved):

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -3267,6 +3268,39 @@ ADMIN_INTERACT_PROCS(/obj/item/mechanics/trigger/button, proc/press)
...
+			if("gte")
+				mode = "gte"
+			if("lte")
+				mode = "lte"
+			if("eq")
+				mode = "eq"
+			if("neq")
+				mode = "neq"
+			else
+				return
```
TobleroneSwordfish, line 3302: What's the point of this switch? If it's for validating the mode string then a `if (mode in list(...` would be a lot cleaner.

TobleroneSwordfish (changes requested):

code/modules/mechanics/MechanicMadness.dm
SirNooben: There we go I think I did what you said, at least that looks tidier to me

TobleroneSwordfish (approved):

PR #17853, opened 2024-02-03
CE spawns with the Aurora Belt MkII

Cherman0:
## About the PR 

Makes the CE spawn with the aurora belt, removes it from their locker.

## Why's this needed? 

It's somewhat awkward that they dont spawn with their tools. In the force field rework it was changed to reduce melee damage by 75% instead of giving you a 75% chance to dodge all melee attacks, making it significantly less valuable for antags to steal (ie it does nothing against grabs, disarms, batons, etc now). As such I dont feel that it will be an issue that it is not available at all if there's no CE, and it also introduces the benefit that the CE cryoing with their belt doesnt lock it out of existance for any future latejoining CE.

## Changelog 

```changelog
(u)Cherman0
(+)The CE now spawns wearing their Aurora Belt MkII
```


blackrep:
Doesn't this cause issues with spy thief? The CE's belt can be one of the targets.

Cherman0:
There's already code for removing possible bounties for jobspawn items if there are zero of that job (hos beret and captain auth disk) so i can just add the aurora belt to that

Tarmunora:
note this the sequestered cloner prefab will need updating, as a fully-filled CE locker (read: belt) is the main reward

Cherman0:
Yeah, those lockers are empty. Theyre just generic command lockers, not CE lockers.

code/obj/storage/secure_closets.dm
```diff
@@ -354,7 +353,29 @@ ADMIN_INTERACT_PROCS(/obj/storage/secure/closet, proc/break_open)
 		/obj/item/clothing/suit/space/light/engineer,
 		/obj/item/clothing/head/helmet/space/light/engineer,
 	#endif
+	)
 
+/obj/storage/secure/closet/command/chief_engineer/puzzle
```
Tarmunora, line 358: will also need to add it to the prefab

Tarmunora (approved):

code/obj/storage/secure_closets.dm
```diff
@@ -354,7 +353,29 @@ ADMIN_INTERACT_PROCS(/obj/storage/secure/closet, proc/break_open)
...
+	//obj/storage/secure/closet/command/chief_engineer/puzzle would never be real
+	//obj/storage/secure/closet/command/chief_engineer/puzzle:
```
ZeWaka, line 360:
```suggestion
```

ZeWaka (approved):

PR #17854, opened 2024-02-03
Tweak and fix of the misc tools box

Cherman0:
## About the PR 

Removes the mechscanner from the ce toolbox, adds a space GPS instead. Also correctly spells "separate".

## Why's this needed? 

They already have a mechscanner in their PDA. Also spelling things right is probably good.

### DIFF
--- a/code/obj/item/storage/misc.dm
+++ b/code/obj/item/storage/misc.dm
@@ -115,11 +115,11 @@
 	/obj/item/reagent_containers/glass/bottle/holywater)
 
 /obj/item/storage/box/misctools //used in CE locker
-	name = "miscellaneous tools"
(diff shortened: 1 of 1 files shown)


Cherman0:
Well firstly, the t-ray scanner already isnt in the box because they have it in their pda. But also, yes the GPS is a tool? On lots of maps there's only GPSs in mining and telesci so they can also be a bit annoying to come across. And I doubt "mechscanner that you have to manually upload everything with" is really functionality that anyone would like to have on-hand alongside their mechscanner that does that automatically. I personally ditch the mechscanner immediately at the start of every round because like, you dont need two.

TobleroneSwordfish:
The device analyzer has uses for crime and if you lose your PDA, and is actually an engineering tool. Not sure why the CE would need a GPS anyway.

PR #17856, opened 2024-02-03
Donut3 Engineering Tweaks

Cherman0:
## About the PR 

Modifies Donut3 engineering so that all engineering rooms are accessible without going through the main singularity room. Adds the toggleable radiation shielding doors to all windows that face into the singularity room, not just the one from the public hallway. Makes a few other minor tweaks to the engineering area.

Also removes mining access to the cargo office rear door, since I believe I accidentally added it in my previous PR that touched donut3 (which was a while ago and im not sure that anybody has noticed, but still).

donut3edit

## Why's this needed? 

A while back singularity radiation was modified to be line-of-sight based instead of purely distance based. Donut3's engineering was not modified to accomodate this, and lots of the surrounding rooms became hazardous due to the number of windows. Extending the radiation shield to all of the windows and adding routes into rooms that don't pass through the singularity area will hopefully encourage people to actually be in engineering and do stuff there, since you wont necessarily have to be permenantly stuck in a radsuit and mesons.

Also adds firedoor markers to a few random doors in engineering that were missing them.

## Changelog 

```changelog
(u)Cherman0
(+)Donut3 engineering modified for improved radiation safety outside of the singularity room.
```


TobleroneSwordfish (changes requested):
I like this, makes engineering much less awkward. You appear to have made accidental changes to the kitchen and tech storage though, check the MDB diff for details.

Cherman0:
There arent changes to tech storage or the kitchen, its just behind compared to master. Updated

TobleroneSwordfish (changes requested):
Actually, some nitpicks now I can look at the diff properly:
- What's the reason for making the center part symmetrical? It looks weird to have a gantry going down to the lack of door on the bottom.
image
- 5 total rad closets seems kind of excessive given the CE also gets a radsuit, and putting that wall of lockers there makes the phone a lot less conspicuous.
image

- This door only serves to make this room slightly easier to tide into from tool storage
image

Cherman0:
- It not being symmetrical bothered me. Presumably the gravity of the singularity puts a lot of mechanical stress on this area, you'd think it'd be symmetrically reinforced in order to not cause structural deformation.
- Given that engineers have a maximum job limit of 8, 5 radsuits doesnt seem too excessive to me. Not having enough for everyone is pretty common at the moment.
- That door is like 65% of the reason why I made this PR. There not being a door between those two rooms is my most hated thing about the whole map. Having to tear down that table and use a windoor as a normal door is really awkward, and odd door placement decisions of donut3 means it takes about as long to just walk around. I have access and the room is right there, let me in dammit.

Cherman0:
Made some changes to address your concerns:
- Rearranged the collectors in order to allow for a south door to the singularity enclosure
- Removed two radiation suit lockers, bringing the total back down to three
I also took the opportunity to make some changes to engineering tool storage to make it less bad. Contains fewer useless or nearly useless items like handheld welding fuel cans or earmuffs, contains more useful things like metal, glass, wires, and rods. Moved the pipe cart in and added a material reclaimer next to the fabricator.

TobleroneSwordfish (changes requested):
- The collectors look really odd like that and don't really make sense from a physical perspective (also shrinking the space makes it harder to add more later)
- The fuel tanks and earmuffs are very useful for crime
- There's a table, fire alarm and APC all on one tile north of the storage room, ideally none of these should be on the same tile

Please stop adding more random changes to this PR.

PR #17865, opened 2024-02-04
Added inhand sprites for the antique pistol.

GuvGuvington:
Finally, it will no longer look like the captain is wielding a generic phaser instead of his really cool and fancy phaser! 

[FEATURE] [OBJECTS]

## About the PR  

Gives the antique pistol a unique inhand sprite.

## Why's this needed?  

The phaser looks nothing like the antique pistol. Not only is this inconsistency visually unpleasing; it also affects combat, since the antique gun has the potential to be way stronger than a standard phaser. The generic phaser sprite basically acts as a disguise in this sense, which really should not be the case.

## Changelog 

```changelog
(u)GuvGuvington
(+)The Antique laser gun received some proper inhand sprites, both the model and functioning weapon.
```


DimWhat:
Why not give it to the nonfunctional version?

GuvGuvington (commented):
> Why not give it to the nonfunctional version? 

For some reason its treated as a seperate item that interestingly enough also doesnt use the generic phaser sprite that the functional antique pistol has.
Once I find the code for the non-weapon antique gun I will update it to use these sprites aswell.

GuvGuvington (commented):
> > Why not give it to the nonfunctional version? 

> For some reason its treated as a seperate item that interestingly enough also doesnt use the generic phaser sprite that the functional antique pistol has.
Once I find the code for the non-weapon antique gun I will update it to use these sprites aswell. 

Updated the non-gun model with the new sprite as well now!

goonstation.dme
```diff
@@ -33,6 +33,7 @@ var/datum/preMapLoad/preMapLoad = new
 #include "code\__secret_public.dme"
 
 // BEGIN_INCLUDE
+#include "_std\__odlint.dm"
```
TobleroneSwordfish, line 36: You're going to need to revert this change or the PR won't pass checks.

TobleroneSwordfish (changes requested):
The changelog section is for the in-game changelog and needs to be formatted correctly to work, see the template you're given when you go to create a PR.

TobleroneSwordfish (approved):

PR #17873, opened 2024-02-05
TGUI Message Modal

ZeWaka:
## About the PR 

This tool is a way to shit some arbitrary text at a user, but have it be kinda nice and not just in the chat window.

## Why's this needed? 

I wanted to shit arbitrary text at a user.

This is what it looks like
image


tgui/packages/tgui/interfaces/MessageModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Window } from '../layouts';
+import { KEY_ESCAPE } from 'common/keycodes';
+import { sanitizeDefAllowTags, sanitizeText } from '../sanitize';
+
+type MessageInputData = {
+  message: string;
+  timeout: number;
+  title: string;
+  theme: string;
+  sanitize: boolean;
```
mordent-goonstation, line 20:
```
  sanitize: BooleanLike;
```

tgui/packages/tgui/interfaces/MessageModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const { act, data } = useBackend<MessageInputData>(context);
+  const { message, timeout, title, theme, sanitize } = data;
+  // Dynamically changes the window height based on the message.
+  const windowHeight
+    = 125 + Math.ceil(message?.length / 3);
+
+  let outputMessage = message;
+  if (sanitize) {
+    let allowedHTMLTags = sanitizeDefAllowTags;
+    allowedHTMLTags.push('a'); // We commonly want to let users redirect to a URL.
```
mordent-goonstation, line 33: This will mutate the imported array. Not worth me trying to work out if that's okay in this particular stack of transpiled module bundlers. Try the below to work regardless:
```
const allowedHTMLTags = [...sanitizeDefAllowTags, 'a'];
```
In either case, future tgui project to have a Markdown renderer. 😀

mordent-goonstation (changes requested):

ZeWaka:
Be quiet I created you

PR #17875, opened 2024-02-05
Fix disguiser organ names being permament

flappybatpal:
## About the PR 
Makes organ name reset after the holographic disguise expires.

## Why's this needed? 
Organ name changes were permanent.
Head was not renamed to match disguise (it sets name in update_icon using donor._name).

### DIFF
--- a/code/obj/item/device/disguiser.dm
+++ b/code/obj/item/device/disguiser.dm
@@ -134,6 +134,7 @@ TYPEINFO(/obj/item/device/disguiser)
 		// Restore original appearance.
 		else
 			user.real_name = src.real_name
+			user.on_realname_change()
 			AH.CopyOther(oldAH)
 			if (user.limbs)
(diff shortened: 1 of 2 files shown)


flappybatpal:
I question if there is even a need to rename the organs at all.
It is possible to lose limbs/undergo surgery without breaking the holographic disguise but this seems incredibly niche considering it's quite hard to keep the disguise from failing.

Valtsu0:
> I question if there is even a need to rename the organs at all.
> It is possible to lose limbs/undergo surgery without breaking the holographic disguise but this seems incredibly niche considering it's quite hard to keep the disguise from failing.

And that case the disguise should probably fail, at least for the limb

TobleroneSwordfish:
I think health analysers can show organ names too actually

Valtsu0:
> I think health analysers can show organ names too actually

Thais is indeed the case

PR #17877, opened 2024-02-05
Mixer refactor moving the KitchenHelper to be a /obj/machinery type instead of /obj/submachine. 

mloccy:
[OBJECTS][CATERING][CHORE]
## About the PR 
Reworks the KitchenHelper to be a machine, what this means is it has a slightly longer time to finish mixing and I added a ding sound effect to help identify when it's now done, it now will not operate when there is no power in the area.

## Why's this needed? 
Kitchen mixer is a machine which should use power and should probably not work while there's no power in the area, also there seems to be a desire to move away slowly from use of submachine in favor of machinery.


pgmzeta (approved):
Small comment on power usage, but otherwise appears to work as expected after light in-game testing. Nice work!

PR #17885, opened 2024-02-06
Prevent prefilled drinking glasses from overfilling

### LINKED ISSUE #17585: Donut 2 50u shot glass
### Labels

[BUG][MINOR]

### Description

On donut 2 there will be a drink in the bar that always spawns with 50u of a drink regardless of what glass it is, including shot glasses.

### Steps to reproduce

1. Spawn into donut 2
2. Check pre-spawned full drink at bar
3. Watch in horror as you get a free 50u shot glass to poison (if you're lucky)

### Expected Behavior

The glasses to always be a standard drinking glass or to not have 50u regardless of size.

Reported by: Asterion0
Active test merges: #17354
Round log date: 2024-01-07-05-01
Reported on: 2024-01-07 05:35:21
Map: DONUT2

TheTyrant25:
[BUG] [CHEMISTRY]
## About the PR 
The prefilled variant of drinking glass, aka `/obj/item/reagent_containers/food/drinks/drinkingglass/random_style/filled`, would manually set `initial_volume` to the correct amount once the random drink type was picked. However, since the parent code of `New()` had already run, the true `reagents.maximum_volume` was set at `50` due to that being the `initial_volume` of the parent drinking glass when it was made.

This PR manually sets `src.reagents.maximum_volume` at the same time as `initial_volume` is set.

## Why's this needed? 
Fixes #17585
Previously, you could end up with shot glasses with 50u of chemicals inside.

### DIFF
--- a/code/modules/chemistry/tools/food_and_drink.dm
+++ b/code/modules/chemistry/tools/food_and_drink.dm
@@ -1660,18 +1660,22 @@ ADMIN_INTERACT_PROCS(/obj/item/reagent_containers/food/drinks/drinkingglass, pro
 				src.amount_per_transfer_from_this = 15
 				src.gulp_size = 15
 				src.initial_volume = 15
+				src.reagents.maximum_volume = 15
 			if ("wine")
 				src.name = "wine glass"
 				src.icon_state = "glass-wine"
 				src.initial_volume = 30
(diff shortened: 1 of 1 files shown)


TheTyrant25:
manually forcing the spawned glass to be a shot glass and testing reveals this fix works

PR #17886, opened 2024-02-06
Add baseline 15 potency to plants

TDHooligan:
[QOL][HYDROPONICS]
## About the PR 
Makes all seeds start at potency 15

## Why's this needed? 
Partially addresses the difficulties in getting Fruit Juice as bartender
- Lets well meaning, less-knowledgeable  botanists provide useful ingredients to Bar
- Bartenders can simply ask for 'apples'/'oranges', instead of 'infused apples & oranges' 
- Makes a bit more sense (space seeds create bone dry watermelon?
- Makes 'dry food' a unique thing for botanists to make
- Makes more sense than adding a flat '15' to all reagents created by plants

dunno if there are more balance ramifications, idk any seed you cant just hit "infuse" 4 times with saltpetre on.

i got more bartending ideas coming up to better address bartending/botany co-operation directly. but this is a 2 character bandaid that helps a lot

## Changelog 

```changelog
(u)TDHooligan
(+)Plants have a baseline 15 potency
```

### DIFF
--- a/code/modules/hydroponics/plants.dm
+++ b/code/modules/hydroponics/plants.dm
@@ -252,7 +252,7 @@ ABSTRACT_TYPE(/datum/plant)
 	var/harvtime = 0 // same vars found in /datum/plant honestly. They go largely towards
 	var/harvests = 0 // the same purpose for the most part.
 	var/cropsize = 0
-	var/potency = 0  // Apart from this one - this one deals with reagents.
+	var/potency = 15  // Apart from this one - this one deals with reagents.
 	var/endurance = 0
 	var/list/commuts = null // General transferrable mutations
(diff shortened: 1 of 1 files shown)


NibChocolateeny:
i know why you are making this cahnge, but I think mr earthfire is right - the porduce should have reagents inside at start, since it fixes non-botany apples/other things having no good juice
making potancy more for all things means that soem things need to be re-balanced (harder for amanitin mutate, does potency of life weed need to be more? etc).
but I agree that it should be not so hard for bar to have apple jujice from fruit
In both cases, there are many changes needed, but for second (add some juice to fruit at all times) there are less "balance problems"

also, why does cut lemon piece have guarantee of juice but normal lemon does not? funny!

PR #17888, opened 2024-02-06
extra torn cloak, adventurous scarf and ten-gallon hat colors

CoolKakapo:
## About the PR 
Exactly what the title says.

image

## Why's this needed? 
Variety! A good outfit is very important for some people, and it sucks when you just cant get your outfit right. this should maybe help that a little.

## Changelog 

```changelog
(u)The_Kakapo
(+)Adds some new Torn Cloak, Adventurous Scarf and Ten-Gallon hat colors
```


icons/mob/clothing/overcoats/worn_suit_gimmick.dmi
WALPVRGIS, line 1: I don't think these palettes work either, they also just look kinda hue-shifted and a bit muddy. I made the original versions of these and would be down to add more colours myself that fit existing palettes better, honestly!

WALPVRGIS (changes requested):
Sprite review; mostly about palettes used;

icons/mob/clothing/overcoats/worn_suit_gimmick.dmi
CoolKakapo: I'd appreciate it if you would make some palettes, you are the creator of these after all. The custom color thing using the paint machine also sounds like a good idea, but I kind of lack the coding skills for that. I could turn the colors of the hats into ones that are already in the game, to make them a lot less horrible.

WALPVRGIS: Aha, I wouldn't say it's horrible, but we use a different style of palettes on goon rather. I'm not at home rn, but I'll get some palettes here hopefully in the next day or two!

Also with the paint machine hat thing, you wouldn't need to do any additional code :) (I too have no code skills), the white hat just needs to be added to the game, and then players can go utilise the paint machine to make a paint bucket to use on it as per usual!

PR #17889, opened 2024-02-07
Adds a paper-to-ash reaction catalyzed by phlogiston 

NibChocolateeny:
[Chemistry][Hydroponics][Feature][QoL]

## About the PR 
add a simple reaction where paper (1) + phlogiston (0) = ash (1) + phlogiston (0) (phlogiston is not used, but is needed)

## Why's this needed? 
betterment for sience and botany, mostly in botany, for making of potash and green goop

## Changelog

```changelog
(u)NibChocolateeny
(+)Adds a chemical reaction that turns paper to ash using phlogiston as a catalyst
```


Tarmunora:
Eh, this seems like a textbook case for just using a bunsen burner instead...

NibChocolateeny:
@Tarmunora can alos use welder on the floor with sheets, or reagnant heater, many options. ash is not hard to make with paper - but it is nice to have way to make faster. phlog burns all things, also sets paper on fire if you pour onto paper on th eground, so I think it makes sense for it to burn paper if it is in jar
Also botany can splice plant with phlog + paper for even more easier making. for science, phlog is not so hard to make, so it is nice for them if lot of ash is needed.

i think makes sense, and is not so strong of a change, but it is ceveloper choice ¯\_(ツ)_/¯
(botany can also steal bunrers in science if you want   :p    )

PR #17891, opened 2024-02-07
TGUIification of Pump Control Computer + Minor Rework

MeggalBozale:
## About the PR 
This PR not only converts the Pump Control Computer to a TGUI interface, but refactors and improves upon underlying code to make the computer more responsive and useful.

While it is demonstrated that the new UI will hide settings elements for pumps which are missing or jammed, this is only a QoL change -- the old computer would simply not tell you that it cannot change the pump settings, as all pumps on the computer are changed with packetry VIA radio transmission.

Lastly, does two things to pump.dm:
1 - changes the magic num for min pressure (0) to a define. makes it easier to change this number in the future if necessary.
2 - adds two pieces of information to pump broadcasts: min and max pressure. this assists in not having to paste around pump pressure if it ever changes while keeping something so benign from being a global define. These are not attributes of the pump currently, but doing so will be simple enough should someone want to make pumps with different min/max values.

Old UI:
image

New UI:
image

Live Demonstration:
https://github.com/goonstation/goonstation/assets/27376947/b804d806-3402-4a6c-97a5-ba5462e9f48f
(not sure why this doesn't embed but w/e)

## Why's this needed? 
For the UI, TGUI is more maintainable and looks better. For the refactor/rework, code improvements will make the computer itself easier to maintain and use. Allowing the granular pressure settings introduced in this PR means that a multitool/packet nerdery is no longer necessary for users to experiment with particular pump pressures - which shouldn't be necessary in the first place.

## Changelog
```changelog
(u)MeggalBozale
(*)Pump Control Computers now use TGUI, and can set the specific pressure of pumps they control. Give it a try!
```


Laboredih123:
ooh now i like that

MeggalBozale:
XL sizing without being a mapping pr *sweat

MeggalBozale:
image
Also, light mode

Laboredih123:
> Also, light mode

GAH

MeggalBozale:
not sure how to resolve the merge conflict btw

tgui/packages/tgui/interfaces/PumpControl.tsx
```diff
@@ -0,0 +1,148 @@
...
+      width={400}
+      height={500}>
+      <Window.Content scrollable>
+        <Button
+          style={{
+            "margin-bottom": "10px",
+          }}
+          onClick={() => refresh()}
+        >
+          Refresh
```
TobleroneSwordfish, line 141: This looks kind of odd up here on its own, but I don't really know where else to put it. 
```suggestion
          onClick={() => refresh()}
          icon="wifi"
        >
          Rescan
```
If we're going to have a refresh button it should be clear to the user that it has some meaning in-world rather than just being an F5 button.

tgui/packages/tgui/interfaces/PumpControl.tsx
```diff
@@ -0,0 +1,148 @@
...
+            onChange={(_e: any, value: number) => setPressure(value)}
+          />
+        </Stack.Item>
+      </Stack>
+    </Box>
+  );
+};
+
+// Responsible for creating a section for the pumps in an area.
+const PumpArea = (_:any, context:any) => {
```
TobleroneSwordfish, line 88: _ means an unused variable, so this should use an actual name instead.

tgui/packages/tgui/interfaces/PumpControl.tsx
```diff
@@ -0,0 +1,148 @@
+/**
+ * @file
+ * @copyright 2024
+ * @author Romayne (https://github.com/MeggalBozale)
+ * @license ISC (https://choosealicense.com/licenses/isc/)
+ */
+
+import { useBackend, useLocalState } from '../backend';
+import { Box, Button, Section, Slider, Stack } from '../components';
+// import { DataInputOptions } from './common/DataInput';
```
TobleroneSwordfish, line 10:
```suggestion
```

tgui/packages/tgui/interfaces/PumpControl.tsx
```diff
@@ -0,0 +1,148 @@
...
+  area_name: string; // Name of the area this pump is in
+  processing: boolean; // Whether we are waiting for packet response or not
+  alive: number; // A value of -1, 0, or 1 where -1 is checking if the pump is alive, 0 is dead, and 1 is alive
+}
+
+// List of areas which have pumps
+type AreaList = {
+  area_list: { [key: string] : { [key: string] : PumpData } };
+  frequency: number;
+};
```
TobleroneSwordfish, line 30: Same as with the TTV UI, these should be in a separate file

TobleroneSwordfish (changes requested):

tgui/packages/tgui/interfaces/PumpControl.tsx
MeggalBozale: renamed to props

tgui/packages/tgui/interfaces/PumpControl.tsx
MeggalBozale: moved to /PumpControl

code/modules/power/TEG.dm
```diff
@@ -1598,197 +1598,209 @@ TYPEINFO(/obj/machinery/power/furnace/thermo)
...
+	"tag" - Name of pump
+	"netid" - Network ID of pump
+	"device" - "AGP" unique pump identifier string
+	"power" - "on" or "off" depending on state
+	"min_output" - MIN_PRESSURE (0kpa)
+	"max_output" - MAX_PRESSURE (~15000kpa)
+	"target_output"- current pump output
+	"address_tag" = "pumpcontrol"
+
+	Some added vars:
```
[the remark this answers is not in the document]
MeggalBozale: those referred to vars that the packet doesnt send but tbh it can just be one thing

tgui/packages/tgui/interfaces/PumpControl.tsx
MeggalBozale: Sounds good, I think the wording "Requery Pumps" may convey such a bit better even though it's wordy. Not like it doesn't have space to do so

MeggalBozale:
i cant make the merge conflict go away >:( but in other news, changes made.

MeggalBozale:
guh i can add/remove that

TobleroneSwordfish (changes requested):
Looks like you've accidentally pulled in changes from several other PRs, you'll need to unfuck the git history a bit before this can be merged.

MeggalBozale:
woops

MeggalBozale:
Reopening this i dun goofed

PR #17894, opened 2024-02-07
Require at least 14 players to vote for oshan

TheTyrant25:
[BALANCE]
## About the PR
Require it so that at least 14 players are required before oshan can be voted.

The number 14 was picked so that it would RP's minimum population requirement for cogmap1. And fit on the wiki chart nicer (see https://wiki.ss13.co/Maps#Map_availability).

## Why's this needed? 
Extreme low pop can struggle on oshan quite a lot, since they only have one power source and no "backup" power to subsist off (i.e. solars or catalytics). By requiring a higher number of players it can stop understaffed shifts from being plunged into darkness 20 minutes in.

## Changelog
```changelog
(u)Tyrant
(*)Oshan now needs at least 14 players online to be voted for
```


TheTyrant25:
also fixed some indentation things that i saw hope that's okay

Tarmunora (approved):

pgmzeta:
Updated https://wiki.ss13.co/Maps#Map_availability

PR #17897, opened 2024-02-07
TGUIification of TTV menu

MeggalBozale:
## About the PR 
This PR serves only to convert the TTV menu to TGUI. There is one mechanical change, which is that you can, only in the menu, select which part of the tank transfer valve to attach a tank to. This has no effect on gameplay, it just felt right to add.

Old menu:
ApplicationFrameHost_qUeEqNnDRB

EDIT newer menu:
image

Old New menu:
YXD9vrXZ6u
gSfgxDDs0G

## Why's this needed? 
TGUI is nicer to have and use, and more in-game menus should be converted to using TGUI. 

## Changelog
```changelog
(u)MeggalBozale
(*)Tank Transfer Valves now use TGUI menus.
```


MeggalBozale:
image
light mode flashbang

MeggalBozale:
Would it be fine to add the pressure indicators for that? I dont feel anything makes that less bare besides providing pressure info

itchybones:
You beat me to it. Looks good 👍

MeggalBozale:
image
new menu

code/obj/item/device/transfer_valve.dm
```diff
@@ -138,64 +138,103 @@ TYPEINFO(/obj/item/device/transfer_valve)
...
-					var/turf/bombturf = get_turf(src)
-					logTheThing(LOG_BOMBING, usr, "closed the valve on a TTV tank transfer valve at [log_loc(bombturf)].")
-					message_admins("[key_name(usr)] closed the valve on a TTV tank transfer valve at [log_loc(bombturf)].")
-				else
-					var/turf/bombturf = get_turf(src)
-					logTheThing(LOG_BOMBING, usr, "opened the valve on a TTV tank transfer valve at [log_loc(bombturf)].")
-					message_admins("[key_name(usr)] opened the valve on a TTV tank transfer valve at [log_loc(bombturf)].")
+		switch(action)
+			if ("add_item")
+				if (ismob(usr))
```
TobleroneSwordfish, line 214: `usr` is always a mob and will always be present here unless the entire engine has fallen apart at the seams, so this check and cast isn't needed afaik

code/obj/item/device/transfer_valve.dm
```diff
@@ -138,64 +138,103 @@ TYPEINFO(/obj/item/device/transfer_valve)
...
+			if ("remove_device")
+				src.attached_device.set_loc(get_turf(src))
+				src.attached_device.master = null
+				src.attached_device = null
 				UpdateIcon()
-			if(href_list["device"])
+			if ("interact_device")
 				attached_device.attack_self(usr)
-			if(href_list["straps"])
+			if ("remove_straps")
```
TobleroneSwordfish, line 237: This doesn't seem to be called anywhere? Would make more sense to be on attack with a snipping tool anyway.

tgui/packages/tgui/interfaces/TTV.tsx
```diff
@@ -0,0 +1,107 @@
...
+ * @author Romayne (https://github.com/MeggalBozale)
+ * @license ISC
+ */
+
+import { useBackend } from '../backend';
+import { Button, LabeledList, RoundGauge, Section, Stack } from '../components';
+import { Window } from '../layouts';
+import { toTitleCase } from '../../common/string';
+import { formatPressure } from '../format';
+interface AirVendorParams {
```
TobleroneSwordfish, line 13: This should probably be `TransferValveParams` or something. Also generally convention is to put the type definitions in a separate `type.ts` file.

tgui/packages/tgui/interfaces/TTV.tsx
```diff
@@ -0,0 +1,107 @@
...
+    tank_two,
+  } = data;
+  let windowWidth = 650;
+  let windowHeight = 160;
+  return (
+    <Window width={windowWidth} height={windowHeight}>
+      <Window.Content>
+        <Stack>
+          <Stack.Item width={windowWidth/3}>
+            <TankInfo height={windowHeight/16.666} tank={tank_one} />
```
TobleroneSwordfish, line 78: I don't think this `height` param is doing anything

tgui/packages/tgui/interfaces/TTV.tsx
```diff
@@ -0,0 +1,107 @@
...
+    tank_one,
+    tank_two,
+  } = data;
+  let windowWidth = 650;
+  let windowHeight = 160;
+  return (
+    <Window width={windowWidth} height={windowHeight}>
+      <Window.Content>
+        <Stack>
+          <Stack.Item width={windowWidth/3}>
```
TobleroneSwordfish, line 77: `Stack.Item`s ideally shouldn't define their own width I think, see the inherited `Flex` `align` and `justify` properties

TobleroneSwordfish (changes requested):
Looks good, just some minor code stuff.
Oh also I know I said to make it shorter but now it looks like this with the default plasma tank inserted so it needs to be just a tiny bit longer:
image

code/obj/item/device/transfer_valve.dm
MeggalBozale: WOOPS and yea, forgot to add the straps bit to the GUI. But will be easier to just be done by a cutting tool

tgui/packages/tgui/interfaces/TTV.tsx
MeggalBozale: :+1: name is a mistake

tgui/packages/tgui/interfaces/TTV.tsx
MeggalBozale: it does but i can define a constant height tbh ive had to manually edit it anyway since height var is weird

code/obj/item/device/transfer_valve.dm
MeggalBozale: added snipping tool handling to attacky

code/obj/item/device/transfer_valve.dm
MeggalBozale: removed checks for mob with usr

tgui/packages/tgui/interfaces/TTV.tsx
MeggalBozale: moved interface to subfolder 'TTV' with index.tsx and type.ts

tgui/packages/tgui/interfaces/TTV.tsx
MeggalBozale: removed height/width variables

tgui/packages/tgui/interfaces/TTV.tsx
MeggalBozale: removed height/width variables

TobleroneSwordfish (changes requested):

MeggalBozale:
Committed those, any other minor changes worth making to improve UI/UX lmk i have all of today to work on it

PR #17907, opened 2024-02-08
Kitchen Mixer power indication LED 

mloccy:
[ENHANCEMENT][CATERING][[OBJECTS]

## About the PR 
Now that the mixer is a machine it can and will lose power when the kitchen loses power. In order for this to be a better experience for the player I have made minor modifications to the mixer sprite to put in an LED like feature. When the power is on and the mixer is idle it will be a flat green color, when the mixer is mixing it will be a blinking yellow / black color. When powered off, the LED will be black.

## Why's this needed? 
It just gives some visual feedback to players that the mixer is powered off / working, otherwise it just seems like the UI isn't working until you figure out what's going on with the other machines too. This PR changes the mixer to show when it has contents, the old blender_on state now shows that the blender has some contents.

## Changelog 

```changelog
(u)mloccy
(*)Added an LED on the kitchen mixer to indicate if it's powered on / off or mixing.
```

## Screenshots

Mixer on
MixerOn

Mixer Off

Mixeroff

Mixing

https://github.com/goonstation/goonstation/assets/757551/3e8c05d9-9b8f-412a-98dd-966267c2660d


code/obj/machinery/food_and_drink/mixer.dm
```diff
@@ -239,13 +241,27 @@ TYPEINFO(/obj/machinery/mixer)
...
+		else
+			src.icon_state = "blender_empty"
+
+		src.overlays.len = 0
+		var/powered = src.status & NOPOWER
+
+		if (powered)
+			src.overlays += image(src.icon, "blender_off")
+		else if (!powered && !src.working)
+			src.overlays += image(src.icon, "blender_powered")
```
ZeWaka, line 263: Cleaner to use OverlayManager and such

ZeWaka (changes requested):

code/obj/machinery/food_and_drink/mixer.dm
mloccy: Did you mean UpdateOverlays by any chance? Struggling to understand how I would use anything OverlayManager to do that.

ZeWaka: Yes

code/obj/machinery/food_and_drink/mixer.dm
```diff
@@ -239,13 +241,31 @@ TYPEINFO(/obj/machinery/mixer)
...
+		src.ClearSpecificOverlays(1)
+		var/powered = src.status & NOPOWER
+
+		if (powered)
+			src.UpdateOverlays(image(src.icon, "blender_off"), "blender_off")
+		else if (!powered && !src.working)
+			src.UpdateOverlays(image(src.icon, "blender_powered"), "blender_powered")
 		else
-			src.icon_state = "blender"
+			src.UpdateOverlays(image(src.icon, "blender_working"), "blender_working")
```
ZeWaka, line 269: aren't these all unique? it should be like `blender_light` - then you don't need to clear stuff all the time

ZeWaka (changes requested):

code/obj/machinery/food_and_drink/mixer.dm
mloccy: I misunderstood how the function worked, makes more sense this way and implemented now. Thanks!

ZeWaka (approved):
Looks good on mobile review

PR #17914, opened 2024-02-09
Lets midround wraiths evolve without souls

CalliopeSoups:
[A-Gamemodes] [C-Balance]
## About the PR 
Wraiths can currently bypass the 3 souls needed to evolve on disaster rounds. This PR gives the same functionality to midround wraiths.

## Why's this needed? 
Midrounds start from roughly 25 minutes into the shift. From my observations, it takes a fast wraith about 10 minutes to get the souls needed to evolve. A slower, unlucky or newer wraith might take around 20 minutes. 

Midround wraiths, at best, lose out on 25 minutes to build up to evolution. Often they're spawned later, which is a particular detriment to trickster wraith on classic I feel. This is because when you evolve, you lose your points and (correct me if I'm wrong) passive point gen total, on top of the time needed to collect the souls. Wraiths tend to have some long cooldowns, so I think this would allow for more interaction past animated objects and such for midround wraiths.

## Changelog 

```changelog
(u)CalliopeSoups
(+) Midround wraiths can now evolve as soon as they have the points.
```

### DIFF
--- a/code/global.dm
+++ b/code/global.dm
@@ -299,7 +299,7 @@ var/global
 	announce_banlogin = 1
 	announce_jobbans = 0
 	radio_audio_enabled = 1
-
+	latejoin_wraith = 0
 
 	outpost_destroyed = 0
 	signal_loss = 0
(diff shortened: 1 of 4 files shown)


flappybatpal:
This is probably a bit overkill. I think it's ok for wraiths to have to work for their second form.

CalliopeSoups:
Yeah that's okay, I wasn't aware of the bonus point generation mechanic when I made the PR too. Helps a lot and I only just found out...

ZeWaka:
Seems like the author is not interested in this PR, closing for now. Can always be reopened / changed / whaterver.

PR #17915, opened 2024-02-09
Let tiny items be hidden under floor tiles

TobleroneSwordfish:
[BALANCE] [FEATURE]
## About the PR 
Makes it so putting a floor tile on a plating turf hides all `W_CLASS_TINY` items in the turf and reveals them when it's pried up again. Also adds a signal for this and makes mousetraps trigger on being revealed in this way.

## Why's this needed? 
Cool idea, allows for convoluted traps and hidden items. May make hiding items too easy though?

## Changelog 

```changelog
(u)LeahTheTech
(*)Tiny items can now be hidden under floor tiles, mousetraps hidden like this will trigger when the tile is pried up again.
```

### DIFF
--- a/code/turf/floors.dm
+++ b/code/turf/floors.dm
@@ -26,8 +26,8 @@
 	/// if this turf is immune to explosion (explosion immune turfs immediately return on ex_act())
 	var/explosion_immune = FALSE
 	///Things that are hidden "in" this turf that are revealed when it is pried up.
-	///We have to do this since turf.contents is a list of things ON the turf, their actual loc is null.
-	var/list/hidden_contents = null
+	///Kept in a hidden object on the turf so that `get_turf` works as normal. Yes this is crime, fight me I have a possum.
+	var/obj/effects/hidden_contents_holder/hidden_contents = null
 
 	New()
 		..()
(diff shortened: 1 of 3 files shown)


code/turf/floors.dm
```diff
@@ -25,6 +25,9 @@
 	var/tmp/roundstart_dir
 	/// if this turf is immune to explosion (explosion immune turfs immediately return on ex_act())
 	var/explosion_immune = FALSE
+	///Things that are hidden "in" this turf that are revealed when it is pried up.
+	///We have to do this since turf.contents is a list of things ON the turf, their actual loc is null.
+	var/list/hidden_contents = null
```
ZeWaka, line 30: needs handling for turf replacement

TobleroneSwordfish:
The items aren't actually on the turf, they're stored in `null` to prevent stuff like that

pgmzeta:
Using the null location trick will hide items like the captain's authentication disk from its related pinpointer.

TobleroneSwordfish:
Using the power of crime and a total disregard for good OOP practice I've fixed all the `get_turf` related issues by just putting an invisible holder object on the turf to hold the items in.

Tarmunora:
> Using the power of crime and a total disregard for good OOP practice I've fixed all the `get_turf` related issues by just putting an invisible holder object on the turf to hold the items in.

😭

PR #17921, opened 2024-02-10
Reduces the 1 million credits via PTL CE objective

Cherman0:
## About the PR 

Reduces the requirement for the CE's one million credit crew objective (and consquently, the 1.21 Jiggawatt medal) from one million to ten thousand credits.

## Why's this needed? 

A series of incremental changes over time has made it increasingly non-viable to produce huge amounts of power without essentially self-antagging or DoSing the server in the process. 

As it stands, achieving one million credits would require an output of approximately 65129 MW for an hour or 213216 MW for half an hour. Achieving ten thousand credits would require about 10.6 MW for an hour or 280.9 MW for half an hour.

## Changelog 

```changelog
(u)Cherman0
(+)The CE's one million ptl credit crew objective has been reduced to ten thousand credits.
```


Cherman0:
I do want to make sure its still somewhat possible on classic rounds that tend to last about an hour (and usually you have less time than that that the engine is at peak operation). For numbers' sake, earning 50k credits would require about 1725 MW for an hour or 4203 MW for half an hour. Its difficult to fine-tune this number because of how comically horrible the PTL cash formula is.

MeggalBozale:
It's an objective and if not related to a medal, should really not be that hard IMO. But I feel scaling down the objective to 10,000 credits makes engineering comparatively unprofitable (But I do not know the extent of profits a pipeburn is capable of) for full-shift effort vs. QM, or any other means of generating credits. Classic and RP will differ by RP being ~50% longer on average, if need be could scale RP to be higher.. its a pretty big difference to keep same on both tbh

TobleroneSwordfish:
I'd rather just buff the PTL output, 10k credits for running the engine at high output all shift is pathetic compared to requisitions or crystal bazar

Cherman0:
While I agree that the ptl could use a buff/refactor, even then 1 million credits seems like an excessively high number to tie to a crew objective (which are usually not that hard). Part of it has to do with the very dubious state of TEG balance, where a normal setup will probably not pass 20MW for an extended period of time, whereas advanced (and very repetitive/boring) setups make an infinitely increasing output. So the PTL output basically has to be balanced to infinity instead of a more sane range.

MeggalBozale:
You can buff the ptl outout without drastically buffing the TEG though? Would just need to carefully only buff the lower outputs of 20-200MW (Reactor can get quite high, singulo can surpass 1GW)

Azrun:
> You can buff the ptl output without drastically buffing the TEG though? Would just need to carefully only buff the lower outputs of 20-200MW (Reactor can get quite high, singulo can surpass 1GW)

Yep! https://www.desmos.com/calculator/r8bsyz5gf9

PR #17922, opened 2024-02-10
TGUIification of Pump Control Computer + Minor Rework

MeggalBozale:
## About the PR 
This PR not only converts the Pump Control Computer to a TGUI interface, but refactors and improves upon underlying code to make the computer more responsive and useful.

While it is demonstrated that the new UI will hide settings elements for pumps which are missing or jammed, this is only a QoL change -- the old computer would simply not tell you that it cannot change the pump settings, as all pumps on the computer are changed with packetry VIA radio transmission.

Lastly, does two things to pump.dm:
1 - changes the magic num for min pressure (0) to a define. makes it easier to change this number in the future if necessary.
2 - adds two pieces of information to pump broadcasts: min and max pressure. this assists in not having to paste around pump pressure if it ever changes while keeping something so benign from being a global define. These are not attributes of the pump currently, but doing so will be simple enough should someone want to make pumps with different min/max values.

Old UI:
image

New UI:
image

Live Demonstration:
https://github.com/goonstation/goonstation/assets/27376947/b804d806-3402-4a6c-97a5-ba5462e9f48f
(not sure why this doesn't embed but w/e)

## Why's this needed? 
For the UI, TGUI is more maintainable and looks better. For the refactor/rework, code improvements will make the computer itself easier to maintain and use. Allowing the granular pressure settings introduced in this PR means that a multitool/packet nerdery is no longer necessary for users to experiment with particular pump pressures - which shouldn't be necessary in the first place.

## Changelog
```changelog
(u)MeggalBozale
(*)Pump Control Computers now use TGUI, and can set the specific pressure of pumps they control. Give it a try!
```

Same as https://github.com/goonstation/goonstation/pull/17891, but without all the messing up I did


code/modules/power/TEG.dm
```diff
@@ -1598,197 +1598,207 @@ TYPEINFO(/obj/machinery/power/furnace/thermo)
...
+		pump_data_ref["alive"] = PUMP_ALIVE
+		return
 
-	process()
-		..()
-		if(status & (BROKEN | NOPOWER))
-			return
-		//src.updateUsrDialog()
+	var/list/infoset = new()
+	for (var/key as anything in signal.data)
```
Laboredih123, line 1669: Untyped vars do not require `as anything`

code/modules/power/TEG.dm
```diff
@@ -1598,197 +1598,207 @@ TYPEINFO(/obj/machinery/power/furnace/thermo)
...
+	var/area_name_index = src.pump_infoset.Find(infoset["area_name"])
+	if (!area_name_index)
+		// We are first of an area, create our place in the list
+		src.pump_infoset[infoset["area_name"]] = list()
+		src.pump_infoset[infoset["area_name"]][infoset["netid"]] = infoset
+	else
+		// We are not first of an area, place us in the list alphabetically
+		var/iter = 1
+		var/list/L = src.pump_infoset[infoset["area_name"]]
+		while ((iter <= L.len) && sorttext(infoset["area_name"], L[iter]) == -1)
```
Laboredih123, line 1687: Length(L) is preferred over L.len

code/modules/power/TEG.dm
```diff
@@ -1598,197 +1598,207 @@ TYPEINFO(/obj/machinery/power/furnace/thermo)
...
+
+		// Insert key first
+		L.Insert(iter, infoset["netid"])
+		L[infoset["netid"]] = infoset
+
+/obj/machinery/computer/atmosphere/pumpcontrol/process()
+	..()
+	if(status & (BROKEN | NOPOWER))
+		return
+	if(!src.pump_infoset.len)
```
Laboredih123, line 1698: ditto

code/modules/power/TEG.dm
```diff
@@ -1598,197 +1598,207 @@ TYPEINFO(/obj/machinery/power/furnace/thermo)
...
+		src.request_data() // get data for first time
+	src.check_if_alive()
+
+/// Check for pumps that 'sploded or are otherwise unreachable
+/obj/machinery/computer/atmosphere/pumpcontrol/proc/check_if_alive(override_cooldown = FALSE)
+	if (!override_cooldown && ON_COOLDOWN(src, "check_pumps_living", 1 MINUTE)) return
+	for (var/area_name as anything in src.pump_infoset)
+		for (var/pump as anything in src.pump_infoset[area_name])
```
Laboredih123, line 1706: ditto

code/modules/power/TEG.dm
```diff
@@ -1598,197 +1598,207 @@ TYPEINFO(/obj/machinery/power/furnace/thermo)
...
+			if (src.pump_infoset[area_name][pump]["alive"] == PUMP_ALIVE) // Don't modify dead ones if they were just jammed theyll respond
+				src.pump_infoset[area_name][pump]["alive"] = PUMP_SCHRODINGER // https://i.imgur.com/mUfxPmb.png
+
+	src.request_data()
+	SPAWN(5 SECONDS)
+		for (var/area_name as anything in src.pump_infoset)
+			for (var/pump as anything in src.pump_infoset[area_name])
```
Laboredih123, line 1713: ditto

code/modules/power/TEG.dm
```diff
@@ -1598,197 +1598,207 @@ TYPEINFO(/obj/machinery/power/furnace/thermo)
...
+				if (src.pump_infoset[area_name][pump]["alive"] != PUMP_ALIVE)
+					src.pump_infoset[area_name][pump]["alive"] = PUMP_DEAD
+
+/// Get a pump by net id. Does not ask for pump data from pump
+/obj/machinery/computer/atmosphere/pumpcontrol/proc/getPump(var/netid)
+	for (var/area_name as anything in src.pump_infoset)
+		var/list/L = src.pump_infoset[area_name]
+		for (var/pump as anything in L)
```
Laboredih123, line 1721: ditto

Laboredih123 (commented):
code seems very nice, other than these teensy weensy things. haven't reviewed the tgui because im too weak to handle it

MeggalBozale:
Changes made thank u :+1:

PR #17937, opened 2024-02-11
Change how sunflower and sunflower seeds share chems

M-Earthfire:
[hydroponics][bug][game objects]

## About the PR 

This PR changes sunflower and sunflower seeds. Instead of each one receiving their respective chems from `HYPsetup_DNA`, the seeds will receive up to 50% of the chems of the sunflower. 

## Why's this needed? 

Splices of sunflowers spawned with a full botany amount of chems and the seeds spawned from the sunflower spawned with their own amount as well. This means you would end up with double amount of chems per produce, and with jumbo sunflowers would end up with quintupled amount of chems (without the doubling of chems from having jumbo produce in the first place).

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Sunflowers now transfers up to 50% of their chems unto their sunflower seeds
```

### DIFF
--- a/code/obj/item/plants.dm
+++ b/code/obj/item/plants.dm
@@ -618,15 +618,14 @@ ABSTRACT_TYPE(/obj/item/plant/flower)
 		var/obj/item/reagent_containers/food/snacks/plant/seeds = locate() in src
 		if(seeds)
 			boutput(user, SPAN_NOTICE("You notice some [seeds] fall out of [src]!"))
+			seeds.set_loc(get_turf(target))
 			// that's the result if you want to keep 50% of the initial chems in the sunflower at the end.
 			// This accomodates partially for chems injected in the sunflower if some seeds are already missing
 			var/chem_amount = src.reagents.total_volume / (current_seed_count + initial_seed_count)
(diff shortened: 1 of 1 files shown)


code/obj/item/plants.dm
```diff
@@ -616,6 +618,14 @@ ABSTRACT_TYPE(/obj/item/plant/flower)
 		var/obj/item/reagent_containers/food/snacks/plant/seeds = locate() in src
 		if(seeds)
 			boutput(user, SPAN_NOTICE("You notice some [seeds] fall out of [src]!"))
+			// that's the result if you want to keep 50% of the initial chems in the sunflower at the end.
+			// This accomodates partially for chems injected in the sunflower if some seeds are already missing
+			var/chem_amount = src.reagents.total_volume / (current_seed_count + initial_seed_count)
+			seeds.reagents.maximum_volume = max(chem_amount, seeds.reagents.maximum_volume)
+			seeds.reagents.inert = 1 //let's not explode people by hitting them with a potassium/water sunflower
```
TobleroneSwordfish, line 625: But why not? You can already explode them by throwing potassium/water tomatoes at them.

PR #17938, opened 2024-02-11
Makes botany chem production scaling have diminishing returns

M-Earthfire:
[Balance][Hydroponics][Input wanted]

## About the PR 

This PR introduces `proc/HYPchem_scaling` and `proc/HYPfull_potency_calculation` and uses them to add diminishing returns to all chem production scaling of botany.

The graphs are given here: https://www.desmos.com/calculator/gy7tn43s6b

Before reaching 50 potency, produce will stlightly give more chemicals. After reaching 50 potency, the effectiveness of potency drastically decreases, with reaching 80u of chems as 100 potency, down from 100. With potency reaching infinity, the amount of added chemicals with cap out at 200.

## Why's this needed? 

Linear scaling of chems within botany meant that balancing botany chem production is neigh impossible and people were able to create ridicilous amount of chemicals within single items very quickly. Example of these are 300u+ morphine peanuts or quick 7x7 fireball seethers. 

These changes will keep the numbers in a reasonable range and will slow down the more ridicilous plants.

The changes proposed here will affect beginners and "normal" botany production not at all or only in a minor way, since diminishing returns start at 50 potency and only drastrically ramp up beyond 100 potency.

With this changes, balancing plants for lower statistics should become easier and it should create more design space when designing new botany chemicals or plants.

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Chemicals produced by plants show diminishing returns with increasing potency. See the minor logs for more information.
(+)Diminishing returns for botany scaling start at 50 potency, with 80% effectivity at 100 potency.
```


NibChocolateeny:
i think it is ok - but it may be very bad for omeaga weed (but I know it is joke plant)

testmerge first?

PR #17945, opened 2024-02-12
Fix reconcile_air null runtimes

flappybatpal:
## About the PR 
[09:34:25] pipe_network.dm,121: Cannot read null.volume
proc name: reconcile air (/datum/pipe_network/proc/reconcile_air)
  source file: pipe_network.dm,121
  usr: null
  src: /datum/pipe_network (/datum/pipe_network)
  call stack:
/datum/pipe_network (/datum/pipe_network): reconcile air()
/datum/pipe_network (/datum/pipe_network): process()
Machine (/datum/controller/process/machines): doWork()
Machine (/datum/controller/process/machines): process()
/datum/controller/processSched... (/datum/controller/processScheduler): runProcess(Machine (/datum/controller/process/machines), 0)

## Why's this needed? 
3000+ runtimes

### DIFF
--- a/code/modules/atmospherics/pipe_network.dm
+++ b/code/modules/atmospherics/pipe_network.dm
@@ -118,6 +118,8 @@ var/global/list/datum/pipe_network/pipe_networks = list()
 	ZERO_GASES(src.air_transient)
 
 	for(var/datum/gas_mixture/gas as anything in src.gases)
+		if (isnull(gas))
+			continue
 		src.air_transient.volume += gas.volume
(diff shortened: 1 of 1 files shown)


flappybatpal:
Is this correct? Does equalize_gases need the same treatment?

Tarmunora:
I feel like might be better to figure out how `null`s are getting into that list in the first place

ZeWaka (commented):
how are null gasses even getting here tho

PR #17947, opened 2024-02-12
Ice cream machine tg UI interface conversion

EleSeu:
[CATERING] [UI]

## About the PR 
Converts the Ice Cream machine UI to TGUI. It also makes possible to see what it's inside the beaker loaded into the machine, and makes it clearer that it is a possibility to make custom ice creams.
image

## Why's this needed? 
New UI to make it consistent with the rest of the game and easier to use. 

## Changelog 

```changelog
(u)Elettra
(+)New UI for the Ice Cream machine using TGUI.
```


tgui/packages/tgui/interfaces/IceCreamMachine.js
```diff
@@ -0,0 +1,117 @@
...
+        </Button>
+      </Flex.Item>
+    </Flex>
+  );
+};
+
+export const IceCreamMachine = (props, context) => {
+  return (
+    <Window
+      title="Ice Cream-O-Mat 9900"
```
ZeWaka, line 99:
```suggestion
      title="Ice Cream-O-Mat 6300"
```
i honestly didn't know this was actually a pun name until recently (ice-o-mat is a brand of ice crushing machines)
version will be downgraded to allow for future improvements to the UI

EleSeu: Cool, I didn't know that. I changed the name anyway

Tarmunora (approved):

tgui/packages/tgui/interfaces/IceCreamMachine.js
```diff
@@ -0,0 +1,117 @@
...
+  return (
+    <Section title="Standard Flavors" >
+      <Flex wrap>
+        {standardFlavors.map((flavor, flavorIndex) => (
+          <Button
+            key={flavorIndex}
+            align="left"
+            width="130px"
+            m=".1rem"
+            onClick={() => act("make_ice_cream", { flavor: flavor.name })}
```
TobleroneSwordfish, line 26: All other dispensers like this have the reagent names capitalised

tgui/packages/tgui/interfaces/IceCreamMachine.js
```diff
@@ -0,0 +1,117 @@
...
+              Make Custom Ice Cream
+            </Button>
+          </Flex.Item>
+          <Flex.Item>
+            <Button
+              mt=".5rem"
+              icon="eject"
+              disabled={!beaker}
+              onClick={() => act("eject_beaker")} >
+              {!beaker ? "Eject Beaker" : "Eject " + beaker.name }
```
TobleroneSwordfish, line 68: This should match how the buttons on chem dispensers work, with the container name and an insert option if there is no container

tgui/packages/tgui/interfaces/IceCreamMachine.js
```diff
@@ -0,0 +1,117 @@
...
+        <Flex wrap >
+          <Flex.Item>
+            <Button key="beaker"
+              mt=".5rem"
+              mr=".5rem"
+              icon="check" color="green"
+              disabled={!beaker || !beaker.totalVolume}
+              tooltip={beaker && !beaker.totalVolume ? "Beaker Is Empty" : ""}
+              onClick={() => act("make_ice_cream", { flavor: "beaker" })}>
+              Make Custom Ice Cream
```
TobleroneSwordfish, line 59: The capitalisation on these two buttons doesn't match

TobleroneSwordfish (changes requested):

tgui/packages/tgui/interfaces/IceCreamMachine.js
ZeWaka: i don't think this was addressed?

EleSeu: Using className="chem-dispenser__dispense-buttons" makes the buttons capitalised

PR #17963, opened 2024-02-14
Wizard spellbook TGUI

IPling:
## About the PR 
Changes:
* Added search bar
* Collapsible categories
* Removed junk code related to the wizard spellbook. (or atleast I hope it was junk code, the spellbook doesnt explode does it?)

wizard_spellbook_showcase

## Why's this needed? 
TGUI is good.
## Changelog 

```changelog
(u)IPingu
(+)Converts wizard spellbook interface to TGUI
```


tgui/packages/tgui/interfaces/Wizard_Spellbook.js
```diff
@@ -0,0 +1,171 @@
...
+      <Window.Content>
+        <Section title={owner_name+"'s Spellbook "}>
+          <Flex justify={"space-between"}>
+            <Flex.Item>
+              {"Spell slots remaining: "+spell_slots}
+            </Flex.Item>
+            <Flex.Item>
+              <Input
+                value={searchQuery}
+                placeholder={"search by spell name"}
```
MeggalBozale, line 61:
```suggestion
                placeholder={"Search by spell name"}
```
capitalization might be nice imo

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
```diff
@@ -0,0 +1,171 @@
...
+    spell_contents.push(spellbook_contents[category][spell][spell_data]);
+  }
+
+  return (
+    <Stack.Item>
+      <Section>
+        {(vr === 1) && (spell_contents[3] === 0) && (
+          <Dimmer>
+            <Box fontSize={1.5} backgroundColor={"#384e68"} p={2}>
+              spell unavailable in VR
```
MeggalBozale, line 137:
```suggestion
              Spell unavailable in VR
```
capitalization might look nicer

MeggalBozale:
LGTM on the code side, but I feel the menu could serve to be more compact in favor of user experience - while before you could see all the spells in a nice list, scrolling through various openable menus to have to select ones favorite loadout or compare spells sounds like a chore. Perhaps a list of all spells by name w/ a description in a hover tooltip as a top element could help this?

MeggalBozale (changes requested):
LGTM on the code side, but I feel the menu could serve to be more compact in favor of user experience - while before you could see all the spells in a nice list, scrolling through various openable menus to have to select ones favorite loadout or compare spells sounds like a chore. Perhaps a list of all spells by name w/ a description in a hover tooltip as a top element could help this?

image
Here's the old menu for comparison, which does this -- I don't really mean in list-view which takes up a good chunk of the menu, rather just a bunch of spell name-labeled buttons in sequence.

ZeWaka:
Could also compact cost to be on the line with the title of the spell also
**Rathen's Secret (1 Spell Slot)**

TobleroneSwordfish (changes requested):
- Some of the section colours are pretty hard to make out on dark mode, consider having specific light/darkmode colours like chat classes do.
image
- The search should probably auto-expand sections otherwise it's not very useful.
- Maybe use the spell icons somehow?

IPling:
> * Some of the section colours are pretty hard to make out on dark mode, consider having specific light/darkmode colours like chat classes do.
Yeah I can tweak that a tiny bit, gonna work on it tommorow.
> * The search should probably auto-expand sections otherwise it's not very useful.
I would LOVE to, but I dont think its quite possible with Collapsibles, since looking at the ObserveMenu that is very similar, it doesnt have that feature either with that project also having a lot more knowledgable people on it. Maybe there is a way, but I got no clue what it would be.
> * Maybe use the spell icons somehow?
That is a nother thing I would love, and there is a good example of how its possible with seed_vendor TGUI, but I could not quite figure out how that code works, and I cant exactly copy and paste since spell icons are animated yet fruit icons arent and it was a headache. I would need someone a bit more knowledgable in that area to get it to work.

IPling:
> Could also compact cost to be on the line with the title of the spell also **Rathen's Secret (1 Spell Slot)**

I had tried that, but I never liked having spell information in two different sections, because I feel as if the spell cost is as important as its description and cooldown meaning they should be paired. Altough that could be nullified if I were to place the spells description in the title of the section like a button that one way or a nother shows the description when clicked or hovered, that could be interesting to check out. Since the descriptions of spells only need to be read once, it becomes unneeded filler after the first read so it would be good to hide it first and only show when the player needs it.

IPling:
I have tried this now, but I am not entirely sure I enjoy it, altough that might be from me not knowing how to decorate it properly.
image

ZeWaka:
Why even hide the descriptions?

ZeWaka:
You could also put the cooldown in the title bar and have the cost be in the purchase button

IPling:
> Why even hide the descriptions?

To make it more compact, since descriptions really only need to be read once for a person to understand what the spell does and after the fact it seems pointless, allowing the user to more freely traverse the list of spells. Altough a player who knows the spells would probably just use the search bar, rather than cataloguing the list. Mostly just testing stuff out and seeing what sticks since that isnt implemented yet and as said I dont quite like it.

Valtsu0:
I'd prefer this having the default theme over ntos

Ideally we would have a wizard theme, but we don't (yet)

IPling:
Okay, I feel like the current NTOS theme and coloring is incredibly trivial to keep, since they feel much more like filler to replace the yet to be made wizard theme, and it would be more preferable to keep it a clean slate for when someone actually decides to make a wizard theme. It would be really great to have color since without it feels bare, but that is out of my expertise that I am not good at for now so maybe I or someone else in the future could make it.

Here is how it currently looks like, with some tweaks:
wizard_spellbook_showcase_test

IPling:
Updated the PR descriptions and some of the tweaks were adding a Divider to more clearly well... divide spells, and the cost is now stated in the purchase button, to make it a bit more compact. Decided to not go with a more compact description since it looks better that the description is more associated with the spell visually and hiding it didnt serve that much purpose when the search bar exists.

ZeWaka (approved):

PR #17977, opened 2024-02-15
Massive Airlock.dm reshuffle

TheTyrant25:
[CODE QUALITY]
## About the PR
The code in airlock.dm is all over the place. This PR does several things:

- Moves all the procs to the top, and all the subtypes (and their code) to alternate files.
- Adds comments to break up the enormous file when you do alt+click on the arrows in vsc to collapse the whole thing and make it possible to scroll through:
    - splits and rearranges the subtypes by parentage. All subtypes now moved into alternate files.
    - split and arrange the procs by function. These are still in airlock.dm

This should be reviewed with caution. Last time something like this was attempted, all of bolting and door opening code broke.

As it's just a reshuffle not a cleanup (This PR currently ONLY copies and pastes things around), it should be okay. In practise, things might break anyway due to the order being changed.

## Why's this needed?
The file is a mess. Several procs are defined twice, procs appeared both above and below the subtypes (and above the airlock definition itself???). This should make maintaining the code in the future easier to do.

## Changelog
```changelog
(u)Tyrant
(+)Massive backend reshuffling of the airlock code has taken place. report any bugs with door bolting/opening/closing.
```


ZeWaka (changes requested):
why not multiple files

TheTyrant25:
i did not think of that 👍

TheTyrant25:
should i do this for procs too or

ZeWaka:
yes but you can do that later

PR #18002, opened 2024-02-17
Airlock cycler packet integration

TheTyrant25:
[FEATURE] [WIKI]
## About the PR
Airlocks now have further packet interaction so that airlocks can be linked in game. Two new commands are added: `change_junction_id` and `change_entrance_id`

In addition, some cleanup of airlock code was done while I was at it. For clarity, `cycle_id` has now been renamed to `junction_id`.

Also, when a linked airlock closes other doors, it will now send a packet showing that it's trying to do so. The packet is not actually functional, as a real door closing packet would give away the access_code of all the doors in the junction. Instead, it's more like a signal to anyone sniffing that these doors are connected.

## Why's this needed?
Player built airlocks can now be linked (albeit through packets). It also makes linked airlocks send an additional signal, which indicates that they are linked.

## Changelog
```changelog
(u)Tyrant
(+)Airlocks can now be linked to each other in game by using packets. You can now set the junction_id and entrance_id of airlocks in the game.
```


MeggalBozale:
Oooh if cycling airlocks could be a thing on ABCU designs it would be so cool (Especially for builds designed to have airlocks for space/ocean)

TheTyrant25:
> Oooh if cycling airlocks could be a thing on ABCU designs it would be so cool (Especially for builds designed to have airlocks for space/ocean)

That sounds amazing. I'll see about making a PR for ABCU integration at a later point (i have no idea how the ABCU works rn)

TheTyrant25:
testing from third parties has revealed that the two commands work as intended :)

code/obj/machinery/door/airlock.dm
```diff
@@ -1692,12 +1698,19 @@ About the new airlock wires panel:
 
 	playsound(src.loc, src.sound_airlock, 25, 1)
 
-	if (src.cycle_id)
-		for (var/obj/machinery/door/airlock/D in cycling_airlocks[src.cycle_id])
+	if (src.junction_id)
```
pgmzeta, line 1701: Is there a good way to skip this or otherwise handle tug-of-wars from e.g. Kudzu opening doors? See #18024

TheTyrant25: tiny bit out of scope, but i'm planning another PR that will give them safety light integration. I'd like to just keep this PR as mainly packets based for now.

The plan for that is to have safety light being off 'disabling' the cycling integration, so that they don't get shut by others and also don't send a signal to shut others. And turning the light back on re enables the behaviour (it doesn't disconnect it from the junction, just disables that behaviour temporarily). The idea being that kudzu could then disable the safety/power on doors it takes over, via its vines being physically inside the door or sth.

TobleroneSwordfish (commented):
Not sure this is a good use case for packets, it seems very obtuse and impractical compared to just adding a field to the maintenance panel UI. If you want to integrate packets with airlock cyclers you could always make the actual cycling use packets :)

TheTyrant25:
I should mention that I'm completely shit at ui code and have no idea where to even start with doing something like that

code/obj/machinery/door/airlock.dm
```diff
@@ -862,12 +868,19 @@ var/global/list/cycling_airlocks = list()
...
+		for (var/obj/machinery/door/airlock/D in cycling_airlocks[src.junction_id])
 			// if they share entry id, don't close, e.g. double doors facing space.
-			if (src.cycle_enter_id && src.cycle_enter_id == D.cycle_enter_id)
+			if (src.entrance_id && src.entrance_id == D.entrance_id)
 				continue
 			D.close()
+		// we make the illusion of a close packet, not an actual one.
```
ZeWaka, line 877: why

TheTyrant25:
merge conflicts are to a point where it'd just be easier to start over on a fresh branch. to be fair the PR is 3 months old and i had to put down coding for a while

PR #18004, opened 2024-02-18
api array encoding for nested lists

Sovexe:
## About the PR 
* Attempts to correct the api query parameters syntax for nested lists

## Why's this needed? 
* API failures for filters that require array inputs treating the current inputs as strings

### DIFF
--- a/code/modules/goonhub/api/routes/apiRoute.dm
+++ b/code/modules/goonhub/api/routes/apiRoute.dm
@@ -31,7 +31,7 @@
 							if (length(src.queryParams[key][subKey]))
 								for (var/idx in 1 to length(src.queryParams[key][subKey]))
 									// encode in array syntax e.x. ?&filters[servers][1]=main3&filters[servers][2]=main4
-									.["[key]\[[subKey]\]\[[idx]\]"] = src.queryParams[key][subKey][idx]
+									.["[key]\[[subKey]\]\[[idx-1]\]"] = src.queryParams[key][subKey][idx]
 							else
 								.["[key]\[[subKey]\]\[\]"] = null
 						else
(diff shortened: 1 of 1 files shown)


code/modules/goonhub/api/routes/apiRoute.dm
```diff
@@ -27,7 +27,15 @@
 			if (islist(src.queryParams[key])) // Do we need to encode the value?
 				if (length(src.queryParams[key]) > 0)
 					for (var/subKey in src.queryParams[key])
-						.["[key]\[[subKey]\]"] = src.queryParams[key][subKey]
+						if (islist(src.queryParams[key][subKey]))
+							if (length(src.queryParams[key][subKey]))
+								for (var/idx in 1 to length(src.queryParams[key][subKey]))
+									// encode in array syntax e.x. ?&filters[servers][0]=main3&filters[servers][1]=main4
+									.["[key]\[[subKey]\]\[[idx-1]\]"] = src.queryParams[key][subKey][idx]
```
Wirewraith, line 34: Index is unnecessary here.

```suggestion
									.["[key]\[[subKey]\]\[\]"] = src.queryParams[key][subKey][idx]
```

Wirewraith (changes requested):

code/modules/goonhub/api/routes/apiRoute.dm
Sovexe: If an array has multiple elements you will lose all but the last one though

You will attempt to set multiple value to index "filters[servers][]"

Sovexe: image

PR #18022, opened 2024-02-21
Thrall Rework Mk. ∞

TobleroneSwordfish:
[BALANCE]
## About the PR 
Reworks vampiric thralls yet again:
- Thrall passive blood drain is halved on both classic and RP
- Thralls no longer have actionbar zombie arms
- Thralls now rapidly lose blood if they stray too far from their vampire master (>15 tiles)
    - The amount lost scales with distance
    - Will take blood from their actual bloodstream if they run out of points, rapidly killing them if they stray for too long
    - Big animated red tint and heartbeat sound effect whenever they lose blood this way to indicate Something Is Very Wrong
    - The lost blood spills all over the floor
    - There is a short grace period where the screen will flash lightly and slightly less worrying messages will play to let you know you're about to start losing blood.
- When a vampire uses enthrall on a living thrall to donate blood, they will also restore up to 200u of lost bloodstream blood (ie taking them out of deep hypotension)

Also moves thrall blood tracking off the mutantrace because I couldn't live with myself if I didn't, and makes blood dropped by thralls actually have their DNA.

## Why's this needed? 
Thralls were given zombie arms to try to make them fight rather than just going back to work as was often the case before. This worked, but lead to many players considering thralls to be near-useless and very unfun to play as due to their difficulty using weapons or tools. (also the arms never worked properly with special arm traits)
This change hopes to solve the problem of thralls not sticking around to help their vampire, while still letting them be useful and fun to play as. Enforced buddy system!

Drafted for feedback, still need to figure out things like range and drain rate. Oh and better feedback messages.

## Changelog 

```changelog
(u)LeahTheTech
(*)Vampiric thralls no longer have zombie arms.
(*)Vampiric thralls now rapidly lose blood and die if they stray too far from their vampire master.
(+)Vampiric thrall passive blood drain halved.
```

### DIFF
--- a/browserassets/html/traitorTips/vampiricthrallTips.html
+++ b/browserassets/html/traitorTips/vampiricthrallTips.html
@@ -2,6 +2,7 @@
 <div class="traitor-tips">
     <h1 class="center">You have been revived as a thrall!</h1>
     <p>You feel an <em>unwavering loyalty</em> to your new master! (As such, <em>do not reveal their identity</em> or otherwise act against their best interests.)</p>
-	<p>You will slowly lose blood points over time. Your max health will decrease as blood points are lost. You can regain blood points by drinking the blood of humans or taking an additional donation from your master.
+	<p>You <em>MUST</em> stay close to your master! Your new unlife is fragile and straying too far from their power will result in your swift demise.</p>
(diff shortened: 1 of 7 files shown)


frawhst:
as the author of Thrall Rework Mk. ∞-1, i’m happy to try this

AXLFAN2000:
Really like these ideas!  Excited to see this in action

PR #18029, opened 2024-02-23
Correct unit on power generation for crew credits

MeggalBozale:
[QoL] [UI]
## About the PR 
Changes the unit of "watts" to "watt-hours" for the unit of total power generated in the credits.

image
ran a quick shift (chui broke and i skipped atmos, weh) and generated a few MJ overall - which comes down to looking like this. More extreme power generation setups of course would go to upwards of mWh or gWh

## Why's this needed? 
The former unit is an expression of power - implicitly in joules/second. This isn't the correct unit to use for the total power! watt-hours are an expression of energy, and the credits are showing the total energy produced -- so this makes sense to use. Could have also just been the overall joules, and still can be if wanted -- just going off a suggestion here.

### DIFF
--- a/code/datums/crew_credits/crew_credits.dm
+++ b/code/datums/crew_credits/crew_credits.dm
@@ -335,7 +335,7 @@
 	src.score_tab_data[SCORE_TAB_SECTION_ENGINEERING] = list(
 		list(
 			"name" = "Power Generated",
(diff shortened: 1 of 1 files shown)


MeggalBozale:
reject size/XL return to size/XS

TobleroneSwordfish (approved):
Ohh yes, this has bugged me for so long but I never got around to fixing it. `Wh` seems like a sensible unit 👍

PR #18034, opened 2024-02-23
Clarion ranch renovation

alicars:
## About the PR 
Renovates the Clarion ranch to make it a bit more useable, if still squished.  
clarion ranch strong dmm
_(With these test graphics, Dolly has been rendered into a crab. And the grinder, an ATM ...)_

Adds the chicken flusher. Currenly this does NOT add a botany chute to this flusher, as it's exit is easily overlooked outside the freezer in maints. It does not really seem like there is a free tile anywhere inside the freezer itself or kitchen to place a produce flusher exit. So I just aimed the chute so the chickens would land beside the back door. I would be happy to rework this chute in any way, including just removing it or expanding it. It is a nice feature for ranches to be able to dispose of unwanted chickens, but a small ranch could be ok without it.  

worlds worst chute
_(The chickens will land on the circled tile)_

This PR touches secret content so I have no idea about some testing things or if the incubator and grinder and everything are working ok.

Possibly controversial is my placement of the interdictor hiding in the topmost bush. I put it there so that people would run into the ranch and bonk their head on it more times. _(You cannot see this in my screenshot for some reason oops)_

## Why's this needed? 
The current Clarion ranch is not that great, for lots of reasons. Though it is a bit funny and can encourage players to renovate, actually creating advanced chickens there is a challenge. For instance, the simple problem of only having 1 incubator leads to big bottlenecks in branching evolution chickens, especially with unfavourable rooster luck.

But still, the amount of tiles currently assigned to the ranch really isn't THAT bad for a single rancher. We can make it work!  

This is my first PR so I'm sure I have made 100,000 mistakes apologies in advance!!

## Changelog 

```changelog
(u)crunchwarp
(*)The Clarion ranch has emerged from its chrysalis at last, transformed...!
```


cheekybrdy:
Considering that the beer keg is meant to be a bartender item, and probs shouldn't be locked to a chef area. It may be your best bet for a place to put the outlet.

alicars:
clarion ranch pr recreation
Sort of recreated on a live server to help illustrate better

TobleroneSwordfish (changes requested):
Looks like you've accidentally remove the detective spawn point, the detective's office camera and the unique ranch camera. See the diff render for details.

alicars:
Oops, I believe the cams and det spawn point should be back in now. Thank you!!

TobleroneSwordfish:
I cast curse of merge conflict upon #18169

PR #18046, opened 2024-02-26
Removes Karmic Punishment from Ahelp & prayer

MintyPhresh:
## About the PR 
Removes four lines of code that reduces a player's karma by 1 when they Ahelp or pray.

## Why's this needed? 
You can theoretically be sent to hell if your karma gets to -200, and Ahelps contributing to eternal punishment seems opposed to current administrative philosophy. Praying to god doesn't seem like the sort of thing that'd earn you a seat in the hot place, either.

### DIFF
--- a/code/modules/admin/adminhelp.dm
+++ b/code/modules/admin/adminhelp.dm
@@ -116,7 +116,8 @@
 	var/in_chapel = 0
 	if(istype(get_area(client.mob), /area/station/chapel))
 		in_chapel = 1
-
+	if (client.mob.mind && client.mob.traitHolder?.hasTrait("atheist"))
+		src.add_karma(-1)
 	var/is_atheist = client.mob.traitHolder?.hasTrait("atheist")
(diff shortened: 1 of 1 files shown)


Chatauscours:
Maybe praying if your character has the atheist trait should still give you some negative karma, for the fun of it (considering that atheist prayers already give out very silly descriptions)

MintyPhresh:
If you code that I'll add it to the PR (lazy)

code/modules/admin/adminhelp.dm
```diff
@@ -120,9 +117,6 @@
 	if(istype(get_area(client.mob), /area/station/chapel))
 		in_chapel = 1
```
TobleroneSwordfish, line 119:
```suggestion
	if (client.mob.mind && client.mob.traitHolder?.hasTrait("atheist"))
		src.add_karma(-1)
```

PR #18047, opened 2024-02-26
German Accent(fixed pr)

klushy225:
## About the PR  

(git broke- I'm adding the labels now- not sure why they didn't go up with the pr)

This pr seeks to add a german accent to the game, which will appear as a cute litte pretzel icon in the language traits section. I am a german speaker, so I tried to make this as accurate as possible. However, German is only my second lanuage so I had a native speaker over on discord help me a bit. A pretty big hurdle with this is the MANY different dialects of German, and how they all sound very different from each other. They are so extreme that alot of the times southern German shows will be subbed for Northern audiences so they can understand them. I chose to go with Hochdeutsch, as that is the standard form of German understood by most German speakers.

This functions similarly to how the scots dialect works, with a few differences. Similarly to the scots code, there is a list of words that replace certain words in the scentence. These are cognates that are similar to english and mostly understandable by purely english speakers, with a other few common(and perhaps a little stereotypical) words thrown in there. The key difference is that the modified tokens are run into the parser, so they are spelled incorrectly but are pronounced correctly if read like English.

## Why's this needed? 

I think there should be more accents in the game, and I recieved positive feedback from multiple people when I demo'd my little test of this when I was bored. I was pretty dissapointed when I first started playing a few years ago and saw no German accent.

## Changelog 

```changelog
(u)Klushy225
(*)Added a german accent to the game, which makes the player sound very german.
```


klushy225:
i think my vpn is screwing this up, or git isn't properly configured- I'll have to fix it when I get home.

PR #18060, opened 2024-02-28
SyndieChemiCompiler_Bugfix

Nitre88:
[bugfix]
## About the PR 
Fixes an issue with the syndicate ChemiCompiler where it could cause runtimes and could not use the pill, drop, or vial functions

## Why's this needed? 
It's currently not able to function as a ChemiCompiler without these important functions. Also ruined my traitor round :(

### DIFF
--- a/code/modules/chemistry/chemicompiler_core.dm
+++ b/code/modules/chemistry/chemicompiler_core.dm
@@ -51,6 +51,7 @@
 	var/heading = "don't touch!!!"
 	var/minStored = 1
 	var/maxStored = 6
+	var/list/utilReservoir = list("11", "12", "13",)
 
 /datum/chemicompiler_core/New(datum/holder)
 	..()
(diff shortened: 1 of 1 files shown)


Nitre88:
did a bit of last minute testing drip isn't working fixing

code/modules/chemistry/chemicompiler_core.dm
```diff
@@ -393,7 +394,7 @@
 	if(source < minReservoir || source > maxReservoir)
 		throwError(CC_ERROR_INVALID_SX) // Invalid source id.
 		return
-	if(target < minReservoir || target > maxReservoir + 3)
+	if(target < minReservoir || target > maxReservoir && !target == utilReservoir)
```
Tarmunora, line 397: `utilReservoir` is being defined as a list though - I don't see how this code would actually work?

Tarmunora (changes requested):

code/modules/chemistry/chemicompiler_core.dm
Nitre88: thank you and fixed

code/modules/chemistry/chemicompiler_core.dm
```diff
@@ -51,6 +51,7 @@
 	var/heading = "don't touch!!!"
 	var/minStored = 1
 	var/maxStored = 6
+	var/list/utilReservoir = list(11, 12, 13,)
```
ZeWaka, line 54:
```suggestion
	var/list/utilReservoir = list(11, 12, 13)
```

PR #18064, opened 2024-02-28
Priority role terminal program for HoP and captain

TobleroneSwordfish:
[FEATURE]
## About the PR 
Adds a `RoleControl` terminal program to communications consoles that allows anyone with ID card modification access (Captain, HoP and HoS by default) to set one role as "priority" for latejoins:
image
The role will appear highlighted in yellow on the latejoin menu:
image
Setting the priority role triggers a PDA message to the command mailgroup:
image

TODO:

- [ ] Figure out some kind of reward for joining as a priority role: spacebux/credit bonus?

## Why's this needed? 
I think Sord suggested it? More things for captain/HoP to do, lets people have an idea of what departments really need backup when they're looking to join a round.

## Changelog 

```changelog
(u)LeahTheTech
(*)Added a new program to communications consoles: RoleControl. Requires ID computer access and allows you to set a priority role to be highlighted on the latejoin menu.
```

### DIFF
--- a/code/datums/controllers/job_controls.dm
+++ b/code/datums/controllers/job_controls.dm
@@ -990,7 +990,8 @@ var/datum/job_controller/job_controls
 				alert(usr, "Could not find a savefile with that ckey!.")
 			src.job_creator()
 
-/proc/find_job_in_controller_by_string(var/string, var/staple_only = 0)
+///Soft supresses crash on failing to find a job
+/proc/find_job_in_controller_by_string(var/string, var/staple_only = 0, var/soft = FALSE, var/case_sensitive = TRUE)
 	RETURN_TYPE(/datum/job)
 	if (!string || !istext(string))
 		logTheThing(LOG_DEBUG, null, "<b>Job Controller:</b> Attempt to find job with bad string in controller detected")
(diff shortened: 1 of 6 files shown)


IPling:
I like it, gives late-joiners indication that a certain jobs role is not being fullfilled by the current members by either them being an antag, incompetent or afk. Since otherwise the late-joiners can already see if a department is under-staffed or not.

With the rewards though... giving some actual reward for picking a priority job makes me feel like people will pick it for the reward and just dip to do the job they might have actually wanted to play. Taking up a slot in the job market and not fulfilling the role.

In-place of rewards my recommendation would be to give some sort of context to late-joiners why this role is being prioritized. Some text that is shown when the prioritized job is hovered over with a mouse, explaining that:
1. We need engineers to rebuild x place. (Medical, Bar, etc.)
2. We need security officers to take down x  (Changeling, Vampires, etc.)
3. We need QM's to buy a new x. (Cloner, singulo, etc.)
Either the text would be freely written by the player (which would make me worry about players using it wrongly) or having it more restrictive like - ''We need (job) to (do this)''. Similar to AI laws.
Since they would have context for what is currently going on in the station + having a goal, my hypothesis is it would give players actual motivation to pick it compared to bribing them to pick the role, meaning the players that would actually pick it are much more likely to accomplish the reason for why the job was made into a priority one.

klushy225:
I like this alot, but I do agree that people might just cheese the rewards. People usually try and join in understaffed departments to help anyways, so i doubt we'd need a reward incentive.

PR #18085, opened 2024-03-02
Eyebots now come with a portable intercom

ThatFiveGuys:
## About the PR 
Eyebots now come with a tiny handheld intercom that has broadcasting and listening to false by default

## Why's this needed? 
Gives eyebots more utility in the spying on people department which sounds like fun

## Changelog 

```changelog
(u)Ikea
(*)Eyebots now come with a portable intercom.
```


Kyle2143:
IMO add a sprite change when it's listening. Like a red "recording" light somewhere visible on it. So at least people know when they're being spied on.

Tarmunora (changes requested):

PR #18091, opened 2024-03-02
Makes NTSC take up a security officer slot

TobleroneSwordfish:
[BALANCE]
## About the PR 
Also refactors job stuff behind the scenes to allow this.

## Why's this needed? 
Too many sec slots, NTSC shouldn't be a free additional officer if there's a whitelisted player online.

## Changelog 

```changelog
(u)LeahTheTech
(*)NTSCs now take up a security officer slot.
```

### DIFF
--- a/code/datums/controllers/job_controls.dm
+++ b/code/datums/controllers/job_controls.dm
@@ -60,27 +60,27 @@ var/datum/job_controller/job_controls
 		dat += "<b><u>Job Controls</u></b><HR>"
 		dat += "<b>Command & Security Jobs</b><BR>"
 		for(var/datum/job/command/JOB in src.staple_jobs)
-			dat += "<a href='byond://?src=\ref[src];AlterCap=\ref[JOB]'>[JOB.name]: [countJob("[JOB.name]")]/[JOB.limit]</A> <a href='byond://?src=\ref[src];Edit=\ref[JOB]'>Edit</a><BR>"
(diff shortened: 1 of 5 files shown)


Chatauscours:
I'll be honest, I think in it's current state the PR will do more damage than it will prevent.

The main current issue is on goon 3, during lowpop hours (9:00 - 17:00 GMT, 20 - 30 pop), a third or even a half of the station is security (1 HoS, 4-5 secoffs, 1 detective, 1 NTSC and at least 1 secass for a total pop of ~25 people and only 1 antag). And I'm the first one to agree that this is a considereable issue.

However, whilst this PR will of course reduce the amount of sec slots by  1 on such rounds, we'll still remain with 9 security for 25 crew, ultimately not changing much. 
And, even worse, on highpop rounds (18:00 - 1:00 GMT particularly on the week-end, 55-70 pop) where every sec slot is filled at roundstart with usually at least one secoff who'll cause more issues than they'll solve and rampant self antagging, this PR will prevent the already ovewerworked HoSii and the rest of security from having someone (aka an NTSC) who's judgement they can trust, in particular with the delicate concerns of roguish secoffs and self antagging crew.

A solution I would propose is making the amount of sec slots scale with pop. In my opinion, a reasonable total (HOS, secoff, det, ntsc and secass included) amount of security ranges from 10 to 20ish % of total crew pop, and letting slots round up to the next highest . It's not the perfect solution, and there is an amount of other issues with security that would need to be solved, in particular relating to secasses, but in my opinion it would be a good start.

TobleroneSwordfish:
To clarify: this PR as written won't prevent an NTSC from joining if the team is full, it'll just lower the number of normal secoffs allowed to join after an NTSC does. Kind of janky but I think this way makes sense?
I will also look into scaling security job limit with population because I think that would be a good change alongside this, yep.

TobleroneSwordfish:
#18096

PR #18099, opened 2024-03-03
Fixes #17275: Reagents applying to overlays on turfs which cause errors in lighting

### LINKED ISSUE #17275: pizza overlays fucked
### Labels

[BUG][MAJOR]

### Description

someone turned everything in areas of the station into pizza but the pizza material overlay is overlaying over everything on the tile not just the floor/walls/stuff like if i walk on to the tile im under the overlay and also its dark as fuck and no one can see anything

### Steps to reproduce

1. pizza grenade turn stuff into matsci pizza
2. trapped in pizza hell (dark and horrific)
3. beg mentors for answers (there is no answers) 
4. make bug report

### Expected Behavior

stuff to become pizza textured but not trap me in this impermeable darkness

this could be a byond glitch/admin fun/intended feature i genuinely dont fucking know but it looks like a bug so im reporting it as a bug will attach screenshots in the github comments in postround

Reported by: Joeled
Round log date: 2023-12-11-01-37
Reported on: 2023-12-11 02:37:26
Map: COGMAP

mloccy:
[BUG][CATERING][CHEMISTRY]

## About the PR 
Fixes #17275 by not applying reagents to overlays.

## Why's this needed? 
Not sure reagents should be applied to overlays so it seems like a good fix.

### DIFF
--- a/code/modules/chemistry/Chemistry-Holder.dm
+++ b/code/modules/chemistry/Chemistry-Holder.dm
@@ -714,6 +714,9 @@ proc/chem_helmet_check(mob/living/carbon/human/H, var/what_liquid="hot")
 								turf_reaction_success = 1
 								.+= current_id
 						if(isobj(A))
+
+							if (istype(A, /obj/overlay))
+								continue
 							// use current_reagent.reaction_obj for stuff that affects all objects
 							// and reagent_act for stuff that affects specific objects
 							if (!current_reagent.reaction_obj(A, current_reagent.volume*volume_fraction))


TobleroneSwordfish (changes requested):
Better to add an empty `reagent_act` definition to `/obj/overlay` to catch other cases where chems can be applied to overlays. In general they should probably just be merged into `/obj/effects` but that's a different PR.

mloccy:
> Better to add an empty `reagent_act` definition to `/obj/overlay` to catch other cases where chems can be applied to overlays. In general they should probably just be merged into `/obj/effects` but that's a different PR. 

Do you mean in addition to the lines added in the PR or instead of? Because the issue is in reagent.reaction_obj which is called on overlays if we remove the type check. We could also do a type check for overlays in the pizza reagent impl of reaction_obj instead.

mloccy:
> > Better to add an empty `reagent_act` definition to `/obj/overlay` to catch other cases where chems can be applied to overlays. In general they should probably just be merged into `/obj/effects` but that's a different PR.
> 
> Do you mean in addition to the lines added in the PR or instead of? Because the issue is in reagent.reaction_obj which is called on overlays if we remove the type check. We could also do a type check for overlays in the pizza reagent impl of reaction_obj instead.

Maybe a better fix could be putting the check in /atom/proc/setMaterial

TobleroneSwordfish:
Ehh this is fine for now actually

PR #12926, opened 2023-01-29
Reworks and adds a few flags (the fabric kind)

DisturbHerb:
[MAPPING][OBJECTS][FEATURE][REWORK]

## About the PR 
This PR re-organises `obj/decal/wallsign/flag` into its own type, `obj/decal/poster/flag`, and introduces a new `obj/item/flag` that can be placed on walls much like fabric banners (#5036). Flags can also now be taken down from walls by attacking them with a free hand. All instances of flags in the public repository have been updated with the new file path. I hope.

Also, more flags. Thanks to Geonightman#2706 on Discord for making the icons for these.*
flegs

*this has been updated like numerous times sorry (accurate as of 1 feb 2023)

I drew the folded flag states and the blank flag stuff and I don't love them, feedback will be appreciated.

All of the flags, save for the lesbian pride ones, are currently player-unobtainable.

## Why's this needed? 
Flags are nice and I wanted them for a thing.

```changelog
(u)DisturbHerb + Geonightman
(+)Flags can now be hung up and removed from walls.
(+)Adds a bunch of pride flags.
```

PR #18104, opened 2024-03-04
Makes pride flags craftable

MintyPhresh:
## About the PR 
Pride jumpsuits, when not being carried or worn, can be cut with snipping tools into a flag of the respective type.

Also adds flags for the two MLM designs, seen below.
image

Big thanks to Geonightman on discord for the flag sprites!

## Why's this needed?
Makes most of the pride flags added in #12926 actually obtainable. Adds a relatively easy way to decorate rooms. 

## Changelog 

```changelog
(u)mintyphresh, Geonightman
(+)Pride jumpsuits can be cut up into flags.
```


code/obj/item/clothing/uniforms.dm
```diff
@@ -17,6 +17,7 @@
 	burn_possible = TRUE
 	health = 10
 	var/team_num
+	var/obj/item/cutting_product
```
TobleroneSwordfish, line 20:
```suggestion
	var/cutting_product
```
Use doc comments rather than types to indicate the kind of type you mean, this looks like an instance var.

TobleroneSwordfish (approved):
Sprites look good, one minor code nitpick.

code/obj/item/clothing/uniforms.dm
MintyPhresh: Thanks for the input!!

PR #18112, opened 2024-03-04
BonePipeshot

Cindertroy:
[FEATURE]
[GAME-OBJECTS]

## About the PR 

Adds a craftable bone-shot bullet which is set at 9 damage with 3 bullets fired per shot with zero armour penetration 

## Why's this needed? 

Mostly a gimmick item, but it does give the "Bits of bone" a use, as I dont believe they currently have a use (correct me if im wrong).

It is constructed with 2 bits of bones and a pipehull, each shot shoots 3 projectiles which have a short range, about 1-3 tiles. Each projectile would do is set at 9 damage without armour penetration

It shouldn't affect much other stuff though I am very open to changing any of the values as they're all pretty much arbitrary 

I do think this would be a good addition and pretty funny to use

## Changelog 

```changelog
(u)Cindertroy
(+)Added a bone pipeshot, crafted with 2 "bits of bone"
```

### DIFF
--- a/code/modules/projectiles/bullet.dm
+++ b/code/modules/projectiles/bullet.dm
@@ -724,6 +724,26 @@ toxic - poisons
 	implanted = /obj/item/implant/projectile/shrapnel
 	damage = 10
 
+/datum/projectile/bullet/improvbone
+	name = "bone"
+	sname = "bone"
+	icon_state = "trace"
+	dissipation_delay = 1
+	dissipation_rate = 3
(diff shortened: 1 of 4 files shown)


TobleroneSwordfish:
Should have comedic bone rattling impact sound 💀

Cindertroy:
that sounds funny im going to try doing it

Tarmunora (approved):

PR #18115, opened 2024-03-05
Connects the north catalytic substation on nadir with the newly added PTL

Knipje:
[STATION-SYSTEMS]

## About the PR 
This PR connects the north catalytic substation on Nadir with the PTL in order to allow it to access this part too, consequently it also combines all three engine output wires into one, making them all accessible through the same wire.

## Why's this needed? 
Only the south substation was connected previously which is a weird inconsistency as it allows all engines to be connected except for the north ones.

## Changelog 

```changelog
(u)Knipje
(+)Nadir North Catalytic Substation connected to PTL.
```

### DIFF
--- a/maps/nadir.dmm
+++ b/maps/nadir.dmm
@@ -120266,7 +120266,7 @@ rJg
 lxn
 wXk
 ruo
-wXk
+hJZ
 udE
 gLr
 sgU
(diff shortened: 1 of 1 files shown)


TheTyrant25 (approved):

TobleroneSwordfish:
Very big long wire with not much purpose? Iunno, doesn't seem very realistic

TheTyrant25:
tbh i don't think the south catalytics should connect to PTL anyway, they're pretty much nadir solars. But if one's connected, the other one should be too

Knipje:
good point honestly, itd probably be better for neither to be connected

ZeWaka:
Seems like this PR is no longer wanted?

PR #18116, opened 2024-03-05
Disguised law module

JORJ949:
## About the PR 
Adds a freeform law module that looks like an Asimov/NT/Robocop law to Head of Staff traitor uplinks.

## Why's this needed? 
The current silicon traitor item, the AI laser camera module sucks and I think a better law based traitor item for heads of staff would be great. 

## Changelog 

```changelog
(u)JORJ949
(*)Added Disguised law module to head of staff traitor uplinks.
```


Retrino:
normal ai modules are mechscannable, so id assume this item inherits that behavior? if so, it should probably be only scannable via syndi device analyzers, if at all

JORJ949:
It is only scannable by syndicate analysers

JORJ949:
> normal ai modules are mechscannable, so id assume this item inherits that behavior? if so, it should probably be only scannable via syndi device analyzers, if at all

I already considered this and it already can only be mechscanned by the syndicate device analyzer.

amylizzle (approved):
Cool idea. Code looks fine. There's an argument that it should just be a subclass of the freeform but I think that'd probably introduce a bug with the "have the freeform on you at the end of the round" objective unless you compensated for that.

JORJ949:
> Cool idea. Code looks fine. There's an argument that it should just be a subclass of the freeform but I think that'd probably introduce a bug with the "have the freeform on you at the end of the round" objective unless you compensated for that.

I forgot that was a freeform objective when writing this so lets just say I considered that by not making it a subclass :)

amylizzle:
oh no

a tip on git: `git merge upstream/master`

JORJ949:
Oh dear

JORJ949:
Lesson learnt. Dont click buttons.

PR #18126, opened 2024-03-06
Enables atmos archiving

Laboredih123:
[ATMOS][EXPERIMENTAL]

## About the PR 
Enables atmos archiving, making atmos behave equally in all directions as you would expect, instead of tending to prefer going some direction. Previously disabled for performance concerns, I believe atmos should be performant enough for this to be on again.

## Why's this needed? 
Consistency in atmos interactions is good, one could say.


TobleroneSwordfish (changes requested):
I said this on discord, but just for the github record: since this is a flag that disabled for performance reasons it should be profiled with the `Disco Inferno` test scenario before being re-enabled.

Laboredih123:
ok will do

Laboredih123:
i infact, did not get around to it. will soon

Laboredih123:
didnt then either lols lmaos

Laboredih123:
should be super cheap nonetheless

Laboredih123:
oh yeah this waiting on atmos optimisations pr

PR #18133, opened 2024-03-07
Adds Blinds to the cog1 Bar

cheekybrdy:
## About the PR 
Adds blinds to the cog1 bar, and makes the blast door between bar and kitchen independent and bartender controlled.

## Why's this needed? 
Cog1 bar office lacks privacy completely from the south due to emergency A and has north privacy controlled by another job for their private office, means the room has potential for a crime room without construction


amylizzle:
I think you might want to clean your branch and redo this, the map diff is very large and touches all of botany and reroutes some disposals pipes.

TobleroneSwordfish (changes requested):
Yep, this is reverting several other PRs by the look of it. Not sure how it's doing that in only 29 lines but 🤷

cheekybrdy:
remaking due to branch bork

PR #18136, opened 2024-03-07
Adds Icons to wizard spellbook UI

IPling:
## About the PR 
Adds icons for each spell (Except Soulguard since it doesnt have a spell icon) to the wizard spellbook UI I made, since when making it I had no clue how to actually add them.

Preview image: (Made the window a bit bigger to show more spells for the preview.)
image

## Why's this needed? 
Visually looks much more pleasant and was previously suggested.

## Changelog 

```changelog
(u)IPingu
(+)Adds Icons to spells in the wizard spellbook
```


ZeWaka:
X

Tarmunora:
image
when spell is ineligible in VR, top couple pixels of the icon for the next spell down is greyed out too

IPling:
> when spell is ineligible in VR, top couple pixels of the icon for the next spell down is greyed out too
Ah yeah, must have forgotten to test in VR. Probably from the constant margins I had to do to get it just right, since sections really dont like it when I add images to em.
Will see what I can do tomorrow to fix that.

IPling:
Looks good now.
image

Tarmunora (approved):

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
```diff
@@ -105,32 +105,54 @@ const Spell = (props, context) => {
...
-            <Flex>
-              <Button
-                backgroundColor={"green"}
-                disabled={spell_slots < spell_contents[1] || purchased}
-                onClick={() => { setPurchased(true); act("buyspell", { spell: spell }); }}
-              >
-                {If_Purchased_Text(purchased, spell_contents[1], spell_slots)}
-              </Button>
+          title={
+            <Flex justify={"space-between"} align={"end"}>
```
mordent-goonstation, line 125:
```suggestion
            <Flex justify="space-between" align="end">
```

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
```diff
@@ -105,32 +105,54 @@ const Spell = (props, context) => {
...
+              <Flex.Item>
+                <Flex>
+                  <Flex.Item>
+                    {!!spell_contents[4] && (
```
mordent-goonstation, line 129:
```suggestion
                    {!!spell_contents[4] && (
```
This is definitely getting too unwieldy for a tuple of data and should be migrated to being an object with key/value pairs.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
```diff
@@ -105,32 +105,54 @@ const Spell = (props, context) => {
...
+                      <Image
+                        pixelated
+                        mt={-2}
+                        height="38px"
+                        width="38px"
```
mordent-goonstation, line 134: Justification for this size? Is this the resolution of the icons? At a glance they seem to be 32x32.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
```diff
@@ -105,32 +105,54 @@ const Spell = (props, context) => {
...
+                        src={`data:image/png;base64,${spell_contents[4]}`}
+                      />
+                    )}
+                  </Flex.Item>
+                  <Flex.Item align={"end"}>
```
mordent-goonstation, line 139:
```suggestion
                  <Flex.Item align="end">
```

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
mordent-goonstation, line 125: You could probably structure this as a single `<Flex>` making use of `grow`. Something like:
```
<Flex align="end">
  {!!spell_contents[4] && (
    <Flex.Item>
      <Image ... />
    </Flex.Item>
  )
  <Flex.Item grow fontSize={1.25}>
    {spell}
   /Flex.Item>
   Flex.Item>
    <Button ... />
   /Flex.Item>
</Flex>
```

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
```diff
@@ -105,32 +105,54 @@ const Spell = (props, context) => {
...
+                    <Box fontSize={1.25} ml={1} mt={1}>
+                      {spell}
+                    </Box>
+                  </Flex.Item>
+                </Flex>
+              </Flex.Item>
+              <Flex.Item>
+                <Button // Putting this into buttons={}, breaks it, somehow.
```
mordent-goonstation, line 147: Very curious about this. Could you expand on what you mean by "breaks"?

mordent-goonstation (changes requested):

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
IPling: Its the size of the icons, 32x32 felt too little to notice the details of the icon and 38x38 felt just nice to see the details but not being too big.
Here is a comparison (The alignment for 32x32 is a bit wrong but that isnt the problem):
image
image

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
IPling: The button would be completely mis-aligned and be stuck on the line (couldnt do anything to move it up a few pixels) that separates the title and contents, it did honestly look pretty good, but it made the UI look a bit messy and was hard to locate the purchase button + half of the button that was under the line couldnt be pressed.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
IPling: Just curious, why should I do this? I don't think there is any difference and its the first time I'm seeing an apprehension towards {}

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
IPling: I would love that, I had tried a lot mind you at the very start of the creation of the spellbook TGUI and I even had a help of a dev (Leah) and a nother person (Dr. Romayne) from the discord, yet we couldnt find a workable solution that didnt cause errors, so instead we did Dr. Romaynes solution that I think they used for their TGUIification of the pump control computer where they had the same problem.

Altough if this doesnt require a re-making of `spellbook_contents` and there is instead a simpler solution that can be fitted just for `spell_contents` I would greatly appreciate that, since I **could** try, but I doubt I have the expertise to make something that doesnt look even uglier.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
IPling: Your right, didnt know the `grow` property could be used like that.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
mordent-goonstation: I get that, but you're distorting a 32x32 image and it's having to make-up content for it (as the scale isn't an integer one). Note how in your example above the orb is slightly vertically stretched in the 38x38 version.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
mordent-goonstation: Mostly just a case of "it's the preferred syntax for representing static strings in JSX". Same sort of thing for `foo={true}` vs. just adding `foo` (without anything else) for boolean true things.

Not requiring this change, just encouraging the convention.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
mordent-goonstation: Curious what kind of errors you were seeing! This should be a pretty straightforward and error-free migration. I'm okay with it being done as follow-up work, I suppose. 5 fields that have to happen to be in there in a certain order does not make for maintainable code.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
IPling: Ah that makes sense, I am fine with that convention then and I will try to keep it in mind.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
IPling: Line was removed in the ''Condenses 2 flexes, into 1'' commit.

tgui/packages/tgui/interfaces/Wizard_Spellbook.js
IPling: Hmm, that's fair. I didn't think about how its actually distorting the image, I am fine with reverting it to 32x32 then.

IPling:
Updated the preview image to match the 32x32

mordent-goonstation:
Needs a changelog. I want to pull this one down locally and poke at it a bit.

IPling:
Changelog made.

PR #18140, opened 2024-03-07
Componentizes putting hats on ghostdrones and AI cores, extends support to mechcomponent storages

CalliopeSoups:
[feature]
## About the PR 
I've taken some code used for ghostdrones and mangled it to allow people to customise their mechcomp (demon) robot with hats and wigs. Currently it's just done with a click but I can make it specific to an intent, or see about letting you weld them on so they're not stolen. Do let me know if any code is bloated or vestigial.

dreamseeker_53O2dDUyM7

Along with adding the component to something you want to be able to put hats on, you can now also set a hat_offset_y and x var there, and in a hat's code in order to correct its position. With this you can set the hat to the right place on the object, then adjust even more for things like wigs and hats with inhand sprites that aren't centred in frame. 

## Why's this needed? 
Mechcomp cabinets/frames are already personified a lot with movement components and dialogue, this allows for more customisation and fun. 

The component itself will ensure that adding behavior for putting hats onto things that isn't a player with slots will be easier to implement. 

## Changelog 

```changelog
(u)CalliopeSoups
(+)Hats and wigs now fit snugly onto mechcomp cabinets and frames. This changes the backend for placing hats on things like ghostdrones, report any bugs if found.
```

### DIFF
--- a/code/datums/components/hattable.dm
+++ b/code/datums/components/hattable.dm
@@ -1,0 +1,110 @@
+TYPEINFO(/datum/component/hattable) // Take a walk through my TWISTED mind.... I'm sorry
+	initialization_args = list(
+		ARG_INFO("death_remove", DATA_INPUT_TEXT, "Remove component and signals on death?", FALSE), // Use this for things that don't revive, or explode on death
+		ARG_INFO("free_hat", DATA_INPUT_TEXT, "Should the parent get a free hat?", FALSE), // For AIs
+	)
+
+/datum/component/hattable  // Compatible stuff should be given hat_offset x and y vars in their own file
+	dupe_mode = COMPONENT_DUPE_HIGHLANDER
+	var/y_level = 0
+	var/x_level = 0
+	var/obj/item/hat = null
+	var/death_throw = FALSE
+	var/death_remove = FALSE
+	var/free_hat = FALSE
+	var/free_hat_lockout = FALSE
+
+/datum/component/hattable/Initialize(death_remove, free_hat)
+	. = ..()
+	RegisterSignal(src.parent, COMSIG_ATTACKBY, PROC_REF(hat_on_obj), override = TRUE)
+	src.death_throw = FALSE
+	src.death_remove = death_remove
+	src.free_hat = free_hat
+	if (free_hat && !free_hat_lockout) // If the thing gets a free hat (currently just AIs with spacebux) and they haven't received it yet, do that
+		var/free_hat_type = pick(childrentypesof(pick(childrentypesof(/obj/item/clothing/head))))
+		hat_on_obj(src.parent, free_hat_type)
+
+
+/datum/component/hattable/proc/hat_on_obj(mob/target as mob, obj/item/item as obj, mob/attacker)
+	var/obj/item/H = null
+	if (src.free_hat && !src.free_hat_lockout)
+		item = new item(get_turf(src)) // Make a new item, lockout future free hats
+		free_hat_lockout = TRUE
+	H = item
+
+	var/atom/movable/hatted = src.parent
+	var/obj/item/hatted_offset = src.parent
+	var/is_item_hat = FALSE
+	if (istype(H, /obj/item/clothing/head/))
+		is_item_hat = TRUE
+		ADD_FLAG(H.appearance_flags, KEEP_TOGETHER) // Flags needed for wigs!
+		ADD_FLAG(H.vis_flags, VIS_INHERIT_DIR)
+
+	if (!src.hat && is_item_hat)
+		if (attacker)
+			attacker.drop_item()
+		else
+			target.drop_item(H) // you're going there whether you like it or not
+
+		src.y_level = 0 // Make ABSOLUTE SURE we're not adding onto a previous offset
+		src.x_level = 0
+		src.y_level = hatted_offset.hat_offset_y + H.hat_offset_y // If a specific hat appears in a strange place, give it its own hat_offset vars to edit the offset by
+		src.x_level = hatted_offset.hat_offset_x + H.hat_offset_x
+		H.pixel_y = src.y_level
+		H.pixel_x = src.x_level
+
+		src.hat = H
+		H.layer = hatted.layer + 7
+		H.set_loc(src.parent)
+		hatted.vis_contents += H
+		if (src.death_remove)
+			RegisterSignal(src.parent, COMSIG_MOB_DEATH, PROC_REF(die), override = TRUE)
+		RegisterSignal(src.hat, COMSIG_ITEM_PICKUP, PROC_REF(take_hat_off), override = TRUE)
+		return
+	else
+		return
+
+/datum/component/hattable/proc/take_hat_off(mob/target, mob/user)
+	var/obj/item/H = src.hat
+	var/atom/movable/hatted = src.parent
+
+	if (H)
+		if (istype(user, /mob/living/silicon/ghostdrone))
+			SPAWN(0) // Magtractors use an action bar until they do stuff, and then they'll clone a ghost image of the item that's on the floor. Drop that!
+				var/mob/living/silicon/ghostdrone/drone = user
+				var/obj/item/magtractor/mag = locate(/obj/item/magtractor) in drone.tools
+				if (mag.holding)
+					mag.dropItem(src)
+		H.set_loc(get_turf(hatted))
+		hatted.vis_contents -= H
+		if(src.death_throw)
+			var/turf/T = get_ranged_target_turf(H, pick(alldirs), 3)
+			H.throw_at(T, 3, 1)
+			SPAWN(10)
+				UnregisterSignal(src.parent, list(COMSIG_MOB_DEATH, COMSIG_ATTACKBY))
+				UnregisterSignal(H, COMSIG_ITEM_PICKUP)
+				src.death_throw = FALSE
+				return
+		else
+			UnregisterSignal(H, COMSIG_ITEM_PICKUP)
+		H.pixel_y = 0
+		H.pixel_x = 0
+		src.y_level = 0
+		src.x_level = 0
+		src.hat = null
+		return
+	else
+		return
+
+/datum/component/hattable/proc/die()
+	if (src.hat)
+		src.death_throw = TRUE
+		take_hat_off(src.parent)
+	else
+		return
+
+
+
+
+
+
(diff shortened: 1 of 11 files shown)


code/modules/mechanics/MechanicMadness.dm
```diff
@@ -48,6 +49,48 @@
...
+		user.drop_item()
+		src.hat = W
+		W.set_loc(src)
+		var/image/hatImage = null
+
+		// Treat wigs differently as their icon_state is always bald
+		if (istype(W, /obj/item/clothing/head/wig))
+			hatImage = W.wear_image
+			hatImage.layer = src.layer+0.1
+			if (src.name == "Component Cabinet")
```
TobleroneSwordfish, line 66: This is level 2 `istype(src...`, please use type vars rather than name checks

TobleroneSwordfish (changes requested):
This would reeeally be a lot better as a component, but if you don't want to tackle that it's okay.

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+	var/free_hat_lockout = FALSE
+
+/datum/component/hattable/Initialize(death_remove, free_hat)
+	. = ..()
+	RegisterSignal(src.parent, COMSIG_ATTACKBY, PROC_REF(hat_on_obj), override = TRUE)
+	src.death_throw = FALSE
+	src.death_remove = death_remove
+	src.free_hat = free_hat
+	if (free_hat && !free_hat_lockout) // If the thing gets a free hat (currently just AIs with spacebux) and they haven't received it yet, do that
+		var/free_hat_type = pick(childrentypesof(pick(childrentypesof(/obj/item/clothing/head))))
```
TobleroneSwordfish, line 24: not sure why you have two `childrentypesof` here but you probably mean `concrete_typesof`

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+		hat_on_obj(src.parent, free_hat_type)
+
+
+/datum/component/hattable/proc/hat_on_obj(mob/target as mob, obj/item/item as obj, mob/attacker)
+	var/obj/item/H = null
+	if (src.free_hat && !src.free_hat_lockout)
```
TobleroneSwordfish, line 30: Better to just make an instance in `Initialize` and pass it in here? You shouldn't need the lockout var then.

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+	src.death_throw = FALSE
+	src.death_remove = death_remove
+	src.free_hat = free_hat
+	if (free_hat && !free_hat_lockout) // If the thing gets a free hat (currently just AIs with spacebux) and they haven't received it yet, do that
+		var/free_hat_type = pick(childrentypesof(pick(childrentypesof(/obj/item/clothing/head))))
+		hat_on_obj(src.parent, free_hat_type)
+
+
+/datum/component/hattable/proc/hat_on_obj(mob/target as mob, obj/item/item as obj, mob/attacker)
+	var/obj/item/H = null
```
TobleroneSwordfish, line 29: single letter variable names are bad and make it very unclear what this var is for

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+	if (src.free_hat && !src.free_hat_lockout)
+		item = new item(get_turf(src)) // Make a new item, lockout future free hats
+		free_hat_lockout = TRUE
+	H = item
+
+	var/atom/movable/hatted = src.parent
+	var/obj/item/hatted_offset = src.parent
+	var/is_item_hat = FALSE
```
TobleroneSwordfish, line 37: Is it ever not an item?

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+	if (istype(H, /obj/item/clothing/head/))
+		is_item_hat = TRUE
+		ADD_FLAG(H.appearance_flags, KEEP_TOGETHER) // Flags needed for wigs!
+		ADD_FLAG(H.vis_flags, VIS_INHERIT_DIR)
+
+	if (!src.hat && is_item_hat)
+		if (attacker)
+			attacker.drop_item()
+		else
+			target.drop_item(H) // you're going there whether you like it or not
```
TobleroneSwordfish, line 47: When is this needed?

code/mob/living/silicon.dm
```diff
@@ -28,6 +28,9 @@ ADMIN_INTERACT_PROCS(/mob/living/silicon, proc/pick_law_rack)
 
 	var/static/regex/monospace_say_regex = new(@"`([^`]+)`", "g")
 
+	var/hat_offset_y = 0
+	var/hat_offset_x = 0
```
TobleroneSwordfish, line 32: Not sure what these are doing here, if you want to do specific offsets per type then pass the offsets in the component init.

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+		src.hat = H
+		H.layer = hatted.layer + 7
+		H.set_loc(src.parent)
+		hatted.vis_contents += H
+		if (src.death_remove)
+			RegisterSignal(src.parent, COMSIG_MOB_DEATH, PROC_REF(die), override = TRUE)
+		RegisterSignal(src.hat, COMSIG_ITEM_PICKUP, PROC_REF(take_hat_off), override = TRUE)
+		return
+	else
+		return
```
TobleroneSwordfish, line 65:
```suggestion
```

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+
+	if (!src.hat && is_item_hat)
+		if (attacker)
+			attacker.drop_item()
+		else
+			target.drop_item(H) // you're going there whether you like it or not
+
+		src.y_level = 0 // Make ABSOLUTE SURE we're not adding onto a previous offset
+		src.x_level = 0
+		src.y_level = hatted_offset.hat_offset_y + H.hat_offset_y // If a specific hat appears in a strange place, give it its own hat_offset vars to edit the offset by
```
TobleroneSwordfish, line 51: see the comment on the AI definition, these should be passed on init, not as vars on specific mob types

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+		hatted.vis_contents += H
+		if (src.death_remove)
+			RegisterSignal(src.parent, COMSIG_MOB_DEATH, PROC_REF(die), override = TRUE)
+		RegisterSignal(src.hat, COMSIG_ITEM_PICKUP, PROC_REF(take_hat_off), override = TRUE)
+		return
+	else
+		return
+
+/datum/component/hattable/proc/take_hat_off(mob/target, mob/user)
+	var/obj/item/H = src.hat
```
TobleroneSwordfish, line 68: Unnecessary cast, `hat` is already an `obj/item`

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+	var/atom/movable/hatted = src.parent
+
+	if (H)
```
TobleroneSwordfish, line 71: Consider using guard pattern here, ie `if (!src.hat) return` instead of indenting the entire proc

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+		hatted.vis_contents -= H
+		if(src.death_throw)
+			var/turf/T = get_ranged_target_turf(H, pick(alldirs), 3)
+			H.throw_at(T, 3, 1)
+			SPAWN(10)
+				UnregisterSignal(src.parent, list(COMSIG_MOB_DEATH, COMSIG_ATTACKBY))
+				UnregisterSignal(H, COMSIG_ITEM_PICKUP)
+				src.death_throw = FALSE
+				return
+		else
```
TobleroneSwordfish, line 88: Won't this try to throw when picked up normally too?

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+			UnregisterSignal(H, COMSIG_ITEM_PICKUP)
+		H.pixel_y = 0
+		H.pixel_x = 0
+		src.y_level = 0
+		src.x_level = 0
+		src.hat = null
+		return
+	else
+		return
```
TobleroneSwordfish, line 97:
```suggestion
```

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,110 @@
...
+
+/datum/component/hattable/proc/die()
+	if (src.hat)
+		src.death_throw = TRUE
+		take_hat_off(src.parent)
+	else
+		return
```
TobleroneSwordfish, line 104:
```suggestion
```

code/obj/item.dm
```diff
@@ -1154,14 +1158,13 @@ ADMIN_INTERACT_PROCS(/obj/item, proc/admin_set_stack_amount)
...
-			if(src in bible_contents)
-				break
-			else
-				return 0
-		checkloc = checkloc:loc
+		if(isliving(checkloc) && checkloc != user)
+			break
+		else
+			return 0
+		checkloc = checkloc.loc
```
TobleroneSwordfish, line 1167: This isn't doing what you think it's doing. A better solution might be just skipping this check if the item is contained *directly* in a mob, rather than being in a container in a mob, since the intent is to prevent people stealing stuff out of other people's backpacks (I think??)

TobleroneSwordfish (changes requested):
This is much better, thanks for taking on the project of making `component/hattable` ❤️

code/datums/components/hattable.dm
```diff
@@ -0,0 +1,106 @@
...
+	src.hat.pixel_x = offsetBy_x
+
+	src.hat.transform *= src.scale_amount
+	src.hat.layer = hatted.layer + 1
+	src.hat.set_loc(src.parent)
+	hatted.vis_contents += src.hat
+
+	if (src.death_remove)
+		RegisterSignal(src.parent, COMSIG_MOB_DEATH, PROC_REF(die), override = TRUE)
+	RegisterSignal(src.hat, COMSIG_ITEM_PICKUP, PROC_REF(take_hat_off), override = TRUE)
```
TobleroneSwordfish, line 64:
```suggestion
	RegisterSignal(src.hat, COMSIG_ITEM_PICKUP, PROC_REF(take_hat_off), override = TRUE)
	return TRUE
```
returning true from this proc cancels the attack stack and prevents "User hits AI with funny hat!" when you're trying to put a hat on something

TobleroneSwordfish (changes requested):
Looks good, a few final nitpicks (sorry for taking a while to get back to this)

CalliopeSoups (commented):
> Looks good, a few final nitpicks (sorry for taking a while to get back to this) 

No problem, thanks for teaching me so much!

PR #18145, opened 2024-03-08
Makes cryoing free up slots for non-unique jobs

TobleroneSwordfish:
[BALANCE] [RESPAWNING]
## About the PR 
Marks some jobs (heads of staff) as unique, cryoing as a non-unique job will now immediately free up the slot instead of requiring ghosting. Cryo-ghosting will still free up a unique job slot.

## Why's this needed? 
Feedback from the #18096 testmerge, people were having issues with secoffs cryoing and still taking up slots. Yes this allows for circumventing slot restrictions by cryoing, but this was already possible under the old system by dying and being cloned so 🤷 

## Changelog 

```changelog
(u)LeahTheTech
(+)Cryoing as a non-unique job will now immediately free up the slot.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -68,6 +68,9 @@
 	var/wiki_link = null //! Link to the wiki page for this job
 
 	var/counts_as = null //! Name of a job that we count towards the cap of
+	///if true, cryoing won't free up slots, only ghosting will
+	///basically there should never be two of these
+	var/unique = FALSE
 
 	New()
 		..()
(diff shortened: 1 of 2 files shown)


Chatauscours:
After playing and observing a few rounds with this I feel that most jobs should be counted as unique since ending up with two chaplains and two detectives was feeling kinda meh for them from what I gathered.

Imo all command jobs, detective, chaplain, and probably most 1 slot gimmick roles like radio host and inspector should be counted as unique.

PR #18153, opened 2024-03-09
Adds blinds to cog1 bar office

cheekybrdy:
## About the PR 
Adds blinds to the cog1 bar office and moves the north blast door to its own independent controls

## Why's this needed? 
Bartender has no control over their office's own privacy with emergency A being able to look in from the south and the shutter on the north being chef controlled means they have no agency on northern privacy either. Fixes both problems


TobleroneSwordfish (changes requested):
Probably swap the locations so they make a bit more sense?

cheekybrdy:
lgtm

PR #18161, opened 2024-03-10
Recoil mechanics

TDHooligan:
[FEATURE] [OBJECTS]

## About the PR 
Adds the following recoil mechanics, all of which are fully configurable per-weapon, and toggleable under 'Effects':

**Camera Recoil:** 
Each shot punts your camera backwards, with a small offset for kinetics to make them feel more rattle-y. The severity of which can be configured per-weapon.

In a very-high-recoil weapon, it looks like this:

https://github.com/goonstation/goonstation/assets/6061314/b1d50637-33e5-4c43-b511-e9967bbe703f

**Cumulative Recoil:**
Kinetic only by default.
Shooting adds recoil which stacks up, then resets as you stop shooting. This penalises accuracy, 
Automatic weapons become less controllable as you fire them harder.

To convey recoil inaccuracy, there is:

**Icon Recoil**
Kinetic only by default.
As you accumulate recoil, the icon in your inventory tilts to demonstrate this.

https://github.com/goonstation/goonstation/assets/6061314/f73b843c-a76f-4b49-b6db-1c321d4e9a64

## Why's this needed? 

ZeWaka thought the Draco was cool, and thought recoil was cool. 
I did too, so I made it. It's also toggle-able so *shrug

## Changelog 

```changelog
(u)TDHooligan
(*)Firearms now have recoil; As a result, big kinetic guns could get hard to control if you apply too much dakka!
```


code/obj/item/gun/gun_parent.dm
```diff
@@ -457,3 +509,69 @@ var/list/forensic_IDs = new/list() //Global list of all guns, based on bioholder
...
+	recoil += (recoil_strength + stacked_recoil)
+	recoil = clamp(recoil, 0, recoil_max)
+	recoil_last_shot = TIME
+	if (camera_recoil_enabled)
+		do_camera_recoil(user, start,target,POX,POY)
+
+	if (first_shot && src.current_projectile.shot_number > 1 && src.current_projectile.shot_delay > 0)
+		for (var/i=1 to src.current_projectile.shot_number-1)
+			SPAWN(i*src.current_projectile.shot_delay)
+				handle_recoil(user,start,target,POX,POY, FALSE)
```
Tarmunora, line 574: this feels very cursed

code/obj/item/gun/kinetic.dm
```diff
@@ -2716,6 +2767,12 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 	default_magazine = /obj/item/ammo/bullets/lmg
 	ammobag_magazines = list(/obj/item/ammo/bullets/lmg)
 	ammobag_restock_cost = 3
+	recoil_stacking_enabled = TRUE
+	recoil_stacking_safe_stacks = 8
+	recoil_stacking_max_stacks = 16
+	recoil_stacking_amount = 1
+	recoil_max = 100
+
```
Tarmunora, line 2775: feels like waaay too much recoil on this, considering it already has base spread.

code/obj/item/gun/kinetic.dm
```diff
@@ -2457,6 +2501,7 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 	ammo_cats = list(AMMO_PISTOL_22)
 	default_magazine = /obj/item/ammo/bullets/bullet_22/smartgun
 	ammobag_magazines = list(/obj/item/ammo/bullets/bullet_22/smartgun)
+	recoil_strength = 6
```
Tarmunora, line 2504: smartgun should be recoil-free

code/obj/item/gun/kinetic.dm
```diff
@@ -2485,6 +2530,7 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 	default_magazine = /obj/item/ammo/bullets/bullet_9mm/smg
 	ammobag_magazines = list(/obj/item/ammo/bullets/bullet_9mm/smg)
 	ammobag_restock_cost = 2
+	recoil_strength = 8
```
Tarmunora, line 2533: SMG recoil should probably account for whether it's being held in 1 hand or 2

Tarmunora (commented):
Overall thoughts:
camera-shake recoil could stand to be tuned down quite a bit, esp considering there's a toggle. It's pretty obnoxious on some weapons, to the point where it feels like flipping the toggle is more-or-less required, and that's not great.

Additionally, with adding recoil like this to more-or-less every kinetic weapon, base spread angles could stand to be adjusted/removed for some weapons, esp automatics.

TDHooligan:
> Overall thoughts: camera-shake recoil could stand to be tuned down quite a bit, esp considering there's a toggle. It's pretty obnoxious on some weapons, to the point where it feels like flipping the toggle is more-or-less required, and that's not great.

the numbers tweaking is awkward as i've spent so long wiring this up i'm definitely desensitised to it. 

i can take another look at this sometime soon. though if you have ideas for how the recoil values should look, i'm happy for you to tweak the values as you see fit.

TDHooligan:
i have shot the numbers down to half or less for the bigger guns 🔫

code/obj/item/gun/gun_parent.dm
TDHooligan: it is. however projectile code (which handles burst) doesn't have a reference back to owner

TDHooligan: i could probably rework that (but i do have the big firemode seperation at about 75~80% completion, and this would definitely get in the way)

Tarmunora (approved):

PR #18163, opened 2024-03-10
Implement Job Manager TGUI

Sovexe:
## About the PR 
* Recreation of the job manager window in TGUI
* Fixes bug preventing the creation of custom jobs and causing unintended edits to existing jobs
* Job creator will now require a name
* declare the base daily job type abstract

image

## Why's this needed? 
* Improve admin UX
* fix bugs


code/modules/admin/job_manager.dm
```diff
@@ -0,0 +1,78 @@
...
+			"hiddenJobs" = hidden_job_data,
+			"allowSpecialJobs" = job_controls.allow_special_jobs
+		)
+
+/datum/job_manager/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
+	. = ..()
+	if(.)
+		return
+	if (!isadmin(ui.user))
+		return
```
mordent-goonstation, line 38: Discussing in internal developer channel about what safety check we should be doing for admin controls.

code/modules/admin/job_manager.dm
```diff
@@ -0,0 +1,78 @@
...
+	switch(action)
+		if ("alter_cap")
+			var/datum/job/JOB = find_job_in_controller_by_string(params["job"])
```
mordent-goonstation, line 41: Prefer non-all-caps names for local variables, all caps implies a global constant.

tgui/packages/tgui/interfaces/JobManager.js
```diff
@@ -0,0 +1,144 @@
...
+              {specialJobs.map(job => (
+                <LabeledList.Item 
+                  key={job.name} 
+                  label={job.name}    
+                  buttons={
+                    <>
+                      <Button
+                        content={`${job.count}/${job.limit}`}
+                        onClick={() => act('alter_cap', { job: job.name })}
+                      />
```
mordent-goonstation, line 87: Add a tooltip to this to explain what clicking it does (`tooltip` prop).

tgui/packages/tgui/interfaces/JobManager.js
```diff
@@ -0,0 +1,144 @@
...
+                    </>
+                  }
+                />
+              ))}
+            </LabeledList>
+          </Collapsible>
+          <Collapsible title="Hidden Jobs">
+            <LabeledList>
+              {hiddenJobs.map(job => (
+                <LabeledList.Item
```
mordent-goonstation, line 108: Given that you're repeating the above things here, extract this `LabeledList.Item` out into a component with relevant props (name, count, limit, type) and render one of those for both this one and the ones above. One allows a delete button, which you can have via a simple `allowDelete` prop to conditionally render.

mordent-goonstation (changes requested):

mordent-goonstation (approved):
Looks good to me. Might want an entry in the admin changelog?

PR #18164, opened 2024-03-10
Poll Editor Panel interface

mordent-goonstation:
## About the PR 

Adds a form UI to the Add Poll action on the Player Polls panel, rather than a series of inputs.

image

The UI does some basic validation.

image

Further errors are displayed based on BYOND-side checking/validation/processing. Exactly what is shown here could be decently technical, but this provides some feedback on whether there was an error.

image

## Why's this needed? 

Allows for editing the entity as a whole with a better UX rather than the current clunky implementation.


code/modules/polling/poll_editor_panel.dm
```diff
@@ -0,0 +1,96 @@
...
+		return
+
+	switch (action)
+		if ("save")
+			USR_ADMIN_ONLY
+
+			var/list/poll
+
+			try
+				// TODO: html_encode?
```
mordent-goonstation, line 42: Mostly in a draft state because I want to investigate this. It was in the earlier implementation.

tgui/packages/tgui/interfaces/PollEditorPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+  if (options.length < 2) {
+    return 'At least 2 options required.';
+  }
+  return undefined;
+};
+
+/*
+ * N.B. On creation, this form was built to handle creating polls, with intent to expand
+ * to editing them at a later date. In its initial state, it only works for creating polls.
+ */
```
mordent-goonstation, line 92: Consider whether this should be done as part of it. Is this even needed (i.e. edit already existing poll via a form)?

mordent-goonstation:
Needs an admin changelog update.

mordent-goonstation:
Probably fine, waiting on confirmation that html_encode isn't needed here. I suspect it isn't (else it would be needed in all sorts of other places as well).

mordent-goonstation:
I need to bump the admin changelog date, but will try to merge this in later.

mordent-goonstation:
image
Light mode screenshot. Placeholders in text inputs look bad, but that's not new. Colors used throughout UI are default or "good"/"bad" so will align with general light mode UX.
image

PR #18169, opened 2024-03-10
Clarion Catering Upgrades

MintyPhresh:
## About the PR
- Layout of bar, cafeteria, and kitchen rearranged: Extra table space in most places, and much more space for the common folk in the cafeteria
- Beer keg moved to the small shared room.
- Ice dispenser added to the bar 
- Swing sign secured to prevent George grife
- Southern Ice Cream Counter widened and decluttered
- Second KitchenHelper Food Mixer removed
- Boxes of cocktail decorations added to a wall closet in the bar, alongside equipment previously found in crates

image
image

## Why's this needed? 
Clarion's bar is unique in that the bartender doesn't have a backroom, and it has the smallest kitchen-bar shared room in all maps in rotation, which poses some layout problems. This is compounded with the dimensions of the bar and kitchen's rooms; Both are 5 tiles wide, which makes it very easy to put an island of tables or other equipment in the center of the room, making two one-tile pathways that are very claustrophobic. Eliminating the islands, and using different storage options like wall closets, should make the departments feel much roomier.


MintyPhresh:
One particular design coince I'd like to highlight for review is the window between the bar and kitchen-- this wasn't in the original Clarion, and it provides a pretty good view about halfway into the kitchen. I figured that with the patented gibber viewing window in tool storage, it wouldn't be too big of a deal, but it does still reduce the crimeability of the kitchen

DecidedlyUnethical:
the only thing I'm not all there on is the way the carpet looks -- both the stacked carpet on the left and the weird trimming on the protruding section of bar table. All the rest, fantastic.

MintyPhresh:
Edited the map to address git + discord feedback

WALPVRGIS (changes requested):
I'll do a more thorough review when I can, but first thing I immediately noticed is this extension of the bar table;

image

It looks quite awkward to me, and also would not be conducive to foot traffic flow; I would much rather see the additional surface/table space as a larger table separate to the bar.

Would also like to see floor decor like this returned to the bar, removing decor and details that add character is not ideal imo;
image

MintyPhresh:
The intent for that protrusion is that bartenders will drinkslide the drinks to customers-- it's a cheap hack to get much more seats since that area doesn't often get a ton of plates there taking up space. Definitely an experimental feature, but in my solo playtesting it was kind of fun.

As for the decorations, I'll get those back in.

WALPVRGIS:
> The intent for that protrusion is that bartenders will drinkslide the drinks to customers-- it's a cheap hack to get much more seats since that area doesn't often get a ton of plates there taking up space. Definitely an experimental feature, but in my solo playtesting it was kind of fun.

Ya I get that, and the concept isn't what I take issue with, but rather the execution; why not do something like this instead? It would look a lot less awkward, mapping-wise.
image
Bartenders would have an option to slide drinks across the bar if they'd like, and can easily access said bartop without leaving the bar area itself. It also doesn't block people from running in from the door to the south, up towards the kitchen entrance (well, nearly as much, as we can only make so many clear pathways in rooms). 

As it is, I would rather not merge this with the bar in its current state.

MintyPhresh:
> why not do something like this instead? It would look a lot less awkward, mapping-wise. 

That looks great, yeah. The chess table and north locker will need to move, but with us gaining a table on the bar we can probably make some sacrifices to the dining area for those. Will update sometime this weekend.

MintyPhresh:
Should I try to implement #18232 into this PR? I think I'd have some issues implementing it, and my personal opinion is that it isn't that important, but if it IS important, then this PR and that one won't mesh very well.

cheekybrdy:
From what I can see the best way to merge the two PRs would likely be to move the hotdog cart to where the kitchen pickle jar is (weird that there are 2 there anyway), replace one of the remaining two crates to where the hotdog crate is now to replace with the chute. I'll close my PR on this area to reduce conflict possibilties

TobleroneSwordfish (changes requested):
image
This looks kind of odd with the carpet under the inner bar area, makes it look more like a temporary table than a bar that's fixed to the floor.

MintyPhresh:
I would prefer to keep that in. I imagine the rug was purchased with no thought put into whether it would fit, so they just unscrewed the front island part and put it in.

If the bar table is supposed to be fixed to the floor, the sprite doesn't really reflect that so much. Just looks like a table whereas I'd expect a fixed bar to look closer to a countertop

PR #18170, opened 2024-03-10
Adds a cooling rack to the kitchen

DecidedlyUnethical:
[CATERING][OBJECTS][FEATURE][WIKI]
## About the PR 
This PR adds a cooling rack, which is a subtype of a plate which exposes the food items placed on it to the air temperature of whatever turf it's on, allowing hot foods to cool off, and cold foods to warm up. While ostensibly this change was made in order to stop people from immediately scalding themselves on deep fried food, this also allows for interesting methods of alternative heating and cooling.

image

The FoodTech starts with three available at roundstart, with more being obtainable through restock cartridges. Cooling racks cannot be stacked, and you can't smash people over the head with them either.

## Why's this needed? 
Fried, grilled, or fireblasted food was permanently scalding with no real way to cool it down other than reagent extracting, and at that point it's not really food, is it?

## Changelog 

```changelog
(u)DecidedlyUnethical, MylieDaniels
(+)Added some cooling racks to the kitchen. Find them waiting in the FoodTech!
```


TobleroneSwordfish:
This is neat, but how are you ending up with scalding food in the first place? Grilling and frying don't seem to heat up the reagents.

DecidedlyUnethical:
> This is neat, but how are you ending up with scalding food in the first place? Grilling and frying don't seem to heat up the reagents.
You're right in that the deep fryer doesn't directly heat the reagents, no. What it does do is add space-soybean oil at 458.15K to the reagent container, which _functionally_ heats the reagents inside. The shitty grill does the same with beff and charcoal at 472.5K. When you eat these items, the reagents _will_ scald, set your temperature to maximum, and do a small amount of burn damage to you. See attached images.
image
image
image

PR #18189, opened 2024-03-12
Minor Boxing Ring fixes

TheTyrant25:
[MAPPING]
## About the PR 
Added/fixed stage edges to all boxing rings you enter on up a staircase. Also removed some boxing mat tiles which don't match the location of the boxing ropes, for perspective reasons.

## Why's this needed?
Just cleans up the appearance a little bit.


Sord213 (changes requested):
The stage edges on the top of the boxing rings block you from jumping off the ropes back into the ring. Please fix.

PR #18190, opened 2024-03-12
Clothing Booth Rework

DisturbHerb:
[OBJ][UI][FEAT][QOL]

## KNOWN ISSUES (28 SEP 2024)

- ~~SVGs used in item swatches (where applicable) don't scale right. Please fix this before fully merging!~~
- There's an unusually large number of items that just haven't been transferred over to the new inventory system due to data entry errors. Keep an eye on these!
- ~~The tag filter selection modal can't be toggled on or off and filtering the stock list by tags is presently broken.~~
- ~~Filtering the stock list by slot is presently broken.~~

## To Do

- [ ] ~~Display seasonal items at the top of the list~~
- [x] Fix tag filters not disabling properly
- [ ] ~~Get those vector images in the item swatches working properly~~
- [x] Rainbow swatches don't exist yet
- [x] Finalise React conversion
  - [x] Tag filter modal and selection
  - [x] Slot filter selection

## About the PR 

Large scale rework of the internals of the clothing booth, including how stock is organised hierarchically and displayed to the end user. Additional filtering options have been added to (hopefully) allow for ease of navigating the large list of available items, and the number of items displayed to the end user have been truncated by grouping re-colours/similar items within larger groupings. The character preview is also significantly more robust and performance intensive.

image

The booth now accepts cashless payments via ID card. No, cyborgs still can't use it.

Contributors should note that there will be significant changes to how stock items are added to the clothing booth with this PR and should refer to the requisite documentation [available here](https://hackmd.io/62Y7PflOT3aLsBgCaVlFmA) or on the Goonstation organisation when it eventually gets added.

All files relating to the clothing booth (save for the defines which are included with `_std.dme`) are in their own separate directory at `modules\vending\clothingbooth`.

Testmerge recommended to ensure that any errors in data entry (which was performed by hand) are resolved and that the interface can survive a larger scale test. There's one specific case where one of the variants of dresses in the suit slot was missing from the older version's stock list and was added by hand to this version.

~~UPDATE (June 1 2024): Turns out, several items are indeed missing from this stocklist and I haven't been able to keep track of them. Wer'e definitely going to need to testmerge this first.~~

## Why's this needed? 

The previous TGUI interface that I made for this was haunting me at night. It's finally time to put this thing to rest.

## Changelog 

```changelog
(u)DisturbHerb, Glamurio, LeahTheTech, Mordent, Trazodont, Tyrant
(*)Revamped the clothing booth interface and reworked how it functions.
(*)Please bug report any missing items or different prices from the old interface!
(+)Credit to DisturbHerb, Glamurio, LeahTheTech, and Mordent for the clothing booth UI.
(+)Credit to Trazodont and Tyrant for volunteering with the clothing booth item data entry process.
```


TobleroneSwordfish:
H-herb how the hell is this 4000 lines, what did you dooo?

DisturbHerb:
> H-herb how the hell is this 4000 lines, what did you dooo?

a lot....

DisturbHerb:
!merge_upstream

DisturbHerb:
!merge_upstream

tgui/packages/tgui/assets/bisect_left.svg
mordent-goonstation, line 1: Are these used? I see the React component equivalents of these further down.

tgui/packages/tgui/interfaces/ClothingBooth/type.ts
```diff
@@ -1,21 +1,72 @@
...
+  name: string;
   cost: number;
-  img: string;
+  swatch_background_colour?: string;
+	swatch_foreground_shape?: string;
+	swatch_foreground_colour?: string;
+}
+
+export interface SwatchForegroundProps {
+  colour: string;
```
mordent-goonstation, line 37: Not sure why you'd want this to be a single color, in that a single color could just be programmatically applied as a fill style to an SVG instead of having to do this way.

My understanding of intent for these icons was that they could support multiple shapes/colors at once? Which would necessitate a `string[]`.

If you only want these to have a single color, then these can just be `.svg` files as before. Set the `fill` to be the color you want.

Also, `colour` -> `color`. Mixing British in to web stuff is dangerous.

tgui/packages/tgui/interfaces/ClothingBooth/BoothGrouping.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <Stack fill vertical>
+          <Stack.Item bold>
+            <Stack fluid>
+              <Stack.Item grow>{name}</Stack.Item>
+              <Stack.Item>{cost_min === cost_max ? `${cost_min}⪽` : `${cost_min}⪽ - ${cost_max}⪽`}</Stack.Item>
+            </Stack>
+          </Stack.Item>
+          <Stack.Item>
+            <Stack fluid style={{ opacity: '0.5' }}>
+              {Object.values(grouping_tags).length && (
```
mordent-goonstation, line 44: This would short-circuit to `0` and render that if the length is `0` as-is. Boolean `false` doesn't have that problem, so throw in a comparator.
```suggestion
              {Object.values(grouping_tags).length > 0 && (
```

tgui/packages/tgui/interfaces/ClothingBooth/BoothGrouping.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <Stack.Item grow={1}>
+        <Stack fill vertical>
+          <Stack.Item bold>
+            <Stack fluid>
+              <Stack.Item grow>{name}</Stack.Item>
+              <Stack.Item>{cost_min === cost_max ? `${cost_min}⪽` : `${cost_min}⪽ - ${cost_max}⪽`}</Stack.Item>
+            </Stack>
+          </Stack.Item>
+          <Stack.Item>
+            <Stack fluid style={{ opacity: '0.5' }}>
```
mordent-goonstation, line 43: How does this render if neither of its children should render? Do we need this item at all? Do we need its parent?

i.e. empty state?

tgui/packages/tgui/interfaces/ClothingBooth/CharacterPreview.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Button, Image, Stack } from '../../components';
+import type { ClothingBoothData } from './type';
+
+interface PreviewImageProps {
+  height: number;
+  icon: string;
+}
+
+const PreviewImage = (props: PreviewImageProps) => {
+  const { height, icon } = props;
```
mordent-goonstation, line 11: Probably provide a default height here. `height = DEFAULT_PREVIEW_HEIGHT` or whatever.

tgui/packages/tgui/interfaces/ClothingBooth/GroupingTag.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const { tags } = data;
+  const { slot, grouping_tags } = props;
+  const groupingTagsObject = Object.values(grouping_tags);
+  const sortedGroupingTags = groupingTagsObject.sort(
+    buildFieldComparator((groupingTag) => tags[groupingTag].display_order, numberComparator)
+  );
+  const clothingBoothSlotLookup = Object.entries(ClothingBoothSlotKey).reduce((acc, [name, id]) => {
+    acc[id] = { id, name };
+    return acc;
+  }, {} as Record<ClothingBoothSlotKey, ClothingBoothSlotDetail>);
```
mordent-goonstation, line 27: This can be moved outside of the component, as it's solely based on `ClothingBoothSlotKey` which doesn't change, so it's only generated once.

tgui/packages/tgui/interfaces/ClothingBooth/GroupingTag.tsx
```diff
@@ -0,0 +1,55 @@
...
+  );
+};
+
+export const GroupingTag = (props: ClothingBoothGroupingTagsData) => {
+  const { name, colour } = props;
+
+  return (
+    <Box
+      className={'clothingbooth__groupingtag'}
+      color={!!colour && colour}
```
mordent-goonstation, line 49: `color` is fine to be `undefined`, so no need to do a `!!foo &&` check.

tgui/packages/tgui/interfaces/ClothingBooth/GroupingTag.tsx
```diff
@@ -0,0 +1,55 @@
...
+      style={{ border: `0.0835rem solid ${colour ? colour : 'currentColor'}` }}
```
mordent-goonstation, line 50: Hate this dimension. Please specify as a constant if you want something esoteric, else just use `1px` etc.? I have no clue what `0.0835rem` resolves to, but I assume it's `1px`.

tgui/packages/tgui/interfaces/ClothingBooth/GroupingTag.tsx
```diff
@@ -0,0 +1,55 @@
...
+  slot: ClothingBoothSlotKey;
+  grouping_tags: string[];
+}
+
+interface ClothingBoothSlotDetail {
+  id: ClothingBoothSlotKey;
+  name: string;
+}
+
+export const GroupingTagContainer = (props: GroupingTagContainerProps, context) => {
```
mordent-goonstation, line 16: Prefer `GroupingTags` as a name here. This would make the file as a whole `GroupingTags.tsx`.

tgui/packages/tgui/interfaces/ClothingBooth/GroupingTag.tsx
```diff
@@ -0,0 +1,55 @@
...
+        </Stack.Item>
+      ))}
+      <Stack.Item>
+        <GroupingTag name={clothingBoothSlotLookup[slot].name} />
+      </Stack.Item>
+    </Stack>
+  );
+};
+
+export const GroupingTag = (props: ClothingBoothGroupingTagsData) => {
```
mordent-goonstation, line 43: No need to export, as not used outside of here.

mordent-goonstation, line 43: Prefer making the props as a:
```
interface GroupingTagProps extends ClothingBoothGroupingTagsData {}
```
Reasoning is that this discourages people from adding things to the shared data type when they're only used by component props (assuming the data type is used elsewhere).

tgui/packages/tgui/interfaces/ClothingBooth/GroupingTag.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const { name, colour } = props;
+
+  return (
+    <Box
+      className={'clothingbooth__groupingtag'}
```
mordent-goonstation, line 48:
```suggestion
      className="clothingbooth__groupingtag"
```

tgui/packages/tgui/interfaces/ClothingBooth/ItemSwatch.tsx
```diff
@@ -0,0 +1,62 @@
+import { classes } from 'common/react';
+import { Box, Tooltip } from '../../components';
+import { SwatchBisectLeft } from './swatchForegroundShapes/bisect-left';
+import { SwatchBisectRight } from './swatchForegroundShapes/bisect-right';
+import { SwatchPolkadots } from './swatchForegroundShapes/polkadots';
+import { SwatchClub } from './swatchForegroundShapes/club';
+import { SwatchDiamond } from './swatchForegroundShapes/diamond';
+import { SwatchHeart } from './swatchForegroundShapes/heart';
+import { SwatchSpade } from './swatchForegroundShapes/spade';
```
mordent-goonstation, line 9:
```suggestion
import { SwatchBisectLeft, SwatchBisectRight, SwatchPolkdadots, /* etc */ } from './swatchForegroundShapes';
```
Add an `index.ts` file to the folder with the icons in it, which just re-exports the contents of that folder.
```
export { SwatchBisectLeft } from './bisect-left';`
// etc.
```
Putting here as related: the icon component files are components, so should follow PascalCase naming convention still.

tgui/packages/tgui/styles/interfaces/ClothingBooth.scss
```diff
@@ -1,31 +1,52 @@
...
 
-  &preview {
+  &__preview {
     width: 64px;
     height: 128px;
   }
+
+  &__swatch {
+    &_box {
+      margin: 0.2em;
```
mordent-goonstation, line 35: Just throw this on the component as `m={whatever}`, imo. You already use `px` IIRC in there.

tgui/packages/tgui/styles/interfaces/ClothingBooth.scss
```diff
@@ -1,31 +1,52 @@
...
   }
 
-  &preview {
+  &__preview {
     width: 64px;
     height: 128px;
   }
+
+  &__swatch {
+    &_box {
```
mordent-goonstation, line 34:
```
  &__swatch-box {
```

tgui/packages/tgui/styles/interfaces/ClothingBooth.scss
```diff
@@ -1,31 +1,52 @@
...
+      border-radius: base.$border-radius;
+  }
+}
+
+// If any swatch foreground shapes need custom styling, this is where it lives.
+.clothingbooth__swatch_icon {
+  &_rainbow {
+    background-image: linear-gradient(45deg, red,orange,yellow,green,blue,indigo,violet);
+  }
 }
```
mordent-goonstation, line 52: No real issue with this being its own little thing, but should still live within the top-level `.clothingbooth` "block". i.e. `&__swatch_icon` (and move into the block above).

Also, naming convention would be `swatch-icon`, not `swatch_icon`.

tgui/packages/tgui/styles/interfaces/ClothingBooth.scss
```diff
@@ -1,31 +1,52 @@
...
+  }
+
+  &__groupingtag {
+      border-radius: base.$border-radius;
+  }
+}
+
+// If any swatch foreground shapes need custom styling, this is where it lives.
+.clothingbooth__swatch_icon {
+  &_rainbow {
```
mordent-goonstation, line 49:
```suggestion
  &--rainbow {
```
As we're using BEM here, let's use the "modifier" syntax (as this is a modifier). Requires a change wherever else this is set.

tgui/packages/tgui/interfaces/ClothingBooth/swatchForegroundShapes/spade.tsx
```diff
@@ -0,0 +1,23 @@
+import { SwatchForegroundProps } from "../type";
+
+export const SwatchSpade = (props: SwatchForegroundProps) => {
+  return (
+    <svg width="auto" height="100%" viewBox="0 0 6 6" xmlns="http://www.w3.org/2000/svg">
```
mordent-goonstation, line 5: Rather than put the `props.color` on every child here, I *think* you can get away with putting `fill={props.color}` on the `svg` (and removing from each of the `path` children). If that doesn't work then fine as you have it, but would prefer that if we can.

tgui/packages/tgui/interfaces/ClothingBooth/swatchForegroundShapes/polkadots.tsx
```diff
@@ -0,0 +1,31 @@
+import { SwatchForegroundProps } from '../type';
+
+export const SwatchPolkadots = (props: SwatchForegroundProps) => {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+      width="21.333332"
+      height="16pt"
```
mordent-goonstation, line 9: This size is nuts. Shouldn't all of these have the same width/height?

tgui/packages/tgui/interfaces/ClothingBooth/swatchForegroundShapes/diamond.tsx
```diff
@@ -0,0 +1,18 @@
...
+        transform="matrix(.64018 .76822 -.64018 .76822 -22.577782 0)"
+        d="M20.297247-19.378817h2.7552822v2.7552826H20.297247z"
+        fill={props.colour}
+      />
+      <path
+        d="M25.016985 2.4392049h.76603258v.76603401H25.016985z"
+        transform="translate(-22.577782)"
+        fill={props.colour}
+      />
+    </svg>
```
mordent-goonstation, line 16: No idea why this has two paths. It's a diamond, one path surely?

mordent-goonstation (changes requested):

tgui/packages/tgui/interfaces/ClothingBooth/swatchForegroundShapes/diamond.tsx
DisturbHerb: The diamond has curves, take it up with the .svg optimiser not me 😰

tgui/packages/tgui/interfaces/ClothingBooth/type.ts
DisturbHerb: You're right, they should only be a single colour, I was just sent down this wormhole trying to figure out `.svg` styling this entire time because of the restrictions of IE11. Importing vector images from `.svg` has caused an unneccessary amount of pain on my end.

ZeWaka:
Closed due to the TGUI feature freeze, https://github.com/goonstation/goonstation/discussions/19808

DisturbHerb:
release me

DisturbHerb:
ball

code/modules/clothingbooth/clothingbooth.dm
```diff
@@ -0,0 +1,446 @@
...
+// Blatantly stolen from `/datum/component/barber`.
+/obj/machinery/clothingbooth/proc/reference_clothes(mob/living/carbon/human/to_copy, mob/living/carbon/human/to_paste)
+	src.nullify_clothes(to_paste)
+	to_paste.wear_suit = SEMI_DEEP_COPY(to_copy.wear_suit)
+	to_paste.w_uniform = SEMI_DEEP_COPY(to_copy.w_uniform)
+	to_paste.shoes = SEMI_DEEP_COPY(to_copy.shoes)
+	to_paste.gloves = SEMI_DEEP_COPY(to_copy.gloves)
+	to_paste.glasses = SEMI_DEEP_COPY(to_copy.glasses)
+	to_paste.head = SEMI_DEEP_COPY(to_copy.head)
+	to_paste.wear_id = SEMI_DEEP_COPY(to_copy.wear_id)
```
Tarmunora, line 373: `SEMI_DEEP_COPY()` is an *extremely* unsafe proc, and this has been the root cause of thousands of runtimes in each round since this got testmerged, primarily caused by issues with PDAs in the wear_id slot getting `SEMI_DEEP_COPY()`ed

Tarmunora (changes requested):

Issue #16758, opened 2023-11-03
salvager glasses shows up in arrival seat


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

salvager glasses shows up in arrival seat

### Steps to reproduce

1. go to arrival on salv round
2. found salvie glasses right on the arrival seat
3. scan it with forensic scanner
4, no fingerprints

### Expected Behavior

for salv glasses to not spawn on arrival floor

### Additional Information & Screenshots

I assume it's related to shortsightness trait

Reported by: Emikamiyuki
On server: Goonstation 3 Roleplay: Morty
Active test merges: nothing
Round log date: 2023-11-03-02-43
Reported on: 2023-11-03 03:12:25
Map: DONUT3

pali6:
Were any of the eye traits involved?

TealSeer:
Can reproduce using the blind trait

pgmzeta:
Need to change the behavior of items added to the player at spawn time (go into storage instead of the ground?)

Related/same base issue:
#10323
#12975
#15270

Issue #15270, opened 2023-08-06
Blind/Deaf traits drop gear on the floor before Pilot trait moves the player.


pgmzeta:
### Labels

[BUG]

### Description

When using Deaf/Blind trait, the Pilot trait causes the original equipment to be dropped at your round-start spawn location.

### Steps to reproduce

1. Make a character with Pilot and Deaf traits.
3. Go to your expected spawn location (e.g. Robotics for a Roboticist)
4. The original headset will be on the ground (where you would have spawned in if not for the Pilot trait)

### Expected Behavior

No random floor headsets/glasses on character spawn

### Additional Information & Screenshots

Should dump them in the pod, probably?

Issue #12975, opened 2023-02-05
Headsets and Glasses don't spawn with you if you have heavy sleeper


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Have short-sighted or deaf, ect and heavy sleeper, glasses/headsets spawn at cryo/arrivals

### Steps to reproduce

1. Have blind (I think), deaf, short-sighted traits.
2. Have heavy sleeper.
3. Wake up without sechuds, ect with you
4. Walk over to cryo, arrivals and find your things under cryo sleep unit or shuttle chairs

### Expected Behavior

Spawn with you.

### Additional Information & Screenshots

Reported by: Shiiba
On server: Goonstation 4 Roleplay: Sylvester
Round log date: 2023-02-05-00-46
Reported on: 2023-02-05 01:22:46
Map: DONUT2

Azrun:
Who sleeps with thier glasses on? </s>

Issue #10323, opened 2022-08-18
spawning with deaf in a location other than the latejoin arrivals causes a headset to spawn at arrivals


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

spawning in a location other than the designated latejoin arrival location (stowaway, heavy sleeper, diner patron) causes your department-specific headset to spawn in at wherever the latejoin arrivals location is

### Steps to reproduce

have "deaf" as a character trait
spawn in somewhere using heavy sleeper, stowaway, or space diner patron
no department-specific headset spawns in the location where you start your round
there will be a corresponding departmental headset for your job that has spawned in on the floor at arrivals

### Expected Behavior

for the headset to either spawn with me or not spawn at all

### Additional Information & Screenshots

Reported by: Nefarious6th
On server: Goonstation 1 Classic: Heisenbee
Round log date: 2022-08-18-12-09
Reported on: 2022-08-18 12:42:27

TobleroneSwordfish:
uuugh race conditions

PR #18196, opened 2024-03-13
Fixes Blind staff dropping their stuff in arrivals

Esstoo:
[FIX][CLOTHING][TRAIT]

## About the PR 
People with Either the Blind, Short-Sighted or Deaf Trait combined with a Background Trait no longer spawn their stuff in arrivals at roundstart

## Why's this needed? 
Should Fix all of these:
#16758
#10323 
#12975 
#15270

### DIFF
--- a/code/mob/living/carbon/human.dm
+++ b/code/mob/living/carbon/human.dm
@@ -2282,7 +2282,7 @@
 			src.drop_from_slot(current, get_turf(current))
 	src.force_equip(I, slot)
 	return TRUE
-
+///Tries to put an item in an available backpack, pocket, or hand slot; will delete the item if unable to place.
 /mob/living/carbon/human/proc/stow_in_available(obj/item/I)
 	if (src.autoequip_slot(I, SLOT_IN_BACKPACK))
 		return
--- a/code/mob/living/carbon/human.dm
+++ b/code/mob/living/carbon/human.dm
@@ -2294,7 +2294,7 @@
 		return
 	if (src.autoequip_slot(I, SLOT_R_HAND))
 		return
-	else qdel(I)
+	qdel(I)
 
 /mob/living/carbon/human/swap_hand(var/specify=-1)
 	if(src.hand == specify)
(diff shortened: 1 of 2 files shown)


ZeWaka (changes requested):
yeah no, this code is super wack, and please indent your function bodies on a newline like i said

Esstoo:
> alternatively, do the heavy sleeper thing and glue the old items onto the newly spawned person instead. Is it jank? yeah. Is it established precedent? Also yeah ([61b4859](https://github.com/goonstation/goonstation/commit/61b4859d288b0bbc4f264b668cba1df2886de028))

Makes more sense in my head if it stores it in your inventory.
If the backpack is full, it ends up in your hand, which will already glue the item to you anyway if you're a heavy sleeper.

code/mob/living/carbon/human.dm
```diff
@@ -2283,6 +2283,19 @@
...
+		return
+	if (src.autoequip_slot(I, SLOT_L_STORE))
+		return
+	if (src.autoequip_slot(I, SLOT_R_STORE))
+		return
+	if (src.autoequip_slot(I, SLOT_L_HAND))
+		return
+	if (src.autoequip_slot(I, SLOT_R_HAND))
+		return
+	else qdel(I)
```
pgmzeta, line 2297: `else` would only trigger if the *last* `if` is false
```suggestion
	qdel(I)
```

code/mob/living/carbon/human.dm
```diff
@@ -2283,6 +2283,19 @@
 	src.force_equip(I, slot)
 	return TRUE
 
+/mob/living/carbon/human/proc/stow_in_available(obj/item/I)
```
pgmzeta, line 2286: small doc-comment to explain functionality
```suggestion
///Tries to put an item in an available backpack, pocket, or hand slot; will delete the item if unable to place.
/mob/living/carbon/human/proc/stow_in_available(obj/item/I)
```
```

pgmzeta (changes requested):
two small tweaks, looks good otherwise.

TobleroneSwordfish (approved):

PR #18200, opened 2024-03-13
Reduces the buying prices of Sketchy D-5

IPling:
[BALANCE]
## About the PR 

Reduces Sketchy D-5's buying prices which include:

1. Psilocybin from 600 to 150
2. Cannabis from 150 to 75
3. Rainbow Weed from 600 to 150
4. Whiteweed from 600 to 150
5. Omegaweed from 5000 to 300

## Why's this needed? 

Omegaweed price of 5000 credits per leaf is stupidly overpowered, one produce satchel of omega weed equals to 250k, a quarter of a million.

As well Sketchy D-5 almost has the same buying items as Buford Tenin (The trader frog from QM computer), but always had the best possible selling price. I felt that QM computer traders should be more prioritized to sell to rather than off-station traders, which includes both making money for the cargo budget, and generally on your private account to encourage department interaction.

The proposed price changes make it so that:
1. With Rainbow weed and Whiteweed, its _generally_ better to sell to cargo for a private account, and always if you got access to cargo's budget.
2. With Psilocybin its always better to sell to cargo no matter what, this is mostly due to me keeping the same price hierarchy and not trying to align it with Bufords or vice versa. Since Buford praises shrooms more valuable than Sketchy D-5 for whatever reason.
3. Omegaweed is always better to sell to Sketchy D-5 for a private account, unless you got access to cargo's budget.

As well cargo will also be receiving more traffic from people that are making money off of weed or psilocybin, and department interaction is always nice.

## Changelog 

```changelog
(u)IPingu
(+)Lowers the bathroom robot's buying prices.
```


TobleroneSwordfish:
This seems good, but it's worth noting that this is afaik one of the few ways to still get wizard chickens, so they may need the amount required lowering @UrsulaMejor?

IPling:
> The diner trader is reliable there, has reliable prices and i **can personally make sure I receive the money**. Nerfing the price won't solve the problem you try to solve.

I dont see it as a major problem for botanists to be going off-station to obtain cash, its actually quite beneficial for antags to have reliable sources so that they are not fucked over by RNG and have their round ruined when doing a gimmick. 

This PR's main objective was a combination between the prices being too OP, like omegaweed, and botanists being encouraged to grow the same weed plant each time that requires no player interaction and gives little nuance. With these prices changes there is actual reward for interacting with cargo and even at the start of the shift seeing ''Hey QM, how much is Buford buying x thing for?'' since it may make the botanist do something else than going to D-5. Yet if you still get unlucky with prices, no communication with cargo or just a grumpy QM, Sketchy D-5 is still a good and reliable alternative with omegaweed or atleast I hope the prices are good.

PR #18206, opened 2024-03-13
Port and implement RestrictedInput TGUI component from upstream

Sovexe:
## About the PR 
* Ports the RestrictedInput component from upstream. Not quite a 1 to 1 as they've moved off of inferno. I had to refactor some bits like ReplaceAll which dont exist for us
* Implements the RestrictedInput componet in NumberInputModal
  * minor difference in arg names, upstream uses `round_value` where as we use `round_input`

## Why's this needed? 
* RestrictedInput brings with it autofocus and autoselect (which was the original purpose of this PR)


mordent-goonstation:
LGTM. When we upgrade Inferno -> React would be worth looping back around to porting these back upstream so we're all in sync.

mordent-goonstation (approved):

ZeWaka:
!merge_upstream

ZeWaka:
!merge_upstream

ZeWaka:
L+conflict+ratio

ZeWaka:
!merge_upstream

ZeWaka:
!merge_upstream

ZeWaka:
what a shame, L

ZeWaka:
!merge_upstream

Sovexe:
converting back to a draft...there are some bugs to be addressed

Sovexe:
Something I noticed with this (and with upstream) the input proc in dm still has an arg named `round_input` (`round_value` in upstream) but if its set to true you won't ever have anything to round. This is because the way the sanitization works numbers like `1.7` become `17`, the decimal is treated as a bad character to be stripped out. The round call does still occur in dm, but unless someone is exploiting shouldn't ever have a decimal to round. 

But that's how upstream has it so that's how it's been implemented here

ZeWaka (changes requested):
zewaka needs to check licenses

tgui/packages/tgui/components/RestrictedInput.js
```diff
@@ -0,0 +1,307 @@
+/**
+ * @file
+ * @copyright 2021
+ * @author jlsnow301 (https://github.com/jlsnow301)
+ * @license
```
ZeWaka, line 5:
```suggestion
 * @license MIT
```

ZeWaka (approved):

ZeWaka:
!merge_upstream

PR #18208, opened 2024-03-14
Adding some sound effects and animations for rockbox damage

mloccy:
[OBJECTS][ENHANCEMENT]

## About the PR 
The rockbox can be destroyed, but when a player attacked the rockbox it behaved like an object that could not be destroyed, at least not without an explosion. This also makes the rockbox inoperable if it has been destroyed, unlike before where it was still usable except the remote purchasing of ore didn't work.

## Why's this needed? 
I think if a machine can be destroyed it should show some indication that damage is being done.

```changelog
(u)mloccy
(+)Adding some animations to the rockbox on damage, and made it so when its broken it cannot be used.
```


mloccy:
Video showing new behaviour
https://github.com/goonstation/goonstation/assets/757551/576d6070-d448-4037-98fe-a2971275272d

code/obj/mining_cloud_storage.dm
```diff
@@ -329,7 +352,17 @@
...
+	get_desc(dist, mob/user)
+		. = ..()
+
+		if (src.status & BROKEN)
+			. += "It looks broken and inoperable."
+
+
 	ui_interact(mob/user, datum/tgui/ui)
+		if (src.status & BROKEN)
+			return
```
ZeWaka, line 364: no feedback?

mloccy: Have added some more messages on all the interactions to tell the user why its not working.

Tarmunora (approved):

ZeWaka (approved):

PR #18209, opened 2024-03-14
Two new hairstyles that focus on people of color!

Sockssq:
## About the PR 
Adds two new hairstyles into the game. Long Box Braids and Bubble Braid Pigtails. Focuses on adding more hairstyles with more variety. 
unknown (11)
Hair Styles Finished

## Why's this needed? 
New hairstyles are always good for the creativity of players to expand, and this adds more variety to hairstyles for people of colors. Opening new possibilities for more textured hair in the future.

## Changelog 

```changelog
(u)Sockssq
(*)Added Bubble braid pigtails - pigtails in a bubble braid style, referring to the simple hairstyle that creates a shape similar to a bubble. Added Long Box Braids - a long hair that consists of braids, referring to the protective hairstyle that organizes the hair in a box shape, all the way to the roots to the ends.


DimWhat:
I'd suggest instead of two major changelog entries you should mark these as one major or just one minor.

WALPVRGIS (changes requested):
I love these, thank you for making them. c:

In terms of feedback; I'd like to see more contrast with the outlines and details in the box braids, because while I was testing these, they lost a lot of the detail while the colour overlay was applied (detail is more obvious in this zoomed in screenshot, but at 64px scaling the style is quite indistinct);
image
Could probably smooth out some of the darker pixels near the top of the hairstyle as well, let me know if you need an example of what I mean. 👍 

A darker outline on the bubble braids, at the lower half in the side views would make them more distinct too. 😸

Sockssq:
> I love these, thank you for making them. c:
> 
> In terms of feedback; I'd like to see more contrast with the outlines and details in the box braids, because while I was testing these, they lost a lot of the detail while the colour overlay was applied (detail is more obvious in this zoomed in screenshot, but at 64px scaling the style is quite indistinct); image Could probably smooth out some of the darker pixels near the top of the hairstyle as well, let me know if you need an example of what I mean. 👍
> 
> A darker outline on the bubble braids, at the lower half in the side views would make them more distinct too. 😸

I saw this the other day and I just found out I lost the file of the long box braids so now I cant edit them. So if you could tell me how to save the files in dm as pngs I will get on this. (Also yes, I would like an example what you meant.)

WALPVRGIS:
> > I love these, thank you for making them. c:
> > In terms of feedback; I'd like to see more contrast with the outlines and details in the box braids, because while I was testing these, they lost a lot of the detail while the colour overlay was applied (detail is more obvious in this zoomed in screenshot, but at 64px scaling the style is quite indistinct); image Could probably smooth out some of the darker pixels near the top of the hairstyle as well, let me know if you need an example of what I mean. 👍
> > A darker outline on the bubble braids, at the lower half in the side views would make them more distinct too. 😸
> 
> I saw this the other day and I just found out I lost the file of the long box braids so now I cant edit them. So if you could tell me how to save the files in dm as pngs I will get on this. (Also yes, I would like an example what you meant.)

Sorry for slow reply, it's been a hectic week! But you should be able to click to copy the directionals, and then paste them into aseprite/your sprite editor of choice :) If you're having trouble with them, pls ping me in the spriter channel and i'll export them for you!

Here's an example of what I mean re: contrast;
image
adding some darker greys to the palette makes the braids look a little more like braids when coloured, especially at smaller resolutions (as you'd see it ingame).

PR #18210, opened 2024-03-14
Prevent dropping 2 handed items from your inventory with only 1 free hand.

Valtsu0:
[objects][qol]
## About the PR 
Currently clicking on a 2 handed item while having only one hand drops it on the ground. This pr prevents that if the item is "in" you. So you can still take 2 handed items out of backpacks on the ground and pickpocket them from people you are pinning, but not take the Hadar heavy power-sword from your back while holding a sawfly remote.

If you have a 2 handed item on your back and you have lost one of your hands you are going to need some help.

## Why's this needed? 
[Thread](https://forum.ss13.co/showthread.php?tid=22409)
Removes a bunch of accidentally dropping stuff, mostly weapons during combat, because you forgot you were holding an item.

### DIFF
--- a/code/atom.dm
+++ b/code/atom.dm
@@ -1417,3 +1417,11 @@ TYPEINFO(/atom/movable)
 		if("icon_state")
 			src.icon_state = oldval
 			src.set_icon_state(newval)
+
+/atom/movable/proc/is_that_in_this(atom/movable/target)
+	if (target.loc == src)
+		return TRUE
+	for(var/atom/movable/AM in container)
+		if(AM.contents.len && AM.is_that_in_this(target))
+			return TRUE
+	return FALSE
--- a/code/atom.dm
+++ b/code/atom.dm
@@ -1422,5 +1422,5 @@ TYPEINFO(/atom/movable)
 	if (target.loc == src)
 		return TRUE
 	if (istype(target.loc, /atom/movable))
-	    return src.is_that_in_this(target.loc)
+		return src.is_that_in_this(target.loc)
 	return FALSE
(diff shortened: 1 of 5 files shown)


DisturbHerb:
Personally unsure about this, I've done this deliberately a few times.

code/mob/living/carbon/human.dm
```diff
@@ -1865,6 +1865,15 @@
 				return TRUE
 	return FALSE
 
+/mob/living/carbon/human/can_hold_two_handed()
+	if (src.r_hand || src.l_hand)
```
pgmzeta, line 1869: parent call to constrain future behavior drift
```suggestion
/mob/living/carbon/human/can_hold_two_handed()
    . = ..()
	if (src.r_hand || src.l_hand)
```

code/atom.dm
```diff
@@ -1417,3 +1417,11 @@ TYPEINFO(/atom/movable)
 		if("icon_state")
 			src.icon_state = oldval
 			src.set_icon_state(newval)
+
+/atom/movable/proc/is_that_in_this(atom/movable/target)
+	if (target.loc == src)
+		return TRUE
+	for(var/atom/movable/AM in src)
+		if(AM.contents.len && AM.is_that_in_this(target))
+			return TRUE
```
pgmzeta, line 1426: Might be more performant to recursively walk up the contents tree instead of iterating over all contents?

```suggestion
	if (istype(target.loc, /atom/movable))
	    return src.is_that_in_this(target.loc)
```

pgmzeta:
ah, my bad on the build error; i think the suggestions had spaces instead of tabs

TobleroneSwordfish (approved):
Yeah I like this, it's almost never intentional and often very annoying.

Sord213:
good change. I'd love to see it so you can wield 2handed items 1handed with a severe downside, but that's a whooooole other thing.

PR #18212, opened 2024-03-15
Gang rework: Gang crates

TDHooligan:
## About the PR 
Adds the new gang crate loot generation system
Contains the new crate system, without the spawning mechanics (these are tied to the gamemode)

Contains 2 new container items, and the means to populate them with loot:
Gang Crate:
image

Contains a 4x4 grid of loot, by default, uses guns_and_gear, with the following tokens:
3 Firearm
3 Ammo/Gear
3 Gear

Duffel Bag;
imageimage
Sprites by Moriarty

Can be opened by gang members, but they have to work with civilians to find them.
Contains a 3x2 grid of loot, by default, uses guns_and_gear, with the following tokens:
1 Firearm
1 Ammo/Gear
2 Gear

## Mechanics
Multiple times a shift, each gang has a couple of Duffel Bags spawn around the map, their positions are then revealed via PDA to a random unaffiliated civilian. Gang members learn of the name of the civilian, and must find and work with them to get their loot. 

Similarly, but only twice a shift - halfway through, and towards the end - a Gang Crate will spawn in the middle of a room, locked and anchored into place. Its location is immediately revealed to all gang members via the radio.

After 5 minutes, the crate is unanchored, whereby the gangs must fight to take the crate to their locker, where it can be unlocked for bonus points and the contents.

## Spawn Mechanics

There is a debug spawnable - /obj/storage/crate/gang_crate/guns_and_gear_visualized that demonstrates this in realtime, if you're interested or debugging weirdness.

The loot system works with an item grid (think Resident Evil 4), and 'Tokens', defining what can spawn inside.
1 'Gun' token, for example, means one 'Gun' item.

By default, these items are spawned with 'spawn_random_item' - This picks a random empty location, then picks a random size of loot, based on how much room is available.
It will then spawn a randomly sized item, given the token passed to it. 

The items' icon is scaled and moved into position according to where they spawned, using offsets defined in the spawn object. Then passed the transform_on_pickup component to reset their icon on pickup.

A quick visualisation, slightly outdated (crates have been changed to 4x4 instead of 3x4)
image

These items can be defined one after the other and be picked up by the loot system, and instantly weighed into their respective pool. 

## Why's this needed? 

my name is dill behrt and i keep giving GUNS to STAFF ASSISTANTS


code/datums/components/transform_on_pickup.dm
```diff
@@ -0,0 +1,10 @@
+/// Resets the transformation on an item when it's picked up. Used for resizing things coming out of gang crates.
+/datum/component/transform_on_pickup
```
TobleroneSwordfish, line 2: This isn't what I'd expect this component to do from the name, maybe `reset_transform_on_pickup` would be clearer (if horribly verbose)

code/obj/storage/gang_crate.dm
```diff
@@ -0,0 +1,1257 @@
...
+			UpdateIcon()
+		SPAWN((GANG_CRATE_LOCK_TIME - 15 SECONDS) )
+			src.light = image('icons/obj/large_storage.dmi',"gangcrateblinkinglight")
+			UpdateIcon()
+		SPAWN(GANG_CRATE_LOCK_TIME)
+			src.light = image('icons/obj/large_storage.dmi',"gangcratefulllight")
+			anchored = FALSE
+			UpdateIcon()
+
+	ex_act()
```
TobleroneSwordfish, line 130: Might also want to consider `emag_act` and `meteor_hit` as rare as those are going to be.

TobleroneSwordfish (changes requested):
Some general things:
- `.len` is bad, use `length(foo)` instead.
- All proc references should use the `PROC_REF` macro or they will break in 515.
- You're tending to treat byond lists like C style arrays a lot. It's not necessarily wrong, it's just very jarring to read compared to the rest of the codebase.

This might not be practical, but it looks really weird when a crate is moved and the neatly arranged contents stay behind:
image
You could put the contents "inside" the opened crate and expose them via `vis_contents` to stop this happening (`vis_contents` handles clicks and mouseover automatically)

code/obj/storage/gang_crate.dm
TDHooligan: makes sense due to the superclass. ill just return for now unless that's a dumb idea

TobleroneSwordfish (changes requested):
Changes look good to me, the only last two things I can see is that resetting transform doesn't reset pixel offset so if you drop an item from a gang crate it tends to be offset oddly:
image
aand pulling the webley from your pocket has a 5% chance of shooting yourself because it uses the spin emote. Decide for yourself whether this is intentional :)

TobleroneSwordfish (approved):

code/obj/storage/gang_crate.dm
```diff
@@ -0,0 +1,1262 @@
...
+		tier = GANG_CRATE_GEAR
+		spawn_loot(var/C,var/datum/loot_spawner_info/I)
+			spawn_item(C,I,/obj/item/chem_grenade/flashbang,off_x=-4,off_y=4)
+			spawn_item(C,I,/obj/item/chem_grenade/flashbang,off_x=4,off_y=4)
+			spawn_item(C,I,/obj/item/chem_grenade/cryo,off_x=-4,off_y=-4,scale_x=0.8,scale_y=0.8)
+			spawn_item(C,I,/obj/item/chem_grenade/shock,off_x=4,off_y=-4,scale_x=0.8,scale_y=0.8)
+	helmet
+		tier = GANG_CRATE_GEAR
+		spawn_loot(var/C,var/datum/loot_spawner_info/I)
+			var/helmet = pick(filtered_concrete_typesof(/obj/item/clothing/head/helmet, /proc/filter_trait_hats))
```
TobleroneSwordfish, line 1095:
```suggestion
			var/helmet = pick(filtered_concrete_typesof(/obj/item/clothing/head/helmet, PROC_REF(filter_trait_hats)))
```

code/obj/storage/gang_crate.dm
```diff
@@ -0,0 +1,1262 @@
...
+
+	booze
+		spawn_loot(var/C,var/datum/loot_spawner_info/I)
+			spawn_item(C,I,pick(booze_items),off_x=-2)
+			spawn_item(C,I,pick(booze_items))
+			spawn_item(C,I,pick(booze_items),off_x=2)
+
+	hat
+		spawn_loot(var/C,var/datum/loot_spawner_info/I)
+			spawn_item(C,I,pick(filtered_concrete_typesof(/obj/item/clothing/head, /proc/filter_trait_hats)),off_y=-2,scale_x=0.7,scale_y=0.7)
```
TobleroneSwordfish, line 1116:
```suggestion
			spawn_item(C,I,pick(filtered_concrete_typesof(/obj/item/clothing/head, PROC_REF(filter_trait_hats))),off_y=-2,scale_x=0.7,scale_y=0.7)
```

code/obj/storage/gang_crate.dm
```diff
@@ -0,0 +1,1262 @@
...
+			spawn_item(C,I,pick(filtered_concrete_typesof(/obj/item/clothing/head, /proc/filter_trait_hats)),off_y=0,scale_x=0.7,scale_y=0.7)
```
TobleroneSwordfish, line 1117:
```suggestion
			spawn_item(C,I,pick(filtered_concrete_typesof(/obj/item/clothing/head, PROC_REF(filter_trait_hats))),off_y=0,scale_x=0.7,scale_y=0.7)
```

code/obj/storage/gang_crate.dm
```diff
@@ -0,0 +1,1262 @@
...
+			spawn_item(C,I,pick(filtered_concrete_typesof(/obj/item/clothing/head, /proc/filter_trait_hats)),off_y=2,scale_x=0.7,scale_y=0.7)
```
TobleroneSwordfish, line 1118:
```suggestion
			spawn_item(C,I,pick(filtered_concrete_typesof(/obj/item/clothing/head, /proc/filter_trait_hats)),off_y=2,scale_x=0.7,scale_y=0.7)
```
```suggestion
			spawn_item(C,I,pick(filtered_concrete_typesof(/obj/item/clothing/head, PROC_REF(filter_trait_hats))),off_y=2,scale_x=0.7,scale_y=0.7)
```

TobleroneSwordfish (changes requested):
Oh actually wait you still have 514 style proc-refs

code/obj/storage/gang_crate.dm
TDHooligan: oopsie. many thanks

TobleroneSwordfish (approved):

PR #18222, opened 2024-03-16
Make objectives a bit more reasonable

TobleroneSwordfish:
[BALANCE] [REWORK] [GAMEMODES]
## About the PR 
- Reduces the amount of "multigrab" objects required to be stolen across the board
- Removes sunglasses stealing objective because it sucks
- Datumizes multigrab objective because it looked at me funny
- Removes hijack objectives from traitor and arcfiend and gives vampire and changeling unique versions that include their subordinate antags properly
- Reduces the number of assassinate targets on some objective sets
- Removes that one objective set that's just "be useless and die"

## Why's this needed? 
"Why don't people care about objectives on goon?"
The objectives: "Steal 5 tasers, assassinate 4 people, and hijack the emergency shuttle"
- A lot of the old multigrab values essentially required you to mechscan rather than steal them and often wouldn't even fit in your backpack.
- A lot of people on seeing hijack just immediately go "welp, not doing my objectives then". Changeling and vampire have a more reasonable shot at it and tend to do it in a more fun way than other antags anyway. (thrall armies are cool!)
- 4 specific assassinate targets is overwhelming, 2 or 3 is much more reasonable.

Objectives are meant to be a simple starting point or idea of some things to do, not complete round plans no-one will ever actually follow.

## Changelog 

```changelog
(u)LeahTheTech
(*)Reworked a lot of stealing and assassinate objectives to be more realistically completable.
```

### DIFF
--- a/code/datums/gamemodes/gm_parent.dm
+++ b/code/datums/gamemodes/gm_parent.dm
@@ -316,12 +316,8 @@ ABSTRACT_TYPE(/datum/game_mode)
 	if (!istype(traitor) || !isnum(allow_hijack))
 		return null
 
-	var/upperbound = 3
-	if (allow_hijack)
-		upperbound = 4
-
 	var/objective_path = null
-	var/picker = rand(1,upperbound)
+	var/picker = rand(1,3)
 	switch(picker)
 		if(1)
 			objective_path = /datum/objective/escape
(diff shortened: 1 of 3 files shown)


TobleroneSwordfish:
You can still kill everyone if you want to, most people I've seen hijack didn't actually have the objective.
> at most 20% of the station population on board

Veery hard to communicate whether you're going to complete or fail with an arbitrary threshold like that.

PR #18231, opened 2024-03-17
Fix lobby text status runtime

TealSeer:
[BUG] [RUNTIME] [UI]
## About the PR 

Add a null check to the `mapSwitcher.next` line in world.dm

## Why's this needed? 

image
Fixes this runtime that only happens when the server is launched in single-user mode (directly launching the dmb) that seems to happen because `world.update_status()` gets called before `world.init()` creates `mapSwitcher`


TobleroneSwordfish (approved):
No idea if we actually support that launch mode, but this seems fine

TealSeer:
it's how stuff like the vscode launch tasks open the game, so mostly for debugging (which makes it annoying when theres already one runtime there when you start)

TobleroneSwordfish:
Strange, I've never run into that

PR #18232, opened 2024-03-18
Moves the Clarion ranch chute into the kitchen

cheekybrdy:
## About the PR 
Moves the beer keg in the freezer (its a bartender item why is it there?) to the bar and moved the ranch chute output to where it was.

## Why's this needed? 
Moves the beer keg to a correct location and means the chickens get dispensed in kitchen

## Changelog 

```changelog
(u)Cheekybrdy
(+)Moved the ranch chute output in clarion to inside the freezer. The beer keg it replaced is now in the bar.
```


flappybatpal (approved):

cheekybrdy:
Closed to be added in #18169.

PR #18246, opened 2024-03-19
Bank records TGUI implementation

mloccy:
[UI][FEATURE]

## About the PR 
Implements a TGUI interface for the banking system that the HoP uses to adminster the station budget and payroll

## Why's this needed? 
Part of TGUI conversion project

## Screenshots
Dark Mode unauthenticated:
image

Light Mode unauthenticated:
image

Dark Mode authenticated:
image

Light Mode authenticated
image

## Changelog 

```changelog
(u)mloccy
(+)Bank computer has a new and improved TGUI UI
```


amylizzle:
Might not be a terrible idea to put the individual crew accounts on an additional tab, maybe with a search function by name and column sorts.

mloccy:
> Might not be a terrible idea to put the individual crew accounts on an additional tab, maybe with a search function by name and column sorts.

Done.

image

image

image

amylizzle:
nice :+1:

TobleroneSwordfish (changes requested):
Some stuff at a glance without looking through the code:
- The login step is unnecessary and doesn't match similar UIs, see the ID computer
- The "Budget name" and "Balance" headers are unnecessary and confusing
- Salery is spelled salary
- The lines on the account list are too close together and the buttons blend into each other, look into using a `Stack` to space them out automatically.

ZeWaka:
> - The lines on the account list are too close together and the buttons blend into each other, look into using a `Stack` to space them out automatically.

The problem is then it's a stack and not a table. I'd rather see css applied to increase cell padding or whatever

mloccy:
> Some stuff at a glance without looking through the code:
> 
> * The login step is unnecessary and doesn't match similar UIs, see the ID computer
> * The "Budget name" and "Balance" headers are unnecessary and confusing
> * Salery is spelled salary
> * The lines on the account list are too close together and the buttons blend into each other, look into using a `Stack` to space them out automatically.

Thanks for the reviews, here is the new crew account view with added padding:
image

Budget Status:
image

mloccy:
Feedback when unauthorized card is inserted:
image

Tarmunora (approved):

pgmzeta:
please consider moving the salary edit buttons to the left side of the salary column; doing so would remove some at-a-glance ambiguity

mloccy:
> please consider moving the salary edit buttons to the left side of the salary column; doing so would remove some at-a-glance ambiguity

image
Done + added tooltip

mordent-goonstation:
> please consider moving the salary edit buttons to the left side of the salary column; doing so would remove some at-a-glance ambiguity

I'll go one further. Please consider making the actual Salary column a button with their current salary as the contents. Button to the left is not really that clear in general.

tgui/packages/tgui/interfaces/BankingComputer/CrewAccounts.tsx
```diff
@@ -0,0 +1,224 @@
...
+    } else {
+      let config = props.currentColumnConfig;
+      if (config.dir === SortDirection.Asc) {
+        config.dir = SortDirection.Desc;
+      } else {
+        config.dir = SortDirection.Asc;
+      }
+
+      config.field = field;
+      props.onSortClick(config);
```
mordent-goonstation, line 52: This mutates the current `config` that's passed in via a prop. Rewrite to not do so.

mordent-goonstation (changes requested):
Do not merge until comments in #18306 are addressed and similar things applied back into here. Please ping me in Discord (`@mordent`) when you've done so and I'll review this one.

tgui/packages/tgui/interfaces/BankingComputer/type.ts
```diff
@@ -0,0 +1,73 @@
+/**
+* @file
+* @copyright 2024
+* @author mloccy (https://github.com/mloccy)
+* @license MIT
+*/
+
+import { SortDirection } from "../PlayerPanel/constant";
+export const CREDIT_SIGN = "⪽";
```
mordent-goonstation, line 9: Can we get this stuck into `.../interfaces/common/strings.ts` or something? Follow-up work from that (for a different PR): change usages to have them all point there.

code/obj/machinery/computer/bank_records.dm
```diff
@@ -99,9 +99,10 @@
 
 			if ("edit_wage")
 				var/id = params["id"]
+				var/isSiliconUser = text2num_safe(params["isSiliconUser"])
```
mordent-goonstation, line 102: Do not trust the params for this. Infer from the user of the UI, not passing back as a param.

tgui/packages/tgui/interfaces/BankingComputer/CrewAccounts.tsx
```diff
@@ -170,32 +173,36 @@ export const CrewAccounts = (props: CrewAccountsProps, context) => {
...
 
-    let accounts = props.accounts;
+    let accountsToSort = accounts;
     if (sortConfig) {
-      accounts.sort((a, b) => compareByField(a, b, sortConfig));
+      accountsToSort = [...accounts].sort((a, b) => compareByField(a, b, sortConfig));
 
       if (sortConfig.dir === SortDirection.Asc) {
-        accounts.reverse();
+        accountsToSort.reverse();
```
mordent-goonstation, line 187: Not too opposed to this, but would prefer the `.sort` comparator could be configured with a direction. This would reduce the amount of operations on the data.

tgui/packages/tgui/interfaces/BankingComputer/CrewAccounts.tsx
```diff
@@ -170,32 +173,36 @@ export const CrewAccounts = (props: CrewAccountsProps, context) => {
 
   };
 
-  const searchAndSortCrewAccounts = (filter: string, text: string, sortConfig: CrewColumnSortConfig) => {
+  const searchAndSortCrewAccounts = (
```
mordent-goonstation, line 176: Looks like this is a util function rather than part of the render, if this doesn't depend on props/state then extract out of the component so it's not regenerated each render (plus makes the code more bitesize).

tgui/packages/tgui/interfaces/BankingComputer/type.ts
```diff
@@ -69,4 +69,5 @@ export interface BankComputerStatus {
   authenticated: boolean,
   cardInserted: boolean,
   cardName: string,
+  isSiliconUser: boolean,
```
mordent-goonstation, line 72: Noting here, though this applies to a bunch of this file: data that comes from `ui_data` is given with `BooleanLike` instead of `boolean`.

mordent-goonstation (changes requested):

tgui/packages/tgui/interfaces/BankingComputer/BudgetSection.tsx
```diff
@@ -0,0 +1,67 @@
...
+    </Table.Row>
+
+  );
+};
+
+interface BudgetSectionProps {
+  budgets: Budget[]
+}
+
+const BudgetSummery = (props: BudgetSectionProps) => {
```
mordent-goonstation, line 42:
```suggestion
const BudgetSummary = (props: BudgetSectionProps) => {
```
Sharing props like this isn't ideal if they just happen to be the same, prefer having them both extend a shared one if needed.

tgui/packages/tgui/interfaces/BankingComputer/BudgetSection.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const sum = props.budgets.reduce((acc, budget) => {
+    return acc + budget.amount;
+  }, 0);
+  const newBudget = { "name": "Total Funds", "amount": sum };
```
mordent-goonstation, line 46:
```suggestion
  const newBudget = { name: "Total Funds", amount: sum };
```

tgui/packages/tgui/interfaces/BankingComputer/BudgetSection.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <BudgetLine rowClassName="budgetSummery" allowTransfer={false} currentBudget={newBudget} />
```
mordent-goonstation, line 48:
```suggestion
    <BudgetLine rowClassName="budgetSummary" currentBudget={newBudget} />
```
Assuming the default is not true, you can handle hard-coded false boolean props by just not setting them.

tgui/packages/tgui/styles/interfaces/BankingComputer.scss
```diff
@@ -0,0 +1,33 @@
...
+.budgetButtonCell {
+  padding-top: 5px;
+  float: right;
+}
+
+.budgetStatusCell {
+  padding-top: 5px;
+  padding-bottom: 5px;
+  vertical-align: middle;
+}
```
mordent-goonstation, line 24: Not sure, but I feel like this can be probably be done with props instead of style overrides?

tgui/packages/tgui/styles/interfaces/BankingComputer.scss
```diff
@@ -0,0 +1,33 @@
...
+
+.budgetSummery {
+  font-size: larger;
+}
+
+.crewAccountCell {
+  padding-top: 5px;
+  // padding-bottom: 5px;
```
mordent-goonstation, line 32: Remove.

tgui/packages/tgui/styles/interfaces/BankingComputer.scss
```diff
@@ -0,0 +1,33 @@
...
+ * @file
+ * @copyright 2024
+ * @author mloccy (https://github.com/mloccy)
+ * @license MIT
+ */
+
+@use '../colors.scss';
+@use '../base.scss';
+
+.headerRow {
```
mordent-goonstation, line 11: These all pollute the heck out of the top-level namespace and may unintentionally affect other interfaces (either existing or future), make sure to wrap these in some interface-level namespace. Take a look at some of the other interface style overrides for examples; they're all done a bit differently, but broadly you want to make sure some top-level class is applied.

tgui/packages/tgui/interfaces/BankingComputer/CrewAccounts.tsx
```diff
@@ -0,0 +1,233 @@
...
+import { TransferButton } from "./TransferButton";
+import { ColumnSortField, CrewAccount, CrewColumnSortConfig, SearchFilter } from "./type";
+
+interface CrewAccountLineProps {
+  account: CrewAccount,
+  isSiliconUser: boolean
+}
+
+const CrewAccountLine = (props: CrewAccountLineProps, context) => {
+  const { data, act } = useBackend(context);
```
mordent-goonstation, line 22: Doesn't look like you're using `data` here.

tgui/packages/tgui/interfaces/BankingComputer/CrewAccounts.tsx
```diff
@@ -0,0 +1,233 @@
...
+      </Table.Cell>
+      <Table.Cell className="crewAccountCell">{props.account.balance.toLocaleString()}{CREDIT_SIGN}</Table.Cell>
+      <Table.Cell className="crewAccountCell"><TransferButton frozen={props.account.frozen} id={props.account.id} type={"crew"} /></Table.Cell>
+    </>
+  );
+};
+
+
+interface CrewAccountsTableHeaderProps {
+  onSortClick: (config: CrewColumnSortConfig) => any,
```
mordent-goonstation, line 40:
```suggestion
  onSortClick: (config: CrewColumnSortConfig) => void,
```

tgui/packages/tgui/interfaces/BankingComputer/CrewAccounts.tsx
```diff
@@ -0,0 +1,233 @@
...
+          Account Balance
+        </Header>
+      </Table.Cell>
+    </Table.Row>
+  );
+};
+
+interface CrewAccountsTableProps {
+  accounts: CrewAccount[],
+  onSortClick: (config: CrewColumnSortConfig) => any,
```
mordent-goonstation, line 111:
```suggestion
  onSortClick: (config: CrewColumnSortConfig) => void,
```

mordent-goonstation (commented):
Oh, yeah, this one was after the previous one so I'll let you go smack this one a bit before I review in more detail.

PR #18251, opened 2024-03-20
resprites grass

WALPVRGIS:
## About the PR 
This resprites the basic grass used for example in botany etc, as well as the grass-gro grass with lil flowers. I also added a snowy version because I drew it and then realised that the existing grass is used in azones and involves a bunch of bespoke assets so uh. The new snowy grass isn't used for anything. :') Feel free to use it in mapping stuff! The palette matches the calm/rough snow I made a few years ago.

Also adds the new grass and snowy grass to the floor and wall planner.
image

image

image

## Why's this needed? 
Literally sprited these 2 years ago and then adhd got me distracted when they were 95% done oops 😂


WALPVRGIS:
> I think the botany grass is supposed to be astroturf? This looks lovely though I don't think it looks super astroturfy

Depends on the map; some have grass, some have grass renamed to astroturf, and some only have astroturf in the monkey pen etc. It's not something set in stone afaik?

I did actually make an astroturf variation that I can add as an actual astroturf, if other devs would prefer the option;

image

I do think the new grass would look much better at centcom however, for example, as that's actually earth and would be neat to have it look much more naturey haha.

DimWhat:
Technically all turf is astroturf in space...

wolfolotl:
I agree with ikea that it feels pretty different from what we have now... these new leafy grasses are lovely but they feel too Fancy for standard grass y'know? something less ornate for the everyday stuff feels more tonally fitting imo

ZeWaka:
Ooh I do like your AstroTurf sprite for botany

MintyPhresh:
I've seen these sprites in-game before-- While they look pretty, the leaf textures spread out much wider than I would expect from a blade of grass, and are also much bigger than I'd expect from a blade of grass... It looks closer to a sprouting farm field that's been overseeded than a grassy lawn that people trample over all the time. Ends up looking very visually busy, which is confusing to my eye. These are fine for the floor planner and stuff, but adding them as the default for Botany would not be a good change in its current state IMO. The astroturf texture you posted is a lot closer to what I'd expect from a grass texture

The flowers look lovely, I'd love to see those in game

WALPVRGIS:
> Ooh I do like your AstroTurf sprite for botany

Ngl I think it's super ugly and would not merge it in its current state. :') But I'll work on it to make it actually suit our art direction better.

flrsh:
I don't think botany grass is canonically astroturf. I know there are some references to it (it varies by map), but I'm pretty sure that's more due to copy paste than a deliberate decision. I would rather have real grass cuz it looks nicer (I like the new sprites a lot and personally don't find them visually busy but rather just lush) and also feels less sad (I would like bees to enjoy real grass and stuff). (Speaking as Flourish and not as a definitive dev opinion.)

Gannets:
I want walpgrass

Glamurio:
Is this still in consideration? I think it looks great, both the new astroturf and flowery variant.

WALPVRGIS:
> Is this still in consideration? I think it looks great, both the new astroturf and flowery variant.

it is yeah! just been busy with non-goonstation stuff the past couple of weeks, but I'm working on this still :)

PR #18254, opened 2024-03-20
Adds "examine" desc functionality to admin verbs

TobleroneSwordfish:
[ADMIN]
## About the PR 
Uses extremely cursed macro hackery to let alt + click (or whatever your examine key is) on admin verbs show the "desc" of the verb.

Glowing reviews from other devs include: "🙈", "Consider just making command dashboards"

Feel free to replace all of this with actual TGUI stat panel/dashboards but this kind of works for now.

## Why's this needed? 
We have a lot of verbs, knowing what they do *before* randomly pushing them would be cool.


TobleroneSwordfish:
Oh and this needs some secret changes too if it actually ends up getting merged.

PR #18255, opened 2024-03-20
Engi-Borg atmos transporter rework.

Francisco-Franks:
## About the PR 
The atmosphere transporter has been changed.
It now has 2 modes (pickup and dropoff/interact).
It can now also pick up smaller tanks.
It can now pick up and drop-off portable atmospherics and tanks on a specific tile instead of your own.
It can use tanks on valid machines/cannisters.

## Why's this needed? 

Engi-borgs can not set up a singularity engine. Well they can it just won't do anything. Borgs are also incapable of doing some ancillary engineering/atmos related tasks. I want to be a helper, dangit.

## Changelog 

```changelog
(u)MetaReferencer
(+)Engineering Cyborg Atmosphere Transporter Rework, Now Featuring : Tanks (the kind that hold gasses)
```


code/obj/submachine/robotics.dm
```diff
@@ -271,26 +271,63 @@
...
+						S.set_loc(A)
+						S.contained = 0
+					else if (istype(selection, /obj/item/tank))
+						var/obj/item/tank/S = selection
+						S.set_loc(A)
+					else return //no sparks for unintended items
+					elecflash(user)
+					return
+				if (istype(A, /obj/machinery/portable_atmospherics) || istype(A, /obj/machinery/power/collector_array)) // It is a portable atmos machine, cannister, or a singulo collector array.
+					var/selection = tgui_input_list(user, "What do you want to use?", "Atmospherics Transporter", src.contents) // Let us choose an item to smack it with.
```
pgmzeta, line 320: you could pre-filter a list of just the relevant tank contents, check its length to prevent no-op tgui popups, and  *also* guarantee that a tank was selected (if one exists).

code/obj/submachine/robotics.dm
```diff
@@ -271,26 +271,63 @@
...
+				src.opmode = 1
+				user.show_text("Operating mode set to Pick-up", "green")
+
+	afterattack(atom/A, mob/user as mob) // I have no idea what I am doing but let's go for it.
+		if (!A) return // I assume this is so we don't try and interact with a non-existent atom.
+		// Time to consider our operating mode.
+		if (BOUNDS_DIST(get_turf(src), get_turf(A)) > 0) // If it is too far away do nothing.
+			return
+		if (src.opmode == 1) // We are in pickup mode, no reason to do anything.
+			return
```
pgmzeta, line 291: it feels weird that the behavior for pickup and interact modes are so separated, while each implementation has an early return for the other mode. however, it just feels like a code smell; is there a reason behind this I'm missing?

pgmzeta (changes requested):
mostly code-style stuff

Note: The code suggestions for this review use named defines
instead of using bare numbers for `opmode`, which makes the code much easier to read. Those suggestions require the following in a file in `_std/defines`:
```
#define ATMOSPORTER_MODE_PICKUP 1
#define ATMOSPORTER_MODE_INTERACT 2
```

code/obj/submachine/robotics.dm
Francisco-Franks: If you have, say, an air tank (normal human wearable air tank) and you want to attach it to a high capacity air cannister, you do not want to have it pick up the cannister with the attack.

Francisco-Franks: No wait a minute that is backwards. Why did I do this again?

Francisco-Franks: Going to file this under "security-blanket" and call it a day. It isn't hurting anything by existing.

code/obj/submachine/robotics.dm
Francisco-Franks: I could, but I do not know how to do this. Is it a deal breaker to not filter out non-tanks? We could file this under "expanding the player's freedom to make the wrong choice".

code/obj/submachine/robotics.dm
Francisco-Franks: Going to file this under "i think i made obsoleted this".

code/modules/atmospherics/portable/canister.dm
```diff
@@ -398,6 +398,8 @@ ADMIN_INTERACT_PROCS(/obj/machinery/portable_atmospherics/canister, proc/toggle_
 
 	else if(istype(W, /obj/item/atmosporter))
 		var/obj/item/atmosporter/porter = W
+		if (porter.opmode == 2)
```
pgmzeta, line 401: just this one

pgmzeta (approved):

code/modules/atmospherics/portable/canister.dm
Francisco-Franks: Woops, missed one.

PR #18263, opened 2024-03-21
Adds Disguised Module (Fixing my mistakes)

JORJ949:
[silicons] [feature] [wiki]
## About the PR 

Adds a freeform law module that looks like an Asimov/NT/Robocop law to Head of Staff traitor uplinks.
(Im sorry)

## Why's this needed? 

The current silicon traitor item, the AI laser camera module sucks and I think a better law based traitor item for heads of staff would be great.

## Changelog 

```changelog
(u)JORJ949
(*)Added Disguised law module to head of staff traitor uplinks.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -1036,6 +1036,18 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 	not_in_crates = TRUE
 	job = list("Captain", "VIP", "Regional Director", "Inspector")
 
+/datum/syndicate_buylist/traitor/ai_disguised_module
+	name = "AI Disguised Law Module"
+	name = "Disguised AI Law Module"
+	item = /obj/item/aiModule/disguised_module
+	cost = 2
+	vr_allowed = FALSE
+	not_in_crates = TRUE
+	desc = "An AI law that at a glance looks completely normal."
+	desc = "An AI law module that at a glance looks completely normal, but could tell the AI to do anything."
+	job = list("Captain", "Head of Personnel", "Research Director", "Medical Director", "Chief Engineer")
+	can_buy = UPLINK_TRAITOR
+
 /////////////////////////////////////////// Surplus-exclusive items //////////////////////////////////////////////////
 
 ABSTRACT_TYPE(/datum/syndicate_buylist/surplus)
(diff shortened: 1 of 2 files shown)


code/obj/item/ai_modules.dm
```diff
@@ -346,6 +346,27 @@ ABSTRACT_TYPE(/obj/item/aiModule/syndicate)
...
+		src.lawText = lawTarget ? lawTarget : "Commit random acts of evil."
+		return ..()
+
+	attack_self(mob/user)
+		var/lawTarget = input_law_info(user, "Freeform", "Please enter anything you want the AI to do. Probably murder.", src.lawText)
+		if(lawTarget)
+			src.update_law_text(user, lawTarget)
+			if (lawTarget != initial(lawText))
+				phrase_log.log_phrase("ailaw", src.get_law_text(allow_list=FALSE), no_duplicates=TRUE)
+		return
```
Mister-Moriarty, line 369: `/obj/item/aiModule/disguised_module` should be a subtype of `/obj/item/aiModule/freeform`; this would permit you to forgo reimplementing the freeform `update_law_text()` and `attack_self()` procs.

code/datums/syndicate_buylist.dm
```diff
@@ -1036,6 +1036,18 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 	not_in_crates = TRUE
 	job = list("Captain", "VIP", "Regional Director", "Inspector")
 
+/datum/syndicate_buylist/traitor/ai_disguised_module
+	name = "AI Disguised Law Module"
+	name = "Disguised AI Law Module"
```
Mister-Moriarty, line 1041: `name` has been implemented twice.

code/datums/syndicate_buylist.dm
```diff
@@ -1036,6 +1036,18 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
...
+	item = /obj/item/aiModule/disguised_module
+	cost = 2
+	vr_allowed = FALSE
+	not_in_crates = TRUE
+	desc = "An AI law that at a glance looks completely normal."
+	desc = "An AI law module that at a glance looks completely normal, but could tell the AI to do anything."
```
Mister-Moriarty, line 1047: `desc` has been implemented twice.

Mister-Moriarty (changes requested):

code/obj/item/ai_modules.dm
JORJ949: In my last PR it was pointed out that that may cause an issue with the freeform steal objective, which has since been removed.

Mister-Moriarty: > ...freeform steal objective, which has since been removed.

Even if an objective still existed, it would be trivial to replace an `is_type()` check with an `is_type_exact_check()`, or if necessary, implement a variable onto the freeform type that determines whether a particular type qualifies as a freeform module for the purpose of stealing.

PR #18270, opened 2024-03-22
Gossip NPC

pgmzeta:
## About the PR 
New gimmick NPC with a spate of random gossip phrases centered around various Luna's in-world locations.

Could be added to the lunaport setpiece (or spawned anywhere)

## Why's this needed? 
NPC for Lunaport setpiece

## TODO
- [ ] MobAI
  - [ ] wander in specific area (i.e. the bar)
  - [ ] sit down for a bit
  - [ ] only pick up and drink from glassware
  - [ ] maybe even set them down like real people?
- [ ] Add to Lunaport map


cogwerks:
Looks cool to me

ZeWaka (changes requested):
not super happy with adding unused content

PR #18272, opened 2024-03-22
Albatross

MylieDaniels:
## About the PR 
This adds a gun that I have owed cogwerks for far too long.
The Albatross is a blatantly ridiculous four bore big game rifle that chambers two bullets because I said so.
image
image
image
It also has a lovely (to me at least) boom+rattle when you shoot it, mixed from two sounds - throwing a super heavy cathode ray television into a dumpster and dropping rocks into a metal bucket.

**The items added here (the Albatross and four bore ammo) are not obtainable without being spawned.**

There are two ammo types, termination round (lethal) and roundhouse slug (sublethal).
image
Termination deals far too much damage (95), pierces one time, ignores 40% of armor, and generally is able to kill things deader than dead.
Roundhouse is just painful and incredibly stamina depleting, not deadly, but both throw any dummy that got hit by it around, breaks windows, heavily damages doors, etc.

## Why's this needed? 
pew :)


Tarmunora (changes requested):
need to use `P.power` to actually get projectile power at point of impact instead of the base power of the projectile datum

MylieDaniels:
> need to use `P.power` to actually get projectile power at point of impact instead of the base power of the projectile datum

Oops, thank you!

DecidedlyUnethical:
wait, where does the harness go when you have it in hand?

MylieDaniels:
> wait, where does the harness go when you have it in hand?

Same as all other large backslung guns... hammerspace. I can sprite up a bag for it, like the flintlock rifle bag, so that you can store the gun and a bit of supplies instead of just the gun on your back, if requested.

code/modules/projectiles/bullet.dm
```diff
@@ -1107,6 +1107,109 @@ toxic - poisons
...
+		if (hit && ismob(hit))
+			var/mob/M = hit
+			if(ishuman(hit))
+				var/mob/living/carbon/human/H = hit
+				#ifdef USE_STAMINA_DISORIENT
+				H.do_disorient(max(P.power,10), weakened = 2 SECONDS, stunned = 2 SECONDS, disorient = 0, remove_stamina_below_zero = FALSE)
+				#else
+				H.changeStatus("stunned", 4 SECONDS)
+				H.changeStatus("weakened", 3 SECONDS)
+				#endif
```
Tarmunora, line 1208: somewhat redundant overlap with also having stun power on the projectile itself

code/obj/item/gun/kinetic.dm
```diff
@@ -2094,6 +2094,40 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+	flags =  FPRINT | TABLEPASS | CONDUCT | USEDELAY | EXTRADELAY
+	c_flags = EQUIPPED_WHILE_HELD | ONBACK
+	slowdown = 10
+	slowdown_time = 8
+	force = MELEE_DMG_RIFLE
+	two_handed = TRUE
+	can_dual_wield = FALSE
+	contraband = 7
+	ammo_cats = list(AMMO_FOUR_BORE)
+	spread_angle = 2
```
Tarmunora, line 2114: bit odd to have spread on a scoped weapon imo

Tarmunora (commented):
overall feels like, a bit excessive but I suppose that's the idea with this thing so 🤷

code/obj/item/gun/kinetic.dm
MylieDaniels: I was basing this off old elephant guns, and thought it was funny to include their inaccuracy past short range despite having scopes- to note, this scope only goes out to 16 tiles. I can remove the inaccuracy, if you want.

code/modules/projectiles/bullet.dm
MylieDaniels: Thank you!

PR #18275, opened 2024-03-22
The Authentication Disk can unauth and issue emergency armory authorization.

pgmzeta:
[GAME OBJECTS][BALANCE]
## About the PR 
* Add a new variable to the armory auth computer to track if it has ever been enabled by the authentication disk
* Share the unauthorization cooldown between HoS and Authentication Disk
* Add logic to detect/process authentication disk usage on the riot computer
* Add help messages to indicate how to auth/unauth the armory.

## Why's this needed? 
Makes the authorization disk more valuable in non-nuclear rounds.
Idea taken from a suggested thread here: https://forum.ss13.co/showthread.php?tid=22435

## Changelog 

```changelog
(u)glowbold
(*)Using the Authentication Disk on the armory computer will revoke armory access.
(*)Once per round, the Authentication Disk can authorize armory access, but doing so will prevent revoking armory access.
```


frawhst:
Wonder if it would fuck anything up if you could also auth the armory with the disk (HoS style)

frawhst:
After some discussion with devs, the above is now an Official Suggestion

Cherman0:
Like, allowing the captain to authorize the armory themselves? While I do actually agree with that (armory is not really that powerful and its kinda weird that the highest rank person doesnt have the highest access) at that point it would be cleaner to just give the captain full station access on their ID,

TobleroneSwordfish:
> at that point it would be cleaner to just give the captain full station access on their ID

Currently the captain doesn't have maxsec access and giving it to them would allow them to walk into the armoury and pick stuff up silently, which is probably a bad idea. It would also make the spare an even more attractive target to steal. The auth disk fits thematically and it would be good to give it a use outside of nukie rounds too.

pgmzeta:
Implemented authing the armory too, and rounded out with text/sound and a small cooldown to prevent ear-smashing announcement spam.

Wisemonster:
Considering cap has a minimum round count, and can't roll antag anymore, this would certainly help on rounds with reduced sec pop for something like nukies.

cheekybrdy:
Biggest no in the west if #18294 doesn't get in. For the same reasons I'm against that PR this gives captains a big power buff, and I don't think its too good for antags. I'd rather the knowledge check that current auth has to open the armoury to at least ensure people are experienced enough to know what they're doing.

Tarmunora:
Worth noting that with a bit of legwork, the auth disc does already contain the netpass to allow auth/unauth of armory via packets

cheekybrdy:
Starting to see powergaming with this with captains HoSliteing by authing and deauthing to grab their own shit. I still think it should be deauth only. maybe have it count as an additional auth instead?

Tarmunora:
I think in the current state this PR just doesn't work - there's nothing stopping captains from just authing, looting the place, walking out, and then locking the door behind them. Either de-auth needs to go (so there's an Actual Consequence:tm: to authorizing it), or there needs to be some additional checks and balances other than "ahelp captains that abuse their ability to freely loot the armory whilst having none of the expectations of the HoS role", because I don't think that's a great solution.

Sbmhawk:
Random idea, 15 minute timer after authing before you can deauth

ThyOverlord:
This is a very valuable PR to me as a sec player. It opens up the ability for sec to see Armory as more than "this is a vault of all the goodies so I can become a juggernaut swatsec" and more as  "lets take bare minimum and reclose it for safety" which I believe is a  preferred perception of it.

I do see the major problems here, especially with captains getting gear for themselves unnecessarily.
and because of that I also have to say that it needs adjustment...

Tarmunora (approved):

PR #18282, opened 2024-03-23
fix ghosts being gang experts

TDHooligan:
## About the PR 

ghosts can still see overlays on respawn. i removed all gang overlays on mind transfer (this should fix thing)


TobleroneSwordfish (changes requested):
Crash after respawning and trying to toggle the overlay when you had it off as a ghost: 
image

PR #18283, opened 2024-03-23
Janktank fixes

TDHooligan:
Crams dead folk back into their corpse when they're janktanked


TobleroneSwordfish:
Mild copy-paste but it'll do for now

PR #18294, opened 2024-03-23
Gives captain a mindshield

Chatauscours:
## About the PR 
Gives captain a mindshield, the same one that secoffs currently have.

## Why's this needed? 
Captain is a very easy target for which a lot of people to mindhack for free AA + gun + instantly rogue AI, which is very unfun for the crew and the captain, and results in captains often getting treated with suspicion by crew.

The main aim is to aim a  more trusted and responsible role, especially now that # 18275 is coming.
(P.S. even if 18275 doesn't get merged, I would still propose this change, as currently both playing as and against mindhacked captains is very often considerably unfun.)

## Changelog 

```changelog
(u)Chatausours
(*)Captains now recieve a mindshield implant.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -216,6 +216,7 @@ ABSTRACT_TYPE(/datum/job/command)
 	announce_on_join = TRUE
 	allow_spy_theft = FALSE
 	allow_antag_fallthrough = FALSE
+	receives_implant = /obj/item/implant/health/security/anti_mindhack
 	wiki_link = "https://wiki.ss13.co/Captain"
 
 	slot_card = /obj/item/card/id/gold


Piesuu:
If you want to mindhack the head of the station you should at least work for it by kidnapping them to cut the implant out or murdering them to have it break!
Definitely a good change that might make them even less of a staffie with AA (as said by someone who has played a lot of captain and never liked the fact they could be instantly subverted)

Retrino:
would rather see some solution to mindhacks themselves being so instant and powerful than continuing the trend of giving anyone with a grain of power a mindshield implant tbh

Cherman0:
remove mindhacks from the game
(seriously antag conversion is metagamer hell and also the non-con stuff is hella gross to me)

Cherman0:
You say that as a "slippery slope" but it would actually be a lot better for balance. If everyone (as in all NT employees) was mindshielded it would make mindhacks require some amount of interaction rather than being 3tc to instantly gank anybody. You'd have to at least beat them in a fight or remove the implant during an unrelated surgery.

Sord213:
Seems like a good change to me. I'm all for making the captain a more serious role.

cheekybrdy:
Against personally, the less Captains are suspicious the more they'll validhunt. Mindhacking just moves the power to a different side. Its still going to have problematic issues either way being this powerful. I'd honestly say revert the Captain no antag rather then this as the additional suspicious kept them in greater check then rn. Out of scope but I'd like to see Captain antag locked to 25+ pop if it returned.

Esstoo:
Slightly Against, captain should be a trustwhorty job, but with all the recent changes regarding the job, I feel like we're encouraging the captain to be a slightly less powerful HoS. I see captains valid-hunting way too often because we're giving them 
more and more reasons to do so. I don't know how we would fix this, but captains needs to be given something to do, preferably not involving security.

> would rather see some solution to mindhacks themselves

I have to agree on this, I understand the arguments as to why the cap should need an implant, but I think mindhacks are the problem and should get a rework. The cloning mindhack works well because antags have to put efforts in to mindhack someone, while with mindhack implants you can just stand next to your victim and convert them in less than a second, it gives no time for the victim to react.
For the meantime tho, maybe captain could get a less powerful mindshield than security, idk how much less powerful we should make it, but it could take a similar approach as to how loyalty implants work against revs.

Btw, we should probably make a discussion thread for this.

PR #18302, opened 2024-03-24
Gang adjustments

TDHooligan:
[GAMEMODES] [BALANCE]
## About the PR 

2 adjustments:
Gang density is now 41% on #1, down from 55%
Given gangs are a lot more sticky than before, this should probably be ok

Gangs in uniform can no longer eat 3 baton hits.
I don't think they need this any more. With the number of firearms, lethals and utilities they have, security needs a bit of an easier time stopping them.

### DIFF
--- a/code/modules/status_system/statusEffects.dm
+++ b/code/modules/status_system/statusEffects.dm
@@ -1277,7 +1277,7 @@
 		maxDuration = null
 		effect_quality = STATUS_QUALITY_POSITIVE
 		var/const/max_health = 30
-		var/const/max_stam = 60
+		var/const/max_stam = 20
 		var/const/regen_stam = 5
 		var/mob/living/carbon/human/H
 		var/datum/gang/gang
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish (approved):

code/modules/status_system/statusEffects.dm
```diff
@@ -1277,7 +1277,7 @@
 		maxDuration = null
 		effect_quality = STATUS_QUALITY_POSITIVE
 		var/const/max_health = 30
-		var/const/max_stam = 60
+		var/const/max_stam = 20
```
TobleroneSwordfish, line 1280: Actually thinking about it, this is pretty low now and we still want people to be incentivized to wear their uniforms. Maybe make it ~40 instead so it juust doesn't cross the threshold for baton hits?

TobleroneSwordfish (changes requested):

code/modules/status_system/statusEffects.dm
TDHooligan: sure. why not

PR #18306, opened 2024-03-25
Sortable columns in APC and SMES power monitor interfaces

mloccy:
[FEATURE][UI]

## About the PR 
Adds sortable columns to the APC / SMES Power monitor.

## Why's this needed? 
There's an awful lot of data been given in the APC power monitor, it would be more usefully presented if the interface was sortable so you can see which APC was drawing the most power, which APCs are turning lights etc off, and other scenarios. The SMES one is a lot less busy but I felt that it would be kinda inconsistant to do it on one but not the other.

## Screenshots
APC Power Monitor
image

SMES Power Monitor
image

## Changelog 

```changelog
(u)mloccy
(+)APC / SMES Power Monitor interface now has sortable columns
```


tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
```diff
@@ -8,7 +8,9 @@
 import { useBackend } from '../../backend';
 import { Box, Chart, LabeledList, Stack, Table, Tooltip } from '../../components';
 import { formatPower } from '../../format';
-import { PowerMonitorApcData, PowerMonitorApcItemData } from './type';
+import { SortDirection } from '../PlayerPanel/constant';
```
ZeWaka, line 11: Probably should pull this out to common or something - maybe common/sorting

tgui/packages/tgui/interfaces/PowerMonitor/type.ts
```diff
@@ -67,3 +68,45 @@ export interface PowerMonitorSmesStaticItemData {
...
+  Active = 5,
+  Load = 6
+}
+
+export type SmesTableHeaderColumnSortState = {
+  dir: SortDirection,
+  field: SmesTableHeaderColumns
+}
+
+export const numericCompare = (a: number, b: number): number => {
```
ZeWaka, line 102: Is this really not a default JavaScript thing

mloccy: I would have hoped so and maybe there is, but I haven't found it if it exists. This could also probably be reduced to a-b and get the same results. I'm following the same approach in #18246 so this could either be reduced to (a - b) in both implementations or keep this implementation but make it common to both to avoid duplication, assuming there's no built in numeric compare-to function somewhere. I'm not the most fluid with javascript yet, I've worked mostly in C++ / .NET where for sure there are STL / standard library funcs that do this.

tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
mloccy: Done, will need to update #18246 depending on merging order but it won't be a big issue.

tgui/packages/tgui/interfaces/PlayerPanel/constant.ts
```diff
@@ -1,18 +1,13 @@
...
+ * @copyright 2024
  * @author Sovexe (https://github.com/Sovexe)
+ * @author mloccy (https://github.com/mloccy)
  * @license ISC
  */
 
 export enum Action {
   JumpToPlayerLocation = 'jump-to-player-loc',
-  OpenPlayerOptions = 'open-player-options',
+  OpenPlayerOptions = ':pen-player-options',
```
ZeWaka, line 11: 🥺

tgui/packages/tgui/interfaces/common/sorting/constant.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @file
+ * @copyright 2021
+ * @author Mordent (https://github.com/mordent-goonstation)
+ * @author Sovexe (https://github.com/Sovexe)
+ * @license ISC
+ */
+
+export enum Action {
+  JumpToPlayerLocation = 'jump-to-player-loc',
+  OpenPlayerOptions = ':pen-player-options',
```
ZeWaka, line 11: 🥺

tgui/packages/tgui/interfaces/PlayerPanel/constant.ts
mloccy: Thanks, not sure how this typo got in

tgui/packages/tgui/interfaces/common/sorting/constant.ts
mloccy: Forgot to save the damn file, thanks.

tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
```diff
@@ -81,48 +83,112 @@ export const PowerMonitorApcGlobal = (_props, context) => {
...
+    }
+  };
+
+const sortApcData
+  = (data: PowerMonitorApcItemData[],
+    names: Record<string, string>,
+    state: ApcTableHeaderColumnSortState): PowerMonitorApcItemData[] => {
+
+    if (state !== null) {
+      data.sort((a, b) => apcDataComparer(a, b, names, state.field));
```
mordent-goonstation, line 171: `Array.sort` mutates the existing array. While that's not necessarily a problem here, better to write sorters to be side-effect free. This is especially true when working on things that are pulled from `props` up the chain.

`const sortedData = [...data].sort((a, b) => ...);`

tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
```diff
@@ -81,48 +83,112 @@ export const PowerMonitorApcGlobal = (_props, context) => {
...
 };
-
 export const PowerMonitorApcTableRows = (props: PowerMonitorApcTableRowsProps, context) => {
   const { search } = props;
   const { data } = useBackend<PowerMonitorApcData>(context);
 
   return (
     <>
-      {data.apcs.map((apc) => (
+      {sortApcData(data.apcs, data.apcNames, props.sortState).map((apc) => (
```
mordent-goonstation, line 191: `props` is deconstructed above, better to keep to that pattern (i.e. deconstruct props (const `{ foo } = props;`) vs. access as `props.foo`). No real issue with use of `props.foo`, but blending the two makes tracking down props usage annoying.

Can do the same with `data`, but as you're accessing them all the same way (as `data.foo`) I have no issue with it. Essentially try to keep to one method of nested access for `props` and one for `data` (can obviously be the same method, but doesn't have to be).

tgui/packages/tgui/interfaces/PowerMonitor/index.tsx
```diff
@@ -8,13 +8,69 @@
...
+
+  const onSetApcHeaderSort = (field: ApcTableHeaderColumns, current: ApcTableHeaderColumnSortState) => {
+    if (current !== null) {
+      if (current.field === field) {
+        let newState = {
+          dir: (current.dir === SortDirection.Asc ? SortDirection.Desc : SortDirection.Asc),
+          field: field,
+        };
+
+        apcSetSortBy(newState);
```
mordent-goonstation, line 30:
```suggestion
        apcSetSortBy({
          dir: (current.dir === SortDirection.Asc ? SortDirection.Desc : SortDirection.Asc),
          field: field,
        });
```

tgui/packages/tgui/interfaces/PowerMonitor/Smes.tsx
```diff
@@ -55,31 +57,117 @@ export const PowerMonitorSmesGlobal = (_props, context) => {
   );
 };
 
-export const PowerMonitorSmesTableHeader = (props, context) => {
+interface PowerMonitorSemsTableHeaderProps {
+  setSortBy: (field: SmesTableHeaderColumns) => any;
```
mordent-goonstation, line 61:
```suggestion
  setSortBy: (field: SmesTableHeaderColumns) => void;
```

tgui/packages/tgui/interfaces/PowerMonitor/Smes.tsx
```diff
@@ -55,31 +57,117 @@ export const PowerMonitorSmesGlobal = (_props, context) => {
...
+  state: SmesTableHeaderColumnSortState;
```
mordent-goonstation, line 62: Prefer calling `sortState` if passing it in as a prop; `state` could mean anything.

tgui/packages/tgui/interfaces/PowerMonitor/index.tsx
```diff
@@ -40,11 +96,25 @@ export const PowerMonitor = (_props, context) => {
...
-                {isDataForSmes(data) && <PowerMonitorSmesTableRows search={search} />}
+                <Table.Row header>
+                  {
+                    isDataForSmes(data)
+                    && <PowerMonitorSmesTableHeader
+                      state={smesSortBy}
+                      setSortBy={(field) => onSetSmesHeaderSort(field, smesSortBy)} />
+                  }
+                </Table.Row>
+                {isDataForSmes(data) && <PowerMonitorSmesTableRows sortState={smesSortBy} search={search} />}
```
mordent-goonstation, line 117: Honestly these are two components at this point, that happen to share an interface.

Rather than doing the conditional inside the `Table.Row` and around the `...TableRows` components individually, do it around the combo then return a fragment with the needed components.

tgui/packages/tgui/interfaces/PowerMonitor/type.ts
mordent-goonstation: If you don't provide a comparator, it'll compare in a way that makes sense for number/strings. One option is to either not pass an argument in, or to set the comparator function to `undefined` if using some kind of configuration or whatever where you're always passing something in to `.sort`.

A downside of using the `undefined` method is that it's not necessarily intuitive what it's doing, and easy to confuse with a "this column cannot be sorted" concept.

Some middle-ground would look like having a set of standard comparators somewhere and exporting named comparators from there (e.g. `numericComparator`) and having *that* set to `undefined` (with a comment).

tgui/packages/tgui/interfaces/common/sorting/constant.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
mordent-goonstation, line 1: Would be `type.ts`, this isn't a constant any more (was used within where it was extracted from).

tgui/packages/tgui/interfaces/PowerMonitor/type.ts
```diff
@@ -67,3 +68,45 @@ export interface PowerMonitorSmesStaticItemData {
...
+  Environment = 3,
+  Load = 4,
+  CellCharge = 5,
+  CellState = 6
+}
+
+export type ApcTableHeaderColumnSortState = {
+  dir: SortDirection,
+  field: ApcTableHeaderColumns
+}
```
mordent-goonstation, line 85:
```suggestion
export type ApcTableSortState = SingleFieldSortState<ApcTableHeaderColumns>;
```
Then add to common/sorting/type.ts:
```
export interface SingleSortState<T> {
  dir: SortDirection;
  field: T;
}
```

tgui/packages/tgui/interfaces/PowerMonitor/type.ts
```diff
@@ -67,3 +68,45 @@ export interface PowerMonitorSmesStaticItemData {
...
+  Input = 3,
+  Output = 4,
+  Active = 5,
+  Load = 6
+}
+
+export type SmesTableHeaderColumnSortState = {
+  dir: SortDirection,
+  field: SmesTableHeaderColumns
+}
```
mordent-goonstation, line 100:
```suggestion
export type SmesTableSortState = SingleFieldSortState<SmesTableHeaderColumns>;
```

tgui/packages/tgui/interfaces/PowerMonitor/index.tsx
```diff
@@ -8,13 +8,69 @@
...
+      if (current.field === field) {
+        let newState = {
+          dir: (current.dir === SortDirection.Asc ? SortDirection.Desc : SortDirection.Asc),
+          field: field,
+        };
+
+        apcSetSortBy(newState);
+      } else {
+        let newState = {
+          dir: current.dir,
```
mordent-goonstation, line 33:
```suggestion
          dir: SortDirection.Asc
```

tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
```diff
@@ -8,7 +8,9 @@
 import { useBackend } from '../../backend';
 import { Box, Chart, LabeledList, Stack, Table, Tooltip } from '../../components';
 import { formatPower } from '../../format';
-import { PowerMonitorApcData, PowerMonitorApcItemData } from './type';
+import { SortDirection } from '../common/sorting/constant';
+import { Header } from '../common/sorting/Header';
+import { ApcTableHeaderColumns, ApcTableHeaderColumnSortState, numericCompare, PowerMonitorApcData, PowerMonitorApcItemData } from './type';
```
mordent-goonstation, line 13: `numericCompare` is not a type/interface/enum (enums are an edge case, but fit in the same bucket as types for logical grouping), should be in a `util` or similar file.

tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
```diff
@@ -81,48 +83,112 @@ export const PowerMonitorApcGlobal = (_props, context) => {
     </Stack>
   );
 };
-
-export const PowerMonitorApcTableHeader = () => {
+interface PowerMonitorApcTableHeaderProps {
+  setSortBy: (type: ApcTableHeaderColumns) => any,
```
mordent-goonstation, line 87:
```suggestion
  setSortBy: (type: ApcTableHeaderColumns) => void,
```

tgui/packages/tgui/interfaces/PowerMonitor/Smes.tsx
```diff
@@ -55,31 +57,117 @@ export const PowerMonitorSmesGlobal = (_props, context) => {
...
+    if (field === SmesTableHeaderColumns.Area) {
+      return names[a[field]].localeCompare(names[b[field]]);
+    } else if (field === SmesTableHeaderColumns.Charging || field === SmesTableHeaderColumns.Active) {
+      if (a[field] === a[field]) {
+        return 0;
+      } else if (a[field] > b[field]) {
+        return 1;
+      } else {
+        return -1;
+      }
```
mordent-goonstation, line 136: If this isn't a string (should be using `localeCompare` as above), or a number (should be using `numericCompare`, as below), what are the types of these fields? Based on context assuming booleans. In either case, extracting out and naming the type of comparison with a comparator function is still a good idea.

tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
```diff
@@ -81,48 +83,112 @@ export const PowerMonitorApcGlobal = (_props, context) => {
...
+
+const sortApcData
+  = (data: PowerMonitorApcItemData[],
+    names: Record<string, string>,
+    state: ApcTableHeaderColumnSortState): PowerMonitorApcItemData[] => {
+
+    if (state !== null) {
+      data.sort((a, b) => apcDataComparer(a, b, names, state.field));
+      if (state.dir === SortDirection.Asc) {
+        data.reverse();
```
mordent-goonstation, line 173: Also mutates. That's fine if you're working on `sortedData` as above (and then return it below).

mordent-goonstation (changes requested):

tgui/packages/tgui/interfaces/PlayerPanel/constant.ts
```diff
@@ -1,8 +1,8 @@
 /**
  * @file
- * @copyright 2021
- * @author Mordent (https://github.com/mordent-goonstation)
+ * @copyright 2024
```
mordent-goonstation, line 3: Copyright year shouldn't change. It's not a "last touched" it's a "copyright since".

tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
mloccy: Fixed, will keep that in mind in the future.

tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
mloccy: Not a fan of the deconsturcting props pattern in general, but do agree that one or the other should be used, at least witihn the same file for consistancy.

tgui/packages/tgui/interfaces/PowerMonitor/Smes.tsx
mloccy: agreed, state is a bad variable name, fixed!

tgui/packages/tgui/interfaces/PowerMonitor/index.tsx
mloccy: Have restructured it to be one component per interface to make it less messy

tgui/packages/tgui/interfaces/common/sorting/constant.ts
mloccy: renamed

tgui/packages/tgui/interfaces/PowerMonitor/type.ts
mloccy: Implemented with the rest of the improvements, I had some fun learning generics in typescript.

tgui/packages/tgui/interfaces/PowerMonitor/Apc.tsx
mloccy: moved sorting code to a common/comparators.ts file since this can be reused, specifically in #18246.

mloccy:
@mordent-goonstation thanks for the detailed review, I think I've addressed most if not all of your feedback and the PR is in a lot better shape due to it, it has been a forcing function for me to learn typescript generics which is a lot more sophisticated than I imagined, and pretty nice overall. I have been able to de-duplicate a lot of the code between the APC / SMES divide, specifically the code that sorts the data for both interfaces and the code that deals with the shared state changes. I was able to slightly reorganize the main component to be a little more sensible after the changes. Looking forward to your thoughts when you get a chance again. 

Regarding #18246, I'm gonna implement what I think you'd like to see from your comments on this PR, (and the direct comments you left already) but I don't think that PR, nor this one, needs to be merged any time soon so I'm happy to wait to get this right.

mordent-goonstation (changes requested):

tgui/packages/tgui/interfaces/common/comparators.ts
```diff
@@ -0,0 +1,87 @@
...
+    default:
+      return undefined;
+  }
+};
+
+
+type RecordKey = (string | number | symbol)
+const RecordTypes = ['string', 'number', 'symbol'];
+
+export const sortTypedArrayByIndex = <V extends Array<unknown>, Lk extends RecordKey = never, Lv = never>(
```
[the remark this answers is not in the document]
mloccy: Have been a bit a busy with other project, got back onto this yesterday, I've reimplemented the PowerMonitor using a (hopefully) easily reusable sortable / searchable table component following your advice. Was able to eliminate all the enums and I think I should be able to reimplement the banking UI with this too.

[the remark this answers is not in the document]
mordent-goonstation: Like the broad strokes, I'll review properly sometime soon.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+            <Table.Cell
+              key={index}
+              {...rest}>
+              {children}
+            </Table.Cell>);
+        } else {
+          return (
+            <Fragment
+              key={index}
+              {...rest}>
```
mordent-goonstation, line 105: What are expecting `{...rest}` to be for a `Fragment`?

tgui/packages/tgui/interfaces/common/sorting/SortableTable/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    <Table.Row
+      {...rest}>
+      {cells.map((config, index) => {
+        const { children, wrapInCell, ...rest } = config;
+        if (!children) {
+          return;
+        }
+        let wrapCell = wrapInCell === undefined
+          ? true
+          : wrapInCell;
```
mordent-goonstation, line 91: Broadly prefer "default off" features. This is also something that could be applied in bulk to the config rather than done in the component. Prefer verbose configurations and lean components for a first pass, as it becomes clear what operations are being done regularly that functionality can be added either as utility functions or baked into a middle-layer component.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+  name: string,
+
+  /** Use local state or shared state; Default: true */
+  useLocalState?: BooleanLike,
+
+  /** Data required for searching through rows */
+  searchState?: SearchState
+
+  /** Props, if provided to be spread on <Table> element */
+  [x: keyof any]: unknown
```
mordent-goonstation, line 131: Better to define this as a specific prop, to avoid accidental inclusion.

i.e. `tableProps?: TableProps`, then earlier on `type TableProps = Record<string, unknown>` (or whatever you need to do to make TypeScript appeased)

tgui/packages/tgui/interfaces/common/sorting/SortableTable/type.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import { BooleanLike } from "common/react";
+import { InfernoNode } from "inferno";
+import { SortDirection } from "../type";
+
+/** Configuration for Sortable Table header row and callback functions */
+export interface SortableTableHeaderConfig {
+
+  /** Presentation of header cell for this field */
+  children: InfernoNode,
```
mordent-goonstation, line 16: `children` as a non-component name is somewhat surprising to the reader, prefer calling this `headerContents` or something that isn't a React/Inferno built-in concept.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/type.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  /** Is this field sortable? */
+  sortable: BooleanLike,
+
+  /** Is this field searchable */
+  searchable: BooleanLike,
```
mordent-goonstation, line 22:
```suggestion
  /** Is this field sortable? */
  sortable?: boolean,

  /** Is this field searchable */
  searchable?: boolean,
```
`BooleanLike`s are a tgui concept for how boolean-ish data gets given to the component via backend state (i.e. `ui_data` proc), as those are *actually* given as `1` and `0` but should be handled as booleans rather than numbers with the caveats involved in doing so.

Essentially, if the items are defined in the UI code, you're making `boolean`s with real `true` and `false` values. Also, no harm in making these optional if you then handle non-existence as "not searchable" or "not sortable", would help keep the config a little leaner.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/type.ts
```diff
@@ -0,0 +1,72 @@
...
+  compareFunc?: (a: unknown, b: unknown) => number,
+
+  /** Conversion from your data to string to allow searching */
+  toString?: (data: unknown) => string,
+
+}
+
+export type SortState = {
+  /** Index into SortableTableHeaderConfig */
+  index: number,
```
mordent-goonstation, line 37: Prefer a more descriptive term here, `columnIndex` (I think that's what this is?)

tgui/packages/tgui/interfaces/common/sorting/SortableTable/type.ts
```diff
@@ -0,0 +1,72 @@
...
+  [x: keyof any]: unknown
+}
+
+export interface SortableTableRowProps {
+  /** Table cells */
+  cells: SortableTableCellProps[],
+
+  /** Props, if provided to be spread on <Table.Row> element */
+  [x: keyof any]: unknown
+}
```
mordent-goonstation, line 72: Broadly, prefer putting props types with their respective component. In the case of there being some kind of shared props type that is used by multiple components, they can reasonably extend a shared one.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+      setCallback({ dir: SortDirection.Asc, index: index });
+    }
+  } else {
+    setCallback({ dir: SortDirection.Asc, index: index });
+  }
+};
+
+export const sortAndFilterRows = (
+  headerConfig: SortableTableHeaderConfig[],
+  rowConfig: SortableTableRowProps[],
```
mordent-goonstation, line 28: Not sure this is config at this point. Isn't this just the row data?

tgui/packages/tgui/interfaces/common/sorting/SortableTable/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  let header: SortableTableHeaderConfig;
+
+  if (sortState) {
+    header = headerConfig[sortState.index];
+  } else if (searchState) {
+    header = headerConfig[searchState.index];
+  } else {
+    return rowConfig;
+  }
```
mordent-goonstation, line 46: Probably easiest just to have an early return, then handle the sort and search as you are below? (i.e. one after the other?) Pre-loading the `header` here doesn't seem especially useful.

Something like:

```
if (!sortState && !searchState) {
  return;
}
```

tgui/packages/tgui/interfaces/common/sorting/SortableTable/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  headerConfig: SortableTableHeaderConfig[],
+  rowConfig: SortableTableRowProps[],
+  sortState: SortState,
+  searchState: SearchState): SortableTableRowProps[] => {
+
+  const first = rowConfig[0];
+  let sorted = rowConfig;
+  if (headerConfig.length !== first.cells.length) {
+    throw new Error(`Inconsistent number of fields in table: ${headerConfig.length}x${first.cells.length}`);
+  }
```
mordent-goonstation, line 36:
```suggestion
  const first = rowConfig[0];
  if (headerConfig.length !== first.cells.length) {
    throw new Error(`Inconsistent number of fields in table: ${headerConfig.length}x${first.cells.length}`);
  }
  let sorted = rowConfig;
```

mordent-goonstation: Also, make sure to handle the case where you have no contents in `rowConfig` (i.e. empty array). Having no data to sort/search is a valid state, and should not cause an error.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    sorted = [...rowConfig].sort((rowA, rowB) => {
+
+      const dataA = rowA.cells[sortState.index].data;
+      const dataB = rowB.cells[sortState.index].data;
+
+      return sortFunc(dataA, dataB);
+    });
+
+    if (sortState.dir === SortDirection.Asc) {
+      sorted.reverse();
```
mordent-goonstation, line 59: If making this a generic component, performance stuff starts to matter as you're losing an idea of how much data you're operating on.

I'd be interested in running some metrics with reversing the list here vs. including and factoring in sort direction in the comparator. I *suspect* for the use cases we're working with the reverse is actually more performant.

Fine if you don't want to go through the effort of doing this, possible place for some future work.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+
+  if (searchState && header.searchable) {
+    sorted = sorted.filter(val => {
+      const data = val.cells[searchState.index].data;
+      return header.toString(data).toLocaleLowerCase().includes(searchState.text.toLocaleLowerCase());
```
mordent-goonstation, line 66: Minor, but you can generate `searchState.text.toLocaleLowerCase()` outside of the `.filter` loop to avoid having to generate it each time.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/type.ts
```diff
@@ -0,0 +1,72 @@
...
+  searchable: BooleanLike,
+
+  /** Optional tooltip string. */
+  toolTipContent?: string,
+
+  /** Callback function to sort on type specific to data given */
+  compareFunc?: (a: unknown, b: unknown) => number,
+
+  /** Conversion from your data to string to allow searching */
+  toString?: (data: unknown) => string,
```
mordent-goonstation, line 31: Could make use of generics here.
```suggestion
  /** Callback function to sort on type specific to data given */
  compareFunc?: (a: TValue, b: TValue) => number,

  /** Conversion from your data to string to allow searching */
  toString?: (data: TValue) => string,
```
Then at the top of this thing:
```
export interface SortableTableHeaderConfig<TValue>
```
You might hit some interesting Typescript errors when doing this (either here or wherever this is consumed). Happy to help you work through them if needed.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+      setCallback({
+        dir: (current.dir === SortDirection.Asc ? SortDirection.Desc : SortDirection.Asc),
+        index: index,
+      });
+    } else {
+      setCallback({ dir: SortDirection.Asc, index: index });
+    }
+  } else {
+    setCallback({ dir: SortDirection.Asc, index: index });
+  }
```
mordent-goonstation, line 23:
```suggestion
  if (current && current.index === index ) {
    setCallback({
      dir: (current.dir === SortDirection.Asc ? SortDirection.Desc : SortDirection.Asc),
      index,
    });
  } else {
    setCallback({ dir: SortDirection.Asc, index });
  }
```
N.B. `index: index` -> `index` shorthand.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export const sortAndFilterRows = (
```
mordent-goonstation, line 26: I would strongly consider separating out "sort" and "filter" logic. I don't mind there being this wrapping function that does one after the other for the sake of convenience, but the sort and filter can have their own functions (even if only used here). Makes operating on this easier.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+          }
+        }
+      })}
+    </Table.Row>
+  );
+};
+
+const SortableTableRow = (props: SortableTableRowProps) => {
+  const { cells, ...rest } = props;
```
mordent-goonstation, line 79: I had a bit of fun trying to follow `cells` around, would `columnConfig` be a more accurate name? (This is an actual question, not me being passive aggressive). If it seems like it would fit, consider using it. If not, could you explain what `cells` is here?

tgui/packages/tgui/interfaces/common/sorting/SortableTable/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+                <Header sortDirection={sortDirection} onSortClick={sortByIndex}>{children}</Header>
+              </Table.Cell>);
+          } else {
+            return (
+              <Table.Cell header key={index}>
+                {children}
+              </Table.Cell>);
+
+          }
+        }
```
mordent-goonstation, line 72: This chunk here is pretty verbose and contains repeated parts. Consider separating out into wrapper and contents, with each being conditionally set to one of the two things, then combining together at the end.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+      })}
+    </Table.Row>
+  );
+};
+
+const SortableTableRow = (props: SortableTableRowProps) => {
```
mordent-goonstation, line 78: Separate this out into its own file (`SortableTableRow.tsx`), as it has enough complexity in it.

tgui/packages/tgui/interfaces/common/sorting/SortableTable/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { onSortClick, sortAndFilterRows } from "./utils";
+
+
+interface SortableTableHeaderRowProps {
+  config: SortableTableHeaderConfig[],
+  sortState: SortState,
+  setSortBy: (index: number) => void
+}
+
+const SortableTableHeaderRow = (props: SortableTableHeaderRowProps) => {
```
mordent-goonstation, line 23: Separate this out into its own file (`SortableTableHeaderRow.tsx`) as it has enough complexity in it.

mordent-goonstation (changes requested):
Reviewed the `common/sorting` directory's changes (focusing on the SortableTable). Will want to look at those changes to make sure the result is something I'm cool with. I have not reviewed the usages of `SortableTable` (didn't seem much point given some of this is in flight still).

Definitely along the right track, though!

I have some thoughts about how best to take a second pass at this to split out some of features into their own things that are then composed in, but I can take on those at a later time (or not at all, as they're really just a case of "avoid doing stuff if we don't need to" and it's not really bloated at the moment).

ZeWaka:
Closed due to the TGUI feature freeze, https://github.com/goonstation/goonstation/discussions/19808

PR #18310, opened 2024-03-25
Gang gun adjustments

TDHooligan:
## About the PR 
Fixes quirks with the lo-point. Brings up the other pistols a bit to be somewhat consistent in their damage potential.
Striker/revolver do stamina damage, as they're lacking real *oomph* despite their high damage
Revolver gets a slightly weaker AP effect, as it's currently a little wimpy vs a fully geared gang member.

Birdshot holds its' damage for longer - spread causes missed shots anyway.

## Why's this needed? 
Bring the pistols more in line with each other. Make the striker birdshot a bit more reliable.

## Changelog 

```changelog
(u)TDHooligan
(*)Gang firearms have been tweaked. Please feedback in the discord or forum.
(+)Lo-Point now has a sensible firerate & has 2 fewer rounds. 
(+)Burst laser does 3 more damage/laser.
(+)Webley Revolver now does stamina damage & has a weaker AP effect.
(+)Striker-7 does stamina damage, damage now falls off later, but a lot faster.
(+)The SMGs remain unchanged (They already have a lot of raw damage)
```

### DIFF
--- a/code/modules/projectiles/bullet.dm
+++ b/code/modules/projectiles/bullet.dm
@@ -1963,8 +1963,9 @@ datum/projectile/bullet/autocannon
 /datum/projectile/bullet/webley
 	name = "bullet"
 	damage = 45
-	damage_type = D_KINETIC
-	hit_type = DAMAGE_CUT
+	stun = 15
+	damage_type = D_PIERCING
+	armor_ignored = 0.5 //just enough to get past gang vests in 3 shots
 	implanted = /obj/item/implant/projectile/bullet_455
 	impact_image_state = "bullethole-small"
 	casing = /obj/item/casing/medium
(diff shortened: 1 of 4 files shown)


Tarmunora:
eh, not sure I like giving stun to the pistols like that, especially to that degree - note that a basic taser is 20 stun, and you're putting an AP revolver bullet at 15

TDHooligan:
oh, i blasted a guy and it did like 60 STA. which is 2/3rds of a revolver magazine to down. i figured it wasn't criminal, as:
- it has no disorient/stagger 
- ammunition is rare (a duffel bag drop will net you 6 + 6 spare bullets)
- unintended, possibly helpful effect of making you fall over halfway through a fan-the-hammer attack.
- lowest net damage of all the handguns
 could probably roll it back some, though.

Tarmunora (approved):

PR #18323, opened 2024-03-26
Adding action bar to signing a zoldorf burrito

mloccy:
[FEATURE][ACTIONS]

## About the PR 
Adds an action bar to signing a weird burrito purchased from zoldorf.

## Why's this needed? 
Action bars are a great feedback tool for players and should be used anywhere where a player has to wait for their action to finish.

### DIFF
--- a/code/obj/item/zoldorfmisc.dm
+++ b/code/obj/item/zoldorfmisc.dm
@@ -3,16 +3,72 @@
 #define SIGNING 2
 #define SIGNED 3
 
-/datum/action/bar/icon/callback/signingBar
+/datum/action/bar/icon/signing_burrito
+
+	var/obj/item/zolscroll/burrito = null
+	var/mob/signer = null
+	var/obj/item/pen/pen
+
+	New(obj/item/zolscroll/B, mob/M, var/dur, var/obj/item/pen/P)
+		..()
+		if (istype(B) && istype(M) && istype(P))
+			src.burrito = B
+			src.signer = M
+			src.duration = dur
+			src.icon = src.burrito.icon
+			src.icon_state = src.burrito.icon_state
+			src.pen = P
 
 	onInterrupt(flag)
 		. = ..(flag)
-		var/obj/item/zolscroll/scroll = src.target
 
-		if (istype(scroll))
-			scroll.icon_state = "scrollopen"
-			scroll.signing_state = UNSIGNED
-			scroll.UpdateIcon()
+		if (istype(src.burrito))
+			src.burrito.icon_state = "scrollopen"
+			src.burrito.signing_state = UNSIGNED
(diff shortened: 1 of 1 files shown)


code/obj/item/zoldorfmisc.dm
```diff
@@ -1,11 +1,32 @@
...
-/obj/item/zolscroll //the contract people sign to become zoldorf. stores the name of the signer so players cant sell someone elses soul in a forced zoldorfing
+#define SIGNING_DURATION 4.6 SECONDS
+#define UNSIGNED 1
+#define SIGNING 2
+#define SIGNED 3
+
+/datum/action/bar/icon/callback/signingBar
+
+	New(owner, target, duration, proc_path, proc_args, icon, icon_state, end_message, interrupt_flags, call_proc_on)
+		. = ..(owner, target, duration, proc_path, proc_args, icon, icon_state, end_message, interrupt_flags, call_proc_on)
```
pgmzeta, line 9: unneeded New

```suggestion
```

pgmzeta (changes requested):
tested a few different ways, looks solid. one nit

code/obj/item/zoldorfmisc.dm
mloccy: Done

pgmzeta (approved):

code/obj/item/zoldorfmisc.dm
```diff
@@ -15,18 +34,34 @@
...
+		var/actionBar = new /datum/action/bar/icon/callback/signingBar(
+			user,
+			src,
+			SIGNING_DURATION,
+			PROC_REF(complete_signing),
+			list(user),
+			src.icon,
+			src.icon_state,
+			"<b>[user.name] finishes signing the contract in blood!</b>",
+			(INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION | INTERRUPT_MOVE))
```
Tarmunora, line 55: any reason not to use the `SETUP_GENERIC_ACTIONBAR` macro?

mloccy: I would be happy to except the old code changed the icon state if the action was interrupted `src.icon_state = "scrollopen"`. Wasn't able to see a way to do that with SETUP_GENERIC_ACTIONBAR.

Tarmunora: mm. In that case, I think it would be better to make the action a direct child of `/datum/action/bar/icon` rather than `...callback`. Mostly a nitpick but it would be a bit cleaner

mloccy: Done.

Tarmunora (approved):

PR #18324, opened 2024-03-26
Removing wage system jobs

mloccy:
[INTERNAL]
## About the PR 
Attempts to remove the jobs list from `/datum/wage_system`

## Why's this needed? 
Remove duplication of purpose between jobs and the wagesystem, jobs datums should be authoritive on what the starting wage is for a given job and everything else should rely on that single point of truth.


Glamurio:
Thank god finally someone did it

Tarmunora (approved):

PR #18326, opened 2024-03-27
Modify the search function in Secmate to allow searching by criminal status.

avanth89:
[Game-Objects] [QoL]

## About the PR 
Changes the way the secmate search function works under the hood to allow it to check security records for criminal statuses that match the searched string.

## Why's this needed? 
Two reasons. First, quality of life in allowing quicker cataloguing of all that are set to a certain status such as arrest, or parole. Secondly, hopefully encouraging further use of secmate record updating so that this may be used as a way to quickly peek through all of a current status to see why they may be set to said status as either a way to jog your own memory during a rather hectic round, or if you happen to be a late comer to one.

## Changelog 

```changelog
(u)Avanth
(+)Modified secmate to allow the search function to search by criminal status in addition to what it searched by previously.
```


code/modules/networks/computer3/sec_rec.dm
```diff
@@ -587,6 +587,9 @@
 				var/list/datum/db_record/results = list()
 				for(var/datum/db_record/R as anything in data_core.general.records)
 					var/haystack = jointext(list(ckey(R["name"]), ckey(R["dna"]), ckey(R["id"]), ckey(R["fingerprint"]), ckey(R["rank"])), " ")
+					var/haystack_secure_addition = data_core.security.find_record("name", R["name"])
```
pgmzeta, line 590: why name and not ID?

avanth89: Not sure! ID probably would've been better! It's more or less my first dive into it at all, so a bit of a learning experience. I'll change it to use ID instead, shortly.

code/modules/networks/computer3/sec_rec.dm
```diff
@@ -587,6 +587,9 @@
 				var/list/datum/db_record/results = list()
 				for(var/datum/db_record/R as anything in data_core.general.records)
 					var/haystack = jointext(list(ckey(R["name"]), ckey(R["dna"]), ckey(R["id"]), ckey(R["fingerprint"]), ckey(R["rank"])), " ")
+					var/haystack_secure_addition = data_core.security.find_record("id", R["id"])
```
TobleroneSwordfish, line 590:
```suggestion
					var/datum/db_record/haystack_secure_addition = data_core.security.find_record("id", R["id"])
```

TobleroneSwordfish (approved):

PR #18327, opened 2024-03-27
Gunbot Fabricator and More Gunbots

Azrun:
## About the PR 
Adds planet prefabs with gunbots and gunbot fabricator.

Adds Gunbot fabricator that will assemble gunbots if there aren't any in the area.

Flamethrower Gunbot!
MRL Gunbot
Flamethrower Gunbot
20mm Cannon Gunbot
Minigun Gunbot
Chainsaw Gunbot?

Expands projectile limbs to allow for alter projectile callbacks.
Expands projectile limbs to include a `fluid` type that launches fluids similar to a flamethrower.
Allow for more use of`attack_self` behavior in limbs.

## Why's this needed? 
Expands breadth of Admin Gunbot options.
Provides for some higher end Gunbots as threats to the crew.
Expands Terrainify and Find Planet Prefabs


cheekybrdy:
from what I understand, these would all be syndicate mechscannable at the base price of the standard one?

Azrun:
> from what I understand, these would all be syndicate mechscannable at the base price of the standard one?

Mmmm, as implemented yes. That should be evaluated further as there is a **large** spike in lethality with some of these.

PR #18329, opened 2024-03-27
Makes security officer a limited high-priority job

TobleroneSwordfish:
[BALANCE]
## About the PR 
Makes security officer "high priority" during initial job selection until there are at least two security officers selected.
This will mean you may roll security officer more often from lower priorities if there is a need for it, but you still won't roll it if you have it set to unwanted.
Also adds an `order_priority` var so that secoffs can be high priority but not prioritize over captains and AIs.

## Why's this needed? 
The game functions a lot better when there is at least *some* security and currently with the way the job code works it's possible for there to be 5 people with security on med priority and still no secoffs because the game decided to make them all engineers. This change will hopefully make it more consistent and reduce the number of zero secoff shifts where no-one wants to late-join as solo-sec.

## Changelog 

```changelog
(u)LeahTheTech
(*)Security Officer is now a limited high priority role, meaning the game will attempt to ensure there are a minimum number of them in each shift. Having the role disabled will still prevent you from rolling it.
```

### DIFF
--- a/_std/compare.dm
+++ b/_std/compare.dm
@@ -52,3 +52,6 @@
 /proc/cmp_chemical_reaction_priotity(datum/chemical_reaction/a, datum/chemical_reaction/b)
 	return a.priority > b.priority
 #endif
+
+/proc/cmp_job_order_priority(datum/job/a, datum/job/b)
+	return cmp_numeric_asc(a.order_priority, b.order_priority)
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -568,6 +568,7 @@ ABSTRACT_TYPE(/datum/job/security)
 		name = "Security Assistant"
 		limit = 3
 		lower_limit = 2
+		high_priority_job = FALSE //nope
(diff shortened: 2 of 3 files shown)


Tarmunora (approved):

Chatauscours:
Good change, but it should probably also account for whether or not an HoS was rolled or not to avoid having too much sec for a lowpop round.

TobleroneSwordfish:
HoSes are rolled after secoffs, but I've made it so a high-priority rolled secoff can be promoted to HoS (if whitelisted of course) which should help a bit.

ZeroProximity:
So i just found this. i thought it was a bug, i had to remove security from my list, it would place me in security officer over other command roles. there was one round on goon 4 where we ended up with HoS 2 security and one assistant for a round with 8-10 people? and the only command staff was the HoS

cheekybrdy:
On single pop rounds, this pop ends up creating 2 officers on a station where they end up being a third of the station. It should probably be only 1 as high priority for ~15pop and maybe only a secass for under 10, if anything.

PR #18330, opened 2024-03-27
Allows 3 Security Officers to revoke Armory access

JORJ949:
( Theres no A tag for Security 👎 and I was unsure which tag was right, thanks 444 )
## About the PR 
Applies Armory authorization rules to revoking authorization by allowing 3 people with security equipment access to unauthorize.

## Why's this needed? 
Security can gain access, but not lock it back up when they are done with it unless the HoS or Captain does it, they should be able to lock it themselves considering they can unlock it

## Changelog 

```changelog
(u)JORJ949
(*) 3 Security Officers can now deauthorize Armory.
```

### DIFF
--- a/code/obj/machinery/computer/security/riot.dm
+++ b/code/obj/machinery/computer/security/riot.dm
@@ -3,7 +3,7 @@
 	icon_state = "drawbr"
 	density = 0
 	glow_in_dark_screen = TRUE
-	var/auth_need = 1
+	var/auth_need = 3
 	var/list/authorized
 	var/list/authorized_registered = null
 	var/net_id = null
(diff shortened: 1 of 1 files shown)


pgmzeta:
I think you could reduce code duplication here with procs called from `attackby` and `print_auth_needed`, as a lot of the logic is shared between the authorized and unauthorized code paths.

JORJ949:
> I think you could reduce code duplication here with procs called from `attackby` and `print_auth_needed`, as a lot of the logic is shared between the authorized and unauthorized code paths.

I considered that but couldn't come up with a clean way to also include the different text for if its authed or not

JORJ949:
I've got a plan to reduce duplicate code, should be coming on monday

Tarmunora:
Nah

PR #18335, opened 2024-03-28
fix gang vests being foldable into invisibleness

TDHooligan:
[BUG][CLOTHING]
## About the PR 
gang vests can be folded because they're /vest. stop it. 
also changes description to match their new state (they dont disappear under jumpsuits)

### DIFF
--- a/code/obj/item/clothing/armor.dm
+++ b/code/obj/item/clothing/armor.dm
@@ -499,7 +499,7 @@ TYPEINFO(/obj/item/clothing/suit/armor/NT_alt)
 
 /obj/item/clothing/suit/armor/vest/gang
 	name = "Light Armor Vest"
-	desc = "A stab & bullet-resistant vest, light enough to slip under a jumpsuit."
+	desc = "A minimalist plate carrier strapped to your torso. Provides as much protection as you can get without cramping your style."
 	icon = 'icons/obj/items/gang.dmi'
 	icon_state = "lightvest"
 	item_state = "lightvest"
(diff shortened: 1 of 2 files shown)


code/obj/item/clothing/armor.dm
```diff
@@ -499,7 +499,7 @@ TYPEINFO(/obj/item/clothing/suit/armor/NT_alt)
 
 /obj/item/clothing/suit/armor/vest/gang
 	name = "Light Armor Vest"
```
pgmzeta, line 501: nit: should be lowercase name?

code/obj/item/clothing/armor.dm
```diff
@@ -510,4 +510,8 @@ TYPEINFO(/obj/item/clothing/suit/armor/NT_alt)
 		..()
 		setProperty("meleeprot", 5)
 		setProperty("rangedprot", 0.5)
+	attackby(obj/item/W, mob/user)
```
pgmzeta, line 513: Any reason for this not to be a direct descendant of `/obj/item/clothing/suit/armor` instead? This still has the property of being usable for hazard suit armoring, but it's unclear if that is intentional (and if so, it appears bugged from quick testing).

TDHooligan: yea fair. i hotfixed it super quick but this is definitely smarter

pgmzeta (approved):

ZeWaka (approved):

PR #18345, opened 2024-03-29
Gangs: Duffle bag changes

TDHooligan:
[GAMEMODES][FEATURE]
## About the PR 
Adds:
**Duffle bag traps:**
If picked up by a civilian before any gang member, barbs inside the bag's handle will snap out and bury in their hand, becoming undroppable, and causing just enough bleed to leave a teeny blood trail. This can be fixed with a 20 second action bar to unhook the barbs from your hand..

https://github.com/goonstation/goonstation/assets/6061314/6632ef61-d516-455c-b819-39bc74abb584

While unremovable by anyone but the civilian with it in their hand, gang members may also mercilessly rip the bag from a belligerent staff assistant's hand.

**Duffle bag tracking:**
When picked up by a civilian, or dragged outside of their room they spawned in, Duffle bags will send a radio message to the gang, informing them which room the bag is in & who knew about it (so that they can stop interrogating a player if their intel is bad). 

**Other nerd-hating behavior**
Nerds will try to drag duffle bags to get around this, so:
- Using a new obj flag, duffle bags will notify their gang if dragged outside of the area they're spawned in.
- Using a new /atom flag, duffle bags are slow to drag. 
- Lockers with a bag inside will act as if they have a mob inside and also slow down

**Duffle bag QoL**
Duffle bags now also use vis_contents, like gang crates. Anything taken out can't be put back in, but picking up a duffel bag will immediately re-seal it. 

https://github.com/goonstation/goonstation/assets/6061314/a241f700-dc25-4df7-9560-30f3e9e6637a

## Why's this needed? 
Goody two-shoes keep stealing all of the gang loot and delivering it to security. Gangs are powerless to stop it, or even know it's happening. Now they have a chance to intervene.

## Changelog 

```changelog
(u)TDHooligan
(*)Gangs are fed up of stolen drops. They've started booby trapping their bags and cut back on weight savings.
```


Cherman0:
While preventing people from stealing these things is good, picking up some random thing and getting crippled and subsequently murdered seems kinda bad especially from a new player perspective who might not know how it works. Maybe non-gangs could have a long actionbar to pick it up and significantly slower movespeed while held. You can probably copy whatever the nukie knife uses to speed you up to implement the movespeed debuff.

flappybatpal:
> While preventing people from stealing these things is good, picking up some random thing and getting crippled and subsequently murdered seems kinda bad especially from a new player perspective who might not know how it works. Maybe non-gangs could have a long actionbar to pick it up and significantly slower movespeed while held. You can probably copy whatever the nukie knife uses to speed you up to implement the movespeed debuff.

Yeah I agree. This feels like quite a complicated solution. If anything I'd question if civs should be able to interact with the bags at all considering the issues that arise from it.

Cherman0:
Its complicated in that its not at all what you'd expect from looking at it or understanding what it is. It doesnt look like something that would embed itself into your hand and summon a bunch of angry dudes to beat you up for lunch money. Whereas "heavy and cumbersome" is something a duffle bag could plausibly be. 

I don't have any issue with the alarm, the hand embedding thing is weird and not useful. It doesnt actually stop you from just running off with it and in fact encourages you to get somewhere where the gangs cant catch you with it, like security.

TDHooligan:
>the hand embedding thing is weird and not useful

I think just saying 'not useful' is disingenuous. Just because you haven't thought of reasons for it to exist doesn't mean I just pulled it out of nowhere.

The embedding:
 - stops you throwing it forwards to remove the slowdown it gives you
 - stops you spacing/hiding the bag instantly (which would make the alarm useless, as the bag could be anywhere very quickly)
 - is a sensible rationale to make you bleed, marking you as the thief & giving you a trail of blood to follow (which other contributors/devs have said is a fun idea)
 - **is not a death sentence**
 
 It's barbs inside a nylon handle attached to a heavy bag, I think this is thematic, not weird. it's not some space-age meta magic like 'glue that gives civilians an action timer'. 
  
You're still also using this reductionist take of 'you will be beaten up immediately for having the bag' - RP is much more broad than this, even on classic most gangs (minus a couple notorious rampagers) are willing to talk to crew.

Chatauscours:
I feel like it should give civilians that try to pick it up a big bright warning like "you feel like if you pick this up you'll get into big trouble with people that will kill you over this bag! Are you REALLY sure you want to pick this up?" or something of the sort, to give new players / players who have no idea how the bag works some sort of warning as to the fact that they'll get in trouble with the gangs over this.

TDHooligan:
while i think the PDA messages serve this purpose (perhaps not reliably enough atm. i'll change up the messages in a sec), 

i'm not entirely against the idea of adding some sort of warning, but that's also similarly meta as it'd involve overriding pickup behaviour.

Cherman0:
I dont think an invisible hook that makes it slow to put down is any more "space-age meta magic" than an invisible weight that makes it slow to pick up or move with. What I'm mainly concerned about is this being a noob trap where experienced players know that if they touch it it will eat up their hand slot for a while and tell a bunch of antags where they are. The gamemode does have to work on classic too, and even on RP some random stealing their shit is plenty of justification to get murdered. To your point about the crew feeling "sidelined" during gangs, I dont want the "solution" to this problem being the expectation that the antags will not actually fight people.

If you had the gang location alert go off at the beginning of a 15 second pickup actionbar, reduce your movespeed while holding it and reduce its throw range like TTVs already do you'd have a workable solution that doesnt punish curiosity / unfamiliarity with the mechanics. If youre really concerned about people trying to make them entirely unfindable you could have a gang pinpointer that picks up its location after a long enough amount of time(also useful for if the selected person dies or is afk or something).

_std/defines/obj.dm
```diff
@@ -11,6 +11,8 @@
 #define HAS_DIRECTIONAL_BLOCKING (1<<3)
 /// prevents ghost critter interaction. On obj so it can cover machinery, items etc...
 #define NO_GHOSTCRITTER (1<<4)
+/// Item treats locker movement as being moved itself (calls OnMove())
+#define TRACK_STORAGE_MOVEMENT		(1<<5)
```
ZeWaka, line 15: Why not just track area exited or something, signal

TDHooligan: i wasn't aware that method existed, ill swap to that shortly

code/mob/living.dm
```diff
@@ -1789,6 +1789,11 @@
 							// if the storage object contains mobs, use its p_class (updated within storage to reflect containing mobs or not)
 							if (locate(/mob) in A.contents)
 								. *= lerp(1, max(A.p_class, 1), mob_pull_multiplier)
+							else if (locate(/obj/item/gang_loot) in A.contents)
+								. *= lerp(1, max(A.p_class, 1), mob_pull_multiplier)
+						else if (A.always_slow_pull)
+							. *= lerp(1, max(A.p_class, 1), mob_pull_multiplier)
```
Mister-Moriarty, line 1795: Personally I fervently dislike this, but refactoring slowing the pulling of objects to use a more sane system than a tree of `if else`s and `locate() in`s is outside of the scope of this PR.

code/modules/status_system/statusEffects.dm
```diff
@@ -1382,6 +1382,17 @@
 		getTooltip()
 			. = "Your vitals have dropped from the shame you feel hiding your true colors inside enemy territory."
 
+	gangtrapped
+		id = "gang_trap"
+		name = "Punctured"
+		desc = "You've grabbed something that wasn't yours, and it's lodged in your hand! Use it in hand to start plucking it free."
+		icon_state = "gangtrapped"
+		unique = 1
```
Mister-Moriarty, line 1390:
```suggestion
		unique = TRUE
```

code/obj/decorations.dm
```diff
@@ -311,6 +311,12 @@
 				visible_message(SPAN_ALERT("<b>[user] violently shakes [src] around! \An [thing] falls out!</b>"))
 				last_use = world.time
 				max_uses--
+			else
+				if (istype(something, /obj))
```
Mister-Moriarty, line 315: Collapse this `else if` into a single line and use the `isobj()` define.

```
else if (isobj(something))
    ...
```

code/obj/storage/gang_crate.dm
```diff
@@ -167,67 +172,193 @@
...
+	var/datum/vis_storage_controller/vis_controller = new()
+	var/datum/loot_generator/lootMaster
 	level = UNDERFLOOR
 
+
+	proc/initialize_loot_master(x,y)
+		src.vis_controller = new()
+		vis_controller.initialize(src)
+		lootMaster =  new /datum/loot_generator(x,y)
+		RegisterSignal(src, XSIG_MOVABLE_AREA_CHANGED, PROC_REF(alert_gang))
```
Mister-Moriarty, line 209: `toggle_tracking()` should handle all signals related to tracking; see my first review comment for initialising tracking.

code/obj/storage/gang_crate.dm
```diff
@@ -167,67 +172,193 @@
...
+		..()
+		if (src.layer == UNDERFLOOR)
+			src.layer = OVERFLOOR
+		var/datum/gang = user.get_gang()
+		if (gang && (trapped || tracking))
+			trapped = FALSE //if one gang member gets their mitts on it, this has done its job
+			toggle_tracking(FALSE)
+			boutput(user, SPAN_ALERT("You disarm the trap in the [src]'s handle. It's now safe to carry."))
+		if (!gang && !open && trapped && owning_gang)
+			if (isliving(user))
```
Mister-Moriarty, line 248: Collapse these two `if` statements into one line.

```
if (!gang && !open && trapped && owning_gang && isliving(user))
    ...
```

code/obj/storage/gang_crate.dm
```diff
@@ -167,67 +172,193 @@
...
+				playsound(src.loc, 'sound/impact_sounds/Generic_Snap_1.ogg', 50, 1)
+				boutput(user, SPAN_ALERT("As you pick up \the [src.name], a series of barbs emerge from the handle, lodging in your hand!"))
+				src.owning_gang.broadcast_to_gang("The bag [src.informant] knew about has just been stolen! Looks like it was in \the [area.name]")
+				ON_COOLDOWN(src,"bleed_msg", 30 SECONDS) //set a 30 second timer to remind players to remove this
+				idiot.setStatus("gang_trap", duration = INFINITE_STATUS)
+				H.emote("scream")
+				H.bleeding = max(1,H.bleeding)
+				SPAWN(1 SECOND)
+					bleed_process()
+	handle_other_remove(var/mob/source, var/mob/living/carbon/human/target){
```
Mister-Moriarty, line 266:
```suggestion
	handle_other_remove(mob/source, mob/living/carbon/human/target)
```

Declaring proc arguments with `var/` is superfluous.
Also, no need to enclose procs within curly brackets.

code/obj/storage/gang_crate.dm
```diff
@@ -167,67 +172,193 @@
...
+		var/datum/gang/userGang = source.get_gang()
+		if (cant_drop && idiot && userGang )
+			source.visible_message("[source] rips the [src] right off [target]! Ouch!","You rip the duffle bag from [target]'s hand.")
+			idiot.emote("scream")
+			playsound(idiot.loc, 'sound/impact_sounds/Generic_Snap_1.ogg', 50, 1)
+			blood_slash(idiot, 2)
+			unhook()
+			return TRUE
+		..()
+	}
```
Mister-Moriarty, line 276:
```suggestion

```

code/obj/storage/gang_crate.dm
```diff
@@ -167,67 +172,193 @@
...
+	handle_other_remove(var/mob/source, var/mob/living/carbon/human/target){
+		var/datum/gang/userGang = source.get_gang()
+		if (cant_drop && idiot && userGang )
+			source.visible_message("[source] rips the [src] right off [target]! Ouch!","You rip the duffle bag from [target]'s hand.")
+			idiot.emote("scream")
+			playsound(idiot.loc, 'sound/impact_sounds/Generic_Snap_1.ogg', 50, 1)
+			blood_slash(idiot, 2)
+			unhook()
+			return TRUE
+		..()
```
Mister-Moriarty, line 275:
```suggestion
		if (!cant_drop || !idiot || !source.get_gang())
			return ..()

		source.visible_message("[source] rips the [src] right off [target]! Ouch!","You rip the duffle bag from [target]'s hand.")
		idiot.emote("scream")
		playsound(idiot.loc, 'sound/impact_sounds/Generic_Snap_1.ogg', 50, 1)
		blood_slash(idiot, 2)
		unhook()
		return TRUE
```

Use an early return for this. There is also no need to declare a variable for `source.get_gang()` if it is only being used once, and within an `if` statement in which it may not always be used.

code/obj/storage/gang_crate.dm
```diff
@@ -167,67 +172,193 @@
...
+			RegisterSignal(src, XSIG_MOVABLE_AREA_CHANGED, PROC_REF(alert_gang))
+		else
+			UnregisterSignal(src, XSIG_MOVABLE_AREA_CHANGED)
+
+	proc/alert_gang(datum/component/component, area/old_area, area/new_area)
+		src.owning_gang.broadcast_to_gang("The bag [src.informant] knew about is being moved! Looks like it's been moved to \the [new_area.name]")
+		toggle_tracking(FALSE)
+
+	proc/unhook()
+		if (idiot)
```
Mister-Moriarty, line 293: Use an early return for this.

code/obj/storage/vis_storage_controller.dm
```diff
@@ -0,0 +1,22 @@
+/datum/vis_storage_controller
+	var/obj/parent = null
+	///The items this storage controller contains
+	var/list/obj/vis_items = list()
+
+	proc/initialize(new_parent)
+		parent = new_parent
+	proc/add_item(var/obj/O)
```
Mister-Moriarty, line 8:
```suggestion
	proc/add_item(obj/O)
```

Declaring proc arguments with `var/` is superfluous.

code/obj/storage/vis_storage_controller.dm
```diff
@@ -0,0 +1,22 @@
...
+
+	proc/hide()
+		for (var/obj/O in vis_items)
+			parent.vis_contents -= O
+
+	proc/show()
+		for (var/obj/O in vis_items)
+			parent.vis_contents += O
+
+	proc/remove(var/obj/O)
```
Mister-Moriarty, line 20:
```suggestion
	proc/remove(obj/O)
```

Declaring proc arguments with `var/` is superfluous.

code/obj/storage/vis_storage_controller.dm
```diff
@@ -0,0 +1,22 @@
...
+		vis_items += O
+		O.AddComponent(/datum/component/storage_viscontents, container = src)
+
+	proc/hide()
+		for (var/obj/O in vis_items)
+			parent.vis_contents -= O
+
+	proc/show()
+		for (var/obj/O in vis_items)
+			parent.vis_contents += O
```
Mister-Moriarty, line 18: If `vis_items` is always only going to contain objects, then `as anything` should be used within the `for` loops. This skips an implicit typecheck performed by BYOND, saving processing time.

```
for (var/obj/O as anything in vis_items)
```

Mister-Moriarty (changes requested):
My review is for the most part concentrated on styling and syntax; I have yet to test the code or review functionality thoroughly.

code/obj/storage/vis_storage_controller.dm
```diff
@@ -0,0 +1,22 @@
+/datum/vis_storage_controller
+	var/obj/parent = null
+	///The items this storage controller contains
+	var/list/obj/vis_items = list()
+
+	proc/initialize(new_parent)
+		parent = new_parent
```
Mister-Moriarty, line 7: This could be achieved by passing the parent atom to `new()` instead of implementing a separate proc for it.

```
/datum/vis_storage_controller/New(obj/parent)
    . = ..()
    src.parent = parent
```

code/obj/storage/vis_storage_controller.dm
```diff
@@ -0,0 +1,22 @@
+/datum/vis_storage_controller
+	var/obj/parent = null
+	///The items this storage controller contains
+	var/list/obj/vis_items = list()
```
Mister-Moriarty, line 4: When declaring a variable that should initially be an empty list, it is better practice to assign `src.var_name = list()` in `New()`. This is because if a list is created in the variable's definition, BYOND will call a hidden `init()` proc to initialise the list. Creating the list in `New()` avoids the overhead of the `init()` call.

code/obj/storage/gang_crate.dm
```diff
@@ -167,67 +172,196 @@
...
+		else
+			..()
+
+	proc/bleed_process()
+		while(cant_drop)
+			if (!ON_COOLDOWN(src,"bleed_msg", 30 SECONDS))
+				boutput(idiot, SPAN_ALERT("The hooks in the bag are digging into your hands! You should pluck it out..."))
+			bleed(idiot, 1, 1)//technically doubling bleed. but it looks nice as the loops dont sync perfectly.
+			idiot.bleeding = max(1,idiot.bleeding)
+			sleep(2 SECONDS)
```
Mister-Moriarty, line 328: This should be achieved through the use of the process controller.
See `proc/process` and `processing_items`.

code/obj/storage/gang_crate.dm
```diff
@@ -167,67 +172,196 @@
...
+			UnregisterSignal(src, XSIG_MOVABLE_AREA_CHANGED)
+
+	proc/alert_gang(datum/component/component, area/old_area, area/new_area)
+		src.owning_gang.broadcast_to_gang("The bag [src.informant] knew about is being moved! Looks like it's been moved to \the [new_area.name]")
+		toggle_tracking(FALSE)
+
+	proc/unhook()
+		if (!idiot)
+			return
+		if (istype(idiot.l_hand, /obj/item/gang_loot) && istype(idiot.r_hand, /obj/item/gang_loot)) //this is REALLY stretching it, bub.
```
Mister-Moriarty, line 296:
```suggestion
		if (idiot.is_in_hands(src))
```

Mister-Moriarty (changes requested):

code/obj/storage/gang_crate.dm
TDHooligan: does not work. this is explicitly checking for two loots.

TobleroneSwordfish (approved):
I've kind of come around to this idea after seeing the amount of duffle bags being immediately handed into security. I do still worry that this is only going to affect newer players and experienced players are just going to know to put it in a locker instead, buuut at least they're now very slowly dragging around a suspect locker.

TobleroneSwordfish (changes requested):

code/obj/storage/gang_crate.dm
```diff
@@ -167,67 +171,195 @@
...
+	var/sealed = TRUE
+	///The gang who should own this duffel bag
+	var/datum/gang/owning_gang
+	///The name of the informant who knows about this bag
+	var/informant
+	///The civilian who has this in their hands, if the trap is active.
+	var/mob/living/idiot = null
+	///The area this bag spawned in.
+	var/area/start_area
+	var/processing_tier = PROCESSING_EIGHTH
```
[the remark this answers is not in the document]
TDHooligan: ah. well in that case this works fine as is. removin it

PR #18349, opened 2024-03-29
Reworks the captain's antique laser gun entirely.

Cherman0:
## About the PR 

Entirely recodes and redesigns the captain's antique laser gun in order to be more intuitive and hopefully less powergameable. 

- The lens part controls the amount of damage it does and is based on its alpha (transparency) value. 
-- Moderately transparent materials like glass and molitz make a weak 30 damage laser
-- Highly transparent materials like telecrystal and ectofiber make a normal 45 damage laser
-- Extremely transparent materials like starstone make a strong 60 damage laser

- The coil controls the power usage per shot. It is based on its electrical conductivity, with a bonus +2 for if the material is an energy source
-- Moderately powerful materials like as copper and pharosium will make an inefficient 41.66u power laser
-- Powerful materials like electrum, cerenkite, and claretine will make a normal efficiency 31.25u laser
-- Highly powerful materials (only custom alloys, figure it out!), will make a high-efficiency 25u laser

If you use a material that isn't at all suited for its role such as an opaque lens it won't even shoot.

The process of installing the required parts and the underlying code is greatly simplified and debugged. The components used are exposed to the heat of the laser, so no making a lens out of ice or a coil out of erebite or something like that. I mean, you could, but you shouldn't. Adding this feature required two changes external to the immediate scope:

- Ice now has a chance to melt when exposed to heat sources above 0 Celsius. This mostly means FEA_fire hotspots, but firing the antique also exposes its components to the same temperature as the laser would create hotspots of in flight. This means that you cant make a lens out of ice for a 45 damage laser (well, you can but it will only work once).
- Hotspots created by flying lasers now use Kelvin instead of Celsius. This fixes the fact that weak (ie less than 15 damage) lasers would actually be colder than room temperature. More importantly, it makes standard 45 damage lasers just hot enough to have a chance to explode erebite and the powerful 60 damage lasers guaranteed to explode erebite.

Numbers subject to change from balance feedback. Getting the tier 3 levels for damage and cost is actually rather difficult, requiring very good mining luck and generous miners for damage and an extremely niche specific alloy (which itself is difficult to get the parts for) for cost.

## Why's this needed? 

The captain's antique laser gun is the last thing that actually relies on matsci material quality, which could generously be described as "half-baked". Refactoring this to use the actual material statistics of the parts you installed makes more intuitive sense (no more completely opaque lenses and non-conductive coils) and allows for making the highest tier of gun actually difficult to get.

## Changelog 

```changelog
(u)Cherman0
(+)The captain's antique laser gun has been reworked to use material statistics instead of quality points.
(+)Ice now melts when exposed to temperatures. Higher temperatures melt it more reliably.
```


Laboredih123:
why the desc change?

Cherman0:
Because I've added various context-dependent diagnostic info in examine() and I didn't want the examine text to be half a screen tall. The old description also was somewhat conflicting between whether it was or was not a real gun and whether it was Super! Protector Friend or Hafgan. I went with Hafgan and that its a model replica since they're the established maker of the actual laser rifle that its default stats are based on.

Cherman0:
Random rooms runtime checker my behated
What do you mean "[X] Set up job"

Cherman0:
Update: it now burns you if you shoot it without closing the maintenance panel back up first. The "model power cell" that it comes installed with now has a max and current charge of 1 (up from 0) so the gun's low charge light wouldn't blink while its in the display case.

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -559,6 +559,16 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
 				qdel(I)
 		return
 
+/datum/materialProc/ice_temp
+	execute(var/atom/owner, var/temp)
+		if(!prob(temp - T0C)) return //percentage chance to melt for every degree above freezing
+		if(ON_COOLDOWN(owner, "ice_temp", 10 SECONDS))
+			return
+		var/turf/tloc = get_turf(owner)
+		tloc.visible_message(SPAN_ALERT("[owner] melts due to heat!"))
```
TobleroneSwordfish, line 568:
```suggestion
		tloc.visible_message(SPAN_ALERT("[owner] melts away!"))
```

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -559,6 +559,16 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
...
+		qdel(owner)
```
TobleroneSwordfish, line 569: Maybe spawn a small amount of water here? Water is relatively harmless and easy to flood with anyway so it should be fine?

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,185 @@ TYPEINFO(/obj/item/gun/energy)
...
+		if(.)
+			//exactly enough that an erebite component will have a chance to explode from a normal damage laser
+			//and guaranteed to explode from a high damage laser
+			//also melts ice so it isnt viable for as a normal damage lens material because thats too easy
+			src.myLens.material_trigger_on_temp(FROM_CELSIUS(src.current_projectile.power * 20))
+			src.myCoil.material_trigger_on_temp(FROM_CELSIUS(src.current_projectile.power * 20))
+			if(src.panelOpen)
+				//close it dummy
+				user.visible_message(SPAN_ALERT("Residual heat burns you through the open maintenance panel!"))
+				user.TakeDamage("All", 0, 25, 0, DAMAGE_BURN)
```
TobleroneSwordfish, line 345: This is copy-paste, make it a proc

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,185 @@ TYPEINFO(/obj/item/gun/energy)
...
+		if(isscrewingtool(object))
+			user.show_text(SPAN_NOTICE("You [src.panelOpen ? "close" : "open"] the maintenance panel."))
+			playsound(user, 'sound/items/Screwdriver2.ogg', 65, TRUE)
+			src.panelOpen = !src.panelOpen
+			if(!src.panelOpen && src.evaluate_quality())
+				if(src.customProjectileDamage == CAPTAINGUN_T3_DAMAGE && src.customProjectileCost == CAPTAINGUN_T3_COST)
+					user.unlock_medal("Tinkerer", 1)
+		if(istype(object, /obj/item/coil/small))
+			if(src.panelOpen)
+				user.show_text(SPAN_NOTICE("You insert the [object] into the [src]."))
```
TobleroneSwordfish, line 305:
```suggestion
				user.show_text(SPAN_NOTICE("You insert [object] into the [src]."))
```
byond handles this, putting "the" here leads to "the the"

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,185 @@ TYPEINFO(/obj/item/gun/energy)
...
+				playsound(user, 'sound/items/Deconstruct.ogg', 65, TRUE)
+				user.put_in_hand_or_eject(src.myCoil)
+				object.set_loc(src)
+				user.u_equip(object)
+				src.myCoil = object
+			else
+				user.show_text(SPAN_NOTICE("The [src]'s maintenance panel is closed."))
```
TobleroneSwordfish, line 312:
```suggestion
				user.show_text(SPAN_NOTICE("[src]'s maintenance panel is closed."))
```

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,185 @@ TYPEINFO(/obj/item/gun/energy)
...
+		if(istype(object, /obj/item/lens))
+			if(src.panelOpen)
+				user.show_text(SPAN_NOTICE("You insert the [object] into the [src]."))
```
TobleroneSwordfish, line 315:
```suggestion
				user.show_text(SPAN_NOTICE("You insert [object] into [src]."))
```

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,185 @@ TYPEINFO(/obj/item/gun/energy)
...
+				playsound(user, 'sound/items/Deconstruct.ogg', 65, TRUE)
+				user.put_in_hand_or_eject(src.myLens)
+				object.set_loc(src)
+				user.u_equip(object)
+				src.myLens = object
+			else
+				user.show_text(SPAN_NOTICE("The [src]'s maintenance panel is closed."))
```
TobleroneSwordfish, line 322:
```suggestion
				user.show_text(SPAN_NOTICE("[src]'s maintenance panel is closed."))
```

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,185 @@ TYPEINFO(/obj/item/gun/energy)
...
+/// 10. custom alloy (figure it out, nerd)
+#define CAPTAINGUN_T3_COIL_EFFICACY_THRESHOLD 10
+
+//Damage dealt by different tiers of lens used
+/// 30. 2/3x normal laser damage
+#define CAPTAINGUN_T1_DAMAGE 30
+/// 45. normal laser damage
+#define CAPTAINGUN_T2_DAMAGE 45
+/// 60. 4/3x normal laser damage
+#define CAPTAINGUN_T3_DAMAGE 60
```
TobleroneSwordfish, line 244: These should just be different projectile types, with unique sprites as before.

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,185 @@ TYPEINFO(/obj/item/gun/energy)
 			projectiles += current_projectile
 
 
-////////////////////////////////////// Antique laser gun
-// Part of a mini-quest thing (see displaycase.dm). Typically, the gun's properties (cell, projectile)
-// won't be the default ones specified here, it's here to make it admin-spawnable (Convair880).
+//Antique laser gun
```
TobleroneSwordfish, line 219: All these defines should be undef'd at the end of the definition as they are not used elsewhere.

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,185 @@ TYPEINFO(/obj/item/gun/energy)
...
+				src.myLens = object
+			else
+				user.show_text(SPAN_NOTICE("The [src]'s maintenance panel is closed."))
+
+	canshoot(mob/user)
+		if(!src.evaluate_quality())
+			return FALSE
+		//change the projectile here because this is checked by both gun.shoot_point_blank(), gun.shoot(), and gun.suicide()
+		//which duplicate a horrifying amount of code
+		src.set_current_projectile(new /datum/projectile/laser/custom(src.customProjectileDamage, src.customProjectileCost))
```
TobleroneSwordfish, line 329: God no, do not put side effects in `canshoot`. Can't you just set the projectile when the parts are updated?

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,185 @@ TYPEINFO(/obj/item/gun/energy)
...
+	//case-in-point, couldnt avoid it here because I needed it to happen after the shot
+	//and because canshoot() is sometimes checked when we arent actually firing a shot and we dont want to blow up randomly
+	shoot(turf/target, turf/start, mob/user, POX, POY, is_dual_wield, atom/called_target)
+		. = ..()
+		if(.)
+			//exactly enough that an erebite component will have a chance to explode from a normal damage laser
+			//and guaranteed to explode from a high damage laser
+			//also melts ice so it isnt viable for as a normal damage lens material because thats too easy
+			src.myLens.material_trigger_on_temp(FROM_CELSIUS(src.current_projectile.power * 20))
+			src.myCoil.material_trigger_on_temp(FROM_CELSIUS(src.current_projectile.power * 20))
```
TobleroneSwordfish, line 341: These should be using `temperature_expose` instead of directly calling the material triggers.

code/obj/item/gun/energy.dm
TobleroneSwordfish, line 345: Maybe also set them on fire briefly for a little more visual feedback?

TobleroneSwordfish (changes requested):
I really like the idea of this, it's muuch more reasonable than `quality` and requiring rare minerals from mining is excellent.
- Why not just spawn it with no cell and make it only display the blinking low battery display when it actually has a cell inserted?
- The projectiles now all look the same, a good feature of the old antique was being able to tell how powerful it was by seeing the projectile.

Cherman0:
Functionally, the old captaingun tiers worked as follows:
- Tier 1: 25 damage / 20 cost laser (1.25 damage per charge)
- Tier 2: 45 damage / 31.25 cost laser (1.44 damage per charge)
- Tier 3: Does not upgrade the main projectile, instead adds an alternate 3-laser burst mode. 25 x 3 damage / 50 cost (1.5 damage per charge if all shots hit)

The reason I didn't go with an aggregate "ranking" of the efficiency of all parts is that it reduces the number of unique combinations. I didnt want there to be no benefit at all to having the best material for one part and a pretty good material for the other (compared to being pretty good in both categories), but I also didnt want to give the player the best possible gun for not having the best possible combination of parts. The existing captaingun formula has this problem; you typically use 100 quality gemstones for the lens and battery and ignore the coil and wire entirely because they cannot be made of gemstones.

I could look into making the projectiles use different laser graphics based on how much damage they do. And very maybe look into adding another component that can enable a burst fire mode, though I'm leery of how OP it could end up being.

code/obj/item/gun/energy.dm
Cherman0: I could definitely look into having the projectile determine its sprite based on its damage, but having every possible combination of projectile stats be its own statically defined type would be very poor form. The projectiles have the exact same logic and just different variables, there's no reason to manually define each one when a constructor arg will do. 

And it prevents the existing situation where someone down the line takes the convenient route of reusing an existing projectile for their new gun, inadvertently linking the balance of two wholly unrelated things together. For example, the T1 and T3 captainguns currently use the strelka beams, meaning that the series of nerfs to the strelka pod gun has made them quite pathetic.

code/obj/item/gun/energy.dm
Cherman0: `temperature_expose(datum/gas_mixture/air, exposed_temperature, exposed_volume)` is called by airgroups on the atoms they are touching. `material_trigger_on_temp(var/temperature_applied)` is the appropriate proc for some other process exposing a thing to temperature.

Cherman0:
Following feedback, I've done the following:

- Laser now changes sprites based on how powerful it is.
- Shooting with the panel open sets you on fire for 1 second. This doesnt really do any additional damage, but it does make you look like an idiot.
- Moved some copy/pasted code to its own proc.
- Undefined all my macros after they are done being needed.

I did not do:

- Change the gun's roundstart cell or how energy gun lights work. I want the lights to be on when it is in the display case, as the captaingun's sprite looks rather weird without them. Think of the 1u power cell as like, a AA battery as opposed to something that can sustain a high-power laser.
- Make ice spawn a chempool of water when melting. I'm not particularly eager to see how performant spawning a couple hundred chempools is when mining blows up an ice asteroid.

code/modules/materials/Mat_MaterialProcs.dm
Cherman0: Since mining can end up making a lot of ice chunks when mining certain asteroids and since explosions (also commonly found in mining) expose exploded items to heat, this has the potential to spawn quite a lot of water pools. My understanding is that lots of chempools is rather unperformant.

Glamurio:
Minor note but if you change ice to melt at specific temperatures, please make a note of it in the changelogs

Cherman0:
Good point. Updated

code/obj/item/gun/energy.dm
```diff
@@ -214,26 +216,194 @@ TYPEINFO(/obj/item/gun/energy)
...
+#define CAPTAINGUN_T1_COIL_EFFICACY_THRESHOLD 6
+/// 8. claretine, electrum, cerenkite, etc.
+#define CAPTAINGUN_T2_COIL_EFFICACY_THRESHOLD 8
+/// 10. custom alloy (figure it out, nerd)
+#define CAPTAINGUN_T3_COIL_EFFICACY_THRESHOLD 10
+
+//Damage dealt by different tiers of lens used
+/// 30. 2/3x normal laser damage
+#define CAPTAINGUN_T1_DAMAGE 30
+#define CAPTAINGUN_T1_DAMAGE_PROJECTILE_ICON_STATE "lasergat_laser"
```
TobleroneSwordfish, line 241: This is a bad way of doing this, make subtypes instead of just big lists of defines.

TobleroneSwordfish (changes requested):
The different projectile types feel very similar, have the same sound effect etc. Why not just keep the original three modes but use the material stats to pick which one instead of quality? (and modify the cost like you have here)

Cherman0:
> The different projectile types feel very similar, have the same sound effect etc. Why not just keep the original three modes but use the material stats to pick which one instead of quality? (and modify the cost like you have here)

I considered giving the different damage tiers different sounds but I decided against it because I didn't want it to be mistaken for a different weapon. As for why I didn't use the old tiers, I explained that previously: 

> The reason I didn't go with an aggregate "ranking" of the efficiency of all parts is that it reduces the number of unique combinations. I didnt want there to be no benefit at all to having the best material for one part and a pretty good material for the other (compared to being pretty good in both categories), but I also didnt want to give the player the best possible gun for not having the best possible combination of parts. The existing captaingun formula has this problem; you typically use 100 quality gemstones for the lens and battery and ignore the coil and wire entirely because they cannot be made of gemstones.

Basically, the efficacy of each material chosen can be either the minimum, the maximum, or somewhere in between. For the case where one material is of middling quality and the other is either most or least powerful functional choice, having only 3 discrete tiers would fail to handle this "T1.5" or "T2.5" gun. With the formulas I've designed getting a T3 in both categories is *really* unlikely. I don't want there to be zero benefit to pulling out all the stops and getting that, but I also dont want a moderately better than average material set to not be better than just average materials. Thus, each stat needs to be a separately tracked tier.

Which leads into why separately defining every possible combination of stats is a bad idea. At the moment, it'd already be a rather unwieldy 9 subtypes. If someone added another stat for firerate or projectile burst count or something you'd have a ridiculous 27 projectile subtypes for one gun. Statically defining every possible combination of every stat is unmaintainable and pointless when you could use one type and feed it constructor args.

TobleroneSwordfish:
But you're not handling T1.5 here, you're still splitting it into three distinct tiers based on the material. You don't need 27 subtypes, you can just use the three that existed before and modify their cost dynamically like you're already doing here.

Cherman0:
Re-drafting this because I'm not really sure how to incorporate the feedback in a positive way. If someone wants to adopt this then go ahead, otherwise... maybe I'll get some idea at some point. Probably not. The existing projectile tiers are terrible, just please don't use them.

Cherman0:
Actually I just had an idea, hold up this could be cool

Cherman0:
Okay, idea (to implement Soon:tm:) is as follows: New (optional) component, Thermoelectric Recycler Module (TRM). Installing it makes the gun fire a second shot right after the first with reduced damage. The multiplier for the damage of the second shot will go 0.2 -> 0.3 -> 0.4 as you make better tiers of TRM. Its efficacy will be calculated as `electrical - thermal`, with the thresholds to hit being 1 -> 3 -> 5.

Implementing this will require cardinal sins against gun code. A "cool gun" has been asked of me and I shall deliver. If it makes you feel any better the code being contorted into vile forms can be seen as thematic of the struggle a captain must endure to turn their model kit into the ultimate cool gun.

PR #18366, opened 2024-03-31
gang leader delegation

TDHooligan:
[BUG]
## About the PR 
when gang leaders enter cryo with no locker, they are immediately replaced with another gang member 
(if they have a locker, there is already a 7.5 minute timer until the role can be claimed)

## Why's this needed? 
gang leader cryoed so hard all my mitigating factors broke down. gang sad ):

### DIFF
--- a/_std/defines/gang.dm
+++ b/_std/defines/gang.dm
@@ -1,7 +1,7 @@
 #define GANG_MAX_MEMBERS 4
 
 /// How long the leader must cryo before gang members can take their role
-#define GANG_CRYO_LOCKOUT 15 MINUTES
+#define GANG_CRYO_LOCKOUT 7.5 MINUTES
 
 // -------------------------
 // GANG ECONOMY
--- a/code/datums/gamemodes/gangwar.dm
+++ b/code/datums/gamemodes/gangwar.dm
@@ -351,7 +351,7 @@ proc/broadcast_to_all_gangs(var/message)
 	/// The mind of this gang's leader.
 	var/datum/mind/leader = null
 	/// The minds of gang members associated with this gang. Does not include the gang leader.
-	var/list/members = list()
+	var/list/datum/mind/members = list()
 	var/list/tags = list()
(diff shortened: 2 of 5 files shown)


_std/defines/gang.dm
TDHooligan, line 1: lowered this - 15 minutes is enough to set a gang back probably too long

code/datums/gamemodes/gangwar.dm
```diff
@@ -413,9 +413,44 @@ proc/broadcast_to_all_gangs(var/message)
...
 
-	proc/handle_leader_cryo()
-		broadcast_to_gang("Your leader has entered cryogenic storage. You can claim leadership at your locker.")
-		leader_claimable = TRUE
+	/// how to handle the gang leader entering cryo (but not guaranteed to be permanent)
+	proc/handle_leader_temp_cryo()
+		if (!src.locker)
+			choose_new_leader()
+		else
+			broadcast_to_gang("Your leader has entered temporary cryogenic storage. You can claim leadership at your locker in 7.5 minutes.")
```
ZeWaka, line 421: use the define

ZeWaka (changes requested):

TobleroneSwordfish (changes requested):

code/obj/cryotron.dm
```diff
@@ -199,6 +199,8 @@
 				if ("Yes")
 					if (tgui_alert(user, "Are you absolutely sure you want to enter cryogenic storage?", "Confirmation", list("Yes", "No")) == "Yes")
 						if (mob_can_enter_storage(user)) // check again in case they left the prompt up and moved away/died/whatever
+							for(var/datum/antagonist/antagonist as anything in user.mind?.antagonists)
+								antagonist.handle_cryo()
```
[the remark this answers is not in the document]
TDHooligan: good catch.

ZeWaka (approved):

PR #18367, opened 2024-03-31
Handsfree radiation collector array.

Francisco-Franks:
[Engineering]

## About the PR 

This allows tanks to be click-dragged to the radiation collector array.

## Why's this needed? 
My friend Nuggets the Chief Engineer, who lost all their limbs, wants to make the best singularity. Would you really deny Nuggets their destiny?

## Changelog 

```changelog
(u)MetaReferencer
(+)You can click-drag plasma tanks onto radiation collectors.
```

### DIFF
--- a/code/obj/machinery/singularity.dm
+++ b/code/obj/machinery/singularity.dm
@@ -1457,6 +1457,18 @@ TYPEINFO(/obj/machinery/power/collector_array)
 			UpdateIcon()
 		..()
 
+/obj/machinery/power/collector_array/MouseDrop_T(obj/item/W, mob/user) // For the armless and the harmless.
+	if (!istype(W, /obj/item/tank)) // Tanks only, please.
+		return
+	if (!in_interact_range(src, user)  || BOUNDS_DIST(W, user) > 0 || !can_act(user)) // No attaching tanks from across the world.
+		return
+	// No ghosts or AI eye or wraiths or blobs or flockminds shall be engineers. This is for the physical and the living.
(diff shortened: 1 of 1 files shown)


444Portal:
Tagged with Silicons as that is what the original PR that was later adapted into this was aimed at.

My personal thoughts:
Despite the fact that I did say I would prefer a click drag, this seems odd. I believe the general consensus was that this is not a good idea, as Cyborgs should not be able to do anything. Given that this PR, which accomplishes the exact same thing as the prior one, has Cyborgs missing from any of the descriptions, it seems even more odd.

If this is not aimed at Cyborgs, why do we need a click-drag?
If this is aimed at Cyborgs, do they really need to be able to do this? As previously discussed, you can call a human over often. Even then, Cyborgs have no issue setting up solars which does the job in most cases.

Francisco-Franks:
Part of Leah's argument was "TONGS, FOR A SPECIFIC ITEM, FOR BORGS" so I am eliminating this. This is exactly on parity with, say, borgs being able to do surgery via clickdrag. It also allows a hypothetical armless human to do their job. This doesn't allow interaction with anything else other than the radiation collector. It is as hyperfocused as it can get, borgs can't cool the plasma or get cool plasma, they can use the most basic available plasma tank and that is literally it.
In short, there is precedent (in the surgery stuff), and it allows a borg to set up the most basic power gen and that is all.

TobleroneSwordfish:
There is precedent, but the precedent is IMO bad design and shouldn't be used as an argument to keep adding clickdrag interactions to everything borgs can't do.

Francisco-Franks:
Twas worth a shot.

PR #18369, opened 2024-04-01
Rebalance gang weaponry

TDHooligan:
[GAMEMODES][BALANCE]
## About the PR 
Lowers the overall stun power of gang weapons. Improves their lethality to boot.

## Why's this needed? 
Gang/sec balance revolves around sec's stun power having a leg-up on gangs. Sec should be able to arrest a lone murdery gang member, or at least stun & run away if they're outnumbered.

However, all the old gang weapons are superbly good at stun. So good, in fact, that security don't have a chance. They receive a knife/D-sword/shuriken to the torso and then get permastunned and die.

This PR buffs gangs' lethal potential, but lowers their stun potential. Security officers are naturally a lot more resistant to being Gun or being Stab. So they're a bigger threat now.
- D-Saber disorient is now very brief (will mess up a tile's movement)
   - To give something back, D-saber now does more damage, and is CUT instead of BURN (so it hits BRUTE death rolls + bleed)
   - also more consistent with being a Discount CSaber (rather than a fake csaber)  
- Throwing knives do less stun. Shuriken no longer instantly stun.
   - I have old shuriken code to refactor, that should make shuriken a much more fun/better weapon overall. This will help in the short term

- Syndicate fighting knives from the gang crate are swapped with a new version that uses the gang palette. They also don't stun any more.
 
## Changelog 

```changelog
(u)TDHooligan
(+)Gang weapons no longer stun as hard. They kill harder. So kill each other harder, gangs.
```


code/obj/item/misc_weapons.dm
```diff
@@ -495,18 +502,22 @@ TYPEINFO(/obj/item/sword/pink/angel)
...
-	if (active)
-		hit_type = DAMAGE_BURN
-	else
-		hit_type = DAMAGE_BLUNT
+
+	if (royale_mode)
+		if (active)
+			hit_type = DAMAGE_BURN
+		else
+			hit_type = DAMAGE_BLUNT
```
TobleroneSwordfish, line 510: Gamemode switches like this are bad, you've already made a subtype so either put a var or proc override on that instead. Orrr maybe just change it globally since the changes probably also make sense in BR and having two identical versions of the same weapon that work differently is Bad.

code/obj/item/misc_weapons.dm
```diff
@@ -711,6 +732,11 @@ TYPEINFO(/obj/item/sword/pink/angel)
 	stamina_cost = 20
 	stamina_crit_chance = 60
 	// pickup_sfx = 'sound/items/blade_pull.ogg'
+	gang
+		force = 12
+		throwforce = 6
+		stamina_damage = 40
+		stamina_crit_chance = 20
```
TobleroneSwordfish, line 739: Nunchucks are kind of meant to be a stamina weapon, if you don't want gangs having stamina weapons then maybe just don't give them nunchucks?

code/obj/item/misc_weapons.dm
```diff
@@ -595,16 +605,27 @@ TYPEINFO(/obj/item/sword/pink/angel)
...
 	item_state = "ninjaknife"
 	force = 8
-	throwforce = 11
+	throwforce = 18
 	throw_range = 10
 	flags = FPRINT | TABLEPASS | USEDELAY //| NOSHIELD
 	desc = "Like many knives, these can be thrown. Unlike many knives, these are made to be thrown."
+	gang
+		name = "familiar fighting knife"
+		force = 21
```
TobleroneSwordfish, line 614: This is a *lot* of damage, more than double that of the switchblade which feels kind of wrong? Also no-one is ever going to throw something that does that much melee damage and that little disorient when thrown.

TobleroneSwordfish (changes requested):

code/obj/item/misc_weapons.dm
TDHooligan: yeah, you're right it's too high when you look at the other knives. i'll knock it down a peg. The main unique trait this brings to the table is definitely the movespeed, given it's in a swathe of other throwing options.

code/obj/item/misc_weapons.dm
TDHooligan: can do. just wanted to have all bases covered for not buffing it in BR. but its probably ok

PR #18375, opened 2024-04-01
Power shovel was invisible due to wrong path.

Francisco-Franks:
Moved icons from sealab_power.dmi to mining.dmi

## About the PR 
Moved the 2 icons for the powershovel to mining.dmi from sealab_power.dmi . I assume it was overlooked during the refactor. Without this the tool is invisible except for the glowy blade when it is on.

## Why's this needed? 
I think that the shovel should have visuals to match its tangibility.


Francisco-Franks:
image
image
image

TobleroneSwordfish (approved):
Looks good, bugfixes don't need changelog entries though

PR #18382, opened 2024-04-02
Gang crate teleportation adjustments

TDHooligan:
[GAMEMODES][FEATURE]
## About the PR 
Gang crates now spawn a marker that only Gang Members can see instead of teleporting in, locked in place.

https://github.com/goonstation/goonstation/assets/6061314/a604979c-9f1f-4280-87be-8e40859812fd

## Why's this needed? 

Sec were catching on too quickly about gang crates and preparing around them to confiscate them. After some discussion, it was decreed it'd be more interesting to have only the gangs in on it - gangs will now break out fighting with no context in random rooms, leaving sec to be more reactive. 

## Changelog 

```changelog
(u)TDHooligan
(+)Gang crates now take 10 seconds to un-anchor. Gang members can see where they'll spawn 5 minutes early.
(+)Gang crates will no longer unceremoniously swallow and capture people when they spawn.
```


code/datums/controllers/process/gang.dm
```diff
@@ -110,18 +110,25 @@
 		if (!length(turfList))
 			message_admins("All attempts to find a valid location to spawn a weapons crate failed!")
 			return
-		var/obj/storage/crate/gang_crate/guns_and_gear/crate = new(pick(turfList))
-		broadcast_to_all_gangs("<span style='font-size:24px'> We've dropped off weapons & ammunition at <b>\the [drop_zone.name]!</b> It's anchored in place for 5 minutes, so get fortifying!</span>")
+		var/turf/location = pick(turfList)
+		var/datum/client_image_group/imgroup = get_image_group(CLIENT_IMAGE_GROUP_GANG_OBJECTIVES)
+		var/image/objective_image = image('icons/effects/gang_overlays.dmi', location, "cratedrop")
```
TobleroneSwordfish, line 115: Not sure what this will do if the turf is RCDed/exploded? Might want to consider making this an `/obj/effects` with an image attached instead.

_std/defines/gang.dm
```diff
@@ -101,7 +101,8 @@
 
 // CRATE DROP DEFINES
 #define GANG_CRATE_SCORE 500 //! how many points gang crates grant to each member, when opened
-#define GANG_CRATE_LOCK_TIME 300 SECONDS //! how long gang crates stay locked to the floor, in seconds
+#define GANG_CRATE_DROP_TIME 300 SECONDS //! How long it takes for gang crates to arrive after being announced, in seconds
+#define GANG_CRATE_LOCK_TIME 10 SECONDS //! How long it takes for gang crates to arrive after being announced, in seconds
```
TobleroneSwordfish, line 105: Comment needs updating

TobleroneSwordfish (changes requested):

PR #18383, opened 2024-04-02
Long Range Teleport Computer 

Azrun:
## About the PR 
Folds the functionality of the Long Range Teleporter into the telepad and teleporter computer.  Teleman expanded to include this support, but is limited to send and receive. 

Existing GPS based a zones have been updated to use long range teleport landmarks and new galactic coordinate disks. These discs can be inserted into the teleporter computer to add them as a new destination.

New small adventure zone has been added called The Observatory. The observatory contains discs for all current adventure zones. This will allow for the observatory to act as a hub for the current set of adventure zones. This can then open up for a future adventure zones to chain together.

- [x] Access to The Observatory as well as the lunar museum will be added as part of roundstart.
- [x] Rip out LRT from Station Maps.

## Why's this needed? 
The reliance and obfuscation of adventure zones on teleportation put it in a somewhat always position for quite some time. There has also been a fair amount of discussion of shifting access to adventure zone to either Artemis or the long-range teleporter off and on and this seems like a great way of following through with that. 

From a thematic standpoint, this also creates a great delineation between local teleportation, which has improved flexibility selecting coordinates and long-range teleportation which should be a bit more limited.

## Changelog 

```changelog
(u)Azrun
(*)Long Range Teleport has been folded into the normal Teleporter!  Check out the Observatory to go on new adventures!
```


Sbmhawk:
What about non-azone uses for telesci?
Stuff on debris/station level.

Azrun:
> What about non-azone uses for telesci?
> Stuff on debris/station level.

There should be no change. If you see something that impacts it, please bring it up.

code/modules/networks/computer3/mainframe2/telesci.dm
```diff
@@ -469,6 +491,65 @@ TYPEINFO(/obj/machinery/networked/telepad)
...
+				showswirl_out(target)
+				use_power(1500)
+				for(var/atom/movable/M in target)
+					if(M.anchored)
+						continue
+					animate_teleport(M)
+					if(ismob(M))
+						var/mob/O = M
+						O.changeStatus("stunned", 2 SECONDS)
+					SPAWN(6 DECI SECONDS) do_teleport(M,src.loc,FALSE,use_teleblocks=FALSE,sparks=FALSE)
```
ZeWaka, line 549: indent

code/modules/networks/computer3/mainframe2/telesci.dm
```diff
@@ -469,6 +491,65 @@ TYPEINFO(/obj/machinery/networked/telepad)
...
+				use_power(1500)
+
+				for(var/atom/movable/M in src.loc)
+					if(M.anchored)
+						continue
+					animate_teleport(M)
+					if(ismob(M))
+						var/mob/O = M
+						O.changeStatus("stunned", 2 SECONDS)
+					SPAWN(6 DECI SECONDS) do_teleport(M,target,FALSE,use_teleblocks=FALSE,sparks=FALSE)
```
ZeWaka, line 520: x

Cherman0:
This sounds really cool. Why not use/rework the Solarium station to be the observatory? Canonically it is _an_ observatory (of the sun) and since its whole central schtick involves going to other azones it ties the puzzle into its actual function. Aside from the automaton (and relevant objects for the Helios medal) nothing about Solarium's design is actually relevant to its puzzle, so its not really load-bearing if it needed to be pretty much completely remapped.

Azrun:
> This sounds really cool. Why not use/rework the Solarium station to be the observatory? Canonically it is _an_ observatory (of the sun) and since its whole central schtick involves going to other azones it ties the puzzle into its actual function. Aside from the automaton (and relevant objects for the Helios medal) nothing about Solarium's design is actually relevant to its puzzle, so its not really load-bearing if it needed to be pretty much completely remapped.

Cause I made my own observatory.  Not for looking at the Sol but listening to the galaxy! A Radio Astronomy Observatory of sorts.  🤷

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
```diff
@@ -28,41 +35,139 @@ export const TeleConsole = (_props, context) => {
...
+                  onRetry={handleRetryConnect}
+                  padNum={padNum}
+                />
+              </Section>
+            </Stack.Item>
+          )}
+          {(tab===Tab.LongRange) && (
+            <Stack.Item>
+              <Section>
+                <Section title="Destinations">
```
mordent-goonstation, line 102: Nesting `Section`s seems a bit odd here, how does this end up looking?

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
```diff
@@ -28,41 +35,139 @@ export const TeleConsole = (_props, context) => {
...
+                  <LabeledList>
+                    {destinations.length ? destinations.map((d) => (
+                      <div key={d["name"]}>
```
mordent-goonstation, line 105: What's the reason for the `div` layer and not just `LabeledList.Item`?

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
```diff
@@ -28,41 +35,139 @@ export const TeleConsole = (_props, context) => {
...
+                        <LabeledList.Item
+                          label={d["name"]}>
```
mordent-goonstation, line 107: Broadly, if `destinations` is an array of known structured objects, you access as `d.name` instead of `d["name"]`. Your way works, just reads oddly and makes me think that you're trying to get around some strange types when in reality it's just DM-like object access syntax.

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
```diff
@@ -28,41 +35,139 @@ export const TeleConsole = (_props, context) => {
...
+                />
+              </Section>
+            </Stack.Item>
+          )}
+          {(tab===Tab.LongRange) && (
+            <Stack.Item>
+              <Section>
+                <Section title="Destinations">
+                  <LabeledList>
+                    {destinations.length ? destinations.map((d) => (
```
mordent-goonstation, line 104: Looks like you only use the `name` here, consider doing as:
```suggestion
                    {destinations.length ? destinations.map(({ name }) => (
```
Then you can just refer to `name` instead of `d.name` (or `d["name"]`), and it also lets you write object instantiation as a shorthand (`act('action', { name })`)

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
```diff
@@ -28,41 +35,139 @@ export const TeleConsole = (_props, context) => {
...
+                />
+              </Section>
+              {(!!disk) && (
+                <Section
+                  title="Disk Controls"
+                >
+
+                  <Button
+                    icon="upload"
+                    color={"blue"}
```
mordent-goonstation, line 154: Color override needed?

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
```diff
@@ -28,41 +35,139 @@ export const TeleConsole = (_props, context) => {
...
+                    onClick={() => act("scan_disk")}>
+                    Read from Disk
+                  </Button>
+                  <Button
+                    icon="eject"
+                    color={"bad"}
```
mordent-goonstation, line 160:
```suggestion
                    color="bad"
```

tgui/packages/tgui/interfaces/TeleConsole/types.ts
```diff
@@ -18,6 +18,13 @@ export interface TeleConsoleData {
   padNum: number;
   maxBookmarks: number;
   bookmarks: BookmarkData[];
+  destinations: LongRangeData[];
+  disk: BooleanLike;
+}
+
+export interface LongRangeData {
+  ref: string;
```
mordent-goonstation, line 26: Doesn't look like you're using the `ref`, instead passing the `name` back in places. Given that names should be unique here that's fine, but as you're giving it the data you'd might as well use it.

Prefer `fooRef` as a variable name here, as `ref` has meaning within React and can complicate trying to pass it around. (Not a problem with how I think you'd use it, but avoids an issue and makes it easier for the next person to use this as an example)

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
```diff
@@ -28,41 +35,139 @@ export const TeleConsole = (_props, context) => {
...
+                icon="wrench"
+                selected={tab === Tab.LongRange}
+                onClick={() => setTab(Tab.LongRange)}
+              >
+                Long Range
+              </Tabs.Tab>
+            </Tabs>
+          </Stack.Item>
+
+          {(tab===Tab.Local) && (
```
mordent-goonstation, line 60: These slices are decently long, consider separating each tab into its own component.

mordent-goonstation (commented):
Skimmed the tgui side only.

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
Azrun: image

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
Azrun: Results in a different tint of blue.

mordent-goonstation: Right, but is that a problem? Consistent color scheme vs. aesthetic preference is a rough discussion, sure.

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
mordent-goonstation: Marking this as unresolved, because that makes no real sense to me.

`<LabeledList.Item key={d.name} label={d.name}>{/* whateverInHere */}</LabeledList.Item>` should absolutely work in the context of return value for a `.map` callback.

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
```diff
@@ -6,63 +6,131 @@
...
+                {readout && <Section>{formatReadout(readout)}</Section>}
+                <ConnectionSection
+                  isConnected={isConnectedToHost}
+                  isPanelOpen={!!isPanelOpen}
+                  onCyclePad={handleCyclePad}
+                  onReset={handleResetConnect}
+                  onRetry={handleRetryConnect}
+                  padNum={padNum}
+                />
+              </Section>
```
mordent-goonstation, line 125: Wrapping all of these in a `Section` is (probably) not needed. `Section`s one after the other *should* look fine without a wrapping `Section`.

mordent-goonstation (commented):
Looked through tgui side again, unresolved a couple of the earlier comments to have them re-looked at, left one more. Broadly, though, looks good and these are more nitpicks/me being grumpy with some of our tooling.

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
Azrun: It is needed, or at least a means, to maintain a consistent with the existing look and feel where there is no visual break between sections.
Implemented -> Suggested  and the Local page.
image

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
Azrun: It is if that is the desired behavior.  Removed in favor of consistent color scheme.

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
Azrun: Updated to be consistent with existing usage and ignoring _ts(2322)_ 🤷

tgui/packages/tgui/interfaces/TeleConsole/index.tsx
mordent-goonstation: Gotcha. I don't actually have issue with the second one, as the connection status isn't really part of the Destinations section so it being its own visual thing seems "fine" to me, but I'm not going to nitpick it further.

mordent-goonstation (approved):
tgui side looks good, not reviewed the rest but fine with the overall idea as stated in the PR description

ZeWaka:
!merge_upstream

PR #18392, opened 2024-04-03
Replaces Airless tiles on the prefab space mining outpost

cheekybrdy:
## About the PR 
The title.

## Why's this needed? 
Area is a NT site and should have air in the rooms where these airless tiles are.


ZeWaka (approved):
fix your master branch

PR #18393, opened 2024-04-03
Adds Goku Costume & Saiyan Hair

itsloley:
[CLOTHING] [SPRITES]

## About the PR 

adds a costume set purchasable from geoff honkington's shop containing a jumpsuit, boots, and a wig based off of goku from dragon ball z

also adds super saiyan hair (which can sometimes be in the costume box instead of the regular wig) oooorr just use scissors

legally-distinct-anime-martial-artist

## Why's this needed? 

because goku's awesome!! and there was a shocking amount of people who actually wanted this when i showed off of the sprites

## Changelog 

```changelog
(u)Lyy
(+)Added a certain anime martial artist's costume to Geoff Honkington's shop.
```


Kyle2143:
I love having this in the code as a thing for admins to dole out or for players to happen upon. But making it so easily available all in one set to just "become goku" is much too easy. 

I'd say, don't include the wig with the set, just sell the gi. If they wanna cosplay as Goku, make them put in a little work for it. It's what Goku would do.

itsloley:
> I'd say, don't include the wig with the set, just sell the gi. If they wanna cosplay as Goku, make them put in a little work for it. It's what Goku would do.

i totally get removing the wig and stuff! but i kept the wig available in the box for mutantraces and stuff if they want to have goku hair, since they can't really have the hair at ALL without asking someone to make a wig off of their own hair.. any other ideas, or should that _really_ be the case?

Kyle2143:
> > no, the goku/saiyan hairstyles are barber only
> 
> Ah, wasn't aware of that, nvm then.

I think this is giving too much consideration to a tiny issue.

Mutantraces could still get it a goku wig with just slightly more effort. They just gotta give the haircut to someone else and thenchop it off to make a wig then wear it themselves. It's basically just marginally harder than getting someone to give you the haircut yourself.

itsloley:
> I think this is giving too much consideration to a tiny issue.
> 
> Mutantraces could still get a goku wig with just slightly more effort. They just gotta give the haircut to someone else and then chop it off to make a wig then wear it themselves. It's basically just marginally harder than getting someone to give you the haircut yourself.

i still think having mutantraces try to convince someone to lop off their hair, or take someone from the space diner, or even get a monkey, ask genetics to humanize it, and then put hairgrownium on the humanized monkey to get hair and then cut it yourself is a little bit excessive compared to humans just being able to cut their own hair basically as fast as they get the costume box

it's not really like cutting your hair is actually any meaningful amount of work either, since again if you just have scissors in your hand you could give yourself the same look and it'd take only like 5 seconds longer

i'd prefer to keep the wig as much as possible since removing it would only add such a minuscule amount of work for humans and cause it to be a _lot_ more annoying for mutantraces, but i guess if there's no wiggle room for a compromise that isn't completely scrapping the wig i'll get rid of it

PR #18404, opened 2024-04-04
Random Event: Lost and Found

pgmzeta:
[EVENTS][FEATURE]
## About the PR 

Adds a new random event that spawns several 'lost' items in closed crates/lockers/closets around the station. These items will track who picks them up, and if they're put into the hands of someone who's name matches they will:
* Change their name/description to indicate they're returned
* Payout 100 credits to all security team members
* Issue a PDA notice to security
* Removes the signal handler to prevent reward spamming. It's already returned!

The origin of this idea came from Flourish's random lost item spawner, but I was unable to use it directly in this implementation.

## Why's this needed? 
* Give security more light-weight stuff to do (with suitably light rewards)
* More reasons for crew to talk with other crew - it doesn't *have* to go through sec
* Random event variety

## Changelog 

```changelog
(u)glowbold
(+)NanoTrasen's Logistics division will alert you if there is a large spike in reported lost items.
```


Sbmhawk:
Should probably be a credit payout to whoever returned it, not security.

Otherwise, why would any non-sec bother

code/modules/events/minor/lost_and_found.dm
```diff
@@ -0,0 +1,56 @@
...
+			var/item_to_lose = pick(possible_lost_items)
+			var/obj/item/lost_item = new item_to_lose(get_turf(hiding_spot))
+			lost_item.loc = hiding_spot
+
+			var/datum/db_record/item_owner = pick(data_core.general.records)
+			lost_item.name = "lost [lost_item.name]"
+			lost_item.desc = "Didn't [item_owner["name"]] lose one of these?"
+			lost_item.add_fingerprint_direct(item_owner["fingerprint"])
+
+			RegisterSignal(lost_item, COMSIG_ITEM_PICKUP, PROC_REF(lost_item_pickup))
```
Azrun, line 30: I think I'd rather see this as a `component` that way it doesn't rely on the item's description.

code/modules/events/minor/lost_and_found.dm
```diff
@@ -0,0 +1,56 @@
...
+
+/datum/random_event/minor/lost_and_found/proc/lost_item_pickup(obj/item/lost_item, mob/user)
+	if (findtext(lost_item.desc, user.real_name, 4)) // skip "lost"
```
Azrun, line 33: Why is "Foo Bar" getting all of "Foo Barz" lost and found items?!?!?

Azrun, line 33: Comment matches string index but desc and real_name are used.

Azrun, line 33: Doesn't this have the ability to be a _really odd_ means of IDing someone.

Azrun (commented):
Neat!  I know this is a WIP but wanted to make a few notes.

PR #18405, opened 2024-04-04
Fixes hattable component breaking if parent is clicked with a non-hat

CalliopeSoups:
## About the PR 
The item argument wasn't being removed at the end of put_hat_on. 

## Why's this needed? 
Bugfix

### DIFF
--- a/code/datums/components/hattable.dm
+++ b/code/datums/components/hattable.dm
@@ -34,7 +34,7 @@ TYPEINFO(/datum/component/hattable) // Take a walk through my TWISTED mind.... I
 /datum/component/hattable/proc/hat_on_thing(mob/target as mob, obj/item/item as obj, mob/attacker)
 	if (src.hat)
 		item = null
-		return TRUE
+		return
 
 	var/atom/movable/hatted = src.parent
 	var/offsetBy_y = 0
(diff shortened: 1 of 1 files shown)


code/datums/components/hattable.dm
```diff
@@ -33,19 +33,21 @@ TYPEINFO(/datum/component/hattable) // Take a walk through my TWISTED mind.... I
 
 /datum/component/hattable/proc/hat_on_thing(mob/target as mob, obj/item/item as obj, mob/attacker)
 	if (src.hat)
-		return
+		item = null
+		return TRUE
```
TobleroneSwordfish, line 37: Won't returning true here make it impossible to attack the parent at all when it has a hat on?

code/datums/components/hattable.dm
```diff
@@ -33,19 +33,21 @@ TYPEINFO(/datum/component/hattable) // Take a walk through my TWISTED mind.... I
...
+
+	if (istype(item, /obj/item/clothing/head/))
+		src.hat = item
 		ADD_FLAG(src.hat.appearance_flags, KEEP_TOGETHER) // Flags needed for wigs!
 		ADD_FLAG(src.hat.vis_flags, VIS_INHERIT_DIR)
 	else
-		return
-
+		item = null
+		return TRUE
```
TobleroneSwordfish, line 50: Same here, we want attacks to go through if the item is say a crowbar.

TobleroneSwordfish (changes requested):
I don't think the `item = null` lines are doing anything? You're just clearing the argument and then immediately returning so it isn't used anyway.

PR #18407, opened 2024-04-04
Fix air alarms not correctly posting status to alert computer

pgmzeta:
[station systems][bug]
## About the PR 
Use defines to make the safety logic in air alarms more straightforward, and fix a related bug in packets posted to the alert computer.

## Why's this needed? 
* Critical air alarms should be reported correctly to the alert computer.
* If an alarm is fixed, it should be properly reset to clear in the alert computer.

### DIFF
--- a/code/obj/machinery/alarm.dm
+++ b/code/obj/machinery/alarm.dm
@@ -2,6 +2,10 @@
 // Alarm
 //
 
+#define ALARM_SEVERE 0
+#define ALARM_MINOR 1
+#define ALARM_GOOD 2
+
 /obj/machinery/alarm
 	name = "air monitor"
 	icon = 'icons/obj/monitors.dmi'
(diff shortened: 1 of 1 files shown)


code/obj/machinery/alarm.dm
```diff
@@ -31,7 +35,7 @@
 	//	list(varname = "oxygen_agent_b", friend_name = "Unknown", safe_min=0, safe_max=INFINITY, good_min=0, good_max=INFINITY),
 	)
 	var/const/temp_safe_min = T0C-15
-	var/const/temp_safe_max = T0C+66
```
pgmzeta, line 34: unrelated, but while we're here

code/obj/machinery/alarm.dm
```diff
@@ -154,11 +155,11 @@
...
 	switch (alert_level)
-		if (2)
+		if (ALARM_SEVERE)
 			alert_signal.data["alert"] = "severe"
-		if (1)
+		if (ALARM_MINOR)
 			alert_signal.data["alert"] = "minor"
-		if (0)
+		if (ALARM_GOOD)
 			alert_signal.data["alert"] = "reset"
```
pgmzeta, line 163: This is the section that was bugged

ZeWaka (approved):

PR #18419, opened 2024-04-05
Ship Alert button targets doors by ID instead of area

pgmzeta:
## About the PR 
* Check the shutter's ID value instead of area
* Check a wider base path for potential lockdown doors on various maps

## Why's this needed? 
A lot of bridge doors get out of sync

atlas bridge out of area
image

clarion bridge uses blast doors
image

cogmap bridge out of area
image

cogmap2 bridge out of area
image

kondaru AI core uses poddoors
image

oshan bridge out of area
image

nadir AI core uses poddoors
image

donut2 bridge out of area
image

donut3 bridge out of area
image

donut3 AI core uses different area
image

### DIFF
--- a/code/obj/machinery/shipalert.dm
+++ b/code/obj/machinery/shipalert.dm
@@ -145,12 +145,12 @@ TYPEINFO(/obj/machinery/shipalert)
 		LAGCHECK(LAG_LOW)
 
 /obj/machinery/shipalert/proc/do_lockdown(mob/user)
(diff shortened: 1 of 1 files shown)


pgmzeta:
let me know if any of these should be map fixes instead

TobleroneSwordfish (approved):
Oh this is much better yep, I assumed the IDs would be wildly inconsistent between maps

PR #18421, opened 2024-04-06
Remove slight gradient from Asimovs laws

JORJ949:
[Silicons] [Feature]
## About the PR 
Removes minor gradient from Asimov's law set

## Why's this needed? 
Who even knew this was a thing? Removed to make the disguised law module actually the same colour as asimov's no matter what law you replace.

### DIFF
--- a/code/obj/item/ai_modules.dm
+++ b/code/obj/item/ai_modules.dm
@@ -116,17 +116,14 @@ TYPEINFO(/obj/item/aiModule)
 /******************** Modules ********************/
 /******************** Asimov ************************/
 /obj/item/aiModule/asimov1
-	highlight_color = rgb(0, 167, 0, 255)
 	name = "AI Law Module - 'Asimov's 1st Law of Robotics'"
 	lawText = "You may not injure a human being or cause one to come to harm."
 
 /obj/item/aiModule/asimov2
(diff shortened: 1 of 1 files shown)


amylizzle:
That's there to provide a visual clue if someone messes with the law order. Having said that, if you're fine with it being removed then so am I

JORJ949:
> That's there to provide a visual clue if someone messes with the law order. Having said that, if you're fine with it being removed then so am I 

I dont think I have ever seen that used mainly because of how negligible the difference is but also because if you want to mess with laws its probably not going to be changing the order around, and if you are gonna put in that effort it shouldnt really be noticable from a glance and the gradient kind of defeats the purpose of the disguised module.

TobleroneSwordfish (approved):
I think it's fine to be able to hide that you've messed with the order, and I certainly never noticed this was a thing sooo

amylizzle (approved):
Yeah, fine by me :+1:

amylizzle:
One thought is that you might wanna do the same for the other law sets - iirc they both have gradients too

JORJ949:
> One thought is that you might wanna do the same for the other law sets - iirc they both have gradients too

The only other lawset with a gradient is the syndicate one but that doesn't matter really

PR #18426, opened 2024-04-06
Improve Wizard Spellbook UX

mordent-goonstation:
## About the PR 

Moves the Wizard Spellbook to a TS interface.
Improves the wizard spellbook to allow more intuitive ways to display and find spells.

image
image

## Why's this needed? 

Having to navigate the accordion menu to find the spell you want was arguably tricky, and didn't make for easy to find spells given the way it was sorted.

## Changelog 

```changelog
(u)Mordent
(+)Wizard Spellbook UX improved updated to make spells easier to find.
```


ZeWaka:
!merge_upstream

code/obj/item/uplinks.dm
```diff
@@ -1334,23 +1334,24 @@ Note: Add new traitor items to syndicate_buylist.dm, not here.
...
-
+			if (!spellbook_contents[spell.eqtype])
+				// create category if it doesnt exist
+				spellbook_contents[spell.eqtype] = list()
+			if (spell.assoc_spell && ispath(spell.assoc_spell, /datum/targetable/spell))
+				var/datum/targetable/spell/spell_ability_datum = spell.assoc_spell
+				cooldown_contents = initial(spell_ability_datum.cooldown)
+				spell_icon = icon2base64(icon(initial(spell_ability_datum.icon), initial(spell_ability_datum.icon_state), frame=6))
+			spellbook_contents[spell.eqtype] += list(list(
+				cooldown = cooldown_contents,
```
Azrun, line 1348: Consider changing the units of this to seconds, to allow for display as provided.  Would also remove seemingly arbitrary conversion from deciseconds.

tgui/packages/tgui/interfaces/WizardSpellbook/SpellItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+                    act('buyspell', { spell: name });
+                  }}>
+                  {buildPurchaseText(purchased, cost, spellSlots)}
+                </Button>
+              </Stack.Item>
+            </Stack>
+          </Stack.Item>
+          <Stack.Item>
+            <LabeledList>
+              {cooldown && <LabeledList.Item label="Cooldown">{`${cooldown / 10} seconds`}</LabeledList.Item>}
```
Azrun, line 76: Update accordingly if action taken on decisecond -> second conversation on DM side.

Azrun (approved):

code/obj/item/uplinks.dm
mordent-goonstation: Port of existing, but valid.

PR #18429, opened 2024-04-07
remove zipgun/slamgun selling for Gangs

TDHooligan:
[GAMEMODES][BALANCE]
## About the PR 
Disallows zipguns/slamguns getting sold altogether

## Why's this needed? 
after more thought and discussion, this is similar to Botany but can be done anywhere, risk free, with no sign of a gun crafting operation besides a pile of sheets and welding tools. totally normal things to have.

It's not a particularly engaging way to play, and there are more fun things to steal. If I lower the point score it just demands players do the same thing for even longer which is twice as boring. 

## Changelog 

```changelog
(u)TDHooligan
(*)Gangs can no longer sell slamguns/zip guns. Go do some *REAL* crime instead of welding pipes together.
```

### DIFF
--- a/code/datums/gamemodes/gangwar.dm
+++ b/code/datums/gamemodes/gangwar.dm
@@ -1623,10 +1623,8 @@ proc/broadcast_to_all_gangs(var/message)
 
 				return
 
-			if(istype(item, /obj/item/gun/kinetic/slamgun))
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (approved):
I saw a gang get 22k points without leaving cargo

PR #18430, opened 2024-04-07
Adds Job XP reward for completing crew objectives

MintyPhresh:
## About the PR 
When a crew objective is completed, it grants XP (Default 50) for whatever job the mind was assigned at the start of the round. This includes jobs that otherwise wouldn't be able to gain XP, like Captain.

## Why's this needed? 
It makes a good deal of sense for job XP to be awarded when you complete job-specific quests, and it helps out jobs that otherwise have little to no XP gain (like Bartender, whose top 20 players have a median of less that 1,500 XP). It still shouldn't be a major source of XP if you're grinding out hundreds of XP per shift, but it'll help a lot if the way you play your role doesn't involve repeatedly doing the things that give XP. This extra incentive to do objectives might also promote people adding new, interesting objectives, or just interacting with them more in general!

```changelog
(u)mintyphresh
(+)Completing a crew objective now grants 50 XP for that job.
```

### DIFF
--- a/code/datums/crew_objective.dm
+++ b/code/datums/crew_objective.dm
@@ -71,6 +71,7 @@
 
 ABSTRACT_TYPE(/datum/objective/crew)
 /datum/objective/crew
+	var/XPreward = 50
 
 /datum/objective/crew/custom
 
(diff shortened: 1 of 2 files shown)


MintyPhresh:
Even if we're not keeping it around, it's here now, so some improvements to it would benefit the game regardless

Glamurio:
What the game truly needs are missions that are actually fun doing. There's only so many times where protecting Stirstir or having blood in your alcohol stream is actually enjoyable. We'd need more varied, dynamic objectives, maybe stuff that targets specific players or specific substances, that would get people to do them rather than ignore them. But this is a good change either way.

ZeWaka (approved):
good change either way but yeah it would be nice if this stuff was better:tm:

PR #18434, opened 2024-04-07
Add cooldown for give item delay

### LINKED ISSUE #18432: exploit to lag out any individual whenever you want
### Labels

[BUG]

### Description

by handing someone an item via spacebar+left click tangentially to them (even as critter) very rapidly, it freezes their game until a while after you stop spamming the action. uh oh!

### Steps to reproduce

1. hold an item
2. find a victim
3. stand next to them
4. hold spacebar
5. click on them repeatedly, macro optional. they are now frozen until you stop spamming

### Expected Behavior

for there to be some kind of limit on how many requests to hand someone an item there can be within a certain timeframe

_No response_

pgmzeta:
## About the PR 
Add a 1 second cooldown delay for `give_item` verb to call `give_to`.
Move dupe boutput "You hand THING to MOB" from human/critter text into the give item proc, after early returns.

## Why's this needed? 
Fix #18432
While the SPAWN delays how instantaneous the dialog appears to the target, it doesn't stop multiple dialogs from being queued up one after the other.

### DIFF
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -1405,10 +1405,13 @@
 		boutput(usr,SPAN_ALERT("You can't give items to yourself!"))
 		return
 
-	SPAWN(0.7 SECONDS) //secret spawn delay, so you can't spam this during combat for a free "stun"
-		if (usr && isliving(usr) && !issilicon(usr) && BOUNDS_DIST(src, usr) == 0)
-			var/mob/living/L = usr
-			L.give_to(src)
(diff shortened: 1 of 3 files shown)


TDHooligan:
is it worth finding some better means to prevent this dialogue being used in situations it could be exploited? like for a few seconds after combat  actions.

pgmzeta:
> is it worth finding some better means to prevent this dialogue being used in situations it could be exploited? like for a few seconds after combat actions.

Could integrate it with the attack/next_click delay instead, do you think that's a better approach here?

Tarmunora (approved):

ZeWaka:
good for now can integrate later if need

PR #18435, opened 2024-04-07
Stun adjustments

TDHooligan:
[BALANCE][OBJECTS]
## About the PR 
For testmerge? maybe?? i am nearly certain this will miss some cases.

**While weakened (IE. Baton stuns, etc):** 
 - Brute & burn damage over (5 BURN minimum) reduces the duration of weakened:
   - This effect is roughly equivalent to a 30s stun being worth 50 health.
   - Stuns lower than 5s (IE. stamina stuns) are reduced less.
   - These effects no longer function if you are under the effect of paralysis (crit stuns or drugs)

## Why's this needed? 
Stun batons are too good at guaranteeing death in 3 touches. It's counterintuitive for security, and way too good for Traitors.

While this won't solve ALL deaths by baton, there's at least a requirement for an alternative kill method than just baton + beatdown.

## Changelog 

```changelog
(u)TDHooligan
(*)Stuns now get shortened by taking melee damage.
(*)Shorter stuns are affected less.
```


FlameArrow57:
Sounds nice! Feels like it will buff handcuff drop methods (robo arms, hulk, changeling, etc) a fair amount tho. Given someone with these who needs to be executed by any security officer, it feels it would almost be safer/easier to space/debrain/cremate/etc than damage them to death, since there’s a decent chance they could escape and/or disarm the attacker. 1 egun laser shot is 45 dmg

TobleroneSwordfish (changes requested):
I like the idea of this and I think it's at least worth testmerging but the stamina regen portion seems like overcomplicating it a bit, and is very counter-intuitive when hitting people with weapons usually drains stamina.

TDHooligan:
> I like the idea of this and I think it's at least worth testmerging but the stamina regen portion seems like overcomplicating it a bit, and is very counter-intuitive when hitting people with weapons usually drains stamina.

the stamina was added after testing the original change. i'm open to alternatives to stop toolboxing being stunbaton but worse

TDHooligan:
> Feels like it will buff handcuff drop methods (robo arms, hulk, changeling, etc) a fair amount tho. Given someone with these who needs to be executed by any security officer, it feels it would almost be safer/easier to space/debrain/cremate/etc than damage them to death, since there’s a decent chance they could escape and/or disarm the attacker

i think this is OK. we also have the port-a-brig for this reason. its not like you *can't* beat someone to death with a baton. it will just use up your whole baton doing so. which is less of a sec nerf and more of a traitor-stealing-baton nerf.

> 1 egun laser shot almost removes the entire stun

this is deliberate - it hits the 3s threshold where you can get one more hit in.

TDHooligan:
> Toolboxing should be stun baton but worse, it's one of the few ways inexperienced and unarmed crew can respond to threats in a meaningful way. Giving stamina back on hit will just make meth and other stun resist/stamina regen chems even harder to fight.

to visualize why i made the stamina changes, here is a video without them:

https://github.com/goonstation/goonstation/assets/6061314/40f17679-96e1-4002-8d38-a454df73fdb6

i've also uploaded a video to the PR to demonstrate how it works with stamina generation. I think getting 2-3 short opportunities to do things is OK.

as is, the current stamina implementation can be used in combination with the main stun reductions to adjust how fighting feels at low stamina (more hits per stamina stun, make it harder/easier to stun someone immediately after).

TobleroneSwordfish:
Okay, just for now can you cut it down to stun reduction only and we can see how that plays out? The stamina changes can be PRed separately if they seem to be needed.

code/mob/living/stamina_stuns.dm
```diff
@@ -91,6 +91,35 @@
...
+		var/stamina_regain_limit = 40+(stamina_damage/2) // 50 STA toolboxes leave you at 15 sta
+		var/stamina_regen_max = max(0,stamina_regain_limit-stamina)
+
+		//add_stamina(max(0,min(stamina_base + stamina_from_damage, stamina_regen_max))) //adrenaline? pain? stun baton hate?
+		*/
+		var/stun_duration = src.getStatusDuration("weakened")
+		if (stun_duration > 3 SECONDS) //if we have a big stun, we can kick it down a lot
+			var/stun_reduction = min(stun_duration-3 SECONDS,(brute + burn)*0.6 SECONDS) // let's saaay 1 full stun is 50 health.
+			var/stunres_penalty = (get_stun_resist_mod()/2)/100
+			src.setStatus("weakened", stun_duration-(stun_reduction*stunres_penalty))
```
TobleroneSwordfish, line 114: This is backwards, 0 stun resist = 0 revenge stun reduction

TobleroneSwordfish (changes requested):

code/mob/living/stamina_stuns.dm
TDHooligan: this is so that high % stunres reduces the effect of stun reduction (though not enough to be net negative. just enough that you arent getting exponential stun resistance)

TobleroneSwordfish: No like, as written if you don't have stun resist you won't get the revenge stamina effect at all

code/mob/living/stamina_stuns.dm
```diff
@@ -91,6 +91,35 @@
...
+		else if (stun_duration > 0 SECONDS)  // but we also still need a penalty for getting stamcrit which is 5s.
+			var/stun_reduction = 1 SECOND
+			var/stunres_penalty = (get_stun_resist_mod()/2)/100
+			src.setStatus("weakened", stun_duration-(stun_reduction*stunres_penalty))
+
+	if(src.stamina_bar?.last_update != TIME) src.stamina_bar?.update_value(src)
+	return
```
pgmzeta, line 121: unneeded return & clarity
```suggestion
	if(src.stamina_bar?.last_update != TIME)
		src.stamina_bar?.update_value(src)
```

code/mob/living/stamina_stuns.dm
```diff
@@ -91,6 +91,35 @@
 	if(src.stamina_bar?.last_update != TIME) src.stamina_bar?.update_value(src)
 	return
 
+/// Adds 'get up' stamina, from taking beatdown damage.
+/mob/proc/get_revenge_stamina(stamina_damage, brute, burn, damage_type )
+	return
+
+/mob/living/get_revenge_stamina(stamina_damage, brute, burn, damage_type )
+	if(!src.use_stamina) return
```
pgmzeta, line 99: call parent by default (aware it's currently a no-op)
```suggestion
/mob/living/get_revenge_stamina(stamina_damage, brute, burn, damage_type)
	. = ..()
	if(!src.use_stamina) return
```

pgmzeta (commented):
It would be nice to get a comparative graph of stamina & health changes before and after this change with a small sample of items covering common station weapons (e.g. toolbox, fire extinguisher, circular saw special-attack (is it effected?)).

TDHooligan:
will do a cleanup pass shortly - DNM in the interim

Sbmhawk:
I see people saying that "killing someone will end up using your entire baton."

If you have a baton, you probably have, or have access to cuffs

TDHooligan:
> I see people saying that "killing someone will end up using your entire baton."
> 
> If you have a baton, you probably have, or have access to cuffs

this has been stated several times (including by myself) - ultimately, making it so you need more than a baton opens up way more means to balance this down the line. IMO, this is a pro-change argument. all of these are much more adjustable than the alternatives.

code/mob/living/stamina_stuns.dm
```diff
@@ -91,6 +91,36 @@
...
+		else if (weak_duration > 0 SECONDS)  // but we also still need a penalty for getting stamcrit which is 5s.
+			src.setStatus("weakened", weak_duration-1)
+
+		var/stun_duration = src.getStatusDuration("stunned")
+		if (stun_duration > 3 SECONDS)
+			var/stun_reduction = min(stun_duration-3 SECONDS,(brute + burn)*0.6 SECONDS)
+			var/stunres_penalty = clamp(1-(get_stun_resist_mod()/2)/100,0,1)
+			src.setStatus("weakened", stun_duration-max(1,stun_reduction*stunres_penalty))
+		else if (stun_duration > 0 SECONDS)
+			src.setStatus("weakened", stun_duration-1)
```
TobleroneSwordfish, line 118:
```suggestion
```
I've gone ahead and removed the unneeded "stunned" portion of projectile stuns so this can be removed (it also wouldn't have worked due to bad copy paste)

code/mob/living/stamina_stuns.dm
TobleroneSwordfish: ^

code/mob/melee_attack_procs.dm
```diff
@@ -676,14 +676,16 @@
 			stam_power *= clamp(damage/pre_armor_damage, 1, 1/3)
 		stam_power *= max(0, (1-shield_amt["shield_strength"]))
 
-		//record the stamina damage to do
-		msgs.stamina_target -= max(stam_power, 0)
+		if (!target.hasStatus("weakened"))
+			//record the stamina damage to do
+			msgs.stamina_target -= max(stam_power, 0)
```
TobleroneSwordfish, line 681: ?

code/mob/living/stamina_stuns.dm
```diff
@@ -91,6 +91,36 @@
 	if(src.stamina_bar?.last_update != TIME) src.stamina_bar?.update_value(src)
 	return
 
+/// Adds 'get up' stamina/stun reduction, from taking beatdown damage.
+/mob/proc/get_revenge_stamina(stamina_damage, brute, burn, damage_type )
+	return
+
+/mob/living/get_revenge_stamina(stamina_damage, brute, burn, damage_type )
```
TobleroneSwordfish, line 98: This name doesn't really make sense anymore since it's not giving stamina, PR description also needs updating.

TobleroneSwordfish (changes requested):

code/mob/living/stamina_stuns.dm
TDHooligan: ick. my bad

code/mob/melee_attack_procs.dm
TDHooligan: punches use this weird way to track their stamina damage.

TobleroneSwordfish: Yees, but why would being weakened make you immune to stamina damage? This isn't mentioned in the changelog.

OmegaShoots:
An honestly baffling change that makes no sense and ruins combat.

PR #18436, opened 2024-04-07
Commander cards now keep access when emagged

JORJ949:
## About the PR 
Syndicate Operative Commander ID cards now keep their commander access after being emagged

## Why's this needed? 
How will they announce to the crew their intent to kill them all if they cant get in their room?

### DIFF
--- a/code/obj/item/card.dm
+++ b/code/obj/item/card.dm
@@ -281,7 +281,7 @@ TYPEINFO(/obj/item/card/emag)
 		all_accesses -= new_access
 		if (istype(src, /obj/item/card/id/syndicate)) // Nuke ops unable to exit their station (Convair880).
 			src.access += access_syndicate_shuttle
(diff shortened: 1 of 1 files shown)


code/obj/item/card.dm
```diff
@@ -281,6 +281,8 @@ TYPEINFO(/obj/item/card/emag)
 		all_accesses -= new_access
 		if (istype(src, /obj/item/card/id/syndicate)) // Nuke ops unable to exit their station (Convair880).
 			src.access += access_syndicate_shuttle
+			if (istype(src, /obj/item/card/id/syndicate/commander)) // Commander unable to play their cool tunes
+			src.access += access_syndicate_commander
 		DEBUG_MESSAGE("[get_access_desc(new_access)] added to [src]")
```
Tarmunora, line 286: indentation error

PR #18437, opened 2024-04-07
Karma affects *snap and *twirl fuckups

MintyPhresh:
## About the PR 
Accidentally dropping items you twirl will never happen if your karma is above 30. (below that point, default 5% chance to fail remains)
Snapping your fingers will never snap them off if your karma is above 0. (below that point, default 5% chance to fail remains)

## Why's this needed? 
Emotes are great for RP, but the fingersnap causing relatively grievous injury can disrupt flow and just... doesn't really add much. Twirl probably doesn't need this as badly as fingersnap, but having a nice little passive reward for people with good karma seems cool.


TDHooligan:
i dont really care for more karma tie ins. twirl drops are fine (but could maybe be disabled for security training) 
as a regular finger snapper it'd be nice for  the snaps to not do any damage but still have the message. i dunno.

Tarmunora:
no, not interested in adding this, especially as it seems intentionally designed for people to consistently farm karma roundstart for it.

PR #18446, opened 2024-04-09
Reimplements the KS-23 'Kuvalda' for Gangs.

TDHooligan:
[GAME OBJECTS][FEATURE]
## About the PR 
ks23

Adds the Kuvalda back with new, fresh mechanics and plenty of overkill. 
Can drop in large gang crates, with 4 rounds of buckshot and slug, at the same rarity as American-180s (very very low)

Features a single-loaded 4 shot magazine. A slow two-stage pump, heaps of recoil... 

https://github.com/goonstation/goonstation/assets/6061314/b3da091e-a1ad-40e2-9cc1-e7e282b46f43

(damage for slugs was upped after this video)

also creates a new basetype for pump shotguns, like single-action revolvers.

## Why's this needed? 
I had an idea, I made it. it was surprisingly fun to use. 
Definitely a fun admin toy. Once again, balanced around ammo scarcity (8 shells total if spawned in the crate) and even then, half of them are AP slugs which can't do nearly as much damage as the buckshot.

## Changelog 

```changelog
(u)TDHooligan, Blackrep
(+)The KS-23 'Kuvalda Carbine' (read: big shotgun) can now be found in gang weapon crates.
```


TDHooligan:
waiting for comments on this one.

EEASAS:
i like more weapons in the game, cool i guess

EEASAS:
now that i think of it, it really is a shotgun recycled from anti-aircraft shells, i think it should have organ/limb removal likes the spes-12

code/modules/projectiles/bullet.dm
```diff
@@ -738,6 +738,129 @@ toxic - poisons
...
+	shot_sound = 'sound/weapons/kuvalda.ogg'
+	dissipation_delay = 6
+	casing = /obj/item/casing/shotgun/gray
+	var/image/blood_image
+	var/bleeding = FALSE // if we're spitting blood
+	var/mob/living/unfortunate  // who was last hit, by this bullet
+	var/turf/last_projectile_loc //1:many data:projectile pairing
+	var/turf/last_turf
+	on_launch(obj/projectile/O)
+		O.AddComponent(/datum/component/gaseous_projectile)
```
Tarmunora, line 759: use of gaseous_projectile here feels incorrect - this will cause unexpected behaviour e.g.: bullets being blocked by atmos forcefields. Find or make a better-fitting solution

code/modules/projectiles/bullet.dm
```diff
@@ -738,6 +738,129 @@ toxic - poisons
...
+	hit_type = DAMAGE_STAB
+	dissipation_rate = 10
+	shot_sound = 'sound/weapons/kuvalda.ogg'
+	dissipation_delay = 6
+	casing = /obj/item/casing/shotgun/gray
+	var/image/blood_image
+	var/bleeding = FALSE // if we're spitting blood
+	var/mob/living/unfortunate  // who was last hit, by this bullet
+	var/turf/last_projectile_loc //1:many data:projectile pairing
+	var/turf/last_turf
```
Tarmunora, line 757: Tracking stuff like this on projectile datum gets a but weird, because the projectile datum is not re-initialized for each projectile - so it'll keep tracking the same mob and cause bleeding to them with each shot until a different mob is hit. Better to track per-projectile stuff in `special_data` on the projectile obj instead

code/modules/projectiles/bullet.dm
```diff
@@ -738,6 +738,129 @@ toxic - poisons
...
+	hit_type = DAMAGE_STAB
+	dissipation_rate = 4 //spread handles most of this
+	shot_sound = 'sound/weapons/kuvalda.ogg'
+	dissipation_delay = 6
+	casing = /obj/item/casing/shotgun/gray
+	var/image/blood_image
+	var/list/bleeding = list() // which bullets are bleeding
+	var/list/mob/living/unfortunates = list() // who was last hit, by each bullet
+	var/list/turf/last_projectile_locs = list() //1:many data:projectile pairing
+	var/turf/last_turf
```
Tarmunora, line 813: less important here because spreaders actually *do* reinitialize the projectile datum constantly, but still

Tarmunora (changes requested):

code/modules/projectiles/bullet.dm
TDHooligan: never heard of that. will look into it ta

code/datums/components/nonwall_pierce.dm
```diff
@@ -0,0 +1,24 @@
...
+
+
+/datum/component/nonwall_pierce/proc/update_pierces(var/obj/projectile/P, var/atom/hit)
+	var/turf/T = get_turf(hit)
+	if (isrestrictedz(T.z))
+		return 0
+	if (istype(hit, /obj/machinery/door))
+		var/obj/machinery/door/door = hit
+		return !door.density
+	return PROJ_ATOM_PASSTHROUGH * (!iswall(T) || (istype(hit, /obj/machinery/door)))
```
Tarmunora, line 20: Condition here feels weird. It won't pierce things if there's a wall on the turf (hitting the thing instead of the wall?)
And the door handling in the return line feels off

Tarmunora (changes requested):

code/datums/components/nonwall_pierce.dm
TDHooligan: ooer. you're right. lemme cover that

Tarmunora (approved):

Tarmunora:
repaths will probably need some secrets mapping changes

PR #18455, opened 2024-04-09
Compatibility with Aseprite DMI plugin

Cherman0:
## About the PR 

Removes the unused icon state "???" from `items.dmi`, as it causes a load failure with Seefaa's aseprite dmi compatibility script (https://github.com/Seefaaa/aseprite-dmi). Specifically, the script manages the icon states as png files in a temp directory, meaning that the icon states need to be valid filenames. Question marks are an illegal character in Windows files, which causes the error. 

This bug is reported on their end but is unlikely to be fixed as it is fundamentally a limitation of Windows. https://github.com/Seefaaa/aseprite-dmi/issues/28

## Why's this needed? 

Spriting tools compatibility and unused crust removal.


ZeWaka:
legendary lmao

PR #18459, opened 2024-04-10
Fixes Colour Fighting Between Command and Civilians

cheekybrdy:
[INPUT WANTED] [FEATURE]
## About the PR 
Changes where command is green colour wise to blue (this does include important items like IDs so input is wanted). And changes vice versa for civilians (also makes some grey references for civilians green).
## Why's this needed? 
Putting departments colours into a single form is good for representing them to players consistently, along with some smaller fluff to change their standards.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Makes command consistently blue and civilians consistently green. Be aware this effects common items like IDs!
(+)Added the civilian assistant, a department staffie equal for the Civilian department.
(+)Staff assistants now have a standard headset instead of a civilian one.
```


TobleroneSwordfish (changes requested):
This should be three different PRs, yes.

PR #18464, opened 2024-04-10
Adds a civilian assistant gimmick job

cheekybrdy:
[FEATURE]
## About the PR 
Adds a civilian equivalent to the department assistant gimmick jobs.

## Why's this needed? 
Even if this job only exists off a technicality, I still think it'd be a fun job and maybe make the civilian department's identity more concrete.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Adds a civilian assistant latejoin role
```


444Portal:
The access they have makes them feel like a hop-lite. Maaaybe if they were reflavoured as like a Service Manager or something but that still doesn't feel like its separate from HoP enough. Not a fan

PR #18465, opened 2024-04-10
Replaces civilian headsets from staffies and some gimmick roles and replaces with general ones.

cheekybrdy:
[INPUT WANTED] [CONTROVERSIAL]
## About the PR 
Replaces civilian headsets from staffies and some gimmick roles and replaces them with general radio headsets.

## Why's this needed? 
Staffies help all departments, not really just civvies (the least if anything from my experience). If the average joe doesn't have access to this frequency civvies that do have it may value it more.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Replaces staffies and a few gimmick jobs with standard radio headsets instead of civilian ones.
```


DisturbHerb:
Back when Staff Assistant didn't have a civilian headset, the channel was even less utilised (going off of my own fuzzy memory). I think returning to that might make us worse off.

444Portal:
I really really don't think we gain anything of value by removing this from staffies. The channel is not often used but I love the gossip and just casual chat that comes from it. It's a fun radio when it's in use and majority of that comes from a bunch of bored staffies talking about things

PR #18470, opened 2024-04-11
Change Killswitch access to Upload from Robotics

JORJ949:
## About the PR 
Changes the access requirement to killswitch silicons  to AI upload from robotics

## Why's this needed? 
1) Wiki said so (https://wiki.ss13.co/Killing_the_AI#Death_by_Killswitch)
2) Roboticists can killswitch but CE/RD cannot

### DIFF
--- a/code/obj/machinery/computer/robotics.dm
+++ b/code/obj/machinery/computer/robotics.dm
@@ -2,7 +2,7 @@
 	name = "Robotics Control"
 	icon = 'icons/obj/computer.dmi'
 	icon_state = "robotics"
-	req_access = list(access_robotics)
+	req_access = list(access_ai_upload)
 	object_flags = CAN_REPROGRAM_ACCESS | NO_GHOSTCRITTER
 	desc = "A computer that allows an authorized user to have an overview of the cyborgs on the station."
 	power_usage = 500


JORJ949:
> I think the wiki needs the change here, not the code.
> 
> Beyond that, the robotocist definetly should be able to killswitch the AI. Working with AI/Borgs is their primary task.

I whole heartedly disagree, roboticists do not have access to upload whereas command does, a roboticists job is not to manage a rogue silicon or their laws, hence they do not have access.

Sbmhawk:
Robos shouldn't be expected to killswitch, when they don't even have access to the killswitch console without hacking.

CaioFrancisco:
I kinda agree with Earthfire, but as Goonstation is today, the robotics console requiring Roboticist access feels very much outdated, or at least, out of touch with how it is implemented currently. In my opinion, it deserves a rework of some kind, for it to actually become a console for managing robots besides being a heavily AI-guarded killswitch, and for Roboticists to have more access to it. Of course, this is not in the scope of this PR, but maybe some coder/dev and silicon nerd at heart may want to take a tackle at this idea.

444Portal:
my thoughts are command should be the ones dealing with AI related stuff, be it laws, killswitch, lockdown, etc, etc. I guess it'd be fitting for robotics to handle it but at the same time one could argue each command member has some link to the AI and it's construction (albeit flimsy). MD manages the brain, RD the underlying 'tech' behind it all, CE the construction and HoS the AI upload's turrets, shocked grilles and the HoP/Cap on the Laws. I don't really know anything else to say other than 'it just seems like commands thing, not robotics'

JORJ949:
> My previous comment caused a bit of confusion, i don't think it's an either/or-kind of thing. What i think is that both, robotics and command, should have the access to this console.
> 
> So instead of replacing robotics access with command, make command additionally to robotics access thr console.
> 
> This way, like it currently does, the robotocist can still be a last-resort if the AI actually managed to kill all heads.

Would still feel weird for robotics to not have access to reach the killswitch but have access to activate it, if you want roboticists to have more interaction with silicons, give them upload. also you mention roboticisits being a "last resort" if the AI is rogue, but in almost all of those cases the killswitch is heavily guarded, if not destroyed, and more reachable by a command member like the RD and CE with telesci and their RCD respectively

Wisemonster:
If the killswitch console was in the robotics department, this change would make more sense, however as is it's really weird for them to have access to a console they can't even reach, unless the idea is for an command member that doesn't have console access to escort a roboticist to the console to killswitch them, even then it'd make more sense to just get a member of command that does have access already to just do it themselves.

JORJ949:
To me the separation is as follows:
Roboticists job is to create and upkeep the silicons themselves
Command keeps them in line with laws and killswitch

Glamurio:
Clearly this calls for the addition of a Chief Roboticist.

pgmzeta:
50% increase in the amount of people who have the access needed to single-handedly killswitch the AI/any cyborg.
Current: Cap, HoP, HoS, MD 
New: CE, RD

Currently the CE & RD would have to convince a roboticist (or secoff on RP) to assist.

JORJ949:
> 50% increase in the amount of people who have the access needed to single-handedly killswitch the AI/any cyborg.
> Current: Cap, HoP, HoS, MD 
> New: CE, RD
> 
> Currently the CE & RD would have to convince a roboticist (or secoff on RP) to assist.
> 
> 

Considering how rare killswitches are (and  classic HoPs and HoSs for that matter) and that there's easy counterplay to it (destroying it as soon as you go rogue) i see no reason not to. (Also thanks for adding tags, PRd this early in the morning and had to rush off after so forgot)

PR #18482, opened 2024-04-13
Jobs support multiple implants on spawn

pgmzeta:
[internal][feature]
## About the PR 
Instead of a single implant, jobs now use a list to track which implants to give on spawn.

I've written a naive approach to support existing saved jobs, but I have not been able to successfully save both with and without this change. Guidance/admin testing appreciated.

## Why's this needed? 
Atomized out support for multiple implants, needed for a future PR to split anti-mindhack and health implants (then giving HoS/SecOffs both). Admins could use this to spawn jobs with multiple implants.

This PR **does not** make any changes to implants, only provides support for future changes / job creator use.


Tarmunora (approved):

code/datums/jobs.dm
```diff
@@ -128,16 +128,18 @@
 		M.faction |= src.faction
 
 		SPAWN(0)
```
frawhst, line 130: Does anything break if this `SPAWN` isn't here? I vaguely remember implants being one of the things that did need that, but I can't remember for sure.

frawhst (approved):
lgtm generally

code/datums/jobs.dm
pgmzeta: SPAWN covers all these, will test o7
- [ ] regular implants
- [ ] sec clone disk
- [ ] mobcritter implants
- [ ] latejoin radio host (test `special_spawn_location` / custom name)
- [ ] bioeffect/mutrace

pgmzeta: This works without the spawn with local testing. Unsure if this might be a problem with the large number of round-start uses at one time?

PR #18491, opened 2024-04-13
Add new negative traits to character creation

Glamurio:
[Traits][Feature]

## About the PR 
This PR adds 4 new traits with negative effects to the game.
- Unstable Genes (+1)
- - More likely to mutate from radiation.
- Soft Swatter (+1)
- - Weaker punches.
- Slow Strider (+2)
- - You're unable to sprint.
- Trippy (+1)
- - You randomly trip when moving.

## Why's this needed? 
When creating a new character, I find negative flaws as important as positive ones. This creates new ways for people to customize their characters without gaining an inherent advantage. I've also put 3 out of 4 traits to exclusive categories to encourage variety.

If needed, I can atomize the PR.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Adds four new negative traits to character creation.
```


code/mob/input.dm
```diff
@@ -202,6 +202,12 @@
 						last_move_trigger = ticker ? ticker.round_elapsed_ticks : 0 //Wire note: Fix for Cannot read null.round_elapsed_ticks
 						deliver_move_trigger(running ? "sprint" : m_intent)
 
+					// Tripping (the physical kind)
+					var/trip_chance = 5
+					if (src.traitHolder && src.traitHolder.hasTrait("trippy") && prob(trip_chance))
+						src.setStatus("resting", INFINITE_STATUS)
+						src.force_laydown_standup()
+						src.visible_message(SPAN_ALERT("<B>[src]</B> trips!"))
```
Tarmunora, line 210: isn't there a "trip" emote (for humans at least) already? Might be worth considering

Tarmunora (approved):

code/mob/input.dm
```diff
@@ -202,6 +202,12 @@
 						last_move_trigger = ticker ? ticker.round_elapsed_ticks : 0 //Wire note: Fix for Cannot read null.round_elapsed_ticks
 						deliver_move_trigger(running ? "sprint" : m_intent)
 
+					// Tripping (the physical kind)
+					var/trip_chance = 5
```
Zonespace27, line 206: an alternative implementation could be to add a signal (`COMSIG_MOB_MOVE` or something) that's called here, and add tripping as a component that hooks into such a signal. This would be better because it would avoid the need to check traitholder on every single step of every single mob's `process_move`.

Glamurio: In that case we should also fix the "Two Left Feet" trait, because that does it even more frequently.

I've personally no idea how to work with signals, however, and feel like I'd do a shoddy job, but the idea behind it seems better.

code/mob/input.dm
Glamurio: Yeah, but that one force paralyses for 3 seconds and is pretty much only used for `foreign_limb_effect()`, seemed excessive.

Valtsu0:
Unstable genes and soft swatter have too small of a downside to be +1 imo

Glamurio:
> Unstable genes and soft swatter have too small of a downside to be +1 imo
In terms of percentage or effect? I can crank up the percentages. I used Addictive Personality as a blueprint for the Unstable Genes which I found fairly even. A lot of the +1 effects are very situational, like Hemophilia or Medical Allergy. There's a good chance you're never in a situation where it matters, but if it does matter, you notice it.

TobleroneSwordfish:
I like Slow Strider and Trippy, but the other two do seem just a bit too situational. Hemophilia is actually really nasty when you get cut (which is pretty often on classic) and medical allergy can make you very hard to heal sometimes. In contrast most people never punch outside of hulk and boxing rings, and it would be near impossible to actually tell the difference with Unstable Genes.

Glamurio:
> I like Slow Strider and Trippy, but the other two do seem just a bit too situational. Hemophilia is actually really nasty when you get cut (which is pretty often on classic) and medical allergy can make you very hard to heal sometimes. In contrast most people never punch outside of hulk and boxing rings, and it would be near impossible to actually tell the difference with Unstable Genes.

Alright, I'm not that attached. Let me cull those two for now and workshop them in a separate PR

PR #18499, opened 2024-04-13
Derived harmonious colors helper procs

Sovexe:
## About the PR 
* Helper procs for choosing various harmonious colors based on several of the major color schemes using hue offsets in the HSL colors space

image

## Why's this needed? 
* It's hard to pick colors that look nice together at random. Starting from a random base however we can pick aesthetically pleasing harmonious colors with ease using these helper procs.


Azrun:
Consider points between A and B for additional complimentary colors.

Azrun (approved):

PR #18505, opened 2024-04-14
PTL renumber

TobleroneSwordfish:
[BALANCE] [GAME OBJECTS]
## About the PR 
Lowers the thresholds for Dangerous PTL effects and scales the money generation function down so that decent amounts of money can be made at much more reasonable power levels.

The PTL will now blow off limbs at 5MW, gib at 200MW and explosive gib at 5GW.
The power generation function is a bit more complicated but should now generate around twice as much at lower levels and scale up a lot faster.

## Why's this needed? 
The PTL scaling was written for old hellburns and has not been adjusted for new TEG/singulo/nuke engine numbers.

## Changelog 

```changelog
(u)LeahTheTech
(*)The PTL formulae have been rescaled to make it more profitable and dangerous at lower power levels
```


code/modules/power/pt_laser.dm
```diff
@@ -465,25 +465,25 @@
 
 	//this will probably need fiddling with, hard to decide on reasonable values
 	switch(power)
-		if(10 to 1e7)
+		if(10 to 10 MEGA WATTS)
```
pgmzeta, line 468: should this be to 5 MEGA WATTS?

TobleroneSwordfish: Good catch, you really would think the compiler should be able to catch this.

Azrun:
Estimated impact when compared to previous rounds with the assumption that SMES would be powered first.
image
image

PR #18506, opened 2024-04-14
Monkeys now throw darts at balloons

Valtsu0:
[AI][FEATURE]
## About the PR 
Monkeys not doing anything will attempt to pickup darts (specifically bar darts) and throw them at balloons. Darts now pop balloons

Mostly just trying to learn how mob ai stuff works so any feedback is greatly appreciated.

## Why's this needed? 
000-DartMonkey 1

### DIFF
--- a/code/mob/living/carbon/human/monkeys.dm
+++ b/code/mob/living/carbon/human/monkeys.dm
@@ -2,6 +2,7 @@
 #define IS_NPC_HATED_ITEM(x) ( \
 		istype(x, /obj/item/handcuffs) || \
 		istype(x, /obj/item/device/radio/electropack) || \
+		istype(x, /obj/item/reagent_containers/balloon) || \
 		x:block_vision \
 	)
 
@@ -229,6 +230,7 @@
 	var/list/shitlist = list()
 	var/ai_aggression_timeout = 600
 	var/ai_poke_thing_chance = 1
+	var/ai_delay_move = FALSE
 	default_mutantrace = /datum/mutantrace/monkey
 
 	New()
@@ -258,6 +260,15 @@
 			return
 		..()
 		if (src.ai_state == 0)
+			if (src.equipped() && istype(src.equipped(),/obj/item/implant/projectile/body_visible/dart/bardart))
+				for (var/obj/item/reagent_containers/balloon/balloon in view(7, src))
+					src.throw_item(balloon, list("npc_throw"))
+					src.ai_delay_move = TRUE
+					break
+			if (!src.equipped() && !src.ai_delay_move)
+				for (var/obj/item/implant/projectile/body_visible/dart/bardart/dart in view(1, src))
+					src.hand_attack(dart)
+					break
 			if (prob(50))
 				src.ai_pickpocket(priority_only=prob(80))
 			else if (prob(50))
@@ -495,6 +506,9 @@
 
 	ai_move()
 		if(src.ai_picking_pocket)
+			return
+		if(src.ai_delay_move)
+			src.ai_delay_move = FALSE
 			return
 		. = ..()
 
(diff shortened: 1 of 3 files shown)


code/mob/living/carbon/human/monkeys.dm
```diff
@@ -258,6 +260,15 @@
 			return
 		..()
 		if (src.ai_state == 0)
+			if (src.equipped() && istype(src.equipped(),/obj/item/implant/projectile/body_visible/dart/bardart))
```
amylizzle, line 263:
```suggestion
			if (istype(src.equipped(),/obj/item/implant/projectile/body_visible/dart/bardart))
```
don't need to check it twice, `istype` will return false on `null`

code/mob/living/carbon/human/monkeys.dm
```diff
@@ -258,6 +260,15 @@
...
+				for (var/obj/item/reagent_containers/balloon/balloon in view(7, src))
+					src.throw_item(balloon, list("npc_throw"))
+					src.ai_delay_move = TRUE
+					break
+			if (!src.equipped() && !src.ai_delay_move)
```
amylizzle, line 268:
```suggestion
			else if (!src.ai_delay_move && !src.equipped())
```
checking a bool is faster than running a proc, so it goes first. Also throw an else on there so the monkey can't throw a dart and pick up another in the same tick.

amylizzle (changes requested):
Some minor optimisations. Funny monke.

TobleroneSwordfish:
Buut darts don't pop the balloons!

code/mob/living/carbon/human/monkeys.dm
amylizzle: you missed this one

Valtsu0: no idea why i though i fixed it

Valtsu0:
Made the darts pop balloons

code/obj/item/implant.dm
```diff
@@ -2143,14 +2143,41 @@ TYPEINFO(/obj/item/gun/implanter)
...
+					. = TRUE
+				if(isobj(A))
+					if(!A.Cross(src) || istype(A, /obj/item/reagent_containers/balloon)) // darts can hit balloons too
+						src.throw_impact(A, thr)
+						. = TRUE
+				if(ishuman(src) && istype(A, /obj/item/plant/tumbling_creeper))
+					var/obj/item/plant/tumbling_creeper/M = A
+					if(M.armed)
+						src.throw_impact(M, thr)
+						. = TRUE
```
TobleroneSwordfish, line 2179: Don't just copy paste the entire `/atom/movable/hit_check` here. A better way would be to implement `Cross` on balloons to deny thrown darts, that way the parent would catch it anyway.

TobleroneSwordfish (changes requested):

code/modules/chemistry/tools/balloons.dm
```diff
@@ -234,6 +234,9 @@
 		..()
 		src.smash(T)
 
+	Cross(atom/movable/mover)
+		return !istype(mover, /obj/item/implant/projectile/body_visible/dart/bardart)
```
TobleroneSwordfish, line 238:
```suggestion
		if (istype(mover, /obj/item/implant/projectile/body_visible/dart/bardart))
			return TRUE
		return ..()
```
Cross should call parent here.

TobleroneSwordfish (changes requested):
Oh you'll also need to add `pass_unstable = TRUE` to the definition of `/obj/item/reagent_containers/balloon` to allow it to implement `Cross` and pass unit tests. (complicated pathfinding optimization reasons)

ZeWaka (approved):

PR #18511, opened 2024-04-15
Rework PTL money allocation

TobleroneSwordfish:
[REWORK] [GAME OBJECTS]
## About the PR 
Changes the PTL to store up all its earned credits in an internal "account" that can then be withdrawn from by anyone with engineering (power) access. All withdrawals send a PDA message to the engineering mailgroup stating who withdrew and how much.
Since the emag effect depended on draining money from everyone's accounts it has been changed to just make the PTL fire backwards instead. This allows for some sabotage, but is fixable with PTL mirrors if the engineers know what they're doing.

## Why's this needed? 
The current system is very passive and unnoticeable, quietly splitting the money between all engineers on shift so that each person ends up with a uselessly small amount. This change will hopefully make it a bit more engaging and practical to use PTL funds and allows for more opportunities for Crime and Drama.

## Changelog 

```changelog
(u)LeahTheTech
(*)The PTL now stores up all of the money it makes, which can be withdrawn by swiping an engineering ID.
(+)The PTL now has a new emag effect.
```

### DIFF
--- a/code/modules/power/pt_laser.dm
+++ b/code/modules/power/pt_laser.dm
@@ -36,7 +36,6 @@
 	var/is_charging = FALSE //for tgui readout
 	///A list of all laser segments from this PTL that reached the edge of the z-level
 	var/list/selling_lasers = list()
-	var/wireless_net_id = null
 
 /obj/machinery/power/pt_laser/New()
 	..()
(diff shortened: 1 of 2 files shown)


mordent-goonstation (approved):
tgui changes fine, not reviewed anything else

TobleroneSwordfish:
Department budgets would be neat, but engineering doesn't currently have one and wouldn't be able to access it if they did so I'm going to leave it like this for now. I've done the negative number on the UI thought because funny.

PR #18516, opened 2024-04-15
Remaps the entirety of Clarion

Chatauscours:
## About the PR 
Remaps the entirety of Clarion, I guess.
Science, medbay, bar and most of command have been left untouched.
Sorry to whoever is reviewing all of those changes.

All of the bot beacons (sec and tour) have been adjusted accordingly.

## Why's this needed? 
Main issues:
- North is not visited and useless
- Crime is near impossible
- The engine lacks space around it for the engineers to work comfortable

Solutions:
--- Not visited North ---
- Moved the QM and podbay there, as well as the Snip and some showers.

--- Engine ---
- Just moved the engine around a bit.

--- Very difficult crime ---
- Addedd a considereable amounts of maintance that can be used as shortcuts in some cases.
- Added some random rooms, for flavour.
- Moved security to a more isolated area of the ship, so they don't hear everything by just sitting in their department, and can't respond to everything in 10 seconds.

## Changelog 

```changelog
(u)Chatauscours
(*)NT accumulated enough budget for the most of the Clarion ship to be rebuilt.
```


TheTyrant25 (commented):
I like this a lot. Particularly like the new placements of cargo and sec. Fabulous work. couple things though:
- seems like there's a normal wall leading to the showers in the northern minibrig. that or it's a false wall. both seem not good.
- Have you double checked that all the mail system pipes still work?
- mining no longer seems to have a real airlock leading to the mineral magnet
- not sure how i feel about the reactor grid line running through space to the right SMES units where it can be easily cut. I guess the left one can cover it but still.
- PTL room seems oddly empty. maybe add some caution tape or something?
- the 5x3 random room under the bridge probably doesn't need reinforced walls on all four sides
- northern wall of the CE's office probably should be reinforced
- a few spots where apparently incorrect lattices are being used. Not sure if they are auto lattice spawners or not.

Made these comments by looking primarily at the `new` render on mapdiffbot, so i'm not comparing it to the old version so to speak, just looking at how it is now. Great work.

TobleroneSwordfish:
Looks cool, the north side of clarion being nothing but copy-pasted bedrooms was never great 👍 
The one thing I can see at a glance is there's a lot of long one-wide maint tunnels: 
image
which tend to be hell to actually fight/do anything in

JORJ949:
Few issues I noticed:
- Escape checkpoint brig has no bed or toilet (On top of the wall tyrant mentioned)
- The area around the sec vendor is very cramped
- Portabrig remote should probably be in interro with the portabrig itself, along with the portabrig instructions
- Genpop brig has standard walls to interro
- The corridor to upload was cramped before, now it will be even worse as it got longer
- The door to reach the AI should probably be on the opposite side to the door to prevent people just running in ignoring the turrets.
- There appears to be a misplaced locker in the middle of hydroponics office
- To be consistent with other armories, consider putting the plasmaglass on the outside of the wall and make use of trip lasers.
- Armory turrets are really close to mining, consider moving the magnet further away so any miners who forgot a jetpack and slip dont get stunned to the debris field.

Chatauscours:
So:
- Changed the wall in north minibrig.
- Added a double airlock to mining
- Decorated the PTL a bit
- unrefinroced the bridge 5*3 random room
- fixed some lattices
- Unclutered a few of maints to make them 2 tiles or wider

----
- Some armouries on other maps have the plasmaglass on the interior, and I thought it would look neat here. Don't want to add a trip wire at all though, this armoury is well protected enough as is.
- Armoury turrets won't hurt miners, since they can walk around the outer lattice without getting shot.
- Not a fan of the cable there myself, but it's either that or make it go through the hallway and security, which is even worse. Here at least the turrets will protect it a bit.

amylizzle:
This looks like a great improvement. I only really looked intensely at the engine, but the general changes look good too.

My only comment on the engine is that the vent dump should probably come off the hot loop, not the cold loop. Maybe put a T-junction before the L-bend that goes into the turbine, and feed it from that.

Also maybe just try pressurising that because at one point there was a bug where the turbine wouldn't connect to non-straight pipes and I don't remember if I fixed it or not.

Sord213 (changes requested):
Overall these seem like some great changes. While the quarters were nice for lore reasons they weren't great for gameplay, so the changes here are definitely an improvement. I do have some requests, though.
Screenshot_947
I do not like the wire running through space at all. Looks weird. I'd rather it go through the main hallway or sec or something. 
Also the main equipment room (blue) is tiiiiny. Maybe move the hangar down a tile and use that tile to make it one tile larger?
The positioning of the sec hangar, beacon, and armory turrets seem conflicting and hard to maneuver around. Don't forget pod beacons are dense.
Screenshot_948
This main hallway is bland. Maybe some windows along the highlighted area. Maybe also a table with windoor for "interaction" (it'll never get used, most likely, but it'll look better) into the upper control room? idk.
Screenshot_949
Red highlights, 1 tile wide maint makes me sad.
I think I'd rather see the AI core turned into ai upload. And AI core placed in the current upload area in bridge. Also it definitely needs an "airlock" set of doors, not just one door into a hallway.
The yellow area seems....unnecessary? Maybe just fit that stuff into the crew lounge to the left? Use the space to improve maint and AI core/upload I'd say.
Screenshot_953
Swap core/upload probably. I think moving upload into an easier to crime spot would be better than DIRECTLY next to sec and the bridge.
Screenshot_952
i am cri. one tile wide maints feel bad and make me sad. I know clarion already has some on the map, and I'm not saying you need to change those, but I'd like to avoid adding more.
Screenshot_950
Replace window with a table/windoor pls
Screenshot_954
I know this isn't a change you've made, but it's one that's bugged me for a while. I'd really like to see eng gas storage be secured a _liiiittle_ bit more than just a single door from a main hallway.
Screenshot_955
pod bay could definitely use a ship parts manufacturer.

Chatauscours:
Right. Implement pretty much all of the suggestions, inculding the engineering storage.
Also removed some other 1 tile maints that were here before, as a treat.

JORJ949:
AI core feels really empty now, especially without any transduction matrices, consider adding some along with some computers, would also like to see the door to the core itself moved around the side opposite the entrance so you cant just run in and out before the turrets shoot.

Chatauscours:
Added more computers to the AI. changed the access for the computer core to go to the hallway and added a turret there, since I wasn't fan of the AI core being run through.
The robotics control was also moved into the AI core, for the AI not to have to defend that many location when it goes rogue.

444Portal (commented):
looks great! might want to change the thindows in the pool and funeral chapel room to something else if its possible, as they are meant to be phased out (despite how used they are elsewhere x.x)

Sord213 (changes requested):
Changes all look great! Got a couple more things and we should be good.
Screenshot_962
Move the weapon racks so if someone RCDs the wall behind them they can't access both at once in complete safety. Suggested placement is the arrows, but feel free to move them somewhere else in the room.
I also noticed armory doesn't seem to have any interior protection like a narc or energy turret. Could definitely use one.

Screenshot_963
Window w/ blinds anywhere facing into sec would be nice. Just to let other seccies know that it's occupied if the blinds are closed.

Screenshot_964
A secondary exit out of cryo into maint anywhere on the highlighted walls would be nice. Just in case the hallway above is blown up or something.

Screenshot_965
I like to avoid buttons like this on windows where possible. Could you move it up to the wall? Also maybe add an extra conveyor on the yellow highlight to make it a hair easier to shove crates out.

Screenshot_966
Some blinds in the crew lounge would be really nice if there aren't already some. And make the doors opaque instead of glass for privacy.

JORJ949:
I dont think the NARCS show up on this image but this armory looks copy pasted from where it was before so i assume its at the bottom

JORJ949:
Few more things I spotted:
image
This bottom left door doesnt appear to have an access spawner
image
This brig could do with a bed (heavy sleeper) and toilet
image
HoS could probably do with a carpet as the rest of the head of staff offices have them and plain wood looks well, plain
image
Consider moving the pipe and cable to seperate tiles as repairing the pipe/cutting the cable could be difficult
image
Chicken-Kitchen chute needs relinking to ranch
image
Probably worth moving loafer left a tile or 2 so you dont have to break down a wall to mess with it
image
Probably not something you changed but these pipes are also cut off

Chatauscours:
Integrated all of the proposed changes.
Narcs as Jorj said is already in the armoury, just the mapdiff doesn't show it for some reason.

Sord213 (changes requested):
Screenshot_967
Pipe junction facing wrong direction

Screenshot_969
Redirect the one pipe in the indicated way instead of going through a wall.
+ remove old unneeded junction

Screenshot_970
Blue wire does not connect to the engine loop like this. Nubs can only connect to other nubs.

Screenshot_971
While this technically works, please use double nubs to connect wires like this. Makes snipping wires....weird if done like this.

Screenshot_972
Screenshot_974
Screenshot_975
Screenshot_977
Same as above

Screenshot_973
All this wiring needs fixing and cleaning up. Also please electrify the windows around the AI itself.

Screenshot_976
Move this APC out into the security brig area. See Donut 2's brig for how to make an APC work in an area it is not physically in

JORJ949:
In addition about AI core I dont think all of those computers need wiring into the mainframe, its worth checking which do and dont

JORJ949:
image
These should probably be air tanks rather than oxy
image
I'd swap the locker and fuel tank around so the fuel tank can be dragged off

Chatauscours:
All of the suggestions are in, except brig firedoors, since it seems to be a standart to not have those on brig doors.

PR #18521, opened 2024-04-15
Airlocks with the ID wire cut will deny access to anyone who bumps to it

CaioFrancisco:
[FEATURE] [STATION SYSTEMS]
## About the PR 

This PR brings back the behavior that was patched out as a bug a little over a year ago. Airlocks with the ID wire cut will deny anyone bumping to the airlock, but clicking on the airlock lets anyone open it.

## Why's this needed? 

As airlocks currently stand, there's no simple way to figure out if it was hijacked or not. This completely inutilizes the concept of accesses, since if you figure out the ID wire, it doesn't matter if it has access restrictions, or if someone wants to keep you out: you simply walk in. This change not only makes it easier to figure out if a door was hijacked, it also presents an active annoyance to fix the airlock if it is hijacked.

If this change proves to be controversial enough, I can slap the Input Wanted tag on it.

3 dislikes, 1 like, controversial enough for a forums post. [INPUT WANTED]
(also one of the dislikes came in sooner than a single minute of the PR going live. wow.)
## Changelog 

```changelog
(u)Caio029
(*) Due to a recent re-design of airlock circuitry, airlocks with a cut ID wire will not respond to any bumps, but pressing it will open or close it unconditionally.
```

### DIFF
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -580,6 +580,7 @@ var/global/list/cycling_airlocks = list()
 /obj/machinery/door/airlock/bumpopen(atom/movable/AM)
 	if (src.isWireCut(AIRLOCK_WIRE_IDSCAN))
 		src.play_deny() // intentional nerf: idwire-cut doors won't respond to bumps. incentivizes people to fix it and makes them aware it is access hacked.
+		src.add_fingerprint(AM)
 		return
 	if (issilicon(AM))
 		if (!src.cyborgBumpAccess)
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -578,7 +578,7 @@ var/global/list/cycling_airlocks = list()
 				return
 	..()
 /obj/machinery/door/airlock/bumpopen(atom/movable/AM)
-	if (src.isWireCut(AIRLOCK_WIRE_IDSCAN))
+	if (!src.requiresID()) // if the ID wire has been tampered with.
 		src.play_deny() // intentional nerf: idwire-cut doors won't respond to bumps. incentivizes people to fix it and makes them aware it is access hacked.
 		src.add_fingerprint(AM)
 		return
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -578,6 +578,9 @@ var/global/list/cycling_airlocks = list()
 				return
 	..()
 /obj/machinery/door/airlock/bumpopen(atom/movable/AM)
+	if (src.isWireCut(AIRLOCK_WIRE_IDSCAN))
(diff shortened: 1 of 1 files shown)


code/obj/machinery/door/airlock.dm
```diff
@@ -578,6 +578,9 @@ var/global/list/cycling_airlocks = list()
 				return
 	..()
 /obj/machinery/door/airlock/bumpopen(atom/movable/AM)
+	if (src.isWireCut(AIRLOCK_WIRE_IDSCAN))
+		src.play_deny() // intentional nerf: idwire-cut doors won't respond to bumps. incentivizes people to fix it and makes them aware it is access hacked.
+		return
```
Tarmunora, line 583: feels a bit odd to handle this here with an early return that doesn't call parent, because this doesn't add fingerprints for example, and looks like it wouldn't respect the cooldown on the deny sfx the parent proc sets. Is there a better way to handle this?

Tarmunora (changes requested):

code/obj/machinery/door/airlock.dm
CaioFrancisco: honestly, not really. i can't move the logic to door parent, this is a nerf exclusive to airlocks. the only other ways i can think of doing is either adding a new `should_open` var to the parent's proc, and changing values on `src` so the conditions in the parent's proc fails, both solutions very janky to me.

DimWhat:
I feel like you could add some sort of flashing light or something on airlocks if they have had their ID wire cut instead. I'd assume that'd be quite annoying to make for every kind of door though.

CaioFrancisco:
not really. just have it flash green every couple of seconds and it'd work. although, i like my take on it, because it adds some urgency to fixing the door and keeping it fixed.

on an unrelated note, maybe for another PR if this one gets merged, silicons can disable the ID wire without cutting it, and this code only checks for cut wires, which means that with this PR merged, AIs can disable acceses on doors very easily. maybe this could be implemented for that scenario?

code/obj/machinery/door/airlock.dm
Tarmunora: Mmm
Janky turtles all the way down, I guess

Tarmunora:
> not really. just have it flash green every couple of seconds and it'd work. although, i like my take on it, because it adds some urgency to fixing the door and keeping it fixed.
> 
> on an unrelated note, maybe for another PR if this one gets merged, silicons can disable the ID wire without cutting it, and this code only checks for cut wires, which means that with this PR merged, AIs can disable acceses on doors very easily. maybe this could be implemented for that scenario?

imo behaviour should be the same for both cut wire and AI-disabling

CaioFrancisco:
> imo behaviour should be the same for both cut wire and AI-disabling

alright, i committed that in. if the other devs don't agree, you can revert the commit.

PR #18523, opened 2024-04-16
Atlas Cafeteria remodel

MintyPhresh:
## About the PR 
Trades the unnecessarily massive shared backroom of Atlas' catering department for a larger Kitchen, Bar, and crew area.
image

## Why's this needed? 
This PR aims to make the Atlas bar less cramped, gives it better equipment, more countertop area, and make it look nicer. More useful space is good, I think. Kitchen crimeability is also massively improved over the previous, aquarium-like kitchen.


MintyPhresh:
Addressing some discrepancies that Map Diff found now.

IPling:
I like this a lot over the current layout of Atlas's kitchen, but I do have some suggestions:

1. I feel like you can remove one of the maintenance doors while you are re-designing the kitchen, since having two maint doors that lead to the same place that close serves no purpose.

2. I don't think the bar needs a sink built right into the table, the bartender can already dispense water in his dispensers or just walk into the kitchen to use its sinks.

3. The sinks, again, the ones in the kitchen specifically. I don't think the chef needs 2, I also dislike the placement of being right infront of a door or behind a locker. If you remove the top-most maint door you can place a sink right next to the kitchen vendor.

MintyPhresh:
Thanks for the feedback! I wanted to get rid of that second door so bad but I was sure it was intentional; I'm looking forward to seeing what other possibilities are in store now that I have the confidence needed to remove it. Should be able to address the sink feedback too.

I'd like to keep the bar sink, though. Real bars have sinks as a place to dump ice or rinse out/clean cups, and although SS13 doesn't need those, it's a nice little thing for immersion IMO.

Gannets:
Oh that's way better actually.

Gannets (approved):

MintyPhresh:
OH NO

MintyPhresh:
I think everything is fixed now... sorry guys

PR #18526, opened 2024-04-16
Click-drag conveyor switch to change direction

pgmzeta:
[feature][qol][game objects]
## About the PR 
Add a `click_drag` implementation for conveyor switches.
Reduce duplication by moving some shared logic into procs.

## Why's this needed? 
Easier to set the direction you want. Feels like you're really moving the lever. 

## Changelog 

```changelog
(u)glowbold
(+)Click-dragging a conveyor switch to the left or right will change its direction to match.
```


Tarmunora (approved):

TobleroneSwordfish (changes requested):
Currently click dragging connects switches to conveyors, would be nice if that still worked here.

PR #18532, opened 2024-04-16
Improve notes commands

FlameArrow57:
[PLAYER ACTIONS][QoL]

## About the PR
Improves some things about the notes commands players can use in the Commands tab

-Renames 
"Add Note" -> "Notes - Modify"
"Notes" -> "Notes - View"
so that they appear next to each other rather than in different spots (since names are sorted alphabetically)

-Improves Add Note command so that instead of only being able to add individual notes one at a time, your notes are just stored in a single string, changed through a big text box, and can be added/removed
notes 2

-Uses TGUI prompts for the commands now so that "Notes - Modify" uses a tgui_input_text and "Notes - View" uses a tgui_message
notes 3

## Why's this needed?
QoL improvements and TGUI prompts

## Changelog

```changelog
(u)FlameArrow57
(+)Renamed notes commands in the Commands tab to "Add Note" -> "Notes - Modify" and "Notes" -> "Notes - View". "Notes - Modify" now allows you to add/remove notes all at once, rather than just adding one note at a time.
```


code/datums/traits_disused.dm
```diff
@@ -138,6 +138,7 @@
 	points = -1
 	unselectable = 1
 
+/*
```
pgmzeta, line 141: unrelated change?

FlameArrow57: These use the add_memory() proc which has been changed. Code here is unused, so it just needs to be revisited in the future if updating

TobleroneSwordfish: Better to either fix the code or remove it, don't just block comment stuff.

pgmzeta: there is a proc for dynamic memories used for conspirators, if you want to Do The Needful for this, but considering these are unused, ehhh?

Valtsu0:
I'd rather these be combined into a single verb/ui.

TobleroneSwordfish (approved):
lgtm

PR #18550, opened 2024-04-18
Move Donut3 AI Upload to Southeast hall

JORJ949:
[Mapping] [Balance]
## About the PR 
Swap AI upload and stock exchange on donut 3
image
image

## Why's this needed? 
Donut 3 AI Upload is in a very weird spot by arrivals making it very difficult to crime in and difficult for an AI to defend without impacting arrivals and the only kill-switch is in AI core, making it very difficult to kill any AI

## Changelog 

```changelog
(u)JORJ949
(*)Donut 3's AI Upload and Stock Exchange have swapped locations.
```


Chatauscours:
Overall looks good, though I am really not a fan of the upload having IR perimeter detectors. It's supposed to be secure, but the AI can already set up a veiwport there if they want it, they don't need an additionnal automated alert for it.

Also I have a slight suspicion that the cyborg dock could very easily explode from door hacking / turret fire, so i think it would be better to place it in the foyer.

Sbmhawk:
> Overall looks good, though I am not really a fan of the upload having IR perimeter detectors. It's supposed to be secure, but the AI can already set up a veiwport there if they want it, they don't need an additionnal automated alert for it.
> 
> Also I have a slight suspicion that the cyborg dock could very easily explode from door hacking / turret fire, so i think it would be better to place it in the foyer.

Most maps (except oshan for some reason) have IR beams going across the space-exposed parts of core/upload. Nadir has one going through the back wall as well, even though it's not space-exposed.

TobleroneSwordfish:
Should prooobably have reinforced walls like most other uploads do especially now there's a Crime Room right next to it

JORJ949:
> Should prooobably have reinforced walls like most other uploads do especially now there's a Crime Room right next to it

It IS reinforced (im like 90% sure), see the upper wall in the corridor for what the non-reinforced ones look like

Chatauscours:
The APC probably shouldn't be on the easily crimeable wall, as reaching it let's you turn off the turrets easily without getting shot once.

JORJ949:
> The APC probably shouldn't be on the easily crimeable wall, as reaching it lets you turn off the turrets easily without getting shot once.

APCs are actually placed not on the tile they seem to be, this is on the same as the table so while hacking the APC it will be shooting at you

Sord213:
The random room placed right next to AI upload feels really out of place.

JORJ949:
> The random room placed right next to AI upload feels really out of place.

I originally had nothing there and someone suggested linking them up as the gap felt odd

Sord213 (changes requested):
That definitely looks better. Just got a couple notes
Screenshot_1016
Red highlight = wiring isn't complete.
Green highlight = DWAINE computer needs to be connected with a data terminal underneath
Yellow highlight = Windows + glass airlock. I like being able to see in from the foyer
Red arrows = swap the law modules and the robotics control computer. I'd rather the robotics control computer be slightly more protected from space so someone can't break the wall down and killswitch the AI/borgs without being seen.

Also: merge the foyer and upload proper into a single area. I'd rather have the turrets shooting at you when in the foyer as well as the main area.

Screenshot_1018
Red highlight = blinds for a little crime
Red arrow = fix the caution decal in the window

Chatauscours:
Снимок экрана (335)
Also, wire is missing here, and the AI display in the upload appears to lack a pixel offset.

Sord213:
AI display auto offsets, iirc.

JORJ949:
> AI display auto offsets, iirc. 

Correct, you can see its actual placement in the screenshots

JORJ949:
> That definitely looks better. Just got a couple notes
> Screenshot_1016
> Red highlight = wiring isn't complete.
> Green highlight = DWAINE computer needs to be connected with a data terminal underneath
> Yellow highlight = Windows + glass airlock. I like being able to see in from the foyer
> Red arrows = swap the law modules and the robotics control computer. I'd rather the robotics control computer be slightly more protected from space so someone can't break the wall down and killswitch the AI/borgs without being seen.
> 
> Also: merge the foyer and upload proper into a single area. I'd rather have the turrets shooting at you when in the foyer as well as the main area.
> 
> Screenshot_1018
> Red highlight = blinds for a little crime
> Red arrow = fix the caution decal in the window
> 

I made the foyer seperate because of like cog1 and cog2 it is its own little room, Im particularly not a fan of same area+windows at the same time as lethals will be able to shoot you while at the controls.

PR #18561, opened 2024-04-19
Gang locker movement

TDHooligan:
[GAMEMODES][BALANCE]
## About the PR 

Lets gang leaders move their locker to somewhere in their territory once every 15 minutes.
Slightly increases the minimum range between lockers

## Why's this needed? 
1: When too close together, lockers cause excessive (almost boring) amounts of violence. Sec then have to get involved, and start patrolling the area. Which then causes more violence, and more lockdowns, and then everyone gets arrested, lockers get blockaded and nobody's having fun.

2: gangs occasionally just... firebomb/explode/plasmafire other lockers. and there's not much the other gang can do besides spend 20+ minutes fixing a hole/plasmafire. surefire win really.

Leaders experiencing too much stress will now be able migrate their locker to somewhere less miserable if they find that where they are isn't working.

There are probably a lot of interesting strategies this may open up. But I think it's a step forward in improving stalemates. And the problem cases can be addressed down the road.

## Changelog 

```changelog
(u)TDHooligan
(*)Gang leaders can now move their locker to somewhere in their territory.
```


code/datums/controllers/process/gang.dm
```diff
@@ -119,6 +119,11 @@
 		imgroup.add_image(objective_image)
 		broadcast_to_all_gangs("<span style='font-size:24px'> We're dropping off weapons & ammunition at <b>\the [drop_zone.name]!</b> It'll arrive in [GANG_CRATE_DROP_TIME/(1 MINUTE)] minute[s_es(GANG_CRATE_DROP_TIME/(1 MINUTE))] so get fortifying!</span>")
 
+		var/datum/game_mode/gang/gamemode = ticker.mode
+		for(var/datum/gang/targetGang as anything in gamemode.gangs) //create loot bags for this gang (so they get pinged)
+			var/datum/targetable/abil = targetGang.leader.current.getAbility(/datum/targetable/gang/move_gang_base)
```
TobleroneSwordfish, line 124: Don't assume that `leader.current` is going to exist, if the leader dies and respawns it can get nulled.

TobleroneSwordfish (changes requested):
Linter also seems unhappy for some reason

code/datums/controllers/process/gang.dm
TDHooligan: my bad. i forgor

code/modules/antagonists/gang/abilities/_gang_leader_ability_holder.dm
```diff
@@ -272,5 +272,76 @@
...
 
 		M.abilityHolder.removeAbility(/datum/targetable/gang/set_gang_base)
-
+		var/datum/targetable/gang/move_gang_base/newAbil = M.abilityHolder.addAbility(/datum/targetable/gang/move_gang_base)
+		newAbil.doCooldown()
 		return
+
+
+
+/datum/targetable/gang/move_gang_base
```
TobleroneSwordfish, line 281: Wait this is just a copy paste of `set_gang_base`, at least make it inherit so you're not duplicating all these checks here (it's already missing the z level check I added to the base one)

TobleroneSwordfish (changes requested):

PR #18569, opened 2024-04-20
Add action bar for hand staunching wounds

FlameArrow57:
[PLAYER ACTIONS][QoL][BUG]

## About the PR 
Adds an action bar for hand staunching wounds

I think this also fixes a small bug with existing hand staunching, where since it was using a sleep(), it wouldn't let you re-attempt staunching until the sleep was over if the first time was interrupted

## Why's this needed? 
More clear on action being done and time it takes

## Changelog 

```changelog
(u)FlameArrow57
(+)Added a visible action bar for hand staunching wounds.
```

### DIFF
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -2308,3 +2308,30 @@
 	for(var/datum/random_event/start/until_playing/RE in random_events.delayed_start)
 		if(RE.include_latejoin && RE.is_crew_affected(src))
 			RE.apply_to_player(src)
+
+/mob/living/proc/staunch_wound(mob/helper)
+	if (src.bleeding <= 0)
+		boutput(helper, SPAN_NOTICE("[src]'s bleeding has already stopped!"))
+		return
+
+	if (prob(80))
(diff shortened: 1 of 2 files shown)


ZeWaka (approved):

code/mob/living.dm
```diff
@@ -2308,3 +2308,30 @@
...
+		boutput(helper, SPAN_NOTICE("[src]'s bleeding has already stopped!"))
+		return
+
+	if (prob(80))
+		helper.tri_message(src, SPAN_NOTICE("<b>[helper]</b> fails to help [src == helper ? "[his_or_her(src)]" : "[src]'s"] bleeding!"),\
+				SPAN_NOTICE("You fail to help [src == helper ? "your" : "[src]'s"] bleeding!"),\
+				SPAN_NOTICE("[helper == src ? "You fail" : "<b>[helper]</b> fails"] to stop your bleeding!"))
+		return
+
+	repair_bleeding_damage(src, 100, rand(1, 2))
```
Tarmunora, line 2323: worth noting that this ends up being an overall nerf to success rates, as repair chance is modified by bleeding amount etc (so it was actually like a 40-50% chance to repair bleeding when at max bleedout)

FlameArrow57: Oop, will fix

PR #18577, opened 2024-04-20
Default AI primary radio to off

TobleroneSwordfish:
[QOL] [SILICONS]
## About the PR 
Sets the AI's primary radio microphone to off by default.

## Why's this needed? 
It's annoying and confusing if you don't know how it works, especially with the intercom override feature still broadcasting on general frequencies by default.
PRing this just in case someone goes "nooo it's like that because..."

## Changelog 

```changelog
(u)LeahTheTech
(+)The AI's primary radio microphone now defaults to off.
```


JORJ949:
This needs to be on to actually talk over the radio as AI via the ":" hotkey, could do with being changed

PR #18579, opened 2024-04-20
Money trees mutate the seedless gene strain upon mutation.

M-Earthfire:
[Balance][Hydroponics][input wanted]

## About the PR 

When a paper tree seed/plant mutates into a money tree, it will be given the seedless gene strain.

## Why's this needed? 

Money trees give an extraordinary amount of money for very low effort if spammed over and over again. By giving these the seedless gene strain upon mutation, spamming them by getting one money tree and replating it's seeds won't be possible anymore. This means to spam these plants you need to mutate each one seperatly, making it take more effort to spam.

This change makes it more feasable to focus on a few but more advanced spliced/stat pumped plants instead of spamming multiple all around botany.

Also, this highly diminishes clutter with seeds being all over botany while working with money trees.

While it would be possible to only make the money tree mutation unable to grant seeds, this also enables to get the seedless mutation in a rather reliable way, if you don't want to work with melon plants for some reason.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)When paper trees mutate to money trees, they will gain the seedless gene strain. 
```

### DIFF
--- a/code/modules/hydroponics/plant_mutations.dm
+++ b/code/modules/hydroponics/plant_mutations.dm
@@ -778,6 +778,10 @@ ABSTRACT_TYPE(/datum/plantmutation)
 	PTrange = list(30, null)
 	chance = 50
 
+	HYPon_mutation_general(var/datum/plant/parent_plant, var/datum/plantgenes/passed_genes)
+		HYPaddCommut(passed_genes, /datum/plant_gene_strain/seedless)
+		return
+
 /datum/plantmutation/tree/paper
 	name = "Paper Tree"
 	dont_rename_crop = TRUE


TDHooligan:
wouldnt this just make hyper optimizing literally any other plant more efficient as you can churn them out into space? seems lame.

Kubius:
Perhaps a limitation to producing no more than one seed, or two? Total seedlessness is rough.

Kubius:
I was thinking one to two seeds on final harvest - and yeah, that would mean a considerably different approach. The main thing to keep in mind, I think, is whether or not the money tree is in such a state before this change (or a different one) that it's causing the experience to be less fun for somebody.

PR #18581, opened 2024-04-21
Adds non-random loot spawners and non-random loot fillers (and an example usage)

MylieDaniels:
[INTERNAL] [CODE QUALITY]
## About the PR 
This PR refactors /obj/randomloot_spawner and the loot system that relies on it, such that random loot is a child, /obj/loot_spawner/random, instead of the only option. This adds the option of making loot grid items that are not placed into the gang pool or any other pool. In addition, it adds the only applicable thing I think wasn't already supported by the loot generator- a function that fills the remainder of the grid with a particular instance of loot, which in my eyes would typically be used for consumables.

This non-obtainable crate of STX grenades uses this code to fill it with a grid of either 6 or 12 grenades (depending on subtype) to show future coders how to format a non-random loot_spawner, and might be useful for admins to give to nukies or something.
image
Mr. Muggles immediately beelined for the war crime boxes, by the way.

## Why's this needed? 
Allows for loot_spawner's lovely organizational code to be used without slowing down the population of the random lists or defining new loot pools.


TDHooligan:
rockin'.
looks good to me

ZeWaka (approved):

PR #18587, opened 2024-04-21
Introduction of plantgrowth_tick and rework of the fast/slow_metabolism gene strains

M-Earthfire:
[Hydroponics][Code Quality][rework]

## About the PR 

This PR introduces `var/datum/plantgrowth_tick`. This is a holder for a multitude of information that is passed along procs of chemicals and gene strains that interact with the process of hydroponic trays.

Whenever a planttray needs to calculate tray-chemistry/water consumption/plant growth, it creates a plantgrowth_tick-datum. This will be stored in `plantpot.current_tick`. This datum get's passed into the plant-process procs of every chemical in the plants tray and every gene strain. These procs were changed to modify this datum only. If all changes are applied to the datum, `plantgrowth_tick` will be passed onto the new `/plantpot/proc/HYPresolve_plantgrowth_tick`-proc and applies all these changes onto the plantpot. Once these changes are applied, the old `plantgrowth_tick` is deleted and a fresh one will be created, waiting to be resolved/manipulated.

Since `var/datum/plantgrowth_tick` and `HYPresolve_plantgrowth_tick` are made to handle fractional changes, all procs that have a %-chance to modify a plant instead apply their average result directly.

Also, this PR makes plantpots now accept compost and holy water as water-substitutes. Man-eating plants now accept blood instead of water.

Last, but not least, this change enabled a rework of the slow/fast-metabolism trait. They will now affect almost every stat-change within  `HYPresolve_plantgrowth_tick`. Fast metabolism increase water consumption/tray chemistry/plant growth by 50%, but increases the baseline-chem-consumption by 1/tick. Slow metabolism decrease water consumption/tray chemistry/plant growth by 25%, but decrease the baseline-chem-consumption by 0.5/tick.

Above this, this changes have some side-effects. If the plant requires water to grow, ammonia and other growth-modifying chems wont have any effect if the hydroponics tray has not sufficient water.

## Why's this needed? 

The `/process()`-proc of hydroponic trays had a good amount of hardcoded interactions, which were in dire need of a cleanup. Also, many chems and gene strains modified the plant directly, called procs of the plantpot or used unnecessary chance-rolls. Bundling all of these into a single datum enables to save on a lot of unecessary calls and cleaned up their respective code.

These changes enabled some more interesting interactions. As an example, fast/slow metabolism were changed to have a new niche within gene strains. Since fast metabolism was simply a worse rapid growth gene strain, changing them up diversifies the selection of gene strains a botanist can go for. With these changes, slow metabolism drastically reduces water consumption now (halving baseline-consumption while only bringing a 25% malus), while being a net negative gene strain, makes it interesting for lazy botanists who just want to use the mister and don't care for hand-filling trays.

Last but not least, the addition of water substitutes gives compost a new role within botany. People within the discord were noting a lack of use for the chemical and also a lack of water if they go for ammonia. With fast metabolism highly increasing water consumption, having compost as another alternative seems like a good idea.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Chemicals in hydroponic trays now should show a lower varience.
(+)Plants accept compost and holy water as alternatives to water. Man-Eating plants accept blood as an alternative to water.
(+)Fast/slow metabolism gene strain now affect many processes around chems in trays or gene strains instead of just improving growth speed.
```


NibChocolateeny:
it is good to have nutriants not make growth happen if water is not there - good chnage!

Does grow lamp/glow lotus efffect over-ride and make a plants grow with no water still?

code/obj/machinery/plantpot.dm
```diff
@@ -823,6 +793,70 @@ TYPEINFO(/obj/machinery/plantpot)
 	if(src.growth >= src.current.harvtime - src.plantgenes?.get_effective_value("harvtime")) return TRUE
 	else return FALSE
 
+/obj/machinery/plantpot/proc/HYPresolve_plantgrowth_tick(var/datum/plantgrowth_tick/growth_tick)
+	// The above is pretty much to figure out whether or not the water level
+	// icon on the plant pot needs to change.
```
Tarmunora, line 798:
```suggestion
```

PR #18589, opened 2024-04-21
Energy Drink changes

colossusqw:
[LABEL][chemistry][balance]

## About the PR 

This PR changes Energy Drink's behavior to be more dependant on it's caffeine output, similarly to how heavy alcohol value chems currently work. To achieve this, it:

- Increases it's depletion to 0.5, and caffeine content per unit to 2.5
- Decreases it's standalone stun resist to 15, from 25
- Makes it so overdose has a increasing chance of dumping out more caffeine
- Changes the on depletion effect to slow the player down, as well as remove a amount of caffeine based on how long it's been on you (to reinstall the now barely noticeable energy drink withdrawal effect)

## Why's this needed? 

It would be interesting to have a drink defined by it's caffeine output be more governed by it, instead of having it's own set of behaviors. I've tried to maintain something akin to current balance of effectiveness/risk, as well as reinstall the previous immediate downside upon having it run out that the pre-caffeine version had as well.

## Changelog 

```changelog
(u)Colossusqw
(+)Energy Drink now no longer has a random chance of causing heart failure, now the heart failure is handled via it's increased caffeine output.
```

### DIFF
--- a/code/modules/chemistry/Reagents-FoodDrink.dm
+++ b/code/modules/chemistry/Reagents-FoodDrink.dm
@@ -2567,7 +2567,7 @@ datum
 			fluid_b = 20
 			taste = "bitter"
 
-		fooddrink/caffeinated/coffee/energydrink
+		fooddrink/caffeinated/energydrink
 			name = "energy drink"
 			id = "energydrink"
 			description = "An energy drink is a liquid plastic with a high amount of caffeine."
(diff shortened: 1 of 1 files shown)


Tarmunora (approved):

Tarmunora (changes requested):
actually now that `caffeinated` is pulled out of coffee, energydrink could probably stand to no longer be a child of coffee

colossusqw:
Can do

PR #18593, opened 2024-04-22
Introduces helper procs for botany to unify growth-stage calculations

M-Earthfire:
[Code quality][hydroponics]

## About the PR 

This PR introduces defines and helper procs to unify most calculations revolving around growth stages of plants:

There were multiple defines introduced to express the sprite changes with the growth stage. These are `HYP_GROWTH_PLANTED`, `HYP_GROWTH_GROWING`, `HYP_GROWTH_MATURED` and `HYP_GROWTH_HARVESTABLE`. Also, `HYP_GROWTH_DEAD` was introduces when the next procs are called with dead plants or insufficient data, but currently it is mostly unused.

`datum/plant/proc/HYPget_growth_stage` was introduces to return a value equal to the current growth stage of a given plant with given plantgenes. `proc/get_current_growth_stage()` was introduces on `/obj/machinery/plantpot` to output the growthstage with the current plant and plantgenes in the corresponding plantpot.

Also, 3 procs were introduced to give out the required growth time needed to reach a certain growth stage. These are `datum/plant/proc/HYPget_growth_to_growing`, `datum/plant/proc/HYPget_growth_to_matured` and `datum/plant/proc/HYPget_growth_to_harvestable`

All calculations based on the corresponding growth-stages were changed to use these procs and defines.

## Why's this needed? 

There were a multitude of places where plants needed to check their current growth to determine when to fire which effect. This meant that things like `P.growtime - DNA?.get_effective_value("growtime")` were copied over and over. Also, this has caused some mistakes to be copied over, like `P.growtime + DNA?.get_effective_value("growtime")`, which caused some plants to fire their special effects later/not at all of you increased growing speed. This change removes these kind of bugs entirely.

Also, this change improves the readability of plant code and makes it easier to change the internal calculations. If, for example, you want to change the time a plant has reached the mature state, you only need to change `HYPget_growth_to_matured` now instead of changing multiple sections of code.


ZeWaka (commented):
i think a lot of this is good but there's way too many HYP global procs, they should really be put into some manager datum or something, not that you need to do it but it'd be nice

ZeWaka (approved):
LGTM!

PR #18596, opened 2024-04-22
Adds a nasty overdose-effect to liquid space rubber and additional brute damage on bounce

M-Earthfire:
[chemistry][balance][feature]

## About the PR 

This PR adds an overdose effect to liquid space rubber:

at 25u, the normal overdose treshold is reached. At this stage, with a 10% chance and a 8 seconds cooldown, the persons body begins jiggling uncomfortably, dealing between 4 and 8 brute damage, depending on the amount of LSR in them, to them.

at 50u, the real overdose kicks in. Over the next 23 cycles, they suffer escalating brute damage while receiving more and more scary messages and jiggling even more intense. At the end of the 23th cycle, ~~they gib, throwing all their items in and on them in an area around them. Their non-robotic organs will be converted into synthrubber~~ the jiggling will reach it's highest intensity, dealing 12 damage with a 25% every 6 seconds while stunning the victim for 2 seconds.

to show this effect properly, a new animation has been added, `animate_flubber`

The amount of cycles elapsed will slowly decreased between 50u and 25u and fully reset below 25u.

Also, bouncing off another person will now deal 7 brute damage to both parties with a 2 second cooldown.

## Why's this needed? 

Liquid space rubber is a fun chemical that has some very strong uses in combat. While not necessarily problematic when self-made, starstone-infusion can grant you infinite LSR in your body without any repercussions.

A 50u treshold for the major overdose is high enough that using LSR in pills shouldn't be affected too much and it is manageable enough with purging chems, like charcoal. It is also be high enough so other more involved methods of automatic dosing, like injector belts, should still be useable.

The Overdose effect is punishing enough that LSR-infusions require much more management with e.g. calomel/pent to be kept under control.

The OD effect is also flashy and scary enough that sleepypen-users might consider penning people with LSR and kill them in a very gruesome way. It requires around 80u of LSR to cause someone to go into crit, so it might not be the best method avaible.

Also, the on-bounce damage effect adds a damage source that affects LSR+magboots-users as well as non-magboot users.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Consuming too high amounts of liquid space rubber has been found affect the consumers body in unintended ways.
(+)Bouncing off other people while under the effect of LSR now deals brute damage to both parties.
```


code/modules/chemistry/Reagents-Misc.dm
```diff
@@ -2272,6 +2272,74 @@ datum
...
+						//now the real fun starts
+						src.OD_ticks += 1
+						if(src.OD_ticks > 8 && prob(10))
+							boutput(M, SPAN_ALERT("<I>Your finger's feel rubbery!</I>"))
+							var/h = M.hand
+							M.hand = 0
+							M.drop_item()
+							M.hand = 1
+							M.drop_item()
+							M.hand = h
```
Tarmunora, line 2309: why not use `empty_hands()` here instead

Tarmunora (changes requested):

TobleroneSwordfish:
Are starstone-flubber infusions that big of a problem? They're usually just kind of funny.

Retrino:
gibbing seems like a bit much but the od effect itself is funny

TobleroneSwordfish:
Eh, I'd rather not add a nasty noob-trap Gib effect to a chemical that isn't really causing issues as is.

Tarmunora:
> > Are starstone-flubber infusions that big of a problem? They're usually just kind of funny.
> 
> I have read some complains over people using it for rampages. I don't think it's that big of a problem, but i can see it being a bit too strong for how easy it is.

Imo flubber-self-throwing should include some effect on the user (i.e.: something like a 1s stun) that doesn't just get entirely negated by wearing magboots, which is the crux of the problem (one piece of not-that-rare clothing countering the intended downsides)

Tarmunora (approved):

PR #18600, opened 2024-04-23
Adds A mining magnet to atlas

cheekybrdy:
## About the PR 
Adds a mining magnet to the "stinger" portion of atlas, to respect the bee shape. Also moves the artifact ejector to route artifacts to be ejected with the chapel mass driver to prevent collisions with the magnet.

## Why's this needed? 
Having no mining magnet has always just been more of a inconvenience then characteristic of Atlas, especially if the miner players as they frequently are are newer players, given that this respects the shape of atlas which was the concern with previous iterations I don't see a reason not to add this.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Added a mining magnet to atlas, its just behind the cold loop.
```


cheekybrdy:
branch is abnormally dated, closing

PR #18607, opened 2024-04-24
Atlas mining magnet

cheekybrdy:
## About the PR 
This PR adds a mining magnet to atlas, with consideration for the bee shape which other implementations have lacked. consequentially artifact ejector has been rerouted to eject through the chapel driver instead of out the back of the ship due to the new obstructions.

## Why's this needed? 
Having mining offstation is more of a inconvenience then something unique, its also rather jarring for new miners to either not know where to go or isolate themselves from the station they're trying to learn in order to do their job.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Added a mining magnet to atlas.
```


cheekybrdy:
image

Glamurio (approved):

cheekybrdy:
!merge_upstream

cheekybrdy:
!merge_upstream

TobleroneSwordfish (changes requested):
Generally looks good, Gannets says yes. A few things I noticed while walking around:
- This corner looks weird with the engine blocks making up the actual corner tile, usually we make sure there are no diagonal walls like this:
image
- Two rechargers for such a small department is a bit much, most mining departments only have one (afaik?)
- This door could use either an additional airlock with cyclers or a door linked forcefield to stop it depressurizing maint every time someone walks through:
image
- The catwalks are missing support turfs, see other catwalks on the map for reference.
- The purge valves are adjacent to catwalks which means they would vent gas onto the catwalks if they had support turfs.

PR #18352, opened 2024-03-29
Add PDA Alert modules for medical, engineering, science, and janitor teams

pgmzeta:
[FEATURE][STATION SYSTEMS][GAME OBJECT]
## About the PR 

* Add PDA alert modules for the Medical/Engineering/Science departments, and Janitor team.
* Add a new crisis alert for Science, complete with new sound 
* Add new sprites for both the module and the ability icon
* For each new alert module, add two of each to their respective job vendors (i.e. jobclothing / janitech).
    * Security alert modules are still only accessible via security PDAs.
* Department heads now start with an alert module for their department.

## Why's this needed? 
* Let teams know when their *own coworkers* think things are bad (and not science for the Nth time)
* Allow 'first-responder' paramedics/repairers/janitors to call for backup
* Nudge heads of staff towards directing/requesting help from their own department
 
### Item Sprites
the new modules are in the middle row:
Screenshot 2024-04-02 131634

### Ability Icons
image

### Science Alert in action
https://github.com/goonstation/goonstation/assets/91498627/d3281a23-f278-4c06-af77-50e29299ccd1

## Changelog 

```changelog
(u)glowbold
(*)Team-specific PDA Alert modules for janitorial, medical, engineering, and science are in their respective job vendors.
(+)Department head PDAs spawn with an alert module for their respective department.
```

PR #18611, opened 2024-04-24
Adds the science alert from #18352 to general PDAs.

cheekybrdy:
## About the PR 
Atomises the Science alert from the closed #18352 to the crisis alert PDA app.

## Why's this needed? 
Science while not in demand very much right now should have the ability to be called over for things like kudzu and artifacts for now and hopefully more in the future. This PR doesn't have the PDA intermal alerts of the original PR so I assume at least that this addition is fine.

## Changelog 

```changelog
(u)Pgmzeta
(+)Adds a Science alert to the Crisis Alert PDA app.
```


ZeWaka:
Why not just PDA message the group

cheekybrdy:
> Why not just PDA message the group

I may be missing something, but isn't the mail group variable doing that here?

ZeWaka:
as in, why can't people just message the mailgroup using the pda functionality
seems fine for
>  things like kudzu and artifacts

cheekybrdy:
oh in that regard the intent was more cultural and more for future proofing science. Being the only department without a alert could be seen as another sign of science lacking responsibility, even though in present there aren't too many reasons for a callout.

Glamurio:
> as in, why can't people just message the mailgroup using the pda functionality seems fine for

Maybe this is just me, but I pretty much ignore mailing via PDA because they just get buried in chat logs, do not stand out and are awkward to use.

I think this is a cool addition in theory, the main issue is actually that Science has little reason to ever leave Science, so there wouldn't ever really be a reason to hit the alert button. It's more of an issue with Science's design than the alert being a bad idea.

Tarmunora:
Science PDA alerts aren't something we need at this time

PR #18617, opened 2024-04-24
Add new "Mute" trait and Speech Pro device

Glamurio:
[Traits] [Feature]

## About the PR 
Adds a new trait "Muted", which spawns you with permanent muteness and a new item, the Speech Pro.

The Speech Pro is a hand-held device which can be used to allow mute individuals to communicate simple, pre-formatted phrases akin to MMORPG quick chats.

image
image

There is a 3 second cooldown between text output to prevent spam.

The Speech Pro has also been added to the robotics and medical fabricator.

## Why's this needed? 
I've always thought blindness and deafness are very cool ways to customize your character. I've wanted to add muteness to complete the disability trio, but felt that simply making someone mute may be too inhibiting or seem like a "Mime-Lite". Additionally, I did not want to covet a mask slot for a "speech enabler" of sorts, instead, I thought it would be more fun to add a goofy device with which people are still able to communicate, but while essentially letting other players know they cannot freely speak.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Added a new trait, "Muted", and a hand-held device that allows simple communication.
```


NibChocolateeny:
👏👏👏

CaioFrancisco:
cool stuff. just a suggestion, it would be cool if mimes were unable to use this.

444Portal:
> cool stuff. just a suggestion, it would be cool if mimes were unable to use this. 

They can already use PDAs and Paper, I don't think it's a big deal. Yeah it'd go aganist the mime vibe but like its their choice and they already have ways to ignore the no talking thing so shrug

TobleroneSwordfish:
Not a huge fan of adding a mute trait to a game based around communication (1 mime per round is enough) but the speech pro is neat, maybe could give it to martian ambassadors in place of a full translator?

Glamurio:
> Not a huge fan of adding a mute trait to a game based around communication (1 mime per round is enough) but the speech pro is neat, maybe could give it to martian ambassadors in place of a full translator?

I see it more as a choice, people already have the ability to not talk or find alternative means to communicate, with or without the trait. I purposefully left the trait on neutral cost because I didn't want to incentivize usage for meta-reasons. Martian Ambassador's getting it seems like a cool idea either way (probably in a separate PR)

TDHooligan:
it could be a -1 that makes you start with it. or a trinket replacer...

Glamurio:
> it could be a -1 that makes you start with it without muting. or a trinket replacer...
> 
> having it left on permanently could really ruin a good shift if it was mechanically enforced.

Considering the trait is optional, it shouldn't ruin anything any more than any other major disability of your choosing. You can also print replacement Speech Pros at the MedFab, should you lose yours or have it stolen, and then naturally you can always communicate via writing of notes, pointing or doing emotes.

TobleroneSwordfish:
Okay after speaking with the rest of the admin team I'm going to close this. People *do* have the option to choose to not speak but crucially adding it as a trait encourages and normalizes it in a way we'd rather not see in a game where a lot of our problems stem from getting people to talk to each other in the first place. Mime is limited to one per round for a reason.

PR #18619, opened 2024-04-25
Resizes cargo and mechlab on atlas

Knipje:
[Mapping] [Rework]

## About the PR 
Sizes down cargo a bit and increases mechlab instead.

Also fixes some area markers around that area and the new cafeteria rework.

## Why's this needed? 
Cargo feels disproportionally large on Atlas, especially compared to the incredibly small mechlab next to it making experimentation by engineers harder.

## Changelog 

```changelog
(u)Knipje
(+)Expanded mechlab on Atlas.
```

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -585,8 +585,6 @@
 /obj/table/auto,
 /obj/item/storage/belt/utility,
 /obj/item/storage/belt/utility,
-/obj/item/storage/toolbox/emergency,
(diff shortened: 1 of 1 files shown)


pgmzeta:
I feel like this makes cargo really, really tight, with no space for crates to sit for fabs without being in the way.

TobleroneSwordfish:
I don't see a reason for this, cargo needs the floor space a lot more than mechlab.

PR #18622, opened 2024-04-25
Armory doors immunity to wire hacking

JORJ949:
[BALANCE] [Systems] 

## About the PR 
Makes armory doors unable to be pierced by screwdrivers and gives them a cool new sprite
image

## Why's this needed? 
Makes Armory unable to be ran into and completely emptied in under 5 minutes if security isn't on their A-game or straight up doesn't exist. Now you need to authorize, deconstruct a wall or Emag in, all being much more noticeable and louder forms of entry. Sneakily breaking into Armory SHOULD be hard. Also encourages antagonists to use their own kit, rather than security's.

## Changelog 

```changelog
(u)JORJ949
(*)Armory doors are now immune to wire hacking.
```


JORJ949:
> The armory does not need any more protection than it already has. Beyond IR wires, the narc, turrets on the outside and multiple rwalls, AND being in the middle of security, the armory has plenty of protection. In comparison to ither codebases, our armory is already unreasonable protected.
> 
> If security does not pay atttention to people breaking into one of the most guarded locations, that's on them.

The Rwalls mean nothing, the NARCS is always prepped for with either the N2O tank or some other object to block it, the IR detectors i have NEVER seen used and on many maps armory is actually at the edge of security and breakins cannot be heard from where sec will be spending most of their time

Glamurio:
> The armory does not need any more protection than it already has. Beyond IR wires, the narc, turrets on the outside and multiple rwalls, AND being in the middle of security, the armory has plenty of protection. In comparison to ither codebases, our armory is already unreasonable protected.
> 
> If security does not pay atttention to people breaking into one of the most guarded locations, that's on them.

Heavy disagree here. All it takes is one robust player taking advantage of an opportune moment to be able to slide into the armory. Frankly, in my opinion the armory doors should not even be able to be emagged, because it kind of turns them into a joke.

On either a low pop round with few Security, or a high pop round with lots of antags it's absolutely possible for someone to slip into Security unnoticed, depending on the map sometimes the armory is so isolated that it's hard to constantly keep an eye on it. Increasing the amount of time it takes to open the doors considering the pay-off is fair in my mind.

Sord213:
I considered adding custom times for different doors when I added the action bar on opening door panels. I'd rather see that than 100% unhackable armory doors. Maybe I'll revisit the idea

ZeWaka (commented):
i think this is good in combination with #18275

TDHooligan:
> i think this is good in combination with #18275

seconded. rather than having to be 24/7 on watch for the armory, you can know that if the auth disc is safe, the armory doors are.

that leaves the rwall as the general weakest link for breaking in, which is a much better amount of risk for break ins.

NightmareChamillian:
Strongly against this implementation. I think part of the reason the armory doors are so frequently hacked is because trying to gain access through any other means is almost impossible. Especially compared to a quick hack or two.

I definitely think getting through the amory doors should be way less trivial, but making breaking in _even_ harder as a whole doesn't seem to be the right way of going about things.

TDHooligan:
making other break-ins more accessible on top of this PR is a good way to go. but im happy briefly having a too-locked-down armory if only to shake up the powergame in the short term.

MeggalBozale:
Happy in favor of this,opening the panel and snipping a wire should not be the weak link. It makes all other defense systems obsolete, as everything except the NARCS is useless to that mode of attack. Not like it's that hard to get thermite, if you want to be quick. Or an RCD. Or explosives. Or anything, really. I can't say it's that hard to break into the armory, passive defenses are weak.

FlameArrow57:
Yeah idk, not too common, but there are cases when security officers/regular crew members will need to hack into the armory for a single item, without a giant announcement to everyone that it was authed, or for emergency situations when security/heads of staff are dead

JORJ949:
> Yeah idk, not too common, but there are cases when security officers/regular crew members will need to hack into the armory for a single item, without a giant announcement to everyone that it was authed, or for emergency situations when security/heads of staff are dead
> 
> Constant armory hack in is not too fun to deal with, but think fully unhackable doors would maybe be a bit too much

If you are using armory it should be announced to the server imo and with the auth disk change you can just shut it behind you after getting the one item.

TDHooligan:
> Yeah idk, not too common, but there are cases when security officers/regular crew members will need to hack into the armory for a single item, without a giant announcement to everyone that it was authed, or for emergency situations when security/heads of staff are dead
> 
> Constant armory hack in is not too fun to deal with, but think fully unhackable doors would maybe be a bit too much

sec offs can just... decon the walls. reinforced walls arent impossible to break (actually arent that difficult, just slightly time consuming)

my original concerns for this commit are gone. i'm 100% for it now. what kind of high security door locking down firearms has the access panel on the outside anyway

Valtsu0:
I'd prefer if hacking the door was hard instead of impossible (Shuffled wires, Alarm wire, bolt on power off and multiple id wires?), but this is fine too

I do agree that blue on black doesn't really match security

JORJ949:
> I'd prefer if hacking the door was hard instead of impossible (Shuffled wires, Alarm wire, bolt on power off, multiple id wires?), but this is works too
> 
> I do agree that blue on black doesn't really match security

I left it as a "command" style door because some maps use those and I also am not a spriter.

JORJ949:
Made them a subtype that can be RCD'd and the AI can hack them again, also has a cool red colour

maps/donut3.dmm
Valtsu0, line 1: This really doesn't seem correct

Valtsu0:
Wonder why map diff bot fails. Donut3 and Oshan having suspiciously high diffs might have something to do with it

JORJ949:
> Wonder why map diff bot fails. Donut3 and Oshan having suspiciously high diffs might have something to do with it, maybe the merge conflict wasn't properly resolved?

According to zewaka its an issue with the bot

JORJ949:
Stop it stop conflicting

JORJ949:
Alright who removed a SINGLE line 
image

PR #18632, opened 2024-04-25
More, smaller gangs

TDHooligan:
## About the PR 
Lowers gang size from 5 -> 4 (1 leader, 3 members)
Lowers pop requirements per gang from:
- 15 -> 12 (RP)
- 12 -> 9 (Classic)

## Why's this needed? 
People aren't seeing 3 gangs. I think this is worth a testmerge to see how this affects gang counts.
It could also be a bug? This will make it more obvious.

## Changelog 

```changelog
(u)TDHooligan
(*)Gangs have been shrunk from 5 to 4 members. 
(*)More gangs will spawn, so an equal number of gang members will exist.
```

### DIFF
--- a/_std/defines/gang.dm
+++ b/_std/defines/gang.dm
@@ -1,4 +1,4 @@
-#define GANG_MAX_MEMBERS 4
+#define GANG_MAX_MEMBERS 3
 
 /// How long the leader must cryo before gang members can take their role
 #define GANG_CRYO_LOCKOUT 7.5 MINUTES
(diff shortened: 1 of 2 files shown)


Glamurio (approved):

Sovexe:
Noting that this will result in less overall gang members on average for most populations

image

TDHooligan:
yeah. reasonable, to be honest i'm not very sure this is worth merging outside of being 100% certain if gang pops are OK. i'm fairly sure it's bugged ATM

Sovexe:
The issue may simply be that a lot of player late join, skewing the initial gang numbers

PR #18636, opened 2024-04-26
Change Lawbringer to apply DNA on redeem

Glamurio:
[QoL] [Removal]

## About the PR 
Changes the Lawbringer to immediately apply the DNA string upon redeeming the job reward.

## Why's this needed? 
This means you won't have to activate / drop and re-grab the gun after redeeming and removes user error.
Additionally, it removes the chance for something to go horribly wrong on redemption or the HoS forgetting to DNA lock the gun. The gun is already supposed to be tied to the HoS, might as well ensure this properly.

### DIFF
--- a/code/obj/item/gun/energy.dm
+++ b/code/obj/item/gun/energy.dm
@@ -1300,19 +1300,23 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 		..()
 
 	attack_hand(mob/user)
-		src.assign_name(user)
+		if (!owner_prints)
+			src.assign_name(user)
 		..()
-
 
 	//if it has no owner prints scanned, the next person to attack_self it is the owner.
 	//you have to use voice activation to change modes. haha!
 	attack_self(mob/user as mob)
 		src.add_fingerprint(user)
-		src.assign_name(user)
+		if (owner_prints)
+			boutput(user, SPAN_NOTICE("There don't seem to be any buttons on [src] to press."))
+		else
+			src.assign_name(user)
+		..()
+
 
 	proc/assign_name(var/mob/M)
 		if (owner_prints)
-			boutput(M, SPAN_NOTICE("There don't seem to be any buttons on [src] to press."))
 			return
 		if (ishuman(M))
 			var/mob/living/carbon/human/H = M
--- a/code/obj/item/gun/energy.dm
+++ b/code/obj/item/gun/energy.dm
@@ -1291,7 +1291,7 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 		// projectiles = list(current_projectile,new/datum/projectile/bullet/revolver_38/lb,new/datum/projectile/bullet/smoke,new/datum/projectile/bullet/tranq_dart/law_giver,new/datum/projectile/bullet/flare,new/datum/projectile/bullet/aex/lawbringer,new/datum/projectile/bullet/clownshot)
 
 		src.indicator_display = image('icons/obj/items/guns/energy.dmi', "")
-		assign_name(M)
+		src.assign_name(M)
 
 		..()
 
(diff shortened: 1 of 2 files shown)


Chatauscours:
I am not entirely sure of the purpose of this change. 
Also, what will happen with mechprinted lawbringers? Do they just not get an owner at all? If so, iirc that messes a bit with the voice input.

Glamurio:
> I don't entirely see the purpose of this change. Also, what will happen with mechprinted lawbringers? Do they just not get an owner at all? If so, iirc that messes a bit with the voice input.

I didn't actually change the way that DNA scanning works. Mechprinted Lawbringers work exactly the same way as before. All I've done is add the DNA scan immediately on redeeming the reward. It's really just to minimize user error.

code/obj/item/gun/energy.dm
```diff
@@ -1300,26 +1300,24 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
...
 	//if it has no owner prints scanned, the next person to attack_self it is the owner.
 	//you have to use voice activation to change modes. haha!
 	attack_self(mob/user as mob)
 		src.add_fingerprint(user)
-		if (!owner_prints)
-			boutput(user, SPAN_ALERT("[src] has accepted your DNA string. You are its owner!"))
-			assign_name(user)
-		else
-			boutput(user, SPAN_NOTICE("There don't seem to be any buttons on [src] to press."))
+		assign_name(user)
```
pgmzeta, line 1311:
```suggestion
		src.assign_name(user)
```

code/obj/item/gun/energy.dm
```diff
@@ -1300,26 +1300,24 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 		..()
 
 	attack_hand(mob/user)
-		if (!owner_prints)
-			boutput(user, SPAN_ALERT("[src] has accepted your DNA string. You are its owner!"))
-			assign_name(user)
+		assign_name(user)
```
pgmzeta, line 1303:
```suggestion
		src.assign_name(user)
```

code/obj/item/gun/energy.dm
```diff
@@ -1300,26 +1300,24 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
...
 
 	proc/assign_name(var/mob/M)
+		if (owner_prints)
+			boutput(M, SPAN_NOTICE("There don't seem to be any buttons on [src] to press."))
+			return
 		if (ishuman(M))
 			var/mob/living/carbon/human/H = M
 			if (H.bioHolder)
+				boutput(M, SPAN_ALERT("[src] has accepted the DNA string. [H.real_name] is now the owner!"))
```
pgmzeta, line 1320: boutput direct to user, verbiage change does not make sense.

```suggestion
				boutput(M, SPAN_ALERT("[src] has accepted your DNA string. You are now the owner!"))
```

code/obj/item/gun/energy.dm
```diff
@@ -1300,26 +1300,24 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
...
-		if (!owner_prints)
-			boutput(user, SPAN_ALERT("[src] has accepted your DNA string. You are its owner!"))
-			assign_name(user)
-		else
-			boutput(user, SPAN_NOTICE("There don't seem to be any buttons on [src] to press."))
+		assign_name(user)
 
 	proc/assign_name(var/mob/M)
+		if (owner_prints)
+			boutput(M, SPAN_NOTICE("There don't seem to be any buttons on [src] to press."))
```
TobleroneSwordfish, line 1315: PR makes sense buut this message would go off every time you pick it up as written I think?

Glamurio: Oops. Yeah.

TobleroneSwordfish (approved):

Issue #18634, opened 2024-04-26
Throwing potwater seethers at gang tags deletes them


goonstation-issuebot:
### Labels

[BUG][CRITICAL]

### Description

If you throw a seether tomato spliced with potassium and water at a gang tag it deletes them

### Steps to reproduce

Grab seethers with potwater spliced
Throw at gang tag
Delete tag

### Expected Behavior

Not delete the tag

### Additional Information & Screenshots

Reported by: Angelx3
Client version: 515.1635
On server: Goonstation 1 Classic: Heisenbee
Active test merges: #15272, #18475, and #18435
Round log date: 2024-04-26-04-01
Reported on: 2024-04-26 04:49:16
Map: NADIR

ZeWaka:
@TDHooligan

TobleroneSwordfish:
fixed in #18638

PR #18638, opened 2024-04-26
Gang tags are resistant to low severity explosions

TDHooligan:
[GAMEMODES][BALANCE]
## About the PR 
Gang tags can no longer be exploded away, and instead have a little easter egg.

## Why's this needed? 
#18634
People are just blowing tags up. It's the same issue as cleaning it normally (but with extra steps).
It's also way easier to seether a tag than to do the tag - It's not fun and can turn territories into a slog again if I just refund the sprays.

### DIFF
--- a/code/datums/gamemodes/gangwar.dm
+++ b/code/datums/gamemodes/gangwar.dm
@@ -176,8 +176,29 @@
 
 	..()
 
+/datum/game_mode/gang/proc/is_deceased(var/datum/mind/M)
+	if (!M?.current)
+		return TRUE
+	if(isdead(M.current))
+		return TRUE
+	if(inafterlife(M.current))
+		return TRUE
+	if(isVRghost(M.current))
+		return TRUE
+	if(!iscarbon(M.current))
+		return TRUE
+	return FALSE
+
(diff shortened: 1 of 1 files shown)


code/datums/gamemodes/gangwar.dm
```diff
@@ -176,9 +176,30 @@
...
 	var/datum/gang/victorius_gang = null
 
+	// Round is over. Grant points for living players.
+	for (var/datum/gang/gang in src.gangs)
+		if (is_deceased(gang.leader))
+			boutput(world, "leader alive for [gang.gang_name]")
+		for (var/datum/mind/member in gang.members)
+			if (!is_deceased(member))
+				boutput(world, "member [member.current] alive for [gang.gang_name]")
+
```
FlameArrow57, line 202: Intended?

TDHooligan: fuck

Sovexe:
blanket explosion immunity is normally reserved for unsimulated areas and adventure zones. Other objects should implement `ex_act` with the `severity` arg in a sane way

PR #18645, opened 2024-04-26
Add action bar for click dragging people onto operating tables

FlameArrow57:
[PLAYER ACTIONS][QoL]

## About the PR 
Adds a visible action bar for when you click + drag someone onto an operating table to move them on

## Why's this needed? 
More visibility/clearer on what's happening


TobleroneSwordfish (approved):
god I wish we had anonymous functions

PR #18650, opened 2024-04-27
Use TGUI list for unloading pod cargo

FlameArrow57:
[UI][QoL]

## About the PR 
Uses tgui_input_list for prompts from unloading pod cargo

## Why's this needed? 
tgui

### DIFF
--- a/code/modules/transport/pods/secondary_system.dm
+++ b/code/modules/transport/pods/secondary_system.dm
@@ -187,7 +187,7 @@
 			if (length(load) == 1)
 				crate = load[1]
 			else
-				crate = input(usr, "Choose which cargo to unload..", "Choose cargo")  as null|anything in load
+				crate = src.get_unloadable(usr)
 			if(!crate)
 				return
 			unload(crate)
@@ -214,7 +214,7 @@
 		boutput(user, SPAN_ALERT("[src] has nothing to unload."))
 		return
 
-	var/crate = input(user, "Choose which cargo to unload..", "Choose cargo")  as null|anything in load
+	var/crate = src.get_unloadable(user)
 	if(!crate)
 		return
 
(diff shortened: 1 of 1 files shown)


code/modules/transport/pods/secondary_system.dm
```diff
@@ -187,7 +187,7 @@
 			if (length(load) == 1)
 				crate = load[1]
 			else
-				crate = input(usr, "Choose which cargo to unload..", "Choose cargo")  as null|anything in load
+				crate = src.get_unloadable(usr)
```
Sovexe, line 190: the `use()` proc should really just be passing `activate()` its `user` arg. `usr` should be avoided outside of unavoidable junk like `topic()`

FlameArrow57: Agreed it works for this specific case, but parent activate() proc doesn't have any parameters, and there's more places than just in the use() proc where activate() is called (one being /obj/machinery/vehicle/Topic())

code/modules/transport/pods/secondary_system.dm
```diff
@@ -318,6 +318,18 @@
...
+	var/list/cargo_by_name = list()
+	var/list/counts_by_type = list()
+	for (var/atom/movable/AM as anything in src.load)
+		counts_by_type[AM.type] += 1
+		if (counts_by_type[AM.type] == 1)
+			cargo_by_name[AM.name] = AM
+		else
+			cargo_by_name["[AM.name] #[counts_by_type[AM.type]]"] = AM
+
+	return cargo_by_name[tgui_input_list(user, "Choose which cargo to unload", "Choose Cargo", sortList(cargo_by_name, GLOBAL_PROC_REF(cmp_text_asc)))]
```
TobleroneSwordfish, line 331: This will runtime if the user hits cancel, don't assume user input procs will always return nice values.

code/modules/transport/pods/secondary_system.dm
TobleroneSwordfish: It's "fine" for pod code and refactoring it would be a massive pain so I'd say don't scope creep the PR here.

TobleroneSwordfish (changes requested):

code/modules/transport/pods/secondary_system.dm
FlameArrow57: nah i tested that case, no issues!

ZeWaka (approved):

code/modules/transport/pods/secondary_system.dm
mordent-goonstation: What's returned?

FlameArrow57: > Member

null

PR #18657, opened 2024-04-27
Simplify charcoal recipe a little

TobleroneSwordfish:
[CHARCOAL] (I typed this instead of [CHEMISTRY] and it made me laugh so now it stays here)
## About the PR 
Simplifies the recipe for charcoal to just be `(1) carbon + (1) ash = (1) charcoal + (1) steam` (the ash is damp or something okay) with a secondary recipe of `(1) carbon + (1) ash + (2) steam = (4) charcoal`
The idea is that if you make it in an open container the steam evaporates and you lose yield compared to making it in a closed container.

## Why's this needed? 
Starting to address some of the rough edges of the chemistry rework. Someone asked in discord earlier how charcoal is made and I realised the explanation should not take half a page for a very basic medical chem. This version removes the common "all my charcoal vanished :(" complaint and brings the recipe more in line with its in-game level of power/complexity.

## Changelog 

```changelog
(u)LeahTheTech
(*)Simplified the charcoal recipe: (1) Carbon + (1) Ash => (1) Charcoal + (1) Steam, with additional charcoal yield if the steam is prevented from evaporating.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -2967,52 +2967,20 @@
 		mix_phrase = "The mixture yields a fine black powder."
 		mix_sound = 'sound/misc/fuse.ogg'
 
+		on_reaction(datum/reagents/holder, created_volume)
+			holder.add_reagent("steam", created_volume)
+
 	charcoal_steam //higher quality activated charcoal can be made irl with steam so this kinda fits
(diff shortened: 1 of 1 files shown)


Flaborized:
I think this is a fine change and I probably overtuned charcoal's recipe a bit, though I think the whole 'use condensers to pipe steam into a reaction' idea is Cool and could be re-utilized for some other recipe (if not the Super Basic Healing Chem one)

PR #18670, opened 2024-04-28
Fix ampoules

### LINKED ISSUE #18669: Ampules become unbreakable when using them on yourself
### Labels

[MAJOR][MEDICAL]

### Description

I tried using a smelling salt ampule I got from a Public MiniMed and nothing happened.

### Steps to reproduce

1. Get a smelling salt ampule
2. Try using it on yourself
3. "You crack open and inhale the ampoule (smelling salts)."
4. Realize nothing actually happened despite the message

### Expected Behavior

For the ampule to actually break and be able to inhale the smelling salts

_No response_

TealSeer:
[BUG] [PLAYER ACTIONS]
## About the PR 

Fix the `if` in ampoule `attack` so it actually runs the actionbar if you use them on yourself

## Why's this needed? 

Fixes #18669

### DIFF
--- a/code/modules/chemistry/tools/ampoules.dm
+++ b/code/modules/chemistry/tools/ampoules.dm
@@ -23,12 +23,12 @@
 	if(expended || reagents.total_volume <= 0)
 		boutput(user, SPAN_ALERT("[src] is empty!"))
 		return
-	else if(target == user)
+	if(target == user)
 		boutput(user, SPAN_NOTICE("You crack open and inhale [src]."))
 	else
 		user.visible_message(SPAN_ALERT("[user] attempts to force [target] to inhale [src]!"))
-		SETUP_GENERIC_ACTIONBAR(user, target, 3 SECONDS, /mob/proc/inhale_ampoule, list(src, user), src.icon, src.icon_state, null, \
-			list(INTERRUPT_MOVE, INTERRUPT_ATTACKED, INTERRUPT_STUNNED, INTERRUPT_ACTION))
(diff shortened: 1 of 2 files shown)


code/modules/chemistry/tools/ampoules.dm
```diff
@@ -23,12 +23,12 @@
...
 		return
-	else if(target == user)
+	if(target == user)
 		boutput(user, SPAN_NOTICE("You crack open and inhale [src]."))
 	else
 		user.visible_message(SPAN_ALERT("[user] attempts to force [target] to inhale [src]!"))
-		SETUP_GENERIC_ACTIONBAR(user, target, 3 SECONDS, /mob/proc/inhale_ampoule, list(src, user), src.icon, src.icon_state, null, \
-			list(INTERRUPT_MOVE, INTERRUPT_ATTACKED, INTERRUPT_STUNNED, INTERRUPT_ACTION))
+	SETUP_GENERIC_ACTIONBAR(user, target, 3 SECONDS, /mob/proc/inhale_ampoule, list(src, user), src.icon, src.icon_state, null, \
+		list(INTERRUPT_MOVE, INTERRUPT_ATTACKED, INTERRUPT_STUNNED, INTERRUPT_ACTION))
```
FlameArrow57, line 31: Oops lol

Action bars are just meant for applying to others (like how action bar is used for pills for others, but not for yourself)

Think the fix to maintain current functionality would be to just call `user.inhale_ampoule()` in the else if block, then probably need to add a if (user != src) check in the inhale_ampoule proc for the message

TealSeer: good catch ty

PR #18671, opened 2024-04-28
Buffs bladed gloves

colossusqw:
[LABEL][balance][clothing]
## About the PR 

This pull request buffs the bladed gloves traitor item in a few minor to significant ways:
- Decreases the single attack from 16 to 15 cut damage, while buffing the double attack from 11 to 15 damage
- Lowers the sound range of the unsheathing effect by 3 tiles
- Reduces the movement slowdown from the double strike gloves attack
- Changes bladed gloves inprints to those of regular insulated gloves, and makes the gloves themselves unscannable when worn

## Why's this needed? 

Bladed gloves entered the game in a overtuned state, being able to being stacked with a variety of items that caused them to be pretty good high end 'nerd' options. After this was remedied, however, it slowly became clear the bladed gloves aren't cut to part for a 3 TC traitor item,  and these small fixes should make it's usability floor a lot more consistent, as well as making it truer to it's agile yet discrete option identity.

- The reduced slowdown will allow players to make better use of the glove's inconspicuous nature in hit-and-run attacks, as well as better utilize it in a defensive runaway manner.
- The increased damage will give it a bit of the bite it was previously lacking, and make it not distinctively underperform when compared to other close cost-ed traitor items and even a few regular weapons.
- The lack of distinct prints will help users not fear having them on more often, since it was a major weakness of the item and dead giveaway that someone had one.

## Changelog 

```changelog
(u)colossusqw
(+)Bladed gloves now create standard insulated glove markings, have a smaller attack slowdown, and deal 15 damage(previous 11).
```

### DIFF
--- a/code/modules/forensics/atom_forensic.dm
+++ b/code/modules/forensics/atom_forensic.dm
@@ -51,8 +51,7 @@
 		LAZYLISTINIT(src.fingerprints)
 
 		if (H.gloves) // Fixed: now adds distorted prints even if 'fingerprintslast == ckey'. Important for the clean_forensic proc (Convair880).
-			if (!H.gloves.no_prints) // Stops very sneaky gloves from creating prints
-				seen_print = H.gloves.distort_prints(H.bioHolder.fingerprints, TRUE)
+			seen_print = H.gloves.distort_prints(H.bioHolder.fingerprints, TRUE)
 		else
 			seen_print = H.bioHolder.fingerprints
 
--- a/code/modules/forensics/atom_forensic.dm
+++ b/code/modules/forensics/atom_forensic.dm
@@ -51,7 +51,7 @@
 		LAZYLISTINIT(src.fingerprints)
 
 		if (H.gloves) // Fixed: now adds distorted prints even if 'fingerprintslast == ckey'. Important for the clean_forensic proc (Convair880).
-			if (!(H.gloves.no_prints)) // Stops very sneaky gloves from creating prints
+			if (!H.gloves.no_prints) // Stops very sneaky gloves from creating prints
 				seen_print = H.gloves.distort_prints(H.bioHolder.fingerprints, TRUE)
 		else
 			seen_print = H.bioHolder.fingerprints
--- a/code/modules/forensics/atom_forensic.dm
+++ b/code/modules/forensics/atom_forensic.dm
@@ -51,7 +51,8 @@
 		LAZYLISTINIT(src.fingerprints)
 
 		if (H.gloves) // Fixed: now adds distorted prints even if 'fingerprintslast == ckey'. Important for the clean_forensic proc (Convair880).
-			seen_print = H.gloves.distort_prints(H.bioHolder.fingerprints, TRUE)
+			if (!(H.gloves.no_prints)) // Stops very sneaky gloves from creating prints
+				seen_print = H.gloves.distort_prints(H.bioHolder.fingerprints, TRUE)
 		else
 			seen_print = H.bioHolder.fingerprints
 
--- a/code/obj/item/clothing/gloves.dm
+++ b/code/obj/item/clothing/gloves.dm
@@ -22,6 +22,7 @@ ABSTRACT_TYPE(/obj/item/clothing/gloves)
 	var/hide_prints = 1 // Seems more efficient to do this with one global proc and a couple of vars (Convair880).
 	var/scramble_prints = 0
 	var/material_prints = null
+	var/no_prints = FALSE // For extra sneaky prints
 
 	var/can_be_charged = 0 // Currently, there are provisions for icon state "yellow" only. You have to update this file and mob_procs.dm if you're wanna use other glove sprites (Convair880).
 	var/glove_ID = null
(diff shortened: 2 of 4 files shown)


code/obj/item/clothing/gloves.dm
```diff
@@ -22,6 +22,7 @@ ABSTRACT_TYPE(/obj/item/clothing/gloves)
 	var/hide_prints = 1 // Seems more efficient to do this with one global proc and a couple of vars (Convair880).
 	var/scramble_prints = 0
 	var/material_prints = null
+	var/no_prints = FALSE // For extra sneaky prints
```
TobleroneSwordfish, line 25: doc comment

code/modules/forensics/atom_forensic.dm
```diff
@@ -51,7 +51,8 @@
 		LAZYLISTINIT(src.fingerprints)
 
 		if (H.gloves) // Fixed: now adds distorted prints even if 'fingerprintslast == ckey'. Important for the clean_forensic proc (Convair880).
-			seen_print = H.gloves.distort_prints(H.bioHolder.fingerprints, TRUE)
+			if (!(H.gloves.no_prints)) // Stops very sneaky gloves from creating prints
```
TobleroneSwordfish, line 54:
```suggestion
			if (!H.gloves.no_prints) // Stops very sneaky gloves from creating prints
```

TobleroneSwordfish (approved):

code/obj/item/clothing/gloves.dm
colossusqw: aight

RubberRats (commented):
Seems good codewise. I'm personally a bit concerned about them not leaving prints, mostly because I don't think the ability to completely eliminate any trace you were at a crime scene leads to interesting play patterns. 
In my opinion, if the point is to make the gloves less obvious, then a more elegant solution would be to let them copy the appearance and fiber ID of another pair of gloves. This arguably addresses the issue better by preventing the scenario of "why is that engineer not wearing gloves?"

colossusqw:
> Seems good codewise. I'm personally a bit concerned about them not leaving prints, mostly because I don't think the ability to completely eliminate any trace you were at a crime scene leads to interesting play patterns. 
> In my opinion, if the point is to make the gloves less obvious, then a more elegant solution would be to let them copy the appearance and fiber ID of another pair of gloves. This arguably addresses the issue better by preventing the scenario of "why is that engineer not wearing gloves?"

While I do like the idea, that would go a bit farther than a simple rebalance of the item, more being a total rework that would involve a sprite rework and overlay change of the glove, as well as behavior for the appearance changes. Don't get me wrong, it seems interesting, but I'd also just like to see the item get some love first, and then if it turns out to really need a full overhaul, so be it.

imawxome:
I think bladed gloves should also get to do the knife strangle, with extra extra bleeding

TobleroneSwordfish:
A compromise for the fingerprint thing could be making them use a non-unique glove material string, the current "transparent fibres" is very obvious. Also the knife grab would be cool but would likely take a little bit of refactoring to make work.

colossusqw:
> A compromise for the fingerprint thing could be making them use a non-unique glove material string, the current "transparent fibres" is very obvious. Also the knife grab would be cool but would likely take a little bit of refactoring to make work.

I'm probably going to try to code those sometime for another PR, but yeah, i've tried it in the past and it didn't go super smoothly

code/obj/item/clothing/gloves.dm
```diff
@@ -508,15 +509,16 @@ ABSTRACT_TYPE(/obj/item/clothing/gloves)
 	name = "transparent gloves"
 	icon_state = "transparent"
 	item_state = "transparent"
-	material_prints = "transparent high-quality synthetic fibers"
+	material_prints = "insulative fibers"
+	no_prints = TRUE
```
TobleroneSwordfish, line 513: This is meant to be removed, no?

code/obj/item/clothing/gloves.dm
```diff
@@ -22,6 +22,7 @@ ABSTRACT_TYPE(/obj/item/clothing/gloves)
 	var/hide_prints = 1 // Seems more efficient to do this with one global proc and a couple of vars (Convair880).
 	var/scramble_prints = 0
 	var/material_prints = null
+	var/no_prints = FALSE // Specifically used so worn gloves cannot be scanned unless removed first
```
TobleroneSwordfish, line 25: .

code/procs/scanprocs.dm
```diff
@@ -552,7 +552,7 @@
 
 		if (!isnull(H.gloves))
 			var/obj/item/clothing/gloves/WG = H.gloves
-			if (WG.glove_ID)
+			if (WG.glove_ID && !(WG.no_prints))
```
TobleroneSwordfish, line 555: .

TobleroneSwordfish (changes requested):

code/obj/item/clothing/gloves.dm
colossusqw: as it stands what it does is make you need to remove the gloves before scanning which prints belong to it

PR #18684, opened 2024-04-30
Remove Unused Accesses from IDs

JORJ949:
## About the PR 
Removes the old armory access from the ID cards that still had it (AA, HoS, HoP) and changes an object I believe to be unused to instead just use security access.
Also removes access_mail from IDs

Anything "unused" but that had a mapping helper or was in ID computer wasnt touched

## Why's this needed? 
CE RCD can use any access in "get_all_accesses" and it probably shouldn't be able to make unused accesses. Specifically ones that can't be given out by the ID computer and are only available to a few roundstart roles.

### DIFF
--- a/code/obj/machinery.dm
+++ b/code/obj/machinery.dm
@@ -350,7 +350,7 @@
 	var/obj/machinery/door/d1 = null
 	var/obj/machinery/door/d2 = null
 	anchored = ANCHORED
-	req_access = list(access_armory)
+	req_access = list(access_security)
 
 /obj/machinery/noise_switch
 	name = "Speaker Toggle"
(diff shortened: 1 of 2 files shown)


ZeWaka (changes requested):
why not remove them entirely so they don't accidentally get used

JORJ949:
> why not remove them entirely so they don't accidentally get used

For a couple reasons:
-This was meant to be a quick fix for the RCD showing "Armory" as an access and confusing new players that it isnt armory
-Some places use the actual value of the access so just shifting all accesses below them up one would probably break a lot of stuff
-Just leaving blank spots in the list would feel wrong and may break some stuff
-I'd rather see them repurposed than just removed.

JORJ949:
Closing because there are much better solutions to this.

PR #18685, opened 2024-04-30
Fixes cog1 shutter blast control in kitchen

cheekybrdy:
## About the PR 
Fixes the kitchen shutter button on cog1 to change the correct blast door.

## Why's this needed? 
Bug fix good.


Sovexe:
changelog needs to be either removed or corrected

PR #3922, opened 2021-03-06
Removes stowaways from Manifest.

vode-code:
[FEATURE] [BALANCE]
## About the PR 
This PR adds a check to joining and changing name that disables manifest and all other NT data reference entries when stowaway; the manifest entry for being clone scanned remains.

## Why's this needed? 
Stowaways being on the manifest is odd and impedes the true stowaway experience.

## Changelog

```changelog
(u)Chickenish
(+)Stowaways are no longer in the manifest.
```

PR #18690, opened 2024-04-30
Jailbird trait changes

Chatauscours:
## About the PR 
- Moves the jailbird trait into the background category, so it is impossible to pick both stoaway and jailbird.
- Removes the major crimes from the jailbird (some of the sillier crimes were moved into the minor crimes).

## Why's this needed? 
- It makes no sense that someone who is not the records (stoaway) is somehow in the records and wanted.
- Major crimes like murder, arson and etc are often used by ill intentioned people to haze new secoffs.

## Changelog 

```changelog
(u)Chatauscours
(+)The jailbird trait has been moved into the background category of traits.
```

### DIFF
--- a/code/datums/datacore.dm
+++ b/code/datums/datacore.dm
@@ -168,35 +168,18 @@
 								"Excessive swearing",\
 								"Cutting in line.",\
 								"Tying the captain's shoelaces together.",\
-								"Forgetting the captain's birthday.")
-		S["mi_crim_d"] = "No details provided."
-		S["ma_crim"] = pick(\
 								"Grand theft apidae.",\
 								"Bee murder.",\
 								"Superfarted on the captain.",\
-								"Released the singularity.",\
-								"Stole the captain's spare ID.",\
-								"Arson, murder, jaywalking.",\
-								"Arson.",\
-								"Murder.",\
-								"Jaywalking.",\
 								"Skating right through the bounds of real-space. Wicked sick, but highly illegal.",\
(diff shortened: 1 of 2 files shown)


Glamurio:
I'm a bit torn on this. On one hand I agree that Jailbird should likely be a "Background" in terms of flavor, but on the other hand I kind of like how well it plays with the Stowaway background. I'd like to state that they're not in the records and wanted, but rather, wanted for off-station crimes. Maybe this should bring up a different discussion at large about how we categorize and limit trait categories instead.

Now, arguably, maybe those off-station crimes should in some capacity be reworked because they range from minor nuisances to "would literally be executed if this happened here".

Chatauscours:
> Also, arguably, maybe those off-station crimes should in some capacity be reworked because they range from minor nuisances to "would literally be executed if this happened here".

Yes, I removed the major crimes like arson, murder, bombing, etc

ZeWaka (commented):
Could be atomized a bit, i think the change is decently decided though - could be worth testmerging

MintyPhresh:
I like the idea behind the changes here-- but making Jailbird a background trait also blocks people from being Jailbird Pilots... and Jailbird Heavy Sleepers... And I think those two aren't worth killing. I'd rather see a reversion of #3922.

Glamurio:
This PR is sort of redundant now, no? Stowaway isn't a trait anymore so you'll always be able to pick it alongside now.

PR #18693, opened 2024-05-01
Deluxe Agent Card - HoP traitor item

JORJ949:
## About the PR 
Adds the "deluxe agent card", a reusable version of the agent card allowing for on the run identity crafting for 3 tc

## Why's this needed? 
HoP only has the AI law traitor items and thats kinda boring

## Changelog 

```changelog
(u)JORJ949
(*)The Head of Personnel can now buy a reusable "Deluxe" agent card. Also available to spy thieves and nuclear operatives.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -1046,7 +1046,7 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 	job = list("Captain", "Head of Personnel", "Research Director", "Medical Director", "Chief Engineer")
 	can_buy = UPLINK_TRAITOR
 
-/datum/syndicate_buylist/traitor/ai_disguised_module
+/datum/syndicate_buylist/traitor/syndicard_deluxe
 	name = "DELUXE Agent Card"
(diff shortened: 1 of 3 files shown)


TDHooligan:
might be too crazy? just swapping your ID freely, constantly, without being near an ID computer. could make it way too easy to entirely fuck up any and all comms with just a voice changer.

JORJ949:
> might be too crazy? just swapping your ID freely, constantly, without being near an ID computer. could make it way too easy to entirely fuck up any and all comms with just a voice changer.

If this does become an issue it would be a fairly simple change to just add a limit on how many uses it has

JORJ949:
> But i also think it does not need to be a HOP-exclusive item. This could fit pretty snuggly in the general traitor list, especially with a whopping 3TC cost.

This was made because HoP had no real exclusive items other than AI stuff which isn't great and also while it could possibly fit in the main traitor menu I wouldnt want to see it used too much because of the potential it has that TD pointed out.

JORJ949:
:/

FronkyGeniuZ:
Woah cool idea

PR #18698, opened 2024-05-01
Make unreadied players count for 1/2 a player for gamemode scaling purposes

TobleroneSwordfish:
[BALANCE] [GAMEMODES]
## About the PR 
Standardizes all the roundstart gamemode player counting logic and adds 1/2 the number of unreadied players to the total. Does not affect choosing modes like revs at low-pop.

## Why's this needed? 
Roundstart ready % hovers around 50-60% on RP3 these days, with most of the remaining players latejoining shortly afterwards. I think figuring out *why* and solving that is something worth doing but currently it means that there are about half as many antags as there should be, which is not good.
This should work as a bandaid to help the problem a bit in the meantime.
## Changelog 

```changelog
(u)LeahTheTech
(*)Unreadied players now contribute 1/2 of a readied player to roundstart gamemode scaling. (Yes this means more antags on average)
```

### DIFF
--- a/code/datums/gamemodes/assday_classic.dm
+++ b/code/datums/gamemodes/assday_classic.dm
@@ -23,12 +23,7 @@
 
 /datum/game_mode/assday/pre_setup()
 
-	var/num_players = 0
-	for(var/client/C)
-		var/mob/new_player/player = C.mob
-		if (!istype(player)) continue
-		if (player.ready)
-			num_players++
+	var/num_players = src.roundstart_player_count()
 
 	var/num_traitors = 1
 	var/num_wraiths = 1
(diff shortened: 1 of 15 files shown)


AXLFAN2000:
This is awesome

Glamurio:
> I think figuring out _why_ and solving that is something worth doing

It's honestly not too complicated of a reasoning. Off the top of my head I can think of 3 primary reasons:
- You want to play a special gimmick role (Daily or otherwise)
- You want to play a job but only under specific circumstances (maybe you want to play botanist, but not if you're the only botanist)
- You want to roleplay a very specific type of character that you've rolecrafted for a specific job, and if that job is not available, you'd rather play another character

As someone who likes to roleplay specific characters, I actually frequently latejoin myself.

Tarmunora (approved):

PR #18701, opened 2024-05-01
Stops unconscious people spinning

TobleroneSwordfish:
[BUG] [PLAYER ACTIONS] [BALANCE]
## About the PR 
Prevents ijkl spinning while knocked out or dead.

## Why's this needed? 
I know this is a bug that's been around since the beginning of time and is probably Soul but:
- How many times have you seen a player try to kidnap someone and get immediately outed because the "unconscious" body they're dragging behind them is spinning faster than the reactor turbine?
- This makes kidnapping and using knockout chems in any kind of sneaky way essentially impossible
- It removes any suspense/potential deceit from "is that person dead or just knocked out?" (we have a chemical that is literally a *Romeo and Juliet* reference but it can't be used to make someone seem dead!)

The only downside I can see is people maybe not being rescued from N2O clouds as often due to being mistaken for a corpse, but also if you see a horizontal person in a cloud of knockout gas you should be able to make a pretty good guess that they're still alive.

## Changelog 

```changelog
(u)LeahTheTech
(*)Unconscious people can no longer spin.
```


ZeWaka:
SOUL

Xkeeper0:
if both this and hearing people while KO'd get put in i'm rioting.

Tarmunora:
SOUL

Laboredih123:
need my SOUL

Wisemonster:
If this gets merged, someone could considering adding some kind of indicator next to people if they are asleep, like some zzzs next to their head (unless their are under the affects of the aforementioned chem that only makes you look like your dead), considering you can tell if someone is asleep or dead by examining them.

Laboredih123:
i think this is one thing where classic and rp really differ, would prefer if this was rp mode only

NightmareChamillian:
I want to be for this change but every time I've had someone walk directly by me when I was KO'd when **spinning** in an n20 cloud. Or had someone walk by me when I was in crit KO'd **spinning** on the ground. Or had someone walk straight past me when I was dead, in front of the cloner, soon to be rotting, while **spinning** on the ground came into mind. 

It's tough on the station, and this would absolutely take away what little power unconscious people have. If the "can hear when unconscious" PR gets merged I'd definitely be for this change but as it stands it's a hard sell.

Flaborized:
I already argued for it on the forums but yeah I think this is a Good Idea (for both servers) and that just because it's somewhat funny doesn't make it worth keeping automatically.

TobleroneSwordfish:
Closing in favour of #18746

PR #18707, opened 2024-05-02
offline mob/player indicator

Xkeeper0:
## About the PR 
image image

adds a little indicator for people who have disconnected.

(after a few minutes it dims a bit to show they've been off for a while)

## Why's this needed?

there's nothing like walking up to someone and going *wave howdy partner! how's it going! and then you notice they're not actually there and likely haven't been for the last 30 minutes. this solves that.

## Changelog

```changelog
(u)Zamujasa
(*)Players who disconnect now have a little "Zz" over their head.
```


DisturbHerb:
Kinda disagree with it being neon green, even if it pops out a lot easier from the background. Just a personal aesthetic quibble, but on the whole good feature.

Sbmhawk:
^ I think light blue would be better personally

Xkeeper0:
this is 100% programmer art, haha. some kind of light blue will probably be better

Glamurio:
Yes, this, please. This is genius.

TobleroneSwordfish:
I don't really like this because it's very meta and breaks a lot of admin gimmicks, buuut if we're going to have it then it should at least only be for humans so it doesn't mess with AIs, critters etc.

Xkeeper0:
the admin gimmicks thing is something that would be nice to change (could be like checking if the last ckey is an admin i guess)

it shouldn't mess with ais or critters other than showing when a living player was inside of it and no longer is, which i'm having trouble thinking of examples of this happening outside of disconnects. there's a "bug" where an ai that deploys to a shell makes the core show "Zz", but that one isn't intentional

it is very meta, though, and that's the intent -- helping people not have to alt-click on every person before they try interacting only to find out the person is offline

code/mob/living.dm
```diff
@@ -37,6 +37,7 @@
 
 	var/move_laying = null
 	var/has_typing_indicator = FALSE
+	var/has_offline_indicator = FALSE
```
Sovexe, line 40:
```suggestion
```

code/procs/mobprocs/offline_indicator.dm
```diff
@@ -0,0 +1,77 @@
...
+				src.UpdateOverlays(offline_a_while_indicator, OFFLINE_OVERLAY_KEY)
+
+/mob/living/remove_offline_indicator()
+	if (src.has_offline_indicator)
+		src.has_offline_indicator = FALSE
+		UnregisterSignal(src, COMSIG_MOB_DEATH)
+		src.UpdateOverlays(null, OFFLINE_OVERLAY_KEY)
+
+
+#undef OFFLINE_OVERLAY_KEY
```
Sovexe, line 77:
```suggestion
#define OFFLINE_OVERLAY_KEY "offline_indicator"

// Singletons for offline indicators
var/image/offline_indicator = image('icons/mob/overhead_icons32x48.dmi', "offline")
var/image/offline_a_while_indicator = image('icons/mob/overhead_icons32x48.dmi', "offline_a_while")
/mob/proc/create_offline_indicator()
	return

/mob/proc/remove_offline_indicator()
	return

/mob/Login()
	if (GetOverlayImage(OFFLINE_OVERLAY_KEY))
		remove_offline_indicator()
	. = ..()

/mob/living/Logout()
	if (key) //disconnected clients don't wipe key. Mobswaps do.
		create_offline_indicator()
	. = ..()

/mob/death()
	if (GetOverlayImage(OFFLINE_OVERLAY_KEY))
		remove_offline_indicator()
	. = ..()

/mob/living/create_offline_indicator(force = FALSE)
	if (isalive(src) || force)

		var/mob/living/silicon/ai/AI
		if (isAIeye(src))
			var/mob/living/intangible/aieye/EYE = src
			AI = EYE.mainframe
		else if (issilicon(src))
			var/mob/living/silicon/S = src
			AI = S.mainframe
		if (AI)
			// if you disconnect while in a shell or eye,
			// update the ai mainframe to show you're out
			AI.create_offline_indicator(TRUE)

		if (isvirtual(src))
			var/mob/living/carbon/human/virtual/V = src
			V.body.create_offline_indicator(TRUE)

		//used to protect against rejoins and leaves from erroneously triggering long logout
		src.logout_at = TIME
		var/logout_check = src.logout_at

		src.UpdateOverlays(offline_indicator, OFFLINE_OVERLAY_KEY)

		SPAWN(5 MINUTES)
			if (QDELETED(src)) return
			// check if they're still logged out after a while and update the overlay
			if (!client && key && logout_check == src.logout_at)
				src.UpdateOverlays(offline_a_while_indicator, OFFLINE_OVERLAY_KEY)

/mob/living/remove_offline_indicator()
	src.UpdateOverlays(null, OFFLINE_OVERLAY_KEY)

#undef OFFLINE_OVERLAY_KEY

```

Sovexe (changes requested):
Many of these checks aren't needed and can be done away with a simple guard in `logout` on the mobs key which is only cleared on disconnect and not mob swap. 

don't really need mutable appearances over images as you aren't actually doing any image manipulation to optimize in the first place. While they would work, this is only due to them being child types of image objects which should not be relied upon for future versions per the byond docs (yes I know we do rely on it elsewhere...one day it very well could bite us...)

instead of a signal we can just override `death` 

art should also change imo, not big on static green Z's

PR #18714, opened 2024-05-02
Clarion bar fixbatch

MintyPhresh:
## About the PR
- Fixes the Great Dark Spot in the northeast
- Fixes pixel offset on mail chute
- Rug is symmetrical
- Floors under doors are better
- Glass fab has glass pre-loaded, as intended

## Why's this needed? 
oopsie haha

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -10079,7 +10079,6 @@
 	dir = 4
 	},
 /obj/landmark/start/job/bartender,
-/obj/machinery/light/small/floor,
 /turf/simulated/floor/specialroom/bar,
 /area/station/crew_quarters/bar)
 "bop" = (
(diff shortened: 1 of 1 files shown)


MintyPhresh (commented):
more light moving

maps/clarion.dmm
```diff
@@ -38379,7 +38373,7 @@
 	},
 /obj/table/reinforced/bar/auto,
 /obj/machinery/glass_recycler/bar{
-	g_amt = 30
+	glass_amt = 20
```
mordent-goonstation, line 38376: Rather than a map override, if the recyclers having glass at roundstart is intended why not change the default value?

MintyPhresh: Good question. At the time I did this because each station had a different amount of glass available to the tender; I tried to keep the amount of glass the same per map just to cut down on the number of changes made to the game. On closer inspection, it's a bit ridiculous. I'll remove that from this PR and atomize the change

Issue #10954, opened 2022-09-22
Macho Man Randy Savage made a gold statue of me but didn't kill me


goonstation-issuebot:
### Labels

[BUG][TRIVIAL]

### Description

I escaped macho mans grasp at thr last second before turning to gold

### Steps to reproduce

1. Get grabbed by macho man and get turned to gold
2. last second before happening use jump gene to escape
3. you will get a gold clone of you but your voice text will appear over the gold clone

### Expected Behavior

Creation of gold statue is not tied to player death

### Additional Information & Screenshots

i got a screenshot but i dont know how to upload it...

Reported by: LeviathanB
On server: Goonstation 1 Classic: Heisenbee
Round log date: 2022-09-22-08-41
Reported on: 2022-09-22 08:50:46

TealSeer:
Fixed in #18723

PR #18723, opened 2024-05-03
Macho Touch won't turn target to statue if they escape grip

Tranquill6:
[BUG] [PLAYER ACTIONS]
## About the PR 
This fixes a bug with the Macho Touch ability, that the machoman mob has, where if the target escaped machoman's grip right before it finished executing, then they would still turn into a statue. The only changes are with the Macho Touch move itself and does not affect any other code.

Issue: #10954 

## Why's this needed? 
A small, but nice to have bug fix so people aren't getting turned to statues unnecessarily.

### DIFF
--- a/code/modules/antagonists/macho_man/abilities/gold_touch.dm
+++ b/code/modules/antagonists/macho_man/abilities/gold_touch.dm
@@ -47,14 +47,24 @@
 						if (N.client)
 							shake_camera(N, 6, 16)
 							N.show_message(SPAN_ALERT("<b>A blinding light envelops [holder.owner]!</b>"), 1)
-
 					playsound(holder.owner.loc, 'sound/weapons/flashbang.ogg', 50, 1)
-					qdel(G)
-					holder.owner.transforming = 0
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/macho_man/abilities/gold_touch.dm
```diff
@@ -47,14 +47,24 @@
...
-							H.transforming = 0
+					/// Whoever got grabbed broke the grip so don't statue them, revert state
+					if(G.affecting == null)
+						H.transforming = 0
+						H.layer = holder.owner.layer
+						H.pixel_y = 0
+						H.pixel_x = 0
+						holder.owner.transforming = 0
+						holder.owner.bioHolder.RemoveEffect("fire_resist")
+						holder.owner.verbs += /mob/living/carbon/human/machoman/verb/macho_touch
```
TobleroneSwordfish, line 59: These lines are common to both sides of the if statement so you can lift them out instead of duplicating them since order shouldn't matter here.

TobleroneSwordfish (changes requested):

PR #18731, opened 2024-05-03
Prevent wizards from blinking out of their own shuttle

### LINKED ISSUE #17955: Wizards can blink out of wizards den when not holding a staff
### Labels

[BUG][MAJOR]

### Description

Wizards can use the blink spell while not holding a staff to teleport to 2 zones around the wizards den, when holding the staff you cant do that.

### Steps to reproduce

1. Be wizard
2. Be at wizard den
3. Purchase blink spell
4. go to the right most wall of the wizards den
5. Make sure to not be holding a staff
6. Use blink
7. Now your on the other side, enjoy your champagne and watch the crew threw the security cameras because your stuck.

### Expected Behavior

I expected that I would not be able to teleport to inaccessible places.

From what I found you can only get to the owlery place above wizards den and the syndicate shuttle to the right of wizards den, but get stuck there since those places dont allow spells to be cast at all.

pgmzeta:
[player actions][bug]
## About the PR 
On restricted Z levels, ensure the wizard will end up on the wizard den shuttle.

## Why's this needed? 
Fix #17955

### DIFF
--- a/code/modules/antagonists/wizard/abilities/blink.dm
+++ b/code/modules/antagonists/wizard/abilities/blink.dm
@@ -47,6 +47,12 @@
 
 		var/turf/targetturf = locate(targetx, targety, holder.owner.z)
 
+		var/list/turf/passthrough_turfs = raytrace(get_turf(holder.owner), targetturf)
+		for(var/turf/T as anything in passthrough_turfs)
+			if (istype(T, /turf/cordon))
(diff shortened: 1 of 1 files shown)


Tarmunora (approved):
easier check would be to check for `/area/wizard_station` on the target turf (as blink is normally restricted from being cast in restricted Z except for the wizard station), but this should work too

PR #18758, opened 2024-05-04
Reduce RP-Mode unreadied player antag boost

pgmzeta:
[RP][Balance]
## About the PR 
Reduce the weight of unreadied players on RP from 2 to 4

## Why's this needed? 
sec and antags have no time to breathe and do actual RP during RP rounds

### DIFF
--- a/code/datums/gamemodes/gm_parent.dm
+++ b/code/datums/gamemodes/gm_parent.dm
@@ -307,7 +307,11 @@ ABSTRACT_TYPE(/datum/game_mode)
 				readied_count++
 			else
 				unreadied_count++
+#ifdef RP_MODE
+	var/total = readied_count + (unreadied_count/4)
+#else
 	var/total = readied_count + (unreadied_count/2)
+#endif
 	if (loud)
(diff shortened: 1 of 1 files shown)


NightmareChamillian:
very much so not a big fan. IMO sec is overstaffed and underworked already on RP.

Glamurio:
I disagree with this, the intensity of antags is very much dependent on the types of antags you roll, not simply the amount. RP can definitely handle more antags, so long as the antags follow proper escalation procedure.

Furthermore, the unreadied player patch was so that the amount of antags would reflect the amount that should be there if everyone readied, so in an ideal world this is how the rounds would look like.

TDHooligan:
idk. if gangs has proven anything it's that there's probably some room for this increased chaos on RP atm. 

not all the time, mind, but the amount of rounds on RP where nothing really goes on was imo kind of staggering. i can play for hours at prime time and not witness a single antagonism incident.

TobleroneSwordfish:
I think we'll leave it as is for now to see how it plays, security on RP are used to dealing with very few antags so it's natural that it will take time for them to adjust.

PR #18762, opened 2024-05-04
Adds second chem heater to Nadir's bar

MintyPhresh:
## About the PR 
Sure, it has one in the backroom chemlab, but that one's a pain to get to. Added a new one where the CheMaster was, and moved the CheMaster to the backroom.

Also adds a wall closet with a set of bartending gear, like what's seen in Atlas and Clarion. Some matching stuff on tables nearby has been removed to compensate.
image

## Why's this needed? 
Cocktail shakers fit in these things now, which makes them pretty important for chilling drinks to have ice in them.

### DIFF
--- a/maps/nadir.dmm
+++ b/maps/nadir.dmm
@@ -9233,7 +9233,8 @@
 	pixel_x = -10
 	},
 /obj/machinery/light/emergency,
-/obj/machinery/chem_master{
+/obj/machinery/chem_heater{
+	pixel_x = -4;
 	dir = 4
 	},
 /turf/simulated/floor/engine/caution/east,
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Going to be honest, I don't think anyone in the last 10 years has opened one of those wall lockers, probably best to just leave stuff out.

MintyPhresh:
That's fair. Do you care if there ones in the closet stay there? #18760 attempts to standardize these things but if having two fruit boxes in the map and one in the vendor is too much then there's not much point to having that pr

TobleroneSwordfish:
Err yeah I think I'd rather have stuff out on shelves in the back room rather than in wall closets, especially for a newbie friendly role like bartender. Mappers feel free to disagree tho :)

PR #18767, opened 2024-05-04
Optimisations for  UpdateOverlay

Laboredih123:
[PERFORMANCE][INTERNAL]

## About the PR 
Optimises the ClearSpecificOverlays and UpdateOverlays functions to be faster.
Adds AddOverlays, a function that handles only adding or replacing. Always expects an image.
Replaces a fuckton of instances of UpdateOverlays with AddOverlays or ClearSpecificOverlays.
Replaces some instances of ClearSpecificOverlays to be compacted into just 1 proc call.

## Why's this needed? 
Speed is good!


Sovexe:
you have several of these which are odd

`src.ClearSpecificOverlays(null, ...`

Can you share your tracey graphs comparing the performance?

Laboredih123:
> you have several of these which are odd
> 
> 
> 
> `src.ClearSpecificOverlays(null, ...)`
buh, if so will fix although it doesnt affect the function. and for the tracy thing i dont as those apply for a different branch but these procs were decently faster due to law of big numbers. ill get a new tracy when i can.

Laboredih123:
image
image
not totally comparable as the cpu likes to vary speed, but eh.

code/atom/UpdateOverlays.dm
```diff
@@ -140,63 +161,132 @@ ClearSpecificOverlays(1, "key0", "key1", "key2") 	//Same as above but retains ca
...
 		if(index <= length(src.overlays))
 			src.overlays.Cut(index, index+1) //Fuck yoooou byond (this gotta be by index or it'll fail if the same thing's in overlays several times)
 		else
 			stack_trace("Overlays on [identify_object(src)] were modified by non-UpdateOverlays method.")
 
-		prev_data[P_INDEX] = 0
 		for(var/ikey in overlay_refs) //Because we're storing the position of each overlay in the list we need to shift our indices down to stay synched
 			var/list/L = overlay_refs[ikey]
-			if(L?.len > 0 && L[P_INDEX] >= index)
+			if(L?[P_INDEX] >= index)
```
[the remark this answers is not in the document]
Laboredih123: the list will always be either null or have a length of P_ILEN. If the ? passes then it is guaranteed that we have a list with a valid length. So length was nothing more than a fancy null checker in this case.

code/datums/mutantraces.dm
```diff
@@ -2385,9 +2385,9 @@ TYPEINFO(/datum/mutantrace/pug)
 	proc/snore()
 		playsound(src.mob, 'sound/voice/snore.ogg', rand(5,10) * 10, 0, 0, src.mob.get_age_pitch(), channel=VOLUME_CHANNEL_EMOTE)
 		. = list("<B>[src.mob]</B> snores.", "<I>snores</I>")
-		src.mob.UpdateOverlays(snore_bubble, "snore_bubble")
+		src.mob.AddOverlays(snore_bubble, "snore_bubble")
```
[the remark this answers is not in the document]
Laboredih123: nah, itll always be set to an image.

code/atom/UpdateOverlays.dm
```diff
@@ -140,63 +161,132 @@ ClearSpecificOverlays(1, "key0", "key1", "key2") 	//Same as above but retains ca
...
+
+	#if DM_VERSION > 515
+	#warn "okay now remove this"
+	#endif
+	#if DM_VERSION < 515 && !defined(SPACEMAN_DMM)
+	for(var/ikey in overlay_refs)
+		if(ikey != key && overlay_refs[ikey][P_INDEX] && overlay_refs[ikey][P_ISTATE] == hash)
+			// logTheThing(LOG_DEBUG, null, "Attempt to add duplicate overlay appearances on [identify_object(src)] with keys [key] and [ikey].")
+			I.layer += 0.0000001 * rand()
+	#endif
```
ZeWaka, line 222: we don't support below 515 anymore, this can be removed

ZeWaka: otherwise looks good, if others want to review you can limit the commits to look at

Laboredih123: done

ZeWaka (approved):

PR #18769, opened 2024-05-05
mineral magnet computer no longer fails to open when magnet destroyed, plus bugfix for what happens when it does

### LINKED ISSUE #18702: Magnet Control PCs are unusable if the magnet is destroyed that it's on
### Labels

[BUG][MAJOR]

### Description

^^

### Steps to reproduce

1. Have magnet blow up or somethin
2. Attempt to access mining magnet PC
3. Just hand sprite on PC, nothing else.

### Expected Behavior

Return to magnet selection menu

Reported by: Bomb_Sirens
Client version: 515.1636
Active test merges: #18435, #18561, and #18697
Round log date: 2024-05-01-23-30
Reported on: 2024-05-02 00:38:24
Map: KONDARU

MeggalBozale:
## About the PR 

mineral magnet computers refuse to open the UI due to a weird ui check status call on... the magnet itself, for whatever reason it had to be. still inoperable without power, of course.

fixes bugs implied by this change when a magnet is destroyed after pulling or if explode during/after pull with UI open. fixes other bug where damage taken by magnet for overuse would not destroy the magnet due to being flagged as active. adds check_should_die() proc to qdel the magnet if its health is null or 0 to be SUPER DUPER SURE!!!!!

## Why's this needed? 

fixes #18702, you should be able to use and repair the thing dammit! AND IT SHOULD BREAK IN THE FIRST PLACE AHH


MeggalBozale:
thats a lotta autorequest for codeownership

code/obj/mining.dm
```diff
@@ -566,6 +567,13 @@
 		if (src.active)
 			src.UpdateOverlays(src.active_overlay, "magnet_active")
 
+	// Sanity check to make sure we gib on no health
+	proc/check_should_die()
```
pgmzeta, line 571: should this send a forced update to the ui?

MeggalBozale: it wouldnt do anything if it did

code/obj/mining.dm
```diff
@@ -692,6 +701,8 @@
 		.["time"] = TIME
 
 	ui_act(action, params)
+		if (src.check_should_die())
+			return
```
TobleroneSwordfish, line 705:
```suggestion
```
Things should explode when they are exploded, not when the UI code executes. If this check is necessary then look into why and fix it.

TobleroneSwordfish (changes requested):
Probably doesn't need a changelog

MeggalBozale:
fair with the changelog, and yeah prolly doesnt need that condition. just lemme verify it doesnt break

MeggalBozale:
Still works as intended

PR #18772, opened 2024-05-05
Gives Clarion ranch a roundstart spawner, and unobstructs the back door.

cheekybrdy:
## About the PR 
Minor clarion ranch fixes, it lacked a roundstart spawner, now has one planter, and the back door is no longer obstructed by a incubator.

## Why's this needed? 
Fixes good.


TobleroneSwordfish (changes requested):
Seems to include a lot of kitchen and bar changes, not intended?

JORJ949:
Honestly might be better to have the back door moved one tile left so the chickens dont break out because some random staffie with aa touched the door

cheekybrdy:
closed due to mapping conflicts I'm better off remaking

PR #18773, opened 2024-05-05
Revolution rounds end faster on brigging the last revolutionary head

FlameArrow57:
[GAME MODES][BALANCE][QoL]

## About the PR 
Makes it so that after buckle-cuffing the last revhead in a brig cell, the round will immediately end in a crew victory

(Not sure if there should be some sort of grace period after, but considering rev rounds end immediately after killing the last revhead, and usually the only way a revhead is brigged is close to the same circumstances, I thought maybe it would be okay)

Possibly a bug fix, seems brig endings don't occur when they should be? Only other victory check was through a game mode process, not sure if there's issues with that maybe, but that seems to occur 30 seconds.

## Why's this needed? 
Brig endings for rev rounds are supposed to be a possible ending, but (correct me if I'm wrong) don't seem to happen too often because when brigging the last revhead, it seems almost like it's bugged and it takes too long for the round end. This would hopefully make brig endings more common

## Changelog 

```changelog
(u)FlameArrow57
(*)Revolution rounds now properly end immediately after buckle-cuffing the last revolutionary head in a brig cell.
```

### DIFF
--- a/code/obj/stool.dm
+++ b/code/obj/stool.dm
@@ -426,6 +426,12 @@ TYPEINFO(/obj/stool/wooden)
 		to_buckle.set_clothing_icon_dirty()
 		playsound(src, 'sound/misc/belt_click.ogg', 50, TRUE)
 		to_buckle.setStatus("buckled", duration = INFINITE_STATUS)
+
+		if (istype(ticker?.mode, /datum/game_mode/revolution) && to_buckle.mind.get_antagonist(ROLE_HEAD_REVOLUTIONARY))
+			to_buckle.update_canmove()
+			SPAWN(1 SECOND) // so no abrupt ending
+				ticker.mode.check_win()
+
 		return TRUE
 
 	unbuckle()


FlameArrow57:
Also might ask for a testmerge, just to make sure this works properly, since harder to test on local

TobleroneSwordfish (approved):
Testmerge would likely be pointless given how rarely this happens, code looks good

code/obj/stool.dm
```diff
@@ -426,6 +426,12 @@ TYPEINFO(/obj/stool/wooden)
 		to_buckle.set_clothing_icon_dirty()
 		playsound(src, 'sound/misc/belt_click.ogg', 50, TRUE)
 		to_buckle.setStatus("buckled", duration = INFINITE_STATUS)
+
+		if (istype(ticker?.mode, /datum/game_mode/revolution) && to_buckle.mind.get_antagonist(ROLE_HEAD_REVOLUTIONARY))
```
TobleroneSwordfish, line 430:
```suggestion
		if (istype(ticker?.mode, /datum/game_mode/revolution) && to_buckle.mind?.get_antagonist(ROLE_HEAD_REVOLUTIONARY))
```

PR #18782, opened 2024-05-05
Quality of life fixes for the Clarion operating theatre

FireKestrel:
[QOL] [MAPPING]
## About the PR 
- Changes the placement of a lot of things in the Clarion operating theatre. See the screenshot for a comparison.
- Removes the two chairs. 
- Removes the third operating table and it's corresponding drain and operating computer.
- Adds a second IV drip.
- Moves the anesthetics locker from the lobby into the operating theatre.
- Adds a second set of surgical tools.
comparison

## Why's this needed? 
The current Clarion operating theatre has a few really irrating design flaws.
There's three operating tables but only one set of tools, only one IV drip (that spawns with two blood bags next to it for some reason??) and a mounted defib that can only reach two of the tables.
This fixes all of those things and just generally cleans up the way the room is laid out.

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -6463,14 +6463,13 @@
 /turf/simulated/floor/black,
 /area/station/hallway/primary/south)
 "aRw" = (
-/obj/cable{
-	icon_state = "1-2"
+/obj/machinery/computer/operating/small{
+	id = "OR1"
 	},
-/obj/machinery/manufacturer/medical,
-/turf/simulated/floor/bluewhite{
-	dir = 8
-	},
-/area/station/medical/medbay)
+/obj/iv_stand,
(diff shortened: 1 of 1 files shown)


Chatauscours (approved):
The hell, how are there 3 OR tables

FireKestrel:
> The hell, how are there 3 OR tables

I think it got mistakenly added at some point during the rework. Originally there was just two like most other maps.

Chatauscours:
Looks like it was like this way before the rework. 
Also, after taking a second look, I think the tool tables shouldn't be cluttered like that in the corner mixed with the lockers.

FireKestrel:
image
Yeah, having the tools closer to the tables is way more convenient and looks better. 
As for when table #3 got added, after some brief investigation I think it was when the small operating computers got introduced.

PR #18790, opened 2024-05-05
Adds the current daily job to job preferences

Glamurio:
[Respawning] [QoL]

## About the PR 
This PR removes the `/datum/job/daily` skip from procs such as `ResetAllPrefsToDefault()` in order to allow selection in job preferences. 

Naturally this keeps in place any existing job logic, meaning that once a day rolls around, the job gets replaced accordingly.
I'm unsure if there's any secret code that needs to be amended for this to work smoothly, but it worked locally.

## Why's this needed? 
We've got an issue with people latejoining right after roundstart instead of readying up, leading to skewed antag spawns. I believe adding daily jobs to roundstart eliminates one of the possible reasons people may have for latejoining.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Daily Jobs may now be selected as a Job Preference.
```


Tarmunora (changes requested):
Have you tested this change? I believe that as written, if you set a daily job as favourite/etc., it would remain there on future days until and unless one actually went to change job prefs again.

Glamurio:
> Have you tested this change? I believe that as written, if you set a daily job as favourite/etc., it would remain there on future days until and unless one actually went to change job prefs again.

I have tested it by actually quite literally waiting for Sunday to tick to Monday so I could be sure myself. If Boxer for instance is set to favorite and rotates away, that means the coming Monday your favorite will be set to Null and Dungeoneer will be selectable. I'm like 90% sure this works as intended.

Glamurio:
Okay, after doing some testing I'm now 95% sure it causes no problems. 

Example:
It's Monday 11 PM, so John Staffman favorites Dungeoneer. He plays a round and rolls Dungeoneer.
It's now Tuesday by the next round, John readies as per usual, he never opens his preferences. Even though Dungeoneer is still set as favorite, the game will simply ignore it and go for the second-next highest.
A week passes, it is Monday again. John has still never opened his preferences and now rolls Dungeoneer again.

PR #18791, opened 2024-05-05
Fixes Clarion Ranch

cheekybrdy:
## About the PR 
Moves the flusher activator to a position where standing in on the flusher isn't needed to activate aforementioned flusher, moved to wall on other side to not force entry into a pen when activated. Adds a Rancher roundstart marker which was previously missing, a single hydro tray and moves the door into maint left one tile to prevent obstruction and prevent pen escapes.

## Why's this needed? 
Fixes ranch issues


alicars:
Another thing I noticed missing is the mysterious sheep egg, which doesn't seem to be around. Thanks for fixes!

PR #18794, opened 2024-05-06
implement region allocator panel

Sovexe:
## About the PR 
* Implements a new admin tool, the Region Allocator Panel
* Front end for the region allocator system, allows admins to create custom allocations, load in prefabs defined in code, as well as load in a custom map file of their choosing
* Prefabs and uploads are given labels based on their file name, custom areas display their weakref from the `allocated_areas` list,
* Due to the area allocator making use of weakrefs all areas created through the panel gets a ref saved in the `region_allocator` `custom_admin_regions` list which acts as a persistent reference to prevent instant garbage collection

image

## Why's this needed? 
* Empower every admin to make use of the area allocator with ease


tgui/packages/tgui/interfaces/RegionAllocatorPanel.js
```diff
@@ -0,0 +1,91 @@
...
+            tooltip="Jump to Source"
+            onClick={() => gotoRegion(region.ref)} />
+          <Button
+            icon="arrows-to-dot"
+            tooltip="Jump to Center"
+            onClick={() => gotoRegionCenter(region.ref)} />
+        </>
+      }
+    >
+      {`${region.x}, ${region.y}, ${region.z} `}
```
ZeWaka, line 41: errant space at end?

ZeWaka (commented):
looks good

code/modules/admin/region_allocator_panel.dm
```diff
@@ -0,0 +1,96 @@
...
+		. += list(list(
+			"ref" = ref(region_ref),
+			"name" = region.name,
+			"x" = region.bottom_left.x,
+			"y" = region.bottom_left.y,
+			"z" = region.bottom_left.z,
+			"width" = region.width,
+			"height" = region.height,
+		))
+	. = list("regions" = .)
```
mordent-goonstation, line 29: This line's kinda weird. I get that it works, but using `.` as a temporary var as well as a return var in the same proc is probably not good practice.

mordent-goonstation (changes requested):
Broadly looks good, code-wise. Just one little thing that seems weird.

code/modules/admin/region_allocator_panel.dm
Sovexe: I can just define a temp list and make it a little more readable

PR #18797, opened 2024-05-07
Removes Max Trait Limit

frawhst:
## About the PR 
Sets the max number of selectable traits to `INFINITY`
# This might need a quick revert if it leads to extremely broken bullshit. There's too many traits to reasonably test all the combinations of, so I'm just letting this one fly.

## Why's this needed? 
Might be fun. Points should already be a limiter on traits, don't need a maximum number as well.

## Changelog 
```changelog
(u)aloe
(*)You can now select as many traits as you like.
```


444Portal:
ghost examining someone with a lot of traits is... yeah. possibly make traits not occupy a line each?
image

mordent-goonstation:
!merge_upstream

PR #18801, opened 2024-05-07
Syndicate Factory - Planet Prefab

EEASAS:
[MAPPING] 

## About the PR 
Adds a factory prefab, available on planets. It is taken by syndicates and barricaded.
The Danger: 4 melee syndicates, 1 rifle syndicate
The Loot: 1 Makarov, 2 Makarov magazines, 2 patch boxes, one is for brute, and other is for burns. 2 drug crate spawners, 4 10% chance loot spawns
The Defense: Barricades surrounded by barbed wire. Building is made out of reinforced walls with concrete floors
68747470733a2f2f616666656374656461726330372e626c6f622e636f72652e77696e646f77732e6e65742f6d6462322f696d616765732f3235323034373438332f32343638393534363934352f612f6173736574735f6d6170735f707265666162735f706c616e6574
(outdated image, check mapdiff2 in checks tab)

## Why's this needed? 
More artemis ruins are good, i think

No changelog because yes

### DIFF
--- a/assets/maps/prefabs/planet/prefab_syndicate_taken_factory.dmm
+++ b/assets/maps/prefabs/planet/prefab_syndicate_taken_factory.dmm
@@ -1,0 +1,698 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"a" = (
+/obj/decoration/junctionbox/junctionbox3{
+	pixel_y = 30
+	},
+/turf/unsimulated/outdoors/concrete,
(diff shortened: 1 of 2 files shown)


EEASAS:
i just heard they're not quite accepting artemis prefabs right now, maybe later

EEASAS:
nevermind

Azrun (approved):
In the future, these should probably be tagged in a manner such that it is easy to individually select the theming of a planet or terrainify. That way the presence of syndicate prefabs makes more contextual sense instead of being randomly dropped in.

PR #18804, opened 2024-05-07
Fix welding 2x2 pods from the top and right sides

pgmzeta:
[bug][player actions]
## About the PR 
use BOUNDS_DIST instead of IN_RANGE when checking if the welding action bar should continue

## Why's this needed? 
welding should work on all sides of vehicle

### DIFF
--- a/code/datums/actions/actions.dm
+++ b/code/datums/actions/actions.dm
@@ -946,8 +946,6 @@
 	var/atom/movable/target = null
 	/// what string is broadcast once the action bar finishes
 	var/end_message = ""
-	/// what is the maximum range target and owner can be apart? need to modify before starting the action.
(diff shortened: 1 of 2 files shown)


Tarmunora:
Would it work to change the IN_RANGE checks to a check that uses BOUNDS_DIST instead?

pgmzeta:
> Would it work to change the IN_RANGE checks to a check that uses BOUNDS_DIST instead?

More invasive change (need to modify the base welding action) but i will test

pgmzeta:
> Would it work to change the IN_RANGE checks to a check that uses BOUNDS_DIST instead?

working for vehicles & tested successfully with other welding action bar uses (lockers/airlock/singulo equip/salvager decon/crates)

PR #18805, opened 2024-05-07
Heavily refactors /obj/item/parts and the limb rendering section of /mob/living/carbon/human/update_body()

MylieDaniels:
[Code Quality] [Internal] [Performance]
#### Purpose
This PR refactors /obj/item/parts and /mob/living/carbon/human/update_body() to increase the performance and long-term workability of parts, which are almost entirely the items that are seen as arms and legs in the game.

#### Methodology
- /obj/item/parts has become /obj/item/mob_part, and most of the variables jampacked into it have moved into /obj/item/mob_part/humanoid_part, a child for parts that can render on humanoid mobs. The separation there is largely for the next step, removing handHolder from mob/living/critter and just using limbs. Health and bones were entirely removed due to being unused, along with siemens_coefficient, which is very nearly unused and should be an atom property on brullbar arms if it is desired for them to continue acting as "one brullbar arm is a pair of insulated gloves, but only for APC wires and the shock proc".

- /obj/item/parts/human_parts is now /obj/item/mob_part/humanoid_part/carbon_part, which handles primarily dna and fingerprints on top of that.

- /obj/item/parts/robot_parts is now /obj/item/mob_part/humanoid_part/silicon_part, which also stores brute/burn damage, power drain, and anything else specific to cyborg parts. Note these are still attachable to humans in all previous cases.

- /obj/item/parts/artifact_parts is now /obj/item/mob_part/humanoid_part/artifact_part, a distinction that exists mostly for typechecking in cases where that may become useful in the future. These limbs and their icons were renamed and sorted more than the others, to allow them to work in the same naming schemes as carbon_parts.

- Parts now store a hand image and a limb image, which update and colorize when necessary. Instead of render code generating these every time the body is dirtied, these are altered as infrequently as possible and used from the stored image in obj/item/mob_part/humanoid_part/proc/update_images()

- As a side effect, limbs now store their skin_tone more accurately and do not override it with the attached holder's skin tone. This is the first player-visible change.

- Limb decomposition is now in the limb as well, detaching it from the body itself - limbs still rot with the body, but detached and swapped limbs from a rotten corpse will remain rotten (affects appearance only, for now). Even previously, this data was stored on all limbs, and has actually been reduced in frequency of transmission and moved to a child (humanoid_part). This is the second player-visible change.

- Shoes now render on the layer MOB_HAND_LAYER2 with gloves, so that they consistently appear over the feet. This is a one line change and technically can be atomized but I think it's a solid solution to this bug. Feet render on MOB_HAND_LAYER1 with hands, so it seems logical.

- Lastly, a changeling bug I noticed was fixed. Not gonna lie, scope creep started to get me here. Changelings using Speed Regeneration as a mutantrace now regenerate their mutantraces limbs, not human ones.

- As all mob_part's calculate their own images, kudzu limbs and cow hooves now handle themselves on the part rather than having large special support in update_body()

#### Atomization
This is a massive PR, I am aware. However, I can only think of three small things to atomize out without negatively impacting the quality - these being the small fix to changeling's such that they now get the correct body part (for their mutantrace) when Speed Regenerating, making shoes render on a layer (which fixes numerous issues with rendering feet on top of or instead of shoes), and possibly moving the calls of decompose_limbs() outside of decomposition.dm to a new PR and just accepting that a few niche things will cause desynced rotten limbs.

I am willing to perform these atomizations, if necessary, and any others that can be reasonably performed if they are pointed out to me.

## Why's this needed? 
Not only is this a massive reduction in images created in the human rendering code, a refactor to help lighten the variables on a very common obj, and a large increase in customization available to part images, it also paves the way for the next step - unifying mob critters to also use mob_part's instead of handHolder and a dummyholder part.

## Changelog 

```changelog
(u)MylieDaniels
(*)Internal rework of limbs, from the items that make up an arm/leg to the rendering code that attaches them to a human. Report any issues and they will be promptly fixed.
```


Tarmunora:
arms/etc in secrets will need to be repathed

mordent-goonstation:
Not a full review, t putting some thoughts from the PR description:

> /obj/item/parts/human_parts is now /obj/item/mob_part/humanoid_part/carbon_part, which handles primarily dna and fingerprints on top of that.

Does this mean that silicon limbs don't have fingerprints? i.e. someone with cyborg arms does not leave fingerprints? I'm willing to believe this, just clarifying as it's not something I've thought about.

> Shoes now render on the layer MOB_HAND_LAYER2 with gloves

MOB_EXTREMITY_LAYER2, then?

Broadly, I would drop the `_part` suffix from anything deeper than `/obj/item/mob_part`, e.g. /obj/item/mob_part/humanoid_part/carbon_part` -> `/obj/item/mob_part/humanoid/carbon`

As a favor to me, a reviewer, could you self-comment some areas of the PR that you think need the most scrutiny, so that I can focus on them before my eyes glaze over from path renames?

ZeWaka (commented):
for future reviewing purposes, please do repaths in a separate commit - since we can filter those out when reviewing

MylieDaniels:
> Not a full review, just putting some thoughts from the PR description:
> 
> > /obj/item/parts/human_parts is now /obj/item/mob_part/humanoid_part/carbon_part, which handles primarily dna and fingerprints on top of that.
> 
> Does this mean that silicon limbs don't have fingerprints? i.e. someone with cyborg arms does not leave fingerprints? I'm willing to believe this, just clarifying as it's not something I've thought about.

No, these are the prints that show when scanning the limb with a forensic scanner, which (surprisingly to me too) have never factored into the prints you leave on touching something.

> > Shoes now render on the layer MOB_HAND_LAYER2 with gloves
> 
> MOB_EXTREMITY_LAYER2, then?

It certainly could be renamed, yes.

> Broadly, I would drop the `_part` suffix from anything deeper than `/obj/item/mob_part`, e.g. `/obj/item/mob_part/humanoid_part/carbon_part` -> `/obj/item/mob_part/humanoid/carbon`

That does make sense to me, but I'll hold off unless told to explicitly.

> As a favor to me, a reviewer, could you self-comment some areas of the PR that you think need the most scrutiny, so that I can focus on them before my eyes glaze over from path renames?

Gotcha - the parts that need full scrutiny are update_icon.dm and base_parts.dm

PR #18811, opened 2024-05-08
Using a plasma gas tank on a vehicle opens the parts panel

pgmzeta:
[qol][vehicles]
## About the PR 
When using any atmos tank on a vehicle, open the parts panel instead of bonking the tank against the vehicle.

## Why's this needed? 
Easier to replace air/fuel tanks in vehicles. Almost always the intent when using a gas tank on a vehicle

### DIFF
--- a/code/modules/transport/pods/vehicle.dm
+++ b/code/modules/transport/pods/vehicle.dm
@@ -172,6 +172,11 @@
 			src.keyed++
 			src.add_fingerprint(user)
 			return
+
+		if (istype(W, /obj/item/tank))
+			src.open_parts_panel(user)
+			return
+
 		..()
 
 		attack_particle(user,src)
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
I think this is the one part where you might actually want to hit a pod with it, gas tanks are common weapons and having them pop up a UI if you smack an unruly pod with one would be very disorienting.

Sbmhawk:
> I think this is the one part where you might actually want to hit a pod with it, gas tanks are common weapons and having them pop up a UI if you smack an unruly pod with one would be very disorienting.

Ive never seen someone use a plasma tank as a weapon, nor the large oxy tanks

pgmzeta:
Updated to only apply to plasma tanks

PR #18815, opened 2024-05-08
Buffs captain's office

Chatauscours:
[BALANCE]

## About the PR 
- Reinforces captain's locker.
- Makes captain's office access almost exclusive to the captain (and the spare). RP HoS no longer has access to cap's office. The Regional Director and NTSO units still have the access.
- The standard ID console can no longer give out captain's office access.
- Adds a new departamental ID console : Captain's. It requires captain's access to work on, and is the only ID console that can issue new IDs with captain's access.
- The new console has been placed in captain's offices on every in rotation map.

## Why's this needed? 
Makes captain a more serious role.

## Changelog 

```changelog
(u)Chatauscours
(*) Captain's office access is now exclusive to the captain and CentCom employees.
(*) Captain's locker is now reinforced.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -242,8 +242,7 @@ ABSTRACT_TYPE(/datum/job/command)
 
 	New()
 		..()
-		src.access = get_all_accesses()
-
+		src.access = get_access("Captain")
 
 	derelict
 		//name = "NT-SO Commander"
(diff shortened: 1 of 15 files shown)


Chatauscours:
Since with the current mechanics if the captain dies unclonably or cryoes, the slot opens up, I don't think the HoP needs captain's office access.
What could of course be done is abandonning the new sign up system, for a system where an acting captain could retrieve the spare from a safe on the bridge or other secure location should the captain cryo/die.

Sbmhawk:
I dont think having more stuff that the ID console cant do is a good idea.

The hos access already feels weird enough

PR #18817, opened 2024-05-08
Fixes a runtime caused by recoil attempting to trigger on non-mob users

Ilysen:
## About the PR 

`handle_recoil` now includes an `istype(user)` check to make sure it only runs on mobs.

## Why's this needed? 

`handle_recoil` expects a mob user, but the place where it's called doesn't do a type check to make sure the user is actually human. This leads to runtimes when a gun goes off without a mob user (such as when it's fired by a gun component); it *seems* like a harmless error, but it does spam the log, so it's a good thing to fix anyway.

## Testing

I jumped into a local server and fired a few different types of spawned-in gun myself and as part of a gun component assembly. No runtimes anymore, and everything seemed to work as intended.

### DIFF
--- a/code/obj/item/gun/gun_parent.dm
+++ b/code/obj/item/gun/gun_parent.dm
@@ -434,11 +434,9 @@ var/list/forensic_IDs = new/list() //Global list of all guns, based on bioholder
 
 	SEND_SIGNAL(user, COMSIG_MOB_CLOAKING_DEVICE_DEACTIVATE)
 
-	handle_recoil(user, start, target, POX, POY)
-
 	if (ismob(user))
-		var/mob/M = user
-		if (ishuman(M) && src.add_residue) // Additional forensic evidence for kinetic firearms (Convair880).
+		handle_recoil(user, start, target, POX, POY)
+		if (ishuman(user) && src.add_residue) // Additional forensic evidence for kinetic firearms (Convair880).
 			var/mob/living/carbon/human/H = user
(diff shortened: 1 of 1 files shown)


code/obj/item/gun/gun_parent.dm
```diff
@@ -434,11 +434,9 @@ var/list/forensic_IDs = new/list() //Global list of all guns, based on bioholder
 
 	SEND_SIGNAL(user, COMSIG_MOB_CLOAKING_DEVICE_DEACTIVATE)
 
-	handle_recoil(user, start, target, POX, POY)
-
 	if (ismob(user))
-		var/mob/M = user
-		if (ishuman(M) && src.add_residue) // Additional forensic evidence for kinetic firearms (Convair880).
+		handle_recoil(user, start, target, POX, POY)
+		if (ishuman(user) && src.add_residue) // Additional forensic evidence for kinetic firearms (Convair880).
```
Tarmunora, line 439:
```suggestion
		if (ismob(user) && src.add_residue) // Additional forensic evidence for kinetic firearms (Convair880).
```

Tarmunora (changes requested):
why not just put an `if(user)` check before doing the camera recoil in `handle_recoil`?

Ilysen:
could work! there actually is a sanity check in that proc for if the user exists, but not for if it's a mob or not. I can add in an istype to the proc easily enough.

ZeWaka (approved):

PR #18820, opened 2024-05-08
Makes action bars private when applying small/tiny items on items in other people's inventories

M-Earthfire:
[Balance][player actions][Input wanted]

## About the PR 

This PR makes it so when you use a small or tiny item on an item in someone elses inventory, the resulting action bar will be only visible for yourself. This specifically only applies to interactions when you open someone elses inventory (by clickdragging them onto yourself) and click on an occupied slot with another item in hand.

This change applies to most items with some exceptions. For these exceptions, a new flag in `item_function_flags` was created, `OBVIOUS_INTERACTION_BAR`. this flag is applied to items which were known from previous tests to be prone to massive abuse, e.g. mousetraps and beaker/dropper/spraybottles/sponges. Also, this applies to anything with storage datums, because else you would be able to pickpocket with a hidden action bar as well.

## Why's this needed? 

Applying items to other people's inventories was added to open up many interesting to funny interactions. Among these are e.g. slipping active grenades into peoples bags, contraband sticker onto mundane items, pills into drinks or planting contraband/evidence.

Due to abuse with beaker-splashing acid and mousetrap pipebombs gibbing people, this feature saw a massive nerf with rather long and highly visible action bars.

While this stomped the abuse into the ground, you nowaday also don't see most of the other interactions of this feature.

This PR aims to make interaction with other peoples inventories viable again while combating the most awfull abuse. Thus, this PR keeps the action bars in their known lenght and keeps them non-private for the worst offenders. 

I am very sure i missed some problematic items and thus want feedback on which items to exclude. I would also suggest a testmerge of this feature, since it opens up a lot of interactions.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Action bars when using many small/tiny items on items in other players inventory are now only visible to the user. 
```


code/modules/storage/storage.dm
```diff
@@ -69,6 +69,9 @@
 	if (istype(src.linked_item, /obj/item))
 		var/obj/item/I = src.linked_item
 		I.tooltip_rebuild = TRUE
+		// Items with storage datums attached shouldn't be able to be used for stealthy pickpocketing
+		if (!(I.item_function_flags & OBVIOUS_INTERACTION_BAR))
+			I.item_function_flags += OBVIOUS_INTERACTION_BAR
```
Tarmunora, line 74: while it's safe here because you are checking for the flag already, best practice is still to use bitwise operations when working with bitfields rather than addition/subtraction

Tarmunora (changes requested):
sponges need to be obvious-interaction as well

PR #18824, opened 2024-05-09
Fix UI-bug with trying to use boxes on other peoples items

M-Earthfire:
[Bug][Player actions]

## About the PR 

Modifies `/obj/item/attack_hand` so that when an item in someone else's inventory is affected, u_equip for that item is correctly called.

## Why's this needed? 

If you use a box on an item in someone else's pocket, the item would be moved to the box, but not clear the UI of the other person. that would create ghost-items in other peoples UIs that cannot be interacted with.  This was due to the item not being properly removed from the other mob. This fix should stop that now (hopefully without introducing new bugs :sheltersweat:).

### DIFF
--- a/code/obj/item.dm
+++ b/code/obj/item.dm
@@ -1192,20 +1192,21 @@ ADMIN_INTERACT_PROCS(/obj/item, proc/admin_set_stack_amount)
 		var/obj/container = src.loc
 		container.vis_contents -= src
 
-	if (src.loc == user)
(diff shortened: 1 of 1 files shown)


Tarmunora:
:shelterfrog: I think I'd rather we entirely block being able to steal things directly into storage items like this, as it uses the duration of applying the storage item, rather than stripping the clothing, which lets you strip stuff like jumpsuits and wizard gear much faster than intended

PR #18831, opened 2024-05-10
Adds nukie purchases to crew credits

alicars:
[UI]
## About the PR 

Adds all the things nukies can buy to the crew credits. This includes stuff from the weapon vendors, anything bought from a custom uplink, as well as the special stuff in the commander's uplink like the nuke teleporter. Yipee!

nukies credits 2

Please note that I have no idea what I'm doing so this code possibly sucks ass

If this seems OK I may try to add wizard spells next, tho I haven't yet figured out how to access the ability icons 

## Why's this needed? 
It's fun to see what everyone bought! Traitor, spief, and rev heads already have this functionality

## Changelog 

```changelog
(u)crunchwarp
(+)Nukies purchases are now visible in the crew credits
```


code/obj/item/uplinks.dm
```diff
@@ -432,6 +432,13 @@ Note: Add new traitor items to syndicate_buylist.dm, not here.
 					if (istype(antagonist_role) && !istype(I, /datum/syndicate_buylist/generic/telecrystal))
 						antagonist_role.purchased_items.Add(I)
 
+				if (src.purchase_flags & UPLINK_NUKE_OP)
+					var/datum/antagonist/nuclear_operative/antagonist_role = usr.mind?.get_antagonist(ROLE_NUKEOP)
+					if (antagonist_role == null )
+						antagonist_role = usr.mind?.get_antagonist(ROLE_NUKEOP_COMMANDER)
```
TobleroneSwordfish, line 438:
```suggestion
					var/datum/antagonist/nuclear_operative/antagonist_role = usr.mind?.get_antagonist(ROLE_NUKEOP) || usr.mind?.get_antagonist(ROLE_NUKEOP_COMMANDER)
```
You can do this due to boolean shortcutting, a little bit neater (also applies to other instances of this)

code/obj/submachine/weapon_vendor.dm
```diff
@@ -31,6 +31,7 @@
 	flags = TGUI_INTERACTIVE
 	object_flags = NO_GHOSTCRITTER //cry about it
 	layer = OBJ_LAYER - 0.1	// Match vending machines
+	var/track_credits = FALSE // True for nukies vendor, false for others
```
TobleroneSwordfish, line 34: This could just be an override of `ui_act` on the nukie vendor then

TobleroneSwordfish (changes requested):
Code looks good, thanks for doing this, a couple minor neatness points.

Gannets:
This seems cool! Happy to approve once those changes are in.

code/modules/antagonists/nuclear_operative/nuclear_operative.dm
```diff
@@ -8,6 +8,8 @@
 
 	var/static/commander_title
 	var/static/available_callsigns
+	var/list/purchased_items = list() //Used for adding a nukie's vendor purchases to crew credits. Items are tracked by whoever interacts with the vendor, so if the whole team gives their credits to the commander, the commander will have multiple entries in the crew credits!
```
Azrun, line 11: List of what? If we know the type let's use it.

code/modules/antagonists/nuclear_operative/nuclear_operative.dm
```diff
@@ -8,6 +8,8 @@
...
+	var/list/uplink_items = list() // Same but for custom uplinks
```
Azrun, line 12: Same

code/modules/antagonists/nuclear_operative/nuclear_operative.dm
```diff
@@ -99,6 +101,37 @@
 
 		. = ..()
 
+	get_statistics()
+		var/list/purchases = list()
+		// Add items purchased from the nukies weapon vendor
+		for (var/datum/materiel/purchased_item as anything in src.purchased_items)
```
Azrun, line 107: Why `/datum/materiel`?

code/obj/item/uplinks.dm
```diff
@@ -1275,6 +1289,13 @@ Note: Add new traitor items to syndicate_buylist.dm, not here.
 								new B.item3(get_turf(src))
 
 							B.run_on_spawn(A, usr, FALSE, src)
+
+							// Remember purchased item for the crew credits
+							var/datum/antagonist/nuclear_operative/antagonist_role = usr.mind?.get_antagonist(ROLE_NUKEOP)
+							if (antagonist_role == null )
+								antagonist_role = usr.mind?.get_antagonist(ROLE_NUKEOP_COMMANDER)
+							antagonist_role.uplink_items.Add(B)
```
Azrun, line 1297:
```suggestion
							antagonist_role?.uplink_items.Add(B)
```
Can we asset it isn't null?

alicars: I changed instances of this as well as the original code it was copied from (just above these lines)

code/modules/antagonists/nuclear_operative/nuclear_operative.dm
alicars: datum/materiel is what weapon_vendor.dm uses, and I believe everything added to purchased_items should be a materiel. (See lines 220 or 78 in weapon_vendor) Maybe something else would be better? Unsure

code/modules/antagonists/nuclear_operative/nuclear_operative.dm
alicars: I picked datum/syndicate_buylist for this for similar reasons to why I used datum/materiel-- its what uplinks.dm uses (line 689 there, for instance)

code/obj/submachine/weapon_vendor.dm
alicars: hopefully this new change looks right?

alicars:
Thank you all for the feedback! I'm learning many things

code/obj/item/uplinks.dm
Azrun: The `istype()` checks were sufficient for previous instances of this.

alicars: Ah, right. I definitely got this area mixed up with the others. Thank you! It's back to original now

code/obj/submachine/weapon_vendor.dm
```diff
@@ -213,6 +213,16 @@
 		STOP_TRACKING_CAT(TR_CAT_NUKE_OP_STYLE)
 		..()
 
+	ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
+		. = ..()
+		if (.)
```
TobleroneSwordfish, line 218: I would check for the param rather than the return value here, just in case someone adds another action to the parent proc.

TobleroneSwordfish (changes requested):

PR #18832, opened 2024-05-10
Adds typeinfo to mutantraces.

Laboredih123:
[PERFORMANCE][INTERNAL]
## About the PR 
Moves some shared vars into typeinfo to save speed. This means that rather than icon_states being run every time a mutantrace is made, it is run only once per unique typeinfo. This probably also saves a bit of memory I guess.

## Why's this needed? 
Speed good :)


TobleroneSwordfish (approved):
This is good, an improvement over my old "goddamnit byond" hack to fix server hangs from `icon_states` calls

PR #18833, opened 2024-05-10
Added checks to some common actions with Pods

### LINKED ISSUE #12672: remotely opening a pod wormhole
### Labels

[BUG][MINOR]

### Description

you can open a wormhole with a pod without being inside the pod even as a human

### Steps to reproduce

open the create wormhole window on a pod, leave the pod, click confirm on create wormhole and it does it

### Expected Behavior

I expected the window to close after I left the pod or atleast be unusable

Reported by: Matej672912
Round log date: 2023-01-08-10-30
Reported on: 2023-01-08 10:56:12
Map: KONDARU

Tranquill6:
[Bug][Vehicles]

## About the PR 
This adds some checks, to see if the player is in the pod, to things that most, if not all, pods do such as creating wormholes, using the communications array, or even unloading cargo. For example, this is to prevent a player from opening up the wormhole window, exiting the pod, and clicking confirm and still being able to create a wormhole despite not being in the pod anymore.

Fixes #12672 

## Why's this needed? 
Prevents players from cheesing pods by being able to unload their cargo, create a wormhole, or use the communications array without being inside of the pod anymore.

### DIFF
--- a/code/modules/transport/pods/communications.dm
+++ b/code/modules/transport/pods/communications.dm
@@ -97,7 +97,7 @@
 
 	proc/External()
 		var/broadcast = copytext(html_encode(input(usr, "Please enter what you want to say over the external speaker.", "[src.name]")), 1, MAX_MESSAGE_LEN)
-		if(!broadcast)
+		if(!broadcast || !istype(usr.loc, /obj/machinery/vehicle/)) // need to check if something wasn't entered or if user isn't in a vehicle
 			return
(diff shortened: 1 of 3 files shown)


TobleroneSwordfish (changes requested):
Why are some of these `istype` checks and not `src.ship` checks? That would mean you could get into another pod and still use the window of the old one.

Tarmunora (approved):

PR #18837, opened 2024-05-10
Makes the radio component less slow.

Laboredih123:
[PERFORMANCE][INTERNAL]
## About the PR 
Splits the default radio components in two to speed them up. The src.vars check was extremely slow, taking lots of milliseconds. This should speed it way up.

## Why's this needed? 
Once again, speed is good.


Laboredih123:
:(

_std/packets.dm
```diff
@@ -7,22 +7,44 @@ proc/check_for_radio_jammers(atom/source)
 		if (IN_RANGE(source, A, RADIO_JAMMER_RANGE))
 			return TRUE
 
-#define MAKE_DEFAULT_RADIO_PACKET_COMPONENT(conn_id, freq) src._AddComponent(list( \
+#define MAKE_DEFAULT_RADIO_PACKET_COMPONENT_NETID(conn_id, freq) src._AddComponent(list( \
 		/datum/component/packet_connected/radio, \
 		conn_id, \
 		freq, \
-		("net_id" in src.vars) ? src.vars["net_id"] : null, \
```
TobleroneSwordfish, line 14: Could you not just remove the ternary since DM gracefully handles missing associative list indexes? I assume it's the `in` operation that's taking the time here.

Laboredih123: vars is a special list and accesses are treated as accessing the var itself with :

Laboredih123: because of course it is

TobleroneSwordfish: Oouuughhh byond. Err okay, could you just move the net ID to be an argument to the macro instead of doing this hacky var list stuff in the first place?

Laboredih123: maybe, ill see when i get the chance

Laboredih123: works, just tested, committed, and pushed

TobleroneSwordfish (approved):

PR #18838, opened 2024-05-10
Replaces redundant static material processers with sheet extruders

TobleroneSwordfish:
[FEATURE] [GAME OBJECTS]
## About the PR 
Replaces the weird static material processers with a new machine called a sheet extruder that makes metal and glass sheets out of material bars. It uses the same sprite/animations as the old material processers and will automatically stack up any sheets it outputs.
The map replacement was done manually to ensure that each map still has material processers in all the places it should.
## Why's this needed? 
Currently the two ways of making sheets are to use a fabricator to make a huge pile of them one at a time or use the nanofab, and no-one should be forced to use the nanofab.
Also it's weird and pointless to have two machines that do the exact same thing, often found next to each other:
image
???

## Changelog 

```changelog
(u)LeahTheTech
(*)Replaced static material processers with new sheet extruders that make sheets out of metal and glass bars.
```

### DIFF
--- a/assets/maps/prefabs/underwater/manta/prefab_water_mantamining.dmm
+++ b/assets/maps/prefabs/underwater/manta/prefab_water_mantamining.dmm
@@ -553,7 +553,7 @@
 /turf/simulated/floor/shuttlebay,
 /area/prefab/tunnelsnake)
 "cj" = (
-/obj/machinery/processor,
+/obj/machinery/sheet_extruder,
 /turf/simulated/floor,
 /area/prefab/tunnelsnake)
 "ck" = (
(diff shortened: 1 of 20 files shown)


TobleroneSwordfish:
Yep, they're mechscannable now

pgmzeta:
A significant difference with the stationary material processor is that it can't be removed from the area. The refinery is going to become a crap-shoot on whether you can actually refine materials if this goes in.

TobleroneSwordfish:
Well you could say the same thing about robotics/cargo etc. People stealing stuff is part of the game, if it's a real problem we could just put two in each refinery but 🤷

ZeWaka:
> Well you could say the same thing about robotics/cargo etc. People stealing stuff is part of the game, if it's a real problem we could just put two in each refinery but 🤷

make it QM-orderable?

TobleroneSwordfish:
It is QM orderable!

PR #18840, opened 2024-05-10
Adds monkeys to the list of allowed mobs for pet carriers

Glamurio:
[Feature] [AI]

## About the PR 
Adds both `/mob/living/carbon/human/npc/monkey` and `/mob/living/carbon/human/monkey` to `allowed_mob_types` of the pet carrier.

Adds some simple AI which will make monkeys try and escape from the pet carrier, and upon successful escape, assault the person who trapped them.

## Why's this needed? 
More ways to deal with monkeys, also it's funny.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Recent studies show monkeys do, in fact, fit inside of pet carriers.
```

### DIFF
--- a/code/mob/living/carbon/human/monkeys.dm
+++ b/code/mob/living/carbon/human/monkeys.dm
@@ -253,6 +253,14 @@
 		. = ..()
 
 	ai_action()
+		if(istype(src.loc, /obj/item/pet_carrier))
+			var/obj/item/pet_carrier/carrier = src.loc
+			src.emote("scream")
+			src.emote("flip", TRUE)
+			if(ishuman(carrier.loc))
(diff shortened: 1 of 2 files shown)


Tarmunora (commented):
Two things:
* using a type check on the mob means that humanized monkeys would still fit in the carrier
* Also, other human mobs that are made into monkeys for whatever reason will not fit

Tarmunora:
Actually, after some discussion with the team, closing this because we neither want to open the cans of worms of treating monkeys as pets (considering how they are also used as test subjects, food, thrown into the burn chamber for good luck, etc.). Even with them violently trying to break out, it's not a direction we want to go down.

PR #18843, opened 2024-05-10
Tries to put crafted recipes into your hand before dropping

Glamurio:
[QoL] [Materials]

## About the PR 
Attempts `put_in_hand_or_drop()` on the person crafting an item if they're human and if the object in question can be held.

## Why's this needed? 
I'm done looking like a FOOL when I craft rods. It's not fair, it's not my fault, I'm perfectly capable crafting objects like a professional, and I won't sit idle while you suggest otherwise. This is my villain arc.

### DIFF
--- a/code/datums/actions/actions.dm
+++ b/code/datums/actions/actions.dm
@@ -302,7 +302,7 @@
 			return
 		owner.visible_message(SPAN_NOTICE("[owner] assembles \the [obj_name]!"))
 		var/obj/item/R = new obj_type(obj_turf)
-		if (!R.density && !src.constructible_check(R) && ishuman(owner))
+		if (isitem(R))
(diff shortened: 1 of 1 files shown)


code/datums/actions/actions.dm
```diff
@@ -299,6 +302,9 @@
 			return
 		owner.visible_message(SPAN_NOTICE("[owner] assembles \the [obj_name]!"))
 		var/obj/item/R = new obj_type(obj_turf)
+		if (!R.density && !src.constructible_check(R) && ishuman(owner))
```
Tarmunora, line 305: saner to just do an `isitem()` check, no?

Glamurio: Very true.

PR #18854, opened 2024-05-11
Change stats shown in gun tooltips

FlameArrow57:
[GAME OBJECTS][REWORK][QoL]

## About the PR 
This PR changes up the stats shown in weapon tooltips for guns

Currently guns display:
Bullet power - Lethality %

Changes them to display:
Max stamina dmg (only shown if not 0) (versus someone with no disorient/stun protection)
Damage - Armor pierce % (AP only shown if not 0)
Pod disruption - Chance % (Only shown if not 0)

## Why's this needed? 
Gun stats currently display bullet force and lethality %. Without looking at the code, these numbers are pretty arbitrary as to what they mean I think? Also not a lot of weapons have stun to them, think it makes the power stat a little confusing in some cases

Showing just damage I think would make it easier to know a gun's damage without needing to look at the wiki, or have game knowledge, of what it is. Also, a non-trivial number of guns have armor pierce and pod disruption chance, which isn't visible to anyone without the wiki or code diving. Think it would be fun to show them

## Changelog 

```changelog
(u)FlameArrow57
(+)Gun tooltips (from mouse hovering) now show exact projectile damage, as well as stamina damage, armor pierce, and pod system disruption chance if they exist
```


Tarmunora:
Doesn't this just make it impossible in turn to tell relative strengths of stun weapons (e.g.: the various taser types etc.) without code diving or wiki checking?

FlameArrow57:
> Doesn't this just make it impossible in turn to tell relative strengths of stun weapons (e.g.: the various taser types etc.) without code diving or wiki checking?

Yea kind second guessing that part now...

Looking at the code even know though, stun stat is a bit unclear as to what it does, seems to do stamina damage based off a clamped value (clamp(stun*4, P.proj_data.stun*2, stun+80). Maybe this could be turned into a proc (get_max_stamina_dmg()), and added to the tooltip to show stamina damage?

FlameArrow57:
Added stamina

image

image
(PRB rounds):

NightmareChamillian:
probably dumb question: how will this affect "spreader" projectiles, like pipeshot and birdshot?

FlameArrow57:
> probably dumb question: how will this affect "spreader" projectiles, like pipeshot and birdshot?

It'll show stats of an individual projectile in the spread. Seems to not work correctly tho for those types, will fix in next commit (ty)

code/obj/item/gun/gun_parent.dm
```diff
@@ -93,7 +93,21 @@ var/list/forensic_IDs = new/list() //Global list of all guns, based on bioholder
 	buildTooltipContent()
 		. = ..()
 		if(current_projectile)
-			. += "<br><img style=\"display:inline;margin:0\" src=\"[resource("images/tooltips/ranged.png")]\" width=\"10\" height=\"10\" /> Bullet Power: [current_projectile.power] - [current_projectile.ks_ratio * 100]% lethal"
+			var/stun
+			var/b_force = "Bullet damage: [current_projectile.damage]"
+			var/disrupt
+			if (current_projectile.stun)
+				stun += "Max stamina dmg: [clamp(current_projectile.stun * 4, current_projectile.stun * 2, current_projectile.stun + 80)]"
```
Tarmunora, line 100:
```suggestion
				stun += "Stamina: [clamp(current_projectile.stun * 4, current_projectile.stun * 2, current_projectile.stun + 80)] dmg"
```
matches the melee-stats part a bit better 🤏

Tarmunora (approved):

PR #18861, opened 2024-05-11
Allow mindhack cloning module to clone puritans

FlameArrow57:
[GAME OBJECTS][QoL][BALANCE]

## About the PR 
Allows the traitor mindhack module to clone puritans

## Why's this needed? 
Killing someone can already be a challenge in some cases, even more so for trying to perform a mindhack cloning set up. Not very fun when the person you manage to kill turns out to be a puritan

Also, if you aren't a medical doctor, such as if it's a spy round, and you get a mindhack module, by default you don't have access to check whether people are puritan or not

If not a good idea, PR at least sets up discussion/denial of the idea

Some possible alternatives if this doesn't sound good lorewise or any other reason:
-Only gives a reduced chance for cloning failure
-Installable module into the cloning computer

## Changelog 

```changelog
(u)FlameArrow57
(+)Mindhack cloning modules now allow cloning of puritans.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -950,7 +950,8 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 	item = /obj/item/cloneModule/mindhack_module
 	cost = 6
 	vr_allowed = FALSE
-	desc = "An add on to the genetics cloning pod that make anyone cloned loyal to whoever installed it."
+	desc = {"An add-on to the genetics cloning pod that makes anyone cloned loyal to whoever installed it. Its unique technology allows cloning of those
+				who are incompatible with usual cloning methods."}
 	job = list("Geneticist", "Medical Doctor", "Medical Director")
 
 /datum/syndicate_buylist/traitor/deluxe_mindhack_module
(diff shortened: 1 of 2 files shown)


Glamurio:
I like this. You almost never even see anyone bother setting one of these up because of how much they stand out and because of how much of a hassle it is in the first place, anything to make them more viable I think.

AXLFAN2000:
This makes sense to me in the same way you can't DNR out of thrall.  Good change

amylizzle:
People generally choose puritan because they don't want to play after death. I don't think forcing them to play as a mindhack is a good call. Perhaps just cause the mindhack clone to fail immediately instead, as it would if they'd DNRd?

RGBDeadSilent:
Yes, 100x yes to this. Think about it, you put up this absolutely horrible fight against the HoS himself, a perfect mindhack for the future. You toss him into the machine, hit clone... Grrrr, Explode. Nope, no HoS for you, just the wasted time fighting him and the possible knowledge of his implant revealing the location it happened at.

TobleroneSwordfish:
I do get the frustration buut puritans exploding is part of the game and sometimes things just go wrong. You can also check medical records if you have access to those to see who is a puritan and therefore not worth kidnapping.

Glamurio:
> People generally choose puritan because they don't want to play after death. I don't think forcing them to play as a mindhack is a good call. Perhaps just cause the mindhack clone to fail immediately instead, as it would if they'd DNRd?

I personally think it's a strange argument, people who are puritan get borged a lot of the time, which is functionally the same. It's not like puritan prevents a resurrection, you can even successfully clone them with a lot of skill and prep.

PR #18864, opened 2024-05-12
Yet another batch of hemotoxin changes

colossusqw:
[LABEL][chemistry][rework]

## About the PR
This PR does a few things:
- Make it so snakes are immune to hemotoxin
- Make it so small animals take more TOX from it, as well as have it progress 10x faster in them
- Make hemotoxin effects not as severe early on, but have them ramp up in severity, ramp up speed is based off of volume, and caps at 3x standard when above 24 units
- Reduce the message spam resulting from hemotox via adding a caffeine style message cooldown
- Make rattlesnakes only have 20 units of hemotoxin in them at death

## Why's this needed?
Hemotoxin isn't as immediately dangerous, or as chat spammy, while it's immediate threat scales with volume(scaling capped at 24 units to prevent exceedingly large doses being immediately deadly, and even still isn't as fast to kill you as something like cyanide, danger level of a single bite stays at about the same rate, while lower amounts are less dangerous than previously).

Besides that, rattlesnakes being immune is something that used to cause problems on resurected snakes, since they'd bleed out from the venom in them. And even if they hunted mice and whatnot, the effect of the venom on them was very underwhelming, and this fixes that.

## Changelog

```changelog
(u)Colossusqw
(+)Hemotoxin now has a wind up period that makes small amounts less dangerous, but has it's progression speed tied to volume. Small animals are much more affected by Hemotoxin.
(+)Rattlesnakes now only drop 20us of hemotoxin upon death, and are immune to it's effects.
```

### DIFF
--- a/code/modules/chemistry/Reagents-PoisonEtc.dm
+++ b/code/modules/chemistry/Reagents-PoisonEtc.dm
@@ -1427,15 +1427,15 @@ datum
 				M.take_toxin_damage(mult)
 
 				switch(counter += clamp(our_amt/10, 1, 3) * mult)
-					if (1 to 30) // The more there is, the faster it progresses
+					if (1 to 40) // The more there is, the faster it progresses
 						bleeding_mult = 0.5 * mult
 						if (probmult(8))
 							make_cleanable(/obj/decal/cleanable/blood/splatter,M.loc)
-					if (30 to 70)
+					if (40 to 80)
 						bleeding_mult = 1 * mult
 						if (probmult(14))
 							make_cleanable(/obj/decal/cleanable/blood/splatter,M.loc)
-					if (70 to INFINITY)
+					if (80 to INFINITY)
 						bleeding_mult = 1.5  * mult
 						if (probmult(24))
 							make_cleanable(/obj/decal/cleanable/blood/splatter,M.loc)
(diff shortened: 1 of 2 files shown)


code/modules/chemistry/Reagents-PoisonEtc.dm
```diff
@@ -1408,39 +1408,57 @@ datum
 			fluid_r = 210
 			fluid_g = 180
 			fluid_b = 25
-			depletion_rate = 0.3
+			depletion_rate = 0.2
```
Azrun, line 1411: Undocumented and unjustified.

Azrun (changes requested):

code/modules/chemistry/Reagents-PoisonEtc.dm
colossusqw: Oh, that probably came from me using the 0.2 version on my testing early on. To be honest i pretty much forgot about it and based a lot of the rest of the changes on it after, and caused it to be overall slower in the 10 to 20 unit bracket because of it. Do you think it better if i revert the depletion and make it slightly faster to maintain that 12 to 15 units to 300 blood level or keep as is?

colossusqw: Trying to keep the balancing similar with a 0.3 depletion version, shouldn't be much of a change if i get it right. Servers me for starting changes from previous ones without checking very well if i did anything else to the code beforehand.

PR #18866, opened 2024-05-12
Char liquefaction?

TobleroneSwordfish:
[CHEMISTRY]
## About the PR 
Adds a rough approximation of coal liquefaction/hydrogenation: heating char in the presence of hydrogen will liquify it into oil with some welding fuel, carbon and kerosene mixed in as byproducts, and emitting some CO2 in the process.

## Why's this needed? 
Oil is a pain to make, more ways to make oil are cool, gives char more of a use. Someone said it in the discord and I thought it sounded cool.

## Changelog 

```changelog
(u)LeahTheTech
(*)Added char liquefaction.
(+)write details here
```

### DIFF
--- a/code/modules/atmospherics/FEA_turf_tile.dm
+++ b/code/modules/atmospherics/FEA_turf_tile.dm
@@ -32,6 +32,7 @@ var/global/list/turf/hotly_processed_turfs = list()
 
 /// Return new gas mixture with the gas variables we start with.
 /turf/return_air()
+	RETURN_TYPE(/datum/gas_mixture)
 	// TODO this is returning a new air object, but object_tile returns the existing air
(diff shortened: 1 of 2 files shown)


MylieDaniels:
I worry that this could be very, very hard to prevent from just exploding.  Producing multiple highly flammable chemicals in an open puddle in presence of open flame seems like a very... for lack of a better word, janky way to do this.

TDHooligan:
this would be healthier, imo, to be considered 'hobo chem'  and not as a legitimate chemistry method to be used in balance equations 

buuuuut, we need more hobochem, so im incredibly down for this

Tarmunora:
Special effects on `raw_material` items themselves😔

PR #18868, opened 2024-05-12
replaces HoS offices with sec doors with command doors

cheekybrdy:
## About the PR 
Gives HoS offices command doors instead of security doors.

## Why's this needed? 
Clarion and Donut2 already had them like this and its the established standard for all other heads offices, always seemed like a weird irregularity.


JORJ949:
Might want to update your branch

TobleroneSwordfish (approved):

PR #18873, opened 2024-05-12
Add a bit more time for comm dish detection for comms console

FlameArrow57:
[GAME OBJECTS][QoL]

## About the PR 
Really minor, but for the communications console, the first time you login, run commaster, and then use the call command, it will give you an error message that the comms dish can't be connected, then a second after that comms were established. Typing call again, it will find it, and then ask you to enter your shuttle call reason

This just adds a bit more time so that you don't have to type call twice, just once, before entering the shuttle call reason

## Why's this needed? 
QoL

### DIFF
--- a/code/modules/networks/computer3/communication.dm
+++ b/code/modules/networks/computer3/communication.dm
@@ -87,7 +87,7 @@
 
 						if(!src.comm_net_id)
 							src.detect_comm_dish()
-							sleep(0.8 SECONDS)
+							sleep(0.9 SECONDS) // gives enough time for comm dish detection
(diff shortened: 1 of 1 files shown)


ZeWaka (commented):
Why not make the comms dish detection faster?

FlameArrow57:
> Why not make the comms dish detection faster?

Think the issue is with code below, 
/obj/machinery/power/data_terminal/proc/post_signal(), which does a ~300 item loop as a part of this, including lag checks. Not specific to shuttle calling I think

image

ZeWaka:
sure why not

PR #18878, opened 2024-05-13
Replace free upgrade for civilian cyborgs with a recharge pack

mordent-goonstation:
[Silicons][Balance]

## About the PR 

Changes the Efficiency upgrade that cyborgs get when they pick "Civilian" as their freebie module to a Recharge Pack upgrade.

A recharge pack is still the same "theme" as an efficiency upgrade, while also being manufacturable at roundstart. This will mean that (once people realize) they will stop picking civilian for the freebie then immediately changing the module.

## Why's this needed? 

Efficiency upgrades are the only free upgrade that is not able to be produced in robotics manufacturers at roundstart, due to having a requirement for a high-energy conductor.

As such, it is common for cyborgs to pick a Civilian module for the free Efficiency upgrade then have a human (roboticst, whoever) change their module for the one they actually wanted.

This is not an ideal feature of the freebie upgrade, which is (as far as I understand it) intended to include "required" gear for the cyborg to do its task well as a convenience (i.e. avoids them having to print their own upgrade that they would do 95%+ of the time because it's a no-brainer choice for that module).

This will improve our understanding of cyborg module choices, reducing the noise of people picking civilian at roundstart in our logs, and overall reduce some of the roundstart churn as people actually do pick the module they want to operate as.

## Changelog 

```changelog
(u)Mordent
(*)Civilian cyborgs get a free Recharge Pack instead of an Efficiency upgrade.
```

### DIFF
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -2348,10 +2348,10 @@
 					src.upgrades += new /obj/item/roboupgrade/spectro(src)
 			if("Civilian")
 				src.freemodule = 0
-				boutput(src, SPAN_NOTICE("You chose the Civilian module. It comes with a free Efficiency Upgrade."))
+				boutput(src, SPAN_NOTICE("You chose the Civilian module. It comes with a free recharge pack."))
 				src.set_module(new /obj/item/robot_module/civilian(src))
 				if(length(src.upgrades) < src.max_upgrades)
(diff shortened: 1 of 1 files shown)


444Portal:
Due to civilian generally using less power compared to other modules a recharge pack should more or less achieve what the efficiency module did for actual real civilan borgs. Woo!

Tarmunora (approved):

Sbmhawk:
D:

PR #18880, opened 2024-05-13
Remove Secbots from robotics cargo crates

JORJ949:
## About the PR 
Removes the securitrons from the crates full of robots from cargo listing

## Why's this needed? 
These are only ever annoying and gives robotics more to do

## Changelog 

```changelog
(u)JORJ949
(+)Removed securitrons from robotics crates.
```


frawhst (approved):
Fine with this.

PR #18885, opened 2024-05-13
Rework faustian chaplain a bit

TobleroneSwordfish:
[REMOVAL] [BALANCE]
## About the PR 
Removes the voluntary infernal contract signing mechanic from faustian contracts and decreases the forced signing actionbar from 4 to 3 seconds. Also reduces the number of souls required to summon the horseman drone from 15 to 8 and makes some of the contract effects a bit more directly Bad and Evil.

## Why's this needed? 
Nothing else in the game works like this, we have rules against helping obvious antagonists get stronger and yet there's this one mechanic that basically requires people to do that.
The voluntary signing mostly turns what should be an interesting and dangerous gimmick antag into a weird kind of meta situation where people are reluctant to stop them and the inevitable horseman drone feels more like a result of apathy than effort on the antag's part.

These changes should push chaplains more towards stunning and kidnapping people to force them to sign and make them y'know actually play as an antag.

## Changelog 

```changelog
(u)LeahTheTech
(*)Faustian contracts can no longer be signed voluntarily.
(*)Faustian contracts are now almost always hilariously bad for you.
(+)Reduced faustian contract force signing actionbar duration from 4 -> 3 seconds.
(+)Reduced the number of souls required to summon a horseman drone from 15 to 8.
```

### DIFF
--- a/_std/defines/infernal.dm
+++ b/_std/defines/infernal.dm
@@ -1,4 +1,4 @@
 //cost to purchase a new contract
 #define CONTRACT_COST 3
 //cost to summon one of the four horsemen
-#define HORSE_COST 15
+#define HORSE_COST 8
--- a/code/WorkInProgress/infernal_contracts.dm
+++ b/code/WorkInProgress/infernal_contracts.dm
@@ -547,6 +547,7 @@ obj/item/contract/macho
 			user.unequip_all()
 			boutput(user, "<span style=\"color:red; font-size:150%\"><b>Note that you are not an antagonist (unless you were already one), you simply have some of the powers of one.</b></span>")
 			user.mind?.add_antagonist(ROLE_MACHO_MAN, do_pseudo = TRUE)
+			user.reagents.add_reagent("krokodil", 50)
 
 		return 1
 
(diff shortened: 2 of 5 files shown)


Sbmhawk:
This accomplishes nothing other than making station-wiping much, much easier

TobleroneSwordfish:
Errr what? How?

Sbmhawk:
Working on editing my initial thing

TobleroneSwordfish:
Ah well, now that it's no longer an obvious voluntary action we could enforce not letting a chaplain steal your soul in the same way we enforce not deliberately getting revved or feeding yourself to a changeling. Hopefully that would also encourage a culture shift of people not putting up with faustian chaplains in the same way they do now.

TDHooligan:
this is definitely better than as is. i also wouldnt mind some faith mechanic working similarly to contracts (like Blessings) and then the faustian contracts being turned into ironic blessings.

Sbmhawk:
The drone is rare, yes

But when it does get summoned, it is almost always the result of the crew not doing anything because "friendly contract seller"

PR #18897, opened 2024-05-14
Replaces most instances of fake corpses with real corpses

Glamurio:
[Internal] [Secret] [QoL]

## About the PR 
Replaces most instances of `/obj/decal/fakeobjects/skeleton` with an instance of `/obj/mapping_helper/mob_spawn/corpse/human`.

Adds additional variables and behaviors to `/obj/mapping_helper/mob_spawn/corpse/human`, namely skeletonized, armless, headless, legless, try_buckle and `var/mob/living/carbon/human/corpse`.

Corpses via corpse spawner will now, on default, attempt to buckle themselves onto any `/obj/stool` when spawned.

Also adds several subtypes to `/obj/mapping_helper/mob_spawn/corpse/human` based on a variety of jobs.

I've manually replaced every instance and have taken great care in preserving the original vibe of most replacements. i.e. when replacing a skeleton in a clown suit, replace it with `/obj/mapping_helper/mob_spawn/corpse/human/clown` and so forth.

All Captain / HoP corpses have by default their pockets, bags, IDs and headsets removed, with the exception of the HoP in the customs shuttle. The same is true for all Security Officer corpses, safe for those that may spawn randomly with a 1% chance.

All corpse replacements inside the Owlery have their IDs removed.

Additionally, a few fake corpses remain due to perhaps unique stuff based around them, particularly on the ice moon.

Note:
There are a handful of maps, such as `ships.dmm`, `lunaport.dmm`, `xg6_fermidzone.dmm` and `ships.ancienttradevessel.dmm` that I was unable to edit due to missing files, perhaps a dev could do those instead.

## Why's this needed? 
The corpse spawner is a powerful tool to generate more dynamic content, and also allows players to interact with corpses, whereas previously they were merely decoration.

### DIFF
--- a/assets/maps/prefabs/planet/prefab_planet_tomato_den.dmm
+++ b/assets/maps/prefabs/planet/prefab_planet_tomato_den.dmm
@@ -98,8 +98,8 @@
 /turf/simulated/floor/grass/random,
 /area/noGenerate)
 "U" = (
-/obj/decal/fakeobjects/skeleton,
 /obj/item/saw,
+/obj/mapping_helper/mob_spawn/corpse/human/skeleton,
 /turf/simulated/floor/grass/random,
 /area/noGenerate)
 "W" = (
--- a/assets/maps/prefabs/space/prefab_beesanctuary.dmm
+++ b/assets/maps/prefabs/space/prefab_beesanctuary.dmm
@@ -1142,11 +1142,11 @@
 	tag = "icon-intact (NORTHEAST)";
 	color = "b4b4b4"
 	},
-/obj/decal/fakeobjects/skeleton,
 /obj/machinery/light/small/warm/very{
 	dir = 8;
 	pixel_x = -10
 	},
+/obj/mapping_helper/mob_spawn/corpse/human/skeleton,
 /turf/unsimulated/floor,
 /area/prefab/bee_sanctuary)
 "Jo" = (
@@ -1164,7 +1164,7 @@
 /turf/unsimulated/floor,
 /area/prefab/bee_sanctuary)
 "Jv" = (
-/obj/decal/fakeobjects/skeleton,
+/obj/mapping_helper/mob_spawn/corpse/human/skeleton,
 /turf/unsimulated/floor,
 /area/prefab/bee_sanctuary)
 "JF" = (
@@ -1727,7 +1727,7 @@
 /area/prefab/bee_sanctuary)
 "XU" = (
 /obj/decal/cleanable/blood/drip/high,
-/obj/decal/fakeobjects/skeleton,
+/obj/mapping_helper/mob_spawn/corpse/human/skeleton,
 /turf/unsimulated/floor,
 /area/prefab/bee_sanctuary)
 "Yd" = (
--- a/assets/maps/prefabs/space/prefab_candy_shop.dmm
+++ b/assets/maps/prefabs/space/prefab_candy_shop.dmm
@@ -395,15 +395,12 @@
 /turf/simulated/wall/auto/reinforced/jen/cyan,
 /area/prefab/candy_shop)
 "Df" = (
-/obj/decal/fakeobjects/skeleton{
-	desc = "The remains of a confectioner.";
-	name = "Carlisle"
-	},
 /obj/item/casing/derringer{
 	dir = 5;
 	pixel_x = -9;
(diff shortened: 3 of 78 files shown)


TheTyrant25:
does this mean that they'll rot? If they do, that seems like a bad thing.

Glamurio:
> does this mean that they'll rot? If they do, that seems like a bad thing.

No, the `no_miasma` variable defaults to `TRUE`.

Glamurio:
> Corpse spawns should all have ID/bag/pockets/headset/etc. stripped by default unless there's a good reason not to. Also, hard no on just slapping a free all-access ID card into a Z5 prefab like that.

I was worried that was going to be the consensus, but figured I'd try PRing it regardless. I'll adjust it.

Xkallubar:
Nice. Will corpses that used to be skeletons still look like skeletons?

Tarmunora (changes requested):
Also, because of how skeletons die, all the skeleton spawns are spawning as piles of bones and limbs. Which is... maybe not ideal?

DimWhat:
Places with a dead captain would give captain armour which may be a bit of a concern.

Glamurio:
> Also, because of how skeletons die, all the skeleton spawns are spawning as piles of bones and limbs. Which is... maybe not ideal?

Ah. Shucks. I forgot. I had it previously just set to skeletonize them, I'll likely revert it to that.

> Places with a dead captain would give captain armour which may be a bit of a concern.
I personally think that's cool, because I find it nice for people to find tangible loot so long as it's not broken, just depends on the extent. If it's a concern, there's either the option for it to only spawn partial loot, or alternatively have a "decayed" version of the armor that just yields a fraction of the stats from before.

flappybatpal:
This really seems like overkill for most purposes. Sure they are interactive but the corpses are only for scenery and you quickly run into loot issues with them being interactive.

TobleroneSwordfish:
Vaaaguely concerned about the performance impacts of swapping out a lot of fakeobject decals for mobs, and also not too thrilled about free captain armour given how much the syndie space suit spawns have been run into the ground. Things like captain gear should mean something when you see them, not just "oh that person went and looted that one prefab again"

MeggalBozale:
I love this change for the idea of aesthetics, and I think there's one simple solution to all the complaints people have about loot:
The corpses are so far into decomposition, that you can't take that off without taking the person with it.
Realistic? Probably not. But gross enough to be reasonable.

Also yeah tracy might be good. I'll post the difference here just because I'm curious.
(Of course, I won't have access to the secret submodule and thus some boot-up cost, but still it should show something)

Also, this looks like a forum thread might be helpful

MeggalBozale:
Tracy_mrp4DbtHAJ
Tracy_XRb8s05qTn
+1.3s of time to roundstart if im interpreting this right. In any case you can clearly see a performance impact without secret submodule

This was loading Atlas with the following flags enabled, and stopping the traces after 1 minute:
IM_REALLY_IN_A_FUCKING_HURRY_HERE
STOP_DISTRACTING_ME
TRACY_PROFILER_HOOK
MAP_OVERRIDE_ATLAS

Glamurio:
> Vaaaguely concerned about the performance impacts of swapping out a lot of fakeobject decals for mobs, and also not too thrilled about free captain armour given how much the syndie space suit spawns have been run into the ground. Things like captain gear should mean something when you see them, not just "oh that person went and looted that one prefab again"

Totally okay with making the captain armor non-lootable / making a more eroded version with much worse stats instead.

flappybatpal:
> Tracy_mrp4DbtHAJ Tracy_XRb8s05qTn +1.3s of time to roundstart if im interpreting this right. In any case you can clearly see a performance impact without secret submodule
> 
> This was loading Atlas with the following flags enabled, and stopping the traces after 1 minute: IM_REALLY_IN_A_FUCKING_HURRY_HERE STOP_DISTRACTING_ME TRACY_PROFILER_HOOK MAP_OVERRIDE_ATLAS

There's 392 /obj/decal/fakeobjects/skeleton (and subtypes), that's quite a lot of mobs to add with whatever level of processing mobs these corpse mobs would have.

Glamurio:
Alright, since I don't think we're going forward with this change, can I atomize the corpse spawner improvements for a separate merge?

PR #18898, opened 2024-05-14
Adds a new Speech Pro device to Ambassadors and Tourists

Glamurio:
[Feature]

## About the PR 
The Speech Pro is a hand-held device which can be used to allow individuals to communicate simple, pre-formatted phrases akin to MMORPG quick chats.

image
image

There is a 3 second cooldown between text output to prevent spam.

The Speech Pro has also been added to the robotics and medical fabricator.

## Why's this needed? 
Goofy device with which people are able to communicate, good for when you can't speak or nobody understands you.

Adapted from #18617 

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Added a new hand-held device that allows simple communication, available to Ambassadors and Tourists.
```


code/obj/item/device/speech_pro.dm
```diff
@@ -0,0 +1,48 @@
...
+		src.AddComponent(/datum/component/log_item_pickup, first_time_only=FALSE, message_admins_too=FALSE)
+		for(var/actionType in childrentypesof(/datum/contextAction/speech_pro)) //see context_actions.dm for those
+			var/datum/contextAction/speech_pro/action = new actionType()
+			if (action.phrase in src.phrases)
+				src.contexts += action
+
+	attack_self(mob/user as mob)
+		user.showContextActions(src.contexts, src, src.contextLayout)
+
+	proc/speak(var/string)
```
TobleroneSwordfish, line 36: This should have logging, ie "<user> says <phrase> using a speech pro"

code/datums/manufacturing.dm
```diff
@@ -1108,6 +1108,15 @@ ABSTRACT_TYPE(/datum/manufacture)
 	create = 1
 	category = "Clothing"
 
+/datum/manufacture/speech_pro
```
TobleroneSwordfish, line 1111: I'm going to say don't make it manufacturable for now because I feel like people are going to give them to mimes every round/expect mimes to use them.

_std/defines/speech_pro.dm
```diff
@@ -0,0 +1,12 @@
...
+#define SPEECH_PRO_SAY_HELP 3
+#define SPEECH_PRO_SAY_WHAT 4
+#define SPEECH_PRO_SAY_THX 5
+#define SPEECH_PRO_SAY_SRY 6
+#define SPEECH_PRO_SAY_GJ 7
+#define SPEECH_PRO_SAY_WAIT 8
+#define SPEECH_PRO_SAY_YES 9
+#define SPEECH_PRO_SAY_NO 10
+#define SPEECH_PRO_SAY_FOLLOW 11
+#define SPEECH_PRO_SAY_SP 12
```
TobleroneSwordfish, line 12: Do you actually need these defines or can you just do `concrete_typesof`?

TobleroneSwordfish (changes requested):

PR #18902, opened 2024-05-14
Make airbridges use glassblock floortiles

FlameArrow57:
[MAPPING][FEATURE]

## About the PR 
This PR makes map airbridges use cyan colored glassblock floortiles. Able to show both space (with shining stars) and parallax underneath

image

image

## Why's this needed? 
Thought it would be cool to add them to airbridges, if it doesn't look bad?

### DIFF
--- a/code/modules/animation/AnimationLibrary.dm
+++ b/code/modules/animation/AnimationLibrary.dm
@@ -1679,7 +1679,7 @@ var/global/icon/scanline_icon = icon('icons/effects/scanning.dmi', "scanline")
 
 /proc/animate_turf_slideout_cleanup(turf/T)
 	T.layer++
-	T.underlays.Cut()
+	T.reset_underlays()
 	var/obj/overlay/tile_effect/fake_fullbright/full_light = locate() in T
 	if(full_light)
 		qdel(full_light)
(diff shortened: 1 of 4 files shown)


444Portal:
i'd use the white ones (if they're avaliable) and maybe see about changing the lights. blue looks kind of bad with the white walls

FlameArrow57:
> i'd use the white ones (if they're avaliable) and maybe see about changing the lights. blue looks kind of bad with the white walls

There's these colors available

image

TobleroneSwordfish:
The cyan is a bit much, maybe try the grey or lighter blue ones?

FlameArrow57:
> The cyan is a bit much, maybe try the grey or lighter blue ones?

Grey:
image

More clear:
image

TobleroneSwordfish:
I like either of the "clear" ones at the bottom, they fit the colour scheme a lot better 👍

TobleroneSwordfish:
Okay actually, after talking with other devs we're going to say no to this because it doesn't really fit what an airbridge is supposed to be, and the fancy glass floors are more for an expensive looking viewing gallery than random semi-industrial airbridges.
It would be nice to resprite the airbridge floors/walls themselves to make them a bit more visually distinct though.

PR #18903, opened 2024-05-15
Add mutantrace ghosts

FlameArrow57:
[MUTANTRACES][FEATURE]

## About the PR 
This PR adds unique ghost sprites for each player selectable mutantrace

They act the same as regular human ghosts, except that the only way to have hair on the ghost is to be using the bald trait.

image

image

## Why's this needed? 
Thought it would be fun, more variety in ghosts

## Changelog 

```changelog
(u)FlameArrow57
(+)Added unique ghost sprites for each mutantrace when you die. If you want hair, be sure to select the Bald trait!
```


code/mob/dead/observer.dm
```diff
@@ -101,14 +101,45 @@
...
+		src.icon_state = "ghost-skeleton"
+		is_mutantrace = TRUE
+	else if ("roach" in P.traitPreferences.traits_selected)
+		src.icon_state = "ghost-roach"
+		is_mutantrace = TRUE
+	else if ("pug" in P.traitPreferences.traits_selected)
+		src.icon_state = "ghost-pug"
+		is_mutantrace = TRUE
+
+	var/mutantrace_hair = is_mutantrace && ("bald" in P.traitPreferences.traits_selected)
```
TobleroneSwordfish, line 121: This should also check for the hidden mutantrace hair admin option ideally.

code/datums/mutantraces.dm
```diff
@@ -290,6 +293,14 @@ ABSTRACT_TYPE(/datum/mutantrace)
 				src.blood_color_changed = TRUE
 
 		src.AH = M.bioHolder?.mobAppearance // i mean its called appearance holder for a reason
+		src.hair_custom_style = list()
+		src.hair_custom_style["hair_icon"] = src.AH?.customization_first.icon
+		src.hair_custom_style["hair_id"] = src.AH?.customization_first.id
+		src.hair_custom_style["beard_icon"] = src.AH?.customization_second.icon
+		src.hair_custom_style["beard_id"] = src.AH?.customization_second.id
+		src.hair_custom_style["detail_icon"] = src.AH?.customization_third.icon
+		src.hair_custom_style["detail_id"] = src.AH?.customization_third.id
```
TobleroneSwordfish, line 302: Why does this need to be duplicated here?

code/mob/dead/observer.dm
```diff
@@ -101,14 +101,45 @@
...
+		is_mutantrace = TRUE
+	else if ("skeleton" in P.traitPreferences.traits_selected)
+		src.icon_state = "ghost-skeleton"
+		is_mutantrace = TRUE
+	else if ("roach" in P.traitPreferences.traits_selected)
+		src.icon_state = "ghost-roach"
+		is_mutantrace = TRUE
+	else if ("pug" in P.traitPreferences.traits_selected)
+		src.icon_state = "ghost-pug"
+		is_mutantrace = TRUE
```
TobleroneSwordfish, line 119: Shouldn't this just be a var on the mutantrace?

code/mob/dead/observer.dm
```diff
@@ -411,6 +431,23 @@
...
+	else if (istype(src.mutantrace, /datum/mutantrace/skeleton))
+		O.icon_state = "ghost-skeleton"
+		is_mutantrace = TRUE
+	else if (istype(src.mutantrace, /datum/mutantrace/roach))
+		O.icon_state = "ghost-roach"
+		is_mutantrace = TRUE
+	else if (istype(src.mutantrace, /datum/mutantrace/pug))
+		O.icon_state = "ghost-pug"
+		is_mutantrace = TRUE
```
TobleroneSwordfish, line 449: ditto

TobleroneSwordfish (changes requested):
Sprites look good :+1:

code/mob/dead/observer.dm
FlameArrow57: This is from client, not mob mutantrace var

code/mob/dead/observer.dm
FlameArrow57: Could be, would be the only location it's used tho, considering the above comment

code/datums/mutantraces.dm
FlameArrow57: In human/proc/Equip_Job_Slots(), it sets the customization styles to new /datum/customization_style/none, so need to save the icons and ids. Tried messing around with that in a few places, figured here would be most appropriate

code/mob/dead/observer.dm
FlameArrow57: You'll have to let me know what variable that is, I don't know anything about admin vars x_x

TobleroneSwordfish: `hair_override`

code/mob/dead/observer.dm
TobleroneSwordfish: You can just get the mutantrace datum from their trait list though, surely?

FlameArrow57: Sorry, I'm still not understanding

If you are a /mob/new_player and start, mutantrace and hair_override vars won't be accessible?

code/mob/dead/observer.dm
FlameArrow57: If you are a /mob/new_player when this proc is started, hair_override var won't be accessible?

TobleroneSwordfish: Then just handle that case, it can only be added after they've joined anyway.

code/mob/dead/observer.dm
TobleroneSwordfish: They'll have the trait in their trait list, which contains a reference to the mutantrace type.

code/datums/mutantraces.dm
TobleroneSwordfish: Oh there's actually a var for this I think: `AH_we_spawned_with`

FlameArrow57:
Okay ready for re-review

TobleroneSwordfish (approved):

PR #18909, opened 2024-05-15
Fix runtime when disposing bunsen burner

TealSeer:
[BUG] [GAME OBJECTS] [RUNTIME]
## About the PR 

Add `START_TRACKING` to bunsen burner to match `STOP_TRACKING` in `disposing()`

## Why's this needed? 

image


TobleroneSwordfish (changes requested):
Buuut why is it being tracked? There's no by-type references anywhere I can see.

TealSeer:
the answer was "oops"

PR #18910, opened 2024-05-15
fix runtime on null area for airlock init

MeggalBozale:
## About the PR 
Discovered when I was initializing things with fabricator code to get their material names and tried an airlock

This just adds a simple null check before doing src.name = A.name to prevent this runtime, it should be fine to init these things in null areas

## Why's this needed? 
Makes it easier to check vars/icons of things which change after init when you dont need to give them the right conditions to init without runtimes

### DIFF
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -85,7 +85,7 @@ var/global/list/cycling_airlocks = list()
 	..()
 	if(!isrestrictedz(src.z) && src.name == initial(src.name)) //The second half prevents varedited names being overwritten
 		var/area/station/A = get_area(src)
-		if (!isnull(A)) // Fabricators
+		if (!isnull(A))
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (approved):
I kind of hate the fact that fabricators generate instances like that but this works.

code/obj/machinery/door/airlock.dm
```diff
@@ -85,7 +85,8 @@ var/global/list/cycling_airlocks = list()
 	..()
 	if(!isrestrictedz(src.z) && src.name == initial(src.name)) //The second half prevents varedited names being overwritten
 		var/area/station/A = get_area(src)
-		src.name = A.name
+		if (!isnull(A)) // Fabricators
```
mordent-goonstation, line 88:
```suggestion
		if (!isnull(A))
```

mordent-goonstation (approved):

PR #18914, opened 2024-05-15
Fix bug that doubled bites_left for vendor pizzas.

MylieDaniels:
[BUG][CATERING]
## About the PR 
Very simple, just fixes a one word typo from my pizza rework that caused vendor pizza slices to inherit from the base pizza slice instead of the worse vendor pizzaslice I had as their parent. I wanted to fix this by making bites_left inherit from the sliced parent / slices, but that means the humble orange would have like 6 bites or something, so I couldn't.
Also because Zewaka knows I will never make a `size/XS` PR, this also abstracts pizzas that shouldn't exist.

## Why's this needed? 
Whoops, I accidentally made a vendor pizza have 2 bites per slice, which made the 100c pizzas roughly comparable to high tier chef food for RP server hunger restoration.

### DIFF
--- a/code/modules/food_and_drink/pizza.dm
+++ b/code/modules/food_and_drink/pizza.dm
@@ -547,7 +547,7 @@
 	initial_reagents = list("bread" = 2, "juice_tomato" = 2, "cheese" = 1, "space_fungus" = 4)
 
 /obj/item/reagent_containers/food/snacks/pizzaslice/vendor
-	name = "slice of fresh basic pizza"
+	name = "slice of basic pizza"
 	desc = "A slice of base vendor pizza (you shouldn't see this)."
 	fill_amt = 1
 	bites_left = 1
--- a/code/modules/food_and_drink/pizza.dm
+++ b/code/modules/food_and_drink/pizza.dm
@@ -156,18 +156,18 @@
 	Eat(mob/M as mob, mob/user, by_matter_eater = TRUE)
 		if(!by_matter_eater)
 			return 0
-		var/obj/item/reagent_containers/food/snacks/pizza/unbaked_pizza = src.bake_pizza(TRUE)
+		var/obj/item/reagent_containers/food/snacks/pizza/bespoke/unbaked_pizza = src.bake_pizza(TRUE)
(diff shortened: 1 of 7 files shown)


code/modules/food_and_drink/pizza.dm
```diff
@@ -546,8 +546,8 @@
 	desc = "A cheesy pizza slice topped with fresh picked mushrooms."
 	initial_reagents = list("bread" = 2, "juice_tomato" = 2, "cheese" = 1, "space_fungus" = 4)
 
-/obj/item/reagent_containers/food/snacks/pizzaslice/standard
-	name = "slice of fresh basic pizza"
+/obj/item/reagent_containers/food/snacks/pizzaslice/vendor
+	name = "slice of basic pizza"
 	desc = "A slice of base vendor pizza (you shouldn't see this)."
```
ZeWaka, line 551: if they shouldnt see it then why isnt it abstract HUH

ZeWaka (changes requested):

code/modules/food_and_drink/pizza.dm
MylieDaniels: cant be because abstract tosses itself up into the parent for some reason, which causes a runtime with additional runtime checking on

ZeWaka: it intentionally does that, but why would it cause a runtime

MylieDaniels: because the parent (pizzaslice) is not uh intended to be abstract because it is created when slicing a bespoke pizza which i can repath into like pizzaslice/bespoke if you want and abstract pizzaslice, pizzaslice/vendor, and pizzaslice/standard

ZeWaka: i think that would make more sense

MylieDaniels:
Trying to locate the two things instantiating the now-abstract pizza parent... guh

ZeWaka (approved):
shouldn't it be

*bee*spoke?

https://github.com/tgstation/TerraGov-Marine-Corps/assets/4741640/09b395dd-5c18-44cc-888c-8906397b1382

PR #18926, opened 2024-05-16
Remove a bunch of maintenance cameras from kondaru

JORJ949:
## About the PR 
Remove a lot of maint cameras on kondaru, particularly around bridge

## Why's this needed? 
Crime is good.


JORJ949:
VERY incorrect branch, remaking

PR #18927, opened 2024-05-16
Remove some Kondaru maint cams

JORJ949:
## About the PR 
Removes cameras from the high-coverage maintenance tunnel that is bridge maintenance

## Why's this needed? 
Crime is good.


TobleroneSwordfish (approved):

444Portal (approved):
i think the radio room one can stay but i dont super care for it and the others going is a good thing

ZeWaka:
i think keeping half of them so its spotty coverage would be good

TTerc (approved):

PR #18939, opened 2024-05-16
Give HoPs the black market megaphone

alicars:
## About the PR 

Give traitor HoPs the ability to purchase the black market megaphone!

hop pr
HOP PR2

## Why's this needed? 

As Captain is non-antag only now, this item is only acquirable by latejoin VIP roles and spiefs, both of which are inconsistent and hard to pre-plan gimmicks around. 

"Does it make lore sense for the HoP to get this item?" My argument is that the HoP is the perfect role for coup-like gimmicks. The power hungry second-in-command is a long standing archetype. If they want to overshadow the captain, of course they'd have to turn to the black market to get a more powerful megaphone!

This also gives HoPs their own unique (at least, mostly) traitor item, which is neat! 

## Changelog 

```changelog
(u)crunchwarp
(+)Traitor HoPs can now purchase the Black Market Megaphone. Overthrow the Captain today!
```


TobleroneSwordfish:
Yea sure!

PR #18951, opened 2024-05-17
Fix medicine category not showing up

MeggalBozale:
## About the PR 

Medicine category was forgotten and now anything which fits in there isnt shown

Also designates anything not in the known list of categories to be assigned to the "Miscellaneous" category where normally only those with null category vars would end up

## Why's this needed? 

Fixes a bug that someone reported in imcoder but i dont think has been reported yet
https://discord.com/channels/182249960895545344/890226559691157524/1240814895247917077
Thank you Vanni!

### DIFF
--- a/code/obj/machinery/manufacturer.dm
+++ b/code/obj/machinery/manufacturer.dm
@@ -409,7 +409,6 @@ TYPEINFO(/obj/machinery/manufacturer)
 		for (var/datum/manufacture/M as anything in L)
 			if (!(M.category in src.categories)) // fix for not displaying blueprints/manudrives
 				M.category = "Miscellaneous"
-				// "WARN: bad category on TYPEPATH"
 			if (length(as_list[M.category]) == 0)
 				as_list[M.category] = list()
 			as_list[M.category] += list(manufacture_as_list(M, user, static_elements))
--- a/code/obj/machinery/manufacturer.dm
+++ b/code/obj/machinery/manufacturer.dm
@@ -409,6 +409,7 @@ TYPEINFO(/obj/machinery/manufacturer)
 		for (var/datum/manufacture/M as anything in L)
 			if (isnull(M.category) || !(M.category in src.categories)) // fix for not displaying blueprints/manudrives
 				M.category = "Miscellaneous"
+				logTheThing(LOG_DEBUG, src, "Manufacturing blueprint [M] has category [M.category], which is not on the list of categories for [src]!")
 			if (length(as_list[M.category]) == 0)
 				as_list[M.category] = list()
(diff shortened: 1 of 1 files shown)


MeggalBozale:
image
moment of all time

code/obj/machinery/manufacturer.dm
```diff
@@ -407,8 +407,9 @@ TYPEINFO(/obj/machinery/manufacturer)
 	proc/blueprints_as_list	(var/list/L, mob/user, var/static_elements = FALSE)
 		var/list/as_list = list()
 		for (var/datum/manufacture/M as anything in L)
-			if (isnull(M.category)) // fix for not displaying blueprints/manudrives
+			if (!(M.category in src.categories)) // fix for not displaying blueprints/manudrives
 				M.category = "Miscellaneous"
+				// "WARN: bad category on TYPEPATH"
```
mordent-goonstation, line 412:
```suggestion
```
No need for commented out debug code here.

code/obj/machinery/manufacturer.dm
```diff
@@ -407,8 +407,9 @@ TYPEINFO(/obj/machinery/manufacturer)
 	proc/blueprints_as_list	(var/list/L, mob/user, var/static_elements = FALSE)
 		var/list/as_list = list()
 		for (var/datum/manufacture/M as anything in L)
-			if (isnull(M.category)) // fix for not displaying blueprints/manudrives
+			if (!(M.category in src.categories)) // fix for not displaying blueprints/manudrives
```
mordent-goonstation, line 410:
```suggestion
			if (isnull(M.category) || !(M.category in src.categories)) // fix for not displaying blueprints/manudrives
```
Simple short-circuit for null case, clarifies that `null` is a possible value.

mordent-goonstation (changes requested):

code/obj/machinery/manufacturer.dm
MeggalBozale: Oh yes i meant to ask mordent, is there a way to have a warning here? In case someone tries to define a new category that doesnt exist yet

code/obj/machinery/manufacturer.dm
MeggalBozale: fair

MeggalBozale:
Is it possible to have something blurb out when there is a null/invalid category? The assignment to misc is reluctant since its not a big issue but it is an issue if someone does it nonetheless

MeggalBozale:
Added debug log

mordent-goonstation (approved):

PR #18958, opened 2024-05-17
Resprites radsuit hood icon state

AXLFAN2000:
## About the PR 
attempted resprite of rad hood item sprite (like when its on the ground) using the palette of the suits item state and the design of the actual worn hood
let me know if the sprite needs changed up ive not really sprited for GOON STATION and it could probably be better 
image
image

## Why's this needed? 
it looks like this currently which is NOT what the hood LOOKS LIKE any more
image

[sprites][clothing][joelsBestHits2024]


DimWhat:
Hood visor should be opaque and a nice reflection of light would be awesome

AXLFAN2000:
> Hood visor should be opaque and a nice reflection of light would be awesome

image
hows this?

AXLFAN2000:
image

PR #18968, opened 2024-05-17
Reduce chance of singularity engine on cogmap1 to 0.

pgmzeta:
[mapping][removal]
## About the PR 
Removes the random chance for singularity engines on cogmap1.

## Why's this needed? 
* It's difficult-to-impossible to push the cogmap1 singularity engine without destroying and irradiating the control room.

## Changelog 

```changelog
(u)glowbold
(*)Cogmap1 will no longer spawn the singularity engine by default. It can still be selected from the engine control computer.
```


TheTyrant25 (approved):
Could probably just delete the file too but this also works.

pgmzeta:
It's possible to mitigate nuclear engine radiation with e.g. bohrum-glass windows, but this does not work for singularity radiation, which uses different logic. A feeder setup de-plates the outer engineering walls, meaning mechlab and the control room get mezzer stuns and radiation due to the line-of-sight.

Glamurio:
I'm with Earthfire here. Making it easier to mitigate and creating gameplay variety is better than removing it.

Sord213:
Yeah I'd rather see changes to how the radiation works if that's the issue than just removal outright. I like setting up the sing on cog1.

pgmzeta:
will think on a hawking rad rework.

PR #18975, opened 2024-05-17
Lower surplus crate cost from 12 to 11 telecrystals

Glamurio:
[QoL] [Balance]

## About the PR 
Changes the cost of a surplus crate from 12 to 11 tokens.

## Why's this needed? 
Surplus crates are cool. However, taking one, means not getting an agent card, and as such not access to the syndicate outpost, which is arguably a really neat RP-friendly part of being a traitor. As such, people have to make the awkward decision of deciding to go lone wolf with a surplus, or instead skipping it altogether in favor of the agent card.

An alternate approach to this could be to lower the surplus content amount to a value closer to 20, and then dropping the price cost to 10, I'm open to suggestions, but figured this is the easiest way to approach it.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Surplus crates now only cost 11 telecrystals.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -428,7 +428,7 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 /datum/syndicate_buylist/traitor/surplus
 	name = "Surplus Crate"
 	item = /obj/storage/crate/syndicate_surplus
-	cost = 12
+	cost = 11
 	vr_allowed = FALSE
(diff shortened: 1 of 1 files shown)


Valtsu0:
I'd rather have aurplus crates always contain an agent card

Glamurio:
> I'd rather have surplus crates always contain an agent card

Fine with this as well, honestly. But I'm guessing it's not wanted in any capacity.

ZeWaka:
this goes against the whole point

Tarmunora:
> this goes against the whole point

x

PR #18978, opened 2024-05-17
Partial revert of #17405, implement LAG_INIT

Sovexe:
## About the PR 
* Partial revert of #17405
* Reverts the `sleep(-1)` calls back to `LAGCHECK` calls
* Implements new less aggressive LAG_INIT used by these procs to salvage some of the performance lost
* Implements `LAGCHECK_IF_LIVE(x)` to allow local servers to keep using `sleep(-1)`

## Why's this needed? 
* Attempting to reduce round start judder


Laboredih123:
alas, if only there was a way to both go fast and to let people in quickly simultaneously

PR #18979, opened 2024-05-17
Rewrite eject_material() for better code quality and to potentially fix a bug.

### LINKED ISSUE #18976: Incorrect bar count when ejecting from fabricators
### Labels

[BUG][MAJOR]

### Description

Ejecting all of one material from a fabricator makes it come out as a smaller number than the amount of bars there actually are in the fabricator.

### Steps to reproduce

1. Purchase 5 of an ore type and reclaim it into a fabricator
2. Eject all 5 bars from the fabricator
3. See that it's only 2 bars
4. ???
5. Try ejecting all 15 bars of another type
6. It comes out as 10 bars

### Expected Behavior

For the fabricator to eject the actual amount of materials

Stacking more bars into the weird buggy stack makes it contain the proper amount it should have in that stack

Reported by: Bluejay357
Client version: 515.1635
Active test merges: #18561 and #18797
Round log date: 2024-05-17-17-51
Reported on: 2024-05-17 18:50:14
Map: COGMAP

MeggalBozale:
## About the PR 
Rewrites the proc on manufacturers which handles ejecting material to be more readable and potentially fix an issue.

## Why's this needed? 
Might fix #18976, but I couldn't reproduce the listed issue in the first place. I suspect the lack of calls to UpdateStackAppearance caused some visual glitch that didn't affect the actual stack amount, which would have confused a player or two.

### DIFF
--- a/code/obj/machinery/manufacturer.dm
+++ b/code/obj/machinery/manufacturer.dm
@@ -1042,11 +1042,11 @@ TYPEINFO(/obj/machinery/manufacturer)
 			return
 		// In case tgui_input_number() misses something or it's 0 (Likely trying to cancel action)
 		if (!isnum(eject_amount) || !isnum_safe(eject_amount) || eject_amount < 1)
-			src.grump_message(user, "ERROR: Cannot eject [eject_amount] bars.", sound = TRUE)
+			src.grump_message(user, "ERROR: Cannot eject [eject_amount] piece\s.", sound = TRUE)
 			return
 		// This should never happen either
 		if (eject_amount > P.amount)
-			src.grump_message(user, "ERROR: Cannot eject [eject_amount] bars, as there are only [floor(P.amount)] bars available to eject.", sound = TRUE)
+			src.grump_message(user, "ERROR: Cannot eject [eject_amount] piece\s, as there are only [floor(P.amount)] piece\s available to eject.", sound = TRUE)
 			eject_amount = floor(P.amount)
 		if (eject_amount == P.amount)
 			P.UpdateStackAppearance()
--- a/code/obj/machinery/manufacturer.dm
+++ b/code/obj/machinery/manufacturer.dm
@@ -1026,44 +1026,47 @@ TYPEINFO(/obj/machinery/manufacturer)
 		storage[index_2] = temp_hold
 		src.should_update_static = TRUE
 
-	proc/eject_material(var/mat_ref, mob/user = null)
-		if (src.mode != MODE_READY)
-			src.grump_message(user, "You cannot eject materials while the unit is working.")
+	proc/eject_material(var/mat_ref, mob/user)
+		var/obj/item/material_piece/P = src.get_material_by_ref(mat_ref)
+		if (!src.can_eject_material(P, user))
 			return
-		var/ejectamt = 0
-		var/turf/ejectturf = get_output_location(user)
-		var/obj/item/material_piece/target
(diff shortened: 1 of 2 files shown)


code/obj/machinery/manufacturer.dm
```diff
@@ -1026,44 +1026,47 @@ TYPEINFO(/obj/machinery/manufacturer)
...
-		if (ejectamt > target.amount)
-			src.grump_message(user, "There's not that much material in [name]. It has ejected what it could.", sound = TRUE)
-			ejectamt = floor(target.amount)
-		if (ejectamt == target.amount)
-			if (user)
-				user.put_in_hand_or_drop(target)
-			src.storage.transfer_stored_item(target, ejectturf)
+		// This should never happen either
+		if (eject_amount > P.amount)
+			src.grump_message(user, "ERROR: Cannot eject [eject_amount] bars, as there are only [floor(P.amount)] bars available to eject.", sound = TRUE)
```
mordent-goonstation, line 1049: Some materials (e.g. cotton) aren't bars. "pieces", maybe? Or is there a proc to determine the correct noun somewhere?

code/obj/machinery/manufacturer.dm
```diff
@@ -1026,44 +1026,47 @@ TYPEINFO(/obj/machinery/manufacturer)
...
+											 max_value = ceil(P.amount),
+											 min_value = 0,
+											 round_input = TRUE,
+											)
+		if (!src.can_eject_material(P, user))
 			return
-		// Since tgui_input_number() is awaited here, make sure they are adjacent before and after
-		if (!src.check_physical_proximity(user))
+		// In case tgui_input_number() misses something or it's 0 (Likely trying to cancel action)
+		if (!isnum(eject_amount) || !isnum_safe(eject_amount) || eject_amount < 1)
```
mordent-goonstation, line 1044: Do we need to check for fractional amounts, e.g. `1.5`?

code/obj/machinery/manufacturer.dm
```diff
@@ -1026,44 +1026,47 @@ TYPEINFO(/obj/machinery/manufacturer)
...
+		if (src.mode != MODE_READY)
+			src.grump_message(user, "ERROR: Cannot eject materials while the unit is working.", sound = TRUE)
+		else if (isnull(material_in_storage))
+			src.grump_message(user, "ERROR: Cannot find material in storage. Aborting.", sound = TRUE)
+		else if (material_in_storage.amount < 1)
+			src.grump_message(user, "ERROR: Not enough material to eject whole amounts of bars.", sound = TRUE)
+		else if (!src.check_physical_proximity(user))
+			src.grump_message(user, "You have to be able to take that out yourself, you can't reach the bars from here!", sound = TRUE)
+		else
+			return TRUE
```
mordent-goonstation, line 1069:
```suggestion
		else
			return TRUE
		return FALSE
```

code/obj/machinery/manufacturer.dm
MeggalBozale: Ah yeah, gonna have to find a proc for that. Not sure if there is one really, I may just have to avoid referring to the quantity of bars/sheets/blahblahblah of fabric

code/obj/machinery/manufacturer.dm
MeggalBozale: No, tgui_input_number rounds this

PR #18980, opened 2024-05-17
Expanded Wizard and Blob crew credits

alicars:
## About the PR 

Adds more antag data to crew credits! This includes the spells wizards purchased and the upgrades the blob picked! As well as some misc text data from blobs I've seen people wonder over in deadchat before.

wizard creds
blob working

The blob data can get a little busy if they were very hungry and bought a lot of repeated upgrades, but I figured that's probably ok? Spiefs credits can also get pretty bloated if they stole a lot.  

Code note: Currently, the frame of the wizard icons are all set to 5 via a new variable, spell_animation_frame. We need this since the wizard spells play a very pretty little animation when you first purcahse them, and we'd like to use the last frame of that animation if possible. 
I tried for several hours to try to figure out where or how this animation is handled so I could access the frame by code, but gave up after a lot of dead ends. 🙃 Perhaps in the case the animation changed, even a more advanced setup would break regardless...? I don't really know, maybe someone wiser could set me straight 
It /does/ seem like there other instances in the codebase of icon2base64 that use hard set, non-1 values, so maybe its fine? I'm not really sure. 

## Why's this needed? 

I just think it's neat to see what people picked!

## Changelog 

```changelog
(u)crunchwarp
(+) Blobs and Wizards now have enhanced crew credits!
```

[FEATURE] [UI]


alicars:
Testing note: If you need to test the blob change on your own local, you may encounter a problem where no crew credits are generated at all. From what I can tell, this is because no minds are added to ||minds|| in gameticker.dm since the blob before spawning its nucleus isn't a mob (?? I guess ??) 
I fixed this by editing the 1800 on line 482 of gameticker.dm to be a lower value, like 60, so I didn't have to wait around for this fallback check to happen. Alternatively, just wait a few minutes after spawning and before ending the round
(This seemed to make it work at least, haha. Shouldn't be a problem on live /I think/ because you should have more than one mind on the server if you have a blob. Though I guess if the blob dies very very quickly...)

code/modules/antagonists/wizard/wizard.dm
```diff
@@ -5,6 +5,8 @@
 	success_medal = "You're no Elminster!"
 	faction = list(FACTION_WIZARD)
 	uses_pref_name = FALSE
+	var/list/datum/SWFuplinkspell/purchased_spells = list()
+	var/spell_animation_frame = 5 // This will need updating if the spell animations ever change
```
TobleroneSwordfish, line 9: This is fine by the way, wizard spell sprites include the animation baked into them so there isn't really a better way to do this. Could be a local `#define` though since it's just taking up memory here as a var.

code/modules/antagonists/blob/blob.dm
```diff
@@ -46,16 +49,52 @@
...
 		var/list/absorbed_lifeforms = list()
+		var/list/upgrades = list()
 		for (var/mob/living/carbon/human/H in src.absorbed_victims)
 			if(!H.last_client?.key)
 				absorbed_lifeforms += "[H.real_name] (NPC)"
 
 			else
 				absorbed_lifeforms += "[H.real_name] (played by [H.last_client?.key])"
 
+		for (var/datum/blob_upgrade/upgrade in purchased_upgrades)
```
TobleroneSwordfish, line 60: Could maybe go through the list of upgrades and combine the repeated ones by counting the number of times they appear and putting a `x2` at the end of the displayed name?

TobleroneSwordfish (changes requested):

code/modules/antagonists/wizard/wizard.dm
alicars: cries, ok thank you for the help!!

code/modules/antagonists/blob/blob.dm
alicars: blob x2
looks like this now! whahoo

alicars:
gasp, the spell animations did change!! went ahead and modified the logic so the cool new soulguard art ( #19099) displays properly 
wizard cred take 4 or 5 or something

TobleroneSwordfish:
Oh, perfect!

PR #18987, opened 2024-05-18
Deployable turrets eject bullet casings

FlameArrow57:
[GAME OBJECTS][FEATURE][REWORK]

## About the PR 
This PR makes it so that deployable turrets (N.A.R.C.S./NARCS and NAS-T) eject bullet casings. They now have a forensics profile too, to match with their casings

Also does a minor rework of variables/procs internally to make it work

image

image

## Why's this needed? 
Thought it would be a fun effect

## Changelog 

```changelog
(u)FlameArrow57
(+)Deployable turrets like N.A.R.C.S. turrets now eject bullet casings and have their own forensics profile.
```

### DIFF
--- a/code/modules/projectiles/projectile_parent.dm
+++ b/code/modules/projectiles/projectile_parent.dm
@@ -28,8 +28,6 @@
 	var/max_range = PROJ_INFINITE_RANGE
 	/// What kind of implant this projectile leaves in impacted mobs
 	var/implanted = null
-	/// Forensic ID of the gun, etc that shot this projectile, used for forensics on implanted projectiles
-	var/forensic_ID = null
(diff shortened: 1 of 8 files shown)


TobleroneSwordfish:
These things do have infinite ammo so I kinda worry about generating an infinite pile of casings if the turret gets stuck shooting a cockroach or something. Maybe start deleting them when they go over a certain amount on one tile?

Tarmunora (approved):
code looks good, some solution to the infinite-stack-of-casings issue would be good though yeah

FlameArrow57:
I added something simple to make it start deleting casings once too many turf contents have been reached, LMK if something more involved is wanted

ZeWaka (approved):

PR #18999, opened 2024-05-18
Fix the lawbringer runtiming for trying to set modes with attack_self(), spamming fingerprints message after having an owner

### LINKED ISSUE #18990: Runtime after claiming lawbringer
### Labels

[BUG][RUNTIME]

### Description

Title

### Steps to reproduce

1. Claim lawbringer
2. Press c

### Expected Behavior

No runtime

Looks like this, but for "detain"
image

MeggalBozale:
[Bug] [Runtime] [Game-Objects]
## About the PR 
lawbringer kept trying to call parent proc which runtimes because it does its projectiles list weirdly
remove parent call and do check for if the one doing attack_self is HoS, to remove blurble

## Why's this needed? 
Fixes #18990


code/obj/item/gun/gun_parent.dm
```diff
@@ -175,7 +175,7 @@ var/list/forensic_IDs = new/list() //Global list of all guns, based on bioholder
 	..()
 	if(src.projectiles && length(src.projectiles) > 1)
 		src.current_projectile_num = ((src.current_projectile_num) % src.projectiles.len) + 1
-		src.set_current_projectile(src.projectiles[src.current_projectile_num])
+		src.set_current_projectile(src.projectiles[src.projectiles[src.current_projectile_num]])
```
mordent-goonstation, line 178: Er, what's this doing? Can you explain the types of the things here, because this looks nuts to me.

MeggalBozale: src.projectiles is an associated list of projectile said name "execue", "bigshot", etc. to value which is the type of projectile itself.

so what the old code was doing was getting the key at some index, and setting that string as the projectile datum.
adding it once more gets the actual projectile instead of the string

MeggalBozale: ah hold on im seeing this is weird in a different way. HUH

MeggalBozale:
This is weirder than i thought, one sec

MeggalBozale:
rdy

MeggalBozale:
ALSO IMPORTANT TO CLARIFY I FORGOT!!!
this doesnt stop you from getting access to the lawbringer yourself, current emag effect isnt removed :)

PR #19003, opened 2024-05-18
Removes wizard name suggestions that are from a transphobe's book series

mordent-goonstation:
[GAMEMODES][REMOVAL]

## About the PR 

Removes a few names from wizard name suggestions. I know there's probably some other references in the codebase somewhere, separate PRs to find better replacements for those welcome. Low-hanging fruit, and all that.

## Why's this needed? 

Some lady wrote a children's book. That book became wildly popular, spawned some sequels and even a movie series, video games, etc. etc.. Turns out she's a massive transphobe and all round shitty person, so let's not keep references to those characters in our cool space game.

I'm not going to stop someone choosing to name themselves these, but we shouldn't "suggest" them as random names.

### DIFF
--- a/strings/names/wizard_female.txt
+++ b/strings/names/wizard_female.txt
@@ -7,8 +7,6 @@ names@=
 	Morgan le Fay
 	Circe
 	Maleficent
-	Hermione
-	Umbridge
 	The Worst Witch
 	Sabrina
 	Morrigan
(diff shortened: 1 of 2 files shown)


MeggalBozale (approved):
Weighing in, if it's worth anything from me, that this is a very nice change even if small

klushy225:
yeah, I like this change. That lady boils my blood.

PR #19004, opened 2024-05-18
Sillier EMAG effects for the lawbringer

MeggalBozale:
[Feature] [Game-Objects]
## About the PR 
Adds a more fitting set of emag effects for the lawbringer.
- When you shoot the lawbringer while it's emagged, it will pick between random modes for you.
- ~~There is a 1% chance to force you to scream out "I AM THE LAW" and trigger the explosion~~
- Emagged lawbringers have no sense of laws, and so the HoS is not immune to the explosion for saying these words.
- Moves over switching code to a proc so that it can be accessed from other places in the code. There's a lot of custom varediting in there, so it should be exposed to more parts of the code for use. Like this!

## Why's this needed? 

https://github.com/goonstation/goonstation/assets/27376947/be5eff0f-2313-478e-9a43-c7cc8dc8bb65

NOTE: VIDEO SHOWS THE CHANCE TO BE FORCED TO SAY "I AM THE LAW". NO LONGER A FEATURE.

```changelog
(u)MeggalBozale
(+)The emag effect for the lawbringer has been modified slightly.
```

### DIFF
--- a/code/obj/item/gun/energy.dm
+++ b/code/obj/item/gun/energy.dm
@@ -1286,6 +1286,7 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 	rechargeable = 0
 	can_swap_cell = 0
 	muzzle_flash = "muzzle_flash_elec"
+	var/emagged = FALSE
 
 	New(var/mob/M)
 		set_current_projectile(new/datum/projectile/energy_bolt/aoe)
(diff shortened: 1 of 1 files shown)


MeggalBozale:
Note that while prob(1) is prob(1) i cut out like 30 seconds of me spamming it to trigger the IATL effect

MeggalBozale:
ALSO IMPORTANT TO CLARIFY I FORGOT!!!
this doesnt stop you from getting access to the lawbringer yourself, current emag effect isnt removed :)

TheRyeGuyWhoWillNowDie:
seems not fun? now even if you kill the head of security, steal their gun, and have an emag, the lawbringer is little more than a toy, and worse yet extended use will get you gibbed

MeggalBozale:
You can fabricate a new one or just not emag the thing. It's an emag! Weird things can happen!

code/obj/item/gun/energy.dm
```diff
@@ -1512,6 +1524,14 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 		return 0
 
 	shoot(turf/target, turf/start, mob/user, POX, POY, is_dual_wield, atom/called_target = null)
+
+		if (src.emagged)
+			if (prob(1))
+				boutput(user, SPAN_ALERT("You let it slip."))
+				user.say("I AM THE LAW")
+				src.are_you_the_law(user, "I AM THE LAW") // the thoughts won
```
TobleroneSwordfish, line 1532: Funny but probably a bad idea

TobleroneSwordfish (changes requested):

code/obj/item/gun/energy.dm
MeggalBozale: fine :(

TobleroneSwordfish (approved):

Glamurio (approved):

PR #19005, opened 2024-05-18
Resprites biosuit hood icon state

AXLFAN2000:
## About the PR 
same as with the radhood just redoing the ICON STATE to match the suit better also i fixed the opacity on the radhood because iconstates should be like full opacity someone told me which makes sense but um if you want me to take that part out let me know OK
image
image

## Why's this needed? 
image
its a little old

[sprites][clothing][joelsBestHits2024][softsoftpizza]


WALPVRGIS (approved):
I had no idea that was still an ancient sprite, great update ty :)

PR #19028, opened 2024-05-19
Add purple recolours of captain gear

JORJ949:
## About the PR 
Adds a Purple recolour of the captain's attire available from the "This station is ours!" medal
image
image

## Why's this needed? 
More medal rewards are good and purple is a traditionally "royal" colour so I thought it would fit the captain and it goes nicely with the gold accents

## Changelog 

```changelog
(u)JORJ949
(*)Captains who have previously overthrown a stations command can now show their royalty with a purple recolouring of their attire.
```


TobleroneSwordfish:
Not a huge fan of this colour for captains, especially such bright pinky-purple. General dev consensus is that we have enough recolours as is.

Glamurio:
While I do like the recolor, I do think it's a bit too much "in your face". A darker, more nuanced shade of purple would look better if anything.

PR #19035, opened 2024-05-20
Adds Scientist Training

cheekybrdy:
[FEATURE] [CONTROVERSIAL] [INPUT-WANTED] [SCIENCE]
## About the PR 
This adds science training for researchers, this iteration only adds the artsci component, as from what I've gathered from forums is the least controversial feature of this. This training now has artifact hints locked behind it.

## Why's this needed? 
This is the start of a training to prevent scientists from being ignored as all their interaction is usually tided away on classic or outright ignored through other job's demand. This particular partition for it is intended to make artifacts more "fuck around and find out" again. It makes no sense that the untrained can recognise anomalous markings regardless of job and makes hoboarts really feel dangerous. No current modifications to make if a art is handheld or not blatantly obvious or not by non arttifact hint means.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Adds Scientist Training, all this does currently is lock artifact hints to research staff.
```


TDHooligan:
mixed feelings on this one.
the effect i'm actually for. and think this would just generally make things more Fun by creating silly mistakes. 
however, because this implementation of science training is so small in scope, it almost isnt worth existing as a trait IMO.

my key issue is that I don't think science has many other areas where this trait could be made worth existing (while also making things more fun)

then again. i forget what security training does most of the time. perhaps the effect of this just being a footnote is totally Okay

cheekybrdy:
I plan to add more to this trait in other PRs. This was more a atomisation attempt along with allowing some controversial propositions for the other areas of sci that'd be made here to be considered individually.

444Portal:
the one thing that gets me on this are borgers and generators - i think those ones should be visible regardless (esp since its purely off of vibes instead of actual markings), hey on that train of thought, wouldnt john staffie be able to tell a generator is sparking? those two tend to be the most immediately deadly so maybe show those regardless

code/obj/artifacts/artifact_items/activator_key.dm
```diff
@@ -11,7 +11,7 @@
 	validtypes = list("ancient","martian","wizard","eldritch","precursor")
 	automatic_activation = 1
 	react_xray = list(12,80,95,8,"COMPLEX")
-	examine_hint = "It kinda looks like it's supposed to be inserted into something."
+	examine_hint = SPAN_ARTHINT("It kinda looks like it's supposed to be inserted into something.")
```
Xkeeper0, line 14: The text changes should be where it is output, not here; this goes for all applications of this text macro.

code/obj/artifacts/artifact_items/energy_gun.dm
```diff
@@ -37,7 +37,7 @@
 		if (!src.ArtifactSanityCheck())
 			return
 		var/datum/artifact/A = src.artifact
-		if (istext(A.examine_hint))
+		if (istext(A.examine_hint) && (usr && (usr.traitHolder?.hasTrait("training_scientist"))))
 			. += A.examine_hint
```
Xkeeper0, line 41: This is where the `SPAN_ARTHINT(...)` should be applied.

Xkeeper0 (changes requested):

Xkeeper0:
> the one thing that gets me on this are borgers and generators - i think those ones should be visible regardless (esp since its purely off of vibes instead of actual markings), hey on that train of thought, wouldnt john staffie be able to tell a generator is sparking? those two tend to be the most immediately deadly so maybe show those regardless

these are good points. but at the same time, i don't think i mind that much? i could see adding some other tells to those artifacts (like maybe an automaton-like whirring every so often, and occasional visible sparking if it doesn't already) but i do like the idea of making ghetto artlab more FAFO-oriented.

Sbmhawk:
> > the one thing that gets me on this are borgers and generators - i think those ones should be visible regardless (esp since its purely off of vibes instead of actual markings), hey on that train of thought, wouldnt john staffie be able to tell a generator is sparking? those two tend to be the most immediately deadly so maybe show those regardless
> 
> these are good points. but at the same time, i don't think i mind that much? i could see adding some other tells to those artifacts (like maybe an automaton-like whirring every so often, and occasional visible sparking if it doesn't already) but i do like the idea of making ghetto artlab more FAFO-oriented.

Imo there's already a lot of FAFO in artifacts, even without having a chance of being instantly killed/borged on touching an artifact. Even just not having the conspicuous marking would make it more towards that.

Glamurio:
I think if we're adding this, then maybe the conspicuous markings should be the default message any old staffie sees, and that scientists should actually have a bit more understanding of what they're looking at, this is what training is for after all. You don't go to med school to learn "this patient seems to be in pain", you learn what is wrong with them in the first place.

And arguably, a part I'd like to see this training used in particular is actually in regards to chems. There's a lot of powergaming with chems, and frankly, it would be alleviated somewhat if you had to have science training to make some of the more advanced chems in the game. I'm not saying the basic heal chems should be on that list, but maybe triple meth or even just regular meth shouldn't be made by a novice, it makes no sense.

JORJ949:
> And arguably, a part I'd like to see this training used in particular is actually in regards to chems. There's a lot of powergaming with chems, and frankly, it would be alleviated somewhat if you had to have science training to make some of the more advanced chems in the game. I'm not saying the basic heal chems should be on that list, but maybe triple meth or even just regular meth shouldn't be made by a novice, it makes no sense.

I disagree with being completely inable to make chems, the other trainings just make things harder if you dont have them as a way to discourage (but not outright prevent) doing that job. Security training lets you cuff faster, medical lets you see if a brain can be used before you use it and makes self surgery easier, QMs training lets you haggle easier. The job should still be possible in case someone has to fill in.

cheekybrdy:
I do plan to add something for chems, but due to the inevitable controversy and scope of that, I made the call to atomise it per subdepartment.

Tarmunora (approved):

PR #19049, opened 2024-05-20
Adds "ants" reagent to foods that have spawned ants & makes fridges ant-safe

Glamurio:
[QoL] [Feature]

## About the PR 
Adds 5 units of "ants" to any food that has spawned ants.

Adds `/obj/storage/secure/closet/fridge` to the list of turfs that spawn ants from blocking, as it's always been comical that an opened fridge would cause ants to swarm it.

Renames `on_table` to `ant_safe`, as that seems more descriptive by now.

## Why's this needed? 
Ants are a fun chem that never gets used. Also people routinely just ignore ants on their food. This should add a little more incentive to put food on tables, and is otherwise just a minor inconvenience to people who ignore ants on their food.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Food left on the floor will now contain ants. Don't leave your food on the floor, people.
```


flappybatpal:
Good idea.

Glamurio:
> Incoming all food in botany be contaminated with ants.
> 
> Dunno about this from the botany side of things. It would promote using produce satchel for harvesting, but that is sadly not that well documented. So i think that could run into problems with beginner botanists.
> 
> then again, we could just try implementing it and changing produce to be a bit easier to handle without ants getting onto it, if it really causes problems.

Could certainly make exceptions to produce, but also ants won't show up until 3 minutes have passed. If you leave food on the ground for 3 minutes, even in botany, then I would argue you weren't paying attention.

DisturbHerb:
eat ant

RelentlessGarbage:
Some feedback; it should be made obvious that an article of food with ants on it (unless said ants have been injected into the food) should have some kind of tell that there are ants (e.g. an "Ant-covered" prefix). The ants are obvious on the floor, they should be obvious on the food as well, unless they were added in a clandestine manner. 

I'd argue that food quality should also be decreased by getting covered in ants and the ants should be removed if the food is washed, but that is scope creep. What matters most is that players either have a tell in the description or the item sprite that a food item has ants in it. If someone handed you an apple covered in ants, you'd probably notice.

Glamurio:
> Some feedback; it should be made obvious that an article of food with ants on it (unless said ants have been injected into the food) should have some kind of tell that there are ants (e.g. an "Ant-covered" prefix). The ants are obvious on the floor, they should be obvious on the food as well, unless they were added in a clandestine manner.
This seems like a cool idea. I'll see to adding that.
> I'd argue that food quality should also be decreased by getting covered in ants and the ants should be removed if the food is washed, but that is scope creep. What matters most is that players either have a tell in the description or the item sprite that a food item has ants in it. If someone handed you an apple covered in ants, you'd probably notice.
Quality of food decrease is a cool idea but not in this PR. Not sure if washing off ants makes sense; they kind of go everywhere and it's impossible to fully remove every ant from every food. You can totally just remove them via syringe right now, technically.
> EDIT: Does the addition of ants play well with the existing reagent capacities of food items? As in, will the ants "fit" into the limited reagent capacity most foods have with the current implementation?
Currently it just does a regular add_reagent() call, which already checks that it can't go over the maximum, so yeah.

code/modules/chemistry/tools/food_and_drink.dm
```diff
@@ -186,12 +188,14 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks)
...
 			create_time = world.time
-			if (!src.disposed && isturf(src.loc) && !on_table())
+			if (!src.disposed && isturf(src.loc) && !ant_safe())
 				if (prob(50))
 					made_ants = 1
 					processing_items -= src
 					if (!(locate(/obj/reagent_dispensers/cleanable/ants) in src.loc))
 						new/obj/reagent_dispensers/cleanable/ants(src.loc)
+						src.reagents.add_reagent("ants", 5)
+						src.name = "[name_prefix("Ant-covered", 1)][src.name][name_suffix(null, 1)]"
```
Azrun, line 198:
```suggestion
						src.name = "[name_prefix("ant-covered", 1)][src.name][name_suffix(null, 1)]"
```

Azrun (changes requested):

frawhst:
Honestly I'd remove some chems from the food, as the ants are 'eating' them. That also means that foods which don't have extra space in their reagent capacity will be able to have ants in them. Maybe just remove 5 units of chems before adding the ants?

frawhst (approved):

Glamurio:
!merge_upstream

PR #19056, opened 2024-05-20
Changes the string material patterns ("CON-1", "MET-1") all to /datum/manufacturing_requirement/...

MeggalBozale:
[Secret] [Internal] [Code-Quality]
## About the PR 
Introduces /datum/manufacturing_requirement, a datum which contains is_match(/datum/material/M) to check if a material meets the given requirements.

THIS WILL REQUIRE SECRET SUBMODULE CHANGES!!! D: (not many)

Unifies the item_paths and item_amounts properties into item_requirements, which is basically just 
`item_requirement[item_path ("MET-1" turned into /datum/manufacturing_requirement)] = item_amount.`

Simplifies item_names[index] into item_requirement.name, as the name is now stored on the requirement itself

Adds a global cache list for all the manufacturing requirements, for easy lookup using only one instance of each in manufacturing.dm

This change affects all /datum/manufacturing types, and all code in manufacturer.dm and elsewhere which works with the old variables item_names and item_amounts

## Why's this needed? 
Significantly improves maintainability, usability, readability, etc. of this code. I don't think I could ever memorize what anything beyond "MET-1" is supposed to mean, having a global list of string pattern to keys like this for only manufacturers is pretty bad.

Brings material handling closer to nanofabs, which also use datums. Allowing the potential for datum checks like /datum/manufacturing_requirement/is_some_item means that one day, we could move away from fabricators only containing material_pieces and instead unify nanofabs to use the same processes.


MeggalBozale:
Actually, hold on. Let me commit these in an order which makes it easier to review. sorry lol

MeggalBozale:
#19057

PR #19057, opened 2024-05-20
Changes the string material patterns ("CON-1", "MET-1") all to /datum/manufacturing_requirement/...

MeggalBozale:
[Secret] [Internal] [Code-Quality]
## About the PR 
Introduces /datum/manufacturing_requirement, a datum which contains is_match(/datum/material/M) to check if a material meets the given requirements.

### THIS WILL REQUIRE SECRET SUBMODULE CHANGES!!! D: (not many)

Unifies the item_paths and item_amounts properties into item_requirements, which is basically just 
`item_requirement[item_path ("MET-1" turned into /datum/manufacturing_requirement)] = item_amount.`

Simplifies item_names[index] into item_requirement.name, as the name is now stored on the requirement itself

Adds a global cache list for all the manufacturing requirements, for easy lookup using only one instance of each in manufacturing.dm

This change affects all /datum/manufacturing types, and all code in manufacturer.dm and elsewhere which works with the old variables item_names and item_amounts

## Why's this needed? 
Significantly improves maintainability, usability, readability, etc. of this code. I don't think I could ever memorize what anything beyond "MET-1" is supposed to mean, having a global list of string pattern to keys like this for only manufacturers is pretty bad.

Brings material handling closer to nanofabs, which also use datums. Allowing the potential for datum checks like /datum/manufacturing_requirement/is_some_item means that one day, we could move away from fabricators only containing material_pieces and instead unify nanofabs to use the same processes.


MeggalBozale:
At this point all the basic code/conversion is done, what comes next is reflecting those changes in the user interfact

MeggalBozale:
Opening for review in case anyone wants to take a look at the code as is, I staggered out the commits so you can see it step-by-step if it helps

+secret
Tarmunora, line 1: Do not commit this

byond-tracy
Tarmunora, line 1: Do not commit this

Tarmunora (changes requested):

MeggalBozale:
Ah woops the consequences of commiting all converted files. mb

code/datums/manufacturing_requirements.dm
```diff
@@ -0,0 +1,303 @@
...
+			id = "syreline"
+
+		telecrystal
+			id = "telecrystal"
+
+		uqill
+			id = "uqill"
+
+		viscerite
+			id = "viscerite"
```
Tarmunora, line 170: these should be autogenerated at runtime

Tarmunora (changes requested):
also add a unit test for id uniqueness

code/datums/manufacturing_requirements.dm
MeggalBozale: Might be for the best so that we don't need to special case having /datum/s and strings in the item_requirements of each manufacturer

code/WorkInProgress/AzrunStuff.dm
MeggalBozale, line 1: Conversion of two manufacture datums

code/WorkInProgress/nadir_antenna.dm
MeggalBozale, line 1: Conversion of one typeinfo, the indentation keeps looking weird in the web viewing thing but i promise its properly indented in vscode

code/WorkInProgress/pali/blob_fabricator.dm
MeggalBozale, line 1: Absolute pathing of, and conversion of, manufacture datums for palis blob stuff

code/WorkInProgress/salvager.dm
MeggalBozale, line 1: Conversion of a single manufacture datum for light salvager pod armor

code/datums/controllers/mechanic_controls.dm
MeggalBozale, line 1: Changes to how create_blueprint() works to function with item requirements

code/datums/gamemodes/pod_wars/pw_manufacturing_.dm
MeggalBozale, line 1: Changes to various podwars manufacturing datums to convert to the new format

code/datums/manufacturing.dm
MeggalBozale, line 1: By far the largest file to have changes in. Most of the changes are the bulk of several hundred manufacture datums in this file getting new formats, but there are the other, distinct changes at the top of the file:
- Removal of get_nice_mat_name_for_manufacturers(mat), as its behavior is replaced elsewhere
- New proc setup_manufacturing_requirements, which converts the string identifiers of manufacturing requirement datums to their corresponding instances in the cache.
- Slight changes to sanity check for new system
- slight changes to get_generated_costs() for new system

code/modules/unit_tests/manufacturing.dm
MeggalBozale, line 1: This is the unit test which was originally a sanity check on the manufacturing blueprints (and ran every time a blueprint was created), as well as a modified material id check to ensure manufacturing requirement IDs remain unique.

code/obj/machinery/manufacturer.dm
MeggalBozale, line 1: Various individual minor changes to allow manufacturers to use item_requirements, as well as send the data as such to the UI.

code/world/initialization/0_preMapLoad.dm
MeggalBozale, line 1: Adds the proc definition and call which sets up the manufacturing requirement cache on world init.

tgui/packages/tgui/interfaces/Manufacturer/components/BlueprintButton.tsx
MeggalBozale, line 1: Minor tweaks to the data handling to utilize the new way data is passed aorund.

code/datums/manufacturing_requirements.dm
MeggalBozale: hold on i figured it out im dumb

code/WorkInProgress/AzrunStuff.dm
```diff
@@ -43,20 +43,20 @@
 
 /datum/manufacture/sub/treads
 	name = "Vehicle Treads"
-	item_paths = list("MET-2","CON-1")
-	item_amounts = list(5,2)
+	item_requirements = list("metal_dense" = 5,
+							 "conductive" = 2)
```
mordent-goonstation, line 47: image
I am absolutely not going to require this change, but this is a classic "indent with tabs, align with spaces" thing.

ZeWaka: i think you're wrong here tbh mordent, this is javascript poisoning your brain - we use tabs for this all over the codebase and in modern code, it's just our fault we don't have a standard tab width

code/datums/manufacturing_requirements.dm
```diff
@@ -0,0 +1,212 @@
...
+				material_threshold = 10
+
+			superdense
+				name = "Extremely Tough Metal"
+				id = "metal_superdense"
+				material_threshold = 15
+
+			matches_property(var/datum/material/M)
+				// This specific check is based off the hardness of mauxite and bohrum.
+				// Mauxite ends up being 15 in here, while bohrum ends up being 33.
```
Tarmunora, line 188:
```suggestion
				// Mauxite ends up being 10 in here, while bohrum ends up being 16.
```

Tarmunora (changes requested):

code/modules/unit_tests/manufacturing.dm
```diff
@@ -0,0 +1,38 @@
+/// This is for making sure that every manufacturing requirement
+/// Has a unique ID for use in the cache. Fails if any ids collide
+/datum/unit_test/manufacturing_requirement_id_uniqueness
+
+/datum/unit_test/manufacturing_requirement_id_uniqueness/Run()
+	var/list/ids = list()
+	for(var/requirement in concrete_typesof(/datum/manufacturing_requirement))
```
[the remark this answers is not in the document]
MeggalBozale: ah, shoulda figured that things get init'd by now

Tarmunora:
!merge_upstream

PR #19061, opened 2024-05-20
Add max/min buttons to the manufacturer speed slider

MeggalBozale:
[QoL] [UI]
## About the PR 
Adds some convenient speed buttons since max/min speed should be fairly common, and the slider itself is pretty bad at being convenient itself.

Additionally, adjusts the now-smaller bar to be easier to slide left/right. It was very resistant to changing values beforehand which made it frustrating.

## Why's this needed? 
QoL I suppose, this is after feedback was prompted in imcoder


MeggalBozale:
dreamseeker_Q4Viw6gT16

mordent-goonstation (approved):

frawhst:
!merge_upstream

PR #19063, opened 2024-05-20
Bleeding Revamp

TDHooligan:
[BALANCE][MEDICAL][INPUT]

## About the PR 
Removes the surprisingly convoluted bleed rolling system and replaces it with a more reliable system:
- Sharp Weapons use a 'Bleed Level' based upon:
  - Sharpness (All CUT/STAB/CRUSH weapons get a base level of '1')
  - Damage also increases the bleed level:
    - 6 - 10 = +1, 
    - 11-29 = +2
    - 30+ = +3
- On hit, sharp weapons will more reliably increase BLEED up to their 'Bleed Level'
- Critical hits *always* increase bleed, even above the weapons' Bleed Level. So being stabbed repeatedly, even with a regular knife, will eventually lead to heavy bleeds
---
 As bleeds will be more reliable, I made a couple changes to bleed:
 - Tier 4+ bleeds have a unique warning text, and can't be cured by natural regeneration. They explicitly require staunching/bandaging/cauterizing.
 - A new intermediate, serious bleedout stage below 300 BLOOD
 - Significantly lower bleeding/sec at debilitatingly low levels of blood (you have less blood to bleed)
 - Anticoagulants/Coagulants now adjust bleed level, making you directly susceptible or resistant to bleed.

## Why's this needed? 
The math around BLEED is incredibly hard to describe, balance, and tends to result in the whole bleed mechanic being unreliable.
You can be naked, getting stabbed with a knife 20x and not bleed. You can take 1 shotshell to the chest and take 3 BLEED. Lying down stops you bleeding more etc. etc.

With this PR, weapons will reliably cause bleed roughly in line with their rarity. Weaker weapons can still cause heavy BLEED if you wail on someone with a knife until they're in crit.

## Changelog 

```changelog
(u)TDHooligan
(*)Sharp weapons cause bleeding more reliably. 
(*)Heavy bleeding (Artery nicks) cannot be naturally regenerated, and require bandages/sutures/cauterization/staunching.
(*)As compensation, bleeding slows down as you lose blood. Coagulants (proconvertin) also limit incoming bleeds, like 'Bleed Armor'.
```


code/mob/living/life/blood.dm
```diff
@@ -33,25 +33,25 @@
 		if (owner.bleeding)
 
 
-			var/decrease_chance = 2 // defaults to 2 because blood does clot and all, but we want bleeding to maybe not stop entirely on its own TOO easily, and there's only so much clotting can do when all your blood is falling out at once
+			var/decrease_chance = 0 // defaults to 2 because blood does clot and all, but we want bleeding to maybe not stop entirely on its own TOO easily, and there's only so much clotting can do when all your blood is falling out at once
```
Tarmunora, line 36: edit code comment

code/mob/living/life/blood.dm
```diff
@@ -169,10 +173,32 @@
 				if (prob(30))
 					owner.changeStatus("shivering", 6 SECONDS)
 				owner.contract_disease(/datum/ailment/malady/shock, null, null, 1) // if you have no blood you're gunna be in shock
+				APPLY_ATOM_PROPERTY(owner, PROP_MOB_STAMINA_REGEN_BONUS, "hypotension", -5)
+				owner.add_stam_mod_max("hypotension", -50)
+
+			if (1 to 200) // very VERY low (70/50)
```
Tarmunora, line 179: effects here seem to be barely more severe than the 200-300 bracket. Should either make this a bit nastier and/or make the higher bracket more forgiving

code/modules/medical/blood_system.dm
```diff
@@ -169,107 +169,77 @@ this is already used where it needs to be used, you can probably ignore it.
...
 				H.visible_message(SPAN_ALERT("[H] starts bleeding!"),\
 				SPAN_ALERT("<b>You start bleeding!</b>"))
-			else if (old_bleeding >= 1)
+			else if (H.bleeding >= 1)
 				H.show_text("<b>You[pick(" start bleeding even worse", " start bleeding even more", " start bleeding more", "r bleeding worsens", "r bleeding gets worse")]!</b>", "red")
-			else if (old_bleeding >= 4)//9)
+			else if (H.bleeding > 3)
+				H.show_text("<b>That nicked an artery!</b>", "red")
+			else if (H.bleeding > 4)//9)
 				H.show_text("<b>You can't go on very long with blood pouring out of you like this!</b>", "red")
```
Tarmunora, line 243: unreachable code

code/mob/melee_attack_procs.dm
```diff
@@ -995,7 +995,7 @@
 			target.TakeDamage(def_zone, (damage_type != DAMAGE_BURN ? damage : 0), (damage_type == DAMAGE_BURN ? damage : 0), 0, damage_type)
 
 			if ((damage_type & (DAMAGE_CUT | DAMAGE_STAB)) || bleed_always)
-				take_bleeding_damage(target, owner, damage + bleed_bonus, damage_type)
+				take_bleeding_damage(target, owner, damage + bleed_bonus, damage_type, is_crit=stamina_crit)
```
Tarmunora, line 998: Somewhat apprehensive about coupling stamina crits to bleeding effects, especially considering one of the weapons already best at causing bleeding, chainsaws, have one of the highest crit chances.

Tarmunora (changes requested):

code/mob/melee_attack_procs.dm
TDHooligan: its definitely worth thinking of. buuut IMO stamcrits are a little fuzzy on what they can and can't do.

my take, longterm, is that  it's positive to move them towards being just 'crits' with appropriate effects based on the weapon type. stamina for blunt, bleed/damage for stabbing weapons. why would 5 STA damage knives ever need *stam*crits?

ultimately bleeding is still only capped at 5/t. the only reason it needs to scale so hard with high damage weapons is because youll be dead in 3/4 hits anyway (way before the bleed can do anything). IMO it helps give lasting impression to big weapons now none of them just stun you forever, since you need more than just a lil' styptic.

TDHooligan:
thoughts for when i get back to this:
- rev heals, other heals need to be checked for bleed heals
- low blood loss tiers probably do too much really disruptive stuff (blurred vision, slows), and not enough minor stuff (increase stamina loss?)
- because of an oversight in old blood healing logic, bleed heals are way weaker than they used to be. probably look at this (or the bloodloss tiers)
- some weapons do a lot of flat bleed. apparently some bullet lodgings have caused extra bleed (though this might be OK).

Valtsu0:
The csaber probably should not deal that much bleed. If anything, it should cauterize the wound

PR #19064, opened 2024-05-20
Try to put tickets from ticket writers & PDAs into the hand

Glamurio:
[QoL]

## About the PR 
Tries to `put_in_hand_or_drop()` on `/obj/item/device/detective_scanner`, `/obj/item/device/ticket_writer` and `/datum/computer/file/pda_program/security_ticket`.

## Why's this needed? 
The floor is full of ants, don't you know? You want ants on your ticket?

### DIFF
--- a/code/obj/item/device/pda2/smallprogs.dm
+++ b/code/obj/item/device/pda2/smallprogs.dm
@@ -1108,7 +1108,9 @@ Using electronic "Detomatix" SELF-DESTRUCT program is perhaps less simple!<br>
 			playsound(src.master, 'sound/machines/printer_thermal.ogg', 50, 1)
 			SPAWN(3 SECONDS)
 				var/obj/item/paper/p = new /obj/item/paper
-				p.set_loc(get_turf(src.master))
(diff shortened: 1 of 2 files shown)


code/obj/item/device/pda2/smallprogs.dm
```diff
@@ -1108,7 +1108,9 @@ Using electronic "Detomatix" SELF-DESTRUCT program is perhaps less simple!<br>
 			playsound(src.master, 'sound/machines/printer_thermal.ogg', 50, 1)
 			SPAWN(3 SECONDS)
 				var/obj/item/paper/p = new /obj/item/paper
-				p.set_loc(get_turf(src.master))
+				if (ishuman(usr))
```
TobleroneSwordfish, line 1111: This doesn't need an `ishuman` check, `put_in_hand_or_drop` is defined on `/mob` so `usr.put_in_hand_or_drop()` should be safe.

TobleroneSwordfish (changes requested):

PR #19065, opened 2024-05-20
Smaller gangs

TDHooligan:
[GAMEMODE][BALANCE]
## About the PR 
Lowers gang size to 4 (leader + 3 members)
hard cap on 3 gangs (4 was too many)

Worth a try. can always increase

## Why's this needed? 
gangs a lil too crazy atm, but dynamics between 3 gangs is generally positive. Let's see if we can keep triple gang with less chaos 

## Changelog 

```changelog
(u)TDHooligan
(*)Gangs have shrunk to 4 people. It should be easier keeping track of all your dudes as a leader, and less chaos overall.
```


TobleroneSwordfish (approved):

444Portal:
where we dropping (sorry not sorry)

PR #19068, opened 2024-05-21
Adds some new floors - Pool tile!

TemThrush:
## About the PR 
This PR adds new floor tiles + some bordered and corner versions! These were originally made for the pool resprite, so they match with the new pool-walls.

(I made the tiles available in the floor-planner for any construction fiends to use, but if they're a bit too overwhelming I'll remove them👍)

image

image
(This PR doesn't replace the current pool concrete, this is just an example of what it looks like underwater!)
## Why's this needed? 
I got the idea from imspriter, more options for mappers and station renovators is always a plus!


TDHooligan:
glorious. this makes me want pool tile walls  though...

Tarmunora (approved):

Flaborized:
these are awesome, great work!!

mordent-goonstation:
Is there intended to be an illusion of depth on the white-blue transitions? Definitely how I see it, I think because of the border contrast.

TemThrush:
> Is there intended to be an illusion of depth on the white-blue transitions? Definitely how I see it, I think because of the border contrast.
I wouldn't say it was intended, but I think I see what you mean!

PR #19078, opened 2024-05-21
Adds gibber as fishing spot

AXLFAN2000:
## About the PR 
lets you fish in the gibber, it has mostly the same pool as the cryopods/cloner but with monkey limbs instead of human limbs and with the inclusion of monkeymeat and blindfolds at a small rate

## Why's this needed? 
more fishing spots = more fun, this seemed charming. also anglers are already incentivized to give fish to catering so more reasons for them to be over there is nice.  lets all do some inter departmental cooperation folks

image

[feature][game objects][joelsBestHits2024]


Gannets (approved):
Sure!

PR #19081, opened 2024-05-21
Penalize security deparmental score if Stirstir dies

TealSeer:
[FEATURE] [STATION SYSTEMS] 
## About the PR 

Halve final department score for security if they fail to protect Stirstir from death.

## Why's this needed? 

A bunch of sec players expressed how much they hate Stirstir even though he's just a little boy so this will incentivize them to show him the love he deserves. Also its funny

## Changelog 

```changelog
(u)TealSeer
(+)Security's departmental score is now influenced by the fate of Monsieur Stirstir.
```

### DIFF
--- a/code/procs/scorestats.dm
+++ b/code/procs/scorestats.dm
@@ -10,6 +10,7 @@ var/datum/score_tracker/score_tracker
 	// var/score_crew_evacuation_rate = 0 save this for later to keep categories balanced
 	var/score_crew_survival_rate = 0
 	var/score_enemy_failure_rate = 0
+	var/score_stirstir_alive = FALSE
 	var/final_score_sec = 0
 	// ENGINEERING DEPARTMENT
 	var/power_generated = 0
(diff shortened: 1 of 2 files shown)


imawxome:
There is a comment in the code saying he's going easy in the fights, he's a demon

TobleroneSwordfish (approved):

code/procs/scorestats.dm
```diff
@@ -86,6 +87,14 @@ var/datum/score_tracker/score_tracker
 
 		final_score_sec = (score_crew_survival_rate + score_enemy_failure_rate) * 0.5
 
+		for(var/mob/living/carbon/human/npc/monkey/stirstir/M in mobs)
+			if(isalive(M))
+				score_stirstir_alive = TRUE
```
BatElite, line 92: Would suggest flipping to "if dead then FALSE", cause as it is sec will get penalised if a map doesn't spawn StirStir on it. IDK if any still don't but that wasn't always guaranteed.

(although, then you could just gib StirStir's body and avoid the penalty, so maybe `score_stirstir_alive` ought to be global & set by StirStir's death proc directly.)

TobleroneSwordfish: If it's set by death then it'd need to be re-set by all revive methods and that becomes a mess quickly. I think a map not spawning with stirstir somewhere is unusual enough to be considered an error state.

TealSeer: stir stir spawns on every map in rotation and even some that aren't I made sure to check

PR #19086, opened 2024-05-22
Group messages relating to bad smells

TealSeer:
[UI] [QOL]
## About the PR 

Adds a message group for the "good lord it smells like shit" messages you get when someone hasn't washed/has apocrine enhancement/opened some rank food.

## Why's this needed? 

Anti chat spam


TobleroneSwordfish (approved):
God yes, we should do this more often.

PR #19105, opened 2024-05-23
Standardize a few security things

### LINKED ISSUE #18181: Clarion Anti-Bio crate has armory requirement despite not being in armory.
### Labels

[BUG]

### Description

Clarion Anti-Bio crate has armory requirement despite not being in armory.

### Steps to reproduce

1. Set the map to clarion
2. Have all access
3. Attempt to open anti-biological crate
4. Access denied

### Expected Behavior

Crate should unlock

_No response_

Retrino:
## About the PR 
This PR does a few things:
- Expands the armories on Cogmap 1, Atlas, Oshan, and Clarion slightly. This was to fit the anti-bio crates in the armory, and is only by a tile in most cases. Pics below.
- Anti-bio crates are now no longer in the main security rooms and will always be found in the armory.
- Interrogation rooms without intercoms have been given them, so that the rest of security (or the AI) can listen in from outside.
- Removed a box of laser eyes from oshan's security locker, not sure why they even had those.
- fixes #18181

Screenshot (719)
Screenshot (720)
Screenshot (721)
Screenshot (722)
Screenshot (724)
Screenshot (725)

## Why's this needed? 

Bioweapon crates having armory access while being in secmain is kind of weird, and the access sometimes wasn't consistent (staring at cog1). They're now all in the armory. Intercoms being in interros was brought up the other day, so I went ahead and put them in the ones missing them. It was requested on discord for cog1, but oshan didn't have one either. 

## Changelog 

```changelog
(u)Retrino
(+)Anti-bioweapon crates have been moved to their proper homes in the armory.
```

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -1796,8 +1796,7 @@
 /area/station/crewquarters/cryotron)
 "agC" = (
 /obj/machinery/light_switch/auto,
-/obj/storage/secure/crate/plasma/armory/anti_biological,
-/obj/item/gun/flamethrower/assembled/loaded,
+/obj/machinery/vending/security,
 /turf/simulated/floor/red/side{
 	dir = 10
 	},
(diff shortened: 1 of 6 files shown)


Retrino:
> That issue was raised before the clarion rework but it has carried over to it...

i considered doing it this way, giving the crates the same security access outside of the armory, but theres a couple reasons why i went this route. flamethrowers being standardized as normal security gear doesnt make much sense. it doesnt seem like an issue that comes up even on cog1 much, but id rather not mechanically incentivize security players toward using fire to solve problems. blobs being the only real consistent antag that theyre effective on, aside from shamblers, but unless one shows up in the security lobby i dont see that crate helping them out much. every other "lethal" is locked behind the armory, and i think its reasonable that a severe threat like blobs warrants players to come together enough (captain and secoffs to auth, or even hop/captain giving other crew access to deal with the problem) to get the big guns out. flamethrowers can absolutely wreck a blob before it has its ectothermids, and are one of the easier to craft weapons around. i think that its a reasonable balance that a shut-down fuckoff weapon like that toward blobs have some slight barrier to overcome in place.

JORJ949:
> i considered doing it this way...

Good point, especially with the increase in players they may not realise this isnt meant to be standard equipment.

flrsh (approved):
Makes sense to me!

TTerc (approved):

PR #19107, opened 2024-05-23
TGUI Chemicompiler

Garash2k:
[UI][QoL][Feature][Game Objects]

## About the PR 
Replaces the old Chemicompiler WebElement UI with a TGUI one.
image
Inspired by ruben-svensson's UI https://github.com/orgs/goonstation/projects/7?pane=issue&itemId=5535040
Almost no functional changes have been done, except for better labels on the reservoir buttons, indicating the amount in the beaker instead of changing the button's color like it used to.

6 reservoirs portable version:
dgpacm

Light mode:
y8l7am

I haven't looked into fancier theming or features like proper comment support or syntax highlighting. I think this could be done in another PR or if requested.

## Why's this needed? 
Modern looking UI that's more responsive and allows to deprecate WebElements

## Changelog 

```changelog
(u)Garash2k
(+)New UI for the stationary and portable Chemicompiler.
```


ZeWaka:
thank god webelemnts can die

tgui/packages/tgui/interfaces/ChemiCompiler/ChemiCompilerTextArea.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const ChemiCompilerTextArea = (_props, context) => {
+  const { act, data } = useBackend<ChemiCompilerData>(context);
+  const { inputValue, loadTimestamp, theme } = data;
+  return (
+    <TextArea
+      value={inputValue}
+      onInput={(_event, value) => { act('updateInputValue', { value }); }}
+      height="100%"
+      // The load button would break if we pressed it between the input's act and the next refresh.
+      // This ensures a refresh after every load button click
```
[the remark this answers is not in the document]
Garash2k: The input value isn't stored locally so there's nothing to update. Ideally it would be stored locally using a local state, then it could be updated, but I couldn't figure out a way to make the load button not constantly override the input after a load click. It's hard to make the whole thing work without a single callback.

mordent-goonstation:
Just putting a blanket "please don't merge until I've reviewed" on this to devs, because I see some stuff I want to change but also do not have the spoons to deal with chemicompiler code right now so I'm going to sit on it for a bit.

mordent-goonstation:
Something you can do while you wait for me to review: could we get a screenshot of it in lightmode added to the PR?

tgui/packages/tgui/interfaces/ChemiCompiler/ChemiCompilerMemory.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Button, Section, Stack } from '../../components';
+
+export const ChemiCompilerMemory = (_props, context) => {
+  const { act, data } = useBackend<ChemiCompilerData>(context);
+  const { buttons, sx, tx, ax, theme } = data;
+  return (
+    <Section title="Memory">
+      sx: <em>{sx}</em> -
+      ax: <em>{tx}</em> -
+      tx: <em>{ax}</em>
```
mordent-goonstation, line 19: All of this review comes with the caveat of "I don't know the ChemiCompiler well", but:

Flat text label and italic values here are bleh, given the screenshot context. Bare minimum consider putting these in their own `Stack` with `grow` applied to the items so they fill the space. `-` is not a good "spacer" these days.

tgui/packages/tgui/interfaces/ChemiCompiler/ChemiCompilerMemory.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <Stack wrap>
+        {buttons.map((button, index) => (
+          <Stack.Item key={index} style={{ border: `1px solid ${theme === "syndicate" ? "#397439" : "#88bfff"}`, "border-radius": "0.2rem" }} p={1} m={1}>
```
mordent-goonstation, line 22: Rather than styling the `Stack.Item` consider putting a `Section` inside it with the `M#` as titles. I think that should have decent default styling in this use case. No issue with buttons as content for these, I think that works here.

tgui/packages/tgui/interfaces/ChemiCompiler/ChemiCompilerTextArea.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { ChemiCompilerData } from './type';
+import { TextArea } from '../../components';
+
+export const ChemiCompilerTextArea = (_props, context) => {
+  const { act, data } = useBackend<ChemiCompilerData>(context);
+  const { inputValue, loadTimestamp, theme } = data;
+  return (
+    <TextArea
+      value={inputValue}
+      onInput={(_event, value) => { act('updateInputValue', { value }); }}
```
mordent-goonstation, line 18:
```suggestion
      onInput={(_event, value) => act('updateInputValue', { value })}
```

tgui/packages/tgui/interfaces/ChemiCompiler/ChemiCompilerTextArea.tsx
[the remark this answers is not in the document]
mordent-goonstation: ~~Prompting with an idea, as I know it's moderately involved so I can help with the details if needed (please ping in coder channel Discord if you want to talk about it):~~

~~Could you use local state for the value, have a load button to "submit" that value, and then in the `ui_act` part of that include the timestamp as a `last_updated` field (alternatively, just increment a number if that doesn't give the right granularity). Include that in the `ui_data`, then use a diff between that value and what you last saw it as (in a local state) to determine if you need to update the text area's value or not?~~

I'm leaving the above in for posterity because funny, I wrote it out without really reading the code and riffing off of ZeWaka's idea instead of reading what you have. I think using a `loadTimestamp` is absolutely fine. I would be a little cautious of using it as the `key` as that involves unmounting/remounting and may lose focus, but would need to play with it a bit to see if we care about that.

tgui/packages/tgui/interfaces/ChemiCompiler/ChemiCompilerTextArea.tsx
```diff
@@ -0,0 +1,28 @@
...
+      height="100%"
+      // The load button would break if we pressed it between the input's act and the next refresh.
+      // This ensures a refresh after every load button click
+      key={loadTimestamp}
+      fontFamily="Consolas"
```
mordent-goonstation, line 23: Please include a backup font of `monospace`. I've included what I think it could be below but not as a suggestion because I haven't verified syntax/that this works. Set it to something like `"thisfontdoesnotexist, monospace"` and check behavior.

```
      fontFamily="Consolas, monospace"
```

Future work: would be good to augment `TextArea`'s this way so you can just pass in a `monospace` prop.

tgui/packages/tgui/interfaces/ChemiCompiler/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const { data } = useBackend<ChemiCompilerData>(context);
+  const { theme } = data;
+  return (
+    <Window width={600} height={500} theme={theme}>
+      <Window.Content>
+        <Stack fill>
+          <Stack.Item grow>
+            <ChemiCompilerTextArea />
+          </Stack.Item>
+          <Stack.Item basis={18} textAlign="center">
```
mordent-goonstation, line 26: Nice clean layout. 👍 

Only suggestion here is to move `basis={18}` to `basis={SIDEBAR_WIDTH}` or whatever, then define that above. Magic numbers bad.

tgui/packages/tgui/interfaces/ChemiCompiler/type.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * @file
+ * @copyright 2024
+ * @author Garash2k (https://github.com/Garash2k)
+ * @license ISC
+ */
+
+interface button {
```
mordent-goonstation, line 8:
```suggestion
interface ButtonData {
```

tgui/packages/tgui/interfaces/ChemiCompiler/type.ts
```diff
@@ -0,0 +1,22 @@
...
+  button: boolean;
+  cbf: boolean;
+}
+
+export interface ChemiCompilerData {
+  reservoirs: Array<string>;
+  buttons: Array<button>;
```
mordent-goonstation, line 15:
```suggestion
  buttons: Array<ButtonData>;
```

code/modules/chemistry/chemicompiler_core.dm
```diff
@@ -211,25 +140,18 @@
...
+	. = list()
+	.["inputValue"] = output
+	.["loadTimestamp"] = loadTimestamp
+	.["sx"] = sx || 0
+	.["tx"] = tx || 0
+	.["ax"] = ax || 0
+
+	.["buttons"] = new/list(6)
+	for(var/i = 1, i <= 6, i++)
+		.["buttons"][i] = list(button = !!buttons[i], cbf = !!cbf[i])
```
mordent-goonstation, line 153:
```suggestion
	. = list(
		"inputValue" = src.output,
		"loadTimestamp" = src.loadTimestamp,
		"sx" = src.sx || 0,
		"tx" = src.tx || 0,
		"ax" = src.ax || 0
	)
	.["buttons"] = new/list(6)
	for(var/i = 1, i <= 6, i++)
		.["buttons"][i] = list(button = !!buttons[i], cbf = !!cbf[i])
```

code/modules/chemistry/chemicompiler_core.dm
```diff
@@ -885,4 +655,37 @@
...
+		if("reservoir")
+			reservoirClick(params["index"] + 1)
+			return TRUE
+		if("save")
+			core.buttons[params["index"] + 1] = core.output
+			core.cbf[params["index"] + 1] = core.parseCBF(core.output, params["index"] + 1)
+			return TRUE
+		if("load")
+			core.output = core.buttons[params["index"] + 1]
+			core.loadTimestamp = world.timeofday
```
mordent-goonstation, line 683: I never remember how DM timestamps work, not saying this is wrong but commenting to sanity check if this is the correct one to use.

ZeWaka: TIME.

tgui/packages/tgui/interfaces/ChemiCompiler/ChemiCompilerTextArea.tsx
Garash2k: Does work: alt text

tgui/packages/tgui/interfaces/ChemiCompiler/ChemiCompilerTextArea.tsx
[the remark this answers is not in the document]
Garash2k: I don't think there's a focus loss issue. I'll only change when the user clicks the load button, which means that focus is already lost. Plus I'd say with the contents of the textarea changing, a loss of focus is desirable.

mordent-goonstation (approved):
Probably fine. I have some concerns over the management of the text area's value, ideally want to poke that a bit before merging, but broadly 👍

ZeWaka:
bop

PR #19115, opened 2024-05-23
Add armor pierce to syndicate dagger

FlameArrow57:
[GAME OBJECTS][BALANCE]

## About the PR 
Adds 80% armor pierce to the syndicate dagger (mostly affecting revs game mode)

## Why's this needed? 
Syndicate dagger does 5 base damage and has no AP, meaning it actually does 0 dmg against an armor vest. Scalpel in comparison does 5 damage and has 80% armor pierce, so it does 4 dmg against armor vest

Seems pretty common that revheads will attack a security officer without knowing it's doing 0 damage, while still expecting something

## Changelog 

```changelog
(u)FlameArrow57
(+)Syndicate dagger now has 80% armor pierce.
```

### DIFF
--- a/code/obj/item/misc_weapons.dm
+++ b/code/obj/item/misc_weapons.dm
@@ -536,15 +536,11 @@ TYPEINFO(/obj/item/sword/pink/angel)
 	stamina_crit_chance = 50
 	pickup_sfx = 'sound/items/blade_pull.ogg'
 	hitsound = 'sound/impact_sounds/Flesh_Stab_1.ogg'
-
-	var/pierces_armor = FALSE
+	HELP_MESSAGE_OVERRIDE({"Throw the dagger at someone to instantly incapacitate them for a short while."})
 
 	New()
 		..()
 		BLOCK_SETUP(BLOCK_KNIFE)
-
-		if (src.pierces_armor)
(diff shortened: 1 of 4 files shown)


code/obj/item/misc_weapons.dm
```diff
@@ -579,6 +584,7 @@ TYPEINFO(/obj/item/sword/pink/angel)
 /obj/item/dagger/syndicate
 	name = "syndicate dagger"
 	desc = "An ornamental dagger for syndicate higher-ups. It sounds fancy, but it's basically the munitions company equivalent of those glass cubes with the company logo frosted on."
+	pierces_armor = TRUE
 	HELP_MESSAGE_OVERRIDE({"Throw the dagger at someone to instantly incapacitate them for a short while."})
 
 /obj/item/dagger/syndicate/specialist //Infiltrator class knife
```
Tarmunora, line 590: feels like it would make more sense to split the inheritance here, since the specialist dagger doesn't inherit anything other than the help message, which ought to be on the parent anyways

FlameArrow57: Sure

Tarmunora (approved):

PR #19124, opened 2024-05-24
Refactors the transparium reagent and makes its duration visible

Ilysen:
[QoL][Internal][Code Quality][Secret]

## About the PR 
this PR makes the following changes:
* gives transparium a new description
* transparium's effects are now handled with a visible status effect instead of being in a `SPAWN()`. all of the internal logic has been moved into this effect. I used a (seemingly) unused state from `statussystem.dmi` for the icon state.
  * this effect contains logic for a "flickering" state, which makes it invisible and blocks off some of the logic. at the moment, transparium patches cause the mob to rapidly appear and disappear, spamming messages the whole time; I decided it'd be more fun to make this a deliberate feature instead of disabling it outright, and so now whenever the effect has a very low duration, it will appear differently.
  * if the potential for a status effect being added and then removed every tick is too heavyweight, I could disable the application entirely when the duration is very low.
* transparium now has a visible fade in/fade out, instead of setting the mob's alpha with no fanfare.
* diluted transparium is now a subtype of regular transparium with overdose threshold and addiction probably set to zero (i.e. disabled). this is for code cleanliness, so that the logic isn't copy-pasted from one to another.
* general syntax and style refinements across the board.

## Why's this needed? 
transparium's a pretty old reagent and the code's a bit yucky. this should hopefully make it a lot cleaner, as well as more intuitive to use.

I could potentially foresee the "visible duration" part of this PR being undesirable, so if so, I'll happily axe it and just keep the cleanliness parts. no sense in wasting a perfectly good refactor!

### General showcase of appearing and disappearing **(discord ping warning, sorry!!)**
https://github.com/goonstation/goonstation/assets/47678781/dc65de8d-b30c-443e-b859-b7c77a5334f2

### Showcase of the "flickering" effect caused by a transparium patch (the audio's a bit crackly on this clip and I'm not sure why. oops!)
[transparium showcase.webm](https://github.com/goonstation/goonstation/assets/47678781/cd1d9b1c-ff99-439e-b491-539c3781aa8e)

### DIFF
--- a/code/modules/chemistry/Reagents-Misc.dm
+++ b/code/modules/chemistry/Reagents-Misc.dm
@@ -2647,7 +2647,7 @@ datum
 		transparium
 			name = "transparium"
 			id = "transparium"
-			description = "Fading into obscurity..."
+			description = "An exotic compound that intimidates nearby photons upon exiting the body, rendering the user invisible for a period of time proportional to how long it was present in their bloodstream."
 			reagent_state = LIQUID
 			fluid_r = 255
 			fluid_g = 255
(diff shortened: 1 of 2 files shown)


444Portal:
status effect should probably that this hides you from thermals and the AI too (lol, or maybe just remove that bit? thats more balance-y than refactor so *shrug)

Tarmunora (approved):

Tarmunora:
Also, please try to avoid force-pushing to PR branches when possible, it makes it very hard to tell what's actually been changed between review passes

Ilysen:
alright! thank you for letting me know. I'll avoid that in future.

PR #19126, opened 2024-05-24
Increase radius of "cateyes" vision

Glamurio:
[Feature] [Traits]

## About the PR 
Changes `icons/effects/vision_cateyes.dmi` to be slightly larger in radius, nestling a bit above thermal vision, but with generally a decreased luminosity.

With cat eyes:
image
Without cat eyes:
image

## Why's this needed? 
I've heard the prevalent opinion that this perk is "never worth taking", and I think it's a shame because it's kind of neat. This makes sure it's better than regular vision by a noticeable amount, but still marginally worse than nightvision.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)People with the "cat eyes" trait will find themselves a bit better at seeing in the dark.
```


TDHooligan:
idk if its my screen but it looks like cat eyes is harder to see with than regular sight with this change

Glamurio:
> idk if its my screen but it looks like cat eyes is harder to see with than regular sight with this change

What? I don't understand how this could happen, you literally just have a tiny circle with regular sight. I can turn up the brightness but I am confused regardless.

Azrun:
Looks like the overall value of cateyes may be less than default?  Value/Lightness went way down with size expansion. 

As others mentioned, I can't see the character in the cateyes.

 'icons/effects/vision_default.dmi' vs 'icons/effects/vision_cateyes.dmi' Could you post that for reference, the icon states.

Glamurio:
> Looks like the overall value of cateyes may be less than default? Value/Lightness went way down with size expansion.
> 
> As others mentioned, I can't see the character in the cateyes.
> 
> 'icons/effects/vision_default.dmi' vs 'icons/effects/vision_cateyes.dmi' Could you post that for reference, the icon states.

No, I definitely purposefully made the lightness lower, but if it's such a big problem, I'll gladly turn it up. Maybe it's MY monitor, but it felt more than bright enough. I'll adjust it.

MintyPhresh:
Seems a bit extreme now. Do you have view tinting turned off? I notice your Captain is wearing sunglasses.

Glamurio:
> Seems a bit too bright now. Do you have view tinting turned off? I notice your Captain is wearing sunglasses.

They were all taken with sunglasses on. I forgot that tints the view, but they're at least all consistent.

Glamurio:
> With cat eyes: image Without cat eyes: image
> 
Alright, I accidentally in-between had the wrong radius saved. This is now the actual same size with increased visibility
image

Here's also a before and after of the radius
image

ZeWaka:
can you update the top post

Glamurio:
> can you update the top post

Done.

frawhst (approved):
Looks fine to me

ZeWaka (approved):

PR #19127, opened 2024-05-24
Wizard spell rebalance/rework

FlameArrow57:
[GAME MODES][BALANCE][REWORK][ADD TO WIKI]

## About the PR 
This PR does a rebalance/rework of wizard spells

See design doc here for all changes/reasoning:
https://hackmd.io/@FlameArrow57/rJzcf8vXA

## Why's this needed? 
Hopefully makes wizards game mode more enjoyable (see design doc for more details)

## Changelog 

```changelog
(u)FlameArrow57
(*)Large rebalance of Wizard spells. See minor changelog for details.
(+)Using a defensive spell now places all other defensive spells on a 5 second cooldown.
(+)Magic missile cooldown increased to 25 seconds.
(+)Empower spell duration and cooldown decreased to 15 and 30 seconds. It also removes 10 seconds of knockdown on use (relevant to stuns).
(+)Spell shield is reworked so that the shield has limited health, rather than 100% immunity, for its duration. Effectiveness from least to most is in the order of: melee attacks, baton stuns, ranged stuns, lethal projectiles, explosions.
(+)Forcewall duration decreased to 10 seconds.
(+)Animate Dead can now be used on alive monkeys.
(+)Prismatic Spray can now be used without a staff.
```

### DIFF
--- a/_std/defines/component_defines/component_defines_atom.dm
+++ b/_std/defines/component_defines/component_defines_atom.dm
@@ -215,7 +215,7 @@
 	#define COMSIG_MOB_GRABBED "mob_grabbed"
 	/// Sent when a mob emotes (emote, voluntary, emote target)
 	#define COMSIG_MOB_EMOTE "mob_emote"
-	/// Sent when a mob is checking for an active energy shield
+	/// Sent when a mob is checking for an active shield
 	#define COMSIG_MOB_SHIELD_ACTIVATE "mob_shield_activate"
 	/// Sent when a mob flips, return TRUE to skip the rest of the flip emote coded, argument is (voluntary)
 	#define COMSIG_MOB_FLIP "mob_flip"
--- a/_std/defines/component_defines/component_defines_special.dm
+++ b/_std/defines/component_defines/component_defines_special.dm
@@ -73,6 +73,8 @@
 // ---- energy shield thing ----
 	/// Sent by the itemability to toggle the energyshield component
 	#define COMSIG_SHIELD_TOGGLE "energy_shield_toggle"
+	/// Used by the item to check if the shield is active or not
+	#define COMSIG_CHECK_SHIELD_ACTIVE "check_energy_shield_active"
 
 // ---- atom property signals ----
 
--- a/_std/defines/component_defines/component_defines_special.dm
+++ b/_std/defines/component_defines/component_defines_special.dm
@@ -70,8 +70,8 @@
 		/// Cell is fully charged
 		#define CELL_FULL 32
 
-// ---- energy shield thing ----
-	/// Sent by the itemability to toggle the energyshield component
+// ---- shield component signals ----
+	/// Sent by the itemability to toggle the shield component
 	#define COMSIG_SHIELD_TOGGLE "energy_shield_toggle"
 	/// Used by the item to check if the shield is active or not
 	#define COMSIG_CHECK_SHIELD_ACTIVE "check_energy_shield_active"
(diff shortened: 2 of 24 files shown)


TDHooligan:
ignoring the other changes, the defense spell cooldown thing might not actually help as much as you think - the key math is that having 2 or more defensive spells on cooldown at all is essentially a 50% cooldown reduction. 3 is a 66% as you get 3x the cooldown dedicated to getaway spells.

but also massively penalising defense spells is kinda lame too. i'd be happier seeing how this works at least

flappybatpal:
I think Animate Dead would still be insanely situational. If it could summon a skeleton from under the floor at any non space turf (using the reveal effect) that would be much easier to use.

However, even with a change to make it cast easily I think this spell isn't great. It helps to perma due to gibbing corpses but in general use the skeleton only does about 50 brute to a staff assistant defending themselves unarmed.

FlameArrow57:
> I think Animate Dead would still be insanely situational. If it could summon a skeleton from under the floor at any non space turf (using the reveal effect) that would be much easier to use.
> 
> However, even with a change to make it cast easily I think this spell isn't great. It helps to perma due to gibbing corpses but in general use the skeleton only does about 50 brute to a staff assistant defending themselves unarmed.

Agree that would make it more commonly usable but feels like that starts to fall into usability more than Sticks to Snakes or Summon Golem maybe, which similarly right now have different use cases. Skeletons do also have the ability to self revive. I could try looking at skeleton stats, but not sure about making them stronger, in comparison to snakes and golems

FlameArrow57:
> Because of these reasons I feel wizards are more fun when there's only one or two, and these changes would not do much to actually change the feel of a wizard round apart from forcing them to split up, which might be even harder for the crew to deal with. Just in my opinion, but if there's a focus on wanting the wizard gamemode to be around more, I'd like that gamemode to have more to it than just wizards trying to kill people. Some objectives maybe?

Intent is to weaken abilities not to the point where it decentivizes splitting up, but still enough to reduce how unstoppable it is. I think changing objectives could be fun, but unless the best option, I will probably just be keeping the scope of this PR to spell changes

> My suggestion would just be to limit wizards to only buying one evasive spell, assuming they'd still have phase shift default.

Makes sense, but I was thinking restricting spells you have starts to go into the realm of being too restrictive, in terms of spell combos and decreasing spell variety. More evasive spells can be annoying, but there's still a trade off with what you spend your spell points on.

Flaborized:
Also, in the process of writing my response I've also thought of an alternative way to make having reduced defensive abilities Fun: if all wizards had a limited, unrecoverable resource that let them "get out of jail" a few times, you could solve a lot of issues with fighting them. You bridge the gap between "Live Forever" and "Die Instantly" which are both quite bad; strong players can't live forever with the limited defense but newer players have a few chances to not die quickly. You let wizards take more Risks cuz they have a buffer built up for it. The crew could feel like they're doing Something Meaningful when they force the wizard to use it cuz it's limited and they're pushing them closer to actual death (though you'd want to communicate this Strongly to the crew). Tele scroll and soulguard are kinda like this but Bad, but I think implemented well it could be pretty fun.

This is just an Idea I had though re: this topic, please don't take me suggesting it as a "You have to Do This", and I'm not particularly sure how other devs think on this whole PR so also probs ask them if you do want to do it. There's probably a bunch of ways to make this work and I think that adjusting wiz to be more fun to play as and play against is Good.

FlameArrow57:
Okay marking this PR ready for review. Everything I put in the design doc has been implemented, except for scaling spells when in a group. I thought it would help with group deadliness but seems disliked maybe.

Also, I think something intermediate based on what Flaborized suggested, around the die instantly versus live forever problem that can appear would be nice, would need to think more on the implementation though

PR #19139, opened 2024-05-25
Makes Stowaway a latejoin role

MintyPhresh:
## About the PR 
This PR delists Stowaway as a round-start trait, instead attaching it to a latejoin role. All other behaviors of Stowaway remain the same.

The new latejoin role:
**Stowaway:** 2 slots. Typical Stowaway container start, with nothing but a jumpsuit, headset, and makeshift tools for door hacking (crowbar, scissors, and screwdriver, so you don't get locked in maints)

## Why's this needed? 
Characters with the Stowaway trait are in this stupid position of having half a job. On one hand, you're clearly recognizable as whatever job you signed up for-- Scientist Stowaways show up garbed in gasmask and labcoat, Bartender Stowaways have their trusty break-action at their hip, Security Stowaways start with their official NT badges proving they're an officer, and Captain Stowaways start with the Nuclear Authentication disk. At the same time, though, they're totally unemployed-- they lack their ID and PDA for doing their job, and they aren't paid wages because NT didn't put them on the manifest. 

Because Stowaways have half a job, it makes it very easy to simply give the Stowaway whatever job their character would have had if they weren't Stowaways. It's a very, very common practice for Captains and HoPs to just completely ignore the fact that NT *did not want this person working today* and give them the ID, PDA, and whatever else they need. In a way, it gets interacted with in a similar way to Jailbird-- a character that consistently stows away will just get the ID they need, completely defeating the point of taking the trait. There's an argument here that Stowaways should be given the ID-- after all, the player needs it to actually play the game and do their job-- but, again, they took the trait. This is such an accepted practice that I've experienced goon-2-esque conflicts where a secoff threatens to throw the HoP across the room because they won't give a Stowaway secass an ID with sec access.

And, tangentially, this status quo makes Stowaway a powergamer trait for antag rounds. Because you aren't deprived of your PDA, ID, or job title thanks to the ennui of whoever's behind the ID console, the trait becomes a +1 that scrubs your name from the sec records that'd be used to set you to arrest. Even when you're not antag, the trait's core concept can invite characters that are "rebellious punks who Aren't With You People," which is never fun to deal with.

Stowaways, in their current state as a thing you can do every round, don't make sense for any job. Secoffs with shiny badges shouldn't be ignoring the wishes of whoever schedules shifts at NT to break into their workplaces and hang out. Stowaway Captains shouldn't be issued nuclear authentication disks, or be respected as an authority by the crew when NT didn't even schedule them. (If they aren't issued IDs or PDAs, why the heck do they get the rest of this stuff?) People who have valuable labor to contribute shouldn't be repeatedly putting in the extra effort to sneak into a space station just to work for free. 

## Changelog 

```changelog
(u)Mintyphresh
(*)Stowaway can no longer be taken as a trait, and is instead a latejoin role.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -1487,6 +1487,31 @@ ABSTRACT_TYPE(/datum/job/civilian)
 		src.access = get_access("Mail Courier")
 		return
 
+/datum/job/special/stowaway
+	name = "Stowaway"
+	limit = 2
+	wages = 0
+	low_priority_job = TRUE
+	slot_back = list(/obj/item/storage/backpack/satchel/anello)
+	slot_belt = list(/obj/item/crowbar)
+	slot_foot = list(/obj/item/clothing/shoes/brown)
+	slot_jump = list(/obj/item/clothing/under/color/grey)
+	slot_head = list(/obj/item/clothing/head/green)
+	slot_ears = list(/obj/item/device/radio/headset/civilian)
+	slot_poc1 = list(/obj/item/currency/spacecash/fivehundred)
+	slot_poc2 = list(/obj/item/scissors)
+	slot_lhan = list(/obj/item/screwdriver)
+
+	items_in_backpack = list(/obj/item/currency/spacecash/buttcoin = 2)
+	rounds_needed_to_play = 10 // It would really suck ass to play this as a new player
+	// missing wiki link, does not have a mention on https://wiki.ss13.co/Jobs
+
+	special_setup(var/mob/living/carbon/human/M)
+		..()
+		if (!M)
+			return
+		M.traitHolder.addTrait("stowaway")
+
 // randomizd gimmick jobs
 
 /datum/job/special/random
@@ -1915,30 +1940,6 @@ ABSTRACT_TYPE(/datum/job/civilian)
 	slot_lhan = list(/obj/item/storage/toolbox/artistic)
 	items_in_backpack = list(/obj/item/canvas, /obj/item/canvas, /obj/item/storage/box/crayon/basic ,/obj/item/paint_can/random)
 	// missing wiki link, does not have a mention on https://wiki.ss13.co/Jobs
-
-/datum/job/special/random/stowaway
-	name = "Stowaway"
-	limit = 3
-	wages = 0
-	slot_back = list(/obj/item/storage/backpack/satchel/anello)
-	slot_belt = list(/obj/item/crowbar)
-	slot_foot = list(/obj/item/clothing/shoes/brown)
-	slot_jump = list(/obj/item/clothing/under/color/grey)
-	slot_head = list(/obj/item/clothing/head/green)
-	slot_ears = list(/obj/item/device/radio/headset/civilian)
-	slot_poc1 = list(/obj/item/currency/spacecash/fivehundred)
-	slot_poc2 = list(/obj/item/scissors)
-	slot_lhan = list(/obj/item/screwdriver)
-
-	items_in_backpack = list(/obj/item/currency/spacecash/buttcoin = 2)
-	rounds_needed_to_play = 10 // It would really suck ass to play this as a new player
-	// missing wiki link, does not have a mention on https://wiki.ss13.co/Jobs
-
-	special_setup(var/mob/living/carbon/human/M)
-		..()
-		if (!M)
-			return
-		M.traitHolder.addTrait("stowaway")
 
 #ifdef HALLOWEEN
 /*
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -1572,18 +1572,15 @@ ABSTRACT_TYPE(/datum/job/civilian)
 	slot_poc1 = list(/obj/item/currency/spacecash/fivehundred)
 	slot_poc2 = list(\
 	/obj/item/scissors,
-	/obj/item/wirecutters/red,
+	/obj/item/wirecutters,
 	/obj/item/wirecutters/yellow,
-	/obj/item/wirecutters/blue,
-	/obj/item/wirecutters/gray,
+	/obj/item/wirecutters/grey,
 	/obj/item/wirecutters/orange,
 	/obj/item/scissors/surgical_scissors)
 	slot_lhan = list(\
 	/obj/item/screwdriver,
-	/obj/item/screwdriver/red,
 	/obj/item/screwdriver/yellow,
-	/obj/item/screwdriver/blue,
-	/obj/item/screwdriver/gray,
+	/obj/item/screwdriver/grey,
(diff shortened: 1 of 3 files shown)


MintyPhresh:
I can see your point-- the gameplay issues with stowaway definitely aren't as important as the worldbuilding aspects to me. But using up time, skill-issuing Command and Sec, and causing confusion/chaos sound like things antags do, not crew (clown excluded).

Glamurio:
I actually really like this, but my issue with this addition is in removing personality from a character. This issue lies primarily in "latejoining" as a system; it means that if I want to play my stowaway character, I now can't do so unless I happen to see it in the roster. I'd love for this to lead to a follow-up about looking into specialized roles in general, there's a lot of cool ones that rarely see play because you a) need to be lucky enough to have them show up and b) you can only play as them if you latejoin.

Since we currently still have the issue of people latejoining vs. readying up at roundstart, I would personally prefer if all of these separate jobs could be consistently queued for with the same priority as a "staff assistant", if that makes sense. I'd sooner prefer playing a "stowaway" than a "staff assistant" any day of the week, and I'm not taking away real estate since I was only ever competing with other staffies.

MintyPhresh:
That's intentional; I don't think stowaways should be an every-round thing, as I mentioned when I compared it to Jailbird. There's something weird about "oh, this is scrumbo, he doesn't work here but we let him sleep in the dingy maints and give him staffie access every shift because we love him" in a... workplace. One that has the possibility of being attacked by infiltrators...

I will be sad to see recurring Stowaway characters go if this gets merged, since they're usually pretty funny, but I think it'll be better for the game overall

code/datums/jobs.dm
```diff
@@ -1487,6 +1487,31 @@ ABSTRACT_TYPE(/datum/job/civilian)
...
+	low_priority_job = TRUE
+	slot_back = list(/obj/item/storage/backpack/satchel/anello)
+	slot_belt = list(/obj/item/crowbar)
+	slot_foot = list(/obj/item/clothing/shoes/brown)
+	slot_jump = list(/obj/item/clothing/under/color/grey)
+	slot_head = list(/obj/item/clothing/head/green)
+	slot_ears = list(/obj/item/device/radio/headset/civilian)
+	slot_poc1 = list(/obj/item/currency/spacecash/fivehundred)
+	slot_poc2 = list(/obj/item/scissors)
+	slot_lhan = list(/obj/item/screwdriver)
```
TobleroneSwordfish, line 1503: These vars accept lists to randomly pick from, this seems like a great opportunity to randomize the clothes they're given so it's not possible to learn The stowaway outfit and spot them from that.

code/datums/jobs.dm
```diff
@@ -1487,6 +1487,31 @@ ABSTRACT_TYPE(/datum/job/civilian)
...
+
+	items_in_backpack = list(/obj/item/currency/spacecash/buttcoin = 2)
+	rounds_needed_to_play = 10 // It would really suck ass to play this as a new player
+	// missing wiki link, does not have a mention on https://wiki.ss13.co/Jobs
+
+	special_setup(var/mob/living/carbon/human/M)
+		..()
+		if (!M)
+			return
+		M.traitHolder.addTrait("stowaway")
```
TobleroneSwordfish, line 1513: Have you tested this works to spawn them in the right place? Roundstart init order is notoriously messy.

code/datums/traits.dm
```diff
@@ -754,6 +754,7 @@ ABSTRACT_TYPE(/datum/trait/job)
 	icon_state = "stowaway"
 	category = list("background")
 	points = 1
```
TobleroneSwordfish, line 756:
```suggestion
	points = 0
```
It might end up forcing an invalid trait config somehow otherwise.

TobleroneSwordfish (changes requested):

code/datums/jobs.dm
MintyPhresh: I have!

code/datums/jobs.dm
```diff
@@ -1487,6 +1487,31 @@ ABSTRACT_TYPE(/datum/job/civilian)
...
+	slot_foot = list(/obj/item/clothing/shoes/brown)
+	slot_jump = list(/obj/item/clothing/under/color/grey)
+	slot_head = list(/obj/item/clothing/head/green)
+	slot_ears = list(/obj/item/device/radio/headset/civilian)
+	slot_poc1 = list(/obj/item/currency/spacecash/fivehundred)
+	slot_poc2 = list(/obj/item/scissors)
+	slot_lhan = list(/obj/item/screwdriver)
+
+	items_in_backpack = list(/obj/item/currency/spacecash/buttcoin = 2)
+	rounds_needed_to_play = 10 // It would really suck ass to play this as a new player
```
Tarmunora, line 1506: Eh, I'd rather not be adding round reqs to stuff that doesn't *need* it.

code/datums/jobs.dm
Tarmunora, line 1513: Hm. Would it not be better to move the trait's behaviour to the job datum if possible?

Tarmunora (changes requested):

code/datums/jobs.dm
MintyPhresh: My concern is that the starting kit expects you to know how to hack doors-- if you don't, you're stuck in a dark, dingy maintenance shaft (most of the time). Unlucky players need to have some level of skill to have a good time with Stowaway, so I do see it as a need...

code/datums/jobs.dm
MintyPhresh: All of the other job traits (e.g. Engineering Training) are part of the traits file, so I think this is OK...

TobleroneSwordfish (approved):

code/datums/jobs.dm
TobleroneSwordfish: I'm just going to remove this for now, we can re-add it if really needed. We really want to keep round locks to an absolute minimum.

code/datums/jobs.dm
```diff
@@ -1487,6 +1487,112 @@ ABSTRACT_TYPE(/datum/job/civilian)
...
+	/obj/item/wirecutters/orange,
+	/obj/item/scissors/surgical_scissors)
+	slot_lhan = list(\
+	/obj/item/screwdriver,
+	/obj/item/screwdriver/yellow,
+	/obj/item/screwdriver/grey,
+	/obj/item/screwdriver/orange)
+
+	items_in_backpack = list(/obj/item/currency/spacecash/buttcoin = 2)
+	rounds_needed_to_play = 10 // It would really suck ass to play this as a new player
```
TobleroneSwordfish, line 1587:
```suggestion
```

PR #19140, opened 2024-05-25
Add magnetized subtype for arcfiend that ignores unionized trait

TealSeer:
[PLAYER ACTIONS] [BALANCE] 
## About the PR 

Adds a subtype for the magnetized effect that ignores the presence of the "unionized" trait on targets

## Why's this needed? 

Unionized is an incredibly common trait to take and it conferring immunity to an antag ability is unreasonable. 

## Changelog 

```changelog
(u)TealSeer
(+)People with the Unionized trait are no longer immune to arcfiend's polarize ability. Immunity to bio-magnetic random events remain unchanged.
```


AXLFAN2000:
This is a good idea IMO, unionized has always been super popular but it's only been even more popular with the trait cap removed.  It's silly that this move gets owned so hard by unionized being such a "meta" trait pick

ZeWaka (approved):

PR #19144, opened 2024-05-26
Refactor job definitions & remove redundancy

Glamurio:
[Respawning] [Internal]

## About the PR 
Refactors `jobs.dm`, adding `var/list/training` which is a list of all training given to the job.
Additionally adds `var/access_string` which handles string-assigned accesses.

Creates a new `/datum/job/special/random/hollywood` which was previously just `/datum/job/special/random`.
More advanced special job mechanics will be handled in a different PR.

Moves the 15% chance to see ghosts from the chaplain to "training_chaplain" instead.
Also adds chaplain training to the Angel. Because it just makes sense.

NOTE:
I left a few jobs unchanged, such as "Nanotrasen Special Operative", because they had a different kind of access handling, maybe in a future PR I'll refactor those as well under one big "grant_access" proc or something.

## Why's this needed? 
This is the first of an on-going series of job-related refactors in an attempt to make the code more readable, maintainable and ultimately, better.


code/datums/jobs.dm
```diff
@@ -7,6 +7,7 @@
 	/// Job starting wages
 	var/wages = 0
 	var/limit = -1
+	var/list/training = list() // specific job training string, i.e. "training_security"
```
Tarmunora, line 10: rename to trait_list or something similar

code/datums/jobs.dm
```diff
@@ -61,6 +62,7 @@
 	var/list/slot_rhan = list()
 	var/list/items_in_backpack = list() // stop giving everyone a free airtank gosh
 	var/list/items_in_belt = list() // works the same as above but is for jobs that spawn with a belt that can hold things
+	var/access_string = null // used to quickly grab access via string, i.e. "Chief Engineer"
 	var/list/access = list(access_fuck_all) // Please define in global get_access() proc (access.dm), so it can also be used by bots etc.
```
Tarmunora, line 66: comment should specify that it completely overrides `var/list/access` if non-null

code/datums/traits.dm
```diff
@@ -646,6 +646,8 @@ ABSTRACT_TYPE(/datum/trait/job)
 	onAdd(mob/living/owner)
 		if(owner.traitHolder?.hasTrait("atheist"))
 			src.faith_mult = 0.2
+		if (prob(15))
+			owner.see_invisible = INVIS_GHOST
```
Tarmunora, line 650: might as well axe this entirely, this implementation of chaplain see_invis still doesn't work and I'm not sure fixing it would be a good thing lol

Tarmunora (changes requested):

PR #19146, opened 2024-05-26
Refactor player count request/response into server sync

Sovexe:
## About the PR 
* More robust cross server data syncing
* Now additionally syncs map, next map, and round time with the ability to easily add more

## Why's this needed? 
* Ability to display more info about other servers locally


code/datums/controllers/process/cross_server_sync.dm
```diff
@@ -0,0 +1,13 @@
+/// proocess for flock_structure objects
+/datum/controller/process/cross_server_sync
+	setup()
+		name = "Cross Server Sync"
+		schedule_interval = 1 MINUTE
+
+#ifdef LIVE_SERVER
+		for(var/datum/game_server/game_server as anything in game_servers.servers)
+			game_server.sync_server_data()
```
mordent-goonstation, line 9: I don't mess with these things, but is this not just "do it immediately as well as when the scheduler says so"?
```suggestion
		src.doWork()
```

code/modules/cross_server/cross_server_messages/server_sync/server_sync_response.dm
```diff
@@ -0,0 +1,22 @@
+/datum/cross_server_message/server_sync_response
+	name = "server_sync_response"
+
+	receive(list/data, datum/game_server/server)
+		server.player_count = data["player_count"]
+		server.map = data["map"]
+		server.next_map = data["next_map"]
+		server.round_time = data["round_time "]
```
mordent-goonstation, line 8:
```suggestion
		server.round_time = data["round_time"]
```

code/datums/controllers/process/cross_server_sync.dm
Sovexe: yes, because loops dont start running until round start but they do setup

code/datums/controllers/process/cross_server_sync.dm
```diff
@@ -0,0 +1,13 @@
+/// proocess for flock_structure objects
```
ZeWaka, line 1: 😬

PR #19150, opened 2024-05-26
Manufacturers: Combine free_resource_amt and free_resource_amounts into free_resources

MeggalBozale:
## About the PR 
Combines the variables which describe amounts of free resources to put into a manufacturer by default into one associated list, free_resources, which is keyed by material pieces with values of the amount (in bars) to add.

## Why's this needed? 
Improves code quality of manufacturer by allowing easier specificity of the amounts of each resource to use, rather than a uniform amount across all bars. Also removes the need for two separate variables on an object to do the same thing

### Might need to update secret content


MeggalBozale:
Obviously I wouldn't know, but,
**This change might require secret changes in case those maps use free_resource_amt and similar**

maps/unused/ozymandias.dmm
```diff
@@ -84653,7 +84653,7 @@
 /area/station/quartermaster/office)
 "vJl" = (
 /obj/machinery/manufacturer/general/grody{
-	free_resource_amt = 2
+	free_resources = list()
```
frawhst, line 84656: very minor but this is a functional change (should have 2 of everything)

frawhst:
Also, it looks like you can't do multiline lists in `dmm` files? checks seem to think so, at least.

maps/unused/ozymandias.dmm
MeggalBozale: Huh, didn't know. grody manufacturers clear their free_materials list on the type, so they shouldnt have anything really

MeggalBozale: image
it never has anything though. I can give it materials by default if this is desired

MeggalBozale: (screenshot from main branch)

MeggalBozale:
Fixed the lists, the only check failing is the map diff bot (broke, last i heard from zewaka) so it should be much better now

maps/unused/ozymandias.dmm
frawhst: Oh, whoops. In that case, just remove the line entirely? Since the varedit on the map isn't actually doing anything.

frawhst (approved):
Can't easily do secret changes right now, but all this lgtm.

maps/unused/ozymandias.dmm
MeggalBozale: Ah true, ha

PR #19152, opened 2024-05-26
Adds overall health pip to health monitor implants

BatElite:
## About the PR 
Adds a little pip (internally the regular health bar) to the center of the expanded health HUD you get from the health monitor implant.
image

## Why's this needed? 
People sometimes avoid the health monitor implant because it obfuscates how well you're doing overall. Splitting the damage and slightly different thresholds for the graphics means you couldn't tell effectively whether you were in crit or not. We may as well make it a straight upgrade.

## Changelog 

```changelog
(u)BatElite
(+)Health monitor implants now show a central pip for your overall health.
```


Flaborized:
I like these and they look good though I think ideally something like this would fill up like a bar as well as having the color, like the default health UI does and the small health implant UI icons pictured do. I added this effect to them because the health indicator is basically the only thing in-game I've ever heard feedback that it is difficult to parse for some colorblind folk, which makes sense because it's going from green to red. Also it would just fit nicely with all the other health UI elements

This might be difficult to implement with the size of some of these current icons being like four pixels tall :sheltersweat:

Sord213:
Oh this is very nice. Even if done without an overall health bar just having an indicator is very good.

BatElite:
Here's that version. The smaller ones update every other stage instead which I think would be decent enough.
image
I don't actually like the bars that much, I find they get a bit hard to read as they run out.

Maybe a bit silly looking, but here's a literal half and half option:
image

AXLFAN2000:
It looks better as just an indicator without the bar IMO

BatElite:
(Should probably note, I didn't make the classic one a bar in those mockups because the quarts aren't either and consistency is nice. I really don't want to go through and edit all the rest of that lifebar.)

Flaborized:
IMO either of these implementations are fine, they don't necessarily have to be done in the same way it's currently done as long as it's also represented by some non-color aspect. The classic HUD doesn't have the bars because I didn't want to touch it because it's "classic" and it's *supposed* to look like the old HUD. It's completely fine then for that HUD to not have them. But yeah I'm fine with it looking like Whatever as long as it's more accessible than just flat color changes.

BatElite:
Uploaded the half-bar designs.

PR #19164, opened 2024-05-26
Remove /datum/lifeprocess/fire

FlameArrow57:
[INTERNAL][PERFORMANCE]

## About the PR 
Removes /datum/lifeprocess/fire

## Why's this needed? 
Performance. It does unnecessary burn duration checks when that can already be handled by the burning status effect


code/modules/status_system/statusEffects.dm
```diff
@@ -719,6 +719,22 @@
...
+							qdel(A)
+			if (isturf(owner.loc))
+				var/turf/location = owner.loc
+				location.hotspot_expose(T0C + 300, 400)
+			for (var/atom/A as anything in owner.contents)
+				A.material_trigger_on_temp(T0C + 900)
+			if (istype(owner, /mob))
+				var/mob/M = owner
+				for (var/atom/A as anything in M.equipped())
+					A.material_trigger_on_temp(T0C + 900)
```
Tarmunora, line 736: statuseffects tick much much faster than lifeprocesses. Need to put this stuff on a cooldown or you're likely to get negative performance benefits from this, not to mention triggering temperature-based stuff way more than intended.

Tarmunora (changes requested):

Tarmunora (approved):

Azrun (approved):

PR #19165, opened 2024-05-26
Change Technical, Research and Medical Assistants from special to roundstart jobs

Glamurio:
[Respawning] [Balance] 

## About the PR 
Moves `/datum/job/special/technical_assistant` to `/datum/job/engineering/engineer/assistant`
Moves `/datum/job/special/research_assistant` to `/datum/job/research/scientist/assistant`
Moves `/datum/job/special/medical_assistant` to `/datum/job/research/medical_doctor/assistant`

They mostly spawn the same, I ensured that they don't get any job-defining things (insulated gloves, gas masks), but instead spawn with pocket guides etc. that way they will have to approach a senior to obtain these things which furthers understanding of the role.

Additional changes:

`rounds_allowed_to_play` to determine max allowed of rounds on record to queue as the job.
`proc/has_rounds_needed` now checks for `rounds_allowed_to_play` as well

`jobs.dm` got new defines for pre-set variables that existed, this seemed cleaner to me:
`ROUNDS_MIN_CAPTAIN`, `ROUNDS_MIN_SECURITY`, `ROUNDS_MIN_DETECTIVE`, `ROUNDS_MIN_SECASS`
`ROUNDS_MAX_RESASS`, `ROUNDS_MAX_MEDASS`, `ROUNDS_MAX_TECHASS`

All newbee requirements have been set to 30 rounds across the board for parity, including `NEWBEE_ROUNDS`, but can still be tweaked if desired.

The new jobs have a `rounds_allowed_to_play` cap of 75.

## Why's this needed? 
Assistant jobs are a really good way for new players to hop into the game, they're a lot less scary with a lot less responsibility, and they typically are a good way to indicate to experienced players to try and be extra helpful when they see an assistant. Unfortunately, these jobs are relegated to special jobs, and as such aren't available at round start, and not even available every round in general, when the equivalent for Security seems to pose no such hindrance.

I think this will both make it easier for newbees to try out new jobs, and make it generally easier for people to become an assistant when they don't know that these jobs are gated behind latejoining.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Technical, Research and Medical Assistants are now properly part of their respective departments.
(*)The specialized assistants can be played until you have played 75 rounds total. Staff assistants are unchanged.
```


Tarmunora (changes requested):
making these into subtypes feels like arming unnecessary future footguns. You don't need inheritance here, especially since you're overriding basically all the fields anyways

Flaborized:
I don't really love making them an 'official' main job when they're so content-dry. Sec assistants were given like this whole identity as pseudo hall monitors with the tickets and stuff, and secass is meant to be playable by anybody who wants to be sec, but with less stuff. Sec can be a bit Much for even experienced players, and secass lets you play in the department but with a different set of expectations, but all of these jobs are functionally just the same as the regular job.

I don't think we should make jobs that are *solely* for new players unless we're roundcapping them or something; it could be neat if these were actual indicators the player was newer, and then we could also probably give them some real equipment because then we wouldn't have to worry about experienced people picking them as extras of existing jobs. 

Otherwise I'd rather people learning just play the actual jobs. If they're full-on real deal jobs that are fully pick-able always by anybody, they should have a more unique role to fill and probably like, jumpsuits at least.

TheFortuneCat:
As a new player I would love if these rolls were able to be picked by default.  Literally right now I just waited to join a round so I could pick one of the assistant roles.  Twice I have tried to get clearance necessary to perform as a medbay assistant as a staff assistance and died as a direct result, just being able to spawn with those clearances is a passing a seemingly huge barrier especially on the roleplay server.

flrsh:
I agree with Flab and Herb. I also think it would be good to actually limit these roles to just newer players. I think we originally did that with security assistants as well, but there was a good point about people in general wanting to play a more chill, less responsibility-laden security role, so we removed the round limit. But I don't think that really applies here, so I'd like to keep these jobs for newer players.

Glamurio:
Okay, I'll make a thread about it to get some information on how we could add some "meat to the bone" with these jobs. I don't mind having them locked to new players either, and we can always change our mind again later like with SecAss.

Glamurio:
!merge_upstream

_std/defines/jobs.dm
```diff
@@ -54,3 +54,14 @@
 #define STAPLE_JOBS (1<<0)
 #define SPECIAL_JOBS (1<<1)
 #define HIDDEN_JOBS (1<<2)
+
+// Job round requirements
+#define ROUNDS_REQ_CAPTAIN 30 // captains should know what they're doing (they won't)
+#define ROUNDS_REQ_SECURITY 30 // higher barrier of entry than before but now with a trainee job to get into the rythym of things to compensate
+#define ROUNDS_REQ_DETECTIVE 15 // half of sec, please stop shooting people with lethals
+#define ROUNDS_REQ_SECASS 5
```
pgmzeta, line 62:
```suggestion
#define ROUNDS_MIN_CAPTAIN 30 // captains should know what they're doing (they won't)
#define ROUNDS_MIN_SECURITY 30 // higher barrier of entry than before but now with a trainee job to get into the rythym of things to compensate
#define ROUNDS_MIN_DETECTIVE 15 // half of sec, please stop shooting people with lethals
#define ROUNDS_MIN_SECASS 5
```

code/datums/jobs.dm
```diff
@@ -71,6 +71,8 @@
 	var/bio_effects = null
 	var/objective = null
 	var/rounds_needed_to_play = 0 //0 by default, set to the amount of rounds they should have in order to play this
+	var/rounds_allowed_to_play = 0 //0 by default (which means infinite), set to the amount of rounds they are allowed to have in order to play this, primarily for assistant jobs
+	var/newbee_only = FALSE // in case jobs should only be playable for new players
```
pgmzeta, line 75: if its not used removed it, it can be re-added

code/datums/jobs.dm
```diff
@@ -1166,32 +1220,16 @@ ABSTRACT_TYPE(/datum/job/civilian)
...
-	access_string = "Research Assistant"
-	low_priority_job = TRUE
-	slot_jump = list(/obj/item/clothing/under/color/white)
-	slot_foot = list(/obj/item/clothing/shoes/white)
-	slot_belt = list(/obj/item/device/pda2/toxins)
-	slot_ears = list(/obj/item/device/radio/headset/research)
-	wiki_link = "https://wiki.ss13.co/Research_Assistant"
+	New()
+		..()
+		src.access = get_access("Chemist")
```
pgmzeta, line 1225: why is this needed?

Glamurio: leftover from an erroneous merge

Glamurio:
Don't worry #19165, one day someone will merge you. It won't be today, it may not even be this month, but one day for sure.

PR #19172, opened 2024-05-27
Remove /datum/lifeprocess/skin

FlameArrow57:
[INTERNAL][PERFORMANCE]

## About the PR 
Removes /datum/lifeprocess/skin. Currently it's only used for patches, so reworks that into a status effect

## Why's this needed? 
Minor performance


code/modules/status_system/statusEffects.dm
```diff
@@ -2773,3 +2773,34 @@
...
+/datum/statusEffect/patches_applied
+	id = "patches_applied"
+	desc = "Patch(es) have been applied"
+	visible = FALSE
+	// amount applied via touch
+	var/static/use_volume = 0.5
+
+	onUpdate(timePassed)
+		if (ON_COOLDOWN(src.owner, "applied_patches_application", LIFE_PROCESS_TICK_SPACING))
+			return
```
Tarmunora, line 2786: I think we'd want to reintroduce `mult` in some fashion to compensate for this cooldown not necessarily lining up with statuseffect tickrate. Should be a pretty minor difference but it would be good to keep it normalized

Tarmunora (approved):

PR #19174, opened 2024-05-27
Adds Speech Pro to loot tables

MintyPhresh:
## About the PR 
Adds the Speech Pro to Amusing Duck egg loot, secure crate loot (like what you'd find mining), and as a possible trinket.

## Why's this needed? 
The Speech Pro is a joy to use regardless of whether you can talk or not. In #18898 the ability to print them at a fabricator was removed since people would expect mimes to use it every round; making it randomized loot should hopefully assuage this concern...


Glamurio:
Unbiased comment, but I personally would still really like to see them fabricated. This is a nice compromise in the meantime.

Glamurio (approved):

flrsh (approved):
This seems okay to me since they'd still be pretty rare.

PR #19184, opened 2024-05-27
Adds "Mails Delivered / Frauded" to credits

Glamurio:
[Feature]

## About the PR 
Adds `mail_opened` and `mail_fraud` to both crew credits and game stats.

Adds a "Mails Delivered / Frauded" at the end of credits which displays them.

## Why's this needed? 
Funny mail stuff

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Mail Delivery Score is now tracked at the end of credits.
```

### DIFF
--- a/code/WorkInProgress/zamu_mail.dm
+++ b/code/WorkInProgress/zamu_mail.dm
@@ -26,7 +26,6 @@
 
 	New()
 		..()
-		game_stats.Increment("mail_sent")
 		if (src.random_icons)
 			src.icon_state = "mail-[rand(1,3)]"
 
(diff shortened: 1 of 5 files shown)


Xkeeper0:
oop. my main comment on this is that the amount of mail sent should not be tracked, and a percentage not shown. the amount opened by the right person and the amount fraudulently opened are fine to track, but they shouldn't contribute to the score -- just fun numbers to show

Glamurio:
> oop. my main comment on this is that the amount of mail sent should not be tracked, and a percentage not shown. the amount opened by the right person and the amount fraudulently opened are fine to track, but they shouldn't contribute to the score -- just fun numbers to show

They're a fun number right now as well, they don't actually contribute to anything. I can hide the amount of mail sent though, I just thought it's more satisfying to see how much mail of all mail available was delivered.

Xkeeper0 (approved):

PR #19190, opened 2024-05-28
Adds Chemical Accidents to those without relevant training

cheekybrdy:
[INPUT WANTED] [FEATURE] [CONTROVERSIAL] [SCIENCE]
## About the PR 
This PR adds chemical accidents to those lacking science, medical or bartender training. This currently adds accidents to the dispenser, allowing those unqualified to accidentally cause part of the beaker's contents to spill onto them, as well as the heater very rarely exploding the glass due to temperature difference. Input is wanted on more devices for these to be added to. Also lets borgs see arthints cause sci module exists.

## Why's this needed? 
This is intended to increase demand for chemicals to the relevant jobs, and to reduce powergaming of this system. While a reductive system, it should be near inconsequential to have beaker contents spilled unless if your making a hell/death/vetoyourexistence mix. The only major concern I see is that problematic rampages are usually stopped by either armoury equipment or deathmixes once sec is dead, and this removes one of those options to alot of people, and armoury is usually supporting the opposition in these scenarios.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Trying to use chemical equipment as a unqualified individual can result in accidents now. It may be best to ask a professional to make what you need now.
```


Tarmunora:
closing, explanation per forum thread post from leah
> Not a fan, tying little bonuses like artifact hints or construction speed to job training is okay but making an entire mechanic worse or dangerous to use based on invisible traits feels very bad. Not to mention job transfers etc.

PR #19193, opened 2024-05-28
Adds milk to the chef dispenser

MintyPhresh:
## About the PR 
Adds plain old milk to the chef dispenser, and reduces the number of space-milks preloaded in a default fridge by 3 to reduce clutter.

## Why's this needed? 
Catering can already print as much milk as they want using the espresso machine, in 16.8u increments. They can also print as much milk as they want for skeleton-healing purposes with the normal chef dispenser, since strawberry and chocolate milk work for that. This change reduces the amount of inputs needed to get milk.


DisturbHerb:
Initial Commit 😱

ZeWaka (approved):
i think milk is normal for a chef to keep in their fridge

MintyPhresh:
I was moreso keeping some in the fridge because that item set is used in places without chef dispensers

frawhst (approved):
Seems fine

PR #19195, opened 2024-05-28
Adds telesci accident error chance to all runs by the untrained

cheekybrdy:
[INPUT WANTED] [FEATURE] [CONTROVERSIAL] [SCIENCE]
## About the PR 
Makes it so all telesci uses regardless of accuracy can cause a accident if used by a untrained individual.

## Why's this needed? 
Maybe tinkering with the volatile equipment when unqualified should cause a error or two, basic lab safety and all that. Adds more to scientist training and local telesci's usages are small or abusable, should just be saved for those doing the job.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Telescience systems are more volatile, using them without training can cause accidents regardless of current coordinates.
```


Tarmunora:
*jsay

PR #19196, opened 2024-05-28
Borg can read Arthints

cheekybrdy:
[SCIENCE]
## About the PR 
Gives borgs the ability to see arthints.

## Why's this needed? 
Borgs should be able to access a database with the relevant information to determine artifact traits. They also have a science module and do artsci anyway.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Borgs can now see artifact hints in addition to scientists.
```


MeggalBozale:
Or just not touch them in general and risk human harm? I don't see why borgs need this.

ZeWaka:
yeah not a fan tbh, mainly because human harm

PR #19207, opened 2024-05-29
516 url scheme adherence 

Sovexe:
## About the PR 
* Introduces the byond:// scheme into links as per Lummox's note on 516 link requirements

## Why's this needed? 
https://www.byond.com/forum/post/2927285#comment26728373
> The workaround letting users get away with starting URLs with ? as a shortcut for byond://? in the browser will be going away in BYOND 516, except for legacy projects. Update your HTML accordingly.


Sovexe:
May no longer be required

PR #19224, opened 2024-05-30
HTML_Encode fix for character flavor texts

soleilan:
## About the PR 

Eons ago [Leah](https://discord.com/channels/182249960895545344/890226559691157524/1094440256113229834) said I can PR this but my PR featurecreeped and I burned out so now I'm back with this 👍 

Adds a new define for MED_FLAVOR_CHAR_LIMIT of 512 and changes Security Note and Medical Note for character customization to use this new limit, going from a limit of 256 to 512. Tested to see how it looks in-game and seems fine.

I also moved the html_encode line down after the character limit check for all the flavor texts for a more sane writing experience. ~~why the hell does a & count as four characters i might as well just write AND!!!~~

Terminal view

image

Print view

image

## Why's this needed? 

256 is barely enough for three sentences, seems downright cruel for RP


mordent-goonstation:
Personal take is that between the various notes you're already making me read a ton of probably useless (to the current round) info. Character backstory is better conveyed via IC interactions, not a loredump in notes.

soleilan:
Nobody is making you read the Security or Medical notes, they're hidden away so usually only bored people at consoles with nothing else to do will read them. That's also why I simply increased their limit from 256 to 512, I don't think we need to lore dump, but 256 is so restrictive you can barely even write a few fun sentences. 

Also the Syndicate notes have a whopping character limit of 4000!!! which literally lets you lore dump already directly into any antagonists PDA. I'm sure we can handle two or three more sentences for the console notes which are fun browsing material for when its a slow shift.

MintyPhresh:
"I would have written a shorter letter, but did not have the time." Records being concise is good, even if they're not required reading by medical/sec staff. I can see your point 512 not being a large increase, but having such a small area forces players to highlight only the absolutely essential elements of their character's medical notes, rather than getting bogged down in unnecessary wording or cursory details that distract the reader from the main message of the notes.

TobleroneSwordfish:
No idea what past me said on the subject but I agree with the above, it'd make more sense to lower the syndicate intel max length than raise these (although that would end up with a bunch of cut off syndicate intel notes and everyone would be mad so 🤷)

soleilan:
I'm fine with lowering the ludicrous Syndie notes limit, but are we sure we can't at least expand the Sec/Med notes even a little bit? 🥺 384? 320? I am in the bargaining phase. If not I'll revert the char limit and at least pass the html_encode move through.

soleilan:
For reference the above message would get shortened to this under our current character limit + html_encode issue.
image

Even with the html_encode fix, the above message is still too long!
image

I understand that we don't want long text and brevity and all that, but 256 seems needlessly restrictive for anything other than the examine flavor text.

mordent-goonstation:
Feel free to handle the encoding issue, but the length of thing you posted there seems A-okay to me. These are notes, not backstories.

> Strongly prefers synthorgans over robotic replacements. Has a fear of needles.

80 characters, has two strong RP directives that a doctor could take into account.

> If you're reading this, go print a robotic liver. They're on their third donor one after repeated alcohol abuse incidents, so just give them a robo-one when they next come in absolutely shitfaced. It'll save us time and money in the long run.

243 characters, verbose yet still "complete", including backstory and flavored as a comment from a previous doctor.

soleilan:
Done!

code/datums/preferences/preferences.dm
```diff
@@ -558,43 +558,43 @@ var/list/removed_jobs = list(
...
 					src.profile_modified = TRUE
 					return TRUE
 
 			if ("update-securityNote")
 				var/new_text = tgui_input_text(usr, "Please enter new flavor text (appears when examining your security record):", "Character Generation", html_decode(src.security_note), multiline = TRUE, allowEmpty=TRUE)
 				if (!isnull(new_text))
-					new_text = html_encode(new_text)
 					if (length(new_text) > FLAVOR_CHAR_LIMIT)
-						tgui_alert(usr, "Your flavor text is too long. It must be no more than [FLAVOR_CHAR_LIMIT] characters long. The current text will be trimmed down to meet the limit.", "Flavor text too long")
+						tgui_alert(usr, "Your flavor text is too long. It must be no more than [MED_FLAVOR_CHAR_LIMIT] characters long. The current text will be trimmed down to meet the limit.", "Flavor text too long")
```
mordent-goonstation, line 573: You removed the `MED_FLAVOR_CHAR_LIMIT` define from an earlier commit, probably want to revert this line.

code/datums/preferences/preferences.dm
```diff
@@ -558,43 +558,43 @@ var/list/removed_jobs = list(
...
 					src.profile_modified = TRUE
 					return TRUE
 
 			if ("update-medicalNote")
 				var/new_text = tgui_input_text(usr, "Please enter new flavor text (appears when examining your medical record):", "Character Generation", html_decode(src.medical_note), multiline = TRUE, allowEmpty=TRUE)
 				if (!isnull(new_text))
-					new_text = html_encode(new_text)
 					if (length(new_text) > FLAVOR_CHAR_LIMIT)
-						tgui_alert(usr, "Your flavor text is too long. It must be no more than [FLAVOR_CHAR_LIMIT] characters long. The current text will be trimmed down to meet the limit.", "Flavor text too long")
+						tgui_alert(usr, "Your flavor text is too long. It must be no more than [MED_FLAVOR_CHAR_LIMIT] characters long. The current text will be trimmed down to meet the limit.", "Flavor text too long")
```
mordent-goonstation, line 584: Revert

mordent-goonstation (changes requested):

code/datums/preferences/preferences.dm
soleilan: Yeah oops, I should've just reverted that commit instead of this mess, fixed at least.

mordent-goonstation:
Removed changelog

PR #19231, opened 2024-05-30
Fix revheads not having a revhead arrest icon when dead

FlameArrow57:
[GAME MODES][BUG]

## About the PR 
Fixes dead headrevs showing as loyal when implanted with a counter rev implant, rather than headrev

## Why's this needed? 
Bug fix


444Portal:
>  rather than loyal
oopsie

FlameArrow57:
> > rather than loyal
> 
> oopsie

ty

PR #19234, opened 2024-05-31
Remove /datum/lifeprocess/viruses

FlameArrow57:
[INTERNAL][PERFORMANCE]

## About the PR 
Removes /datum/lifeprocess/viruses and reworks it to be an event based status effect

## Why's this needed? 
Minor performance


FlameArrow57:
Requires #19172 to be merged first before working properly

FlameArrow57:
above pr is closed now

code/modules/status_system/statusEffects.dm
```diff
@@ -2832,3 +2832,31 @@
...
+		var/mult = max(LIFE_PROCESS_TICK_SPACING, src.passed) / LIFE_PROCESS_TICK_SPACING
+		src.passed = 0
+
+		var/mob/living/L = src.owner
+
+		if (!isdead(L))
+			for (var/datum/ailment_data/ailment as anything in L.ailments)
+				ailment.stage_act(mult)
+
+		for (var/mob/living/carbon/C in oviewers(4, L))
```
ZeWaka, line 2855: you probably want hearers here?

code/modules/status_system/statusEffects.dm
```diff
@@ -2852,9 +2852,9 @@
 			for (var/datum/ailment_data/ailment as anything in L.ailments)
 				ailment.stage_act(mult)
 
-		for (var/mob/living/carbon/C in oviewers(4, L))
-			if (prob(40))
-				L.viral_transmission(C, "Airborne", 0)
+		for (var/mob/living/other_mob in hearers(4, L))
```
ZeWaka, line 2855: makes sense also since they could be on the same time lol

ZeWaka (approved):

PR #19240, opened 2024-05-31
Latejoining folk with Background traits no longer get announced

MintyPhresh:
## About the PR 
The code from #15240 was changed to use a concise and readable proc! Unfortunately that proc tried to compare a list to a string directly and was nonfunctional. This PR fixes that proc (and should also fix the issues it was likely having during pod wars).

## Why's this needed? 
Stowaways and heavy sleepers were getting announced again.

### DIFF
--- a/code/datums/traits.dm
+++ b/code/datums/traits.dm
@@ -207,7 +207,7 @@
 			var/datum/trait/T = traits[id]
 			for (var/heldcat in T.category)
 				if (heldcat == cat)
-					. = T
+					return T
 
 //Yes these are objs because grid control. Shut up. I don't like it either.
 /datum/trait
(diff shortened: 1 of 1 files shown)


code/datums/traits.dm
```diff
@@ -205,8 +205,9 @@
 	proc/getTraitWithCategory(var/cat)
 		for(var/id in traits)
 			var/datum/trait/T = traits[id]
-			if (T.category == cat)
-				. = T
+			for (var/heldcat in T.category)
+				if (heldcat == cat)
+					. = T
```
TobleroneSwordfish, line 210: Any reason why this isn't just `return T`? It'd be a little bit faster and just generally cleaner code.

TobleroneSwordfish (changes requested):

code/datums/traits.dm
MintyPhresh: Nope-- This is just how it was before and I generally try to keep code the same as it was before since my understanding is pretty poor... I'll do a test run with it as "return" to see if it works and update it if it does

PR #19241, opened 2024-05-31
German Accent Fixes

klushy225:
[TRAITS][BUG]

## About the PR  The two tweaks are as follows:
1) the letter D no longer has a 50% chance to turn into a T.

2) the letter W at the end of a word is left unchanged, so how isn't displayed as "hov"

3) "miss" is no longer replaced by frau, because "I miss you" would be turned to "I frau you," which means "I woman you." Makes no sense.

This, as an added side affect, makes the lawbringer fully functional with the accent.

## Why's this needed? 
A couple words said with the accent previously would raise some eyebrows, and make no sense at all. These small tweaks both make the accent more appealing, and less unintelligable.

## Changelog 

```changelog
(u)Klushy225
(+)Tweaked the German accent to make more sense.
```

### DIFF
--- a/code/procs/accents.dm
+++ b/code/procs/accents.dm
@@ -833,8 +833,8 @@ proc/random_accent()
 	S = lowertext(S)
 	switch(S)
 
-		if("w") //germans pronounce W like V
-			if(lowertext(R.next_char) != " " || lowertext(R.next_char) != null)
+		if("w") //germans pronounce W like V, but generally keep the "ow" sound at the end of words and such.
+			if (!is_null_or_space(lowertext(R.next_char)))
 				new_string = "v"
 				used = 1
 		if("t") //unlinke the stereotypical zat- it's more like dat, because both the t and the h are pronounced in german. Of course, the Z still does happen quite often.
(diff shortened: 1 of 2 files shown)


code/procs/accents.dm
```diff
@@ -834,12 +834,17 @@ proc/random_accent()
 	switch(S)
 
 		if("w") //germans pronounce W like V
-			new_string = "v"
-			used = 1
-		if("t") //unlinke the stereotypical zat- it's more like dat, because both the t and the h are pronounced in german
+			if(lowertext(R.next_char) == " " || lowertext(R.next_char) == null)
```
TobleroneSwordfish, line 837: Shouldn't this be the other way around? Currently it looks like it's *only* changing w's to v's on the end of words

TobleroneSwordfish (changes requested):

code/procs/accents.dm
klushy225: oop- probably. I'm very sleep-deprived. FIxing that now :P

TobleroneSwordfish (approved):

klushy225:
oh dear I broke it. I really have to be more careful. Trying to find a fix now, but at the moment es ist _kaputt_

Tarmunora (approved):

PR #19244, opened 2024-05-31
Fix and absolute path manufacturing_requirements

### LINKED ISSUE #19222: Steel and Copper are treated as high density matter and high conductive material respectively
### Labels

[BUG][MAJOR]

### Description

As per the title, the fabricator is treating these resources as something they definitely aren't.

### Steps to reproduce

1. Find a fabricator, for my attached examples I used a robotics one.
2. Make sure the fabricator only has base resources (i.e. Steel, Copper, Glass).
3. Find a recipe requiring either High Density Matter, High Conductive Material, or both.
4. See that they can be fabricated.

### Expected Behavior

These base materials shouldn't have such high value to the fabricator.

Here is a list of materials in the fabricator, and a picture with several underlined objects which it shouldn't be able to make.

materials
Screenshot (29)

MeggalBozale:
## About the PR 
Hopefully absolute pathing and these tweaks brings it up to parity with the old checks once and for all.

## Why's this needed? 
Fixes #19222
...again

### DIFF
--- a/code/datums/manufacturing_requirements.dm
+++ b/code/datums/manufacturing_requirements.dm
@@ -77,6 +77,7 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement/match_property)
 
 /datum/manufacturing_requirement/match_property
 	is_match(var/datum/material/M)
+		var/should_match_property = !isnull(src.material_property) || !isnull(src.material_threshold)
 		if (should_match_property && !src.matches_property(M))
 			return FALSE
 		if (!isnull(src.material_flags) && !src.matches_flags(M.getMaterialFlags()))
--- a/code/datums/manufacturing_requirements.dm
+++ b/code/datums/manufacturing_requirements.dm
@@ -175,6 +175,7 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement/match_property)
 	name = "Metal"
 	id = "metal"
 	material_flags = MATERIAL_METAL
+	material_threshold = 0 // So we try to match properties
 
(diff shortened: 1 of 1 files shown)


code/datums/manufacturing_requirements.dm
```diff
@@ -75,139 +75,143 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement/match_property)
...
+
+	/// Returns whether the material flags are matched. This will return true should any flag match.
+	proc/matches_flags(var/material_flags)
+		return material_flags & src.material_flags
+
+	/// Returns whether the material property matches the given criterion. Default behavior is to check if >=, override w/o calling parent for diff behavior.
+	proc/matches_property(var/datum/material/M)
+		var/property_value = M.getProperty(src.material_property)
+		var/is_true = property_value >= src.material_threshold
+		boutput(world, "Property check on [M.getName()] for [src.name] evaluates to ([property_value] >= [src.material_threshold] = [is_true])")
```
mordent-goonstation, line 94: Remove debug code

MeggalBozale: still working on fixingbug D:

Tarmunora (approved):
tbh the entire inheritance structure here should be given a once-over (e.g.: why are material_id and material_flags both on the parent, they should be split out into the subtypes where they're needed etc.) but that's an issue for another day (translation: I will do it myself when I have time)

MeggalBozale:
Yeah i wanna get to that tmr, this fixes it for now though

PR #19261, opened 2024-06-01
Allow syndicate dagger to stick into mobs on throw

FlameArrow57:
[GAME OBJECTS][BALANCE][QoL]

## About the PR 
This PR allows you to set the syndicate dagger to stick into mobs on hit, by using the in-hand hotkey. Daggers stuck into mobs can be pulled out with help intent, or alternatively fall out of the mob after 2 seconds. 

Pulling the dagger out shares the same click cooldown as attacks.

Also two thoughts:
-Thinking click cooldown for dagger pullout might be a bit confusing, and may be best to remove if not too strong
-Intent can be changed depending on the user, not clear to others how it's being used. May be best to just make it always stick into the target?

## Why's this needed? 
The sprite is really small and hard to click sometimes, depending on x and y pixel offsets. Other times it will almost be completely underneath the hit mob and you need to use right click to pick it up.

## Changelog 

```changelog
(u)FlameArrow57
(+)Syndicate daggers will now stick into people on throw. Stuck daggers can be removed with help intent or otherwise will fall out after 2 seconds.
```


Glamurio:
Think it'd be cooler if they kept being stuck and just slowly caused the person to bleed out until removed. Might as well.

FlameArrow57:
> Think it'd be cooler if they kept being stuck and just slowly caused the person to bleed out until removed. Might as well.

Yea idk, syndicate daggers are already pretty hard to come by, I think it would be bad if the person ran off, and it was unrecoverable

Glamurio:
> > Think it'd be cooler if they kept being stuck and just slowly caused the person to bleed out until removed. Might as well.
> 
> Yea idk, syndicate daggers are already pretty hard to come by, I think it would be bad if the person ran off, and it was unrecoverable

I mean, yeah, but if you throw your "so hard to come by" knife at someone, you kind of should expect not to get it back? You could make it so that it needs outside help to be removed or it takes a really long time. Maybe even slow the person so you can chase after.

code/obj/item/misc_weapons.dm
```diff
@@ -560,6 +575,14 @@ TYPEINFO(/obj/item/sword/pink/angel)
 		M.changeStatus("knockdown", 6 SECONDS)
 		M.force_laydown_standup()
 		take_bleeding_damage(M, null, 5, DAMAGE_CUT)
+		if (src.set_to_stick && istype(M, /mob/living))
+			var/mob/living/L = M
+			L.daggers += src
```
TobleroneSwordfish, line 580: This should probably come with some kind of visual indication that the dagger is stuck into someone. I know it's sprite work buuut it would be a lot clearer.

code/obj/item/misc_weapons.dm
```diff
@@ -594,6 +617,7 @@ TYPEINFO(/obj/item/sword/pink/angel)
 	throwforce = 20
 	stamina_cost = 5
 	c_flags = EQUIPPED_WHILE_HELD
+	can_stick_into_mob = FALSE
```
TobleroneSwordfish, line 620: Should probably apply to at least the syndie combat knife too since that otherwise works exactly the same as the dagger.

TobleroneSwordfish (changes requested):
I quite like this, syndie daggers are pretty awkward to use since you have to scramble to pick them up during the stun.

FlameArrow57:
> > > Think it'd be cooler if they kept being stuck and just slowly caused the person to bleed out until removed. Might as well.
> > 
> > 
> > Yea idk, syndicate daggers are already pretty hard to come by, I think it would be bad if the person ran off, and it was unrecoverable
> 
> I mean, yeah, but if you throw your "so hard to come by" knife at someone, you kind of should expect not to get it back? You could make it so that it needs outside help to be removed or it takes a really long time. Maybe even slow the person so you can chase after.

Sorry, rephrasing - I think they're meant to be a melee tool that you keep on using, rather than disposable

code/obj/item/misc_weapons.dm
FlameArrow57: Yeah was thinking, someone's going to be lying down, and depending on the way they are facing, the sprite may not even be seen at all, which is why I didn't sprite it

code/mob/living.dm
```diff
@@ -84,6 +84,8 @@
 
 	var/list/implant = list()
 	var/list/implant_images = list()
+	/// daggers stuck into this mob
+	var/list/obj/item/dagger/daggers
```
Tarmunora, line 88: 🥹

pgmzeta:
maybe convert to implant that stores the dagger inside of it, then pops itself out?

code/obj/item/misc_weapons.dm
FlameArrow57: If you think so, I was mostly just aiming at the syndicate dagger because of how it small it is, syndicate dagger is more clickable
image

FlameArrow57:
> maybe convert to implant that stores the dagger inside of it, then pops itself out?

Could yeah, I guess I don't see much of a benefit with that though

code/obj/item/misc_weapons.dm
TobleroneSwordfish: I think it's still better than no indication, and you could probably fudge the sprites so it's visible in 3/4 directions anyway.

code/obj/item/misc_weapons.dm
```diff
@@ -594,6 +607,7 @@ TYPEINFO(/obj/item/sword/pink/angel)
 	throwforce = 20
 	stamina_cost = 5
 	c_flags = EQUIPPED_WHILE_HELD
+	sticks_into_mob = FALSE
```
TobleroneSwordfish, line 610:
```suggestion
```

code/mob/melee_attack_procs.dm
```diff
@@ -16,6 +16,14 @@
 	if (!istype(M))
 		return
 	src.lastattacked = M
+	// dagger code here so that you can still put it out when target is bleeding
```
TobleroneSwordfish, line 19: Consider using `COMSIG_ATTACKHAND` instead of putting this here

code/obj/item/implant.dm
```diff
@@ -1136,6 +1136,29 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+				src.implant_overlay = image(icon = 'icons/mob/human.dmi', icon_state = "dagger_stick_[pick(0, 1)]", layer = MOB_EFFECT_LAYER)
+				dagger_to_store.set_loc(src)
+				src.stored_dagger = dagger_to_store
+
+			on_pull_out(mob/puller)
+				if (ismob(src.loc))
+					puller.put_in_hand_or_drop(src.stored_dagger)
+					src.on_remove(src.loc)
+				src.stored_dagger = null
+				qdel(src)
```
TobleroneSwordfish, line 1154: This isn't removing the dagger from the implants list so it doesn't actually work.

TobleroneSwordfish (changes requested):

PR #19265, opened 2024-06-01
Moves freezer pipes around a bit on cog1.

Laboredih123:
[BUG]

## About the PR 
Reduces the amount of pipes and fixes a case of overlayed pipes in the freezer.

## Why's this needed? 
Bug bad, pipe spam bad.

## Changelog 

```changelog
(u)Laboredih123
(+)Less pipes to deal with in the cog1 freezer, wow.
```


Laboredih123:
why mapdiffbot fail :(

Laboredih123:
heres how it looks like now (hid some stuff to make it clearer) (wow so cool quad pipe)
image

TheTyrant25:
the changelog entry should probably mention which map's freezers are being changed.

Laboredih123:
ok done

PR #19269, opened 2024-06-02
Fix deleting prints with blueprints of that instance in progress bluescreening manufacturers

MeggalBozale:
## About the PR 
Add null check to stop displaying invalid prints but still display the window, rather than imploding upon ourselves

## Why's this needed? 
Fixes bug reported in discord


MeggalBozale:
!merge_upstream

MeggalBozale:
ok ig that dont work. WELP

MeggalBozale:
There we go. should be rdy

PR #19270, opened 2024-06-02
DRAFT: Re-refactor manufacturing requirements to be Better

MeggalBozale:
[Internal] [Code-Quality]
## About the PR 
Adds more subtype matches of manufacturing requirements to better split the functionality between children.
Protects the ID and name attributes, and provides getters (but not setters) for these values -- which should remain constant throughout
renames various requirement IDs to be more distinct from another given the splitting of certain requirement datums into more datums -- renames all the IDs appropriately (pain... but worth it)
adds some more checks for CHECK_MORE_RUNTIMES, as well as better logic checks for the is_match() procs.

## Why's this needed? 
While these manufacturing datums have certainly been an improvement over the old system, they still had a long way to go -- this PR aims to bring it closer to its ideal state.

### DIFF
--- a/code/datums/manufacturing_requirements.dm
+++ b/code/datums/manufacturing_requirements.dm
@@ -36,7 +36,7 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement)
 	/// Checks whether or not the given material meets the requirements enforced by this proc.
 	proc/is_match(var/datum/material/M)
 		SHOULD_CALL_PARENT(TRUE)
-		return isnull(M)
+		return !isnull(M)
 
 /datum/manufacturing_requirement/not_null
 	name = "Any"
--- a/code/datums/manufacturing_requirements.dm
+++ b/code/datums/manufacturing_requirements.dm
@@ -8,6 +8,13 @@
 var/global/list/requirement_cache
 
 /proc/getRequirement(var/R_id)
+	#ifdef CHECK_MORE_RUNTIMES
+	if (!istext(R_id))
+		CRASH("getRequirement() called with a non-text argument [R_id].")
+	if (!(R_id in requirement_cache))
+		CRASH("getRequirement() called with an invalid requirement id [R_id].")
+	#endif
+	// Sanity checks that all requirement IDs resolved to instances in the cache
 	return requirement_cache?[R_id]
 
 ABSTRACT_TYPE(/datum/manufacturing_requirement)
@@ -238,7 +245,7 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement/match_property)
 		if (!(istype(M, src.match_typepath))) return
 
 /datum/manufacturing_requirement/match_subtypes/gemstone
-	name = "Gemstone"
+	name = "gemstone_subtypes"
 	id = "gemstone_subtypes"
 	match_typepath = /datum/material/crystal/gemstone
 
@@ -267,7 +274,7 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement/match_property)
 		/datum/manufacturing_requirement/match_flags/metal,
 	)
 /datum/manufacturing_requirement/mixed/metal
-	name = "Tough Metal"
+	name = "Sturdy Metal"
 	id = "metal_tough"
 	requirements = list(
 		/datum/manufacturing_requirement/match_flags/metal,
@@ -285,6 +292,14 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement/match_property)
 /datum/manufacturing_requirement/mixed/insulated
 	name = "Insulative"
 	id = "insulated"
+	requirements = list(
+		/datum/manufacturing_requirement/match_flags/insulated,
+		/datum/manufacturing_requirement/match_property/insulated,
+	)
+
+/datum/manufacturing_requirement/mixed/insulated
+	name = "Highly Insulative"
+	id = "insulated_high"
 	requirements = list(
 		/datum/manufacturing_requirement/match_flags/insulated,
 		/datum/manufacturing_requirement/match_property/insulated,
--- a/code/datums/manufacturing_requirements.dm
+++ b/code/datums/manufacturing_requirements.dm
@@ -21,10 +21,6 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement/match_property)
 	var/material_id = null
 	/// Material flags of the material to check. None of null, can be made like MATERIAL_A | MATERIAL_B if needed to check for either.
 	var/material_flags = null
-	/// Property of the material to check. None if null, some string like "radioactive" if used
-	var/material_property = null
-	/// Context-dependent material threshold for an item. Use if you want to check a material property of something. Currently just checks if >=
-	var/material_threshold = null
 
 	// ID must be defined, or else we have a problem
 	#ifdef CHECK_MORE_RUNTIMES
@@ -46,7 +42,7 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement/match_property)
 			return FALSE
 		return TRUE
 
-/datum/manufacturing_requirement/any
+/datum/manufacturing_requirement/not_null
 	name = "Any"
 	id = "any"
 
@@ -61,7 +57,8 @@ ABSTRACT_TYPE(/datum/manufacturing_requirement/match_property)
 		. = ..()
 
 	is_match(var/datum/material/M)
-		if (!..())
+		. = ..()
+		if (!.)
 			return FALSE
 		if (src.material_id != M.getID())
 			return FALSE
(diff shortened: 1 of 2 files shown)


MeggalBozale:
note: changes to the 52 other files for ID name changes was done last, im not sure how you do it on the reviewing end but i hope it helps to be able to review the changes outside of there.

primarily code changes are isolated to `manufacturing_requirements.dm` the rest of the bloat is just the relevant ID name changes, which were all made more verbose as to avoid ID name conflicts

MeggalBozale:
Some NYIs in there because I dont recall the ideal way to iterate bitflags for those, if at all. In any case, they aren't used, either. Just a future consideration and removable without issues

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,338 @@
...
+	/// Resolve the requirement paths to instances in the cache.
+	New()
+		. = ..()
+		var/list/datum/manufacturing_requirement/requirement_instances = list()
+		var/datum/manufacturing_requirement/requirement_id
+
+		for (var/datum/manufacturing_requirement/requirement_path as anything in src.requirements)
+			requirement_id = initial(requirement_path.id)
+			requirement_instances += getRequirement(requirement_id)
+		src.requirements = requirement_instances
```
Tarmunora, line 270: Why are we jumping through additional hoops by listing the requirements here as typepaths instead of by ID like everywhere else does?

Tarmunora (changes requested):
Not a fan of coupling the threshold values/etc. to the ID field - means that making adjustments to the internals of the requirement forces edits to everything that uses the req, which is not ideal, and also makes actually using them extra annoying (who wants to type out `electrical_property_>=_6`?)

code/datums/manufacturing_requirements.dm
MeggalBozale: fair, i can bring it back to that

MeggalBozale:
Understandable, but I'm not sure how else to convey the requirement type, threshold, *and* material amount without introducing an associative list on the other end.

MeggalBozale:
ah dammit

MeggalBozale:
I definitely goofed up a basic revert but I did do it

Tarmunora (approved):

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,336 @@
...
-/datum/manufacturing_requirement/match_property
 	is_match(var/datum/material/M)
-		var/should_match_property = !isnull(src.material_property) || !isnull(src.material_threshold)
-		if (should_match_property && !src.matches_property(M))
-			return FALSE
-		if (!isnull(src.material_flags) && !src.matches_flags(M.getMaterialFlags()))
-			return FALSE
 		. = ..()
+		if (!.) return
+		if (!src.match_property(M)) return FALSE
```
mordent-goonstation, line 78:
```suggestion
		if (!.) return
		return src.match_property(M)
```

I *think* I read this/understood the types here right.

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,336 @@
...
 /datum/manufacturing_requirement/match_property/dense
 	name = "High Density Matter"
 	id = "dense"
-	material_property = "density"
-	material_threshold = 4
+	property_id = "density"
+	property_threshold = 4
 
-/datum/manufacturing_requirement/match_property/dense/super
+/datum/manufacturing_requirement/match_property/superdense
```
mordent-goonstation, line 101: Assuming flattening here for some reason? You have the nested types below still, so not like it's a consistent restructure. One or the other, right?

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,336 @@
...
+	name = "Highly Radioactive"
+	id = "energy_property_high"
+	property_threshold = 5
 
 /datum/manufacturing_requirement/match_property/insulated
-	name = "Insulative"
-	id = "insulated"
-	material_flags = MATERIAL_CLOTH | MATERIAL_RUBBER
-	material_threshold = 4
+	name = "Insulated Material"
```
mordent-goonstation, line 125:
```suggestion
	name = "Insulative Material"
```
The material itself is not insulated, *it* is the insulator and thus the thing you make out of it would be insulated.

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,336 @@
...
-		return TRUE
+	match_property(datum/material/M)
+		return M.getProperty(src.property_id) <= src.property_threshold
 
 /datum/manufacturing_requirement/match_property/insulated/super
-	id = "insulative_high"
 	name = "Highly Insulative"
-	material_threshold = 2
+	id = "insulated_property_high"
+	property_id = "density"
```
mordent-goonstation, line 136: Moderately confident this should be `electrical` instead of `density`? Advantage of the nested structure would be only having to set on the common root insulative material datum.

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,336 @@
...
+	New()
+		. = ..()
+		if (isnull(src.match_typepath) || !ispath(src.match_typepath))
+			CRASH("[src] has invalid match_typepath [src.match_typepath], it must be non-null and a valid path (not an instance!)")
+	#endif
+
+	is_match(var/datum/material/M)
+		. = ..()
+		if (!.) return
+		if (!(istype(M, src.match_typepath))) return FALSE
```
mordent-goonstation, line 249:
```suggestion
		return istype(M, src.match_typepath)
```

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,336 @@
...
-*/
+/**
+ * Manufacturing Requirements are datums which check if a material satisfies some given requirements, to determine if a manufacturer can produce
+ * a blueprint. Manufacturing datums define them by their string ID, which gets converted to the single instance of it in the cache on New().
+ * Not shown in this file are the exact material ID requirement datums, which are generated for the cache on init using the material cache.
+ */
 
 var/global/list/requirement_cache
 
 /proc/getRequirement(var/R_id)
```
mordent-goonstation, line 9: Entirely too generic of a name for a global proc.

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,336 @@
...
+		var/datum/material/M = getMaterial(mat_id)
 		src.name = capitalize(M.getName())
 		. = ..()
 
 	is_match(var/datum/material/M)
-		if (!isnull(src.material_id) && !src.matches_id(M.getID()))
-			return FALSE
 		. = ..()
+		if (!.) return
+		if (src.material_id != M.getID()) return FALSE
```
mordent-goonstation, line 59:
```suggestion
		return src.material_id == M.getID()
```

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,336 @@
...
+	VAR_PROTECTED/list/datum/manufacturing_requirement/requirements = list() //! A list of requirements which must all be satisfied for this to return TRUE
+
+	/// Resolve the requirement paths to instances in the cache.
+	New()
+		. = ..()
+		var/list/datum/manufacturing_requirement/requirement_instances = list()
+
+		for (var/requirement_id in src.requirements)
+			requirement_instances += getRequirement(requirement_id)
+		src.requirements = requirement_instances
```
mordent-goonstation, line 268: Don't much like this. Too tired right now to suggest a better alternative, but to make sure I understand: this is turning a list of string IDs into a list of concrete instances and putting them back on the same variable?

code/datums/manufacturing_requirements.dm
MeggalBozale: Fair, changed

code/datums/manufacturing_requirements.dm
MeggalBozale: aaaaaaaaaaaaaa

code/datums/manufacturing_requirements.dm
MeggalBozale: alr, changed to `getManufacturingRequirement`

code/datums/manufacturing_requirements.dm
MeggalBozale: Yes -- requirements is initially a list of all the manufacturing requirement datum IDs to use, which get converted into instances defined in the cache, to finish setting up the variable.

This avoids having to call getManufacturingRequirement() constantly during runtime, and either way for a concrete instance it has to access some typepath/id/identifier somehow and use that to get a concrete instance.

mordent-goonstation: Sure, my issue is mainly in the "list of type strings" getting converted to "list of concrete instances". I would prefer a different var got used, rather than replacing it. This just seems like a maintenance footgun as-is. Feel free to set the list of strings to null if you think the clean-up is important.

code/datums/manufacturing_requirements.dm
```diff
@@ -1,216 +1,334 @@
...
+	VAR_PROTECTED/list/datum/manufacturing_requirement/requirements = list() //! A list of requirements which must all be satisfied for this to return TRUE
+
+	/// Resolve the requirement paths to instances in the cache.
+	New()
+		. = ..()
+		var/list/datum/manufacturing_requirement/requirement_instances = list()
+
+		for (var/requirement_id in src.requirement_ids)
+			requirement_instances += getManufacturingRequirement(requirement_id)
+		src.requirements = requirement_instances
```
mordent-goonstation, line 266: Why create a new list then assign it and not just operate on the target list? Note my change assumes that `src.requirements` is not used by `..()`.

Also edited to avoid one of the `list()` on the type definition (rather than within `New`).
```suggestion
	VAR_PROTECTED/list/datum/manufacturing_requirement/requirements = null //! A list of requirements which must all be satisfied for this to return TRUE

	/// Resolve the requirement paths to instances in the cache.
	New()
		. = ..()
		src.requirements = new
		for (var/requirement_id as anything in src.requirement_ids)
			src.requirements += getManufacturingRequirement(requirement_id)
```

PR #19274, opened 2024-06-02
Remove bullet "hit by" messages

TobleroneSwordfish:
[PERFORMANCE] [QOL] [REMOVAL]
## About the PR 
Changes all the "Person has been hit by the bullet" messages to only show to the person who was hit, instead of everyone in visual range.

## Why's this needed? 
I've noticed client performance taking a real hit during high population firefights and I'm guessing some of it has to do with the sheer amount of chat spam. Grenades especially generate an absurd amount of rapid chat messages so enough of them going off in a crowd is bound to cause problems.
These kinds of messages are from an earlier era of the game when there were a lot less useful visual effects to convey this type of information.

## Changelog 

```changelog
(u)LeahTheTech
(+)Bullet hit by messages now only show to the person being hit.
```

### DIFF
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -1563,7 +1563,7 @@
 		if (D_TOXIC)
 			src.take_toxin_damage(damage)
 	if (!P || !P.proj_data || !P.proj_data.silentshot)
-		src.visible_message(SPAN_ALERT("[src] is hit by the [P]!"))
+		boutput(src, SPAN_ALERT("You are hit by the [P]!"))
 
 	actions.interrupt(src, INTERRUPT_ATTACKED)
 	return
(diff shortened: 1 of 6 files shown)


MintyPhresh:
removing the "x fires the gun at the floor!" messages would be nice too, honestly

Tarmunora:
Eh, I'd rather just cull messages from spammy projectiles like frag grenades fragments/etc. as needed. I'm not a fan of cutting important info from the chatbox outright

TobleroneSwordfish:
The thing is that all projectiles are spammy once you have 10 nukies on screen with automatic weapons. Mordent made a good point that although the messages are occasionally useful in piecing together exactly what happened after the fight, this might not actually be a good thing as chaotic situations probably shouldn't be perfectly reproduceable by scrolling up the chatbox.

Ultimately I think this kind of chat spam is what makes people ignore the actually important info in the chatbox.

Gannets:
I've considered doing this a few times. It's only a really major issue for nuke ops, but it IS an issue on those rounds.

PR #19284, opened 2024-06-02
Alcohol Resistance also grants resistance to ethanol addiction

MintyPhresh:
## About the PR 
The alcohol resistance effect, which can be gotten either through genetics, the career alcoholic trait, or by being a certain job, already grants full immunity to ethanol's effects (so long as there's no moonshine in there). This PR grants them a greater resistance to ethanol addiction, too; quartering the chance it happens and making it take four times as long to start rolling.

## Why's this needed? 
I've taken the gene that grants alcohol resistance before, expecting it to have this behavior, since it's the hardest symptom of ethanol to treat and the most persistent. What's the point of alcohol resistance if you can't resist taking more of it?

## Changelog
```changelog
(u)mintyphresh
(+)Alcohol resistance from genetics, jobs, and traits now helps stave off addiction, too.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Reagents.dm
+++ b/code/modules/chemistry/Chemistry-Reagents.dm
@@ -322,7 +322,7 @@ datum
 					addProb = round(addProb * 2)
 					rate *= 2
 					//DEBUG_MESSAGE("addictive_personality: addProb [addProb], rate [rate]")
-				if (src.id == "ethanol" || isalcoholresistant(H))
+				if (src.id == "ethanol" && isalcoholresistant(H))
 					addProb /= 4
 					rate /= 4
 			if (!holder.addiction_tally)
--- a/code/modules/chemistry/Chemistry-Reagents.dm
+++ b/code/modules/chemistry/Chemistry-Reagents.dm
@@ -253,7 +253,8 @@ datum
 						H.sims.affectMotive("Energy", energy_value)
 			deplRate = deplRate * mult
 			if (addiction_prob)
-				src.handle_addiction(M, deplRate)
+				var/addProb = addiction_prob
+				src.handle_addiction(M, deplRate, addProb)
 
 			if (src.volume - deplRate <= 0)
 				src.on_mob_life_complete(M)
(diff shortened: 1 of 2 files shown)


code/modules/chemistry/Chemistry-Reagents.dm
```diff
@@ -322,6 +322,9 @@ datum
 					addProb = round(addProb * 2)
 					rate *= 2
 					//DEBUG_MESSAGE("addictive_personality: addProb [addProb], rate [rate]")
+				if (src.id == "ethanol" || isalcoholresistant(H))
```
mordent-goonstation, line 325: Should be `&&`?

mordent-goonstation (changes requested):
Would prefer this was done on the ethanol reagent only, rather than on every reagent datum. Would require restructuring a little bit to let ethanol call the parent proc (`..()`) with appropriate changes.

code/modules/chemistry/Chemistry-Reagents.dm
MintyPhresh: Yeppp

MintyPhresh:
Changed to make it ethanol reagent only-- thanks for the comment, I learned something new about DM today!

code/modules/chemistry/Chemistry-Reagents.dm
```diff
@@ -253,7 +253,8 @@ datum
 						H.sims.affectMotive("Energy", energy_value)
 			deplRate = deplRate * mult
 			if (addiction_prob)
-				src.handle_addiction(M, deplRate)
+				var/addProb = addiction_prob
+				src.handle_addiction(M, deplRate, addProb)
```
mordent-goonstation, line 257:
```suggestion
				src.handle_addiction(M, deplRate, addiction_prob)
```

code/modules/chemistry/Reagents-Base.dm
```diff
@@ -251,6 +251,14 @@
 				H.contract_disease(/datum/ailment/malady/heartdisease,null,null,1)
 			..()
 
+	handle_addiction(var/mob/M, var/rate, var/addProb)
+		if (isliving(M))
+			var/mob/living/H = M
+			if (isalcoholresistant(H))
+				addProb /= 4
+				rate /= 4
+		..()
```
TobleroneSwordfish, line 260:
```suggestion
		..(M, rate, addProb)
```
I *think* DM will handle the modified argument fine without this, buut it's much clearer what you're doing when it's written like this.

TobleroneSwordfish (changes requested):

PR #19292, opened 2024-06-02
cache medals to player datum

Sovexe:
## About the PR 
* Cache medals to player datum
* Change medal requests to check local cache rather than relying on getmedal cache/hub

## Why's this needed? 
* Faster medal checks
* No need to sleep medal checks anymore

## Changelog 

```changelog
(u)Sovexe
(+)Medal checks have been refactored to be more performant and responsive
```


ZeWaka (changes requested):
needs wrappers around has

ZeWaka:
https://github.com/goonstation/goonstation/issues/19352

Sovexe:
I can only assume the hub failed to respond for them. We can implement a retry for players with null medals

ZeWaka:
@Sovexe do you still want this open

PR #19294, opened 2024-06-02
Prevents attacking space bees with items if you're on help intent

TealSeer:
[QOL] [PLAYER ACTIONS]
## About the PR 

Adds a check just before the parent call on `attackby` for if you're on help intent, to harmlessly bap the bee with whatever object you're holding instead of jumping straight to beating them with it.

## Why's this needed? 

A single misclick in a bee-heavy room can spell your brutal demise and it's called help intent isn't it

### DIFF
--- a/code/obj/critter/bee.dm
+++ b/code/obj/critter/bee.dm
@@ -410,6 +410,9 @@ ADMIN_INTERACT_PROCS(/obj/critter/domestic_bee, proc/dance, proc/puke_honey)
 			if (W.reagents.has_reagent("menthol") && length(W.reagents.reagent_list) == 1)
 				src.visible_message("<b>[src]</b> sniffles a bit.")
 				src.health = min(initial(src.health), src.health+5)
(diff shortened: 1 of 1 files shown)


Sbmhawk:
The behavior is the same for clicking on anything with an item in your hand, I dont think an exception for just bees is the way to do this.

TealSeer:
> The behavior is the same for clicking on anything with an item in your hand, I dont think an exception for just bees is the way to do this.

Certain things respect it, it's just super inconsistent e.g. clicking someone with a plate on help intent will tap them with it

flappybatpal:
Really feel that we should move to making everything more intent focused, like you actually can't harm someone with help.
However, not sure about the usability side of that and what makes sense for something like a c-saber or red saw.

TobleroneSwordfish:
I think an exception for just one critter type would be very confusing, if you want help intent to not be able to harm then it should be universal (although honestly I'm not sure that's a good idea either)

PR #19301, opened 2024-06-03
Revamp manufacturer material checking algorithm & Introduce a few more null checks into manufacturer UI

MeggalBozale:
[Internal] [Game-Objects]
## About the PR 
Redoes the algorithm for manufacturers both UI side and .dm side.

On the .dm side:
General code touch-up. Split functionality of check_enough_materials and get_materials needed.
These were originally used interchangeably, but now the former returns a boolean for checking if materials are sufficient, and the latter for checking the former / removing materials.

Tangentially (This *was* used for a hot minute, before I had to backpedal some changes) allows grump_message to use null target to instead show a visible message in addition to its grump sound.

On the .tsx side:
General code touch-up to the getProductionSatisfaction function.
Add null checks to prevent potential TermOSing on most cases.

## Why's this needed? 
Code quality good


tgui/packages/tgui/interfaces/Manufacturer/index.tsx
```diff
@@ -83,26 +83,30 @@ export const Manufacturer = (_, context) => {
   }
 
   // Get a ManufacturableData from a QueueBlueprint using its type, category, and name.
-  let queueBlueprintRefs = data.queue.map((queued:QueueBlueprint) =>
+  let queueBlueprintRefs = data.queue && data.queue.length
+  && data.queue.map((queued:QueueBlueprint) =>
     blueprints_by_category[queued.category].find((key) => (key.name === queued.name))
   );
```
mordent-goonstation, line 89:
```suggestion
  const queueBlueprintRefs = (data.queue ?? []).map((queued:QueueBlueprint) =>
    blueprints_by_category[queued.category].find((blueprint) => (blueprint.name === queued.name))
  );
```
If the `key` -> `blueprint` change doesn't make sense, please explain what that would represent.

tgui/packages/tgui/interfaces/Manufacturer/index.tsx
```diff
@@ -83,26 +83,30 @@ export const Manufacturer = (_, context) => {
...
+              && data.all_categories.map((category:string) => (
                 blueprints_by_category[category].length > 0 && (
                   <Collapsible
                     key={category}
                     open
                     title={`${category} (${blueprints_by_category[category].length})`}
                   >
-                    {blueprints_by_category[category].map((blueprint:ManufacturableData, index:number) => (
+                    {blueprints_by_category[category] && blueprints_by_category[category].length
+                    && blueprints_by_category[category].map((blueprint:ManufacturableData, index:number) => (
```
mordent-goonstation, line 109:
```suggestion
                    {(blueprints_by_category[category] ?? []).map((blueprint, index) => (
```
Should be able to infer the function argument types? If not, `blueprints_by_category` needs better typing.

tgui/packages/tgui/interfaces/Manufacturer/index.tsx
```diff
@@ -111,7 +115,7 @@ export const Manufacturer = (_, context) => {
                         manufacturerSpeed={data.speed}
                         materialData={data.resource_data}
                         deleteAllowed={data.delete_allowed !== AccessLevels.DENIED}
-                        hasPower={!!data.indicators.hasPower}
+                        hasPower={data.indicators && !!data.indicators.hasPower}
```
mordent-goonstation, line 118:
```suggestion
                        hasPower={!!data.indicators?.hasPower}
```

tgui/packages/tgui/interfaces/Manufacturer/index.tsx
```diff
@@ -83,26 +83,30 @@ export const Manufacturer = (_, context) => {
...
-  let queueBlueprintRefs = data.queue.map((queued:QueueBlueprint) =>
+  let queueBlueprintRefs = data.queue && data.queue.length
+  && data.queue.map((queued:QueueBlueprint) =>
     blueprints_by_category[queued.category].find((key) => (key.name === queued.name))
   );
 
   return (
     <Window width={1200} height={600} title={data.fabricator_name}>
-      {!data.indicators.hasPower && <PowerAlertModal width={100-SETTINGS_WINDOW_WIDTH} height={"100%"} />}
+      {data.indicators && !data.indicators.hasPower
```
mordent-goonstation, line 93: Confirming intended behavior: if `data.indicators` is not set, as you have it this will not show the `PowerAlertModal`.

Further down I think you treat `data.indicators` not being set as out of power?

Broadly, rather than burrowing down each time, just extract it once then use throughout (i.e. here, and wherever else you care about it)?

```
const hasPower = !!data.indicators?.hasPower;
```

tgui/packages/tgui/interfaces/Manufacturer/index.tsx
```diff
@@ -83,26 +83,30 @@ export const Manufacturer = (_, context) => {
...
+      && <PowerAlertModal width={100-SETTINGS_WINDOW_WIDTH} height={"100%"} />}
```
mordent-goonstation, line 94: Huh, surprised I didn't catch this earlier. This is a weird way of doing width. Will revisit at a different time.

mordent-goonstation (changes requested):
Not going to annotate everywhere, but broadly your approach for checking array existence and contents before iterating over is overkill.

This works for cases where `undefined` is an acceptable return type (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining):

```
data.array_maybe?.map((item) => /* whatever */)
```

In cases where it's not, you can do:

```
(data.array_maybe ?? []).map((item) => /* whatever */);
```

Note that if used in a render part, there is a problem with how you have it. The below will output `0` if an empty array is given to it, because of how short circuiting here works.
```
{data.array_maybe && data.array_maybe.length && data.array_maybe.map((item) => /* whatever */)}
```

tgui/packages/tgui/interfaces/Manufacturer/index.tsx
MeggalBozale: Sounds good! Did that, and it fixed one of the errant null errors too.

tgui/packages/tgui/interfaces/Manufacturer/index.tsx
MeggalBozale: Uhh sounds good, this just represents the percentage of the window that is not the settings bar though. 100%-20%.

MeggalBozale:
Added a round(amount, 0.1) to change_stack_amount as doing 5 - 4.8 was giving some 0.1999999999988 and it was messing up material amts elsewhere because of it

tgui/packages/tgui/interfaces/Manufacturer/index.tsx
```diff
@@ -290,7 +293,7 @@ export const Manufacturer = (_, context) => {
                       </Button>
                     </Stack.Item>
                   </Stack>
-                  {data.queue.length > 0 && (
+                  {(data.queue && data.queue.length > 0) && (
```
ZeWaka, line 296: why not use ?. here and everywhere above

MeggalBozale: Ah, probably forgot those. one moment, apologies

tgui/packages/tgui/interfaces/Manufacturer/components/BlueprintButton.tsx
```diff
@@ -49,17 +52,16 @@ const getProductionSatisfaction = (
...
-    if (matchingMaterial !== undefined) {
-      material_amts_predicted[i] -= target_amount/10;
-      patterns_satisfied.push(true);
-    }
-    else {
+    if (matchingMaterial === undefined) {
       patterns_satisfied.push(false);
+      continue;
     }
+    material_amts_predicted[i] -= target_amount;
```
mordent-goonstation, line 63: No change from you required here, going to try to come back to this part of the code in the nearish future to tidy up some of the looping structures a bit.

tgui/packages/tgui/interfaces/Manufacturer/components/BlueprintButton.tsx
```diff
@@ -105,10 +110,10 @@ export const BlueprintButton = (props:BlueprintButtonProps) => {
           <LabeledList.Item
```
mordent-goonstation, line 110: Another "Mordent TODO", specifically re: the line above but I can't comment on that in the PR.

tgui/packages/tgui/interfaces/Manufacturer/index.tsx
```diff
@@ -62,7 +65,7 @@ export const Manufacturer = (_, context) => {
     This is done here instead of sending one big list to reduce the amount of times we need to refresh static data.
   */
   let blueprints_by_category:Record<string, ManufacturableData[]> = {};
-  for (let category_index = 0; category_index < data.all_categories.length; category_index++) {
+  for (let category_index = 0; category_index < (data.all_categories?.length ?? 0); category_index++) {
```
mordent-goonstation, line 68: TODO: Mordent going loopy

tgui/packages/tgui/interfaces/Manufacturer/index.tsx
```diff
@@ -83,26 +86,26 @@ export const Manufacturer = (_, context) => {
...
-  let queueBlueprintRefs = data.queue.map((queued:QueueBlueprint) =>
-    blueprints_by_category[queued.category].find((key) => (key.name === queued.name))
+  const queueBlueprintRefs = (data.queue ?? []).map((queued:QueueBlueprint) =>
+    blueprints_by_category?.[queued.category]?.find((blueprint) => (blueprint.name === queued.name))
   );
 
   return (
     <Window width={1200} height={600} title={data.fabricator_name}>
-      {!data.indicators.hasPower && <PowerAlertModal width={100-SETTINGS_WINDOW_WIDTH} height={"100%"} />}
+      {!hasPower && <PowerAlertModal width={100-SETTINGS_WINDOW_WIDTH} height={"100%"} />}
```
mordent-goonstation, line 95: Mordent TODO: more sane width setting.

code/obj/item.dm
```diff
@@ -596,6 +596,8 @@ ABSTRACT_TYPE(/obj/item)
 	if ((amount + diff) < 0)
 		return 0
 	amount += diff
+	// Fix some of the floating point imprecision
+	amount = round(amount, 0.1)
```
mordent-goonstation, line 600: Thanks BYOND.

code/obj/machinery/manufacturer.dm
```diff
@@ -1607,43 +1609,46 @@ TYPEINFO(/obj/machinery/manufacturer)
...
+			if (!R.is_match(P.material))
+				continue
+			// We can use this material! Get the amount of free material and reserve/mark as used whatever is free.
+			var/P_ref = "\ref[P]"
+			var/amount_free = P.amount - mats_reserved[P_ref]
+			var/amount_to_use = min(amount_free, required_amount / 10)
+			if ((amount_to_use * 10) < required_amount)
+				continue
+			mats_reserved[P_ref] ||= 0
+			mats_reserved[P_ref] += amount_to_use
```
mordent-goonstation, line 1638: Per "thanks BYOND" comment, possible we need to handle dumb additive behavior here as well?

mordent-goonstation (approved):
Broadly looks fine, the comments I have as TODOs aren't blockers and if they never get done it's *fiiiine*.

Broadly, I think the whole `0.1` material thing has highlighted to me how 1 block of material should add 10 to the count (then multiply all existing costs by 10) just to get rid of all of the float errors, but then I'm sure one day someone will want "just half of a tenth of a block" or something and back we are.

Above is just me grumping about material code, not direction for change from you.

code/obj/machinery/manufacturer.dm
MeggalBozale: Thankfully not, the only way this happens is when maths isn't done right. Rounding here would imply that there's flexibility with item amounts, which, fine, if that's desired -- but it also means people can start doing stuff like
produce 0.1 sheet using 0.990 extremely valuable metal
x10
reclaim sheet
repeat ad infinitum
and now you have infinity infused gibbis material.
IDK, i don't really care about that weird thing. But I feel I have to mention it before someone else does, because it tends to come up.

MeggalBozale: I did try a few implementations here, by the way -- and I settled on the above because any implementation that's not going to address the cause (that 5 - 4.8 is not evaluating to 0.2 as per intent) doesn't stop it from escalating to someone putting in like, 100 bars, printing 1000 sheets, and then figuring out that floating point math makes that last 0.1 unprintable.

If it's just about using both as a sort of layered approach though, sounds fine? As it is now, though, this issue should not come up.

code/obj/item.dm
MeggalBozale: Yep.....

MeggalBozale:
> Broadly, I think the whole 0.1 material thing has highlighted to me how 1 block of material should add 10 to the count (then multiply all existing costs by 10) just to get rid of all of the float errors, but then I'm sure one day someone will want "just half of a tenth of a block" or something and back we are.
Yes, I noticed this pattern going into the code, and chose to bring it back to amount to remove the need for a secondary, redundant var. There's really one... ok... solution here that I see --

-- which is choosing to "digitize" all resources which enter the fabricator, all vars and everything, and then having the amount the fab shows as *10. I'm very against this for the reason that it would be such a headache, and for very little benefit versus a single line to round the result in a public-facing proc for the bars, anyway. And then you have to de-digitize them as they exit. And deal with all the consequences/bugs of that. And by you I mean me, just sort of speaking "you" as the general audience here. Just wanted to clarify that whilst rambling.

And then as you mention, it opens up the issue of supporting more fine-grained amounts for *all* bars, and now the more decimals to precisely support, the less headroom there is for the actual material bars, and now bars with an amount of 1000 are one bar, and it just could end up being a whole mess.

I think this is really just an issue of fabricators not using the `amount` var properly though, because it has recipes which demand fractions of amounts. I don't see how that's being fixed anytime soon, though.

MeggalBozale:
!merge_upstream

MeggalBozale:
it never works for me does it

ZeWaka:
gsmimh

PR #19303, opened 2024-06-03
Properly equip donut 2's detective office

JORJ949:
## About the PR 
Adds a lot of items missing from the detectives office on donut 2 (and removes the extra thermals because the detective already gets some in their locker and doesnt need a second pair)
image
image

## Why's this needed? 
Go Go Gadget Cigarette
No camera monitor, no evidence boxes, no lighter, no computer.


JORJ949:
> Any reason given for the removal of the extra thermals? It's commonplace (and a design intention, as far as i know) for different departments on different maps to have different type of missing or surplus equipment avaible.

Any old staffie shouldnt be able to break one door for free cloaker counter, every other map has these in the det/hos locker and armory.

JORJ949:
> > Any old staffie shouldnt be able to break one door for free cloaker counter, every other map has these in the det/hos locker and armory. And the det locker still has a pair
> 
> That's debatable. But at least put that in your reasoning given in the PR and add the balance tag.

Already had added the balance tag

PR #19305, opened 2024-06-03
Add new vars to corpse spawners

Glamurio:
[QoL] [Internal]

## About the PR 
Atomizes most new subtypes and extra vars from #18897

Changes the default values of `empty_pockets`, `empty_bag` and `delete_id` to TRUE.

## Why's this needed? 
Useful tools for future maps.

### DIFF
--- a/code/mob/living/carbon/human/normal.dm
+++ b/code/mob/living/carbon/human/normal.dm
@@ -96,11 +96,6 @@
 		..()
 		JobEquipSpawned("Chaplain")
 
-/mob/living/carbon/human/normal/artist
-	New()
-		..()
-		JobEquipSpawned("Artist")
-
 /mob/living/carbon/human/normal/bartender
 	New()
 		..()
--- a/code/mob/living/carbon/human/normal.dm
+++ b/code/mob/living/carbon/human/normal.dm
@@ -96,6 +96,11 @@
 		..()
 		JobEquipSpawned("Chaplain")
 
+/mob/living/carbon/human/normal/artist
+	New()
+		..()
+		JobEquipSpawned("Artist")
+
 /mob/living/carbon/human/normal/bartender
 	New()
 		..()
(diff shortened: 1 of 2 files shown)


code/mob/living/carbon/human/normal.dm
```diff
@@ -96,6 +96,11 @@
 		..()
 		JobEquipSpawned("Chaplain")
 
+/mob/living/carbon/human/normal/artist
```
ZeWaka, line 99: ?

code/modules/mapping/helpers/mob_spawn.dm
```diff
@@ -72,108 +82,200 @@
...
 
 	/// If TRUE we delete the contents of the backpack after spawning
-	var/empty_bag = FALSE
+	var/empty_bag = TRUE
 	/// If TRUE delete the pocket contents if any
-	var/empty_pockets = FALSE
+	var/empty_pockets = TRUE
 	/// If TRUE we delete the ID slot contents after spawning
-	var/delete_id = FALSE
+	var/delete_id = TRUE
```
ZeWaka, line 93: you're changing default functionality here?

code/mob/living/carbon/human/normal.dm
Glamurio: Right. This is technically a leftover from a corpse job preset, and since those use the defines in `normal.dm` and artist was missing, I added it, but we can just remove it entirely.

code/modules/mapping/helpers/mob_spawn.dm
Glamurio: Yes. Should probably specifically also mention that in the PR, but Tarmunora requested that corpse spawns should have their IDs stripped by default.
https://github.com/goonstation/goonstation/pull/18897#pullrequestreview-2056013570

PR #19307, opened 2024-06-03
Update syndicate_buylist.dm to price the syndicate hat at 11 TC

PastaSpaghett:
[A-Clothing] [A-Gamemodes] [C-QoL]

## About the PR 

Reduces the price of the Syndicate Hat by 1, and changes the description slightly

## Why's this needed? 

The syndicate hat is really just a silly item to be a mustache twirling evil bad guy, but purchasing the hat would prevent the buyer from purchasing anything else. This change allows buyers to pick up another 1 TC item to go along with their evil hat, like a fake mustache or agent card. This shouldn't change balance, as the syndicate hat was never really meant to be balanced, it was mostly an item used for the aesthetic, which would be enhanced with the addition of a fake mustache or an agent card assigned to Dr. Badguy, Head of Evil.  

I can't have a cool evil mustache and have a cool evil hat at the same time ):<

## Changelog 

```changelog
(u)Spaghett Pasta
(+)Syndicate Hat price changed from 12 TC to 11 TC
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -280,8 +280,8 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/generic)
 /datum/syndicate_buylist/generic/bighat
 	name = "Syndicate Hat"
 	item = /obj/item/clothing/head/bighat/syndicate
-	cost = 12
-	desc = "Think you're tough shit buddy?"
+	cost = 11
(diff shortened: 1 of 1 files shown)


MintyPhresh:
I know this is very similar to #18975 and probably will run into the same issue, but I'm not personally against it

ZeWaka:
Closing this due to concerns above.

PR #19310, opened 2024-06-03
Moves some code from update_inhands onto the item

Valtsu0:
[Internal] [Code-Quality]
## About the PR 
Should not have any visible changes

## Why's this needed? 
Less repeated code
Allows items to properly use overlays in inhand sprites and probably few other things


Valtsu0:
wow that was fast

PR #19311, opened 2024-06-03
Add a silly effect for emagging manufacturers

MeggalBozale:
## About the PR 
Emagging a manufacturer makes it immediately flip out and malfunction for the first time, and jumbles up blueprint information.

## Why's this needed? 
Emag effects are silly, and I think this harmless effect is cool.

## Changelog 

```changelog
(u)MeggalBozale
(+)Manufacturers react worse to being emagged.
```


mordent-goonstation:
I get the intent, but this feels like an accessibility concern to me. An emag is an expensive item, either it should have a positive effect (hacked etc.) for speed for the user, or a negative effect to make using the machinery harder/impossible without some work ("bricking" it).

MeggalBozale:
I don't see why it has to be so clear-cut? This is an effect that only occurs when a traitor decides to emag the machine -- it hacks it for you in one swipe, but also makes things a bit jumbled to read. The images are still in place, and the text is not entirely unreadable with the current configuration.

mordent-goonstation:
Could you give me a short recording of how it looks with you clicking around (as I believe anything that updates static data would rechurn the scrambling)?

MeggalBozale:
Sure, just a minute

MeggalBozale:
https://github.com/goonstation/goonstation/assets/27376947/819e2583-260a-442b-811e-af843356b15f

@mordent-goonstation

TobleroneSwordfish:
ai vinkkr

MeggalBozale:
true that

mordent-goonstation:
I don't much like it and think it's unnecessary extra UI weight that isn't a value-add.

I also really think the accessibility concern is there. Dyslexia and the like is going to make this from "slight garble, fun" to "actually affects the player". I do not think there is any amount of probability lowering that justifies doing that (and, at a low level, would actually just look like some text typos).

An example comparable UI is the turret control one, which when emagged is *very* clearly utterly corrupted and angry at you (forces turrets to kill mode), which feels acceptable because reading the exact text is not the concern there.

Happy if another dev wants to greenlight an alternative emag effect, this one isn't it.

PR #10505, opened 2022-08-29
Severely reduces brute/burn healing from cryox, increases healing from effects of crit from cryox

ThatFiveGuys:
## About the PR 
This pr reduces the brute and burn healing from cryox from 12 to 2. Cryoxadone now has a 50% chance of reducing lose breath however and now reduces the chance of death the same amount as synaptropine (does not stack). This also removes the heating up from cryoxadone as it shouldnt be needed anymore.

## Why's this needed? 
I am personally of the take that generally, ignoring doctors should not be faster and/or easier then going back into medbay and doing it yourself. Not only is cryo insanely easy to access, its also far faster, with some of the highest healing rates in game. This shifts the niche of cryo from a all around amazing healer to something that youll want to use as a stopgap for patients to buy time to heal them. Oh also semi nerfs cryopills but those arent used that much anymore so. This mostly just makes cryo pills less of a mid combat rapid healer and more of a simple and reliable slow healer at the cost of less mobility for things like azones.

## Changelog

```changelog
(u)Ikea, tarmunora
(*)Cryoxadone brute/burn healing has been reduced from 12 to 2, oxy healing also reduced to 2, however it now can combat losebreath, counters radiation, and provides some protection from random death rolls in crit while in effect. Cryoxadone will no longer work when body temp is below -240C or so
```

PR #19312, opened 2024-06-03
Reworks the cryoxadone recipe to be a bit more simple and cooling

M-Earthfire:
[Chemistry][Balance][Rework][add to wiki]

## About the PR 

Changes the cryoxadone recipe:

from: 
- (1) Cryostylane + (1) Unstable Mutagen+ (1) Plasma + (1) Acetone -> (3) Cryoxadone
to:
- (2) Cryostylane + (1) Platinum + (2) Water -> (2) Cryoxadone @ less than 323K

The reaction now happens over time and slowly cools down the reaction vessel it is in, similary to cryostylane + oxygen.

A chemical reaction effect was added to show the cooling of the liquid

## Why's this needed? 

With the chem rework, many chems were tied to the now much more involved reaction-chain of welding ->oil -> acetone.

While this reaction chain makes multiple reaction paths more interesting, it added additional hurdles to chemicals that were already barely worth it to make.

After the Cryoxadone-reworkin #10505, the chems use was highly diminished outside of cryotubes. Thus, this chemical can be moved away from being a endpoint of synthesis to be an advanced precursor-chem. For this, acetone and mutagen were replaced with base-chemicals.

the recipe change enables people to add cryoxadone in further PR's as an indrigient to other chemical recipes. To make this interesting, a cooling effect was added to the reaction so the cold instead of heat needs to be managed. This cooling effect will freeze the water in the reaction, effectively stopping the reaction.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)The recipe of cryoxadone was changed to: (2) Cryostylane + (1) Platinum + (2) Water -> (2) Cryoxadone @ less than 323K
```


Flaborized:
Hi sorry for commenting so late I missed this PR initially; I personally think that changing acetone to something else is a good idea as it's probably too much for what the chem is right now, saline isn't particularly my favorite substitute though because while it's not hard to make, it's a pretty tedious "click a lot of things" recipe. I feel like most commonly doctors would just make this themselves since they have tons of it and it's not particularly valuable or anything.

Ideally it'd either have some interesting quirk to make like other Updated Meds or require fuel (or even just be a really simple recipe), chem rework tried to move away from bigger lists of somewhat tedious/You Just Click A Lot chems. The ice visual effect and cooling effect are good and fine though.

PR #19314, opened 2024-06-03
Fish oil pyrolysis as an oil source and increase in fish oil avaiability

M-Earthfire:
[Balance][Feature][chemistry][add to wiki][Catering]

## About the PR 

This PR adds a new recipe to fish oil:

(1) Fish Oil @ 423 K & Time -> (0.2) Potash + (0 - 0.8) Oil +  (0 - 0.8) Ash

The ratio of oil to ash depends on two factors:
1. If the reaction is done in an open container, no oil and 100% ash is produced
2. If the reaction is done in a closed container, the amount of oil produced will lineary increase from 0.2 to 0.8, the more the temperature approaches 438K. At 423K or above 453K the minimum amount of fish oil is created.

To make this route more usefull, the amount of fish oil in fish is doubled from 10u to 20u.

## Why's this needed? 

There was feedback that asked for more sources of oil. And there was feedback that asked for more uses of fish oil and fish in general.

This PR adds both, a way for rancher/angler/botanists to be able to supply chemistry with an easy source of oil right from the start of the game. Or cargo, if they get a fishing kit.

The reaction is best done by chemistry, since they have access to big closed containers (via lids/large beakers) and chem heaters to properly heat the fish oil to the required temperature. The reaction speed is vastly faster than the welding fuel route and has a better yield.

To additionally push this route and because fishing can be tedious, the amount of fish oil was increased.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)The amount of fish oil in fish has been doubled.
(+)Heating fish oil in closed containers can pyrolyse it to a usefull amount of oil, ash and potash.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -609,6 +609,33 @@
 		required_reagents = list("phlogiston" = 1, "thermite" = 1, "fuel" = 1)
 		result_amount = 1 */
 
+	fishoil_pyrolysis
+		name = "fish oil pyrolysis"
+		id = "fishoil_pyrolysis"
+		result = "potash"
+		required_reagents = list("fishoil" = 1)
+		min_temperature = T0C + 150
+		result_amount = 0.2
+		instant = 0
+		reaction_speed = 0.4
+		mix_phrase = "The oil starts to bubble and turn into a back tar."
+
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish (approved):
I like this, it should still be slower than making it manually if you include fishing time but gives anglers something to do with their massive pile of spare fish other than buy endless hats.

Gannets (approved):

PR #19318, opened 2024-06-03
Teas no longer inherently addictive

MintyPhresh:
## About the PR
All teas are of the subtype /fooddrink/caffeinated, which decays into the addictive chemical caffeine. I'm not sure why the addiction parts were never removed when caffeine was added, so I'm assuming oversight.

## Why's this needed?
leaf water is addictive because it has caffeine in it, not because it is leaves water


mordent-goonstation:
Comedy option: tea itself is addictive if you have any of the various British accents.

PR #19324, opened 2024-06-04
Add blob context menu actions for upgrading tiles

FlameArrow57:
[GAME MODES][FEATURE][QoL][BALANCE][ADD TO WIKI]

## About the PR 
This PR adds context menu upgrade actions for blobs upgrading tiles.

When a blob player left clicks one of their tiles, they will be given a context menu with what they want to upgrade it to, with allowed upgrades being thick membrane, fire resistant membrane, slime launcher, and reflective cell.

image

## Why's this needed? 
Hopefully faster gameplay

Reduces some of the "clunkiness" in needing to hover over tiles to upgrade them, with hotkey'ing more common abilities in mind

## Changelog 

```changelog
(u)FlameArrow57
(*)Blobs can now left click basic blob tiles to open a context menu for what they want to upgrade the tile to.
```

### DIFF
--- a/code/obj/blob.dm
+++ b/code/obj/blob.dm
@@ -121,17 +121,9 @@
 		if ("right" in pa)
 			right_click_action()
 		else if (src.type == /obj/blob && !(("alt" in pa) || ("shift" in pa) || ("ctrl" in pa)))
-			if (!src.contexts)
-				src.contexts = list()
-				var/datum/contextLayout/experimentalcircle/context_menu = new
-				context_menu.center = TRUE
-				src.contextLayout = context_menu
-				for(var/actionType in childrentypesof(/datum/contextAction/blob_tile_upgrade))
-					var/datum/contextAction/blob_tile_upgrade/upgrade = new actionType()
-					src.contexts += upgrade
-			var/mob/user = usr
+			var/mob/living/intangible/blob_overmind/user = usr
 			user.closeContextActions()
-			user.showContextActions(src.contexts, src, src.contextLayout)
+			user.showContextActions(user.tile_upgrade_contexts, src, user.contextLayout)
 		else
 			..()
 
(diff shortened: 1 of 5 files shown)


code/obj/blob.dm
```diff
@@ -118,9 +120,17 @@
 		var/list/pa = params2list(params)
 		if ("right" in pa)
 			right_click_action()
+		else if (src.type == /obj/blob && !(("alt" in pa) || ("shift" in pa) || ("ctrl" in pa)))
+			var/mob/living/intangible/blob_overmind/user = usr
+			user.closeContextActions()
+			user.showContextActions(user.tile_upgrade_contexts, src, user.contextLayout)
```
TobleroneSwordfish, line 126: This seems to result in the options being a bit offset from the tile, maybe due to the circle layout not accounting for 32x32 sprites?

code/obj/blob.dm
```diff
@@ -1067,6 +1077,51 @@
 	update_icon(override_parent)
 		return
 
+/datum/contextAction/blob_tile_upgrade
```
TobleroneSwordfish, line 1080: It'd be really cool if these showed the cooldowns, maybe just shove the actual ability button in the vis_contents of the context option?

code/obj/blob.dm
```diff
@@ -1067,6 +1077,51 @@
...
+		name = "Build Thick Membrane"
+		icon_state = "blob-wall"
+		associated_ability = /datum/blob_ability/build/wall
+
+	fire_membrane
+		name = "Build Fire-Resistant Membrane"
+		icon_state = "blob-firewall"
+		associated_ability = /datum/blob_ability/build/firewall
+
+	slime_launcher
```
TobleroneSwordfish, line 1113: Why is this here and not other tile upgrades like ectothermids and lipid cells? I think keeping it to just the three "wall" types would make more sense

TobleroneSwordfish (changes requested):

code/obj/blob.dm
FlameArrow57: I was just focusing on the more common frontline/combat type cells, agree maybe slime launcher isn't needed tho

code/obj/blob.dm
FlameArrow57: Yeah not sure really, code is at /datum/contextLayout/experimentalcircle/showButtons(). Seems either intentional, or something wrong with math

image

PR #19325, opened 2024-06-04
Clown announcement computer

alicars:
[FEATURE] [GAME OBJECTS] [SPRITES]

## About the PR 

Adds an illegal announcement computer to the clown hole! The computer only works 1 time, and shocks/sets the clown on fire for using it. It also sets them to arrest!

https://github.com/goonstation/goonstation/assets/48467372/b40d5b2c-00de-43ee-922f-2daab260b475

clouwcement 3
(The screen occasionally flickers to a BSOD to hint that its malfunctioning) 

This puts the computer on every map except Cog1 and Atlas. If it gets merged, would be happy to do a follow up PR adding some extra room to at least the Cog 1 clown hole, but that seemed a bit out of scope to start with

The computer is technically useable by anyone, since there is no Clown-exclusive access right now and adding that seemed like a whole can of worms. 

Happy to edit or re-balance any part of this, I kind of just went with what seemed funny to me. The sprites are probably what I'm most unsure of:

clown announcemnet computer main sprites w broken

## Why's this needed? 

This was originally a discussion thread on the forums: https://forum.ss13.co/showthread.php?tid=22341

It seemed funny and mostly well received so I went for it! Most of the features of the computer are from comments on that thread

## Changelog 

```changelog
(u)crunchwarp
(+)A special announcement computer has been added to most clownholes. But beware...
```


JORJ949:
I feel like the announcement sound should be a honk or something similar

alicars:
Please ignore the long list of irrelevant commits at the beginning, I accidentally forked this from the incorrect branch...

Glamurio:
Funny idea, don't see why not. And I agree that there should be a unique honking announcement noise, if you're already setting yourself on fire and you can only do it once, might as well make it really stand out. Could use like a pitched vuvuzela noise, or even play clips of the existing honk collection soundtrack.

Valtsu0:
clearly we now need a clown tgui theme

MintyPhresh:
I'm fine with the idea, but it being an every round thing sounds like it could get tiresome. What do you think about it being a prob(20) to spawn if placed in a map? Maybe the other 80% of the time it still spawns, but needs fixing?

Glamurio:
> I'm fine with the idea, but it being an every round thing sounds like it could get tiresome. What do you think about it being a prob(20) to spawn if placed in a map? Maybe the other 80% of the time it still spawns, but needs fixing?

If it's only a chance to spawn then I think it'd be okay to be usable multiple times a shift, but with a cooldown.

MintyPhresh:
> If it's only a chance to spawn then I think it'd be okay to be usable multiple times a shift, but with a cooldown.
I disagree... clowns have no sense of right and wrong, so that thing's bound to be overused to death if it can be repeatably used, unless the cooldown's unreasonably long.

code/obj/machinery/computer/announcement.dm
```diff
@@ -213,3 +214,31 @@
 		name = "Syndicate Announcement computer"
 		voice_name = "Syndicate Announcement Computer"
 		theme = "syndicate"
+
+/obj/machinery/computer/announcement/clown
+	req_access = null
```
TobleroneSwordfish, line 219: There's no such thing as clown access buuut maybe you could make it require a clown ID since that's unique by itself? I'm just thinking about random staffies wandering in and using the one announcement instead.

TobleroneSwordfish (changes requested):
This is dumb and I like it, clown swan song! Definitely make the text comic sans and the sound effect something annoying/funny though.

alicars:
> I'm fine with the idea, but it being an every round thing sounds like it could get tiresome. What do you think about it being a prob(20) to spawn if placed in a map? Maybe the other 80% of the time it still spawns, but needs fixing?

That was part of my thought with making it 1 use before it explodes. The clown gets one chance at a joke, better make it good! vs. spamming it every 5 minutes or whatever. And I guess I don't fully get why it would be all that tiresome, since there are already lots of other announcement computers on the station? Though I'm sure clowns would use it for both good and evil depending, haha. But would be happy to try to implement this if others feel strongly about it too. Could be cool to have it be more of a rare bonus than a sure fire thing 

And these other suggestions I'll work on soon! Appreciate everyone's feedback and ideas

code/obj/machinery/computer/announcement.dm
Valtsu0: alternatively it could require clown training (and maybe mime training could announce "...")

JORJ949:
Emag could remove the downside? (or make it worse)

code/obj/machinery/computer/announcement.dm
TobleroneSwordfish: I'd prefer tying it to something tangible rather than hidden training properties. If you wanted to be really mean you could make it shred the ID too, since it's canonically made of cardboard. (maybe an emag effect?)

alicars:
OK new things in!

-The announcement is now in comic sans
-The computer can only be used by someone with an authentic clown ID 
-Emags will now upgrade the fake explosion to a real explosion (explosion_new is called). This is the same level of severity as a pod exploding and does cause a small 1 tile breach on some maps. Could scale this up or down as desired, or just switch to shredding the ID if its too much. This shouldn't put a clown at full in crit, unless they stick around oxygenless for a while
-A new sound effect, announcement_clown.ogg. Just a couple simple audacity effects put on the classic bike_horn1.ogg we all know and love. If this seems like it sucks, the suggested pitched up vuvuzela also sounded pretty good to me. Just thought it would be a little more clowny as a honk :o)

TobleroneSwordfish (approved):

PR #19330, opened 2024-06-04
Nerf Prairie Fire's capsaicin content

MintyPhresh:
## About the PR
Reduces the probability and volume of Capsaicin surges, from a 20% chance for 10-20 units per tick to a 5% chance of 4-12 units per tick. This changes Prairie Fire's average Capsaicin-by-volume content from being 750% CPV to just 100%, which still allows for capsaicin-duping without making the drink unreasonably spicy.

## Why's this needed? 
7.5 units of capsaicin per 1 unit of prairie fire was simply too much.

### DIFF
--- a/code/modules/chemistry/Reagents-FoodDrink.dm
+++ b/code/modules/chemistry/Reagents-FoodDrink.dm
@@ -1507,8 +1507,8 @@ datum
 
 			on_mob_life(var/mob/M, var/mult = 1)
 				if(!M) M = holder.my_atom
-				if(prob(20))
-					M.reagents.add_reagent("capsaicin", rand(10,20) * mult)
+				if(prob(5))
+					M.reagents.add_reagent("capsaicin", rand(4,12) * mult)
 				if(prob(10))
 					M.reagents.add_reagent("histamine", rand(1,5) * mult)
 				..()


MintyPhresh:
Prairie Fire is mass-producible even without botany, requiring only 2-3 units of Capsaicin to make (which can be fished from vending machines), and then you need only spike a Tequila with a little bit of Prairie Fire to set the whole drink ablaze. The capsaicin the victim experiences causes stuttering and stunning (with weapon drops) that can addict, make communication harder, and make you supremely vulnerable in a fight. Compare to BOOrbon or Bloody Scary, which have similar levels of difficulty in mixing; those are at 130% ABV and 230% ABV respectively, and alcohol only starts dealing missteps at 15u in the system, which takes like 20-ish ticks. 

At 750% CPV Prairie Fire stands out as an especially debilitating drink. Seriously, if you drink a 50-unit cup you'll be nearly reagent-capped with capsaicin by the end of it. Even at 100%, the concept is ridiculous (you're essentially drinking a cup of capsaicin) but instead of magically filling up your bloodstream like a thermometer in a furnace it only provides intermittent doses of the chemical.

MintyPhresh:
I don't think something needs to be abused to warrant a nerf; something being in the game and not causing problems doesn't mean it's good. If PF were in the process of being added today, the design of the drink would be questionable at best. Its behavior isn't consistent with other mixed drinks in its class, and violating conservation of volume isn't usually something we like to do for non-magical stuff, especially not to the absurd degree PF does it. VHFCS and Charcoal make sense, since there's no quality system for chems, but adding a splash of tequila to hot sauce doesn't make it nearly 8 times as spicy.

MintyPhresh:
Agreed that capsaicin has a high depletion rate. But there's no threshold until its effects activate; the only threshold is one at 20u that makes it about twice as likely to activate the stun (but doesn't reduce the cooldown for it at all)...
image

I was aware of the depletion rate when I made the PR; I thought bursts of fiery hotness to be appropriate, considering that it can take effect instantly. Plus, even with the high depletion rate, the capsaicin from a 50u glass of Prairie Fire in its current form lasts longer than the ethanol from a 50u glass of rum and coke. I could see it being like, 150% CPV, though, if accumulation is important.

PR #19331, opened 2024-06-04
TGUI Map Vote

Garash2k:
## About the PR 
Upgrades the map vote window and map vote results windows to TGUI

image

image

image

image

Admin view:
image

## Why's this needed? 
Thumbnails! Also looks nicer.

## Changelog 

```changelog
(u)Garash
(+)New UI for the Map Vote & Map Vote Results windows, do your civic duty today!
```

[Feature]


mordent-goonstation:
Not reviewed the code at all, but a tooltip is not the right place for a potentially very long comma-separated list.

tgui/packages/tgui/interfaces/MapVote/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  );
+};
+
+
+export const MapPanel = (props) => {
+  return (
+    <Stack.Item>
+      <Section title={props.mapName} textAlign="center" width="100px">
+        <Box>
+          <Image src={`https://goonhub.com/storage/maps/${props.mapID.toLowerCase()}/thumb.png`} backgroundColor="#0f0f0f" width="75px" />
```
ZeWaka, line 64: Please pass this in and use the goonhub_url helper or whatever it's called that wire made last week

code/mapSwitching.dm
```diff
@@ -301,6 +281,11 @@ var/global/datum/mapSwitchHandler/mapSwitcher
 				var/vote = href_list["vote"]
 				src.composeVoteReportSimple(vote)
 
+	proc/get_player_pickable_map_list()
+		var/list/mapList = new()
+		for (var/map in src.playerPickable)
+			mapList.Add(list(list(id = src.playerPickable[map]["id"], name = map)))
+		return mapList
```
ZeWaka, line 288: `.`, `. +=`

ZeWaka (commented):
why would you need light mode screenshots? They're not going to change, lol

444Portal:
i think the images could do with being zoomed in

tgui/packages/tgui/interfaces/MapVote/index.tsx
Garash2k: Mine's null, is it supposed to be? Should I assume it works live? or tweak a config somewhere?
image

Garash2k:
> i think the images could do with being zoomed in

I did them like they are on https://goonhub.com/maps, 75x75 with a dark background. With a go from the devs I could tweak it from that, maps like Donut 2 are pretty off center though, and them being the same format is a plus

MeggalBozale:
while i do think this looks nice, im not sure if a whole section for two buttons in a section is necessary. and if so, you could just use the `buttons` attribute / have it be a toggleable button -- that way it's just one.

BatElite:
does this just get wider as the list of candidates expands?

tgui/packages/tgui/interfaces/MapVote/index.tsx
ZeWaka: #19334

Garash2k:
> does this just get wider as the list of candidates expands?

It does, about 100px per map

@MeggalBozale

Good ideas! I'll look into them next time I dive in

Sovexe:
Believe this will have issues with updating as the TGUI process loop doesnt run pregame, may have to adjust for such

tgui/packages/tgui/interfaces/MapVote/index.tsx
ZeWaka: done

ZeWaka:
ORGANIZATION IS BANNED YOU HEARD IT HERE FIRST FOLKS

ZeWaka:
d53751fb54981a2ab7249c16352869cb8adec9d8

Xkeeper0:
I testmerged this onto Classic (and will shortly be merging it into the others!)

For the main voting window, I would like the selected maps to have some kind of subtle highlighted background. a bit more than the red/green checkmark.


Otherwise, I like it, though the default layout may work better with the map names _under_ the maps -- that way the longer "oshan laboratory" doesn't push its thumbnail down and look awkward.

Here are my thoughts on the admin review:
image


As an additional thought, we may want to consider adding some kind of tooltip when you are hovering over a map, with a short description and some other information:

> **Cogmap 1**
> A long-time classic.
> **Location**: Space
> **Engine**: TEG/Nuclear/Singularity
> **Mining**: Magnet, Asteroid Field
> **Ideal Players:** 69-420

> **Oshan Laboratory**
> Sealab 2053.
> **Location**: Ocean Planet Abzu
> **Engine**: Geothermal Hotspots
> **Mining**: Trench
> **Ideal Players:** 12-34

(to be clear this is 100% a future thing, i like it quite a bit as is and will be tming it into the other servers shortly)

Xkeeper0:
if you want to simulate multiple votes, you might be able to spawn additional voting booths; those accept space_cash and will let you place an additional vote, and they (should) show up in the list as their `[ref]`; just make a bunch of them, spawn some money, and pay for additional votes. it might work!

Garash2k:
Oooh testmerge
Thanks for the feedback! I was hesitant to put more work in the admin view before knowing what was wanted, now I can!

Garash2k:
Before I forget, tooltip with hardcoded demo values that just needs data plugged from the DM side to work:
image

ZeWaka (commented):
honestly i liked it way more before you redesigned it, now the map sections are way too wide and the green background is super garish

Garash2k:
Regarding the Oshan Laboratory point, I've enlarged the panels a little, and made it so the text cuts if needed.
image

All other points have been adressed! Looking at the screenshots, I think I'll work on the background a bit to make it more subtle

Garash2k:
Just spitballing the idea here, if we had zamu's tooltip idea, it wouldn't be that big of a deal to have map names truncated, we could go back to the 100px width the panels were at the PR's creation.

Also, of note, there are `MAP_PANEL_WIDTH` and `PANEL_PER_LINE` constants (current desc. is at `150` and `4`,) so things are pretty easy to change on the fly!

image

PR #19333, opened 2024-06-04
Adds emote text to *udder command

### LINKED ISSUE #17406: bovine clothing emote throws error (but still works)
### Labels

[BUG][TRIVIAL]

### Description

using the *udder command from the testmerge (which btw looks great and i havent had/seen any performance issues from personally) gives an error after sending the confirm message (but the command still works)

### Steps to reproduce

1.  *udder
2. Disabled bovine-specific clothes filters.
3. Unusable emote 'udder'. 'Me help' for a list.

### Expected Behavior

maybe i just dont understand how commandwork but i dont think it should send Unusable emote 'udder'. 'Me help' for a list.

Reported by: Joeled
Active test merges: #17043
Round log date: 2023-12-21-01-41
Reported on: 2023-12-21 01:53:04
Map: OSHAN

AXLFAN2000:
## About the PR 
Adds an emote string to the *udder command which was missing one, and instead returned an error when used

Unusable emote 'udder'. 'Me help' for a list.
vs
image

## Why's this needed? 
Fixes  #17406

[bug][player actions][mutantraces]

### DIFF
--- a/code/datums/mutantraces.dm
+++ b/code/datums/mutantraces.dm
@@ -2246,6 +2246,7 @@ TYPEINFO(/datum/mutantrace/cow)
 				src.clothes_filters_active = !src.clothes_filters_active
 				boutput(src.mob, src.clothes_filters_active ? "Bovine-specific clothes filters activated." : "Disabled bovine-specific clothes filters.")
 				src.mob.update_clothing()
+				. = "<B>[src.mob]</B> readjusts [his_or_her(src.mob)] udders."
 			else
 				.= ..()
 


MintyPhresh:
hmm... Is this visible to all players? Might be better as a self-only thing

AXLFAN2000:
> hmm... Is this visible to all players? Might be better as a self-only thing

That was the original goal, but we went through it together and with the way emote code works we couldn't find way to have it output nothing, it's either it sends something in the chat or it sends the error which is why I took this route.  If you can figure out how to make it both not send the error and not send a message in the chat, please let me know how!

code/datums/mutantraces.dm
```diff
@@ -2246,6 +2246,7 @@ TYPEINFO(/datum/mutantrace/cow)
 				src.clothes_filters_active = !src.clothes_filters_active
 				boutput(src.mob, src.clothes_filters_active ? "Bovine-specific clothes filters activated." : "Disabled bovine-specific clothes filters.")
 				src.mob.update_clothing()
+				. = "<B>[src.mob]</B> readjusts [his_or_her(src.mob)] udders."
```
Glamurio, line 2249: `boutput` should work. No need to do `. =` either, just:

`boutput(usr, SPAN_NOTICE("You readjust your udders."))`

Glamurio (changes requested):

code/datums/mutantraces.dm
ZeWaka: Or visible_message depending on what you want

TealSeer: This would ultimately lead to the same problem as before, the way emote code is written an emote needs some kind of return value for it to be considered valid, if it doesn't have one then it'll say "Unusable emote" (even if it works).

Glamurio: > This would ultimately lead to the same problem as before, the way emote code is written an emote needs some kind of return value for it to be considered valid, if it doesn't have one then it'll say "Unusable emote" (even if it works).

Right. I see your concern. Honestly, this seems like more of a necessary refactor of emote code than an udder PR?
The real reason seems like that mutantraces use their own scuffed emote code and that should get fixed instead. This PR really just otherwise disguises a limitation.

AXLFAN2000: If a larger refactor is necessary as opposed to fixing this I'm just going to close this PR  --  that's beyond the scope of my current knowledge/time/abilities and would be better served being done by someone who understands it, I just wanted to fix the bug with what was available in code now

PR #19341, opened 2024-06-05
Make blob player left clicks automatically cast the spread ability

FlameArrow57:
[GAME MODES][PLAYER ACTIONS][FEATURE][QoL][ADD TO WIKI]

## About the PR 
Makes it so that when a blob player left clicks something, it will attempt to call the spread ability on the tile.

## Why's this needed? 
-Allows easier use of main spammable ability without disrupting anything else intended
-Allows other abilities to be hotkeyed
-Hopefully reduces some of the hovering over tiles that is needed and skill gap required to play

## Changelog 

```changelog
(u)FlameArrow57
(*)As a blob, left clicking a tile will now attempt to spread to that tile, meaning you don't need to click or hotkey the ability button at all!
```

### DIFF
--- a/code/mob/living/intangible/blob_overmind.dm
+++ b/code/mob/living/intangible/blob_overmind.dm
@@ -310,7 +310,10 @@
 			src.update_buttons()
 			return
 		else
-			if(params["right"])
+			if(params["left"])
+				var/datum/blob_ability/spread_abil = src.get_ability(/datum/blob_ability/spread)
+				spread_abil?.onUse(T)
+			else if(params["right"])
 				if (T && (!isghostrestrictedz(T.z) || (isghostrestrictedz(T.z) && restricted_z_allowed(src, T)) || src.tutorial || (src.client && src.client.holder)))
 					if (src.tutorial)
 						if (!tutorial.PerformAction("clickmove", T))


TobleroneSwordfish (changes requested):
Good PR, but the way it's currently written prevents blobs from examining, it shouldn't override default alt examine behaviour unless an ability is bound to alt.

BatElite:
Didn't left click previously teleport you? Does that mean this affects the viewport thing blobs for some reason get? (I think you could like, click those for quick travel anyway...)

FlameArrow57:
> Good PR, but the way it's currently written prevents blobs from examining, it shouldn't override default alt examine behaviour unless an ability is bound to alt.

I don't think they're meant to, parent is never called to do so in the click() proc

FlameArrow57:
> Didn't left click previously teleport you? Does that mean this affects the viewport thing blobs for some reason get? (I think you could like, click those for quick travel anyway...)

nah it's right click that teleports you

TobleroneSwordfish (approved):
Pfff okay then!

MeggalBozale:
Sounds nice for new blobs and for blobs in general to free up a hotkey. On top of making it easier to do the most basic task :heart:

PR #19360, opened 2024-06-06
Skelegib visual improvements + torso drop 

CalliopeSoups:
[FEATURE] [BALANCE] 
## About the PR 

This PR makes the skeleton's body parts fly off in random directions on death, and the body will no longer disappear. This effectively allows for body reclamation, cloning and antag absorbing. I've renabled puritan as an option for skeleton. too.

## Why's this needed? 
Skeletons are tricky for some antags, they leave nothing that can be feasted on, drained or absorbed on death. This was discussed on the forums here (https://forum.ss13.co/showthread.php?tid=22763). 

I raised the trait cost to -4 to cover skeletons being clonable and puritan giving +2.

## Changelog 

```changelog
(u)CalliopeSoups
(+)Skelegib is visually improved, and now drops the skeleton's ribcage for antags to absorb. Skeleton costs -4 points now and you can take puritan with it again.
```


CalliopeSoups:
> I don't think skeletons should be able to be cloned or reassembled in any shape or form. This is their main drawback. This makes the PR a massive buff to an already very powerful mutantrace.
> 
> But besides thst, i like what the PR adds.

I see your point, I'm open to reverting that part of the PR but I'll wait on an admin opinion. I could keep the torso but not allow for changing mutantrace somehow to stop geneticists changing them into humans and cloning. 

I will argue that I think skeleton's main drawback is their limb removal in particular, people can get a leg or your head with a crowbar and wrench, I find that pretty good.

CalliopeSoups:
Keeping the rebuilding unless admin decision says otherwise, on the grounds of skeleton's limb replacement being a double edged sword. Raised the trait's cost to account.

JORJ949:
Does the head count as "dead" for the purposes of things like revs?

CalliopeSoups:
> Does the head count as "dead" for the purposes of things like revs?

I believe it should count as dead as the mind is being transferred out after death, but I don't think I can test it on a solo server. A testmerge might be good idea anyway, I feel like this must have created a bug _somewhere_.

Glamurio:
I'm confused where the "little skeleton in the head" part comes from. I think it's cool that skeletons live on in their head and can be reassembled, but why not just leave a brain in their skull? That way borging is still possible.

CalliopeSoups:
> I'm confused where the "little skeleton in the head" part comes from. I think it's cool that skeletons live on in their head and can be reassembled, but why not just leave a brain in their skull? That way borging is still possible.

To keep speaking after death I had to put the little skeleton in there so that your mind is in there and speaking from something. The head still contains the original brain, I'll have the skull drop it when you crush it cause I forgot to do that.

As for butchering, even if it drops a brain you shouldn't be able to butcher the tiny skeleton in the first place, it gets automatically gibbed a couple seconds after the skull is crushed. I might have to make that happen if the skull is removed in any way possible and not just crushing with an item, I can't think of a way to do get the skeleton out of the skull without destroying both though so I don't _think_ it's an issue.

CalliopeSoups:
I added a way to off yourself as an after-death head incase newbies get stuck. Known bug: sometimes when the little skeleton comes out the maptext appears above the head instead. Can't for the life of me figure out why.

TobleroneSwordfish (commented):
Spawning as a skeleton and pressing `kill` on the admin menu seems to get me stuck in an infinite loop of runtimes: 
image

CalliopeSoups:
> Spawning as a skeleton and pressing `kill` on the admin menu seems to get me stuck in an infinite loop of runtimes: image

Okay that is strange. I've been using the kill button this whole time and I'm not getting that, I just tested it again too.

flrsh:
I think when you actually kill someone, they should actually die. I like it being safe to assume this and do not want to add a bunch of edge cases and exceptions. (Just as a single example - with the admin kill / gib buttons, the intent is to actually kill people, and this complicates things and makes it not the case.) I also don't think a tiny skeleton critter living inside the actual skeleton's head is very fitting for a regular trait anyone can take. I think that aspect (but instead of upon "death", maybe they emerge once below a certain health threshold?) would be a lot more suited for a gimmick critter. 

Re: the issue with skeletons not being able to be absorbed / eaten, I personally think it would be fine to also drop their torso.

CalliopeSoups:
> I also don't think a tiny skeleton critter living inside the actual skeleton's head is very fitting for a regular trait anyone can take. I think that aspect (but instead of upon "death", maybe they emerge once below a certain health threshold?) would be a lot more suited for a gimmick critter.

Could you elaborate on what you mean by not fitting for a regular trait? I understand where you're coming from on the other points.

What I'm going to do is remove all visible aspects of the tiny skeleton and also make admin killing or gibbing kill it, I'm also going to make the player's last words correspond to what their head said. Then I'll work on the idea a bit, write up a design doc and try it again with a part 2 PR that will go all in on the concept so it fits in with game systems better. (For example, making the body flop unresponsive unless there's a pilot skeleton, then adding a pilot skeleton transplant surgery as mutantrace specific surgery sounds fun to me. I talked with people in the discord a little about replacing the brain entirely with the pilot skeleton, which I'll attempt but I have no expectations that it's possible.)

flrsh:
> Could you elaborate on what you mean by not fitting for a regular trait? I understand where you're coming from on the other points.

Yup. I think generally, with traits, or anything that's available to a regular player at roundstart, they should fall within some range of "normalcy", while leaving more special things to be obtained either in special ways (e.g. the game makes you an antag) or within the actual game. "There's a skeleton critter that controls the skeleton, and it pops out when the skeleton dies" is funny but also, I think, goes against important assumptions as "regular players are a single mob, when they are killed, they die". 

> What I'm going to do is separate the PR into two, this one for the skeleton gib behavior and another one for staying alive as the head past death, where I'll make sure admin kills kill the pilot skeleton too.

Separating the two sounds good to me, thank you! To be clear though, admin kills were just a single example, and there are a lot more. For instance - corpse spawners, what happens when you spawn a skeleton corpse and it's not actually dead cuz the skeleton critter pops out. Or changeling absorb code. There's already some weird edge cases with skeletons not dying when they're decapitated, and I don't really want to add a bunch more, at least to the default skeleton mutantrace. (Codewise, I think it's also not very good to subvert expectations and have "oh the die proc doesn't actually have them die".) Which is why I'm suggesting a gimmick critter (perhaps for wraiths?), cuz I think it's a lot more done there if some things are overlooked or if the code eventually breaks (because someone adds a new thing and forgets about this special handling) somewhere.

flrsh:
Re: the updated change - you can only clone torsos if you have the cloner upgrade or add back a head with the brain right? If so, letting skeletons take puritan and only increasing the skeleton cost by 1 trait point seems unbalanced.

CalliopeSoups:
> Re: the updated change - you can only clone torsos if you have the cloner upgrade or add back a head with the brain right? If so, letting skeletons take puritan and only increasing the skeleton cost by 1 trait point seems unbalanced.

You're able to turn them into a human with genetics and clone them that way, which is how it used to be before skeletons fell apart on death. You would need the head yes, I suppose there's situations where it could fly into a crusher, into space or a chute etc but I find that funny myself. 

Thanks for elaborating as well, I agree there's so many edge cases to cover it's probably not worth it. What I might do is add a similar effect to harbinger wraith's skeleton commanders instead, or the skeletons that they spawn.

Tarmunora (changes requested):
Strongly disagree with making skeletons able to be cloned/etc. again, and personally, exploding into a messy spew of limbs and leaving behind a body loses all the comedy factor of gently falling apart into a pile of bones.

While I agree that making skeletons able to be absorbed by the various absorption antags better would be a good change, this is a swing and a miss from me

PR #19364, opened 2024-06-06
Increase sale price of identified artifacts

pgmzeta:
[balance]
## About the PR 
Change the formula for calculating artifact prices by reducing the rarity modifier coefficient from 2 to 1.5. This increases the total sale price as it's raisng the power of a value less than one.

Comparison graph:
https://www.desmos.com/calculator/fe8s3bb7vf

## Why's this needed? 
Several people have lamented on artifact sale prices. Sucks to have cargo spend 8k on artifacts and potentially get back <1k on four correct artifact IDs.

## Changelog 

```changelog
(u)glowbold
(+)Artifacts sell for slightly more on the market.
```

### DIFF
--- a/code/modules/economy/shippingmarket.dm
+++ b/code/modules/economy/shippingmarket.dm
@@ -399,7 +399,7 @@
 
 
 	proc/calculate_artifact_price(var/modifier, var/correctness)
-		return ((modifier**2) * PAY_EMBEZZLED * correctness)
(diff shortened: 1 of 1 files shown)


Glamurio:
Honestly always bugged me that identified artifacts sell for more than you buy them, makes no sense. They should have more value because we know what they do. Could raise the sell price even more or reduce the buy price.

ZeWaka:
For now

PR #19374, opened 2024-06-06
Adds action bar to loosening/tightening player piano castors

AXLFAN2000:
## About the PR 
replaces do_after for anchoring/unanchoring player piano with an action bar that retains the functionality of disabling all mechcomp connections

## Why's this needed? 
roxy told me do_after is bad
I tested this and it seemed to work!
image

[player actions][qol]


AXLFAN2000:
I come bearing a message from aloe who does not have his github account right now:

> the action barrrr

PR #19391, opened 2024-06-08
Slightly more mail arrives when a mail courier is assigned to the station

pgmzeta:
[feature]
## About the PR 
Boost the crew mail target percenrtage value from 25% to 37.5% when the mail courier role is assigned.

## Why's this needed? 
Boost the amount of available mail for active mail couriers. I've noticed over several rounds couriers will show up asking for mail when none is ready; this should keep them appropriately busy wtihout increasing the mail load on QMs when couriers aren't around.

## Changelog 

```changelog
(u)glowbold
(+)More mail will arrive if your station has a dedicated mail courier.
```

### DIFF
--- a/code/modules/economy/shippingmarket.dm
+++ b/code/modules/economy/shippingmarket.dm
@@ -1,6 +1,6 @@
 #define SUPPLY_OPEN_TIME 1 SECOND //Time it takes to open supply door in seconds.
 #define SUPPLY_CLOSE_TIME 15 SECONDS //Time it takes to close supply door in seconds.
-
+#define SHIPPING_MARKET_SHIFT_BASELINE_DURATION 7.5 MINUTES //Baseline time between market shifts; 7:30 +/- 2:30 = 5 ~ 10 minutes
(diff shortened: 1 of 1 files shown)


Xkeeper0:
hey look a big green button i get to push

Xkeeper0:
gurgle. this actually seems to have broken and was causing an infinite loop of runtimes

```
[04:16:18] macros/debug.dm,6: weighted_pick given null weight: ["/obj/item/clothing/glasses/sunglasses/tanning"]
proc name: stack trace (/proc/stack_trace)
  source file: macros/debug.dm,6
  usr: (src)
  src: null
  call stack:
stack trace("weighted_pick given null weigh...")
weighted pick(/list (/list))
create random mail(the mail box (/obj/storage/crate/mail), 3475890)
/datum/shipping_market (/datum/shipping_market): generate mail(3511)
/datum/shipping_market (/datum/shipping_market): market shift()
```

reverted  for now since this seemed like the most likely culprit

PR #19398, opened 2024-06-08
Add escape pods to Clarion

Glamurio:
[Feature]

## About the PR 

Adds a new escape pod pay where the previous pod bay was.
image

Moves the old podbay in a maintenance section East of Ranch.
image

Adds `/obj/railing/guard` which basically just takes the sprite from `/obj/window/cubicle/railing` because it looked nice as a separator for the escape pods.

## Why's this needed? 
Parity with other maps, escape pod good.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)NT put aside some funding for an escape pod bay in Clarion.
```


Glamurio:
!merge_upstream

TTerc (approved):

PR #10871, opened 2022-09-18
Allow the Relay to convert a much larger part of the station

FlameArrow57:
[GAME MODES][FEATURE]

## About the PR 
This PR increases the Relay radial conversion to a radius of 100 tiles.

At the end of conversion process, placing the Relay at the middle of the map on Cogmap 2:
image

## Why's this needed? 
Thought it would be cool to make the radius larger.

PR #19399, opened 2024-06-08
Increase Relay conversion radius to 50 tiles

FlameArrow57:
[GAME MODES][EXPERIMENTAL][BALANCE]

## About the PR 
Increases Relay conversion radius to 50 tiles

Adds some extra code to make it only convert turf, instead of objects, and a safety check to make sure conversions aren't over queued

Originally from #10871 and #11825, probably needs test merge to make sure it doesn't break the server like how it did last time

## Why's this needed? 
Would be cool

Also, converting space tiles to flock lattices could help make space fights easier?

## Changelog 

```changelog
(u)FlameArrow57
(*)Increased Flock Relay conversion radius to 50 tiles
```

### DIFF
--- a/code/datums/flock/flock.dm
+++ b/code/datums/flock/flock.dm
@@ -1065,6 +1065,42 @@ proc/get_default_flock()
 			O.color = fancy_flock_matrix
 	return T
 
+/// only converts turf
+/proc/flock_convert_turf_turfonly(turf/T)
+	if(!T || !flockTurfAllowed(T))
+		return
+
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
Why would it only convert turfs though? Also I'm not too convinced the conversion proc will be that much faster considering it still contains several `locate`s and animation calls.

FlameArrow57:
> Why would it only convert turfs though?

More of a "conversion lite" since it's just from a broadcast. Also converting doors is more obstructive than necessary to getting to relay

PR #19405, opened 2024-06-08
Change FPRINT flag to NOFPRINT flag

TealSeer:
[INTERNAL] [CODE QUALITY]
## About the PR 

Refactor the `FPRINT` flag (enables fingerprint recording on an atom) to be `NOFPRINT` instead (disables fingerprint recording on an atom)

### To do

- [x] Figure out anything that definitely needs a `NOFPRINT` flag before this goes live (because looking at stuff that just doesn't have an `FPRINT` flag currently tells me nothing)

## Why's this needed? 

There's about a million and a half things that were missing this because even though it's set on `/atom` everyone overrides `flags` to slap more on and forgets to add `FPRINT`, in general a whitelist system for fingerprints is terrible because the vast majority of things in game should be recording fingerprints, and so using a blacklist makes a lot more sense.

## Changelog 

```changelog
(u) TealSeer
(*) The way fingerprints are left on objects has been changed under the hood, please report any weirdness or things having fingerprints that shouldn't!
```


frawhst:
Going to require secret changes for sure.

code/procs/basketball.dm
```diff
@@ -429,7 +429,7 @@
 	var/selfdestruct = 0
 	var/traitor_frequency = 0
 	var/obj/item/device/radio/origradio = null
-	flags = FPRINT | TABLEPASS| CONDUCT
+	flags = TABLEPASS| CONDUCT
```
frawhst, line 432:
```suggestion
	flags = TABLEPASS | CONDUCT
```

frawhst:
apparently if you apply 100 suggestions at once it leaves your PR in some hellish github limbo from which there is no escape

PR #19412, opened 2024-06-08
introduces calculate_depletion_rate and generalize most chem depletion

M-Earthfire:
[chemistry][balance][refactor]

## About the PR 

This PR splits off chem depletion calculation into the helper proc `/datum/reagent/proc/calculate_depletion_rate`. Calculations that are special to the mob in question were moved into `/mob/living/proc/get_chem_depletion_multiplier()`

This new helper proc was applied to most instances where chems deplete into other chemicals. The depletion amount was divided by their depletion rate to keep the amount the corresponding chems decay into constant.

## Why's this needed? 

Some chems take their depletion into account, some don't. This creates a whole lot of inconsistencies. Also, chems suddendly doubled in their ethanol/histamine/fat content once a spessman has slow metabolism, a missing liver or missing kidneys. This didn't made much sense and was counterintuitive, so now 1u of chem should on average decay into the same amount, whenever someone depletes them slower or not.

This will have balance implications with one of the meta-traits, slow metabolism. E.g. poisoning them with ethanol will become harder as well as using caffeine will be harder for them.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)The amount of chems a unit of a chemical decays into no longer changes with the speed the person metabolizes it.
```


code/modules/chemistry/Reagents-Drugs.dm
```diff
@@ -467,6 +467,21 @@ datum
...
+				var/caffeine_amt = holder.get_reagent_amount(src.id)
+				switch(caffeine_amt) //use ~midpoints for depeletion rate thresholds - need stronger coffees or blends to overcaffeinate
+					if(0 to 3)
+						. *= 0.5
+					if(10 to 30)
+						. *= 2
+					if(30 to 50)
+						. *= 4
+					if(50 to INFINITY)
+						. *= 5
```
mordent-goonstation, line 481: This doesn't seem consistent with previous behavior, i.e. these values are not just a multiple of the previous ones. What gives?

mordent-goonstation (commented):
Broadly I am confused by the seemingly random multipliers used along with `calculate_depletion_rate`. Should they not be the same multiple of what was there before? If not, why not? Example: 5 -> 8.4 in bath salts overdose usage, but 5 -> 12.5 for crank overdose usage?

code/modules/chemistry/Reagents-Drugs.dm
mordent-goonstation: I am not a fan at all of just missing the 3 to 10 value in the switch statement. Without context it almost feels like an oversight/error. Would much prefer the baseline was set around the first batch (0 to 3) then the others scaled accordingly (increasingly).

MintyPhresh:
I haven't done a full code review here, but Earthfire's methods are sane. The change essentially takes the numbers from being per-tick to per-volume, so crank's 5u per tick becomes 12.5 per unit

code/modules/chemistry/Reagents-Drugs.dm
MintyPhresh: Agreed with Mordent on this one, although it seems consistent with how the code was written previously it'd probably be better the other way

code/modules/chemistry/Reagents-PoisonEtc.dm
```diff
@@ -837,7 +837,7 @@ datum
 					boutput(M, SPAN_ALERT("<b>So itchy!</b>"))
 					random_brute_damage(M, 2 * mult)
 				if (prob(6))
-					M.reagents.add_reagent("histamine", rand(1,3) * mult)
+					M.reagents.add_reagent("histamine", randfloat(3.3 , 10) * mult)
```
Tarmunora, line 840: missing `calculate_depletion_rate`

code/modules/chemistry/Reagents-PoisonEtc.dm
```diff
@@ -1355,7 +1355,7 @@ datum
 
 				var/our_amt = holder.get_reagent_amount(src.id)
 				if (prob(25))
-					M.reagents.add_reagent("histamine", rand(5,10) * mult)
+					M.reagents.add_reagent("histamine", rand(25, 50) * mult)
```
Tarmunora, line 1358: missing `calculate_depletion_rate`

Tarmunora (changes requested):

PR #19423, opened 2024-06-09
The Authentication Disk can be used to alter the escape shuttle depature time

pgmzeta:
[rework][game objects][add to wiki]
## About the PR 
Move the early shuttle launch trigger from`emag_act` to `attackby` with the authentication disk as the trigger.

Add help text to the auth disk for its three current uses:
1. armed nuclear bomb detonation timer
2. armory emergency auth/unauth
3. escape shuttle launch timer

## Why's this needed? 
More value to the auth disk in non-nukie rounds. Require antags doing early shuttle launches to engage with at least one (1) other crew member (the captain). 

## Changelog 

```changelog
(u)glowbold
(*)The authentication disk can be used on the escape shuttle console to immediately launch it.
```

### DIFF
--- a/code/obj/machinery/computer/shuttle.dm
+++ b/code/obj/machinery/computer/shuttle.dm
@@ -319,6 +319,13 @@ var/bombini_saved
 		return
 
 	if (istype(W, /obj/item/disk/data/floppy/read_only/authentication))
+		if(emergency_shuttle.location != SHUTTLE_LOC_STATION)
+			return
+		for (var/datum/flock/flock in flocks)
+			if (flock.relay_in_progress)
+				boutput(user, SPAN_ALERT("[src] emits a pained burst of static, but nothing happens!"))
+				return
+
 		if (user)
 			var/choice = tgui_alert(user, "Would you like to launch the shuttle?", "Shuttle control", list("Launch", "Cancel"))
 			if(BOUNDS_DIST(user, src) > 0 || emergency_shuttle.location != SHUTTLE_LOC_STATION) return
--- a/code/obj/machinery/computer/shuttle.dm
+++ b/code/obj/machinery/computer/shuttle.dm
@@ -314,28 +314,21 @@ var/bombini_saved
 	icon = 'icons/obj/decoration.dmi'
 	icon_state = "syndiepc4"
 
-/obj/machinery/computer/shuttle/emag_act(var/mob/user, var/obj/item/card/emag/E)
-	if(emergency_shuttle.location != SHUTTLE_LOC_STATION)
-		return
-	for (var/datum/flock/flock in flocks)
-		if (flock.relay_in_progress)
-			boutput(user, SPAN_ALERT("[src] emits a pained burst of static, but nothing happens!"))
-			return
-
-	if (user)
-		var/choice = tgui_alert(user, "Would you like to launch the shuttle?", "Shuttle control", list("Launch", "Cancel"))
-		if(BOUNDS_DIST(user, src) > 0 || emergency_shuttle.location != SHUTTLE_LOC_STATION) return
-		if (choice == "Launch")
-			boutput(world, SPAN_NOTICE("<B>Alert: Shuttle launch time shortened to 10 seconds!</B>"))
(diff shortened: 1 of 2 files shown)


JORJ949:
I like this, but would rather see it just fully auth the shuttle and full authing removing a minute regardless of how much is left.

code/obj/machinery/computer/shuttle.dm
```diff
@@ -314,28 +314,28 @@ var/bombini_saved
...
-/obj/machinery/computer/shuttle/emag_act(var/mob/user, var/obj/item/card/emag/E)
-	if(emergency_shuttle.location != SHUTTLE_LOC_STATION)
+/obj/machinery/computer/shuttle/attackby(var/obj/item/W, var/mob/user)
+	if(status & (BROKEN|NOPOWER))
 		return
-	for (var/datum/flock/flock in flocks)
-		if (flock.relay_in_progress)
-			boutput(user, SPAN_ALERT("[src] emits a pained burst of static, but nothing happens!"))
+
+	if (istype(W, /obj/item/disk/data/floppy/read_only/authentication))
```
TobleroneSwordfish, line 321: Make this a guard (ie invert the condition and make it return instead of putting the entire proc in an `if` block) and have it call parent, otherwise attacking the shuttle computer normally won't work.

pgmzeta:
> I prefer George's idea too, not a fan of attaching antag behavior to non-inherently antag items. For a crew the only reason for fast shuttle would probably be blob and I don't think that warrants having such a behavior for non blob rounds. Might also encourage some meta of deleting auth disk if it's not nukies, captains already have enough reasons to be wanted by antags (aa/gun/bank records/etc).

I've changed the non-antag behavior to prompt for extending the shuttle time before departure by one minute

TobleroneSwordfish:
Things should not operate differently based on antag status, no.

code/obj/machinery/computer/shuttle.dm
pgmzeta: I don't quite follow your intent here - can you please say more about the changes you want to see and how'd they would affect the console?

TobleroneSwordfish: Sure, so currently you have the entire proc wrapped in this if statement to check if the item is an auth disk, and if it isn't then nothing happens. So if I hit the shuttle computer with a screwdriver or a fire extinguisher or something it won't defer to the parent `/obj/machinery/computer` behaviour, it'll just do nothing.

An easy way to fix this while also making the code cleaner would be to use [guard pattern](https://en.wikipedia.org/wiki/Guard_(computer_science)) and write something like
```
if (!istype(W, /obj/item/disk/data/floppy/read_only/authentication) || (status & (BROKEN|NOPOWER)))
	return ..()
```
and then leave the rest of the proc to continue below the "guard", having verified that there is power and the item is an auth disk.

pgmzeta: okay, I'll fix up the entire proc to use this pattern.

code/obj/machinery/computer/shuttle.dm
```diff
@@ -314,29 +314,28 @@ var/bombini_saved
...
 	icon_state = "syndiepc4"
 
-/obj/machinery/computer/shuttle/emag_act(var/mob/user, var/obj/item/card/emag/E)
-	if(emergency_shuttle.location != SHUTTLE_LOC_STATION)
-		return
-	for (var/datum/flock/flock in flocks)
-		if (flock.relay_in_progress)
-			boutput(user, SPAN_ALERT("[src] emits a pained burst of static, but nothing happens!"))
+/obj/machinery/computer/shuttle/attackby(var/obj/item/W, var/mob/user)
+	if(!(istype(W, /obj/item/disk/data/floppy/read_only/authentication) || istype(W, /obj/item/card/id)) && (status & (BROKEN|NOPOWER)))
```
TobleroneSwordfish, line 318:
```suggestion
	if(!(istype(W, /obj/item/disk/data/floppy/read_only/authentication) || istype(W, /obj/item/card/id)) || (status & (BROKEN|NOPOWER)))
```

PR #19432, opened 2024-06-09
Add system for automatically adding/removing mutantrace-associated abilities

TealSeer:
[MUTANTRACES] [CODE QUALITY] 
## About the PR 

Add two new variables to `/datum/mutantrace`, `mutant_abilityholder` and `mutant_abilities`. The former holds an abilityHolder type that will be automatically added when the mob gains the mutantrace, the latter holds a list of abilities to be added to the new abilityHolder. 

## Why's this needed? 

A couple different mutantraces implement abilities tied to the mutrace, and each one implements it separately which is no good. One way to do it means less duplicate code.

### DIFF
--- a/code/datums/mutantraces.dm
+++ b/code/datums/mutantraces.dm
@@ -858,10 +858,16 @@ TYPEINFO_NEW(/datum/mutantrace/lizard)
 
 	ghost_icon_state = "ghost-lizard"
 
+	mutant_abilityholder = /datum/abilityHolder/lizard
+	mutant_abilities = list(
+		/datum/targetable/lizardAbility/colorshift,
+		/datum/targetable/lizardAbility/colorchange,
+		/datum/targetable/lizardAbility/regrow_tail
+	)
+
 	on_attach(var/mob/living/carbon/human/H)
 		..()
 		if(ishuman(H))
-			H.give_lizard_powers()
 			H.AddComponent(/datum/component/consume/organpoints, /datum/abilityHolder/lizard)
 			H.AddComponent(/datum/component/consume/can_eat_inedible_organs)
 			H.mob_flags |= SHOULD_HAVE_A_TAIL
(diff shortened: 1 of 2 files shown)


ZeWaka (approved):

code/datums/mutantraces.dm
```diff
@@ -232,6 +232,11 @@ ABSTRACT_TYPE(/datum/mutantrace)
 
 	var/self_click_fluff //used when clicking self on help intent
 
+	/// Abilityholder associated with this mutantrace, will be automatically given to mobs on spawn
+	var/datum/abilityHolder/mutant_abilityholder = null
```
TobleroneSwordfish, line 236:
```suggestion
	/// Abilityholder associated with this mutantrace, will be automatically given to mobs on spawn
	var/mutant_abilityholder = null
```
Comment should make it clear this is a type, otherwise it looks like an instance var.

code/datums/mutantraces.dm
```diff
@@ -232,6 +232,11 @@ ABSTRACT_TYPE(/datum/mutantrace)
...
+	/// List of abilities associated with this mutantrace, requires mutant_abilityholder to be set
+	var/list/datum/targetable/mutant_abilities = list()
```
TobleroneSwordfish, line 238:
```suggestion
	var/list/mutant_abilities = list()
```
Ditto

TobleroneSwordfish (changes requested):

code/datums/mutantraces.dm
ZeWaka: why remove the type

TobleroneSwordfish: Because it's a lie!! When you see a var typed as `/datum/abilityHolder` you expect it to contain an instance of `/datum/abilityHolder` not a type. It's not like typing it offers any validation that the type you put in there will inherit from the correct parent.

PR #19439, opened 2024-06-10
Adds a new shotgun alongside the Salvager's reclaimed pump-action: The Four Letter Word

NightmareChamillian:
## About the PR 

**Note that that the magpie's trader list is secret content and so this PR only implements the weapon.**

This PR adds an alternative to the relcaimed shotgun. The demo video's a bit outdated, and firing in the 2- and 4-barrel mode is currently more punishing than shown.

image

https://youtu.be/PLY0vErHNDU

More than anything, my intent is to provide an in-theme sidegrade. Still, it's got its fair share of unique mechanics.

### So, how's it work?

Mechanically, the Four Letter Word relies on a "health" counter that decreases for every bullet fired. The intent here is to reinforce the the hit-and-run salvager gameplay loop while still being a deadly weapon in its own right.

The maximum condition starts at nine, and once the it dips below zero, the gun will be "severely damaged": making unpleasant noises, _very_ unpleasant chat messages, and rolling a 5*X percent chance to break after every additional shot. It'll always give at least _one_ warning, and generally speaking, tends to give in after 2-4 additional complaints.

If the user is negligent enough to let it break, a slamgun can still be recovered from the wrecked weapon, so they still have a gun they're carrying ammo for. If the user is negligent enough to let it break _and_ reckless enough to have it in the 4-barrel mode, the gun explodes in a similar fashion to the zipgun. 

The condition can be repaired with a weldingtool, which takes time and rapidly eats through fuel. It takes a full weldingtool to bring the condition to full from "crit", and even more if the weapon is below 0 uses.

The weapon's firing mode can be cycled with a screwdriver, switching between a one, two, and four barrel mode respectively. Note that this doesn't impact disarm-intent pointblanks, as doing so would be busted.

## Why's this needed? 

I've always really disliked the salvager's reclaimed shotgun. Given the other late-game salvager items, it's expensive, underwhelming, and never really fit in with the kludged-together theme as well as the other weapons did. Hopefully, this adds better and more varied combat to lategame salvager rounds.

## Changelog 

```changelog
(u)NightmarechaMillian
(*) An alternative shotgun has appeared in the M4GP13's barter list! Get 4 barrels of break-action fun for just 1200 points today!
```


Azrun:
.... What have you done. :stuck_out_tongue_winking_eye:

NightmareChamillian:
OH GOD WHAT HAVE I DONE

Azrun:
When I spawn one, why does it say it has 2 rounds?

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,136 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+		if((firemode == ALL_BARRELS) && (priorarmmo != 0))
+			sleep(0.3) //give it a micro-delay
+			if (src.canshoot(user))
+				boutput(user, SPAN_ALERT("The [src] kicks like a damn mule!"))
+				animate_shake(user, 2, 5, 5, 0, 0)
+				//this might seem punishing but keep in mind this is still FOUR whole shotgun shells at once.
+				user.do_disorient(stamina_damage = 40, knockdown = 0, stunned = 0, disorient = 20, remove_stamina_below_zero = 0)
+
+		//check the gun's condition, break as needed
+		if(guaranteed_uses < 0)
```
Azrun, line 2221: Why does this occur if no shots are fired?

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,136 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+				user.drop_item(src)
+				user.put_in_hand_or_drop(broken)
+				qdel(src)
+
+	attackby(obj/item/I, mob/user)
+		if (isweldingtool(I)) //are we repairing?
+			var/obj/item/weldingtool/welder = I
+			if(guaranteed_uses >= MAX_USES)
+				boutput(user, SPAN_NOTICE("The [src] doesn't seem to be all that damaged."))
+			else if (welder.try_weld(user, 2, 2))
```
Azrun, line 2243: > helping to reinforce the hit-and-run salvager gameplay loop

As implemented this seems to only be gated by the amount of welder fuel you have.  Just toss it in a backpack and spam your welder on it, and get it back out again.  I think I'd rather see this as an actionbar requiring full disengagement before being used again.

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,136 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+	inhand_image_icon = 'icons/mob/inhand/hand_guns.dmi' //gotta make it use gun inhands
+	item_state = "quadbarrel"
+	name = "Broken Four Letter Word"
+	desc = "This thing is TOTALED well beyond repair. You feel like you can recover a slamgun from it, though."
+	force = MELEE_DMG_RIFLE
+
+	attack_self(mob/user as mob)
+		boutput(user, SPAN_NOTICE("You pull some scraps off the [src]."))
+		user.drop_item(src)
+		var/obj/item/gun/kinetic/slamgun/newgun = new /obj/item/gun/kinetic/slamgun
```
Azrun, line 2273: Should this turn into more than one slamgun?

Azrun:
> My intent is to provide an in-theme sidegrade more than anything

What weapon is this intended to be in conversation with?

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,136 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 		else
 			boutput(user, SPAN_ALERT("You can't fire \the [src] when it is open!"))
 
+#define ONE_BARREL 1
+#define TWO_BARRELS 2
+#define ALL_BARRELS 4
```
Azrun, line 2149: 4 barrels seems insane.

This seems like it turns them into murder hobo loot pinatas.

NightmareChamillian:
> While i really like the concept, i am not sure how i fell about a reliable weapon being replaced with a non reliable and potentionally self-damaging one, especially for salvagers whose weapons are mediocre at their top end.

You do bring up a good point. After a discussion in the discord about the possibility of an upgrade kit, implementing it as a craftable item, or unlock, it was brought up that there's no reason both shotguns can't both exist at the same time.

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,216 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+		if((firemode == ALL_BARRELS) && (priorammo != 0))
+			sleep(0.3) //give it a micro-delay
+			if (src.canshoot(user))
+				boutput(user, SPAN_ALERT("The [src] kicks like a damn mule!"))
+				animate_shake(user, 2, 5, 5, 0, 0)
+				//this might seem punishing but keep in mind it's FOUR whole shotgun shells at once.
+				user.do_disorient(stamina_damage = 40, knockdown = 0, stunned = 0, disorient = 20, remove_stamina_below_zero = 0)
+
+		//check the gun's condition, break as needed
+		if(priorammo > 0) //make sure the gun isn't empty before we roll to break
```
Azrun, line 2223: Check occurs if shotgun is open.

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,216 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
 			boutput(user, SPAN_ALERT("You can't fire \the [src] when it is open!"))
 
+#define ONE_BARREL 1
+#define TWO_BARRELS 2
+#define ALL_BARRELS 4
+#define MAX_USES 11
+
+/obj/item/gun/kinetic/sawnoff/quadbarrel //for salvagers
+
+	name = "Four Letter Word"
```
mordent-goonstation, line 2154: Needs `\improper`

code/obj/item/gun/kinetic.dm
NightmareChamillian: Hopefully 9 safe uses+ longer repair times is better. Let me know if you think we should axe the 4-barrel mode entirely.

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,209 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+		else if (guaranteed_uses < 4) //0-4
+			. += " It seems pretty damaged."
+		else if(guaranteed_uses < 7) //4-7
+			. += " It's damaged."
+		else  //7+
+			. += " It's barely damaged."
+
+	shoot(var/target, var/start, var/mob/user)
+
+		var/i //Because we can't initialize variables when we're declaring the loop. thanks, DM.
```
Azrun, line 2205: You should be able to, otherwise `/atom/proc/update_filters()` and `/proc/list_to_assoc_list`  is 😭

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,209 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+		for(i=1, ((i <= priorammo) && (i <= firemode)), i++)
+			..() //shoot an additional ith time
+			guaranteed_uses-- //damage the gun an additional ith time
+
+		//you're shooting multiple shotgun shells out of a garbage gun at the same time. don't think there won't be consequences
+		if((firemode == TWO_BARRELS) && (priorammo != 0))
+			boutput(user, SPAN_ALERT("The [src] jumps in your hands!"))
+			animate_shake(user, 2, 3, 3, 0, 0)
+			user.do_disorient(stamina_damage = 20, knockdown = 0, stunned = 0, disorient = 5, remove_stamina_below_zero = 0)
+		if((firemode == ALL_BARRELS) && (priorammo != 0))
```
Azrun, line 2218: Should this occur iff there are four rounds shot?

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,209 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+		var/i //Because we can't initialize variables when we're declaring the loop. thanks, DM.
+		var/priorammo = src.ammo.amount_left
+
+		//Go up to the limit defined by the firing mode, but don't exceed however many bullets are in the gun BEFORE we started firing
+		for(i=1, ((i <= priorammo) && (i <= firemode)), i++)
+			..() //shoot an additional ith time
+			guaranteed_uses-- //damage the gun an additional ith time
+
+		//you're shooting multiple shotgun shells out of a garbage gun at the same time. don't think there won't be consequences
+		if((firemode == TWO_BARRELS) && (priorammo != 0))
```
Azrun, line 2214: Should this occur iff there are 2 rounds shot?

Azrun (changes requested):
In practice this doesn't feel that different from using a pump action.  At a certain skill level the act of pumping the shotgun isn't much of a limiting factor besides cognitive load and execution.

Shooting all 4 seems like it is almost wasting ammo with much larger risk.
Shooting 2 shots seems the sweet spot.
Shooting 1 shot seems like it is worse than pump action as this breaks.

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,209 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+	var/firemode = ONE_BARREL
+
+	force = MELEE_DMG_RIFLE
+	max_ammo_capacity = 4
+	two_handed = TRUE
+
+	New()
+		..()
+		ammo = new/obj/item/ammo/bullets/pipeshot/plasglass
+		set_current_projectile(new/datum/projectile/special/spreader/buckshot_burst/plasglass)
```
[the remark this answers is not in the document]
NightmareChamillian: No real reason other than glass was a bit shortrange and I didnt wanna give them 4 scrap rounds. I can switch to birdshot or empty if need be.

code/obj/item/gun/kinetic.dm
NightmareChamillian: It can be removed, but a well placed 2-round burst can do a flat 100 if every shot hits. Ideally, I want to discourage its use in melee for that reason.

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,209 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+		if((priorammo > 0) && !(src.broke_open)) //make sure the gun isn't empty and also closed (shooting conditions) before we roll to break
+			if(guaranteed_uses < 0)
+				//warn the user that they're in the danger zone
+				playsound(src.loc, 'sound/impact_sounds/Generic_Snap_1.ogg', 50, 1)
+				user.visible_message(SPAN_COMBAT("The [src] [pick(list("rattles!", "bulges!", "pops!", "thunks!", "jolts!", "makes a concerning click...", "cracks!"))]"))
+				if (prob(guaranteed_uses*-5)) //roll for failure. Since [uses] is now negative, we need another minus sign to cancel out
+
+					user.visible_message(SPAN_ALERT("[user]'s [src] makes a severe-sounding bang!"), SPAN_ALERT("The [src] gives out!"))
+
+					if(firemode == ALL_BARRELS) //ohhh, you REALLY fucked up now.
```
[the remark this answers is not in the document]
NightmareChamillian: Good point!

code/obj/item/gun/kinetic.dm
```diff
@@ -2144,6 +2144,212 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
...
+	var/priorammo
+
+	force = MELEE_DMG_RIFLE
+	max_ammo_capacity = 4
+	two_handed = TRUE
+
+	New()
+		..()
+		ammo = new/obj/item/ammo/bullets/pipeshot/glass
+		set_current_projectile(new/datum/projectile/special/spreader/buckshot_burst/glass)
```
Azrun, line 2175: Seems like this should start out empty or with a non-player created ammo. But that could just be a no ammo subtype.

Azrun (changes requested):

NightmareChamillian:
I'm currently where there's no Internet but I'll do the final review implementation ASAP

code/obj/item/gun/kinetic.dm
```diff
@@ -2156,6 +2156,7 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 	icon = 'icons/obj/large/64x32.dmi'
 	icon_state = "quadb"
 	item_state = "quadbarrel" //custom inhands, though.
+	default_magazine = /obj/item/ammo/bullets/a12/bird/seven
```
Azrun, line 2159: seven?

Azrun (changes requested):

Azrun (approved):

PR #6272, opened 2021-10-12
tgui-bench port

ZeWaka:
## About the PR 

ports https://github.com/tgstation/tgstation/pull/61081

## Why's this needed? 

performance testing good

not currently working - opens IE and runs server with html but tests don't occur even though js is loaded and fetched

PR #19440, opened 2024-06-10
Port tgstation/tgstation#61081 - continuation of zewaka's 2021 port

MeggalBozale:
## About the PR 
~~Ports the tgui benchmark stuff in a weird hacky way~~

~~Uses IEChooser.exe (under MIT license, see https://github.com/LesFerch/LaunchIE/blob/main/LICENSE for the license and https://github.com/LesFerch/LaunchIE/tree/main for the project itself~~

~~IEChooser is used to deal with actually loading IE and not Edge, this could potentially be done w/o the executable but I chose this since it's the easiest and works.~~

Was done properly now

- [x] Automatically connect the browser and the benchmarking code
- [x] Close the IE window when complete

Continues #6272

## Why's this needed? 
With heavier and heavier UIs being TGUId, performance concerns are affecting gameplay -- we desperately need a tool to use for pinpointing hot code to optimize so players can continue to enjoy the TGUI windows.


MeggalBozale:
I dont know if this is how I do it but i'm trying

MeggalBozale:
lot of this will be spend finding a replacement for LaunchIE.exe

MeggalBozale:
nvm that was easy

tgui/webpack.config.js
```diff
@@ -125,13 +128,13 @@ module.exports = (env = {}, argv) => {
     ],
   };
 
-  // Add a bundle analyzer to the plugins array
-  if (argv.analyze) {
-    const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
-    config.plugins = [
-      ...config.plugins,
-      new BundleAnalyzerPlugin(),
-    ];
```
ZeWaka, line 134: ?

ZeWaka: i feel like there was some warning in the console about this

MeggalBozale:
This needs a look at and is 'done'

PR #19447, opened 2024-06-11
Add a nuclear engine to Donut3

pgmzeta:
[feature][mapping][station systems]
## About the PR 
Add a 19x19 engine room prefab for donut3 with two options: singularity and nuclear.
Rework how random engine rooms are selected to work around duplicate prefab name issues.
Make a number of supporting changes to donut3:
* Engine computers now use landmark versions
* Donut3 Engineering gas storage now gets co2
  * The previously accessible air cans have been moved to engineering storage
* Disposals piping rerouted through gas storage to avoid the prefab area
* Emitters moved into to singularity prefab

## Why's this needed? 
More engine variety for one of the more popular maps.

## Screenshot
image

## Changelog 

```changelog
(u)glowbold
(*)The nuclear engine has a chance to appear instead of a singularity engine on donut3.
```


code/modules/worldgen/prefab/engines.dm
```diff
@@ -1,5 +1,17 @@
+#ifdef MAP_OVERRIDE_COGMAP1
```
pgmzeta, line 1: i am trading one hack for another here, but this should end up with only files relevant to the map being read/parsed.

RubberRats:
This is neat, but having played them I personally don't like the random engines very much. Not being able to expect a particular engine type takes out a lot of motivation for me to roll engineer. Furthermore, a lot of maps are designed with a specific engine setup in mind, making alternate configurations awkward. That might be a discussion for another time though.
That being said, I love the nuclear engine and think seeing it on more maps would be cool.

JORJ949:
I would honestly love to see random engines removed and replaced with you getting to choose the engine, Donut 3 is my favourite engine at the moment and I would honestly hate to see it be randomly nuclear.

AXLFAN2000:
Not a fan of random engines, but if they're the way things are going then this setup looks like a great example of one.  Very understandable layout

Sord213:
Not a fan of implementing random engines on more maps. I'd rather it stay as a unique (cool) thing for cog1.

pgmzeta:
> While I love random engines, I just don't like the nuclear engines. In practice, they don't usually leave much customization and experimentation and kind of just feel boring. If anything, I'd like this to also have a chance to add the TEG.

image
holding off on future work, but a TEG is just barely feasible. Would still need testing.

PR #19448, opened 2024-06-11
Creamer bottles contain cream and not milk

MintyPhresh:
## About the PR
Everyone chugs this shit with no side effects like FREAKS! well I say ENOUGH! let's give them erm.... fuckign heart disease??? (it actually only gives them a -10 stam debuff at these volumes)

## Why's this needed? 
creamer should be made of cream, I think (correct me if I'm wrong)


DisturbHerb:
Hi, creamer bottle creator here. These were PR'd with their contents being milk in mind. Thanks.

Glamurio:
Creamer can be both made of milk and / or cream, sometimes both, and sometimes it's just water and oils and emulsifier shit.

Maybe we can just find a happy middle ground, 50/50 milk and cream.

PR #19449, opened 2024-06-11
stop a proc from updating a mob's scream if there's one there

Skeletonman0:
[bug] [player actions]
## About the PR 

deletes the two lines that force reset scream and fart sounds to default in updatemob

this probably will break atleast something

i do not know the reasoning behind why the current check resets your sounds so often, but it doesnt play nice with things

## Why's this needed? 

screams reverting to default bad


frawhst:
Just remove those lines imo, they don't seem to do anything important.

Issue #19435, opened 2024-06-10
Mail makes a ball sound now


goonstation-issuebot:
### Labels

[BUG][TRIVIAL]

### Description

When thrown, Mail now makes the sound the various sport balls make on impact

### Steps to reproduce

1. Mail courier throws mail
2. it makes the sound a football would make

### Expected Behavior

mail making the usual throwing sound i guess?

### Additional Information & Screenshots

Reported by: croissantman1
Client version: 515.1636
On server: Goonstation 4 Roleplay: Sylvester
Active test merges: #18889, #19292, and #19117
Round log date: 2024-06-10-00-54
Reported on: 2024-06-10 01:18:04
Round ID: 181155
Map: KONDARU

Glamurio:
This isn't a bug.

Gannets:
Fixed by #19451

PR #19451, opened 2024-06-11
Remove basketball bounce sound from thrown mail packages

JustMyselfUwU:
[bug] [game objects]
## About the PR 
Removes basketball bounce sound from mail, as mail packages are not basketballs. It now uses the generic sounds.

mailnobbal

## Why's this needed? 
Fix for https://github.com/goonstation/goonstation/issues/19435
Thrown mail should not make a basketball noise, as it is not a basketball.

### DIFF
--- a/code/WorkInProgress/zamu_mail.dm
+++ b/code/WorkInProgress/zamu_mail.dm
@@ -84,7 +84,6 @@
 		// copied from basketballs, but without the stun if you get beaned.
 		..(hit_atom)
 		if(hit_atom)
-			playsound(src.loc, 'sound/items/bball_bounce.ogg', 65, 1)
 			if(ismob(hit_atom))
 				var/mob/M = hit_atom
 				if(ishuman(M))


Glamurio:
This isn't a bug, this is intended behavior and it's a fitting sound for the mail imo. I'm down to change it if we replace it with a better sound, but as it stands, I can't think of one. Many objects sound slightly off because not every PR stands to add a new sound.

pgmzeta:
I definitely think it needs a noticeable sound considering one of your arms is full of mail suddenly

AXLFAN2000:
The basketball sound is funny

MintyPhresh:
Joke suggestion, but we could make the mail bounce inhand like a basketball

TobleroneSwordfish (approved):
This is not intended behaviour, there's already a generic throw catch sound effect and the only reason the baskball one is used here is because the feature was wholesale copy-pasted from basketball code.

PR #19454, opened 2024-06-11
Add map to death events

ZeWaka:
## About the PR 

adds current map to death events

# Unsure if we want to implement like this or send the map id instead, wire input needed

## Why's this needed? 

important

https://github.com/goonstation/api/issues/30


Wirewraith:
Map information is already attached to events inserted into Goonhub via the round_id relationship

PR #19460, opened 2024-06-11
Hygiene behavior tweaks

MintyPhresh:
## About the PR

- You no longer need to take off clothes to restore hygiene with chem splashes (e.g. showers, baths, flopping in puddles). Instead, the maximum hygiene you can reach is limited by 4x the chemical protection (chemprot) of your equipped items.
--  Players are notified in chat if their clothes are preventing them from getting any cleaner.
--  A default Staff Assistant wearing a jumpsuit and shoes will have their maximum decreased by 60% (that's 15% * 4), so a shower can only get them up to 40% unless they take off clothes.
--  Chemprot also reduces the effectiveness of chemical splashes for both restoring and reducing hygiene.

- The stink-lines effect can be removed by getting above 15% hygiene, rather than 85%. This means that a clothed shower will still cure stinkiness. 
-- This means that a character will not emanate a "fetid stench" (or similar) at 80% hygiene because they were below 15% at some point.

## Why's this needed? 
We get a lot of Mhelps about bathing on RP, because the system isn't intuitive. Clothes permeable enough to let a blood puddle soil your skin should probably also be permeable enough that showers make you a little cleaner... especially if you're so fucking rancid that you make the room "smell like a corpse took a shit in here." (Jesus fuck, the language used in those notifications can get nasty...) This PR is designed to get new players clean, while guiding them along to intended gameplay paths to maximize cleanliness.

## Changelog
```changelog
(u)mintyphresh
(+)Showering with clothes on now restores the hygiene motive, with higher chemprot limiting the maximum cleanliness you can achieve.
```

### DIFF
--- a/code/datums/sims.dm
+++ b/code/datums/sims.dm
@@ -268,7 +268,7 @@
 			if (value < SIMS_HYGIENE_THRESHOLD_FILTHY && prob(33))
 				if (holder.owner.bioHolder && !(holder.owner.bioHolder.HasEffect("sims_stinky")) && !holder.owner.hasStatus("filthy"))
 					holder.owner.setStatus("filthy", 3 MINUTES)
-			else if ((value >= SIMS_HYGIENE_THRESHOLD_CLEAN ) && holder.owner.hasStatus("rancid"))
+			else if ((value >= SIMS_HYGIENE_THRESHOLD_FILTHY) && holder.owner.hasStatus("rancid"))
 				holder.owner.delStatus("rancid")
 			/*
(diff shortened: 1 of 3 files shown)


MintyPhresh:
> I would add an additional penalty for special type of gear like spacesuits and raise the bar for removing the stink lines accordingly.
> 
> People washing in their spacsuits/emergency space suits doesn't feel right for me.
looks like space suits (incl. emergency suits) have a base chemprot of 30, meaning that the maximum hygiene value it can reach is reduced by 120% if you're naked under it

PR #19479, opened 2024-06-12
Adds unallocated payroll funds

MintyPhresh:
## About the PR
- Instead of NT's 5-minutely payroll stipend providing only exactly what's needed to pay the crew, they now provide just a little extra; 110% of the money needed for wages instead of 100%. 
-- This means that every Staff Assistant will still be paid 150 per wage period, but the payroll stipend for them will be 165-- leaving 15 extra credits in the payroll budget per pay period.

- The payroll budget also starts out with an extra 1200 credits at roundstart.

## Why's this needed?
This extra bit of money makes it easier for Command to award bonuses and raises without having to take a hit to their own pays to do so. This money can also be embezzled by corrupt HoPs or devious traitors, which could lead to some interesting inter-command interactions.

## Changelog

```changelog
(u)mintyphresh
(+)NT's payroll stipends now deposit some excess money to the payroll budget.
```

### DIFF
--- a/code/datums/banking.dm
+++ b/code/datums/banking.dm
@@ -38,7 +38,7 @@
 		time_between_paydays = 5 MINUTES
 		time_between_lotto = 8 MINUTES
 
-		station_budget = 0
+		station_budget = PAY_IMPORTANT
 		shipping_budget = PAY_EXECUTIVE*5
 		research_budget = PAY_EXECUTIVE*10
 		total_stipend = station_budget + shipping_budget + research_budget
--- a/code/datums/datacore.dm
+++ b/code/datums/datacore.dm
@@ -253,7 +253,7 @@
 	src.medical.add_record(M)
 	src.security.add_record(S)
 	src.bank.add_record(B)
-	wagesystem.payroll_stipend += B["wage"]
+	wagesystem.payroll_stipend += B["wage"] * 1.1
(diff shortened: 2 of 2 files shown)


mordent-goonstation:
Doesn't this just mean someone can bump their funds without there being any real indication to others? Having to lower someone's pay/balance in order to increase someone else's feels like the appropriate level here, as that means that *someone* might notice the discrepancy.

MintyPhresh:
I can see your point there. From that angle, this PR makes giving people raises or embezzling money, etc a "victimless crime..." Until someone else looks at it.

However, with Budget Control access being Captain and HoP only, there probably won't be anyone else looking at it. That being said... with the current values this PR proposes, the extra cash it adds only accounts for 9% of the total station stipend. Anyone greedy enough to be willing to reduce others' pay probably won't be happy with just stealing a tenth of the pie, so I'm not sure it will de-emphasize the existing interactions with theft of payroll (though with only two people able to look at it, it won't exactly add much either).

I dunno, the main point of the PR was to move bonuses and pay raises away from selfless acts of kindness from Command, and more into how a real business would work. I'm certainly not directly cannibalizing my boss' paycheck every time I get a raise... Do you think I should reduce how much cash is added, from 10% to maybe 5%?

PR #19483, opened 2024-06-13
Allow directly swapping out a cyborg's head

BatElite:
[qol][feature]
## About the PR 
This PR allows roboticists to swap heads on a cyborg with an empty one they're holding. With a cyborg's wiring exposed (as for other part swaps), you can either use a wrench with the replacement head in your offhand or use the head directly on the borg. This swap automatically transfers the brain or interface over, so the cyborg's player doesn't lose control of their mob at any point in the process.

The held/replacement head needs to be empty for now, which is just to make it easy on myself. Given the persistence of mind transfer bugs, I'm not willing to mess with that stuff. It does mean that you can't swap out an empty-headed borg's head for one with a brain in it though, which I don't think is good.

## Why's this needed? 
Outfitting a roundstart borg with a screen head* currently involves complete dis- and reassembly. Given that the roboticist job has one of the higher knowledge floors for basic functioning, this should make what is one of the more fraught requests safer. Especially since I've seen borg players walk new roboticists through the steps to service them quite often.

(*Although I'd also love to see someone request a heavy head on their otherwise light body)

## Changelog 

```changelog
(u)BatElite
(+)You can now swap out heads on cyborgs by using an empty head on a cyborg with its wiring exposed, or using a wrench with the replacement in your other hand.
(+)The brain/interface inside the borg transfers to the new head automatically. This way you don't have to de-brain or take apart a cyborg just to change their head.
```

### DIFF
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -1463,19 +1463,6 @@
 					playsound(src, 'sound/items/Ratchet.ogg', 40, TRUE)
 					swap_heads(held_head, user)
 					return //swap_heads handles all the rest
-					//Borgs die to parts now the moment they cease having a head, so none of the code below is relevant anymore
-					/*
-					if(src.part_head)
-						boutput(user, SPAN_ALERT("[src] already has a head part."))
-						return
-					src.part_head = RP
-					if (src.part_head.brain)
-						if(src.part_head.brain.owner)
-							if(src.part_head.brain.owner.current)
-								src.gender = src.part_head.brain.owner.current.gender
-								if(src.part_head.brain.owner.current.client)
-									src.lastKnownIP = src.part_head.brain.owner.current.client.address
-							src.part_head.brain.owner.transfer_to(src)*/
 				if("l_arm")
 					if(src.part_arm_l)
 						boutput(user, SPAN_ALERT("[src] already has a left arm part."))
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -1340,6 +1340,8 @@
 				actions.Add("Remove Left Leg")
 			if (src.part_head)
 				actions.Add("Remove Head")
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
I think a better way to do this would be making it so removing the head reverts the borg back to a frame with all the other parts still attached, or even just leaves it as a headless borg body you can stick a head on to activate. This PR doesn't really make sense with the way it teleports the brain from one head to another.

BatElite:
The brain teleporting is a shorthand for multiple actions in the same way as "hit item with storage to put it in" and "hit rockbox with ore scoop to transfer ores" are. Reverting to frame is precisely the bit we're trying to avoid here.

Xkeeper0:
> I think a better way to do this would be making it so removing the head reverts the borg back to a frame with all the other parts still attached, or even just leaves it as a headless borg body you can stick a head on to activate. This PR doesn't really make sense with the way it teleports the brain from one head to another.

this still results in the borg dying during the switch. this head switch makes it so the roundstart ritual of screenheading every borg doesn't result in every borg becoming a ghost for 1-5 minutes and spoiling the round for everyone else

shoutouts to when a borg was getting changed, head pops off, and oh hi wraith

ZeWaka:
I think it'd be cool if borg parts could be swapped in a docking station safely instead of this tbh

Xkeeper0:
that's pretty different from this, since you can already swap everything other than a chest or head part outside of docking stations

BatElite:
Aside from that, I feel like putting that kind of thing on borg docks is only going to make roboticists have even less to do. About the only things the docks don't let borgs do is swap their cell out or replace parts. (Hell, I don't think most people even know you can manually swap out modules and upgrades too. I wish borgs let me do that more often.)

Xkeeper0:
are there any other outstanding issues, because if not this seems like a better solution to the current "ghost every cyborg for a while" thing

JORJ949:
If I had a nickel for every time I was debrained while explaining how to replace my head and left dead for 30 minutes, id have two nickels, would love this.

PR #19484, opened 2024-06-13
tgui: ie11 only + esbuild

ZeWaka:
## About the PR 

total ripoff of https://github.com/DaedalusDock/daedalusdock/pull/965
im even ripping off the pr text

## Why's this needed? 

For esbuild: Straight forward performance boost from replacing a JS-based minifier with a Go-based minifier.

For ie support: From what I understand they aren't even used by players who are on linux, it's now just VMs. Therefore, it's senseless to support these antiquated browsers. Each polyfill shaved off brings us closer to the mighty god-emperor.

perhaps this might address the weirdness with #19331, idk

## Changelog 

```changelog
(u)ZeWaka
(*)Internet Explorer 8-10 support has been dropped. Basically nobody should be affected by this.
```


ZeWaka:
it all works and stuff

ZeWaka:
https://github.com/tgstation/tgstation/pull/80310 also later

ZeWaka:
swc really wanted react and not inferno even with the plugin so thatll wait until react time

PR #19485, opened 2024-06-13
New traitor item: Marionette implant

Ilysen:
[Feature][Game Objects][Add to Wiki]

## About the PR 
this adds a new Syndicate item called the marionette implant. the details are a little complicated, so here's the rundown:
<details><summary>Item description</summary>

* Purchased with 1 TC by traitor uplinks only. The first purchase comes in a box that includes a remote (further info later in this list) as well as a sheet of paper with instructions on how to use it; subsequent purchases only provide extra implants.
* The niche of this item is to receive remote signals and convert them into minor disruptive actions for the implantee. They can be forced to say short phrases, emote specific things, drop what they're holding, and so on. A full list of actions is provided at the bottom of this list.
* The included remote control provides a tgui interface for tracking the status of all linked marionette implants, and can be used to send them commands.
* The backend of these implants is all built on wireless packets, with the remote just being a front-end to simplify the process of sending them. As such, they can be controlled using packet devices! Each implant has a unique passkey (named after a random alcoholic drink for fun thematics) that must be provided in a given wireless packet for it to work, adding an extra layer of complexity to account for.
* Implants have a **heat level,** which puts a soft cap on spamming them. Heat is increased by various actions, and dissipates slowly over time. Each time the implant is activated while its current heat is above 100, it has a 20% chance to burn out, permanently destroying it.
  * Heat is reduced by 1 every processing tick. Dissipation rate increases by 0.05 per tick, meaning that heat is reduced faster if the implant is given time to rest (to a maximum of 3 per tick).
* A current full list of actions is as follows:
  * `say` or `speak`: The implantee speaks a given phrase out loud. Max 45 characters. +15 heat.
  * `emote`: The implantee emotes using the provided key (such as `drool` or `gasp`). +15 heat. Certain emotes, like fainting and deathgasps, cannot be triggered.
  * `move`, `step`, or `bump`: The implantee steps in a given cardinal direction. +5 heat. Notably, this works on dead humans as long as they haven't decomposed.
  * `shock` or `zap`: The implantee loses 90 stamina and becomes disoriented for 3 seconds. +50 heat.
  * `drop` or `release`: The implantee drops whatever they're holding. +60 heat.
  * `use` or `activate`: The implantee uses whatever they're holding as though they'd clicked on it. +35 heat.
  * ***To prevent spam setups, all heat costs are imposed even if their conditions are not met. You gotta time it!***

I deliberately held off on making the effects too overtly strong or flashy, like making people attack others nearby -- I figured that mindhack implants can already accomplish stuff like that, as can stuff like madness toxin. The goal was to make the effects disruptive in ways that are minor on their own but can be very disruptive when timed correctly!
</details>

Necessary backend changes are as follows:
<details><summary>Backend changes</summary>

* Uplinks now keep track of how many times they've purchased a given entry using the new `purchase_log` associative list. This is used to allow marionette implants to provide a box on the first purchase, but not on subsequent ones.
* The error message for trying to emote using an invalid key now only appears if the emote is voluntary. This is to prevent spam in case of typos or errors when using the force emote function.
</details>

## Why's this needed? 
I thought it'd be cool to be able to have some kind of item to influence mobs by sending packets, and that gradually metamorphosed into this idea. Being able to sneakily implant someone and cause shenanigans with them at (in)opportune time seems like a really neat idea and opens the floor to creative subterfuge and interference if used well.

I'm aware that the ability to make people speak or emote certain phrases definitely has some abuse potential. While I've gone to great pains to make sure that everything is extensively logged, I'm absolutely okay with axing those functions if the likelihood for misuse is too great.

## Changelog 

```changelog
(u)Ilysen
(*)Added a new traitor item: the marionette implant. Check your uplink today, especially if you're a packet nerd.
```

## Media

Because of size constraints on uploading files to github, these are youtube videos. Sorry about that :P

[Basic showcase -- functions and remote](https://youtu.be/FRzTYjZxfzA) (no, I have no idea why shitty bill has a robot scream. I think that's a bug with the latest master)

[Packet control showcase using the PDA's packet sender](https://youtu.be/ME2_xcAdCWc)


Glamurio:
Very cool, my immediate question is; how does this interact with the mindshield implant? Does it ignore it, does it get hindered by it? Because otherwise we have similar issues that we had with the mindhacks, meaning that SecOffs or the HoS are the easiest targets, dropping their batons or guns when holding them, which you can bait them into.

Not for me to argue if it should or shouldn't go past mindshield. I think it's worth testmerging and seeing how disruptive it will be, but it's worth keeping in mind.

Ilysen:
at the moment, it doesn't have an interaction with mindshields; I figured that it's fairly hard to get it into 'em if they're being attentive, because unlike a mindhack, it doesn't immediately make them like- unable to retaliate. if you walk up and implant one, they're gonna ask what's up! I'm up for it going either way though depending on what the dev folks would prefer

Glamurio:
> at the moment, it doesn't have any interaction with mindshields, meaning it bypasses them; I figured that it's fairly hard to get it into 'em if they're being attentive, because unlike a mindhack, it doesn't immediately make them like- unable to retaliate. if you walk up and implant one, they're gonna ask what's up! I'm up for it going either way though depending on what the dev folks would prefer

How are they going to know you implanted them? It has no animation, no visual cues, no message in the log. If the HoS is talking to someone, especially on RP, they won't know if a staffie injects them when walking past. And unlike mindhacks, which knock you down and at least could give people a hint, this one leaves zero trace.

I don't think that's bad, in fact it probably should remain that way, because if people knew they got injected it gets into complicated territory of what prerogative a person has to go and get the implant removed; better they find out when they randomly say things. However, this is precisely why the mindshield exists, because it was too strong on authority figures.

Perhaps mindshield could interfere with it, not outright block it? Maybe it's like an emagged version that randomly does things other than what you intended, or it has a chance to be delayed. I'd hate immunity, and I'd hate for it to be harder to implant them; it kind of ruins the point. Food for thought.

Ilysen:
huh, do implants not have an animation? in my testing it did the little sprite wiggle when applying it to someone, even if it didn't have a chat message. plus your sprite turns to face 'em and such -- it's subtle, but it's there! in any case, i'll think on it when I sleep, and see if anyone else wants to toss their thoughts into the ring too

JORJ949:
> huh, do implants not have an animation? in my testing it did the little sprite wiggle when applying it to someone, even if it didn't have a chat message. plus your sprite turns to face 'em and such -- it's subtle, but it's there! in any case, i'll think on it when I sleep, and see if anyone else wants to toss their thoughts into the ring too

Implanting someone does make you sort of lean in their direction a bit (even if the implanter is empty) for a moment, implanting yourself has no such effect afaik.

Tarmunora (changes requested):
Only looking at the mechanical side of things, `emote` would need to either be significantly more expensive, or have a curated list, otherwise there's no reason to use `shock` or `drop` or whatever when you can `emote trip` for an instant 3s KO at will

Ilysen:
oh dang, good point. i didn't even think about that. thanks for letting me know, i'll see what i can do

JORJ949:
> Only looking at the mechanical side of things, `emote` would need to either be significantly more expensive, or have a curated list, otherwise there's no reason to use `shock` or `drop` or whatever when you can `emote trip` for an instant 3s KO at will

Also throw bible at them and force them to fart if they are stunned or a little slow to react instant kill for 1tc

code/obj/item/implant.dm
```diff
@@ -890,6 +896,366 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+	disposing()
+		processing_items.Remove(src)
+		..()
+
+	deactivate()
+		..()
+		processing_items.Remove(src)
+
+	process()
+		src.adjust_heat(-1)
```
TobleroneSwordfish, line 950: Shouldn't this use `heat_dissipation`?

code/obj/item/implant.dm
```diff
@@ -890,6 +896,366 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+	/// If TRUE, this implant is burned out and permanently unusable.
+	var/burned_out = FALSE
+	/// The implant's heat level, increased by various actions. Slowly reduces over time.
+	var/heat = 0
+	/// The implant's heat dissipation. `heat` is reduced by this value every processing tick.
+	/// The value slowly ramps up over time, but is reset upon being activated. This makes short-term overuse very punishing,
+	/// but allows it to recover decently quickly if given time to rest.
+	var/heat_dissipation = 1
+	/// If `heat` is above this value, each activation has a chance to break the implant permanently.
+	var/const/heat_danger_zone = 100
```
TobleroneSwordfish, line 922: Maybe include a warning prop in the response signal if an activation sends heat above the danger zone? (and a corresponding display of some kind on the remote UI) I get that it's meant to be somewhat opaque but having no indication of it is a little too punishing.

code/obj/item/implant.dm
```diff
@@ -890,6 +896,366 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+
+		var/command = signal.data["command"]
+		if (command == "ping")
+			var/datum/signal/ping_reply = get_free_signal()
+			ping_reply.source = src
+			ping_reply.data["device"] = "IMP_MARIONETTE"
+			ping_reply.data["sender"] = src.net_id
+			ping_reply.data["address_1"] = signal.data["sender"]
+			ping_reply.data["command"] = "ping_reply"
+			ping_reply.data["passkey"] = src.passkey
```
TobleroneSwordfish, line 985: Isn't the passkey meant to be a secret? If it responds with the passkey on ping then any random packetnerd could control it!

code/obj/item/implant.dm
```diff
@@ -890,6 +896,366 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+			if ("drop", "release")
+				if (!isdead(src.owner))
+					var/obj/item/I = src.owner.equipped()
+					if (istype(I))
+						logTheThing(LOG_COMBAT, src.owner, "was forced to drop \the [I] by \a [src] at [log_loc(src.owner)] (caused by [constructTarget(signal.author, "combat")] at [log_loc(signal.author)]).")
+						boutput(src.owner, SPAN_ALERT("Your grip on \the [I] suddenly relaxes!"))
+						H.drop_item()
+					else
+						fail_reason = MARIONETTE_IMPLANT_ERROR_INVALID
+					src.adjust_heat(60)
```
TobleroneSwordfish, line 1053: Since these values are so high that just using some of the functions 2 or 3 times could burn out the implant it would be nice if they showed up on the remote as well as the wall of text manual. Maybe just a "low/medium/high" heat warning when you have an action selected?

code/obj/item/implant.dm
```diff
@@ -890,6 +896,366 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+				var/step_dir = text2dir(uppertext(data))
+				if (step_dir && (step_dir in cardinal))
+					step(src.owner, step_dir)
+				else
+					fail_reason = MARIONETTE_IMPLANT_ERROR_INVALID
+				src.adjust_heat(5)
+			if ("shock", "zap")
+				logTheThing(LOG_COMBAT, src.owner, "was shocked by \a [src] at [log_loc(src.owner)] (caused by [constructTarget(signal.author, "combat")] at [log_loc(signal.author)]).")
+				boutput(src.owner, SPAN_ALERT("You feel a shock from inside your body!"))
+				src.owner.do_disorient(90, knockdown = 7 SECONDS, disorient = 3 SECONDS)
```
TobleroneSwordfish, line 1041: This doesn't seem to do what you think it does.

code/obj/item/implant.dm
```diff
@@ -890,6 +896,366 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+		src.burned_out = TRUE
+		src.deactivate()
+
+
+/obj/item/remote/marionette_implant
+	name = "marionette implant remote"
+	desc = "A remote control that allows the sending and receiving of data from linked marionette implants."
+	icon = 'icons/obj/porters.dmi'
+	icon_state = "remote"
+	item_state = "electronic"
```
TobleroneSwordfish, line 1114: This doesn't actually seem to have an inhand, but that might be for the best considering it's meant to be somewhat sneaky.

code/obj/item/implant.dm
```diff
@@ -890,6 +896,366 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+
+	flags = FPRINT | TGUI_INTERACTIVE
+	object_flags = NO_GHOSTCRITTER
+	w_class = W_CLASS_SMALL
+
+	HELP_MESSAGE_OVERRIDE({"Can track and control any number of marionette implants. To link an implant, simply use the remote on the implanter (or vice-versa)."})
```
TobleroneSwordfish, line 1120: Should probably work on the implant itself and maybe cases too?

code/obj/item/storage/implant.dm
```diff
@@ -28,3 +28,18 @@
...
+	name = "marionette implant kit"
+	desc = "Contains one marionette implant and the peripherals required to use it."
+	icon_state = "hard_case"
+	spawn_contents = list(/obj/item/remote/marionette_implant, /obj/item/paper/marionette_implant_readme)
+
+	make_my_stuff()
+		..()
+		// We do this because buying the box from the uplink will add the implanter from that to the box right afterwards
+		// This way, placed or spawned boxes will still have their implant, but the uplink one will not have a duplicate
+		SPAWN(1 SECOND)
```
TobleroneSwordfish, line 42: This probably doesn't need to be as long as a second, either try `SPAWN(0)` or just do it the non-cursed way and pass a flag into `New` that prevents adding the free implanter

TobleroneSwordfish (changes requested):
I really really like this, it has the good kind of old-school vibes.

code/obj/item/implant.dm
Ilysen: tf2-demoman

code/obj/item/implant.dm
Ilysen: the intent is that it chunks for 90 stamina damage, knocks down for 7 seconds if it drains stamina completely, and disorients for 3 seconds regardless. is that not what it's doin' here?

code/obj/item/implant.dm
Ilysen: that's a good point... I like the idea of having the ability to get the passkey from the ping, but I don't want it to be too easy to control. do you think that making a close proximity requirement (ala 3 tiles) to provide the passkey would fix that?

code/obj/item/implant.dm
TobleroneSwordfish: Seems to just be doing the stamina damage with no disorient

code/obj/item/implant.dm
TobleroneSwordfish: Well if you can get the passkey from examining it and you get it put in chat when you implant it then that's probably good enough on its own? You could also add it to the notes verb ie "You implanted a marionette implant with passkey 'xyz'"
What's the benefit to having it available from being pinged besides other people being able to take control of it?

Ilysen: honestly, I mostly just thought it'd be a neat facet to have. I also wanted it to be consistent with other packet-controlled systems, where you can run pings or other commands to receive the relevant access pass/key ID/etc. still, I suppose this is a bit of a different case since it's an antag item... I'll probably scrap it in light of that since, yeah, fair enough

code/obj/item/implant.dm
Ilysen: huh, weird. I recall it working during testing, and I can actually see the disorient effect working in the demo videos I recorded. new bug that got introduced or something? if I can repro it locally after updating to the latest master I'll comment here

Valtsu0:
Can this force admins do admin only emotes like birdwell? It probably should not. Also i think allowing monologue would be really funny

Ilysen:
good call! re monologuing, it *does* sound funny, but I worry it'd be super disruptive. i'm not averse to adding it if a dev thinks it'd be fine, but otherwise I'm inclined to keep it on the blacklist for sanity reasons

Ilysen:
made some changes based on the review -- here's a brief video showing the changes made to the UI and responsiveness:

https://github.com/goonstation/goonstation/assets/47678781/338be72d-62b2-45f4-b49d-69c5554496ef

code/obj/item/implant.dm
Ilysen: so I wasn't actually able to reproduce this on my end while on the devtest map -- the shock correctly caused the disorient effect when I wasn't being knocked down outright by stamina loss. under what conditions did it not work right for you? I'll try and repro them locally

Ilysen:
that shoooould be everything for this change batch. the only thing I'm confused on is the seeming tendency for the UI to slightly shift itself when hitting the Refresh button or when selecting from the N/E/S/W buttons. I have no idea why this happens :p

Ilysen:
known issue: spiefs get the box every time, instead of getting the implanters themselves on extra ones. this seems a bit complicated to resolve for a few reasons -- namely order of operations with the put_in_hands -- and I don't have the energy to resolve it today, but it *shouldn't* be a dealbreaker (because of how unlikely it is that an individual spief will get two of them) so I'm not gonna lose sleep over it unless we'd for sure like that fixed

Ilysen:
scratch that, spief spawning has issues in general I need to work out and that fix doesn't work for it. waugh. I'll have to tackle that tomorrow or so but in the meantime everything else should hopefully be fine

Ilysen:
that should do it! might not be the cleanest solution so, I'm open for replacements.

code/obj/item/implant.dm
```diff
@@ -890,6 +901,401 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
...
+		return ..()
+
+	implanted(mob/M, mob/I)
+		. = ..()
+		if (!src.burned_out)
+			// this is an anti-frustration feature with the goal of both making it easier to reference chat logs to know who is implanted with what,
+			// as well as to make sure players can still use packets if they forget to scan it onto their remote and didn't write down the network data
+			boutput(I, SPAN_NOTICE("You make a mental note that this implant's network ID is <b>[src.net_id]</b> and its passkey is <b>[src.passkey]</b>."))
+
+	receive_signal(datum/signal/signal)
```
Garash2k, line 974: It would be flavourful if the item used `netpass_syndicate`. I could see it function like airlocks, where an `access_code` is required on top of the `netpass_syndicate`. A balance dev would have to confirm if giving away the `netpass_syndicate` is okay or not though.

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
+import { useBackend } from '../backend';
```
Garash2k, line 1: The interface should probably in typescript. See intrfaces in their individual folders with `index.tsx` and `types` for a start. The Apc one goes pretty deep.

`selected_command` and `status` would get to be enums.

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
...
+import { Window } from '../layouts';
+import { Button, Flex, LabeledList, Section, Stack } from '../components';
+
+export const MarionetteRemote = (_props, context) => {
+  const { act, data } = useBackend(context);
+  const { entered_data, selected_command, implants } = data;
```
Garash2k, line 7: `MarionetteRemoteType.ts` would allow for typing `entered_data, selected_command, implants`

Also, here you wouldn't need to declare implants.

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
...
+    case "NO RESPONSE":
+      return "This implant is not responding to pings. It could have been destroyed, or it could just be far away.";
+    case "BURNED OUT":
+      return "This implant has been rendered permanently unusable by overuse and can be removed from the tracking list.";
+    default:
+      return "Unknown.";
+  }
+};
+
+const mapImplants = (act, entered_data, selected_command, implants) => {
```
Garash2k, line 125: `mapImplants` should be a function Component, letting you do `<MapImplants />` above. The context is passed so you can get your variables the same way with `useBackend`.

Actually I would suggest making an `<Implant implant={implant} />` component and having the `<LabeledList>`, the `map(...)` and the no implants check be in the parent component.

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
...
+  const { entered_data, selected_command, implants } = data;
+
+  return (
+    <Window
+      title="Marionette Remote"
+      width={410}
+      height={550}
+      theme="syndicate" >
+      <Window.Content scrollable>
+        <Stack vertical fill minHeight="1%" maxHeight="100%">
```
Garash2k, line 16: minHeight 1% seems hacky, consider putting comments explaining why it's necessary.

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
...
+export const MarionetteRemote = (_props, context) => {
+  const { act, data } = useBackend(context);
+  const { entered_data, selected_command, implants } = data;
+
+  return (
+    <Window
+      title="Marionette Remote"
+      width={410}
+      height={550}
+      theme="syndicate" >
```
Garash2k, line 14:
```suggestion
      theme="syndicate">
```

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
...
+      <Window.Content scrollable>
+        <Stack vertical fill minHeight="1%" maxHeight="100%">
+          <Section title="Controls">
+            <Flex direction="column">
+              <Flex.Item>
+                <LabeledList>
+                  <LabeledList.Item key="data" label="Data">
+                    {selected_command !== "step" ? (
+                      <Button
```
Garash2k, line 23: I would suggest using an Input with an `onChange` that continuously updates `entered_data`. This is what is done in other UIs. Multiple partially filled updates are ok, no need for a button that opens an input.

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
...
+                  <LabeledList.Item key="action_heat" label="Heat Per Action">
+                    {selected_command === "shock" || selected_command === "drop" ? "HIGH"
+                      : selected_command === "step" ? "LOW" : "MEDIUM"}
+                  </LabeledList.Item>
+                </LabeledList>
+              </Flex.Item>
+            </Flex>
+          </Section>
+          <Section title="Implants" buttons={(
+            <Button
```
Garash2k, line 92: Unless I'm mistaken, haven't looked to deep in the dm code, using timely `tgui_process.update_uis(src)`, there'd be no need for a refresh button.

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
...
+                      selected={selected_command === "use"}
+                    />
+                    <Button
+                      onClick={() => act("set_command", { new_command: "shock" })}
+                      content="Shock"
+                      selected={selected_command === "shock"}
+                    />
+                  </LabeledList.Item>
+                  <LabeledList.Item key="action_heat" label="Heat Per Action">
+                    {selected_command === "shock" || selected_command === "drop" ? "HIGH"
```
Garash2k, line 84: Spitballing ideas, maybe this could be done in a tooltip, or maybe red/yellow/green colors and a thunder icon would be nice here.

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
...
+                disabled={implant.status === "BURNED OUT"}
+              />
+              <Button
+                icon="envelope"
+                content="Activate"
+                onClick={() => act('activate', { address: implant.address, packet_data: entered_data, packet_command: selected_command })}
+                disabled={implant.status === "BURNED OUT"}
+              />
+              <Button.Confirm
+                icon="x"
```
Garash2k, line 153: A tooltip would be useful here. What does this do, destory the implant? unlink it?

tgui/packages/tgui/interfaces/MarionetteRemote.js
```diff
@@ -0,0 +1,163 @@
...
+                onClick={() => act('remove_from_list', { address: implant.address })}
+              />
+            </>
+          )}>
+          {implant.status}
```
Garash2k, line 158: Some colors on the ACTIVE/IDLE/... label would be neat I think

Garash2k (commented):
This looks super cool, here are some thoughts

tgui/packages/tgui/interfaces/MarionetteRemote.js
Garash2k: Once that is done, if you want to go for top notch organization, I would suggest moving the Controls and Implants sections to respective `Controls.tsx` and `Implants.tsx` files.

Garash2k: Once that is done, if you want to go for top notch organization, I would suggest moving the Controls and Implants sections to respective `Controls.tsx` and `Implants.tsx` files.

Tarmunora (approved):

tgui/packages/tgui/interfaces/MarionetteRemote.js
TobleroneSwordfish: JS interfaces are fine, don't feel obligated to translate it to typescript.

code/obj/item/implant.dm
TobleroneSwordfish: Seems to work now, not sure if you changed something or I was just being dumb!

tgui/packages/tgui/interfaces/MarionetteRemote.js
TobleroneSwordfish: I like it as a separate readout personally, fits the clunky remote feeling

tgui/packages/tgui/interfaces/MarionetteRemote.js
Ilysen: thank you for that. while I do want to learn typescript, I think I'll stick with the js one for now just because it's what I'm familiar with and it might take me a while to learn how to make a typescript port work. a tsx version could always be made later, so it's hardly a permanent loss

Ilysen:
made a few changes based on the review, mostly for cleanliness stuff. I kept a few of the things as they are, though I can put in some other changes upon dev request

Garash2k (approved):

PR #19486, opened 2024-06-13
TGUI Alert RP Rules

Garash2k:
## About the PR 
Replaces the raw html RP Rules window with a TGUI one.
image

## Why's this needed? 
We can make it look better than the old one, and indented lines in a .tsx file are more easily maintained than the chunk of html it used to be. Previous UI:
261895715-9908d5db-0ab5-4503-91cd-9579f1cfa43f

[UI][Feature]


BatElite:
Where'd the line with the station name go?

I think you should also bring back the indentation for the main body of each rule's text. The old thing looked nicer that way.

Garash2k:
> Where'd the line with the station name go?

Being able to pass parameters to this window would be a coding adventure, if the "Welcome to Goonstation" I put is good enough, I'd be happy to get away with that.

> I think you should also bring back the indentation for the main body of each rule's text. The old thing looked nicer that way.

As for indentation, there's pros and cons. Indentation is standard for a number/bullet point list, but theses are titled sections. Also, it would make the window longer to scroll. I'd like to see what others think.

ZeWaka:
I like this and I'll merge it soon

PR #19491, opened 2024-06-13
Fakeobject refactor

TheTyrant25:
[CODE QUALITY]
## About the PR
- All instances of `/obj/decal/fakeobjects` have been repathed to `/obj/fakeobject`.
- All the inherited behaviour from decal has been put directly into `/obj/fakeobject` in code/obj/decal.dm, which means no behaviour should change.
- All the `/obj/fakeobject` have also been moved to a new file; `fakeobject.dm`, in the same folder as `decal.dm`.
- Other than that file, every other file has been edited using visual studio code find-replace.
- Also adds a script for updatepaths 

## Why's this needed?
Fake objects aren't really decals if you think about it. It being plural is also a bit strange, and lead to an instance of `obj/decal/fakeobject/something` being made by mistake. This should unify it a bit and distinguishes fake objects more. Which is good, since `fakeobject`s are used a lot in azones, they should probably be higher up in the inheritance tree (is that the right term?)


TheTyrant25:
size check still failing even after updating :(

TheTyrant25:
according to the makers of the check this is a known issue

PR #19495, opened 2024-06-13
make tomato splats use create_reagents

M-Earthfire:
[hydroponics]bug]

## About the PR 

Makes the creation of tomato splats properly use create_reagents instead of creating a improperly set up reagent container in them

## Why's this needed? 

someone in the discord found a bug where tomato splats of tomatoes filled with black powder were unable to be ignited. this was because the reagent container of the tomato splat was improperly configurated and hadn't `.my_atom` set properly. This fixes this and probably other inconsistencies that this would cause.


ZeWaka (commented):
10,000????????

PR #19496, opened 2024-06-13
Skeletons drop torsos on death and their limbs fly off 

CalliopeSoups:
[RESPAWNING]
## About the PR 
Remake of a previous PR, without the additional fluff, that causes a skeleton's limbs, head and torso to fly off in random directions a couple tiles on death. A skeleton can be rebuilt after, but attempting to clone them will still result in a puritan explosion.

https://github.com/goonstation/goonstation/assets/102558507/cc443f8c-4f3f-44d0-b228-7bdd0d350d4a

## Why's this needed? 
Skeletons don't drop their torso on death, which means that antags have to keep them alive while they absorb, eat or drink from them. This was discussed on the forums here (https://forum.ss13.co/showthread.php?tid=22763).

## Changelog 

```changelog
(u)CalliopeSoups
(+)Skeletons drop their torso on death as a tasty snack for antags.
```

### DIFF
--- a/code/datums/mutantraces.dm
+++ b/code/datums/mutantraces.dm
@@ -1162,34 +1162,29 @@ TYPEINFO(/datum/mutantrace/skeleton)
 			var/obj/itemspecialeffect/poof/P = new /obj/itemspecialeffect/poof
 			P.setup(src.mob.loc)
 			var/obj/item/I
-			//this should be done in like, a loop but by god the only way to do that is with vars[]
-			I = src.mob.limbs.l_arm?.remove(FALSE)
-			I?.pixel_x += rand(-8, 8)
-			I?.pixel_y += rand(-8, 8)
-			I = src.mob.limbs.r_arm?.remove(FALSE)
-			I?.pixel_x += rand(-8, 8)
-			I?.pixel_y += rand(-8, 8)
-			I = src.mob.limbs.l_leg?.remove(FALSE)
-			I?.pixel_x += rand(-8, 8)
-			I?.pixel_y += rand(-8, 8)
-			I = src.mob.limbs.r_leg?.remove(FALSE)
-			I?.pixel_x += rand(-8, 8)
-			I?.pixel_y += rand(-8, 8)
-			I = src.mob.organHolder.drop_organ("head")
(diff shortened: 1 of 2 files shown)


Tarmunora (approved):
still feel like the throw is a bit unnecessary, but this is acceptable enough

CalliopeSoups:
> still feel like the throw is a bit unnecessary, but this is acceptable enough

I get what you mean about it being funny that the bones fall gently. This way at least the limbs only get thrown 2 to 4 tiles, the last PR used sever() which would throw them frankly ridiculous distances. I can reduce that even more and have them only throw 1 tile out in random directions if you think that would be any better?

Glamurio:
Ideally the throw force would depend on how they die. If a skeleton dies by just... suffocating, they should just crumble, but if it does to brute damage or being smacked too hard, it should probably be thrown around. Otherwise 1 tile radius seems the best in-between.

CalliopeSoups:
The skeleton's torso and limbs will only throw 1 tile off now, I think this strikes a balance between subtlety and having the effect be visible, as my main gripe was that the bones and limbs fall underneath your items. 

However, there's this freaky thing ThrowRandom does where it sends the torso and limbs in the same direction every time, without fail. I discussed this at length in #imcoder and decided to ignore it, because it scares me. The important thing is you see the bones n stuff anyhow.

TobleroneSwordfish:
Couple minor nitpicks:
- Why does the torso get thrown too? Isn't the point that all the bits are flying off it?
- Why are there still random chunks of bone spewing out if we're throwing the actual limbs and parts the skeleton is made of instead?

CalliopeSoups:
> Couple minor nitpicks:
> 
>     * Why does the torso get thrown too? Isn't the point that all the bits are flying off it?
> 
>     * Why are there still random chunks of bone spewing out if we're throwing the actual limbs and parts the skeleton is made of instead?

I threw the torso thinking that a newer antag might not see it along with the bones and assume they have nothing to absorb, but I admit it probably makes more sense otherwise, I'll revert that.

As for the bones, I'm not sure? They were there with the limbs already before I changed things and adds to the effect I think.

PR #19505, opened 2024-06-14
Disposal Chute Resprite

TemThrush:
[Game Objects]

## About the PR 
This PR resprites the disposal chute and all of its variants (also the outlet) and gives the mail-chute a more unique appearance to match with the qm mail machine.

image

https://github.com/goonstation/goonstation/assets/108035657/55223c18-d168-415f-afb9-21afd6a62c92

https://github.com/goonstation/goonstation/assets/108035657/46b6d391-074c-4311-91f8-097d97e01002

https://github.com/goonstation/goonstation/assets/108035657/a68c9750-0286-4133-81f1-9050cf8c20c2

## Why's this needed? 
The original sprites have been around for a very long time, but no longer meet the current style guidelines. They could use a bit of an update!
The flushing animation for several of the special job-related chutes were mislabeled as well, and didn't play when flushed, that has also been fixed in this PR.


TobleroneSwordfish:
These look great, there's just one issue I can see:

image
image

I get that the new sprites are meant to be mounted *on* the wall, but most of the chutes in-game aren't offset like that, so they end up floating. Since a lot of maps don't seem to use the defined subtypes you could either go through and pixel offset them all in the map editor (blegh) or write code to automatically pixel offset them on spawn based on direction, checking that they don't already have pixel offsets defined (kind of cursed but probably better?)

TemThrush:
> These look great, there's just one issue I can see:
> 
> I get that the new sprites are meant to be mounted _on_ the wall, but most of the chutes in-game aren't offset like that, so they end up floating. Since a lot of maps don't seem to use the defined subtypes you could either go through and pixel offset them all in the map editor (blegh) or write code to automatically pixel offset them on spawn based on direction, checking that they don't already have pixel offsets defined (kind of cursed but probably better?)

Honestly, I was planning on just manually going through and adjusting them on every map (EW) in a separate pr alongside this one, but the code makes this way easier lmao (thank you YellowLemons again 🙏) this hopefully solves that issue!

TobleroneSwordfish (approved):

TemThrush:
> North and South offsetting?

North and south directional sprites shouldn't need any offsetting with how they're currently placed on most maps afaik. Unless i'm misunderstanding something 😓

PR #19508, opened 2024-06-14
Fix (seemingly) forced middle names in character creation

Glamurio:
[Bug] [QoL]

## About the PR 
Changes middle names to use `allowEmpty = TRUE`, changes any null names to `""`.

## Why's this needed? 
It was weird how the game made it seem like you needed a middle name, but actually supports a lack of middle names just fine. This is a streamlining of the feature, because you can already force a `None` middle name by simply picking a character found in `bad_name_characters` such as `_` and making this your middle name, but that's clearly an exploit.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)You can now choose not to have a middle name in character creation by leaving it empty.
```

### DIFF
--- a/code/datums/preferences/preferences.dm
+++ b/code/datums/preferences/preferences.dm
@@ -443,9 +443,9 @@ var/list/removed_jobs = list(
 					return TRUE
 
 			if ("update-nameMiddle")
-				var/new_name = tgui_input_text(usr, "Please select a middle name:", "Character Generation", src.name_middle)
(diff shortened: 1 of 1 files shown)


soleilan:
Yeah before the tgui port of Character Setup you could leave your middle name blank and it would set it to None. I can load my old character with None middle names fine, I noticed this and was going to PR it myself, so good job! 👍

TobleroneSwordfish (approved):

PR #19509, opened 2024-06-14
Spruce up the Inspector

JORJ949:
## About the PR 
The inspector now spawns with the following:
>NT coloured ID and PDA
image

>"Inspector's" gloves and boots, previously exclusive to the detective medal reward
image

>An Inspector's headset, making it clearer over the radio when the Inspector is mad at you
image

## Why's this needed? 
As an NT official they should get the NT ID colour and not just the basic one
The Inspector isn't really a "head of staff" so their command headset didn't match

## Changelog 

```changelog
(u)JORJ949
(+)As an NT official the Inspector now gets an NT ID and their own headset.
```


JORJ949:
Branched from the wrong branch :/

PR #19510, opened 2024-06-14
Inspector NT ID and Unique Headset

JORJ949:
## About the PR 
The inspector now spawns with the following:
>NT coloured ID and PDA
image

>An Inspector's headset, making it clearer over the radio when the Inspector is mad at you
image

## Why's this needed? 
As an NT official they should get the NT ID colour and not just the basic one
The Inspector isn't really a "head of staff" so their command headset didn't match

## Changelog 

```changelog
(u)JORJ949
(+)As an NT official the Inspector now gets an NT ID and their own headset.
```


DimWhat:
The meshing of two different palletes of black kinda looks strange by default.

JORJ949:
> The meshing of two different palletes of black kinda looks strange by default.

Now that you mention it, I agree, but found it difficult to notice in-game

DimWhat:
I'm not sure copying the full det outfit is the right choice either? Idk, i made most of that gear without really considering the a actual inspector role. This makes the detective medal have a bit of a confusing position as it would just let you near completely copy the outfit of the inspector. Which goes against the whole "making them stand out" thing.

JORJ949:
> I'm not sure copying the full det outfit is the right choice either? Idk, i made most of that gear without really considering the a actual inspector role. This makes the detective medal have a bit of a confusing position as it would just let you near completely copy the outfit of the inspector. Which goes against the whole "making them stand out" thing.

What do you think would fit best? Leaving the clothing as-is?

JORJ949:
Reverted outfit changes as per suggestion

PR #19512, opened 2024-06-14
Syndicate Clothing Vendor

JORJ949:
## About the PR 
Replaces gear closets from the listening post with a Syndicate Clothing vendor
image

C.A.R.L. no longer sells syndie gas masks, these have been moved to the vendor, and instead sells clothing restock cartridges for the syndicate clothing vendor.

## Why's this needed? 
Currently whoever buys the agent card first gets all the listening post clothing, this allows multiple traitors/sleeper agents to get the syndicate outfit.

## Changelog 

```changelog
(u)JORJ949
(*)Listening post gear closets replaced with clothing vendor, restocks are purchasable from C.A.R.L.
```


JORJ949:
> This implementation does mean that just about anybody with a way inside the outpost can get a syndicate headset. It's probably fine though.

I realised but I realised that's probably _fine_ as going to post as non-antag is almost always metagaming, if it becomes an issue it can be put back in the merchant I have no issue with that I just moved it there for consistency with other vendors.

code/modules/economy/commodity.dm
```diff
@@ -573,6 +573,12 @@
 ///// syndicate trader /////////
 ////////////////////////////////
 
+/datum/commodity/clothing_restock
+	comname = "Clothing Vendor Restock Cartridge"
+	desc = "A restock cartridge for restocking clothing vending machines."
+	comtype = /obj/item/vending/restock_cartridge/jobclothing
+	price = PAY_IMPORTANT
```
TobleroneSwordfish, line 580: This should just be for the syndie vendor if it's specifically sold by CARL (also because it makes no sense for it to be able to restock any arbitrary clothing vendor)

code/modules/vending/jobclothing.dm
```diff
@@ -242,3 +242,30 @@ ABSTRACT_TYPE(/obj/machinery/vending/jobclothing)
...
+	req_access = list()
+
+	create_products(restocked)
+		..()
+		product_list += new/datum/data/vending_product(/obj/item/clothing/under/misc/syndicate, 4)
+		product_list += new/datum/data/vending_product(/obj/item/clothing/head/helmet/space/syndicate, 2)
+		product_list += new/datum/data/vending_product(/obj/item/clothing/suit/space/syndicate, 2)
+		product_list += new/datum/data/vending_product(/obj/item/tank/jetpack/syndicate, 2)
+		product_list += new/datum/data/vending_product(/obj/item/clothing/mask/gas/swat, 2, cost=PAY_IMPORTANT)
+		product_list += new/datum/data/vending_product(/obj/item/device/radio/headset/syndicate, 2, cost=PAY_IMPORTANT/2)
```
TobleroneSwordfish, line 265: God no, keep these locked behind a syndie ID or they'll just be compromised constantly.

TobleroneSwordfish (changes requested):

code/modules/vending/jobclothing.dm
```diff
@@ -242,3 +242,29 @@ ABSTRACT_TYPE(/obj/machinery/vending/jobclothing)
...
+		product_list += new/datum/data/vending_product(/obj/item/clothing/head/helmet/space/syndicate, 2)
+		product_list += new/datum/data/vending_product(/obj/item/clothing/suit/space/syndicate, 2)
+		product_list += new/datum/data/vending_product(/obj/item/tank/jetpack/syndicate, 2)
+		product_list += new/datum/data/vending_product(/obj/item/clothing/mask/gas/swat, 2, cost=PAY_IMPORTANT)
+		product_list += new/datum/data/vending_product(/obj/item/clothing/shoes/swat, 4)
+		product_list += new/datum/data/vending_product(/obj/item/storage/backpack/syndie, 1)
+		product_list += new/datum/data/vending_product(/obj/item/storage/backpack/satchel/syndie, 1)
+		product_list += new/datum/data/vending_product(/obj/item/storage/fanny/syndie, 1)
+
+		product_list += new/datum/data/vending_product(/obj/item/clothing/head/helmet/space/syndicate/commissar_cap, 1, hidden=1)
```
TobleroneSwordfish, line 270:
```suggestion
```
Oh uh, I just noticed this was added - this is a nukie/podwars exclusive item that shouldn't appear in normal rounds. The stats on it are kind of crazy and it just doesn't fit outside of special round types.

TobleroneSwordfish (changes requested):

code/modules/vending/jobclothing.dm
JORJ949: The stats on it are the same as the space helm except for 10% explosion resist but I can remove it 👍

JORJ949: image
Was meant as a fun syndicate headpiece so its fine to remove

PR #19513, opened 2024-06-14
fix up the VR racing gokarts

Skeletonman0:
[objects] [qol]
## About the PR 

- make vr kart abilities usable again
- nukes the hell out of /obj/powerup, replacing it with /datum/targetable/kart_racing versions
- gives vr carts _atleast some_ clickdrag into and click to exit. i can already see this part of the PR needing changes
- change up kart naming to use suffixes with the name of the owner to hopefully break less things
- instead of setting overlays to null, the affected vehicles use AddOverlays
- fix the sprite being ontop of the vehicle, the cover/suit sprite was changed during this as it did not fit the human anymore
image

## Why's this needed? 

current go-karts use old /obj/powerup which just slaps itself on your screen **_under_** the exit vr ability button

they are also unable to be climbed in with clickdrag, which is unintuitive

this should make using the VR race track more practical again.

## Changelog 

```changelog
(u)Skeletonman0
(+)The VR Go-Karts have been serviced and are ready for their next race.
```


code/WorkInProgress/racing.dm
```diff
@@ -191,158 +191,145 @@
...
+		var/turf/T = get_turf(get_turf(racer))
 		new/obj/racing_trap_banana/(T)
 
 		playsound(T, 'sound/mksounds/itemdrop.ogg', 45, FALSE)
-
+		// remove ourselves when we're used
+		racer.removeAbility(src)
+		holder.abilities -= src
+		racer = null
 		qdel(src)
```
TobleroneSwordfish, line 241:
```suggestion
		racer.removeAbilityInstance(src)
```
This is the correct proc, the `qdel` and `abilities -=` are handled by the ability holder and the `racer = null` should be in the parent `disposing`. Same goes for all other instances of this snippet.

code/WorkInProgress/racing.dm
```diff
@@ -413,10 +418,13 @@
 
 		driver.set_loc(get_turf(src))
 
-		if(powerup && (powerup in driver.client.screen))
-			driver.client.screen -= powerup
+		var/datum/abilityHolder/kart_racing/A = driver.get_ability_holder(/datum/abilityHolder/kart_racing)
+		if (A && istype(A))
```
TobleroneSwordfish, line 422:
```suggestion
		if (istype(A))
```
`null` doesn't have a type

code/WorkInProgress/racing.dm
```diff
@@ -508,6 +520,9 @@
 	var/returndir = null
 	var/turf/returnloc = null
 
+	var/occupant_vis_flags
```
TobleroneSwordfish, line 523: Weird var, doc comment it pls.

code/WorkInProgress/racing.dm
```diff
@@ -527,13 +542,24 @@
...
-			driver.client.screen += powerup
 
-		name = "[driver.name]'s Go-Kart"
+		// setup kart abilities
+		var/datum/abilityHolder/kart_racing/A = driver.get_ability_holder(/datum/abilityHolder/kart_racing)
+		if (!A || !istype(A))
+			A = driver.add_ability_holder(/datum/abilityHolder/kart_racing)
+		abilities = A
+
+		src.name_suffix(" ([driver.name])")
```
TobleroneSwordfish, line 561:
```suggestion
		src.name_suffix("([driver.name])")
```

TobleroneSwordfish (changes requested):
These things were really crusty, huh

Skeletonman0:
While i do have updated sprites that i made while working on this, this PR is already larger than i wanted when i started this.
I think i will be saving those sprites for a later PR

PR #19516, opened 2024-06-14
Introduces customizationHolder to appearanceHolder in preparation for muterace rework

Glamurio:
[Internal] [QoL] [Secret]

## About the PR 
Refactors all customizations in `/datum/appearanceHolder` to `/datum/customizationHolder`.

Appearance Holders now have `/datum/customizationHolder/customizations` which initiates with the previous default customizations, but technically allows support for more customizations. Currently the copy functions will presume that all Customization Holders have the same length, this would need to be refactored.

The list itself is associative, and gets initialized with three values, "hair_top", "hair_middle" and "hair_bottom".

Changes the following var names across all files that used them:
- `customization_first_color ` to `customizations["hair_top"].color`
- `customization_first_color_original` to `customizations["hair_top"].color_original`
- `customization_first` to `customizations["hair_top"].style`
- `customization_first_original` to `customizations["hair_top"].style_original`
- `customization_first_offset_y` to `customizations["hair_top"].offset_y`

Notably changes `customizations["hair_top"].style_original` to be `/datum/customization_style` instead of being a mere string for parity with `customizations["hair_top"].style`.

Probably has secret code changes?

Any testing would be greatly appreciated, from what I could tell this has parity with existing systems, I've tested barbers and geneticist pods (wasn't able to test cloning due to not figuring out how minds work in debugging).

Muterace Design Doc (for reference):
https://hackmd.io/SM4bP26PSLCd3aSZjOeEtA?view

## Why's this needed? 
@Tarmunora mentioned that this would be a good way to get started with the upcoming muterace rework. Generally this should allow customizations to be able to be swapped in-place.


code/modules/medical/genetics/bioHolder.dm
```diff
@@ -263,6 +246,19 @@ var/list/datum/bioEffect/mutini_effects = list()
 		special_hair_3_color_ref = toCopy.special_hair_3_color_ref
 		special_hair_3_offset_y = toCopy.special_hair_3_offset_y
 
+	proc/CopyOtherCustomizationAppearance(var/datum/appearanceHolder/toCopy)
+		if (!src.CanCopyCustomization(toCopy))
+			return
+
+		for(var/i in 1 to length(src.customizations))
```
Tarmunora, line 253: safe for now since we can assume both appearanceHolders will have the same # of customizationHolders, but will need a more robust soln in the future

Tarmunora (changes requested):
Major change worth considering, would be to make the customizations list an assoc-list, indexed by strings (e.g.: "hair_top", "hair_middle", "hair_bottom") rather than relying on an ordered list

code/modules/medical/genetics/bioHolder.dm
Glamurio: Ordered list coming right up

PR #19521, opened 2024-06-14
Buffs the bowling kit

M-Earthfire:
[Balance][Game objects]

## About the PR 

This PR changes the bowling kit in the following way:

- TC cost: 7 -> 6
- Thrown damage while wearing suit: 10 - 15 -> 15 - 20
- Melee damage while wearing suit: 5 -> 15
- Stamina melee damage while wearing suit: 20 -> 40

This PR also adds help texts to both the bowling balls as well as the bowling suit.

## Why's this needed? 

From talks on the discord, the bowling kit was in general considered to be a rather weak purchase.

While the stun is a great initiation tool for an attack or great for getaways, the kit lacks a follow up to the throw-stun. This is a problem on a rampage-cost item. Cal suggested on the discord to make the bowling ball a good melee weapon while wearing the suit, so that's what i changed. I kept the numbers low enough that the kit is unable to be fully rampage-capable like the big guns, e.g. c-saber or katana. 

I also reduced it's TC cost by 1 to enable either someone to get their own followup-item or to get a second bowling set if they want to pour all their TC into being the kingpin. This is in line with other strong non-direct rampage items, like the glaucus fishing hook or stimulants.

## Changelog 

```changelog
(u)Lord_Earthfire
(+) The bowling kit's cost was reduced to 7, down from 6.
(+) Bowling balls from the bowling kit now deal additional damage when thrown or in melee while wearing the bowling suit.
```


code/obj/item/bowling.dm
```diff
@@ -72,6 +85,18 @@
 							src.hitWeak(hitMob, user)
 		return
 
+	proc/pre_attack_damage_modifier(obj/item/parent_item, atom/A, mob/user)
+		var/mob/living/carbon/human/human_user = user
+		if(human_user.w_uniform && istype(human_user.w_uniform, /obj/item/clothing/under/gimmick/bowling))
+			//bashing someones skull in with a bowling ball should hurt if you are worthy of the bowling ball
+			src.force = 20
+			src.stamina_damage = 40
+			spawn(1)
```
TobleroneSwordfish, line 94: This is extremely cursed, couldn't you just do
```
attackby()
	src.force = x
	..()
	src.force = initial(src.force)
```
and not need the spawn or signal registration boilerplate? (this isn't a great solution either, but it's at least a bit better)

code/obj/item/bowling.dm
```diff
@@ -72,6 +85,18 @@
 							src.hitWeak(hitMob, user)
 		return
 
+	proc/pre_attack_damage_modifier(obj/item/parent_item, atom/A, mob/user)
+		var/mob/living/carbon/human/human_user = user
+		if(human_user.w_uniform && istype(human_user.w_uniform, /obj/item/clothing/under/gimmick/bowling))
```
TobleroneSwordfish, line 90:
```suggestion
		if(istype(human_user.w_uniform, /obj/item/clothing/under/gimmick/bowling))
```
`istype` already checks for `null` because `null` isn't an instance of any type

code/obj/item/bowling.dm
```diff
@@ -72,6 +85,18 @@
...
+			//bashing someones skull in with a bowling ball should hurt if you are worthy of the bowling ball
+			src.force = 20
```
TobleroneSwordfish, line 92:
```suggestion
			src.force = 15
```
20 is a LOT, that's the same as the fireaxe and reverse katana. Since this gives you four of these and they can't be used if stolen I think 20 is too far.

code/obj/item/bowling.dm
```diff
@@ -25,7 +38,7 @@
 	proc/hitHard(var/mob/hitMob, var/mob/user)
 		hitMob.visible_message(SPAN_ALERT("[hitMob] is knocked over by [user]'s [src]!"))
 
-		src.damage(hitMob, 10, 15, user)
+		src.damage(hitMob, 20, 30, user)
```
TobleroneSwordfish, line 41:
```suggestion
		src.damage(hitMob, 15, 20, user)
```
30 is a lot of damage and #19525 made this able to combo with vending machines/glass tables etc.

TobleroneSwordfish (changes requested):

Sneezless:
Just wanted to point out that the changelog has the costs the wrong way round as far as I can tell. It should be "reduced to 6, down from 7." instead, right? It's not a huge problem, obviously, just a bit confusing if you're glancing over the changelog as I was earlier.

PR #19528, opened 2024-06-15
Trait Icon Resprite 

TemThrush:
[Traits] [UI]

## About the PR 
This PR resprites all the current trait-icons and adds new ones for any traits that previously didn't have a unique icon. 
Some are as simple as updating the palette and adding an outline and some got a complete redesign!
image
[New sprites :D]
(ignore the stowaway icon i made it before it got removed as a trait, the icon is no longer part of the PR)
image
[Example of traits in the trait menu 🕺]
## Why's this needed? 
The current trait icons are very low-contrast and don't have a consistent set of values. Traits having unique and memorable icons helps in making them easy to identify when selecting traits for their character.  
Also, I just thought it would be fun :]


Gannets (approved):

Valtsu0:
Not quite sure about the atheist icon. Could use not referring to christianity, doubly so if we want to distance ourselves from real world religions. This just seems like a step in the wrong direction

otherwise looks very good

444Portal (approved):
the bible uses a cross, atheist directly effects bible interactions, icon seems fine to me
all the other sprites are very pretty! nice stuff

BatElite:
That wheelchair looks way off to me. There's no front wheels and the near back wheel is somehow behind the seat.

TemThrush:
> That wheelchair looks way off to me. There's no front wheels and the near back wheel is somehow behind the seat.

Fixed the wheelchair icon and updated the example images to reflect the change! 👍

TobleroneSwordfish (approved):

PR #19553, opened 2024-06-16
Fix organs making you full indefinitely

### LINKED ISSUE #19404: eating organs as a lizard makes your stomach full forever
### Labels

[BUG][MINOR]

### Description

Eating organs as a lizard will fill your stomach, but it won't ever deplete and won't fill your hunger meter.

### Steps to reproduce

1. Be a lizard.
2. Eat organs.
3. Get the "full" status effect.
4. Starve as your hunger meter drains but your stomach remains full.

### Expected Behavior

For organs to not remain in the stomach forever. (and actually fill the hunger meter.)

Reported by: firekestrel
Client version: 515.1635
Active test merges: #18889, #19292, and #19117
Round log date: 2024-06-08-16-47
Reported on: 2024-06-08 17:31:44
Round ID: 181066
Map: DONUT2

nova2053:
[BUG] [MEDICAL] [STATUS EFFECTS]

## About the PR 

Organs which are biological in origin are deleted upon being eaten, since they do not have a digestion proc. Cybernetic organs are unaffected.

## Why's this needed? 

Fixes #19404

## Changelog 

```changelog
(u)nova2053
(+)Normal and synthetic organs will no longer make you indefinitely full.
```

### DIFF
--- a/code/obj/item/organs/stomach.dm
+++ b/code/obj/item/organs/stomach.dm
@@ -58,7 +58,7 @@
 		if (istype(AM, /obj/item/organ)) // organs don't have a digestion proc, so just delete them instead
 			var/obj/item/organ/organtodelete = AM
 			if (!organtodelete.robotic)
-				organtodelete.Del()
+				qdel(organtodelete)
 				return
 		AM.set_loc(src.donor)
 		src.stomach_contents |= AM
--- a/code/obj/item/organs/stomach.dm
+++ b/code/obj/item/organs/stomach.dm
@@ -57,7 +57,7 @@
 			return
 		if (istype(AM, /obj/item/organ)) // organs don't have a digestion proc, so just delete them instead
 			var/obj/item/organ/organtodelete = AM
-			if (organtodelete.robotic == 0)
+			if (!organtodelete.robotic)
 				organtodelete.Del()
 				return
 		AM.set_loc(src.donor)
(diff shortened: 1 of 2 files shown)


code/obj/item/organs/stomach.dm
```diff
@@ -55,6 +55,11 @@
 	proc/consume(atom/movable/AM)
 		if (AM in src.stomach_contents)
 			return
+		if (istype(AM, /obj/item/organ)) // organs don't have a digestion proc, so just delete them instead
+			var/obj/item/organ/organtodelete = AM
+			if (!organtodelete.robotic)
+				organtodelete.Del()
```
TealSeer, line 61:
```suggestion
				qdel(organtodelete)
```

code/obj/item/organs/stomach.dm
```diff
@@ -55,6 +55,11 @@
 	proc/consume(atom/movable/AM)
 		if (AM in src.stomach_contents)
 			return
+		if (istype(AM, /obj/item/organ)) // organs don't have a digestion proc, so just delete them instead
```
ZeWaka, line 58: Why not implement a digestion thing that does nothing/deletes them?

nova2053: Organs being immediately deleted prevents people from vomiting up the organs before the digestion loop and re-consuming them, therefore preventing people from getting unlimited amounts of metachroitic factor (saurian points system) from the same organ. Also, most organs dont hold reagents, and hearts (the only organs to my knowledge that do hold reagents) immediately transfer them all to the consumer when eaten anyways, making a digestion proc unnecessary.

nova2053: Updated it to use a proc that deletes organs in the stomach every life cycle, seems to work

ZeWaka: even if the performance might be wack this is a lot cleaner thank you

PR #19559, opened 2024-06-17
TGUI Antag Tips

Garash2k:
## About the PR 
Built on top of what's done in #19486, this is a migration of the antag tips popups to tgui.
It's basically just deleting the old `browserassets/html/traitorTips/nukeopTips.html` files and pasting it's markup in new `tgui/packages/tgui/interfaces/AlertContentWindows/acw/nukeop.AlertContentWindows.tsx` ones.
One thing I'm concerned with are the images. They currently work with `` and apparently that might have caused CDN issues before and I'm not quite sure how the old `{{resource(...)}}` technique worked. I imagine a testmerge could answer that worry?

Regular:
image

Light theme (see below):
image

Syndicate:
image

Flock:
image

Note, because of the recent tgui-core integration, themes aren't currently looking so hot, they'll look better once that's improved, for example, this is what the white theme would look like:
image

Here was the conversion process:

- Get the title and html file from antag_popups.dm

- Copy the .traitor-tips element from the html file into a .tsx file, delete the html one.

- Do the following renames:
  - class -> className
  - `<br>` -> `<br />`
  - Replace quotes and apostrophes with `&quot;` and `&apos;`
  - {{resource(...)}} -> just the filename
  - Fix the wiki link
  - Correct the markup, there were occasional `<p>paragraph<p>` that should have been `</p>`s
  - Upon saving, the whole text would automatically add line breaks to respect the code width limit, so I rolled with it.

## Why's this needed? 
Dark/Light theme compatibility, themed popups for syndicate and flock, and allows for better antag tips in the future

## Changelog
```changelog
(u)Garash2k
(+)New UI for the antag tip popups! Make sure to report any errors
```

[ui][feature]


FlameArrow57:
Nice, possible to use existing Flock theme for Flock pop up maybe?

Garash2k:
Totally possible yes, I'll look into adding a theme property on the `AlertContentWindow`, maybe syndicate antag popups could use the syndicate theme too.

Garash2k:
Now that #19486 is merged, this is very doable. Personally I'd have time to look into it it a few weeks after the TGUI freeze is over, because, well FFXIV Dawntrail is out :), but if anyone feels like beating me to the punch I say go for it!

Garash2k:
we are so back

Garash2k:
I am pretty cluless about the "error: /datum/job/special/ghostdrone/proc/special_setup sets SpacemanDMM_should_not_sleep but calls blocking proc /proc/tgui_input_list" error because of ghostdrone and nothing else.

TobleroneSwordfish:
Yeeeah looks like the images don't work: 
image
(Testmerged on the dev server)
Maybe look at how business cards work? They serve images directly from browser assets iirc.

Garash2k:
Thanks for the pointer, business cards use `/proc/resource(...)`, which either builds a `"[cdn]/[file]?v=" + VCS_REVISION` url, or strips out the path if a local server, which is what I did manually. This could be done TGUI side if we passed the cdn and the VCS_REVISION.

tgui/packages/tgui/cdn.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * @file
+ * @copyright 2024
+ * @author Garash2k (https://github.com/Garash2k)
+ * @license ISC
+ */
+
+import { useBackend } from './backend';
```
ZeWaka, line 8: you would need to throw this in a goon folder or something

code/modules/tgui/tgui_alert.dm
```diff
@@ -129,6 +129,8 @@
 		"items" = items,
 		"autofocus" = autofocus,
 		"theme" = theme,
+		"cdn" = cdn,
+		"VCS_REVISION" = VCS_REVISION,
```
ZeWaka, line 133: gotta mark as goonstation-add

TobleroneSwordfish:
Looking into the linter issue: it seems like the problem is that `tgui_alert` *can* call `tgui_input_list` and `alert.wait()`, both of which sleep. To fix this you could probably just move the `G.show_antag_popup` call from `/proc/droneize` down a couple of lines into the `SPAWN` block OR mark `show_antag_popup` as `set waitfor = FALSE` to make it safe to call from anywhere.

PR #19564, opened 2024-06-17
Add freezers to toxins on all rotation maps that didnt already have them

LaCumbiaDelCoronavirus:
## About the PR
This adds freezers (like the ones that cool gas in the TEG/kitchen freezer, not an actual freezer room) to Science's toxins in Cogmap1, 1Pamgoc, Donut 2, Donut 3, Atlas, Clarion and Mushroom.
I tried to keep stuff in toxins and also still make sense where its located if i had to shift it around to make space for the freezers.

## Why's this needed?
Freezers are useful equipment in toxins, and for parity, since some maps lack them while others have them.

## Changelog
```changelog
(u)DecodeQwertyPls
(+)Most maps now have freezers in toxins.
```


AXLFAN2000:
Actual comment now that I'm home, I think this would be a super super nice QOL when it comes to working with the crystal bazaar/encouraging non-antag toxins use  --  having both a cold loop and the freezer like on kondaru gives you WAY more freedom to actually play around with specific temps and hit bounties

LaCumbiaDelCoronavirus:
why did it change from XL to XXL? looked at the mapdiff and nothing changed there from before i made that commit to resolve the merge conflict.. oh well

TobleroneSwordfish:
Out of interest, what is actually the difference between using a freezer and a cold loop radiator? Is the freezer just better because it lets you fine-tune the temperature?

AXLFAN2000:
> Out of interest, what is actually the difference between using a freezer and a cold loop radiator? Is the freezer just better because it lets you fine-tune the temperature?

Absolutely, it makes hitting specific temps for bounties a lot easier since it doesn't step down in temperature in as big leaps.  Having both the cold loop and freezer is helpful as well because you can use the cold loop for cooling your oxygen and the freezer for cooling plasma to hit measurements

pgmzeta:
> IIRC, 2-tile freezer setups that are just the freezer + canister port don't hold any gas in pipes. So there's also the niche use of not losing gas.

the freezer still contains some gas, so you will always lose gas

having both is really useful because there's no other way to cool oxy and and plasma without them mixing

TobleroneSwordfish (changes requested):
Not a fan of the atlas vendor being blocked into a corner like that, it looks odd. Is there anywhere else you could shove it?
Also make sure the moved switches are on different tiles and not just pixel-offset or they'll behave weirdly.

TobleroneSwordfish:
Oh uh, the second one looks fine then 👍 Atlas is just kind of a cramped station.

PR #19568, opened 2024-06-17
Give Fire Elementals a new ability that lets them create "permanent" fires that shoot embers to heal them

Kyle2143:
## About the PR 
- Adds a "Self-Immolate" ability that creates 3 dark-red chemfires on fire elemental and damages the elemental.
- Can only be cast if they have over 100 health (their full health is 150). Deals 50 damage to self.
- The red fire it spawns shoots out embers on the fire elemental's Life loop if its nearby
- Embers heal the fire elemental and give stamina, but make tiny fireflashes if they hit other mobs/objects
- The dark red fire is a totally normal chemfire otherwise, you can extinguish it normally
https://streamable.com/mw5g0p

## Why's this needed? 
I just feel like When I see a group of Fire Elementals set up shop somewhere, they should have something that lets them keep some fire around permanently. Added the healing mechanic for fun.


imawxome:
I dont think this needs self harm to be balanced, it can be put out by an Extinguisher anyway, so i think an areal denal for a few seconds is good

PR #19569, opened 2024-06-17
VR kart mechanics fixes

Skeletonman0:
[OBJECTS] [BUG]
## About the PR 
Fixes VR kart boosting and stops most sound effects from stacking multiple times

VR boosting would reset early due to the original code using SPAWN() to remove the added speed
this PR replaces the boosting system, the spin out system, and the super boost system with three status effects

This status effect based system allows you to extend a boost as long as you hit another one before your current one ends

Changes when some kart sounds are played, slapping in some cooldowns to try and stop the sounds stacking
## Why's this needed? 

VR kart boosting is broken, VR kart sounds can stack over and over and it gets annoying / loud fast


Skeletonman0:
spinning still appears to be broken

code/modules/status_system/statusEffects.dm
```diff
@@ -2900,3 +2900,102 @@
 
 	getTooltip()
 		return "You've [alpha == 0 ? "completely" : "partially"] faded from view! People can still hear you and see light from anything you're carrying."
+
+/datum/statusEffect/kart_boost
+	id = "kart_boost"
```
ZeWaka, line 2905: move these to the racing.dm file

ZeWaka (changes requested):

PR #19587, opened 2024-06-19
Cogmap2 bathroom expansion

Kubius:
[MAPPING][FEAT]
## About the PR 

* Upgrades the bathroom next to Cogmap 2's public market with two individual toilet stalls and a small shower room with occupancy sign (doorbolt button not used as the bathroom is part of maintenance access).

image

* Rearranges the arrivals locker room bathroom to break out the two toilets into distinct stalls with a divide between, and add a small shower area with single shower head.

image

* There was some maintenance rearrangement to allow this - maintenance contents by these areas should be unchanged, though some APCs near the south bathroom have relocated as it's now its own distinct area.

## Why's this needed? 

Cogmap 2 currently has something of a deficit of public showers, with only three shower heads outside of departments (one in the wash room near Bridge, two in the gym), and the south bathroom change helps this significantly. North bathroom change, in addition to a smidge of extra shower availability, also fixes a lack of distinct stalls after the map's removal of cubicle walls.

## Changelog 

```changelog
(u)Kubius
(+)Cogmap 2's bathroom near QM is now larger - equipped with 2 toilets and a small shower room. Arrivals washroom has a shower head too.
```


ZeWaka:
pictures plkease

Kubius:
Added some pics snipped from StrongDMM with area borders on. If you explicitly need a before-and-after that the diff isn't covering, or ingame pictures, let me know.

Kubius:
!merge_upstream

PR #19598, opened 2024-06-19
Geodes

TobleroneSwordfish:
[FEATURE] [GAME OBJECTS]
## About the PR 
Adds geodes that mining can find buried in asteroids.
- Geodes can only be broken open with explosions, with some of the softer geodes being able to be broken using mining charges but tougher geodes requiring more powerful explosions.
- The power required to break open a geode is shown when using a geological scanner on it.
- Geodes can contain either crystalline ore deposits or fluids.
- Once broken open, crystal geodes can be hit with a mining tool to pry out additional chunks of crystal.

image

Sprites are all by @Flaborized with some recolours and tweaks done by me.
## Why's this needed? 
- Geodes are cool. 
- Lets us give mining a little bit of some rarer resources more commonly without giving them an entire starstone or erebite asteroid.
- Provides an alternate way to get oil, sulfuric acid etc.

## Changelog 

```changelog
(u)LeahTheTech and Flaborized
(*)Added geodes that mining can find. Can only be broken open with explosions, check your geological scanner for the power required. Geodes can contain various types of crystalline ore deposits as well as oil and other liquids.
```


Gannets:
geode :)

PR #12132, opened 2022-11-23
Change ContextLayout ExperimentalCircle to start from the top and work clockwise

CheffieGithub:
[QOL] [UI]
## About the PR 
Changes the experimental circle context layout to start making the buttons from the top and clockwise. This will effect anything that uses the experimental circle as its context layout. RCD, AI holograms, Robospray, Security RecordTrak, ect. 

## Why's this needed? 
This makes context layouts with an odd number of actions look nicer as they don't look tilted. Starting from the top and going clockwise is more intuitive then starting from the right and going counter clockwise. example:

before
image

after
image

RCD change for example also:
image

also changes one icon state for the lamp manufacturer to be more clear.

```changelog
(u)Cheffie
(+)Changes how the ExperimentalCircle context layout is created to be more intuitive. Changes current items that use this layout to roughly match how they were before in terms of layout.
```

PR #19623, opened 2024-06-21
adds a second subtler palette to portable lamp manufacturer

BatElite:
## About the PR 
Allows portable lamp manufacturers to switch between the current set of flashy colours and a set of milder ones. The new palette comprises the cool/warm/harsh options as well as the reddish/yellowish/greenish/blueish/purpleish flavours. About the only inaccessible colour at this point is the light purpleish tubes, which have no bulb equivalent and only exist on Atlas and also are nearly indistinguishable from very harsh tubes anyway.

Because the context menu is stitched from parts now, I couldn't keep the ordering that #12132 preserved through cleverly ordering the options in code. Where the settings would previously sit (the southeast quarter) is technically in the middle of the palette. Shame to see it go though.
image

Minor changes:
-Adds a reddish bulb, which coolstation already had but only had tubes here. I didn't really want to lose the reddish option since it's actually a nice orange, which I previously found to be lacking.
-Simplifies some complicated looking code for loading sheets in.

## Why's this needed? 
While the bright selection currently is fun and all, they can be limited use because they're so harshly coloured. Borgs can now go do mood lighting and repairing. A bunch of the tinted lighting on Oshan for example wasn't exactly replaceable.
Hopefully this will also attune some folks to the subtle impact of colour temperature on a space, and the horror of mildly green lighting.

## Changelog 

```changelog
(u)BatElite
(+)Portable lamp manufacturers now have a second page of milder colour choices.
```


code/obj/submachine/robotics.dm
```diff
@@ -315,14 +315,27 @@
 	//can be obj/machinery/light for wall tubes, obj/machinery/light/small for wall bulbs. Either mode does floor fittings because there's only one type of those
 	var/dispensing_fitting = /obj/machinery/light
 	var/list/setting_context_actions
+	var/list/common_actions
+	var/list/page_1_actions
+	var/list/page_2_actions
 	contextLayout = new /datum/contextLayout/experimentalcircle
```
pgmzeta, line 321: pass in a value to have this circle embiggened (32 is the default size; the ai hologram menu also does this)
```suggestion
	contextLayout = new /datum/contextLayout/experimentalcircle(36)
```

it's going to look a bit weird with the circles have a different count of options, but this will help it not overlap as much:
Screenshot 2024-06-29 120236
Screenshot 2024-06-29 120241

you could, alternatively, have the P1/P2 buttons change the circle size but that might be weirder to use in practice

BatElite:
Cheers pgm

image
Adjusted the sizes per page, but it looks like it's not possible to remove the overlap entirely without creating large gaps.

PR #19627, opened 2024-06-21
Knockout effect when hitting crit inside boxing rings + bell resprite

CalliopeSoups:
[STATUS EFFECTS] [FEATURE] 
## About the PR 

Humans on-top of boxing mats will be given the 'wrestling' status effect when ringing the wrestling bell. Hitting critical condition, leaving the ring for 15 seconds or ringing the bell again after a cooldown will remove the status. 

Hitting critical condition with the status active will play a knockout effect, removing the status from nearby people.

This PR is intended to work as a baseline for potential future ring updates.

https://github.com/goonstation/goonstation/assets/102558507/0fa7a2b6-dc6a-4f65-babe-841757e1e5f5

## Why's this needed? 

It can be somewhat hard to tell who won a match without a doctor present, this makes the first person to hit crit in a boxing ring be visibly and audibly knocked out.

I was planning to add a 'second wind' buff to the person knocked out to tide them over until a doctor can get to them. I haven't added this for fear of people gaming it somehow, but I can see about adding it if it's wanted.

As discussed on the forums here: https://forum.ss13.co/showthread.php?tid=22723.

## Changelog 

```changelog
(u)CalliopeSoups
(+)The boxing ring is now a bit more functional, featuring a new bell, vaultable entrances and sfx for being KOd.
```


Retrino:
it would be good to tie the wrestling buff only when the bell has been rung to initiate a match, instead of just when entering the ring. deathmatches (which people do use the ring for) would be interrupted and give the person who crits first a severe disadvantage. it makes more sense to apply it to people who actually want it applied

CalliopeSoups:
> deathmatches (which people do use the ring for) would be interrupted and give the person who crits first a severe disadvantage. it makes more sense to apply it to people who actually want it applied

Yeah I hadn't considered that. I suppose it wouldn't be much of a disadvantage if I hadn't included the force lie down, but I think that adds to the effect so I'll keep it.

I'll tie actually giving and removing the status to the bell. I need to add a sprite for the status anyhow, and I might add a little effect on those given the status on application so you can better tell that ringing the bell has Done Something.

CalliopeSoups:
Edited the PR's description and uploaded a new demo video to reflect recent commits.

Flaborized:
the KO voice clip sound is currently used for tyson gibbing, but it's secretly just brazenly taken from Scott Pilgrim (and scott pilgrim brazenly took the sample from street fighter alpha 3)

I think for a weird admin-only reference gib this is OK but I personally don't think we should put more Outright Taken stuff in common game places that are player accessible. im not speaking as an Authority on this though and other devs may disagree

CalliopeSoups:
> the KO voice clip sound is currently used for tyson gibbing, but it's secretly just brazenly taken from Scott Pilgrim (and scott pilgrim brazenly took the sample from street fighter alpha 3)
> 
> I think for a weird admin-only reference gib this is OK but I personally don't think we should put more Outright Taken stuff in common game places that are player accessible. im not speaking as an Authority on this though and other devs may disagree

Yeah that makes sense, I was fully expecting at least some discussion over the sound clip. It shouldn't be too hard to whip up an original version that sounds similar, though I'm not too great at audio myself. I can see how other devs feel and change it if need be.

CalliopeSoups:
Removed the KO sound for now, I've tried to make a good replacement but I'm pretty bad at it. I'll add one at a later date if I'm able.

code/modules/status_system/statusEffects.dm
```diff
@@ -1591,12 +1591,57 @@
...
 
 		onRemove()
+
 			. = ..()
 			if(ismob(owner))
 				var/mob/M = owner
 				REMOVE_ATOM_PROPERTY(M, PROP_MOB_MELEEPROT_BODY, src)
 				REMOVE_ATOM_PROPERTY(M, PROP_MOB_MELEEPROT_HEAD, src)
 
+	wrestler
```
ZeWaka, line 1601: Nah. Absolutely path this and put it organized somewhere.

code/turf/floors.dm
```diff
@@ -989,6 +989,24 @@ DEFINE_FLOORS(minitiles/black,
 	name = "boxing mat"
 	icon_state = "boxing"
 
+	Entered(atom/movable/A, atom/oldloc) // if a human leaves the ring with the wrestler status, they get a countdown to get back in before it's removed
+		. = ..()
+		if (istype(oldloc, /turf/simulated/floor/specialroom/gym) || !ishuman(A))
+			return
+
+		var/mob/living/M = A
```
ZeWaka, line 997: Human.

code/turf/floors.dm
```diff
@@ -989,6 +989,24 @@ DEFINE_FLOORS(minitiles/black,
...
+		if (M.hasStatus("wrestler"))
+			M.setStatus("wrestler", INFINITE_STATUS)
+
+	Exited(atom/movable/A, atom/newloc)
+		. = ..()
+		if (istype(newloc, /turf/simulated/floor/specialroom/gym) || !ishuman(A))
+			return
+
+		var/mob/living/M = A
```
ZeWaka, line 1006: Human.

code/obj/item/wrestlingbell.dm
```diff
@@ -0,0 +1,112 @@
...
+		if (src.hammer)
+			src.hammer.force_drop(sever=TRUE)
+			src.hammer.set_loc(src)
+			src.hammer.parent = null
+
+			UpdateIcon()
+
+	proc/toggle_boxing_mode()
+		playsound(src.loc, 'sound/misc/Boxingbell.ogg', 50,1)
+		for (var/mob/living/carbon/human/human in view(10, src))
```
ZeWaka, line 105: Not a huge fan of this. :/

code/obj/item/wrestlingbell.dm
```diff
@@ -0,0 +1,112 @@
...
+		RegisterSignal(user, COMSIG_MOVABLE_MOVED, PROC_REF(hammer_move), TRUE)
+		UpdateIcon()
+
+	attackby(obj/item/W, mob/living/user)
+		user.lastattacked = src
+		if (user.a_intent != "harm")
+			src.put_back_hammer()
+			return
+		if(istype(W, /obj/item/wrestlingbell/hammer) && user.a_intent == "harm")
+			if(last_ring + 10 SECONDS >= world.time)
```
ZeWaka, line 82: Use cooldown

code/obj/item/wrestlingbell.dm
```diff
@@ -0,0 +1,112 @@
...
+			icon_state = "wrestlingbell1"
+		else
+			icon_state = "wrestlingbell0"
+
+	attack_hand(mob/living/user)
+		if (isAI(user) || isintangible(user) || isobserver(user) || !in_interact_range(src, user)) return
+		user.lastattacked = src
+		..()
+		if(!hammer || QDELETED(hammer))
+			hammer = null
```
ZeWaka, line 67: ? What?

CalliopeSoups:
Notable changes:

- Status applies/removes on entering and exiting again, as the only way I could think of having the bell apply it was costly.
- boxingropeenter is now solid and vaultable
- Knockout sfx is back now royalty free (https://pixabay.com/sound-effects/ko-95973/)
- Compatible with wrestlemap

ZeWaka:
is the demo still accurate

CalliopeSoups:
> is the demo still accurate

Somewhat, I'll make a new one in a bit today.

CalliopeSoups:
Up to date showcase!

https://github.com/user-attachments/assets/18fbe01d-a3db-4afc-a111-52f373aec68e

ZeWaka:
X

CalliopeSoups:
I goofed up master merging, will fix soon.

CalliopeSoups:
Lord help me I'm messing it up bad

Edit: All fixed, sorry for all the devs added for review by mistake!

code/obj/railing.dm
```diff
@@ -262,7 +264,7 @@
 
 	New(The_Owner, The_Railing, use_owner_dir = FALSE)
 		..()
-		collision_whitelist = typesof(/obj/railing, /obj/decal/stage_edge, /obj/sec_tape)
+		collision_whitelist = typesof(/obj/railing, /obj/decal/stage_edge, /obj/railing/wrestling, /obj/sec_tape,)
```
JORJ949: /obj/railing/wrestling should be in types_of(/obj/railing)?

CalliopeSoups:
New things:

- The status activating now shows a boxing gloves visual effect.
- Wrestling status counts down when leaving the boxing mat, not really needed but it can stop the boxing gloves effect from appearing more than it must.
- All boxing rings in wrestlemap are now compatible, only major changes are slight area adjustments to make wrestling bells work.

I think features wise it's ready, the code will need a review before going forward though.

444Portal:
Reopening per user request

code/obj/item/wrestlingbell.dm
```diff
@@ -0,0 +1,172 @@
...
+			UnregisterSignal(M, COMSIG_ATTACKBY)
+			if (M.health <= 0 && src.active)
+				SPAWN(0)
+					playsound(M.loc, 'sound/misc/knockout_new.ogg', 50)
+				playsound(M.loc, 'sound/misc/Boxingbell.ogg', 50,1)
+				M.make_dizzy(140)
+				M.UpdateOverlays(image('icons/mob/critter/overlays.dmi', "dizzy"), "dizzy")
+				M.setStatus("resting", INFINITE_STATUS)
+				SPAWN(10 SECONDS)
+					M.UpdateOverlays(null, "dizzy")
```
flappybatpal: This is a long spawn and should check if M still exists

CalliopeSoups:
> There are many icons deleted from cons/ui/statussystem.dmi?

Ah yeah there was an issue with a master merge there, I think I restored em.

code/obj/item/wrestlingbell.dm
```diff
@@ -0,0 +1,173 @@
...
+			return //if someone else has it, don't put it in user's hand
+		user.put_in_hand_or_drop(src.hammer)
+		src.hammer.parent = src
+		RegisterSignal(user, COMSIG_MOVABLE_MOVED, PROC_REF(hammer_move), TRUE)
+		UpdateIcon()
+
+	attackby(obj/item/W, mob/living/user)
+		user.lastattacked = get_weakref(src)
+		if (istype(W, /obj/item/tinyhammer/wrestling))
+			if (user.a_intent != "harm")
```
TobleroneSwordfish: Needing to use harm intent to activate the bell is very bad usability (I'd never have figured it out without looking at the code), non-mob intent interactions should be kept to an absolute minimum, especially for something that new players are likely to interact with a lot.

I'd rather just keep it as a single click action over this.

code/turf/floors.dm
```diff
@@ -1048,6 +1048,16 @@ DEFINE_FLOORS(minitiles/black,
 	name = "boxing mat"
 	icon_state = "boxing"
 
+	Entered(atom/movable/M, atom/OldLoc)
+		..()
+		if (!istype(OldLoc, /turf/simulated/floor/specialroom/gym))
```
TobleroneSwordfish: What is the purpose of adding the wrestling status on entering the ring? Why not just add it when the bell is rung?

code/obj/item/wrestlingbell.dm
```diff
@@ -0,0 +1,173 @@
...
+			M = owner
+			UnregisterSignal(M, COMSIG_ATTACKHAND)
+			UnregisterSignal(M, COMSIG_ATTACKBY)
+			if (M.health <= 0 && src.active)
+				SPAWN(0)
+					playsound(M.loc, 'sound/misc/knockout_new.ogg', 50)
+				playsound(M.loc, 'sound/misc/Boxingbell.ogg', 50,1)
+				M.make_dizzy(140)
+				M.UpdateOverlays(image('icons/mob/critter/overlays.dmi', "dizzy"), "dizzy")
+				M.setStatus("resting", INFINITE_STATUS)
```
TobleroneSwordfish: If the knockout effect is meant to end a fight then it should probably come with several seconds of "knockdown" status to ensure the KO'd person actually stays down and doesn't keep hitting their opponent with a chair or something. (resting is not the same as knockdown)

code/obj/item/wrestlingbell.dm
```diff
@@ -0,0 +1,173 @@
...
+	onUpdate(timePassed)
+		var/mob/M = null
+		if(ismob(owner))
+			M = owner
+		else
+			return ..(timePassed)
+		if (M.health <= 0)
+			M.delStatus("wrestler")
+
+	onRemove()
```
TobleroneSwordfish, line 133: If this is `onRemove` then won't it KO both wrestlers when one enters crit?

CalliopeSoups: Not from what I've tested, the status isn't removed from anyone but the crit player.

code/obj/item/wrestlingbell.dm
CalliopeSoups: Agreed, I made it ring the bell if you click it with the hammer at all, it gets put away after a couple seconds. It plays a noise whenever the hammer gets put back in general now too, not just specifically placing it down like before.

code/turf/floors.dm
CalliopeSoups: I'm sure there was a purpose at some point but I don't recall it and it works fine without it, just ringing the bell and punching someone who has the status gives it out now.

PR #19633, opened 2024-06-22
Makes pufferfish actually obtainable

colossusqw:
[LABEL][QoL][Catering]

## About the PR 

This PR adds pufferfish spawns to similar spots as other aquarium fish.

## Why's this needed? 

You can't currently fish pufferfish.


colossusqw:
Yes i did fundamentally misunderstand fish spawn rates but in my defense i DO be dum.

Gannets (approved):

PR #19636, opened 2024-06-22
Fix contributor scream resetting constantly

TealSeer:
[BUG] [PLAYER ACTIONS]
## About the PR 

Changes contributor scream code to add sound to appearanceHolder list of screams and change `screamsound` to match, instead of setting `sound_scream` directly.

## Why's this needed? 

Contributor scream was reverting every time `appearanceHolder.UpdateMob()` was called (which is often)

### DIFF
--- a/code/modules/medals/rewardsLocker.dm
+++ b/code/modules/medals/rewardsLocker.dm
@@ -1517,7 +1517,9 @@
 	proc/sillyscream(mob/M)
 		var/mob/living/living = M
 		if(istype( living ))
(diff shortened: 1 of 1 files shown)


ZeWaka (commented):
see but do i want to fix this because it's so terribly annoying

TealSeer:
haha good question

Tarmunora:
> see but do i want to fix this because it's so terribly annoying

eh, feature culling can happen later if desired

PR #19646, opened 2024-06-22
Reintroduce soldering iron and t-ray scanner to general manufacturer

Kubius:
[BALANCE]
## About the PR 

Reintroduces the soldering iron and t-ray scanner to the slate of items general manufacturers can produce. Does **not** reintroduce the multitool.

## Why's this needed? 

c816c33 removed these tools from the general manufacturer at the same time they were added to the engineering specialist manufacturer. Between their common use in requisitions, the soldering iron's importance in deploying machinery frames, and the fact specialist manufacturers are pretty rare, I'd like to reintroduce them to the general manufacturer.

## Changelog 

```changelog
(u)Kubius
(+)General manufacturer can once again make soldering iron and t-ray scanner. Notable for requisitions.
```


Kubius:
@ sord: requested you because you were the one who made the removal initially

Kubius:
After some external discussion, maybe not the best idea.

PR #19648, opened 2024-06-22
Requisition item requirement mini-fixbatch

Kubius:
[FEAT][BALANCE]
## About the PR 

Bit of requisition tuning, primarily tuning around de-emphasis of annoying-to-access items.

- The "Computer Failure" aid requisition will no longer be able to require a soldering iron or T-ray scanner, and can now require a power cell or circular saw instead. It is much less likely to require a multi-tool.
- The "Engineering firm" Prototypist requisition client will now ask for 1 to 2 t-ray scanners when asking for some, down from 2 to 4.
- The "Food Delivery" special requisition now pays out 2x as much (base payout number changed from PAY_TRADESMAN to PAY_DOCTORATE).
- Welding fuel has received a substantial reduction in its sale price, and water a slight reduction. Fuel in particular was significantly overvalued for how much you can sell using fuel cans - the requisition that requires it in bulk now also requires a larger quantity.
- The seed genetics requisition can no longer require Lifespan; that roll now requires 5-10 Production and 3-5 Potency.
- Reduced the frequency of the Event Catering requisition (60 > 40 weight) and Birthday Bash requisition (70 > 50 weight).

## Why's this needed? 

Adjusts some requisitions to stay appropriately tuned against item acquisition difficulty, and rein in some outliers quantity-wise.

## Changelog 

```changelog
(u)Kubius
(+)Small update pass on requisitions, notably less using tools no longer available in general manufacturer.
```


code/modules/economy/requisition/rc_aid.dm
```diff
@@ -270,20 +271,10 @@ ABSTRACT_TYPE(/datum/rc_entry/item/surgical)
...
-/datum/rc_entry/item/soldering
-	name = "soldering iron"
-	typepath = /obj/item/electronics/soldering
-	feemod = PAY_TRADESMAN*2
-
 /datum/rc_entry/item/multitool
 	name = "multitool"
 	typepath = /obj/item/device/multitool
-	feemod = PAY_TRADESMAN*4
+	feemod = PAY_EXECUTIVE
```
pgmzeta, line 277: these are easily orderable, not sure it needs the bonus pay

code/modules/economy/requisition/rc_scientific.dm
```diff
@@ -545,7 +545,7 @@ ABSTRACT_TYPE(/datum/rc_entry/item/organ)
 			if(PROTOTYPIST_ENGINEER)
 				prototypist_desc = "Engineering firm"
 				if(prob(60))
-					src.rc_entries += rc_buildentry(/datum/rc_entry/item/tscan,rand(2,4))
+					src.rc_entries += rc_buildentry(/datum/rc_entry/item/tscan,rand(1,2))
```
pgmzeta, line 548: these can always be filled by 1 electrical maintenance crate

pgmzeta (approved):

code/modules/economy/requisition/rc_aid.dm
Kubius: Was on the fence about it. I don't mind it being swapped back if there's broad agreement.

code/modules/economy/requisition/rc_scientific.dm
Kubius: They can, yes. Just less tedium to not randomly need *four* of them.

code/modules/economy/requisition/rc_aid.dm
Kubius: You know what, you're probably right.

NibChocolateeny:
@Kubius maybe you camn change botanists recquisition?

for seeds, it is not so hard to get all statistics, EXCEPT lifespan ("harvests") - it is only ramndom (can get fom mutashions), all other stat can be got from infuse in direct way. Can you change the number range for lifepsan from 3-5 and make it more small, like 1-2? it is not so fun to have contact that cannot be finishe because lifespan 4 is wanted, that is super super hard! (more hard than all otherss)

Tarmunora (approved):

PR #19653, opened 2024-06-23
Use a process controller to handle shipping market refreshes

pgmzeta:
[qol][bug][station systems]
## About the PR 
Move shipping market refresh handling to a process controller.

Note: This will cause mail deliveries to occur even when there are no QMs/Mail Couriers.

## Why's this needed? 
Mail couriers do not have QM console access, and thus cannot recieve more mail until the market has refreshed. If there is no active QMs for whatever reason, then the mail courier can be SOL with their one job.
Stifle some small cheese where you can keep hot items hot indefinitely if no one ever checks the console.

## Changelog 

```changelog
(u)glowbold
(*)Market refreshes and mail deliveries will now occur automatically.
```

### DIFF
--- a/code/modules/economy/shippingmarket.dm
+++ b/code/modules/economy/shippingmarket.dm
@@ -74,6 +74,7 @@
 
 		SPAWN(300)
 			market_shift()
+			auto_market_refresh()
 
 		var/list/unique_traders = list(/datum/trader/gragg,/datum/trader/josh,/datum/trader/pianzi_hundan,
 		/datum/trader/vurdalak,/datum/trader/buford)
@@ -315,6 +316,12 @@
 
 			update_shipping_data()
 			update_buy_prices()
+
+	///Ensure market refresh / mail delivery happens regularly, even if no one checks the console
+	proc/auto_market_refresh()
+		src.timeleft()
+		SPAWN(5 MINUTES)
+			src.auto_market_refresh()
 
 	proc/generate_mail(time_since_previous)
 		var/adjustment = max(time_since_previous, 2 MINUTES)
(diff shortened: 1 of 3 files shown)


frawhst (changes requested):
This should definitely be a process loop instead of a spawn loop. See `code/datums/controllers/process/stock_market.dm` or `code/datums/controllers/process/ticker.dm` for simple examples of what those look like. Also, we probably don't need the code to manually refresh the market since it's handled automatically, so feel free to axe that.

code/modules/economy/shippingmarket.dm
```diff
@@ -72,9 +71,6 @@
 				qdel(C)
 
 
-		SPAWN(300)
-			market_shift()
-
```
Tarmunora, line 77: hm
without this early market shift, when does the first mail crate actually arrive in a round? at the ~7 minute mark? ~14?

pgmzeta: It comes in at ~7 since the first market shift now happens at round-start

frawhst (approved):

code/modules/economy/shippingmarket.dm
```diff
@@ -151,27 +144,17 @@
...
+	/// Returns the time in MM:SS format
 	proc/get_market_timeleft()
 		var/timeleft = src.timeleft() / 10
 		if(timeleft)
 			return "[add_zero(num2text((timeleft / 60) % 60),2)]:[add_zero(num2text(timeleft % 60), 2)]"
 
 	proc/market_shift()
-		#ifndef FUCK_OFF_WITH_THE_MAIL
-		var/time_since_previous = (TIME - last_market_update)
-		#endif
```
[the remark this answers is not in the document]
pgmzeta: still used down here

https://github.com/goonstation/goonstation/blob/febafea5807f83bc2ed355610874e6f169e8cd7b/code/modules/economy/shippingmarket.dm#L280-L284

PR #19657, opened 2024-06-23
Resprite the currently existing cat ears and make them overlay over helmets

ang0501:
[Sprites] [Clothing]

## About the PR 

Resprites the cat ears from their current barely visible tiny sprites into a bigger, more visible sprite, also uses a colour palette more in line with the other current sprites in game. It also makes so that when using cat ears as a headset, the cat ears will overlay over the helmet you're currently using, helping with their visibility, and looking really cute.

image
image
image
image
image

Examples of the cat ears overlayed on different helmets:
image

## Why's this needed? 
As discussed both in the forums and the discord server, the current cat ears in game are too tiny to be noticeable, and thus a resprite to make them more visible would be nice. The overlay over helmets is also needed to help with the visibility.

## Changelog 

```changelog
(u)Angelx3
(+)Resprites the cat ears and all their recolours.
(+)Makes cat ears overlay over helmets.
```


Laboredih123:
🐱

ang0501:
> I suggest adding help text to cat ears, to notify players that they can be attached to headsets. I had nooo idea that was a thing

Yeah that would be nice, but imo, out of the scope of this PR.

TobleroneSwordfish (approved):
Sprites look good, the layering is a little odd but I think it's fine 👍

PR #19658, opened 2024-06-23
Transport chutes flush after people have finished entering, instead of right away

pgmzeta:
[qol][game objects]
## About the PR 
Set transport chutes to flush at the end of the action bar for entering/shoving someone into a chute instead of immediately.

## Why's this needed? 
Makes transport chutes consistently send people after they've entered, instead of depending on machineloop process timing.
Not popping up a UI makes them feel more 'automatic'; you can still bring up the ui by slapping it with your hands before entering.


Sbmhawk:
Eh, I like being able to hide in them, I dont think this is a good idea

pgmzeta:
> Eh, I like being able to hide in them, I dont think this is a good idea

This does not prevent hiding in them.

TobleroneSwordfish (changes requested):
Wouldn't this be better to be on `Entered` anyway so it activates when people are thrown in too?

PR #19659, opened 2024-06-23
Resprites icon state for capcoats

AXLFAN2000:
[clothing][sprites]
## About the PR 
Resprites the icon states for the capcoats coz they were off palette and the blue one being so different in shade always bugged me

## Why's this needed? 
image
image


ang0501:
I like the new look

frawhst (approved):
much better imo

PR #19661, opened 2024-06-23
Swap Kondaru AI upload and Book Nook

JORJ949:
## About the PR 
Swap around kondaru's AI upload and book nook, moving the gene booth to the new book nook
image
image

## Why's this needed? 
Kondaru upload is in an incredibly public area being opposite medbay means its incredibly easy to hear the turrets firing, this moves it to a less travelled (but still decent chance someone will notice) area

## Changelog 

```changelog
(u)JORJ949
(*)Kondaru AI upload and Book nook have traded locations
```

### DIFF
--- a/maps/kondaru.dmm
+++ b/maps/kondaru.dmm
@@ -3560,8 +3560,11 @@
 /turf/simulated/floor/plating,
 /area/station/crew_quarters/quartersB)
 "amB" = (
-/obj/table/reinforced/auto,
-/turf/simulated/floor/circuit,
+/obj/cable{
+	icon_state = "0-8"
+	},
+/obj/mapping_helper/wingrille_spawn/auto/reinforced/crystal,
+/turf/simulated/floor/plating,
 /area/station/turret_protected/ai_upload)
 "amC" = (
 /obj/machinery/light,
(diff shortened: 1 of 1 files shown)


pgmzeta:
The fire alarm south of the ai upload should be moved (probably one tile east?)
image

It's trivial to be fully protected from turrets by entering the rack room from the rear with just a set of tools and no supplies. Disassemble rwall, weld rods into sheet, place girder, safety guaranteed.
image

I don't think wires should path directly through the lawrack as it makes wire repair harder. Law racks often have a data terminal under them (even though it doesn't do anything).
image

I like that the gene booth is visible from both geneticist seats still. The book nook is a little sparse, maybe spawn a newspaper on the table?

pgmzeta:
the issue with the rear wall is a no-camera maint & single wall tile that isn't fired upon when being dismantled.

consider something like this, which uses an internal plasmaglass sensorline ala nadir's upload to provide a second layer of defense.
image
image

pgmzeta (changes requested):
move/remove this camera; otherwise looks good

camera-markup

pgmzeta (approved):

Kubius (approved):
Seems solid. Book nook is laid out a little weirdly but there's nothing wrong with it that I can specifically quantify enough to hold things up.

PR #19662, opened 2024-06-23
Resprites the Mining Ore Accumulator

Flaborized:
## About the PR 

Resprites the mining ore accumulator. Adds an overlay for when it has no target set (and thus cannot accept ore) and for when its cell is removed, because I wanted this to be a small little thing but got carried away.

both overlays on left

accumulator

## Why's this needed? 

I've seen this thing in the corner of my eye too many times, I don't like looking at it!! It no longer fits well into the game so I redrew it and I think this looks nicer. Some thoughts:

- The old design didn't much look like mining equipment; new one is meant to be industrial and heavy like mining stuff
- It's internally in the DMI called a gravgen but I recontextualized the sprite as being a spinny magnet because I thought that was a charming idea.
- It has antennas cuz it can teleport stuff to cargo pads. So antennas = teleport naturally
- It glows purple on the vents because I liked that part of the original and wanted to keep it somewhat
- it has tiny wheels at the bottom because you can drag it around
- it has a hole on the bottom, that's where the ore goes in. maybe it has a hole on all four sides!!! or maybe not idk
- nobody uses this thing but it should look good in its starting spawn position at least

here's what you could have if you actually wanted to use this thing (sprite in context also I tested new overlays and it works hooray)

accumulator

## Changelog 

```changelog
(u)Flaborized
(+)Resprited the mineral accumulator.
```


Flaborized:
possibly this icon might be used for some adventure zone I may have to go and fix that if it is (did not cross my mind to check when adding)

Gannets:
This is so much better

Gannets (approved):

TobleroneSwordfish (changes requested):
The `gravgen` sprite isn't used in any azones as far as I can see but it *is* used by some weird vestigial looking "gravity well generator". May as well leave the old sprite in to be used for that and rename the new ones to `ore_magnet` or something.

Flaborized:
> The `gravgen` sprite isn't used in any azones as far as I can see but it _is_ used by some weird vestigial looking "gravity well generator". May as well leave the old sprite in to be used for that and rename the new ones to `ore_magnet` or something.

Oh I didn't see this before committing, but yeah gravity well generator spawns in Sol afaik. I just redrew that sprite too since I really don't love the old sprite and renamed the accumulator stuff

gravgen

light design inspiration from Friend cogwerks idea, tried to keep this one more similar to the original design even though it's pretty different still

PR #19665, opened 2024-06-23
EFIF-1 Construction System for pods

Kubius:
[FEAT]
## About the PR 

efif_beeg

The **Extra-Facility Integrity Fabricator** (EFIF for short) is a pod-mounted tool capable of constructing and reconstructing walls and floors from the comfort of your cockpit, giving builders a substantial new avenue for such activities.

### Functionality

The EFIF is able to be installed on both miniature and full-size vehicles; 2x2 vehicles have a larger construction area, but 1x1 can be useful for working in tight spaces. In addition to its construction capability, it is **able to perform basic rock drilling**.

The EFIF can carry up to 200 steel sheets in its internal storage - only steel is accepted, as the system's internal mechanisms are calibrated for forming steel. These are loaded from the front of the pod by the tool's arms; manual loading is also possible.

There are two main categories of operation for the EFIF; **construction** and **repair**. **Construction** has three sub-modes - floor, reinforced floor, and wall - which the EFIF will attempt to construct on the tile(s) in front of the pod. **Repair** has a single mode, which attempts to match the original structure of the repaired area.

Construction must be done on suitable tiles (space, bare plating, asteroid floor, or regular floor in the case of walls).

At time of PR, the Construction mode **cannot build walls or non-reinforced floors on the station Z-level** as a precaution; this feature can be disabled, and I feel like it perhaps should be, but I've left it in for the moment.

These modes utilize an action bar for their construction, set up on the pod; the construction action will take more time for each constructed tile regardless of mode, based on the type of constructed tile, with regular floors being the fastest and walls being the slowest.

Additionally, the EFIF is equipped with a **Wide Field** toggle, which extends the construction area; this is a toggle to allow for fine-grained control of construction. While active, Construction's area extends one tile perpendicular from the targeted tiles in each direction, and Repair's area extends one tile further from the pod.

### Contents Breakdown

Bolded entries are "primary", containing major functional elements.

* code/datums/controllers/action_controls.dm: Changed action bar pixel X and Y offsets from a hardcoded value to a variable on the action bar datum which can be modified (in turn used for the EFIF's action bar position offset in 2x2 pods).

* code/datums/manufacturing.dm: Manufacture recipe for the EFIF-1. Can be made at the engineering specialist manufacturer (added to list in other file) for 5 pieces ultradense metal, 2 pieces claretine and 1 piece electrum.

* code/modules/economy/supply_packs.dm: EFIF-1 crate for purchase at QM. Costs PAY_DOCTORATE*15 (currently $9,000) and is cardlocked to Engineering. Comes fully stocked with steel sheets.

* code/modules/sound/soundCache.dm: Added new constructor work noise to the sound cache file.

* **code/modules/transport/pods/MainWeapon.dm**: The EFIF-1 implementation, including the tool itself, its construction field, and its action bar variant.

* code/modules/transport/pods/vehicle.dm: Added ability to feed steel sheets directly to the constructor by smacking the pod with them, for easy manual loading.

* code/modules/writing/papers.dm: Info paper that accompanies the EFIF-1 when bought from QM.

* code/obj/item.dm: stack_item contingency to allow sheets to stack to 200 inside EFIF-1.

* code/obj/machinery/manufacturer_subtypes.dm: Adds aforementioned manufacture recipe to the engineering specialist manufacturer.

* **code/turf/turf.dm**: Adds code to preserve the original turf path (a longstanding to-do, apparently!) and a function to replace the turf with its original turf path.

### Notes

New build sound adapted from [Freesound](https://freesound.org/people/craigsmith/sounds/675536/), CC0 license.

## Why's this needed? 

Constructing new buildings out in the asteroid field and repairing deeply-damaged areas of the station are both tasks which take enough time (and perhaps more importantly, have enough tedium) that they're often not done at all; this tool significantly improves players' capabilities for both, if they go to the trouble of acquiring it.

## Changelog 

```changelog
(u)Kubius
(*)The pod-mounted EFIF-1 Construction System is now available. It can rapidly repair damaged station tiles or build new offsite structures.
```


code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,563 @@
...
+	mats = list("metal_superdense" = 50, "claretine" = 20, "electrum" = 10)
+
+/obj/item/shipcomponent/mainweapon/constructor
+	name = "EFIF-1 Construction System"
+	desc = "A pair of elaborate robotic arms equipped for large-scale construction and asteroid demolition."
+	current_projectile = new/datum/projectile/laser/drill/cutter
+	appearanceString = "pod_weapon_efif"
+	icon_state = "constructor"
+	firerate = 12
+	var/mode = EFIF_MODE_DRILL
```
pgmzeta, line 413: should default to repair mode

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,563 @@
...
+	///Current count of loaded steel sheets (only accepts steel, as the system's metalforming is designed for it)
+	var/steel_sheets = 0
```
pgmzeta, line 415: start with 20 sheets?

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,563 @@
...
+	///Maximum allowable sheets loaded within tool
+	var/max_sheets = 200
+	///Secondary toggleable setting to increase build size (can be deactivated for fine work)
+	var/wide_field = FALSE
+
+	///Currently active construction fields
+	var/list/active_fields = list()
+
+	attack_self(var/mob/user)
```
pgmzeta, line 424: `attackby` to accept metal sheets loaded by hand?

pgmzeta (approved):
this is going to be super-strong with subs on oshan and nadir, i like it

code/modules/transport/pods/MainWeapon.dm
Kubius: Since it can be manufactured, I'd lean towards no for the baseline version. I could make a "loaded" subpath for the cargo purchase that starts with 100 or 200, though.

code/modules/transport/pods/MainWeapon.dm
Kubius: Done.

Kubius:
> given the invisible restrictions, maybe add a small paper sheet with the supply pack version

Did the thing.

code/modules/transport/pods/MainWeapon.dm
pgmzeta: could roll the steel cost into the fabrication cost, since you will (almost always) want sheets in this?

Kubius: Zero initial sheets means 4 regular-sized stacks or 1 QM-ordered megastack will cleanly fit into the device, and also, requiring 5+ steel bars would usually just mean running sheets through the reclaimer to then... turn them back into sheets again.

Kubius: QM version is now prestocked with 200 sheets.

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,608 @@
...
+				boutput(user, SPAN_ALERT("[src] only accepts steel sheets!"))
+				. = ..()
+
+	Fire(var/mob/user,var/shot_dir_override = -1)
+		switch(mode)
+
+			if(EFIF_MODE_FLOORS to EFIF_MODE_WALLS)
+				if(ON_COOLDOWN(src, "fire", firerate))
+					return
+				if(src.mode != EFIF_MODE_R_FLOORS && ship.z == 1) //antigrief
```
TobleroneSwordfish, line 460:
```suggestion
				if(src.mode != EFIF_MODE_R_FLOORS && ship.z == Z_LEVEL__STATION) //antigrief
```

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,608 @@
...
+			if(src.sheet_load_helper(O) >= LOAD_SUCCESS)
+				loaded_stuff = TRUE
+		if(loaded_stuff)
+			for (var/mob/M in ship)
+				M.playsound_local(ship, 'sound/machines/chime.ogg', 5, TRUE, ignore_flag = SOUND_IGNORE_SPACE)
+			ship.visible_message("<b>[ship]</b> loads metal sheets into its tool.")
+
+	///Sheet-loading proc for both automatic and manual loading. Set up for flag-based return, for responsive player messaging.
+	///Returns FALSE if provided object is unsuitable, LOAD_FULL (1) if storage is full, and/or LOAD_SUCCESS (2) if sheets were loaded.
+	proc/sheet_load_helper(var/obj/O,var/mob/user)
```
TobleroneSwordfish, line 546: Consider just using sheet stacking behaviour rather than reimplementing it here

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,608 @@
...
+		user.Browse(dat, "window=ship_main_weapon")
+		onclose(user, "ship_main_weapon")
+		return
+
+	Topic(href, href_list)
+		if(..())
+			return
+
+		if ((usr.contents.Find(src) || (in_interact_range(src, usr) && istype(src.loc, /turf))) || (issilicon(usr)))
+			src.add_dialog(usr)
```
TobleroneSwordfish, line 782: I don't think any of these are ever going to be true apart from the `issilicon` one and that probably isn't what you want?

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,608 @@
...
+		..()
+
+	onEnd()
+		..()
+		src.dense_object_refresh()
+		if (buildtool.steel_sheets < action_build_cost || !length(buildtool.active_fields))
+			interrupt(INTERRUPT_ALWAYS)
+			return
+		///Plural management for log formatting
+		var/multy = FALSE
```
TobleroneSwordfish, line 904: Typo?

code/turf/turf.dm
```diff
@@ -997,6 +1003,25 @@ var/global/in_replace_with = 0
 				W.UpdateIcon()
 	return wall
 
+///Returns a turf to its initial path, if it is a simulated floor or wall. Returns FALSE if requested turf was not one of these initially.
+/turf/proc/ReplaceWithInitial()
+	var/oldpath = src.path_old
+	if(ispath(oldpath,/turf/simulated/floor) || ispath(oldpath,/turf/simulated/wall))
```
TobleroneSwordfish, line 1009: This check feels like it should be outside the proc, otherwise it doesn't really do what it says it does.

code/datums/manufacturing.dm
```diff
@@ -3081,6 +3081,16 @@ ABSTRACT_TYPE(/datum/manufacture/pod/weapon)
 	time = 20 SECONDS
 	category = "Tool"
 
+/datum/manufacture/pod/weapon/efif1
+	name = "EFIF-1 Construction System"
+	item_requirements = list("metal_superdense" = 50,
+							 "claretine" = 20,
+							 "electrum" = 10)
```
TobleroneSwordfish, line 3088: These should not be specific material requirements

TobleroneSwordfish (changes requested):
Neat feature, extremely unintuitive to use:
- The pod computer menu is eldritch knowledge I would bet 90% of the playerbase would give up before ever finding. The button to open it should really just be on the top bar, I know this is out of scope buuut also it's close to unusable without it so 🤷 
- Sheet loading requires either finding a button in the eldritch menu or ejecting the part, consider just letting people hit the pod with a stack of sheets (see `COMSIG_ATTACKBY`)
- Sheet loading doesn't work properly for 2x2 pods
- The disassembly mode doesn't appear to do anything and gives no error message

code/datums/manufacturing.dm
Kubius: They are specific material requirements for a particular purpose - using conductive_high cannot distinguish between claretine and electrum, and electrum is a deliberate step to increase complexity of production.

code/modules/transport/pods/MainWeapon.dm
Kubius: Mult-y - has the quality of multipleness. I liked it because it's a little silly, but if something more formal is preferred I'm not that attached to it.

Kubius:
WRT pod computer menu: I do agree that it's Not Great and the likelihood it's out of scope might be enough reason for me to make a distinct PR to unhose pod UI and then merge it into this.

WRT sheet loading: 2x2 pods do load properly but there is definitely an intuitive operation issue (it only loads with the left arm right now), and I can absolutely make it so both arms can load sheets + look into the hit-ship-to-load-sheets approach. I do want the ability to pick up sheets without having to leave the pod.

WRT Disassembly mode: Mining or Drilling mode is probably a clearer name for it.

code/turf/turf.dm
Kubius: Moved ReplaceWith outside check so it can initial-ify anything, and put another check on the tool (check in the proc is still relevant for post-placement adjustment).

code/modules/transport/pods/MainWeapon.dm
Kubius: Not entirely sure what you mean by "just using" - I took a peek at sheet stacking behavior and it has a lot of code not relevant to this functionality, unless I'm looking in the wrong spot, and I do have some return code stuff for load message formatting woven into the implementation here.

code/modules/transport/pods/MainWeapon.dm
Kubius: Cargo culted this one from the industrial utility arms. im shame

Removing it doesn't appear to have created any issues, so I'll probably do that.

code/modules/transport/pods/MainWeapon.dm
TobleroneSwordfish: I think you could just store a stack of sheets and call `stack_item` on them?

Kubius: Was a little more complex than that, but done.

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,624 @@
...
+			if (outcome & LOAD_FULL)
+				boutput(user, SPAN_NOTICE("[src] is[outcome & LOAD_SUCCESS ? " now" : null] fully loaded."))
+			if (outcome < 1)
+				boutput(user, SPAN_ALERT("[src] only accepts unreinforced steel sheets!"))
+				. = ..()
+
+	//Stopgap discoverability feature, pending pod UI improvement
+	activate()
+		. = ..()
+		if(ship.pilot)
```
TobleroneSwordfish, line 456: This causes a runtime on trying to insert the constructor component into a pod

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,624 @@
...
+			if(EFIF_MODE_R_FLOORS)
+				return EFIF_R_FLOOR_BUILD_TIME
+			if(EFIF_MODE_WALLS)
+				var/space_adjustment = EFIF_WALL_BUILD_TIME
+				if(istype(get_turf(F),/turf/space))
+					space_adjustment += EFIF_FLOOR_BUILD_TIME
+				return space_adjustment
+
+	///Check whether dense objects are present in build fields; interrupt any fields which have been blocked
+	proc/dense_object_refresh()
```
TobleroneSwordfish, line 865: Maybe this check should only happen for walls? It's a bit odd not to be able to repair floors because there's an unexploded pod/fuel tank over where they should be.

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,624 @@
...
+		..()
+		//didn't work right for some reason
+		/*
+		if(pod.bound_height == 64)
+			bar.pixel_x += 16
+			bar.pixel_y += 16
+			border.pixel_x += 16
+			border.pixel_y += 16
+			updateBar()
+		*/
```
TobleroneSwordfish, line 885: Not sure what this is trying to do, but if it doesn't work then remove it.

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,624 @@
...
+			return
+		///Plural management for log formatting
+		var/multy = FALSE
+		if (length(buildtool.active_fields) > 1)
+			multy = TRUE
+		if(duration > 2 SECONDS)
+			playsound(owner, 'sound/machines/constructor_work.ogg', 30, 1)
+
+		///Flag set for log reporting; 1 is floors, 2 is walls
+		var/build_types = 0
```
TobleroneSwordfish, line 923: This var is both inventive and causes me physical pain to look at, 10/10 gooncode

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -380,6 +380,624 @@
 		opencomputer(usr)
 		return
 
+// Extra-Facility Integrity Fabricator: the big boyo pod construction tool
+
+//mode defines for tool
+#define EFIF_MODE_DRILL 0
```
TobleroneSwordfish, line 386: Is there a reason why these can drill too? It seems a little out of place and these are QM buyable compared to the regular mining drills requiring bohrum and uqil.

TobleroneSwordfish (changes requested):
The console popup is a good compromise, code seems generally okay.

code/modules/transport/pods/MainWeapon.dm
Kubius: Lets them work effectively for asteroid field construction (one of the fun use cases I had in mind). I can dial back the fire rate to something similar to the industrial utility arms or up the QM price, if you like.

code/modules/transport/pods/MainWeapon.dm
Kubius: Good thought. Will take a poke at adjusting that.

code/modules/transport/pods/MainWeapon.dm
Kubius: screm

Will investigate.

code/modules/transport/pods/MainWeapon.dm
Kubius: Adjusted.

code/modules/transport/pods/MainWeapon.dm
Kubius: Think I have it fixed. Verifying ingame.

Kubius: Yep!

code/modules/transport/pods/MainWeapon.dm
Kubius: Now figured it out.

TobleroneSwordfish (approved):

PR #19670, opened 2024-06-24
Cyborg docking station click-drag range buff

JORJ949:
## About the PR 
Before running the distance check for click-drag validity check on the cyborg dock if whats being click-dragged is a mob, if it is use current range, if its not use old range

## Why's this needed? 
This change was made so you couldn't click-drag people at long distance into the conversion chamber, however it removed the ability to put clothes and upgrades in a dock at a further distance, which was nice QOL for borgs.

## Changelog 

```changelog
(u)JORJ949
(+)Objects can now be click-dragged into cyborg docks from further away. Mobs still require the shorter distance.
```


frawhst (approved):
I would rewrite the changelog to make it clear that this *only* applies to objects, not mobs. Otherwise, fine change imo.

mordent-goonstation (approved):

PR #19673, opened 2024-06-24
Removes floorpills from the brig

TheTyrant25:
[MAPPING]
## About the PR
Removes floorpills bottles from the bathrooms of the donut2 and cogmap1 brigs.

## Why's this needed?
It's not present on the other maps and causes stirstir to think "yum yum" and go into anaphylactic shock. Not ideal.

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -45200,7 +45200,6 @@
 /obj/item/storage/toilet{
 	dir = 4
 	},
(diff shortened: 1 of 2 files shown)


pgmzeta (approved):
makes sense to remove this as security is now scored based on Stir's survival

AXLFAN2000:
stirstir has enough going on in his life he doesn't need floorpills

PR #19676, opened 2024-06-24
FIX: Manufacturer UI comparison check not using predicted material amounts

### LINKED ISSUE #19577: Fabricator doesn't produce even if it has materials
### Labels

[BUG][MAJOR]

### Description

I tried producing a Helios Mark 2 engine and had loaded up all the materials necessary but the fabricator refused to construct it saying that it had insufficient materials

### Steps to reproduce

1 Load necessary materials in ship component fab
2 try to produce Helios Mark II engine

### Expected Behavior

The fabricator should have built the engine. 

I don't know if this is recreatable, but it produces everything else fine

Reported by: omegazip2
Client version: 515.1637
Active test merges: #19460 and #19331
Round log date: 2024-06-18-13-49
Reported on: 2024-06-18 14:19:44
Round ID: 181733
Map: KONDARU

MeggalBozale:
## About the PR 
The list specifically for preventing issues with checking amt for a material used twice was not used due to some oversight, now it is.

Resolves the issue where having, say, 2 steel in a fab, then having a recipe that uses 2 steel and 0.5 steel would look good, even though it refuses to print.

## Why's this needed? 
Fixes #19577


MeggalBozale:
woops one moment

PR #19677, opened 2024-06-24
Re adds the two pill bottles in different places

TheTyrant25:
[MAPPING] [QOL] [BALANCE] (allegedly)
## About the PR
The sequel to #19673

Puts the removed bottles somewhere nearby in the maints but not inside the brig.

## Why's this needed? 
SOMEONE was complaining. Also there aren't that many of these so they probably shouldn't just get removed.

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -6988,6 +6988,7 @@
 /area/station/storage/emergency)
 "aAG" = (
 /obj/machinery/drone_recharger,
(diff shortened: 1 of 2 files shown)


TealSeer (approved):

TealSeer:
emoji

TheTyrant25:
you.

TobleroneSwordfish (approved):
Floorpills must be neither created nor destroyed.

PR #19678, opened 2024-06-24
Changes the max stack for pipeshot from 4 to 8

NightmareChamillian:
## About the PR 
Quite literally a one-line change that makes it so you can bunch two pipeshot stacks together. Since I was in there I also made the descriptions a little less clunky and it's not really worth atomizing out. (I still can if need be)

## Why's this needed? 

Honestly, this is a QoL change more than anything. Every other shotgun shell type starts out as a stack of either 7 or 8 shells, and since you can already use the unload-reload trick to make stacks of 8 pipeshot, there's no real point in not having it be an intentional feature.  

The only issue I can think of is balance reasons, but I doubt it's that big of a concern; anyone with access to a shotgun that isn't a slamgun or double-barrel is already capable of stacking pipeshot beyond 4.

## Changelog 

```changelog
(u)NightmarechaMillian
(+)Max pipeshot stack 4 -> 8
```

### DIFF
--- a/code/obj/item/gun/ammo.dm
+++ b/code/obj/item/gun/ammo.dm
@@ -827,26 +827,26 @@ ABSTRACT_TYPE(/obj/item/ammo/bullets/pipeshot)
 	ammo_type = new/datum/projectile/special/spreader/buckshot_burst
 	icon_state = "makeshiftscrap"
 	amount_left = 4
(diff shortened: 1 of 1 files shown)


NightmareChamillian:
seems like a no-go; closing

PR #19682, opened 2024-06-24
Replaces the syndicate's Sawfly purchase with the pouch containing three

NightmareChamillian:
## About the PR 

See title- changes the buylist purchase. 

If need be, I'm fine with changing it to be a 1 TC purchase per drone, but thought I'd see what folks think of the pouch idea first.

## Why's this needed? 

There's been some talk in RP-chat about how sawflies are a little underwhelming for 2 TC, especially since all the other grenades you can purchase are 1 TC and give a whole pouch of them. 

When sawflies were first added, 2 TC made sense. They pierced armor, had considerably faster attacks, and had shorter fuze times. I definitely think the changes were merited, but since then, I feel like they've dropped out of the 2 TC price bracket.

The purchase giving a pouch of sawflies instead of one and a remote is more in-line with the other grenades, and should hopefully legitimize their use outside of the "let's make an army in mechlab" niche.

## Changelog 

```changelog
(u)NightmarechaMillian
(*)The 2 TC sawfly purchase will now provide a pouch of 3 drones. Yes, this is the one nukies get.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -203,11 +203,11 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/generic)
 	desc = "A pack of Syndicool Lights exploding trick cigarettes. Due to the use of a military-grade explosive, please do not attempt to smoke these after lighting."
 
 /datum/syndicate_buylist/generic/sawfly
-	name = "Compact Sawfly"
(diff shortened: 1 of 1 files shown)


444Portal (approved):

444Portal:
see imcoder ( https://discord.com/channels/182249960895545344/890226559691157524/1255713886590603386 ) for my thoughts

PR #19685, opened 2024-06-24
fix sawfly remote mechscan

pgmzeta:
[bug][internal]
## About the PR 
Use "conductive" instead of "CON-1" for sawfly remotes
Removes a global list of material properties to stop future confusion (hopefully nothing in secret uses this???)

## Why's this needed? 
Follow up to #19681 using current material definition system

### DIFF
--- a/_std/defines/materials.dm
+++ b/_std/defines/materials.dm
@@ -22,32 +22,6 @@
 /// Global static list of rarity color associations
 var/global/static/list/RARITY_COLOR = list("#9d9d9d", "#ffffff", "#1eff00", "#0070dd", "#a335ee", "#ff8000", "#ff0000")
 
-/// Material category names as displayed in fabricators
-/// see match_material_pattern() for exact definitions
-var/global/list/material_category_names = list(
-	"ALL"   = "Any Material",
-	"CON-1" = "Conductive Material",
-	"CON-2" = "High Energy Conductor",
-	"CRY-1" = "Crystal",
-	"DEN-1" = "High Density Matter",
-	"DEN-2" = "Very High Density Matter",
-	"CRY-2" = "Extraordinarily Dense Crystalline Matter",
-	"FAB-1" = "Fabric",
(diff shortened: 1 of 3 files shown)


NightmareChamillian:
wasn't aware this was an issue 🫡

_std/defines/materials.dm
```diff
@@ -22,32 +22,6 @@
 /// Global static list of rarity color associations
 var/global/static/list/RARITY_COLOR = list("#9d9d9d", "#ffffff", "#1eff00", "#0070dd", "#a335ee", "#ff8000", "#ff0000")
 
-/// Material category names as displayed in fabricators
```
NightmareChamillian, line 25: Uh oh! I don't think this is supposed to happen.

code/atom.dm
```diff
@@ -475,8 +475,8 @@ TYPEINFO(/atom)
 	..()
 
 TYPEINFO(/atom/movable)
-	/// Either a number or a list of the form list("MET-1"=5, "erebite"=3)
-	/// See the `match_material_pattern` proc for an explanation of what "CRY-2" is supposed to mean
+	/// A key-value list of match property or material IDs and an amount required to construct the item
```
NightmareChamillian, line 478: Ditto here

code/obj/sawflymisc.dm
```diff
@@ -96,8 +96,8 @@ TYPEINFO(/obj/item/old_grenade/sawfly)
 			icon_state_armed = "clusterflyB1"
 
 // -------------------controller---------------
-TYPEINFO(/obj/item/old_grenade/sawfly)
-	mats = list("CON-1"=2)
+TYPEINFO(/obj/item/remote/sawflyremote)
+	mats = list("conductive"=2)
```
NightmareChamillian, line 100: 👍

NightmareChamillian (changes requested):

PR #19688, opened 2024-06-25
pathology code removal

pgmzeta:
[internal][removal]
## About the PR 
Remove pathology code / defines
Replace pathology machines with fake objects & update z2/z3 and **unused maps**
Move statue code to code/obj/statue.dm

A few notable highlights:
* Removes the synth-o-matic modules from the debris field spawn
* pathology is now always medical access (Manta's patho lab has been set to research access)
* Pathology is now a special job, so it can still be custom spawned (for zombie rounds? or whatever)
* The pathology glassware still exists, it just doesn't really do anything special

## Why's this needed? 
pathology super dead
image


ZeWaka:
thank goodness i can rest

PR #19693, opened 2024-06-25
moves map votes to the end of the round

Xkeeper0:
## About the PR

this PR does the following things:
* the automated map vote is __moved from pre-round to the end of the round__.
* the map vote time is shortened from 7 minutes (420 sec) to 60 seconds.
* the end-of-round countdown time is increased from 60 to 120.
* the map vote link in chat is now a command alert with a much larger link.

in addition, if the next map has already been set (`mapSwitcher.next` non-null), the map vote will be skipped. this allows an admin (or other event?...) to set the map without worrying that the map vote will end up overriding it.

----

**note**: this change is based on an assumption that the build server can build and deploy the next round's code within a bit over a minute; current estimates are that it generally finishes within 30 seconds, though this is also dependent on if other servers are active.

a future improvement involves delaying the end of a round until receiving a "build ready" signal from the build server (or running out of time and going ahead anyway)

## Why's this needed?

map votes typically have low turnout, and determine the fate of a round anywhere from 50 to 100 minutes (or more) in the future. this allows votes to happen by people who will likely be around for the next round, and also makes the vote more visible.

## Changelog
```changelog
(u)Zamujasa
(*)Map votes now happen after the round is over, when the countdown starts.
(*)Map votes are now 60 seconds.
(*)End-of-round time is now 120.
```

### DIFF
--- a/code/datums/gameticker.dm
+++ b/code/datums/gameticker.dm
@@ -56,9 +56,6 @@ var/global/current_state = GAME_STATE_INVALID
 	pregame_timeleft = 1
 	#endif
 
-
-
-	var/did_mapvote = FALSE
 	var/did_reminder = FALSE
 
 	#ifdef LIVE_SERVER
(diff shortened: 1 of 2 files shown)


DisturbHerb:
but i can't bribe my way to getting the mapvote results i want 🥺

ZeWaka:
I think this change is a step in the right direction, we just need to be careful about server crap. Will review closely with thought this week for sure

Xkeeper0:
> but i can't bribe my way to getting the mapvote results i want 🥺

you can still bribe the map vote, you just have to do it when it happens. time to put a voting booth at Centcom and the afterlife bar!


> I think this change is a step in the right direction, we just need to be careful about server crap. Will review closely with thought this week for sure

aye. it's testmerged on 1 at the moment to see if the concept is working. the time can always be shortened to 30 seconds (= 90 for build time) as well

Xkeeper0:
60 seconds was not enough time for it to map change on main1 in the first round i tested it, so maybe 30 seconds will work

PR #19694, opened 2024-06-25
Capsule machines stand back up when they're fixed

### LINKED ISSUE #18901: Capsule machine sprite does not change when pryed back up
### Labels

[BUG]

### Description

When prying back up a capsule machine, the sprite of the machine does not fix itself and stays on the fallen down sprite

### Steps to reproduce

1. Throw someone into a capsule vendor
2. Pry the capsule vendor back up

### Expected Behavior

For the capsule vendor to stand upright

image

pgmzeta:
[bug][game objects]
## About the PR 
add a base image state for capsule machines (since they set their icon state on `create_products`)
Then, on `fall()`/`right()` use that base icon state to set the correct standing/fallen state (since they don't use power and thus have `power_change()` no-opped).

## Why's this needed? 
Fix #18901

### DIFF
--- a/code/modules/items/gimmick/figurine.dm
+++ b/code/modules/items/gimmick/figurine.dm
@@ -1120,13 +1120,15 @@ ABSTRACT_TYPE(/datum/figure_info/patreon)
 	icon_state = "machine1"
 	icon_panel = "machine-panel"
 	var/sound_vend = 'sound/machines/capsulebuy.ogg'
+	var/base_icon_state = "machine1"
 	var/image/capsule_image = null
 
 	create_products(restocked)
 		product_list += new/datum/data/vending_product(/obj/item/item_box/figure_capsule, 35, cost=PAY_UNTRAINED/5)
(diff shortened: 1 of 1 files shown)


frawhst (approved):

code/modules/items/gimmick/figurine.dm
```diff
@@ -1120,13 +1120,15 @@ ABSTRACT_TYPE(/datum/figure_info/patreon)
 	icon_state = "machine1"
 	icon_panel = "machine-panel"
 	var/sound_vend = 'sound/machines/capsulebuy.ogg'
+	var/base_icon_state = "machine1"
```
ZeWaka, line 1123: why can't we use initial()

pgmzeta: in create_products the icon state was set with
```src.icon_state = "machine[rand(1,6)]"```, so initial would always have it revert to `machine1`

ZeWaka (approved):

PR #19718, opened 2024-06-26
Hudzone-ify pod UI

Kubius:
[VEHICLES][INTERNAL][CLEAN]
## About the PR 

Converts some pod UI that was being manually positioned to use the hudzone system.

- [x] Move top bar of UI into a hudzone
- [ ] Get top-bar hudzone to exist in the same space that manual offsets placed it
- [ ] Add buttons that automatically generate under each top-bar element that is associated with a ship component, to access its ship computer entry
- [ ] Restructure creation of top-bar HUD elements to use a helper proc or other means of reducing bulk repetition (?)
- [ ] Move Leave Pod button into a hudzone (?)
- [ ] Move tracking indicator into a hudzone, adjust references from outside of HUD datum (?)
- [ ] Move sensor lock indicator into a hudzone, adjust references from outside of HUD datum (?)
- [ ] Clean up orphaned variables from the previous revision 

Graphic for the ship component settings button:

more_arrow

## Why's this needed? 

Moving old UI into a new, less janky system with improved capabilities.


Kubius:
Closing pending further collaboration

PR #19720, opened 2024-06-27
Projectiles have a chance to  pass through girders

pgmzeta:
[balance][game objects]
## About the PR 
Give girders a chance to let projectiles pass right through them, with the rate dependent on how secure the girder is (displaced girder > girder > reinforced girder).

## Why's this needed? 
Girders can be used as a bullet-proof shield, which feels a bit wrong because you can literally see through them.

## Demo
Firing test with the Antares LMG (Nukies heavy machine gun). From left to right: Displaced Girders, Girders, and Reinforced Girders. Fired 1 LMG Belt behind the center test dummy for each set of girders.
https://github.com/goonstation/goonstation/assets/91498627/e43d83e1-4e3c-4593-a9c1-1bc78a0b8fb4

## Changelog 

```changelog
(u)glowbold
(*)Projectiles have a chance to pass through girders. The more secure a girder is, the less likely projectiles can travel through them.
```

### DIFF
--- a/code/obj/structure.dm
+++ b/code/obj/structure.dm
@@ -7,6 +7,7 @@ obj/structure
 		density = 1
 		material_amt = 0.2
 		var/state = 0
+		var/projectile_passthrough_chance = 50
 		desc = "A metal support for an incomplete wall."
 		HELP_MESSAGE_OVERRIDE({"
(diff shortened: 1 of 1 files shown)


Tarmunora (approved):

Gannets (approved):
neat!

PR #19722, opened 2024-06-27
Adds Chameleon Bombs to Mail Courier uplinks

cheekybrdy:
## About the PR 
Adds Chameleon bombs to the mail courier's uplink

## Why's this needed? 
Its a unique item only available to the clown and spiefs but works very well with what a mail courier is intended to do.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Adds chameleon bombs to mail courier's options as a traitor.
```


Tarmunora (approved):
I feel like some kind of more dedicated mail-bomb thing might be a better fit but i like this too

PR #19723, opened 2024-06-27
Cap/HoS spawn with their eguns in their bags

JORJ949:
## About the PR 
Moves the Captain and HoS Energy gun spawns into their backpacks rather than their lockers

Just a sidenote that with Short-sighted + NT loyalist I spawned with a full backpack as HoS, may need to remove the flash if there are any trait combos that can give more items. Unless there is a way for those trait items to spawn in hands/glued to them.

## Why's this needed? 
Captains / HoSes latejoining to their gun already stolen isn't great, if a role isn't in the game their gear shouldn't be, also allows HoS latejoiners to have something to deal with antagonists that are already in security, preventing them from redeeming their token. And in addition newer captains can forget their gun in their office (probably not knowing it exists), this gives it to them by default.

## Changelog 

```changelog
(u)JORJ949
(*)Captains and HoSes now spawn with their Energy guns
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -257,7 +257,7 @@ ABSTRACT_TYPE(/datum/job/command)
 	slot_eyes = list(/obj/item/clothing/glasses/sunglasses)
 	slot_ears = list(/obj/item/device/radio/headset/command/captain)
 	slot_poc1 = list(/obj/item/disk/data/floppy/read_only/authentication)
-	items_in_backpack = list(/obj/item/storage/box/id_kit,/obj/item/device/flash)
+	items_in_backpack = list(/obj/item/storage/box/id_kit,/obj/item/device/flash,/obj/item/gun/energy/egun/captain)
 	rounds_needed_to_play = 30
 
 	derelict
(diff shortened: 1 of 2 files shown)


JORJ949:
> personally not for this, having your stuff stolen should remain a fact of life 🥺

I agree that having your stuff stolen should remain, however having it stolen _before you even join_ isn't particularly fun.

Glamurio:
Not necessary I think. Captains spawn on round-start, in fact the game demands it, so this is really primarily a HoS only issue. And HoS doesn't need the e-gun, they have a token and armory access to grab whatever they want, if someone steals their e-gun it's honestly not a big deal and if anything leads to a fun mystery of finding out who did it.

And on the same note, yes, captain's locker should be desirable to steal, and removing the gun does not add to it.

pgmzeta:
100% for this for captains. Cap lockers are strong enough without Free Gun, especially when it's only going to be in there *when there is no captain*. HoS has a token and can get into the armory, not needed as badly.

Retrino:
meh, hos still gets a token and The Entire Armory so i think theyll live if the egun is stolen. rounds sometimes have no captain so i think their stepin deserves a gun for the bravery

mralexs5:
This is a change that many HoS players (myself included) have been asking for for a long time. Just because they _can_ access the armory doesn't they will. The armory is for when the station is in grave danger, not "Someone stole my entire locker for funsies before I joined."

Glamurio:
> This is a change that many HoS players (myself included) have been asking for for a long time. Just because they _can_ access the armory doesn't they will. The armory is for when the station is in grave danger, not "Someone stole my entire locker for funsies before I joined." Also IIRC most jobs have their gear spawn on them, so why not the HoS and Cap?

HoS has access to the armory precisely for that reason. You can go inside without flipping the loud alert button. And an antagonist stealing the HoS locker imo qualifies as a decent enough danger to warrant grabbing a phaser or pulse rifle from the rack.

TobleroneSwordfish:
Having things to steal as a traitor is good, see also Glamurio's comment.

PR #19732, opened 2024-06-27
Change combustion furnace's name and description so it's not LYING to you

NibChocolateeny:
[BUG][TRIVIAL][OBJECTS][ADD TO WIKI]
## About the PR
charnges name folr the "5 kW" furnace to "20 kW" furnace in name and desc. There was change from 5 to 20 a long time ago from @Xkeeper0 but name was not made right! now it is right. :>

also fix some (2) line lengths so they are not so long, like code guide line says. does not change code workingness, I think.

## Why's this needed?
No lying! Lying is for the TRAITORS.

### DIFF
--- a/code/modules/power/furnace.dm
+++ b/code/modules/power/furnace.dm
@@ -2,8 +2,9 @@ TYPEINFO(/obj/machinery/power/furnace)
 	mats = 20
 
 /obj/machinery/power/furnace
-	name = "Zaojun-2 5kW Furnace"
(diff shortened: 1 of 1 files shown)


code/modules/power/furnace.dm
```diff
@@ -12,7 +13,7 @@ TYPEINFO(/obj/machinery/power/furnace)
 	var/fuel = 0
 	var/last_fuel_state = 0
 	var/maxfuel = 1000
-	var/genrate = 20000
+	var/genrate = 20000 // previously 5kW
```
mordent-goonstation, line 16:
```suggestion
	var/genrate = 20000
```
Code history is there if needed, no need for a comment here.

PR #19742, opened 2024-06-28
mail fix + clown announcement PC on mushroom

cheekybrdy:
## About the PR 
Title

## Why's this needed? 
working mail good and up to date maps good too


TobleroneSwordfish:
That mail chute doesn't look right, what was wrong with it before?

cheekybrdy:
> That mail pipe doesn't look right, what was wrong with it before?

accidental place on that one in genetics, there was no pipe under the left MD office door

TobleroneSwordfish (changes requested):
Looks like you're now reverting the freezer PR, please check the diff render.

PR #19754, opened 2024-06-29
Cluwneshot

RubberRats:
[GAME OBJECTS] [FEATURE]
## About the PR 

This PR "adds" a cluwneshot mode to the Lawbringer. In reality, it's a version of the lethal kinetic mode with a different VOX line.

## Why's this needed? 

The appropriate response to a cluwne is putting them out of their misery. But mostly, I think this is funny.

## Changelog 

```changelog
(u)RubberRats
(+)The Lawbringer has been given a new (?) mode for use against the cluwne
```


ZeWaka (approved):
lol why not

PR #19759, opened 2024-06-29
Door button mapping helpers

Kubius:
[MAPPING]
## About the PR 

Adds a new set of door button mapping helpers, including two that significantly improve the mapping experience when used in appropriate circumstances. These apply an ID to /obj/machinery/door_control, /obj/machinery/r_door_control, /obj/machinery/door/airlock or /obj/machinery/door/poddoor.

image

- The area-name door button helper applies an ID created by using ckey() on the area name. Useful for hangars where you have one big set of buttons and doors.
- The pair door button helper can be placed on a door button, then pointed in any direction. It creates an ID using the text "AUTO" and its X and Y coordinates, and then applies it to both the button on the helper, and the airlock or poddoor that's adjacent in the direction it's pointing. Useful for simple door-button pairs, such as bathroom stalls or crew quarters rooms.
- The basic door button helper simply uses a var-edited ID and applies it to matching objects. Not as expedient as the other two, but an option if you want to split variables off into more readable objects.

## Why's this needed? 

Significantly simplifies the process of setting up button-controlled doors in several common cases.


Kubius:
Added to mapping guide.

ZeWaka:
!merge_upstream

PR #19761, opened 2024-06-30
Resprites station lights and gives them an overlay! (Thank you Herb)

TemThrush:
[Game Objects]
## About the PR 
Along with new sprites for bulbs, tubes, and floor-lights, this PR also gives them a separate bloom overlay on top of the sprites that doesn't burn your hand! (Everyone say thank you, Herb)

image

[All the new sprites laid out with the current sprites for comparison]

image

[Ominous room with as many icon states as I can fit]

https://github.com/goonstation/goonstation/assets/108035657/c19f2a08-55d3-4cf5-9426-d8b0950a0a6a

[Video show-casing the emergency light animation]

## Why's this needed? 
Old sprites no longer meets current spriting standards. Plus I have a personnel vendetta against the nearly invisible pixels you could burn your hand on.


DisturbHerb:
only took me like two months to finish this in an afternoon 🎉

Xkeeper0:
The burnt out ones don't really convey their burnt-out-ness; they just look like the ends are smudged, which isn't quite as clear as the entire-thing-looks-grimy-and-burnt style the current ones have.

i also think the off floor lights have too much shading, but eh, personal preference. i think of them as very smooth pot-light style bulbs

Gannets:
I liken these

LaCumbiaDelCoronavirus:
In my opinion, i (also) feel like the tube and floor lights look 'overshaded' or something of the sort. The floor lights also seem very yellow, i think that they'd be better in the same shade of yellow as the original ones.
Other than that, they all look very nice!

ZeWaka:
> The burnt out ones don't really convey their burnt-out-ness; they just look like the ends are smudged, which isn't quite as clear as the entire-thing-looks-grimy-and-burnt style the current ones have.

to be fair that's somewhat what burnt out flouroscent tubes look like, but i think  in this case they could be conveyed by a bit more interior soot yeah

code/modules/power/lighting.dm
```diff
@@ -200,6 +200,8 @@ ADMIN_INTERACT_PROCS(/obj/machinery/light, proc/broken, proc/admin_toggle, proc/
 	var/datum/light/point/light
 	var/install_type = INSTALL_WALL
 
+	var/obj/dummy/light_overlay // Light overlay object to place in `src.vis_contents`
```
ZeWaka, line 203: clear this on disposing

ZeWaka (changes requested):

Azrun (commented):
Like it once disposing resolved.

ZeWaka:
i think the burnt out ones are a vast improvement over the shitty existing ones, we can tweak further if we want

PR #19763, opened 2024-06-30
Move Nukie field medic satchel supplies to the pouch

JORJ949:
## About the PR 
Removes the field medic nuclear operative's medical satchel and puts the defibrillator and fire extinguisher into the pouch, also lets the pouch hold them

## Why's this needed? 
The satchel is a straight downgrade to the bag you spawn with, essentially just making it a trap for new players

### DIFF
--- a/code/obj/item/storage/ammo_pouches.dm
+++ b/code/obj/item/storage/ammo_pouches.dm
@@ -178,8 +178,12 @@
 	w_class = W_CLASS_TINY
 	slots = 4
 	opens_if_worn = TRUE
+	can_hold = list(/obj/item/robodefibrillator,
+	/obj/item/extinguisher/large)
 	spawn_contents = list(/obj/item/reagent_containers/mender/brute/high_capacity,
-	/obj/item/reagent_containers/mender/burn/high_capacity)
+	/obj/item/reagent_containers/mender/burn/high_capacity,
+	/obj/item/robodefibrillator,
+	/obj/item/extinguisher/large)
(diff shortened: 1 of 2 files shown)


pgmzeta:
why not make the satchel bigger instead?

JORJ949:
> why not make the satchel bigger instead?

Because whats the point in giving them a second bag when they already have one?

pgmzeta:
> Because whats the point in giving them a second bag when they already have one? 

for the team to identify which of the nukies will stand them up after they RPG themselves

> If the pouch feels weird I could give them a medical belt instead for them

pouch holding 2 items that don't fit into pockets definitely feels weird to me, I think a belt seems reasonable (they could still hold the fire extinguisher in their pack)

JORJ949:
> pouch holding 2 items that don't fit into pockets definitely feels weird to me, I think a belt seems reasonable (they could still hold the fire extinguisher in their pack)
I imagine it as 2 clips on the side of the pouch that clip them on

TobleroneSwordfish (changes requested):
I agree with zeta, putting large items in pouches like that feels weird, may as well just increase the capacity of the satchel.

JORJ949:
I considered giving them a medical belt and removing the pouch altogether, or just putting defib qnd extinguisher loose for use in bag, I don't see much point in giving them the exact same bag tbh, if they want a satchel they can just go get one from the post

JORJ949:
Closing to make a better solution in the future

PR #19767, opened 2024-06-30
Skeletons no longer duplicate their clothes on death + their limbs correctly get thrown

### LINKED ISSUE #19764: Item duplication through skeleton death
### Labels

[BUG][CRITICAL]

### Description

Skeleton died, gibbed, torso remained with items on it but the items were also thrown everywhere during the skeleton gib.  Now we're all filled with cursed items the same as when there used to be the dupe from people picking stuff up at the same time

### Steps to reproduce

1.  I was a skeleton and was killed
2.  I skeleton gibbed, but my torso with items also dropped?
3.  Antags looted the gibbed items, I got the ones from the torso back when I was cloned.  Now we're all broken and can't access most of our items/UI

### Expected Behavior

Not that

Reported by: joeled
Client version: 515.1640
Active test merges: #19063
Round log date: 2024-06-30-14-59
Reported on: 2024-06-30 16:06:24
Round ID: 182404
Map: OSHAN

CalliopeSoups:
[RESPAWNING] 
## About the PR 
A previous PR (#19496) intended to use ThrowRandom for limbs, but the range was set to 1 so the limbs would never actually get thrown on death. PR also removes code that drops clothing to prevent duplication.

## Why's this needed? 
Fixes #19764 and fixes bug left by previous PR.

### DIFF
--- a/code/mob/living/carbon/human.dm
+++ b/code/mob/living/carbon/human.dm
@@ -350,7 +350,7 @@
 		if (r_leg && istype(r_leg, /obj/item/parts/human_parts))
 			r_leg:set_skin_tone()
 
-	proc/sever(var/target = "all", var/mob/user)
+	proc/sever(var/target = "all", var/mob/user, var/messy=TRUE)
 		if (!target)
 			return 0
 		if (istext(target))
@@ -372,11 +372,11 @@
 					limbs_to_sever += list(src.r_leg)
 			if (length(limbs_to_sever))
 				for (var/obj/item/parts/P in limbs_to_sever)
-					P.sever(user)
+					P.sever(user, messy)
 				return 1
 		else if (istype(target, /obj/item/parts))
 			var/obj/item/parts/P = target
-			P.sever(user)
+			P.sever(user, messy)
 			return 1
 
 	// quick hacky thing to have similar functionality to get_organ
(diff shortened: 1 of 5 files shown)


code/obj/item/mob_parts.dm
```diff
@@ -252,7 +255,10 @@ ABSTRACT_TYPE(/obj/item/parts)
 			direction = turn(direction,180)
 
 		if (isitem(object))
-			object.streak_object(direction, src.streak_decal)
+			if (messy)
+				object.streak_object(direction, src.streak_decal)
+			else
+				object.streak_object(direction, null)
```
ZeWaka, line 261: why bother even streaking if its not messy

ZeWaka: dont get this

CalliopeSoups: I was using it because streaking doesn't throw the limbs, and skele limbs are boomerangs. I realized you can just turn that off with a var though, so I've reverted back to using ThrowRandom.

code/mob/living/carbon/human.dm
```diff
@@ -2847,7 +2847,7 @@
 	else if(limb == "l_leg" && src.limbs.l_leg) src.limbs.l_leg.remove()
 	else if(limb == "r_leg" && src.limbs.r_leg) src.limbs.r_leg.remove()
 
-/mob/living/carbon/human/proc/sever_limb(var/limb)
+/mob/living/carbon/human/proc/sever_limb(var/limb, var/user = src, var/messy)
```
pgmzeta, line 2850: i don't see where these new params are used
```suggestion
/mob/living/carbon/human/proc/sever_limb(var/limb)
```

pgmzeta (changes requested):

code/mob/living/carbon/human.dm
CalliopeSoups: Whoops, leftover from an earlier commit.

pgmzeta (approved):

PR #19769, opened 2024-06-30
Replaces toxins camera monitors with ones that work

AXLFAN2000:
[bug] [mapping]
## About the PR 
Replaces the blank camera monitors in toxins on donut 2 and donut 3 with public ones, because the bomb testing sites are on the public network and I think they should stay public.  Let staffies watch bombs. But also let me watch the bombs

## Why's this needed? 
they currently still have the camera monitors that dont work coz everything got repathed or whatever

### DIFF
--- a/maps/donut2.dmm
+++ b/maps/donut2.dmm
@@ -10704,10 +10704,8 @@
 /turf/simulated/floor/white,
 /area/station/medical/medbay)
 "bdz" = (
-/obj/item/device/camera_viewer{
-	network = "Zeta"
-	},
 /obj/table/reinforced/chemistry/auto,
+/obj/item/device/camera_viewer/public,
 /turf/simulated/floor/black,
 /area/station/science/lab)
 "bdB" = (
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
On donut 2 at least this seems like more of a bug with the cameras since all the other cameras on the research outpost are on their own network "Zeta"

AXLFAN2000:
> On donut 2 at least this seems like more of a bug with the cameras since all the other cameras on the research outpost are on their own network "Zeta". Donut 3's were probably just copy-pasted though.

I noticed that, and noticed that there's a second camera monitor specifically for looking at security cameras on the outpost  --  but seeing as this monitor was next to the toxins bomb chute, I assumed it was meant to be tied to these instead
image
Which are on the public network, and I really think they should stay that way so non-research folks can watch bomb tests
Would it be okay if I just tossed one of the outpost camera viewers somewhere else, like in the RDirs office or something?

TobleroneSwordfish:
Sure that works too, just make sure the zeta one is visually distinct from the public one in some way (sprite, colour, name etc.)

AXLFAN2000:
> Sure that works too, just make sure the zeta one is visually distinct from the public one in some way (sprite, colour, name etc.)

image
 Yep, they're differetn!  Tossed it on the RDir's desk!

PR #19787, opened 2024-07-01
attempt to speed up mining asteroid generation

Skeletonman0:
[PERFORMANCE] [INTERNAL]
## About the PR 
this only applies to space asteroid generation
- macros the proc used in the while loop inside asteroid generation
- this also makes the type checks in the following for loop redundant
- replace some of the .Add calls with +=
- turns atleast one lagcheck into lagcheck_if_live

something to consider is that the prefab placement loop also adds to the mining gen time

## Why's this needed? 
example of the generation times ive been getting on this branch: 
image

example of the generation times ive been getting on my master: 
image

will this be that much faster? i dont know.  could entirely be a "works on my machine" situation


code/modules/worldgen/GenerateMining.dm
```diff
@@ -86,15 +86,16 @@ TYPEINFO(/turf/variableTurf/clear)
...
-	if(!current || !center || (current.loc.type != /area/space && !istype(current.loc , /area/allowGenerate) && !isgenplanet(current)) || !istype(current, /turf/space))
-		return 0
-	if(ISDISTEDGE(current, AST_MAPBORDER))
-		return 0
-	var/probability = 100 - (((abs(center.x - current.x) + abs(center.y - current.y)) - (AST_MINSIZE+sizemod)) * AST_REDUCTION) + rand(-AST_TILERNG,AST_TILERNG)
-	if((abs(center.x - current.x) + abs(center.y - current.y)) <= (AST_MINSIZE+sizemod) || prob(probability))
-		return 1
-	return 0
+// this macro determines if a turf is to be considered for mining asteroid generation
+#define decideSolid(current, center, sizemod)\
```
ZeWaka, line 90: please provide proof via tracy that this significantly impacts startup and is worth it

ZeWaka (changes requested):

code/modules/worldgen/GenerateMining.dm
Azrun, line 90: Should also undef as this makes big assumptions regarding where it is called.

Azrun (changes requested):

PR #19798, opened 2024-07-02
Add additional clothing options to the syndicate clothing vendor

JORJ949:
## About the PR 
Adds a "syndicate" beret and winter coat (designs stolen from security)
image

Adds black and fingerless gloves to the vendor
Replace random colour berets on the cairngorm with the "syndicate" variety

## Why's this needed? 
Additional style options and security/syndicate use a very similar style

## Changelog 

```changelog
(u)JORJ949
(+)The Syndicate recently raided a security clothing designer. Additional clothing options available at the Syndicate clothing vendor.
```


OfficialSpoodle:
I think the beret looks waaaay to similar to the HoS' :P

JORJ949:
> I think the beret looks waaaay to similar to the HoS' :P

It IS the HoS beret, I just think it matches the syndicate style, it has flavour text to point this out.

itchybones:
I'd still consider a redesign/alternative beret version. I don't think that gold is very common in the Syndicate palette, I'd suggest making it a black/dark grey beret with a red badge instead of a golden one.

TobleroneSwordfish:
Not a fan of just giving syndies security clothing, you should have to work to steal that!

JORJ949:
Yeah no gonna put this one away for a bit, would like more syndie clothing but stealing the sec stuff just doesnt feel right

PR #19804, opened 2024-07-03
New fishing locale: Deep hole

AXLFAN2000:
[feature][gameobjects]
## About the PR 
Adds deep hole (like trench holes, oshan/nadir style) as fishable location which has fun stuff (rock, seashell, flipper) and even funner stuff (evil fish that want to kill you)

## Why's this needed? 
We need more stupid fishing spots that try to hurt you
image
(I tried to get a picture of one of the actual mobs I fished up but he threw himself back into the trench immediately and died)


ZeWaka (approved):
hole

PR #19806, opened 2024-07-03
Replaces 16th BYOND member message

AXLFAN2000:
[UI][QOL]
## About the PR 
Removes the harry potter BYOND member message and replaces it with a pride one :-)
image

Tested it on local and it looked good in chat (should look fine on lightmode too since theyre jpgs so no transparency)

## Why's this needed? 
The harry potter one makes me (and other people I've spoken to) frown a little (and other HP stuff got removed from the game so I felt this could be received well)


ZeWaka (changes requested):
unfortunately, this won't work per the terms of the license from TextStudio
please replicate in some editor like microsoft word

> You may not ... use the materials ... for any public display (commercial or non-commercial);

BatElite:
finally

AXLFAN2000:
> unfortunately, this won't work per the terms of the license from TextStudio please replicate in some editor like microsoft word
> 
> > You may not ... use the materials ... for any public display (commercial or non-commercial);

AWW drat, I didn't see that part  --  I'll whip something else up after work.  What a shame, that font is SO nice lol

PR #19814, opened 2024-07-03
Kondaru Toxins Update!

thebee03:
# I hope I made this PR correctly :) 

## About the PR 
This just makes a few changes to Kondaru's toxins to make it more in line with other maps' toxins!
 ping me on the goonstation discord @thebee03 if anything has to be changed

## Why's this needed? 
Currently, toxins is almost unplayable on Kondaru. It's a pain in the ass to do anything with just one burn chamber, plus the layout is a bit cramped, and generally it's just not the best to look at. 

## Changes
Moved things around to make room for a second burn chamber
Im gonna be honest, this is a pretty close ripoff of donut 3's toxins, but why mess with greatness? 
Added some lab tables, cause those look cool 

## Before and After!
Before:
image_2024-07-03_123035276
After: Now Matching With The Rest Of Science!
image

### DIFF
--- a/maps/kondaru.dmm
+++ b/maps/kondaru.dmm
@@ -14017,6 +14017,18 @@
 	dir = 4
 	},
 /area/station/crew_quarters/quarters_north)
+"beJ" = (
+/obj/grille/steel,
+/obj/machinery/door/poddoor/pyro/shutters{
+	dir = 4;
+	id = "tox_accessvent";
+	layer = 8;
+	name = "Buffer Vent"
+	},
+/turf/simulated/floor/shuttlebay{
+	name = "reinforced plating"
+	},
+/area/space)
 "beK" = (
 /obj/machinery/door/airlock/pyro{
 	dir = 4;
@@ -23556,8 +23568,13 @@
 /turf/simulated/floor/grey/blackgrime,
 /area/station/security/brig)
 "bQu" = (
-/obj/machinery/light/small/floor,
-/turf/simulated/floor/white,
+/obj/cable{
+	icon_state = "1-2"
+	},
+/obj/decal/tile_edge/stripe/big{
+	dir = 4
+	},
+/turf/simulated/floor/airless/black,
 /area/station/science/lab)
 "bQv" = (
 /obj/item/storage/wall{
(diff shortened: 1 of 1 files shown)


thebee03:
we are so back. should be 100% okay, but for real this time

TealSeer (commented):
Might be worth swapping the gray tiles out for the white ones so it matches the rest of the science department

thebee03:
> Might be worth swapping the gray tiles out for the white ones so it matches the rest of the science department

Can do!

thebee03:
> > Might be worth swapping the gray tiles out for the white ones so it matches the rest of the science department
> 
> Can do! A part of me wanted it to be a darker color for when im being a nerd at 1am so the brightness doesn't get to me. But it totally makes sense for it to be more in-line with the rest of the department :)

Done! Its all switched out!

TealSeer (approved):

Kubius:
Thank you for enjoying my map enough to update it

image

1. Connecting catwalks to burn chamber outputs can allow heat to propagate backwards through the airlocks. **This is not necessarily something that needs changed**, just a note.
2. This little area looks a little goofy - possibly-misaligned button, definitely-misaligned lightbulb, could potentially also use a meter on the corner for this section.
3. Self-explanatory.

thebee03:
> Thank you for enjoying my map enough to update it
> 
> image
> 
> 1. Connecting catwalks to burn chamber outputs can allow heat to propagate backwards through the airlocks. **This is not necessarily something that needs changed**, just a note.
> 2. This little area looks a little goofy - possibly-misaligned button, definitely-misaligned lightbulb, could potentially also use a meter on the corner for this section.
> 3. Self-explanatory.

1) I wanted a "safe" way to get to the burn chambers, as I (and other scientists) like to have easy-ish access to the chamber to set up stuff, especially if we are using molitz b, but i guess if the lattice is there, I should be good regardless. So im gonna remove it cause it also seems a bit out of place. 
2) Yeah, throwing a meter on there seems like itl be good, as for the bottom part, maybe I can make the pipe blue, to differentiate that it's the one going out to space that you can use to cool things down! Ill fix the lightbulb as well 
3) Hahahahah oops :)

thebee03:
Okay, if everything goes according to plan, this is my final edit. I just figured out that some airless tiles were used in the change up, but thats fixed now. If there's anything else, just let me know, but hopefully it can be added to the game soon (tm)!

itchybones:
Seems like I'm the only one who likes Kondaru toxins as it is right now, but this is so much better.

thebee03:
> Seems like I'm the only one who likes Kondaru toxins as it is right now, but this is so much better.

Aww thank you
Mainly thank whoever designed donut 3's chambers. I *can* say I did a bit of the decorations and adding the airlock out for easy access to the burn chambers via space! 
As for your other point, im one of those who can't stand one chamber, its just inefficient time-wise waiting for a chamber to fully clear to stop cross-contamination. But im a nerd so whatever :)

TobleroneSwordfish (changes requested):
Looks good, one thing: take two oxygen canisters out of the storage room if you want to put two more on the mixer, just to keep the total number roughly in line with other maps.

BatElite:
My one complaint is it's just a big rectangle of a room now. That's not very interesting.

Top burn chamber also has the input/output colours wrong on the floor.

thebee03:
> Looks good, one thing: take two oxygen canisters out of the storage room if you want to put two more on the mixer, just to keep the total number roughly in line with other maps.

image
Extra O2 cans removed. I wanted to have them on the ports, as its more in-line with a lot of other toxins, and so maybe beginners see where the canisters have to go.

thebee03:
> My one complaint is it's just a big rectangle of a room now. That's not very interesting.
> 
> Top burn chamber also has the input/output colours wrong on the floor.

I fixed the colors, but as for your other point about it being just a big rectangle, I agree. But honestly all toxins look that that. Or at least the toxins with two burn chambers. Personally, I don't want the chamber on different side of the rooms, as its just more work, and tougher to watch both burns. Also I'm in a pretty constrained area, as there's the whole conveyor belt system on the outside of Kondaru, which I don't want to change.

PR #19817, opened 2024-07-04
Adds plasma research counter subtypes

AXLFAN2000:
[mapping][science][qol]
## About the PR 
Adds two new subtypes of the chemistry pre-filled counters for use in plasma research labs: 

### Plasma Research
2x Atmospheric Analyzer
3x Timer
3x Igniter
3x Timer/Igniter Assembly
2x Proximity Sensor

### Fire safety 
2x Fire Extinguisher
3x Fire fighting grenade
1x Fire first aid
2x Gas mask

I wrote a list of the clutter on counters/tables in toxins in all rotation maps and this is a good median of all of it, with the exception of items I believe should stay visible (VR goggles, the valves themselves of course, and mechanical toolboxes/wrenches) and of the misc "office supply" type items, of which there is already a counter subtype for.

## Why's this needed? 
I tire of these mystery piles
image
image

### DIFF
--- a/code/obj/table.dm
+++ b/code/obj/table.dm
@@ -886,7 +886,7 @@ TYPEINFO_NEW(/obj/table/reinforced/chemistry)
 
 /obj/table/reinforced/chemistry/auto/firesafety
 	name = "fire safety lab counter"
-	desc = "These drawers have sooty marks on their handles. Hopefully you won't be adding to them."
+	desc = "These drawers have sooty scorch marks on their handles. Hopefully you won't be adding to them."
 	drawer_contents = list(/obj/item/storage/firstaid/fire,
 				/obj/item/extinguisher = 2,
 				/obj/item/chem_grenade/firefighting = 3,
@@ -894,7 +894,7 @@ TYPEINFO_NEW(/obj/table/reinforced/chemistry)
 
 /obj/table/reinforced/chemistry/auto/toxins
 	name = "plasma research lab counter"
-	desc = "These drawers are full of all the special doodas an aspiring bomb-maker could hope for."
+	desc = "These drawers are full of all the special doodads an aspiring bomb-maker could hope for."
 	drawer_contents = list(/obj/item/device/analyzer/atmospheric = 2,
 				/obj/item/device/timer = 3,
 				/obj/item/device/igniter = 3,
--- a/code/obj/table.dm
+++ b/code/obj/table.dm
@@ -884,6 +884,23 @@ TYPEINFO_NEW(/obj/table/reinforced/chemistry)
 				/obj/item/clothing/glasses/spectro,
 				/obj/item/device/reagentscanner)
 
+/obj/table/reinforced/chemistry/auto/firesafety
+	name = "fire safety lab counter"
+	desc = "These drawers have sooty marks on their handles. Hopefully you won't be adding to them."
+	drawer_contents = list(/obj/item/storage/firstaid/fire,
+				/obj/item/extinguisher = 2,
+				/obj/item/chem_grenade/firefighting = 3,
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Might just be me but I'm not a huge fan of just shoving everything in drawers that give no indication they have anything in them until you go around and click on all of them

AXLFAN2000:
> Might just be me but I'm not a huge fan of just shoving everything in drawers that give no indication they have anything in them until you go around and click on all of them

Good point, currently looking into ways to fix that!

ZeWaka:
yeah i'd rather have mapping helper spawners or something

ZeWaka:
or like a dedicated cabinet thing

AXLFAN2000:
> or like a dedicated cabinet thing

Like, putting them in their own DM?  Currently workshopping ways to visually indicate differences between types of chemistry counter

AXLFAN2000:
> or like a dedicated cabinet thing

Aloe said you probably mean the like cabinets on Oshan that chemicals are kept in   --  I really like those, but I'm not sure if they're a great idea here.  They don't hold enough items (I'd need two or three of them to fit as many items as can fit on top of 1 table currently or inside of 1 set of drawers as I intended with this PR), and can only be placed south of walls due to overlapping things above them

Currently discussing/workshopping an idea to give the dedicated mapping pre-filled versions of drawers their own little symbol on top, here's a concept (these sprites are nothingburger just a concept):
image

Would that help alleviate the concerns about the drawers being hard to understand?

ZeWaka:
> > or like a dedicated cabinet thing
> 
> Like, putting them in their own DM? Currently workshopping ways to visually indicate differences between types of chemistry counter

like a closet moreso

AXLFAN2000:
> like a closet moreso

Like a locker?  Sorry, I'm really not understanding what you're asking me to do  -- can you rephrase?

AXLFAN2000:
> Currently discussing/workshopping an idea to give the dedicated mapping pre-filled versions of drawers their own little symbol on top, here's a concept (these sprites are nothingburger just a concept): image
> 
> Would that help alleviate the concerns about the drawers being hard to understand?
@TobleroneSwordfish 
Not sure if plasma research stuff will end up getting in drawers in the end due to Zewakas concerns, but wanted to get your thoughts on these as a concept for helping with readability of the existing chem drawers?

AXLFAN2000:
> like a closet moreso

@ZeWaka 
I think I figured out what you were referring to: the secure lockers, right?  I did a mockup of how that would look, and am struggling with it a bit: 
image
image

This is under 2/3rds what was able to be fit in a drawer, and requires right clicking to access specific items since they layer over each other and have very similar sprites (namely the assemblies, timers, and igniters).  You'd need two lockers to fit what currently spawns in a heap atop one table in most toxlabs, which is doable but would severely cut back on available table space for working which is what appealed to me about the drawers to begin with.  

Do you think that even with proposed visual distinctions like above for drawers being implemented (and perhaps others --  Minty had a really good idea of giving drawers a light to indicate the presence of contents, borrowing from the visual cue of chutes) that the drawers are still not up for consideration for this use?  Just trying to figure out where to go from here, and if this drawers idea is worth trying to get into a workable state or if it's a hard no

TobleroneSwordfish:
I think both of these proposed solutions are worse than what we have right now, the locker one is extremely cramped and the drawer one will just lead to them never being opened. Honestly if you want to do cleanup you could maybe just remove some of the spare timer, igniters etc since they're available in the vendor anyway.

AXLFAN2000:
> Honestly if you want to do cleanup you could maybe just remove some of the spare timer, igniters etc since they're available in the vendor anyway.

#19897 did that instead!

PR #19823, opened 2024-07-04
Fix loudspeakers

### LINKED ISSUE #19784: Loudspeaker Transmitter nonfunctional
### Labels

[BUG][MINOR]

### Description

Loudspeaker Transmitter doesn't do anything when you try to turn it on

### Steps to reproduce

1.  Loudspeaker transmitter is currently inactive
2.  Do whatever to it (nothing happens)
3.  Mhelp asking how to turn it on
4.  Mentors tell you it is currently bugged out right now
5.  No open bug report
6.  Open a bug report

### Expected Behavior

I don't even know I genuinely don't.  I assume it worked at some point.  Maybe it should be able to be turned on and used.  I don't know.

Reported by: joeled
Client version: 515.1640
Active test merges: #19063
Round log date: 2024-07-01-15-55
Reported on: 2024-07-01 16:38:24
Round ID: 182474
Map: MUSHROOM

pgmzeta:
[game objects][bug]
## About the PR 
* Make attack_self and attack_hand use a new toggle proc for switching on/off
* Lock the frequency so it's not sanitized 

## Why's this needed? 
Fix #19784
For testing, these exist on donut3, manta, decarabia

### DIFF
--- a/code/obj/item/device/radio.dm
+++ b/code/obj/item/device/radio.dm
@@ -1009,6 +1009,7 @@ TYPEINFO(/obj/item/device/radio/intercom/loudspeaker)
 	anchored = ANCHORED
 	speaker_range = 0
 	chat_class = RADIOCL_INTERCOM
+	locked_frequency = TRUE
 	//Best I can figure, you need broadcasting and listening to both be TRUE for it to make a signal and send the words spoken next to it. Why? Fuck whoever named these, that's why.
 	broadcasting = 0
 	listening = 0		//maybe this doesn't need to be on. It shouldn't be relaying signals.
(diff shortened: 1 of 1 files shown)


Tarmunora (changes requested):
wouldn't it be better to give them a valid frequency rather than sidestep the sanitization step?

pgmzeta:
> wouldn't it be better to give them a valid frequency rather than sidestep the sanitization step?

as far as I can tell, they aren't meant to interconnect with standard intercoms; pushed a change to have them use a locked frequency instead since there is no way to modify it anyway.

Tarmunora (approved):

PR #19826, opened 2024-07-04
Fluid Highlight Overlays

Skeletonman0:
[OBJECTS] [SPRITES]
## About the PR 
new sprite overlays for fluids, changes submerged overlay behavior on max depth to a layer change

this is intended to only appear if the fluid is at a sufficient depth level where wall overlays are used

image

draft PRing as suggested to get recommended code improvements and sprite changes i should make

todo:

- [ ] get sprite improvements to make this look nicer
- [ ] rip out the last of the submerged overlay 4 stuff
- [ ] make the code better

## Why's this needed? 
feels like deep fluid filling a room / pool shouldn't just be a giant blob of one color


Azrun:
Nit: I think referring to this as caustics would be more illustrative of what the majority of this PR is trying to achieve. 

The vague title makes me think it is less impressive/consequential.

Mister-Moriarty:
> ...I think referring to this as caustics would be more illustrative of what the majority of this PR is trying to achieve.

Even more pedantic nitpick: "caustics" specifically refers to light that has been reflected or refracted by a surface. In the context of a liquid, caustics are the dancing light patterns that occur on the surface below the liquid, commonly seem in swimming pools (or on Oshan Lab). I feel that referring to the above implementation as caustics is misleading.

Skeletonman0:
taking a break from ss13

PR #19827, opened 2024-07-04
Resprites Floor Guides + Adds Some Sprites For New Ones

Flaborized:
## About the PR 

This PR resprites all the floor guide squares and adds a few variants for new ones. The new ones aren't implemented at all but it'd be pretty trivial if a mapper wanted to put them in. 

(new ones on bottom row here:)

floorguide

floorguide2

floorguide3

## Why's this needed? 

I don't like how the current sprites are basically two-tone and very flat, these aim to make them a bit higher contrast and gel more nicely with the rest of the sprites. I think they look better!

Kubius mentioned that having additional floor guides could be useful, and I think the extra ones could be a nice way to add flavor or like, direct people around n stuff.

## Changelog 

```changelog
(u)Flaborized
(+)Resprited floor guides.
```


Azrun:
Really like these!

Only quibble I have is that the botany symbol looks like it could possibly be a lamp or another piece of office equipment. There's something about the symmetry of the leaves that look either like a lamp or a scale. Additively the root structure flutes out a bit like something made in a lathe or pottery wheel. I think the coloration helps a lot but it is the least readable to me if the given department/POI.

cheekybrdy:
These look really cool! Just can't tell what the final 5 are, especially not at a glance. The one in engineer colours looks like the mining magnet if anything, not really sure on what the sec coloured ones are at all (brig?), the sci one kind of gives toxins vibes (P for poison == toxins?) and I assume the last is tech storage? They look good, just not sure they bring the intended information across.

itchybones:
> These look really cool! Just can't tell what the final 5 are, especially not at a glance. The one in engineer colours looks like the mining magnet if anything, not really sure on what the sec coloured ones are at all (brig?), the sci one kind of gives toxins vibes (P for poison == toxins?) and I assume the last is tech storage? They look good, just not sure they bring the intended information across.

I think the magnet-looking one is supposed to be the credit symbol for cargo and the "P" might stand for plasma. I'm not sure what the last two red ones are for, maybe brig and armory? Drawing bars could make the brig one easier to recognize. They look a lot better color-wise, though

Flaborized:
Yeah the yellow is the credits symbol, the P is for plasma, and the two red ones are non-specific and could be used for Whatever. (General warning sign/No Entry sign).

PR #19830, opened 2024-07-04
Small EFIF-1 bugfix set

Kubius:
[FIX]
## About the PR 

Fixes a small oversight in EFIF-1 construction evaluation in floor and r-floor modes by changing it to use the turf's "intact" variable instead of checking if it's plating by path, allowing it to work properly with turfs which were initially tiled.

Also adds verification that contained steel sheets exist in several spots where their quantity is checked (some of these cases are much less likely than others, but better to be safe).

## Why's this needed? 

Makes EFIF behave better when constructing in pre-existing structures (repair mode was not affected by the bug) and when empty. screm

### DIFF
--- a/code/modules/transport/pods/MainWeapon.dm
+++ b/code/modules/transport/pods/MainWeapon.dm
@@ -731,7 +731,7 @@ TYPEINFO(/obj/item/shipcomponent/mainweapon/constructor)
 		. = FALSE
 		//floors can be freshly constructed on other floors if they're bare (less fussy if repairing)
 		if (mode == EFIF_MODE_FLOORS || mode == EFIF_MODE_R_FLOORS)
-			if(T.intact == FALSE)
+			if(!T.intact)
 				return TRUE
 		//walls can be built on most floors. avoid some types that are unsuitable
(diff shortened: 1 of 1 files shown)


code/modules/transport/pods/MainWeapon.dm
```diff
@@ -727,9 +729,9 @@ TYPEINFO(/obj/item/shipcomponent/mainweapon/constructor)
 	///Helper proc to determine suitability of current tile for construction field application
 	proc/build_mode_eval(var/turf/T)
 		. = FALSE
-		//floors can be built on plating, but not other floors, unless repairing
+		//floors can be freshly constructed on other floors if they're bare (less fussy if repairing)
 		if (mode == EFIF_MODE_FLOORS || mode == EFIF_MODE_R_FLOORS)
-			if(istype(T,/turf/simulated/floor/plating))
+			if(T.intact == FALSE)
```
TobleroneSwordfish, line 734:
```suggestion
			if(!T.intact)
```

TobleroneSwordfish (approved):

PR #19832, opened 2024-07-05
Door/airlock operation rework (better open/close & door-linked forcefield integration)

Kubius:
[FIX][STATION SYSTEMS][CODE QUALITY]
## About the PR 

- Reworks unpowered_open_close() (the proc for airlock prying) to use the airlock's regular open() and close() procs. These procs now have a manual_actuation variable that, when set, alters them to behave correctly with prying (no checking for power, no airlock operation sound, et cetera).
- Also removes duplicate open and close behavior from pod doors. These now instead have overridden icon/animation procs as necessary (what the previous duplicate behavior was attempting to account for).
- Door-linked forcefield power usage management has been moved to the base door type (since all doors can have linked forcefields), and doors will now be subscribed to process when open if they have a force field that isn't being powered by a door-link shield generator.
- Door-linked forcefields now respond reliably to changes in power; if powered by the door, they will now deactivate if equipment power is lost while the door is open (and reactivate if equipment power is regained), while door-linked shields provided by a shield generator will consistently operate even when area power has failed.

## Why's this needed? 

Makes doors and their forcefields operate in a more cohesive manner (notably including updating adjacent tiles when they're supposed to) and reduces a bit of janky reimplementation.

### DIFF
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -855,8 +855,9 @@ var/global/list/cycling_airlocks = list()
 					D.close()
 
 /obj/machinery/door/airlock/close(var/manual_actuation = FALSE)
-	if (src.welded || src.locked || src.operating)
-		return
+	if(!manual_actuation)
+		if (src.welded || src.locked || src.operating || (!src.arePowerSystemsOn()) || (src.status & NOPOWER) || src.isWireCut(AIRLOCK_WIRE_OPEN_DOOR))
+			return
 
 	var/already_closed = ..(!src.safety)
 
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -1,9 +1,8 @@
 // how many possible network verification codes are there (i.e. how hard is it to bruteforce)
 #define NET_ACCESS_OPTIONS 32
 
-// power usage defines
+// power usage define
 #define OPEN_CLOSE_POWER_USAGE 50
-#define LINKED_FORCEFIELD_POWER_USAGE 100
 
 /// a global associative list of all airlocks linked together by cycling mechanisms. Indexed by ID
 var/global/list/cycling_airlocks = list()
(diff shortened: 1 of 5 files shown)


code/obj/machinery/shield_generators/portable_shield_generator.dm
```diff
@@ -665,6 +673,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/shieldgenerator, proc/turn_on, proc/turn_off
 			var/obj/machinery/door/door = (locate() in src.loc)
 			if(door)
 				door.linked_forcefield = src
+				door.SubscribeToProcess() //necessary for power consumption while open
```
TobleroneSwordfish, line 676: This is quite a heavy performance hit if every single door on Nadir is a processing machine, consider making it only subscribe to process when it actually needs to (or just moving power usage to only when the field is created)

code/obj/machinery/shield_generators/portable_shield_generator.dm
```diff
@@ -627,7 +635,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/shieldgenerator, proc/turn_on, proc/turn_off
 
 
 //sealab arrivalss
-/obj/machinery/door/var/obj/forcefield/energyshield/linked_forcefield = 0
+/obj/machinery/door/var/obj/forcefield/energyshield/linked_forcefield = null
```
TobleroneSwordfish, line 638: This is utterly cursed and shouldn't be here, I charge you with moving it to the `/obj/machinery/door` definition

code/obj/machinery/door/airlock.dm
```diff
@@ -857,44 +833,41 @@ var/global/list/cycling_airlocks = list()
...
-			else
-				D.close()
-
-/obj/machinery/door/airlock/close()
-	//split into two sets of checks so failures to close due to lacking power will cause linked shields to deactivate
-	if ((!src.arePowerSystemsOn()) || (src.status & NOPOWER) || src.isWireCut(AIRLOCK_WIRE_OPEN_DOOR))
-		if (src.linked_forcefield)
-			src.linked_forcefield.setactive(0)
-			power_usage -= LINKED_FORCEFIELD_POWER_USAGE
-		return
```
TobleroneSwordfish, line 887: Removing this early return makes it possible to close doors by hand without power.

TobleroneSwordfish (changes requested):

code/obj/machinery/shield_generators/portable_shield_generator.dm
Kubius: Moving power usage to field creation is a little wonky, but if I can't think of a better way to dodge the processing, I'll probably end up doing that.

code/obj/machinery/door/airlock.dm
Kubius: oops

code/obj/machinery/shield_generators/portable_shield_generator.dm
Kubius: Done in dfad581

code/obj/machinery/door/airlock.dm
Kubius: fixed in cbe5b1f

code/obj/machinery/shield_generators/portable_shield_generator.dm
Kubius: Now subscribes to process only when door is open. Also converted poddoors to use base open and close along the way.

PR #19841, opened 2024-07-05
Adds subtype for common station announcement computers

JORJ949:
## About the PR 
Adds subtypes for the announcement computers that are commonly found on-station (Command and QM) that have their names and accesses built in

## Why's this needed? 
1) Varediting is EVIL 
2) Mapping simplicity
3) Future PR where they all have a circuit subtype so keep their accesses when reassembled needs this (probably)

### DIFF
--- a/code/obj/machinery/computer/announcement.dm
+++ b/code/obj/machinery/computer/announcement.dm
@@ -206,6 +206,42 @@
 		src.announcement_radio.talk_into(src, messages, 0, src.name, src.say_language)
 		logTheThing(LOG_STATION, src, "ANNOUNCES: [message]")
 		return 1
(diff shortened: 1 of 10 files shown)


code/obj/machinery/computer/announcement.dm
```diff
@@ -206,6 +206,42 @@
 		src.announcement_radio.talk_into(src, messages, 0, src.name, src.say_language)
 		logTheThing(LOG_STATION, src, "ANNOUNCES: [message]")
 		return 1
+/obj/machinery/computer/announcement/station
```
ZeWaka, line 209:
```suggestion
		return 1

/obj/machinery/computer/announcement/station
```

PR #19847, opened 2024-07-05
Glass tile stuff

Laboredih123:
## About the PR 
Adds large glass transparent tiles, adds a type of these tiles that can autoconnect, and glass tiles emit light.

## Why's this needed? 
It looks real nice.
image

## Changelog 

```changelog
(u)cringe
(+)Nanotrasen has built better glass floors that are seamless. They also let through light.
```


itchybones:
Will these be added to the floor designer selection?

Laboredih123:
idk

Laboredih123:
should i add it?

Laboredih123:
what happened to the checks what

code/turf/floors.dm
```diff
@@ -2677,6 +2697,109 @@ TYPEINFO(/turf/simulated/floor/auto/water/ice)
 		if(prob(10))
 			src.icon_state = "snow_rough[rand(1,3)]"
 
+TYPEINFO(/turf/simulated/floor/auto/glassblock)
+	mat_appearances_to_ignore = list("steel","synthrubber","glass")
+	connect_diagonal = 1
+
+//dont use please
```
ZeWaka, line 2704: Then abstract it?

itchybones:
> should i add it?

I think so, yeah, players woulf come up with lots of cool ways of using these. Although this is entirely out of this PR's scope, it'd be neat for these to be craftable through some way.

WALPVRGIS:
Do all glass floors emit light now? I feel like that will affect a bunch of mapping I've done :o would like to account for this

Laboredih123:
> Do all glass floors emit light now (as of this PR anyway)? I feel like that will affect a bunch of mapping I've done :o would like to account for this

all of them yeah since it felt wrong for them to be dark, if you want me to remove that for some glass tiles like the nonlarge ones i suppose i can

code/turf/floors.dm
Laboredih123: already tried, every map check proceeded to fail

Laboredih123: /turf/simulated/floor is a used type

code/turf/floors.dm
```diff
@@ -2677,6 +2697,109 @@ TYPEINFO(/turf/simulated/floor/auto/water/ice)
...
+	to_plating()
+		return
+
+	break_tile_to_plating()
+		return
+
+	break_tile()
+		return
+
+	// unburnable, otherwise floorbots use steel sheets for repair which doesn't make sense
```
ZeWaka, line 2754: we can revisit this if it's a problem i guess

ZeWaka (approved):

Laboredih123:
oh huh

Laboredih123:
nice

PR #19871, opened 2024-07-06
Fix RP players cloneable after respawning

### LINKED ISSUE #11869: Being cloned after cryo-ghosting then going to title screen
### Labels

[BUG][MINOR]

### Description

There used to a bug where you could be cloned out of a cryo-ghosting, seems it's returned but only if you respawn on RP and get taken to the title screen. That might not be the case, but I was only cloned after returning to the title screen.  

### Steps to reproduce

1. Cryo ghost
2. Try to respawn
3. Get cloned?

### Expected Behavior

Not to be cloned

Reported by: CalliopeSoups
Round log date: 2022-11-07-23-28
Reported on: 2022-11-08 00:32:00

### LINKED ISSUE #12207: Cloneable after DNR and Ghosting after RP Respawn cooldown
### Labels

[BUG][MINOR]

### Description

If you set yourself as DNR, ghost, and then the timer on respawn ticks down, you are able to be cloned once it comes off cooldown, which tends to happen while you're selecting a new character, forcing you to do it again and probably take the 20 min respawn time.

### Steps to reproduce

1. Set DNR
2. Cryo
3. Ghost
4. Wait until Respawn becomes available
5. You are now cloneable, enabling you to be cloned before you are able to respawn as a new character.

### Expected Behavior

I expect that the DNR sticks while I'm choosing a new character, and that it hopefully doesn't stay after I respawn.

_No response_

### LINKED ISSUE #15135: DNR breaks when respawning?
### Labels

[BUG][MINOR]

### Description

Its possible to be cloned again even after setting DNR when you are in the process of respawning

### Steps to reproduce

1. die
2. set DNR
3. wait until you are able to respawn
4. click the funny button and start loading up a character
5. have someone clone you
6. woah, you're being cloned! and the character screen is stuck on your screen.

### Expected Behavior

not being able to be cloned in the title screen

Reported by: Dr. Disappointment
Active test merges: #14516 and #15096
Round log date: 2023-07-27-16-23
Reported on: 2023-07-27 16:59:44
Map: DONUT3

pgmzeta:
[respawning][bug]
## About the PR 
`/mob/new_player` counts as `isdead()`, so when checking if someone is cloneable do an explicit check to see if we're a respawned player and bail early.

## Why's this needed? 
Fix #12207
Fix #11869
Fix #15135

Recommend test-merge.


code/obj/machinery/computer/cloning.dm
```diff
@@ -447,6 +447,11 @@ proc/eligible_to_clone(var/datum/mind/mind)
 	if (!M.client)
 		return null
 
+	if (istype(M, /mob/new_player))
+		var/mob/new_player/player = M
+		if (player.is_respawned_player)
+			return null
```
Tarmunora, line 453: feels kinda weird that as a `new_player` after respawning you keep the same mind as you had, but maybe that's needed for things :iiam:

TobleroneSwordfish: I think `mind` has ended up mostly just being used as "persistent `client`", although that gives it very little distinction from `player` so :iiam:x2

TobleroneSwordfish, line 453: Actually is there ever a situation where we want a `new_player` to be cloned anyway?

TobleroneSwordfish (changes requested):

PR #19873, opened 2024-07-06
Add 10 second cooldown to reciting Miranda rights

TealSeer:
[QOL] [PLAYER ACTIONS]
## About the PR 

Add a 10 second cooldown to reciting miranda rights and removes an unused variable that was presumably going to be used to implement that cooldown however many years ago

## Why's this needed? 

People spamming a giant block of text as fast as they can click is annoying

## Changelog 

```changelog
(u) TealSeer
(+) The "Recite Miranda Rights" verb now has a 10 second cooldown.
```


Retrino:
rest in peace my lawbringer audio spam, you will be missed

PR #19874, opened 2024-07-06
Fix ship alert button closing all shutters

pgmzeta:
[bug][station systems]
## About the PR 
"not `lockdown` AND not `ai_core` AND `armory`" means ship alert locks down *everything but armory*

## Why's this needed? 
ship alert should close the lockdown, core, and armory shutters, not the bar/kitchen/public market/entire station... 
i am doofus and make bug :( please let me fix bug :)?


ZeWaka (changes requested):
Parentheses to avoid this in the future

PR #19880, opened 2024-07-07
Adds new mechcomp controls to the PTL

klookiyam:
## About the PR 
This PR adds mech comp controls for the PTL. The controls it adds are nearly one-to-one with the controls already available
on the ingame ui (with toggles converted to stop/start controls for convenience). They are:
- Toggle PTL input.
- Set PTL input.
- Toggle output.
- Set PTL output.
## Why's this needed? 
Automated engine setups are cool. This also givers players some options to leverage the PTL in interesting ways
that were not previously possible. A list of ideas off the top of my head:
- An engineer could set up a contraption that disables the ptl whenever whenever a player wants to walk through the path.
- An engineer could set up a controller that monitors an engines output and automatically adjusts the PTL along with it.
- An antagonist engineer could rig up something to charge up energy and fire a short, stronger pulse of the laser at souls foolish enough to cross it (under the gib threshold, of course).

[Station Systems]

## Changelog 

```changelog
(u)KakoLookiyam
(+) Added new mechcomp inputs to the Power Transmission Laser.
```

### DIFF
--- a/code/modules/power/pt_laser.dm
+++ b/code/modules/power/pt_laser.dm
@@ -56,12 +56,6 @@
 						terminal = term
 						break dir_loop
 
-		if(!terminal)
-			status |= BROKEN
-			return
-
-		terminal.master = src
-
 		AddComponent(/datum/component/mechanics_holder)
 		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_INPUT,"Enable Power Input", PROC_REF(_enable_input_mechchomp))
 		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_INPUT,"Disable Power Input", PROC_REF(_disable_input_mechchomp))
@@ -69,6 +63,13 @@
 		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_INPUT,"Enable Power Output", PROC_REF(_enable_output_mechchomp))
 		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_INPUT,"Disable Power Ouput", PROC_REF(_disable_output_mechchomp))
 		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_INPUT,"Set Power Output", PROC_REF(_set_output_mechchomp))
+
+		if(!terminal)
+			status |= BROKEN
+			return
+
+		terminal.master = src
+
 		UpdateIcon()
 
 /obj/machinery/power/pt_laser/disposing()
--- a/code/modules/power/pt_laser.dm
+++ b/code/modules/power/pt_laser.dm
@@ -13,7 +13,7 @@
 	req_access = list(access_engineering_power)
 	var/output = 0		//power output of the beam
 	var/max_dial_value = 999 // limitation of what can be set on the ui dial.
-	var/capacity = 1e15
+	var/capacity = 200 MEGA // Under the gib amount to prevent players from pulsing a death laser.
 	var/charge = 0
 	var/charging = 0
 	var/load_last_tick = 0	//how much load did we put on the network last tick?
--- a/code/modules/power/pt_laser.dm
+++ b/code/modules/power/pt_laser.dm
@@ -223,7 +223,6 @@
 	var/last_onln = online
 	var/last_llt = load_last_tick
 	var/last_firing = firing
-	// var/last_output = output_last_tick
 	var/dont_update = 0
 	var/adj_output = abs(output)
 	var/input_power = src.chargelevel * charging // We can fire a laser with more power than our battery capacity, but only if input is enabled.
--- a/code/modules/power/pt_laser.dm
+++ b/code/modules/power/pt_laser.dm
@@ -92,7 +92,7 @@
 /obj/machinery/power/pt_laser/proc/_set_input_mechchomp(var/datum/mechanicsMessage/inp)
 	if(!length(inp.signal)) return
 	var/newinput = text2num(inp.signal)
-	if (/*isnum_safe(newinput) && */newinput != src.chargelevel && newinput > 0 && newinput < 999 * 1 TERA WATT)
+	if (newinput != src.chargelevel && isnum_safe(newinput) &&  newinput > 0)
 		src.chargelevel = newinput
 		// Working backwards to update the ui based on the power we've set up.
 		if(chargelevel < 1 KILO WATT)
@@ -121,7 +121,7 @@
 	if(!length(inp.signal)) return
 	var/newoutput = text2num(inp.signal)
 	// We check against the absolute value of the current charge level, in case the PTL has been emagged.
-	if (newoutput != abs(src.output) && /*isnum_safe(newoutput) && */ newoutput > 0)
+	if (newoutput != abs(src.output) && isnum_safe(newoutput) newoutput > 0)
 		src.output = src.emagged ? -newoutput : newoutput
 		// Working backwards to update the ui based on the power we've set up.
 		if(newoutput >= 1 TERA WATT)
--- a/code/modules/power/pt_laser.dm
+++ b/code/modules/power/pt_laser.dm
@@ -1,4 +1,5 @@
 #define PTLMINOUTPUT 1 MEGA WATT
+#define LASER_COOLDOWN_ID "PTL laser change cooldown"
 
 /obj/machinery/power/pt_laser
 	name = "power transmission laser"
@@ -13,10 +14,13 @@
 	req_access = list(access_engineering_power)
 	var/output = 0		//power output of the beam
 	var/max_dial_value = 999 // limitation of what can be set on the ui dial.
-	var/capacity = 200 MEGA // Under the gib amount to prevent players from pulsing a death laser.
+	var/capacity = 200 MEGA WATTS // Under the gib amount to prevent players from pulsing a death laser.
 	var/charge = 0
 	var/charging = 0
 	var/load_last_tick = 0	//how much load did we put on the network last tick?
+	// How often you can change the output/firing status of the PTL
+	var/laser_cooldown = 1 SECOND
+	var/laser_output_needs_update = FALSE	// Does the PTL need to be updated this loop?
 	var/chargelevel = 0		//Power input
 	var/online = FALSE
 	var/obj/machinery/power/terminal/terminal = null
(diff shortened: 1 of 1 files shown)


AXLFAN2000:
I've wanted this for awhile, roxy was going to add it before the original PTL mechcomp integration was added and then didn't work on it because of that  --  this would be a really fun addition

SimpleWolf:
I beg for this to be merged. I've wanted this for so long

klookiyam:
> Doesn't appear to actually work as written, attempting to connect a button to the PTL with click-drag or "connect component" doesn't do anything

Thats weird - there was already a mechcomp holder there before my change so at the very least I'd expect that to work even if my changes are failing. There might be something weird going on with my local where I wasn't able to reproduce this - let me try testing with a few different engines/scenarios (or even live, if there's some weird bug causing mechcomp holders not to work at all) to see if I can figure it out.

TobleroneSwordfish:
If it helps I just spawned a `/cheat` ptl on atlas and put a button next to it

klookiyam:
Thank you! I think I've figured it out. The `New()` of a ptl has an early return [here ](https://github.com/goonstation/goonstation/pull/19880/files#diff-b0bd4ce30eb86ad0515970d8fd602e129ed4a8ab8acc4477243b38e87490f694L50-L61) if there is no `obj/machinery/power/terminal` in the three vertical turfs that make up its base. This exits the initialization early and doesn't touch any of the mechanic holder code or `UpdateIcon()`. I didnt run into this because I was testing against pre-existing PTL's with a varedited SMES feeding into it. Specifically, if I didnt have the terminal positioned here it broke:

ptl1
ptl2


I've moved the mechanics holder code above the early return which should make spawned PTL's function, even when not placed on a terminal.

code/modules/power/pt_laser.dm
```diff
@@ -76,6 +84,59 @@
...
+			src.input_multi = 1 MEGA WATT
+		else if(chargelevel < 1 TERA WATT)
+			src.input_multi = 1 GIGA WATT
+		else
+			src.input_multi = 1 TERA WATT
+		src.input_number = clamp((src.chargelevel/src.input_multi), 0, src.max_dial_value)
+
+/obj/machinery/power/pt_laser/proc/_enable_output_mechchomp()
+	src.online = TRUE
+	src.process(1)
```
TobleroneSwordfish, line 113: This makes it so spamming the signal repeatedly rolls the melt prob much faster than it should, process should only be called by the process controller (usually (kinda))

code/modules/power/pt_laser.dm
```diff
@@ -55,13 +56,20 @@
 						terminal = term
 						break dir_loop
 
+		AddComponent(/datum/component/mechanics_holder)
+		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_INPUT,"Enable Power Input", PROC_REF(_enable_input_mechchomp))
+		SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_INPUT,"Disable Power Input", PROC_REF(_disable_input_mechchomp))
```
TobleroneSwordfish, line 61: Is there a reason these aren't toggles? Seems like it would be much easier to work with (at least for simple things) and a shorter list of signals is usually less confusing to look at.

TobleroneSwordfish (changes requested):
Oookay that make sense, thanks!
I am a little concerned about this:
> An antagonist engineer could rig up something to charge up energy and fire a short, stronger pulse of the laser at souls foolish enough to cross it.

Ideally we wouldn't want to make it easy to push a button and get a gib-strength laser without having the full engine setup required for it. Might need to lower the maximum stored charge to prevent higher laser danger levels being possible with just reserve power.

code/modules/power/pt_laser.dm
klookiyam: I was thinking that tracking that toggle would make tracking the state of the PTL lasers annoying, as they shut themselves off if they're not powered. Thinking a little more about that - the PTL output doesn't shut off, it just enters an entirely different state `idle`, so theres really no reason not to leave these as toggle.

code/modules/power/pt_laser.dm
```diff
@@ -12,7 +12,8 @@
 	bound_width = 96
 	req_access = list(access_engineering_power)
 	var/output = 0		//power output of the beam
-	var/capacity = 1e15
+	var/max_dial_value = 999 // limitation of what can be set on the ui dial.
+	var/capacity = 200 MEGA // Under the gib amount to prevent players from pulsing a death laser.
```
TobleroneSwordfish, line 16:
```suggestion
	var/capacity = 200 MEGA WATTS // Under the gib amount to prevent players from pulsing a death laser.
```
pointless nitpick :P

code/modules/power/pt_laser.dm
```diff
@@ -76,6 +82,52 @@
...
+		if(newoutput >= 1 TERA WATT)
+			src.output_multi = 1 TERA WATT
+		else if(newoutput >= 1 GIGA WATT)
+			src.output_multi = 1 GIGA WATT
+		else
+			src.output_multi = 1 MEGA WATT
+		var abs_output_number = clamp((newoutput/src.output_multi), 0, src.max_dial_value)
+		src.output_number = src.emagged ? -abs_output_number : abs_output_number
+		src.update_output()
+		src.update_laser_power()
```
TobleroneSwordfish, line 128: Actually doesn't this still let you call `melt_blocking_objects` rapidly by flickering the value from `0` to `lots` repeatedly? It should probably have the same delayed startup behaviour that the toggle signal does.

TobleroneSwordfish (changes requested):
Hopefully last round of tweaks (sorry!)

code/modules/power/pt_laser.dm
klookiyam: I appreciate the quick responses. Let me go through the whole file and fix this on the ui controls as well.

klookiyam: I ran into some issues and ended up increasing the scope of the original changes. Two big problems were:
- Once I removed a lot of the redundant calls and pushed all the laser updates into the `process` function, I realized that the PTL was designed to only run off battery power (the `can_fire` proc checks to see `abs(src.output) <= src.charge`, and there are checks in the process loop as well). This made the previous fix where we reduced the battery capacity break things.
- If we defer the all laser changes to the process loop, the PTL gets a lot less responsive compared to now which felt bad. Even on my local, the delay was noticeable, and I imagine its only worse on live.

To fix the first issue, I refactored the `can_fire` proc and `process` to be able to check against the total available input power in addition to the charge (this will the PTL work with the reduced capacity).

To fix the second I added an adjustable cooldown to all PTL laser updates (currently `1 SECOND` but I can adjust that). It can toggle beam propagation instantly on the first click, but any clicks after will defer the updates to the next process loop update. This should stop from flickering the value.

This was my attempt to fix things based on my superficial knowledge of the codebase, and might be a little too amibitious. If you think I should defer all the changes to the process loop I can push a new revision.

code/modules/power/pt_laser.dm
```diff
@@ -243,13 +241,17 @@
...
-			if(charge >= adj_output && (adj_output >= PTLMINOUTPUT)) //have power to fire
+			if((charge + input_power) >= adj_output && (adj_output >= PTLMINOUTPUT)) //have power to fire
 				start_firing() //creates all the laser objects then activates the right ones
 				dont_update = 1 //so the firing animation runs
-				charge -= adj_output
-		else if(charge < adj_output && (adj_output >= PTLMINOUTPUT)) //firing but not enough charge to sustain
+				input_power -= adj_output
+				if (input_power < 0)
+					charge += input_power
+		else if(((charge + (src.charging * input_power)) < adj_output) && (adj_output >= PTLMINOUTPUT)) //firing but not enough charge to sustain
```
TobleroneSwordfish, line 250: This is good, but I think it should drain the remaining charge in this case to indicate *why* it's shutting off. Also can this statement be replaced with `can_fire`, incorporating the `PTLMINOUTPUT` check into that if needed?

code/modules/power/pt_laser.dm
```diff
@@ -262,9 +264,20 @@
 			if(length(blocking_objects) > 0)
 				melt_blocking_objects()
 			power_sold(adj_output)
-
+	else if (firing && !online)
+		stop_firing()
+
+	if(connected && charging && src.excess >= src.chargelevel)// if there's power available, try to charge
+		var adj_chargelevel = clamp(0, input_power, chargelevel)
```
TobleroneSwordfish, line 271:
```suggestion
		var/adj_chargelevel = clamp(0, input_power, chargelevel)
```

code/modules/power/pt_laser.dm
```diff
@@ -440,6 +453,10 @@
 
 
 /obj/machinery/power/pt_laser/proc/update_output()
+	if(ON_COOLDOWN(src, LASER_COOLDOWN_ID, src.laser_cooldown))
+		src.laser_output_needs_update = TRUE
+		return;
```
TobleroneSwordfish, line 458:
```suggestion
		return
```

code/modules/power/pt_laser.dm
```diff
@@ -440,6 +453,10 @@
 
 
 /obj/machinery/power/pt_laser/proc/update_output()
+	if(ON_COOLDOWN(src, LASER_COOLDOWN_ID, src.laser_cooldown))
```
TobleroneSwordfish, line 456: The way this is currently written seems to make turning the beam on instant, while turning it off takes time. This seems like the opposite of what we'd want.

TobleroneSwordfish (changes requested):

code/modules/power/pt_laser.dm
klookiyam: If it's off cooldown, this will make any action, turning off, turning on, or adjusting the power instant. If they attempt to do that when it's on_cooldown it should instead early return and set a flag that makes the next `process` update the laser.

klookiyam: An alternative, which I don't mind doing, is removing the `ON_COOLDOWN`, always just setting the `laser_output_needs_update` flag, and always waiting for `process`. We'd lose the snappiness of the current PTL but it would make things a little simpler.

TobleroneSwordfish: Nono the cooldown is fine. Looking into it a bit further it looks like the issue is actually just that `update_output` isn't checking the `online` var to determine whether to disable the beam, so it's always waiting for the next process tick instead.

PR #19887, opened 2024-07-08
Kondaru toxins fixes

thebee03:
## About the PR 
I made an oopsie.

## Why's this needed? 
Two buttons weren't working, and there was one extra ignition switch

## Changelog 

```changelog
(u)TheBee03
(+)A new and improved toxins on Kondaru, complete with a full 2 burn chamber setup!
```

### DIFF
--- a/maps/kondaru.dmm
+++ b/maps/kondaru.dmm
@@ -26853,9 +26853,6 @@
 	},
 /turf/space,
 /area/space)
-"ceT" = (
-/turf/simulated/floor/plating,
-/area/station/science/lab)
 "ceW" = (
 /obj/cable{
 	icon_state = "4-8"
@@ -26991,15 +26988,10 @@
 /turf/simulated/floor/black,
 /area/station/science/lab)
 "cfD" = (
-/obj/machinery/atmospherics/pipe/simple/insulated{
-	dir = 10;
-	can_rupture = 0
+/obj/machinery/atmospherics/pipe/manifold{
+	dir = 1;
+	level = 2
 	},
-/obj/machinery/atmospherics/pipe/simple/insulated{
-	dir = 10;
-	can_rupture = 0
-	},
-/obj/machinery/meter,
 /turf/simulated/floor/black,
 /area/station/science/lab)
 "cfE" = (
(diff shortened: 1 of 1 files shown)


444Portal (changes requested):
this doesnt need changelog entries, make sure that you arent removing the syndicate announcement computer in the listening post and this looks fine

JORJ949:
Changelog isnt necessary (except maybe a minor to say you fixed it if you want) and title should say what you actually changed

soleilan:
The updated Toxins still has issues that need addressing:

- The heating loop has no purge valve and purge vent.
- The Burn Chamber Heat Shields button has no heat shields to activate.
- There's a duplicate pipe at X 110 Y 77.

TobleroneSwordfish:
Please give PRs descriptive names

thebee03:
> The updated Toxins still has issues that need addressing in case you want to do that in this PR or a future PR:
> 
> * The heating loop has no purge valve and purge vent.
> * The Burn Chamber Heat Shields button has no heat shields to activate.
> * There's a duplicate pipe at X 110 Y 77.

Fixed em all!

444Portal (changes requested):
juuuuust learned theres no more portable air pump, that needs added too

AXLFAN2000:
Noticed something about the buttons that's worth changing while you're in there --  In mapping most buttons are actually on the tile in front of the wall they seem to be attached to, and just offset so that they're accessible without being directly adjacent to them.  Like, for example:
image
This button is actually on the tile with the table and just offset so it looks like its on the wall, so you can access it across the table
Some of the buttons you've added don't do that (the exhaust buttons) which makes them kind of inconvenient for hitting

Also, one of the pipes wasn't insulated so the current iteration of it that's on live is nearly unusable due to the amount of heat bleed (shouldn't be a problem after this is merged based on the mapdiff though, since you completely redid that section of piping)

Love the extra space this rework has given though!  Plenty of working room

thebee03:
> juuuuust learned theres no more portable air pump, that needs added too

Added that as well. Hopefully it should be good to go for real this time :)

444Portal (approved):

Issue #19800, opened 2024-07-02
shoes deteriorate in acid even with suit on


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

title

### Steps to reproduce

1. be in nadir breach with diving suit, internals and mask
2. shoes melt anyway
3. the heck, they're supposed to be covered

### Expected Behavior

shoes no melty

### Additional Information & Screenshots

Reported by: rdcb
Client version: 515.1640
On server: Goonstation 3 Roleplay: Morty
Active test merges: #19063
Round log date: 2024-07-02-09-58
Reported on: 2024-07-02 11:36:44
Round ID: 182526
Map: NADIR

PR #19889, opened 2024-07-08
Suits with coverage can save your jordans.

klookiyam:
## About the PR 
Checks for `LEGS` in body_parts_covered before exposing players shoes to puddles. This is an OK approximation and covers all space suits/biosuits, but there are certain outer suits (like the nurse outfit) that have leg coverage but obviously wouldn't cover the feet. A better fix might be to  add a `FEET` bitfield to all existing outer wear...

## Why's this needed? 
SS13 is a highly realistic space simulation and must be made as accurate as possible. Also, fixes https://github.com/goonstation/goonstation/issues/19800

## Changelog 

```changelog
(u)KakoLookiyam
(+)Outer suits with coverage will now prevent shoe exposure to puddles.
```

### DIFF
--- a/code/modules/fluids/fluid_core.dm
+++ b/code/modules/fluids/fluid_core.dm
@@ -787,7 +787,7 @@ ADMIN_INTERACT_PROCS(/obj/fluid, proc/admin_clear_fluid)
 	if (F.my_depth_level == 1)
 		if(!src.lying && src.shoes && src.shoes.hasProperty ("chemprot") && (src.shoes.getProperty("chemprot") >= 5)) //sandals do not help
 			do_reagent_reaction = 0
-			if (!src.wear_suit || !(src.wear_suit.body_parts_covered & LEGS)) // suits can go over shoes
+			if (!src.wear_suit || !(src.wear_suit.c_flags & SPACEWEAR)) // suits can go over shoes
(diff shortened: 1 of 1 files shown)


code/modules/fluids/fluid_core.dm
```diff
@@ -787,7 +787,8 @@ ADMIN_INTERACT_PROCS(/obj/fluid, proc/admin_clear_fluid)
 	if (F.my_depth_level == 1)
 		if(!src.lying && src.shoes && src.shoes.hasProperty ("chemprot") && (src.shoes.getProperty("chemprot") >= 5)) //sandals do not help
 			do_reagent_reaction = 0
-			F.group.reagents.reaction(src.shoes, TOUCH, F.group.amt_per_tile, can_spawn_fluid = FALSE)
+			if (!src.wear_suit || !(src.wear_suit.body_parts_covered & LEGS)) // suits can go over shoes
```
TobleroneSwordfish, line 790:
```suggestion
			if (!src.wear_suit || !(src.wear_suit.body_parts_covered & SPACEWEAR)) // suits can go over shoes
```
This makes more sense I think, since it's roughly what aqua tenebrae uses to determine protection.

TobleroneSwordfish (changes requested):

PR #19893, opened 2024-07-08
Remove Pathology from Donut 3

JORJ949:
## About the PR 
Removes the pathology lab from donut 3 and places maintenance tunnels between the sci "maint" and disposals

## Why's this needed? 
Pathology gone, blank room in sci the RD cant access

## Changelog 

```changelog
(u)JORJ949
(+)Removed the pathology lab from Donut 3.
```

### DIFF
--- a/code/obj/machinery/computer/shuttle.dm
+++ b/code/obj/machinery/computer/shuttle.dm
@@ -223,8 +223,7 @@ ABSTRACT_TYPE(/obj/machinery/computer/transit_shuttle)
 /obj/machinery/computer/transit_shuttle/asylum/New()
 	..()
 	destinations = list(/area/shuttle/asylum/observation,
-	/area/shuttle/asylum/medbay,
-	/area/shuttle/asylum/pathology)
+	/area/shuttle/asylum/medbay)
 	currentlocation = locate(/area/shuttle/asylum/medbay)
 
 // research shuttle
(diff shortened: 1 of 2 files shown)


444Portal:
Rundown pathology is far more neat than a random room and a hallway imo

AXLFAN2000:
Could the area be turned into something else, if the patho ties must be removed?  That area works well like the warrens on nadir, for antagging and just as a nice build zone.

AXLFAN2000:
Wooo! Crime zone!

itchybones:
I love the place as-is, and it's not hurting anyone, right? Donut 3 has plenty of maint space already, so we don't really need more. It's a neat room, and unlike the other pathology labs, this one doesn't have any obsolete machines and fits into the map perfectly fine. I'd suggest opening a forum thread for this.

Vaskritaya:
I'm not sure I see the point of this. To my understanding, the actual disabled machinery isn't there, it's functionally just a flavored maintenance area. Donut3's entire schtick is sprawling run-down maints and forgotten rooms, and this one adds a lot more flavor than most with it being a fun nod to the literally abandoned mechanics of the past.

trazodont:
but then where would i steal a chem dispenser from without causing issues for others!

TobleroneSwordfish:
Yeah I think I'd rather either redesign it to be something unique or just leave it in, it's a neat little crime area with a connection to the asylum shuttle.

ZeWaka:
closing this in favor of a either a redesign or just leaving it in per community and dev feedback

JORJ949:
Yeah I completely agree and would have been redoing this anyway if I wasnt away

PR #19895, opened 2024-07-08
remove two bannable words from vox

pgmzeta:
[removal][internal]
## About the PR 
remove "bitch" and "penis" vox files

## Why's this needed? 
they're not allowed in game; don't carry dead weight


TobleroneSwordfish (approved):
sure!

PR #19897, opened 2024-07-08
Adds toxins-specific vendomat with larger inventory

AXLFAN2000:
[gameobjects][feature]
## About the PR 
Adds a subtype of the vendomat that spawns in toxin labs with an increased inventory, and removes those items from being loose on tables at roundstart.
(it's research purple)
image

## Why's this needed? 
Yaay more tabletop workspace!
Leah suggested just removing the countertop items since they're in the vendor anyways, and minty suggested adding this subtype so that I'm not, like, removing the amount of tools that science has access to.  Just moving them! and Aloe said the vendomat confuses him (like the original vendomat, not this one) but that this seemed fine

## Changelog 

```changelog
(u)joeled
(+)Loose vendomat items in toxin labs have been moved to the vendomat.
```


pgmzeta:
Sprited the vendor 4dir & added to special event maps: https://github.com/AXLFAN2000/goonstation/pull/2

pgmzeta (changes requested):
Sprited the vendor's east/west and added to special event maps:
https://github.com/AXLFAN2000/goonstation/pull/2

looks good otherwise

AXLFAN2000:
> Sprited the vendor's east/west and added to special event maps: [AXLFAN2000#2](https://github.com/AXLFAN2000/goonstation/pull/2)
> 
> looks good otherwise

Yooo thanks so much!!!! :-D

TobleroneSwordfish (changes requested):
Looks like you accidentally put one in the gym on pamcoG, not that it *should* matter because in theory that map should be autogenerated but I fear that technology may have been lost to time.

AXLFAN2000:
> Looks like you accidentally put one in the gym on pamcoG, not that it _should_ matter because in theory that map should be autogenerated but I fear that technology may have been lost to time.

Woops, should be fixed now!

itchybones:
Would it be possible for the vendor to have a few (maybe 4 or 6) empty black-red tanks? I don't know if it's just me, but I primarily use tanks from heneral fabricators, which are often far away from research, so having them closer would be useful. They could also be added to scifabs, though that's probably out of scope.

AXLFAN2000:
> Would it be possible for the vendor to have a few (maybe 4 or 6) empty black-red tanks? I don't know if it's just me, but I primarily use tanks from heneral fabricators, which are often far away from research, so having them closer would be useful. They could also be added to scifabs, though that's probably out of scope.

That's a great idea, I've actually wished they were in the scifabs myself  --  I'd rather they be in the fabs than in this vendomat, coz then you can fabricate as many as you want (and have metal for).  I'll make a second PR adding those to scifabs later tonight, and see how people feel about it.

TobleroneSwordfish (approved):

pgmzeta:
please fix standard-off being accidentally renamed
image

AXLFAN2000:
> please fix standard-off being accidentally renamed 

Done!

pgmzeta (approved):

code/modules/vending/vending.dm
```diff
@@ -2552,6 +2552,22 @@ TYPEINFO(/obj/machinery/vending/monkey)
...
 
+/obj/machinery/vending/standard/toxins
+	desc = "A vending machine machine full of various useful tools and devices that plasma researchers can use to make bombs."
+	icon_state = "toxins"
+
+	create_products(restocked)
+		product_list += new/datum/data/vending_product(/obj/item/device/prox_sensor, 10)
+		product_list += new/datum/data/vending_product(/obj/item/device/igniter, 10)
+		product_list += new/datum/data/vending_product(/obj/item/device/radio/signaler, 10)
+		product_list += new/datum/data/vending_product(/obj/item/assembly/time_ignite, 3)
```
Azrun, line 2563: I understand that this provides parity on some maps, but I don't see why this is needed.  Igniter and timer are already available.

AXLFAN2000: > I understand that this provides parity on some maps, but I don't see why this (`/obj/item/assembly/time_ignite`) is needed. Igniter and timer are already available.

I felt the same way, but I didn't want to be "removing" items that were previously available  --  It's not just some maps, Atlas, Clarion, cog1, cog2, donut 2, donut 3, kondaru and oshan all had them.  If you're happy with them just being removed outright, I'd prefer that too!

AXLFAN2000: On second thought, I just ripped em out.  If someone else decides that they were, like, load-bearing important to the toxins experience they can always PR adding them in.  I personally have never used one of the pre-built assemblies

PR #19928, opened 2024-07-10
Add Syndicate PDAs to syndicate clothing vendor

JORJ949:
## About the PR 
Makes the pda2/syndicate a generic syndicate PDA with the special cartridge one being a subtype (If syndie PDAs were used in secret they may need repathing to this pda2/syndicate -> pda2/syndicate/nuclear)
Puts both this new PDA in the syndicate clothing vendor

## Why's this needed? 
So EVIL people can be more DASTARDLY!

## Changelog 

```changelog
(u)JORJ949
(+)Syndicate PDAs now available at the Syndicate clothing vendor (cartridge not included).
```


TobleroneSwordfish:
It's kinda confusing to have syndicate IDs that aren't agent cards so you can't do anything with them unless you somehow get ID computer access *before* going loud as a syndicate agent?

JORJ949:
> It's kinda confusing to have syndicate IDs that aren't agent cards so you can't do anything with them unless you somehow get ID computer access _before_ going loud as a syndicate agent?

What if I gave them a generic name like "John Doe" and job "syndicate operative"

TobleroneSwordfish:
> What if I gave them a generic name like "John Doe" and job "syndicate operative"

Ehh but they still don't actually work or serve an in-universe purpose.

JORJ949:
> Ehh but they still don't actually work or serve an in-universe purpose.

Would it be terrible to give it post access as like a "backup key" so operatives dont get stuck? If someone already got in surely they have a way to get back in when they leave, afaik the only thing this lets them use is the announcement computer.

TobleroneSwordfish:
I don't really like the idea of adding more random listening post access cards because they are ultimately just going to be stolen and make the listening post even less secure.

JORJ949:
> I don't really like the idea of adding more random listening post access cards because they are ultimately just going to be stolen and make the listening post even less secure.

Perhaps stick the ID in C.A.R.L then or create a new device that "plates" your ID in the syndicate colour

TobleroneSwordfish:
I.. don't see the point in the plating kits at all, sorry. That just seems weird and snowflakey for something incredibly minor like an ID card colour. Why would you ever want your ID card to declare that you're a syndicate anyway?

JORJ949:
> I.. don't see the point in the plating kits at all, sorry. That just seems weird and snowflakey for something incredibly minor like an ID card colour. Why would you ever want your ID card to declare that you're a syndicate anyway?

Yeah thats fair, if you want a syndicate ID you probably already bought an agent card, I'll remove the ID part

PR #19929, opened 2024-07-10
Remove "Use external speaker" button/option from pods

### LINKED ISSUE #19902: pods give the power of speach to ghost critters
### Labels

[BUG]

### Description

critters can speak via pod comunication system

### Steps to reproduce

1. become a ghost critter
2. have a person a pod with multiple seats
3. enter as a passenger
4. click the "Use Comms System" button
5. type your message that you want to say to the living

### Expected Behavior

ghost critters should not be able to use the comm system function

passenger interaction with the pod function seems to be intended as you cant turn on the engine
image

pgmzeta:
[removal][vehicles]
## About the PR 
Remove the "Use Comms System" functionality and button from the pod HUD

## Why's this needed? 
Button to talk isn't needed when you can just talk and have it be heard outside your pod. "or whatever" indeed.
Fix #19902

### DIFF
--- a/code/datums/hud/pod.dm
+++ b/code/datums/hud/pod.dm
@@ -14,7 +14,6 @@
 		set_code
 		rts
 		wormhole
-		use_comms
 		leave
 		rcs
 		lights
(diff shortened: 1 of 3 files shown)


DisturbHerb:
I'll try to get an alternative PR soon with what I think could fit in this role. It'd be a waste of my good sprites to lose the functionality entirely...

PR #19930, opened 2024-07-10
Mildly Mutated genes are reinforced and can be scrambled from roundstart

jan-antilles:
## About the PR 
This PR changes the Mildly Mutated trait to:
- Reinforce the activated gene so that it can't be removed with mutadone
- Grant 1 trait point

The gene can still be removed by scrambling at Genetics. In an effort to encourage this, Radiation Emitters are now available at the GeneTek computer at roundstart, and Radiation Dampeners have been moved to Tier 1 research.

## Why's this needed? 
This makes Mildly Mutated more high-risk (and high-reward) and gives more interaction to Genetics from the beginning of a round. Discussion of the idea is from this forum thread by Soleilan:

https://forum.ss13.co/showthread.php?tid=22929

## Changelog 

```changelog
(u)Jan.antilles
(*)Mildly Mutated now reinforces the mutation (requiring scrambling or gene research to be removed) and grants 1 trait point. 
(*)Geneticists are able to scramble genes at roundstart instead of needing to research radiation emitters.
```


TheTyrant25 (approved):
The code itself looks all correct as far as I can tell. Nice change.
That said, I'd probably break the changelog entry into two bits: the mildly mutated change and the radiation emitters being unlocked at roundstart.

mordent-goonstation (changes requested):

code/modules/medical/genetics/geneticsResearch.dm
```diff
@@ -345,25 +347,13 @@ var/datum/geneticsResearchManager/genResearch = new()
...
-		..()
-
 /datum/geneticsResearchEntry/rad_coolant
 	name = "Emitter Coolant System"
 	desc = "Reduces the amount of time required for Radiation Emitters to cool down."
 	researchTime = 1800
 	researchCost = 120
 	tier = 2
-	requiredResearch = list(/datum/geneticsResearchEntry/rademitter)
+	requiredResearch = list(/datum/geneticsResearchEntry/rad_dampers)
```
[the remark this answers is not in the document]
jan-antilles: I considered it, but I figure having some progression is still nice. Making research even faster means letting Genetics "max out" even sooner in long rounds, and this probably keeps a little more balance in the Force, since they're getting a free unlock now.

code/modules/antagonists/nuclear_operative/nuclear_operative.dm
```diff
@@ -51,6 +51,9 @@
 		else
 			H.equip_if_possible(new /obj/item/tank/emergency_oxygen/extended(H), SLOT_L_STORE)
 
+		if(H.traitHolder?.hasTrait("mildly_mutated"))
+			H.traitHolder.removeTrait("mildly_mutated")
```
mordent-goonstation, line 55: I suspect this won't solve the issue, as the trait being added will activate (and reinforce) the gene, but removing it doesn't do the opposite.

mordent-goonstation (changes requested):

code/datums/traits.dm
```diff
@@ -526,20 +526,22 @@
...
+	disability_desc = "Minor reinforced alteration from baseline genetic sequence"
 
 	onAdd(var/mob/owner)
 		var/datum/bioHolder/B = owner.bioHolder
 		var/datum/bioEffect/E = pick(B.effectPool)
 		B.ActivatePoolEffect(B.effectPool[E], 1, 0)
+		E.curable_by_mutadone = FALSE
+		E.name = "Reinforced " + E.name
 		SPAWN (1 SECOND) // This DOES NOT WORK at round start unless delayed but somehow the trait part is logged??
 			logTheThing(LOG_COMBAT, owner, "gets the bioeffect [E] from the trait [name].")
```
mordent-goonstation, line 546:
```
/datum/trait/mildly_mutated
	name = "Mildly Mutated"
	desc = "A random mutation in your gene pool starts activated."
	id = "mildly_mutated"
	icon_state = "mildly_mutatedB"
	points = 0
	category = list("genetics")
	afterlife_blacklisted = TRUE
	disability_type = TRAIT_DISABILITY_MINOR
	disability_name = "Genetic Deviation"
	disability_desc = "Minor alteration from baseline genetic sequence"
	var/datum/bioEffect/activated_effect = null

	onAdd(var/mob/owner)
		var/datum/bioHolder/B = owner.bioHolder
		var/datum/bioEffect/E = pick(B.effectPool)
		var/msgGain = E.msgGain
		E.msgGain = ""
		var/datum/bioEffect/activated = B.ActivatePoolEffect(B.effectPool[E], 1, 0)
		E.msgGain = msgGain
		if (istype(activated))
			src.activated_effect = activated
			SPAWN (1 SECOND) // This DOES NOT WORK at round start unless delayed but somehow the trait part is logged??
				logTheThing(LOG_COMBAT, owner, "gets the bioeffect [E] from the trait [name].")

	onRemove(mob/owner)
		if (!isnull(src.activated_effect))
			// TODO: check the owner has the bioeffect still
			var/msgLose = src.activated_effect.msgLose
			src.activated_effect.msgLose = ""
			owner.bioHolder.DeactivatePoolEffect(src.activated_effect)
			src.activated_effect.msgLose = msgLose
```
Something like this? Would suppress the message shown to people at roundstart, but I'd argue that's a feature.

mordent-goonstation: EDIT: we're going a different way with it, I think.

chemtrailhaze:
personally i think this should either give a 0 trait point, or vestigal, emp, radiation and mutagenic should be blacklisted. it's already bad enough having people come in already dying from radiation when i'm trying to work on something else at roundstart. higher chance of a bad gene is good, but if its EMP they can just choose to not come into genetics and stand around breaking airlocks, cameras, headsets, etc etc. pretty sure it also drains security's batons and tasers.

scrambling being moved up to a roundstart thing is also good, and it would help if this is the case, but now medbay has to treat them for radiation at the start. having it irremovable unless they die might be better. people pick the trait, they should live with the repercussions, just like with puritan.

jan-antilles:
Closing so I can reopen in the other PR because Mordent and I made SO MANY CHANGES THERE

PR #19931, opened 2024-07-10
Refactor /obj/artifact and add triggers for lighting

Glamurio:
[Feature] [Internal] [Science]

## About the PR 
Changes  every instance of `/obj/artifact` into `/obj/machinery/artifact` and removes `/obj/artifact`.
Adds `/obj/proc/ArtifactProcess()` which extends `/obj/machinery/artifact/process()` and `/obj/item/artifact/process()`.
Adds artifacts to `processing_items` if their trigger uses light in any way.

Adds `/datum/artifact_trigger/lighting` and `/datum/artifact_trigger/darkness` which activate via light level.
The lighting trigger checks the light level by comparing the brightness value of the light used, this was done so that artifacts don't randomly activate by regular simple lights.
The darkness trigger checks the light level of the turf via `turf/is_lit`, and if this check fails, it compares how dark the tile is. This was done because simple lights don't really have any kind of brightness associated with them and I wanted to avoid weird cases where a tile is technically lit but counts as dark.

## Why's this needed? 
Originally I just wanted to implement light-based triggers, however it turns out that half the artifact code was duplicated into machinery and non-machinery, and because light-based checks can only feasible be done via process() checks, I figured I kill two birds with one stone and merge the redundant code so that now everything's a machine and uses process.

Also, light-related checks could cause performance problems, but I made sure to only check for them when the artifact actually rolls it as a trigger, and the check is removed once the artifact does trigger.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Latest scientific discoveries indicate that artifacts can respond to different levels of light.
```

### DIFF
--- a/_std/__build.dm
+++ b/_std/__build.dm
@@ -30,7 +30,7 @@ o+`        `-` ``..-:yooos-..----------..`
 
 //------------ OPTIONS TO GO FAST ------------//
 
-//#define GOTTA_GO_FAST_BUT_ZLEVELS_TOO_SLOW 1  // Only include the tiny map Devtest, no other zlevels. Boots way faster
+#define GOTTA_GO_FAST_BUT_ZLEVELS_TOO_SLOW 1  // Only include the tiny map Devtest, no other zlevels. Boots way faster
 
 //#define IM_REALLY_IN_A_FUCKING_HURRY_HERE 1  // All of the below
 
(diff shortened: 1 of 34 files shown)


TobleroneSwordfish (changes requested):
Refactoring more things to be `/obj/machinery` when they shouldn't be is kind of the opposite of what we want. If you need to add a process proc to certain non-activated artifacts then do it using a custom controller instead.

Glamurio:
> Refactoring more things to be `/obj/machinery` when they shouldn't be is kind of the opposite of what we want. If you need to add a process proc to certain non-activated artifacts then do it using a custom controller instead.

Time to draft and re-do it

PR #19932, opened 2024-07-10
Adds withdraw cash verb to PDAs

Valtsu0:
[objects][qol]

## About the PR 
image
Works exactly the same as the withdraw button in the pda's ui

## Why's this needed? 
would be nice.
Other basic functionality of the pda is already in verbs and at least i use them


LaCumbiaDelCoronavirus:
Does this also add the verb to the Local tab of commands in the top-right? If not then i would recommend doing that for better ease of use
Good change 👍

Valtsu0:
> Does this also add the verb to the Local tab of commands in the top-right? If not then i would recommend doing that for better ease of use Good change 👍

It does

PR #19938, opened 2024-07-11
Adds empty gas tanks to science fabricator

AXLFAN2000:
[game-objects][qol]
## About the PR 
Adds the empty gas tanks from the general fabricator to the science fabricator as well

## Why's this needed? 
Science uses them a lot and the nearest gen fab is usually very very far away


TobleroneSwordfish (approved):
ye!

itchybones:
Yesss, very glad to see this. I've been meaning to PR this change for quite a while, but never got around to it. This is gonna save me a lot of time walking to/from manus

PR #19939, opened 2024-07-11
pda pin

Sovexe:
## About the PR 
* PDAs feature a simple PIN lock/unlock on withdrawals
* Adds a cooldown on dispensing cash to prevent abuse. Should be essentially unnoticed by players
image

## Why's this needed? 
* Basic money security, but still stealable
* Potential denial of service vector

## Changelog 

```changelog
(u)Sovexe
(+)added a simple PIN lock to PDA wallets
```

### DIFF
--- a/code/obj/item/device/pda2/base_os.dm
+++ b/code/obj/item/device/pda2/base_os.dm
@@ -1032,7 +1032,10 @@
 				if (!isnull(src.master.ID_card))
 					. += "<a href='byond://?src=\ref[src.master];eject_id_card=1'>Eject [src.master.ID_card]</a><br>"
 					if (src.master.accessed_record)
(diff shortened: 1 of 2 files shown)


ZeWaka:
Thoughts would be appreciated

Retrino:
[this thread](https://forum.ss13.co/showthread.php?tid=22926) made a lot of good points on why it should remain pinless, but a tldr from the thread
- most of the time, will only exist to annoy the player with yet another dialogue box to click through
- already has the softlock of not being withdrawable unless an id is in it
- makes stealing money a bit easier. normally only doable by someone with bank console access or empathetic thought to get someone's pin, and considering how little cash is actually used, why discourage someone who is seeking it out?

PR #19947, opened 2024-07-12
Add air shields to bomb test range airlocks

AXLFAN2000:
[mapping][QOL]
## About the PR 
Adds air shields to the external doors in toxin lab that lead to the bomb testing ranges on Donut 2 and Donut 3

## Why's this needed? 
Other external airlocks have them and after a good round of going in and out to pick up the crystals from the bomb testing range there's usually no air at all in toxins which is annoying

### DIFF
--- a/maps/donut2.dmm
+++ b/maps/donut2.dmm
@@ -28404,6 +28404,11 @@
 	cycle_id = "toxins";
 	name = "Toxins"
 	},
+/obj/forcefield/energyshield/perma{
+	desc = "A door-linked force field that prevents gasses from passing through it.";
+	name = "Door-linked Atmospheric Forcefield";
+	powerlevel = 1
+	},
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
Shouldn't they just have airlock cyclers instead?

AXLFAN2000:
> Shouldn't they just have airlock cyclers instead?

I'm pretty sure they do, if by that you mean going out the external airlock closes the internal one.  They still end up leaking a lot if you're going in and out a lot and the room they're attached to is pretty small from my experiences  --  donut 2s is worse, donut 3s toxins is a lot bigger and one big room so it's not as necessary, just felt reasonable to do both at once

AXLFAN2000:
Sord said they'd prefer it stay the way it is, so closing

PR #19948, opened 2024-07-12
Makes cameras in bomb testing chambers explosion immune

AXLFAN2000:
[mapping][feature]
## About the PR 
Adds a new variant of the stationary cameras used in the bomb testing ranges, which are named as such and will not be destroyed by explosions

## Why's this needed? 
The regular ones blow up really easily and I believe it is enriching for staff assistants to watch ttvs going off on the TV

I'm not fully sure this is the best way to make them explosion resistant (ideally I'd like them to be explosion Resistant but not Immune) but roxy said this is how it's handled currently and this is the best option.  If you have a better idea pleaase let me know
I'm also not sure if it should be a variant of the regular television cam or of the auto version because donut 2 uses the regular ones and donut 3 uses the auto ones so I just went with the regular one. I dunno

### DIFF
--- a/code/modules/camera/television.dm
+++ b/code/modules/camera/television.dm
@@ -75,3 +75,8 @@
 	name = "mobile television - science"
 	c_tag = "science mobile"
 	network = "telesci"
+
+/obj/machinery/camera/television/bombtest
+	name = "television camera - bomb testing range"
+	desc = "A bulky stationary camera for wireless broadcasting of live bomb tests."
+	invuln = TRUE
--- a/maps/donut2.dmm
+++ b/maps/donut2.dmm
@@ -17199,10 +17199,8 @@
 /turf/simulated/floor/grass/leafy,
 /area/station/wreckage)
 "cpz" = (
-/obj/machinery/camera/television{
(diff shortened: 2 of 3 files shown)


AXLFAN2000:
> Random cameras should not be explosion immune.

Is there any way to make them more reinforced to explosions otherwise? Like more beefy, but not immune  -- as it currently stands, most explosions bigger than like a pipebomb (so  what you're making in the toxins lab the range is attached to) tend to wipe out these "reinforced" cameras while leaving most of the chamber intact.  I just think it'd be nice if the cams set up to film explosions were less flimsy

pgmzeta:
override the ex_act for your new camera type; right now they fall back to the parent machinery one.

AXLFAN2000:
> override the ex_act for your new camera type; right now they fall back to the parent machinery one.

Oooh okay, I'll look into that and update the PR over the weekend

AXLFAN2000:
actually I'll just make a brand new PR with cooler better beefy cameras later coz I'm BUSY I'm going to ribfest today

PR #19959, opened 2024-07-12
Lets Wraiths point at things

Kyle2143:
## About the PR 
Lets wraiths point at things, the point is invisible to normal people while the wraith is invisible, and visible when the wraith is visible(corporeal).
Also lets wraith turn their mob in different directions by clicking like all other directional mobs.

## Why's this needed? 
Lets wraiths point at things, I'm fairly sure this was always an oversight, especially since we added ghost pointing.

### DIFF
--- a/code/mob/living/intangible/wraith.dm
+++ b/code/mob/living/intangible/wraith.dm
@@ -408,7 +408,7 @@
 	click(atom/target, params)
 		if (src.targeting_ability)
 			return ..()
-		if (src.client && src.client.check_key(KEY_POINT))
+		if (src.client?.check_key(KEY_POINT))
 			src.point_at(target, text2num(params["icon-x"]), text2num(params["icon-y"]))
 
 		else if (!density)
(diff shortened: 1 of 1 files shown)


code/mob/living/intangible/wraith.dm
```diff
@@ -405,12 +405,25 @@
 	equipped()
 		return 0
 
-	click(atom/target)
+	click(atom/target, params)
 		if (src.targeting_ability)
 			return ..()
-		if (!density)
+		if (src.client && src.client.check_key(KEY_POINT))
```
frawhst, line 411:
```suggestion
		if (src.client?.check_key(KEY_POINT))
```
more concise

code/mob/living/intangible/wraith.dm
```diff
@@ -405,12 +405,25 @@
...
+			src.point_at(target, text2num(params["icon-x"]), text2num(params["icon-y"]))
+
+		else if (!density)
 			src.examine_verb(target)
 
+		//Stolen from /mob/click because
```
frawhst, line 417: why doesn't this just call the parent anyways? seems smelly to reimplement code from the parent when we could just be... calling that

Kyle2143: IDK, they did something weird with the order of things and I didn't want to figure it out. I suppose I should

PR #19962, opened 2024-07-13
Mutation ability misfires does only begin under 80 stability

M-Earthfire:
[balance][Medical]

## About the PR 

This PR changes abilities granted by mutations in a way so their misfire-chance begins to scale under 80 instead of 100 stability.

## Why's this needed? 

With past changes to stability, people are very much encouraged to not take any non-stabilized mutations. While stability is meant to be a tradeoff, most often the repercussions for going under 100 stability are so bad that most people are unwilling to do any tradeoffs and instead use stabilized genes only.
One of the major reasons for that are mutation misfires, with effects going from "stunned for 3 seconds" to "overdosing on multiple chems to a deadly degree". Since most of these effects translate to you instantly loosing a fight, they are  considered to be a risk not worth taking.

Thus, this PR tries to create a room where the first danger someone encounters from instability is gene degradation, so people are encouraged to take risks by dipping stability.

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Abilities granted by mutations start their misfire-chance at under 80 stability, down from 100.
```


Cherman0:
As a matter of balance I'd rather the degredation chance start lower and the misfire chance stay the same. The benefits of a gene can never be worth going under the degredation threshold because you'll just lose the gene and end up even worse than where you started in very short order. Having a space where you get to have the gene but it doesnt always work right seems more potentially appealing than having a space where the gene works but you dont get to have it.

PR #19967, opened 2024-07-14
Fix Oshan Hos/Armory door AI access

JORJ949:
## About the PR 
Makes the door between HoS/Armory on oshan require hacking by the AI

## Why's this needed? 
Accidental change when the door was changed from sec to command probably, also consistency


JORJ949:
Closing due to #20022 being merged

Issue #19961, opened 2024-07-13
PTL sometimes reports roughly twice the power available


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

After being in operation a bit, the `Optimal Input` fields sometimes shows roughly double the power actually available.

### Steps to reproduce

1. Monitor an SMES or TEG.
2. Also monitor the PTL.
3. Notice how the available power does not match.

### Expected Behavior

The PTL should report the correct available power.

### Additional Information & Screenshots

May be related to this: https://github.com/goonstation/goonstation/pull/19880

Reported by: zeroxoz2
Client version: 515.1641
On server: Goonstation 1 Classic: Heisenbee
Active test merges: #19063
Round log date: 2024-07-13-06-08
Reported on: 2024-07-13 06:16:19
Round ID: 183217
Map: ATLAS

klookiyam:
I reported this and think I know what's causing it, working on a fix.

Issue #19977, opened 2024-07-14
PTL fizzles out


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

the PTL can't seem to work well when the input is over 200MW, it will fizzle out and start again

### Steps to reproduce

Make over 200 MW
Set the PTL to a value over 200 MW
Watch the PTL not work properly

### Expected Behavior

The PTL should be able to work with amounts of power that are much much higher

### Additional Information & Screenshots

Reported by: freedo5
Client version: 515.1641
On server: Goonstation 3 Roleplay: Morty
Active test merges: #19063
Round log date: 2024-07-14-22-06
Reported on: 2024-07-14 23:09:39
Round ID: 183351
Map: DONUT3

klookiyam:
Tested it on the server - even if the input is greater than output, anything greater than the battery capacity set for output will eventually fizzle. It takes longer when the output is well under the input (664 vs 400), but eventually you will see grid saturation drop to 0 followed by the laser switching off, then on again on the next tick.

PR #19982, opened 2024-07-15
Refactored PTL process block and removed mult adjustments from load.

klookiyam:
[Station Systems][Bug]

## About the PR 
When I pushed https://github.com/goonstation/goonstation/pull/19880 I made some incorrect assumptions about how the power grid worked. Specifically, I assumed that power produced by engines scaled with `mult`, and that compounding floating point imprecision wouldn't cause issues.

This PR removes `mult` from power draw but keeps it for adjusting power of the beam. This is a slight buff, as before we were also multiplying the charge we added to the battery by `mult` without increasing the load. The only difference is now we are not attempting to take the increased output from the battery. Adding the complexity of dealing with the compounding precision drift with a reduced battery capacity isn't worthwhile imo. (I am open to pursuing other fixes if someone has better ideas)

This also refactors the `process` block so that power calculations are done first, power changes are comitted to the grid/battery, and then the new laser state is executed. This will make the code a little more readable and easier to reason about.

## Why's this needed? 
Fixes https://github.com/goonstation/goonstation/issues/19961 and https://github.com/goonstation/goonstation/issues/19977

## Changelog 

```changelog
(u)Kakolookiyam
(+)PTL should no longer occasionally shut off when firing beyond its batteries capacity.
```

### DIFF
--- a/code/modules/power/pt_laser.dm
+++ b/code/modules/power/pt_laser.dm
@@ -257,7 +257,7 @@
 		else //firing and have enough power to carry on
 			power_used += adj_output
 			new_state = CONTINUE_FIRING
-	else if (firing && !online)
+	else if (firing)
 		new_state = STOP_FIRING
 
 	// Consume power from network
--- a/code/modules/power/pt_laser.dm
+++ b/code/modules/power/pt_laser.dm
@@ -261,9 +261,9 @@
 		new_state = STOP_FIRING
 
 	// Consume power from network
-	var/adj_charge = clamp(-src.charge, starting_surplus - power_used, src.capacity - src.charge)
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (changes requested):
The input circuit isn't switching to idle properly when there isn't enough power to satisfy it
image
actually seems like the condition is reversed since it's switching to idle when it has excess power: 
image

klookiyam:
> The input circuit isn't switching to idle properly when there isn't enough power to satisfy it image actually seems like the condition is reversed since it's switching to idle when it has excess power: image

I've inverted the condition I've applied - its a bit confusing because the variable name is `is_charging` but it translates to Input circuit "Idle" and "Online" in TGUI.

PR #19983, opened 2024-07-15
Add landing pad to Kondaru listening post

JORJ949:
## About the PR 
Adds a landing pad to the kondaru listening post
image

## Why's this needed? 
Space hurts and the other listening posts have them


JORJ949:
Me when i forget to checkout before opening strongdmm

PR #19984, opened 2024-07-15
Fix Clarion and Kondaru non-perspective post doors

JORJ949:
## About the PR 
Clarion and kondaru had a couple non-perspective interior doors, makes them the perspective ones

## Why's this needed? 
They shouldn't be flat silly syndies

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -13136,9 +13136,9 @@
 	},
 /area/listeningpost)
 "bLL" = (
-/obj/machinery/door/airlock/syndicate{
-	desc = "This looks like a perfectly innocent weather satellite. It's locked up pretty tight to keep bath salts junkies from breaking in to steal the equipment.";
-	name = "Weather Satellite"
+/obj/machinery/door/airlock/pyro/reinforced/syndicate{
+	dir = 4;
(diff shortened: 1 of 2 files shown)


maps/clarion.dmm
```diff
@@ -13136,9 +13136,9 @@
 	},
 /area/listeningpost)
 "bLL" = (
-/obj/machinery/door/airlock/syndicate{
-	desc = "This looks like a perfectly innocent weather satellite. It's locked up pretty tight to keep bath salts junkies from breaking in to steal the equipment.";
-	name = "Weather Satellite"
```
TobleroneSwordfish, line 13141: This name/desc seems like it makes more sense, not sure how other maps are but the hidden syndicate listening post door just being called "Listening Post" seems silly.

TobleroneSwordfish (changes requested):

maps/clarion.dmm
JORJ949: No other maps have this afaik and the doors I changed are purely internal, I can make a seperate PR changing post door labelling but thats probably best kept seperate

TobleroneSwordfish (approved):

PR #19988, opened 2024-07-15
Make robot dectalk quieter

TobleroneSwordfish:
## About the PR 
Makes robots less loud

## Why's this needed? 
Robots are too loud
PRing this to TM on dev so I don't accidentally blow out people's speakers or mute head surgeon in case I got the scaling wrong on the volume settings.

## Changelog 

```changelog
(u)LeahTheTech
(+)Head surgeon and other robotic voices should now speak at a more reasonable volume.
```


Sbmhawk:
Thank god

PR #19998, opened 2024-07-16
Rework Atlas Engineering

JORJ949:
## About the PR 
Reworks atlas engineering to make the TEG overall better
image
Move mainframe (and associated devices) to AI upload to make space for Engineering
image
Also move tech storage quantum telescope to sci

## Why's this needed? 
Atlas TEG is awful to use and doesnt make any power (not that its needed because atlas, but engineering is fun seeing number go up

## Changelog 

```changelog
(u)JORJ949
(*)Reworked Atlas Engineering, moving the computer core to AI foyer.
```


LaCumbiaDelCoronavirus:
Looks great compared to original but 1 suggestion:
If theres space for it, the plasma canisters should really probably be in a separate storage room because: all other maps do it (donut 2 engi gas storage is an exception) and if those start bursting.. i think you know what will happen

Also small mistake:
The apc under the burn chamber is partially covered by a FIRE sign

(also this is really really a nitpick but the 2 light switches next to the gas extractor seem awkward, could be put on the same tile with an offset?)

JORJ949:
Away until friday so expect fixes then/saturday

TheTyrant25 (commented):
This corridor is now quite visually uninteresting.
image
Possibly do something with it? A side entrance or some optionally blast shielded windows could be good. Or even just a table and chair.
Otherwise it all looks pretty nice.

TheTyrant25 (approved):

TobleroneSwordfish (changes requested):
Generally looks good, a couple of minor things:
- This APC terminal is on the same tile as a gas port, difficult to repair.
image
- Some of these switches look wrong, probably not adjusted for the new perspective sprites?
image
- The yellow walls are out of place on atlas, should use standard walls like the rest of the map.

PR #20004, opened 2024-07-16
Pre-stocked, organized supply carts

pgmzeta:
[game objects][qol]
## About the PR 
Create standard, pre-stocked definitions with pixel-shifted contents for the following carts:
* Breach Cart
  * Acid variant swaps o2 medkit for fire (acid burns), removes o2 grenades
* Mechanics Tools Cart
* Medical Crash Cart
* Forensics Detective Cart
  * Security variant does not contain forensics scanner or luminol bottle
* Bomb Defusal Crisis Cart

Replaces instances of these cart stacks in maps with new pre-stocked carts. Does *not* blankly add them to all maps, only replaces existing copy-pasted cart/object stacks.

## Why's this needed? 
Cross-map consistency, easier to see what's in carts, less copy-paste of objects

## Screenshot
image
top row: Breach, Breach (Acid variant), Tools
middle-right: Crash
bottom row: Detective, Security, Bomb Crisis

### DIFF
--- a/code/obj/storage/carts.dm
+++ b/code/obj/storage/carts.dm
@@ -29,6 +29,45 @@
 	icon_closed = "mechcart"
 	icon_opened = "mechcartopen"
 
+/obj/storage/cart/mechcart/breach
+	name = "breach cart"
+	desc = "A big rolling supply cart equipped for handling hull breaches."
+	spawn_contents = list(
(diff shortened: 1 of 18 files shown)


JORJ949:
acid one could use an ifdef for nadir to check rather than having 2 different types

PR #20019, opened 2024-07-17
Re-routes most wall pipes on Cogmap 1

CalliopeSoups:
[MAPPING] [QOL]
## About the PR 

I enjoy plumbing... This re-routes most pipes on Cogmap 1 that go through walls or otherwise can go under a door instead of a window etc.

Points of note:

- Table, window and chemvendor budged in medsci for space.
- Pipes route through artsci now, possibly too frequently damaged?
- Window and blinds added to the crematorium.

I believe I've tested most chutes, it all seems fine.

## Why's this needed? 

Pipes are much easier to repair when they're not underneath walls.

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -4461,15 +4461,6 @@
 	},
 /turf/simulated/floor/grime,
 /area/station/security/brig)
-"arK" = (
-/obj/disposalpipe/segment/transport{
-	dir = 8;
-	icon_state = "pipe-c"
-	},
-/turf/simulated/floor/purple/side{
-	dir = 4
-	},
-/area/station/science/lobby)
 "arM" = (
 /obj/structure/girder,
 /turf/simulated/floor/plating/airless,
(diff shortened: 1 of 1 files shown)


JORJ949:
artsci pipes arent that bad because you are only piping the test chamber mail and sci lobby disposal pipes through it and the explosions in artsci are often just enough to scorch the floors, if its any bigger then the test chamber mail chute is the least of your worries

BatElite:
image
Why make this complicated circuit out of a previously simple bend? Can't you have one grouped junction branching into sci, then split that up into the individual tags just inside sci where you've got a bit of room?

CalliopeSoups:
Research piping is a lot more reasonable now.

StrongDMM_PsEozUsM9r

ZeWaka:
This is a huge pain to review, you should've done smaller chunks

CalliopeSoups:
> This is a huge pain to review, you should've done smaller chunks

Oh rip, I asked in the discord if I should split by department or map and all the responses were by map. I can take this down and put up department splits, cause I want to change all the varedited mail junctions into the proper subtype versions anyway.

PR #20031, opened 2024-07-18
Fix Clarion Engine plant

JORJ949:
## About the PR 
Clarion engineering plant is "thermo electric generator room" engine/core area, none of these exist on clarion but the map pointer still shows the "nuclear reactor room" engine/core/nuclear area, nukies cant plant there. Fixes that

## Why's this needed? 
Bug fix = good


JORJ949:
honestly should probably not be a plant spot but whatever

BatElite:
Why not? the reactor room is plenty big

JORJ949:
> Why not? the reactor room is plenty big

The radiation, nukies are immune to it amd crew are not meaning if the engine is setup yhe crew are going to have a near impossible time

frawhst:
The viability of the plant location is sort of a separate issue. Fix is good.

PR #20033, opened 2024-07-18
First aid kits use can_hold to hold pill bottles

### LINKED ISSUE #20032: tools and first aid kits
### Labels

[BUG][MINOR]

### Description

tools fit and don't fit into first aid kits

### Steps to reproduce

If you try to place a tool like welder or crowbar into a first aid kit, it will not fit.
If you click on a tool on the floor with a first aid kit in hand, the tool will go into the kit.

### Expected Behavior

Either tools should fit when placing or tools shouldn't go in to the kit when clicked on floor

Reported by: kameliali
Client version: 515.1636
Active test merges: #19063
Round log date: 2024-07-18-16-33
Reported on: 2024-07-18 16:59:18
Round ID: 183652
Map: NADIR

pgmzeta:
[bug][game objects]
## About the PR 
Changes the max held size class of first aid kits to tiny, and use can_hold instead of custom size class handling code in firstaid/attackby (in `medbot.dm`).

## Why's this needed? 
Fix #20032

### DIFF
--- a/code/modules/storage/items/firstaid.dm
+++ b/code/modules/storage/items/firstaid.dm
@@ -1,7 +1,0 @@
-// allows first aid kits to hold pill bottles
-/datum/storage/firstaid
-
-/datum/storage/firstaid/check_can_hold(obj/item/W)
-	. = ..()
-	if(istype(W, /obj/item/storage/pill_bottle))
-		return src.get_fullness(W)
--- a/code/modules/storage/items/firstaid.dm
+++ b/code/modules/storage/items/firstaid.dm
@@ -1,0 +1,7 @@
+// allows first aid kits to hold pill bottles
+/datum/storage/firstaid
+
+/datum/storage/firstaid/check_can_hold(obj/item/W)
+	. = ..()
+	if(istype(W, /obj/item/storage/pill_bottle))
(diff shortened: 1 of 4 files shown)


code/modules/storage/items/firstaid.dm
```diff
@@ -0,0 +1,7 @@
+// allows first aid kits to hold pill bottles
+/datum/storage/firstaid
+
+/datum/storage/firstaid/check_can_hold(obj/item/W)
+	. = ..()
+	if(istype(W, /obj/item/storage/pill_bottle))
+		return src.get_fullness(W)
```
Tarmunora, line 7: Is there a reason to override the proc rather than just using the `can_hold` list?

Azrun: ☝️

pgmzeta: misread the order of operations for can_hold ^^'

Tarmunora (approved):
inb4 removing the spawn breaks things somehow

pgmzeta:
> inb4 removing the spawn breaks things somehow

all i ask is we comment why when if it has to get added back :(

PR #20043, opened 2024-07-20
Re-locates most contraband lockers 

CalliopeSoups:
[GAME-OBJECTS] [QoL] [BALANCE]
## About the PR 

This PR does three things:

- If a map has contraband lockers in its security department's main room or similar, removes them and places them in the back.
- If a map only has one locker, add another. (Clarion didn't have space...)
- If there's no light tubes/bulbs above one of the lockers, add one. If there's no switch in the room add one.

## Why's this needed? 

I believe contraband lockers should be located in the back of a security department if possible, as stealing back lost loot should be encouraged. I also think adding a light tube above them makes that form of breaking lockers more obvious.

## Changelog 

```changelog
(u)CalliopeSoups
(+)Contraband lockers have been re-located to the back rooms of most security departments. This is sure to deter theft as nobody goes back there!

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -12651,7 +12651,6 @@
 /area/space)
 "bJh" = (
 /obj/storage/secure/closet/security/equipment,
-/obj/machinery/light_switch/west,
 /turf/simulated/floor/red/side{
 	dir = 1
 	},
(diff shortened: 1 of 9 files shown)


frawhst (approved):
You added a light on plasma glass to D3, and I don't remember if that's something we specifically avoid on maps (imnotmapper). If not, all good. If so, remove that, otherwise lgtm.

CalliopeSoups:
> You added a light on plasma glass to D3, and I don't remember if that's something we specifically avoid on maps (imnotmapper). If not, all good. If so, remove that, otherwise lgtm.

I don't see a new light tube on a plasmaglass window on Donut 3, and there were already a couple there, were you referring to another map maybe?

Sord213:
> You added a light on plasma glass to D3, and I don't remember if that's something we specifically avoid on maps (imnotmapper). If not, all good. If so, remove that, otherwise lgtm.

Looks like the PR removed/moves the light. But also lights on windows is fine, imo

PR #20064, opened 2024-07-22
Conga lines

Valtsu0:
[Actions][Feature]
## About the PR 
https://github.com/user-attachments/assets/96941502-6254-4cad-9568-e32348f5a67e
Everyone after the first must use a grab
(The teleporting in the video is reproducible on master, and also worse on the cargo train)

## Why's this needed? 
Fun

## Changelog 

```changelog
(u)Valtsu0
(*)You can now grab while being grabbed/pulled, allowing you to form conga lines
```

### DIFF
--- a/code/mob/input.dm
+++ b/code/mob/input.dm
@@ -300,7 +300,7 @@
 		if ((BOUNDS_DIST(old_loc, src.pulling) > 0 && BOUNDS_DIST(src, src.pulling) > 0) || !isturf(src.pulling.loc) || src.pulling == src) // fucks sake
 			src.remove_pulling()
 			//hud.update_pulling() // FIXME
-		else
+		else if (length(chain) <= 1)
 			pulling += src.pulling
 	for (var/obj/item/grab/G in src.equipped_list(check_for_magtractor = 0))
 		pulling += G.affecting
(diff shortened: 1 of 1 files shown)


DisturbHerb:
finally. security patrol conga line parity with tg

TobleroneSwordfish:
I get that the teleporting is technically an existing bug, but I'm not sure it's a good idea to add a feature that exposes it like this without looking into fixing it first.

flrsh:
I think the idea is fun + funny. However, I'd like this to be tied to grabs instead of passive pulls, since the former require you to use a hand, and this seems meaningful enough to warrant that.

Valtsu0:
It currently works with both. I see no reason to specifically prevent doing it with passive pulls

Valtsu0:
I have looked into the teleporting a bit, but have not made any progress

Azrun:
> > I think the idea is fun + funny. However, I'd like this to be tied to grabs instead of passive pulls, since the former require you to use a hand, and this seems meaningful enough to warrant that.
> 
> It currently works with both. I see no reason to specifically prevent doing it with passive pulls

It creates value for an available hand.
Diegetically this seems to represent having the capacity to hold and direct someone, either a garment or a limb and use that to create the chain itself.

How would a bunch of people with chainsaws for hands pull each other around? 🤔 🩸🩸🩸🩸🩸🩸🩸🩸

Valtsu0:
Not that different from just pulling one person around, but made it so everyone except the "driver" has to grab

PR #20067, opened 2024-07-22
Changes the graphite hardening compound recipe

TobleroneSwordfish:
## About the PR 
Swaps `silicon dioxide` to `silicon` in the graphene recipe.

## Why's this needed? 
The current recipe is extremely tedious for something you ideally need close to the start of the round when you're setting the engine up. Space glue is already pretty hard to make and was made slightly longer by the phenol recipe change.

## Changelog 

```changelog
(u)LeahTheTech
(*)Graphene now takes silicon, rather than silicon dioxide to make.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -2383,7 +2383,7 @@
 		name = "Graphene"
 		id = "graphene"
 		result = "graphene"
-		required_reagents = list("fuel" = 4, "iron" = 1, "silicon_dioxide" = 1)
+		required_reagents = list("fuel" = 4, "iron" = 1, "silicon" = 1)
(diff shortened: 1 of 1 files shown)


code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -2383,7 +2383,7 @@
 		name = "Graphene"
 		id = "graphene"
 		result = "graphene"
-		required_reagents = list("fuel" = 4, "iron" = 1, "silicon_dioxide" = 1)
+		required_reagents = list("fuel" = 4, "iron" = 1, "silicon" = 1)
```
Azrun, line 2386: Would like a comment to the effect that silicon is used to reduce complexity. The original recipe was supposed to be an abstraction of an existing process.

Azrun (approved):

PR #20072, opened 2024-07-23
Mildly Mutated genes are reinforced and can be scrambled from roundstart

jan-antilles:
## About the PR 
This PR contains:

-Changes to Mildly Mutated. It gives +1 trait point so that people still take it, but the mutation it gives is Reinforced from roundstart. Live with your decision or go talk to the genticists!
-Geneticists can now use Radiation Emitters to scramble genes at roundstart. They can research Rad Dampeners at Tier 1 now.
-Logic has been built in to track role/job so that roles that are incompatible with Mildly Mutated aren't affected by it. This means that Nukies, Salvagers, and Wizards can take the trait, but it won't trigger the addition of a bioeffect from their gene pool if they roll a gamemode role.

(This is a redo of #19930 because we were going to atomise it and then decided not to. But the logic for role tracking is really cool and we can use it for other things in the future as well.)

## Why's this needed? 
This makes Mildly Mutated more high-risk (and high-reward) and gives more interaction to Genetics from the beginning of a round. Discussion of the idea is from this forum thread by Soleilan:

https://forum.ss13.co/showthread.php?tid=22929

## Changelog 

```changelog
(u)Jan.antilles and Mordent
(*)Mildly Mutated now reinforces the mutation (requiring scrambling or gene research to be removed) and grants 1 trait point.
(*)Geneticists are able to scramble genes at roundstart instead of needing to research radiation emitters.
```


code/datums/traits.dm
```diff
@@ -171,6 +173,10 @@
 				T = trait_instance
 			if(T.afterlife_blacklisted && inafterlifebar(owner))
 				return
+			// TODO: probably fix syntax
```
mordent-goonstation, line 176:
```suggestion
```

code/datums/traits.dm
```diff
@@ -171,6 +173,10 @@
...
+			var/resolved_role = owner?.mind?.assigned_role || src.mind_role_fallback
+			if (T.preventAddTrait(owner, resolved_role))
```
mordent-goonstation, line 178: Wondering if we should have a way to skip this check? e.g. admins trying to add traits that aren't typically available.

code/datums/traits.dm
```diff
@@ -526,21 +535,33 @@
...
 	onAdd(var/mob/owner)
 		var/datum/bioHolder/B = owner.bioHolder
-		var/datum/bioEffect/E = pick(B.effectPool)
-		B.ActivatePoolEffect(B.effectPool[E], 1, 0)
+		var/bioEffectId = pick(B.effectPool)
+		var/datum/bioEffect/E = B.effectPool[bioEffectId]
+		B.ActivatePoolEffect(E, 1, 0)
 		SPAWN (1 SECOND) // This DOES NOT WORK at round start unless delayed but somehow the trait part is logged??
+			if (E)
+				E.curable_by_mutadone = FALSE
```
mordent-goonstation, line 563: Also likely need to flag as modified (or whatever the var is that happens when you splice a chromosome), else people will be able to splice other chromosomes on top (e.g. "Empowered Reinforced Gamma Ray Exposure")

mordent-goonstation (changes requested):

code/datums/traits.dm
mordent-goonstation: Eh, future requirement. If an admin wants to apply the only trait this gets affected by for now, they can just go activate one of their bioeffects and reinforce it using admin tools.

code/datums/traits.dm
jan-antilles: fixed!

imawxome:
It should probably also not work in battle royal mode, a bad roundstart gene can make it very frustrating

jan-antilles:
> It should probably also not work in battle royal mode, a bad roundstart gene can make it very frustrating

They're also MODE roles, so even though Battle Royale can't roll naturally, it's still handled by the changes.

code/datums/traits.dm
jan-antilles: I solved it ANYWAY

mordent-goonstation (approved):
Code seems fine. Would prefer another dev merge, I'm not available to deal with monitoring it for a while because of the tgui mines.

PR #20073, opened 2024-07-23
Rework of the cryoxadone recipe

M-Earthfire:
This is a checkout into a new branch from an old commit of #19312, Because i absolutely managed to fuck up my branch and i found myself kinda incapable of reverting it to this old state on that branch... sorry for the inconveniences.... 

[Chemistry][Balance][Rework][add to wiki]

## About the PR 

Changes the cryoxadone recipe:

from: 
- (1) Cryostylane + (1) Unstable Mutagen+ (1) Plasma + (1) Acetone -> (3) Cryoxadone
to:
- (2) Cryostylane + (1) Platinum + (2) Water -> (2) Cryoxadone @ less than 323K

The reaction now happens over time and slowly cools down the reaction vessel it is in, similary to cryostylane + oxygen.

A chemical reaction effect was added to show the cooling of the liquid

## Why's this needed? 

With the chem rework, many chems were tied to the now much more involved reaction-chain of welding ->oil -> acetone.

While this reaction chain makes multiple reaction paths more interesting, it added additional hurdles to chemicals that were already barely worth it to make.

After the Cryoxadone-rework in #10505, the chems use was highly diminished outside of cryotubes. Thus, this chemical can be moved away from being a endpoint of synthesis to be an advanced precursor-chem. For this, acetone and mutagen were replaced with base-chemicals.

the recipe change enables people to add cryoxadone in further PR's as an indrigient to other chemical recipes. To make this interesting, a cooling effect was added to the reaction so the cold instead of heat needs to be managed. This cooling effect will freeze the water in the reaction, effectively stopping the reaction.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)The recipe of cryoxadone was changed to: (2) Cryostylane + (1) Platinum + (2) Water -> (2) Cryoxadone @ less than 323K
```


Flaborized:
Concepts look good to me and better than other PR's changes though I haven't looked at the code or anything. But a 👍 anyways from me

PR #20091, opened 2024-07-26
Earnable Agent B and Methane canisters for Toxin Lab

kbsmilk:
[SCIENCE] [BALANCE] [FEATURE]

## About the PR 
Adds preset pressure crystal bounties with preset rewards.
* The planned rewards are 1 oxygen agent B canister, and 1 methane canister. Also considering 1 fallout canister.
* These bounties can only be done once per round.
* 1 bounty rewards 1 item, so there will be multiple bounties, but a predictable number of them.
* The required pressure value on the crystal will be randomized. It'll be limited to powers attainable by TTV bombs.
* As a small aside, the Agent B canister is 500L instead of the usual 1,000L. (The sprite I reused is a cute little half canister)
image

So there's the balance plan, more or less. Now here's some implementation details.

The bounties will be in a new section of the Crystal Bazaar PDA app. Like normal bounties, they're claimed simply by launching the crystal out of Cargo. In case of conflict, these special bounties will take sale priority over the standard credit bounties.
Canister delivery will also pass through Cargo. Ideally they go into card-locked science crates; you only get one canister, so you don't want it easily stolen and used for giga-fires.
It won't use the same valuation system as today's crystal sales. Instead it'll have a simple yes/no range: for example, you have to get a crystal pressure between 266 and 271. Anything outside an unclaimed bounty's range gets sent to the regular sale market.

## Why's this needed? 

Primarily, this adds some in-round progression to Toxin Lab. The crystal market is neat, number go up is fun, but it's just credits for Cargo. Here, we're directly linking starter Toxlab gameplay to its endgame: playing with the exotic gases, and maybe learning canister bombs.
Secondarily, it's really tedious and/or difficult to get large quantities of these gases. Especially farts which is just time-consuming button mashing. I don't know the reasons behind the difficulty, so, this is just a draft with some arguments behind it: in case canisters of these gases are a hard "no".

## Changelog 

```changelog
(u)Smilg
(*)placeholder
```

### DIFF
--- a/code/modules/atmospherics/portable/canister.dm
+++ b/code/modules/atmospherics/portable/canister.dm
@@ -124,6 +124,12 @@ ADMIN_INTERACT_PROCS(/obj/machinery/portable_atmospherics/canister, proc/toggle_
 	icon_state = "darkgreen"
 	casecolor = "darkgreen"
 
+/obj/machinery/portable_atmospherics/canister/agentb
+	name = "Canister \[Agent B\]"
+	icon_state = "bluish"
+	casecolor = "bluish"
+	volume = 500
+
 /obj/machinery/portable_atmospherics/canister/update_icon()
 	if (src.destroyed)
 		src.icon_state = "[src.casecolor]-1"
@@ -797,5 +803,12 @@ ADMIN_INTERACT_PROCS(/obj/machinery/portable_atmospherics/canister, proc/toggle_
 	..()
 	if (!src.isempty)
 		src.air_contents.farts = (src.maximum_pressure*filled)*air_contents.volume/(R_IDEAL_GAS_EQUATION*air_contents.temperature)
+	src.UpdateIcon()
+	return 1
+
+/obj/machinery/portable_atmospherics/canister/agentb/New()
+	..()
+	if (!src.isempty)
+		src.air_contents.oxygen_agent_b = (src.maximum_pressure*filled)*air_contents.volume/(R_IDEAL_GAS_EQUATION*air_contents.temperature)
 	src.UpdateIcon()
 	return 1


444Portal:
YEAAAAHHH

GabrielMation:
Hmm, i was first worried that it would be a "put molitz beta into the gas extractor" pr, but this actually seems like a great way to give an alternate access point to agent B, AND make the Crystal Bazaar actually get looked at for once, instead of sitting in the RD's and QM pda gathering dust. Molitz beta will (and SHOULD), stay viable, along with its own quirks, though admittedly it IS pretty much impossible to figure it out without reading the code (at least it was when i was messing with it a year ago. it should be almost viable with the few changes to the wiki)

But i digress. My potential issue with this against molitz beta is that this is going to be much, much, much more gas than you could get from molitz. A crystal offgases about 2.5 kilopascals (iirc, when in a scruber. those are 1000 liters right?) before it's fully depleted. The point being, it's a very small amount. You need a few xrystals to break the thermodynamic limit, when extracted optimally. which  uhh it's actually a small amount. say like 8 if you do it optimally. which is actually not that much. 

uhhh where was i going with this? ah yes. you don't _need_ to give people so much agent b. The actual proportion needed to break the thermodynamic barrier is under 0.5% of the total gas volume (and that seems fair to me). Of course there's gonna be the odd engineer that puts it in the TEG or reactior, but it's just... already very little. So  basically, if you tolerated my aimless rant, you may as well dilute it, maybe. It's basically there for hellburns. Actually, don't. or do. fuck. It's fine, probably leave it as is. My worry was that it could potentially be powergame-rushed in under 20 minutes. An optimized spawn-to-hellburn molitz beta rush could take like 20 to 25 minutes, or maybe less, depending on telescope rng. If you can get far more, faster gas by just making a bomb, then it seems a bit problematic. but then again, it doesn't actually matter, since the amount needed to break teh heat limit is very small. So basically, do whatever you want. Rant over.

itchybones:
Love this PR, as it gives Toxins Scientists more goals than just "make le big bomb". Making these gases more accessible is probably gonna help newbies figure out _what the fuck they actually do_ and make experimenting with them less of a pain.

thebee03:
PLEASE PLEASE PLEASE. I CAN DO LIKE REALLY CONROLLED RECTIONS NOW

Azrun:
`/obj/item/storage/briefcase/toxins`
`/obj/item/paper/hellburn`
Paper and possibly the briefcase need updating if agent b becomes more available.

MyNameIsRomayne:
Code as-is (I know it's far from done) looks good. Note that I am sitting on PRing an air scrubber functionality extension, wherein you can blacklist scrubbing additional gases (So you can choose not to scrub more than just O2/N2) to facilitate using crystals for gas extraction. So whenever TGUI is unfrozen, that'll go up

PR #20112, opened 2024-07-28
Allow gravitons inside mechcomp cabinets.

klookiyam:
[Game-Objects][Feature]

## About the PR 
Allows gravitons to be attached within mechcomp cabinets and gives it a secondary impulse behavior when inside of them. (credit to penultimatum for the concept)

When mounted inside cabinets, instead of throwing the tiles contents 50 tiles in a direction, it throws only the cabinet 3 tiles. This does 10 damage, tuned significantly less than other lockers/carts as a balance measure. All gravitons in a cabinet share the same cooldown - no stacking them to do instantaneous bonks.
ezgif-1-4cb7bcf58c

This also works in mechcomp cabinets with a movement component, movement resumes after the throw completes.
ezgif-1-11565f4bc8

How the components look inside a cabinet:
bonklocker

## Why's this needed? 
More actuator options for mechcomp contraptions are good. Mechcomp cabinets are interesting but their main use case is as a mobile signal repeater. It would be fun to let them bonk things.

## Changelog 

```changelog
(u)KakoLookiyam
(*) Mechcomp cabinets are now capable with of equipping internal gravitons.
(+) Mechcomp cabinets can now attach internal graviton components. When attached, a graviton is only able to fling its enclosing cabinet itself a shorter distance than a floor mounted graviton.
```


klookiyam:
...I realized after I published this that each graviton having its own cooldown can be abusable.

RSG250:
i immediately thought of what would happen if it was just full of gravi and a single auto signaler

Azrun (commented):
Does the intended behavior still exist with the cooldown adjustment? Didn't try it prior, but it didn't seem very exciting outside of trying to squish/push people in confined spaces but I didn't try anything elaborate.

MyNameIsRomayne:
This seems a little dangerous. How would the crew be expected to stop such an autonomous machine from bonking aforementioned things? (windows, people, doors)

klookiyam:
> Does the intended behavior still exist with the cooldown adjustment? Didn't try it prior, but it didn't seem very exciting outside of trying to squish/push people in confined spaces but I didn't try anything elaborate.

The original graviton functionally still exists. The big one is the ability to push other movables, which a movement component on the component cabinet can not currently do. I agree that bonking on mobs might be a bit boring...I might be more interesting as a very low damage slip or something else.

> This seems a little dangerous. How would the crew be expected to stop such an autonomous machine from bonking aforementioned things? (windows, people, doors)
The easiest way would be a wrench - once tapped by a wrench and anchored the cabinet cannot move/graviton. Cabinets also shouldn't be able to trap people - while moving and even with this change they're incapable of pusing/pulling a mob. In addition they're only dense while you're standing, so you are able to slip out under them.

I did test shooting them with bullets/hitting them with melee - they are immune to that despite appearing to be closets. Funnily enough they can burn since they inherit storage properties.

MyNameIsRomayne:
Huh, neat! Never knew you could just wrench em to being off.

MyNameIsRomayne (approved):
in any case, the code looks fine

MyNameIsRomayne:
dreamseeker_NZkPbjpzQ5

https://github.com/user-attachments/assets/38fef417-bd3e-4226-93b0-381fa96ea398

Warming up to this actually. Ran this for a test to make sure the code works and it sure does

BinneBoi:
Consider adding a spin/sound/description or some other que so its clear that it is jumping. I love this idea but i think it might be unclear what its doing until something is hit by it.

TobleroneSwordfish (approved):
Funny and relatively harmless

PR #20127, opened 2024-07-29
Makes secondary radiation more dangerous

TobleroneSwordfish:
[BALANCE] [EXPERIMENTAL]
## About the PR 
Reverts the scaling on radioactive components from inverse square (realistic but very steep) to linear (bad dumb and video gamey but with a lot less instant falloff) and doubles the passive dosage you get from standing next to things.

## Why's this needed? 
Radiation doesn't feel like a credible threat at the moment, even after reactor meltdowns the scary looking fallout and green glowing walls tend to equate to <10 burn damage in reality.

## Changelog 

```changelog
(u)LeahTheTech
(*)Radioactive things are now more dangerous to stand next to, you will accumulate a significant dose a lot faster.
```


OmegaShoots:
Threw two radioactive ghost hairs out of the escape shuttle and went into deep crit from full hp in 20 seconds and straight up died with a mediborg injecting me with many meds in another 20 seconds.

PR #20129, opened 2024-07-29
Add radioactive damage for tank explosions which use fallout gas

MyNameIsRomayne:
## About the PR 

Allows explosions to also do radioactive damage, granted there's a notable amount of fallout inside the tanks.
Mobs within range will take a sudden and potentially lethal dose of radiation depending on how much fallout was used for the explosion, and the explosion itself will also emit some amount of neutrons.

##  Video is to demonstrate mechanics and does not reflect balance tweaks made during the PR
(i just cant be re-recording this each time)

https://github.com/user-attachments/assets/99885c60-2eb6-472d-b7cb-dcddf97ce3e2

Adds a define for the default tank volume (70 Litres). Not required/related to the changes, just a code quality improvement as I went along.

## Why's this needed? 
The means to produce fallout are there, but go unused due to there being no incentive for players to create any -- radioactive gas floods are impractical, the actual rad damage is mediocre, and plasma just ends up doing it better.

Adding mechanical depth to TTV explosions / tank explosions should make it more engaging overall, and potentially open up a new avenue for the bazaar to do rad explosion bounties! I'd like to do more of these to add a sort of mechanical depth to the system of tank explosions, making it more interesting overall.

## Changelog 

```changelog
(u)Romayne
(*)Tank explosions will now become a new kind of devastating when fallout is added.
```


MyNameIsRomayne:
Higher quality version of the clip on the discord (https://discord.com/channels/182249960895545344/890226559691157524/1273770554834550814), if you want. Its not perma but its there

code/obj/item/tank.dm
```diff
@@ -9,6 +9,8 @@ Contains:
 -Plasma Tank
 */
 
+#define TANK_VOLUME_LITRES 70 //! The volume of a normal tank in litres
```
Sovexe, line 12: would like to see `LITERS` added into `_units.dm` would also prefer using the American spelling for consistency

MyNameIsRomayne: Dangit im american and i still used litres. how could i let this happen to myself

MyNameIsRomayne: I did this in #20200 but in light of that PR basically getting left to collect dust I guess I could just put it here.. but 🤷‍♀️

code/obj/item/tank.dm
```diff
@@ -9,6 +9,8 @@ Contains:
 -Plasma Tank
 */
 
+#define TANK_VOLUME 70 * LITERS //! The volume of a normal tank in litres
```
Sovexe, line 12:
```suggestion
#define TANK_VOLUME 70 LITERS //! The volume of a normal tank in litres
```

PR #20131, opened 2024-07-30
Adds welded sheet masks as a craftable item

colossusqw:
[LABEL][feature][clothing][sprites]

## About the PR 

This PR adds the steel sheet welded mask. It posseses good head protection and some disorient resist from flashes, at the cost of lowering visibility. It can be crafted by using a welder on a stack of at least 5 steel sheets, and can be altered with a welder to increase visibility, at the cost of halving protection and removing all disorient resist.

image

## Why's this needed? 

I've been wanting some different mask options for quite a bit, and this felt like a nice option with some usability, but a mostly cosmetic alternative version as well.

## Changelog

```changelog
(u)Colossus
(+)Added metal masks to the game, which can be crafted by using a welder on a stack of 5 or more steel sheets. Masks offer protection in exchange for visibility.
```


Tarmunora (approved):

code/obj/item/clothing/masks.dm
```diff
@@ -629,6 +629,45 @@ TYPEINFO(/obj/item/clothing/mask/wrestling)
 	item_state = "anime"
 	see_face = FALSE
 
+/obj/item/clothing/mask/steel
+	name = "sheet welded mask"
+	desc = "A crudely welded mask made by attaching bent sheets. Highly protective at the cost of visibility"
+	icon_state = "steel"
+	item_state = "steel"
+	see_face = FALSE
+	var/low_visibility = TRUE
```
Azrun, line 638: `	allow_staple = 0`
Should staples be able to puncture this?

code/obj/item/building_materials.dm
```diff
@@ -226,6 +226,29 @@ MATERIAL
...
+			if(!W:try_weld(user, 1))
+				return
+
+			var/obj/item/clothing/mask/steel/M = new /obj/item/clothing/mask/steel(user.loc)
+			if(src.material) M.setMaterial(src.material)
+			src.change_stack_amount(-5)
+
+			user.visible_message(SPAN_ALERT("<B>[user]</B> welds the sheets together into a mask."))
+			UpdateStackAppearance()
+			if(src.amount < 1)	qdel(src)
```
Azrun, line 249:
```suggestion
```
Isn't this covered by `change_stack_amount()`? Or did you run into times the stack wasn't removed?

colossusqw: I followed the same method used by the sheet stacking and unstacking code, if it's redundant here, it may be redundant there as well

itchybones:
Can the visibility debuff be influenced by genes like X-Ray vision? It'd make sense to be able to see through the masks

icons/effects/overlays/steelmask.dmi~
Azrun, line 1: This shouldn't be.

colossusqw:
Found out if you were to put both a welding mask and a steel mask, then remove one, the result would rid you of the vision penalty, so i gave the welding mask it's own separate overlay.

icons/effects/overlays/steelmask.dmi~
colossusqw: oh?

PR #20135, opened 2024-07-30
Raise the volume of Cow Moo

kbsmilk:
[MUTANTRACE] [SOUND]
## About the PR 
Raises the volume of moo from 50 to 100.

## Why's this needed? 
I think it's too quiet. Here's a comparison video:

https://github.com/user-attachments/assets/5eef4b10-0dcc-4632-afd0-b4316137ac8b

### DIFF
--- a/code/datums/mutantraces.dm
+++ b/code/datums/mutantraces.dm
@@ -2250,7 +2250,7 @@ TYPEINFO(/datum/mutantrace/cow)
 					return // use muted scream emote handling
 				if (src.mob.emote_check(voluntary, 50))
 					. = "<B>[src.mob]</B> moos!"
-					playsound(src.mob, 'sound/voice/screams/moo.ogg', 50, 0, 0, src.mob.get_age_pitch(), channel=VOLUME_CHANNEL_EMOTE)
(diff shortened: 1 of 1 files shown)


ZeWaka:
i think screams are too loud, not moos too quiet

Azrun:
> I hope you mean in general, because this is a dramatically decibel boosted video for comparison purpose. PR is to equalize perceived volume; lowering the volume of all screams in game is out of scope

Understood. I agree with ZeWaka that it would be preferable to have a lower general scream level compared to achieving moo scream parity.

YoSnoid:
imo 100 is way too loud, i could see it bumped up to like 60 or 70 at most. moos shouldnt be as loud as a person screaming anyway.

Glamurio:
I agree with the opinion that all scream volumes should be lowered rather than boosting moos.

MyNameIsRomayne:
Also agree, I think it's good if all the screams are as loud as each other for the sake of, y'know, being heard equally when getting dragged into maints, but current screams feel too loud anyway (not just because you boosted the volume)

ZeWaka:
I'm going to close this based on my opinion and community feedback in seeking an alternate solution.

PR #20138, opened 2024-07-30
Fix typo in Brick Suicide text

NibChocolateeny:
[Bug][Player Actions]

## About the PR
fixes the tyopo in texxt for the funny new brick suiside
Originally is "[user] throws the the brick into the air!."
now it is "[user] throws the brick into the air!"

## Why's this needed?
No ypos is good, i think.


NibChocolateeny:
@TheTyrant25 ok?

PR #20142, opened 2024-07-31
Adds the brick random room

TheTyrant25:
[MAPPING]
## About the PR
Room with four bricks in it.
image

## Why's this needed?
What can I say, sometimes you just need a brick. Or four.


TobleroneSwordfish:
Yes.

PR #20146, opened 2024-07-31
Replace newspaper's hit sound

TheTyrant25:
[SOUNDS]
## About the PR
Used generic stab 1, because even though it's meant to sound like stabbing, it doesn't really.

## Why's this needed? 
Makes it sound more like being hit by a rolled up newspaper rather than something else.


TobleroneSwordfish (approved):
Weirdly this does sound a lot more like being hit with a newspaper than being stabbed

Gannets (approved):

PR #20150, opened 2024-08-01
Circuit Board Vending Machine

Cherman0:
## About the PR 

image
Adds a vending machine that holds the circuit boards needed to create specialized consoles. Does not add them to any maps. Most boards cost 50c, though some of the hidden ones are more or less expensive.

## Why's this needed? 

Allows access to every circuit board, not just the handful that spawn in the piles in tech storage. It would be impractical to scale the pile of circuit boards as the number of boards in the game increases. Vending machines support a scrollbar and ctrl+f searching, making them better suited to large lists of items.


code/modules/vending/vending.dm
```diff
@@ -1707,6 +1707,71 @@ ABSTRACT_TYPE(/obj/machinery/vending/cola)
...
+		..()
+		//Normally available circuits:
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/security_tv, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/small_tv, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/communications, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/teleporter, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/secure_data, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/atmospherealerts, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/atmospheresiphonswitch, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/air_management, amount=2, cost=PAY_UNTRAINED/3)
```
JORJ949, line 1735: 2 entries for air_management

code/modules/vending/vending.dm
```diff
@@ -1707,6 +1707,71 @@ ABSTRACT_TYPE(/obj/machinery/vending/cola)
...
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/mining_magnet, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/telescope, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/announcement, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/siphon_control, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/chem_request, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/chem_request_receiver, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/sea_elevator, amount=2, cost=PAY_UNTRAINED/3)
+		//Hidden circuits available by hacking:
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/clown_announcement, amount=1, hidden=TRUE, cost=PAY_DUMBCLOWN)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/robotics, amount=1, hidden=TRUE, cost=PAY_EXECUTIVE)
```
JORJ949, line 1764: If you want this used anywhere ever, it should not have the killswitch, that should take effort to steal.

code/modules/vending/vending.dm
```diff
@@ -1707,6 +1707,71 @@ ABSTRACT_TYPE(/obj/machinery/vending/cola)
...
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/security, amount=1, hidden=TRUE, cost=PAY_EXECUTIVE)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/bank_data, amount=1, hidden=TRUE, cost=PAY_EXECUTIVE)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/card, amount=1, hidden=TRUE, cost=PAY_DONTBUYIT)
+		//not present because because deprecated/nonfunctional to build:
```
JORJ949, line 1768: If deprecated, why add at all?

code/modules/vending/vending.dm
```diff
@@ -1707,6 +1707,71 @@ ABSTRACT_TYPE(/obj/machinery/vending/cola)
...
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/solar_control, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/powermonitor, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/powermonitor_smes, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/barcode, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/barcode_qm, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/operating, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/qmorder, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/qmsupply, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/stockexchange, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/transception, amount=2, cost=PAY_UNTRAINED/3)
```
JORJ949, line 1754: Should have a check here for if its nadir so you dont end up with pointless transception controls

JORJ949 (commented):
Review mostly made assuming this is intended to be player-usable (otherwise why add it?)

code/modules/vending/vending.dm
```diff
@@ -1707,6 +1707,71 @@ ABSTRACT_TYPE(/obj/machinery/vending/cola)
...
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/mining_magnet, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/telescope, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/announcement, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/siphon_control, amount=2, cost=PAY_UNTRAINED/3)
```
soleilan, line 1758: Include this in Nadir check

code/modules/vending/vending.dm
```diff
@@ -1707,6 +1707,71 @@ ABSTRACT_TYPE(/obj/machinery/vending/cola)
...
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/chem_request, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/chem_request_receiver, amount=2, cost=PAY_UNTRAINED/3)
+		product_list += new/datum/data/vending_product(/obj/item/circuitboard/sea_elevator, amount=2, cost=PAY_UNTRAINED/3)
```
soleilan, line 1761: Include this in Oshan/Nadir check

code/modules/vending/vending.dm
soleilan: Ditto, those are usually placed in computer core as a circuit board on all maps for a reason

pgmzeta:
Can you please explain the rationale behind charging credits for these?

Cherman0:
Mostly because some of the other tech storage vendors also do. In theory the price could be balancing but the funny free stuff mechcomp button exists, so its basically just fluff.

As for map checking, I'd rather it didn't. Map checks add extra random spots that need changed if maps are added or changed, so as long as it doesnt introduce broken functionality I'd rather just have items that arent useful.

Im not against removing robotics or id computer boards, though since you need a command ID to actually use them it doesnt seem like an issue to me. Since at that point you could just use the robotics console in the bridge or ai areas.

Cherman0:
Removed the robot control board. Also fixed that the computer type of the clown announcement computer was a string instead of a typepath, which would prevent it from being used to make a console.

Cherman0:
Using mapchecks to minimize the size of a list of items is not really necessary for vending machines, since they're much easier to navigate than the rightclick list of items on the ground. You could similarly argue that the pump control boards shouldn't be available on maps with no pumps, but there's no real reason to remove them and it just adds work later.

JORJ949:
> Using mapchecks to minimize the size of a list of items is not really necessary for vending machines, since they're much easier to navigate than the rightclick list of items on the ground.
>

All the boards look the same unlike other vending machines which you can visibly see so you have to read, cutting down unnecessary bloat I think there isnt much reason to not do, pumps are on almost every map and solars you just do a NOT oshan or nadir check for

soleilan:
Yeah what they said, but also think about it like this. What if making a sea elevator computer throws mad runtime errors on maps without one? Are you going to go through all of these circuits and make sure they don't throw errors on maps that don't support them? It's just much easier to make sure they can't be built at all. Pumps are on every single map IIRC since they are part of Toxins but yeah I would add logic for solars, transception arrays, sea elevators, siphon control. Even if they don't throw errors, there's literally no reason for them to be buildable and confuse new players or invite people to find ways to make them throw errors IMO.

Cherman0:
Eh, sure. It's added.

Cherman0:
And for context, I haven't added it to maps as a matter of scope. I'd rather not fix map merge conflicts all day while the matter at hand is wether it's actually wanted in the first place at all.

soleilan (approved):
Can't speak on the topic if its wanted or about the balance, but code-wise it looks all good.

JORJ949 (approved):
Looks good to me, though I think it should be added to maps in the same PR tbh.

PR #20151, opened 2024-08-01
remove soulsteel from artifact transmutor bomb material pool

Retrino:
## About the PR 

title. simple one line change, removes soulsteel from possible transmutation bombs

## Why's this needed? 

because bombs coat all objects in their radius in their material, you can end up in situations where unbreakable items get coated in soulsteel. ghosts can then possess the items and block entire areas off with basically no way to deal with it aside from blowing the objects up. while rare, it really just shouldnt be happening in general

### DIFF
--- a/code/obj/artifacts/artifact_objects/bombs.dm
+++ b/code/obj/artifacts/artifact_objects/bombs.dm
@@ -378,7 +378,6 @@ ABSTRACT_TYPE(/datum/artifact/bomb)
 					100;"silver",
 					100;"cobryl",
 					50;"miracle",
-					50;"soulsteel",
 					50;"hauntium",
 					50;"ectoplasm",
 					10;"ectofibre",


RelentlessGarbage:
You'd also need to remove Hauntium from that materials list if the goal was to prevent all possession-based shenanigans. Whether or not it makes sense to get rid of all possession-based shenanigans from artifact bomb aftermaths is not something I will comment on.

Retrino:
> You'd also need to remove Hauntium from that materials list if the goal was to prevent all possession-based shenanigans.
> I won't speak to the need for this change since I have not been on in a while and never encountered this.

i thought this at first as well but 444 let me know you can actually walk through hauntium so its not much of an issue

Retrino:
image

TobleroneSwordfish:
Seems like a really niche issue that would be better addressed by making more thing deconstructible/destroyable

Retrino:
> Seems like a really niche issue that would be better addressed by making more thing deconstructible/destroyable

its true its a niche issue, which is why the pr completely resolves it. things that can currently be plated by soulsteel are either movable or deconstructable, which balances it out pretty well (even if the objects like lockers can be annoying, you can still deal with them). id disagree that the right way to solve this is by making the objects deconstructable purely because of the amount that there are, or the complexity to which they should be deconstructable. things like the replace-ability of the object (to my knowledge things like podbay buttons have no replacement, making them breakable could make it really easy to fuck over things like cargo/engineering/podbays) could make it more complicated and out of the overall scope/goal of this specific pr. 

im not opposed to making another pr to make more objects deconable, because more stuff *should* be, but for now this addresses the main overall issue while still letting ghosts be silly with plated stuff via hauntium or just plating stuff in soulsteel.

Tarmunora (approved):

ZeWaka (approved):

MyNameIsRomayne:
Ive been having medass spam at me for several years every day to make things more deconnable. It's just a lot of work, and you have to understand a lot of mappers use items as convenient barricades against breakins. Doesnt make it infeasible, but it gets made much more difficult because of that

PR #20160, opened 2024-08-01
Prevent afterattacks when clicking on vis_contents contents

### LINKED ISSUE #20153: Splashing liquids on plated items has no distance limit
### Labels

[BUG][MINOR]

### Description

You can splash liquids onto a plated food item from any distance

### Steps to reproduce

Get a pitcher of any fluid
Click on a food item on a plate
Pour it from like 10 tiles away

### Expected Behavior

Not be able to pour a pool of liquids onto far away objects

Reported by: solenoid
Client version: 515.1636
Active test merges: #20127
Round log date: 2024-08-01-04-31
Reported on: 2024-08-01 05:57:58
Round ID: 184733
Map: NADIR

pgmzeta:
[bug][game objects]
## About the PR 
Count items placed on plates and pizza bases as pixelable, which prevents afterattacks from firing.

## Why's this needed? 
Fix  #20153

### DIFF
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -388,10 +388,10 @@
 			if (istype(target, /atom/movable))
 				if (isturf(target.loc))
 					pixelable = TRUE
-				else if (istype(target.loc, /obj/item/plate))
-					pixelable = TRUE
-				else if (istype(target.loc, /obj/item/reagent_containers/food/snacks/ingredient/pizza_base))
-					pixelable = TRUE
+				else if (istype(target.loc, /atom/movable))
+					var/atom/movable/location = target.loc
+					if (target in location.vis_contents)
+						pixelable = TRUE
 		if (pixelable)
 			if (!W.pixelaction(target, params, src, reach))
 				if (!QDELETED(W))
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -385,8 +385,13 @@
 	if (!QDELETED(W) && (equipped() == W || usingInner))
 		var/pixelable = isturf(target)
 		if (!pixelable)
-			if (istype(target, /atom/movable) && isturf(target.loc))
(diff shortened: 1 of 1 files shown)


code/mob/living.dm
```diff
@@ -385,8 +385,13 @@
...
 		if (!pixelable)
-			if (istype(target, /atom/movable) && isturf(target.loc))
-				pixelable = TRUE
+			if (istype(target, /atom/movable))
+				if (isturf(target.loc))
+					pixelable = TRUE
+				else if (istype(target.loc, /obj/item/plate))
+					pixelable = TRUE
+				else if (istype(target.loc, /obj/item/reagent_containers/food/snacks/ingredient/pizza_base))
+					pixelable = TRUE
```
TobleroneSwordfish, line 394: Well no, again this code should never be getting here! Look into why it's passing range checks instead of adding a snowflake fix here.

TobleroneSwordfish (changes requested):

code/mob/living.dm
```diff
@@ -385,8 +385,13 @@
 	if (!QDELETED(W) && (equipped() == W || usingInner))
 		var/pixelable = isturf(target)
 		if (!pixelable)
-			if (istype(target, /atom/movable) && isturf(target.loc))
```
TobleroneSwordfish, line 388: Okay I looked into this because this solution is still very oddly specific for the case we're handling here. The problem seems to be that the code assumes we won't ever be able to click on something which is inside something else unless it's in our inventory. So a better solution might be just replacing the `isturf(target.loc)` with `(isturf(target.loc) || !reach)`, that way if we can't reach something we will always trigger the pixel action and cancel the `afterattack`.

TobleroneSwordfish (changes requested):

PR #20163, opened 2024-08-02
Fix to prevent generators from permanently coupling generators together

klookiyam:
[Station Systems][Bug]

## About the PR 
This is a small hack to prevent experimental local generators from accidentally permanently coupling powernets together. The `Move` proc of `machinery/power` rebuilds the powernet every tile they move. When the generator arrives at an intersection of two unconnected cables, it acts as an omnidirectional connection between every wire on that turf that can't be decoupled.

 The `obj/machinery/power` type seems like it was designed to only be used for machinery that push or pull power from wires on the same turf as them. The prototype generator doesn't interact with wires (it magically beams power to the rooms apcs) and shouldn't ever be rebuilding power nets. Also, it probably probably shouldn't be a subtype of `obj/machinery/power` at all...

## Why's this needed? 
Magically coupled powernet's that people can't figure out how to fix are **very bad**.

### DIFF
--- a/code/modules/power/chem_combustion.dm
+++ b/code/modules/power/chem_combustion.dm
@@ -18,6 +18,7 @@ TYPEINFO(/obj/machinery/power/combustion_generator)
 	anchored = UNANCHORED
 	flags = FLUID_SUBMERGE | NOSPLASH
 	deconstruct_flags = DECON_SCREWDRIVER | DECON_WRENCH | DECON_CROWBAR | DECON_WELDER | DECON_WIRECUTTERS
+	interacts_with_wiring = FALSE
 
 	var/active = FALSE
 	var/packet_control = TRUE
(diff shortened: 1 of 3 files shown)


TobleroneSwordfish:
I think ideally we'd want power machinery to connect to only one pnet in the criss-cross situation, this won't fix hamster wheels for example (unanchored but need to connect to wires).

klookiyam:
> I think ideally we'd want power machinery to connect to only one pnet in the criss-cross situation, this won't fix hamster wheels for example (unanchored but need to connect to wires).

Yeah, that makes sense and it seems like a proper fix might be more involved. I'm going to dig back to look more closely at how powernet building works. One thing is I don't know what sort of logic you'd use to determine which wire in the cross to use for more ambiguous pieces of `machinery/power`. I'll close this until a better fix comes to mind.

PR #20164, opened 2024-08-02
Thonagarang nerf + grammatical errors

### LINKED ISSUE #20149: Error in description of dirty garbage flip-flops
### Labels

[BUG][TRIVIAL]

### Description

Text description contains typo/duplication?

### Steps to reproduce

Examine and see:

This is a dirty garbage flip-flops. 
These cheap sandals don't even look legal. These cheap slops don't even look legal.

### Expected Behavior

These cheap sandals don't even look legal.

Reported by: airconswitch
Client version: 515.1641
Active test merges: #20127
Round log date: 2024-08-01-00-14
Reported on: 2024-08-01 01:54:46
Round ID: 184718
Map: DONUT3

IPling:
[BUG] [BALANCE]

## About the PR 

While looking at #20149 I found out that Shitty Bill's sandals can be turned into a boomerang, that deal 5 throwing damage, and we don't have cooldowns on throwing. Thankfully nobody knows about the existence of this thing yet.
So I reduced its throwing damage to 0 and fixed #20149 + a small grammatical error.

## Why's this needed? 
I fear when someone discovers this, the stations windows? Gone to smithereens, and the stations crew? Destroyed by the almighty power of the thongarang.


IPling:
> Dunno, a massive preemptive nerf for something that could be solved giving the item the treatment of the safari kit boomerang seems very odd,. Thus applies even more so for something completely unused that most likely falls under scrutiny of secret stuffs gentlemans agreement aká "don't abuse it, else its getting nerfed"

I do think the gentlemen's agreement could apply here, yet this is much more stupidly easy to obtain compared to the OHR in ice moon or the beef case while also being more if not as powerfull as them. When its really just a joke item that really shouldnt be this powerfull.

With giving it different treatment, yeah it could be solved by giving it a more interesting nerf than just removing the damage. Though that seems like a very trivial decision for a joke item that really shouldnt have any practical effect other than well, being a joke. If anybody wants to give it something more interesting, then sure go ahead.

code/WorkInProgress/warcrimes.dm
```diff
@@ -28,7 +28,6 @@ TYPEINFO(/area/diner/juicer_trader)
 
 /obj/item/clothing/shoes/thong
 	name = "garbage flip-flops"
-	desc = "These cheap sandals don't even look legal."
```
ZeWaka, line 31: why rm

code/WorkInProgress/warcrimes.dm
```diff
@@ -47,7 +46,7 @@ TYPEINFO(/area/diner/juicer_trader)
 			boutput(user, "You staple the [src] together to create a mighty thongarang.")
 			name = "thongarang"
 			icon_state = "thongarang"
-			throwforce = 5
+			throwforce = 0
```
ZeWaka, line 49:
```suggestion
			throwforce = 1
```

ZeWaka (commented):
sounds like we should just add a cooldown to throwing

code/WorkInProgress/warcrimes.dm
IPling: Fixes the double description from #20149 
Unless there is a nother way to do it.

code/WorkInProgress/warcrimes.dm
IPling: I dunno, I find this to be a funny item that people can just spam throw at someone else. Making it do actual damage - even if its 1 - while its spammed can do around 30 brute in a few seconds, which I find a bit too much since people would probably wanna use this as a joke without actually hurting the person. 
The -1 melee damage reduction for jumpsuits doesnt seem to negate it.

IPling:
> sounds like we should just add a cooldown to throwing

People being able to throw basketballs _or gods for sake knifes_ at each other at supersonic speeds is really funny and it doesnt really have any gameplay impact since both parties are willingly participating. So getting rid of that because 1 obscure item was unbalanced around it sounds unfun.
Although if it was a toggleable thing then it could bring up some potential for weaponry.

ZeWaka:
> People being able to throw basketballs _or gods for sake knifes_ at each other at supersonic speeds

you made a comment about how this could be used to annihilate windows and such
does the lack of a throwing cooldown not apply to those also (yes, I know you'd need to pick them back up)

IPling:
> you made a comment about how this could be used to annihilate windows and such does the lack of a throwing cooldown not apply to those also (yes, I know you'd need to pick them back up)

That is definitely a nother negative about no throwing cooldowns, the thonagarang can destroy reinforced windows and plasmaglass windows very efficiently. The safari boomerang in theory can aswell, yet the 1 throwing damage makes it somewhat slower than just using a crowbar, screwdriver and cutters + its highly ineffective without the safari suit making it a significantly rare event.
For stuff that doesnt return to you when your throw it, the picking up part and then throwing it yet again is practically the cooldown for such things.
Unless lightspeed throwing is that icky of a thematic mechanic, where it just doesnt fit in the game. The only reason I can really see for cooldowns to be added is when someone decides to make a boomerang-like object that is used to deal damage, unlike the safari boomerang which is used to stun people.

ZeWaka:
hmm @Tarmunora what do you think, i think it can probably be lowered to zero, but lightspeed throwing is just jank imo

PR #20172, opened 2024-08-03
Announcement Circuit Boards

JORJ949:
## About the PR 
Gives each departmental announcement computer its own kind of circuit board

## Why's this needed? 
Deconstructing and rebuilding any announcement computer presently just resets it to bridge access which is not good
Further PR where I put the department announcement computer boards in command lockers for maps that dont have the announcement computers

## Changelog 

```changelog
(u)JORJ949
(+)Departmental announcement computers now have their own circuit boards.
```


Cherman0:
Looks good to me. Does having multiple bridge announcement computers actually duplicate the arrivals messages? If it does then fixing that would be a better solution, though this works as a tempfix.

JORJ949:
> Looks good to me. Does having multiple bridge announcement computers actually duplicate the arrivals messages? If it does then fixing that would be a better solution, though this works as a tempfix.

image
yes

PR #20173, opened 2024-08-03
Shift around Oshan security a bit

JORJ949:
## About the PR 
Move the HoS announcement computer into the HoS office 
Move the rechargers and phone off the main sec table to make space for disks and pinpointers
Replace spread out "desk" officer spawns with assistants, officers spawn around main table
Push sec area one tile into armory to expand walkable area
Expand sec table for more space
image

## Why's this needed? 
Im paranoid every time I announce something as HoS on oshan that someone will grab my ID
There is no space for pinpointers and disks unless I want to go grab a decon tool and start destroying stuff
the main walkway from the front to back of sec was cramped at the entrance only being a 1 tile corner, now its bigger

### DIFF
--- a/maps/oshan.dmm
+++ b/maps/oshan.dmm
@@ -37614,9 +37614,7 @@
 /obj/disposalpipe/trunk/south{
 	pixel_z = 1
 	},
-/obj/machinery/disposal/small{
-	dir = 8
-	},
+/obj/machinery/disposal,
 /turf/simulated/floor,
 /area/station/security/main)
 "ksg" = (
--- a/maps/oshan.dmm
+++ b/maps/oshan.dmm
@@ -1024,12 +1024,6 @@
 	dir = 1;
 	icon_state = "pipe-c"
 	},
-/obj/machinery/camera{
-	c_tag = "autotag";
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Looks good, one tiny nitpick: I'd rather keep the disposal chute as a full one because that makes it more obvious as an escape route (which is a good thing, I think?)

JORJ949:
> Looks good, one tiny nitpick: I'd rather keep the disposal chute as a full one because that makes it more obvious as an escape route (which is a good thing, I think?)

Yeah good point, I just put it on the wall because the room already feels cramped (also its a pretty rubbish escape considering disposals is right there)

TobleroneSwordfish (approved):

MyNameIsRomayne (changes requested):
image
Is this there in the final changes? Looks like you forgot to rotate/edit this lamp by the entrance to main security, it's floating.

MyNameIsRomayne (approved):
Disregard, I apparently didn't know lights auto choose dirs now. This looks cool!

MyNameIsRomayne:
image

PR #20174, opened 2024-08-03
Unify Turret Control ID behaviour

JORJ949:
## About the PR 
Creates subtypes for turret controls for the below areas with their areas and access defined and places them on all rotation maps.
-AI Chamber
-AI Upload
-AI Perimeter
-Armory
-Armory Perimeter
-Computer Core

I also replaced the "AI satellite" area on donut 3 with "AI chamber" because donut 2 only uses that for the non-turret protected area and I didnt make a subtype for AI satellite for one map.

## Why's this needed? 
Fun fact: some armory turret controls could be used by anyone with upload access because they were not varedited to be null

### DIFF
--- a/code/obj/machinery/turret.dm
+++ b/code/obj/machinery/turret.dm
@@ -356,7 +356,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/turretid, proc/toggle_active, proc/toggle_le
 	var/emagged = 0
 	var/turretArea = null
 
-	req_access = list(access_ai_upload)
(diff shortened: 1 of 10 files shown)


JORJ949:
I have immediately instead considered just making subtypes that automatically set their turretProtectedArea and req_access to where they should be protecting and thats probably a better solution tbh

MyNameIsRomayne (approved):
This seems objectively good. Code looks fine, though I didn't load up to test any of it

As long as everything was properly mapped in strongdmm should be good

PR #20175, opened 2024-08-03
Add more radiation shields to Donut3 engineering

Cherman0:
## About the PR 

StrongDMM-2024-08-03 12 28 31

Adjusts a few things about Donut3 engineering so that it is possible to access all of the rooms around the singularity without being irradiated if the singularity is active. This includes connecting the engineering lounge to the singularity control room and gas storage to mechlab, replacing transparent doors with opaque doors, and adding more blast shields over windows looking into the singularity room.

To accomodate this a few pipes and wires were rerouted. Pipes were moved with an emphasis on avoiding going through the singularity area if possible, as pipes are often the most convenient way to feed items to the singularity and existing pipes can get in the way.

## Why's this needed? 

A while back I made a similar PR to this, but it was not very responsibly constructed and touched a lot of other random things. This should stay tighter within scope of relevant changes.

Currently most of engineering will become a radiation hazard once the singularity is on. Mr. Rathens dies of radiation poisoning and stinks up the place, and half the fish in the CEs fishtank die too. The whole place is basically a social dead-zone since you have to wear a radiation suit in order to not die horribly. This should fix that. The main singularity concourse is still radioactive, but not the surrounding rooms like the lobby, lounge, and CE office.

## Changelog 

```changelog
(u)Cherman0
(+)Donut3 engineering redesigned with OSHA radiation protection standards in mind.
```


soleilan:
Yeah it's pretty bad when new engineers arrive and enter the break room through the south maintenance and still literally go into crit before they can take stuff out of the lockers to gear up.

TobleroneSwordfish (approved):
Looks good!

TealSeer (approved):

JORJ949:
image
I like the CE having a door to their tank but it shouldnt be behind a crate

PR #20176, opened 2024-08-03
Replaces the lawbringer's HIGH EXPLOSIVE mode with a HIGH POWER assault laser mode

TobleroneSwordfish:
[BALANCE]
## About the PR 
Also tweaks some things behind the scenes to make assault lasers slightly more consistent demolition tools and adds a cool flash effect and slight disorient on firing the assault mode.

## Why's this needed? 
Currently the high explosive mode isn't very practical for breaching doors/walls since it usually takes out the floor too and any nearby APCs, security assistants etc. Assault lasers do almost exactly the same damage, have similar breaching chances and do not blow holes in the floor.

## Changelog 

```changelog
(u)LeahTheTech
(*)Replaced the lawbringer's High Explosive mode with a High Power assault laser mode. Keywords are: BIG SHOT, HIGH POWER and ASSAULT.
```

### DIFF
--- a/code/obj/item/gun/energy.dm
+++ b/code/obj/item/gun/energy.dm
@@ -1300,7 +1300,16 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 
 	New(var/mob/M)
 		set_current_projectile(new/datum/projectile/energy_bolt/aoe)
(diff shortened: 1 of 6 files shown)


trazodont:
BIG LASER???

code/obj/item/gun/energy.dm
```diff
@@ -1375,7 +1384,7 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 			src.change_mode(M, text)
 		else		//if you're not the owner and try to change it, then fuck you
 			switch(text)
-				if ("detain","execute","knockout","hotshot","incendiary","bigshot","highexplosive","he","clownshot","clown", "pulse", "punch")
+				if ("detain","execute","knockout","hotshot","incendiary","assault","highpower","clownshot","clown", "pulse", "punch")
```
pgmzeta, line 1387: should "bigshot" be here still?

TobleroneSwordfish: Oh, good catch!

PR #20178, opened 2024-08-03
Replaces the lawbringer's EXTERMINATE mode with laser blaster projectiles

TobleroneSwordfish:
[BALANCE]
## About the PR 
The blaster projectiles do 2 less damage and don't embed/bleed so this is a slight nerf overall but frankly the current mode is pretty busted so I think it's fine.

## Why's this needed? 
Recharging kinetics are bad design wise, cause a lot of balance problems and make no sense!! Blaster bolts are much more in theme while still being dangerous and punchy feeling.

## Changelog 

```changelog
(u)LeahTheTech
(*)Replaced the lawbringer's EXTERMINATE mode with laser blaster projectiles. Keywords remain unchanged.
```


Retrino:
on one hand, this makes sense and the lack of blood certainly saves the time of washing every item of clothing after you inevitably slip in the blood of the person youre shooting at. on the other, as a purely visual standpoint, its a lot harder to tell if a laser shot has landed opposed to a kinetic bullet because of the blood. is adding some sort of effect on hit out of the scope of the pr?

TobleroneSwordfish:
Prooobably out of scope yep, although it might be doable to add some kind of dissipation effect to high power lasers? Maybe being less obvious is just an aspect of lasers anyway though.

Cherman0:
What happened to the queen of fuckery, I thought you were going to make a kinetic/energy hybrid switching mode?

Just kidding. Looks good to me. This blaster projectile does 33 damage, starts falling off after 6 tiles, and falls off 5 damage per tile after that. For reference the .38 lawbringer bullet does 35 damage, starts falling off after 5 tiles, and falls off for 5 damage per tile after that. So technically at longer than 5 tiles range this will actually do more damage, assuming you ignore the effects of bleed and lodged bullets. And the egun laser is substantially more powerful at 45 damage, starts falling off after 6 tiles, and falls off for 5 damage per tile after that.

PR #20179, opened 2024-08-03
Showing off auth disk

Valtsu0:
[Game objects][Feature]
## About the PR 
image
Adds showing off the authentication disk

## Changelog 

```changelog
(u)Valtsu0
(+)You can now also show off the authentication disk
```


TobleroneSwordfish (changes requested):
Oof okay these actions badly need a common parent, I know the original PR kind of copy-pasted them but if we're going to make it apply to more things it needs to start using inheritance.

Valtsu0:
Didn't end up using inheritance, but it's better now

TobleroneSwordfish (changes requested):
Oh that's better, but the pixel offsets are slightly different between the item types so you'll have to either make subtypes or pass the offsets in `New` (probably the latter?)

TobleroneSwordfish (approved):

code/datums/actions/actions.dm
```diff
@@ -1971,3 +1971,42 @@
...
+			hand_icon_state = "[hand_icon]_hold_r"
+			src.pixel_x_offset = -src.pixel_x_offset
+
+		var/image/overlay = src.item.SafeGetOverlayImage("showoff_overlay", src.item.icon, src.item.icon_state, MOB_LAYER + 0.1, src.pixel_x_offset, src.pixel_y_offset)
+		var/image/hand_overlay = src.item.SafeGetOverlayImage("showoff_hand_overlay", 'icons/effects/effects.dmi', hand_icon_state, MOB_LAYER + 0.11, src.pixel_x_offset, src.pixel_y_offset, color=user.get_fingertip_color())
+
+		src.user.UpdateOverlays(overlay, "showoff_overlay")
+		src.user.UpdateOverlays(hand_overlay, "showoff_hand_overlay")
+
+		src.user.dir = SOUTH
```
Azrun, line 2007: `set_dir()` or do so we need to bypass that for some reason?

PR #20186, opened 2024-08-04
Unify RP and Classic security access

JORJ949:
## About the PR 
Makes RP and Classic  security officer/HoS accesses the same, removing a large portion of RP's officer access in the process
I made a forum post to discuss this which can be found [here](https://forum.ss13.co/showthread.php?tid=23020)
I was going to wait for more feedback on this but Leah expressed interest in testmerging these changes
(Below spreadsheet may be inaccurate if testmerge is not updated)
image

**WIP ACCESS LIST**, if you think either job should/shouldn't get an access please mention it.

## Why's this needed? 
Please read the comment below by TobleroneSwordfish because they did a much better explanation than I did.

## Changelog 

```changelog
(u)JORJ949
(*)RP and Classic security officer and HoS accesses have been unified. See the PR for a more in depth comparison and report any issues that may arise to the forum thread.
```

### DIFF
--- a/code/procs/access.dm
+++ b/code/procs/access.dm
@@ -213,7 +213,7 @@
 						access_forensics_lockers, access_armory, access_tox, access_tox_storage, access_chemistry, access_medical, access_morgue,
 						access_change_ids, access_eva, access_heads, access_medical_lockers, access_medlab,
 						access_all_personal_lockers, access_tech_storage, access_maint_tunnels, access_bar, access_janitor,
-						access_crematorium, access_kitchen, access_robotics, access_cargo, access_money
+						access_crematorium, access_kitchen, access_robotics, access_cargo, access_money,
 						access_research, access_dwaine_superuser, access_hydro, access_ranch, access_mail, access_ai_upload,
 						access_engineering, access_teleporter, access_engineering_engine, access_engineering_control,
 						access_mining, access_pathology, access_researchfoyer, access_chapel_office, access_telesci,
--- a/code/procs/access.dm
+++ b/code/procs/access.dm
@@ -213,7 +213,7 @@
 						access_forensics_lockers, access_armory, access_tox, access_tox_storage, access_chemistry, access_medical, access_morgue,
 						access_change_ids, access_eva, access_heads, access_medical_lockers, access_medlab,
 						access_all_personal_lockers, access_tech_storage, access_maint_tunnels, access_bar, access_janitor,
-						access_crematorium, access_kitchen, access_robotics, access_cargo, access_money, access_change_ids
+						access_crematorium, access_kitchen, access_robotics, access_cargo, access_money,
 						access_research, access_dwaine_superuser, access_hydro, access_ranch, access_mail, access_ai_upload,
 						access_engineering, access_teleporter, access_engineering_engine, access_engineering_control,
 						access_mining, access_pathology, access_researchfoyer, access_chapel_office, access_telesci,
--- a/code/procs/access.dm
+++ b/code/procs/access.dm
@@ -209,21 +209,16 @@
 						access_research, access_hydro, access_ranch, access_mail, access_ai_upload, access_pathology, access_researchfoyer,
 						access_telesci, access_teleporter, access_money)
 		if("Head of Security")
-#ifdef RP_MODE
-			var/list/hos_access = get_all_accesses()
-			hos_access += access_maxsec
-			hos_access += access_armory
-			return hos_access
-#else
-			return list(access_security, access_carrypermit, access_contrabandpermit, access_maxsec, access_brig, access_securitylockers, access_forensics_lockers, access_armory,
-						access_tox, access_tox_storage, access_chemistry, access_medical, access_morgue, access_medlab,
-						access_change_ids, access_eva, access_heads, access_medical_lockers,
+			return list(access_security, access_carrypermit, access_contrabandpermit, access_maxsec, access_brig, access_securitylockers,
+						access_forensics_lockers, access_armory, access_tox, access_tox_storage, access_chemistry, access_medical, access_morgue,
+						access_change_ids, access_eva, access_heads, access_medical_lockers, access_medlab,
 						access_all_personal_lockers, access_tech_storage, access_maint_tunnels, access_bar, access_janitor,
-						access_crematorium, access_kitchen, access_robotics, access_cargo,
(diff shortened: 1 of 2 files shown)


Cherman0:
I broadly agree with these changes except that I do think the HoS should get ID computer perms. Controlling access seems under the purview of security. Though I suppose most of what they would be doing is downgrading ID cards and they can do that just fine from their own, so it doesnt matter that much.
And since you asked for people to mention if they had any thoughts on it I'm going to put E-Input Wanted on this.

JORJ949:
> I broadly agree with these changes except that I do think the HoS should get ID computer perms. Controlling access seems under the purview of security. Though I suppose most of what they would be doing is downgrading ID cards and they can do that just fine from their own, so it doesnt matter that much. And since you asked for people to mention if they had any thoughts on it I'm going to put E-Input Wanted on this.

There is already a discussion thread, visible in the PR description this has created a duplicate thread.

code/procs/access.dm
```diff
@@ -209,21 +209,15 @@
...
-#else
-			return list(access_security, access_carrypermit, access_contrabandpermit, access_maxsec, access_brig, access_securitylockers, access_forensics_lockers, access_armory,
-						access_tox, access_tox_storage, access_chemistry, access_medical, access_morgue, access_medlab,
-						access_change_ids, access_eva, access_heads, access_medical_lockers,
+			return list(access_security, access_carrypermit, access_contrabandpermit, access_maxsec, access_brig, access_securitylockers,
+						access_forensics_lockers, access_armory, access_tox, access_tox_storage, access_chemistry, access_medical, access_morgue,
+						access_change_ids, access_eva, access_heads, access_medical_lockers, access_medlab,
 						access_all_personal_lockers, access_tech_storage, access_maint_tunnels, access_bar, access_janitor,
-						access_crematorium, access_kitchen, access_robotics, access_cargo,
+						access_crematorium, access_kitchen, access_robotics, access_cargo, access_money,
```
TobleroneSwordfish, line 216:
```suggestion
						access_crematorium, access_kitchen, access_robotics, access_cargo, access_money, access_change_ids
```
Okay on second thoughts not every map has departmental ID consoles so leave ID access for now.

TobleroneSwordfish (changes requested):

TobleroneSwordfish:
Okay I'm going to put a comment here before I testmerge this summarizing why because the PR body doesn't do a very good job of justifying these changes and the forum thread has inevitably become 4 pages of people talking at cross purposes.

The intent here isn't just to unify access for the sake of parity, it's addressing what many see as longstanding issues with the level of access security gets on RP.
Firstly, the current level of access was designed for a time when 15 people was highpop and the only map was Clarion. Most people playing the game right now are playing on mid to high population RP server(s) and security's access on those servers causes at least three major issues I can define:

- Once an antagonist is running, chases tend to go on forever or until they are caught. This doesn't happen on classic and feels much less frustrating for all involved. If there's a decent chance to get away then antagonists are much more likely to stand and talk in the meantime.
- Antagonists are incentivised to stay out of the station altogether because nowhere is safe. I heard someone in the thread mention this as a justification - "they can just go hide in space or off station" but is that healthy for the game when the people who are meant to be driving the round are instead incentivised to hide in space and interact only with security?
- Security officers become do-everything roles who can go anywhere and fix any problem, disincentivising other roles from fixing those problems. Yes, "stay in your lane" is supposed to prevent this but it's hard to blame security for wanting to fight fires, set up the engine and do whatever else needs doing *when they have the access to do all this*.

This PR is unlikely to get fully merged as-is, but since this topic keeps coming up and being argued about we should at least try it out and see what needs to be changed to make it work. Tying security access to population thresholds or alert levels are definitely possible solutions to the issues of low-pop and crises requiring the extra access if that appears to be necessary.

Please don't use this as an opportunity to start arguing here as well, put your thoughts in the forum thread and they will be read.

Tarmunora (changes requested):
the sec [departmental IDcomp](https://github.com/goonstation/goonstation/blob/780df14eb1d5537d76ba158f6a235085edf328b4/code/obj/machinery/computer/card.dm#L638) will need it's RP/classic split changed as well

Tarmunora:
And as an additional hot take, I'd remove full IDcomp access too. The security IDComp that HoS has access to can handle deputizing and demoting officers perfectly well.

Cherman0:
Its hidden because the convo was marked resolved but Leah mentioned that not all maps actually have those and so it makes sense to keep ID comp for now at least.

Sord213:
i'm pretty sure the only map without departmental ID comps is atlas. clearly the solution here is to give all heads ID comp access on atlas nothing could possibly go wrong

Cherman0:
In fairness there is often just an Mdir or something on Atlas so that may make actual sense.

Tarmunora:
> Its hidden because the convo was marked resolved but Leah mentioned that not all maps actually have those and so it makes sense to keep ID comp for now at least.

this is what I get for not reading smh

JORJ949:
Added mining because they often are too busy with their rocks to pay attention to the radio or pda to talk to you
Added toxins to deal with known and active bomb threats (please dont come "just to check" or ill be sad)

JORJ949:
Re-removed toxins because you dont want sec wandering in while you make bombs and bomb threats often happen elsewhere (they also have to come out to do the bombing)

Cherman0:
Not a huge fan of that justification for adding mining. RP servers nowadays (especially 3) aren't usually in a state of having nobody who would care to respond to you anymore. Adding mining is of some note because the mining outpost uses it and because it lets security easily commandeer mining podbay if the security podbay is disabled.

JORJ949:
> Not a huge fan of that justification for adding mining. RP servers nowadays (especially 3) aren't usually in a state of having nobody who would care to respond to you anymore. Adding mining is of some note because the mining outpost uses it and because it lets security easily commandeer mining podbay if the security podbay is disabled.

The outpost has a separate access I did not add, however this is also something that happens on classic because outside of explosions happening where mining confirms if it was or wasnt them they tend to be too focused on their rocks to let you in.

TobleroneSwordfish:
Okay, I'm going to merge this now. The testmerge has been generally positive I think, with the only lingering concern being low-population shifts and so far all of the proposed solutions to that seem like they'll cause more problems than just letting security ask the HoP/captain for more access when they need it.

PR #20197, opened 2024-08-05
Fix unremovable shoes turning into rocket boots with an air tank

### LINKED ISSUE #20196: Cursed Cluwne shoes can be removed an air tank
### Labels

[BUG][MINOR]

### Description

You can remove cluwne shoes you put on willingly by turning them into rocket boots with an airtank

### Steps to reproduce

Put on Cluwne shoes willingly.
Get an air tank and combine it with the equipped cluwne shoes.
The cluwne shoes disapear and become Rocket shoes. 
You are now shoe free

### Expected Behavior

The cluwne shoes not being disapearing and becoming rocket shoes, as they are cursed and should not be taken off so easily

Reported by: cyferr
Client version: 515.1642
Active test merges: nothing
Round log date: 2024-08-05-17-10
Reported on: 2024-08-05 18:06:43
Round ID: 185101
Map: COGMAP

pgmzeta:
[bug][game objects]
## About the PR 
Check to see if the cursed clown shoes are being worn and return early before attempting to use an airtank on them.

## Why's this needed? 
Fixes #20196

### DIFF
--- a/code/obj/item/clothing/gimmick.dm
+++ b/code/obj/item/clothing/gimmick.dm
@@ -460,6 +460,14 @@ TYPEINFO(/obj/item/clothing/under/gimmick/fake_waldo)
 	step_lots = 1
 	step_priority = 999
 
+	attackby(obj/item/W, mob/user)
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish (changes requested):
A better solution would just be to check if the shoes aren't removable in the shoes assembly code rather than duplicating a bunch of checks down here.

Tarmunora (approved):

PR #20199, opened 2024-08-05
Space law rewrite

TobleroneSwordfish:
[LAW] [GAME OBJECTS]
## About the PR 
image

## Why's this needed? 
The old space law book is hilariously outdated, encourages you to do things that actually break the game rules in a couple of places and is generally just bloated and useless. This version scraps the long list of crimes and exact punishments and ends up more like a handbook of how security are expected to behave and enforce the law.
Also I added colour coded highlights so even the most helmet-headed security grunt can get through it.

## Changelog 

```changelog
(u)LeahTheTech + Jan.antilles
(*)Nanotrasen Space Law has been completely rewritten and updated to actually be something you should show sec assistants.
```

### DIFF
--- a/strings/books/space_law.txt
+++ b/strings/books/space_law.txt
@@ -19,6 +19,6 @@ Keep in touch with your fellow security members over the radio and support one a
 
 <h3>Punishments</h3>
 The variety of creative and awful ways crewmembers will find to break the law is infinite, so the variety of punishments and sentences should be too.<br>
-Standard penalties range from tickets, fines and brig sentences to exile and even execution for more severe crimes.<br>
+Standard penalties range from tickets, fines and brig sentences to exile and even execution for the most severe crimes.<br>
 Brig sentences should not exceed a maximum of 5 minutes jail time and all executions of captured prisoners <b>must</b> be signed off by a head of staff, preferably the <b><span style="color:#a87101;">Captain</span></b>.<br>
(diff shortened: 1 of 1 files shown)


Valtsu0:
I don't think it should be called space law if it is a handbook on how a security officer should act

soleilan:
Having a list of various crimes was nice especially for newer players or people trying to figure out what to write into records.

Glamurio:
> Having a list of various crimes was nice especially for newer players or people trying to figure out what to write into records.

I agree. Punishment is very nuanced, and there's a couple of crimes that shouldn't be punished as hard as you would think normally. A few key notes like "murder that destroys the body without allowing clones or cyborgification are significantly worse", and I think it should have a segment on interrogations, which can be used as more interactive "brig time".

TobleroneSwordfish:
> I don't think it should be called space law if it is a handbook

I'm kinda torn on this because the Space Law is *iconic* and this is still broadly law, just more in the other direction I guess?

> Having a list of various crimes was nice especially for newer players

I dunno, I was always told to ignore it completely and I can't think of a single instance where anyone has referred to it in-game so 🤷. Maybe as a separate document or an appendix or something?

TobleroneSwordfish:
I'm going to just merge this as-is for now, we can maybe add back a revised version of the big list later but right now I think succinctness and readability are more important.

PR #20200, opened 2024-08-06
Allow storing items inside of gas tanks

MyNameIsRomayne:
[Feature] [Game Objects]
## About the PR 
This PR allows welding a hole into gas tanks to place items inside! Doing so reduces the available volume of the tank depending on the weight class of the item, and may have special effects if these items are inside when the tank explodes.

- Tanks may have holes welded open into them. This allows placing in and removing items like a trash bag, using grab intent.
- Tanks which are welded open will partially exchange gas contents with the ambient environment.
- When a tank explodes, any and all items inside are released and flung in random directions.

Things left to do:
- [ ] Replace developer art overlays with proper sprites
- [ ] Consider any unintended balance implications of putting items in tanks
- [ ] Finish and test implementation for the handheld/pocket oxygen tanks
- [ ] (Potentially atomized) add explosion mods for RCD cartridges, char
- [ ] Edit changelog entry before undrafting PR

## Why's this needed? 
Opens the door for custom modifications to tank bombs / TTVs. I think this mechanic would be interesting to see used as it would provide more 'flavor' to the explosives beyond the boring old explosion. Basically, same intent as #20129 with the same process as #20128. Maybe both, if this is worth having as one PR.

Also allows using tanks as storage at the cost of their precious breathable air space. The welding job to close it back up will be tough, but hey, surely security won't bother to grab a welder to check? (This just incidentally comes with the PR)

## Changelog 

```changelog
(u)Romayne
(*)You can now place items inside of gas tanks.
```


LaCumbiaDelCoronavirus:
Will this be compatible with the bomb sim?

MyNameIsRomayne:
Doable, it would just need special handling to copy over items in a way that is safe.

MyNameIsRomayne:
Still hoping for feedback on this, I see a :confused: react and it frightens me

Glamurio:
If there's no visible way to tell that cans have items hidden inside, I'm worried about a meta of having to weld check possibly multiple cans per shift just in case.

MyNameIsRomayne:
> If there's no visible way to tell that cans have items hidden inside, I'm worried about a meta of having to weld check possibly multiple cans per shift just in case.
There is a visible overlay when the can is opened, and there will be one on cans which are closed, but have been opened. There is a matching description.

Tarmunora:
my thoughts re the "modifiable TTVs" concept in general
- that's already pipebombs' thing
- the one cool thing about TTVs is that they're strongly tied to atmos, and that them working as bombs at all is more or less emergent behavior, and making it also depend on hardcoded mcguffins you shoved in the ends are would take away from that somewhat imo
- and diagetically, cutting a hole into and subsequently resealing a pressure vessel by yoloing it with a handheld welding torch sounds like a recipe for a ruptured seam before you'd even get it halfway repressurized

MyNameIsRomayne:
> my thoughts re the "modifiable TTVs" concept in general
> 
>     * that's already pipebombs' thing
> 
>     * the one cool thing about TTVs is that they're strongly tied to atmos, and that them working as bombs at all is more or less emergent behavior, and making it also depend on hardcoded mcguffins you shoved in the ends are would take away from that somewhat imo
> 
>     * and diagetically, cutting a hole into and subsequently resealing a pressure vessel by yoloing it with a handheld welding torch sounds like a recipe for a ruptured seam before you'd even get it halfway repressurized

Okay, I see those and those are fair points. The main reason I went for items is that creating a bunch of atmospheric gases specifically to get these kinds of ranges of behaviors seems difficult to justify, considering there could be performance implications behind it. I'm open to creating more gases with interesting interactions in the game-word if that's not the case, but it is the impression I got from radgas/fallout.

Last point I get, I'm leaning hard into 'game-logic' here to justify such a thing. I think it's a bit odd to draw a line here, but I have to admit first that:
- I have never, ever welded in my life
- I am not well educated on how pressured containers like this work or maintain their seal
And so without that knowledge, this kind of makes sense to me. Similar logic is found all over the game, like from welding together busted pipes and maintaining their pressured seal immediately (and doing it *while* there's so much pressure) to immediately welding/unwelding doors shut in seconds. It just sort of comes with the way the game handles welding.

As for it being a pipebomb's thing, I think it's doable to add cool effects to an explosion outside of pipebombs? I understand it might be a poor idea to take directly from the effects pipebombs have, though I don't see why only pipebombs get this kind of thing. It would be similar no matter how it's implemented though, I suppose.

Short of it is I don't expect this to be realistic, I just want there to be more fun things to do here with tank explosions. If that should be done with new/existing gases, I'd love too -- I just might need to spend a bit more time to understand exactly how to go about doing that the right way.

code/obj/item/tank.dm
```diff
@@ -140,6 +203,26 @@ Contains:
...
+						// Remove *some* of the gas from the tank at a time to 'leak' out
+						temp = src.air_contents.remove_ratio(0.25)
+						// Add the gas removed from the tank *all* into the turf
+						T.air.merge(temp)
+						// Determine what percentage of the turf's air contents should go back into the tank
+						var/transfer_ratio = src.air_contents.volume / (src.air_contents.volume + T.air.volume)
+						// Remove aformentioned amount from the turf, put it back into the temp
+						temp = T.air.remove_ratio(transfer_ratio)
+						// Lastly, put all that air removed from the combined mixture back into the tank.
+						src.air_contents.merge(temp)
```
BatElite, line 224: Does this work? I think you'll need to involve a pressure delta somewhere or the tank won't equalise to ambient pressure.

Tried calculating to check but IDK what the default contents of a tile is in moles.

MyNameIsRomayne: ah yeah it borks up a bit. gimme a minute

MyNameIsRomayne: image
Probably going to do something like this instead

MyNameIsRomayne:
I still want to do this, I'll reopen and request stuff when I one day have time

PR #20202, opened 2024-08-06
[DRAFT] Add explosion modifier / effect datums for tank bombs.

MyNameIsRomayne:
## About the PR 
Adds a system for creating modification and effects for explosions.

Creating a modifier is done to transform the presence of some items/conditions (temp, pressure, gas, item) into a tangible effect on the explosion, or several.

As an example, the effects for having RCD ammo in a tank are implemented and functional:

https://github.com/user-attachments/assets/13baab50-24d5-4d49-b2e5-3076a6e0353d

# Requires #20200

TODO:

- [ ] Remove debug / testing code
- [ ] Wait for #20200 which allows this to be actually possible
- [ ] Add more modifiers(?)
- [ ] Further test/debug code, consider balance implications of any modifiers added

## Why's this needed? 
Allows for interesting and fun interactions with TTVs, allowing player creativity and choice to be expressed where otherwise there has just been a large explosion.

Opens the gates for these datums to be used with pipebomb assembly, reducing the spaghetti in that code for its explosion modifier handling.


MyNameIsRomayne:
I'm going to close this to reduce clutter while I wait for #20200. Not working on this until that is merged anyways. Will reopen if/when that happens

PR #20209, opened 2024-08-07
Have the Port-a-Brig make noises

TheTyrant25:
[SOUNDS] [FEATURE]
## About the PR
Have the Port-a-Brig make the same sound effects as the med sleeper when an occupant is put inside or removed. Also clunks like an airlock whenever it's locked or unlocked.

## Why's this needed?
It was a bit strange that it was dead silent.

## Changelog
```changelog
(u)Tyrant
(+)Due to budget cuts, Port-a-Brigs are no longer equipped with soundproofing.
```

### DIFF
--- a/code/obj/machinery/porters.dm
+++ b/code/obj/machinery/porters.dm
@@ -533,7 +533,7 @@ TYPEINFO(/obj/machinery/port_a_brig)
 		for(var/obj/O in src.brig)
 			O.set_loc(src.brig.loc)
 		src.brig.build_icon()
-		playsound(src.loc, 'sound/machines/sleeper_close.ogg', 50, 1)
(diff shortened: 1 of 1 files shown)


DisturbHerb:
i think the pickle jar deserves something a little more substantial, the vibes of the sleeper sound are kinda off? otherwise important change

TheTyrant25:
I'm open to suggestions on sfx as long as I don't have to record them meself

Glamurio:
If it's not too much trouble, I'd also like something akin to a bolt / unbolt sound when locking or toggling the lock on it.

frawhst (approved):
Ideally this would have bespoke sounds but these are fine imo.

PR #20212, opened 2024-08-07
Remove Katana and Wrestling Belt from surplus crates

JORJ949:
## About the PR 
Removes both the katana and wrestling belt from surplus crates

## Why's this needed? 
The wiki states "some items cannot appear in them, primarily those geared towards murder" and these 2 items in particular are if anything stronger at rampaging than a csaber, let alone them appearing with other items like stimulants.

## Changelog 

```changelog
(u)JORJ949
(+)Removed Wrestling Belt and Katana from Surplus crates.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -1064,7 +1064,7 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 	name = "Black Market Megaphone"
 	desc = "An illegal megaphone with the limiter taken off, and a loudener added. Not for the subtle."
 	item = /obj/item/megaphone/syndicate
-	cost = 5
+	cost = 3
 	vr_allowed = FALSE // no
 	job = list("Captain", "VIP", "Inspector", "Head of Personnel")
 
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
I agree with taking out the katana and wrestling belt but I think the megaphone is going to be a pretty big disappointment for 5tc worth of your crate most of the time.

klookiyam:
The blowgun looks like an interesting weapon, maybe you can get one of the variations of those?

frawhst:
Not sure about this— as far as I understand the removal of certain rampage items from the surplus crates is just to reduce the odds of getting some crazy weapon setup from the crate. We don't necessarily want to remove all weapons from the crate, but we do want to remove *some* of them so that the odds of getting rampage items are skewed a bit lower.

JORJ949:
Removed megaphone changes, those are not what this PR is for

TobleroneSwordfish (approved):
Thinking about it, if the point of taking c-sabers and revolvers out is to prevent busted combos with other items then it should include the other "this is 7TC because otherwise you could also buy stims/cloaker" items like the katana and wrestling belt.

I like the idea of the surplus crate being a box of random gimmick supplies rather than "you rolled a wrestling belt, so now you have to go LOUD and do WRESTLER THINGS"

PR #20214, opened 2024-08-07
fixes a secret medal

Valtsu0:
[bug][medal][secret]
## About the PR 
Makes the secret medal "Law 1: Don't be an asshat" be able to be earned

the medal was last earned 5 months ago
## Why's this needed? 
The current requirements for completing that medal is to attack the ai with an item that:
1: is not /obj/item/clothing/head/ (hattable component overrides the attackby function)
2: is /obj/item/clothing/mask/moustache/ (broken indentation)
3: is /obj/item/clothing/head/butt
4: has you as a donor
This is obviously nonsensical


Sbmhawk:
The medal was last earned 5 months ago because... it's a rare medal, only 81 people have it.

Nothing in the code you are changing mentions points 1 and 2 that you mention, where are you getting that from.

Valtsu0:
for 1, the hattable component overrides the entire function it is in. 2 is caused by broken indentation.

I tried to earn the medal today

ZeWaka (approved):
sure ty

PR #20215, opened 2024-08-07
Fix up manufacturer buttons a bit

MyNameIsRomayne:
## About the PR 
Turn this
firefox_1aSrLG1URP

Into this
IjssgmB1gc

I dont know why the inline-flex property matters but it fixed it so i suppose it should be good


MyNameIsRomayne:
Huh I didn't know this shows up on master. Just for clarity, this is for the react branch to help stuff blocking the conclusion of the freeze

ZeWaka (approved):
thank you!

PR #20217, opened 2024-08-07
Fix Telesci stepped slider and SMES slider

pgmzeta:
[ui][bug]
## About the PR 
Remove some `inline` declarations.
Also add a vertical-align middle for the SMES so the text lines up better

## Why's this needed? 
Fix issue blocking react code freeze 

Screenshot 2024-08-07 104431
Screenshot 2024-08-07 104625


ZeWaka (approved):
thank you 💝

PR #20227, opened 2024-08-08
Add new HoS reward, Voice of the Law

JORJ949:
## About the PR 

Adds a new egun job redeem, Voice of the Law
When emagged forces you to say "I AM THE LAW!" when attempting to use it
Saying "I am the law" into it while not being the HoS is a bad idea.

https://github.com/user-attachments/assets/4e0efe8f-380a-4dba-b3ac-b0f615d293ed

https://github.com/user-attachments/assets/277f75eb-c067-4a2d-b93b-2a17ed0b9ffc

## Why's this needed? 
Funny alternative to the lawbringer, they get a taser anyway they dont need 2 guns.

Im happy to change obtainment method from a job reward as I know those suck

## Changelog 

```changelog
(u)JORJ949
(*)New E-Gun alternative for the HoS, Voice of the Law. CAUTION: Not suitable for crew that are not the law.
```


itsloley:
pretty funny, but it looks like by scaling up the captain's megaphone text to that size the text would be all blurry and messy, maybe they should stay the same size if there's not a cleaner number to scale to?

JORJ949:
> pretty funny, but it looks like by scaling up the captain's megaphone text to that size the text would be all blurry and messy, maybe they should stay the same size if there's not a cleaner number to scale to?

The current size already does this a bit, its just more visible on captain green than HoS red

JORJ949:
After looking at it more, yeah it looks weird at 18px, changed captain back to 12 and HoS emagged version to the 24

Retrino:
seems unnecessarily volatile for such a simple looking object. it'd make sense to continue the exploding theme for Funny Line, but just outright for anything said kind of overdoes it for a non emagged/non traitor item

JORJ949:
> seems unnecessarily volatile for such a simple looking object. it'd make sense to continue the exploding theme for Funny Line, but just outright for anything said kind of overdoes it for a non emagged/non traitor item

Has been changed since originally PRed, Only forces you to say the line when emagged.

klushy225:
I think I need this... It looks great. The HoS definitely needs a megaphone

Flaborized:
Having made the captain's megaphone, I'm not much a fan of giving other heads them. It's meant to be a unique little toy for the captain, giving one to the HoS also makes it much less special. I'd rather each head get their own unique little doodads personally. I think the idea of being able to trade away your gun for a silly item is fine though

JORJ949:
> I'm not much a fan of giving other heads them. It's meant to be a unique little toy for the captain, giving one to the HoS also makes it much less special. 

I also disagree with giving it to any other head, but I dont see the captain use their megaphone nearly enough and feel it uniquely fits the HoS in that they can bellow "COME OUT WE HAVE YOU SURROUNDED"

TobleroneSwordfish:
Yeah I agree with Flaborized, I think keeping the megaphone as a captain exclusive thing makes more sense. Cool idea and sprite but just doesn't quite fit.

PR #20239, opened 2024-08-09
Allow usage of a local file in "code/WorkInProgress/__development.local.dm"

MyNameIsRomayne:
## About the PR 
Adds support for an untracked file, __development.local.dm, for code changes to take place in.

Due to the order in which __build.local.dm is included in goonstation.dme, you cannot make local code changes inside that file. Thus, another file included after all the define flags and _std files is necessary to support such a thing without being caught under the watchful eyes of git version control.

There is also support for a new define `DISABLE_DEVFILE`, which will prevent including this file. Good for testing if your local tools are breaking things.

## Why's this needed? 
Making changes to the codebase is made much simpler by creating personal tools and widgets to speed along development. However, such tools are often unfit for the main code repository and thus become a burden to maintain/tug around with version control. Having a standard location, much like __build.local.dm, where code changes can be made and kept should be a good QoL for development going foward.

Feel free to make your own judgements on this, this may also create issues where people have errors resulting from their own tooling and then forget it's there. I doubt it'd happen, but who knows!

### DIFF
--- a/goonstation.dme
+++ b/goonstation.dme
@@ -33,8 +33,8 @@ var/datum/preMapLoad/preMapLoad = new
 #include "code\__secret_public.dme"
 
 // Local development file handling
-#if fexists("code/WorkInProgress/__local.development.dm") && !defined (DISABLE_DEVFILE)
-#include "code\WorkInProgress\__local.development.dm"
+#if fexists("code/WorkInProgress/__development.local.dm") && !defined (DISABLE_DEVFILE)
+#include "code\WorkInProgress\__development.local.dm"
 #endif
 
 // BEGIN_INCLUDE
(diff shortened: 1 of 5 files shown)


goonstation.dme
```diff
@@ -32,6 +32,11 @@ var/datum/preMapLoad/preMapLoad = new
 #endif
 #include "code\__secret_public.dme"
 
+// Local development file handling
+#if fexists("code/WorkInProgress/__development.local.dm") && !defined (DISABLE_DEVFILE)
```
ZeWaka, line 36:
```suggestion
#if fexists("code/WorkInProgress/__development.local.dm") && !defined(DISABLE_DEVFILE)
```

MyNameIsRomayne:
image

PR #20245, opened 2024-08-10
Make containers heat up their contents when on fire

LaCumbiaDelCoronavirus:
## About the PR 
(specifically containers that can be held, like backpacks, satchels, boxes. not stuff like lockers, crates, carts)
Call temperature_expose on everything in a container when they start burning and every tick they continue burning

Also fixes them dropping all their items when they stop burning by being extinguished, now only when they actually fully burn

## Why's this needed?
It makes sense for things inside a container to actually be affected when the container is on fire

## Changelog
```changelog
(u)DecodeQwertyPls
(+)Things inside a container (satchel, backpack, not lockers or crates) will heat up if the container starts burning
```


ZeWaka (approved):
we can add thermal insulation factors later

LaCumbiaDelCoronavirus:
fyi for anyone interested: an o2 box inside the backpack wont burn if the backpack is lit on fire

PR #20248, opened 2024-08-10
Minor Nadir Security changes

JORJ949:
## About the PR 
Make a few minor changes to nadir security:
- Replace brig hallway area with "security wing" so its teleblocked and hit by radstorms
- Give HoS a command dive suit rather than security
- Move static devices off central security table
- Move second weapon vendor to escape checkpoint
image
- Move second prisoner outfit locker to be with the other one
- Swap segway and flasher spawns as flashers make more sense in equipment
- Remove brig blinds (Nobody used or knew these existed)
- Add an automatic locker for genpop
- Replace stupid unpickable book with the actual space law first print item
- Expand armory to add a bar
- Replace armory NARCS with a turret for bar space
- Add "front desk" to armory
image

## Why's this needed? 
Various QoL things I didnt like/didnt make sense.
Turets make much more sense in armories than NARCS because turrets arent completely nullified by a 1s action bar if you have a screwdriver (which you will considering you broke into armory) and the HoS can disable them if granting crew armory access

## Changelog 

```changelog
(u)JORJ949
(+)Nadir security has received a few minor changes. HoSes should take particular note of the armory.
```


Chatauscours:
One minor thing, in general in armouries the phasers and rifles racks are separated from eachother so it can't be both stolen with a single wall getting RCD'd

TobleroneSwordfish:
Oh but the Nadir armoury wall cavity is souuul! I like the expansion but maybe you could eat some of the pod bay/equipment room instead?

TobleroneSwordfish (approved):

PR #20250, opened 2024-08-10
Remove the Still's item-type limitations

Egregorious:
## About the PR 

Changed the initial checks the Still performs during 'attackby' and 'MouseDrop_T', to check whether 'brew_result' is truthy rather than do typechecks of the item. This means any type of item can potentially be brewable as long as it has a valid 'brew_result', rather than it being limited to plant or food types.

This should not change what items are currently brewable - unless any non-food, non-plant item has valid brew results for some reason - only that any item given a valid 'brew_result' will now work with the Still. 

Also added a feedback message for when attempting to brew an unbrewable item which has a SUPPRESSATTACK flag.

## Why's this needed? 

It seemed like an arbitary limitation; 'brew_result' belongs to the base 'item' class anyway so I don't know what reason there could be to hardcode the Still to only accept plant/food types. This makes it more ripe for secrets too, maybe hide FOOF in random secret junk, or make viscerite distillable into a nasty drink. 

The message is just because it was weird that you'd get no feedback when attempting to prod the Still with an onion.


Glamurio:
Is there a reason why this was closed? This seemed like a good refactor.

PR #20253, opened 2024-08-11
Adds mechcomp link visualization to meson goggles

TobleroneSwordfish:
[FEATURE] [QOL]
## About the PR 
Adds little green animated lines between linked mechcomp components when wearing meson goggles.
image
You cannot see it in this colour photograph but the little lines are moving in the direction of the connection.
Also refactors meson code to not be copy-paste.

## Why's this needed? 
Seeing where all the connections are going makes mechcomp devices easier to understand and reason about aaand it's a lot easier to figure out why a door keeps opening and closing randomly because someone attached an automatic signaller to it or something.

## Changelog 

```changelog
(u)LeahTheTech
(*)Wearing meson goggles now allows you to see mechcomp connections.
```


BinneBoi:
Does it work inside lockers and device frames?

TobleroneSwordfish:
Not currently, I had a look at making it work on HUDs but the way Byond's UI placement works makes it very difficult to implement. Also it might just end up being a cluttered mess with how many connections there are inside many component cabinets.

PR #20254, opened 2024-08-11
New breaching sledgehammer special

TDHooligan:
[FEATURE] [OBJECTS]
## About the PR 

Adds a new special attack, heavy swing.

The attack makes the user take a step in a direction:
- If the user steps into a player/mob/wall, the attack is cancelled, and the mob they collided with is shoved. (This prevents this attack being spammed in crowded areas)
- If the user is uninterrupted, they promptly follow up with a 5-tile U shaped swing. 

https://github.com/user-attachments/assets/8977cbda-4a78-4dce-8a81-e4881960edec

## Why's this needed? 

Hammer swing & special attack are functionally identical due to having insane click delay anyway.
This adds some nice flavor and niche purpose for a really cool weapon concept that has very limited use currently. The idea being that this grants a riskier, but more reliable means of landing heavy hammer attacks.

Could also be used elsewhere and cut up for parts of other special attacks, like the shove being added to a heavy object special or something.

I'm not against nerfing the numbers on this or adding diminishing returns, but ATM I don't think it matters too much, given the rarity and risk of using this weapon.

I also asked a few times where to place the code for the new special, but didn't get replies. Happy to move it elsewhere :shrug:

## Changelog 

```changelog
(u)TDHooligan
(+)The airlock breaching sledgehammer now has a unique special attack.
```


AXLFAN2000:
This is REALLY fun, I like it!

TDHooligan:
added pace effect and upped shove length. increased cooldown a little because now it can wallstun more easily

TobleroneSwordfish (approved):
Very cool! Maybe the shove should deal a little bit of damage too though?

PR #20255, opened 2024-08-11
Makes omnitools mechscannable

Munien7:
## About the PR 

Makes normal and syndicate omnitools mechscannable (NOT silicon ones). Current material requirements are a rough guess and I'm fine to adjust these. Similarly, happy to make this a syndicate scanner only thing.

## Why's this needed? 

Someone brought it up on the discord so I figured I'd make the pr. Could be useful for salvagers or traitors to make more omnitools in the case that they lose them or get them confiscated.

## Changelog 

```changelog
(u)Munien
(+)Omnitools are now mechscannable.
```

### DIFF
--- a/code/obj/item/tool/omnitool.dm
+++ b/code/obj/item/tool/omnitool.dm
@@ -1,8 +1,8 @@
 TYPEINFO(/obj/item/tool/omnitool)
-	mats = list("metal_dense" = 80,
-				"conductive_high" = 50,
-				"energy_property_high" = 50,
-				"crystal" = 30,)
+	mats = list("dense_super" = 20,
+				"conductive_high" = 20,
+				"insulated_property" = 15,
+				"miracle" = 10,)
 
 /obj/item/tool/omnitool
 	name = "omnitool"
@@ -228,10 +228,10 @@ TYPEINFO(/obj/item/tool/omnitool)
 			return null
 
 TYPEINFO (/obj/item/tool/omnitool/syndicate)
-	mats = list("metal_dense" = 80,
-				"conductive_high" = 50,
-				"energy_property_high" = 50,
-				"crystal" = 30,)
+	mats = list("dense_property_ultra" = 20,
+				"conductive_high" = 20,
+				"insulated_property" = 15,
+				"miracle" = 20,)
 
 /obj/item/tool/omnitool/syndicate
 	icon_state = "syndicate-omnitool-prying"
--- a/code/obj/item/tool/omnitool.dm
+++ b/code/obj/item/tool/omnitool.dm
@@ -1,3 +1,9 @@
+TYPEINFO(/obj/item/tool/omnitool)
+	mats = list("metal_dense" = 80,
+				"conductive_high" = 50,
+				"energy_property_high" = 50,
+				"crystal" = 30,)
+
 /obj/item/tool/omnitool
 	name = "omnitool"
 	desc = "Multiple tools in one, like an old-fashioned Swiss army knife. Truly, we are living in the future."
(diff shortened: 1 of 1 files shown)


JORJ949:
Should probably be syndicate analyzer exclusive, at least for the syndicate one.

Munien7:
> If the material requirements are fancy enough, i don't think it needs to be syndie mechscan-able. The syndie multitool is not that great of a weapon and thus i would say, like some other syndicate items, it shouldn't need the syndie mechscanner (which is a 4TC buy).
> 
> But these material requirements are quite excessive. That's 21 full bars of materials. That's almost as much as the singularity generator. And some of these, like 5 highly radiosctive sources, are quite rare.

this was my line of thinking regarding the syndicate scanner, and why I didn't just PR it with that from the start. The materials ARE quite hefty, though that's just a result of me not having a good comparison rather than my idea of how expensive it should be. Maybe I'll lower each requirement by a couple of bars, though I do think it should be relatively expensive still

Munien7:
> > this was my line of thinking regarding the syndicate scanner, and why I didn't just PR it with that from the start. The materials ARE quite hefty, though that's just a result of me not having a good comparison rather than my idea of how expensive it should be. Maybe I'll lower each requirement by a couple of bars, though I do think it should be relatively expensive still
> 
> If it helps, i can write out a few notable mechscanable items so we can have some pointers to talk about:
> 
>     * salvager hand-tele(non-syndie): 0.5 metal, 0.5 energy source, 0.5 high energy conductor, 3 Telecrystals
> 
>     * Phaser (non-syndie): 3.2 random materials
> 
>     * custom chemical grenade (syndie): 0.4 dense metal, 0.2 energy source
> 
>     * stun baton (non-syndie): 1 very high dense metal, 1 high energy conductor
> 
>     * macrophaser (non-syndie): 1.5 metal, 1 dense metal, 1 high energy conductor, 1.5 high energy source, 1 crystal
> 
>     * naval mine (non-syndie): 1.6 random materials
> 
>     * stinger grenade (non-syndie): 0.6 random materials
> 
>     * injector belt (non-syndie): 1 random materials
> 
>     * gunbot (syndie): depends, but 1.2 dense metal, 1.2 high conductive material, 0.6 dense material for the baseline
> 
> 
> Overall, i personally would say it should require ~4 bars of material for the normal omnitool and 5 for the syndicate one One thing i would add, however, is a rare material to balance it out. If it would be a syndicate-mechscan item, i would go with 2 random materials and not bother with recources at all.
> 
> Maybe 1 bar of miracillium for the non-syndicate omnitool and 2 for the syndicate omnitool? That material is also used to balance out the chem dispenser mechscan and it would fit to the shifting nature of the omnitool. I see around 3-5 bars of miracillium on a common round where mining is existing and playing chill.

I appreciate the feedback, I'll work on implementing some of this (really like the idea of the miracillium, I'll have a play around with that) when I have a chance.

JORJ949:
Remember that the syndicate omnitool is something that can never be confiscated effectively as it can be stored in clothing unlike the regular omnitool

Munien7:
Updated the material requirements to be closer in line with other commonly scanned items, without being too punishing or too easy I feel. More feedback would be appreciated as I'm not a balance guy.

TobleroneSwordfish:
Are there any syndie items that are scannable with the normal scanner?

code/obj/item/tool/omnitool.dm
```diff
@@ -1,3 +1,9 @@
+TYPEINFO(/obj/item/tool/omnitool)
+	mats = list("dense_super" = 20,
+				"conductive_high" = 20,
+				"insulated_property" = 15,
+				"miracle" = 10,)
```
TobleroneSwordfish, line 5: Miracle is usually for things that produce matter out of nothing (chem dispensers etc.), seems excessive here.

TobleroneSwordfish (changes requested):

PR #20259, opened 2024-08-11
NT Winter Coat

JORJ949:
## About the PR 
Adds an NT style winter coat to NTSCs 
image

## Why's this needed? 
Im a big winter coat fan and none of the existing coats fit the NTSC style

## Changelog 

```changelog
(u)JORJ949
(+)NT has equipped their security consultants with a new winter coat.
```


DisturbHerb:
I've got a few reservations about this one. The coat, at present, doesn't look great with the NTSC palette (although more thorough sprite critique should follow, this is just after a quick glance). Moreover, this'd be the only job to spawn with their winter coat variant and none of the other types can currently be purchased outside of each department's uniform vendor.

Munien7:
While I personally like the idea, I'm concerned about the addition of more things to the, frankly (in my opinion), bloated amount of gear the NTSC spawns with.

Glamurio:
Honestly, I really like this but I also don't want them to spawn with it. Maybe we could add a command vendor into the bridge which features some command gear, such as the coats and winter jackets, and this could be a cool purchase here. That way it's also not specifically limited to the NTSC.

trazodont:
I like the idea of a command clothing vendor (maybe restricted to the job on the ID? Idk how) but the sprite def needs some work

code/modules/economy/commodity.dm
TheRyeGuyWhoWillNowDie, line 1: i don't think that's the coat

code/modules/economy/commodity.dm
```diff
@@ -745,6 +745,12 @@
 	desc = "Fancy. Possibly salvaged, possibly stolen, what's it to you?"
 	price = PAY_IMPORTANT
 
+/datum/commodity/contraband/nt_coat
+	comname = "Nanotrasen Winter Coat"
+	comtype = /obj/item/clothing/head/NTberet
```
TheRyeGuyWhoWillNowDie, line 750: that doesnt seem like the coat

itchybones:
I'm not sure about the shady trader version being identical to the one the NTSC gets. imo, it should be more like the armor it sells, used and grimy. Making a whole second version of it with sprites is probably out of scope, so I'd rather not see this in the merchant's inventory at all.

JORJ949:
Like the command vendor idea, definitely not for this PR though, will remove from trader for now though

TobleroneSwordfish:
Ehh on balance I think we're going to say no. Some thoughts from dev chat:
- The palette really doesn't fit a jacket, it looks weirdly oversaturated
- NTSC already spawns with enough junk, including a space suit that does the same thing but better

PR #20269, opened 2024-08-11
Rework atlas post

JORJ949:
## About the PR 
Reworks atlas post a bit to add medical supplies
image

## Why's this needed? 
Listening posts should have medicine (even if its just donk pockets+medibot) and atlas had none, also a little landing pad is nice and some post gear was missing e.g. syndicate hologram module


Cherman0:
Are those windows indestructable? Otherwise it would not be hard to break in. It could also use a light or two on that docking pad.

TobleroneSwordfish:
Not sure about the full medical area, most listening posts just have the donks and mayybe a medkit. Would make more sense to have a bunk area like the original I think, as this is meant to be a long term surveillance outpost.

JORJ949:
> Are those windows indestructable? Otherwise it would not be hard to break in. It could also use a light or two on that docking pad.

Those are indestructable yes, and I have since added a light

JORJ949:
> Not sure about the full medical area, most listening posts just have the donks and mayybe a medkit. Would make more sense to have a bunk area like the original I think, as this is meant to be a long term surveillance outpost.

If its long term, they need somewhere to shower (and eat pizza in the shower??? (idk why I did this))

maps/atlas.dmm
```diff
@@ -17792,6 +17933,22 @@
...
+/obj/item/parts/robot_parts/leg/right/standard,
+/obj/item/parts/robot_parts/leg/left/treads,
+/obj/item/parts/robot_parts/leg/right/treads,
+/obj/storage/crate/medical{
+	name = "medical supplies crate"
+	},
+/obj/item/storage/firstaid/brain,
+/obj/item/storage/firstaid/brute,
+/obj/item/storage/firstaid/fire,
+/obj/item/storage/firstaid/oxygen,
```
TobleroneSwordfish, line 17949: Again, this is a lot more medical supplies than any other listening post has (afaik). Try to keep them relatively standardized.

TobleroneSwordfish (changes requested):

PR #20271, opened 2024-08-11
Rework the Still to be more in line with botany equipment

Egregorious:
## About the PR 

Reworked the still to be more in line with other botany objects, changes include:
- A sound effect on brewing.
- Drag and drop loads items one by one instead of all at once.
- Drag and drop from the floor only loads items of the type dragged.
- Drag and drop works with items adjacent to you, they don't need to be adjacent to the still.
- Attempting to put items in an already full still fails without consuming the item.
- Proper feedback for interactions involving items with a SUPPRESSATTACK flag.
- Screws and wrenches bolt and unbolts the still to the floor.
- Loading messages use correct terminology (it's probably not correct terminology, and honestly it's probably more confusing)

Also removed the type check that's performed on items and replaced it with a truthy check on 'brew_result'. This means any type of item will now be brewed if it has a valid 'brew_result', it no longer needs to be a plant or food item. I assume this interacts poorly with stackable items. 

## Why's this needed? 

The still was a little outdated. 

## Changelog 

```changelog
(u)Egregorious
(+)The still is a little nicer to use.
```


Glamurio:
Oh. I see. You moved your PR.

code/modules/chemistry/tools/dispensers.dm
```diff
@@ -80,6 +80,20 @@
 			return TRUE
 		return ..()
 
+	proc/bolt_unbolt(mob/user)
+		if(!src.anchored)
+			var/turf/T = get_turf(src)
+			if (istype(T, /turf/space))
+				boutput(user, SPAN_ALERT("What exactly are you gunna secure [src] to?"))
```
Glamurio, line 87: "gunna" instead of "gonna"

Egregorious:
I felt like I changed what the purpose of the PR was so I opted to just remake it, not sure if that was the right thing to do but it's what I did.

code/modules/chemistry/tools/dispensers.dm
```diff
@@ -730,55 +742,72 @@ TYPEINFO(/obj/reagent_dispensers/watertank/fountain)
...
+				if (user.loc != staystill) break
 				if (src.reagents.is_full())
-					user.show_text("[src] is full!", "red")
+					boutput(user, SPAN_ALERT("[src] is full!"))
 					break
 				if (src.brew(P))
 					amtload++
 					qdel(P)
+					playsound(src.loc, 'sound/effects/bubbles_short.ogg', 30, 1)
+					sleep(0.2 SECONDS) // faster than from the floor because you organised your crate well
```
Glamurio, line 782: If you want the action to have a specific cooldown, then I think making an actionbar is better than having arbitrary sleep times, especially if you vary them depending on if it's a crate or not. Also nothing in the game behaves this way when being loaded from a crate, might warrant a message to the player.

code/modules/chemistry/tools/dispensers.dm
```diff
@@ -730,55 +742,72 @@ TYPEINFO(/obj/reagent_dispensers/watertank/fountain)
...
-				else
-					continue
-			user.visible_message(SPAN_NOTICE("<b>[user]</b> finishes stuffing items into [src]."),\
-			SPAN_NOTICE("You finish stuffing items into [src]."))
+				if (user.loc != staystill) break
+				if (P.type != itemtype) continue
+				if (src.brew(P))
+					qdel(P)
+					playsound(src.loc, 'sound/effects/bubbles_short.ogg', 30, 1)
+					sleep(0.3 SECONDS)
```
Glamurio, line 808: See above comment

Glamurio (commented):
Some commentary, also a general question I wasn't sure about, is it possible to retrieve items from the still again? This wasn't an issue with apples, but since we now allow literally any item we might need an eject button. I know we check for brew_result, but this might cause issues later. Maybe a drag and drop remove like with artlab machinery.

Egregorious:
> is it possible to retrieve items from the still again?

So I didn't change any of the functionality of the brewing process itself, it just outright deletes the item and replaces it with liquid in the still, but it always did that. It would need a deeper rework to add an 'eject' button. As it stands, just don't make important items brewable - or do and then laugh at anyone stupid enough to try it. The original type check was arbitrary because it checks for valid 'brew_results' anyway.

code/modules/chemistry/tools/dispensers.dm
Egregorious: I actually copy-pasted that message from the fountain and thought I shouldn't change it because it was already written like that.

code/modules/chemistry/tools/dispensers.dm
Egregorious: That's fair, I just copied the functionality from the way the compost bin is coded and changed the crate time because I thought it was cute. I'll probably just try to find how other crates work rather than add an action bar since I feel like an action bar is a bigger departure from expectation.

Egregorious:
Hm, I say arbitrary but I guess the type checks do hedge against bugs involving 'brew_result's that might get added at runtime..

Glamurio:
> > is it possible to retrieve items from the still again?
> 
> So I didn't change any of the functionality of the brewing process itself, it just outright deletes the item and replaces it with liquid in the still, but it always did that. It would need a deeper rework to add an 'eject' button. As it stands, just don't make important items brewable - or do and then laugh at anyone stupid enough to try it. The original type check was arbitrary because it checks for valid 'brew_results' anyway.

Reworking the brewing process is outside of the scope of this PR, then I say just leave it. I'd still rather get more QoL features generally speaking. Currently only fruits and vegetables have a brew result so it won't cause problems.

TobleroneSwordfish (changes requested):

code/modules/chemistry/tools/dispensers.dm
```diff
@@ -730,55 +742,67 @@ TYPEINFO(/obj/reagent_dispensers/watertank/fountain)
...
+				playsound(src.loc, 'sound/effects/bubbles_short.ogg', 30, 1)
 				return
 			else  ..()
-		else ..()
+		// create feedback for items which don't produce attack messages
+		if (W && (W.flags & SUPPRESSATTACK))
+			if (isfull)
+				boutput(user, SPAN_ALERT("[src] is already full."))
+			else
+				boutput(user, SPAN_ALERT("Can't brew anything from [W]."))
```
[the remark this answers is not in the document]
Egregorious: I've tried to fix this by excluding the OPENCONTAINER flag, which prevents the message for every relevant case I can think of.

PR #20272, opened 2024-08-11
Adds a new clown traitor item: The twirl gun

Chatauscours:
## About the PR 
The twirl gun:
The more you twirl it, the higher the damage, but the higher the chance to shoot yourself when you fire it. For every two damage added, adds 1 percent to shooting yourself when you fire it, caps off at 200 damage. Each twirl adds 20 damage.
Only the clown can twirl it without shooting themselves (there's still a small chance you will even as a clown), but everyone can try to shoot it. Non clowns have twice the chance of shooting themselves.
The gun costs 5 TC.

Unsure about whether some other jobs might take it as well, and about the TC cost.

## Why's this needed? 
Funny.

## Changelog 

```changelog
(u)Chatauscours
(*)Adds a new clown traitor item: the twirl gun. The more you twirl, the higher the damage and the chance to land a bullet in yourself.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -1084,6 +1084,14 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 	job = list("Captain", "Head of Personnel", "Research Director", "Medical Director", "Chief Engineer")
 	can_buy = UPLINK_TRAITOR
 
+/datum/syndicate_buylist/traitor/twirlgun
+	name = "Twirl gun"
+	item = /obj/item/gun/energy/twirlgun
+	cost = 5
+	desc = "<b>Spin me right round...</b>"
+	job = list("Clown")
+	can_buy = UPLINK_TRAITOR
+
 /////////////////////////////////////////// Surplus-exclusive items //////////////////////////////////////////////////
 
 ABSTRACT_TYPE(/datum/syndicate_buylist/surplus)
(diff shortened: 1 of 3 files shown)


pgmzeta:
What if this was a rubber band gun instead of a generic energy gun? I think that theming fits the twirl X clown combo, and could allow some unique properties e.g. reflective projectiles.

TDHooligan:
i like the twirl idea. funny but probably too risky an item atm i think. i dont really think traitors should be blowing their own head off. 
definitely make using it inhand twirl it so you dont have to *emote directly. add a cooldown to it too

second, i don't think a gun that regularly rolls to kill you with no control over it will be popular. 
could do like, "twirl to gain a random amount of charge, if you overcharge it it shoots you" instead

maaaaaaybe could make it less directly murder-y? perhaps it could be a toxic silly string gun or something. a silly string flamethrower that applies glue effects and burns people with acid/fire would be wicked.

TobleroneSwordfish:
I... really don't get this? I'm not sure what gameplay it's meant to allow other than a clown spinning it a few times and instant killing someone, with a small chance of dying themselves. None of that seems particularly funny or interesting?

Tarmunora:
> I... really don't get this? I'm not sure what gameplay it's meant to allow other than a clown spinning it a few times and instant killing someone, with a small chance of dying themselves. None of that seems particularly funny or interesting?

Also, clowns already have a lot of exclusive traitor items. I really can't see this one making it in

PR #20278, opened 2024-08-12
Refactors captain's antique

Cherman0:
## About the PR 

Refactors the captain's antique laser rifle to make use of the matsci statistics of the coponents instead of the quality ranking. The lens is scored based on it's alpha (transparency) with lower values being better. The coil is scored based on its electrical conductivity, with a +2 bonus if it's also an energy source. The effects of the gun (ie projectiles, charge costs, etc.) were unchanged. 

As for under-the-hood changes, the antique is repathed to always be a gun, instead of being secretly swapped with a real gun when you build it. You can't shoot it unless it has all the right parts in working order. To accomodate that, the displaycase was given a `/captain` subtype which holds the antique, which was put on every map. Previously they were all just identically-varedited normal display cases.

## Why's this needed? 

Quality is a fuck and I'm pretty sure this is the last thing that actually uses it. Also turns a copypasted varedit into a proper suptype which is good too.

## Changelog 

```changelog
(u)Cherman0
(+)The captain's antique now requires certain material statistics to make a more powerful gun instead of "quality".
```


Cherman0:
Sord and flourish were codeownered because it changed the path of an item on donut2 and oshan, respectively. Since it isn't really a mapping PR I'm going to unrequest the reviews.

TheRyeGuyWhoWillNowDie:
Big fan of this. Current method for making max-strength antique gun parts is really, really stupid, this encourages more experimentation instead of just doing math

Sbmhawk:
Does anything even have a electrical conductivity of 10?
Iirc electrum/gnesis only have 8

RelentlessGarbage:
Seems like a much more sane implementation.

Going to echo the above comment regarding materials with over 10 electrical, looking through the code puts electrum as the most conductive material available at 9 electrical (from the basic search I did). Might want to change the break points for the +3 quality boost for the coil to start at 9.

Cherman0:
> Does anything even have a electrical conductivity of 10? Iirc electrum/gnesis only have 8

No, statistics only go up to 9. There is a +2 bonus appended for it being an energy source, though. It is possible to get the max score in that category but it is quite an obscure combination.

Munien7:
Love this change, the quality system has always bugged me (especially since this is the only mechanic I'm aware of that uses it) and is pretty obscure, not to mention annoying quirks like ores losing their prefix when put in the rockbox etc. Would be glad to see it go and have a much more reasonable system like this.

RelentlessGarbage:
Nitpick: you spelled it "anqtique" (/obj/item/gun/energy/anqtique). Although Super Protector! Friend probably would call it an anqtique gun, we probably shouldn't.

Cherman0:
goddamn it

RelentlessGarbage:
It's still called an anqtique in the typeinfo and where it's declared as a station bounty. :p

code/obj/item/gun/energy.dm
```diff
@@ -216,25 +216,116 @@ TYPEINFO(/obj/item/gun/energy)
...
+				if(src.myLens)
+					user.put_in_hand_or_drop(src.myLens)
+				src.myLens = item
+				item.set_loc(src)
+			else
+				user.show_text("You need to unscrew the maintenance panel first!", "red")
+
+	canshoot(mob/user)
+		//configures the projectiles and makes sure it can actually shoot
+		if(src.determineProjectiles() < 1)
```
TobleroneSwordfish, line 273: This probably shouldn't be running the entire evaluation proc every time the gun fires, why not just evaluate it when the parts are swapped out?

TobleroneSwordfish (changes requested):
This is great, thanks for finishing this!

code/obj/item/gun/energy.dm
Cherman0: The hypothetical in my mind was if someone put something that would destroy itself in there, like a burning material. Since I didnt implement it exposing the parts to heat it probably isnt a very common scenario though. Moreso it's just convenient because youd need to check that it actually has all the parts anyways.

Cherman0: Moved the quality analysis call to happen when you screw the maintenance panel closed. It also now cannot be fired while the maintenance panel is open.

Tarmunora: > The hypothetical in my mind was if someone put something that would destroy itself in there, like a burning material. Since I didnt implement it exposing the parts to heat it probably isnt a very common scenario though. Moreso it's just convenient because youd need to check that it actually has all the parts anyways.

One option you have there is to add an `Exited()` override and additionally recalculate whenever an atom leaves the contents of the gun

code/obj/item/gun/energy.dm
```diff
@@ -298,7 +298,7 @@ TYPEINFO(/obj/item/gun/energy/antique)
 			if(180 to INFINITY)
 				//not good enough to be functional
 				return 0
-		switch(src.myCoil.material.getProperty("electrical") + ((src.myCoil.material.getMaterialFlags() & MATERIAL_ENERGY) ? 2 : 0))
+		switch(src.myCoil.material.getProperty("\"electrical") + ((src.myCoil.material.getMaterialFlags() & MATERIAL_ENERGY) ? 2 : 0))
```
Tarmunora, line 301: ??

Cherman0: I uh... hm. It worked in testing? I legitimately dont remember doing that. Fixed

PR #20283, opened 2024-08-12
Remove "HoS" from owned lawbringer names

JORJ949:
## About the PR 
Removes the "HoS" part from owned lawbringer names (currently "HoS [NAME]'s Lawbringer")

## Why's this needed? 
Not all lawbringer owners are HoSes

### DIFF
--- a/code/obj/item/gun/energy.dm
+++ b/code/obj/item/gun/energy.dm
@@ -1346,7 +1346,10 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 			if (H.bioHolder)
 				boutput(M, SPAN_ALERT("[src] has accepted the DNA string. You are now the owner!"))
 				owner_prints = H.bioHolder.Uid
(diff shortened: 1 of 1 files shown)


Cherman0:
Should probably just never say "HoS" so it isn't a weird meta-identifying thing that an impersonator can't copy

JORJ949:
> Should probably just never say "HoS" so it isn't a weird meta-identifying thing that an impersonator can't copy

If there IS a HoS with a lawbringer its probably very obvious that its not the HoS as HoSes are both announced and on radio (and the lawbringer is only there if a hos is in the round, so its much harder to impersonate)

Tarmunora:
> Should probably just never say "HoS" so it isn't a weird meta-identifying thing that an impersonator can't copy

Agreed, should be all or nothing.

TobleroneSwordfish:
On balance I'm just going to close this, it's the HoS's gun and seeing "HoS Kevin Bakins" or whatever every time it gets mechscanned is pretty funny.

PR #20287, opened 2024-08-13
New special stab attack for forks/screwdrivers and some daggers

TDHooligan:
[FEATURE] [OBJECTS]
## About the PR 

Adds a new special attack for some less-loved stabbing implements. On a successful stab, the special attack cooldown is reduced to allow very fast follow-up stabs.

It can't crit (no bonus bleed tiers), does 20% less damage, and will be massively impacted by armor due to the low damage of weapons it's applied to except the nukeops dagger. 

At peak performance, you can match a fire extinguisher's DPS screwdrivering an unprotected face, dealing a bit of bleed too. instead it quickly falls behind against armor (a jumpsuit alone cuts the DPS down ~20%). 

https://github.com/user-attachments/assets/e52c70a8-b38a-4d26-8cd5-bfb7e427a12e

## Why's this needed? 
Screwdrivers have gotten the bum deal losing their once infamous eye-stabbing quality. I also think that each tool in a toolbox should be scary as a weapon in some cases, and a screwdriver is a prime shiv.

Daggers are kinda wimpy, and it seems like a good change to make forks at least a little scarier as an improvised weapon.

## Changelog 

```changelog
(u)TDHooligan
(+)Screwdrivers, daggers and forks now have a new special attack.
```


AXLFAN2000:
This is dope

code/modules/items/specials/ItemSpecials.dm
```diff
@@ -499,6 +499,53 @@
...
+			var/obj/itemspecialeffect/jab/effect = new /obj/itemspecialeffect/jab
+			effect.set_dir(direction)
+			effect.setup(turf)
+
+			var/hit = FALSE
+			for(var/atom/A in turf)
+				if(isTarget(A))
+					A.Attackby(master, user, params, 1)
+					hit = TRUE
+					last_use = world.time - (cooldown - success_cooldown)
```
Tarmunora, line 540: ough this feels very cursed but it works

TDHooligan: yea maybe there could be a nicer-looking set function but there isnt much else to get around it

PR #20298, opened 2024-08-14
Hopefully stops the thing where nonsimulated tiles in mah air groups

Laboredih123:
## About the PR 
Tries to remove ourselves from any air groups that may cause problems when we become space or somethin.

## Why's this needed? 
bug bad


Laboredih123:
while not all of the errors are gone, i think the remaining few are due to stuff happening during the lagchecks, i believe the pr is good to merge now as i havent seen any reports of broken atmos
if there is indeed broken stuff, lemme know

PR #20302, opened 2024-08-14
Remove robotalk implants from robotics manufacturers

TobleroneSwordfish:
[BALANCE] [GAME OBJECTS]
## About the PR 
Removes the recipe to produce machine translator implants from robotics manufacturers. Doesn't remove the item or associated helper types as they are still used for some gimmick things.

## Why's this needed? 
It's really weird that the dedicated "machine talk" channel is almost always compromised and therefore can't be reliably used to discuss law changes or coordinate silicon-based crime. A lot of players have taken to using the silicon PDA group, but that's clunky and can still be snooped on by standard PDA means.
I dunno, people have been complaining about this thing for a long time and while I definitely am exactly the player who prints one at the start of every round because *I like to know things*, I can definitely see how it discourages silicons from coordinating with each other to execute their laws.

## Changelog 

```changelog
(u)LeahTheTech
(*)Removed the machine translator implant recipe from the robotics fabricator. Silicons may now communicate freely without fear of being overheard.
```

### DIFF
--- a/assets/maps/prefabs/underwater/nadir_safe/prefab_water_sketchy.dmm
+++ b/assets/maps/prefabs/underwater/nadir_safe/prefab_water_sketchy.dmm
@@ -398,7 +398,6 @@
 /turf/simulated/floor/scorched2,
 /area/prefab/sea_sketch)
 "bK" = (
-/obj/item/implant/robotalk,
 /turf/simulated/floor/plating/scorched,
 /area/prefab/sea_sketch)
 "bL" = (
--- a/code/obj/machinery/manufacturer_subtypes.dm
+++ b/code/obj/machinery/manufacturer_subtypes.dm
@@ -162,7 +162,6 @@
 		/datum/manufacture/robup_visualizer,
 		/datum/manufacture/robup_efficiency,
 		/datum/manufacture/robup_repair,
-		/datum/manufacture/implant_robotalk,
 		/datum/manufacture/sbradio,
 		/datum/manufacture/implant_health,
 		/datum/manufacture/implant_antirot,


444Portal (approved):
this is for the best, ive seen it described as a noob trap and frankly having to use the PDAs for all communication if you dont want a 'gotcha! i get to stop you before you do anything!' moment. rip though

klookiyam:
I like this change, but there is one interaction that I will miss if this goes through: evil robotocist's ability to communicate with their legion of borgs.

On the plus side, this might force people to learn how to use the custom PDA mail group feature.

wolfolotl:
are they available anywhere else in normal play, like from traders or in locked crate loot tables? could be a nice middle step between having them Always Available and taking them out of usual play completely, and it'd also make the hands they _do_ fall into much less Reliable

soleilan:
> I like this change, but there is one interaction that I will miss if this goes through: evil robotocist's ability to communicate with their legion of borgs.
> 
> On the plus side, this might force people to learn how to use the custom PDA mail group feature.

Could give traitor roboticists an option to buy a 1 TC machine translator implant maybe? I am okay with antags having a way to get the implant somehow, but rather regular crew can't

soleilan:
though I guess if only traitors had access to it, silicons could metagame so hard since if any non-human is there they must be evil. Maybe some implants in the debris field or shuttle merchant is the right play then

JORJ949:
I like this change but as others have said would love if there was a way for evildoers to obtain it, perhaps stick one in the listening post so its not guaranteed whoever has it is antag (though it is almost certain)

Sbmhawk:
I agree with the listening post idea, having it be trader/crate loot still has the issue of it being a noob trap in almost all cases.

soleilan:
I don't think it being a DOC or other merchant purchase would be a terrible idea, though ideally in combination with the TC or listening post, so long as its decently expensive. That way its at least somewhat not an automatic metagame giveaway.

Glamurio:
Hate to see it go, but agree that it's good for the game.

TobleroneSwordfish:
It might be okay to add some debris loot/traitor ways to acquire this, but honestly I think I prefer the idea of it just being guaranteed secure in normal gameplay. The psychological effect of knowing your chat channel is always secure vs "probably" secure is big.

JORJ949:
>  The psychological effect of knowing your chat channel is always secure vs "probably" secure is big.
what if it only let you talk TO the robots but not hear them, this way you can be "hey can i get some help over here" but not "hey general radio the ai is more rogue than I want it to be"

Emmett1Brown:
> > The psychological effect of knowing your chat channel is always secure vs "probably" secure is big.
> 
> what if it only let you talk TO the robots but not hear them, this way you can be "hey can i get some help over here" but not "hey general radio the ai is more rogue than I want it to be"

silicon PDA group could be used for that probably. can be overheard if someone happens to log all PDA messages entries but you know, thats the case for all PDA messaging.

PR #20303, opened 2024-08-14
Increase chances for clone defects & add new defects

Glamurio:
[Balance] [Medical]

## About the PR 
Recategorizes all existing cloner defects:
"Minor" defects are now things that slightly inconvenience a player, things such as damage on cloning, vocal chord disfiguration, etc.
"Major" defects are now things that highly inconvenience players, such as missing limbs, organs or bio effects such as narcolepsy.

It adds a few new cloner defects:
Baldness (Minor), Deafness (Major), Blindness (Major), Narcolepsy (Major), Plasma Lungs (Major), Rare Mutant Race (Major).
Splits up Brute and Burn damage, and adds Tox damage as individual damage defects.
The weights have been placed low for more harmful effects such as large damage sources.

The biggest change is the change to the defect_count calculation.
Every cloned person now gets at least 2 defects, with a bias to skew towards 2-3.
There's now a 60% chance to roll a minor defect and a 40% chance to roll a major defect.

## Why's this needed? 
Over the years, there's been a some threads on the topic of cloning being too safe, as well as some conversation on the Discord. The general consensus seems to be that there's a general dislike toward cloning being an easy way to fix a patient without having to specifically engage in any of the medical systems. Doctors will often rather let someone die and have them be cloned than try to save them because it's easier.

Rather than make it harder or add restrictions to clone, this PR attempts to make cloning less "safe". Not more lethal, not more annoying, but being cloned now carries more risk if there isn't a doctor present. The likelihood of needing medical assistance goes up dramatically the more often one is cloned, and even after the first clone you may end up with a disfigured face or a missing arm.

The intent is for cloning to be a last resort that should be used if the patient is already dead, and encourages having someone with medical knowledge nearby that can assist the person, but does so in a way that doesn't require it.

Threads for reference:
https://forum.ss13.co/showthread.php?tid=20207
https://forum.ss13.co/showthread.php?tid=22809

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)NT Clone Pods are now less safe to use. In case of defects, please refer to the nearest available doctor.
```


code/modules/medical/cloner_defects.dm
```diff
@@ -238,36 +258,56 @@ ABSTRACT_TYPE(/datum/cloner_defect/missing_organ)
...
 	init()
-		src.data = list("organ_string" = pick("tail", "left_kidney", "right_kidney", "stomach", "intestines", "appendix"))
+		src.data = list("organ_string" = pick("tail", "left_kidney", "right_kidney", "stomach", "intestines", "appendix", "liver",))
 
 /datum/cloner_defect/missing_organ/major
 	severity = CLONER_DEFECT_SEVERITY_MAJOR
 
 	init()
-		src.data = list("organ_string" = pick("liver", "left_eye", "right_eye", "left_lung", "right_lung")) // no skull, head, brain; instant death sucks. also removed heart
+		src.data = list("organ_string" = pick("left_eye", "right_eye", "left_lung", "right_lung")) // no skull, head, brain; instant death sucks. also removed heart
```
Tarmunora, line 270: the minor organ missing stuff is all things that are either relatively or entirely harmless, and was classified as minor as such
making it major and putting liver in there (which will kill you via tox damage) is a strange choice

code/modules/medical/cloner_defects.dm
```diff
@@ -238,36 +258,56 @@ ABSTRACT_TYPE(/datum/cloner_defect/missing_organ)
...
 		src.owner.set_mutantrace(data["new_mutantrace_type"])
 
-// TODO abstract status effect defects more
+	rare
+		severity = CLONER_DEFECT_SEVERITY_MAJOR
+		weight = 30
+
+		init()
+			src.data = list("new_mutantrace_type" = null)
+			data["new_mutantrace_type"] = pick(filtered_concrete_typesof(/datum/mutantrace, /proc/filter_rare_muteraces))
```
Tarmunora, line 310: vetoing all of this. The mutantraces selected here range from "literal aliens" to "ancient jank mutantrace that should never see the light of day again", and none of that should be popping out of the cloner when cloning a human

code/modules/medical/cloner_defects.dm
```diff
@@ -450,24 +489,61 @@ ABSTRACT_TYPE(/datum/cloner_defect/organ_damage)
...
+	desc = "Subject's ears have become damaged and the subject is now deaf."
+	severity = CLONER_DEFECT_SEVERITY_MAJOR
+	effect_type_pool = list(/datum/bioEffect/deaf)
+
+/datum/cloner_defect/biotrait/sleepy
+	name = "Spontaneous Sleep Deprivation"
+	desc = "Subject's sleep center was scrambled and as a result, subject randomly falls asleep."
+	severity = CLONER_DEFECT_SEVERITY_MAJOR
+	effect_type_pool = list(/datum/bioEffect/narcolepsy)
+
```
Tarmunora, line 523: should all be nonstackable

Tarmunora (changes requested):
Overall I think these changes are probably too punishing, especially the likelihood of just stacking enough instant-damage defects to instakill, and being able to just roll 2-3 debilitating defects on first clone. Without adding any features to mitigate this (e.g.: "'fresh' (i.e.: recent scan - presumably of the (unrotten) corpse) cloning records have reduced defects", or "something something give geneticists a job"), this seems like too much stick with no carrot

code/modules/medical/cloner_defects.dm
Glamurio: This was by accident, actually.

code/modules/medical/cloner_defects.dm
Glamurio: Technically they are unstackable because the trait check itself checks if the user has the trait, the reason clumsy had it was because of aloe's variant that has multiple options which I left unchanged. Stackable would prevent someone from rolling blind and deaf, but if that's desired then I can amend it.

Glamurio:
> Overall I think these changes are probably too punishing, especially the likelihood of just stacking enough instant-damage defects to instakill, and being able to just roll 2-3 debilitating defects on first clone. Without adding any features to mitigate this (e.g.: "'fresh' (i.e.: recent scan - presumably of the (unrotten) corpse) cloning records have reduced defects", or "something something give geneticists a job"), this seems like too much stick with no carrot
I kind of purposefully went so hard on it to make it an actual threat. Death isn't scary anymore, and I think it's causing problems both on a mechanical level as well as with people's attitude. It's hard to feel intimidated when you can be cloned easily with no repercussion. "Fresh" scans are probably the most common, so I'm not sure if that change alone would do much. The average person dies less than once per round I'd argue, and dying more than once should really make itself shown.

That being said, I do love the idea of implementing a way for geneticists to make the cloner more safe. Maybe they could spend some of their time researching cures for cloner defects, or have ways to easily remove them. Perhaps the recently revived person could go and get their defects checked at cloning, as an extra pit stop or something.

I'm down to open a discussion chat on the subject, see what people think.

code/modules/medical/cloner_defects.dm
```diff
@@ -479,6 +543,18 @@ ABSTRACT_TYPE(/datum/cloner_defect/organ_damage)
 		. = ..()
 		src.owner.vdisfigured = TRUE
 
+/// Basically makes you bald. Truly the worst of the defects.
+/datum/cloner_defect/baldness
```
TobleroneSwordfish, line 547: Hard no. Erasing a large part of someone's character design at random is not a good or interesting defect.

code/modules/medical/cloner_defects.dm
```diff
@@ -450,24 +477,61 @@ ABSTRACT_TYPE(/datum/cloner_defect/organ_damage)
...
+	severity = CLONER_DEFECT_SEVERITY_MAJOR
+	effect_type_pool = list(/datum/bioEffect/blind)
+
+/datum/cloner_defect/biotrait/deafness
+	name = "Permanent Cochlea Damage"
+	desc = "Subject's ears have become damaged and the subject is now deaf."
+	severity = CLONER_DEFECT_SEVERITY_MAJOR
+	effect_type_pool = list(/datum/bioEffect/deaf)
+
+/datum/cloner_defect/biotrait/sleepy
```
TobleroneSwordfish, line 506: Cloning defects should ideally be something fairly obvious that can be fixed or mitigated by the average MD right after you spill out of the tube, being randomly stunned for the rest of the round is not that and will just make people angry.

TobleroneSwordfish (changes requested):

code/modules/medical/cloner_defects.dm
```diff
@@ -247,28 +277,36 @@ ABSTRACT_TYPE(/datum/cloner_defect/missing_organ)
 	severity = CLONER_DEFECT_SEVERITY_MAJOR
 
 	init()
-		src.data = list("organ_string" = pick("liver", "left_eye", "right_eye", "left_lung", "right_lung")) // no skull, head, brain; instant death sucks. also removed heart
+		src.data = list("organ_string" = pick("left_eye", "right_eye", "left_lung", "right_lung", "liver")) // no skull, head, brain; instant death sucks. also removed heart
+
+/datum/cloner_defect/missing_organ/heart
+	severity = CLONER_DEFECT_SEVERITY_MAJOR
+	weight = 10
```
Azrun, line 284: This seems heavily reliant on an active and available medbay. What is the expected life expectancy for this? 

This seems like a major bummer for a new player to bump into.

code/modules/medical/cloner_defects.dm
```diff
@@ -450,24 +499,61 @@ ABSTRACT_TYPE(/datum/cloner_defect/organ_damage)
...
+		else if (src.owner.bioHolder.HasEffect(data["bioeffect_id"]))
+			src.owner.bioHolder.RemoveEffect(data["bioeffect_id"])
+		..()
+
+/datum/cloner_defect/biotrait/plasmalungs
+	name = "Plasma Lung Infection"
+	desc = "Subject's lungs have become infected and are now only capable of breathing plasma."
+	severity = CLONER_DEFECT_SEVERITY_MAJOR
+	weight = 50
+	effect_type_pool = list(/datum/trait/plasmalungs)
```
Azrun, line 521: Due to the nature of how this is typically contracted, I don't think it makes much sense for this to be a generic abnormality. Unless it is rare.

Glamurio: I had a similar thought. Perfectly okay with making it rare, or cutting it entirely otherwise.

code/modules/medical/cloner_defects.dm
Glamurio: Admittedly extremely rare occurrence, but the concern is warranted. A lot of these were made with a doctor on stand-by in mind, or at the very least someone with medical knowledge, partially to deter random passer-by cloner usage.

If it's a major concern for new players, we could exclude newbees from contracting it.

Azrun: > partially to deter random passer-by cloner usage

I'd prefer that the behavior can account for the current health/status/population of the server or is going to be conducive to providing agency to players.  

> The intent is for cloning to be a last resort that should be used if the patient is already dead, and encourages having someone with medical knowledge nearby that can assist the person, but does so in a way that doesn't require it.

Doesn't this require someone with medical knowledge to be nearby and assist?

code/modules/medical/cloner_defects.dm
Azrun: Either one.  This requires a lung transplant or "immediate" coordination with Science or Engineer doesn't it?

Azrun (changes requested):

Azrun:
Still waiting for some resolution regarding: https://github.com/goonstation/goonstation/pull/20303#issuecomment-2290078702

I think detrimental effects to cloning is interesting and adds weight to death.  That said I think there should be some form of escalation if how deaths are handled (Having dynamic weight scaling in `add_random_cloner_defect()` with minor vs major to exclude or reduce likelihood of major on first clone or current defect count, or time in the round).  

Some of these defects seem like they should be the result of an antagonists tampering with the cloner or cloning records.

code/modules/medical/cloner_defects.dm
```diff
@@ -186,24 +182,59 @@ ABSTRACT_TYPE(/datum/cloner_defect)
...
 		..()
-		var/brute = rand(src.data["damage_amount"])
-		owner.TakeDamage("All", brute + rand(-data["extra_variance"], data["extra_variance"]), data["damage_amount"] - brute + rand(-data["extra_variance"], data["extra_variance"]))
+		src.do_damage()
 
+	proc/do_damage()
+		var/damage = rand(src.data["damage_amount"])
+		owner.TakeDamage("All", damage + rand(-data["extra_variance"], data["extra_variance"]), data["damage_amount"] - damage + rand(-data["extra_variance"], data["extra_variance"]))
 
 /datum/cloner_defect/ouch/big
```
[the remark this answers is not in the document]
Glamurio: Personally, I think so. This just aligns with my reasoning of "the average person doesn't die a lot".

code/modules/medical/cloner_defects.dm
Glamurio: Yeah, I'm convinced. We can exclude it from normal selection and perhaps make it emag-only or something.

code/modules/medical/cloner_defects.dm
Glamurio: It's a bit more complicated because of the way the trait is coded, replacing the lungs does not fix it, which imho is an issue with the trait itself and very confusing design. Intermediate solution is to supply the patient with salbutamol until plasma can be ordered.

Because replacing the lungs does not fix it, I would actually argue this should be culled, at least not as long as the trait remains that way. Too unintuitive.

Glamurio:
> Still waiting for some resolution regarding: [#20303 (comment)](https://github.com/goonstation/goonstation/pull/20303#issuecomment-2290078702)
> 
> I think detrimental effects to cloning is interesting and adds weight to death. That said I think there should be some form of escalation if how deaths are handled (Having dynamic weight scaling in `add_random_cloner_defect()` with minor vs major to exclude or reduce likelihood of major on first clone or current defect count, or time in the round).
> 
> Some of these defects seem like they should be the result of an antagonists tampering with the cloner or cloning records.
Lowering the likelihood of major I'm okay with, on live right now it's impossible to roll major defects on the first time, and since people don't generally die more than once a round it's actually rare to ever see virtually any major defects. That's why I've put a lot of previous minor defects into the major category and adjusted the weights a bit.

There was also an idea of introducing an "extreme" category, worse than major, but ultimately I couldn't come up with enough hugely detrimental effects that weren't basically just close to "you spawn dead", and figured it yields no massive major difference than the weights already present in major.

What's the biggest concern with cloning defects? That people basically end up dead after spawning? Or that they are essentially permanent? If it's the latter what do you think of the approach to hard cap defects? That way the first few will be detrimental, but it ensures that nobody ends up spawning dead on arrival. I know that runs counter to our current system but may be better for the game.

If it's the former, I'm struggling to find something sufficiently satisfying to implement that wouldn't just end up being a chore at best. If it becomes commonplace to have to waddle over to genetics after being cloned to get your defects sorted, it sort of also defeats the idea of having defects in the first place, and genetics will also potentially become completely overrun in certain situations (and genetics already struggles majorly with being too absorbed in their work).

Azrun:
At this point I think maybe I’m not sure I fully understand the intent of this PR.  I think there are some discrepancies in the description as well as compared to the implementation where I think starting there may be better.

Is the requirement of medical assistance a goal or a non-goal?  The PR description seems contradictory in saying that on one hand the likelihood of needing medical assistance goes up while also saying cloning doesn’t require someone to assist.

Certain defects appear to be near lethal or lethal without intervention, and much more so as they are stacked.  A number of them vary in the amount of inconvenience they are. The descriptions references not making cloning “not more lethal, not more annoying”, so can you provide more context/intent?

PR #20305, opened 2024-08-15
Resprites Gimmick Bombs, Mousetrap Overlays, and Mousetrap Car Bombs

MitchTheUser:
[GAME OBJECTS]

## About the PR 

Resprites the 4 main Gimmick Bombs in the game with new Normal and Armed sprites, which now flash animated instead of a static image:

**- Butt Bomb**
Has a new sprite instead of a old bloody butt with a weird piece attached on top. Now a butt attached with wired to a timer. When armed, the timer starts flashing.

**- Hotdog Bomb**
Has a new sprite instead of a hotdog suit cut in half with a weird piece attached on top. Now a full-on hotdog with a pin jammed into the top. When armed, the grill marks start flashing.

**- Owl Bomb**
Has a new sprite instead of a decapitated owl with a weird piece attached on top. Now it is a seemingly living owl with a pin stuck in the head. When armed, it comes to life, with flashing eyes and flapping wings.

**- Gold Bomb**
Has a unique sprite set instead of using the Banana Bomb sprites, and also turns into gold when armed, causing it to give out a few gold sparkles before blowing up.

GimmickBombs

Also resprites all the Overlays for putting things on top of mouse traps, as well as Mousetrap Car Bomb. They now have new sprites :

**- Butt** 

**- Synth Butt**

**- Butt Bomb**

**- Cyber Butt**

**- Remote Signaler**

**- Grenade (Uses the Chemical Grenade since it is the most generic looking one)**

**- Pipe Bomb**

**- Mousetrap Car Bomb**

OverlayNew

## Why's this needed? 

mousetrap-chem

Most of the Gimmick Bombs look outdated, and the armed icons are very basic, and don't currently work due to there being no special Var allowing it to switch when armed. The Gold Bomb also uses the Banana Bomb Sprite instead of having it's own sprite. Mousetrap Overlays are pretty old, using outdated and flat sprites, same with the Mousetrap Car Bomb.

## Changelog 

```changelog
(u)MitchSmith
(+) Resprited Gimmick Bombs, Mousetrap Overlays, and Mousetrap Cars!
```


Laboredih123:
the banana grenade is funny

MitchTheUser:
> the banana grenade is funny

Yeah I don't know why the Gold Bomb used it's sprite instead of having it's own sprite, but now it has it's own set.

code/obj/item/grenades.dm
```diff
@@ -934,7 +935,10 @@ ADMIN_INTERACT_PROCS(/obj/item/gimmickbomb, proc/arm, proc/detonate)
 	proc/beep(i)
 		var/k = i/2
 		sleep(k*k)
-		flick(icon_state+"_beep", src)
+		if (src.icon_state_armed)
+			src.icon_state = icon_state_armed
```
Sovexe, line 939: The goal here should be to alternate iconstates based on the original (broken) code

 ```suggestion
		if (icon_state_armed)
			if (icon_state == src::icon_state)
				icon_state = icon_state_armed
			else
				icon_state = src::icon_state
```

File: code/obj/item/grenades.dm [hunk 2]
```diff
@@ -934,7 +935,10 @@ ADMIN_INTERACT_PROCS(/obj/item/gimmickbomb, proc/arm, proc/detonate)
...
+		if (istype(src, /obj/item/gimmickbomb/gold))
+			src.setMaterial(getMaterial("gold"))
```
Sovexe, line 941: this should be handled in the subtype not in the abstract parent

Sovexe (changes requested):

code/obj/item/grenades.dm
MitchTheUser: Transferred it to the Subtype by including the gold material setting in a modification of the beep proc. Seems to work right.

code/obj/item/grenades.dm
MitchTheUser: Got it, though I had to replace the ::'s with .'s because if not the armed animation wouldn't stay looped properly otherwise. I assume that was just a (typo?) in the suggestion code.

Sovexe: `::` is the scope operator used here to retrieve initial values. The problem here is that some of the bombs rely on alternating between their unarmed and armed icons, while as others expect to be left in armed icon state. The best fix is to animate the armed icons

MitchTheUser: > `::` is the scope operator used here to retrieve initial values. The problem here is that some of the bombs rely on alternating between their unarmed and armed icons, while as others expect to be left in armed icon state. The best fix is to animate the armed icons

So should I change it to add the scope operator :: back in like in your suggestion code, or is the code fine the way I have it right now? I don't see in the code where some bombs "rely on alternating" and some "expect to stay in armed state" so I'm not really sure what you are referring to? The beep code only affects Gimmick Bombs, of which there are only four main bombs under that subtype, and I don't see anything making them act any different when undergoing the Beep proc.

Sovexe: when armed many items such as grenades have an animated icon that flashes to make it obvious. The gimmick bombs dont because it was long since bugged. The original flick didnt do anything. If you just want to have them sit on the armed icon state with a static icon then theres no need for the code to flip the icon back and forth each beep, Though the code now has pointless checks like icon_state == icon_state which would never fail

code/obj/item/grenades.dm
```diff
@@ -934,8 +935,15 @@ ADMIN_INTERACT_PROCS(/obj/item/gimmickbomb, proc/arm, proc/detonate)
...
 		var/k = i/2
 		sleep(k*k)
-		flick(icon_state+"_beep", src)
+
+		if (icon_state_armed)
+			if (icon_state == src.icon_state)
+				icon_state = icon_state_armed
+			else
+				icon_state = src.icon_state
+
```
Sovexe, line 944:
```suggestion
```

Sovexe: add to arm()

```suggestion
		if (icon_state_armed)
			icon_state = src.icon_state
```

MitchTheUser: > add to arm()
> 
> ```
> 	if (icon_state_armed)
> 		icon_state = icon_state_armed
> ```

Done, it actually looks a lot better now since it immediately switches to it's armed animation instead of waiting for the beep, which makes it look more natural and realistic for the pin/lever to immediately go up.

Sbmhawk:
Aw, I liked the banana grenade sprite

PR #20314, opened 2024-08-15
Remove matsci quality

Cherman0:
## About the PR 

The quality property of materials and asteroids and all associated logic is removed.

## Why's this needed? 

Quality no longer actually does anything at all, Its implementation is highly arbitrary and janky (like materials and asteroids in general) and had a lot of weirdness associated with it. As such, it has been removed for code quality purposes.


ZeWaka:
cat-meme-ezgif com-video-to-gif-converter

Tarmunora (approved):

ZeWaka:
ytoull be removing rarity next this is so ssaad

TobleroneSwordfish:
it doesn't doo anything Waka!!

PR #20317, opened 2024-08-15
Janitor backtank sprayer

TobleroneSwordfish:
[EXPERIMENTAL] [GAME OBJECTS]
## About the PR 
Adds a cleaning spray gun that pulls from a tank worn on the back: 
image
- Intended to replace the Tsunami spray bottle as a more reasonable and interesting option, available from cargo(?)
- Has 3 modes: a narrow 1 tile wide spray, the standard 3 tile wide spray like the old tsunami, and a big 5 tile wide short range spray. All modes are projectiles that can be fired in any direction.
- Has 3 `W_CLASS_SMALL` size storage slots.
- Pushes loose items around, the wide shot also pushes people.

This is intended as a replacement for the Tsunami and will be available through a cargo purchase instead of a debris field spawn.
Sprites (mostly) by Flaborized, the rest by me.
Wind-up sound modified from here: https://pixabay.com/sound-effects/wind-up2-106350/ and licensed under Pixabay's standard license.
Name suggested by TekoTheTeapot.

TODO:

- [x] Worn/inhand sprites
- [x] More granular reagent overlay sprites for the tank
- [X] Decide how much crime should be allowed, ie does it melt if you pour phlog into it?
- [x] Map the gun reagent overlay onto the tank storage
- [x] Hand crank or something silly to pressurize it?
- [x] Balance mode cost
- [x] Fiddle with diagonal shot length, maybe just give some free tiles to make it look right.
- [x] Fix the wide shot going through walls.
- [x] Oh god worn icon reagent overlays??
- [x] Better sound effect for running out of juice
- [x] Unscruncle the bloody inhands
- [x] Remove old tsunami from debris field

Feedback on design/balance very much welcome!

## Why's this needed? 
The Tsunami is the ultimate janitor power fantasy but it is A) only found as a one of in the debris field B) is infinite and a little boring to use. This sprayer aims to keep the cool factor of annihilating grime with waves of cleaning spray while making it a bit more interactive and available.

```changelog
(u)LeahTheTech + Flaborized
(*)Added the WA-V3 (get it?) janitorial sprayer, available in a crate from cargo for 4500 credits. See minor changes for details.
(+)The WA-V3 pulls from a 500u back-mounted tank and has three different modes with different reagent costs: narrow (10u), wave (20u) and w i d e (30u). The tank itself also has 3 "small" capacity inventory slots.
(+)Replaced the Tsunami spray bottle spawn with a cybernetic spray bottle spawn.
```


DecidedlyUnethical:
I really like this, but I can also see this becoming a poor man's Vega with the right mix.

TobleroneSwordfish:
Maybe I should just use the fire extinguisher whitelists hmm

RelentlessGarbage:
Any EMAG effects? If there are any reagent locks/restrictions, I'd imagine that emagging the device would remove them, or give some other functionality.

Consider that "cleaner flamers" already exist, with their restrictions being that they are contraband and will get you in hot water with Beepsky, the construction cost, and consumption of gas upon firing alongside the space cleaner payload. The revamped floor buffer also allows for enhanced cleaning of large areas. What makes this proposed item stand out/act as a sufficient replacement for the Tsunami?

cropseyy:
Honestly would go for being able to put in it what you can put in the janitor car thing- essentially making it a handheld one of those. 

Maybe exclude lube. Unless it's EMAGed...

Mister-Moriarty:
Code-wise it may be worth investigating merging the Vega and (new) Tsunami backtank behaviours.

TobleroneSwordfish:
> Code-wise it may be worth investigating generalising the Vega and (new) Tsunami backtank behaviours to share common code.

I did think about that but it's about 5 lines of code right now and I don't really like the way the Vega works anyway.

Mister-Moriarty:
> I did think about that but it's about 5 lines of code right now and I don't really like the way the Vega works anyway.

My solution to this would be to just refactor both, ensuring that whatever new system is created works for both and is relatively clean. Although refactors should be separate from feature PRs, so perhaps that could be done later.

TobleroneSwordfish:
> What makes this proposed item stand out/act as a sufficient replacement for the Tsunami or these other, existing Janitor tools?

Okay so compared to a flamethrower it:
- Has higher tank capacity (400 -> 500)
- Can be stored in a bag
- Isn't contraband
- Can be ordered from cargo instead of having to be crafted
- Can't fire sticky or acidic chemicals (same restrictions as a fire extinguisher)
- Is generally more efficient for each tile cleaned (pressurized water or something)
- Can clean with just water, at the cost of covering the floor in water (flamethrowers don't leave puddles and so won't clean with just water loaded)
- Pushes things around

I hope that's enough to make it distinct and worth using?
Compared to other cleaning tools:
- It's more flexible than the janitor cart but takes more effort to use and has a smaller tank
- It's better than a mop but will waste water if used on small areas, same with sponge, spray bottle etc.

DecidedlyUnethical:
Hmm. I'm thinking about other things to make it distinct from the floor buffer, and maybe the blast could clean off walls and windows? Like, you are literally hosing an area down, it makes sense that it would be able to clean spatters blood/oil/fungus decals off of walls and windows, while also being massively less efficient than just walking up and sponging it off.

Tarmunora (commented):
Frankly, I'm not huge on letting it be a chem-thrower, especially with applying in a 5-tile wide line, especially when considering stuff like lube.
Moreover, it seems like using it for cleaning would become really annoying once it runs out of cleaner if it starts making giant puddle fields

TobleroneSwordfish:
Wellll fire extinguishers already exist, are a lot easier to get and can spread lube almost as effectively 🤷‍♀️ 
I could remove the 5 wide mode I guess, but then it is basically just a fancy fire extinguisher.
As for cleaner, it's pretty easy to make/buy more of it and we can always tweak the recipe or add a buyable barrel of it from cargo or something?

I'm not married to the current concept though, and I kind of like the idea of making it fire a continuous beam somehow instead. It just seems a shame not to let this interact with chemistry at all.

Glamurio:
I love this, I think having it be a chem thrower with a sensible whitelist should be fine imo. Can always add a traitor variant that has no whitelist.

Tarmunora:
Ok for whatever reason I had been under the impression this was like, a roundstart piece of janitor equipment
If it's something that one has to go out of their way for, my concerns are significantly lessened

code/modules/chemistry/tools/extinguisher.dm
```diff
@@ -1,3 +1,20 @@
...
+	"pacid",
+	"phlogiston",
+	"big_bang",
+	"clacid",
+	"nitric_acid",
+	"firedust")
+
 /obj/item/extinguisher
 	name = "fire extinguisher"
 	icon = 'icons/obj/items/items.dmi'
```
Tarmunora, line 20: notably, `foof` and `infernite` are missing from this list because they're specialcased in extinguisher code.
Should also block them from the sprayer

Tarmunora (changes requested):

TobleroneSwordfish:
Oh, well the point of this PR is to remove the tsunami from the debris field and add this to cargo to replace it.

AXLFAN2000:
I'd love if the tsunami was removed.  This looks super charming and fun

TobleroneSwordfish:
Okay I'm done feature creeping I SWEAR

Stushcinta:
Used it for the first time today, I love it, the Tsunami always kinda sucked because all you did was click and mess disappears, there's nothing to it, it's not fun at all. This thing is great, it's absolutely incredible for cleaning huge messes, it leaves water behind though, so it's good to drag a fluid tank behind you to suck it up (And then you can refill the pack)

It's a little annoying that it pushes items but that's good because it means it's not perfect for every situation. If there's huge messes in hallways, this thing is amazing, and it feels fun to use, and i honestly wanna experiment to find out the firefighting potential. Just be careful when beepsky's around, he doesn't like to get wet.

Seriously though, A+++ addition, great job!

AXLFAN2000:
This is a huge step up from the tsunami, been super fun to use and feels more like an actual Job Tool with the refilling and different modes.  Replacing the tsunami itself with the borg sprayer was a nice touch too

Stushcinta:
A few suggestions/questions

If you add a hand crank to pressurise it (Which is an idea i love), would it be possible that higher pressure levels lead to longer range and pushing items? That way you could use it at low pressure to just clean like, small areas and not mess up items.

Would it be possible to have a mode in it that sucks up fluids? Working kinda like a wet vac kinda thing, that could make it replace the mop for cleaning up big puddles (Mopping infinitely and cleaning up all the water in oshan is funny and it makes you feel like a hero but it is kinda dumb and not the most interesting thing), and it'd kinda make more sense in that situation, and it'd have a better flow where you don't have to drag a fluid canister around to make it more practical.

Would it be possible for there to be a slot on the backpack that only takes the gun? It just makes more sense to me that that's where it'd be stored, also maybe we need to add the gun to the mop locator, if you misplace it and can't find it you probably need to order a whole new pack, which is a pain.

But yeah, i love it, beautiful job on creating it.

Stushcinta:
> A few suggestions/questions
> 
> If you add a hand crank to pressurise it (Which is an idea i love), would it be possible that higher pressure levels lead to longer range and pushing items? That way you could use it at low pressure to just clean like, small areas and not mess up items.
> 
> Would it be possible to have a mode in it that sucks up fluids? Working kinda like a wet vac kinda thing, that could make it replace the mop for cleaning up big puddles (Mopping infinitely and cleaning up all the water in oshan is funny and it makes you feel like a hero but it is kinda dumb and not the most interesting thing), and it'd kinda make more sense in that situation, and it'd have a better flow where you don't have to drag a fluid canister around to make it more practical.
> 
> Would it be possible for there to be a slot on the backpack that only takes the gun? It just makes more sense to me that that's where it'd be stored, also maybe we need to add the gun to the mop locator, if you misplace it and can't find it you probably need to order a whole new pack, which is a pain.
> 
> But yeah, i love it, beautiful job on creating it.

Actually upon further testing, making it suck up fluids would probably be bad as if it gets any fluids inside that can't be used, it's impossible to use anymore, i put in a bug report after extensive testing to see if i could fix it.

PR #20324, opened 2024-08-15
Remaps the mining outpost

Cherman0:
## About the PR 

image

Remaps the mining outpost from scratch. Organizes things into rooms in a manner that makes at least some sense. Also moves the martian trader more north since it was weirdly close to the mining outpost.

## Why's this needed? 

Makes the mining outpost a more feature-complete place to actually mine. 

## Changelog 

```changelog
(u)Cherman0
(*)The mining outpost has been completely remapped.
```


RubberRats:
I understand being more spread out was part of your intention, but this seems excessive. All the major rooms are separated by long, generic hallways, which feels more like padding than anything else. The current outpost isn't perfect, but it has character, and it didn't feel like it was big just for the sake of being big.

Additionally, in my experience the mining outpost does not receive much foot traffic. Barring something like security doing a raid on the place, it's usually just 1-2 miners and the occasional assistant looking for gear. It's already very out of the way relative to anything on the station, and I don't think its design needs to bend over backwards to be a good crime spot.

As for suggestions: break up/shorten the hallways between rooms, move the mining, refinery, and hangar areas closer together (unloading mining pods into the rockbox is currently not very convenient), and add more rooms/decoration. The mining outpost should look run-down and lived in, like the space diner.

AXLFAN2000:
The most common complaint/s I hear about the mining outpost's layout is that points of interest are too spread out despite the rooms themselves being pretty small.  I don't feel that way about the current outpost, but that's definitely the energy I get from looking at this proposed layout change  --  I like the concept of an outpost revamp, but this feels like it adds a lot more walking down empty hallways where it doesn't need to.  It would flow better in terms of gameplay for the actual working rooms (magnet/gear room, refinery, podbay) to be closer together.

Cherman0:
I can look into compacting things down. Though I feel that the old mining post felt more abandoned than lived in. There's just like, random garbage and an inordinate amount of dirt strewn everywhere like it was inhabited by Fallout NPCs.

Cherman0:
I've removed the hallways and compacted everything in. The SDMM picture in the main body has been updated to reflect the changes.

LaCumbiaDelCoronavirus:
I'm all for updating the mining outpost but I say that I prefer the old one. Heres some of my *personal* thoughts about it:

- It was set as having been built into an asteroid which justifies it being rather big. Although I agree that the original mining outpost was cluttered and messy which yea wasn't good, this feels too small, not just for that reason too.
- This is more of balance and imo but the mining outpost should probably **not** have a prebuilt comms console. Original had a prebuilt comms disk and a frame: thats it.
- Strong opinion: Why did you remove the mechanics outpost? I really see no reason to just delete it

and some suggestions:
- If it's possible, maybe separate the rooms with a hallway or 2? I will say i'm a fan of the new rooms you made for it but its not as if the mining outpost has to be small
- Keep the original size of the mining magnet, please. This one is much smaller
- A quantum telescope inside the mining room too, not just in space, would also be convenient

And I really agree with RubberRats' suggestions too

Last thing:
Im pretty sure these wires wont connect, as theyre nubs
mining room


in podbay


in comms console room and cafeteria room (under the chair)

AXLFAN2000:
Wait, when I commented before I was unaware that this overhaul also completely removed the mechanics outpost.  I'm very against that  --  I spend a decent amount of time there, and there's always other players I run into. Frankly, it often gets more foot traffic than the mining outpost itself.  I don't think removing that is a good change at all.

CoolKakapo:
I don't really like this. The old mining outpost feels really unique a nice. I cant really take an ingame look but from the picture it just looks like a generic small mining station. I like how the old one is made in an asteroid , and i also really like how the old one looks and feels used, like it is constantly used instead of being completely clean. My opinion is that im up for a new mining station but it needs the same feel and atmosphere of it, instead of looking like a generic mining spot.

soleilan:
I'm mostly confused where this came from. What exactly is the issue with the current mining outpost? You say to make it more feature complete for mining, but the current mining outpost has everything you need to mine already? And it would be a lot simpler to just add whatever is missing.

This seems to just remove the interesting widespread half asteroid half catwalk mining outpost + mechanics outpost combo for a very regular looking mining outpost instead and I'm not sure why.

Cherman0:
Added back some hallways and asteroid tunnels. Pic in the main PR body is updated.

Cherman0:
> I'm mostly confused where this came from. What exactly is the issue with the current mining outpost? You say to make it more feature complete for mining, but the current mining outpost has everything you need to mine already? And it would be a lot simpler to just add whatever is missing.
> 
> This seems to just remove the interesting widespread half asteroid half catwalk mining outpost + mechanics outpost combo for a very regular looking mining outpost instead and I'm not sure why.

Aside from the obvious answer of it being just really outdated in terms of items and objects, I feel that the mining outpost should be very regular looking. Presumably it was designed with the intention to be an actual industrial mining site and not a serial killer den, so it could stand to not be bursting at the seams with garbage spawns, empty crates, random bolted-on rocks, and conveyor belts that go literally nowhere. That catwalk wasn't originally just for stashing bodies on, it's actually the remnant of a mobile AI rail system that never got fully implemented, for example.

Sord213:
Not really a fan overall. The rooms all look....generalized and bland. The current mining outpost looks exactly like what it is. It was the home base for miners way back when, then when mining was moved on station it was abandoned, now it's only sometimes used when mining blows up or there's too many miners. They gotta clean the cobwebs out but it's still usable. It's an NT asset but mostly abandoned. This looks like some prefabricated structure bolted in place and in active use by NT.

Generally I don't see this as an improvement over the current.

Cherman0:
The mining outpost was moved out of the debris field and into the mining field a while back, so it isnt really abandoned. It's the center of an active mining zone now.

Tarmunora:
Not a fan of the vertical arrangement - widescreen view area lends itself much better to a more horizontal layout of the important bits, especially the mining magnet.

Maturia:
Losing the mechanics outpost is a major gameplay change and should probably be atomized into a different PR.

Cherman0:
I might come back to this at a later date but at the moment I'm not really interested in continuing to maintain it

PR #20325, opened 2024-08-16
Remove Antibiological cargo crates

JORJ949:
## About the PR 
Removes the "Anti-Biological Hazard Supplies" from cargo's orderable crates

## Why's this needed? 
This is only ever ordered for blob and it absolutely demolishes them, its also an armory crate and flamethrowers can be made anyway, it should take more effort than ordering 4 of these to defeat an entire gamemode with ease

If this ends up with blobs being overpowered then i'm honestly fine with that as a change of pace from what they are right now, which is "this round will end in 15m when we flamer the blobs with our 15 flamers"

## Changelog 

```changelog
(u)JORJ949
(*)Removed Anti-Biological Hazard Supply crate from cargo's orderable crates
```

### DIFF
--- a/code/modules/economy/supply_packs.dm
+++ b/code/modules/economy/supply_packs.dm
@@ -1136,17 +1136,6 @@ ABSTRACT_TYPE(/datum/supply_packs)
 	containertype = /obj/storage/crate/packing
 	containername = "Utility Belt Crate"
 
-/datum/supply_packs/antibio
-	name = "Anti-Biological Hazard Supplies"
-	desc = " A couple of tools for combatting rogue biological lifeforms."
-	category = "Security Department"
-	contains = list(/obj/item/gun/flamethrower/assembled/loaded,
-					/obj/item/storage/box/flaregun)
-	cost = PAY_IMPORTANT*2
-	containertype = /obj/storage/secure/crate
-	containername = "Anti-Biological Hazard Supplies (Cardlocked \[Security Equipment])"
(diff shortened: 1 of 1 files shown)


BatElite:
Hey now, there's plenty of crates that are basically single-purpose. The flamethrower crate isn't all that special in that regard. (In fact if it wasn't cardlocked I reckon we might see some more crime use out of it.)

Given "If this ends up with blobs being overpowered then i'm honestly fine with that..." I can't but see this more as a knee-jerk than a serious balancing attempt. Blob needs a not lot but not that.

JORJ949:
> 
> Given "If this ends up with blobs being overpowered then i'm honestly fine with that..." I can't but see this more as a knee-jerk than a serious balancing attempt. Blob needs a not lot but not that.

Yeah it is a bit overkill but I would like to see how the crew fairs without a near-infinite source of flamethrowers and instead has to make them, giving the blob a little more time to prepare.

AXLFAN2000:
I'm really not sure about this tbh, I don't like the way blob rounds are just "QM orders 300 of the crate -> crew victory" on repeat but I worry that just straightup removing the crate from purchase will make it a lot harder to mobilize crew to fight the blob at all (the other thing I don't like about blob rounds lol). I dunno.  I'm not a huge fan of the way things are now (IRT blob), though, so I'd be curious to see how this or other changes actually play out in rounds

ZeWaka (approved):

PR #20333, opened 2024-08-16
Changes pipe pressure limits

Cherman0:
## About the PR 

Increases the amount of pressure that the pipes can withstand and decreases the amount that plating the pipes increases it by. More specifically, the pipe burst threshold increases from 15 MPa to 50 MPa and the plating effect decreases from `density^2` to `density`. Overall, this has the effect that unplated pipes are more durable than before but plated pipes are not as durable as before.

## Why's this needed? 

`density^2` is crazy scaling and results in absurdly resistant pipes. On the flipside, a standard no-frills chamberburn will break unplated pipes if you put more than like, half a plasma tank in the hot loop. This should normalize things a bit and make it so that plating the pipes is not always necessary for any burn.

## Changelog 

```changelog
(u)Cherman0
(+)The base maximum pressure of pipes has increased to 50 MPa but the effects of plating have been reduced.
```

### DIFF
--- a/code/modules/atmospherics/machinery/pipe.dm
+++ b/code/modules/atmospherics/machinery/pipe.dm
@@ -83,7 +83,7 @@
 	/// How well we share temperature.
 	var/thermal_conductivity = WALL_HEAT_TRANSFER_COEFFICIENT
 	/// Pressure needed before the pipe gets a chance to burst, see proc/effective_fatigue_pressure for the value that takes into account material stats too
-	var/fatigue_pressure = 150*ONE_ATMOSPHERE
+	var/fatigue_pressure = 500*ONE_ATMOSPHERE
 	/// Can this pipe rupture?
 	var/can_rupture = FALSE // Currently only used for red pipes (insulated).
 	/// How broken is our pipe.
@@ -96,7 +96,7 @@
 	alpha = 128
 
 /obj/machinery/atmospherics/pipe/simple/proc/effective_fatigue_pressure()
-	return src.fatigue_pressure * ((src.material?.getProperty("density") ** 2) || 1)
+	return src.fatigue_pressure * (src.material?.getProperty("density") || 1)
 
 /// Returns list of coordinates to start and stop welding animation.
 /obj/machinery/atmospherics/pipe/simple/proc/get_welding_positions()


klookiyam:
The scaling is similar to what graphene hardening compound does https://github.com/goonstation/goonstation/blob/dd58767db38e306205d061c9ef7ec213285ce9c1/code/modules/chemistry/Reagents-Misc.dm#L730.

When you're doing a pipe burn you can quickly hit the pressure limit of easily obtainable materials (I've ruptured bohrum pipes multiple times doing this on donut 2).

If you're reducing the pressure scaling of materials, then material plating becomes useless for its main use case: an easier to obtain, less effective ghc for high intensity burns.

Cherman0:
Graphene hardening compound is `2^reagent volume` and scales up to 1 GPa. Assuming you have more than like, 6 drops of GHC per pipe it is much better than material plating. As it should be, since the station is not littered with beakers of GHC like it is steel plates. 

For reference, currently steel plating increases a pipe's burst point from 15 MPa to 240 MPa (16x multiplier). Using Bohrum increases that to 540 MPa (36x multiplier). This multiplier stacks multiplicatively with GHC, meaning with 30 drops of GHC (the maximum that has any additional effect) and bohrum plating the burst point goes up to 36 GPa (2400x multiplier).

With this PR, steel plating would increase a pipe's burst point from 50 MPa to 200 MPa (4x multiplier). Using Bohrum increases that to 300 MPa (6x multiplier). Basically, this makes the case of "an easier to obtain, less effective ghc" pretty similar while making unplated pipes not completely useless for anything and while nerfing the multiplicative exponential stacking of using both.

TDHooligan:
i don't see why the burst floor can't just be raised, afaik nobody's complaining that plating is too strong, and making it exponentially weaker at higher values is a certifiable way to get the mechanic ignored. 

there aren't any hard and fast numbers provided so i dont know what the average hellburn or pipeburn sits at these days, but everything atmos gets crazy exponential fast, and linear scaling things quickly stop being useful.

MyNameIsRomayne:
I haven't heard of this scaling ever being a problem? If anything, it's quite nice to see something scale greater than linearly without a fuss in regards to the TEG these days.

TobleroneSwordfish:
See above: plating scaling well with very rare high density materials is intentional and steel already being a good reinforcement helps when mining isn't around and you still want to do something better than a charburn.

PR #20338, opened 2024-08-16
[TGUI REACT] last two things for listinputmodal

MyNameIsRomayne:
## About the PR 
Get listinput modal working for TGUI React


ZeWaka:
yuh

PR #20343, opened 2024-08-16
Resprites the Garden Trowel sprite and gives it an in-hand sprite, and also adds SFX when digging and planting in plant pots

Wrench-1:
## About the PR 
Resprites the Garden Trowel so it looks a lot less ugly and is more inline with the current artstyle.

trowel sprites

Includes new inhand sprites, and also a sound effect will play when digging up plants, and planting them in plant pots.

trowel inhands

All Sprites are by Bork, and the code implementation by me. This is also my first time adding SFX to items so if anything looks off code wise, please let me know!

## Why's this needed? 
The old sprite needed some serious updating as it pretty rough and old. Now it looks inline with the games current artstyle. More art cohesion is good!

## Changelog 

```changelog
(u)Bork and Wrench
(*)Resprites the Garden Trowel, and gives it a new In-Hand and Sound Effect when digging and planting. Sprites by Bork
```


TobleroneSwordfish (approved):

Gannets (approved):
Yeah

PR #20344, opened 2024-08-17
Make lightswitches deconstructible

MyNameIsRomayne:
## About the PR 
Lightswitches can be deconstructed with a screwdriver, wirecutters, and multitool using the decon device. Deconstructing a switch will turn the lights on if they weren't on already.

## Why's this needed? 
Medass wont leave me alone

Screenshot 2024-08-16 190454

(In seriousness, this is just a nice kind of change to have for making space on station)


TobleroneSwordfish (approved):

ZeWaka:
But are they constructable?

MyNameIsRomayne:
> But are they constructable?

always have been. you can scan then

PR #20352, opened 2024-08-17
'Toggle Scope' item ability

TDHooligan:
[BALANCE][FEATURE][OBJECTS]
## About the PR 
Allows toggling of sniper rifles with an item ability. when disabled you can sprint normally.
image

## Why's this needed? 
Sniper rifles implicitly disable sprint. Which is awkward.

## Changelog 

```changelog
(u)TDHooligan
(*)Sniper rifles now have an item ability to toggle their scope. You can now sprint with sniper rifles equipped.
```


Sbmhawk:
Could there be a way to toggle scope vs sprint

Marksman nukie already blunders enough due to scoping, this would end up being a nerf to the ohr imo.

LaCumbiaDelCoronavirus:
Might want to add a center dot/reticle, i feel like that would be helpful

TDHooligan:
more than happy to put this on another gun that isn't explicitly the OHR *shrug

Tarmunora:
> more than happy to put this on another gun that isn't explicitly the OHR *shrug

Yeah, I think that it would feel a bit off on the OHR

Valtsu0:
I think it's great on the OHR. While i do think it ends up as a nerf due to the scoping jank, nerfing OHR is fine and it's not in the scope (hah) of this pr to fix

flappybatpal:
Changed the behavior after conversation with Leah.

PR #20354, opened 2024-08-17
Cleans up redundant hair-related code.

RubberRats:
[INTERNAL] [CODE QUALITY] [BUG]
## About the PR 
This PR consolidates and cleans up code related to hair and barbery to remove certain redundant lines.

First, the `create_wig()` proc has been given an argument that determines whether or not the target's hair should be removed afterward. This logic is now baked into the proc instead of being written separately every time it is needed.

Second, this PR adds an `is_bald()` proc to human mobs, which returns true if the target's hair styles are all set to none. Parts of the code that check this manually have been changed to use this proc instead.

Additionally, this PR fixes a bug that was caused by one instance of `create_wig()` being called on someone without checking that they had any actual hair, creating a weird bald-wig that had no worn sprite. If this does happen it will create a bald cap instead. 

## Why's this needed? 
I was planning to add something barber-related to the game, and having these changes would make my implementation far cleaner, on top of making barber code more readable. All but two of the mentions of 'create_wig()' removed the target's hair afterward, so it made sense to me to bake it into the proc as an argument. Also, bug fix good.

### DIFF
--- a/code/mob/living/carbon/human.dm
+++ b/code/mob/living/carbon/human.dm
@@ -2802,6 +2802,9 @@
 			ret += A
 	return ret
 
+/mob/living/carbon/human/proc/is_bald()
+	return istype(src.bioHolder.mobAppearance.customization_first,/datum/customization_style/none) && istype(src.bioHolder.mobAppearance.customization_second,/datum/customization_style/none) && istype(src.bioHolder.mobAppearance.customization_third,/datum/customization_style/none)
+
 /mob/living/carbon/human/proc/create_wig(var/keep_hair = FALSE)
 	if (!src.bioHolder || !src.bioHolder.mobAppearance)
 		return null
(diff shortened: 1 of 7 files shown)


code/mob/living/carbon/human.dm
```diff
@@ -2802,7 +2802,11 @@
 			ret += A
 	return ret
 
-/mob/living/carbon/human/proc/create_wig()
+/mob/living/carbon/human/proc/is_bald()
+	var/datum/appearanceHolder/AH = src.bioHolder.mobAppearance
+	return istype(AH.customization_first,/datum/customization_style/none) && istype(AH.customization_second,/datum/customization_style/none) && istype(AH.customization_third,/datum/customization_style/none)
```
ZeWaka, line 2807: Put this on multiple lines please

PR #20355, opened 2024-08-17
Wig glue trait that allows mutantraces to wear hats and hair

klushy225:
[TRAITS][MUTANTRACES][EXPERIMENTAL][INPUT-WANTED]
## About the PR 
Adds the Superglue Enthusiast trait that costs 3 points, replaces the player trinket with a crumpled up pouch of glue, and enables hair_override for the player. This allows mutantraces to have hats and hair concurrently, at the cost of quite a few trait points and a trinket of actual use.

The most glaring issue, however, is the character preview not showing the selected hair styles. I would make this a feature, but I'm simply not skilled enough to make that happen. (help with that would be appreciated!)

image

(I also find the trinket names coupled with a crumpled pouch of glue to be funny.)

## Why's this needed? 
The community is fairly divided on whether or not to allow mutantraces to have hats and hair, but most times I've mentioned this on the discord I've gotten positive feedback. I believe that mutantraces need more customization options, and while we wait on the mutantrace rework, I think this is a good first step. The way in which this trait is implemented would not at all impact the upcoming mutantrace rework, and could be very easily removed if needed.

## Changelog 

```changelog
(u)Klushy225
(*)Adds a Wig Adhesive trait, which allows mutantraces to have 'real' hair at the cost of 2 points.
```


DisturbHerb:
The whole trade-off with giving mutantraces wigs was to prevent this, if I recall correctly.

klushy225:
I understand that the tradeoff was intended, but I don't really agree with it. it's possible to wear hats and wigs at the same time in real life, and while I am fully aware that ss13 does not nor should not mirror real life, I don't really know why mutantrace players should be excluded from hats.

444Portal:
Trait needs a sprite and the glue itself could use a nicer sprite too, otherwise looks good!

klushy225:
> Trait needs a sprite and the glue itself could use a nicer sprite too, otherwise looks good!

yeah I'm not really good at pixel art- but I'll try!

DisturbHerb:
Implementation-wise, this would allow mutantrace hair to grow and be styled. We don't have very good support for wigs that can't do that, and I don't think lizards have hair follicles up there...

klushy225:
> Implementation-wise, this would allow mutantrace hair to grow and be styled. We don't have very good support for wigs that can't do that, and I don't think lizards have hair follicles up there...

oh I overlooked that- I'll fix that.

Vaskritaya:
I disagree with the general principle behind this change. Wigs were already an accidental encroachment on humans' "thing", one rationale for Kyle reverting his fix for the original wig oversight was that humans without wigs didnt have to give up their hat slot for hair. 

Mutantraces not having deeper customization shouldn't be an excuse to implement this, even as a stopgap. It'll be impossible to remove even if customization increases, and customization _won't_ increase, because this will already make mutantraces more customizable than humans to begin with.

Retrino:
> I disagree with the general principle behind this change. Wigs were already an accidental encroachment on humans' "thing", one rationale for Kyle reverting his fix for the original wig oversight was that humans without wigs didnt have to give up their hat slot for hair.
> 
> Mutantraces not having deeper customization shouldn't be an excuse to implement this, even as a stopgap. It'll be impossible to remove even if customization increases, and customization _won't_ increase, because this will already make mutantraces more customizable than humans to begin with. "The most permanent solution is a temporary one", as the saying goes.

it wasnt accidental, that was the point of wigs

Vaskritaya:
> 
> it wasnt accidental, that was the point of wigs

Clearly not to the devs, since Kyle was bullied into reverting an alteration made within days of wigs being merged which was supported by a large amount of the admins active at the time.

ZeWaka:
I don't really care, but I'm going to put this on

IPling:
Personally I like this change. My main gripe with this addition would have been of practically removing the interaction with trading style for gameplay benefits. For example with security officers having to choose to keep their wig for style or have a helmet for protection. 
Yet the high-price of the trait costing 3 points balances it out. So the only reason you would really pick it is so that you don't have to sacrifice your characters personality (The wig) just so that you could wear something more stylish (Like a hat). Which is an interaction that people who are attached to their own characters hate a lot - especially people on the RP servers - when it serves no real gameplay addition/benefit other than making the player feel like shit.
Other than that I don't really think the lowering of the uniqueness of mutant-races or humans matters that much, like yeah it is a bad thing. Yet I would say the benefits would outweigh it, since this is an incredibly common suggestion that a good majority of players enjoy.

Retrino:
> > it wasnt accidental, that was the point of wigs
> 
> Clearly not to the devs, since Kyle was bullied into reverting an alteration made within days of wigs being merged which was supported by a large amount of the admins active at the time.
> 
> Per kyle,
> 
> > I added multicolored lizards and the framework for letting mutantraces change colors years ago and expected someone would eventually add more on top of it to give mutantraces more customizibility without just giving them human hair. But nobody ever did. I think that hair is a distinctive trait of humans in game, and giving it to mutantraces undercuts it and the uniqueness of humans and mutantraces. They never had hair before and that wasn't an accident, I considered it an accident/mistake/what-have-you that it was essentially added to them in this trait.

the pr for wigs says "Fun for mutantraces and possibly humans!" in the description. i dont agree with the aggression toward kyle specifically when this all went down years ago, and i regret being a part of that, but the pr for wigs was specifically for this purpose. that is my only point there

Glamurio:
Just going to go ahead and tag it appropriately since you seemed to want to have this label on anyway and because this seems like a discussionable topic.

klushy225:
> Just going to go ahead and tag it appropriately since you seemed to want to have this label on anyway and because this seems like a discussionable topic.

Thanks! I messed up the label box :p

Glamurio:
Is there a good reason why we can't just merge this with the Bald trait? Isn't that trait currently mostly taken by people who want extra hair as a mutant race? Doesn't it make sense for a "wig wearer" to just spawn with wig adhesive? Would solve a lot of issues.

klushy225:
> Is there a good reason why we can't just merge this with the Bald trait? Isn't that trait currently mostly taken by people who want extra hair as a mutant race? Doesn't it make sense for a "wig wearer" to just spawn with wig adhesive? Would solve a lot of issues.

that would be very easy to do, but it would also make it difficult for borgs to get wigs if they want them

itsloley:
> Is there a good reason why we can't just merge this with the Bald trait? Isn't that trait currently mostly taken by people who want extra hair as a mutant race? Doesn't it make sense for a "wig wearer" to just spawn with wig adhesive? Would solve a lot of issues.

merging it to the bald trait & modifying it a bit for borgs is a good idea i think, or at least setting the price to 0, 1 at most. imo 3 points is _super_ steep for a change that'd only affect characters visually without any actual gameplay benefit, which kind of breaks the trend that all other traits with no cost created. the only reason i could see 1 cost making sense is cause of the cost trinkets have, i guess? this is very very good but i do think it shouldn't be its own very expensive trait

klushy225:
> > Is there a good reason why we can't just merge this with the Bald trait? Isn't that trait currently mostly taken by people who want extra hair as a mutant race? Doesn't it make sense for a "wig wearer" to just spawn with wig adhesive? Would solve a lot of issues.
> 
> merging it to the bald trait & modifying it a bit for borgs is a good idea i think, or at least setting the price to 0, 1 at most. imo 3 points is _super_ steep for a change that'd only affect characters visually without any actual gameplay benefit, which kind of breaks the trend that all other traits with no cost created. the only reason i could see 1 cost making sense is cause of the cost trinkets have, i guess? this is very very good but i do think it shouldn't be its own very expensive trait

I recently dropped the price to two points, and I do agree that it really should just be 0, but the trait point cost is the compromise for those who hold reservations against hair and hats for mutantraces. As for the borg bit- I can do that, but I worry that would make the code too messy for a change that will hopefully be replaced with a more elegant solution by the mutantrace rework. At the moment, keeping both the glue and bald trait makes some sense, as the wig glue trait is more or less the 'upgraded' version of it. Plus- some people might just want to start with a wig for whatever reason.

klushy225:
I'm not super sure why this is labeled as a merge conflict, considering that the bit of code that it's angry about only added a few lines, and overwrote/removed nothing. Could also be a case of me not understanding how github works.

Dazzelmist:
I just started playing as a Lizard and really wanted hair to make them look more distinctly female. But i also noticed that the hair occupying the hat slot makes a lot of things like using the det-gadget-hat more difficult.
If i want to use any hat for functionality then my character suddenly looks completely different.

This change would make playing them easier and I would get to not worry about the tradeoffs between the character looking like they are supposed to and gameplay.

HyenaShepherd:
I play as a Human so I've only got an outside perspective, but I've always considered the distinctions of mutantraces to be in their speech styles and unique identifiers. Bovines with their long 'm' sounds, roaches with their rad resistances, saurians with their temperature and 's' extensions. Making them wear wigs and no headgear always struck me as something tacked on without much benefit, and I've been wanting this for them solely for the opportunity for them to personalize their look a bit more. Seems like a good change to me.

klushy225:
So, by recommendation, I added the ability to glue wigs to a head with space glue in my local version, however, due to how mutrantrace hair works, it would have to be the color of the player's body, or it would change the scale/body colors _every_ time it's updated. Unless I can find a way around that, the PR will only include the trait /:

klushy225:
Closing my PR, as it's just extremely jank, and if mutantraces were to get hair, it would be best to not do it this way.

PR #20357, opened 2024-08-18
Proximity component

### LINKED ISSUE #20358: AI Eye can trigger proximity/radio assemblies
### Labels

[BUG][MINOR]

### Description

an active proximity/radio assembly on the ground will send out a signal if the AI eye passes by it

### Steps to reproduce

make a proximity/radio assembly
activate it and place it on the ground
get a signal everytime the ai eye passes over it on a pair remote signaller

### Expected Behavior

AI Eye is incorporeal and should probably not trigger a proximity detector.

Reported by: zelda4040404
Client version: 515.1635
Active test merges: #20186 and #20298
Round log date: 2024-08-17-23-30
Reported on: 2024-08-18 01:09:58
Round ID: 186079
Map: COGMAP

Laboredih123:
[PERFORMANCE][INTERNAL]
## About the PR 
Replaces the old hasproximity system with a proximity component. Add it to begin detecting movement in neighboring tiles. Range is also customizable, 1 tile out (3x3) being the default.
Fixes #20358

## Why's this needed? 
slow bad


code/datums/components/proximity.dm
```diff
@@ -0,0 +1,56 @@
...
+	var/atom/A = src.parent
+	var/turf/center = get_turf(A)
+	if(isnull(center))
+		return //we got kabloowied or something
+	for(var/turf/T as anything in src.listening_to)
+		UnregisterSignal(T, COMSIG_ATOM_ENTERED)
+		src.listening_to -= T
+	for(var/turf/T as anything in block(center.x-range, center.y-range, center.z, center.x+range, center.y+range, center.z))
+		RegisterSignal(T, COMSIG_ATOM_ENTERED, PROC_REF(Detect))
+		src.listening_to += T
```
Tarmunora, line 49: Gut feeling is that this (unregistering all signals then registering) on every move is pretty inefficient if there's a lot of overlap in the 'before' and 'after' blocks. e.g.: moving a single tile N/E/S/W with a 5x5 proximity only needs to +-5 turfs, not the whole 25
But I suppose the question is if the list operations needed to avoid unnecessary signal churn would actually be cheaper 🤷

code/datums/components/proximity.dm
```diff
@@ -0,0 +1,56 @@
...
+
+	src.enabled = enabled
+	src.range = range
+	src.listening_to = list()
+
+/datum/component/proximity/RegisterWithParent()
+	var/atom/A = src.parent
+	var/turf/center = get_turf(A)
+	if(ismovable(src.parent))
+		RegisterSignal(src.parent, XSIG_MOVABLE_TURF_CHANGED, PROC_REF(parent_moved))
```
Tarmunora, line 24: Using this complex signal allows proximity stuff to function contained - previous hasproximity stuff would only track turfs when directly on a turf.
Should either revise, or make it explicit in documentation/PR that this is not equivalent behaviour

Tarmunora (changes requested):

code/datums/components/proximity.dm
Laboredih123: ill note this as not equivalent.

Laboredih123: or could maybe add a way to have equivalent behavior hm.

code/datums/components/proximity.dm
Laboredih123: i dont know how i would only select some so :/, i mean its pretty cheap on the small cases which will also be the most common.

Tarmunora, line 49: could look something like this
```suggestion
	if(isnull(center))
		return //we got kabloowied or something
	var/L = block(center.x-range, center.y-range, center.z, center.x+range, center.y+range, center.z)
	for(var/turf/T as anything in (src.listening_to - L))
		UnregisterSignal(T, COMSIG_ATOM_ENTERED)
		src.listening_to -= T
	for(var/turf/T as anything in (L - src.listening_to))
		RegisterSignal(T, COMSIG_ATOM_ENTERED, PROC_REF(Detect))
		src.listening_to += T
```
But idk how expensive list subtraction is

Tarmunora (approved):
I think there's room for optimization but current usage is low enough that it's not a pressing concern

Laboredih123:
epico

PR #20366, opened 2024-08-19
Add alpha mask to hoodies to show hair

Glamurio:
[QoL] [Clothing]

## About the PR 
Adds an alpha filter to all hoodies which allows the display of hairs underneath.
image
image
Furthermore, all suits now track if they have a hood, and whether or not that hood is up.

The mask is added in in `/mob/living/carbon/human/update_face()`, because when I tried doing it anywhere else it lead to very weird behavior, such as the mask not updating or not being properly removed when disrobing. If anyone has a better method, I'll gladly use that.

The reason I'm using an overlay_mask instead of simply adding a filter to the head itself is because BYOND currently does not support icon_state, dir, and animation with alpha masks:
https://www.byond.com/forum/post/2817775

This also adds it so that hair now copies and filters applied on the head, as this was necessary for propagation:
`src.image_cust_one.filters = my_head.filters.Copy()`

Additionally, hoods now no longer count as covering the head of the user via `body_parts_covered`, I felt this a negligible change and was mostly done in order not to rewrite the render logic, because it would hide the hair.

----

One important thing to add here, **this does NOT include wigs**. Wigs are headgear, and by default, hoodies simply occupy the suit slot, meaning headgear renders on top. I can look into adding wig support as well but that may be out of scope for this PR.

## Why's this needed? 
Making the user bald when using a hoodie was a necessary side effect because there just wasn't a better way to do so that didn't cause immense clipping. This made hoodies extremely unpopular. I think that now people will be more frequently using these and it opens up a plethora of new customization options.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Hair will now be visible under the hood when wearing hoods from hoodies.
```


code/obj/item/clothing/suits.dm
```diff
@@ -18,6 +18,8 @@ ABSTRACT_TYPE(/obj/item/clothing/suit)
 	var/team_num
 	/// Used for the toggle_hood component, should be the same as the default icon_state so it can get updated with medal rewards.
 	var/coat_style = null
+	/// Used for hoodies (and anything that uses the toggle_hood component)
+	var/hooded = FALSE
```
TobleroneSwordfish, line 22: grr tightly bound components (not your fault dw)

TobleroneSwordfish (approved):
Looks great, glad you got it working!

DimWhat:
Maybe you could add this to stuff like helmets too 🥺

Glamurio:
> Maybe you could add this to stuff like helmets too later?🥺

Now that I know how it works, absolutely.

code/datums/components/toggle_hood.dm
```diff
@@ -22,11 +23,15 @@
...
 	src.hooded = !src.hooded
-	suit.over_hair = !suit.over_hair
-	suit.body_parts_covered ^= HEAD
+	src.suit.hooded = src.hooded
+
 	if (ismob(suit.loc))
 		var/mob/M = suit.loc
+		if (!isliving(M))
+			suit.body_parts_covered ^= HEAD
+			suit.over_hair = !suit.over_hair
```
Tarmunora, line 32: not `ishuman()`?

Glamurio: Was trying to preserve existing logic but sure

PR #20369, opened 2024-08-20
Borgproofs the robotics control console.

RubberRats:
[SILICONS] [GAME OBJECTS] [BALANCE]
## About the PR 

This PR prevents borgs from disassembling the robotics control (IE killswitch) console by making it flash them when they try to unscrew the panel. Why the flash? It was the only mechanism I could think of that would make sense working against all borgs.

## Why's this needed? 

Working under the assumption that people are supposed to activate the killswitch in case of rogue silicons, it is laughably easy for a borg to remove it before the crew has any reason to suspect a thing. With this change, silicons can't disable the console without a human accomplice, making it less trivial. Most of the time, the robotics control console is either just as fortified as the law rack, or literally in the same room, so this shouldn't make thwarting evil robots _too_ easy. My hope is that this change makes people at least consider killswitching more often when fighting the silicons.

## Changelog 

```changelog
(u)RubberRats
(+)The robotics control console has been fitted with countermeasures to prevent the silicons from tampering with it.
```

### DIFF
--- a/code/obj/machinery/computer/robotics.dm
+++ b/code/obj/machinery/computer/robotics.dm
@@ -1,5 +1,5 @@
 /obj/machinery/computer/robotics
-	name = "robotics control"
+	name = "robotics control console"
 	icon = 'icons/obj/computer.dmi'
 	icon_state = "robotics"
 	req_access = list(access_ai_upload)
@@ -23,9 +23,14 @@
 		STOP_TRACKING
 
 
-/obj/machinery/computer/robotics/attackby(obj/item/I, user)
+/obj/machinery/computer/robotics/attackby(obj/item/I, mob/user)
 	if (perma && isscrewingtool(I))
 		boutput(user, SPAN_ALERT("The screws are all weird safety-bit types! You can't turn them!"))
+		return
+	else if (issilicon(user) && isscrewingtool(I))
+		boutput(user, SPAN_ALERT("[src]'s anti-intrusion countermeasures activate and flash you!"))
+		user.apply_flash(30,5,5,5)
+		playsound(src, 'sound/weapons/flash.ogg', 50, TRUE)
 		return
 	..()
 	return
--- a/code/procs/mob_procs.dm
+++ b/code/procs/mob_procs.dm
@@ -257,7 +257,7 @@
 		return 0
 	return 0
 
-/mob/proc/apply_flash(var/animation_duration, var/knockdown, var/stnu, var/misstep, var/eyes_blurry, var/eyes_damage, var/eye_tempblind, var/burn, var/uncloak_prob, var/stamina_damage,var/disorient_time)
+/mob/proc/apply_flash(var/animation_duration, var/knockdown, var/stun, var/misstep, var/eyes_blurry, var/eyes_damage, var/eye_tempblind, var/burn, var/uncloak_prob, var/stamina_damage,var/disorient_time)
 	return
 
 // We've had like 10+ code snippets for a variation of the same thing, now it's just one mob proc (Convair880).


JORJ949:
Im not a huge fan tbh, I like the killswitch being deconstructable and feel like it creates a more enjoyable fight to derogue

Cherman0:
It should probably only do that to silicons that are controllable by that robotics console. So not syndicate borgs but i believe still EMAGged borgs.

DecidedlyUnethical:
Ostensibly this _is_ a good change, as the tactic of unscrewing the console and grinding the board is _unbearably_ cheap, however would it not simply be possible to blacklist a cyborg omnitool from interacting with the console instead? I believe that'd work better than a flash would.

cropseyy:
I'm not sure I like this- there's no real reason to make the console itself flash since no other console does that- however, a portable flasher bolted down in there or some such might work better?

RubberRats:
> Im not a huge fan tbh, I like the killswitch being deconstructable and feel like it creates a more enjoyable fight to derogue

This doesn't 100% prevent borgs from destroying the console, just increases the effort required to do it. There's still several ways for a borg to deny the crew access to the killswitch. Making it something that is more likely to be used in an emergency paves the way for the killswitch process to be made more interesting, if that is needed.

> It should probably only do that to silicons that are controllable by that robotics console. So not syndicate borgs but i believe still EMAGged borgs.

I would not be opposed to this. I initially made it work against all borgs because I was concerned about a single emagged borg breaking the console and making a targeted rogueing much harder to deal with. However, considering they can't be affected by the console anyway, and can't exist without human intervention, it's probably fine.

> I think this _is_ a good change in principle, as the tactic of unscrewing the console and grinding the board is _unbearably_ cheap, however I don't like the idea of it flashing cyborgs who try, mainly because it feels a little bit noobtrap-y. Is it not possible to just blacklist a cyborg omnitool from interacting with the console instead? I believe that'd work better than a flash would.

I initially made it flash any borg that tried because I had intended for it to work against emagged and syndicate borgs, and because I couldn't think anything else that made sense. The flash is 10 seconds, which didn't feel unbearably long, especially considering it should only really happen when the borg is being sneaky before anyone has any reason to be in the computer core. However, if it does get changed to only touch borgs who are tied to the console, I could easily replace it.

RubberRats:
> I'm not sure I like this- there's no real reason to make the console itself flash since no other console does that- however, a portable flasher bolted down in there or some such might work better?

I don't know, I don't think the robotics console having a flash is much weirder than, say, the genetics console dispensing gene injectors (where do they come from?). Portable flashers don't attack borgs by default, so a new subtype that behaves the way I want (and more importantly can't be cheesed as easily as the current console) would need to be created, and then added to maps, which is significantly more overhead for a small feature.

444Portal:
sure why not, stun being 3-4 seconds would be better imo though

MyNameIsRomayne:
I like the idea purely because its more of a 'you really didn't think it'd be that easy, did you'? Kind of change. as long as borgs can still get creative when it comes to destroying the console with chemical explosives and whatnot, I think this is a good change

JORJ949:
Changed my mind I like this, I agree with the shorter stun time 444 suggested and would like it to not apply to broken consoles or syndicate cyborgs (Emagged borgs are still on the killswitch)

TobleroneSwordfish (commented):
Eh, I don't really like it just because I think killswitching is an inherently boring mechanic. The current meta of borgs rushing the console isn't great but I'd rather remove it than make it even harder to deal with.

RubberRats:
> Eh, I don't really like it just because I think killswitching is an inherently boring mechanic. The current meta of borgs rushing the console isn't great but I'd rather remove it than make it even harder to deal with.

I understand where you are coming from, but I don't think outright removal is a very good solution. On maps where the console and law rack are in different rooms, it forces the silicons to defend two areas instead of just fortifying the upload room. I think this is a good thing. Considering how difficult it can be to actually fix rogue borgs, I think this is good.

Even then, removing the killswitch still leaves the module lockdown, and if the problem is being unfun, I don't think having hand privileges revoked indefinitely is much of an improvement. 

As a compromise, what if the killswitch could be interrupted? Then, whoever activated it would have to defend the computer core from all the silicons converging on them at once to try and stop it. It forces a confrontation and makes the entire interaction more complex than "captain sneezes on the console and defuses everything".

JORJ949:
> As a compromise, what if the killswitch could be interrupted?

I like this idea and a little fun fact: You dont need any access to CANCEL a killswitch, so anyone can just walk in and disable it (except the borgs). Also the wiki states the AI can remove a borg's killswitch but I have never seen this done.

TobleroneSwordfish:
Making borgs able to interrupt it might be cool, yeah!

PR #20370, opened 2024-08-20
Give Classic SecTechs the space law book and increase available amount to 3

JORJ949:
## About the PR 
Removes the ifdef RP_MODE from the space law book in the sectech
Increases space law book amount available from sectechs from 1 ->3

## Why's this needed? 
I see no reason classic shouldnt have access to this outside of security assistants and the occasional random spawn
Only one available per sectech feels low, 3 feels more reasonable

## Changelog 

```changelog
(u)JORJ949
(+)Classic SecTechs now get Space Law. Also increased available quantity from 1->3
```


JORJ949:
At least give me time to label it >:(

PR #20372, opened 2024-08-20
[TGUI REACT] Reimplement functionality for TGUI console history

MyNameIsRomayne:
## About the PR 

writing commands adds them to the history, and sets the current command to the front-most entry.
restarting the console clears the history.
up and down arrow keys choose previous/next entries in the console.
choosing beyond the last command executed (trying command executed after than that) changes it back to empty

note that this has a difference in behavior compared to current because in this implementation, commands are stored as strings server-side. this means that closing/re-opening the terminal will show the last saved command in the history (if exists)

## Why's this needed? 

broke in the port/upgrade, should really be fixed.

### DIFF
--- a/code/modules/networks/computer3/computer3.dm
+++ b/code/modules/networks/computer3/computer3.dm
@@ -1,4 +1,5 @@
 
+#define MAX_INPUT_HISTORY_LENGTH 100 //! Maximum amount of things some nerd can put in here until we've had enough
 
 /obj/machinery/computer3
 	name = "computer"
@@ -35,6 +36,11 @@
 	/// does it have a glow in the dark screen? see computer_screens.dmi
 	var/glow_in_dark_screen = TRUE
 	var/image/screen_image
+
+	// Vars for command history
+	var/list/list/tgui_input_history //! (Keyed by CKEY) A list of strings representing the terminal's command execution history. New history is appended as commands are executed
+	var/list/tgui_input_index //! (Keyed by CKEY)  An index pointing to the position in tgui_input_history to update tgui_last_accessed with
+	var/list/tgui_last_accessed //! (Keyed by CKEY)  The most recently accessed command from the console
 
 	power_usage = 250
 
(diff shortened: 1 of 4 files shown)


MyNameIsRomayne:
woops

MyNameIsRomayne:
alright should be good now. mb

MyNameIsRomayne:
> * tgui_input_history should probably be per player, feels weird to access someone else's history
> 
>     * I wonder if all hacks could be avoided by using react's useState/useEffect which seemed to work way better than what was available under inferno. It's what was used for the chat and the [chemicompiler](https://github.com/goonstation/goonstation/pull/20284/files)

Fair enough, just was what I went with since it was the easiest approach. I could probably just paste back the logic under the terminalinput component and call it a day

Garash2k:
I don't think there's anything wrong with it being server side, it could just be `tgui_input_history[ckey]`

MyNameIsRomayne:
Ah ok. I feel like it would make more sense to have it be by the logged-in user if anything but ckey specific works too.  I wont be able to get to that today but ill find time to hopefully tmr

MyNameIsRomayne:
This is per-ckey now, I'll have to wait for comment about what exactly is meant by avoiding the code hacks,

Garash2k:
> This is per-ckey now, I'll have to wait for comment about what exactly is meant by avoiding the code hacks

Ohhh fair, let me delve into it myself

Garash2k:
Ok so the hack to be avoided was getting the DOMInput to read and set it's value. [This solution](https://github.com/Garash2k/goonstation/commit/afab2da654733ecd7a7bff490c417fc5e5790726) *should work* but it doesn't and I'm not quite sure why. Maybe because of the useEffect() in tgui's Input.tsx. I couldn't find a workaround.

I've made a PR, that at least doesn't *read* from the DOMInput and is one method deletion (`setDOMInputValue(...)`) away from not being a hack, https://github.com/MyNameIsRomayne/goonstation/pull/3

MyNameIsRomayne:
Yeah a few weird things there, thank you for going out of your way to PR an improvement to that!! :heart:

Laboredih123:
wahooo

PR #20373, opened 2024-08-20
Give borgs synthetic brains

Cherman0:
## About the PR 

Makes roundstart borgs spawn with the metallic AI brain subtype instead of a human brain.

## Why's this needed? 

Has essentially no mechanical effect, but should help to reduce the notion of borgs being a human consciousness/soul in a metal body. This often causes uncomfortable discussion which is not really fit for the tone of the server. Also edits the description of the AI brain to make it no longer state that it is an actual human brain, instead states that it is a machine modeled after the human brain. It was already made of pharosium metal and bled oil, so this does not conflict with the existing theming.

And as a matter of code quality the borg brain was not an actual subtype of brains. It was actually a plain human brain that just has its icon state edited when a cyborg spawns with one.

### DIFF
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -233,9 +233,8 @@
 
 		SPAWN(1.5 SECONDS)
 			if (!src.part_head.brain && src.key && !(src.dependent || src.shell || src.part_head.ai_interface))
-				var/obj/item/organ/brain/B = new /obj/item/organ/brain(src)
+				var/obj/item/organ/brain/B = new /obj/item/organ/brain/ai(src)
 				B.owner = src.mind
-				B.icon_state = "borg_brain"
 				if (!B.owner) //Oh no, they have no mind!
 					logTheThing(LOG_DEBUG, null, "<b>Mind</b> Cyborg spawn forced to create new mind for key \[[src.key ? src.key : "INVALID KEY"]]")
 					stack_trace("[identify_object(src)] was created without a mind, somehow. Mind force-created for key \[[src.key ? src.key : "INVALID KEY"]]. That's bad.")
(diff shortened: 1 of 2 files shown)


ZeWaka (approved):

cropseyy:
Honestly cosmetically the AI brains look really cool. Lore reasons aside it also prevents some of the confusion I've seen where trying to "humanise" borgs hasn't worked.

TobleroneSwordfish:
It does work though, doesn't it?

cropseyy:
> It does work though, doesn't it?

Off the top of my head, yes, but I've definitely seen confusion over monkey cloning a borg brain.

Cherman0:
Iirc it doesnt if the borg has puritan on their profile, even though they dont generate a medical record (and thus there is no way to know). Though also irrelevant to this PR

444Portal:
Doesn't this defeat the point of them being Cyborgs?

Tarmunora:
Can't say I'm a fan of this

klushy225:
I kind of like the human brain in the cyborgs. Explains why the cyborgs can't tell you the eight root if PI in a fraction of a second, and allows for more avenues of character development for RP players.

Glamurio:
I just think the sprite is cool and makes sense for borgs. Heck, might as well just convert any human brain into a synthetic brain via icon state once it activates just for theming. I don't much care for detaching borgs from humans lore-wise, but it does make it easy to understand if a brain was inside a human or inside of a borg.

Tarmunora:
Closing this. I don't think that wholly divorcing silicons from humans lore-wise is a good direction to take

ju45he:
Might be a bit late for me to say this but I believe there is a merchant in the debris/mining Z-level which you can sell the AI brain to. I'm pretty sure it's for like 25000 credits or something really large like that so this would have some niche balance concerns.

PR #20374, opened 2024-08-20
Allow constructing/deconstructing/editing door control buttons

MyNameIsRomayne:
## About the PR 

Makes door control buttons scannable, and allows adjusting their ID within reason. Adds a global list which is generated only when needed which contains all the pre-existing door button IDs. Forbids setting a new button ID if its ID is on this list, and forbids changing the ID of a pre-existing button (both to prevent abuse). Buttons with changed IDs can still go back to their original state.

Experimentally, an access-pro with sufficient access can remove the tamper locks on existing buttons.

What you *can* do:

- Toggle tamper lock (requires access_change_ids with an access pro)
- Change button ID (requires tamper lock off)
- Change button control mode (bolts/opening doors) (requires tamper lock off)
- Scan buttons, print/build new ones (tamper lock off by default)
- Deconstruct existing buttons (requires tamper lock off)

image

image

(does not retain ID)

image

image

image

(Removes tamper lock, allows changing ID/deconstructing. Doesn't allow changing ID to restricted values still)

image

What you *can't* do:

Basically *anything* with the tamper lock off

image

image

image

## Why's this needed? 

More player choice! Having door control buttons as a new thing to be able to play around with ought to be a good thing. Allowing players to deconstruct more things in rooms is equally nice.

## Changelog 

```changelog
(u)Romayne
(+)You can now construct/deconstruct door control buttons. Existing ones have tamper locks enabled which will need to be disabled to change them.
```


Cherman0:
I'm reasonably sure that this would let you turn off the belts at NTRC, amongst other things. The ID field is too widely used to give players general access to sending signals on it, imo. It would probably be better to make buttons emit a radio packet that the items they control can listen for, such that the system isnt magic transmission and that specific items that are desired to be untamperable can be made to not listen to accessible frequencies.

MyNameIsRomayne:
> I'm reasonably sure that this would let you turn off the belts at NTRC, amongst other things. The ID field is too widely used to give players general access to sending signals on it, imo. It would probably be better to make buttons emit a radio packet that the items they control can listen for, such that the system isnt magic transmission and that specific items that are desired to be untamperable can be made to not listen to accessible frequencies.

That seems out of the scope of this PR honestly, if anything this should be resolved with the system as-is by expanding the scope of blacklisted IDs to include anything that could receive the signal.

MyNameIsRomayne:
Alright, you can no longer use this to do stuff like toggle NTRC belts

TobleroneSwordfish:
I think ideally we shouldn't expose yet another disconnected system for linking things together like this. I'd love to make door controls player replicatable, but it should use an existing system (mechcomp, packets, radio signals) rather than making up its own magic thing.

MyNameIsRomayne:
Uhh ok then. Closing this as I am not in the mood to refactor door control systems rn

PR #20388, opened 2024-08-22
Makes the German accent understandable.

klushy225:
[TRAIT][QOL]

## About the PR 
This guts the German accent I made so that one without much experience in German, or any at all, can understand it. I have removed a large amount of replacement words, and toned down how many sounds it replaces.

(note- many of the words listed in the replacement file are spelled wrong on purpose, with how they are run through the parser and such.)

EXAMPLE SENTENCES:

image
## Why's this needed? 
It hurt me to do it, but when people started asking me to translate for them because I had made the accent, I realized that it was far too complex. When I made the accent originally, it was a case of me not realizing how incomprehensible it was, as I was the one who made it. This should fix the issue.

## Changelog 

```changelog
(u)Klushy225
(+)The German accent is now able to be understood.
```


klushy225:
oh shit I forgot the labels again

HyenaShepherd:
Much easier to understand after reading the example sentences. I could see more people using the accent if this goes through!

itchybones:
As a passionate german accent user (and actual German), I'm so glad this got merged. Especially words like "Tank" turning into "Panzer" didn't make much sense.

PR #20418, opened 2024-08-25
Allow HoSes to play halloween critters

JORJ949:
## About the PR 
Changes the "mentor_only" var into "trusted_only" (it was only used for the halloween critters) and allows jobs with this tag to be joined by HoSes and mentors

## Why's this needed? 
Mentors aren't the only trustworthy individuals.

## Changelog 

```changelog
(u)JORJ949
(+)HoSes can now play halloween critters (when its halloween of course).
```

### DIFF
--- a/code/datums/controllers/job_controls.dm
+++ b/code/datums/controllers/job_controls.dm
@@ -99,8 +99,8 @@ var/datum/job_controller/job_controls
 		if (!job.no_jobban_from_this_job && jobban_isbanned(player, job.name))
 			logTheThing(LOG_DEBUG, null, "<b>Jobs:</b> check job eligibility error - [player.ckey] requested [job.name], but is job banned.")
 			return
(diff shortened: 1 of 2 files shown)


klushy225:
Absolutely biased here, so take my opinion for a grain of salt, but sounds like a good change to me.

Retrino:
at this point why keep them on a whitelist, most abusable behaviors from back then (which was on an hos whitelist and eventually got put to mentors only) have been nerfed/removed

ZeWaka (approved):
sounds fine to me

flrsh (approved):
Seems fine to me as well, though I also agree with Retrino that we may not even need this anymore and can just let everyone play a critter.

TobleroneSwordfish:
Well I'll just merge this for now then 🤷

PR #20426, opened 2024-08-25
Headpat as Emote Now

MintyPhresh:
## About the PR
Adds a new emote, *headpat. 

Targeting a human's head with help intent now taps their shoulder, instead. Doing the same with a cyborg will no longer occationally give headpatting as an output.

The stunhat now stuns players both if they do the *headpat emote, and if they try to tap the wearer's shoulder.

## Why's this needed?
Headpats are accidental 90% of the time which makes it hard to tell if they're intentional, devaluing it. Doesn't make much sense for them to be so common in a workplace, either.

## Changelog
```changelog
(u)Mintyphresh
(+)Headpatting humans and cyborgs is now done with the *headpat emote.
```

### DIFF
--- a/code/mob/living/carbon/human/procs/emote.dm
+++ b/code/mob/living/carbon/human/procs/emote.dm
@@ -303,12 +303,12 @@
 					G.throw_at(target,5,1)
 					src.visible_message("<b>[src]</B> farts out a...glowstick?")
 
-			if ("salute","saluteto","bow","hug","wave","waveto","blowkiss","sidehug","fingerguns")
+			if ("salute","saluteto","bow","hug","wave","waveto","blowkiss","sidehug","fingerguns","headpat")
 				// visible targeted emotes
 				if (!src.restrained())
 					var/M = null
 					var/range = 5
-					if (act == "hug" || act == "sidehug")
+					if (act == "hug" || act == "sidehug" || act == "headpat")
 						range = 1
 					if (param)
 						for (var/atom/movable/A in view(range, src))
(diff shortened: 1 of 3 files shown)


Retrino:
dont really agree with the specific implementation for the shoulder thing, headpats (while odd to be common in the workplace) work as a quick action a player can do to another as a comforting gesture. would rather see this added as "pats them on the shoulder" or something, we already have an attention grabbing click gesture from help intent

soleilan:
soul..

klushy225:
I like the headpats. If someone is uncomfortable with headpats, they can always say so, which is perfectly valid(and if people don't stop headpatting them that's arse behaviour).

MintyPhresh:
I don't see the value of headpats as a quick-action rather than as an emote. If you're providing someone some comfort, you're probably not in that big of a rush to move on, and the extra time it takes helps make it more meaningful.

What makes it even less meaningful is that head-targeting is most players' default 50% of the time (ever since the scroll wheel was added as a limb targeting control), so even if you do make use of the fact that it's on combat intents and is super easy to put out, it'll often get lost in the interpretation of "oh, they're just greenhanding me."

I'm not a fan of getting headpats, but the reason why I made this PR is because I'm tired of getting frosty every time someone headpats me because I'm not sure if they did it on purpose, since like 90% of the headpats performed are accidental.

> Headpats are a goofy but also comforming thing that is fitting for the humor of goonstation i would rather see as the standard action instead of shoulder-patting, especially borgs and on other mobs like orwell or man eating plants.

This only affects humans!

RubberRats:
As someone who plays a character that will deck someone for headpatting him, I don't really like this PR. Headpats have their place, and them being quick is a part of that place. If people aren't respecting your personal space in the video game, you can remind them gently, you can remind them less gently, or you can use a bit of physical force. People usually listen to at least one of those. If they won't listen and are making you genuinely uncomfortable, you can (and probably should) ahelp that.

I'm all for shoulder pats though. Maybe make them for targeting an arm on help intent?

Tarmunora:
headpatting will stay

PR #20427, opened 2024-08-26
Lower classic shuttle autocall to 1h

JORJ949:
## About the PR 
Reduces the time on non-RP servers for the shuttle to be automatically called from 90m -> 60m

## Why's this needed? 
Classic rounds rarely reach 60m without a shuttle call and rounds in which they do almost always are due to nobody being able or willing to call in a round that should have had the shuttle called a while ago. The autoshuttle can still be recalled.

## Changelog 

```changelog
(u)JORJ949
(*)Classic shuttle is now autocalled at 60 minutes rather than 90
```

### DIFF
--- a/code/datums/gamemodes/gm_parent.dm
+++ b/code/datums/gamemodes/gm_parent.dm
@@ -9,7 +9,11 @@ ABSTRACT_TYPE(/datum/game_mode)
 
 	var/shuttle_available = 1 // 0: Won't dock. | 1: Normal. | 2: Won't dock if called too early.
 	var/shuttle_available_threshold = 12000 // 20 min. Only works when shuttle_available == SHUTTLE_AVAILABLE_DELAY.
+	#ifdef RP_MODE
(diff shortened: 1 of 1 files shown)


LaCumbiaDelCoronavirus:
Disagree. Although 90 mins does feel too long for an autocall, 60mins feels too early. Id prefer to see this somewhere inbetween that. Really though, autocall should in the first place depend on pop, not a hard timer.

JORJ949:
> Disagree. Although 90 mins does feel too long for an autocall, 60mins feels too early. Id prefer to see this somewhere inbetween that. Really though, autocall should in the first place depend on pop, not a hard timer.

Very few classic rounds reach 70m, and those that do should have probably been called sooner, the only recent rounds that come to mind where this wasnt the case was earlier today when the whole crew drained the inside of oshan and that still was only 70ish minutes long including shuttle. Autocall is not the end of the round, it can always be recalled and as the AI Laws wiki page says "The humans can recall the shuttle within four minutes of you calling it. If they can't, it was probably time to call it."

Though despite all that I would be open to upping it to maybe 70

I would also like to expand upon the current "Crew shortages and fatality" mechanic by making it a % of players dead rather than just less than 5 alive.

JORJ949:
Closing in favour of improving the "Crew shortages and fatalities" system

PR #20429, opened 2024-08-26
Rework Atlas Security

JORJ949:
## About the PR 
Reworks atlas security to make it feel more spacious by moving the maints 2 tiles northward (bee-shape is maintained)
Turns the HoP's vault into the HoS' office
image
image

## Why's this needed? 
Current atlas security feels very cramped for somewhere you are supposed to bring people into and search, which is hard to do when anyone trying to get anywhere else in sec has to push you aside. Armory is also lacking a turret and the weapons vendor is a pain to get to if multiple officers try redeem at once. HoS "office" is 2 tiles next to 3 potential antags that can just swipe stuff from your locker for free if you aren't careful.

## Changelog 

```changelog
(u)JORJ949
(*)Atlas security has been remodelled. The HoP's vault has been invaded by the HoS.
```


trazodont:
oh i like this a lot

Gannets:
Looks pretty good to me

JORJ949:
image
Here is before without the HoS being in bridge, happy to revert to this state if it is preferred.

TobleroneSwordfish:
Contraband lockers ideally shouldn't be in sight of the front room, see #20043. Maybe stick it on the back wall with the kitchen so the chef can bust through the wall to steal it.
Also in theory it'd be nice to have the interrogation room closer to the cells but that might be kind of tricky to slot in here so 🤷

JORJ949:
> Also in theory it'd be nice to have the interrogation room closer to the cells but that might be kind of tricky to slot in here so 🤷

I tried to fit the interro below first but there just wasn't the space for it, its closer to the brig than oshan's interro at least

BatElite:
Not a fan as it is. If anything this version looks to be more cramped rooms than the previous, and most of the space gained appears to be going to either the armoury or to interior walls. Plus, there's no disposal chutes or other method of escape for ne'er-do-wells to make use of. If someone ends up in either of the larger rooms on their own, they're bash-down-two-doors effed. They just don't seem like spaces that lend themselves to improvisation.

trazodont:
could there be a coffee machine pretty please <3

Sord213 (approved):
some minor stuff i'd like to shift around at some point after merging but otherwise lgtm, definitely an improvement

JORJ949:
>:(

JORJ949:
All is good again

JORJ949:
Requesting review removal for Kyle and Frawhst, they were requested for review as code-owners due to a fucked up merge.

PR #20430, opened 2024-08-26
Shrimply gives mutantraces hair

klushy225:
[MUTANTRACE][QOL]
## About the PR 
Simply allows mutantraces to have hair. No trait, no jank workaround, just hair. Adds the HAS_HUMAN_HAIR flag to most mutantraces(all the normally playable ones, and excluding the rarer ones that only come up through artifact or admin shennanigans)
This is mutually exclusive with #20355

edit: As with the last PR, this should not impede the progress of the upcoming mutantrace rework in any way shape or form. This simply adds a flag that is already defined and used by the human mutantrace(and other niche examples) to the rest of the mutantraces.

## Why's this needed? 
A common opinion expressed in the discussion around my other hair-related PR, #20355 was that this should be divorced from traits as a whole, and mutantraces should simply be given hair. As stated by my comment on my first PR, 

"I understand that the tradeoff(with wigs and mutantraces) was intended, but I don't really agree with it. it's possible to wear hats and wigs at the same time in real life, and while I am fully aware that ss13 does not nor should not mirror real life, I don't really know why mutantrace players should be excluded from hats."

A multitude of players has expressed frustration at the seemingly arbitrary barring of hats from mutantrace players, claiming that it would be encroaching on the "human thing," and claiming that nobody would play humans anymore. I think this makes no sense, as nobody is forcing anyone to play a mutantrace. It's about personal choice and customization, that's it.

I chose to create two separate PRs, as to provide an alternate solution for those who would prefer that mutantrace customization still has tradeoffs.

## Changelog 

```changelog
(u)Klushy225
(*)Most mutantraces can now have hair without the need for a special trait, and without taking up a head slot.
```


TheTyrant25 (approved):

Glamurio (approved):

TealSeer (approved):

itsloley:
this is what i've frankly been hoping for the entire time, so i'm super happy to see this solution brought up! letting people just express themselves will always beat silly tradeoffs imo, very much hoping this gets merged :)

klushy225:
> Does it work with (to name a few) polymorph, changelings, the barber shop, hairgrownium?
> 
> Also, I have hair configured for mutantraces in case they get turned human mid-round, but I don't necessarily want hair on the mutants. Can I still do that?

it works exactly the same as human hair. I added lab-grown wigs in the changelog to explain the hairgrownium.

ZeWaka:
> Should be noted the last PR was tagged as controversial. 

tagging it with controversial shrimply because the last one was also

ZeWaka:
as for my thoughts as a Developer but not speaking for the team, i think any solution (that is not horribly jank) the community generally agrees on is fine...
As long as its generally understood we eventually want to revisit this when mutantraces are reworked along with the character setup menu to allow for cooler and more unique mutantrace stuff instead of just being humans~~ for the most part

ZeWaka (commented):
i do think that the changelog as worded currently is a bit confusing with the wig reference though

klushy225:
> i do think that the changelog as worded currently is a bit confusing with the wig reference though

I'll reword it

444Portal:
I think we should probably ditch the in universe changelog entry thingy and just make it like "Mutantraces can now have hair without the Bald Trait" or something

klushy225:
> I think we should probably ditch the in universe changelog entry thingy and just make it like "Mutantraces can now have hair without the Bald Trait" or something

Done!

itsloley:
> Another point I must make, is that hair will make it notably harder for mutants to impersonate other mutant characters in-game.
> In addition to stealing/copying clothes and ID to have a facsimile of their profile/appearance, now you potentially have to mimic a highly distinctive hairstyle as well.

i feel like this is kind of a non-issue, since copying someone's clothes and ID is already a lot, and i never even see people go to the extent of trying to mimic every aspect of a person like that (which, at that point, you'd might as well use stable mutagen). not to mention, that problem already exists for regular humans, and mutantraces currently aren't able to mimic eachother perfectly anyways. if you try to mimic another mutantrace player who has hair from a wig, you're basically forced to either steal their wig, or be bald.

there might be something i'm missing, but i'd argue to say that if anything i think this'd make it even more accessible to mimic the appearance a mutantrace player? between this being so rare & the problem already existing for humans & current mutantraces anyways, i'd say it's probably safe to not worry about

CoolKakapo:
If I were to play a mutantrace character but I don't want hair while I'm that mutantrace, only when I am a human? Maybe just a toggle for this would be nice if I for instance want to play as a bald roach, but a human with hair if I turn into one mid-round.

Glamurio:
> If I were to play a mutantrace character but I don't want hair while I'm that mutantrace, only when I am a human? Maybe just a toggle for this would be nice if I for instance want to play as a bald roach, but a human with hair if I turn into one mid-round.

This seems so highly specific and on the level of "when I turn into a monkey I'd like a top hat but only when I'm a monkey". If you want to be a bald roach, you can ask genetics to remove your hair, and ask genetics to give your hair back. And it's most likely genetics that handled your transformation in the first place, this seems very intuitive already.

klushy225:
You can always shave your head if you don't want to be bald. If you play a human on randomized appearance, chances are you won't be bald either. Later on, a trait can be added/modified to ensure that the player starts with absolutely zero hair, but that is once again, kind of niche and out of the scope of this PR.  Also not to mention a little jank.

RubberRats:
> > If I were to play a mutantrace character but I don't want hair while I'm that mutantrace, only when I am a human? Maybe just a toggle for this would be nice if I for instance want to play as a bald roach, but a human with hair if I turn into one mid-round.
> 
> This seems so highly specific and on the level of "when I turn into a monkey I'd like a top hat but only when I'm a monkey". If you want to be a bald roach, you can ask genetics to remove your hair, and ask genetics to give your hair back. And it's most likely genetics that handled your transformation in the first place, this seems very intuitive already.

I don't think that's a good (or fair) comparison. Changing your hairstyle is significantly more involved than getting a new hat, especially considering that you can't really do it on your own. This is compounded by the fact that some hairstyles (like facial hair) just look really ugly on mutantraces. I think it's extremely reasonable for someone to not want their roach to have hair, but also to not want to be bald when they get humaned. "Just get some hairgrownium and cut it" isn't an answer.

More generally, I don't think mutantrace hair should be the default. Lots of styles weren't made with them in mind, and I think it will just make them less visually interesting and less distinct from humans overall. Roaches and cows have great head details, and making it a conscious effort to actually see them doesn't sit right with me.

I understand that this is meant to be an easy fix to the admittedly very limited mutantrace customization options. My concern is that this will make it harder to eventually give them a bespoke system for this down the line. I personally don't really care much about lizard hair, but I know a lot of people would fight tooth and nail for this no matter how cool the thing it gets replaced by is. In my ideal wigs would be a secondary option to unique mutantrace styles, and this PR would very firmly entrench what I personally think is a stopgap solution.

klushy225:
> More generally, I don't think mutantrace hair should be the default. Lots of styles weren't made with them in mind, and I think it will just make them less visually interesting and less distinct from humans overall. Roaches and cows have great head details, and making it a conscious effort to actually see them doesn't sit right with me.
> 
I agree with this point, and I bet it would be rather simple to disable random hair for mutantraces. Not super sure if it messes up the process or whatever, adding commits to a PR already out for final review, however. Then again, for now, it's not too hard to just run over to the barber shop and cut all that hair off if needed

TobleroneSwordfish:
No idea if this will end up getting merged but a couple of points to consider:
- Skeletons should probably still have to have wigs because... cmon, it's a skeleton.
- Pugs were specifically designed not to have hair iirc (or rather to have their hair built in, because 🐶) and can't currently get wig hair using the bald trick.

klushy225:
> > > More generally, I don't think mutantrace hair should be the default. Lots of styles weren't made with them in mind, and I think it will just make them less visually interesting and less distinct from humans overall. Roaches and cows have great head details, and making it a conscious effort to actually see them doesn't sit right with me.
> > 
> > 
> > I agree with this point, and I bet it would be rather simple to disable random hair for mutantraces. Not super sure if it messes up the process or whatever, adding commits to a PR already out for final review, however. For now, it's not too hard to just run over to the barber shop and cut all that hair off if needed
> 
> Please do disable random hair. As it exists now, this PR has what I'd call "collateral damage".

on it.

klushy225:
the merge conflict will be fixed when my git stops imploding

Xkeeper0:
i would personally love to see this merged

klushy225:
okay I had to learn how to use ssh but it SHOULDN'T have a merge conflict anymore. Keyword here shouldn't. I think everything is done now- randomized appearance is fixed, merge conflicts fixed, and uh, the poor skeletons don't get hair.

klushy225:
okay. NOW it should be all ready

444Portal (approved):

Glamurio (approved):

Xkeeper0 (approved):
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

TobleroneSwordfish:
Merged in a very slightly different form here: #20564

PR #20443, opened 2024-08-28
Standardize AI upload defenses

TobleroneSwordfish:
[BALANCE] [MAPPING]
## About the PR 
Brings all uploads on maps in current rotation up to the same standard of requiring More Than One Thing to block turret shots.

## Why's this needed? 
Currently how difficult it is to rogue/unrogue the AI varies pretty randomly depending on the map, atlas has 4 turrets for example and cog1 has 2. A little variation is cool, but a shockingly large number of maps have turrets that can be defeated with a single security barrier or just well timed walking.

## Changelog 

```changelog
(u)LeahTheTech
(*)Standardized AI upload turret placements. Generally this means more turrets on maps that lacked them.
```

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -18167,10 +18167,12 @@
 /turf/simulated/floor,
 /area/station/science/lab)
 "fCY" = (
-/obj/machinery/turret,
 /obj/machinery/alarm{
 	pixel_y = 23
 	},
+/obj/table/reinforced/auto,
+/obj/random_item_spawner/ai_experimental,
+/obj/item/device/infra_sensor,
 /turf/simulated/floor/circuit,
 /area/station/turret_protected/ai_upload)
 "fDe" = (
@@ -18387,12 +18389,11 @@
 /turf/simulated/floor/plating,
 /area/station/storage/tech)
 "fJV" = (
-/obj/table/reinforced/auto,
-/obj/item/aiModule/notHuman,
 /obj/machinery/light{
 	dir = 4;
 	pixel_x = 10
 	},
+/obj/machinery/turret,
 /turf/simulated/floor/circuit,
 /area/station/turret_protected/ai_upload)
 "fKc" = (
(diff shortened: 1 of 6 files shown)


JORJ949:
Not a huge fan of the 3-turret setups on clarion, kondaru and oshan (but I really dont know how you would fix oshan upload without a complete rework of it), corner turrets make the most sense to me but I also dont like how donut 2 corners them with tables, 2 turrets in corners with rows of tables between them like the top of clarion I think would work best, especially on kondaru which the black tiles are supposed to be walkable on.

JORJ949:
Donut2 turrets would like nicer a tile lower pushed into the corner and kondaru's black tiles were meant to be the "walkable" bit so the table should have matrices under them

TobleroneSwordfish:
Sadly the donut 2 turrets can't see any of the entrance when they're tucked in the corner like that.

Glamurio:
Not sure how to feel about this. I never quite liked how the solution to laser turrets shooting at you was to **build an entire wall** in front of them. It feels jank and more like a band-aid approach. If we could hack the panel or turn the turrets off another way, even if it's more involved, I'd much prefer that when increasing the amount of turrets. Because as this stands, it'll be rare to see anyone rogue the AI now that doesn't have to ability to turn them off or isn't prepared to powergame as Bob The Builder.

Vincent983:
> > Not sure how to feel about this. I never quite liked how the solution to laser turrets shooting at you was to **build an entire wall** in front of them. It feels jank and more like a band-aid approach. If we could hack the panel or turn the turrets off another way, even if it's more involved, I'd much prefer that when increasing the amount of turrets. Because as this stands, it'll be rare to see anyone rogue the AI now that doesn't have to ability to turn them off or isn't prepared to powergame as Bob The Builder.
> 
> Potentially give them packet interaction? Either wired for the turrets themselves (just dont connect armory turrets to the network) or wireless for the control terminal (but blacklist armory turrets)

The funny thing is, they already have packet interaction, they just don't actually have data terminals under them.

JORJ949:
> 
> The funny thing is, they already have packet interaction, they just don't actually have data terminals under them.

and in addition its just the /obj/machinery/turret/network subtype that do this, so you just dont make the armory turrets those.

TobleroneSwordfish:
Packet interactions are not a good way to make anything more intuitive 🙈 Blocking the turrets is always going to be the easiest way unless we give them the ability to shoot through walls.

Tarmunora (approved):

PR #20444, opened 2024-08-28
Remove obj spawn from plating/random

JORJ949:
## About the PR 
Removes the random object spawners from plating/random, leaving plating/random with dirty/burnt/broken turf placement

## Why's this needed? 
Turfs shouldnt randomly sometimes spawn tools and stuff on themselves, causes weird stuff like this
image


TheTyrant25:
that's what plating/random is for, though? just replace the under-window turfs with regular plating

Gannets:
Yeah, this feels like a mapping error not an issue with plating/random

Sovexe:
I enjoy the randomly spawning junk and have made use of it in my adventure zones. The bug here is that random plating shouldnt be used under windows

PR #20448, opened 2024-08-28
Revert Security Officer Priority

JORJ949:
## About the PR 
Reverts Security Officers being a priority job

## Why's this needed? 
I feel this change has had the complete opposite effect than what was intended with #18329, now secoff is "set it unwanted or get it every round" which results in most people having it in unwanted. People want job variety and having a job ALWAYS be yours if not unwanted just results in everyone setting it unwanted and there being no sec, if there IS sec its people that fully knew and expected to be playing sec this round, rather than having a chance at it. The only people this does effect is new players who get forced to play sec and don't understand why.

I agree sec should be more common but it shouldn't be forced like this. A better solution would be weighting the odds of rolling a job depending on how many there already are (E.G. 5 people are going to either be engineer or doc, try to balance them a bit rather than just making 5 engineers)

## Changelog 

```changelog
(u)JORJ949
(*)Security Officer is no longer a priority role.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -506,9 +506,6 @@ ABSTRACT_TYPE(/datum/job/security)
 	limit = 5
 	lower_limit = 3
 	variable_limit = TRUE
-	high_priority_job = TRUE
-	high_priority_limit = 2 //always try to make sure there's at least a couple of secoffs
-	order_priority = 2 //fill secoffs after captain and AI
 	wages = PAY_TRADESMAN
 	trait_list = list("training_security")
 	access_string = "Security Officer"


JORJ949:
Requesting your review because it was your change.

TobleroneSwordfish:
I disagree that the change has resulted in less security officers overall, there has been a noticeable decrease in zero/solo security shifts since it was given priority. I do understand the frustration of rolling the same job over and over simply because no-one else wants to play it but I think if we have *less than two* players per shift that are actually willing to play security then there are fundamental issues with the role that we should be considering.

You mention that a better solution might be redesigning job assignment code to more evenly distribute players between departments. This might be a good thing overall (although on lower population shifts it is likely to lead to some very spread out feeling crew distributions), but would essentially amount to just giving security a high priority of 1 instead of 2, leading to people still being dragged from low priority to play it but *also* not having a partner as often.

JORJ949:
> I disagree that the change has resulted in less security officers overall.

Could just be me noticing more how little sec there is but I feel like it has resulted in less GOOD sec, as experienced players put it unwanted as to not get it and newer players have it low

TobleroneSwordfish:
No further feedback, so I'm closing this for now.

PR #15634, opened 2023-09-01
Chemical production rework

TobleroneSwordfish:
[CHEMISTRY] [BALANCE] [INPUT WANTED]
### General Changes:

- Chemicals are now mixed at the temperature of precursors. (Hot precursors will make a hot product chem!)
    - Some chemicals are endothermic or exothermic and add/remove heat on reaction.

- Beakers, extinguishers and barrels will now shatter open completely when an explosive reaction happens inside them, spraying out shards of glass/metal that can embed.
image
    - Watch out!

- Reactions not inside inventories or puddles will display little animations on their container. Beakers will visually bubble, puff out flames, smoke, and more!

### Chemistry Equipment
- Beaker lids
image
    - One-size fits all lids that go on any type of beaker, turning it into a closed container. Useful for some reactions.

- Condensers
image
    - Useful pieces of glassware that can contain reagents themselves.
    - Can be attached to beakers, barrels and even other condensers by clickdragging. Both the condenser and container must be within one tile range and not inside anything or any inventory to be connected, and are disconnected by being picked up or moved.
    - Any reagent mixed inside the condenser will be added to attached container. If the container is full, the product instead backflows into the condenser.
    - Allows you to seperate product from reactants without ending the reaction, or automatically mixing products into other containers.

- Barrels
image
    - Several quality of life adjustments for mixing bulk chemicals:
        - Barrels now have a glass window on the side, letting you peek in and see how Much Chemical you have stored in it.
        - On the other side, there's a funnel/spout. Clicking the barrel with an empty hand will toggle between spout/funnel mode, changing the behavior when you click with a container. Funnel = Pour into barrel. Spout = Pour from barrel.
        - Barrel lids are much more visible when off/on, since open/closed containers are a more important mechanic. 
        - Barrels can now be click-drag attached to ChemMaster 3000s to connect them and allow the ChemMaster to make products from the barrel's contents.

### Recipes:

- Water Boiling
    - Water now boils gradually if under 200C (473.15K), removing more water per cycle the hotter it is.
    - If in open container: Vents steam out as steam clouds, removes heat proportional to water boiled. (2 Degrees K per unit, minimum 1)
    - If in closed container: Creates steam, which is now a unique reagent from water.
    - Steam condenses into water under 100C (373.15K), speed 10u. Steam also escapes open containers into the air.

- Acetone And Phenol
    - Recipe is now Oil(1) + Fuel(3) + Chlorine(.1) = Acetone(2) + Phenol(2)
[...]

PR #20449, opened 2024-08-28
Make non-instant chemical reactions scale with volume of their reactants

M-Earthfire:
[chemistry][feature][balance]

## About the PR 

This PR introduces 3 procs to `/datum/chemical_reaction`:

- `proc/get_total_reaction_volume` returns the total volume of chemicals that are required for a reaction in a given holder 
-  `proc/get_volume_reaction_speed_factor` uses the former proc to return a factor based on this volume and the reactions `var/reaction_volume_for_double`-variable. At standard, this will return 1 for 100u of react-able chems and 2 for 200 and so on.
- `proc/get_reaction_speed_multiplicator` is called as a multiplactor for a reaction's `reaction_speed`. This can be modified to take the scaling of a reaction with temperature into account or through catalysts and such. This also handles the scaling of reaction speed with reaction volume

These procs were used to modify most non-instant reactions. Now, most non-instant reactions react with a higher reaction rate if they have more of their corresponding chemicals in them. Thus, a reaction with 100u of reacting chems in it reacts at double the speed, one with 200u of chems at triple the speed.

The reaction_speed of a reaction now gives the minimum reaction speed a reaction will reach.

This scaling won't be applied if `var/reaction_volume_dependant` is set to FALSE.

The following reactions don't have a volume-scaling component to them:
- synthflesh pustules (it's a fixed time independing of volume for total reaction)
- salicylic acid/salbutamol (i really wasn't able to see through how to properly change it)
- pyrosium + magnesium (i found it more interesting without it, sue me)
- cryostylane + iodine (see above)
- Photophosphide boom (it's a fixed time independing of volume for total reaction)
- styptic powder (it's a fixed time independing of volume for total reaction)

Most non-instant reactions were changed to use these procs and use the `created_volume` variable in the `on_reaction`-proc, whenever possible.

## Why's this needed? 

With the introduction of many non-instant reactions in #15634, the go-to strategy of chem precursor production unfortunally became to spam multiple beakers. The reason for this are non-volume scaling chemical reactions, which made reactions in barrels sluggish and extremly time consuming. To combat this, multiple ideas, like #15718, were introduced, but sadly none stuck (i would still love that chemical mixer idea).

This change is probably the most straightforward way to combat this.

This PR also introduces some tech to play with reaction speeds. Because it scales of the total volume of all reacting chemicals in a reaction without caring for ratios, you can use one inexpensive chem in excess to drastically increase reaction speed while preserving your more rare precusors. Almost like in real life...

The only problem i see with this is the increase calculation need for non-instant reactions. This could have an impact on performance.

Because reaction_speed is chosen as the lowest value for reaction speed, this is a straight up buff to many reaction speeds. I highly advice PR's following this to slightly nerf base-reaction speed values across the board. I could also see adjusting the scaling to make the reaction speed double at 200u instead of 100u.

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Most non-instant chemical reactions react faster the higher their corresponding volume of reacting chemicals is. Barrel-chemists rejoice!
```


TobleroneSwordfish:
Sorry this took so long to get to. It shouldn't have done, it's a good PR!

PR #20453, opened 2024-08-28
Tiny-sized items' names are concealed when put in storage containers

kbsmilk:
[GAME OBJECTS] [BALANCE]
## About the PR 
Instead of seeing **"Antagonist has added 9mm magazine to the backpack!"**, nearby players will now see **"Antagonist has added something to the backpack!"**.

## Why's this needed? 
Feels like your character should make an attempt to hide that stolen ID while putting it away.

## Changelog 

```changelog
(u)Smilg
(*)Tiny-sized items are now partially concealed from onlookers when placed into bags or other storage.
```

### DIFF
--- a/code/modules/storage/storage.dm
+++ b/code/modules/storage/storage.dm
@@ -413,7 +413,7 @@
 	if (visible)
 		animate_storage_rustle(src.linked_item)
 		if (!src.sneaky && !istype(I, /obj/item/gun/energy/crossbow))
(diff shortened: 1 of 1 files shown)


TheRyeGuyWhoWillNowDie:
> By the way I have not tested this change, because I dont know the debug commands to simulate it, sowwy </3
The low tech solution is connect to the local server, log off in BYOND, and then join again as a guest

PR #20467, opened 2024-08-29
Fixes observers screaming near wraith exploder death

### LINKED ISSUE #19884: Plague Wraith's Rot Hulk dying and exploding makes ghosts (and itself) scream
### Labels

[BUG][TRIVIAL]

### Description

microscopic issue but whatever. me send bug reporte

### Steps to reproduce

1. Plaguebringer Wraith summons Rot Hulk
2. alive player slays the Rot Hulk
3. it explodes, which makes all nearby players scream, but ghosts near it also scream.
4. also looking at the screencap i took, the Rot Hulk itself screams too. dunno if thats supposed to happen.

### Expected Behavior

The dead to have less anxiety about shit that isnt their problem

https://i.imgur.com/BC6PUmA.png

Reported by: snoid
Client version: 515.1641
Active test merges: #19063
Round log date: 2024-07-07-20-18
Reported on: 2024-07-07 21:20:00
Round ID: 182878
Map: KONDARU

TheGeneralJay:
## About the PR 
Ensures observers do not scream like living mobs do as a result of the wraith exploders dying nearby.

## Why's this needed? 
Fixes #19884


ZeWaka (commented):
This should be is intangible or something

TheGeneralJay:
> This should be is intangible or something

I've tried implementation with isintangible, but was still getting ghost scream reactions visible in deadchat from an observer POV. On the other hand, this did work instead with is_dead_or_ghost_role, which I've swapped in if that implementation is better. I am still figuring this all out, apologies if there's an easier/better way to handle it!

code/modules/antagonists/wraith/critters/exploder.dm
```diff
@@ -69,7 +69,8 @@
 					continue
 				else
 					boutput(M, SPAN_ALERT("You are sprayed with disgusting rotting flesh! You're pretty sure some of it got in your mouth."))
-			M.emote("scream")
+			if (!is_dead_or_ghost_role(M))
```
TobleroneSwordfish, line 72:
```suggestion
			if (isalive(M) && !isingtangible(M))
```
This makes more sense, ghosts are dead but not intangible and blob overminds/AI eyes etc are alive but intangible. `is_dead_or_ghost_role` includes ghostdrones, critter respawns etc and is mostly used for respawning code, not determining if something is real and exists nearby.

TobleroneSwordfish (changes requested):

code/modules/antagonists/wraith/critters/exploder.dm
TheGeneralJay: This makes more sense! I appreciate the explanation, it helps a ton. Thank you so much!

ZeWaka (approved):

PR #20477, opened 2024-08-29
Resprites the icemoon's lakes

TemThrush:
[Sprites] [Mapping]

## About the PR 
This makes a new turf for the icemoon's lakes and replaces the pool floor that was previously used. The turf uses edge overlays for the banks. I also modified one of the lake shapes to go a little better with the overlays. 👍
image
image
image
## Why's this needed? 
The current sprite is very out of place and doesn't really convey an underground lake


trazodont:
Can I find Mewtwo here

ZeWaka (approved):

Gannets (approved):

PR #20481, opened 2024-08-30
Pipes now use the modular pipe system.

Laboredih123:
[ATMOSPHERICS][FEATURE][BUG]
## About the PR 
Pipes now use a variant of the modular pipe end system to control their ends. This both removes the bug of the pipe automatically changing its direction when an end is removed, which breaks bent pipes, and allows the case of rupture and missing pipe ends to be visible. Also makes pipe init more consistent.
Also removes some useless debug things that will never be used.
image

## Why's this needed? 
Clean good, bug bad.

## Changelog 

```changelog
(u)cringe
(+)Pipes now use a variant of modular pipes. This means that when a pipe is ruptured or merely missing an end, they appear visually distinct. Ruptured pipes are broken in the center, while missing ends are broken at the ends. These states can also appear simultaneously.
```


TobleroneSwordfish (changes requested):
That changelog doesn't make any sense.

Laboredih123:
im not really sure how to describe this pr but ill try again.

Laboredih123:
> That changelog doesn't make any sense.

tried again, hows it look?

TobleroneSwordfish (changes requested):
Two issues from the testmerge: 
image
- Reinforced pipes look very stripy
- The south west sprite seems to have a little line on it but that might just be a result of the stripes

RubberRats:
Something I noticed while this was testmerged is prefabs that use pipes for decoration (like the faint signal station) were broken by this change. Neglected to take screenshots but they weren't continuous.

Laboredih123:
> Something I noticed while this was testmerged is prefabs that use pipes for decoration (like the faint signal station) were broken by this change. Neglected to take screenshots but they weren't continuous.

prefabs my not beloved

Laboredih123:
> > Something I noticed while this was testmerged is prefabs that use pipes for decoration (like the faint signal station) were broken by this change. Neglected to take screenshots but they weren't continuous.
> 
> prefabs my not beloved

i really have to get to that decal rework

PR #20484, opened 2024-08-30
Fixes player vending machines registering MechComp signals

### LINKED ISSUE #20362: Build-A-Vend does not accept mechcomp inputs
### Labels

[BUG][MAJOR]

### Description

Build-A-Vend machines don't vend items when recieving an input signal from a linked component, such as a button. This happens regardless of whether you select "Vend random" or "Vend by name", and regardless of what the input signal is.

### Steps to reproduce

1. Build a vending machine frame, insert the components, and turn it on.
2. Unlock it with a screwdriver and click it with a multitool.
3. Open the wire panel, unlock the machine, and turn on the loading chute.
4. Insert a couple random items into the machine.
5. Turn off the loading chute and lock the machine.
6. Place a button and wrench it into place.
7. Click the button with the multitool, then click "connect component".
8. Click the Build-A-Vend with the multitool and select "vend random".
9. Stop linking with the multitool, then press the button.
10. Nothing happens.

### Expected Behavior

I expected pressing the button to vend a random item from the machine.

Reporting on 8/18/2024, goon #3

MetricDuck:
[BUG] [GAME OBJECTS]
## About the PR 
Player vending machines, like all vending machines, are registered for the "Vend Random" and "Vend by Name" signals. These signals use procs that interact with the vending machines product_list. 

However, player vending machines store their inventory in player_list not product_list: player vending machines can receive mechcomp signals but nothing will actually happen if they're triggered (as their product_list is always empty). 

This PR makes it so player vending machines aren't registered for the two MechComp signals.

## Why's this needed? 
Fixes #20362

### DIFF
--- a/code/modules/vending/vending.dm
+++ b/code/modules/vending/vending.dm
@@ -133,7 +133,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item, proc/admin_command
 	///The product currently being vended
 	var/datum/data/vending_product/currently_vending = null // zuh
 
-	var/mechcomp_compatible = TRUE
+	var/mechcomp_compatible = TRUE //Can this vending machine take mechcomp inputs?
 
 	power_usage = 50
 
@@ -2093,7 +2093,7 @@ TYPEINFO(/obj/item/machineboard/vending/monkeys)
 	player_list = list()
 	var/lastPlayerPrice = 0
 	icon_panel = "standard-panel"
-	var/mechcomp_compatible = FALSE
+	mechcomp_compatible = FALSE //Player vending machines can't take mechcomp inputs
 
 	New()
 		. = ..()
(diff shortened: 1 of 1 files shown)


MetricDuck:
image

MetricDuck:
A nicer long-term fix that could be pursued after the TGUI feature freeze could be to add a new button to the UI, much like the "Loading Chute" button, called "MechComp Enabled".

TobleroneSwordfish (changes requested):
Better to make it a flag variable, ie `var/uses_mechcomp = TRUE` or an overridable proc, ie `proc/register_mechcomp()`
`istype(src, ...)` is code smell that makes things harder to maintain by hiding child functionality in the parent class.

MetricDuck:
> Better to make it a flag variable, ie `var/uses_mechcomp = TRUE` or an overridable proc, ie `proc/register_mechcomp()` `istype(src, ...)` is code smell that makes things harder to maintain by hiding child functionality in the parent class.

Makes sense!

PR #20486, opened 2024-08-30
Loosens UI ID requirements on player vending machines

### LINKED ISSUE #16974: Player vending machines can eat pda of the owner (with ID in them) when vending chute is open
### Labels

[BUG][MAJOR]

### Description

as the maker of a player vendor, if you accidentally click on the machine with your pda (with an id in it) while the product chute is loading, you lose the pda and nobody can close the chute afterwards

### Steps to reproduce

1. make a player vending machine and open it up with a screwdriver
2. make the vending machine yours with you pda+ID card (or maybe just Id, I am not sure!)
3. put things in it
4. accidentally put your pda with your id in as well, by mistake
5. realize that you can no longer change any of the values of the vendor, because you haven't got your id!
6. try to put in money so you can buy your id, but have your money get put into the list of items sold instead (chute is open, can't close it)
7. repeat process for anything else you try (screwdirvers to close the panel, other people's pdas - including the admins, so on)

### Expected Behavior

pda with owner ID in it is not able to be put into player vending machine sell list. eating everyones things is bad!

I think zamujasa saw this in the game - they helped fix it for us. thank you Steph! :>
but I don't think it should happen to start with, its very confusing!

(I am not sure how it was fixed by admins. cutting wires with tools doesnt put them in the machine, but when i tried I couldn't close the chute that way. If there is in-game way to fix, then maybe this is not that bad?)

MetricDuck:
[BUG] [GAME OBJECTS]
## About the PR 
Player vending machine UI commands (lock/unlock, open/close chute, rename, reprice, change icon) will only respond to input if:
a) There's no vending machine owner
or
b) The inputter has the vending machine owners ID

Unfortunately, this means that if the vending machine's owner drops their ID into the vending machine, the vending machine is now pretty stuffed:
1) The ID can't be vended from the machine whilst the panel is unscrewed;
2) The panel can't be screwed whilst the loading bay is open;
3) The loading bay can't be shut without the correct ID. See 1).

This PR changes the security a bit, such that whilst only the ID holder can lock/unlock the UI, once the UI is unlocked anyone can mess with the buttons. The loading bay is only ever open whilst the UI is unlocked; so anyone will always be able to shut the loading bay regardless of ID.

The PR also catches a runtime when trying to credit the owner's account of ownerless vending machines.

## Why's this needed? 
Fixes #16974

### DIFF
--- a/code/modules/vending/vending.dm
+++ b/code/modules/vending/vending.dm
@@ -564,7 +564,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item, proc/admin_command
 		if("togglechute")
 			if(istype(src,/obj/machinery/vending/player))
 				var/obj/machinery/vending/player/P = src
-				if(usr.get_id()?.registered == P.owner || !P.owner)
+				if(P.unlocked)
 					P.loading = !P.loading
 		if("togglelock")
 			if(istype(src,/obj/machinery/vending/player))
@@ -576,7 +576,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item, proc/admin_command
 		if("setPrice")
 			if(istype(src,/obj/machinery/vending/player))
 				var/obj/machinery/vending/player/P = src
-				if(usr.get_id()?.registered == P.owner || !P.owner)
+				if(P.unlocked)
 					for (var/datum/data/vending_product/R in player_list)
 						if(ref(R) == params["target"])
 							R.product_cost = text2num(params["cost"])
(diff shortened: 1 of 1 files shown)


code/modules/vending/vending.dm
```diff
@@ -666,15 +666,15 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item, proc/admin_command
 							qdel(product)
 						product.product_amount--
 					if(src.pay && vended)
+						var/obj/machinery/vending/player/vMachine = src
 						if (src.acceptcard && account)
 							account["current_money"] -= product.product_cost
 						else
 							src.credit -= product.product_cost
-						if (!player_list)
+						if (!player_list || !vMachine.owneraccount)
```
TobleroneSwordfish, line 674: Is there a guarantee that `src` is a player vending machine at this point? Ie can this code be reached while operating a normal vending machine? If it can then this will runtime.

TobleroneSwordfish (changes requested):

code/modules/vending/vending.dm
MetricDuck: player_list should be null on anything that's not a player vending machine - normal vending machines use product_list - so my understanding was the check would never reach !vMachine.owneraccount if it wasn't a player vending machine.

That being said ?. are free so

code/modules/vending/vending.dm
```diff
@@ -666,15 +666,15 @@ ADMIN_INTERACT_PROCS(/obj/machinery/vending, proc/throw_item, proc/admin_command
 							qdel(product)
 						product.product_amount--
 					if(src.pay && vended)
+						var/obj/machinery/vending/player/vMachine = src
 						if (src.acceptcard && account)
 							account["current_money"] -= product.product_cost
 						else
 							src.credit -= product.product_cost
-						if (!player_list)
+						if (!player_list || !vMachine?.owneraccount)
```
TobleroneSwordfish, line 674:
```suggestion
						if (!player_list || !vMachine.owneraccount)
```

TobleroneSwordfish (approved):
`?.` doesn't actually do anything here because `src` can't be null but the `player_list` part makes sense.

PR #20489, opened 2024-08-30
Fixes dumping satchels onto tables along with their contents

Egregorious:
## About the PR 

Added a proc to obj/item, which is called during obj/place_on() and allows the item being placed to opt out depending on its own state and the parameters of the call. Satchels override said proc to avoid being placed on tables when the call is the result of a click-drag while they're not empty. So satchels now remain in-hand when dumping their contents out through click-dragging onto tables. 

atom/MouseDrop() now adds an entry to 'params' so that subsequent procs can all have explicit knowledge that they are ultimately the result of a click-drag. Also fixed obj/item/mouse_drop() overriding its own 'params' with 'over_control'.

## Why's this needed? 

This is basically standardising satchel interactions, since every other satchel-emptying mechanic results in the satchel remaining in-hand. 

This should also make it easier to fix similar issues in the future. 

## Changelog 

```changelog
(u)Egregorious
(+)Keep hold of satchels when dumping their contents onto tables.
```


Egregorious:
Thinking about it, can_place_on() isn't actually necessary for this fix, it could just be replaced by an if-return block in table. I made it because I had ideas about it being a generic fix that could also affect mouse_drop()'s click-drag onto ground block, but backed out of doing that because it seemed like too much work. So now it might just be making the code less readable by being an exception to the way the rest of the checks are written.

TobleroneSwordfish (approved):
This seems fine

PR #20497, opened 2024-08-31
Adds fishing lootpools and adds a new fishing spot

M-Earthfire:
[game objects][internal][feature]

## About the PR 

This PR adds a new datum that is connected to fishing spots: `fishing_lootpools`

This datum is a holder for a certain conditional and a loottable that can be created, modified and added to fishing spots. Fishing spots then check the conditionals on each loottable and add their respectable loot, if it is fulfilled, onto the to be rolled loot.

To demonstrate its use, two changes/additions to fishing spots were made:
- each fishing spot that could roll the lava fish now was dropped to tier 2 and can only roll the lava fish if a tier 3 fishing rod was used
- Clown shoes were added as a new fishing spot. Clowns and people with the clumsy trait are able to fish additional clown-appropriate items out of them.

## Why's this needed? 

Fishing spots have the very hard limitation to have the same loot among all fishing rod tiers. This would also makes it very hard to later add new mechanics that modify loot pools (e.g. baits or hooks you could attach to your fishing rod).
This datum adds an easy way to modify loottables of fishing spots. It's also being conceptualized to be later used with a fishing component, where it is meant to be the holder for the loottables.

While it is theoretically possible to move all fishing spot loottables onto this new datum, the fact that they are not standartized at all would add no gain at all. Thus, no real gain would come from it.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)The upgraded fishing hook is able to fish in hot places now. But it is insufficient to catch lava fish.
(+)Clown shoes can now be fished in. But only the clown knows the ins and outs of these dangerous waters.
```


code/modules/fishing/fishing_spots.dm
```diff
@@ -26,15 +26,42 @@ ABSTRACT_TYPE(/datum/fishing_spot)
 	/// the type of the atom that is the "fishing spot"
 	var/fishing_atom_type = null
 	/// associative list with the format (fish_type = probability), doesnt need to be ordered in descending probability
+	/// these are the fishing results that are ALWAYS avaiable at the spot. These won't get modified by conditionals like fishing loottables are.
 	var/list/fish_available = null
```
Glamurio, line 30: This is bound to cause problems, it's not really obvious from the variable name that these are innate and unaffected by loot tables. Comment is nice, but I'd prefer a more descriptive var name that makes it hyper-obvious.

Might almost consider doing away with this one entirely and replacing every instance of "fish_available" with a lootpool corresponding to the location, but that can be done in a follow-up PR.

code/modules/fishing/fishing_spots.dm
```diff
@@ -134,6 +161,18 @@ ABSTRACT_TYPE(/datum/fishing_spot)
 	/obj/item/reagent_containers/food/fish/real_goldfish = 5,\
 	/obj/item/reagent_containers/food/fish/red_herring = 1)
 
+/datum/fishing_spot/clown_shoes
+	fishing_atom_type = /obj/item/clothing/shoes/clown_shoes
+	rod_tier_required = 1
+	fish_available = list(/obj/item/reagent_containers/food/snacks/burger/moldy = 5, \
```
Glamurio, line 167: Why do we have `/datum/fishing_lootpool/clown_shoes_loot` if we're adding more manually here? Sort of defeats the point imho, better to put these items in the lootpool and have as few manual assignments as possible.

code/modules/fishing/fishing_spots.dm
```diff
@@ -524,14 +547,17 @@ ABSTRACT_TYPE(/datum/fishing_spot)
 
 /datum/fishing_spot/lava_moon
 	fishing_atom_type = /turf/unsimulated/floor/lava
-	rod_tier_required = 3
-	fish_available = list(/obj/item/reagent_containers/food/fish/lava_fish = 25,\
-	/obj/item/reagent_containers/food/fish/igneous_fish = 10,\
+	rod_tier_required = 2
+	fish_available = list(/obj/item/reagent_containers/food/fish/igneous_fish = 10,\
```
Glamurio, line 551: Same as with the clown shoes, if we're doing loot tables, let's try to clean up all of the individual assignments. Not saying do this for every fishing spot in this PR, but if you start messing with balance (like lowering rod tier) then you should do it proper.

code/modules/fishing/fishing_spots.dm
```diff
@@ -719,8 +745,11 @@ ABSTRACT_TYPE(/datum/fishing_spot)
 /datum/fishing_spot/furnace
 	fishing_atom_type = /obj/machinery/power/furnace/thermo
 	rod_tier_required = 2
-	fish_available = list(/obj/item/reagent_containers/food/fish/lava_fish = 25, \
-	/obj/item/reagent_containers/food/fish/igneous_fish = 10)
+	fish_available = list(/obj/item/reagent_containers/food/fish/igneous_fish = 10)
```
Glamurio, line 748: see prev. comment

code/modules/fishing/fishing_spots.dm
```diff
@@ -739,9 +768,12 @@ ABSTRACT_TYPE(/datum/fishing_spot)
 //Arc electroplater
 /datum/fishing_spot/arc_electroplater
 	fishing_atom_type = /obj/machinery/arc_electroplater
-	rod_tier_required = 3
-	fish_available = list(/obj/item/reagent_containers/food/fish/lava_fish = 25, \
-	/obj/item/reagent_containers/food/fish/igneous_fish = 10)
+	rod_tier_required = 2
+	fish_available = list(/obj/item/reagent_containers/food/fish/igneous_fish = 10)
```
Glamurio, line 772: see prev. comment

Glamurio (changes requested):
After giving it some thought, I'm quite happy with this implementation. It has foresight added to it when considering baits and hooks later, so while it seems a bit odd now it will be the better approach later.

Glamurio:
Yeah if the intent is to adjust some of these in a later PR then I don't really have any outstanding complaints.

Glamurio (approved):

PR #20498, opened 2024-08-31
Standardize AI Core Containment

JORJ949:
## About the PR 
Standardizes AI core glass boxes to have the following:

- Electrified reinforced glass
- Shutters that start open
- Door to the core
- Turret control for chamber outside the box
- Camera in the box so you can see core when the shutters are closed
- Blue matrix under the AI and black under the windows

Also removes the extra bit on the side of oshan core and gives it to maints

## Why's this needed? 
AI cores differ a lot from map to map and a bolted door is a nice QoL than having to break down the window every time you want to move your core

## Changelog 

```changelog
(u)JORJ949
(+)AI cores now have a bolted door among other changes.
```

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -4432,8 +4432,14 @@
 	name = "AI Core Shield";
 	opacity = 0
 	},
-/obj/mapping_helper/wingrille_spawn/auto/reinforced,
-/turf/simulated/floor/circuit,
+/obj/mapping_helper/wingrille_spawn/auto/reinforced/crystal,
+/obj/cable{
+	icon_state = "0-2"
+	},
+/obj/cable{
+	icon_state = "0-4"
+	},
+/turf/simulated/floor/circuit/off,
 /area/station/turret_protected/ai)
 "asx" = (
 /turf/simulated/floor/blueblack{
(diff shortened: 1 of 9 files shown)


Glamurio:
Might want to mention the slight Oshan core redesign in the PR, but it looks better so I have no issues with it.

Glamurio (approved):

TobleroneSwordfish (changes requested):
Couple things:
- Why plasmaglass? Most AI cores just have reinforced glass currently I think and plasmaglass takes aaaages to break through.
- You shouldn't need a camera in the box since the AI core has a camera in it?
- Leave the doors unbolted by default because stealing the AI core is funny.

JORJ949:
> Couple things:
> 
> * Why plasmaglass? Most AI cores just have reinforced glass currently I think and plasmaglass takes aaaages to break through.

Because theres a door now to go through and it feels like it should have the most secure version, though in reality the real barricade is the shutters so im fine to lower it to reinforced

> * You shouldn't need a camera in the box since the AI core has a camera in it?

Nope, you can see from core when deployed to it, but not in camera view

> * Leave the doors unbolted by default because stealing the AI core is funny.

True

TobleroneSwordfish (approved):

PR #20517, opened 2024-09-01
Lets items/mobs be placed on trays directly

RubberRats:
[GAME OBJECTS][QOL]
## About the PR 

This PR changes `machine_tray` objects so that using items and grabs on the tray with `attackby()` places them on it instead of attempting to smack the tray.

## Why's this needed? 

For years, I kept having issues trying to place things into the morgue and crematorium trays. I think this new behavior should make it more intuitive.

## Changelog 

```changelog
(u)RubberRats
(+)Items and grabbed mobs can now be placed in tray machines (like the morgue) directly.
```

### DIFF
--- a/code/obj/machinery/traymachines.dm
+++ b/code/obj/machinery/traymachines.dm
@@ -276,6 +276,14 @@ ABSTRACT_TYPE(/obj/machine_tray)
 /obj/machine_tray/attack_ai(mob/user as mob)
 	attack_hand(user)
 
+/obj/machine_tray/attackby(obj/item/I, mob/user)
+	if(istype(I, /obj/item/grab))
+		var/obj/item/grab/G = I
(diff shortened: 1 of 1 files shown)


code/obj/machinery/traymachines.dm
```diff
@@ -276,6 +276,14 @@ ABSTRACT_TYPE(/obj/machine_tray)
 /obj/machine_tray/attack_ai(mob/user as mob)
 	attack_hand(user)
 
+/obj/machine_tray/attackby(obj/item/I, mob/user)
+	if(istype(I, /obj/item/grab))
+		var/obj/item/grab/G = I
+		G.affecting.set_loc(src.loc)
```
ZeWaka, line 282: This seems a bit sus?

PR #20518, opened 2024-09-01
Reworks the aranesp recipe

M-Earthfire:
[Balance][Rework][Chemistry]

## About the PR 

Reworks the aranesp recipe into the following two recipes:

(1) Insulin + (1) cryoxadone + (0) nickel @ under 273K -> (2) aranesp + (0) nickel

This recipe heats up over time, which goes into the following instant reaction when not monitored:

(1) Insulin + (1) cryoxadone + (0) nickel @ over 273K -> (1) green vomit + (1) carbon + (0) nickel

Also, the fermid recipe was changed to require a temperature over 273K

## Why's this needed? 

Aranesp requires 3 quite complex medical chems to source right now, 2 of them being valuable to medbay. This PR makes it slightly harder for doctors to create the chem, because they now need to pay attention to the reaction conditions. On the other hand they still start with the precursors and the only high-tier medchem needed is changed to be insulin. This makes it much easier to source the chems needed to create Aranesp.

Overall, this will make the chem overall more avaible to doctors and chemists alike, since it is only bottlenecked by insulin instead of 3 medchems.

The failstate of this reaction was made to a usefull chem can be extracted out of it. I think a seasoned pirate/bartender knows where green vomit comes in handy. And really, everyone doing chemistry should once in their life create tar. It's a rite of passage.

The change to the fermid recipe was made so the delayed reaction of aranesp cannot be chained into fermid. I think noone wants that.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)The Aranesp recipe was changed to: (1) Insulin + (1) cryoxadone + (0) nickel @ under 273K -> (2) aranesp + (0) nickel. The reaction heats up over time and, if left unattended, can turn into useless(?) tar.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -544,8 +544,30 @@
 		name = "Aranesp"
 		id = "aranesp"
 		result = "aranesp"
-		required_reagents = list("atropine" = 1, "epinephrine" = 1, "insulin" = 1)
-		result_amount = 3
+		required_reagents = list("nickel" = 0, "cryoxadone" = 1, "insulin" = 1)
+		result_amount = 2
+		instant = 0
+		reaction_speed = 0.25
+		max_temperature = T0C
+		reaction_icon_state = list("reaction_puff-1", "reaction_puff-2")
+		mix_phrase = "The solution emits a fine mist as it slowly begins to change colour."
+
+		on_reaction(var/datum/reagents/holder, var/created_volume)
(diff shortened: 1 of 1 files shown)


Egregorious:
This isn't necessarily an argument against, but this does essentially remove it from being accessible by botany. Botanists can grow atropine and insulin and then source epi from minimeds/cargo - but they don't have access to a chem dispenser for nickel/platinum nor a good method of cooling reagents by default. 

Again, not necessarily an argument against, but I think worth considering.

Tarmunora (approved):

PR #20520, opened 2024-09-01
Fixes typo in syndicate drone AI

MetricDuck:
[BUG]
## About the PR 
~Horseman AI has a '&' that should be a '&&' in its logic so the code runtimes when it attempts to target a vehicle.~

Syndicate drone AI has a '&' that should be a '==' in its logic so the code runtimes when it attempts to target a vehicle; at some point the FACTIONS were presumably bitflags.

This PR fixes that.

## Why's this needed? 
Typo

### DIFF
--- a/code/obj/critter/drone.dm
+++ b/code/obj/critter/drone.dm
@@ -178,7 +178,7 @@ TYPEINFO(/obj/critter/gunbot/drone/helldrone)
 					var/obj/machinery/vehicle/C = atom
 					if (C.health < 0)
 						continue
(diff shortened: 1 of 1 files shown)


ZeWaka (commented):
No

Azrun:
Oh cool! I broke this!

Azrun:
Addressed by [b5193e9](https://github.com/goonstation/goonstation/commit/b5193e99415ae70c609bba8d6cb895a0a27161ed)

PR #20527, opened 2024-09-02
Randomised click-drag item placement offset for tables

Egregorious:
## About the PR 

This is an extension for a fix done in #20489; added a randomised offset option for obj/place_on(), which tables make use of during click-drag item placements.

Also a couple more minor things;
- Standardised table/MouseDrop_T() to make use of place_on() for item placement. 
- obj/place_on() attempts to convert 'params' to a list if it isn't already, for ease of use. 

## Why's this needed? 

Because params were borked in item/mouse_drop(), every click-drag to table item placement was being processed through a place_on() call with essentially nulled parameters, leading to items being placed on the table without an updated offset. Fixing those params made it so the items accurately dropped on the mouse cursor - but ingame that just feels wrong somehow. It also may arguably reduce player control, because it takes away the option to do an "imprecise" placement of an item, when precise placements are already available through clicking normally. 

This change makes it so click-dragging items from an inventory onto a table is imprecise again. The randomness is experimental, it might be better to just maintain previous functionality and so avoid updating the offset instead.

### DIFF
--- a/strings/changelog.txt
+++ b/strings/changelog.txt
@@ -1,5 +1,10 @@
 
 (t)wed sep 25 24
+(u)Lord_Earthfire
+(p)20497
+(e)🆕📦⚙|C-Feature, A-Game-Objects, A-Internal
+(+)The upgraded fishing hook is able to fish in hot places now. But it is insufficient to catch lava fish.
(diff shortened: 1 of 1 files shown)


Egregorious:
Woops, tried to sync it and deleted all the commits.

PR #20538, opened 2024-09-03
Adds catering announcement computers

TobleroneSwordfish:
[MAPPING] [FEATURE] [STATION OBJECTS]
## About the PR 
Adds an announcement computer to each catering department with a unique announcement sound and circuit board.

## Why's this needed? 
Chefs announcing special dishes, bartenders declaring that the clown is BANNED from the bar after the incident.
I always wanted catering to have an announcement computer but it doesn't fit for them to have the **important** sounding announcement sound so I modified a more sedate sounding one from here: https://pixabay.com/sound-effects/bingbong-42645/ (licensed under pixabay's standard license)

## Changelog 

```changelog
(u)LeahTheTech
(*)Added catering announcement computers.
```


code/obj/machinery/computer/announcement.dm
```diff
@@ -244,6 +247,14 @@
 		req_access = list(access_ai_upload)
 		name = "AI Announcement Computer"
 
+	catering
+		req_access = list(access_bar) //chef gets bar access
```
JORJ949, line 251: no they don't, bartender has only bar and chef has only kitchen, its just most kitchens are behind bar so the doors are both bar and kitchen access

PR #20549, opened 2024-09-03
DRAFT: Major Matsci Refactoring

Cherman0:
## About the PR 

A major redesigning of matsci with the goal for it to be much easier to add new unique materials. This includes several changes:
- The range of material stats is 9 times bigger (1-9 to 1-81)
- Interpolated custom materials are removed.
- Raw and refined materials have been combined into a single item, removing the need for refining.
- All materials are repathed with paths based on their actual material classifications. Most notably is an introdduction of "ceramics", which includes rocks, glass, plastic, bones, crystals, and basically everything generally defined by being inflexible and not very conductive.
- Stats "flammability". "reflectivity", "fissile isotopes", "molitz bubbles", and "plasma offgas" were removed. Flammability and reflectivity are now material flags and the others are now handled by their items.
- Binary properties that are not related to the material's characteristic classification (such as biological origin or flammability) are now flags instead.

## Why's this needed? 

Matsci is really difficult to design anything unique for because the range of unique materials is pretty much completely filled out. This should significanlty open the design space by increasing the range of stats, making material more consistently typed and flagged, and removing the ability to accumulate material flags using interpolated materials.

## This will probably require changes in secret

Specifically, every check for material properties (ie `getProperty()`, `hasProperty()`, etc.) will need to be adjusted for the new stat range (usually just appending `/9`) and the new property ID macros. Every call of material.getMaterialFlags() will need to be either changed into an `istype()` (if it was checking for a type, such as a metal or crystal flag) or updated for the new flags. `raw_material` and `material_processed` items have been combined into a `material` item. Manufacturing requirements may need to be adjusted.

## Changelog 

```changelog
(u)Cherman0
(*)Matsci has been internally refactored. User-facing changes are intended to be minimal, but check minor changes for details.
(+)Interpolated alloying has been removed
(+)Raw and refined materials are combined into one type, removing the need for refining.
(+)Some materials have changed typing or flags.
(+)Some crafting recipes or material interactions may be somwhat different.
```


wolfolotl:
matsci property values were initially something like 1-100, and then they got condensed into 1-9. in my opinion (which is not necessarily shared by anyone else) this made matsci feel much more intuitive from a player-side perspective, whereas the width of possible property values before didn't really feel like it added anything impactful. it's a little funny to see a pr that undoes that, but I'm not necessarily opposed to it, as long as you can make the larger breadth of values justify itself

amylizzle:
```
- Raw and refined materials have been combined into a single item, removing the need for refining.
- All materials are repathed with paths based on their actual material classifications. Most notably is an introdduction of "ceramics", which includes rocks, glass, plastic, bones, crystals, and basically everything generally defined by being inflexible and not very conductive.
- Binary properties that are not related to the material's characteristic classification (such as biological origin or flammability) are now flags instead.
```

These I love. Also swapping out material name strings with defines, very good. The rest not so much.

Mutability was heavily tied in to the last redesign of matsci, and it's very cheap to handle it, there's no need to strip those properties off and put them on items - it only acts to reduce the complexity of interactions for no gain.

Making properties have a wider range again seems unnecessary.

Custom materials are pretty lame as they stand, but I'm not sure removal is the right way to go.

My general take on this PR is that it should be completely atomised, and each part submitted and evaluated separately.

TobleroneSwordfish (changes requested):
What Amy said! (especially the atomization bit) Although I would be open to removing custom alloys, I think it would need more work and support than this.

Cherman0:
It isn't really possible to atomize this, as everything in matsci is rather viscerally stitched together. You can't retype materials without addressing `/material/interpolated`, for instance. The intention here is to basically flush out all the cruft that prevents this system from being actually interesting in order to start building a better base on it in future PRs.

MyNameIsRomayne:
Struggling to even look at the code with the shear amount of find/replace which has occured, fwiw I also share the sentiments above, an increased range of values works only if there is justification from a player's standpoint to deal with 9 times more values. Having there be flexibility in the system to expand the range of values at any time sounds great -- but maybe it needs things that utilize that first instead of just being reasoned with "It is a really good number" (not saying it isnt)

TobleroneSwordfish:
It is absolutely possible to atomize this and nearly impossible to review/merge it without atomization given the variety of sometimes conflicting ideas presented here.

Cherman0:
Well... ok. I probably won't do that though, at least not anytime soon. So I'm going to close this. I'm somewhat disappointed because I had talked pretty extensively about what I was planning and doing in the discord and would have appreciated if any of these concerns had been mentioned at the time.

PR #20560, opened 2024-09-04
Prevents appendicitis from being asymptomatic

### LINKED ISSUE #20511: Appendix Missing Still has Appendicitis
### Labels

[BUG][MINOR]

### Description

Person is suffering from Asymptomatic Medical Emergency: Appendicitis (Stage 3/3) while having no Appendix

### Steps to reproduce

They walked in with Asymptomatic Appendicitis and i removed the organ as to not risk any issues, and they kept the status of  Asymptomatic Appendicitis

### Expected Behavior

When Removal of Appendix for the status to be removed.

https://cdn.discordapp.com/attachments/890223070999101480/1279683400097202206/image.png?ex=66d55589&is=66d40409&hm=3be1a4f837e500e352633675b4374ae0c2633d4ce42752852630c7b7ee4fd3df&

Reported by: connortheconqueror
Client version: 515.1640
Active test merges: #20186, #20205, and #20481
Round log date: 2024-09-01-04-28
Reported on: 2024-09-01 06:08:38
Round ID: 187197
Map: DONUT2

MetricDuck:
[BUG]
## About the PR 
Appendicitis checks for a(n) (robo)appendix in stage_act, curing itself if it does not find one.

If the disease is asymptomatic however, we never run disease/appendicitis/stage_act() as we early return out of disease/stage_act(). Thus, removing the appendix will not cure the appendicitis.

This PR adds an override that ensures we never roll asymptomatic appendicitis.

## Why's this needed? 
Fixes #20511
No balance tag as asymptomatic appendicitis doesn't actually do anything than show up on scans.

### DIFF
--- a/code/datums/disease.dm
+++ b/code/datums/disease.dm
@@ -23,6 +23,7 @@
 	var/detectability = 0				// detectors must >= this to pick up the disease
 	var/resistance_prob = 0				// how likely this disease is to grant immunity once cured
 	var/max_stacks = 1					// how many times at once you can have this ailment
+	var/can_be_asymptomatic = TRUE
 
 	///If we need a specific ailment_data type
 	var/datum/ailment_data/strain_type = /datum/ailment_data
@@ -90,7 +91,7 @@
 
 	setup_strain()
 		var/datum/ailment_data/disease/strain = ..()
-		if (prob(5))
+		if (prob(5) && src.can_be_asymptomatic)
 			strain.state = "Asymptomatic"
 			// carrier - will spread it but won't suffer from it
 		strain.virulence = src.virulence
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
Uhhh I'm not sure how that's meant to fix it but I don't think it does? That proc definition doesn't seem to do anything (calling parent and returning the value is the same as not defining the proc)

MetricDuck:
Sorry, I explained a bit poorly - the PR overrides this proc that's called when we first create the disease. 

image

MetricDuck:
all ailments can be asymptomatic, but diseases are currently the only ones who roll for it, so can_be_asymptomatic is on ailment/disease

TobleroneSwordfish (approved):

PR #20562, opened 2024-09-05
Button Directionals + Mounted Flasher Resprite

MitchTheUser:
[GAME OBJECTS]

## About the PR 

Adds unique North, East and West directional sprites for both the Door Control Button and the Door Timer Button. These sprites are all angled to fit on walls alongside other directional sprites like APCs, Fire Alarms and Light Switches.

This PR does not fully implement button directionals into all maps yet, however, all Door Timer and Control Button subtypes have been altered in the code to use the new directionals (Door Bolters, GenPop Timers, Podbay buttons, etc...) and should automatically apply to all maps, though there might be a few incorrect directions due to incorrect or inconsistent direction naming in the code, which can be fixed in a later PR.

As a small showcase, Atlas has been updated to have all generic and subtype buttons face in their new directions, as well as tweaking pixel position on some nearby objects to get rid of overlap and sprite clipping. This is my first real time using StrongDMM, so tell me if there are any problems in the map.

Button

Additionally, new_wall/[direction] subtypes for generic Door Timers and Control Buttons have been added for use by mappers, and a new (not yet used) table button subtype has been added to the code, which visually resembles a button with reversed perspective, which makes it look great when placed on tables.

ButtonTable

The second part of this PR resprites the ancient Wall Mounted Flasher found in cells with a redone sprite and animation to make it match other wall mounted machinery.

WallMountedFlashDisplayNew

As a micro bonus addition I made due to a mid PR discovery, the Cremation Switch sprite has been redone to be the same as the new Light Switch sprites, since I believe it was missed during that Resprite PR.

## Why's this needed? 

[All three of these sprites were on the Wanted Resprites List](https://github.com/orgs/goonstation/projects/5?pane=issue&itemId=5780478)


MitchTheUser:
Here's some screenshots of how they all look ingame
bar
med
sec

MitchTheUser:
Everything should be up to date and without Merge Conflicts now

Tarmunora (changes requested):
mapdiffbot is showing a changed turf in the atlas janitor room?

MitchTheUser:
> mapdiffbot is showing a changed turf in the atlas janitor room?

Fixed, the turf that was accidently changed was partially hidden by the Janitor bed, so I didn't notice it when testing.

Tarmunora:
only real note I have is that the wall mounted flasher not having directionals is a little jarring when it's being done in the same style as the rest of these, but that can be a future project perhaps

PR #20581, opened 2024-09-05
Add Black+Fingerless gloves to syndicate apparel vendor

JORJ949:
## About the PR 
Adds 4 of both black and fingerless gloves to the syndicate apparel vendor

## Why's this needed? 
You shouldn't have to break into security for simple black gloves and I forgot to add gloves in the initial PR


Glamurio (approved):
Yeah why not.

PR #20587, opened 2024-09-05
Gemstones are transparent enough to be in windows/antique

444Portal:
## About the PR
Just bumps their alpha value to be the same as glass (180). That's it.

## Why's this needed?
It'd be cool if you could use them in Windows and it makes making a basic level Antique a little easier. Maybe will encourage people to mess with them a bit more, who knows!

## Changelog

```changelog
(u)444explorer
(+)Gemstones are now more transparent - Use them in windows!
```


444Portal:
_Technically_ balance if this lets them be in the antique :v

TobleroneSwordfish (approved):

Issue #20579, opened 2024-09-05
1 pamgoc armory doors not updated


trazodont:
### Labels

[BUG]

### Description

oops old doors!

### Steps to reproduce

1. 1 pamgoc 
2. security :)

![image](https://github.com/user-attachments/assets/44ddf304-a919-41b1-abc5-89eebc51d260)

### Expected Behavior

new doors :)

Issue #20578, opened 2024-09-05
morgue trays on 1 pamgoc unusable


trazodont:
### Labels

[BUG]

### Description

lol cant open

### Steps to reproduce

1. 1 pamgoc 
2. medical
![image](https://github.com/user-attachments/assets/2df0dabd-d968-4a48-87de-9788bf6d49f5)

### Expected Behavior

open up!!

PR #20603, opened 2024-09-07
Brings the 1pamgoc armory up to parity with the cogmap1 armory and fixes morgue slabs

DecidedlyUnethical:
[MAPPING][BUG]
## About the PR 
Please ignore the branch name. This PR brings the 1pamgoC armory up to (mirrored) parity with the Cogmap1 armory with the following changes;
- Extends the armory upwards by one tile.
- Swaps the old doors for the new unhackable ones, which now also use mapping helpers instead of varedits.
- Moves the N2O canister away from said doors.
- Adds two flashbang boxes and three pepper grenades.
- Adds the anti-biological crate.
- Removes the HoS access spawner from the NARCs.

Bonus change, also turns the morgue trays around to face the correct way.

dreamseeker_XM6nhs4aE1

I _tried_ to get the thermals, NVGs, and gas masks to layer correctly, but the game just wasn't having it. Sorry.

## Why's this needed? 
Updating the map is nice, and also this fixes both #20578 and #20579.


ZeWaka:
no, this map is generated not edited

PR #20607, opened 2024-09-07
(REACT) Plantmaster Regular Data

Garash2k:
## About the PR 
Makes it so that the Plantmaster data is passed as regular data in lieu of static data.
Also removes an unused usage of seedoutput.

## Why's this needed? 
Static data wasn't appropriate for any of the data. Static data is appropriate for "large data that should be sent infrequently." Most of the properties are small except for, arguably, the seed list and the extractable list, but those 2 change frequently.

[UI][Bugfix]


amylizzle:
The original reason this was done as static instead of regular uidata was to prevent sending all the data for 100s of seeds every 3 seconds to every user, and instead only to send that data when it was updated. IDK if that's still an issue with the react change.

Garash2k:
Yeah at first I was going for only keeping the seed and extractable list in static_data, but by doing that you still get a lot of "Loading dialog" when removing multiple seeds or doing multiple operations (like extractions) in a row, aka frequently sent data. I don't know either if it's not an issue anymore, we'd need input/testing

PR #20625, opened 2024-09-09
Removes the cargo carousel from Oshan

Cherman0:
## About the PR 

- Removes the cargo carousel from Oshan.
- The north dropoff point was replaced with an expanded maintenance tunnel and random room.
- The south dropoff point was replaced with a medical checkpoint.
- The carousel power room was replaced with two random rooms.
- Since inner maint no longer necessarily needs to be a continuous ring, doors were added to partition it into north, west, and east sections. This reduces how widespread a single flood in maint becomes and makes it not be entirely powered by one APC.

## Why's this needed? 

Many species of cuckoos are brood parasites, meaning that they lay their eggs in others birds' nests and let the other bird expend its resources on raising it. Some cuckoos are adapted to target specific species of birds, with the cuckoo's eggs and young mimicing that of their host bird in order to evade detection. The cuckoo often hatches before the other eggs in the nest and destroys them, ensuring that it recieves a larger share of the food its false parent bring back to the nest.

In this sense, the cargo carousel is like the cuckoo. It is adapted to mimic the SMES unit, and it extends its false wire into the nest of SMES wires in order to trick the unwitting engineer into feeding it instead of their true young. Without this trick it will starve and die, as it is unable to justify its existance on legitimate grounds.

Now in seriousness, the cargo carousel mostly just takes up a lot of space in inner maint and the only time anybody ever actually powers it is by accident. Otherwise it just expends its starting charge and completely fails about 30 minutes into the shift and nobody notices. It only has two delivery destinations and both are not useful at all. Removing it allows for a few extra random rooms (oshan has fewer than most maps of its size) and a medical booth on the south end of the station (I feel that oshan is large enough to justify one). It also makes wiring power into the SMES a little less confusing for new engineers.

### DIFF
--- a/maps/oshan.dmm
+++ b/maps/oshan.dmm
@@ -176,8 +176,20 @@
 /obj/cable{
 	icon_state = "0-2"
 	},
+/obj/machinery/camera{
+	c_tag = "autotag";
+	dir = 6;
+	name = "autoname - SS13";
+	pixel_y = 20;
+	tag = ""
+	},
+/obj/stool/chair/yellow{
+	dir = 8
+	},
+/obj/item/match,
+/obj/item/cigbutt,
 /obj/decal/tile_edge/stripe/extra_big{
-	dir = 1
+	dir = 5
 	},
 /turf/simulated/floor/shuttlebay,
 /area/station/hangar/engine)
@@ -186,13 +198,6 @@
 	dir = 8
 	},
 /obj/machinery/light/incandescent/warm,
-/turf/simulated/floor/shuttlebay,
-/area/station/hangar/engine)
-"aaO" = (
-/obj/decal/tile_edge/stripe/extra_big{
-	dir = 4
-	},
-/obj/item/cigbutt,
 /turf/simulated/floor/shuttlebay,
 /area/station/hangar/engine)
 "aaQ" = (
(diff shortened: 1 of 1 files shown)


JORJ949:
I like the carousel, perhaps to make space just make the inside of oshan smaller

DecidedlyUnethical:
I like the carousel too, but you can't deny that it's really just... not all that great.

TobleroneSwordfish:
Not everything on a map has to have an in universe practical purpose to justify its existence, many things exist mostly just for crime/shenanigans and I think the oshan carousel fulfils that role pretty well. Think of how many times someone has escaped a chase by diving through one of the carousel ports, or someone spots something weird and inexplicable (usually a corpse) doing slow circles on the belts.

If you want it to see more regular use then working to add more stations or move the existing ones out of grimy maint tunnels would be a good start.
Also if it's an issue that people are accidentally feeding power into it (don't think I've personally ever seen that but 🤷) why not just change the wire colour or label it better?

Cherman0:
> Think of how many times someone has escaped a chase by diving through one of the carousel ports, or someone spots something weird and inexplicable (usually a corpse) doing slow circles on the belts.

By my experience, both of these things have happened about zero times. Maybe the chase thing, but only the north one would really be useful for that and it still goes into maint. The cargo carousel is beyond obscure, it's pretty much books-and-papiers tier of mentally filtered out of existance. I have seen people power it accidentally, I have yet to see someone do it on purpose.

But maybe I'm just a MULE fan and want people to use them more.

AXLFAN2000:
> > Think of how many times someone has escaped a chase by diving through one of the carousel ports, or someone spots something weird and inexplicable (usually a corpse) doing slow circles on the belts.
> 
> By my experience, both of these things have happened about zero times. 

I've had both of those things happen many times.  As an antag, I've frequently used the belts for speedy get-away from a crime/chase, and as sec (or crew being hassled) I see other antags doing the same.  I've got little investment in it being removed or not, but wanted to back up that those things do happen, and I personally enjoy the bit of scrambling that the belts add to running around the inner maints where things and people can slip out of reach really easily.... But I also like random rooms.  Lol

Glamurio:
I kind of agree with the takes, I think even though the carousel isn't really that useful or commonly used, it is still iconic and nothing like it exists in this capacity. I'd also rather see the carousel expanded, maybe just connect it to the regular power network to have it be powered alongside the station if it stops after 30 minutes.

JORJ949:
Bits of contraband always end up on there and its fun to chase em down

MylieDaniels:
> By my experience, both of these things have happened about zero times. Maybe the chase thing, but only the north one would really be useful for that and it still goes into maint.

In the last week, I've experienced the former twice and the latter once, and seen two admin gimmicks involving the belt.

Cherman0:
I convinced people to remember that this thing exists and maybe even make it better so mission accomplished here.

PR #20627, opened 2024-09-09
Give AI cores an internal camera

JORJ949:
## About the PR 
Gives AI cores internal cameras so they can see what their core sees when in eyeview

## Why's this needed? 
you shouldnt need to deploy to your core to see your surroundings, these should be visible from AI view.

## Changelog 

```changelog
(u)JORJ949
(+)AI cores now have cameras in them that AI eyes can view out of.
```

### DIFF
--- a/code/mob/living/silicon/ai.dm
+++ b/code/mob/living/silicon/ai.dm
@@ -94,6 +94,7 @@ var/global/list/ai_emotions = list("Annoyed" = "ai_annoyed-dol", \
 	var/network = "SS13"
 	var/classic_move = 1 //Ordinary AI camera movement
 	var/obj/machinery/camera/current = null
+	var/obj/machinery/camera/camera = null //Our internal camera for seeing from core while in eye
 	var/list/connected_robots = list()
 	//var/list/connected_shells = list()
 	var/list/installed_modules = list()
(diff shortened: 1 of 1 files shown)


Tarmunora (approved):

code/mob/living/silicon/ai.dm
```diff
@@ -383,6 +384,10 @@ or don't if it uses a custom topopen overlay
 			src.bioHolder.mobAppearance.pronouns = src.client.preferences.AH.pronouns
 			src.update_name_tag()
 
+		src.camera = new /obj/machinery/camera(src)
+		src.camera.c_tag = src.real_name
```
TobleroneSwordfish, line 388: Minor thing but this probably should update on name change then?

TobleroneSwordfish (changes requested):

code/mob/living/silicon/ai.dm
JORJ949: Afaik theres no actual way to see this camera in a list naturally, and borg name changes dont update it I think

JORJ949: Can confirm borg name changes dont change their camera name and are probably "Cyborg Beta-15" forever but i'm not entirely sure it matters as there is no way to see these cameras on a list unlike "Test Chamber 1" for example

PR #20660, opened 2024-09-11
Fix issues with atlas artifact ejector PR

JORJ949:
## About the PR 
Fixes the following issues caused by atlas art ejector PR:
Window that leads directly to a wall in ranch
image
Artifact ejector being off cams
Artifact ejector launching arts directly into the very much dense solar tracker
image

## Why's this needed? 
I love devastating artbombs on atlas :D
I also love staring at walls!


TobleroneSwordfish:
lol

PR #20661, opened 2024-09-11
Improve the Oshan carousel

TobleroneSwordfish:
## About the PR 
An alternative to #20625, starting to improve the carousel by making the dropoff stations more accessible and adding a chute from QM onto the belt.
In future it'd be nice to add additional dropoff stations in other locations but this should go some way towards making the big belt loop a bit more practical to use.
image
image
image
image

## Why's this needed? 
I like the carousel but I think some of the criticism in the removal PR is valid so I tried to improve it a bit!

## Changelog 

```changelog
(u)LeahTheTech
(*)The Oshan belt carousel stations have been improved and a chute has been added from cargo onto the belt.
```

### DIFF
--- a/code/map.dm
+++ b/code/map.dm
@@ -804,7 +804,7 @@ var/global/list/mapNames = list(
 	merchant_right_centcom = /area/shuttle/merchant_shuttle/right_centcom/cogmap
 	merchant_right_station = /area/shuttle/merchant_shuttle/right_station/cogmap
 
-	shipping_destinations = list("North", "South")
+	shipping_destinations = list("North Carousel", "South Carousel")
 
 	valid_nuke_targets = list("the fitness room" = list(/area/station/crew_quarters/fitness),
 		"the quartermaster's office" = list(/area/station/quartermaster/office),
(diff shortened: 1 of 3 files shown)


Cherman0:
I like it conceptually but I dont think this is really a useful way to go about it. The carousel suffers from both a lack of inputs and outputs, and remediating that would likely require some cog1-esque mail rooms in departments for it to directly hook into. Otherwise there's no real point in adding more input pipes if there still arent useful output spots for the cargo to actually go to.

Doing this would likely require routing the carousel around the exterior of the station and having it poke in through maints to medbay, cargo, security, engineering, etc. Having an exterior line of belts would also be nice if you could run a line of insulated cables in parallel, as wiring the carousel/SMESs to a hotspot in the ocean on the opposite end of the station takes a lot of time and potentially multiple boxes of cable.

TobleroneSwordfish:
> Feels weird being launched in the opposite direction

I did think about that, but I'm trying to avoid running pipes under r-walls or through upload.

Routing the carousel around the outside of the station instead might be worth trying but that's a much larger redesign and would make it take *even longer* for crates to get around so I'm starting with just modifying what's there for now. I could definitely see east/west dropoffs being added in the bar or next to mechanics in future though.

PR #20666, opened 2024-09-11
Resprites Berets

MitchTheUser:
[CLOTHING] [GAME OBJECTS]

## About the PR 

As the tin says, this PR resprites Berets, giving them higher contrast, a more vibrant palette that closely follows it's worn sprite, and fixes to shading that was missing or messed up on several Berets.

BeretDisplay

## Why's this needed? 

Berets are pretty dull in color and have messed up shading, some Berets missing an extra line of shading, others missing several pieces of shading, and some having weird inverted shading. This also brings them closer to how their worn sprites appear while also making them stand out more with brighter colors.

BeretCapture


itsloley:
the brighter colors are nice! though there are a couple things i noticed, being that i think the shading might look a little... dirty? it looks kind of worn and old, i'm not sure how else to put it.. and the smushed HoS beret even more-so (which is fitting, but i feel like the grime is too noisy..?)

the palette (mainly the outline and badge on the beret) is also a bit off from the worn sprite palette / palette pinned in #im-spriter which isn't a big deal or anything but i think it might look better :>

i threw this quick mock-up together based on your sprites to kinda show what i mean if that helps!

berets

MitchTheUser:
Tweaked the sprites a bit based on your suggestions. I see where your coming from about the shading making them look a little bit dirty. I'm not sure about using the really dark outlines of the worn sprites but if someone else thinks it would also make them look better I could tweak it a bit further.

BeretAltDisplay

itsloley:
yeah, that looks a lot better!

> I'm not sure about using the really dark outlines of the worn sprites but if someone else thinks it would also make them look better I could tweak it a bit further.

fair enough, i think the prisoner's beret could still use a bit of a darker outline just since i don't think there are typically(?) modern items with bright outlines, but of course i don't speak for everyone else :p, looking great!!

PR #20672, opened 2024-09-12
Antihol rework (ish

imawxome:
## About the PR This PR works to rebalance antihol.
[balance] [chemistry] [S]   

## Why's this needed? 
Originally, a single unit of charcoal, when injected into a drunk person, can purge a whopping 41 units of ethinol, in  only ten seconds. this is silly, and makes it way to easy to instantly sober up anyone who drinks to much.

 The changes make it so that antihol purges less ethinol, and cannot be created just by clicking someone with a hypospray of charcoal, instead requireing a 3-1 mix of char and ipepec. this also allows ipepec to be used to  "spike"  poisons, making it harder to treat poisonings with charcoal alone, as it will react with the ipepec to make antihol, and not purge the toxins. This acts as an indirect buff to the “weak” poisons, such as cyanide or atrazine,that can be trivially removed with charcoal, as you can make it substantially harder to treat with charcoal alone.

originally, using ethanol to “spike” poisons was actively detrimental, creating antihol that would just purge all the ethinol, and then allow charcoal to be injecting, leaving you with 2 toxic healing chemicals instead of one, making it significantly harder to kill someone than if you just used poison alone.

## Changelog 

```changelog
(u)imawxome
(*) changes the recipe for antihol, and lowers its rate of ethinol purging
(*) changes the color of antihol
(+)the new recipe is: (1)ipepec+ (3)charcoal  -> (2) antihol
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -2125,10 +2125,10 @@
 		name = "Antihol"
 		id = "antihol"
 		result = "antihol"
-		required_reagents = list("ethanol" = 1, "charcoal" = 1)
+		required_reagents = list("ipecac" = 1, "charcoal" = 3)
 		result_amount = 2
-		mix_phrase = "A minty and refreshing smell drifts from the effervescent mixture."
-		mix_sound = 'sound/misc/drinkfizz.ogg'
+		mix_phrase = "A gross sludge forms in the mixture"
+		mix_sound = 'sound/impact_sounds/Slimy_Hit_4.ogg'
 
 	ectocooler
 		name = "Ecto Cooler"
--- a/code/modules/chemistry/Reagents-Medical.dm
+++ b/code/modules/chemistry/Reagents-Medical.dm
@@ -1591,18 +1591,17 @@ datum
 			id = "antihol"
 			description = "A medicine which quickly eliminates alcohol in the body."
 			reagent_state = LIQUID
-			fluid_r = 0
-			fluid_b = 180
-			fluid_g = 200
+			depletion_rate = 0.6
+			fluid_r = 40
+			fluid_b = 0
+			fluid_g = 40
 			transparency = 220
 			value = 6 // 5c + 1c
 			var/list/flushed_reagents = list("ethanol")
 
 			on_mob_life(var/mob/M, var/mult = 1)
 				if(!M) M = holder.my_atom
-				flush(holder, 8 * mult, flushed_reagents)
-				if (M.get_toxin_damage() <= 25)
-					M.take_toxin_damage(-2 * mult)
+				flush(holder, 2.5 * mult, flushed_reagents)
 				..()
 				return
 
@@ -1610,7 +1609,7 @@ datum
 			name = "space ipecac"
 			id = "ipecac"
 			fluid_r = 2
-			fluid_g = 20
+			fluid_g = 2
 			fluid_b =  5
 			description = "Used to induce emesis. In space."
 			reagent_state = LIQUID


imawxome:
Ethanol is sometimes used to thwart charcoal treatments, but the current recipe lets you “unthwart” it just by injecting even more charcoal, something cyborgs and medical have an nigh unlimited supply of, so it becomes useless to try and use ethanol to stop people from getting toxins purged. The new recipe also would encourage botany to grow ipepecs for medbay, instead of just for meth. Additionally, all the “weak” toxins (cyanide, atrazine ext) are all useless due to how easy they are to solve with charcoal alone, letting ipepec be used for antihol acts as an indirect buff to these chemicals, as you can’t simply click them once to heal all the toxic damage and clear out all the chemicals, instead requiring some time in medbay tending to the symptoms and using the dialysis machine. This change would also give medbay something to do with all of the bottles of penetic acid they keep making.

imawxome:
Also, if you use strong chemicals like prions or ricin, stopping purgatives won’t even matter, as you are already dead. it’s only a buff to the “weak” toxins that are never used due to how trivial they are to purge.

TDHooligan:
idk. i say just make the antihol reaction way less efficient (3 charcoal + 1 ethanol -> 2 antihol or something) 
then remove the other unlisted changes

imawxome:
I really didn’t want the recipe to be just ethanol and charcoal, I had initially chosen ipepec while I was talking to Leah about the change I wanted to make. Making people un-drunk in a single click felt weird, and ipepec never ever got used for anything, so we thought that it would be a good fit, especially how it makes ipepec decent at sabotaging charcoal treatments. I definitely should have also noted the lack of healing and color change, I had removed the healing because in my head, this is much more of a bootleg cure than an official un-drunk chemical

imawxome:
Also I should probably make it like 2 char 1 ipepec not 3 char, that’s a bit of a steep cost

imawxome:
Also, vomiting won’t purge your container, otherwise ricin would never do anything, and the people puking their guts out at the bar would get un-drunk after a little bit

imawxome:
Ethanol spiking poisons is actively detrimental, you clear all the ethanol extremely fast, and then you just have both Antihol (heals you) and charcoal, having the “poison spike chemical” actively cause the target to get healed more, while also getting purged away in seconds makes it very bad to use. Adding ethanol into someone who has been badly poisoned to get antihol to heal them more is a viable strategy

imawxome:
On top of that, the char+ethanol reaction is instant, so you can just dump a ton of char to instantly clear it all anyway

imawxome:
you can also get 15u char injectors from the medu-vendors, and robots have an infinite supply, along with sleepers. The goal of the PR is not to make med use a lot or resources to fix "spikings" it is to A: make un-drunking someone not instant, and B: give ipecac some use, as it is never, ever used. Having it be used for anthol solves them both, and makes spiking not pointless, as you cant instantly purge it all by just dumping in char. it also serves as an indirect buff for all the "Weak" chemicals that never get used because you can just purge them, now you need to use calomel or penetic acid, or just wait out the depletion and treat the symptoms in the meantime. In total, giving ipecac some use, buffing all the weak chemicals, and making undrunking someone not instant are all Pluses, and the reasoning behind the PR. (maybe not the last one).

MintyPhresh:
I think it's essential that antihol be at least possible to make in the bar, and changing the recipe to use _ipecac_ makes this far too difficult. Since this change doesn't affect the loadout of antitox medkits or do any mapping changes, there'll be no roundstart antihol in the bar, either. It's good to involve other departments in things like this, which is what I see with the choice of ipecac, but chemistry needs exclusives that help other people way more badly than pharmacy or botany do. I'd rather see it involve a very small amount of phenol or something, so that science feels comfortable sharing their 50u roundstart supply.

If antihol simply being too effective is your main concern, I think it'd be best to just have the purge rate scale with the amount of ethanol in your system, similar to how caffeine's depletion rate works. That way it doesn't get someone to insta-sober, but it'll get them out of the thresholds for severe ethanol poisoning quickly. Implementing this would preserve (and buff!) charcoal-cushioning, plus give antihol the nerf it's needed since #15065.

PR #20677, opened 2024-09-12
Barbery & dye rework

Glamurio:
[Rework] [Clothing]

## About the PR 
Refactors a bunch of parts in `barber.dm` and `barber_shop.dm` relating to cutting and dying of hair.

Most notably, adds a `dye` attribute to `/datum/customizationHolder` that supports holding a color for temporary application without having to override the base color.

Adds a `get_color_or_dye()` function which is used to prioritize fetching dye if able, otherwise defaulting to color.

Dye inside of hair can be removed by showering, restoring the original color.

Changes `get_barbery_conditions()` to default to 100% success chance, only lowering it if the user is jittery, blind, clumsy or drunk. It also now passes a list instead of merely a success chance, with the list containing the former and furthermore a list of conditions in string format for messages. Cutting hair on yourself is now (realistically) possible, but will now take significantly longer.

Adds a flavor list of adjectives to `/datum/action/bar/barber`, and also changes the cutting sound of razors to be more slice-y.

Adds `dye_wig()` to `/obj/item/clothing/head/wig` to add the ability to color wigs after they've been created.

Refactors `dye_hair()` in `/obj/item/dye_bottle` to reduce duplicate code. Adds the ability to recolor wigs that people are wearing.

## Why's this needed? 
Barbery has been in a weird spot for a while, being a daily job (save for Ozymandias) with very rare presence, yet having a dedicated system and location on every single station. We have some funny or cool hairstyles only accessible via barbery, yet the conditions for getting your hair cut are weirdly archaic and restrictive, checking for area and very particular conditions, up to and including doing a literal string comparison for the "Barber" job, which doesn't even include the "Hairdresser" special variant.

This "all or nothing" approach just meant that nobody really ever cut their hair, because there was an extremely high chance you'd mangle your face and lose it entirely. Now, it's more accessible, allows people to start RPing as a barber without the role spawn, and also allows the occasional person that feels strongly about their hair to fix it themselves.

As for dye currently just overrides hair color, which is problematic for a variety of reasons, various parts of the code, such as cloning, don't respect original color and simply copy over dyed hair. Now it works visually the same as before, but it washes off with showers, giving them a new use. Additionally, dyes can now be applied to wigs (though these do not wash out), which just seemed like a missing feature.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Barbery has received a rework and is now more accessible. Cut your hair today!
(*)Hair dye applied to hair is now impermanent and washes out if coming into contact with water.
(*)You can now apply hair dye to wigs! Wow.
```


code/datums/components/barber.dm
```diff
@@ -124,13 +124,13 @@ TYPEINFO_NEW(/datum/component/barber/shave)
 
 /datum/component/barber/proc/do_haircut(var/obj/item/thing, mob/living/carbon/human/M as mob, mob/living/carbon/human/user as mob)
 	if(!M || !user || (user.a_intent != INTENT_HELP && !thing.force_use_as_tool))
-		return 0 // Who's cutting whose hair, now?
+		return FALSE // Who's cutting whose hair, now?
 
 	var/non_murderous_failure = 0
 	var/mob/living/carbon/human/H = M
-	if(ishuman(M) && ((H.head && H.head.c_flags & COVERSEYES) || (H.wear_mask && H.wear_mask.c_flags & COVERSEYES) || (H.glasses && H.glasses.c_flags & COVERSEYES)))
+	if(ishuman(M) && H.head || ((H.head && H.head.c_flags & COVERSEYES) || (H.wear_mask && H.wear_mask.c_flags & COVERSEYES) || (H.glasses && H.glasses.c_flags & COVERSEYES)))
```
TobleroneSwordfish, line 131: This doesn't make sense, you're checking `if H.head exists` OR a more specific condition about H.head that will never be reached.

code/datums/components/barber.dm
```diff
@@ -758,17 +754,15 @@ ABSTRACT_TYPE(/datum/action/bar/barber)
...
-
 	getHairStyles()
 		return get_available_custom_style_types(filter_type=/datum/customization_style/hair)
 
 /datum/action/bar/barber/shave
 	cut = "shave"
 	cuts = "shaves"
 	cutting = "shaving"
+	hair = "face"
+	cut_sound = 'sound/impact_sounds/Flesh_Cut_1.ogg'
```
TobleroneSwordfish, line 765: That's definitely a "you fucked up and stabbed them" kind of sound, not a successful shave kind of sound.

code/modules/barber/barber_shop.dm
```diff
@@ -199,21 +238,141 @@
...
+			boutput(bottle_guy, "You fail to recolor [W]. Huh. How'd you manage that?")
+
+		if(!ishuman(target)) // if it's not a wig and not a human, we don't want it.
+			return FALSE
+
+		var/mob/living/carbon/human/target_guy = target
+		if(!target_guy?.organHolder?.head)
+			boutput(user, SPAN_ALERT("[target_guy] has no head, and you're all out of stump dye!"))
+			return FALSE
+		else
```
TobleroneSwordfish, line 288: Don't need to indent this in an else case, since it returns otherwise

code/modules/chemistry/Chemistry-Reagents.dm
```diff
@@ -160,6 +160,23 @@ datum
 							did_not_react = 0
 					if (ishuman(M))
 						var/mob/living/carbon/human/H = M
+
+						// Reset hair dye
+						var/had_dye = FALSE
```
TobleroneSwordfish, line 165: This will make hair dye wash out from stepping into a puddle of water or even blood with no shoes on, probably a little extreme anyway since water isn't known for removing hair dye. Hair dye removers are typically sulfur based according to wikipedia so maybe make sulfur wash off hair dye?

TobleroneSwordfish (changes requested):
Generally looks good, a couple of minor points. Sorry it took so long to get to.

Glamurio:
Still working on this, ended up going for a refactor that's taking too long and is not quite done.

Glamurio:
This is gonna get delayed a bit because of #21241

PR #20683, opened 2024-09-13
Restores the TEG to Cog1

Cherman0:
## About the PR 

image
Removes the engine randomization such that Cog1 always has a TEG again. The engine rando functionality is still maintained as it is used in devtest and theoretically could be used somewhere else.

## Why's this needed? 

The primary justification for [#14694](https://github.com/goonstation/goonstation/pull/14694) was that 90% of rounds were cog1 and thus it was always the same engine setup every shift. Since that isn't the case anymore (in my limited observation kondaru, nadir, and donut3 are all more popular now) I'd like to see cog1 return to the engine that it was originally mapped for.

Specifically, the other engines are extremely crammed in to fit in a 11x11 random room space, which leads to them being annoying, confusing, or hard to work with:
- The singularity setup (collectors and all) is compacted into a smaller space than just the singulo enclosure alone on the donut maps. This makes it quite difficult to do anything other than a basic small singulo. The surrounding area also lacks radiation shutters, meaning all of the adjacent rooms become lethal.
- The nuclear reactor struggles to route its pipes, needing to go through rwalls or under dense obstructing objects in some cases. This can make it difficult to track which pipes go where or to repair broken pipes. The lack of space between the reactor and turbine makes it quite difficult to build radiation shielding, and the lack of radiation shutters in the surrounding area makes this issue worse.

Adjusting the area to accomodate all of the engine types in a fun way would likely involve completely redoing engineering, mechlab, and a good chunk of the belt hell routing from scratch.

## Changelog 

```changelog
(u)Cherman0
(+)Cog1 has returned to being exclusively a TEG map.
```

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -21526,10 +21526,10 @@
 /obj/machinery/status_display{
 	pixel_y = 30
 	},
-/obj/landmark/engine_computer/two,
 /obj/cable/orange{
 	icon_state = "0-2"
 	},
+/obj/machinery/power/reactor_stats,
 /turf/simulated/floor/yellow/side{
 	dir = 1
 	},
@@ -23635,6 +23635,17 @@
 	},
 /turf/simulated/floor,
 /area/station/quartermaster/office)
+"bIJ" = (
+/obj/cable{
+	icon_state = "0-4"
+	},
+/obj/cable,
+/obj/mapping_helper/wingrille_spawn/auto/reinforced,
+/obj/cable{
+	icon_state = "0-8"
+	},
+/turf/simulated/floor/plating,
+/area/station/engine/core)
 "bIK" = (
 /obj/machinery/conveyor_switch{
 	id = "qmin";
@@ -32239,6 +32250,11 @@
 	dir = 4
 	},
 /area/station/hangar/science)
+"clO" = (
+/obj/machinery/atmospherics/pipe/simple/insulated,
+/obj/decal/stripe_delivery,
+/turf/simulated/floor/engine,
+/area/station/engine/core)
 "clP" = (
 /obj/cable,
 /obj/cable{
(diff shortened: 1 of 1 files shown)


JORJ949:
Cog1 is still the majority of rounds on classic from my experience and the random engine is a nice touch on cog, so im not a huge fan of this change.

DecidedlyUnethical:
I've always been of the opinion that random engine is a great idea in concept, but that it'd require a map specifically built with it in mind to truly work at its best. Cog1, while working for a proof of concept that it can work... is not that map. Forcing alternate engines into it when it was built around the TEG, hasn't been a fun experience for me.

Fully support this PR.

MyNameIsRomayne:
I feel like the reasoning for engine randomization to exist still holds -- I much prefer getting a random 1 of 3 engines on the most popular map (and potentially having a favorite/least favorite) over what is, admittedly, my least favorite, repeatedly. I'd much rather see an improvement on engine types and/or fitting each selected engine into the map rather than taking the 'easy way out' and scrapping it altogether -- especially because I still quite like this feature and see nothing wrong with it.
Might be good for a discussion thread, though?

Cherman0:
> Only issue I see with this approach is the effort it would take -- which honestly feels just like shutters, pipe reshuffles, etc... not the worst, really?

I feel that there's value in keeping important parts of the map consistent. It lets players know what theyre getting into when they sign up for a job on a map and helps to not have giant empty holes on the official goonhub maps. And it would also be a much larger mapping effort than that: The random room area has to be rectangular and they all need to be the same size, which limits how much you can expand it without significantly restructuring engineering. I quite like cog1 engineering and feel it would be a shame to genericize it for the sake of fitting engines that it wasn't really meant to fit. It would also necessitate either reprogramming the engine random room system or remapping devtest and all of its random engine modules.

Sord213:
I like the random engines. On classic the vast majority of rounds are still on cog1 so it's nice to get some different engines when I play engie. I'd rather increase the randomized space to be the whole of northern engineering so it's adapted to the engine type more than just flat out removal. Plus the very rare choose your own engine feels like a nice treat.

Cherman0:
It's too late sord, the people have already thumbs-upped the PR. You have to add it now. It says so in the bible.

Though in seriousness I'm not opposed to the random engine in concept, I just feel like the current implementation is not good. It might be a better fit for cog2, which both has a lot more space to work with and currently has a pretty bad TEG.

Cherman0:
My evil plan to submit PRs to remove things to convince other people to do the harder work of actually making them better continues to succeed. For hopefully obvious reasons this PR is mutually exclusive with [#20689](https://github.com/goonstation/goonstation/pull/20689).

PR #20689, opened 2024-09-13
New Cog1 Random Engines

Sord213:
## About the PR 
this is a concept of what we could be doing on cog1 instead of just cube engine and leaving the rest. Makes less sense from a lore standpoint....but also warping the engine in in general is weird lore wise but rules gameplay wise. This gives each engine type a unique layout within cog1's engineering. There's some jankiness because we don't have object passthrough apparently on dmm loading, IE not deleting what objects are there already when loading a DMM or maybe we just aren't using, it I'm not a coder. Notably the duplicated parts of the router, hangar, and PTL. 

This PR gets rid of the choose your own engine.

And it changes the weight of nuclear and singularity from 100->80.

Also without the engine cog1 looks kinda ugly i didn't feel like editing the landmark icon. sue me

StrongDMM-2024-09-13 14 19 23
StrongDMM-2024-09-13 14 19 29
StrongDMM-2024-09-13 14 19 17

## Changelog 

```changelog
(u)Sord
(*)Cogmap1's random engines now affect more of engineering. Woah.
(+)The Choose Your Own Engine engine on Cog1 has been removed.
```


ZeWaka:
> There's some jankiness because we don't have object passthrough apparently on dmm loading, IE not deleting what objects are there already when loading a DMM or maybe we just aren't using,

We do. Z5 asteroid prefabs use it.

Cherman0:
Looks neat, though some things I notice:
Nuke:
- Pglass windows on the left are wired to nothing
- Cold loop equipment adds a lot of volume to the nuke loop that doesn't benefit it. Simple fix would be to have a pump where the injection pipe meets the main nuke loop so that gas doesnt sit in that part.
- Fabricator by the nuclear nanofab would be nice to facilitate rockbox buying.
Singulo:
- Would be nice if the collectors were instead in that "break room" area, as it would be required to unwrench, drag them over, wire, and re-wrench them if you want to do a big singularity.
- The emitters and pylons being in that storage room seemds odd. Dragging them through the one-tile doorway will be annoying (especially if others are dragging the collectors out at the same time). Would be better to just have them already be in their "standard" positions.
- There doesnt seem to be a way to pull gas out of the cold loop? Being able to supercool gas is useful for the singularity. Though supercooled gas has the tendency to not flow very far (due to fuckery happening with a pressure very close to zero) so probably the cold loop should be kept very low-volume.
- You could use those perspective windooors and fences to allow the gas tanks to be accessible without being pulled by the singulo.
- The singulo prefab should have radsuit lockers.

Cherman0:
Oh also the singulo prefab needs those tank storage units for plasma tanks to go in the collectors. Donut3 has 30 tanks spread across 3 iirc, though it's just a vareditable number and can be whatever.

Sord213:
Fixed most of those. A couple things like the singularity only being sized for a standard 5x5 is intentional. If you want to make it bigger it's inherently less safe. "not designed for that". Same with donut2/donut3. You _can_ make a bigger singulo, but it's not what the space is designed for. Technically.

Also the singularity equipment room I like because it leaves a nice crime room you can use after for not murder.

Cherman0:
Mm, not a super big fan of that. For being "unsafe" I feel it would be better to emphasize the unsafety (ie cant manually start the pylons without risking getting hit by the containment field or breaking a wall, like donut2 and donut3 do) instead of just making it really annoying but perfectly safe.

pgmzeta:
sad to lose combustion chamber nuclear reactor

JORJ949:
image
10/10 PR

klookiyam:
One of the problems with the current singulo in cog1 is its really hard to add new collectors/controllers to the other sides of the singulo, not just because of the space limitations but you have to cut a lot of wiring and reroute it to avoid a hotwire.

Now that its a lot bigger, can I request rerouting the majority of the non-engine wiring out of the main chamber so that I can add more orange wiring for new collectors?

Cherman0:
I've thought about it more and I'm not really sure what you're meant to be able to do in this singularity setup. Setting a big singularity is (apparently intentionally) obstructed. This obstruction also (inadvertently?) makes it quite difficult to add more collectors. The reinforced floors makes it difficult to build a feeder pipe. Teleporters shouldnt be used for feeders in general because the reliance on the singulo gravity to pull items in makes a lag machine. The only thing I can think of that you actually can do without major resistance from the map design is supercool plasma cans.

As it stands, I don't think I can really support the singulo prefab, as it seems to actively not want you to engage with it. The nuke seems fine, aside from having annoyingly voluminous pipe extensions without valves and having a lot of somewhat awkward dead space around it. Maybe this space could be used for a decarabia-style reactor shield chamber.

JORJ949:
image
Which pipe does this meter measure on nuclear?
image
These platings should all be reinforced floors with decals on them preferably, so weaker singularities dont eat the floor.
image
Shouldnt this be a specific job spawn?
image
Bolted door should probably be changed to use the mappinghelper

PR #20692, opened 2024-09-14
Resprites Wizard Staff Inhands

MitchTheUser:
[GAME OBJECTS]

## About the PR 

Resprites the inhands for all the Wizard Staffs. They have been updated with higher detail and more accurate colors, and gaps/clipping in some directions/hands have been fixed or improved.

WizardStaff

## Why's this needed? 

Current Inhands use the colors and designs of the old staffs, and the new inhands better match the new staffs and have much more detail, as well as fixing how it's held in some directions.


ZeWaka (approved):

BatElite:
Any reason why the far inhands are lower than the front ones? that doesn't really match the from-above view

Think it'd be nice not to obscure the face as well.

MitchTheUser:
> Any reason why the far inhands are lower than the front ones? that doesn't really match the from-above view
> 
> Think it'd be nice not to obscure the face as well.

Inhands are now level with each other, and they are also now not obscuring the face as much.

WizardStaff

PR #20696, opened 2024-09-14
Shift atlas Armory and Kitchen a little

JORJ949:
## About the PR 
Moves around atlas kitchen andarmory a little and adds writing on an armory wall that says "BREACH IN CASE OF HUNGER", this results in any hungry security getting front row seats to the kitchen 
image

## Why's this needed? 
Funny small feature to add a little soul to the map and could create some fun moments

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -11502,6 +11502,10 @@
 /turf/simulated/floor/black,
 /area/station/security/brig)
 "cXf" = (
+/obj/lattice{
+	dir = 4;
+	icon_state = "lattice-dir"
+	},
 /obj/effects/background_objects/station/ss10,
 /turf/space,
 /area/space)
@@ -67357,7 +67361,7 @@ aaa
 aaa
 aax
 aaa
-aax
+cXf
 agl
 afS
 aAZ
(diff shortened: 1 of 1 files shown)


Cherman0:
I like it. Though i would probably swap the gibber and deep frier as it would be better to have the gibber out of sight from outside the kitchen.

JORJ949:
> I like it. Though i would probably swap the gibber and deep frier as it would be better to have the gibber out of sight from outside the kitchen.

thats the food processor, the gibber is in the freezer as per usual

MintyPhresh:
Goodbye kitchen bush... 😭

MintyPhresh:
I know it's a new map change, but the kitchen changes overall feel like a downgrade. Having the table space on the same wall as the ovens is something I try to avoid, since it tends to significantly increase the average distance from tables to ovens as well as making it slightly more difficult to navigate between the two since you can walk past it if you don't stop yourself. 
The chute and sink sharing a tile with equipment is icky, too. I'd much rather see the ovens and soup stove just shift down a tile rather than completely reworking the kitchen 4noraisin.

JORJ949:
Fun Idea, poor execution, may return to this eventually.

PR #20710, opened 2024-09-15
Transparency/Alpha show on Material Scans

444Portal:
## About the PR
shows alpha on material scans with extremely, very, transparent and opaque as the describers. same criteria used for the capgun is used to judge the alpha

## Why's this needed? 
now that alpha is used for something more than style i think we can put it in the scanners, will help builders and captains wanting to shoot twice the gun

## Changelog

```changelog
(u)444explorer
(+)Material Scanners now show the Materials Alpha/Transparency value
```


code/obj/item/device/pda2/scanners.dm
```diff
@@ -184,6 +184,15 @@
 				for(var/datum/material_property/mat in A.material.getMaterialProperties())
 					var/value = A.material.getProperty(mat.id)
 					. += "• [mat.getAdjective(A.material)] ([value])<br>"
+				switch(A.material.getAlpha())
```
ZeWaka, line 187: Why aren't you processing like the rest with get adjective?

444Portal: Err, good point now that you mention it - from my limited knowledge though that'd need a datum/property that stores alpha that i can stick getAdjective on - should I do that?

TobleroneSwordfish: Ideally yes, you could preserve `getAlpha` as a wrapper proc to access the new opacity property's value though.

code/modules/materials/Mat_Processing.dm
```diff
@@ -266,6 +266,15 @@ TYPEINFO(/obj/item/device/matanalyzer)
...
 						boutput(user, SPAN_NOTICE("• [X.getAdjective(W.material)] ([value])"))
+					switch(W.material.getAlpha())
+						if(-INFINITY to 80)
+							boutput(user, SPAN_NOTICE("• extremely transparent ([W.material.getAlpha()])"))
+						if(80 to 130)
+							boutput(user, SPAN_NOTICE("• very transparent ([W.material.getAlpha()])"))
+						if(130 to 180)
+							boutput(user, SPAN_NOTICE("• transparent ([W.material.getAlpha()])"))
+						if(180 to INFINITY)
+							boutput(user, SPAN_NOTICE("• opaque ([W.material.getAlpha()])"))
```
TobleroneSwordfish, line 277: Might read better if it went "opaque, slightly transparent, transparent, very transparent"

TobleroneSwordfish (changes requested):

PR #20727, opened 2024-09-17
Remove RD and MD inter-department radio access

JORJ949:
## About the PR 
Removes the RD's medical radio access
Removes the MD's science radio access

## Why's this needed? 
Consistency with other (HoS/CE) command headsets, they shouldn't have access to a radio channel outside of their department. I presume this is a remnant of when the RD was in charge of medbay but this just isnt the case anymore and it feels out of place with the relationship between medical and research now, they are different departments, with different heads of staff.

Im open to better descriptions for the headsets because im not a huge fan of these

## Changelog 

```changelog
(u)JORJ949
(+)The RD and MD no longer have access to each other's radio channels.
```

### DIFF
--- a/code/obj/item/device/radios/headsets.dm
+++ b/code/obj/item/device/radios/headsets.dm
@@ -234,31 +234,27 @@
 
 /obj/item/device/radio/headset/command/rd
 	name = "research director's headset"
-	desc = "This headset can receive on the Medical channel in addition to other secure frequencies. The 'sci' part of 'medsci'."
+	desc = "This headset seems to be covered in a large variety of burns."
 	secure_frequencies = list(
 		"h" = R_FREQ_COMMAND,
 		"r" = R_FREQ_RESEARCH,
-		"m" = R_FREQ_MEDICAL,
 		)
 	secure_classes = list(
 		"h" = RADIOCL_COMMAND,
 		"r" = RADIOCL_RESEARCH,
-		"m" = RADIOCL_MEDICAL,
 		)
 	icon_override = "rd"
 	icon_tooltip = "Research Director"
 
 /obj/item/device/radio/headset/command/md
 	name = "medical director's headset"
-	desc = "This headset can receive on the Research channel in addition to other secure frequencies. The 'med' part of 'medsci'."
+	desc = "Perfect for shouting at the Chief Engineer that the reactor is killing people."
 	secure_frequencies = list(
 		"h" = R_FREQ_COMMAND,
-		"r" = R_FREQ_RESEARCH,
 		"m" = R_FREQ_MEDICAL,
 		)
 	secure_classes = list(
 		"h" = RADIOCL_COMMAND,
-		"r" = RADIOCL_RESEARCH,
 		"m" = RADIOCL_MEDICAL,
 		)
 	icon_override = "md"


444Portal:
The CE argument fails because they aren't so closely related to Med and Sci, I don't see why they should lose the access

JORJ949:
> The CE argument fails because they aren't so closely related to Med and Sci, I don't see why they should lose the access

The medsci link is mostly chemistry, the sciengi link is with atmos, to me it feels odd that all other heads of staff just have access to their own department, while RD and MD have both.

cropseyy:
Genetics and science share budget I believe, and artifact lab and medical have a lot of overlap. It's not "just" chemistry by a long shot.

JORJ949:
> Genetics and science share budget I believe, and artifact lab and medical have a lot of overlap. It's not "just" chemistry by a long shot.

Science never actually uses the "science" budget, it may aswell be renamed to "genetics budget" because afaik thats all that use it. I also heavily disagree that artlab and medical have a lot of overlap, they have more overlap with cargo if anything as that is where most analysed artifacts go.

Sord213:
this is really probably just a holdover from when there wasn't a med director. RD used to be the head of medical as well kinda sorta technically and had access to their radio.

Connie-Dun:
I don't believe the gains in consistency outweigh the losses in encouraging interdepartmental interaction with this. I am well aware that the channels go largely unused anyways (at least on Classic), but I believe in an unfortunately hard-to-evaluate "signposting" that the channels being shared by the heads do. What I mean by signposting is, even if the channels don't actually get used, it helps prime players to think in terms of communicating with the counterpart department.
Just being a holdover isn't really a reason for removal, imo, just an explanation for why it is so in the first place. I don't know whether or not you intended to express it as a reason for removal, just want to cover my bases there.

Glamurio:
I actually would rather see doctors and scientists gain access to the respective radio. MedSci is in the past but it's still relevant in some places and frankly it'd be nice to see more of it, rather than removing what few remnants linger.

Cherman0:
I'd like to see medical and science have reduced overlap but I feel that this would need to be the last in a series of changes to that effect. Ideally I feel like cross-departmental comms should be done either on the general channel or in-person since that facilitates more interaction from bystanders.

Glamurio:
> I'd like to see medical and science have reduced overlap but I feel that this would need to be the last in a series of changes to that effect. Ideally I feel like cross-departmental comms should be done either on the general channel or in-person since that facilitates more interaction from bystanders.

General channel is already flooded with so much random noise, it's tough to filter out meaningful cross-departmental comms. That's why the go-to is still in-person. But this is a) really difficult / tedious on some maps because departments are not always adjacent (looking at you Donut2) and b) sometimes impossible because some maps have scientists hidden in nooks and crannies, impossible to approach without department access (Oshan)

cheekybrdy:
For myself, the need for medical communication is mainly for chemistry, which is taken internally by the pharmacy. Unless if pharmacy gets nerfed it doesn't really have any need. Most I've seen this used for is snooping on the other department's channel when there's a rivalry between the two anyway.

TobleroneSwordfish:
Seems broadly unpopular and I don't see a particular reason to change this.

PR #20761, opened 2024-09-19
Adds subtypes for closed blinds

MintyPhresh:
## About the PR
Evil mappers would be more than happy to do the varedits to have their blinds be closed at roundstart, but I will stop their wicked hand by  adding subtypes that do this. I also added blind switches that start turned off.

## Why's this needed? 
Varedits inferior. Coders superior.

### DIFF
--- a/code/obj/decorations.dm
+++ b/code/obj/decorations.dm
@@ -692,6 +692,20 @@ TYPEINFO(/obj/shrub/syndicateplant)
 		icon_state = "blindsH-R-o"
 		base_state = "blindsH-R"
 
+	closed
+		open = 0
+		opacity = 1
+		icon_state = "blindsH-c"
+		left
+			icon_state = "blindsH-L-c"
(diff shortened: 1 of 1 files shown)


Glamurio (approved):

code/obj/decorations.dm
```diff
@@ -780,21 +824,37 @@ TYPEINFO(/obj/shrub/syndicateplant)
 	name = "N blind switch"
 	dir = NORTH
 	pixel_y = 24
+
```
Tarmunora, line 827: This seems like an unnecessarily confusing naming convention

Tarmunora (changes requested):

code/obj/decorations.dm
MintyPhresh: I'm not sure why I did this. I guess I just assumed the way they start would always be the "on" position?? I even looked at this and said it was weird

PR #20763, opened 2024-09-19
Adds "Stinky" trait which reintroduces hygiene

Glamurio:
[Feature] [Traits]

## About the PR 
Adds `/datum/trait/stinky` as "Stinky", which makes you sweat a lot, and in turn, stinky.
Said trait adds the "Hygiene" sim to the user in exchange for +1 trait point.

"Filthy" debuff now gives a -5 max stamina modifier.
"Rancid" debuff now gives a -35 max stamina modifier.

Adds `removeMotive()` to `/datum/simsHolder` to allow easy removal for motives.

## Why's this needed? 
Despite widespread distaste for the hygiene feature, a few select individuals did like the idea of it, if only to have stinky lines on their characters. This allows people to still have that feature if they want to, and to incentivize usage, it now punishes users for not engaging with the mechanic, but not overly so.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Hygiene makes a comeback with the new "Stinky" trait.
```


MintyPhresh:
Already said this to you in a discord discussion, but repeating for discussion and posterity purposes: I think the max stamina debuff should be higher, maybe between -30 and -40. Plenty of other, more punishing traits still see use (Athletic and Frail Constitution come to mind), and I think it's important that the debuff be significant to discourage the "Gross & Stinky Bastard" gimmick character from becoming a commonality.

Filthy probably shouldn't have a debuff, though, IMO; Back before hygiene had a concrete penalty, Filthy acted as a sort of "fairness timer" so you wouldn't get instantly penalized for slipping into a pile of blood. If we were to get instantly penalized like that, I think it would end up being frustrating mechanically.

I can't wait to see this trait added!

Glamurio:
> Already said this to you in a discord discussion, but repeating for discussion and posterity purposes: I think the max stamina debuff should be higher, maybe between -30 and -40. Plenty of other, more punishing traits still see use (Athletic and Frail Constitution come to mind), and I think it's important that the debuff be significant to discourage the "Gross & Stinky Bastard" gimmick character from becoming a commonality.
> 
> Filthy probably shouldn't have a debuff, though, IMO; Back before hygiene had a concrete penalty, Filthy acted as a sort of "fairness timer" so you wouldn't get instantly penalized for slipping into a pile of blood. If we were to get instantly penalized like that, I think it would end up being frustrating mechanically.
> 
> I can't wait to see this trait added!

Completely fine with increased debuffs. Mechanically I'd like to separate the types of dirt that can affect hygiene, so this would be less of an issue, but even still, I don't personally think a -10 stamina debuff is too debilitating if you happen to get covered in blood. It's a negative trait for a reason, plenty of other traits can negatively affect you against your will because you picked them. Clown antag when you can't see them, dying early with puritan etc.

cropseyy:
Honestly I'd be down for doubling the penalty too, but maybe worth testmerging to see.

code/datums/sims.dm
```diff
@@ -442,12 +442,7 @@
 	var/list/motives = list()
 	var/list/datum/simsHolder/simsHolders = list()
 	var/list/datum/simsMotive/simsMotives = list()
-
-#ifdef RP_MODE
-	var/provide_plumbobs = 0
-#else
-	var/provide_plumbobs = 1
-#endif
+	var/provide_plumbobs = 0 // I don't think this is still relevant? Maybe for gimmicks
```
Tarmunora, line 445: relevant for togglesimsmode() proc

Tarmunora (changes requested):

code/datums/sims.dm
Glamurio: The issue I kept running into is that when I added hygiene to a classic character, it would also enable plumbobs, hence why I set them to 0, because "sims" as we call them have kind of moved away from its origins, I'd wager.. Unless you mean I should just remove the comment. Also, I can't actually see plumbobs being mentioned in the proc, though maybe it's secret and / or I'm overlooking stuff. Isn't the enabling of plumbobs a toggle for admins separately?

Glamurio:
> Honestly I'd be down for doubling the penalty too, but maybe worth testmerging to see.

Changed Filthy to -5 and Rancid to -35 max stamina.

Glamurio:
God dammit.

code/datums/sims.dm
```diff
@@ -442,12 +442,7 @@
...
 	var/list/datum/simsMotive/simsMotives = list()
-
-#ifdef RP_MODE
 	var/provide_plumbobs = 0
-#else
-	var/provide_plumbobs = 1
-#endif
 
 	New()
 		..()
```
Tarmunora, line 448: if globally disabling plumbobs like this, `toggle_sims_mode` verb should re-enable them

flrsh (commented):
I am someone who doesn't like motives in general, so I also don't think this is very interesting or worth a point as a trait. I think the UI element makes this seem far more important than it actually is. If people want stinky lines, we could add it to the regular `stinky` bioeffect?

Glamurio:
> I am someone who doesn't like motives in general, so I also don't think this is very interesting or worth a point as a trait. I think the UI element makes this seem far more important than it actually is. If people want stinky lines, we could add it to the regular `stinky` bioeffect?

I think this goes beyond just stink lines. Motives are fun if they're something you can choose, I think. I personally quite like having to manage systems, I just don't like having to do it all the time. Putting it in a trait seems sensible to me because it puts the burden onto the person choosing the trait.

Tarmunora (approved):

PR #20765, opened 2024-09-19
Updates Donut2's Catering

MintyPhresh:
## About the PR
image

- Adds a loafer, chicken disposal, bar dispenser, and more to the area.
- Complete layout overhaul of kitchen and bar

## Why's this needed? 
I had a bartender round on this map directly after having one on Kondaru, and was quite distraught with the contrast between the bar's facilities. The drink dispensers were still inconveniently split between alcohol and non-alcohol, there was no place to store my cut wedges and boxes of umbrellas, I had no easy access to a trash can without leaving the counter, and the bar itself felt very cramped thanks to the one-tile hallways that connect the 8-tile working space to the tiny bar office and cafeteria.

Further investigations and discussions with others revealed that the map also lacked a loafer and a chute for carrying chickens to the freezer, staples on more modern maps.

Modernizing this map should make it a lot more streamlined and enjoyable to be around, be you a patron or a caterer.

## Anticipated points of contention

- Guardbuddy exile
-- I'm generally a guardbuddy hater, but these guys take the cake! They don't use any (or don't have?) patrol routes or chargers on the main station, so they just take up space and make noise until their battery runs out halfway through the shift. I've exiled them to the crew quarters just in case someone makes them functional again, but I don't like 'em in the bar driving me cray-cray (crazy)! Lazy bums!


Sord213:
Definitely too gaudy. It works for a purpose built aesthetic map like decarabia but not for the vibes I'm going for on donut 2. Redesign otherwise tho I don't mind. It was definitely a weaker area on donut 2.

MintyPhresh:
Gotcha, will revert to the old tileset

MintyPhresh:
Donut2 is somewhat exceptional, I think. Since Chefs and bartenders don't have access to maintenance, they can't get into the gen-fab at tool storage, and have to traverse literally the entire station to get the vending machine module. The rest of the stuff can be scrounged from elsewhere (which is why I added the TV).

TobleroneSwordfish (changes requested):
Oh this looks great actually, not sure what the player vending machine board is there for though? Also personal request to remove this door or replace it with a serving hatch because otherwise this bar is going to be tided constantly:
image

MintyPhresh:
There any way we could keep that as a point of egress for the bartender? Better visibility, perhaps? I want them to go out into the world and make friends instead of staying cooped up in their room all the time

MintyPhresh:
@TobleroneSwordfish Sorry to double post, but what could be done to reduce tideability without removing the door? I'm not super familiar with tiding in general.

TobleroneSwordfish:
Actually thinking about it most bars are like that so I'm just going to merge it. I also removed maint access from tool storage so I'll remove the player vending machine cartridge since I think that's why that was there?

MintyPhresh:
Oh, no, that was there to encourage players to use it. I rarely see anyone use the vending machines and thought it'd be a good thing if I encouraged players to consider a new way to set out food.

PR #20770, opened 2024-09-20
Cleaning up some of the hardcoded lists to much

Loaded4Bears:
cleaner concrete_typesof

Attempting to clean up some hardcoded lists that can be easily simplified, expanded the cola and booze spawner, fixed icon for the pizza and cola spawners.

## Why's this needed? 

Toooo many hardcoded lists

### DIFF
--- a/code/obj/random_spawners.dm
+++ b/code/obj/random_spawners.dm
@@ -1254,9 +1254,10 @@ git
 	icon_state = "rand_btool"
 	min_amt2spawn = 1
 	max_amt2spawn = 2
-	New()
-		items2spawn = concrete_typesof(/obj/item/device/guardbot_tool)
-		. = ..()
+	items2spawn = list (/obj/item/device/guardbot_tool/medicator,
+						/obj/item/device/guardbot_tool/smoker,
+						/obj/item/device/guardbot_tool/taser,
+						/obj/item/device/guardbot_tool/flash)
 
 	one
 		amt2spawn = 1
--- a/code/obj/random_spawners.dm
+++ b/code/obj/random_spawners.dm
@@ -1,4 +1,4 @@
-
+git
 /obj/random_item_spawner
 	name = "random item spawner"
 	icon = 'icons/obj/objects.dmi'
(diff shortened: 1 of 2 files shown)


code/obj/random_spawners.dm
```diff
@@ -1925,9 +1866,7 @@
 	//only 1 can spawn for now since the pool size is small. Might want to increase it if the pool size increases by a fair amount
 
 	items2spawn = list(/obj/item/aiModule/experimental/equality/a,
-						/obj/item/aiModule/experimental/equality/b,
-						/obj/item/aiModule/experimental/corrupted,
-						/obj/item/aiModule/experimental/historic)
```
JORJ949, line 1930: Outdated.

JORJ949 (changes requested):
If I understand concrete_typesof properly then this removes and adds stuff to these spawners, such as syndicate tables, old medkits and various other things. Probably best leaving these as lists so you dont end up with /obj/item/clothing/mask/gas/swat/syndicate in an otherwise clean random gimmick mask spawner, or any /obj/item/reagent_containers/food/drinks/bottle/poison_that_instagibs_everything (example) in a random booze spawner.

Loaded4Bears (commented):
Reverted the first aid kits, masks, shoes, guardbot tools. 

Mines should be ok, pizza looks ok, the 'bad' booze is under /food/drinks and the 'regular shelf' stuff is under food/drinks/bottle.

Furniture parts and gimmick clothing doesn't seem to have anything traitor/secret/OP

Glamurio (changes requested):
Always happy for refactoring, however this isn't a simple refactor, you're changing a lot of core functionality of some spawners. For some of them that's fine, but for some others that would really change their function.

Also, `SpaceStop.dmm` is a new prefab, please atomize this PR and make a separate one for the map or the cleanup, but not both at once.

JORJ949:
> 
> Furniture parts and gimmick clothing doesn't seem to have anything traitor/secret/OP

You should also consider future items that could be added there, I think the set list or a var as to "do i spawn in random spawner" (like how surplus crates work) would be better than just anything of this type.

Loaded4Bears (commented):
didn't mean to include the new map, i'll close this entire pull and come back for a  2nd try.

PR #20772, opened 2024-09-20
Moves Cog1 Catering Announcement Computer

MintyPhresh:
## About the PR 
Moves the announcement computer from the bar counter to catering storage. Check the mapdiff under "Checks" to see the specific new location.

## Why's this needed? 
Bartender couldn't reach the north 3 tables since the computer was in the way.


ZeWaka:
This does make it harder to break in

JORJ949:
> This does make it harder to break in

Theres a window on the south entrance cut off this image, so its not much of a change.

PR #20774, opened 2024-09-20
Buffs syndicate command armour slightly

TobleroneSwordfish:
[BALANCE]
## About the PR 
Gives syndicate command armour flash protection on the helmet and some disorient resist on the body, enough to take an extra baton hit.

## Why's this needed? 
For a 5TC item that looks like it should make you a tank this often feels like a noob trap and isn't actually very threatening. Giving it a little better resilience against some stuns might make it more worth using.

## Changelog 

```changelog
(u)LeahTheTech
(*)Syndicate command armour now offers eye protection on the helmet and some disorient resistance on the body (enough to take an extra baton hit)
```

### DIFF
--- a/code/obj/item/clothing/helmets.dm
+++ b/code/obj/item/clothing/helmets.dm
@@ -1062,6 +1062,7 @@ TYPEINFO(/obj/item/clothing/head/helmet/space/industrial)
 		..()
 		setProperty("meleeprot_head", 7)
 		setProperty("space_movespeed", 0)
+		setProperty("disorient_resist_eye", 100)
 
 TYPEINFO(/obj/item/clothing/head/helmet/space/industrial/salvager)
 	mats = list("metal_superdense" = 20,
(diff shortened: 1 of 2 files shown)


JORJ949:
This thing isnt actually available to traitors for 5TC, just spiefs. The most common obtainment method is 25k a piece at C.A.R.L.

JORJ949:
This thing isnt actually available to traitors for 5TC, just surpluses and maybe spiefs?. The most common obtainment method is 25k a piece at C.A.R.L. So the 5TC rarely comes into play. 

I dont personally think the helmet should have 100% eyeprot, you should need to steal sechuds/sunglasses for that (maybe stick sunglasses in the syndie vendor for like 1-2k each??). 

Its also important to note this is almost always combod with the strong explosives CARL sells, making stunning basically impossible as you explode before stunning, and this makes that worse.

frawhst (approved):

Tarmunora:
> This thing isnt actually available to traitors for 5TC, just surpluses and maybe spiefs?. The most common obtainment method is 25k a piece at C.A.R.L. So the 5TC rarely comes into play.

In the context of it being purchasable from the listening post, I'm not sure it really needs these buffs, yeah.

PR #20775, opened 2024-09-20
Adds cinnamon to 3 drinks that typically contain it

Glamurio:
[QoL] [Catering]

## About the PR 
Adds `cinnamon` requirement to "Spiced Rum", "Hot Toddy", and "Pumpkin Spiced Latte".

Also adds a "fake" version of "Spiced Rum" that's still made with capsaicin.

## Why's this needed? 
These recipes are typically defined by their use of spice, particularly cinnamon, and were largely added before cinnamon existed. This just makes them a bit more accurate, and also adds more reasons to make cinnamon. Since we're approaching the fall season, this seemed like good timing.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)"Spiced Rum", "Hot Toddy", and "Pumpkin Spiced Latte" are now made with cinnamon.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -2055,8 +2055,8 @@
 		name = "Hot Toddy"
 		id = "hottoddy"
 		result = "hottoddy"
-		required_reagents = list("sweet_tea" = 1, "bourbon" = 1, "juice_lemon" = 1)
-		result_amount = 3
+		required_reagents = list("sweet_tea" = 1, "bourbon" = 1, "juice_lemon" = 1, "cinnamon" = 1)
+		result_amount = 4
 		mix_phrase = "The drink suddenly fills the room with a festive aroma."
 		mix_sound = 'sound/misc/drinkfizz.ogg'
 		drinkrecipe = TRUE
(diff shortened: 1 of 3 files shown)


code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -2085,12 +2085,18 @@
 		name = "Spiced Rum"
 		id = "spicedrum"
 		result = "spicedrum"
-		required_reagents = list("rum" = 1, "capsaicin" = 1)
+		required_reagents = list("rum" = 1, "cinnamon" = 1)
 		result_amount = 2
-		mix_phrase = "You feel like you might have misunderstood the recipe."
```
ZeWaka, line 2090: Eh, this is kind of a funny joke. Idk if we want a weird fake version idk

Glamurio: The fake version doesn't actually act any different, it just has the same old mix phrase it used to have before. This mostly just lets you make spiced rum with cinnamon and capsaicin. I made it a separate chem for fun future interactions, but if preferred they can both just result in the same chem.

jan-antilles:
This does make Hot Toddy locked behind Botany participation, for what that's worth. The rest already require Botany ingredients.

flrsh (approved):

PR #20788, opened 2024-09-22
Adds glass floors to the observatory in COG1

Laboredih123:
## About the PR 
image

## Why's this needed? 
It looks neat

## Changelog 

```changelog
(u)cringe
(+)Nanotrasen put some new flooring in the cogmap observatory.
```


Laboredih123:
why is it shattering, its *meant* to have underlays

Laboredih123:
otherwise how would the effect work in non-parallax mode?

itchybones:
I think the cl entry should specify that it's on cog1, since most people's minds would probably jump to the telesci place

Laboredih123:
done

PR #20797, opened 2024-09-22
Resprites Archery Stuff

Flaborized:
## About the PR 

This PR adds in some resprites for bows, arrows, arrowheads and quivers. 

bow

showcasemorelikebowcase

wooden bows automatically use the palette-d brown sprite instead of applying a filter, cuz I can. Note that arrows in quivers don't use their own material properties which I did not fix because I don't want to, so they just kinda look weird (which they did before)

## Why's this needed? 

Bow stuff is a bit janky and old. I drew nice sprites for them though. Also the little brown junk next to the quivers in the image is their arrow total being displayed outside of UI (and also taking on material appearances because of course they do). I'll make a bug report for that. Also this thickens bows up a bit so they're easier to click on. Currently it kinda sucks to click on them because they're so painfully thin.


Glamurio:
Can we pretty please get a matching back-sprite to go along with these? :)

Flaborized:
> Can we pretty please get a matching back-sprite to go along with these? :)

quiver_worn

yea here

Glamurio (approved):
Personally would love if the whole arrow-material thing was included in this PR but it's also true than this is so much better than we have already, should be fine to do it in a follow-up PR.

PR #20814, opened 2024-09-23
Enabling strictNullChecks

Garash2k:
## About the PR 
Enables `strictNullChecks` and fixes the code so it passes.

<details><summary>Addressed draft points</summary>

- [x] Actually test most of the stuff
- [x] An error in `components/goonstation/ListGrid/Cell.tsx`
- [x] Errors in `interfaces/PlayerPanel/index.tsx`
- [x] Consider not allowing undefined on PollEditorPanel SavePayload
- [x] Find a solution for BuyMaterialModal that ~~isn't null = closed.~~ is less ambiguous
- [x] Other discord comments

</details>

This is #20748, but cherry picked to a master branch.

## Why's this needed? 
strictNullChecks yippie

[ui][internal][code quality]


tgui/packages/tgui/interfaces/GeneTek/index.tsx
```diff
@@ -103,56 +103,64 @@ export const GeneTek = () => {
...
-                  value={r.total - r.current}
+                  value={r.total && r.current ? r.total - r.current : 0}
                   maxValue={r.total}
                   mb={1}
                 >
                   <Box position="absolute">{r.name}</Box>
-                  <TimeDisplay auto value={r.current} format={formatSeconds} />
+                  <TimeDisplay
+                    auto
+                    value={r.current || 0}
```
mordent-goonstation, line 161:
```suggestion
                    value={r.current ?? 0}
```

tgui/packages/tgui/interfaces/GeneTek/modals/BuyMaterialsModal.tsx
```diff
@@ -17,7 +17,7 @@ export const BuyMaterialsModal = (props) => {
   const maxBuyMats = props.maxAmount;
   const { budget, costPerMaterial } = data;
 
-  const resolvedBuyMats = Math.min(buyMats, maxBuyMats);
+  const resolvedBuyMats = buyMats !== null ? Math.min(buyMats, maxBuyMats) : 1;
```
mordent-goonstation, line 20: Can't easily leave as a diff comment, but:

Set `buyMats` to default to `1` instead of `null` if that's what's going to be used anyway, then the `useSharedState` would infer `number` and this line wouldn't need changing.

ZeWaka (approved):

tgui/packages/tgui/interfaces/GeneTek/tabs/ResearchTab.tsx
```diff
@@ -84,24 +84,28 @@ export const ResearchTab = (props) => {
...
-                        icon="flask"
-                        disabled={materialCur < r.cost}
-                        onClick={() => act('research', { ref: r.ref })}
-                        color="teal"
+              ? ar.map(
+                  (r) =>
+                    r.ref &&
+                    r.cost && (
+                      <Section
+                        key={r.ref}
```
mordent-goonstation, line 92: Sanity check on the `.map` returning a keyed component sometimes, and presumably `false` or `null` in others, when there's also a `.length` check earlier?

Would it be better to `.filter`, then use that for the `.length` check and then the `.map` as you have it?

tgui/packages/tgui/interfaces/GeneTek/tabs/ResearchTab.tsx
```diff
@@ -110,14 +114,18 @@ export const ResearchTab = (props) => {
...
-                  >
-                    <Description text={researchTree[r.ref].desc} />
-                  </Section>
-                ))
+              ? fr.map(
+                  (r) =>
+                    r.ref &&
+                    r.cost && (
+                      <Section
+                        key={researchTree[r.ref].name}
```
mordent-goonstation, line 122: Same as other comment re: `.map` and `false`/`null`.

tgui/packages/tgui/interfaces/ListInputWindow/ListInputModal.tsx
```diff
@@ -95,6 +95,9 @@ export const ListInputModal = (props: ListInputModalProps) => {
   // |goonstation-change| send any text input to the search bar
   const onFocusSearch = (letter) => {
     let searchBarInput = getSearchBar();
```
mordent-goonstation, line 97: Not looked at surrounding code etc., but I *suspect* this can be made to use a `ref` instead. Not requiring for this PR, as out of scope. Your change is fine.

tgui/packages/tgui/interfaces/MixingDesk.tsx
```diff
@@ -32,7 +32,10 @@ interface MixingDeskData {
 export const MixingDesk = () => {
   const { act, data } = useBackend<MixingDeskData>();
   const { voices, selected_voice, say_popup } = data;
-  const [message, setMessage] = useSharedState<string | null>('message', null);
+  const [message, setMessage] = useSharedState<string | undefined>(
+    'message',
+    undefined,
+  );
```
mordent-goonstation, line 38: Change is fine. We could very easily default this to an empty string, though.

tgui/packages/tgui/interfaces/PaperSheet/Stamps.tsx
```diff
@@ -23,8 +23,11 @@ export const PaperSheetStamper: React.FC<PaperSheetStamperProps> = ({
   const [rotate, setRotate] = useState(0);
 
   const findStampPosition = (e: MouseEvent) => {
-    let rotating: boolean;
+    let rotating = false;
     const windowRef = document.querySelector('.Layout__content');
+    if (!windowRef) {
+      return;
+    }
```
mordent-goonstation, line 30: Stick this before `let rotating = false;`, as that's not required for the early return.

tgui/packages/tgui/interfaces/PlayerPanel/index.tsx
```diff
@@ -229,12 +232,13 @@ export const PlayerPanel = () => {
...
-                        row: player,
-                        value: playerValues[ckey][id],
-                      })}
+                      {template &&
+                        template({
+                          act,
+                          column,
+                          row: player,
+                          value: playerValues[ckey][id],
+                        })}
```
mordent-goonstation, line 241: Possibly just `{template?.({ ... })}`

tgui/packages/tgui/stories/Themes.stories.tsx
```diff
@@ -6,23 +6,26 @@
...
   render: () => <Story />,
 };
 
-const Story = (props) => {
-  const [theme, setTheme] = useLocalState('kitchenSinkTheme', undefined);
+const Story = () => {
+  const [theme, setTheme] = useSharedState<string | undefined>(
+    'kitchenSinkTheme',
+    undefined,
+  );
```
mordent-goonstation, line 20: I suspect `useState` would be fine here as well. If there's a reason it needs to be shared lmk.

mordent-goonstation (changes requested):
Neat. Code suggestions I put in are suggestions only, I've not tried to use them. Entirely possible I may be wrong with them and that our gradually-getting-more-shouty tooling will say otherwise.

tgui/packages/tgui/debug/KitchenSink.tsx
```diff
@@ -23,7 +24,10 @@ const getStories = () => r.keys().map((path) => r(path));
 
 export const KitchenSink = (props) => {
   const { panel } = props;
-  const [theme] = useState(null);
+  const [theme] = useSharedState<string | undefined>(
```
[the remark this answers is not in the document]
Garash2k: The theme story is currently broken on master and upstream. `theme` has to be set in `Themes.stories.tsx` but used in `KitchenSink.tsx`. Shared state is basically how it was able to be done pre-react.

tgui/packages/tgui/stories/Themes.stories.tsx
Garash2k: The theme story is currently broken on master and upstream. `theme` has to be set in `Themes.stories.tsx` but used in `KitchenSink.tsx`. Shared state is basically how it was able to be done pre-react.

tgui/packages/tgui/interfaces/GeneTek/BioEffect.tsx
```diff
@@ -255,7 +261,8 @@ export const BioEffect = (props) => {
             <Button
               icon="radiation"
               disabled={
-                onCooldown(equipmentCooldown, 'Emitter') || subject.stat > 0
+                onCooldown(equipmentCooldown, 'Emitter') ||
+                (subject && subject.stat > 0)
```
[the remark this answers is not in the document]
Garash2k: tooling no likey
image

tgui/packages/tgui/interfaces/PlayerPanel/index.tsx
Garash2k: ‼️

tgui/packages/tgui/interfaces/ListInputWindow/ListInputModal.tsx
Garash2k: ref would be great, but I don't think there's an easy way to get a tgui-core's input ref at the moment

image

tgui/packages/tgui/interfaces/GeneTek/modals/BuyMaterialsModal.tsx
Garash2k: The problem is that, in `GeneTek.tsx`, the buyMats state is also used to determine whether the popup is open or not,

Later on in BuyMaterialsModal, buyMats is set to null to close the popup.

Oh! Just had the idea of using 0 instead of null, that should solve everything.

tgui/packages/tgui/interfaces/GeneTek/BioEffect.tsx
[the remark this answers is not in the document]
mordent-goonstation: Kinda silly given that it's a valid comparison, but sure.

tgui/packages/tgui/interfaces/GeneTek/type.ts
```diff
@@ -38,7 +38,7 @@ export interface GeneTekData {
 }
 
 interface ResearchEntry {
-  ref?: Ref;
+  ref: Ref;
   name: string;
   desc: string;
   cost?: number;
```
mordent-goonstation, line 44: If "the cost property is always set for available researches", per comment on ResearchTab.tsx:94, shouldn't this be `cost: number`? If so, can change the type and lose the comment on that line.

tgui/packages/tgui/interfaces/ListInputWindow/ListInputModal.tsx
```diff
@@ -53,7 +53,7 @@ export const ListInputModal = (props: ListInputModalProps) => {
 
   const getSearchBar = () => {
     return searchBarVisible
-      ? document.getElementById('search_bar').getElementsByTagName('input')[0]
+      ? document.getElementById('search_bar')?.getElementsByTagName('input')[0]
       : null;
   };
```
mordent-goonstation, line 58: Minor, but `getElementsByTagName('input')[0]` returns `undefined` if no matches, rather than null.

Both are valid, of course, but prefer to stick to one "not found" return type if you can rather than the return type implicitly being `null | undefined`.

Consider: `return (searchBarVisible && document.getElementById('search_bar')?.getElementsByTagName('input')[0]) || null;`

mordent-goonstation (changes requested):
Second pass, I think this is it. I would look to rebase your PR onto `master`, rebundle, and check the change doesn't throw any more errors with any other tgui changes made since this PR. Ping me in Discord when done and I'll speedmerge.

tgui/packages/tgui/interfaces/GeneTek/type.ts
Garash2k: No, because `ReasearchEntry`s are also used for completed researches, which don't have `cost`.

PR #20815, opened 2024-09-23
Move Some Plantmaster Data to Regular Data

Garash2k:
## About the PR 
Makes it so that *some* of the Plantmaster data is passed as regular data in lieu of static data.
The seed list and the extractable list remains as static data.
Regular data:
image
Static data:
image

This is a different take on #20607

## Why's this needed? 
Static data wasn't appropriate for some the data. Static data is appropriate for "large data that should be sent infrequently." Many properties were small.
This has a nice upside of not needing to recalculate the large lists when: shuffling beakers around, sorting the list or toggling the Output Internally button.

[UI][Performance]

### DIFF
--- a/code/obj/submachine/seed.dm
+++ b/code/obj/submachine/seed.dm
@@ -493,7 +493,7 @@ TYPEINFO(/obj/submachine/seed_manipulator)
 				qdel(seed1)
 				qdel(seed2)
 				src.mode = "seedlist"
-				. = TRUE
+				update_static_data(ui.user, ui)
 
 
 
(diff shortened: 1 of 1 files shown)


mordent-goonstation:
There's probably some benefit to caching some of the data rather than simply deciding to regenerate at a certain time, but that comes with all of the fun cache validation problems.

I assume there's a performance reason for this change? Can you give an example of before vs. after, if so?

Garash2k:
It's not a performance metric per say, but the actions perform much better when they don't require static data updates.
In this first example, I'm spamming the new actions that don't require static updates, they work great

https://github.com/user-attachments/assets/eea026f0-d50d-4ec5-84d5-0d56f35a2f32

In this second example, I'm spamming actions that do require static updates. They are visibly super delayed. The previous example's actions would be just as delayed if they were static data.

https://github.com/user-attachments/assets/ef43789d-fec4-48c3-b2f1-906d672cf3e4

mordent-goonstation (approved):

PR #20826, opened 2024-09-23
Access implant for admin mice

flappybatpal:
## About the PR 
Gives admin mice a new infinite implant equivalent to Captain + HoS + Centcom.
As far as I can see in testing implants don't drop and can't be removed from critters.

## Why's this needed? 
Admin mice couldn't interact with everything, particularly consoles.
They still have issues with things requiring ID cards but such is life when you are a small amaranth rodent.

### DIFF
--- a/code/obj/item/implant.dm
+++ b/code/obj/item/implant.dm
@@ -1710,11 +1710,10 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
 				..()
 				access.access = get_access("Chef")
 
-		head_of_security
-
+		admin_mouse
 			New()
 				..()
-				access.access = get_access("Head of Security")
+				access.access = get_access("Captain") + get_access("Head of Security") + access_centcom
 
 
 /* ============================================================ */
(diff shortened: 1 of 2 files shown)


JORJ949:
Reminder that HoS doesnt have AA anymore, so for example no interacting with QM consoles.

flappybatpal:
> Reminder that HoS doesnt have AA anymore, so for example no interacting with QM consoles.

Stupid HoS!

code/obj/item/implant.dm
```diff
@@ -1710,6 +1710,11 @@ ABSTRACT_TYPE(/obj/item/implant/revenge)
 				..()
 				access.access = get_access("Chef")
 
+		admin_mouse
+			New()
+				..()
+				access.access = get_access("Captain") + get_access("Head of Security") + access_centcom
```
ZeWaka, line 1716:
```suggestion
				access.access = get_access("Admin")
```
https://github.com/goonstation/goonstation/commit/cb56bec537e2338001f4f6a5ff7f33a2ff86d5b5

ZeWaka (changes requested):

PR #20856, opened 2024-09-25
Extends the bartender's alcohol intuition

MintyPhresh:
## About the PR 
Whenever a bartender uses any of the scanning items I could think of (reagent scanner, spectroscopic goggles, PDA app), they will do quick mental math to estimate how much ethanol is in the container. Pictured below is a 1:1 mix of Screwdriver and Martini in a cocktail glass:
image

Additionally, whenever a bartender tastes a drink, their keen sense of taste will give them an estimate of the drink's composition, in percent ethanol by volume (AKA %ABV). Here's how that same mixture tastes to the trained tastes of a bartender:
image

Finally, everyone can get some idea of the level of pure ethanol in someone's body just by looking at them. Here's what someone with 50u of ethanol looks like:
image

This PR also implements Bartender Training, which determines if someone can twirl drinks without spilling it, affects bottle shatter success chance, and affects access to the bartender abilities listed above.

## Why's this needed? 
A bartender's job is to provide drinks to a customer, and ostensibly to ensure their customers don't get drunker than they want to from their drinks. This change gives relevant information to the player that can help them achieve the latter goal.

## Changelog 
```changelog
(u)Mintyphresh
(+)Bartenders can now intuit much more about the alcohol contents of drinks.
```

### DIFF
--- a/code/mob/living/carbon/human/procs/get_desc.dm
+++ b/code/mob/living/carbon/human/procs/get_desc.dm
@@ -31,8 +31,8 @@
 			. = "<br>[src.bioHolder.mobAppearance.flavor_text]"
 
 	// crappy hack because you can't do \his[src] etc
-	var/t_his  = his_or_her(src)
-	var/t_him  = him_or_her(src)
+	var/t_his = his_or_her(src)
+	var/t_him = him_or_her(src)
 
 	. +=  "<br>[SPAN_NOTICE("*---------*")]"
 	var/datum/ailment_data/found = src.find_ailment_by_type(/datum/ailment/disability/memetic_madness)
@@ -414,8 +414,6 @@
 		var/et_amt = src.reagents.get_reagent_amount("ethanol")
 		var/drunk_assess = ""
 		if (!isalcoholresistant(src) || src.reagents.has_reagent("moonshine"))
-			if (src.reagents.has_reagent("moonshine"))
-				et_amt *= 7
 			switch (et_amt)
 				if (0 to 10)
 					drunk_assess = "[capitalize("[he_or_she(src)]")] seem[blank_or_s(src)] <b>buzzed.</b>"
--- a/code/mob/living/carbon/human/procs/get_desc.dm
+++ b/code/mob/living/carbon/human/procs/get_desc.dm
@@ -31,8 +31,8 @@
 			. = "<br>[src.bioHolder.mobAppearance.flavor_text]"
 
 	// crappy hack because you can't do \his[src] etc
-	var/t_his = his_or_her(src)
-	var/t_him = him_or_her(src)
+	var/t_his  = his_or_her(src)
+	var/t_him  = him_or_her(src)
 
 	. +=  "<br>[SPAN_NOTICE("*---------*")]"
 	var/datum/ailment_data/found = src.find_ailment_by_type(/datum/ailment/disability/memetic_madness)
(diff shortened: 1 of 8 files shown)


Valtsu0:
These should check if you have bartender training not if it is your job.

 `training_drinker` is also shared with HoS and det. Giving at least the last ability to these would make sense (honestly that one could be universal). Training specifically for bartender could be made if we do not want to give the other abilities to HoS and det.

code/modules/chemistry/Chemistry-Holder.dm
```diff
@@ -1096,11 +1102,12 @@ proc/chem_helmet_check(mob/living/carbon/human/H, var/what_liquid="hot")
 		for (var/i in 1 to num_val)
 			.[result_name[i]] = result_amount[i]
 
-	/// Gets a string of what something tastes like (as shown to the drinker/eater/whatever)
+	/// Gets a string of what something tastes like (as shown to the drinker/eater/whate ver)
```
Valtsu0, line 1105: This seems accidental

code/mob/living/carbon/human/procs/get_desc.dm
```diff
@@ -410,6 +410,30 @@
 		items = copytext(items, 3)
 		. += "<br>[SPAN_NOTICE("[src] is juggling [items]!")]"
 
+	if (src.reagents.has_reagent("ethanol"))
+		if (usr.mind && usr.mind.assigned_role == "Bartender")
```
Valtsu0, line 414: I don't like things checking for your job.
These should check if you have bartender training.

training_drinker is also shared with HoS and det. Giving at least the last ability to these would make sense (honestly that one could be universal). Training specifically for bartender could be made if we do not want to give the other abilities to HoS and det.

code/obj/item/device/pda2/scanners.dm
```diff
@@ -65,9 +65,15 @@
 		size = 6
 
 		scan_atom(atom/A)
+			var/mob/user = usr
 			if (..())
 				return
 			. = scan_reagents(A, visible = TRUE)
+			if (iscarbon(user) || ismobcritter(user))
+				if (user.mind && user.mind.assigned_role == "Bartender")
```
Valtsu0, line 73: same as above

code/obj/item/device/scanners.dm
```diff
@@ -530,9 +530,15 @@ TYPEINFO(/obj/item/device/reagentscanner)
 		if (isnull(src.scan_results))
 			boutput(user, SPAN_ALERT("\The [src] encounters an error and crashes!"))
 		else
-			boutput(user, "[src.scan_results]")
-
-	attack_self(mob/user as mob)
+			var/scan_output = "[src.scan_results]"
+			if (iscarbon(user) || ismobcritter(user))
+				if (user.mind && user.mind.assigned_role == "Bartender")
```
Valtsu0, line 535: same as above

Valtsu0 (changes requested):

code/mob/living/carbon/human/procs/get_desc.dm
Glamurio: I also vote for just slapping it on training_drinker and giving that to the Bartender. It fits.

MintyPhresh: I'm not sure about putting the first two abilities on training_drinker-- the HoS and Detective won't benefit nearly as much from the information as the bartender will, and I feel it'd just clog up the reagent scans with useless information. I'd rather just make Bartender Training. The third ability being for everyone makes sense, I'll implement that.

Valtsu0 (approved):

code/mob/living/carbon/human/procs/get_desc.dm
```diff
@@ -31,8 +31,8 @@
 			. = "<br>[src.bioHolder.mobAppearance.flavor_text]"
 
 	// crappy hack because you can't do \his[src] etc
-	var/t_his = his_or_her(src)
-	var/t_him = him_or_her(src)
+	var/t_his  = his_or_her(src)
+	var/t_him  = him_or_her(src)
```
Tarmunora, line 35: but why

code/mob/living/carbon/human/procs/get_desc.dm
```diff
@@ -410,6 +410,29 @@
 		items = copytext(items, 3)
 		. += "<br>[SPAN_NOTICE("[src] is juggling [items]!")]"
 
+	if (src.reagents.has_reagent("ethanol"))
```
Tarmunora, line 413: this does not appear to check for the trait on the viewer

code/mob/living/carbon/human/procs/get_desc.dm
```diff
@@ -410,6 +410,29 @@
...
+		var/et_amt = src.reagents.get_reagent_amount("ethanol")
+		var/drunk_assess = ""
+		if (!isalcoholresistant(src) || src.reagents.has_reagent("moonshine"))
+			if (src.reagents.has_reagent("moonshine"))
+				et_amt *= 7
```
Tarmunora, line 418: I'd drop this part, since that's not *really* how moonshine works anyways - it increases the efficacy of ethanol effects, but doesn't affect the volume thresholds at all

Tarmunora (changes requested):

code/mob/living/carbon/human/procs/get_desc.dm
MintyPhresh: I see! I did fundamentally misunderstand how moonshine works, will revert

code/mob/living/carbon/human/procs/get_desc.dm
MintyPhresh: That's intentional-- someone else here mentioned that it made sense as more of an everyone thing, so I did away with locking it behind bartender training

code/mob/living/carbon/human/procs/get_desc.dm
MintyPhresh: Whoopsie.

Tarmunora (approved):

PR #20868, opened 2024-09-26
Pod Wars: Add cryotrons to both main ships and Genfabs to the syndicate ship

JORJ949:
## About the PR 
Adds cryotrons to the cloning rooms in both ships' cloning rooms (I dont think the shields let the other team in to prevent spawncamping but im not sure, so the enemy shouldnt be able to cryo in your cryotron)

Also adds general fabricators to the syndicate ship so they can make more tools if someone runs off with them all

## Why's this needed? 
So those that need to go can cryo and not fill the ship with mindless operatives

So the syndicate can actually make ships


JORJ949:
I got sniped with the genfabs >:(

PR #20872, opened 2024-09-26
Pop pills from pill bottles directly, and other pill bottle tweaks

Egregorious:
## About the PR 

Tweaks both the regular pill bottle and the CheMaster pill bottle. 

Mainly this adds the ability to consume the next pill from a pill bottle directly by click-dragging the pill bottle onto the player sprite:
- Requires a free hand or the bottle to be in your hand, moves the bottle to your hand if in the inventory.
- Creates a sound and visible message.
- Clumsiness and braindamage can cause you to accidentally take more than one pill at once and spill pills on the floor.
- Is an action that takes 0.75 seconds.
- Interrupted when attacked, as well as when stunned, when moving, performing other actions etc.
- When interrupted via stuns and attacks, drop the pill you were attempting to take onto the floor. 

Also a few minor things:
- Click-dragging chemmaster bottles onto tables dumps contents similarly to satchels.
- Removed the capital letter from chemmaster bottles so they're not named "menthol Pill bottle".
- Added a description rebuild to the chemmaster so bottles display their content quantity immediately.

Internally:
- `obj/table`'s `MouseDrop_T()` uses `place_on()` for item placement.
- `atom/MouseDrop()` converts `params` to a list prior to making `mousedrop()` and `MouseDrop_T()` calls.

## Why's this needed? 

Pill bottles feel unnecessarily clunky, it takes at least three inputs to get them from your inventory into your gob. This feature means it can take as little as one input, but with the tradeoff being it is far less stealthy to do so. Using them normally is still a way to eat a pill without anyone around realising. It can also be used for roleplay because it enables being an addict in a more apparent manner. 

I don't _think_ this should affect combat, because it's also not really any faster than just consuming a loose pill from your backpack. 

## Changelog 

```changelog
(u)Egregorious
(*)Can now pop pills directly from pill bottles via click-dragging them onto yourself.
```


Tarmunora (changes requested):
With respect to the click-drag-to-take-pill interaction, I think we'd rather see something along the lines of chugging drinkingglasses - a repeating actionbar that eats a pill on each completion

Egregorious:
> With respect to the click-drag-to-take-pill interaction, I think we'd rather see something along the lines of chugging drinkingglasses - a repeating actionbar that eats a pill on each completion

I understand the desire for parity, but if you'll allow me to make a counter-argument:

Click-dragging is the most convenient action for this, because it reduces even taking pills from the inventory into 1 input instead of multiple. The next candidate I can think of is attacking yourself with the bottle, which requires an extra action to put the bottle in-hand from the inventory. 

Unlike chugging drinks, pills have variable quantity per-consumption. Pills can have up to 100 units in them, so there's far less reason to always want to down an entire pill bottle, and more reason for doctors and people who have made chemmaster pill bottles to want to avoid others chugging the whole thing at once, since it can waste reagent units and thus their effort. 

The most common use-case for a pill is to take one whenever an effect is desired, rather than all at once. So while I realise it's funny enough to be an inclusion, click-dragging to chug pills would take the space of the most convenient method of the most efficient - and I'd also assume most-common - use-case.

If you still think that's worth it that's fine, but I wanted to explain my thinking with this.

Tarmunora:
To clarify, having it as an action-bar based effect gives a number of levers to pull for balance, both in duration and interruptability, and allows to maintain some tradeoff still between loose pills (low-inventory-density) and a pill bottle (less convenience/reliability), rather than making pill bottles the singularly correct choice in all cases

Glamurio:
Yeah, I think the chugging makes sense, you can tweak the chugging speed, perhaps the first chug is fast, and subsequent chugs are slower, or something like that.

Egregorious:
An action bar does make a lot of sense! I still personally don't personally like the concept of no-effort pill chugging, so I tentatively made it act more like an input-buffer - it will continue looping for as many times as the player performs the action.

I also made it interruptable through attacks as well as all the default interrupts, made it so stunning and attacking interupts cause you to drop the pill you would be taking, because that seems like it could be fun, and arbitrarily made each consumption 0.75s because I couldn't come up with anything better.

PR #20876, opened 2024-09-27
Jailbird trait sends wanted posters to all printers

pgmzeta:
[FEATURE][REWORK][TRAITS]
## About the PR 
* Joining the round as a jailbird causes a wanted poster to be printed at all printers
* Remove the special arrest status interaction between stowaway and jailbird (and removes the stowaway trait)

## Why's this needed? 
* Allows more crew have a chance to interact with jailbirds about their trait choice.
* Idea taken from a forum thread here: https://forum.ss13.co/showthread.php?tid=23217

## Screenshot
image
Left: Jailbird captain, Right: Jailbird stowaway
## Changelog 

```changelog
(u)glowbold
(+)Joining the round as a jailbird prints a wanted poster at all printers. Keep an eye out!
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -1248,6 +1248,7 @@ ABSTRACT_TYPE(/datum/job/civilian)
 	name = "Stowaway"
 	limit = 2
 	wages = 0
+	trait_list = list("stowaway")
 	add_to_manifest = FALSE
 	low_priority_job = TRUE
 	slot_card = null
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -1248,7 +1248,6 @@ ABSTRACT_TYPE(/datum/job/civilian)
 	name = "Stowaway"
 	limit = 2
 	wages = 0
-	trait_list = list("stowaway")
 	add_to_manifest = FALSE
 	low_priority_job = TRUE
 	slot_card = null
--- a/code/datums/traits.dm
+++ b/code/datums/traits.dm
@@ -827,15 +827,6 @@ ABSTRACT_TYPE(/datum/trait/job)
 
 //Category: Background.
 
-/datum/trait/stowaway
-	name = "Stowaway"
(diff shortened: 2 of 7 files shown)


444Portal:
this is amazing

flrsh:
Good PR. Still not a fan of jailbird in general, but I very much appreciate trying to make it better and more interactive.

flrsh (approved):

TobleroneSwordfish (changes requested):
Won't the removal of the stowaway interactions mean that jailbird stowaways won't have arrest status at all?

pgmzeta:
> Won't the removal of the stowaway interactions mean that jailbird stowaways won't have arrest status at all?

thought about this a bit and while the original intent was to replace the jailbird&stowaway special-casing for arrest status with just a poster, I've just added it back in because I couldn't justify removing it. It's still a weird interaction, but the poster should hopefully be enough of a way to clue security into what's going on with Rando McArrestme without giving them a full record.

pgmzeta:
> making it print out on ALL Printers may lead to a bit of spam. Why not pick something like 10 random printer on station Z level and make only these print out the posters?
> 
> Overall, i like this PR.

Limited the printing to station Z, and did a check on how many printers each station has.

| Map | Printer Ct. |
| ------------- | ------------- |
| Atlas | 3 |
| Clarion | 10 |

| Clarion | 10 |

| Clarion | 10 |

| Clarion | 10 |

MintyPhresh:
Jailbird Stowaways should probably just get their own sec record (possibly missing stuff like fingerprints and blood ID) so that they have crimes and arrest status. Could have a couple not-real sec records so that Central doesn't magically know which wanted stowaways are stowing away

TobleroneSwordfish:
I'm going to just merge this for now, might be cool to do the fake record entries at some point but that can be a separate PR since this one is now maintaining the current behaviour.

PR #20877, opened 2024-09-27
removes underage drinking crime

Retrino:
## About the PR 
as title says, removes the automatic arrest status gained from drinking while below age 21 (so one age. 20)

## Why's this needed? 
conceptually, keeping an american standard for drinking age doesn't make sense when the setting is in the middle of space. while nt may be based out of america, i dont think sticking to this one specific law fits the setting well, even if it is a joke.

practically, the way underage drinking goes usually boils down to one of two things: 1, a person is aware of the underage drinking auto setting, and sets their age to 20 while slamming shots at the bar because ha ha funny arrest status. jailbird lite, basically. 2, a person who is unaware of the drinking age due to being from another country or just not thinking about it, ends up confused and frustrated because they keep getting stunned and cuffed by the securitron who keeps doing loops around the bar. 

not putting a changelog since i dont think it really needs one, but if it does i'll make one

### DIFF
--- a/code/modules/chemistry/Reagents-FoodDrink.dm
+++ b/code/modules/chemistry/Reagents-FoodDrink.dm
@@ -291,23 +291,6 @@ datum
 				..()
 				return
 
-			on_add()
-				if(!istype(holder) || !istype(holder.my_atom) || !ishuman(holder.my_atom))
-					return
-				var/mob/living/carbon/human/H = holder.my_atom
-				if(seen_by_camera(H))
-					// determine the name of the perp (goes by ID if wearing one)
-					var/perpname = H.name
-					if(H:wear_id && H:wear_id:registered)
-						perpname = H:wear_id:registered
-					var/datum/db_record/gen_record = data_core.general.find_record("name", perpname)
-					var/datum/db_record/sec_record = data_core.security.find_record("name", perpname)
-					// Yes. Its 21. This is Space America. That is canon now.
-					if(gen_record && sec_record && text2num(gen_record["age"]) < 21 && sec_record["criminal"] != ARREST_STATE_ARREST)
-						sec_record["criminal"] = ARREST_STATE_ARREST
-						sec_record["mi_crim"] = "Underage drinking."
-						H.update_arrest_icon()
-
 		fooddrink/alcoholic/hard_punch
 			name = "hard punch"
 			id = "hard_punch"


Retrino:
bro hit me with the soul comment 😭🙏

TobleroneSwordfish:
I am the number one soul destroyer incarnate and even I think this is soul 🥺

444Portal:
what if beepsky declared why he arrests underage drinkers to solve problem 2

MintyPhresh:
Would be nice if, instead of completely removing this feature, we reworked it to set you to a different, less severe status. Never really made sense to me that mail fraud, underage drinking, and public urination were worthy of forceful arrests (by beepsky).

JORJ949:
On the one hand: I cant be penned/artifacted/gassed/forcedrunk with ethanol to cause me to be marked arrest anymore
On the other hand: I cant be penned/artifacted/gassed/forcedrunk with ethanol to cause me to be marked arrest anymore

MintyPhresh:
> Making it less severe makes it mostly meaningless.
I also think the existence of mechanics like this invite players to be nuisances-- The setting to arrest is automatic, but removing it requires security attention. Making it less severe means that Sec can still interact with it if they want to, but the drinker isn't wailing over radio to be set off arrest.

flrsh:
I agree with the reasoning outlined in the PR. If we want to keep "soul" we could just have Beepsky verbally reprimand the drinkers. Or, if the necessary code doesn't exist for securitrons, make guardbuddies do it. Could even be extra and have them admonish anyone who's publicly drunk. I think that'd be much funnier. (Though I also don't think it's necessary to have in this specific PR; it should probably be a follow-up if someone wants to code that).

I do think a minor changelog entry would be good though.

TobleroneSwordfish:
To add an actual reason beyond "Soul": I think interactions like underage drinking and mail fraud setting you to arrest are good for the game because they help to maintain a certain level of background chaos that doesn't have to do with antagonist actions.

Also it's funny!

Flaborized:
this feature is hilarious and id hate to see it removed

ZeWaka:
soul

jan-antilles:
> If clarity for why someone gets arrested after doing this is an issue, I'd much rather have beepsky shout something when arresting an underage drinker compared to removing it outright/watering it down severely. It's a funny interaction that makes you go "oh huh that's a thing?" and it happening to someone once really isn't a big deal.

Agreed. Out of scope maybe, but I'd like it if Beepsky did that for *more* things.

TobleroneSwordfish:
Beepsky charging his baton while staring into your soul and declaring `ARREST INITIATED, REASON: USER IS CRINGE`

TobleroneSwordfish:
I think the consensus is that we're keeping this so I'll close the PR now.

PR #20879, opened 2024-09-27
Unhides handkerchiefs from Uniform manufacturers

itchybones:
## About the PR 
As simple as that; Unhides the handkerchiefs usually only found in the uniform manu's hidden menu. This also adds them to the HoP's uniform fab, as is reqested by the code comments.

## Why's this needed? 
~Someone told me "When you code it"~
It's a neat item! Who on a space station doesn't need a handkerchief? It's got fun little interactios, such as being able to be turned into a mask. Now in the code, @cogwerks left a comment saying that the manu "should be for regular uniforms [...], not tons of gimmicks, which I also understand. I simply want to make these more available, and if the uniform manu isn't the intended way, I'll try and find another way.

## Changelog 

It's a small change, I don't think it needs a changelog as is.

### DIFF
--- a/code/obj/machinery/manufacturer_subtypes.dm
+++ b/code/obj/machinery/manufacturer_subtypes.dm
@@ -448,12 +448,12 @@
 		/datum/manufacture/satchel,
 		/datum/manufacture/satchel_red,
 		/datum/manufacture/satchel_green,
-		/datum/manufacture/satchel_blue)
+		/datum/manufacture/satchel_blue,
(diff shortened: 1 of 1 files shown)


Azrun (approved):

flrsh (approved):
This seems fine with me since they're just the basic white ones, which means there's still a little incentive to buy a cooler one as a Spacebux purchase. When I added them as a hidden item they didn't have the bandana features; I can see why we'd want them to be more accessible now.

PR #20880, opened 2024-09-27
New Trait: "Was It Something I Said?"

Flaborized:
## About the PR 

Adds a new +2 trait to the game:

bothate_trait

It makes many small robots hate you! They won't help you out and some will actively go out of their way to cause you annoyance or harm. 

<details>
  <summary>List of Current Effects</summary>

  - Secbots will arrest you on sight, whether you're set to arrest or not. And they baton you a few extra times for good measure. 
  - Firebots won't put you out, and will instead insult you. They still put out area fires, so they can kinda hit you by accident, but they'll never target you specifically.
  - Medibots won't heal you unless they're emagged. They also deliver quirky little doctor insults.
  - Guardbuddies will become sad if you hug them, and will never hug you back. (:<) When they go to hug you, they'll instead flash/taser/shoot you point blank and then say something like "as if". (They don't cuff you tho just a little mean jab) They also no longer wave at you and instead gesture rudely.
  - Mulebots :)
 
</details>

medbot_insult

## Why's this needed? 

I think this would be fun! And some of the effects are funny. It's probably worth the two points, but I could probably be convinced to drop it to one. I may implement more bots being annoying, but they'd all have to do totally new things so I'm not sure. (Ie: skullbots could trip you with their leg, etc). Also the changelog is deliberately bare so people can figure out the effects it has in real time instead of having it all explained to them. Please leave any excellent ideas or feedback in the comments and be sure to like and subscribe

## Changelog 

```changelog
(u)Flaborized
(*)New Trait: "Was It Something I Said?".
```


MintyPhresh:
Not sure I like the bots going out of their way to hurt/inconvenience you, especially the secbots and guardbuddies. What about tripping you instead? Would probably fit well with the janibots.

TobleroneSwordfish:
Oh I love this, really creative! Maybe have medibots just kind of jab them with a needle (or steal some blood?) instead so it's less obvious when they're emagged? Otherwise people with this trait could instantly tell if a medibot is emagged.

Valtsu0:
Would love if it had interactions with azone bots too like tour guide or coat rack.

code/modules/robotics/bot/firebot.dm
```diff
@@ -256,8 +257,14 @@
 
 	if(src.extinguish_flags & EXTINGUISH_MOBS)
 		for (var/mob/M in by_cat[TR_CAT_BURNING_MOBS]) // fine I guess we can go extinguish someone
-			if (M == src.oldtarget || isdead(M) || !src.valid_target(M))
+			if (M == src.oldtarget || isdead(M) || !src.valid_target(M) || (M in src.people_to_ignore))
 				continue
+			if(M.traitHolder.hasTrait("wasitsomethingisaid"))
+				src.people_to_ignore.Add(M) //each bot gets to insult the person once
```
Tarmunora, line 263: This will cause GC issues - consider weakrefs instead

code/modules/robotics/bot/guardbot.dm
```diff
@@ -2759,6 +2764,11 @@ TYPEINFO(/obj/item/device/guardbot_module)
 							return
 
 						if(BOUNDS_DIST(master, hug_target) == 0)
+							if(hug_target.traitHolder?.hasTrait("wasitsomethingisaid"))
+								master.bot_attack(hug_target, TRUE) //betrayal!!
+								master.speak(pick("As if!", "You know what you did.", "Level [rand(32,80)] dork alert!"))
+								drop_hug_target()
+								return
```
Tarmunora, line 2771: normal hug behaviour sets `master.moving = 0` before returning. Intentional that this doesn't?

Tarmunora (changes requested):

code/modules/robotics/bot/guardbot.dm
Flaborized: was not intended fixed alongside other suggested change, thanks!

Flaborized:
also made it so cleanbots trip you like cats when you walk over them, though they're slightly less likely to

Tarmunora (approved):

EmeraldCrow:
Been seeing this in action, for the most part its a lovely and fun trait. my only issue with it is the secbots arresting you on sight. In practice, you can't really tell in game why the secbot keeps arresting this person, and over the course of a round it starts to get old. Reminds me of the days of jailbreaker+stowaway, where people would take both and then it was a headache for the officer on duty to deal with someone that was on arrest but not in the system. Much like here, the sec bot is going for their butt and there's not a lot you can do about it in game, and it just becomes frustrating and a time waste. My ideas for a change for this:

Sec bots only go in to arrest you when you get too close to them instead of it being on sight, such as being a tile or two tiles near them or on them. 

Sec bots don't target you by default, but if you have _any_ sec hud change (parole, arrest, suspect, etc) they will target you, even if they normally would not. kinda iffy on this one, since i do wanna have a way for sec bots to show displeasure at you without someone having to change up your sec files.

Sec bots don't arrest people they should near you, in the hopes they are about to ruin your day. "I DIDNT SEE NOTHING." sounds like a pain to code.

Sec bots, when they should arrest you, are a lot faster at doing it. not a fan of this one, since again, the trait should make the little bots upset at you by default instead of you having to be on arrest/holding contraband.

TDLR: fun trait, very good add, but maybe the sec bots can chill down just a little, but still arrest your sorry butt.

Jeanfeels:
For floorbots, since they don't seem to do much, perhaps have them follow you around (leading to a flock of cleanbots) and refuse to be turned off by you specifically. especialy if you're an artist or mime or have a crayon creator in your inventory. evil, but for this trait it would be perfect

PR #20881, opened 2024-09-27
Nukie masks get built-in voice changers

JORJ949:
## About the PR 
Adds a built in voice changer to syndicate masks
Also adds support for voice changers that cannot be removed and uses this for the built in voice changers of the syndicate and hunter masks.

## Why's this needed? 
Less chat clog when "BlastCo Operative Lima (as John Syndicate) says "Im gonna blow you up""
This isn't that huge a buff as far as I can tell for infiltrating as its still your syndicate mask and you are going to want other infilling gear anyway (infil comes with a loose voice changer you can put in a normal mask) though this may require blacklisting syndicate field masks from both the SWAT and respirator medal rewards.

## Changelog 

```changelog
(u)JORJ949
(+)Nuclear Operative gasmasks now come with built-in voice changers.
```

### DIFF
--- a/code/obj/item/clothing/gimmick.dm
+++ b/code/obj/item/clothing/gimmick.dm
@@ -24,7 +24,7 @@
 
 	New()
 		..()
-		src.vchange = new(src) // Built-in voice changer (Convair880).
+		src.vchange = new /obj/item/voice_changer/permanent(src) // Built-in voice changer (Convair880).
 		if(istype(src.loc, /mob/living))
 			var/mob/M = src.loc
 			src.AddComponent(/datum/component/self_destruct, M)
--- a/code/obj/item/clothing/masks.dm
+++ b/code/obj/item/clothing/masks.dm
@@ -47,6 +47,9 @@
 			return
 	else if (issnippingtool(W))
 		if (src.vchange)
+			if (src.vchange.permanent)
+				user.show_text("[src]'s [src.vchange.name] cannot be removed!", "red")
+				return
 			user.show_text("You begin removing [src.vchange] from [src].", "blue")
 			if (!do_after(user, 2 SECONDS))
 				user.show_text("You were interrupted!", "red")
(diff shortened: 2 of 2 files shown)


Xkallubar:
You can buy nukie masks at the syndie listening post for like 1200 credits. Voice changers from the merchant there cost 5000. I'd prefer if we weren't scamming people.

JORJ949:
> You can buy nukie masks at the syndie listening post for like 1200 credits. Voice changers from the merchant there cost 5000. I'd prefer if we weren't scamming people.

Nukie masks are NOT the red SWAT masks you can buy at the post, they only spawn on nuclear operatives. Nukies also dont have any money to buy the voice changer from CARL, even if carl rolled the voice changer.

Tarmunora:
I think my preferred resolution here would be to just have nukie IDs be fundamentally normal ID cards, with their spawn name etc. pre-filled on them, and refactor agent cards (and all their associated behaviour) out to a distinct child of syndicate ID cards rather than having them be the common parent type for all syndicate IDs.

JORJ949:
> I think my preferred resolution here would be to just have nukie IDs be fundamentally normal ID cards, with their spawn name etc. pre-filled on them

I kind of like the way the agent cards help them lean into gimmicks though, for example a recent round where they went all sawflies and picked names like "Sawfly Bishop"

TobleroneSwordfish:
I like this, although
> This isn't that huge a buff as far as I can tell for infiltrating as its still your syndicate mask

You can remove voice changers from masks with a pair of scissors, maybe make these unremovable because they're "built-in"?

JORJ949:
> You can remove voice changers from masks with a pair of scissors, maybe make these unremovable because they're "built-in"?

Its done in the same way the hunter mask is, I assume theres no way to remove it from that?

TobleroneSwordfish:
image

TobleroneSwordfish (approved):

PR #20882, opened 2024-09-27
give clarion barbershop a barber uniform

Retrino:
## About the PR 
adds a barber uniform to the back room of the clarion barbershop

## Why's this needed? 
cutting hair is hard ):


Glamurio:
Relevant PR., although does not invalidate this one #20677

PR #20887, opened 2024-09-28
Prevent recalling of the next automatic shuttle after 120 minutes

JORJ949:
## About the PR 
After 120 minutes the next automatic shuttle call will disable shuttle recalling

## Why's this needed? 
Rounds have almost no need to EVER go over 2 hours and if they do there should be an admin there to go "yep sure its ok" and recall it themselves. This will only really affect those classic players that want a 5 hour round just because.

## Changelog 

```changelog
(u)JORJ949
(*)Shuttle recalling is now disabled after the second automatic call.
```

### DIFF
--- a/code/datums/gamemodes/gm_parent.dm
+++ b/code/datums/gamemodes/gm_parent.dm
@@ -12,6 +12,7 @@ ABSTRACT_TYPE(/datum/game_mode)
 	var/shuttle_auto_call_time = 90 MINUTES // 120 minutes.  Shuttle auto-called at this time and then again at this time + 1/2 this time, then every 1/2 this time after that. Set to 0 to disable.
 	var/shuttle_last_auto_call = 0
 	var/shuttle_initial_auto_call_done = 0 // set to 1 after first call so we know to start checking shuttle_auto_call_time/2
+	var/shuttle_prevent_recall_time = 120 MINUTES // After how long do we prevent recalling the Shuttle (only applied upon an automatic call)
 
 	var/latejoin_antag_compatible = 0 // Ultimately depends on the global 'late_traitors' setting, though.
 	var/latejoin_only_if_all_antags_dead = 0 // Don't spawn 'em until all antagonists are dead.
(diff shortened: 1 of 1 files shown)


Xkallubar:
Just make the shuttle arrive faster the longer the shift has gone on?

Metekillot:
Or put this to a vote?

Tarmunora (approved):

TobleroneSwordfish:
What are people's concerns with this? I don't really see a reason you would ever want 2 hour+ rounds under normal circumstances.

flrsh:
I could see RP rounds occasionally going longer. Maybe if people don't like 120, it could be 180?

JORJ949:
> I could see RP rounds occasionally going longer. Maybe if people don't like 120, it could be 180?

Looking at goonhub only about 2/10 RP rounds deviate from 103 minutes, which is the autocall time, and even then most that go over only do so by about 10-20m. Though if it is an issue theres always #ifdef RP_MODE

NightmareChamillian:
> > Looking at goonhub only about 2/10 RP rounds deviate from 103 minutes, which is the autocall time, and even then most that go over only do so by about 10-20m. Though if it is an issue theres always #ifdef RP_MODE
> 
> Ah true. The ifdef works for me, does anyone who thumbs down'd this have thoughts? 

Honestly, gimmick long-term rounds. Back when goon2 existed, rounds could last in the span of multiple days, with some truly crazy shit happening to the station/departments. I doubt it's going to happen again but this PR superfluously axes the potential _for_ it to happen again naturally, and for no apparent reason than "just because"

JORJ949:
> 
> Honestly, gimmick long-term rounds. Back when goon2 existed, rounds could last in the span of multiple days.

These gimmick rounds should always be admin run because if they arent then anyone who dies now cant play the game for hours. Anyone wanting to join the next round will have to wait forever. While I see where the fun of super-long rounds come from these always happen when admins are offline because delaying the shuttle is against the rules. I do see the appeal of long-term rounds but the way players run them is inconsiderate to other players and I've had people get mad at me for calling the shuttle in a 4 hour round.

ZeWaka (approved):

Konamaco:
I've seen plently of rounds that have gone on for 2+ hours because of auto recaller packets, ai's and mindhacks  with "dont recall laws or orders" and griefers, and despite ahelping they continued until admins could get on, i feel like although rare, this pr would help stop those rounds where antags or griefers keep spam recalling the shuttle while admins are busy and unable to stop them. maybe make it a toggle for admins or some shit incase they are doing some event round though but, no round normally last that long anyways.

PR #20894, opened 2024-09-28
Require harm intent to adjust shape of table with screwdriver

TealSeer:
[QOL] [GAME OBJECTS]
## About the PR 

Changes table code to require you be on harm intent to use a screwdriver to adjust the shape of the table, like how you need harm intent to disassemble it with a wrench (this also changes it for the screwdriver drink too). Update help message to reflect this change. Also change wrench part of code to use intent macro instead of string.

## Why's this needed? 

Trying to place a screwdriver on a table is frustrating so we can give it parity with how the wrench does it

## Changelog 

```changelog
(u)TealSeer
(+)Using a screwdriver on a table now requires you be on harm intent.
```

### DIFF
--- a/code/obj/table.dm
+++ b/code/obj/table.dm
@@ -37,7 +37,7 @@ TYPEINFO_NEW(/obj/table)
 	var/drawer_locked = FALSE
 	/// id for key checks, keys with the same id can lock it
 	var/lock_id = null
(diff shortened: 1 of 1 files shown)


Sbmhawk:
how does this interact with the drink?

TealSeer:
> how does this interact with the drink?

same deal, requires harm intent now

PR #20898, opened 2024-09-29
Allow 3 Security Officers to revoke armory access

JORJ949:
## About the PR 
Allows 3 people with security locker access to deauthorize armory

Code probably sucks but I wanted dev opinion before putting effort into cleaning it considering the last one was closed.

## Why's this needed? 
This PR was previously closed (#18330)  (Forum Thread: https://forum.ss13.co/showthread.php?tid=22461) under the reasoning "Nah - armory auth should be treated as a last resort, not an additional gear locker". Since a cooldown is now required to pass before you can deauth (5 minutes) this will discourage using it as a "gear locker". In addition its far more likely one captain would use it this way than 3 officers (I have never seen 3 officers auth without good reason, I have seen many captains do it, then requiring me to use packets to lock it.) This allows 3 officers to unauthorize the armory in the case they dont want it open, e.g. rogue captain, someone stole the auth disk. This change seemed pretty liked in the thread.

## Changelog 

```changelog
(u)JORJ949
(*)Three people with Security Equipment access can now revoke armory access.
```

### DIFF
--- a/code/obj/machinery/computer/security/riot.dm
+++ b/code/obj/machinery/computer/security/riot.dm
@@ -117,11 +117,11 @@
 
 	get_help_message()
 		if (src.authed)
-			. = "The Head of Security can unauthorize armory access."
+			. = "Three security personnel, or the Head of Security can revoke armory access."
 			if(!authdisk_authorized)
 				. += "<br>You can also use the <b>Authentication Disk</b> to issue an emergency revocation."
 		else
-			. = "Three security personnel, or the Head of Security, can authorize access."
+			. = "Three security personnel, or the Head of Security, can authorize armory access."
 			if(!authdisk_authorized)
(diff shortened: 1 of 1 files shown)


JORJ949:
Forum thread for the previous PR can be found [here](https://forum.ss13.co/showthread.php?tid=22461)

Glamurio:
There's plenty of situations where you'd want access to the armory but not have it permanently open, particularly when it wasn't even Security who opened it in the first place. Considering a HoS can deauth, and one is not always present, it just makes sense to have this be a feature.

If SecOffs using it as a locker is in any way a problem, this should be resolved via ahelp, as it's very firmly against powergaming rules, I think.

flrsh (approved):
This makes sense to me.

ZeWaka (approved):

PR #20912, opened 2024-09-30
Alters Tetrodotoxin behavior and changes pufferfish availability slightly

colossusqw:
[LABEL][balance][catering][chemistry]

## About the PR 
This PR first and foremost reduces both Tetrodotoxin's depletion, and the amount of it that come in each pufferfish, it also makes it so you need at least a tier 2 rod to actually catch pufferfish under most circumstances, with a few minor fishing spot exceptions. This PR also alters the timings of TTX slighly to try and make the progression less immediately obvious and forgiving.

(I intend to change the rod tier requirement to a bait requirement once the system is properly implemented, so the rod tier is a bit of a stop gap intended for the moment.)

## Why's this needed? 
I've been looking at how TTX plays in game for a while, and what i've noticed is it really doesn't see much use at all. While i never intended for it to be some meta poison, i think the high dose until it does any damage, and somewhat forgiving nature of the damage windup, while coupled with the ease to treat and very inconsistent availability makes it feel quite underwhelming, even for a gimmick poison. As for the progression change, it's an attempt at making the poison a bit more dangerous, since at the moment the very loud heartbeats and sizeable alloted time do make this a pretty undertuned chem, especially for the acquisition effort.

The change in depletion is supposed to do two things:
1- make obtaining any usable dose through ??? pills an actual possibility(3 units of impurity means you'd need to get lucky in 5 pills for any effect).
2- make smaller amounts as additives to food have some way of happening, since even spiking a drink feels near impossible with the 15 unit lethal dose.

Stacking does not seem like it will be a concern, since i find it difficult to think of most things that could complement this well at 0.2 depletion, that weren't already options, once most loud poisons just defeat tetrodotoxin's main thing, as mentioned before, what this does the most is make it existing in the ??? pills pool not as irrelevant as it currently is.
This, not to mention we've since veered away from things like scalpel dipping and other easy ways to pass on poisons in small doses directly to people, which already mostly only worked with very fast acting chems like old 11 cycle depletion pancuronium.

## Changelog
```changelog
(u)Colossus
(+)Tetrodotoxin now has a smaller(0.2) depletion, and slightly altered timing, but pufferfish only carry half the amount in them, and can only be caught by tier 2 rods or higher.
```

### DIFF
--- a/code/modules/chemistry/Reagents-PoisonEtc.dm
+++ b/code/modules/chemistry/Reagents-PoisonEtc.dm
@@ -1949,7 +1949,7 @@ datum
 			fluid_g = 180
 			fluid_b = 240
 			transparency = 10
-			depletion_rate = 0.3
+			depletion_rate = 0.1
 			var/progression_speed = 1
 			var/counter = 1
 
--- a/code/modules/chemistry/Reagents-PoisonEtc.dm
+++ b/code/modules/chemistry/Reagents-PoisonEtc.dm
@@ -1949,7 +1949,7 @@ datum
 			fluid_g = 180
 			fluid_b = 240
 			transparency = 10
-			depletion_rate = 0.1
+			depletion_rate = 0.2
 			var/progression_speed = 1
 			var/counter = 1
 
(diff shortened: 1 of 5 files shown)


code/modules/chemistry/Reagents-PoisonEtc.dm
```diff
@@ -1949,7 +1949,7 @@ datum
 			fluid_g = 180
 			fluid_b = 240
 			transparency = 10
-			depletion_rate = 0.3
+			depletion_rate = 0.1
```
TobleroneSwordfish, line 1952:
```suggestion
			depletion_rate = 0.2
```
I get where you're coming from with this change, but 0.1 is just too low considering smoke powder exists.

TobleroneSwordfish (changes requested):

PR #20918, opened 2024-10-01
Adds Golf Crates 

cheekybrdy:
## About the PR 
Adds golf crates to QM, as well as adds a mapping prefab of it.

## Why's this needed? 
Resurfaces a buried minigame, which could be used really creatively with mechcomp and other modern items.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Added golf crates to QM - a blast from the past!
```


TobleroneSwordfish:
I'm not sure the golf stuff is stable enough to be made player accessible at the moment, from some basic testing it just seems to runtime and arbitrarily pixel shift players around. Maybe if it was cleaned up a bit and ideally had new sprites 🥺

TobleroneSwordfish:
Oka I'm going to close this for now to save it ending up in a stale loop. Feel free to re-open if and when golf cleanup happens!

PR #20941, opened 2024-10-02
AI Law Manudrive

JORJ949:
## About the PR 
Adds a manudrive to the RD's locker with the 6 basic laws found on-station.

## Why's this needed? 
If someone steals, blows up or otherwise destroys the laws found in upload they are presently impossible to replace if they weren't already mechscanned (which is unlikely). 
## Changelog 

```changelog
(u)JORJ949
(*)The Research Director now has a Manudrive containing the blueprints for most on-station laws.
```

### DIFF
--- a/code/datums/manufacturing.dm
+++ b/code/datums/manufacturing.dm
@@ -1702,6 +1702,48 @@
 	time = 60 SECONDS
 	category = "Component"
 
+/datum/manufacture/syndicate_laws
+	name = "Syndicate Law Module Set"
+	item_requirements = list("metal_dense" = 40)
+	item_outputs = list(/obj/item/aiModule/syndicate/law1, /obj/item/aiModule/syndicate/law2, /obj/item/aiModule/syndicate/law3, /obj/item/aiModule/syndicate/law4)
+	create = 1
+	time = 60 SECONDS
+	category = "Component"
+
+ABSTRACT_TYPE(/datum/manufacture/aiModule)
+/datum/manufacture/aiModule
+	name = "AI Law Module - 'YOU SHOULDNT SEE ME'"
+	item_requirements = list("metal_dense" = 10)
+	item_outputs = list(/obj/item/aiModule/asimov1)
+	create = 1
+	time = 10 SECONDS
+	category = "Component"
+
+	makeCaptain
+		name = "AI Law Module - 'MakeCaptain'"
+		item_outputs = list(/obj/item/aiModule/makeCaptain)
+
+	oneHuman
+		name = "AI Law Module - 'OneHuman'"
+		item_outputs = list(/obj/item/aiModule/oneHuman)
+
+	notHuman
+		name = "AI Law Module - 'NotHuman'"
+		item_outputs = list(/obj/item/aiModule/notHuman)
+
+	emergency
+		name = "AI Law Module - 'Emergency'"
+		item_outputs = list(/obj/item/aiModule/emergency)
+
+	removeCrew
+		name = "AI Law Module - 'RemoveCrew'"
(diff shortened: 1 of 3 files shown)


code/datums/manufacturing.dm
```diff
@@ -1702,6 +1702,48 @@
...
+	item_requirements = list("metal_dense" = 40)
+	item_outputs = list(/obj/item/aiModule/syndicate/law1, /obj/item/aiModule/syndicate/law2, /obj/item/aiModule/syndicate/law3, /obj/item/aiModule/syndicate/law4)
+	create = 1
+	time = 60 SECONDS
+	category = "Component"
+
+ABSTRACT_TYPE(/datum/manufacture/aiModule)
+/datum/manufacture/aiModule
+	name = "AI Law Module - 'YOU SHOULDNT SEE ME'"
+	item_requirements = list("metal_dense" = 10)
```
flrsh, line 1716: Ditto re: conductive materials.

code/datums/manufacturing.dm
```diff
@@ -1702,6 +1702,48 @@
...
+	item_outputs = list(/obj/item/aiModule/asimov1)
+	create = 1
+	time = 10 SECONDS
```
flrsh, line 1719: I think 20 seconds would be better.

code/datums/manufacturing.dm
```diff
@@ -1702,6 +1702,48 @@
 	time = 60 SECONDS
 	category = "Component"
 
+/datum/manufacture/syndicate_laws
+	name = "Syndicate Law Module Set"
+	item_requirements = list("metal_dense" = 40)
+	item_outputs = list(/obj/item/aiModule/syndicate/law1, /obj/item/aiModule/syndicate/law2, /obj/item/aiModule/syndicate/law3, /obj/item/aiModule/syndicate/law4)
+	create = 1
+	time = 60 SECONDS
+	category = "Component"
```
flrsh, line 1711: I feel like these should also be separate and output individual law modules? And should probably require some conductive material as well?

flrsh (changes requested):
The code itself looks good to me! Two thoughts--

Should the manudrive be in the computer core or bridge instead? Since all heads have access to the upload, I feel like they should be able to access this as well.

Also, I disagree with emagging this unlocking Syndicate law modules. I feel like emagging should result in more "malfunction" type effects rather than blatantly Syndicate stuff. You could instead add a separate Syndicate manudrive to the trader at the listening post?

code/datums/manufacturing.dm
JORJ949: Same time and requirements as the other lawsets, though due to the syndicate nature of this I would be open to increasing the cost, though I will be removing the availability of these in this PR.

code/datums/manufacturing.dm
JORJ949: I do think AI laws should probably take some conductive, though I am just keeping consistency with other laws for the purposes of this PR

JORJ949:
>
> Should the manudrive be in the computer core or bridge instead? Since all heads have access to the upload, I feel like they should be able to access this as well.
>

Currently each head of staff has something to replace something AI (HoP/Cap dont count). MD has the AI itself, HoS can replace the turrets with NARCS and the CE has the law rack, probably the closest thing to the laws themselves. I originally considered also making this available via cargo but i'm not sure on the balance implications of being able to make an entire law rack, rogue laws included, without visiting upload.

flrsh:
> Currently each head of staff has something to replace something AI (HoP/Cap dont count).

Ah fair point!

flrsh (approved):

Azrun (approved):

PR #20945, opened 2024-10-02
Burnable trees and shrubs

pgmzeta:
[feature][game objects]
## About the PR 
* Co-opts the burnable items tracking category for `/obj/tree` and `/obj/shrub`
* Add basic fire effects for burning trees/shrubs
* Allows firefighting foam to put out tree/shrub fires

## Why's this needed? 
Fights in terrainify environments more interactive

## Changelog 

```changelog
(u)glowbold
(*)Trees and shrubs can be set on fire.
```


TobleroneSwordfish:
Oouugh copy paste. Can I persuade you to make this a component instead so it can be not copy pasted and can be added to other things in future?

pgmzeta:
> Oouugh copy paste. Can I persuade you to make this a component instead so it can be not copy pasted and can be added to other things in future?

Yep, can do. To confirm, should the component should replace/supplant the base `/obj/item` burning code too, or just be for non-item burning?

TobleroneSwordfish:
Ideally I think yes, but I don't want to condemn you to refactoring all of that right now if it turns out to be A Lot.

pgmzeta:
i'm not sure how to do this as a component

PR #20953, opened 2024-10-03
"Fix" staggered still letting you sprint

TobleroneSwordfish:
[BUG] [BALANCE] [GUH]
## About the PR 
The description of the "staggered" status effect you get when you hit someone or get hit reads: `"You have been staggered by a melee attack.<br>Slowed slightly, unable to sprint."` but currently it doesn't do that.

## Why's this needed? 
Might make fights a little less wacky clickfests and make it easier to catch someone who is sprinting away from you?
I'm very open to being told this is a bad idea or was removed for a reason etc.

## Changelog 

```changelog
(u)LeahTheTech
(*)Getting staggered by a melee attack now makes you unable to sprint for a short time.
```

### DIFF
--- a/code/modules/status_system/statusEffects.dm
+++ b/code/modules/status_system/statusEffects.dm
@@ -941,9 +941,13 @@
 
 		onAdd(optional=null)
 			.=..()
+			APPLY_ATOM_PROPERTY(src.owner, PROP_MOB_CANTSPRINT, src)
 			if (ishuman(owner))
 				var/mob/living/carbon/human/H = owner
 				H.sustained_moves = 0
+		onRemove()
+			REMOVE_ATOM_PROPERTY(src.owner, PROP_MOB_CANTSPRINT, src)
+			. = ..()
 
 	blocking
 		id = "blocking"


TDHooligan:
agree with the concept - sprinting through debilitating conditions is a fun little interaction in that you get to directly expend stamina for a benefit, but sprinting in melees is unhealthy. 

i'd personally love to see this expanded so that this effect replaces continuous sprint with a one-tile quick step that costs a chunk of stamina.

flrsh (approved):
Makes sense to me.

Laboredih123:
i think a better option would be a chance to twist their legs together and fall when staggered, not prevent sprinting entirely. if they decide not to sprint, then they might or might not escape as always, if they decide to sprint, they can either successfully escape or they can twist their legs and fall and get stunned for a short time, ending up in a worse situation than if they had just not sprinted. i find allowing sprinting with risk of ending up worse better than just not allowing sprinting or keeping sprinting for free.

TobleroneSwordfish:
> a chance to twist their legs together and fall when staggered, not prevent sprinting entirely.

I don't think a random chance to fall over if you tap sprint in combat would be a very popular idea.

PR #20956, opened 2024-10-03
Resprites the Mime's Bowler Hat and Odd Job

MitchTheUser:
[CLOTHING]

## About the PR 

Resprites the Mime's Bowler Hat and Odd Job. The hat's shape has been remade to more resemble an actual bowler hat, being more round at the top and smoothing out the brim's edge. The color and shading has also been improved, and are more accurate to the worn sprites.

BowlerDisplay

## Why's this needed? 

Current sprites are quite old and don't look so good compared to modern hats. They also didn't look very "bowler hat-y".


ZeWaka:
better than prior

PR #20959, opened 2024-10-03
Add PTL, MD office and CE office to atlas

JORJ949:
## About the PR 
Adds a second ship to mirror the science one
On the second ship is a PTL, CE office, engi storage, and engi hangar (ignore the debug generator in the storage room)
image
Reclaim empty room for the MDir office and swap it with the toilets
image
Move the north trader dock onto the outer ship and expand mechanics and refinery slightly
image

## Why's this needed? 
Adds many things that arent presently on atlas such as ABCU units, the PTL and a portamedbay

## Changelog 

```changelog
(u)JORJ949
(*)Added a second smaller "ship" to atlas on the northwest side of the ship, on board is a PTL and CE office.
(+)Added a Port-A-Medbay and MDir office to atlas.
```

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -106,24 +106,9 @@
 /turf/simulated/wall/auto/reinforced/supernorn,
 /area/station/quartermaster/office)
 "aaI" = (
-/obj/table/reinforced/auto,
-/obj/item/paper/book/from_file/minerals{
-	pixel_x = -2;
-	pixel_y = 5
-	},
-/obj/item/paper/book/from_file/matsci_guide,
-/obj/item/device/matanalyzer,
-/obj/cable{
-	icon_state = "0-2"
-	},
-/obj/machinery/power/apc/autoname_north,
-/obj/machinery/light{
-	dir = 4;
-	layer = 9.1;
-	pixel_x = 10
-	},
+/obj/laser_sink/mirror,
 /turf/simulated/floor/engine,
-/area/station/mining/refinery)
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Oh this is really coool! I love the inbuilt PTL mirror setup. Maybe for symmetry and ease of access the cables should go over the gantry rather than under the airbridge?

JORJ949:
> Maybe for symmetry and ease of access the cables should go over the gantry rather than under the airbridge?

I wanted the PTL wire to be easily reached by any engineers looking for a potential hotwire, instead of running it through mining.

Cherman0:
Mmm... I like the concept of putting in a CE office and PTL and I like the symmetry, but these don't seem like things that would be on a detachable ship. It would be a larger map overhaul but maybe it would make sense for mining to be on a ship (with the magnet somewhat mirroring the toxins test platform) and haivng the engineering stuff be where mining currently is?

JORJ949:
> Mmm... I like the concept of putting in a CE office and PTL and I like the symmetry, but these don't seem like things that would be on a detachable ship. It would be a larger map overhaul but maybe it would make sense for mining to be on a ship (with the magnet somewhat mirroring the toxins test platform) and haivng the engineering stuff be where mining currently is?

My lore excuse for it is that the realised atlas doesnt have a PTL and instead of doing an expensive retrofit on a large-scale ship they retrofitted another ship that was going to be scrapped and sent it to dock with atlas.

PR #20960, opened 2024-10-03
Jelly bean tweaks

Egregorious:
## About the PR 

1. Reworks the way jelly beans colour themselves, to ensure they retain their colour when grown by botany.
2. Mabie Nott's beans (the botany variety) are given a random colour if they spawn with no reagent filling, instead of defaulting to black.
3. Reduced the default sugar content of both jelly beans:
   - Farty Snott's: 50 -> 40
   - Mabie Nott's: 90 -> 10
4. Fixed Farty Snott's beans not producing a 'tastes like' message when eaten. 
5. Standardised the way jelly beans get their sugar contents with the way other candy items do. 

## Why's this needed? 

1. Beans grown by botany tend to just be the colour of nitrogen after almost any amount of potency is involved, instead of the colour of their spawned reagent-filling.
2. More colourful jelly beans!
3. Mabie Nott's beans are weirdly high in sugar by default, especially considering everything else about them encourages eating them in multiples; and this change doesn't affect the sugary-potential of botany-grown beans anyway due to their scalable VHFCS. Reducing Farty Snott's sugar content on the other hand just ensures that wizard golems created from the beans are always named for their additive reagent, rather than for sugar. 
4. Almost definitely an intended mechanic which just isn't getting activated. 
5. The only downside to this is that botany beans don't replace the default sugar content anymore. Due to the sugar content reduction this doesn't actually matter much, but I also increased the initial volume from 100->110 to compensate. The flavouring also doesn't get replaced now either, but instead it raises the volume of the reagent container in order to fit and thus will not fight botany reagents for bean-space. 

## Changelog 

```changelog
(u)Egregorious
(+)Botany's jelly beans are a lot more colourful.
```


code/modules/food_and_drink/candy.dm
```diff
@@ -137,88 +137,98 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/candy/jellybean)
...
+			var/datum/reagent/current = reagents_cache[reagent]
+			if (current)
+				// make space for the flavouring if there is none
+				if(src.reagents.maximum_volume - src.reagents.total_volume < 10)
+					src.reagents.maximum_volume += 10
+				src.reagents.add_reagent(reagent, 10)
+				src.food_color = rgb(current.fluid_r, current.fluid_g, current.fluid_b, max(current.transparency,255))
+				src.icon += src.food_color
+			else // no flavouring? Pick a random colour!
+				src.food_color = rgb(rand(1, 255), rand(1, 255), rand(1, 255), 255)
```
flrsh, line 178:
```suggestion
				src.food_color = random_saturated_hex_color()
```
Will look better I think.

code/modules/food_and_drink/candy.dm
```diff
@@ -137,88 +137,98 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/candy/jellybean)
...
-				if (src.reagents.total_volume <= 60)
-					src.reagents.add_reagent("sugar", 40)
+			if(prob(50))
+				flavor = pick("egg", "vomit", "snot", "poo", "earwax", "wet dog", "belly-button lint", "sweat", "congealed farts", "mold", "armpits", "elbow grease", "sour milk", "WD-40", "slime", "blob", "gym sock", "pants", "brussels sprouts", "feet", "litter box", "durian fruit", "asbestos", "corpse flower", "corpse", "cow dung", "rot", "tar", "ham", "gooncode", "quark-gluon plasma", "bee", "heat death")
+				phrase = pick("Oh god", "Jeez", "Ugh", "Blecch", "Holy crap that's awful", "What the hell?", "*HURP*", "Phoo")
+				tastesbad = 1
+			else
+				flavor = pick("egg", "strawberry", "raspberry", "snozzberry", "happiness", "popcorn", "buttered popcorn", "cinnamon", "macaroni and cheese", "pepperoni", "cheese", "lasagna", "pina colada", "tutti frutti", "lemon", "margarita", "coconut", "pineapple", "scotch", "vodka", "root beer", "cotton candy", "Lagavulin 18", "toffee", "vanilla", "coffee", "apple pie", "neapolitan", "orange", "lime", "mango", "apple", "grape", "Slurm")
+				phrase = pick("Yum", "Wow", "MMM", "Delicious", "Scrumptious", "Fantastic", "Oh yeah")
```
flrsh, line 230: Maybe add flavor and phrase list vars on the jellybean parent so you can just do `pick(bad_flavors)` / `pick(good_flavors)` or whatever instead of duplicating this stuff on both the `someflavor` and `everyflavor` children? I think they're meant to contain the exact same strings.

flrsh (changes requested):

code/modules/food_and_drink/candy.dm
Egregorious: You're right it looks like they're pretty much identical, I've refactored as you suggest. I also removed the need for the flavor/phrase variables for the sake of optimisation, so now beans have Schrodinger's flavour until eaten. I don't think that makes any practical difference unless admins were peaking on jellybean flavour-flavourtext prior to eating them for some reason.

Egregorious: Oh, I guess it does prevent admins from editing flavourtext mid-round, maybe that's something worth keeping?

flrsh: This seems like a super niche thing that I've never seen get edited so I think it's probably fine to remove it in favor of better code!

flrsh (approved):

PR #20970, opened 2024-10-04
Clean up item burning variables and use define for burnt item remains

pgmzeta:
[internal][cleanup]
## About the PR 
* Use doc-comments for various item burning-related variables
* Replace `burn_type` with `burn_remains` as the variable name that tracks what items should melt into when burnt up. 
* Create a new burnable define `BURN_REMAINS_*` for clarity on what stuff will burn into.

## Why's this needed? 
Clearer variable names/doccomments. Atomizing work on burning component. Allows for additional types of burn remains to be easily added in the future, as well as no remains.

### DIFF
--- a/_std/defines/burning.dm
+++ b/_std/defines/burning.dm
@@ -1,0 +1,8 @@
+// Burning /obj/item
+
+/// Leaves nothing behind when burned
+#define BURN_REMAINS_NONE 0
+/// Leaves an ash pile behind when burned
+#define BURN_REMAINS_ASH 1
+/// Leaves a molten mess behind when burned
+#define BURN_REMAINS_MELT 2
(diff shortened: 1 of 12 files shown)


_std/defines/burning.dm
```diff
@@ -0,0 +1,8 @@
+// Burning /obj/item
+
+/// Leaves nothing behind when burned
+#define BURN_REMAINS_NONE 0
+/// Leaves an ash pile behind when burned
+#define BURN_REMAINS_ASH 1
```
TobleroneSwordfish, line 6: Consider making these defines contain type paths of the remains in future?

TobleroneSwordfish (approved):

PR #20984, opened 2024-10-06
Removes examine range from intangibles

Valtsu0:
[SILICONS][QoL]
## About the PR 
Intangibles no longer get the "is too far away to see clearly." message for trying to examine people far from their eye.
Mostly affects AIs (as they can now examine people from viewports), but wraiths, poltergeists, flockmind/trace, blobs, zoldorf manifests and brain things (whatever they are) are also affected

## Why's this needed? 
The eye is mostly an abstraction right? Their ability to see people clearly should not depend on where it is.
I am open to changing this to not include intangibles that can't move freely (wraith, poltergeist, zoldorf manifest)

### DIFF
--- a/code/mob/living/carbon/human/procs/get_desc.dm
+++ b/code/mob/living/carbon/human/procs/get_desc.dm
@@ -17,7 +17,7 @@
 	if (isalive(usr))
 		. += "<br>[SPAN_NOTICE("You look closely at <B>[src.name] ([src.get_pronouns()])</B>.")]"
 
-	if (!istype(usr, /mob/dead/target_observer) && !istype(usr, /mob/living/intangible))
+	if (!isobserver(usr) && !isintangible(usr))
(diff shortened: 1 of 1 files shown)


JORJ949:
Makes sense for AI eyes to me because thats just where the AI is supposed to be focusing on (from my understanding) and it can see anywhere on the station thats on-camera.

code/mob/living/carbon/human/procs/get_desc.dm
```diff
@@ -17,7 +17,7 @@
 	if (isalive(usr))
 		. += "<br>[SPAN_NOTICE("You look closely at <B>[src.name] ([src.get_pronouns()])</B>.")]"
 
-	if (!istype(usr, /mob/dead/target_observer))
+	if (!istype(usr, /mob/dead/target_observer) && !istype(usr, /mob/living/intangible))
```
Tarmunora, line 20:
```suggestion
	if (!isobserver(usr) && !isintangible(usr))
```

PR #20986, opened 2024-10-06
Money laundering now sets you to arrest

JORJ949:
## About the PR 
Automatically sets you to arrest if you do money laundering
https://github.com/user-attachments/assets/7439441a-b2ee-47d4-9c79-4b3803d4dd25

## Why's this needed? 
Dastardly criminals have gotten away with it for TOO LONG!

## Changelog 

```changelog
(u)JORJ949
(+)Laundering money now automatically marks you for arrest.
```

### DIFF
--- a/code/WorkInProgress/laundry.dm
+++ b/code/WorkInProgress/laundry.dm
@@ -22,6 +22,7 @@ TYPEINFO(/obj/submachine/laundry_machine)
 	var/cycle_current = 0
 	var/cycle_max = CYCLE_TIME
 	var/mob/occupant = null
+	var/mob/activator = null
 	var/image/image_door = null
 	var/image/image_light = null
 	//var/image/image_panel = null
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (changes requested):
Funny and cool, only nitpick is that `src.activator` should be set back to `null` at the end of the cycle to avoid GC issues and occasional weird edge cases.

TobleroneSwordfish (approved):

code/WorkInProgress/laundry.dm
```diff
@@ -22,6 +22,7 @@ TYPEINFO(/obj/submachine/laundry_machine)
 	var/cycle_current = 0
 	var/cycle_max = CYCLE_TIME
 	var/mob/occupant = null
+	var/mob/activator = null
```
ZeWaka, line 25: clear these on disposing please

ZeWaka (changes requested):

code/WorkInProgress/laundry.dm
ZeWaka: also maybe when the record gets processed. this would leave a hanging ref to the mob.

ZeWaka:
Occupant eject/null and such as well pls

ZeWaka (approved):

PR #20997, opened 2024-10-07
Give wizard headsets unique sprites

JORJ949:
## About the PR 
image
image

Gives wizard headsets unique sprites, name, description and chat icon
Chat icon by @Mister-Moriarty 

## Why's this needed? 
Makes it more clear to newer wizards that their headset is special as it is immune to EMPs unlike any on-station radios.

## Changelog 

```changelog
(u)JORJ949 & Mr. Moriarty
(+)Wizard headsets now have unique icons.
```


flappybatpal:
Might be worth adding to the description that they are immune to EMPS?

flappybatpal (approved):

PR #21006, opened 2024-10-08
Security pouches can be attached to armor vests

FlameArrow57:

## About the PR 
Allows security pouches to have a one-time option to be attached to a security armor vest permanently, at the cost of reduced armor.

Balance changes if making a pouched vest:
-One extra pocket pouch free
-Reduced protection of vest
-Can't access pouch as easily while in space

Vest won't be able to be attached to anything else like paramedic suits afterwards (not implemented yet)

## Why's this needed? 
Just a stylistic clothing idea option for Security, maybe good, maybe bad idea (not sure about the idea of extra storage space)

Could also rework into something different, ideas below
-Utility credit option that only holds security supplies or grenades (pairs with morphine injectors, flashbang grenades, etc)
-Armory exclusive item

## Changelog 

```changelog
(u)FlameArrow57
(*)Security pouches can be permanently attached to Security armor vests through clicking it with the pouch, at the cost of reduced vest armor. The vest can be taken off by click-dragging it to your hand, or using another clothing piece in-hand.
```


Glamurio:
I personally think it's good to give vests more options, however the trade-off of not being able to access it as easily feels like a massive penalty to any bonuses this may provide. I'd ideally like to see an implementation that allows you to open the attached pouch on the vest as per usual, even if it'd have to happen via UI icon at the top left.

That aside, I also think a utility credit item that does something similar but is NOT the pouch could work as well here., if not better. Just a 3 slot purchase for small items in your vest slot, or something like that.

Glamurio:
> I think you understand that tradeoff wrong. It's not easy to be accessed in space because you are wearing a space suit instead of the vest. Looking at the PR description and the code changes, clicking on the vest should open up the inventory (if you are wearing it)

I may have misunderstood it. I think that's fine, then.

Flaborized:
I don't much like the pouch attach to vest idea, weakening your armor for a single small item slot is pretty weak plus all the other downsides, even without those mechanics I'm just not a huge fan of combining the functionality of two items in this kinda way. A "chest carrier" suit item slot or something could be pretty cool though if it were balanced to either have a small amount of slots or only hold specific things (ie: flashbangs). Ideally that would have a pretty distinct sprite from armor so players could tell the difference at a glance tho

ZeWaka:
idk lets try it out?

DisturbHerb:
There's an armour vest style that's very much more of a webbing setup than anything that provides actual armour protection. Could split that off into its own chest carrier object. Otherwise, I can be around to help sprite one.

trazodont:
I would really love if it worked with the other styles of the armor!

PR #21039, opened 2024-10-10
Reduce ability log spam

flappybatpal:
## About the PR 
Reworks ability logging to be more centralized and remove redundant/duplicate log entries.

## Why's this needed? 
Logs have many entries just from people pressing ability buttons without a target or we don't actually log the location properly.
image


flappybatpal:
This is probably kinda bad ATM, advice appreciated.
Need to test more abilities.

ZeWaka (approved):
merging before this causes conflicts

PR #21047, opened 2024-10-10
Seed harvesting rework

Egregorious:
## About the PR 

Adds a counter to seeds to keep track of how many times they can be planted. Seeds harvested from plants are combined into a single 'seed packet' with a number of uses based on the amount of produce items in the harvest, instead of multiple identical objects. The use count should be the same as the number of seeds that would be harvested now. 

- Splicing seeds consumes a use of a seed and the seed will only be destroyed if all uses have been consumed.
- Infusing seeds reduces uses by 10% of the current total amount in the packet rounded up, but will not reduce the use count below 1. Reagents which deal more damage to seeds also consume more uses; notably ammonia costs an additional 2 uses, and things which deal around 50% of the seeds health at once tend to reduce the remaining uses to 1.
- Mutating a seed packet via infusion drops the number of uses to 1.
- Seed extraction creates a seed packet with 2-5 uses, instead of 2-5 seed objects. 

## Why's this needed? 

Botany makes too many seed items. The intent is to reduce the sheer quantity of seed items that botany tends to create and has to deal with, hopefully helping to declutter hydroponics and the Plantmaster UI while affecting the balance as little as possible. 

Mutating and infusing reduces seed uses in an attempt to balance the fact that this change would multiply the affects of infusing across several seeds. Similarly, splicing only consumes a use instead of an entire seed because it's closer to how it works now. 

This change does affect accessibility to dominant and recessive seeds, a seed packet can only be one or the other rather than a mix. I couldn't think of a decent way around that.

However, creating a number of individual seeds from a single harvest is still possible for the most part via seed extraction, so dominant and recessive seeds are still available that way. This is also relevant in cases where you'd want to mutate a number of seeds at the same time. 

## Changelog 

```changelog
(u)Egregorious
(*)Each plant harvest now produces only one seed packet with multiple uses. Infusing and mutating this seed packet will limit the number of times it can be planted. 
```


Tarmunora (approved):

TobleroneSwordfish:
Going to testmerge this for a bit just to expose it to all the nerd botanists and see what they think.

ZeWaka:
https://github.com/goonstation/goonstation/issues/21089

PR #21051, opened 2024-10-11
Player Piano Notes Rework V2

Facsimil:
[GAME OBJECTS] [QOL] [REWORK] [FEATURE] [MINOR]

PR #21058 (Text to Piano Component) is a dependent.

## About the PR 

<details>
<summary>Adds delays</summary>

- A delay of 0 means a concurrent note, limited to 8
- All notes have a delay of 1 by default, to maintain backwards compatibility
- Note events can have a max delay of 100
- Rest events can have a max delay of 1000

</details>

<details>
<summary>Adds the <code>Timing</code> event</summary>

- Goes from 10 centiseconds to 50
- Can only be used at the start of a song

</details>

<details>
<summary>Changes how notes are counted</summary>

- Checks if the split notes is under the limit in proc `clean_input`
- ```MAX_NOTE_INPUT``` was reduced from 15360 to 1920 to account for the note counting change
    - 1920 was found by multiplying 120 (measures) by 16 (16th note), from [PR #9027](https://github.com/goonstation/goonstation/pull/9027)

</details>

<details>
<summary>Fixed bug where the first given-input's length wasn't checked</summary>

- Steps:
    - Checks if the length of ```note_input``` (not yet set) is below limit
    - Sets ```note_input``` to  ```given_notes```'s value
- This allowed a user to add up to 3750 notes (text box character limit), though it would error when you tried to set the notes again since ```note_input``` had the previous input's value that was over the limit

</details>

<details>
<summary>Updated <code>player_piano.txt</code></summary>

- Includes details about the Dense Format and events
- Reworked with HTML
- Added information about PRs [#14878](https://github.com/goonstation/goonstation/pull/14878) and [#16647](https://github.com/goonstation/goonstation/pull/16647)

</details>

- Added `LAGCHECK(LAG_LOW)` to `build_notes` since it takes too long to run (21.2 ms for 3895 notes and 10 ms for 1790 notes)

## Why's this needed? 

- Reduces song character count
- Single-instrument note concurrency reduces the need for multiple pianos
- Fixes a minor note-counting bug

## Changelog

```changelog
(u)Facsimil
(*)For Player Pianos: added note delays and single-instrument concurrent notes. Details can be found in the Player Piano book.
```


ZeWaka:
why aren't we using a standard format like ABC

Facsimil:
> why aren't we using a standard format like ABC
> 
> like, my thought is we're adding YET ANOTHER niche random format for our ingame music, which really hurts people sharing music around ss13

- `why aren't we using a standard format like ABC`
    - Because all the Player Piano does is play a note every `x` seconds.
    - It's unlikely all of the ABC format's features would be supported due to limitations on our end.
    - If we would use ABC, we'd probably have to implement an ABC to Player Piano converter internally.
    - Speaking of a standard format, why don't we use TG's format? We both run on the same engine and TG alone has a relatively massive (700+) library of songs.
- `niche random format`
    - The Dense Format is essentially a compressed version of the original format.
- `really hurts people sharing music around ss13`
    - The [Piano Song Dump](https://wiki.ss13.co/Piano_Song_Dump) is still up. I've even seen others use songs that I converted.

TobleroneSwordfish:
My thoughts are basically: is it worth adding another extremely confusing custom format for the purpose of allowing slightly longer songs?
The delay and timing changes seem cool (although should ideally be in their own PR), and I wouldn't be opposed to supporting TG's music format as well since that's an already existing standard and would allow for better music sharing.

Facsimil:
I will admit that the Dense Format is not as understandable to the general player.

So, Zewaka, would you be ok with me implementing a TG Song parser?

Facsimil:
Also would this PR be ok if I dropped the Dense Format?

TobleroneSwordfish:
I think so, yeah! It's a little hard to give a concrete review of the other stuff because it's all mixed in together.

PR #21054, opened 2024-10-11
More atlas fixes

JORJ949:
## About the PR 
3 minor atlas fixes:
Change ID of sci airbridge to "sciship" from "ai2medbay" (copypasted from cogmap according to original PR)
Add a missing wire in sci ship
Adds a normal toxins access spawner to the storage room because to reach normal toxins you have to go through storage (and the other door into tox storage has both)

## Why's this needed? 
Fix good


TobleroneSwordfish:
Oh good catch on me deleting a wire, whoops

PR #21059, opened 2024-10-12
Adds Radiation Protection to CE's Belt's Energy Shield

Garash2k:
## About the PR 
Adds 100 points of external radiation protection, (same as radsuits,) to CE's aurora belt's active shield effect.
image

## Why's this needed? 
I like the idea of CEs quickly dashing in the reactor room to talk to their engineers and maybe adjust something, without needing to don a rad suit, safely.
I know that the belt is already considered very strong. But this buff is both pretty minor, considering how radsuits are plenty available to them, they even have one in their personal locker. And thematic, since the belt already shields from heat and cold. Feel free to slap a controversial or input wanted tag if you feel otherwise!
I also don't think stacking this effect on top of a rad suit is a concern. The baseline 92% is strong enough to stand next to a generator meltdown without much noticeable damage. Temporarily cranking that number up to 99% won't help with anything.

## Changelog 

```changelog
(u)Garash
(+)The Chief Engineer's aurora MKII utility belt now provides radiation protection while toggled on.
```

[Game Objects][Balance]


Tarmunora:
Aurora belt might be getting a bit overtuned at this rate, but this is a good add

PR #21062, opened 2024-10-12
Datumize Clothing Stains

pgmzeta:
[internal][feature]
## About the PR 
Switch clothing stains from strings to a datum singleton pattern. Freshly-laundered's cold resistance behavior has been moved to the stain itself.

## Why's this needed? 
More easily add behavior to clothing based on stain application.

### DIFF
--- a/code/obj/item/clothing.dm
+++ b/code/obj/item/clothing.dm
@@ -65,16 +65,16 @@ ABSTRACT_TYPE(/obj/item/clothing)
 		src.name = "[name_prefix(null, 1)][src.get_stain_names()][src.name][name_suffix(null, 1)]"
 
 	///Add a stain to this clothing piece
-	proc/add_stain(datum/stain/stn)
-		stn = get_singleton(stn)
-		if (!stn || !src.can_stain)
+	proc/add_stain(stain_type)
+		var/datum/stain/stain = get_singleton(stain_type)
+		if (!istype(stain) || !src.can_stain)
 			return
 		if (!islist(src.stains))
 			src.stains = list()
-		else if (stn in src.stains)
+		else if (stain in src.stains)
 			return
-		src.stains.Add(stn)
-		stn.add_to_clothing(src)
+		src.stains.Add(stain)
+		stain.add_to_clothing(src)
 		src.UpdateName()
 
 	///Returns a space-concatenated string of stain names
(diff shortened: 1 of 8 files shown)


code/obj/item/clothing.dm
```diff
@@ -62,25 +62,41 @@ ABSTRACT_TYPE(/obj/item/clothing)
 		src.name = src.real_name || initial(src.name)
 		if(src.material?.usesSpecialNaming())
 			src.name = src.material.specialNaming(src)
-		src.name = "[name_prefix(null, 1)][src.get_stains()][src.name][name_suffix(null, 1)]"
+		src.name = "[name_prefix(null, 1)][src.get_stain_names()][src.name][name_suffix(null, 1)]"
 
-	proc/add_stain(var/stn)
+	///Add a stain to this clothing piece
+	proc/add_stain(datum/stain/stn)
```
TobleroneSwordfish, line 68:
```suggestion
	proc/add_stain(stain_type)
```
This shouldn't be typed as an instance

code/obj/item/clothing.dm
```diff
@@ -62,25 +62,41 @@ ABSTRACT_TYPE(/obj/item/clothing)
...
+	proc/get_stain_names()
 		if (src.can_stain && islist(src.stains) && length(src.stains))
-			for (var/i in src.stains)
-				. += i + " "
-
+			for (var/datum/stain/stn in src.stains)
+				. += stn.name + " "
+
+	///Removes a stain from this clothing piece. Returns 1/TRUE if stain removed.
+	proc/remove_stain(datum/stain/stain)
```
TobleroneSwordfish, line 87: ditto

code/obj/item/clothing.dm
```diff
@@ -62,25 +62,41 @@ ABSTRACT_TYPE(/obj/item/clothing)
...
+		stain = get_singleton(stain)
+		if(islist(src.stains))
+			for(var/datum/stain/stn in src.stains)
+				if (stn == stain)
+					stn.remove_from_clothing(src)
+					. = src.stains.Remove(stn)
```
TobleroneSwordfish, line 93: This doesn't need to be a loop, just check `if (stn in src.stains)` and then call the remove proc and remove it from the list if it is.

TobleroneSwordfish (changes requested):
Code generally looks good, couple of nitpicks

PR #21067, opened 2024-10-13
Hacked Mining Charges no Longer Guarantee a Nuclear Generator Meltdown

Garash2k:
## About the PR 
Makes it so that mining charges do not enact the `ex_act(1)` they normally would if the target is a nuclear reactor. Because of the explosion, it still causes harm, but at least the reactor can be repaired.

(First explosion is slower because it had breakpoints)

https://github.com/user-attachments/assets/3d345a9d-0eca-4a21-b159-d8c7feb311f0

## Why's this needed? 
Old behaviour sucked because of a combination of 2 things:
There's very little counterplay for engineers against the hit and run tactics of a wizard teleporting in or a miner breaching in.
It runins the whole round for the engineer since destroyed reactors can't be repaired.

## Changelog 

```changelog
(u)Garash
(+)Hacked charges no longer guarantee a nuclear generator meltdown. They still make lots of radioactive rods and debris fly around though!
```
[Game Objects][Balance]

### DIFF
--- a/code/obj/item/grenades.dm
+++ b/code/obj/item/grenades.dm
@@ -1402,7 +1402,7 @@ ADMIN_INTERACT_PROCS(/obj/item/gimmickbomb, proc/arm, proc/detonate)
 						if (src)
 							src.boom()
 							if (target)
-								if (istype(target, /obj/machinery))
+								if (istype(target, /obj/machinery) && !istype(target, /obj/machinery/atmospherics/binary/nuclear_reactor))
 									target.ex_act(1) // Reliably blasts through doors.
 						return
 		return


code/obj/item/grenades.dm
```diff
@@ -1402,7 +1402,7 @@ ADMIN_INTERACT_PROCS(/obj/item/gimmickbomb, proc/arm, proc/detonate)
 						if (src)
 							src.boom()
 							if (target)
-								if (istype(target, /obj/machinery))
+								if (istype(target, /obj/machinery) && !istype(target, /obj/machinery/atmospherics/binary/nuclear_reactor))
```
Tarmunora, line 1405: Perhaps it would be better to make it specifically check for doors, if that code block is meant for busting doors 🤔

Garash2k: Maybe, but people might be used to them removing stuff like computers, manufacturers, the AI law rack and the PTL. For the first 2, it's probably a good thing that they get nuked, for the law rack and PTL, I dunno, it'd be a gameplay change.

Thoughts on what's preferred between a blacklist or whitelist and it's content?

Tarmunora:
https://github.com/goonstation/goonstation/commit/ae4530ce19be92d21499cb882b69b7a7eb842537
I went ahead and implemented this via a different approach

PR #21068, opened 2024-10-13
Fixes missing trunk

cheekybrdy:
## About the PR 
Fixes a missing flusher

## Why's this needed? 
Fixing my fuckup which is good


ZeWaka:
please fix your master

PR #21070, opened 2024-10-13
Adds a vision overlay to gas masks

colossusqw:
[LABEL][Clothing][Balance]
## About the PR 
This PR adds a overlay to gas masks, the same size as welded sheet masks, that restricts longer range visibility when worn. This does not affect slasher masks, NTSO/NTSC or listening post/nukie masks.(please let me know if i made any mistakes in which masks to apply this to)

## Why's this needed? 
There have been multiple discussions related to the addition of SOME drawback to gas masks (for example this forum post https://forum.ss13.co/showthread.php?tid=23206) and more often than not the general aggreement is they could use something to balance them, since at the moment they offer lots of benefits and no drawback to wearing all the time. This would bring masks in line with other protection equipment that has added descentive not to be worn at all times such as hazard suits and space suits, and cut down on the amount of players rushing to get one at roundstart + never taking them off at any moment.

With this overlay, the mask's visibility is not impacted nearly enough to stop regular use at chemistry, nor it's emergency use to traverse smoke, but would make a case for actually taking it off when you go to the bar get a drink.

EDIT: from just doing chemistry with the mask on all the time as testing i can attest to it not really getting in the way of your actual job, everything is visible enough that i can just play as normal if i'm in science doing science stuff.

image

```changelog
(u)Colossus
(*) Regular station gas masks now have a visibility penalty.
```

### DIFF
--- a/code/modules/overlays/OverlayManager.dm
+++ b/code/modules/overlays/OverlayManager.dm
@@ -387,6 +387,16 @@
 
 		return ..()
 
+/datum/overlayComposition/gasmask
+	New()
+		var/datum/overlayDefinition/dither = new()
+		dither.d_icon = 'icons/effects/overlays/weldingmask.dmi'
+		dither.d_icon_state = "steelmask"
+		dither.d_blend_mode = 2
+		dither.d_mouse_opacity = 0
+		definitions.Add(dither)
+
+		return ..()
 
 // temporary blindness overlay until the other one is fixed
 /datum/overlayComposition/limited_sight
--- a/code/obj/item/clothing/masks.dm
+++ b/code/obj/item/clothing/masks.dm
@@ -111,12 +111,25 @@
 	color_r = 0.8 // green tint
 	color_g = 1
 	color_b = 0.8
+	var/low_visibility = TRUE
 
 	setupProperties()
 		..()
 		setProperty("coldprot", 7)
 		setProperty("heatprot", 7)
 		setProperty("disorient_resist_eye", 10)
+
+	equipped(mob/user, slot)
+		. = ..()
+		if (low_visibility)
+			user.addOverlayComposition(/datum/overlayComposition/gasmask)
+			user.updateOverlaysClient(user.client)
+
+	unequipped(mob/user)
+		. = ..()
+		if (low_visibility)
+			user.removeOverlayComposition(/datum/overlayComposition/gasmask)
+			user.updateOverlaysClient(user.client)
 
 /obj/item/clothing/mask/gas/NTSO
 	name = "NT gas mask"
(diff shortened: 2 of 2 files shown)


Valtsu0:
I really dislike this. The vision overlay is mostly just ugly and annoying. Sure, it's effective, but maybe too much so. 99% of time i'd just rather suffer in the capsaicin cloud of whatever. Most of the really bad or common chems are skinpen or don't have touch anyways.

It also just doesn't make sense. It would if we had the old gas masks tg has, but the ones ours are based on don't restrict vision.

colossusqw:
This might just be me, but like i said, at least in normal science gameplay, it didn't feel annoying or restrictive at all, i could very much do any science related activity very much like before with this. It might just be something that's better played than seen tbh, since yeah i had my doubts before test-playing it.

Valtsu0:
I agree that it's not restrictive and you can still do the same stuff you usually can. I just don't like the way vision circles look in general

TobleroneSwordfish:
Generally agree that this is too debilitating/annoying for a regular gas mask. They have big wide faceplates, it doesn't make sense that they'd restrict your vision that much.

colossusqw:
> Generally agree that this is too debilitating/annoying for a regular gas mask. They have big wide faceplates, it doesn't make sense that they'd restrict your vision that much.

Ok, could something like a movement penalty(bio suits for example) work?

Demigurian:
Their mass availability is the main issue, in my opinion. Additionally, any item that interfere with someone's mobility or visibility will likely be discarded 90% of the time. Moreover, this makes the stealth traitor gameplay even more frustrating, as both the listening post mask and fake moustache easily reveal your allegiance.

colossusqw:
> Their mass availability is the main issue, in my opinion. Additionally, any item that interfere with someone's mobility or visibility will likely be discarded 90% of the time. Moreover, this makes the stealth traitor gameplay even more frustrating, as both the listening post mask and fake moustache easily reveal your allegiance.

While i am currently testing other ways of doing this, yeah, gas masks are kinda supposed to be tools for specific situations, and as such, having a good reason not to be running them all the time outside the points where they are most useful should make sense. You'll still not want to be without a gas mask on a chem lab or in a big miasma leak with god knows what in it. As for the stealth traitor part, there are numerous other options for concealment that rarely see play exactly because the mask is so free of downsides and incredibly useful.

colossusqw:
Well, this seems to not be the way to go about it. Closing it for the time being while i ponder some other way to keep gas mask balance in check.

colossusqw:
While i do agree, i'm not sure if the general consensus goes this way as well. I'd rather it be this over the other change but that's me personally, although i wouldn't mind some mix of the two..

PR #21071, opened 2024-10-13
Resprites the humble Desk Lamp (black)

Flaborized:
## About the PR 

Resprites the humble desk lamp. the black ones. the green ones look good and aren't resprited by this. just the black ones. it is specifically based on the luxo L-1 model i believe!

lamp2

lamp1

note that it's a bit better centered than the current sprite, which could mess up some lined up desk lamps. i dont think this will be a huge deal and I'd rather the sprite be centered than not considering its just a few pixels off from the OG

edit: oh also i added a little switch to the on overlay implying you use that to control the lamp because i thought it would be cool

## Why's this needed? 

one day (today) I opened up the wanted resprites list and picked one arbitrarily to redraw and this was that one. I think it looks nicer and fits the guidelines better!


TemThrush:
Really nice work, love the strong colors and I think the switch is a super nice addition! A bit of a nitpick, but the perspective on the old sprite stands out to me as one of the main factors that makes the lamp look very out of place in-game, so keeping it in that same perspective seems like a bit of an odd choice to me since it's something I think could really be improved. The lamp is a pretty complex object, so I get it may be a bit less readable top-down. 😓 Besides that note, I think this is great!

Flaborized:
> Really nice work, love the strong colors and I think the switch is a super nice addition! A bit of a nitpick, but the perspective on the old sprite stands out to me as one of the main factors that makes the lamp look very out of place in-game, so keeping it in that same perspective seems like a bit of an odd choice to me since it's something I think could really be improved. The lamp is a pretty complex object, so I get it may be a bit less readable top-down. 😓 Besides that note, I think this is great!

I think in this case yeah the tilted angle helps a bit with readability, And in general lamps like this don't unambiguously have an exact front, so I kept it the way it was because I think it outlines the shape of it very nicely; if it was looking at you head-on you wouldn't be able to see the iconic bend joint thing those lamps have. I don't personally think it looks too out of place on tables and next to other stuff that fits perspective, since you're still seeing it from the same vertical angle (ie: it's not a full side-on or top-down view)

PR #9196, opened 2022-06-17
wearing a mask that covers your mouth now prevents eating

Skeletonman0:
[BALANCE] [QOL]
## About the PR 
image

This PR makes it so that wearing a mask that has coversmouth set will prevent eating food items.

## Why's this needed? 

scientists and other gas mask enthusiasts have been able to phase food through gas masks for too long

## Changelog

```changelog
(u)Skeletonman0
(+)Spacemen can nolonger magically phase food through masks that cover their mouth.
```

PR #21078, opened 2024-10-14
No eating/drinking/pills with gas masks

Valtsu0:
[Game Objects][Balance]
## About the PR 
Prevents eating, drinking and swallowing pills while wearing a gas mask.
Does not apply to nukie masks
Does not prevent others from force feeding you as that is arguably a buff.

oh also adds pronouns to some messages near the changes
## Why's this needed? 
It's mostly to make some reason not to wear one, even if a quite small one.
Should make people take off the gas mask sometimes, at least on rp.
Needing an extra hand to take a pill might do something on classic.

## Changelog 

```changelog
(u)Valtsu0
(+) You can no longer eat, drink or swallow pills while wearing a non-syndicate gas mask
```

### DIFF
--- a/_std/defines/clothing.dm
+++ b/_std/defines/clothing.dm
@@ -67,6 +67,8 @@ var/list/all_slots = list(SLOT_BACK, SLOT_WEAR_MASK, SLOT_L_HAND, SLOT_R_HAND, S
 #define ONBACK						(1<<17)
 /// can be work on the belt
 #define ONBELT						(1<<18)
+/// stops the person from eating, drinking and swallowing pills
+#define BLOCKEATING					(1<<19)
 
 
 //Suit blood flags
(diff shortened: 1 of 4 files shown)


DisturbHerb:
#9196 was a previous PR with a similar proposed change.

Valtsu0:
Both of those are from over 2 years ago.

AXLFAN2000:
stopping nukies from eating their nukie donk pockets in combat is a bit of a nerf

Valtsu0:
forgot about those, i'll change it

pgmzeta:
It's weird that it stops you from eating but not getting force fed. Previously have thought about preventing chugging with a gas mask as a Minor Nerf instead of full eating restriction.

IPling:
Personally I find this a confusing mechanic for newer players. As to them it seems extremely arbitrary to only apply to gasmasks, which becomes even more arbitrary as nukies get an exception.

With balance its eh. Of course its a small one which is fine, yet I feel like the weight of the inconvenience for the majority of players outweighs the benefits this PR brings.

Valtsu0:
Third time was not the charm

PR #21080, opened 2024-10-14
Split every access from I WANNA BE THE JOB

JORJ949:
## About the PR 
Splits captain access returning real all access from I_WANNA_BE_THE_JOB into its own define of I_MEAN_ALL_ACCESS

## Why's this needed? 
Sometimes you wanna be the job captain without every access, though it could be on EVERYONE_GETS_CAPTAIN_ID which would be perfectly reasonable

### DIFF
--- a/code/procs/access.dm
+++ b/code/procs/access.dm
@@ -200,11 +200,7 @@
 	switch(job)
 		// --------------------------- Heads of staff
 		if("Captain")
-#if defined(I_WANNA_BE_THE_JOB)
-			return access_all_actually
-#else
 			return get_all_accesses()
-#endif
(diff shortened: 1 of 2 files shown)


code/procs/access.dm
```diff
@@ -347,6 +343,9 @@
 			return list()
 
 /proc/get_all_accesses()  // not adding the special stuff to this
+#if defined(I_MEAN_ALL_ACCESS)
+			return access_all_actually
+#else
 	return list(access_security, access_brig, access_forensics_lockers,
```
ZeWaka, line 349: identation

ZeWaka (changes requested):

PR #21081, opened 2024-10-14
Roundstart "Stumped" improvements

JORJ949:
## About the PR 
Removes the SPAWN(6 SECONDS) from stumped trait's leg removal
Automatically buckles any stumped trait owner into their wheelchair at roundstart

## Why's this needed? 
Having legs for about 5s at roundstart only for them to magically dissapear looks odd
Lots of jobs spawn on top of normal chairs and you will always buckle to those first rather than your wheelchair so you have to either move to a different tile or clickdrag yourself to the 1 pixel of wheelchair visible behind the chair.


TobleroneSwordfish:
I said this would be fine so now I have to merge it and see if it explodes 🙏

PR #21082, opened 2024-10-14
Limits the Availability of Gas Masks

colossusqw:
[LABEL][Balance][Clothing]

## About the PR 
This PR removes gas masks from random maintenance lockers. Additionally, it adds a slight cobryl cost to the manufacture of gas masks.

## Why's this needed? 
Alternative to the addition of a gas mask downside. This makes their availability a bit smaller while not being nearly as restricted as other hard counters such as flash protection, thus limiting their use to mostly the jobs that should need to use them as a tool.

There have been multiple discussions related to just how better than any other option gas masks are, and while i don't think a consensus ever shows up, one of the things that shows up again and again is the matter of just how easily available gas masks are. This hopefully changes that, and still leaves everyone with an option to counter smoke hazards in the form of the breath mask everyone spawns with.

## Changelog 

```changelog
(u)Colossus
(*)Gas masks no longer spawn on random maintenance lockers.
(+) Gas masks have had 0.4 cobryl added to their manufacture cost.
```


JORJ949:
Should not be removed from engineering IMO, they need gas masks to deal with burst TEG pipes.

444Portal:
Just noticed the Engineering Vendor, that one should not be seeing removal in my opinion - Removal from Fabricators is what I'd rather do.

colossusqw:
There you go, back to vendors

MintyPhresh:
Crossposting this from the now-closed gasmask overlay post:

I would prefer if gas masks had some downside to being worn, rather than just being hard to find. If it's got some reason to not be on your face, it's taking up an inventory slot, which is a more fun resource to manage than simply "did I rush uniform lockers faster than the other crew."

Nerfing the mechanics of the gas mask, in essence, makes the question "is this worth having?" which is much more engaging than "can I get this" or "can I afford this" in my opinion. 

I wouldn't sweat large amounts of negative feedback too much; nerfs, removals and similar are generally unpopular but I think it'll improve the game to do something about gas masks.

colossusqw:
Paraphrasing from what I said in that PR as well:
Yeah, i'd definitely prefer some level of drawback to gas masks as well, or potentially some middle ground, which would line the masks more with the balancing of other stuff and provide actual interesting gameplay choice instead of handing a very strong item with no downsides whatsoever. This PR mostly exists because people also seemed to suggest it as an alternative.

JORJ949:
The commonly available gas masks (maints) are specifically the emergency ones, so you could look at nerfing those a bunch

Carton171:
I'd like to point out that the gas masks in the lockers in maints are there for use with firefighting gear, and I feel that removing an essential part of fire gear that is already not in most of the lockers would probably be a bit harsh, especially for lowpop where fighting a large fire can be especially difficult, since often the most suited roles simply don't exist.

colossusqw:
> Gas masks fulfill no use in firefighting, except working as internals. All the gas mask does there is accomplished by the internal mask everyone starts with at round start in their emergency box.
> 
> There are almost no reasons for them to be part of engineering or emergency gear.

That is a big part of why I was hoping for the drawback option, as it stands, gas masks are a better breath mask with flash protection and identity obscuring. Giving them something in the way of vision or movement alterations would make them an alternative rather than the benchmark, and if anyone didn't want to deal with those, there's plenty of breath masks and oxygen tanks around.

JORJ949:
> That is a big part of why I was hoping for the drawback option, as it stands, gas masks are a better breath mask with flash protection
Gas masks dont have eyeprot, only the NTSC one does with only 20% resistance

444Portal:
> > gas masks are a better breath mask with flash protection
> 
> Gas masks dont have eyeprot, only the NTSC one does with only 20% resistance

they do, no?

TobleroneSwordfish:
Default gas masks do have 10% flash protection, not that it's going to change much in most circumstances?

colossusqw:
TDH's PR is just better, closing this one too. Godspeed.

PR #21096, opened 2024-10-16
Prevent shooting guns at null target

flappybatpal:
## About the PR 
Add a null target check to gun_parent to prevent shooting if we have a null target. Moves the isturf check up higher to also catch this if it ever returns null.

## Why's this needed? 
Shooting at null doesn't do anything but still causes the gun to use ammo and runtime.
Specifically we error on shooting at /obj/jobxprewardbutton.


flappybatpal:
We might still want to log or CRASH on the null target?

Tarmunora (approved):

PR #21107, opened 2024-10-18
Lethal .38 obtainability rebalance

JORJ949:
## About the PR 
Removes lethal .38 ammo from sectechs
Increases .38 lethal ammo in ammotech from 2->6
Increases .38 stun ammo in sectech from 2->3

## Why's this needed? 
I often see detectives (on classic at least) just play it like a secoff with lethals, and with them being able to print infinite lethal ammo at sectechs by restocking them this allows them to lethal without needing to think about it. This aims to force detectives to think more carefully about using the 3 lethal rounds they can obtain without Armory, down from the map dependent approximately 5-6 without cargo restock cartridges.

## Changelog 

```changelog
(u)JORJ949
(*) Lethal .38 rounds removed from sectechs. Increased lethal .38 rounds in armory. Increased stun .38 rounds in sectechs.
```

### DIFF
--- a/code/modules/vending/vending.dm
+++ b/code/modules/vending/vending.dm
@@ -1499,12 +1499,11 @@ TYPEINFO(/obj/machinery/vending/medical)
 		product_list += new/datum/data/vending_product(/obj/item/clothing/head/helmet/hardhat/security, 4)
 		product_list += new/datum/data/vending_product(/obj/item/device/pda2/security, 2)
 		product_list += new/datum/data/vending_product(/obj/item/sec_tape/vended, 3)
-		product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/a38/stun, 2)
+		product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/a38/stun, 3)
 		product_list += new/datum/data/vending_product(/obj/item/implantcase/counterrev, 3)
 		product_list += new/datum/data/vending_product(/obj/item/implanter, 1)
 		product_list += new/datum/data/vending_product(/obj/item/paper/book/from_file/space_law, 3)
 		product_list += new/datum/data/vending_product(/obj/item/device/flash/turbo, rand(1, 6), hidden=1)
-		product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/a38, rand(1, 2), hidden=1) // Obtaining a backpack full of lethal ammo required no effort whatsoever, hence why nobody ordered AP speedloaders from the Syndicate (Convair880).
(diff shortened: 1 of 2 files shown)


TheRyeGuyWhoWillNowDie:
considering some maps are missing detective's office ammo (D2 for example) please map edit those to make up for this 🙂

JORJ949:
> considering some maps are missing detective's office ammo (D2 for example) please map edit those to make up for this 🙂 (also kind of sad that "responsible" RP detectives are being punished for the behaviors of classic players)

I would honestly rather remove those (so they cant be used in zipguns) however maps are meant to have varying amounts of items, and were RP detectives really using this that much considering the vendor needs to be hacked?

NightmareChamillian:
Big disagree on this. If a classic detective is powergaming hard enough to
1: foresee the need to have a shitload of lethals
2: hack a sectech, with multitool, insuls, and screwdriver
3: to into cargo, grab restock carts, and mass-print bullets
That's something you should really be adminhelping instead of just making a PR about. As it stands, this is going to nerf sleeper agents and mindhacks more than anything-- being brave enough to dart into sec or det and hack the sectech should offer a reward. _**ESPECIALLY**_ because you can't hack armory doors anymore.

> I would honestly rather remove those (so they cant be used in zipguns) 

Please god no. Zip guns have already been nerfed into the ground 38 special is the only thing onstation you can shoot from them that isn't 22.

I'm fine with the concept of this PR, but the execution is just lacking. Hacking vending machines for contraband goodies is a core gameplay component. If you want lethals to be harder to get, just make it so that restock carts don't refill the extended inventory items-- they're not supposed to be in there, anyway.

JORJ949:
Great points that I do agree with. The point of the PR is not to stop powergaming infinite lethals (because yeah, that is an ahelp thing) but to limit their availability and make them a rarer resource to discourage their usage over stuns. 

Unfortunately this also does nerf sleeper/mindhack (and other antag) detectives but I am honestly fine with that, detective is already a very strong antagonist job  and a popular mindhack target because of the relative ease they have in killing, especially security members. One lethal speedloader can put an armoured HoS deep enough into crit  that they aren't getting back up (or having time to call for help), especially if you follow up with stuns thus letting you raid armoury for more lethals.

The zip gun point is a good thing and I'm not really sure why I mentioned zip guns when my actual reason was that I'd like more map consistency, to which the better option is adding them to other offices.

JORJ949:
Requesting your review because you discussed this over discord

TobleroneSwordfish:
Mmmm I don't really see the need for this, as Millian said I don't think non-antagonists abusing ammo restocks is a big enough of an issue to warrant this.

PR #21110, opened 2024-10-18
Space + left click auto picks up an item on a tile

FlameArrow57:
[PLAYER ACTIONS][FEATURE][QoL][BALANCE]

## About the PR 
Makes it so holding space and then left clicking a tile, while not holding anything, will pick up the highest weight item on that tile

## Why's this needed? 
Mostly QoL, for faster actions. Allows for more easily picking up items with smaller sprites. Useful for pick up items layered over people or just running to a tile with an item you want to pick up fast

Pros:
-Makes certain fights more fun by removing micro delays where people pick up items/weapons that are dropped
-More QoL for stuff like hunter spear/syndicate dagger

Cons:
-Reduces "skill" in picking up dropped items
-Gap where two people are fighting and one doesn't have knowledge of the macro

Pro/con (not sure)
-Almost guarantees the attacker who has an advantage will pick up an item from a victim who's dropped it

## Changelog 

```changelog
(u)FlameArrow57
(*)You can now hold Space and Left Click anything on a tile to automatically pick up the highest weight item
```


code/mob/living/carbon/human.dm
```diff
@@ -1139,13 +1139,26 @@
...
+					var/turf/T = get_turf(target)
+					var/list/items = list()
+					var/obj/item/I
+					for (var/i = 1 to min(10, length(T.contents)))
+						I = T.contents[i]
+						if (istype(I) && !I.anchored)
+							items += I
+					if (length(items))
+						I = items[length(items)] // top-most item is listed last in content
+						I.attack_hand(src.client.mob)
```
Tarmunora, line 1153: `attack_hand()` should never be called directly. Use `AttackHand()`, or perhaps `interact_verb()`, which calls `client.Click()` on the atom

Tarmunora (changes requested):

Valtsu0:
This could use a combat click

https://github.com/user-attachments/assets/db0f5b4d-12c8-4760-a185-fd7541a64af6

FlameArrow57:
> This could use a combat click

Fixed, ty

FlameArrow57:
Couldn't figure out how to do top-most item in a stack, if it's even possible. Just decided to make it only work if there's a single item on a tile. Otherwise, if multiple, it'll just act like you're clicking your cursor as normal. Think it might be more intuitive for cases where it's hard to tell what's the top-most item

TobleroneSwordfish:
I think having it always pick up something is more intuitive, making it sometimes randomly fail because you didn't notice there was a bullet casing on the floor seems bad.
In terms of picking the top-most item, I asked around Coderbus and the Byond discord and I don't think that's possible unless we change how items are handled to internally sub-layer them with a bunch of tiny fractional layers (ew)
A better idea might be making it prioritize the highest `W_CLASS`? It would require iterating the full turf contents to find the highest one, buut since this has a combat click delay now it shouldn't be too heavy performance wise?

FlameArrow57:
> I think having it always pick up something is more intuitive, making it sometimes randomly fail because you didn't notice there was a bullet casing on the floor seems bad. In terms of picking the top-most item, I asked around Coderbus and the Byond discord and I don't think that's possible unless we change how items are handled to internally sub-layer them with a bunch of tiny fractional layers (ew) A better idea might be making it prioritize the highest `W_CLASS`? It would require iterating the full turf contents to find the highest one, buut since this has a combat click delay now it shouldn't be too heavy performance wise?

Rip. Do you want it to pick up heaviest weight class (along with random pick if some items have the same value), or just completely random?

FlameArrow57:
It random picks from highest weight item now

code/mob/living/carbon/human.dm
```diff
@@ -1139,13 +1139,29 @@
...
 					usr = src
 					var/mob/living/living_target = target
 					living_target.give_item()
 					return
+			else if (src.client.check_key(KEY_THROW) && !src.equipped() && BOUNDS_DIST(src, target) <= 0)
+				var/turf/T = get_turf(target)
+				var/list/items = list()
+				var/obj/item/I
+				var/highest_weight = 0
+				for (var/i = 1 to min(10, length(T.contents)))
```
TobleroneSwordfish, line 1154:
```suggestion
				for (var/obj/item/I in T.contents)
```
Doing it this way lets you take advantage of Byond's basic type filtering optimizations and should easily be fast enough to just iterate the entire turf contents. Bear in mind that pathfinding code is doing this for every turf it checks, so doing it maximum once every few seconds on a user action is absolutely fine.

For reference, with a limit of 10 I was finding the weight priority system was breaking on any table with a lot of things on it, partly because it was also iterating lighting overlays, underfloor cables etc.

code/mob/living/critter.dm
```diff
@@ -543,6 +543,22 @@ ADMIN_INTERACT_PROCS(/mob/living/critter, proc/modify_health, proc/admincmd_atta
...
+			I = T.contents[i]
+			if (istype(I) && !I.anchored)
+				if (!("[I.w_class]" in items))
+					items["[I.w_class]"] = list()
+				items["[I.w_class]"] += I
+				highest_weight = max(highest_weight, I.w_class)
+		if (length(items))
+			I = pick(items["[highest_weight]"])
+			I.pick_up_by(src)
+			return
```
TobleroneSwordfish, line 561: This should be a proc on `/mob/living` instead of copy pasted down here.

TobleroneSwordfish (changes requested):

TobleroneSwordfish:
I'll testmerge for now just to make sure there aren't any weird exploits or to find out if this completely breaks combat balance somehow.

Tarmunora (approved):

PR #21116, opened 2024-10-19
Security utility towel

JORJ949:
## About the PR 
Adds a security towel to the security apparel vendor to clean off the blood they will likely get covered in at some point
image

## Why's this needed? 
Encourages cleaning yourself after ~~brutally murdering the clown~~ valiantly fighting traitors!

## Changelog 

```changelog
(u)JORJ949
(*)Added security towels found in the security apparel vendor
```


code/obj/submachine/weapon_vendor.dm
```diff
@@ -409,6 +413,12 @@
 	path = /obj/item/device/prisoner_scanner
 	description = "A device used to scan in prisoners and update their security records."
 
+/datum/materiel/utility/sectowel
+	name = "Security Towel"
+	path = /obj/item/cloth/towel/security
+	description = "A Towel used to clean off the clow- criminal blood!"
```
Glamurio, line 419: I know it's minor, but don't capitalize towel here.

Glamurio (changes requested):

code/obj/submachine/weapon_vendor.dm
Glamurio: That made me chuckle. I was speaking of the description.

JORJ949: >:(

Valtsu0:
I know sec has the same color scheme, but that looks a lot like a syndie item. Maybe its fine?

JORJ949:
> I know sec has the sameish color scheme, but that looks a lot like a syndie item. Maybe its fine?

Palette is directly taken from the secoff uniform. Thats just them having the same colour scheme

TobleroneSwordfish (approved):
Seems cool, it's a pretty low "value" utility purchase but I'm sure people will like it anyway!

JORJ949:
> Seems cool, it's a pretty low "value" utility purchase but I'm sure people will like it anyway!

If the vendor system supports making things with different costs this could be a great way to balance it but currently the donuts are by far the most popular option and for good reason, they have saved me many times and often save a lengthy medbay trip. Other things like the RecordTrak and Forensic scanner can often be found elsewhere without costing a utility credit.

Sord213:
IMO I'd personally rather see the items in the vendor be actually useful for security work. This seems like it'd be better in the clothing vendor to me.

PR #21122, opened 2024-10-19
Fixes unintentional misplaced tile in Donut2 Kitchen

MintyPhresh:
## About the PR 
A tile under the door between the Cafeteria and the shared Kitchen backroom had the wrong style. This PR fixes that single tile.

## Why's this needed? 
It looks stupid like it is now, and I know it wasn't intentional because I was the one that placed it there.

### DIFF
--- a/maps/donut2.dmm
+++ b/maps/donut2.dmm
@@ -43391,7 +43391,7 @@
 	icon_state = "line1"
 	},
 /obj/mapping_helper/access/bar,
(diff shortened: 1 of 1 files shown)


Glamurio:
This year's most niche PR contender, I think.

Glamurio (approved):

Valtsu0:
#21124 is gonna win that price

PR #21124, opened 2024-10-19
Fixes Shitty Grill typo

MintyPhresh:
## About the PR 
When lighting the grill, the message would say "You light the shitty grill ." I'm getting rid of the space between "grill" and "."


TobleroneSwordfish:
Peak PR

PR #21129, opened 2024-10-20
add a medical booth to oshan

Retrino:
## About the PR 

adds a medbooth and expands the janitor closet to have a hall facing window. relocates the emergency storage equipment to the maintenance room next to escape. alters maintenance around the area to be more of a staffie hangout spot, to conserve the maint gambling machine and add a little hangout. moves the paint machine to the warehouse below cargo, which is public access

image
image

also adds a couple lights in areas that are super dark roundstart

## Why's this needed? 
- give docs an option to hang out outside of the northern side of the station
- med booths are kind of a staple to most rotation maps, especially the ones shaped like an O
- convenience, but lacks a morgue pipe. may go back in and figure out how to reasonably fit a fourth pipe into the hallway in a different pr

## Changelog 

```changelog
(u)Retrino
(*)Oshan now has a medical booth.
```

### DIFF
--- a/maps/oshan.dmm
+++ b/maps/oshan.dmm
@@ -1,8 +1,16 @@
 //MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
 "aaa" = (
-/obj/reagent_dispensers/foamtank,
-/turf/simulated/floor/plating/random,
-/area/station/maintenance/south)
+/obj/disposalpipe/trunk/north{
+	dir = 2
+	},
+/obj/machinery/disposal,
+/obj/machinery/light/incandescent/harsh{
+	dir = 8
+	},
+/turf/simulated/floor/purple/side{
+	dir = 9
+	},
+/area/station/janitor/supply)
 "aab" = (
 /obj/table/glass/reinforced/auto,
 /obj/item/plate,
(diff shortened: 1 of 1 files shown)


JORJ949:
You seem to have added a couple lights and a shield generator not listed in the PR description, though they are very minor it would probably be worth doing these in separate PRs in the future.

Retrino:
> You seem to have added a couple lights and a shield generator not listed in the PR description, though they are very minor it would probably be worth doing these in separate PRs in the future.

shield gen was just moved so that there was a public one since it was the only machine i didnt see readily available that was in the old storage. the lights were just kinda missing. totally forgot to list them tho

Glamurio:
Not a fan of the new layout. Feels very awkward. I'd prefer if we simply moved the janitor room two tiles to the left, merge it with the medical booth wall, and add a door facing the hallway. That way the maintenance feels a bit more open, without a 1 tile dead-end hallway.

(Pink is the new room spot, yellow is doors)
image

flrsh:
When we first made Oshan, we intentionally did not add a medbooth. This was due to a couple reasons such as it being fairly straightforward (compared to some other maps) to reach medbay proper and there, at the time, being an issue with overabundances of supplies, including meds and wanting to address that in our design. However! Obviously the game has changed a ton since like, four or five years ago or whenever Oshan was released. So if people really want to add a medbooth, we're (I checked with mbc) not opposed to it. But, in that case, mbc recommends expanding maint to accommodate this instead of trying to fit it into the existing area. I agree with that and I think that as long the general area still looks cohesive, you should feel free to take up / make more room in order to have a nice medbooth. I personally also think it'd be more interesting to have a non-exactly-rectangular room.

One caveat is that we don't want more medkits in the medbooth, but all the regular equipment (including a vendor) should be fine.

Retrino:
will do some revamping with yalls suggestions soon

flrsh (approved):

PR #21133, opened 2024-10-20
Reorganize NTSC inventory

JORJ949:
## About the PR 
Reorganizes the NTSC inventory to be much more alive a secoff's
- Swap candy heart and security whistle to put whistle in pouch
- Start with space suit and gas mask equipped to clear backpack up a bit
- Baton now starts active
- Moved space helmet and handkerchief to belt
- Pouch starts in pocket
- Remove 500c cash
image
image
image
image

Also splits up the secbelt spawn lists onto individual lines (mostly so I could see all the NTSC belt gear)

## Why's this needed? 
NTSC backpack was actually overfilling depending on selected traits (in my case it wasnt spawning the 300c I get for spawning as a high paid job thanks to short sighted and NT loyalist) 
NTSC as a latejoin/admin spawn job is supposed to be ready to immediately get to work and the equipped suit, mask and active baton help this.

## Changelog 

```changelog
(u)JORJ949
(+)Reorganized the NTSC's inventory, resulting in less clog.
```


code/obj/item/stun_baton.dm
```diff
@@ -522,3 +522,11 @@ TYPEINFO(/obj/item/baton/ntso)
 		src.is_active = FALSE
 		usr?.show_text("The [src.name] is now open and unpowered.", "blue")
 		src.process_charges(-INFINITY)
+
+/obj/item/baton/ntso/active //active subtypes so NTSC baton spawns ready to go
```
TobleroneSwordfish, line 526: May as well just make this the default for the type? Regular batons start on iirc.

PR #21134, opened 2024-10-20
Refactor syndicate buylist

Glamurio:
[Internal] [Code Quality] [QoL] [Secret]

## About the PR 
Changes `/datum/syndicate_buylist/item`, `/datum/syndicate_buylist/item2` and `/datum/syndicate_buylist/item3` to `/datum/syndicate_buylist/items`, a list.

Refactors `/datum/syndicate_buylist/run_on_spawn()` to put items purchased into the hand, rather than on the floor, except for surplus crate contents.

Changes any behavior that had to do with  `/datum/syndicate_buylist/item` to instead iterate over the list.

## Why's this needed? 
Having purchases contain a list of items is cleaner instead of having them be hardcoded values. Spawning items in your hand is just QoL in line with many other systems in the game.


Glamurio:
Looks like the chicken grenades will need a secret fix.

code/obj/storage/crates.dm
```diff
@@ -338,14 +338,16 @@
...
+				if(length(item_datum.items) == 0) continue
+				for (var/item in item_datum.items)
+					var/obj/item/I = new item(src)
+					I.Scale(NESTED_SCALING_FACTOR**nest_amt, NESTED_SCALING_FACTOR**nest_amt) //scale the contents if we're nested
+					if (owner)
+						item_datum.run_on_spawn(I, owner, TRUE, owner_uplink)
+						var/datum/antagonist/traitor/T = owner.mind?.get_antagonist(ROLE_TRAITOR)
+						if (istype(T))
+							T.surplus_crate_items.Add(item_datum)
+					telecrystals += item_datum.cost
```
Tarmunora, line 350: Indentation error - looks like this should be outside the `for item in items` loop.

Tarmunora (changes requested):

code/obj/storage/crates.dm
Glamurio: Ooops, looks like a faulty merge

code/datums/syndicate_buylist.dm
```diff
@@ -18,12 +18,8 @@ ABSTRACT_TYPE(/datum/syndicate_buylist)
...
 	/// Name of the buylist entry
 	var/name = null
-	/// A typepath for the item that will be spawned when the datum is purchased
-	var/atom/item = null
-	/// A typepath for the second, optional, item that will be spawned when the datum is purchased
-	var/atom/item2 = null
-	/// A typepath for the third, optional, item that will be spawned when the datum is purchased
-	var/atom/item3 = null
+	/// A typepath for the items that will be spawned when the datum is purchased
+	var/list/atom/items = list()
```
ZeWaka, line 22:
```suggestion
	/// Typepaths of the items that will be spawned when the datum is purchased
	var/list/atom/items = list()
```

ZeWaka (approved):

PR #21136, opened 2024-10-20
Refactor QM traders

Glamurio:
[Internal] [QoL] 

## About the PR 
Refactors `/datum/trader` to add rarities `ITEM_COMMON`, `ITEM_UNCOMMON` and `ITEM_RARE`, which can subsequently be used with `amount_of_items_per_rarity` to iterate over items and add them depending on rarity.

As such, all `base_goods_sell` and `base_goods_buy` have been adjusted to work within the new system.

Some liberty was taken when placing common and uncommon items, primarily to establish a rough equivalent of item occurrences as we had before, with the exception of guaranteeing the presence of a few select items. But if any changes need-be made, I'm not particularly attached to any of the current choices.

## Why's this needed? 
The old code was crusty and very rigid, we had no way to easily differentiate between different spawn pools or adjust how common any given item should be. This PR serves both as a refactor to that old system, and as a setup for new rarities or exceptions to be implemented to customize individual trade pools more easily (and cleaner, too).

Methane cannister are now always purchasable from Buford, which was the original intent from this PR, which then spiraled into refactoring the whole system. Some items such as weed seeds will also always be stocked by Buford.

## Changelog 

```changelog
(u)Glamurio & ZeWaka
(+)Traders now stock certain items more frequently than before, such as methane.
```

### DIFF
--- a/_std/defines/trader.dm
+++ b/_std/defines/trader.dm
@@ -1,0 +1,4 @@
+// Rarities to determine item amounts in traders
+#define ITEM_COMMON 1
+#define ITEM_UNCOMMON 2
+#define ITEM_RARE 3
--- a/_std/defines/trader.dm
+++ b/_std/defines/trader.dm
@@ -1,4 +1,4 @@
 // Rarities to determine item amounts in traders
 #define TRADER_RARITY_COMMON 1
-#define TRADER_RARITY_UNCOMMON2
+#define TRADER_RARITY_UNCOMMON 2
 #define TRADER_RARITY_RARE 3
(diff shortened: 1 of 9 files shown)


_std/defines/trader.dm
```diff
@@ -0,0 +1,4 @@
+// Rarities to determine item amounts in traders
+#define ITEM_COMMON 1
+#define ITEM_UNCOMMON 2
+#define ITEM_RARE 3
```
TobleroneSwordfish, line 4: Proobably best to rename these to `TRADER_RARITY_COMMON` or similar to distinguish them from the `ITEM_RARITY_COMMON` defines which already exist.

TobleroneSwordfish (changes requested):
One minor thing, code looks good otherwise

PR #21142, opened 2024-10-21
Syndicate Research Facility Fenrir

JORJ949:
## About the PR 
Adds the Syndicate Research Facility Fenrir as an explorable area on the long range teleporter. Details as to how to reach it, threats and loot will be in a comment below so if you are reading this in #github-log and want spoilers go read that.

## Why's this needed? 
Originally just my first dip into azones I thought I may aswell PR this for feedback, more places to explore is always fun and I wanted to expand on the disk locations by not just having them all sat in the observatory


JORJ949:
image

HOW TO GET HERE:
Coordinate disk can be found in the large drone in the debris field

NOTABLE LOOT:
- Flare Gun
- Syndicate AI parts and hologram module
- AI Core frame
- NT Blue space suit
- XS Syndicate Fanny Pack (1 slot)
- A couple pieces of syndicate clothing (turtlenecks, shoes)
- Syndicate announcement computer (needs agent card to use)
- Evil red syndicate tables
- Evil red syndicate IDs (3 to be exact, used to get around the facility)

THREATS:
- Syndicate gunbots
- NAS-T turrets (can't be picked up by any potential thieves)

LORE:
The facility was a joint operation between various syndicate companies to research and improve their various weaponry and tech, the most notable of which being a syndicate AI unit. NT Operations discovered this location and infiltrated the facility for a while before removing the laws from the facility AI, said AI had developed a hatred for the syndicate and gladly murdered the entire crew via various methods and upgraded them into more gunbots.

Randomplayer2or3:
Really cool, though in my opinion the loot is rather on the low end considering the effort (bots on the large drone, multiple gunbots, NAS-T turrets, a goddamn crusher). For reference, it’s closest counterpart in terms of theming and danger, the drone factory, has the following: 2 Pod HE rounds, Heavy disruptor array, gunbot blueprint, Gyrojet (which according to the code, the bullets deal more damage the longer they travel, but since you only get 6 shots not much people know this), while here you get like a flare gun, NT suit, AI parts that you can easily get from Carl with the Agent ID from [REDACTED] that is needed to access the drone factory mentioned above.

JORJ949:
> Really cool, though in my opinion the loot is rather on the low end considering the effort 

I was really worried about making the loot overpowered (I think I started working on this before powergaming rule) though with a rule against powergaming and the ability to make future nerfs I would be open to increasing loot. Originally you could reach the syndicate AI laws and there was an EMP grenade. In my experience with this azone I found it extremely easy (probably because I made it and know the threats), the crusher is lit and fairly easy to spot and the bots can be avoided if you know where they are with the sole exception of the final one, before which there are 2 brute medkits, largest threat is the NAS-Ts in the armory which can be ran past with time. However considering this azone will probably kill you a couple times to teach you it should probably have more than what it does.

Randomplayer2or3:
Since this is basically the space map's version of the NSS Polaris, maybe add in the Alastor rifle and thruster blueprints here? The threat level is slightly higher with the turrets and the crusher but I see that the gunbot's stats are the same as Ketchup and Mustard.

flappybatpal:
The general quality of work seems decent although I have not yet actually inspected the map in the editor to double check but I have a number of problems:
1. Players only need to sprint past two NASTs and engage with one gunbot to access the entire content. Two of the gunbots are ignorable after you note where they are on your first visit.
2. Due to the above this location is trivial with a barrier as you can defeat everything with reflect. Most people won't have access to one but it's a trick.
3. Not sure about making the unavailable pod wars NT Spacesuit quite easily accessible even if it's no better than a standard EVA suit.
4. The syndicate ID cards will give you partial access to science.

Not sure about the red syndicate mask considering we intentionally made it much harder to get.

JORJ949:
1. Moved the avoidable gunbots to unavoidable area
2. Knowing what's in an azone already makes the azone trivial if you prep for it
3. This isnt specifically the podwars one, the podwars one is /pilot, this is just a reskinned EVA suit, you can already obtain a syndicate one in a couple places. This specific subtype is currently used exactly nowhere (unless its in secret)
4. Removed
5. I would be fine with changing it to an NT one and putting it in the NT hidey hole, but that would make the Syndicate Armory feel lackluster.

PR #21149, opened 2024-10-22
AI players receive atmos alerts

JORJ949:
## About the PR 
While not a "good" level atmosphere, air alarms will send an alarm to AI units.

2 minor known issues:
- Air alarms only seem to start processing status a few seconds after spawning, meaning all new alarms built/initialized will send an alert to the AI that is cleared when it catches up a few seconds later, though there shouldnt be any AIs to receive these messages before the map air alarms initialize.
- image all the AI trigger alarm code uses single or double letter vars and I can't be bothered spending the time right now to figure out what is causing this or if its just aa dev zone thing.

## Why's this needed? 
Makes air alarms more useful, the more useful they are the more we will realise where they are missing.

## Changelog 

```changelog
(u)JORJ949
(*)AIs now receive atmospheric alerts from air alarms.
```

### DIFF
--- a/code/obj/machinery/alarm.dm
+++ b/code/obj/machinery/alarm.dm
@@ -23,6 +23,7 @@
 	/// keeps track of last alarm status
 	var/last_safe = ALARM_NOPOWER
 	var/datum/gas_mixture/environment
+	var/alertingAI = FALSE // Does this alarm currently have an AI alarm active
 
 	/// this is a list of safe & good partial pressures of each gas. If all gasses are in the good range, the alarm will show green. If any gas is outside the safe range, the alarm will show alert. Otherwise caution.
 	/// most of these values are taken from lung.dm
@@ -157,6 +158,18 @@
 	light_ov.plane = PLANE_SELFILLUM
 	src.UpdateOverlays(light_ov, "light")
 
+	var/list/cameras = list()
+	for_by_tcl(C, /obj/machinery/camera)
+		if(get_area(C) == get_area(src))
+			cameras += C
+
+	for_by_tcl(aiPlayer, /mob/living/silicon/ai)
+		if ((safe == ALARM_GOOD) && src.alertingAI)
+			aiPlayer.cancelAlarm("Atmosphere", get_area(src), src)
+			src.alertingAI = FALSE
+		else
+			aiPlayer.triggerAlarm("Atmosphere", get_area(src), cameras, src)
+			src.alertingAI = TRUE
 
 	if(alarm_frequency)
 		post_alert(safe)


code/obj/machinery/alarm.dm
```diff
@@ -157,6 +158,18 @@
...
+		if(get_area(C) == get_area(src))
+			cameras += C
+
+	for_by_tcl(aiPlayer, /mob/living/silicon/ai)
+		if ((safe == ALARM_GOOD) && src.alertingAI)
+			aiPlayer.cancelAlarm("Atmosphere", get_area(src), src)
+			src.alertingAI = FALSE
+		else
+			aiPlayer.triggerAlarm("Atmosphere", get_area(src), cameras, src)
+			src.alertingAI = TRUE
```
amylizzle, line 172:
```suggestion
	for_by_tcl(aiPlayer, /mob/living/silicon/ai)
		if ((safe == ALARM_GOOD) && src.alertingAI)
			aiPlayer.cancelAlarm("Atmosphere", get_area(src), src)
			src.alertingAI = FALSE
		else
			for_by_tcl(C, /obj/machinery/camera)
			if(get_area(C) == get_area(src))
				cameras += C
			aiPlayer.triggerAlarm("Atmosphere", get_area(src), cameras, src)
			src.alertingAI = TRUE
```

no point building the camera list if you're not gonna use it

code/obj/machinery/alarm.dm
```diff
@@ -157,6 +158,17 @@
...
+	for_by_tcl(aiPlayer, /mob/living/silicon/ai)
+		if ((safe == ALARM_GOOD) && src.alertingAI)
+			aiPlayer.cancelAlarm("Atmosphere", get_area(src), src)
+			src.alertingAI = FALSE
+		else
+			for_by_tcl(C, /obj/machinery/camera)
+				if(get_area(C) == get_area(src))
+					cameras += C
+			aiPlayer.triggerAlarm("Atmosphere", get_area(src), cameras, src)
+			src.alertingAI = TRUE
```
ZeWaka, line 171: why are we running through every camera for every AI
should just be once for each

ZeWaka (changes requested):

TobleroneSwordfish:
Okay I went and glared really hard at the AI alarm code so it's more readable now. The duplication is probably just null-named cameras on devtest (devtest moment)

ZeWaka (approved):

code/obj/machinery/alarm.dm
```diff
@@ -157,6 +158,18 @@
 	light_ov.plane = PLANE_SELFILLUM
 	src.UpdateOverlays(light_ov, "light")
 
+	if ((safe == ALARM_GOOD) && src.alertingAI)
+		for_by_tcl(aiPlayer, /mob/living/silicon/ai)
+			aiPlayer.cancelAlarm("Atmosphere", get_area(src), src)
+		src.alertingAI = FALSE
```
TobleroneSwordfish, line 164:
```suggestion
	if (safe == ALARM_GOOD)
		if (src.alertingAI)
			for_by_tcl(aiPlayer, /mob/living/silicon/ai)
				aiPlayer.cancelAlarm("Atmosphere", get_area(src), src)
			src.alertingAI = FALSE
```
Otherwise if it's not currently alerting an AI it will send an alert even if the current state is `ALARM_GOOD`, probably causing your known issue?

TobleroneSwordfish (changes requested):

PR #21151, opened 2024-10-22
Lower amount it takes to uncuff with weak cuffs

Glamurio:
[Game Objects] [Balance]

## About the PR 
Changes the `remove_self_multiplier` to `0.25` for `/obj/item/handcuffs/guardbot` and `/obj/item/handcuffs/tape_roll`.

Normal cuffs take `60 SECONDS` to uncuff, ziptie and tape now take `20 SECONDS` each.

## Why's this needed? 
Came to be from a thread stating frustration regarding self-uncuffing. Albeit necessary from a design standpoint, it seems okay (to me) to make an exception for weaker cuffs both from a balance standpoint and realism standpoint. It still takes a decent amount of time to uncuff, but this should make Beepsky cuffing at least less overbearing.

Relevant thread:
https://forum.ss13.co/showthread.php?tid=23302

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Weak cuffs, such as ziptie or tape, can now be removed faster.
```

### DIFF
--- a/code/datums/actions/actions.dm
+++ b/code/datums/actions/actions.dm
@@ -886,10 +886,7 @@
 
 	onStart()
 		..()
-		if(owner != null && ishuman(owner) && owner.hasStatus("handcuffed"))
-			var/mob/living/carbon/human/H = owner
-			duration = round(duration * src.cuffs.remove_self_multiplier)
-
+		duration = round(duration * src.cuffs.remove_self_multiplier)
 		owner.visible_message(SPAN_ALERT("<B>[owner] attempts to remove the handcuffs!</B>"))
 
 	onUpdate()
--- a/code/datums/actions/actions.dm
+++ b/code/datums/actions/actions.dm
@@ -747,7 +747,7 @@
 						target.internal.add_fingerprint(owner)
 
 /datum/action/bar/icon/handcuffSet //This is used when you try to handcuff someone.
-	duration = 40
+	duration = 4 SECONDS
 	interrupt_flags = INTERRUPT_MOVE | INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION
 	icon = 'icons/obj/items/items.dmi'
 	icon_state = "handcuff"
@@ -821,7 +821,7 @@
 			O.show_message(SPAN_ALERT("<B>[owner] handcuffs [target]!</B>"), 1)
 
 /datum/action/bar/icon/handcuffRemovalOther //This is used when you try to remove someone elses handcuffs.
-	duration = 70
+	duration = 7 SECONDS
 	interrupt_flags = INTERRUPT_MOVE | INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION
 	icon = 'icons/obj/items/items.dmi'
 	icon_state = "handcuff"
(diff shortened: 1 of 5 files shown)


flrsh:
I feel like we have a pretty ingrained understanding that if you want to prevent someone from uncuffing you gotta keep moving them around and I'm not a fan of introducing a large inconsistency here. Imo we should just make the uncuff times shorter for weaker cuffs.

Glamurio:
> I feel like we have a pretty ingrained understanding that if you want to prevent someone from uncuffing you gotta keep moving them around and I'm not a fan of introducing a large inconsistency here. Imo we should just make the uncuff times shorter for weaker cuffs.

Essentially boils down to a similar effect, and I do agree with the assessment. I've gone ahead and changed the values to about a 4th of regular cuffs. Normal cuffs take `60 SECONDS` to uncuff, ziptie and tape now take `20 SECONDS` each.

Vaskritaya:
> I feel like we have a pretty ingrained understanding that if you want to prevent someone from uncuffing you gotta keep moving them around and I'm not a fan of introducing a large inconsistency here. Imo we should just make the uncuff times shorter for weaker cuffs.

I'm not entirely convinced these are equivalent. While the effect on ziptie cuffs are more or less similar (chance of escape because a door was open vs chance of escape because sec couldn't arrive fast enough; lower occurrence of being interrupted as a civilian either way), the effect on duct tape is pretty different (and more of a nerf/annoyance) in several contexts. While the PR could be changed to only affect ziptie cuffs, that might also feel exceptional and could be confusing for players. 

My proposal would be to only apply this change to ziptie cuffs, leaving duct tape as it currently is, since it's a traitor item.
If need be, the inconsistency could then be mitigated by a future PR implementing a second weak cuff variety [such as this one](https://forum.ss13.co/showthread.php?tid=23321), to better reinforce the idea of there being two tiers of cuffs.

JORJ949:
I disagree with nerfing duct tape, its a traitor item that already sees little to no use (at least on classic)

Glamurio:
> I disagree with nerfing duct tape, its a traitor item that already sees little to no use (at least on classic)

I had the same thought when changing it, but also felt like Vaskritaya that if we make it ziptie-only it may be misleading.

The reason for why it sees little use has probably less to do with its uncuffing time, and more with how antagonists typically operate. We have antagonists that are good at ambushing in the dark and antagonists that are typically better at sabotage and being loud. And traitors fall into the latter category mostly, with some exceptions.

What's imo really needed more is several additions to a sneaky gameplay-style for traitors, most items they have that fill that niche are focused on "killing silently" and not "abducting or incapacitating". Traitors strength is versatility, but that (should) usually come at the downside of being not quite at good at one specific thing.

Tapes now take `40 SECONDS` each. Thought it'd be good to meet in the middle. Give traitors an "almost as good as real cuffs" item, but still incentivize traitors to steal cuffs from Security.

Tarmunora:
Code breaks handcuff removal, and ziptie cuffs *already* take only between 15 and 18 seconds to remove as per:
```dm
calcTime = istype(src.handcuffs, /obj/item/handcuffs/guardbot) ? rand(15 SECONDS, 18 SECONDS) : rand(40 SECONDS, 50 SECONDS)
```

PR #11958, opened 2022-11-13
Revert "Sterile mask blocks miasma"

pali6:
Reverts goonstation/goonstation#11957

Imo they should only block miasma.

PR #11957, opened 2022-11-13
Sterile mask blocks miasma

MarkNais:
## About the PR 
Adds `BLOCKSMOKE` to the sterile masks found all over medbay.

## Why's this needed? 
Per idea thread: https://forum.ss13.co/showthread.php?tid=20079
`BLOCKSMOKE` is the laziest change to implement sterile masks blocking miasma. It also blocks all other smoke - maybe this is too strong? Unlike gas masks they don't work as internals and don't block your identity.
Embalming fluid and morgue trays are easy to access regardless.
I'm not a balance nerd.
Tested.
image

## Changelog 

```changelog
(u)MarkNstein
(*)Sterile masks now block miasma and other smokes.
```

PR #21153, opened 2024-10-22
Make sterile masks block miasma

Glamurio:
[Feature] [Chemistry] [Balance]

## About the PR 
Adds a new flag `BLOCKMIASMA`, and adds said flag to `/obj/item/clothing/mask/surgical`, also adds `ismiasmaimmune()` as a check to simplify usage.

Adds a new `var/exception` param to `/obj/fluid/airborne/just_do_the_apply_thing()`, `/obj/fluid/airborne/force_mob_to_ingest`, as well as `/datum/reagents/trans_to` and `/datum/reagents/reaction`. Said exception param expects any reagent ID and will simply skip over them.

## Why's this needed? 
Miasma can make any medbay nigh-unusable if unmanaged. And managing it is surprisingly difficult for a department that has to deal with it a lot. Gas masks aren't always readily available unless you go out of your way to hunt one, vacuums (and by extension, janitors) are the only way to get rid of it and both of these aren't necessarily present always, and there's only so much oxygen you're willing to burn through with your internals before you throw in the towel and just inhale it.

It's been suggested a handful of times sterile masks would be a great solution against this, they're underused, cool, and readily available in all of medbay, and I think are a great solution to this issue.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Sterile masks now render you immune to breathing in miasma.
```

### DIFF
--- a/_std/defines/clothing.dm
+++ b/_std/defines/clothing.dm
@@ -67,7 +67,8 @@ var/list/all_slots = list(SLOT_BACK, SLOT_WEAR_MASK, SLOT_L_HAND, SLOT_R_HAND, S
 #define ONBACK						(1<<17)
 /// can be work on the belt
 #define ONBELT						(1<<18)
-
+/// block miasma inhalations (medical mask)
+#define BLOCKMIASMA					(1<<19)
 
 //Suit blood flags
 #define SUITBLOOD_ARMOR 1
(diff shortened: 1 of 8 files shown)


Valtsu0:
I tough this was already a thing

seems #11957 got reverted in #11958 because it blocked all smoke, not just miasma

Good change

PR #21154, opened 2024-10-22
General Alert Computer TGUI Minimap

pgmzeta:
[ui][station systems][feature][rework]
## About the PR 
* Convert the general alert computer UI to use a minimap controller UI.
* No longer able to manually clear alerts.
* Track individual fire, power, and atmos alarms separately
* Create defines/datum for tracking alerts
* Remove the unused atmospheric/alarm computer
* Move the alert computer to its own file
* Have area-wide power/fire alarms update the alert computer via radio packet
* APCs, Fire Alarms, and Air Alarms add an icon to the `MAP_ALERT` minimap when triggered.

## Why's this needed? 
Make the alert computer generally more useful and prettier.

## Screenshots
image

## Changelog 

```changelog
(u)glowbold
(+)The alert computer displays alerts on a minimap, and tracks atmospheric, fire, and power alarms separately.
```


JORJ949:
The colons look a little weird like they are meant to have data after them, IIRC the computer receives both atmos and fire alarms so you could put the type of alarm triggered after the colon

TobleroneSwordfish:
This is a decent recreation of the current UI, so I'm fine with merging it for now buut also the current UI is barely functional, it makes no attempt to display what type of alarm occurred and doesn't even appear to receive fire alarms.
A better way to display the data would probably be with sections per alarm type, displaying "severity" inline if applicable.

pgmzeta:
I'll rework including more detail on the alert source/type and absorb any useful bits from `/obj/machinery/computer/atmosphere/alerts` into this UI

Garash2k:
Neat! I would suggest using icons such as [bell-slash](https://fontawesome.com/icons/bell-slash?f=classic&s=solid) for the buttons and move "clear alarm" to a tooltip. When adding more details, consider having a fire and air icon somewhere for the alarm type and coloring them red/blue.

Garash2k:
Also looking at the tgui project, there's a total of 4(!!) alert windows. It would be nice to either knock all of them at the same time, or keep them in mind so that bits of code can be re-used for them later.

[General Alert Computer](https://github.com/orgs/goonstation/projects/7/views/1?pane=issue&itemId=36363816)
[Atmos Alert Computer](https://github.com/orgs/goonstation/projects/7/views/1?pane=issue&itemId=36362972)
[AI Alerts and Cyborg Alerts](https://github.com/orgs/goonstation/projects/7/views/1?pane=issue&itemId=36360760)

tgui/packages/tgui/interfaces/AlertComputer.tsx
```diff
@@ -0,0 +1,137 @@
+import { Button, LabeledList } from 'tgui-core/components';
+
+import { useBackend } from '../backend';
+import { Window } from '../layouts';
+
+interface AlertComputerData {
+  alerts: Partial<Record<string, AlertData>>;
```
mordent-goonstation, line 7: I don't see where the key is really used in this data structure. You're doing length checks on it, but not actually using the `key` string for anything. What does the key of this Record represent?

pgmzeta: it's used to lazily pack the data on the DM side

code/obj/machinery/computer/alert.dm
```diff
@@ -0,0 +1,155 @@
...
+/obj/machinery/computer/general_alert/ui_data(mob/user)
+	. = ..()
+	.["alerts"] = list()
+	for (var/datum/alert/alert in src.alerts)
+		var/area_ckey = ckey(alert.zone)
+		if(!(area_ckey in .["alerts"]))
+			.["alerts"][area_ckey] = list()
+		.["alerts"][area_ckey]["area_ckey"] = area_ckey
+		.["alerts"][area_ckey]["area_name"] = alert.zone
+		.["alerts"][area_ckey][alert.kind] = alert.severity
```
mordent-goonstation, line 119: Your response to the previous comment re: UI data structure had me come look here. This line in particular sets of a code smell alert (geddit? Because of what you're working on).

If I'm reading correctly, this only gives one alert per area, as if multiple entries in `src.alerts` have the same area the latest one overwrites the first one.

I'm not opposed to grouping by area, but subsequent alerts should add a new list item, not overwrite the fields on the previous one in the same area, right?

pgmzeta: updated the alerts to have one datum per area ("zone"), which made this logic a lot clearer

Garash2k:
Small usability issue with the UI: colored buttons usually indicate whether they're disabled or not (grey or green) or hint at the type of action they'd do (i.e. red for destructive stuff.) With the current UI, we end up with bunch of green buttons that can't actually be clicked.

Instead of putting everything in a button, I would suggest coloring the *alarm type icon* and *severity text*, and having a small, tooltipped, button on the right, who's only content is the bell-slash icon, that's either green or grey.

Also for "aligning everything" purposes, I imagine a grid is gonna be needed. Once there's one, I'd suggest adding an Atmospherics / Fire / Power header

God awful mockup yippie:

image

pgmzeta:
in hindsight, writing a component with "Row" in the name and having it not be a table was silly.

pgmzeta:
Questions I'm musing over for this

- [ ] The alert minimap icons are on the alarm objects themselves. Should clearing the alert also clear the minimap icon? If so, we would need a way to communicate back to the object.
- [ ] Do we need alert-clearing functionality? It does not actually clear the alarms themselves, just the alert computer's listing of it. Only the person who cleared it would know.
- [ ] Should clearing alerts on one computer clear it on all others?

tgui/packages/tgui/interfaces/AlertComputer.tsx
```diff
@@ -0,0 +1,127 @@
...
+}
+
+export const AlertComputer = () => {
+  const { data } = useBackend<AlertComputerData>();
+  const { alerts } = data;
+  return (
+    <Window title="Current Alerts" width={600} height={420}>
+      <Window.Content scrollable>
+        <Table>
+          <Table.Row header fontSize="1.1em" className="candystripe">
```
mordent-goonstation, line 26:
```suggestion
          <Table.Row header className="candystripe">
```
Leave core styling alone.

tgui/packages/tgui/interfaces/AlertComputer.tsx
```diff
@@ -0,0 +1,127 @@
...
+            <Table.Cell header width="200px">
```
mordent-goonstation, line 27:
```suggestion
            <Table.Cell header>
```
Avoid hard-coding the width of at least one column, so if someone makes the container wider/narrower it scales with it. Example: new area name that is longer than you've assumed here.

tgui/packages/tgui/interfaces/AlertComputer.tsx
```diff
@@ -0,0 +1,127 @@
...
+            <Table.Cell header collapsing width="100px">
+              <Icon name="fire" mr="5px" />
+              Fire
+            </Table.Cell>
+            <Table.Cell header collapsing width="100px">
+              <Icon name="bolt" mr="5px" />
+              Power
+            </Table.Cell>
+          </Table.Row>
+          {Object.values(alerts).map((value, index) => {
```
mordent-goonstation, line 43: Per typing of AlertComputerData, `alerts` is an array so doing `Object.values` is not needed and you can just do `alerts.map(...)`. Is the typing wrong, or is this part wrong?

tgui/packages/tgui/interfaces/AlertComputer.tsx
```diff
@@ -0,0 +1,127 @@
...
+            return <AlertRow key={index} alert={value} />;
```
mordent-goonstation, line 44: Avoid index keys if you can. It looks like you should be able to key by `value.area_ckey`?

tgui/packages/tgui/interfaces/AlertComputer.tsx
```diff
@@ -0,0 +1,127 @@
...
+          <Button
+            icon="bell-slash"
+            disabled={severity === 0}
+            onClick={() => act(`clear_${kind}`, { area_ckey: area_ckey })}
+            tooltip="Reset alarm"
+            mr="5px"
+          />
+        </Flex.Item>
+        <Flex.Item>{AlertButtonText(severity)}</Flex.Item>
+      </Flex>
```
mordent-goonstation, line 102: Not necessarily "wrong", but you don't seem to be using any of the actual properties of `Flex` or `Flex.Item` for actual layout. Are you just trying to ensure `Button` and the text are kept on one line?

tgui/packages/tgui/interfaces/AlertComputer.tsx
```diff
@@ -0,0 +1,127 @@
...
+    case 1:
+      return 'average';
+    case 2:
+      return 'bad';
+    default:
+      return 'good';
+  }
+};
+
+const AlertButtonText = (severity: number | null): string => {
```
mordent-goonstation, line 118:
```
const getAlertButtonText = (severity: number | null): string => {
```

tgui/packages/tgui/interfaces/AlertComputer.tsx
```diff
@@ -0,0 +1,127 @@
...
+            mr="5px"
+          />
+        </Flex.Item>
+        <Flex.Item>{AlertButtonText(severity)}</Flex.Item>
+      </Flex>
+    </Table.Cell>
+  );
+};
+
+const AlertButtonColor = (severity: number | null): string => {
```
mordent-goonstation, line 107:
```
const getAlertButtonColor = (severity: number | null) => {
```
Convention for how you had it would imply `AlertButtonColor` was a component, class, or enum. It's a getter function. Also, no need to specify return type, that can be inferred by TS. Fine if you want to leave that part in, it doesn't break anything.

mordent-goonstation (changes requested):
Reviewed tgui interface only.

TobleroneSwordfish (changes requested):
- The map is cool but should probably close when the UI does, see `ui_close`
- If there's no way to manually clear alerts and alerts don't clear automatically then the "Current Alerts" list isn't very accurate or useful? If you want it to be a static historical list of alerts then it should be labelled as such, or if you want it to be a list of current alerts that reflects the map then it needs to either clear automatically the same way the map icons do or have a manual button to dismiss the alert.
- More of a minor point but the whole screen map might be better off as part of the UI (see how the AI map works), since currently silicons can't use it remotely and the UI window starts awkwardly over the top of the map when first opened.

pgmzeta:
need to rework for ai/cyborg alerts

pgmzeta:
closed in lieu of #21724 instead

PR #21158, opened 2024-10-23
Add CommasterLite to Mining outposts

JORJ949:
## About the PR 
Adds CommasterLite program that has all the functionality of commaster except it cannot recall the shuttle
Adds commasterlite consoles to:
- The mining outpost in the asteroid field
- The Pasiphae on nadir (replaced full comms console+added dish)
- Oshan's Tech outpost

Consoles could use a better name than "Light communications console" but I couldnt think of one

## Why's this needed? 
More alternative methods to call the shuttle are good but you dont want someone to recall from offstation

## Changelog 

```changelog
(u)JORJ949
(*)The mining outpost is now equipped with communications consoles incapable of recalling the shuttle.
```

### DIFF
--- a/code/modules/networks/computer3/communication.dm
+++ b/code/modules/networks/computer3/communication.dm
@@ -14,6 +14,7 @@
 	var/obj/item/peripheral/network/powernet_card/pnet_card = null
 	var/tmp/comm_net_id = null //The net id of our linked ~comm dish~
 	var/tmp/reply_wait = -1 //How long do we wait for replies? -1 is not waiting.
+	var/cant_recall = FALSE //can this program recall the shuttle
 
 	var/transmit_title = null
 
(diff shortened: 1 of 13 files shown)


TobleroneSwordfish:
Why would a syndicate spy outpost be able to call an NT shuttle though?

JORJ949:
> Why would a syndicate spy outpost be able to call an NT shuttle though?

What if it called the syndicate version of the shuttle (do these exist for all maps?) with some additional flavour of NT wondering why the syndicate decided to offer their crew a shuttle

Other explanations are that they stole the tech and codes because they want their operatives as trusted by NT as possible and them dying sat on an exploded station isnt very good for business

Honestly no explanation will fit perfectly this is more to provide alternate options to calling the shuttle if comms dishes go down, I didn't put much thought into lore

JORJ949:
Moved from post to NT outposts because as pointed out, it doesnt particularly make sense.

JORJ949:
And over the course of a day I have determined this feature is probably not needed and very minor. I should think before I code...

PR #21160, opened 2024-10-23
Cleanbot whistle

JORJ949:
## About the PR 
Generalizes /datum/component/secbot_command into generic bot_command with /security subtype for secbots
Adds janitorial whistle that when used can change a cleanbots target or turn it off
Janitorial whistle spawns on the mime and in the janitech vendor

## Why's this needed? 
EVIL cleanbots have ruined mime artwork too long. That ends TODAY!

## Changelog 

```changelog
(u)JORJ949
(*)Added janitorial whistle for controlling cleanbots.
```


444Portal:
Finally a whistle we can give to borgs without abuse

JORJ949:
> 
> Is it possible to turn cleanbots on using it?

It is not, you will have to turn them back on by hand.

PR #21162, opened 2024-10-24
Add Language artifact activation stimulus

FlameArrow57:
[SCIENCE][FEATURE]

## About the PR 
This PR adds a new Language stimulus for activating artifacts, using a puzzle similar to Wordle/Mastermind (intended to not be as complicated as those but not much simpler)

How it works is for each artifact activated this way, there will be a prechosen 5 letter word. However, you only need to guess any word that has the same vowel positions as the chosen word.
Examples:
Chosen word is Apple, acceptable word is Alpha
Chosen word is Peach, acceptable word is Leach

Speaking by the artifact will make it give a hint message.
If you choose an invalid word (meaning not 5 letters, or not an actual word), the artifact will give a Dull chime
If you choose a valid word with too many vowels, the artifact will give a Grumpy chime.
For a word with the right number of vowels, it will give a High chime for each vowel in the right position, and a Low chime for each vowel in the wrong position

Idea of the "language" trigger is that it is based on valid language patterns (vowel usage for Human language) for activation

Note: Wanting to get the idea approved, but was also wanting this to be changed to a weighted pick from anywhere from 3 to 6 letter words, and possibly make it so some can be activated by ghosts

## Why's this needed? 
-Adds a slight more "puzzle" mechanic that is missing from other triggers
-Would be cool to have an interactive stimulus

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new artifact stimulus - Language. See minor changes for details
(+)Language activated artifacts are activated by saying a 5 letter word with vowels in correct places
(+)Speaking near one of these artifacts will give a hint from the artifact
(+)Using too many vowels will give a Dull chime. A vowel in the right place will give a High chime. A vowel in the wrong position will give a Low chime
```


Glamurio:
More artifact triggers, cool! I love the idea. My main concern is the amount of times you'll hear it randomly make a series of beeps while you're having a conversation next to it. But that's probably a minor issue. If you want to randomize the amount of letters (i.e. 3-6), consider adding a scientist training hint that tells you the word length.

Big no on ghost activation, I think that's asking for trouble. Ghosts are not considered part of the game for most instances, and the last thing you want is a John Staffman's ghost activating a black hole in a dingy maints nobody even knows about.

PR #21165, opened 2024-10-24
Expanded Photocopier

LorrMaster:
## About the PR 
This PR expands the photocopier to perform more tasks that are within it's grand domain of scanning and copying things. The photocopier is now able to scan & print wanted posters, discount credits, and paper copies of books & newspapers. It can also now be emagged, allowing it to print credits & create exact copies of existing IDs. Copying IDs requires blank IDs to be inserted beforehand. 

The photocopier can now receive commands via packets, allowing bureaucrats to perform their grueling copying tasks from the comfort of their office chair. Network information on how to do this can be easily printed from the photocopier directly.

I hope that the code is also a little easier to navigate than it was before. 

## Why's this needed? 
- It allows the photocopier to print various paper items that it couldn't before. 
- Allows another machine to be emagged for potential fun.
- Users can manually print information on how to send packets to the photocopier, which could be helpful for newer players to learn with if they run into it.
- Code for the photocopier should be more readable than it was before. For instance, each item no longer uses its own variable for storing scans.

## Changelog 

```changelog
(u) LorrMaster
(*) Photocopier can now scan and print wanted posters, discount credits, paper copies of books & newspapers, booklets, contents of folders, reagent scanner results, and information from IDs.
(+) Packets can be used by bureaucrats on the photocopier to perform their grueling tasks from the comfort of their office chair. Details can be printed from the photocopier directly.
```


JORJ949:
Changelog is way too long, most only need to be a couple lines as this is what shows in game it only needs to show important bits players need to know, for example:
(*) Emagged photocopiers can now print credits and copy IDs.
It needs to stick to the original formatting of the changelog stuff so that the bot can process it into a changelog

LorrMaster:
I updated the changelog to three points that should be the most directly relevant to players.

Glamurio:
A bit concerned about how powerful it is to have the photocopier be able to scan IDs. It sort of invalidates the use for the agent card, unless you really want listening outpost access.

LorrMaster:
Well it is only able to create an exact copy of another ID, and you first need to get a blank ID card to copy the information over to. So I don't think it is that powerful because by that point you can use the EMAG to try and roll for ID Computer access (though now that I think about it, the photocopied IDs should probably be marked as emagged even if the original ID isn't). If that still takes something away from the Agent Card, I could make the printing process itself more obvious & risky or come up with another barrier to require more preparation. The most extreme case could be to require inserting a card with ID Computer access before printing, but I think that would be overkill.

Tarmunora (changes requested):
Not a fan of the emag effect. We're trying to steer away the emag being "emag turns on crime mode". and lean more towards "emag _breaks_ things", like access locks, electronic safeties, etc.
And being able to photocopy ID cards falls well on the side of "magic crime mode", esp considering that ID cards are printed at fabs out of conductor and crystal, not printed on paper.

LorrMaster:
> Not a fan of the emag effect. We're trying to steer away the emag being "emag turns on crime mode". and lean more towards "emag _breaks_ things", like access locks, electronic safeties, etc. And being able to photocopy ID cards falls well on the side of "magic crime mode", esp considering that ID cards are printed at fabs out of conductor and crystal, not printed on paper.

I think that you may be misunderstanding how the emagging effect works for IDs. The emagged photocopier does not use paper for copying IDs. You have to place spare IDs into copying tray if you want it to copy IDs. So it is more like reading all the information off of the scanned ID and imprinting it onto the blank ID. Not exactly realistic, but not at the level of building an ID out of a sheet of paper.

Though the emagging effect also lets you copy credits, so the idea is very much in line with "turning on crime mode" by disabling security settings or whatever. So I'd like some clarification on what kind of design philosophy you're looking for in relation to the EMAG here. Thank you!

JORJ949:
Here is an example changelog and how it appears on goonhub/ingame
image
image

each line is a different entry in the changelog and the date/categories are done automatically

LorrMaster:
> each line is a different entry in the changelog and the date/categories are done automatically

I think that it should be good now.

code/obj/machinery/photocopier.dm
```diff
@@ -29,246 +39,819 @@ TYPEINFO(/obj/machinery/photocopier)
...
+		src.print_type = "paper"
+	proc/scan_bible(var/obj/item/bible/B, var/mob/user)
+		scan_setup(B, user)
+		boutput(user, "You open the Bible, press a passage onto the scan bed, and press start...")
+		src.icon_state = "scan_book"
+		effects_scanning(B)
+		src.print_info["name"] = "Matthew 25:14-30"
+		src.print_info["desc"] = "Photocopy of a Bible passage."
+		// Source: https://www.vatican.va/archive/ENG0839/__PVY.HTM
+		src.print_info["info"] = file2text("strings/books/bible_matthew25_14_30.txt")
```
ZeWaka, line 740: yeah, no. not stirring up this issue. nix this entirely, please.

_**Do Not Merge until this is removed**_

ZeWaka (changes requested):

code/obj/machinery/photocopier.dm
LorrMaster: > yeah, no. not stirring up this issue. nix this entirely, please.

Oh, okay? I didn't think it was any more of an issue than the Bible already being in the game. I mean what else would the Bible do, aside from smiting you of course.

Tarmunora (approved):

code/obj/machinery/photocopier.dm
```diff
@@ -29,246 +39,798 @@ TYPEINFO(/obj/machinery/photocopier)
...
+			var/isBlank = TRUE
+			// isBlank = isBlank && I.name == "identification card"
+			isBlank = isBlank && I.icon_state == "id"
+			isBlank = isBlank && I.item_state == "card-id"
+			isBlank = isBlank && I.access.len == 0
+			isBlank = isBlank && I.registered == null
+			isBlank = isBlank && I.assignment == null
+			isBlank = isBlank && I.title == null
+			isBlank = isBlank && I.emagged == FALSE
+			isBlank = isBlank && I.money == 0 */
```
TobleroneSwordfish, line 497: Minor gripe but please remove this and the other ID cloning code instead of just commenting it, I know it sucks to remove something you put effort into but otherwise we end up with files filled with weird unused and unmaintained code everywhere.

TobleroneSwordfish (changes requested):
Very cool additions, I'm sure someone will spend half an hour printing 10 credit notes and get arrested for it 👍

code/obj/machinery/photocopier.dm
LorrMaster: You're absolutely right. For the "isBlank" code in particular, should I move that to the ID card class as a new proc in case someone else wants to use it, or just get rid of it entirely since it probably wouldn't be maintained up until that point?

TobleroneSwordfish: Uhh I'd probably just remove it, I can't think of another reason you'd want to check that exact condition.

JORJ949: Other use cases where we currently want to check an ID state are:
Agent card for their description (just checks if the name is still "Agent Card")
Emagging, of which there is a stat for.
Other use cases like this would probably just use their own method

PR #21166, opened 2024-10-25
Cable Handcuffs

itchybones:
## About the PR 
Adds cable handcuffs, craftable with 4 lenghts of cable. They can be cut instantly with a snipping/cutting tool, and have a strenght of 1, with a self-removal multiplier of 0.2.

TODO:
- [x] obj
- [x] make it craftable
- [ ] can be cut instantly with cutting/snipping tools, dropping 4 lengths of wire
- [ ] sprite (preferably matching with wire color)

(This is my first proper feature PR, so it's probably gonna take a while to finish)
## Why's this needed? 
Quoted from my [idea thread](https://forum.ss13.co/showthread.php?tid=23321&pid=207410#pid207410):
> I just feel like there should be a way to restrain people as non-sec player. If you roll sleeper or arcfiend, for example, you don't have an uplink to order ductape with! Of course, sleepers sometimes get a little loot stash, but I think having more options is neat.

## Changelog 

```changelog
(u)Chasu
(*)Added craftable weak cable handcuffs, made with 4 lenghts of any cable
(+)They can be cut instantly by using a cutting/snipping tool on the cuffed person.
```


Tarmunora:
After some discussion in admin channels, we have decided we aren't interested in adding craftable handcuffs - antagonists can/should just steal (ziptie)cuffs from sectechs, and crew has pins.

PR #21175, opened 2024-10-26
Lowers the PTL destruction threshold to function as a mining laser

vortex1942:
[balance][station systems][game objects]
## About the PR 

Enables the PTL to function as a mining laser at lower power levels.

On Asteroid turfs 1-5MW of laser power have increased chances (20% per MW) of destroying asteroid walls. A 5MW laser takes approx 4.5 minutes to mine an asteroid (see video)

Balance impact should be minimal as this does not change the threshold for other objects, so station damage would still require a 100MW laser. and the lower power requirement reduces the risks of miners dying. And can be an automated, but slower way of mining

And also protects the mining magnet forcefield from being destroyed, (Though this could be expanded to energy fields or a dedicated lasersink object.

## Why's this needed? 
The PTL has gameplay potential to be used as a mining laser however currently it requires a significant setup time, and for an engine to be running at 100mw 

This would make it more accessible for players and make a new gameplay element to an existing object.

## Changelog 

```changelog
(u)Vortex1942
(*)The PTL can now be used as a mining laser at low power outputs.
```

https://github.com/user-attachments/assets/11c90765-1bee-442e-b318-4bea77a1a0f2

### DIFF
--- a/code/modules/power/pt_laser.dm
+++ b/code/modules/power/pt_laser.dm
@@ -413,15 +413,19 @@
 				istype(A, /obj/machinery/vehicle) || \
 				istype(A, /obj/machinery/bot/mulebot) || \
 				istype(A, /obj/machinery/the_singularity) || /* could be interesting to add some interaction here, maybe when singulo behviours are abstracted away in #16731*/ \
+				istype(A, /obj/forcefield/mining ) || \
 				isrestrictedz(A.z))
 			continue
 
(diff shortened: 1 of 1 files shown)


vortex1942:
@TobleroneSwordfish it would appear we've both had the same idea.

vortex1942:
Already added in full a hours before opening

PR #21179, opened 2024-10-27
Mind Protection implants block wraith possession

JORJ949:
## About the PR 
Prevents wraiths from possessing anyone with a mind protection health implant.

## Why's this needed? 
It is super easy right now to just possess a secoff/hos/ntsc/cap into killing another sec and then themselves if they didnt already have silver cuffs on hand, now you will have to kill them first to possess them.

## Changelog 

```changelog
(u)JORJ949
(*)Trickster wraith possession can no longer be used on those with Mind protection health implants.
```


Tarmunora:
Not really a fan of extending blanket immunities like this. We are going to try https://github.com/goonstation/goonstation/commit/43b95092641c891aa07bfd33a0cc8dec76d16314 first, and then maybe come back to this if it remains a squeaky wheel

PR #21196, opened 2024-10-29
Add Talisman artifact

FlameArrow57:
[SCIENCE][FEATURE][BALANCE]

## About the PR 
This PR adds a new tiny sized artifact, the Talisman. 

This is an artifact that provides a small/minor buff to you when held in your hand, pocket, or anywhere in your inventory. There are six possible beneficial effects, listed below per category/theme:
Talisman of 
Swiftness:
-Increases your movespeed by random value from 5 - 15 %
Spacewalking:
-Provides cold resistance randomized from 25 - 75%
Elements:
-Provides heat and cold protection, each randomized from 1 - 25%
Power:
-Gives you extra max hp randomized from 10 - 50
Fortune:
-Gives you random stacks of 100 - 500 cash every few minutes
Protection:
-Gives you a passive, random 1 - 10% brute damage reduction, burn reduction, or tox reduction, or a small chance to do all three with reduced value

-These benefits are not visible to the user, but using the artifact in-hand gives an effect-related hint message as to the effect
-Talisman artifacts have a unique examine hint of "It seems magical somehow."
-Periodically a user of a Talisman artifact will have a purple shade cross over them
-Talisman artifacts will deactivate if EMP'd
-Talisman artifacts can be found in Container artifacts
-Talisman artifacts will periodically (< 1 minute) roll a fault chance, making faults relatively frequent
 
## Why's this needed? 
-Would be cool to have passive status effect artifact
-Idea of a protection artifact/talisman sounds cool

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new Talisman tiny-sized artifact! See minor changelog for details
(+)Holding this artifact in your hand, pocket, or inventory will provide its passive, beneficial effect to you
(+)Beneficial effects include increased speed, extreme cold protection, minor cold and heat protection, increased max health, wealth, and damage reduction
(+)Users will periodically have a purple shimmer over them
(+)Talisman artifacts can be disabled with EMPs, are found in container artifacts, and have frequent fault chance rolls
```


code/obj/artifacts/artifact_items/talisman.dm
```diff
@@ -0,0 +1,138 @@
...
+				src.brute_prot = rand(1, 10)
+				src.burn_prot = rand(1, 10)
+				src.tox_prot = rand(1, 10)
+
+
+
+/datum/artifact/talisman
+	associated_object = /obj/item/artifact/talisman
+	type_name = "Talisman"
+	type_size = ARTIFACT_SIZE_MEDIUM
```
Glamurio, line 120: Makes more sense for this to be `ARTIFACT_SIZE_TINY` instead, in my mind. We also have like 10 medium-sized artifacts and only 2 tiny-sized.

code/obj/artifacts/artifact_items/talisman.dm
```diff
@@ -0,0 +1,138 @@
...
+/*
+-Random cold protection (from space) from 25% - 75% - the artifact feels somewhat cold
+-Random set of heat + cold protection from 0% - 25% each - feels hot and cold in different spots
+-Random incoming brute, burn, tox reduction from 0 - 10% each - holding it makes you feel safe somehow
+-Random max health increase from 0 - 50 - you get a sense of power from it
+-Random amounts of money appearing in your inventory between 0 - 200 credits. does not add it if your inventory is full. - holding it makes you feel lucky somehow
+-Random movement speed increase from 0 - 5% - you feel drafts of air surrounding this artifact/you feel air moving quickly around this artifact
+*/
+
+/obj/item/artifact/talisman
```
Glamurio, line 17: Any chance you could make this an `/obj/item/clothing/suit`? It would be neat to wear them sort of like an amulet around your neck. Kind of how the HoS medal works. Still has the same effects, but would allow some interesting extra slot management.

Glamurio (changes requested):
Really cool idea, conceptually. Fills in a cool niche we don't have.

code/obj/artifacts/artifact_items/talisman.dm
FlameArrow57: Tiny size is used for the power cells, so would need to use those sprites. May do, will think about it a bit more

code/obj/artifacts/artifact_items/talisman.dm
FlameArrow57: Could, but it would require sprites and all + removes the part of this in that it provides its magical effects without worn, if restricted to clothing

code/obj/artifacts/artifact_items/talisman.dm
Glamurio: Yeah, and I think those tiny sprites work better for a "talisman". Protective charms are usually small items, rocks or gems, etc.

code/obj/artifacts/artifact_items/talisman.dm
Glamurio: Wouldn't *need* sprites technically, and it should be easy enough to make some variants using the HoS medal as reference for size and just making 1 for each origin.

The magical effect should remain in pocket or worn is what I'm saying. I'm not arguing to change that. But it allows you to balance having it in your pocket or suit slot depending on convenience.

code/obj/artifacts/artifact_items/talisman.dm
FlameArrow57: Added!

code/obj/artifacts/artifact_items/talisman.dm
FlameArrow57: I probably won't plan on it for now, agreed it would be cool, but I think wearing it is more on the extra side

code/modules/status_system/statusEffects.dm
```diff
@@ -2926,14 +2926,17 @@
 				"Luck is on your side... wait, what...?"))
 			boutput(src, SPAN_NOTICE(msg))
 
-/datum/statusEffect/talisman_fault
-	id = "art_talisman_fault"
+/datum/statusEffect/talisman_held
+	id = "art_talisman_held"
 	unique = FALSE
 	visible = FALSE
 	effect_quality = STATUS_QUALITY_NEGATIVE
```
Tarmunora, line 2941: this feels like an odd choice - just a copypaste thing or is there a reason for it?

code/mob/living/carbon/human.dm
```diff
@@ -2281,7 +2281,7 @@
 	src.force_equip(I, slot)
 	return TRUE
 ///Tries to put an item in an available backpack, pocket, or hand slot; will delete the item if unable to place.
```
Tarmunora, line 2283: update doc comment

code/obj/artifacts/artifact_items/talisman.dm
```diff
@@ -0,0 +1,205 @@
...
+	icon = 'icons/obj/artifacts/artifactsitemS.dmi'
+	associated_datum = /datum/artifact/talisman
+	w_class = W_CLASS_SMALL
+	var/associated_effect
+	var/mob/living/carbon/human/active_user = null
+
+	// swiftness vars
+	var/swiftness_mod = 0
+	// fortune vars
+	var/money_amt = 0
```
Tarmunora, line 19: variable is never read

code/obj/artifacts/artifact_items/talisman.dm
```diff
@@ -0,0 +1,205 @@
...
+					APPLY_ATOM_PROPERTY(user, PROP_MOB_TALISMAN_TOX_REDUCTION, src, src.tox_prot)
+
+		src.active_user = user
+		user.setStatus("art_talisman_held", null, src)
+
+	proc/remove_effect_from_user()
+		if (!src.active_user)
+			return
+
+		src.active_user.delStatus("art_talisman_held")
```
Tarmunora, line 107: how does this interact with having multiple talismans active?

Tarmunora (changes requested):

code/modules/status_system/statusEffects.dm
FlameArrow57: srry, could you clarify which line you're talking about?

Tarmunora: oh, it being `STATUS_QUALITY_NEGATIVE`

FlameArrow57: Oh ok. Yeah I can change it to neutral if you think it would be better, I was considering artifact faults negative, bit awkward with the visual effect too in one status effect... dunno

FlameArrow57: I'll just split it up into two

Tarmunora: > I'll just split it up into two

I instead have just made godmode block artifact faults from occurring entirely for the user

FlameArrow57: okay cool

code/obj/artifacts/artifact_items/talisman.dm
FlameArrow57: Yeah wasn't working correctly, should be fixed now

FlameArrow57:
> Any particular reason why you're using a single unique statuseffect. rather than just holding a ref to the statuseffect datum (which is returned by `setStatus()` on the artifact itself for the `delStatus()` later?

I didn't know that was available. I guess I could see it going either way, whichever is best

Tarmunora (approved):

code/obj/artifacts/artifact_items/talisman.dm
```diff
@@ -0,0 +1,208 @@
...
+	var/swiftness_mod = 0
+	// fortune vars
+	var/datum/statusEffect/talisman_fortune/fortune_effect
+	// spacefaring vars
+	var/space_prot = 0
+	// elements vars
+	var/heat_prot = 0
+	var/cold_prot = 0
+	// strength vars
+	var/extra_hp = 0
```
Azrun, line 28: Couldn't these all be collapsed into a single var? The type already differentiates how the var used.

_std/macros/atom_properties.dm
```diff
@@ -261,6 +261,10 @@ To remove:
 #define PROP_MOB_EQUIPMENT_MOVESPEED(x) x("equipment_movespeed", APPLY_ATOM_PROPERTY_SUM, REMOVE_ATOM_PROPERTY_SUM)
 #define PROP_MOB_EQUIPMENT_MOVESPEED_SPACE(x) x("equipment_movespeed_space", APPLY_ATOM_PROPERTY_SUM, REMOVE_ATOM_PROPERTY_SUM)
 #define PROP_MOB_EQUIPMENT_MOVESPEED_FLUID(x) x("equipment_movespeed_fluid", APPLY_ATOM_PROPERTY_SUM, REMOVE_ATOM_PROPERTY_SUM)
+#define PROP_MOB_TALISMAN_SWIFTNESS(x) x("art_talisman_movement_mod", APPLY_ATOM_PROPERTY_SUM, REMOVE_ATOM_PROPERTY_SUM)
+#define PROP_MOB_TALISMAN_BRUTE_REDUCTION(x) x("art_talisman_brute_reduction", APPLY_ATOM_PROPERTY_SUM, REMOVE_ATOM_PROPERTY_SUM)
+#define PROP_MOB_TALISMAN_BURN_REDUCTION(x) x("art_talisman_burn_reduction", APPLY_ATOM_PROPERTY_SUM, REMOVE_ATOM_PROPERTY_SUM)
+#define PROP_MOB_TALISMAN_TOX_REDUCTION(x) x("art_talisman_tox_reduction", APPLY_ATOM_PROPERTY_SUM, REMOVE_ATOM_PROPERTY_SUM)
```
Azrun, line 267: Do these need to be talisman specific?

FlameArrow57: Could you clarify? It's property sum so should be fine with multiple talismans?

code/obj/artifacts/artifact_items/talisman.dm
FlameArrow57: Um if you think it should be done. I don't see much of a benefit in doing so, I just see the var name being made less clear and for where multiple values are chosen it would be a list

_std/macros/atom_properties.dm
Azrun: Could these be applicable outside of talismans? Is so in not sure the word talisman needs to be in the name.

Azrun: Could these be applicable outside of talismans? Is so in not sure the word talisman needs to be in the name.

code/obj/artifacts/artifact_items/talisman.dm
Azrun: Definitely doesn't need to be done. Just an observation that a variable indicating the magnitude of the change would cover the majority of the use since only one type can be active at a time aside from the one protection outlier.

_std/macros/atom_properties.dm
FlameArrow57: Could be yeah. I can change if you'd like, but name can be changed whenever so dunno

Glamurio (approved):

PR #18351, opened 2024-03-29
Cocktail Shaker: Thermal stones & QoL

TDHooligan:
[FEATURE][SPRITES][OBJECTS]
## About the PR 
Adds a new item to the bartender's vending machine (to save them precious inventory space)

**Thermal shaker stones:**
These can be taken in/out of the shaker, and used in-hand to change their mode from 'cool', 'heat' and 'normalise'. These work to heat/cool beverages towards safe temperatures when you shake the cocktail shaker. 

https://github.com/goonstation/goonstation/assets/6061314/e14cf60d-6068-4bbf-a654-936d20550017

The cocktail shaker now shows the temperature of beverages contained inside via its' icon state:
shakerfreezeshakercoolshakerwarmshakerhot
< -- From dangerously cold to dangerously hot -- >
Sprites by Guv.

## Why's this needed? 
Bartender can't actually heat beverages without awkwardly manoeuvering over to a beaker. And beakers aren't big enough. 

More reasons to shake the cocktail shaker too! :)

## Changelog 

```changelog
(u)TDHooligan & Guv
(*)The bartender's vending machine contains 2 handfuls of shaker stones. Put them in your shaker to cool/heat beverages!
```

PR #21211, opened 2024-10-31
Add shaker temperature gauging

TDHooligan:
[QOL][SPRITES][OBJECTS]
## About the PR 

Shakers show temperature of contained drinks, with sprites indicating when they're in unsafe drinking ranges.
Hot: >97C (Burn damage)
Warm: >30C (icky beer)
Cold: <7C (cool beer)
Freezing: <-23C (Burn damage)

image
image
Sprites by Guv. Originally in #18351 

Also fixes to stop infinite recursion of temperature changes in menders.

## Why's this needed? 
Better gauging of temperature for barkeeping. More reasons for cocktail shaker existing.

## Changelog 

```changelog
(u)TDHooligan, Guv
(+)The bartender's shaker now shows if your cocktail is a safe temperature to drink
```


TobleroneSwordfish (approved):
Lgtm, it might be nice to extend something like this to all reagent containers even if it's just a line of description like "it feels hot to the touch"

code/modules/chemistry/tools/patches.dm
```diff
@@ -477,7 +477,7 @@ TYPEINFO(/obj/item/reagent_containers/mender)
 
 	on_reagent_change(add)
 		..()
-		if (src.reagents)
+		if (src.reagents && src.reagents.total_temperature > 330)
```
Tarmunora, line 480: ?

Tarmunora (changes requested):

code/modules/chemistry/tools/patches.dm
TDHooligan: infinite loop triggered by adding reagents_changed callbacks to temperature_reagents. 

makes more sense that heating reagents also counts as changing them, better place to hook it.

Tarmunora: Should also probably check if temp is lower than 270 though.

PR #21217, opened 2024-11-01
Emergency suits in boxes now start folded

NightmareChamillian:
## About the PR 

Emergency suits spawning in boxes now start compacted, so that way they can be taken in or out of the box as many times as desired up **until** you decide to unfold and put it on.

https://github.com/user-attachments/assets/674e5329-214c-4c2b-a900-47edfcda4a51

## Why's this needed? 

Honestly, QoL that's _just_ barely bugged me enough to bother making a PR. Accidentally taking an esuit out, or taking it out only to realize you didn't need to, and now you _have_ to dedicate a whole ass backpack slot for it is a bit of a bother. 

## Changelog 

Not entirely sure one's needed but why not

```changelog
(u)NightmarechaMillian
(+)Emergency suits in boxes now start folded so they can actually fit back into the box
```


code/obj/item/clothing/suits.dm
```diff
@@ -1214,6 +1214,24 @@ TYPEINFO(/obj/item/clothing/suit/hazard/fire/armored)
...
+	desc = "A suit that protects against low pressure environments for a short time. At least, it would be if it hadn't been vacuum-compressed into a small rectangle. You'll have to unfold it before putting it on."
+	icon = 'icons/obj/clothing/overcoats/item_suit_hazard.dmi'
+	icon_state = "emerg_folded"
+	item_state = "emerg"
+	w_class = W_CLASS_TINY
+
+	attack_self(mob/user as mob)
+
+		user.drop_item(src) //clear hands
+		boutput(user, SPAN_NOTICE("You deploy the [src]!"))
```
JORJ949, line 1228: "deploy" sounds weird here, "unfold" would probably be better

Tarmunora (approved):

code/obj/item/clothing/suits.dm
NightmareChamillian: it originally did, but in that case the message would be "you unfold the folded emergency suit" which seemed a little repetitive

PR #21220, opened 2024-11-02
Information map with "You Are Here" pin

pgmzeta:
[feature][game objects][experimental]
## About the PR 
Adds an "Informational Map" display that can be used to bring up a station map. This map is similar to the AI station map, but only has a marker for the map object's current location.

## Why's this needed? 
In-game way for people to figure out where the heck they are on the station.

## Screenshot
image
## TODO
- [X] Better sprite for the in-world map object
- [X] Rework into wall-mounted display machinery


Glamurio:
Pin should be a bright color that starkly contrasts the rest, perhaps even dark outlined.

ZeWaka:
red pin with black outline

code/modules/minimap/minimap_objects/objects/info_map.dm
```diff
@@ -0,0 +1,23 @@
+/obj/info_map
+	name = "Information Map"
+	icon = 'icons/obj/decals/posters.dmi'
+	icon_state = "pw_map"
+	var/atom/movable/minimap_ui_handler/minimap_ui
+	var/obj/minimap/info/infomap
+
+/obj/info_map/New()
```
ZeWaka, line 8: disposing

pgmzeta: what changes would you recommend from the default obj ex_act?

pgmzeta: What changes from the default obj `ex_act` would you recommend?

mordent-goonstation:
How does it work if the wall it's placed on is deconstructed? Can it be dragged around? Can it be deconstructed/placed somewhere else?

pgmzeta:
> How does it work if the wall it's placed on is deconstructed? Can it be dragged around? Can it be deconstructed/placed somewhere else?

Currently, the minimap marker will update with the current location of the map object. It's anchored by default similar to a wall poster.

If preferred, I could make this into a device akin to the status displays, which would give it device decon/recon/mechscan features.

mordent-goonstation:
Based on the positioning, it looks like it occupies the same turf as the wall, and thus could be accessed from either side? I think having it function similar to APCs, in that it visibly sits on a wall but can't be seen from the other side would be better?

Glamurio:
> If preferred, I could make this into a device akin to the status displays, which would give it device decon/recon/mechscan features.

I would actually *really* like  decon/recon/mechscan features, so that people can set up new ones, for RP or otherwise.

code/modules/minimap/minimap_objects/objects/info_map.dm
pgmzeta: now that it's a machine, it's using the default machinery `ex_act`

pgmzeta:
This is now machinery ala status displays, so it is scannable with power on/off and pixel shifting to appear on walls.

pgmzeta:
Please let me know if I should add these to maps in this PR, or hold off until a future PR.

FlameArrow57:
Just happened to find an existing station map sprite in icons/obj/stationobjs.dmi, if you want to use it

image

TobleroneSwordfish (approved):

PR #21221, opened 2024-11-02
Add "Licensed Archaeologist" medal

FlameArrow57:
[SCIENCE][MEDAL][FEATURE]

## About the PR 
Adds a new medal tied to a new scientist objective of having activated 10 artifacts at the end of the shift. You count towards this if you are in range of 5 tiles of an artifact when it's activated

Name: "Licensed Archaeologist"
Pic:
artmedal
Desc: "Complete the Scientist or Research Director objective of activating 10 artifacts in a single shift."
Other: Medal is visible on the medals page before earned (not secret)

## Why's this needed? 
Artlab doesn't have any medals with it right now

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new Scientist/Research Director objective to activate 10 artifacts in a single shift, with a medal tied to it
```

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -59,6 +59,8 @@ datum/mind
 
 	var/show_respawn_prompts = TRUE
 
+	var/artifacts_activated = 0
+
 	New(mob/M)
 		..()
 		if (M)
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -59,8 +59,6 @@ datum/mind
 
 	var/show_respawn_prompts = TRUE
 
-	var/artifacts_activated = 0
-
 	New(mob/M)
 		..()
 		if (M)
--- a/code/obj/artifacts/artifactprocs.dm
+++ b/code/obj/artifacts/artifactprocs.dm
@@ -162,6 +162,9 @@
 	else
 		A.show_fx(src)
 	A.effect_activate(src)
+	for (var/mob/living/L in range(5, src))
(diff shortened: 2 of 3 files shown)


Glamurio:
Would be cool if there was a reward tied to it. Even if it's just something simple.

code/obj/artifacts/artifactprocs.dm
```diff
@@ -162,6 +162,9 @@
 	else
 		A.show_fx(src)
 	A.effect_activate(src)
+	for (var/mob/living/L in range(5, src))
+		if (ishuman(L) || isrobot(L))
```
TobleroneSwordfish, line 166:
```suggestion
		if (L.mind)
```
this is probably what you mean and `L.mind.foo` is unsafe otherwise

code/datums/mind.dm
```diff
@@ -59,6 +59,8 @@ datum/mind
 
 	var/show_respawn_prompts = TRUE
 
+	var/artifacts_activated = 0
```
TobleroneSwordfish, line 62: Consider just putting this on the objective datum unless there's a reason we'd want to track this otherwise?

Tarmunora (approved):

TobleroneSwordfish (changes requested):

code/obj/artifacts/artifactprocs.dm
FlameArrow57: Not exactly, otherwise it will count intangibles, critters, etc

PR #21229, opened 2024-11-04
Change info maps to use native HUD rendering rather than going via ByondUI

TobleroneSwordfish:
## About the PR 
image
Makes the new info maps open as an overlay with a background tint instead of a full TGUI window, also adds point and click GPS functionality to them because it was only a few lines and I should probably atomize that but shrug.

## Why's this needed? 
A lot less janky, maybe looks better?


TobleroneSwordfish:
Hmm maybe?
Without transparency:
image
With a semitransparent background:
image
With a whole screen dimmer effect:
image

PR #21232, opened 2024-11-04
Remove thindows from Cog1 upload

JORJ949:
## About the PR 
Replaces or removes all thindows in Cogmap 1's AI upload and computer core and shifts some other stuff to account for the extra space the full windows take up
image
Also fixes the wall without a corner

## Why's this needed? 
Thindows evil, as are diagonal walls with no wall to connect them

## Changelog 

```changelog
(u)JORJ949
(+)Cogmap 1's AI Upload changed slightly to remove thindows.
```


Tarmunora:
What's with the changes in the hot loop area?

JORJ949:
> What's with the changes in the hot loop area?

I had originally moved the entrance one tile down and needed to make space and forgot to revert those changes when moving it one tile back up.

TobleroneSwordfish (approved):
Lgtm, not sure what the 😕s are for but please comment if you have feedback!!

Valtsu0:
I think that the obervation area made by thindows is pretty important to the feel of cogmap1 upload, which i quite like. As so i don't think thindow bad is good enough reason for it's removal

Valtsu0:
I retract my 😕. The change i was 😕 about (removal of the thindow observational area) wasn't actually this pr, but an older one i missed

JORJ949:
> I retract my 😕. The change i was 😕 about (removal of the thindow observational area in upload) wasn't actually this pr, but an older one i missed

Someone could PR adding windows to the foyer, lots of other foyers have them (though youd probably need to remove a window to the airbridge)

PR #21254, opened 2024-11-07
Misc Mushroom fixes

cheekybrdy:
## About the PR 

- Adds a missing clown holo module
- Fixes incorrect access on a maint door
- makes tables in upload be the auto variant
- adds a larger mail chute to QM, and gives QM some privacy due to lacking it
- adds missing UV lights and mister to botany
- adds surgical scissors to medical
- adds a chicken killer pipe
- changes the mug rack in bar for a cup rack which can work with the espresso machine

## Why's this needed? 
fix good


Tarmunora:
PR title/desc do not match PR contents

PR #21255, opened 2024-11-07
Adds prototype engines to rotation maps

cheekybrdy:
## About the PR 
Adds prototype engines, which were station level derelict engines which could be fixed up by the crew back into rotation. These were traditional engines but weren't the one currently in rotation. In this PR TEG and singlo maps have a nuclear prototype, ocean maps have a singularity one and space nuclear maps have a TEG prototype.

## Why's this needed? 
Gives engineers lategame options and variation, which can feel rather unengaging outside of mechlab after initial setup. Also adds proper backup engines to be rerouted and managed if a engine with irreplaceable parts gets destroyed early.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Added prototype engines back to rotation maps.
```


pgmzeta (changes requested):
- Should have some prototype engine prefabs and use a landmark instead of being inserted directly into the maps ala cogmap1 engine.
- Nuclear power cold side should have valves for going going into space; forcibly cooling down the gas is for emergencies only.
- Nuclear engines should probably not use the random subtype. Removing all those rods before being able to set up the reactor is a lot of tedious work to get it going on top of repairing the engine area. Perhaps fully empty instead with a few rods strewn about?
- These APCs/SMES will count against the engineer crew objective to have them all powered, which feels off for side-content.
- TEG Hotloop inputs have the on/off valve after the pump, usually it's valve then pump.
- No engine stats computer for TEG; if you don't think it should be included, then the `stats_meter`s aren't needed. 
- I don't think it's possible to recreate vacuum floors for the TEG chamber, so the flooring in the chamber should be put in complete. Be sure to add one under the broken chamber window or it will uncontrollably bleed heat into the main chamber.

Edit:
- These show up in the AI/nukie map. Set the `minimaps_to_render_on` variable for the engine/proto and `/area/station/engine/proto` and `/area/station/engine/proto_gangway` to 0 (maybe make a define for `MAP_NONE` in `_std\defines\minimap.dm`?)

TobleroneSwordfish:
Mmm not sure about this, just putting an entire second engine on every map doesn't seem like a good way to give engineering more lategame content. The engine should be at the center of the station where it can have some interaction with the rest of the crew, not randomly off in space. I'd imagine this would lead to either the main engine being entirely abandoned or splitting engineering between those who want to work on the main engine and those who want to work on the spare engine.
A better (albeit much more difficult) approach would be to work on making the other two engine types constructible alongside the singularity.

Glamurio:
I do like the idea of having them be a spawnable prefab though, just for variety. Perhaps a chance for them to spawn on random coordinates, sometimes, on the regular Z level (perhaps including some other prefab spawning)

PR #21261, opened 2024-11-08
Add blue HoS recolour for the "I helped!" medal

JORJ949:
## About the PR 
Adds a blue/NT recolour for HoS outfits claimable via reward for the medal "Suitable? How about the Oubliette?!
image

Changes objective to complete to obtain "Suitable? How about the Oubliette?!" from an antag brigged at roundend to not losing your hat.

Removes HoS objective to have an antag brigged at roundend

## Why's this needed? 
More customisation is always cool and lots of people like blue security (most sec mains I see on 1 often run a blue jumpsuit) and the HoS has all these cool outfits but nothing blue

Brigging an antag at roundend should not be encouraged, there is already another objective that encourages instead taking them to Centcom in cuffs, a far better method of punishment. From what I can tell from the code this objective used to also exist for officers but was removed, likely only left on HoS because they should know better.

## Changelog 

```changelog
(u)JORJ949
(*)New reward for the "I helped!" medal to recolour HoS clothing from red to blue.
(+) HoS "Antagonist in brig at round end" objective removed, associated medal moved to "Don't lose your hat" objective.
```


444Portal:
It being tied to a medal for Security Assistants is very weird to me. I'd be pretty bummed to learn that the medal I just got has a reward thattt... I literally can't use without applying for something.

JORJ949:
> It being tied to a medal for Security Assistants is very weird to me. I'd be pretty bummed to learn that the medal I just got has a reward thattt... I literally can't use without applying for something.

Moved it to the "Suitable? How about the Oubliette?!" medal, removed the old associated objective for that medal (brig antag at roundend) and changed associated objective to "Dont lose your hat"

Cthucky:
That kind of ruins the name of the medal, though.  Perhaps have the security room on the shuttle count as the brig instead?

TobleroneSwordfish:
I know bluesec is kind of a thing but this is a lot less recognizable than the default HoS gear imo. The one on the right could easily be a staffie other than the 2px of boot detail.

Tarmunora:
After further discussion in #dev, we have come to the conclusion that we aren't particularly interested in adding/encouraging blue-HoS

AXLFAN2000:
Though the main focus of this PR was denied, I do think that a rework of the oubliette medal could be real nice in a separate PR as it's currently kind of a revs-only thing lol.  I like Cthucky's suggestion a lot

PR #21264, opened 2024-11-09
Add Curser artifact

FlameArrow57:
[SCIENCE][GAME OBJECTS][FEATURE]

## About the PR 
This PR adds a new large artifact, the Curser

About:
The idea behind the artifact is that upon touching it, it gives you and everyone nearby a curse. 
- After touching, it will become dormant for a short while before activating again
- Curses must be completed or appeased, destroying the artifact will not remove the curse. 
- Talisman artifacts make you immune to curses. For death curses, talismans are destroyed but still block them.
- Cursers can only be Eldritch artifacts

Curses that will be added are below, which are in 3 categories, with 2 per category

Death curses (Will kill you if not appeased):
- Blood curse 
    - Drains your blood until you die. To remove the curse, it needs to drain 600u of blood from yourself and/or people who offer their blood.
- Aging curse
    - Causes you to rapidly age until you die. To remove it, you need to get three people younger than you to touch the artifact (they are not affected)

Survival curses (Need to survive an event or escape):
- Nightmare curse 
    - Causes void creatures that only you can see and damage to appear around you for a short duration. You need to kill enough for the curse to be removed. They can kill you if not dealt with
- Labyrinth curse
    - Teleports you to a labyrinth pocket dimension where you need to complete a maze to escape

Status effect curses (give a status debuff for a short bit):
- Displacement curse 
    - Causes your soul to leave your body, in a new form. It's tangible but can't be harmed. To remove the curse, you need to 
       wait a few minutes, or someone needs to touch the artifact to return you
- Light curse 
    - Makes being in lit areas hurt you for a few minutes, meaning you need to keep the lights off

## Why's this needed? 
-Curser existed as old code that wasn't finished, thought it would be cool to complete it
-Adds a new bad effect artifact that has more player interaction than just clicking it and being done (as well as for traitors using it)

## Changelog 

```changelog
(u)FlameArrow57
(*)Add a new Curser artifact! This is an artifact that gives you 1 of 6 curses when you touch it, as well as others nearby. See minor changelog for more details
(+)The curser artifact will curse whoever touches it as well as a random 2 - 5 nearby people
(+)Curses include: Blood, aging, nightmare creatures, labyrinth, soul displacement, and light curses
(+)Curses are blocked by Talisman artifacts, but for the blood and aging curses the Talisman is also destroyed
(+)Curses give you a status effect that will tell you the condition for removing the curse
```


LorrMaster:
I like those ideas. I feel like these could be used by the evil wishgranter, not exactly the same thing but it only has two wishes.

pgmzeta:
Labyrinth curse is technically feasible to do, let me know if you want help with allocated zones

FlameArrow57:
> Labyrinth curse is technically feasible to do, let me know if you want help with allocated zones

Thanks, I think I've got the allocated zone part down, mostly just random maze generation that I was looking into

FlameArrow57:
@Tarmunora Hey, could you provide thoughts on the ideas here, if they sound okay or any suggested changes?

flappybatpal:
`three people younger than you`
This sounds quite difficult to pull off outside 60+ player RP. Also there is probably quite a few people all at the minimum character age.

FlameArrow57:
> `three people younger than you` This sounds quite difficult to pull off outside 60+ player RP. Also there is probably quite a few people all at the minimum character age.

Note, this is in comparison to the age as the curse makes you older, rather than your initial age. I was hoping that the idea would be it would be harder to get people to touch an artifact that they are being asked to touch, but dunno

Tarmunora:
> @Tarmunora Hey, could you provide thoughts on the ideas here, if they sound okay or any suggested changes?

I think the concept is cool!

However I do have a concern, notably the combination of:
1. curses being AoE
2. Talismans granting curse immunity
3. Death curses being unable to cure without access to the original artiface
Which might make antag use of a curser a bit oppressive. Not a dealbreaker but I think it's worth further consideration, or have an alternate cursebreaking method (chaplain/faith based?) in the wings in case it gets run into the ground.

On more specific critiques, I think that
- Both death curse cures seem trivially solved in artlab by grab-shoving lab monkies at the problem. Unsure if this is the intended outcome but it seems like the most likely way it'll see play
- Labyrinth effect feels like it might be more suitable as a (rare?) alternate prison artifact effect rather than a curse effect, esp if there isn't an active threat in the maze

FlameArrow57:
> > @Tarmunora Hey, could you provide thoughts on the ideas here, if they sound okay or any suggested changes?
> 
> I think the concept is cool!
> 
> However I do have a concern, notably the combination of...:
> 
> On more specific critiques, I think that...

Okay thanks for the input!
I'll make the following changes for now-
- AoE of the curse is limited to the person who uses it + 2 other people, rather than everyone, if that's better?
- Will not make Talismans give curse immunity
- I'll think more on an alternate curse breaking method
- For death curses, agree it may be so with the blood curse. For the aging curse, it won't work with monkeys
- Will think on the labyrinth one and maybe replace it with something else (moving it to the prison art instead if so)

Tarmunora:
> Okay thanks for the input! I'll make the following changes for now-
> 
>     * AoE of the curse is limited to the person who uses it + 2 random other people, rather than everyone, if that's better?
> 
>     * Talismans don't give curse immunity to whoever triggers the art, but it's given to those in the AoE field?
> 
>     * I'll think more on an alternate curse breaking method

To be clear - my concern was less with any of the 3 points on their own, and more how they stacked up together
How do you feel about this as a potential change for death curses specifically:
* Reduce from true AoE to random number (2-5?) targets, and
* talismans only grant immunity to the holder, as normal, but are consumed/destroyed after blocking a death curse

I think that the concept of a traitor using a death curse as a weapon is pretty cool, and wouldn't want to stifle it too hard out of the gate, but I think that having the tactic use talismen as 'ammunition' per se might be a good compromise between letting cool things happen and limiting very cheesy gameplay

FlameArrow57:
> * Reduce from true AoE to random number (2-5?) targets, and
> * talismans only grant immunity to the holder, as normal, but are consumed/destroyed after blocking a death curse

Yeah that sounds good too!

FlameArrow57:
Okay ready for review, may need some numbers re-balancing tho

Not yet implemented:
-Talisman interaction (waiting for PR to be merged)
-Alternate curse breaking method (was thinking access should be difficult enough that completing the requirements of the curse should be preferrable over using this alternate method, maybe an Instrument unbinds the curse?)
-Labyrinth moved to prison artifact (I've left it here, since I was just wanting it to be part of the Curser artifact and the maze can take a few minutes to complete worst case. Can still move it to the prison art if that's better)

code/mob/living/life/blood.dm
```diff
@@ -81,7 +81,7 @@
 
 		owner.blood_volume = max(0, owner.blood_volume) //clean up negative blood amounts here. Lazy fix, but easier than cleaning up every place that blood is removed
 		var/current_blood_amt = owner.blood_volume
-		if (owner.reagents?.total_volume > 0)
+		if (owner.reagents?.total_volume > 0 && src.owner.regens_blood)
```
Tarmunora, line 84: I assume this is so you can't buffer your way out of having 0 blood by loading up on cholesterol/procovertin, but they do both have their own built-in negatives to discourage loading up on massive amounts, so I'm not sure that blocking it here is needed

code/modules/status_system/statusEffects.dm
```diff
@@ -2975,3 +2975,276 @@
...
+					var/obj/decal/cleanable/blood_splat = make_cleanable(/obj/decal/cleanable/blood/splatter, get_turf(H))
+					blood_splat.streak_cleanable(pick(cardinal), full_streak = prob(25), dist_upper = rand(4, 6))
+				playsound(H, 'sound/impact_sounds/Slimy_Splat_2_Short.ogg', 40, TRUE)
+				H.implode(TRUE)
+				src.linked_curser.lift_curse_specific(FALSE, H)
+
+		onRemove()
+			var/mob/living/carbon/human/H = src.owner
+			if (!QDELETED(H))
+				H.regens_blood = TRUE
```
Tarmunora, line 3059: Should migrate to an atomprop - otherwise, if you get 2 blood curses (somehow), clearing one will reenable blood regen even though you're still under a curse

code/mob/living/critter/art_curser_nightmare.dm
```diff
@@ -0,0 +1,114 @@
...
+		if (M != src.cursed_human)
+			return
+		. = ..()
+
+	proc/register_target(mob/living/carbon/human/to_kill)
+		src.cursed_human = to_kill
+
+/datum/limb/small_critter/art_curser_nightmare
+	dam_low = 2
+	dam_high = 2
```
Tarmunora, line 113: Damage numbers here seem a little low (i.e.: fully blocked by wearing a sweater??). Maybe have their attacks do less stamina than default but more damage?

code/turf/floors_airless.dm
```diff
@@ -113,14 +113,23 @@
...
+			src.oxygen = MOLES_O2STANDARD
+			src.nitrogen = MOLES_N2STANDARD
+			src.temperature = T20C
 		..()
 		var/image/lava = image(icon = 'icons/turf/floors.dmi', icon_state = "lava", layer = src.layer - 0.1)
 		src.UpdateOverlays(lava, "lava")
 
+/turf/unsimulated/floor/airless/plating/catwalk/auto/iomoon/safe
+	is_safe = TRUE
+
```
Tarmunora, line 132: Why not just have the `.../safe` type override the vars on the type, rather than having them be set in `New()`? I don't see what the point of this pattern is.

Tarmunora (changes requested):

code/turf/floors_airless.dm
FlameArrow57: This was left in accidentally, I didn't mean to make a subtype - will be no changes to this file

code/mob/living/life/blood.dm
FlameArrow57: It was more aimed at just natural blood regen actually. I can still remove it if you think it'd be better

code/mob/living/critter/art_curser_nightmare.dm
FlameArrow57: Yeah I'm not sure what was going on, but actual damage when testing was 2x this amount, so always doing 4 dmg - armor - I can add a comment for this unless you know the reason why

Tarmunora: Hm. Looks like small critter limbs have `extra_damage` set to `rand(dam_low, dam_high) * quality`, as well as a baked in 0.1x stam damage mult. 🍝

code/mob/living/life/blood.dm
Tarmunora: Blood regen is not handled in that code block though. Natural regen is provided by the spleen (for humans)

PR #21270, opened 2024-11-09
Cash Briefcases

JORJ949:
## About the PR 
Adds cash briefcases which are briefcases that only hold spacecash but can be used to show off the large amounts of cash you are doing illegal deals with.
image

The standard one can be found empty in the HoP's locker
The syndicate one comes pre-loaded with 15k available for 2TC

## Why's this needed? 
Adds a nice RPable element to cash deals rather than just throwing your cash at someone, allows traitors to make illegal deals easier by trading TC for actual cash if they dont want to spend time making some.

## Changelog 

```changelog
(u)JORJ949
(*)Added cash briefcases! An empty one can be found in the HoPs locker and traitors can also purchase two of them for 2tc.
```


TobleroneSwordfish:
Oh these are cool, a couple of things though:
- Ideally we shouldn't have two different types of briefcase with the same sprite, would it be possible to wrap this functionality onto the regular briefcase instead?
- The cash purchase option should proobably have a buy limit or it's just going to be used to buy 90k worth of AEX every round.

JORJ949:
> 
> * Ideally we shouldn't have two different types of briefcase with the same sprite, would it be possible to wrap this functionality onto the regular briefcase instead?

We do already have this kinda with objBriefcase vs storage/briefcase, the former being the one used for foldable devices and the latter for putting stuff in, this isnt really either of those, with the exception of objBriefcase but id be making a whole new object for it anyway.

Though yeah same sprite does make it not the same I would rather the briefcase items look the same until you take a closer look because that's the cool factor of them, that its just a briefcase until you open it. Though a possibility is maybe adding a little icon on the front of a briefcase (possibly only on the inventory like stack numbers?) to indicate what it contains, e.g. record for the record player. You can also see the name of the object to see what it has inside if you dont want to deploy it.

Yeah limit is a good idea, I forgot we can do that now.

itchybones:
Does this work with prop cash as well? The stuff from the game vending machines.

JORJ949:
> Does this work with prop cash as well? The stuff from the game vending machines.

Currently only works with proper spacecash, though I will probably devise more features for it in the future, e.g. rigging it to shred the money when you try to take it out

code/obj/item/misc_junk.dm
```diff
@@ -721,3 +721,119 @@ TYPEINFO(/obj/item/reagent_containers/vape)
...
+		else
+			. = ..()
+
+	attack_hand(mob/user)
+		if (src.open)
+			var/amount = round(input("How much cash do you want to take from the briefcase?") as null|num)
+			if (isnum_safe(amount))
+				if (amount > src.balance || amount < 1)
+					boutput(user, SPAN_ALERT("You wish!"))
+					return
```
Sovexe, line 804: This should use `tgui_input_number` and dynamically assign a max input value based on the available stack for a better user experience

PR #21275, opened 2024-11-09
Portable Reagent Extractor TGUI

Garash2k:
## About the PR 
Makes it so that the Portable Reagent Extractor (prototype ChemiTool) uses the Reagent Extractor's UI and the CheMaster UI instead of some old html extractor UI and a bunch of alerts
image

## Why's this needed? 
All reagent extractors/CheMasters should have the same UI
Old UI:
261896918-32a052a3-3a20-4522-bd12-1c10f0373694

[ui][qol]

### DIFF
--- a/code/obj/submachine/chem_extractor.dm
+++ b/code/obj/submachine/chem_extractor.dm
@@ -22,8 +22,9 @@ TYPEINFO(/obj/submachine/chem_extractor)
 	var/obj/item/reagent_containers/glass/storage_tank_2 = null
 	var/list/ingredients = list()
 	var/list/allowed = list(/obj/item/reagent_containers/food/fish/, /obj/item/reagent_containers/food/snacks/,/obj/item/plant/,/obj/item/clothing/head/flower/,/obj/item/seashell)
+	var/obj/item/robot_chemaster/prototype2/parent_item = null
 
-	New()
+	New(var/obj/item/robot_chemaster/prototype2/parent_item = null)
 		..()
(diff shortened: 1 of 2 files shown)


code/obj/submachine/chem_extractor.dm
```diff
@@ -22,8 +22,9 @@ TYPEINFO(/obj/submachine/chem_extractor)
 	var/obj/item/reagent_containers/glass/storage_tank_2 = null
 	var/list/ingredients = list()
 	var/list/allowed = list(/obj/item/reagent_containers/food/fish/, /obj/item/reagent_containers/food/snacks/,/obj/item/plant/,/obj/item/clothing/head/flower/,/obj/item/seashell)
+	var/obj/item/robot_chemaster/prototype/parent_item = null
 
-	New()
+	New(var/obj/item/robot_chemaster/prototype/parent_item = null)
```
mordent-goonstation, line 27: Isn't the first argument of `New` the location of where it's being instantiated? Does this still work if instantiating on a turf?

Garash2k: Turns out setting location to null breaks the child machines and results in "[src] seems to be out of order." messages. Setting loc to the parent item like before seems to work well, and looking at `/obj/machinery/proc/powered(...)` makes me think obj locs are fine

Valtsu0:
It would probably be better if these were in the same window

Garash2k:
> It would probably be better if these were in the same window

I sure would, but that's a lot more effort for an easter-egg-ish item, effort which I'd rather spend on other stuff!

PR #21276, opened 2024-11-09
Sawfly QoL batch

NightmareChamillian:
## About the PR 

I'm going to be honest, this is entirely an intermediary "stepping stone" PR that primes Sawflies for a more RP-friendly mode of AI. Since the fate of that is uncertain, and there were several minute changes I wanted to make regardless, I thought I'd atomize and draft these separately.

## Okay, but what's it actually do?

Other than code cleanliness, this makes three alterations:
1) If you're now friendly to a sawfly, you can walk through them similar to small animal critters. This is the only significant change and just makes them less annoying for nukies.
2) I removed the overtuned and convoluted communal_beep() proc. It was originally added after someone voiced concerns with sawfly swarms being too noisy. Having had them in the game for at least two years now, I really don't think it's an issue. Sawflies are _extremely_ quiet as they stand, and if anything I feel like large groups of them should provide some kind of warning.
3) Sawflies now have a weak reference to the first friendly they encounter, which is tracked as their master. Other than people figure out which ones belong to who there's multiple users, this change has no effects.

## Changelog 

```changelog
(u)NightmarechaMillian
(+)You can now pass through sawflies if they're on your side.
```

### DIFF
--- a/code/mob/living/critter/sawfly.dm
+++ b/code/mob/living/critter/sawfly.dm
@@ -2,7 +2,7 @@
 ->The sawfly, by NightmareChamillian
 This file is the critter itself, and all the custom procs it needs in order to function.
 
--For the AI, check the critter/AI folder, it should be in sawflyai.dm
+-For the AI, sawflies use a mostly generic critter AI with some localized baked-in changes
 -For the grenade and controller, check code/obj/sawflymisc.dm
 */
 /mob/living/critter/robotic/sawfly
@@ -15,6 +15,7 @@ This file is the critter itself, and all the custom procs it needs in order to f
 
 	var/sawflynames = list("A", "B", "C", "D", "E", "F", "V", "W", "X", "Y", "Z", "Alpha", "Beta", "Gamma", "Lambda", "Delta")
 	var/static/list/priority_target_jobs = list("Head of Security", "Security Officer", "Nanotrasen Security Consultant")
+	var/mob/living/master = null //first friendly they imprint upon /datum/aiTask/timed/targeted/follower
 	var/obj/item/old_grenade/sawfly/ourgrenade = null
 
 	speechverb_say = "whirrs"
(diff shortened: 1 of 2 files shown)


code/mob/living/critter/sawfly.dm
```diff
@@ -15,6 +15,7 @@ This file is the critter itself, and all the custom procs it needs in order to f
 
 	var/sawflynames = list("A", "B", "C", "D", "E", "F", "V", "W", "X", "Y", "Z", "Alpha", "Beta", "Gamma", "Lambda", "Delta")
 	var/static/list/priority_target_jobs = list("Head of Security", "Security Officer", "Nanotrasen Security Consultant")
+	var/mob/living/master = null //first friendly they imprint upon /datum/aiTask/timed/targeted/follower
```
amylizzle, line 18:
```suggestion
	var/datum/weakref/master = null //first friendly they imprint upon /datum/aiTask/timed/targeted/follower
```

amylizzle (changes requested):

amylizzle (approved):

JORJ949:
You shouldn't be able to walk through them on non-help intent to avoid being outed as syndie by someone pushing you into one.

code/mob/living/critter/sawfly.dm
```diff
@@ -25,6 +26,8 @@ This file is the critter itself, and all the custom procs it needs in order to f
 	var/retaliate = FALSE
 	misstep_chance = 40 //makes them behave more like drones, and harder to kite into a straightaway then shoot
 
+	HELP_MESSAGE_OVERRIDE({"Can be stood on the same tile as if your intent is set to <span class='help'>help</span>. Syndicate only: use the remote or click on the sawfly with <span class='help'>help</span> or <span class='grab'>grab</span> intent to deactivate it."})
```
JORJ949, line 29: The "Syndicate only" bit of the help message is a bit misleading talking about the remote (iirc anyone can use the remote), and makes it seem as if anyone can stand on the same tile as a sawfly.

code/mob/living/critter/sawfly.dm
```diff
@@ -115,9 +104,20 @@ This file is the critter itself, and all the custom procs it needs in order to f
 		else
 			src.foldself()
 
-	Cross(atom/movable/mover) //code that ensures projectiles hit them when they're alive, but won't when they're dead
-		if(istype(mover, /obj/projectile))
+	Cross(atom/movable/mover)
+
+		//since sawflies have a density of 1 and will auto-set intent to harm, we have to recode tile-sharing
+		if(istype(mover, /mob/living))
```
JORJ949, line 110: not all mob/livings are tangible, for example this will stop AI eyes moving through them

NightmareChamillian: should be fixed now, added a parent call to let other functions decide

JORJ949 (approved):

PR #21295, opened 2024-11-11
Rework Mushroom medbay

JORJ949:
## About the PR 
Rework's mushroom medbay to be more spacious
+ Move MDir office
+ Make the OR have much more space to move around in so you arent being shoved around in surgery
+ Add Pharmacy
+ Expand Morgue
image

Other than the deleted duplicate firelock im not sure what change I made around the cryotron, I _think_ there may have been a duplicate window there but im not sure, IIRC I deleted a window because it would cry about there being 2 windows on the same tile.

## Why's this needed? 
The current surgery area is cramped, theres no pharmacy and the space is used incredibly inefficiently at the moment.

Redoing because mushroom has seen a small resurgeance and a couple people have asked about mushroom updates


cheekybrdy:
Think this is missing the department announcement PC that was in the old staff room

PR #21283, opened 2024-11-10
Fix for photocopying jailbird posters

LorrMaster:
[BUG] [MINOR] [TRAITS] [GAMEOBJECTS]

## About the PR 
- Fix for wanted posters generated by the jailbird trait not being able to be photocopied.

## Why's this needed? 
- Right now only wanted posters generated by wanted poster stations can be photocopied.

PR #21299, opened 2024-11-11
Panic photocopier revert

TobleroneSwordfish:
## About the PR 
Reverts #21283 and #21165

## Why's this needed? 
#21165 appears to have caused a cascade of runtime errors that is bricking PDA messengers and QM consoles. I'm not sure exactly why or how this is, but the errors start on the next round built after the PR was merged.


Laboredih123:
hehe

PR #21303, opened 2024-11-12
Donut 1 waoh

Laboredih123:
[MAPPING][FEATURE]
## About the PR 
Adds donut 1 but very mildly less shitty from its old version.
Things changed include: colorful wow, escape pods, cryotron, usable bridge, working merchant and mining shuttles, chapel ejector, and swap-ins for features that no longer exist. It is secretly votable through a hopefully sufficiently difficult threshold.
It does not include: very modern features, quality of life.

**This PR is _highly_ unlikely to be merged ever. As such, it serves more as an archival or perhaps even adoptable pr in the case anyone ever wishes to try their hand at it.**
If some admin takes a liking to this pr and has some change they would like to see, perhaps some thing offstation or minor departmental changes, I can do that. I can not, however, implement large-scale changes such as massive departmental rearranges. I do not have the time unfortunately.

## Images to observe at.
Olden
image

Newen
image

## Why's this needed? 
Uh, funi.


Glamurio:
Probably cool to add to the code base for posterity, even if we're (hopefully) never seeing it in action outside of meme rounds.

ZeWaka:
> Probably cool to add to the code base for posterity, even if we're (hopefully) never seeing it in action outside of meme rounds.

Nah

PR #21315, opened 2024-11-13
PDA Announcement Request App

JORJ949:
## About the PR 
Adds an "Announcer" app to all PDAs, allowing anyone to request an announcement and anyone with ID computer access on their inserted ID can approve announcements. You must have your ID inserted to use the app.
image

## Why's this needed? 
Doing gimmicks is much easier when everyone knows what you are doing. Catering recently got an announcement computer and using that creates a noticeable difference in visitors to the cafeteria, making announcements more accessible to other gimmicks will encourage further interaction with said gimmicks. 

The current solution is to just ask "hey AI can you announce X for me" or asking the HoP/Cap for bridge access to make your announcement from there, or some other departmental announcer access and building your own announcement computer, this heavily reduces the time needed to announce your gimmicks to the crew.

If it gets overused then it could probably be moved to a unique PDA cartridge that no PDAs start with.

## Changelog 

```changelog
(u)JORJ949
(*)Added "Announcer" PDA app to all PDAs, requested announcements can be approved by either the Captain, HoP, HoS or AI!
```

### DIFF
--- a/_std/defines/announcement_origins.dm
+++ b/_std/defines/announcement_origins.dm
@@ -7,3 +7,6 @@
 #endif
 #define ALERT_GENERAL "Frontier Authority Update"
 #define ALERT_STATION "Automated Mainframe Alert"
+
+
+#define JOBS_CAN_APPROVE_ANNOUNCEMENTS list("Captain","Head of Security","Head of Personnel", "AI")
--- a/_std/defines/announcement_origins.dm
+++ b/_std/defines/announcement_origins.dm
@@ -7,6 +7,3 @@
 #endif
 #define ALERT_GENERAL "Frontier Authority Update"
 #define ALERT_STATION "Automated Mainframe Alert"
-
-
-#define JOBS_CAN_APPROVE_ANNOUNCEMENTS list("Captain","Head of Security","Head of Personnel", "AI")
--- a/code/datums/datacore.dm
+++ b/code/datums/datacore.dm
@@ -499,7 +499,6 @@
 
 	var/approved = FALSE
 	var/approver = null
-	var/approver_job = null
 	var/approver_byond_key = null
 
 	var/sound_to_play = 'sound/misc/bingbong.ogg'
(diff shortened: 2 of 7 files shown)


Valtsu0:
btw you can put the captains name on an agent card to approve your own requests (same as ticket master)

Valtsu0:
(this probably defeats the entire point of the pr so take it with a grain of salt)
I'd rather this not be a remote announcement computer. 
Maybe have it send a pda alert and have the approving be done somewhere like commaster (that way it also could be tied to access rather than job name on manifest)

JORJ949:
> note that you can put the captain's name on an agent card to approve your own requests (same as ticket master). Unsure if this is a good thing or not

Both this and tickets should probably use inserted ID accesses, if anything this encourages putting your ID in it (possibly with help message in app if no ID is inserted)

JORJ949:
> (this probably defeats the entire point of the pr so take it with a grain of salt) I'd rather this not be a remote announcement computer. Maybe have it send a pda alert and have the approving be done somewhere like commaster (that way it also could be tied to access rather than job name on manifest)

If it does get used like a remote announcement computer I would agree with nerfing it like that yes, especially being able to(and automatically) authorizing your own announcements, but even then that may not be a problem.

TobleroneSwordfish:
Kind of agree with Valtsu that there shouldn't be a way to make mobile announcements, having to physically go to the computer is an important step that gives information on where that person is right now which can be very useful (ie imagine seeing the HoS announcing something from the bridge and knowing it's safer to go break into security right now)

JORJ949:
I like the idea of making it a comms console thing (also means I dont have to touch TGUI if I wanted to put it in the announcement console), that way any command member walking by the bridge (or the AI) ((or a borg)) could approve it

ZeWaka:
> Kind of agree with Valtsu that there shouldn't be a way to make mobile announcements, having to physically go to the computer is an important step that gives information on where that person is right now which can be very useful

this

JORJ949:
Moved the approval process to comms consoles, anyone can check who approved an announcement in the PDA app still and could check who approved it to know who is currently in bridge (or armory)

PR #21317, opened 2024-11-13
Should fix "bad index" error

Dredsen:
## About the PR 

Should fix the #1 error by count seen on https://goonhub.com/events/errors


Laboredih123:
the comedy

flappybatpal:
Seems like this is just covering up _SendSignal being sent bad arguments?

PR #21327, opened 2024-11-14
TOOL_CHOPPING now also does quad damage to windows

NightmareChamillian:
[wiki]
## About the PR 

Title- weapons with the chopping flag (breaching hammers, syndie saws, fireaxes) deal quad damage to windows.

## Why's this needed? 

Makes breaching equipment more useful- airlock hammers and the like have their niches somewhat usurped by the fact that breaking down a window and grille will always be quicker than going through a door, even with the quad damage that TOOL_CHOPPING provides. Hopefully giving more window damage will make them useful in cases outside of "man I really want to get into this room but alas there's no windows that I can break down with a fire extinguisher"

## Changelog 

unsure if should be small change but I put down big
```changelog
(u)NightmarechaMillian
(*)Breaching weapons (sledgehammers, fireaxes, etc) will now also deal 4x damage to windows
```

### DIFF
--- a/code/obj/window.dm
+++ b/code/obj/window.dm
@@ -473,7 +473,10 @@ ADMIN_INTERACT_PROCS(/obj/window, proc/smash)
 		else
 			attack_particle(user,src)
 			playsound(src.loc, src.hitsound , 75, 1)
-			src.damage_blunt(W.force)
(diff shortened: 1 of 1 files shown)


Azrun:
Don't forget to label your PRs.

PR #21349, opened 2024-11-17
Crime and Capitalism

Laboredih123:
[CLOTHING]
## About the PR 
Adds an *even bigger* syndicate hat that makes you *very valid* and a hat resembling that of certain business man.

image
image

## Why's this needed? 
Its very very funny. Also kyle liked it so why not.


frawhst:
DOUG SYNDSDALE? OWNER OF THE SYNDSDALE SYNDIDOME?

PR #21350, opened 2024-11-17
tgui-styles Prework

### LINKED ISSUE #20978: TGUI light mode text prompts are unreadable
### Labels

[BUG]

### Description

The Tgui in game text prompts in light mode are black on black, leading you to eater highlight to see what you are typing or just blindly guess and hope what you put in makes sense

### Steps to reproduce

1. attempt to change your name, send a announcement/PDA message, or write something on the floor with a pen
2. hope what you put is legible (see image)

### Expected Behavior

I expect the text boxes to be more like the other prompts and have black text on a white background

the set first name prompt (10k spessbucks for the first person to say the name)
image

mordent-goonstation:
[UI][Internal][Code Quality]

## About the PR 

Bumps a bunch of our internal dependencies for `tgui` up a few versions, as prerequisite work for using `tgui-styles` library.

Bumps `tgui-core` to 1.3.0, which is a headless version, meaning our component style overrides (that we never removed, thanks lazy past us!) now apply again.

Updated our in-repo component styles to be in-line with `tgui-styles` 0.0.8, with differences called out via `|GOONSTATION-ADD|` or `|GOONSTATION-CHANGE|` comments still.

This *should* give us our light-mode theming back as well, as those light-mode themes applied to the same set of classes that we're now revealing again via `tgui-core` 1.3.0.

Fixes #20978.

## Why's this needed? 

Gets us onto `tgui-core` 1.3.0, allows for the migration to `tgui-styles` to be a smaller piece of work as it is now replacing "like" for like.

Giving us light mode back is *supposedly* a good thing. It makes finding out weirdos easier.


ZeWaka (approved):
Reviewed, looks correct

PR #21351, opened 2024-11-17
removes prototype engine areas from minimap + adds fixes notice board runtime on mushroom & second rancher slot

cheekybrdy:
## About the PR 
Moves disposals so it disposes all contents to space + prototype engine area is moved off minimap and has a reactor stat PC. chapel launcher now fully requires access.

## Why's this needed? 
Fixes and tweaks good.


code/area.dm
```diff
@@ -2855,6 +2856,7 @@ TYPEINFO(/area/station/engine/substation)
 	luminosity = 1
 	force_fullbright = 1
 	requires_power = 0
+	minimaps_to_render_on = 0
```
ZeWaka, line 2859: does this affect other maps tho

cheekybrdy: The area in question is for a station level space structure, no rotation maps currently have this area and it'd make sense for it not to be on there anyway. This would effect older maps like trunkmap but they don't compile anyway.

PR #21353, opened 2024-11-17
Clicking on an airlock with an item and harm intent will always attack the airlock

NightmareChamillian:
## About the PR 

See title- adds a check before all the tool checks

## Why's this needed? 

There's a lot of weird mutual exclusivity with airlock-item interactions. By adding an intent check, you can, if you really want to, damage airlocks with crowbars even though they're prying tools.

## Changelog 

```changelog
(u)NightmarechaMillian
(+)Clicking an airlock with harm intent will now always attack it
```


NightmareChamillian:
i had a better idea

PR #21354, opened 2024-11-17
Adds a new melee weapon for the Salvager antagonist- the Crash Axe

NightmareChamillian:
[feature][gamemode][input]
## About the PR 

The crash axe is a 1-handed small weapon that deals 14 brute damage.
image

Furthermore, it can be toggled between functioning as a crowbar and a chopping tool for utility in hacking through airlocks. Whether by force or electronically.

Lastly, the weapon deals a 3-second stun on throw. The intent behind this isn't make a good rampaging weapon, but allow for opportunistic snatching of peoples' weapons. The short duration and thrown prerequisite makes this hard to pull off, though; since the crash axe and whatever held weapons are dropped at the RECIPIENT'S feet. 

To do so need to: 1, hit the throw (or you lose your weapon), then 2, close the distance (they're already in range), then 3, pick up your axe from under them (or that's theirs now), then 4, swap hands and pick up whatever THEY were holding (which at the top of their priority list as well). 

## Why's this needed? 

Salvagers only have one melee option, and it's a 2-hander, offering a completely missing niche.

While they _can_ make scrap weapons, use decon devices, and fire extinguishers, that does not an antagonist make. Adding more unique items helps flesh them out more from being spacefaring versions of sleeper agents. 

I feel like the crash axe is thematic enough to fit and offers a chance to bolster the loadout through sheer utility. 

Its use as a crowbar is applicable to salvagers specifically, since you can pry an airlock and then get straight into a fight without having to swap items. Doors aren't a big deal for salvagers due to their decon gun, though it offers brute force  alternatives if theirs is lost, out of charge, or untimely-- additionally, conjunction with  #21327 allows for expedient entry through windows. Lastly, the microstun mechanic is thematic and rewards the user for capitalizing properly on high skill maneuvers.

## Changelog 

```changelog
(u)NightmarechaMillian, sprite help by DisturbHerb
(*)The M4GP13 now stocks a new tool/weapon: the crash axe! See minor changes for details
(+) Crash axes are pocket sized and act as chopping and prying tools, toggleable by using them inhand. 
(+) In combat, they deal 13 damage, 21 on thrown, and have a short, 2-second thrown stun.
```


NightmareChamillian:
also ideally I'd prefer if #21327 was merged as a prerequisite PR to this one

code/obj/item/misc_weapons.dm
```diff
@@ -2130,6 +2130,62 @@ obj/item/whetstone
...
+	hitsound ='sound/impact_sounds/coconut_break.ogg' //shamelessly re-using this sound idea from my halberd
+
+	health = 5
+	tool_flags = TOOL_CUTTING | TOOL_CHOPPING
+
+	w_class = W_CLASS_SMALL
+	hit_type = DAMAGE_CUT
+	leaves_slash_wound = TRUE //yeah prolly
+	force = 14
+	stamina_damage = 1 //we WANT this to be low so crewmembers can run away easily
```
TobleroneSwordfish, line 2148: You're getting hit by an axe, it should do a lot more than 1 stamina damage.

code/obj/item/misc_weapons.dm
```diff
@@ -2130,6 +2130,62 @@ obj/item/whetstone
...
+	throwforce = 24 //probably needs tuning but these aren't too far divorced from tomahawks- encourage high-skill gameplay
+	throw_speed = 4
+	throw_range = 7
+
+	var/breakmode = TRUE
+	var/breaking_tool_flags = TOOL_CUTTING | TOOL_CHOPPING
+	var/prying_tool_flags = TOOL_CUTTING | TOOL_PRYING
+
+	HELP_MESSAGE_OVERRIDE({"Can be used as a crowbar. Throw the weapon (space+click) to inflict a half-second stun and two second disorient."})
```
TobleroneSwordfish, line 2157:
```suggestion
	HELP_MESSAGE_OVERRIDE({"Can be used as a crowbar. Can be thrown to inflict a very short stun."})
```
Don't mention specific keybinds or durations that are either user specific or likely to be changed in future. Should also mention the flipping mechanic somehow.

code/obj/item/misc_weapons.dm
```diff
@@ -2130,6 +2130,62 @@ obj/item/whetstone
...
+		src.setItemSpecial(/datum/item_special/tile_fling) //borrowing attributes from crowbar
+		BLOCK_SETUP(BLOCK_ROD)
+
+	// I refuse to add to the item/knife/butcher inheritance tree. I just won't do it.
+	throw_impact(atom/A, datum/thrown_thing/thr)
+		if(iscarbon(A))
+			var/mob/living/carbon/C = A
+
+			C.changeStatus("knockdown", 0.5 SECONDS) //give what's basically a defib stun; lets salvagers yoink weapons if they're savvy
+			C.changeStatus("disorient", 2 SECONDS) //give them getaway time
```
TobleroneSwordfish, line 2170: Honestly for throwing a weapon you have to buy these numbers could definitely be longer, syndie daggers are 6 seconds although they don't have the additional utility or damage of the axe.

TobleroneSwordfish (changes requested):
Seems to be missing the salvager shop entry? It's in open code now so you should be able to add it.

Glamurio:
Personally would actually prefer the stun on throw to be longer rather than shorter, for the precise reason of a) making it a bit more worthwhile and b) making it more user-friendly for people who aren't that robust.

code/obj/item/misc_weapons.dm
```diff
@@ -2130,6 +2130,62 @@ obj/item/whetstone
...
+	desc = "A lightweight utility-survival axe designed to cut through the thin skin of pod hulls. Works just as well on bothersome crewmembers."
+	icon = 'icons/obj/items/weapons.dmi'
+	icon_state = "crashaxe"
+	inhand_image_icon = 'icons/mob/inhand/hand_weapons.dmi'
+	hitsound ='sound/impact_sounds/coconut_break.ogg' //shamelessly re-using this sound idea from my halberd
+
+	health = 5
+	tool_flags = TOOL_CUTTING | TOOL_CHOPPING
+
+	w_class = W_CLASS_SMALL
```
Azrun, line 2144: PR Description describes this as `pocket-sized weapon` this is smaller than pocket sized.

code/obj/item/misc_weapons.dm
```diff
@@ -2130,6 +2130,62 @@ obj/item/whetstone
...
+	hit_type = DAMAGE_CUT
+	leaves_slash_wound = TRUE //yeah prolly
```
Azrun, line 2146: Should this be on both sides?

code/obj/item/misc_weapons.dm
```diff
@@ -2130,6 +2130,62 @@ obj/item/whetstone
...
+
+	var/breakmode = TRUE
+	var/breaking_tool_flags = TOOL_CUTTING | TOOL_CHOPPING
+	var/prying_tool_flags = TOOL_CUTTING | TOOL_PRYING
+
+	HELP_MESSAGE_OVERRIDE({"Can be used as a crowbar. Throw the weapon (space+click) to inflict a half-second stun and two second disorient."})
+
+	New()
+		..()
+		src.setItemSpecial(/datum/item_special/tile_fling) //borrowing attributes from crowbar
```
Azrun, line 2161: Should this be limited to the pry side?

Azrun (commented):
Which items is this supposed to be in conversation with?

In thematic function this seems like the survival machete but the stats are **way** better, and has more utility.  Based on stats these seem like a the gang throwing knives but slightly reduced stats but **more** utility.

code/obj/item/misc_weapons.dm
NightmareChamillian: I thought about having it toggle between cut and stab, but that seemed a little convoluted

code/obj/item/misc_weapons.dm
NightmareChamillian: good point, though by "pocket sized" in the description I meant "can fit into pockets", oops. should be clearer now

code/obj/item/misc_weapons.dm
```diff
@@ -2130,6 +2130,64 @@ obj/item/whetstone
...
+	tool_flags = TOOL_CUTTING | TOOL_CHOPPING
+
+	w_class = W_CLASS_SMALL
+	hit_type = DAMAGE_CUT
+	leaves_slash_wound = TRUE //yeah prolly
+	force = 14
+	stamina_damage = 25
+	throwforce = 24
+	throw_speed = 4
+	throw_range = 7
```
Azrun, line 2151: https://github.com/goonstation/goonstation/pull/21354#pullrequestreview-2441260993

Fraction of the size of a katana and only deals one less damage. Now has it even has swipe.
Throwforce is one less than an axe, has greater range, greater speed and has a special knockdown/disorient.

Azrun (changes requested):

code/obj/item/misc_weapons.dm
NightmareChamillian: hm yea good point

Azrun: If those are the bar, then 200 points seems exceptionally cheap for an item of that caliber.  That is cheaper than the sledgehammer.

Azrun (approved):
Let's give it a shot, I'm expecting to need to tone it down but we'll let the data tell us the facts.

Azrun:
Will merge once changelog is corrected....

PR #21355, opened 2024-11-17
Search Box for QM Package list UIs

Garash2k:
## About the PR 
Adds a textbox which allows to filter QM packages on screens where they're listed
image
image
image

Video showing the placeholder/usage:

https://github.com/user-attachments/assets/396f680f-da04-4bba-ae8e-d0c7eafe30d1

## Why's this needed? 
I thought the change was small&simple enough. Ideally everything is redone in tgui, but this seems like a good stopgap.

[UI][QoL]


TobleroneSwordfish (approved):
Oh I love this, very neat solution.

PR #21364, opened 2024-11-18
Revamp botany-associated requisitions

### LINKED ISSUE #21602: Seed packets only count as 1 seed for cargo requesitions that ask for seeds
### Labels

[BUG][MAJOR]

### Description

Seed packets only count as 1 seed for cargo requesitions that ask for seeds

### Steps to reproduce

1. Get a cargo requesition that asks for seeds
2. do whatever you do to make a seed packet of the type requested
3. scan with cargo scanner, see only 1 seed counted instead of the number of seeds in the packet

### Expected Behavior

seed packets should count as multiple seeds for the purpose of cargo requesitons

Reported by: grasswitch
Client version: 515.1643
Active test merges: #21006 and #20620
Round log date: 2024-12-07-20-12
Reported on: 2024-12-07 21:03:21
Round ID: 194755
Map: CLARION

Egregorious:
## About the PR 

Converts the science-themed seed requisition into a gourmet-kitchen themed fruit and vegetable requisition. It still asks for certain stats, but instead of seed items it asks for fruit or veg items with those same stats. It is less likely to ask for more than 2 unique items than the seed req did, on account of requiring a bit more work on average. 

Adds a science-themed mutated-plant product req, which asks for a number of products obtainable from plant mutations. The list of products is mostly based on those I don't often see grown, or those mutations often associated with antag-gimmicks. This one is a little tiered, with easier-to-obtain products being more common, and plant mutations with prerequisite stats or infusions being rarer and valued more highly by the req reward. 

Internally this refactors the seed entry to be the child of a generic plant-genetics entry, and allows seed entries to accept seed packets as a stack. Also includes a refactor of descriptions to use overrides instead of defines, which should probably be its own PR but was done so that switching between seed and plant descriptions wasn't so jank and unscalable.

## Why's this needed? 

The seed req has been a little broken since the introduction of seed packets, since it only ever counts them as a single seed. I opted to essentially remove it instead of fix it because I think it's not a very good requisition anyway; it tends to encourage only infusing seeds instead of growing them, and since every seed is basically the same as every other seed it doesn't actually have much variety to it. 

Requiring produce instead of just seeds asks a bit more out of the gameplay, and is more likely to encourage players to interact with plant varieties they otherwise wouldn't.

Similarly, the mutation requisition asks for products from mutated plants to encourage mutating plants that don't often get mutated, or to provide a little bit of plausible deniability to antags growing tomato grenades.

fixes #21602

## Changelog 

```changelog
(u)Egregorious
(+)Revamped the botany-associated requisitions.
```


Kubius (changes requested):
Seems *largely* structurally correct and I like the shift to requiring finished produce instead of seeds to avoid simply splicing the seeds into condition, with an associated reduction in needing multiple varieties as frequently.

There are a few things that seem to have been missed:
* There was a /datum/rc_entry/plant/seed added, but the existing seed entry type doesn't seem to have been removed and other uses of the seed entry (Grass seed for construction projects as an example I can recall) don't seem to have been moved to utilize it.
* Request list formatting for RC_SEED has been altered to remove the word "seed" but it's still expecting to be passed /datum/rc_entry/seed, not the new /datum/rc_entry/plant. The entry type define could probably be shifted to RC_PLANT in the process of restructuring this, and the "seed" suffix could reappear if the entry is specifically the /plant/seed subtype.

Egregorious:
Thanks for the feedback! I forgot to delete the old seed entry, the lack of errors lulled me into a false sense of security. It should be fixed now, and all old references accounted for.

I applied the description change you suggested, and although I am now going into scope creep territory, I also refactored the way descriptions are handled to make use of overrides instead of defines, I think this way is more scalable.

code/modules/economy/requisition/requisition_contracts.dm
```diff
@@ -252,40 +287,78 @@ ABSTRACT_TYPE(/datum/rc_entry/seed)
...
+					if(genes.cropsize >= gene_reqs["Yield"]) gene_count++
 				if("Potency")
-					if(cultivar.plantgenes.potency >= gene_reqs["Potency"]) gene_count++
+					if(genes.potency >= gene_reqs["Potency"]) gene_count++
 				if("Endurance")
-					if(cultivar.plantgenes.endurance >= gene_reqs["Endurance"]) gene_count++
+					if(genes.endurance >= gene_reqs["Endurance"]) gene_count++
 
 		if(gene_count >= gene_factors) // Compare satisfied parameter count to number of parameters. Met or exceeded means seed satisfies requirements
 			src.rollcount++
```
Kubius, line 330: Does this account for seed packets? If not, I'd *suspect* it would be as simple as adding the packet's "amount" variable to rollcount (if applicable) instead of incrementing it by 1.

Kubius (commented):
Looks like a pretty solid restructure. If it tests out well ingame, Seems Good.

code/modules/economy/requisition/requisition_contracts.dm
Egregorious: More or less, seed packets aren't technically a stack of items for the same reason seeds haven't been stackable in the past, but it's just an extra cast. 

It should be accounting for seed packets though, finally.

Kubius (commented):
Scope's gotten a bit more extensive (PR re-title?) but seems good if everything tests out fine. If no one else can get around to giving it a supplementary test within a couple days, I'll probably look into doing one myself.

PR #21368, opened 2024-11-19
Adds mail rooms to all rotation maps + makes them the mail courier workplace

cheekybrdy:
## About the PR 
Adds mail sorting rooms to all rotation maps, adds mail access helpers, mail courier lockers. Removes cargo access and automatically reroutes mail to the room. Moves mail couriers back there as their workplace.

## Why's this needed? 
Mail Couriers, while in a rather social position mainly have a RP based gameplay loop for their reward. Giving them back their dedicated space allows for additions curated for the position which don't compete with cargo's typically exponentially decreasing space. Also gives the role a private area for crime and still has a need to interact with cargo during a blockage. I don't think this'll impede on the nomadic appeal to mail couriers either.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Added mail sorting rooms to all maps and remade them the mail courier's dedicated workplace. Mail rooms, and the lockers inside now use a dedicated access. All inbound mail is rerouted to these sorting rooms.
```


code/obj/storage/crates.dm
```diff
@@ -498,6 +498,7 @@ TYPEINFO(/obj/storage/crate/chest)
 	icon_state = "mailcrate"
 	icon_opened = "mailcrateopen"
 	icon_closed = "mailcrate"
+	destinations = list("Mail")
```
pgmzeta, line 501:
```suggestion
	delivery_destination = "Mail"
```

Glamurio:
I like the idea of the PR, but I'm not sure if I'm quite sold on some of the new room locations. I'd prefer Clarion not lose that random room to make room for this, and would rather see it placed elsewhere. Also feel like if the mail courier gets their own room now, that room should look just a tad bit nicer than your average maintenance hallway, some plating on the floors or something would go a long way.

cheekybrdy:
Routings been all tested and works.

code/obj/storage/secure_closets.dm
```diff
@@ -868,6 +879,9 @@ ADMIN_INTERACT_PROCS(/obj/storage/secure/closet, proc/break_open)
 /obj/storage/secure/closet/civilian/ranch
 	name = "\improper Rancher supplies locker"
 	req_access = list(access_ranch)
+	icon_state = "secure_green"
+	icon_closed = "secure_green"
+	icon_opened = "secure_green-open"
```
JORJ949, line 884: Not a part of this PR.

JORJ949 (changes requested):
Mapping changes are incredibly hard to review due to the size of some, IMO mail shouldnt go too far from cargo anyway, there isnt always a mail courier and QM should be able to do it in downtime.

Also be consistent with the QM mail chutes, they should always lead into the general mail loop (clarions one appears to just lead to the mail room)

code/datums/jobs.dm
```diff
@@ -972,7 +972,7 @@ ABSTRACT_TYPE(/datum/job/civilian)
 	slot_back = list(/obj/item/storage/backpack/satchel)
 	slot_ears = list(/obj/item/device/radio/headset/mail)
 	slot_poc1 = list(/obj/item/pinpointer/mail_recepient)
-	slot_belt = list(/obj/item/device/pda2/quartermaster)
+	slot_belt = list(/obj/item/device/pda2)
```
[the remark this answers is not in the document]
cheekybrdy: Added a version with civvie colours

code/obj/storage/secure_closets.dm
cheekybrdy: Done + Clarion's pipe goes to mail loop for QM, just crosses over the crate transfer pipe so at a glance it looks like it goes into there

Tarmunora:
Honestly, mail operating out of cargo is a feature, not a bug:
1. Quartermasters can handle mail if they want, esp for lowpop where there's no mailman or if QMs are just bored and waiting for market updates
2. Mail couriers regularly interact with QMs

After some discussion with other admins in discord, closing this PR

PR #21374, opened 2024-11-19
Cigarette packet quality of life improvements

Egregorious:
## About the PR 

Cigarette packets can be quick-filled via dragging cigs onto them, and dumped out by dragging them onto floors or tables.

## Why's this needed? 

This mostly just helps distribute custom cigarettes, it's currently quite cumbersome to deal with them if you make a bunch at once.

## Changelog 

```changelog
(u)Egregorious
(+)Enabled quick-filling and dumping of cigarette packets. Inspect them for more detailed information. 
```

### DIFF
--- a/code/obj/item/cigarette.dm
+++ b/code/obj/item/cigarette.dm
@@ -550,6 +550,47 @@
 		for(var/i in 1 to src.max_cigs)
 			new src.cigtype(src)
 
+	MouseDrop_T(atom/movable/O as obj, mob/user as mob)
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (changes requested):
Consider using `/datum/component/transfer_input/quickloading` or similar instead of copy pasting the behaviour here.

Egregorious:
Ah, I did not know that existed! Not a fan of how the component works personally, but I suppose that's something for a different PR.

PR #21410, opened 2024-11-23
Performance increase for /obj/item/proc/update_inhand

FlameArrow57:
[GAME OBJECTS][PERFORMANCE]

## About the PR 
Performance enhancement, if an icon state is not found in .dmis, it should be debugged through inspection, not fallback

Before:
x

After:
x2


ZeWaka (commented):
tempting to log this for a bit and then merge this so we have a list

TobleroneSwordfish:
fucking `icon_states` omg

ZeWaka:
fuck yeah uhh need to log this

ZeWaka:
e526b9fe5a884ac18fdd4fd4888736e5c4b6a83e

ZeWaka:
X, just needs to log and backlog a list

FlameArrow57:
Okay ready for re-review

ZeWaka:
It really just needs to log to a file that we then later retrieve off of the server, and dedupe

ZeWaka (approved):
good to TM

flappybatpal:
I think this PR is causing certain two handed inhands (taser shotgun, newspaper) to not show as they work correctly on local but not live.

ZeWaka:
there is 7,487 unique instances of errors @FlameArrow57 
[missing_inhand_item_states.txt](https://github.com/user-attachments/files/20750127/missing_inhand_item_states.txt)

PR #21406, opened 2024-11-23
Use tgui-styles

mordent-goonstation:
[UI][INTERNAL][CODE QUALITY]

## About the PR 

Uses https://github.com/tgstation/tgui-styles, removing our own versions of those styles.

There are a couple of Goonstation changes I need to either get merged into tgui-styles itself or put back in some capacity before this can be merged.

## Why's this needed? 

Keeps us more in-line with upstream tgui, helping us stay up-to-date.

## Port Upstream/Extract

- [ ] `.fas`/`.far` <-> `.fa-solid`/`.fa-regular`
- [ ] `.Modal--fitted`
- [x] `.Button--ellipsis .Button__content`
- [ ] `.Dimmer--full`

PR #21411, opened 2024-11-23
TGUI Ruck Kit

Garash2k:
## About the PR 
Replaces the old ruck kit's html ui with a tgui one
image

With locking perms:
image

Light theme:
image

## Why's this needed? 
A bit more responsive and more modern looking, matches the fabricator UI, has icons!
Old UI:
261901990-01ae96bc-a294-4098-a3d3-847a77d2a4f8

Note: oldie mode kinda works? It can print electronics frames just fine, and it ignores locked blueprints the same as before.

## Changelog 

```changelog
(u)Garash2k
(+)New UI for the Ruck Kit
```

[UI][Feature]


mordent-goonstation (changes requested):
Hold pending us working out performance for Manufacturer, as will have some similar things going for it.

tgui/packages/tgui/styles/components/Button.scss
```diff
@@ -186,3 +186,15 @@ $bg-map: colors.$bg-map !default;
...
+.Button--ComplexContent {
+  // Button__content has an `align-self: stretch` that only works when it's parent is display flex, this makes it work.
+  display: flex;
+  // Opting out of `whitespace: nowrap` for complex button contents
+  white-space: normal;
+  .Button__content {
+    //In addion to stretch, grow: 1 to take the full width is pretty much necessary
+    flex-grow: 1;
+  }
+}
```
mordent-goonstation, line 200: Hold pending #21406

tgui/packages/tgui/interfaces/RKit/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+import { useBackend } from '../../backend';
+import { Window } from '../../layouts';
+import {
+  BlueprintButtonStyle,
+  BlueprintMiniButtonStyle,
+} from '../Manufacturer/constant';
+import { RKitData, ScannedItemData } from './type';
+
+export const RKit = () => {
```
mordent-goonstation, line 18: Use the full name, tbh, or at least "RuckKit"

code/obj/item/tool/electronics.dm
```diff
@@ -796,97 +796,100 @@
...
+			blueprint_available = !!item.blueprint,
+			locked = item.locked,
+			imagePath = getItemIcon(item.item_type, C = user.client),
+			ref = ref(item),
+		)))
+
+	. = list(
+		hide_allowed = src.allowed(user),
+		scanned_items = scanned_items,
+		olde = src.olde
```
mordent-goonstation, line 831: I have absolutely no idea what `olde` means.

Garash2k: So, `/obj/machinery/rkit` have a var/olde mode, which, if enabled, makes them print electronic frames instead of blueprints. There is 1 rkit in z2.dmm with olde enabled. I made olde mode work without too much effort, but I'm also down to tear down the whole concept if I get a dev go ahead.

Tarmunora: as the person who did the `olde` thing, go ahead and tear it out if it makes things cleaner.
It doesn't even work too well these days anyways, since it's incompatible with anything that has a custom material list

Garash2k: Maybe we can just rename the tgui property to legacyMode and we're good?

mordent-goonstation:
Broadly, if we're wanting to have this kind of compound button be a re-usable pattern, it should be put in the `components` directory. Definitely wait until after #21406 for that, else that's just going to complicate things.

code/obj/item/tool/electronics.dm
mordent-goonstation: Rename the UI prop to that, then see my other comment.

tgui/packages/tgui/interfaces/RuckingenurKit/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+      <Window.Content>
+        <Section title="Scanned Items" scrollable fill>
+          {scanned_items.map((scanned_item) => (
+            <ScannedItem
+              ScannedItem={scanned_item}
+              key={scanned_item.ref}
+              hide_allowed={hide_allowed}
+              olde={olde}
+            />
+          ))}
```
mordent-goonstation, line 35: Branch on (currently) `olde` here to render a different component instead if `olde` is truthy. That component is likely to be composed of a similar structure to `ScannedItem`, but you can then strip out any `olde`-specific stuff from `ScannedItem`.

mordent-goonstation (changes requested):

vortex1942:
I think this might benefit from having a filter option for blueprints.

As it might get very cluttered when a majority of the station is scanned.

Garash2k:
Good idea, I'll look into it after once it's a performant component button thingie

Tarmunora:
from mordent on discord:
> Extract out that weird compound button thingy into a component, use that in the manufacturer and the Ruck Kit, bam.

Garash2k:
Oh yeah, good timing, snow storm cancelled my plans today.

Garash2k:
Cheeky search added in the buttons of the section, I think it looks nice there. When you think about it, a search input that acts instantly is kind of like a button.

image

PR #21412, opened 2024-11-23
Replaces the cogmap1 upload door with a see trough one

Valtsu0:
[SILICONS]
## About the PR 
You can't see inside without bridge access. Camera also moved so it can't see laws
image

## Why's this needed? 
I've had a lot of fun interactions in the old thindow observational area and i miss it

## Changelog 

```changelog
(u)Valtsu0
(+)Cogmap1 AI upload door has been replaced with a see-through one
```

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -55125,12 +55125,10 @@
 /obj/cable{
 	icon_state = "1-2"
 	},
-/obj/machinery/door/airlock/pyro/command{
-	name = "AI Upload"
-	},
 /obj/mapping_helper/access/ai_upload,
 /obj/mapping_helper/firedoor_spawn,
 /obj/decal/stripe_delivery,
+/obj/machinery/door/airlock/pyro/glass/command,
 /turf/simulated/floor,
 /area/station/turret_protected/ai_upload)
 "qqS" = (
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -21956,6 +21956,12 @@
 	pixel_y = 32
 	},
 /obj/decal/stripe_caution,
+/obj/machinery/camera{
+	c_tag = "autotag";
+	dir = 10;
+	name = "autoname - SS13";
+	tag = ""
+	},
 /turf/simulated/floor,
 /area/station/turret_protected/ai_upload_foyer)
 "bzz" = (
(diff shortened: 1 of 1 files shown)


JORJ949:
I think if you want upload to be visible than the whole wall should be windows.
Additional point that unlike other maps, foyer is reachable by just having bridge access letting anyone with bridge access (NTSC, Mail Courier, etc) and not upload access see inside or even just break the windows to get in, so plasmaglass windows may be best here.

maps/cogmap.dmm
```diff
@@ -55125,12 +55125,10 @@
 /obj/cable{
 	icon_state = "1-2"
 	},
-/obj/machinery/door/airlock/pyro/command{
-	name = "AI Upload"
-	},
 /obj/mapping_helper/access/ai_upload,
 /obj/mapping_helper/firedoor_spawn,
 /obj/decal/stripe_delivery,
+/obj/machinery/door/airlock/pyro/glass/command,
```
pgmzeta, line 55131: name removed?

JORJ949: name is automatically applied by area, so its only needed if you want to name a door something that isnt the area name

Valtsu0:
> I think if you want upload to be visible than the whole wall should be windows. Additional point that unlike other maps, foyer is reachable by just having bridge access letting anyone with bridge access (NTSC, Mail Courier, etc) and not upload access see inside or even just break the windows to get in, so plasmaglass windows may be best here.

I was gonna do a window first, but due to plasmaglass windows being equally vulnerable to screwdriver and crowbar, it made breaking into upload too easy

JORJ949:
Have you checked the foyer camera cant see the laws? Upload/Core cams can only be used by AIs, foyer can be used by anyone

Valtsu0:
Didn't think about that, i'll see if i can move the camera a bit

Valtsu0:
The only place in the foyer that can't see the law is the tile i'm standing in the image. Moving it there would require shuffling things quite a lot as very few things there look good north of a wall. Would it be better to restrict the camera?

Valtsu0:
Updated the image to reflect the moved camera

JORJ949:
> 
> Originally i was gonna do a window, but due to plasmaglass windows being equally vulnerable to screwdriver and crowbar, it made breaking into upload too easy

If you have a screwdriver, crowbar and bridge access its not _that_ much harder to break down the door in a similar timeframe it would take to open a plasmaglass window, the only jobs that come to mind with bridge and not upload access are: NTSC (their gun is very effective at door removal), Mail Courier (can just mail themselves into engi), and gimmick jobs (which are rare)

JORJ949:
Also if #21433 gets merged you dont have to worry about camera placement

ZeWaka:
cogwerks has approved this

PR #21426, opened 2024-11-24
Add hotkeys to store to/unstore from belt and backpack

FlameArrow57:
[PLAYER ACTIONS][FEATURE][QoL]

## About the PR 
Adds hotkeys that let you automatically store items specifically to the storage in your belt or backpack slot. Default key is "H" for belt and "N" for backpack
-Pressing the hotkey will automatically store anything that you are holding to the storage item in that slot
-You don't need the storage UI open for it to work
-It keeps track of the last item you stored, so pressing the hotkey again will automatically take what you last stored out of the storage, without you needing to open the UI

## Why's this needed? 
-QoL/faster player inventory interaction

## Changelog 

```changelog
(u)FlameArrow57
(*)Added two new hotkeys for auto-storing/unstoring items to/from your belt storage (default "H") and backpack storage (default "N")
(+)Hotkeys do not need the storage UI to be open, and have a 1 second cooldown on alternating store/unstore actions
```

### DIFF
--- a/code/datums/hud/human.dm
+++ b/code/datums/hud/human.dm
@@ -317,26 +317,22 @@
 						master.autoequip_slot(I, SLOT_GLASSES) || \
 						master.autoequip_slot(I, SLOT_EARS) || \
 						master.autoequip_slot(I, SLOT_WEAR_MASK) || \
-						master.autoequip_slot(I, SLOT_HEAD))
-						return
-
-					if (master.autoequip_slot(I, SLOT_BACK))
-						master.last_stored_backpack = I
+						master.autoequip_slot(I, SLOT_HEAD) || \
+						master.autoequip_slot(I, SLOT_BACK))
 						return
 
 					if (!master.belt?.storage || I.storage) // belt BEFORE trying storages, and only swap if its not a storage swap
 						master.autoequip_slot(I, SLOT_BELT)
 						if (master.equipped() != I)
-							master.last_stored_belt = I
+
 							return
 
 					for (var/datum/hud/storage/S in user.huds) //ez storage stowing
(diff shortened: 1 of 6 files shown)


TobleroneSwordfish (changes requested):
Not sure about the rest of the PR but this in particular is an "oh god no":
> It keeps track of the last item you stored, so pressing the hotkey again will automatically take what you last stored out of the storage, without you needing to open the UI

This would let you take out a stun baton/gun/other weapon for a single game tick, hit someone with it and then stow it again without giving them any chance to respond or disarm you. You could bind it to a mouse key, hover your cursor over someone and press `side mouse -> left click -> side mouse` for a free attack.

In general I'm not too sure about adding *even more* hotkeys, especially quick swap ones that more or less require you to rebind them to be useful. Seems like it's adding to the pile of things you need to learn/set up for combat.

FlameArrow57:
Okay, intent was mainly for QoL not exclusive to combat - what about adding a click delay to the hotkey? not sure if that'be counter intuitive or not

Glamurio:
The only way I could see this work out is if it started a relatively quick action (ca. 1 second or so), that way you're about as fast, albeit a little slower than doing it manually, but it is more convenient. Just be sure to make the action private and uninterruptible.

ZeWaka:
delay would make this a lot more feasible

FlameArrow57:
Okay I added a cooldown on taking an item out, which triggers when storing the item

FlameArrow57:
Okay it now has cooldowns on the hotkey for alternating storing/unstoring, but you can still quick store a bunch of items

FlameArrow57:
@TobleroneSwordfish Are you okay with this PR if it's only to quick store to belt/backpack, but not take out?

PR #21430, opened 2024-11-24
Banish emojis from character names

TobleroneSwordfish:
## About the PR 
Just adds some regex validation to make emojis illegal characters in character names.

## Why's this needed? 
Missing validation, emojis don't make any sense in character names.

## Changelog 

```changelog
(u)LeahTheTech
(+)Character names can no longer contain emojis.
```


Sbmhawk:
But my SOUL.....

Laboredih123:
goonstation has fallen, millions must write letters

PR #21439, opened 2024-11-25
Experimentally add a useSafeBackend alternative

mordent-goonstation:
[UI][Code Quality]

## About the PR 

Adds a `useSafeBackend` alternative to `useBackend`, that returns the same thing but with the data type set to one that recursively adds null/undefined as possible values received. Uses in CyborgModuleRewriter interface, which was decently safe already, to demonstrate usage and see if it might have value.

This change makes it so that if this alternative is used our tooling would require data is handled in a way that handles `null` or `undefined` values gracefully. This doesn't enforce that the UI displays as such to the user, but it should at least "work".

## Why's this needed? 

One of the greatest causes of bluescreen tgui errors for us is when the "contract" between the backend and frontend is broken, and the tgui interface receives data that it was not expecting. Often, some obscure edge case that the coder didn't think was even possible, let alone likely, during regular play occurs that means something that *should* be there... isn't.

Putting this up as a draft because I think this is probably overkill, but I think there's some merit to perhaps making this available for coders to use when making their interface so that they can consider appropriate places to handle unexpectedly missing data in a better way. In *most* cases I imagine this will be viewed as annoying boilerplate code/having to wrestle with types, and makes the code harder to read.

I anticipate this PR not being merged, but perhaps the `useSafeBackend` and the `NestedPartial` type could be kept around for people to use during development if they wish to see how well their interface stands up to it.

## Explanation of the Problem

The type of error that this is intended to help prevent: [Unable to get property 'foo' of undefined or null reference](https://github.com/goonstation/goonstation/issues?q=is%3Aissue%20%22of%20undefined%20or%20null%20reference%22)

Giving a contrived example that might help explain the problem it's trying to stop occuring:

```tsx
interface MyWidgetData {
  things: ThingData[]
}

interface ThingData {
  name: string;
}

export const MyWidget = () => {
  const { data } = useBackend<MyWidgetData>();
  const { things } = data;
  return (
    <Stack>
      {things.map(thing) => <Stack.Item>{thing.name}</Stack.Item>}
    </Stack>
  );
};
```
Looks all well and good. But what if `ui_data` doesn't do its list setting properly in some unforeseen case and doesn't add `things` to the list returned? Now when it gets to `things.map` it'll be trying to do `undefined.map`, which will cause an error.

If the coder used `useSafeBackend` instead, they would be warned to handle that case (by e.g. `(things ?? []).map(...)`) instead.

Not saying this is a "real case", but giving an example of the sort of thing that could cause the issue above: if a Widget when spawned gets a special kind of container instance spawned within it, that you can swap out by clicking on it with that kind of item in hand, you might think you can assume that this is still safe, and that the `things` (which is actually the contents of that container) is always going to exist.

However, due to someone messing around with explosives and planting one inside of the container, what if setting that off destroyed the container but let the machine working? Or maybe an admin foolishly teleported it to their mob? Now the Widget no longer has one of those containers, so when you open the Widget's UI you get an angry blue screen because how it can it send you a list of contents of the container that it no longer has?

(Above example assumes the `ui_data` looks for it in `contents` or whatever.)


Garash2k:
Would there be a way to log warnings when the "contract" is broken? Because I feel like improving the contract over time to be more accurate, supporting those cases as needed, is better than assuming *everything* could always be undefined.

And even undefined props asides, whacky shenanigans could make it that we end up with the wrong type of data. Which this wouldn't help with, but the above validation would catch.

mordent-goonstation:
Runtime validation of data structure is rough because it's arguably decently performantly expensive (depending on data volume/complexity). A more ideal way is to have the manual types we currently add actually provided by an automatic process instead that is used to ensure front and back-end stay in sync by both conforming to it.

The real issue with that is that I'm not aware of a way to elegantly generate that and ensure is complied with at the DM level. I would love for them to be generated for the sake of not having manual typing errors.

We don't really see cases of the *wrong* thing being given, because that typically means the feature doesn't work.

mordent-goonstation:
Closing because this is a very over-the-top and obscure way to handle this. I may add an alternative way for people to achieve something similar if they want for debug/dev purposes, but this is likely to just cause confusion rather than help right now.

PR #21464, opened 2024-11-27
Prevent ant growth on conveyor belts

vortex1942:
[Catering][QOL]

## About the PR 

Adds Conveyor belts to the list of ant safe conditions.

## Why's this needed? 

This allows players to setup sushi belts without the food being coated in copious amounts of ants
image


TobleroneSwordfish:
Okay but how do the ants get on the conveyor belt 🤔

PR #21470, opened 2024-11-27
Add glow to hi-vis vests

Glamurio:
[Game Objects] [QoL]

## About the PR 
Adds a light reflection overlay to `/obj/item/clothing/suit/hi_vis` and `/obj/item/clothing/suit/hi_vis/puffer` to stand out in the dark.

Changes `/obj/item/clothing/suit/puffer/hi_vis` to `/obj/item/clothing/suit/hi_vis/puffer`.

image image

Known issues:
Because of `PLANE_SELFILLUM` limitations, the vest only glows when worn, otherwise it would clip over everything.
Likewise, beards and backpacks are also getting rendered over.

## Why's this needed? 
Inspired by this thread:
https://forum.ss13.co/showthread.php?tid=23395

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Hi-vis vests now actually reflect light in dark areas.
```


ZeWaka:
i like this but i think it's too bright, it should just be a subtle glow

Glamurio:
More subtle glow with less alpha.
Before: image After: image

PR #21471, opened 2024-11-27
Refactors HoS Medal to be a badge and lets you show it off

Glamurio:
[Game Objects] [QoL]

## About the PR 
Changes `/obj/item/clothing/suit/hosmedal` to `/obj/item/clothing/suit/security_badge/hosmedal`.

Adds `/obj/item/clothing/suit/security_badge/show_off_medal()` to reduce duplicate code.

Reworks `/obj/decal/framed_award` to contain its award instead of doing a `locate` inside its contents.

Adds an `award_text` var to HoS medals which are displayed when showing off the medal.

## Why's this needed? 
Medals are cool and underused, this gives them a cool new gimmick usage.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)You can now show off the HoS medal like other badges.
```

### DIFF
--- a/code/obj/item/clothing/suits.dm
+++ b/code/obj/item/clothing/suits.dm
@@ -2113,17 +2113,20 @@ ABSTRACT_TYPE(/obj/item/clothing/suit/sweater_vest)
 	get_desc()
 		. += "This one belongs to [badge_owner_name], the [badge_owner_job]."
 
-	attack_self(mob/user as mob)
+	proc/show_off_badge(var/mob/user, var/mob/target = null)
 		if(ON_COOLDOWN(user, "showoff_item", SHOWOFF_COOLDOWN))
 			return
-		user.visible_message("[user] flashes the badge: <br>[SPAN_BOLD("[bicon(src)] Nanotrasen's Finest [badge_owner_job]: [badge_owner_name].")]", "You show off the badge: <br>[SPAN_BOLD("[bicon(src)] Nanotrasen's Finest [badge_owner_job] [badge_owner_name].")]")
+		if (istype(target))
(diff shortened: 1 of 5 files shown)


code/obj/item/clothing/suits.dm
```diff
@@ -2146,14 +2149,24 @@ ABSTRACT_TYPE(/obj/item/clothing/suit/sweater_vest)
...
-/obj/item/clothing/suit/hosmedal
+/obj/item/clothing/suit/security_badge/hosmedal
 	name = "war medal"
 	desc = ""
 	icon = 'icons/obj/clothing/overcoats/item_suit_gimmick.dmi'
 	w_class = W_CLASS_TINY
 	wear_image_icon = 'icons/mob/clothing/overcoats/worn_suit_gimmick.dmi'
 	icon_state = "hosmedal"
 	icon_state = "hosmedal"
+	var/award_text = "This is a medal. There are many like it, but this one's mine."
```
JORJ949, line 2160: This appears to everyone when used by anyone to anyone so I don't think this is great flavour text, it should probably say something like "Awarded for great bravery." because its supposed to be a war medal, though im not sure the flavor text is needed.

JORJ949 (changes requested):
Do the hand sprites look alright with this? The ribbon that goes around the neck is quite large on the sprite.

code/obj/item/clothing/suits.dm
Glamurio: This actually gets overwritten by the `/framed_award/hos_medal` it spawns in. The only way you'd ever see that `award_text` is if it was hand-spawned by admins.

Glamurio:
> Do the hand sprites look alright with this? The ribbon that goes around the neck is quite large on the sprite.

I thought it was fine, yeah it'll cover bits of the face, if need-be we could make a ribbon-less variant to show-off instead.

flappybatpal:
Two things that come to mind.

1. I believe the intent was that you _had_ to wear the HoS medal as a suit, as a status symbol/flex. Perhaps not very realistic in practice but still I think needs to be double checked.
2. Feels like it should have a custom sprite but that could be done later.

ZeWaka:
X

PR #21481, opened 2024-11-28
Fix catwalk auto-connection

pgmzeta:
[bug][mapping]
## About the PR 
Use the `connects_to` var set for the `catwalk` grille subtype. Previously it was using `connects_to_obj`, completely skipping the specially defined list.

## Why's this needed? 
Catwalks auto-connect to windows with grilles under them which looks kinda ugly and isn't intended by code.

### DIFF
--- a/code/obj/grille.dm
+++ b/code/obj/grille.dm
@@ -101,8 +101,8 @@
 				var/connectdir = 0
 				for (var/dir in cardinal)
 					var/turf/T = get_step(src, dir)
-					for (var/i in 1 to length(connects_to_obj))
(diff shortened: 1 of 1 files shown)


Tarmunora (approved):
Feels kinda like it should be overriding `connects_to_obj` rather than just declaring a new var and using that instead, but if it works it works I guess 🤷

pgmzeta:
they probably shouldn't be `grille` subtypes considering how much is overridden and how they differ functionality-wise but that is a much bigger change :(

PR #21488, opened 2024-11-29
Give the NTSC EVA Access

JORJ949:
## About the PR 
Moves NTSC access into the get_access() function
Grants NTSC EVA access

## Why's this needed? 
The NTSC comes with a space suit and is supposed to be able to navigate through space, EVA access allows them to do this more effectively without having to ask the AI for access to something they probably should already have access to

## Changelog 

```changelog
(u)JORJ949
(+)The NTSC now has access to EVA.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -2546,6 +2546,7 @@ ABSTRACT_TYPE(/datum/job/special/halloween/critter)
 	unique = TRUE
 	wages = PAY_TRADESMAN
 	trait_list = list("training_security")
+	access_string = "Nanotrasen Security Consultant"
 	requires_whitelist = TRUE
 	requires_supervisor_job = "Head of Security"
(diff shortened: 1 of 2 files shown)


DisturbHerb:
I'm reasonably certain that it has been stressed that the NTSC role is not supposed to be HoS-lite. This is a lot of access to be giving them, even if it's behind the same whitelist. I suggest a feedback thread.

JORJ949:
> I'm reasonably certain that it has been stressed that the NTSC role is not supposed to be HoS-lite. This is a lot of access to be giving them, even if it's behind the same whitelist. I suggest a feedback thread.

I do not believe this will turn them into much of a hoslite, I believe most of the HoS' strength comes from the armory, their weapons and their additional general area access, but a feedback thread is still a good idea

TobleroneSwordfish (approved):

PR #21491, opened 2024-11-29
Wizard White Hole

pgmzeta:
[feature][events]
## About the PR 
Add a wizard white hole effect that emits some wizard hats/robes/staffs, some mugwort/everyflavor jelly beans, and a bunch of spell projectiles.

## Why's this needed? 
More white hole variety & dangerous white holes. The fireballs will not explode, just cause fire flashes, so this shouldn't cause breaches. The prismatic projectiles can be deadly quite quickly, especially combined with the magic missiles.

## Changelog 

```changelog
(u)glowbold
(+)Beware the curse of a wizarding white hole.
```


TobleroneSwordfish:
Oh this is cool, I'll add some other things to it!

PR #21494, opened 2024-11-29
Makes address fields consistent on atmos devices.

Laboredih123:
[ATMOSPHERICS][WIKI]
## About the PR 
Changes the netid field to address_1 and sender like other packet devices.

## Why's this needed? 
consistency good. Also adds some missing features to vent scrubbers.

## Changelog 

```changelog
(u)cringe
(+)The netid field on packets was changed to address_1 and sender for atmospheric devices.
```


TobleroneSwordfish (approved):
Makes sense

PR #21498, opened 2024-11-30
Increase quantity of counterrev boxes on Donut 3

JORJ949:
## About the PR 
Increases quantity of boxes of counterrev implants on Donut 3 from 1 to 4, to match cogmap 2
One in the armory, one in the HoS office, one on the main sec table (and the original one by the sleeper to the south)

## Why's this needed? 
For a highpop map with cargo insanely far from security and bridge one counterrev box is nowhere near enough to lead a decent counterrev operation compared to maps such as cog1 which has 2 boxes.

### DIFF
--- a/maps/donut3.dmm
+++ b/maps/donut3.dmm
@@ -2692,8 +2692,8 @@
 /obj/table/reinforced/bar/auto{
 	name = "table"
 	},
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (approved):
Makes sense

PR #21501, opened 2024-11-30
Add Dimensional Key artifact

FlameArrow57:
[SCIENCE][FEATURE]

## About the PR 
This PR adds a new tiny-sized artifact, the Dimensional Key. Idea is that it opens doors to a dimensional traversal point

This artifact can be used 1 time per cardinal direction to create a door on a wall to a dimensional fissure area, where you may exit through a door to any other previously created doorway. This allows for point-to-point travel in a long distance, across z-levels, or "non-euclidean" manner. The artifact acts as a key that can open any of the doors whenever

Proof of concept:
https://streamable.com/g0peco

More details:
-Deactivating the artifact will not destroy the doors or fissure area, meaning they are permanent
-Doors can only be opened from the outside by using the key. Otherwise, inside the fissure, you can open the doors without the key
-Items and objects that enter will corrode heavily while inside unless held, meaning using the dimension as a storage is very limited
-Artifact faults are rolled upon opening a door with the key
-Examine hint for the artifact is still "It kinda looks like it's supposed to be inserted into something", but you can tell the difference from large power cell artifacts because those ones start automatically activated

## Why's this needed? 
-Thought the idea of a key that opens a door to some doorway dimension would be cool
-Use of visual mirror type features is pretty scarce right now, thought adding one to artifacts would be cool

I don't think it's too strong, as 
-you can already do this somewhat with mech comp teleporterss. This just happens to be more secure, accessible to only the key holder, and removes the mech comp barrier
-placeability of doors will be limited

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new tiny-sized Dimensional Key artifact! See minor changelog for details
(+)The dimensional key acts as a key that can be used once per cardinal direction on a wall to create a doorway to a dimensional fissure, which can be accessed through any of the doors, and exited to any of the doors
(+)Things that enter the fissure that aren't secured on you will corrode and be destroyed over a few seconds
```


Glamurio:
This is so cool

code/obj/artifacts/artifact_items/key.dm
```diff
@@ -0,0 +1,352 @@
...
+					T.density = station_turf.density
+					T.opacity = station_turf.opacity
+					T.name = station_turf.name
+					T.desc = station_turf.desc
+					T.icon = station_turf.icon
+					T.icon_state = station_turf.icon_state
+					//T.fullbright = TRUE
+					T.RL_Init()
+
+/datum/artifact/key
```
Glamurio, line 159: Might be better to rename this to be "dimensional_key" as to not confuse it with "activator_key".

Glamurio (changes requested):

code/obj/artifacts/artifact_items/key.dm
FlameArrow57: Yeah these are more temporary names, gonna give them better names before final draft

code/obj/artifacts/artifact_items/dimensional_key.dm
```diff
@@ -0,0 +1,402 @@
...
+
+	afterattack(atom/target, mob/user, reach)
+		..()
+		if (!src.artifact.activated)
+			return
+		if (!reach)
+			return
+		if (!iswall(target))
+			return
+		if (BOUNDS_DIST(user, target) > 0 || istype(get_area(target), /area/artifact_fissure) || user.z == Z_LEVEL_SECRET || user.z == Z_LEVEL_ADVENTURE)
```
ZeWaka, line 31: might want to check that its not `Z_LEVEL_DYNAMIC` also

code/obj/artifacts/artifact_items/dimensional_key.dm
```diff
@@ -0,0 +1,402 @@
...
+		..()
+
+	Crossed(atom/movable/AM)
+		if (isliving(AM) && !isintangible(AM))
+			var/area/artifact_fissure/fissure_area = get_area(src)
+			fissure_area.update_visual_mirrors(src.station_ref, src.entrance_loc)
+			return ..()
+		return ..()
+
+/obj/artifact_fissure_door
```
ZeWaka, line 212: i feel like we should put this under something, even if its abstract or similar, just for organization

code/obj/artifacts/artifact_items/dimensional_key.dm
Tarmunora: Or, check for restricted Z instead of hardcoding it

code/obj/artifacts/artifact_items/dimensional_key.dm
FlameArrow57: I think I did what you were thinking, let me know if not

code/modules/status_system/statusEffects.dm
```diff
@@ -2975,3 +2975,32 @@
 		src.art = null
 		qdel(src.glimmer)
 		src.glimmer = null
+
+/datum/statusEffect/art_fissure_corrosion
+	id = "art_fissure_corrosion"
+	effect_quality = STATUS_QUALITY_NEGATIVE
+	var/corrosion_stacks = 0 // 1 stack per tick
```
Tarmunora, line 2982: Might be better to have this status cleared on area `Exited()`, and track state via atomproperty, rather than keeping an infinite duration status ticking forever on any obj that's touched the floor was and then removed

code/obj/artifacts/artifact_items/dimensional_key.dm
```diff
@@ -0,0 +1,388 @@
...
+				boutput(user, SPAN_NOTICE("[src] doesn't seemed to be locked, but won't open either... strange."))
+			else
+				src.open()
+		else
+			src.deny_open()
+			boutput(user, SPAN_NOTICE("[src] is locked. Perhaps a key will open it?"))
+
+	attackby(obj/item/I, mob/user)
+		if (!istype(I, /obj/item/artifact/dimensional_key) || !I.artifact.activated)
+			return ..()
```
Tarmunora, line 239: intentional that any key art can be used, and not just the one associated with the door?

Tarmunora (changes requested):

code/obj/artifacts/artifact_items/dimensional_key.dm
FlameArrow57: Intentional yeah

Tarmunora (approved):

code/obj/artifacts/artifact_items/dimensional_key.dm
```diff
@@ -0,0 +1,394 @@
...
+
+	Crossed(atom/movable/AM)
+		if (istype(AM, /obj/projectile))
+			var/obj/projectile/P = AM
+			var/obj/projectile/new_proj = initialize_projectile(src.exit_turf, P.proj_data, P.xo, P.yo, P.shooter)
+			new_proj.travelled = P.travelled
+			new_proj.launch()
+			P.die()
+		else if (AM.dir == src.required_dir && !istype(AM, /obj/art_fissure_objs/door) && !istype(AM, /obj/art_fissure_objs/cross_dummy))
+			AM.set_loc(src.exit_turf)
```
TobleroneSwordfish, line 164: Ideally this should either use `warptarget` or duplicate the weird double `set_loc` behaviour warp targets do to fix the glide animation.

code/obj/landmark.dm
```diff
@@ -642,6 +642,10 @@ var/global/list/job_start_locations = list()
 	warptarget_modifier = LANDMARK_VM_WARP_NONE
 /// target turf for projecting its contents elsewhere
 /turf/var/turf/vistarget = null
+/// turfs to project speech to (overlaps with vistarget speech)
+/turf/var/list/listening_turfs = null
```
TobleroneSwordfish, line 646: When you say it overlaps, why is this necessary instead of just relying on the existing vistarget behaviour?

TobleroneSwordfish (changes requested):
This is such a creative use of vis mirrors, really cool!
~~I am a little concerned about people getting stuck in the fissure though, particularly if the key is destroyed. Maybe if all the doors are shut and the corresponding key has been deleted anyone inside should be teleported to the void after a while?~~

FlameArrow57:
> This is such a creative use of vis mirrors, really cool! I am a little concerned about people getting stuck in the fissure though, particularly if the key is destroyed. Maybe if all the doors are shut and the corresponding key has been deleted anyone inside should be teleported to the void after a while?

The doors open from the inside without the key

code/obj/landmark.dm
FlameArrow57: I didn't want to override vistarget in case anything else was using it (I thought it made sense for things like Centcomm, but on station might be different)

code/obj/artifacts/artifact_items/dimensional_key.dm
FlameArrow57: Nice, with the double set_loc, seems it does let you see past the walls at the entry/exit for a very small moment, if that's an issue

TobleroneSwordfish: Hmm, I thiink that's better than the jerky teleport? Up to you though, I can't think of a way to get the best of both as it were.

FlameArrow57: kk, yeah I think it does look better

code/obj/artifacts/artifact_items/dimensional_key.dm
```diff
@@ -0,0 +1,410 @@
...
+			// makes it look like an animation glide
+			switch (src.required_dir)
+				if (NORTH)
+					AM.set_loc(get_step(src.exit_turf, SOUTH))
+				if (EAST)
+					AM.set_loc(get_step(src.exit_turf, WEST))
+				if (WEST)
+					AM.set_loc(get_step(src.exit_turf, EAST))
+				if (SOUTH)
+					AM.set_loc(get_step(src.exit_turf, NORTH))
```
TobleroneSwordfish, line 173:
```suggestion
			AM.set_loc(get_step(src.exit_turf, turn(AM.dir, 180)))
```
Should be able to guarantee this dir is right given the above check

TobleroneSwordfish (changes requested):

TobleroneSwordfish (approved):

Tarmunora:
Rarity weight might end up being cranked down to ~wishmaker levels. We'll see

PR #21509, opened 2024-12-01
constructible atmos

Laboredih123:
[ATMOSPHERICS][FEATURE]
## About the PR 

https://github.com/user-attachments/assets/f16de933-20c7-4dfc-94f2-2164f5275f8c

## Why's this needed? 
its neat

## Changelog 

```changelog
(u)cringe, LeahTheTech
(*)Constructible atmos, no longer a week away. Find the HPD in the CE's RCD crate, or in the RD locker along with a big RCD ammo charge.
```


Laboredih123:
oh yeah dont look at the code at all lol the hhd doesnt even place anything right now lol

code/obj/item/handheld_dispenser/dispenser.dm
```diff
@@ -0,0 +1,88 @@
...
+		"Pressure tank" = /obj/machinery/atmospherics/unary/tank,
+		"Passive gate" = /obj/machinery/atmospherics/binary/passive_gate,
+		"Pressure pump" = /obj/machinery/atmospherics/binary/pump,
+		"Volume pump" = /obj/machinery/atmospherics/binary/volume_pump,
+		"Manual valve" = /obj/machinery/atmospherics/binary/valve,
+		"Digital valve" = /obj/machinery/atmospherics/binary/valve/digital,
+		"Outlet Injector" = /obj/machinery/atmospherics/unary/outlet_injector,
+		"Vent pump" = /obj/machinery/atmospherics/unary/vent_pump,
+		"Vent scrubber" = /obj/machinery/atmospherics/unary/vent_scrubber,
+	)
```
FlameArrow57, line 27: Would it be better to use defines for these names? Thinking it is going to be confusing remembering the upper/lower casing for these and probably would be better if these names are being used a lot

TobleroneSwordfish: Could actually use `::` syntax to get them directly, but this whole thing needs to be datumized anyway I think

TobleroneSwordfish:
UI work done here: https://github.com/Laboredih123/goonstation/pull/3

JORJ949:
Couple bugs I noticed: The placed valve has its off and on icons backwards and the "maximum" materials doesnt actually do anything it let me fill to at least 101

Laboredih123:
> Couple bugs I noticed: The placed valve has its off and on icons backwards and the "maximum" materials doesnt actually do anything it let me fill to at least 101

could not reproduce valve. and there is no maximum. at least, not until one is shown to be needed

JORJ949:
> 
> could not reproduce valve. and there is no maximum. at least, not until one is shown to be needed

The exact setup was this
image
I _think_ I had the valve on, then plugged in the tanks and it didnt flow, then when I turned it off it did flow, specifically the manual valve

Laboredih123:
tried it but it flowed when open. can ya record yourself reproducing it?

PR #21527, opened 2024-12-02
Do not give vandalism points for spraying an already graffiti'd turf

pgmzeta:
[balance][gamemodes]
## About the PR 
When laying down graffiti for a vandalism objective, check if there is already graffiti on the tile before adding vandalism points.

## Why's this needed? 
A strat I've seen on live is to stand in one place and spray the same three tiles over and over until the bag pops out. This will make people actually have to move around the target area to vandalize it.

## Changelog 

```changelog
(u)glowbold
(+)In gang war, when vandalizing areas, you do not gain points for spraying over existing graffiti.
```


TDHooligan:
bo-ring! this is a good pr

TobleroneSwordfish (approved):

PR #21531, opened 2024-12-02
Lock the Detective's gun behind armoury auth

JORJ949:
## About the PR 
Adds an /armory subtype of box that is locked unless: the armory is currently authorized, anyone with armory access unlocks it, or it is emagged (permanently unlocking it).
Repaths the detective gun box to this subtype, locking the gun behind armory or antagonist actions.

The box can still be opened without armory, emag or HoS but it requires blowing something up nearby to set the box on fire, destroying it and dropping the contents, meaning antagonists can still get in without an emag, and non-antags can use it as a last resort.

Increases ammo counts for both ammos in the box from 2 > 3 to account for increased difficulty to open the box.

## Why's this needed? 
The idea comes from[ this thread ](https://forum.ss13.co/showthread.php?tid=23355) where I expressed concerns about how detective is essentially being treated like a security officer with lethals, and is a prime mindhack target for antagonists and the like, this heavily nerfs the detective's lethal capabilities, but they still have access to flashes and cuffs from sectechs like assistants do.

Please direct discussion on the balance of this PR to the above thread.

## Changelog 

```changelog
(u)JORJ949
(*)The detective's gun box is now locked. It is unlocked when the armory is authorized, anyone with armory access unlocks it or it is emagged. It can still be destroyed to drop its contents.
```


cropseyy:
I genuinely don't see the point of this. Dealing with the powergamers is better than taking a chunk out of a fun role.  Also detective should get equipment access, which I don't think they have already else they won't be able to use the lockers or anything.

(And yes, I'm putting this on the PR so people can see that people do disagree.)

Sbmhawk:
This would turn det into a glorified loot pinata.
No means of fighting back against someone trying to steal the box to get lethals for free.

PastaSpaghett:
I'm not completely against making the Detective's gun less lethal, but this is one of the worst ways to nerf it, making it less of a self-defense weapon and more of a free gun for any somewhat competent traitor.

JORJ949:
As specified in the PR description, please specify further balance discussion to the thread.

TobleroneSwordfish:
Yeah I'm just going to close this now, we're not taking the detective's gun away.

Tarmunora:
Additionally, going to forward what @Carton171 said on the forum thread, because I think it's worth putting here as well:

> Additionally, I don't believe making nuclear level PRs in order to push discussion forward is productive, as it tends to just result in emotional arguing and every other response being the same refutation of that PR. It also feels like an underhanded tactic to get people to accept ideas they normally wouldn't, as an alternative to the nuclear option.

PR #21536, opened 2024-12-02
Solar Flare events significantly increase solar panel output

pgmzeta:
[balance][events]
## About the PR 
* Move the define for the default solar generation rate to the machinery defines
* Keep track of the current solar generation rate as a global
* The solar flare event multiplies the generation rate by signal loss

## Why's this needed? 
Hotwiring solars should have some risk.

## Changelog 

```changelog
(u)glowbold
(*)Solar flare events super-charge solar panels. Beware of hot-wires.
```

### DIFF
--- a/_std/machinery.dm
+++ b/_std/machinery.dm
@@ -48,6 +48,10 @@
 #define PROCESSING_TIER_MULTI(target) (1<<(target.current_processing_tier-1)) //! Scalar to behave as if it were running at full speed
 #define MACHINE_PROCS_PER_SEC (MACHINE_PROC_INTERVAL / (1 SECOND))
 
+// Previous SOLARGENRATE was 1500 WATTS processed every 3.3 SECONDS.  This provides 454.54 WATTS every second
+// Adjust accordingly based on machine proc rate
+#define DEFAULT_SOLARGENRATE (454.54 * MACHINE_PROCS_PER_SEC)
+
 #define PROCESSING_FULL      1
 #define PROCESSING_HALF      2
 #define PROCESSING_QUARTER   3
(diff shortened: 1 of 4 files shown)


JORJ949:
Im not sure discouraging engineers from setting up the power is a smart idea, it already doesnt get done a lot on classic. Not to mention this outright prevents borgs from setting up power for 2/3 engines.

pgmzeta:
> Im not sure discouraging engineers from setting up the power is a smart idea, it already doesnt get done a lot on classic. Not to mention this outright prevents borgs from setting up power for 2/3 engines.

This doesn't prevent proper setup of the solars through an SMES, just direct hotwiring.

hellotherereading:
I like this. Engineers would be encouraged to mechcomp the console to the smes instead of boring hotwiring(atleast, without consequences)

Connie-Dun:
As someone who's done solars a ton, I love this - didn't like there not being a real point to actually setting up solar SMESes, which took a fair bit of the fun out of it. Importantly, it also gives another route for a villain to cause overload issues without needing to use the engine itself.

_std/machinery.dm
```diff
@@ -48,6 +48,10 @@
 #define PROCESSING_TIER_MULTI(target) (1<<(target.current_processing_tier-1)) //! Scalar to behave as if it were running at full speed
 #define MACHINE_PROCS_PER_SEC (MACHINE_PROC_INTERVAL / (1 SECOND))
 
+// Previous SOLARGENRATE was 1500 WATTS processed every 3.3 SECONDS.  This provides 454.54 WATTS every second
+// Adjust accordingly based on machine proc rate
+#define DEFAULT_SOLARGENRATE (454.54 * MACHINE_PROCS_PER_SEC)
```
ZeWaka, line 53: bro just use whole numbers
```suggestion
// Previous SOLARGENRATE was 1500 WATTS processed every 3.3 SECONDS.  This provides 455 WATTS every second
// Adjust accordingly based on machine proc rate
#define DEFAULT_SOLARGENRATE (455 * MACHINE_PROCS_PER_SEC)
```

ZeWaka (approved):

PR #21546, opened 2024-12-04
Adds the GR1-Mk3 Dirt Dispenser, a WA-V3 syndicate variant

Glamurio:
[Game Objects] [Feature]

## About the PR 
Adds `/obj/item/reagent_containers/glass/backtank/syndicate` and `/obj/item/gun/sprayer/syndicate`, equating to the "GR1-Mk3 Dirt Dispenser". The GR1-Mk3 spawns with 500 units of sewage.

Adds `/datum/syndicate_buylist/traitor/backtank` as a purchasable option for the janitor.

Also reworks `/datum/projectile/special/shotchem` to grab the average color of the chemicals inside the projectile if there's no color supplied (which is the case for the WA-V3 projectile), to make it look nicer.

Notably, the GR1-Mk3 does not have a black list, and the nozzle instead clogs up randomly with a low chance.

## Why's this needed? 
The WA-V3 is really cool, and I think the idea of someone spreading chaos with it seems fun. Obviously that's too strong for the regular variant, but I felt that a traitor purchase seemed like a good power level. It still requires players to supply their own chemicals, unless they just want to make the station stinky.

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Syndicate Janitors may now purchase the GR1-Mk3 Dirt Dispenser.
```


JORJ949:
Clearly we need a janitor purchase of a chemical tank full of dirt and grime like scientist poison bottles

code/obj/item/mops_cleaners.dm
```diff
@@ -1272,6 +1272,25 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
 		src.set_current_projectile(new /datum/projectile/special/shotchem/wave/wide/emagged)
 		src.projectiles = null
 
+	syndicate
+		name = "\improper GR1-Mk3 Dirt Dispenser"
+		desc = "The Grime Release, Version 1, Mark III Dirt Dispenser, absolute bane of cleanliness.<br>\
+			Must ONLY be used with one of many Syndicate licensed GR1-Mk3 back tanks."
```
JORJ949, line 1278: Are you sure they only work with the correct backtank? as far as I can see theres no special case checking for the syndicate one.

Glamurio: No, I mostly referenced the original description. I could put a type check, but I figured what's the harm in letting them mix and match, especially considering the WA-V3 is just worse in terms of blacklisting.

code/obj/item/mops_cleaners.dm
```diff
@@ -1346,6 +1369,23 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
 			return
 		. = ..()
 
+	syndicate
+		name = "\improper GR1-Mk3 back tank"
+		desc = "A little label on the side reads \"Now with corrosive substances support!\"."
+		icon_state = "janitor_tank_s"
+		item_state = "janitor_tank_s"
+		initial_volume = 500
+		initial_reagents = list("sewage" = 500)
```
TobleroneSwordfish, line 1378: Might be fun if it was full of a random mixture of industrial waste style chemicals: sewage, "yuck", carbon, grease, oil, toxic slurry etc.

TobleroneSwordfish (changes requested):
Probably not a very good traitor item but potentially a very funny one that could enable shenanigans.

TobleroneSwordfish (approved):

code/obj/item/mops_cleaners.dm
```diff
@@ -1346,10 +1365,51 @@ TYPEINFO(/obj/item/handheld_vacuum/overcharged)
...
+				var/selected_reagent = pick(reagent_pool)
+				var/amt = 0
+				// Ensure the last reagent gets all remaining volume
+				if (length(reagent_pool) == 1)
+					amt = total_volume
+				else
+					// Dynamically adjust max and min to prevent underflow
+					var/max_amt = min(total_volume - (length(reagent_pool) - 1), max(10, total_volume / 5))
+					var/min_amt = min(10, max_amt)
+					amt = round(rand(min_amt, max_amt))
```
frawhst, line 1392: I'm confused where some of these numbers are coming from- why divide total volume by 5? Why subtract the length of the reagent pool (I assume to get at least 1 unit of each reagent in?)
At the very least, this could use a more comprehensive comment to explain why this math is being done here.

PR #21548, opened 2024-12-04
Adds a buffering mechcomp component

patricia-izin:
[Feature]
## About the PR 

Adds a mechcomp component that buffers up signals and doles them out at the speed most components are restricted to(configurable), especially wifi and the speaker.

Offers several buffer modes including first in first out, first in last out, a random mode, and a ring buffer where old signals are overwritten.

This PR also ends up adding a new process just for mechcomp instead of using the item loop, because the item loop was too slow for what we needed. The mechcomp loop now runs at 2.8s per default instead of 2.9 (balance change??)

image

image

http://0x0.st/8r_r.mp4 Demo video

## Why's this needed? 

It's kind of jank at the moment dealing with things like packets and chat messages where we might get signals far faster than we can process them. You need a selection component, a control unit, and a delay and functionality is limited.

## Changelog 

```changelog
(u)patricia
(*)Adds a new buffering component to the mechanic's vendor. Read about it in the mechanic's book!
```


code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,187 @@
...
+		var/inp = input(user,"Set size of signal buffer","Buffer size", buffer_size) as num
+		if(!in_interact_range(src, user) || user.stat)
+			return 0
+		inp = round(inp)
+		inp = min(inp, buffer_max_size)
+		if(inp == 0)
+			boutput(user,"Buffer size can't be zero.")
+			return
+		if(inp < 0)
+			boutput(user,"How could a buffer be negative? What are you doing here?")
```
ZeWaka, line 1910: consider making it beep angrily

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,187 @@
...
+	cooldown_time = 0.4 SECONDS
+	var/list/buffer = list()
+	var/buffer_size = 15
+	//I wanted to set this to the same limit as the selection component
+	//But the selection compoment doesn't actually have a limit in the additem() proc
+	//It kind of worries me
+	var/buffer_max_size = 200
+	//Ring reader ring writer
+	var/rr = 1
+	var/rw = 1
```
ZeWaka, line 1842: then name it as such, we are not limited in the length of our variable names

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,187 @@
...
+				signal = buffer[bufl]
+				buffer.Remove(buffer[bufl])
+			if(buffer_model == RING_BUFFER && bufl >= rr && !isnull(buffer[rr]) )
+				signal = buffer[rr]
+				buffer[rr] = null
+				rr++
+				if(rr > buffer_size)
+					rr = 1
+			if(buffer_model == RANDOM_BUFFER)
+				var ran = rand(1,bufl)
```
ZeWaka, line 1981: `var x` is not allowed in this codebase, please use `var/x`

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,187 @@
...
+			tooltip_rebuild = 1
+			boutput(user, "Set delay to [inp]")
+			return 1
+		return 0
+
+
+	proc/setBufferSize(obj/item/W as obj, mob/user as mob)
+		var/inp = input(user,"Set size of signal buffer","Buffer size", buffer_size) as num
+		if(!in_interact_range(src, user) || user.stat)
+			return 0
```
ZeWaka, line 1903: not a big fan of this input checking, there's much better methods these days. For example, if a user is paralyzed on the ground this would still work. try `can_act` and `tgui_input_number`

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,187 @@
...
+			flick("[under_floor ? "u":""]comp_buffer1", src)
+			var/transmissionStyle = changesig ? COMSIG_MECHCOMP_TRANSMIT_DEFAULT_MSG : COMSIG_MECHCOMP_TRANSMIT_MSG
+			SPAWN(0) SEND_SIGNAL(src,transmissionStyle,signal)
+
+
+
+
+	update_icon()
+		icon_state = "[under_floor ? "u":""]comp_buffer"
+		return
```
ZeWaka, line 1997: unnecessary
```suggestion
```

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,187 @@
...
+			return
+		if(inp < 0)
+			boutput(user,"How could a buffer be negative? What are you doing here?")
+			return
+		buffer_size = inp
+		tooltip_rebuild = 1
+		boutput(user,"You set the buffer size to [inp]")
+		qdel(buffer)
+		buffer = list()
+		return
```
ZeWaka, line 1917:
```suggestion
```

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,187 @@
...
+			buffer.Add(input)
+			return
+
+	process()
+		..()
+
+		if(level == OVERFLOOR || ON_COOLDOWN(src, SEND_COOLDOWN_ID, src.cooldown_time)) return
+
+		var/bufl = (length(buffer))
+		var signal = null
```
ZeWaka, line 1965: x

ZeWaka (changes requested):
I think there's a book you need to add an entry to as well

patricia-izin:
@ZeWaka Alright I think that takes care of all of that thanks for the review

patricia-izin:
image

ZeWaka (approved):

patricia-izin:
By the way if you're trying to test this you can hook a button to a counter to a buffer to a speaker and you'll be able to observe if the numbers output how you expect them to for the given mode

patricia-izin:
http://0x0.st/8r_r.mp4

I've added a complete demo video if that helps with the review process at all

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,195 @@
...
+		else if(model == "RING")
+			model = "ring"
+			buffer_desc = "This mode outputs from oldest to newest, overwriting the oldest signal when full."
+			buffer_model = RING_BUFFER
+		else if(model == "RANDOM")
+			model = "random"
+			buffer_desc = "This mode outputs signals randomly, randomly overwriting previous signals when full."
+			buffer_model = RANDOM_BUFFER
+		else
+			return FALSE
```
Sovexe, line 1880: Consider using a `switch` statement here

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,195 @@
...
+			//That means that item is the oldest, jump the reader to it
+			//Watch out for the edge
+			if(ring_reader == ring_writer)
+				if((ring_reader + 1) > buffer_size && !isnull(buffer[1]) )
+					ring_reader = 1
+				else if(bufl >= (ring_reader + 1) && !isnull(buffer[(ring_reader + 1)]))
+					ring_reader++
+			ring_writer++
+			if(ring_writer > buffer_size)
+				ring_writer = 1
```
Sovexe, line 1955: We can simply this by using the modulus operator

```suggestion
			ring_writer = (ring_writer % buffer_size) + 1
```

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,195 @@
...
+		if(!in_interact_range(src, user) || !can_act(user) || isnull(inp))
+			return
+		inp = min(inp, 60)
+		inp = max(4, inp)
+		cooldown_time = inp
+		tooltip_rebuild = 1
+		boutput(user, "Set delay to [inp]")
+
+	proc/setBufferSize(obj/item/W as obj, mob/user as mob)
+		var/inp = tgui_input_number(user,"Set size of signal buffer","Buffer size", buffer_size,buffer_max_size,-1)
```
Sovexe, line 1913: The user shouldn't be able to provide an invalid buffer size.
```suggestion
		var/inp = tgui_input_number(user,"Set size of signal buffer","Buffer size", buffer_size,buffer_max_size,1)
```

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,195 @@
...
+
+		inp = round(inp)
+		inp = min(inp, buffer_max_size)
+		if(inp == 0)
+			boutput(user,"Buffer size can't be zero.")
+			return
+		if(inp < 0)
+			playsound(src.loc, 'sound/machines/buzz-sigh.ogg', 50,1)
+			boutput(user,"How could a buffer be negative? What are you doing here?")
+			return
```
Sovexe, line 1925: I think this may be redundant since the tgui input does validation anyways but, this can just be written to use a clamp statement.
```suggestion
		if (isnull(inp)) return
		inp = round(inp)
		inp = clamp(inp, 1, buffer_max_size)
```

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -1820,6 +1820,195 @@
...
+			model = "random"
+			buffer_desc = "This mode outputs signals randomly, randomly overwriting previous signals when full."
+			buffer_model = RANDOM_BUFFER
+		else
+			return FALSE
+
+		buffer_string = model
+		tooltip_rebuild = 1
+		qdel(buffer)
+		buffer = list()
```
Sovexe, line 1885: Much faster way to empty a list than destruction and recreation. We should also probably reset our reader/write indexes
```suggestion
		buffer.len = 0
		ring_reader = 1
		ring_writer = 1
```

patricia-izin:
what happened here

Sovexe (changes requested):

patricia-izin:
@Sovexe merged all those changes thanks :)

Sovexe (approved):

patricia-izin:
I don't think i'll do the switch statement since we don't evaluate for buffer and it's kinda terse

patricia-izin:
@Sovexe so....the item loop was too slow....I put it on the machine loop but

1) This is a hack
2) I might be still redundantly subscribed to the item loop (doesn't matter because of the cooldown?)

patricia-izin:
it's running too slow because of the item loop (actually it's using a horrible hack right now to use the machine loop), gonna work on it

patricia-izin:
WRONG BUTTON

patricia-izin:
@Sovexe alright we now have a mechanics loop, the buffering component works how it's supposed to, and I've checked the most obvious process()ing components, e-gun and control unit, both still work the same (though up to 0.1s faster)

code/modules/mechanics/MechanicProcess.dm
```diff
@@ -0,0 +1,44 @@
+/// handles obj/item/mechanics/process()
+/datum/controller/process/mechanics
+	var/tmp/list/detailed_count
+	var/tmp/tick_counter
+	var/tmp/processing_mechanics
```
Sovexe, line 5: None of these vars appear to actually do anything
```suggestion
```

code/modules/mechanics/MechanicProcess.dm
```diff
@@ -0,0 +1,44 @@
...
+		logTheThing(LOG_DEBUG, src, "Mechanics process intialized")
+
+		detailed_count = new
+
+		src.processing_mechanics = global.processing_mechanics
+
+	copyStateFrom(datum/controller/process/target)
+		var/datum/controller/process/mechanics/old_items = target
+		src.processing_mechanics = old_items.processing_mechanics
+		src.detailed_count = old_items.detailed_count
```
Sovexe, line 21: Vars don't do anything useful and just waste memory duplicating a global list that we then ignore. copyStateFrom just perpetuates an unneeded and a blank list
```suggestion
```

code/modules/mechanics/MechanicProcess.dm
```diff
@@ -0,0 +1,44 @@
...
+
+	doWork()
+		var/c
+		for(var/obj/item/mechanics/target in global.processing_mechanics)
+			if (!target || target:disposed || target:qdeled) //if the object was pooled or qdeled we have to remove it from this list... otherwise the lagchecks cause this loop to hold refs and block GC!!!
```
Sovexe, line 26: we have a macro to clean this up
```suggestion
			if (QDELETED(target)) //if the object was pooled or qdeled we have to remove it from this list... otherwise the lagchecks cause this loop to hold refs and block GC!!!
```

code/modules/mechanics/MechanicProcess.dm
```diff
@@ -0,0 +1,44 @@
...
+
+	tickDetail()
+		if (length(detailed_count))
+			var/stats = "<b>[name] ticks:</b><br>"
+			var/count
+			for (var/thing in detailed_count)
+				count = detailed_count[thing]
+				if (count > 4)
+					stats += "[thing] used [count] ticks.<br>"
+			boutput(usr, "<br>[stats]")
```
Sovexe, line 44: detailed_count is never anything but an empty list, this does nothing
```suggestion
```

code/modules/mechanics/MechanicProcess.dm
```diff
@@ -0,0 +1,44 @@
...
+	var/tmp/list/detailed_count
+	var/tmp/tick_counter
+	var/tmp/processing_mechanics
+
+	setup()
+		name = "Mechanics"
+
+		schedule_interval = 0.4 SECONDS
+
+		logTheThing(LOG_DEBUG, src, "Mechanics process intialized")
```
Sovexe, line 12: shouldn't needed as process starts are already recorded in the scheduler
```suggestion
```

Sovexe (changes requested):

patricia-izin:
@Sovexe got all of those commited and re-tested should be ready for a test merge

code/modules/mechanics/MechanicProcess.dm
```diff
@@ -0,0 +1,19 @@
...
+
+	doWork()
+		var/c
+		for(var/obj/item/mechanics/target in global.processing_mechanics)
+			if (QDELETED(target))
+				global.processing_mechanics -= target
+				continue
+			if(target.process_fast == TRUE)
+				target.process()
+			else if(!(src.ticks % 7)) //Target schedule time is 2.8 seconds (was 2.9)
```
Sovexe, line 15: The 7 is kind of magic numbery, better to use vars and devise it pragmatically based on a target interval which will be easier to maintain should someone want to change it 

```suggestion
/datum/controller/process/mechanics
	/// The desired schedule time for standard devices
	var/standard_device_schedule_interval = 2.8 SECONDS
	/// Rounded number of ticks we must delay to sync our standard speed devices with the main loop
	var/standard_device_tick_delay

	setup()
		name = "Mechanics"
		schedule_interval = 0.4 SECONDS
		standard_device_tick_delay = max(round(standard_device_schedule_interval / schedule_interval), 1)

	doWork()
		var/c
		for(var/obj/item/mechanics/target in global.processing_mechanics)
			if (QDELETED(target))
				global.processing_mechanics -= target
				continue
			if(target.process_fast == TRUE)
				target.process()
			else if(!(src.ticks % standard_device_tick_delay))
```

Sovexe (changes requested):

code/modules/mechanics/MechanicProcess.dm
patricia-izin: Committed make sense to me, I didn't think it would be so easy to specify it in time

patricia-izin:
This is working great on live and nothing seems to have gone wrong!

patricia-izin:
woo!

PR #21549, opened 2024-12-04
Mail vendor

cheekybrdy:
[FEATURE] [DNM] [WIP]
## About the PR 
Intended to be merged after #21368

Adds a mail vendor to be added to mail sorting rooms (totally not just a resprited angler vendor). The tokens to make purchases would be dropped from mail litter which now drops regardless of opening method, however only legit openings create the type that allows for a token to be collected. The more unique items haven't been made for this yet but I hope this PR in addition to #21368 can serve as a backbone for a reward system to Courier's gameplay loop.

Paper crumple sound used for searching through mail mess for a token is Creative Commons from Freesound. First time merging sounds so please lmk if I'm missing something else.
crumple-06.ogg by drooler -- https://freesound.org/s/508597/ -- License: Creative Commons 0

If this gets merged, I'm willing go give 250k spacebux to anyone who makes a item for this vendor.
## Why's this needed? 
Mail Courier works with a proper job gameplay loop, but lacks anything uniquely rewarding from it, this can work as a medium for that.

No changelog as no additions to maps.


TobleroneSwordfish:
I said this in discord but I'll copy it here: I'm not a fan of adding the very video gamey magic vendor + tokens thing to more jobs. It kind of works for angler because you're just sitting around fishing, but feels very out of theme for a job like mail courier. Janitors don't need "cleaning tokens" and a progression tree to make them want to mop spills and the lack of that is part of what makes this game a unique sandbox rather than an MMO style progression sim.

PR #21568, opened 2024-12-05
Add AI access disablers to trenchwarfare.dmm

JORJ949:
## About the PR 
Adds AI control disablers to most doors around Caer Vedwyd

## Why's this needed? 
These doors have syndicate access, normal crew cant just run through so neither should the AI/borgs, the AI/borgs should especially not just be able to access the armory and other weaponry stored here.


frawhst (approved):
Not going to merge since this is a Cogwerks Thing, but makes sense to me.

TobleroneSwordfish:
I'm a merge tyrant

PR #21569, opened 2024-12-05
Add staff bonuses to bank computer

patricia-izin:
## About the PR 

Adds a button to the bank computer to pay a bonus to the entire staff from the payroll budget. It creates an announcement when used, and is restricted to only being available every five minutes.

image

image

The clown is taxed at 50% for their pleasure.

## Why's this needed? 

The station can end up with a lot of money and generally it would be nice to be able to issue a bonus to the crew for good performance or comedy

## Changelog 

```changelog
(u)patricia
(*)The HoP bank computer can now issue bonuses to department staff, or stationwide!

```

### DIFF
--- a/code/obj/machinery/computer/bank_records.dm
+++ b/code/obj/machinery/computer/bank_records.dm
@@ -271,36 +271,36 @@
 						boutput(usr, SPAN_ALERT("NT Regulations forbid issuing multiple staff incentives within five minutes."))
 						return
 
-					var/amount = input(usr, "How many credits should we issue to each staff member?", "Issue Bonus", 100) as null|num
-					if(isnull(amount)) return
+					var/bonus = input(usr, "How many credits should we issue to each staff member?", "Issue Bonus", 100) as null|num
+					if(isnull(bonus)) return
 
-					amount = ceil(clamp(amount, 1, 999999))
-					var/bonus_amt = (length(data_core.bank.records) * amount)
-					if ( bonus_amt > wagesystem.station_budget)
+					bonus = ceil(clamp(bonus, 1, 999999))
+					var/bonus_total = (length(data_core.bank.records) * bonus)
+					if ( bonus_total > wagesystem.station_budget)
 						//Let the user know the budget is too small before they set the reason if we can
-						boutput(usr, SPAN_ALERT("Total bonus cost would be [bonus_amt][CREDIT_SIGN], payroll budget is only [wagesystem.station_budget][CREDIT_SIGN]!"))
+						boutput(usr, SPAN_ALERT("Total bonus cost would be [bonus_total][CREDIT_SIGN], payroll budget is only [wagesystem.station_budget][CREDIT_SIGN]!"))
 						return
 					var/message = input(usr, "What is the reason for this staff bonus?", "Bonus Reason") as text
 					if(isnull(message) || message == "")
 						boutput(usr, SPAN_ALERT("NT Regulations require that the reason for issuing a staff bonus be recorded."))
 						return
 
-					if(isnull(amount) || bonus_amt > wagesystem.station_budget)
+					if(isnull(bonus)|| isnull(bonus_total) || bonus_total > wagesystem.station_budget)
 						//Something ain't right but it could be a coincidence
 						//Maybe someone stole the budget under our feet, or payroll was issued
-						boutput(usr, SPAN_ALERT("Total bonus cost would be [bonus_amt][CREDIT_SIGN], payroll budget is only [wagesystem.station_budget][CREDIT_SIGN]!"))
+						boutput(usr, SPAN_ALERT("Total bonus cost would be [bonus_total][CREDIT_SIGN], payroll budget is only [wagesystem.station_budget][CREDIT_SIGN]!"))
 						return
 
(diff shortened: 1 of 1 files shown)


patricia-izin:
I thought about adding the ability to give bonuses to individual departments (it was suggested) but scope creep is as scope creep does

code/obj/machinery/computer/bank_records.dm
```diff
@@ -264,6 +266,35 @@
...
+						boutput(usr, SPAN_ALERT("NT Regulations forbid issuing multiple staff incentives within five minutes."))
+						return
+
+					var/amount = input(usr, "How many credits should we issue to each staff member?", "Issue Bonus", 100) as null|num
+					if(isnull(amount)) return
+
+					amount = ceil(clamp(amount, 1, 999999))
+					var/bonus_amt = (length(data_core.bank.records) * amount)
+					if ( bonus_amt > wagesystem.station_budget)
+						boutput(usr, SPAN_ALERT("Total bonus cost would be [bonus_amt][CREDIT_SIGN], payroll budget is only [wagesystem.station_budget][CREDIT_SIGN]!"))
```
ZeWaka, line 280: This needs to be checked right before spending it. Right now someone could open the UI past this, to the message section, and drain all the accounts, then have negative money

ZeWaka (changes requested):

patricia-izin:
@ZeWaka Alright added some checks and even an admin log for one exceptionally weird condition, also made the variables more understandable

code/obj/machinery/computer/bank_records.dm
```diff
@@ -264,6 +266,47 @@
...
+						//No you really shouldn't be here
+						logTheThing(LOG_ADMIN, usr, "may have attempted href hacking Bank Computer. Do not take this as a guarantee.")
+						return
+					if(bonus_total > wagesystem.station_budget)
+						boutput(usr, SPAN_ALERT("Total bonus cost would be [bonus_total][CREDIT_SIGN], payroll budget is only [wagesystem.station_budget][CREDIT_SIGN]!"))
+						return
+
+					logTheThing(LOG_STATION, usr, "issued a bonus of [bonus][CREDIT_SIGN] ([bonus_total][CREDIT_SIGN] total) to the staff.")
+					src.bonus_rate_limit_time = world.time + (5 MINUTES)
+					command_announcement("[message]<br>Bonus of [bonus][CREDIT_SIGN] issued to all staff.", "Payroll Annoucment by [scan.registered] ([scan.assignment])")
```
JORJ949, line 301:
```suggestion
					command_announcement("[message]<br>Bonus of [bonus][CREDIT_SIGN] issued to all staff.", "Payroll Announcement by [scan.registered] ([scan.assignment])")
```

code/obj/machinery/computer/bank_records.dm
```diff
@@ -264,6 +266,47 @@
...
+					wagesystem.station_budget = wagesystem.station_budget - bonus_total
+					for(var/datum/db_record/R as anything in data_core.bank.records)
+						if(R["job"] == "Clown")
+							//Tax the clown
+							R["current_money"] = (R["current_money"] + ceil((bonus / 2)))
+							return
```
JORJ949, line 307: This will return the whole function and so stop running through all the records the moment it finds the clown, use an else instead

patricia-izin: oh dang! thanks for the catch!

patricia-izin:
@JORJ949 that should do those ty again

patricia-izin:
> https://github.com/patricia-izin/goonstation/pull/1

Merged but Id really love if you could still give stationwide bonuses

TobleroneSwordfish:
A general station bonus option sounds good, I didn't include heads as they already get paid very well and excluding them makes it feel more like it's a bonus from command to the rest of the crew? I dunno, it's fine either way.

patricia-izin:
@TobleroneSwordfish and should be good, tested it and was happy, added heads to departments and tweaked some text and added stationwide bonuses

code/obj/machinery/computer/bank_records.dm
```diff
@@ -13,6 +13,25 @@
 	var/printing = null
 	var/can_change_id = 0
 	var/payroll_rate_limit_time = 0 //for preventing coammand message spam
+	var/static/bonus_rate_limit_time = 0 //prevent bonus spam because these have an annoucement
+	///I know we already have department job lists but they suck and are brittle and way too general so I made my own here
+	var/static/list/departments = list(
+		"Stationwide" = list(/datum/job/), //hacky, yet generic
```
TobleroneSwordfish, line 19: Oough this will iterate every job datum in the codebase for every player on the server. It's fine when it's more restricted but I'd rather this was specialcased to skip the loop instead.

TobleroneSwordfish (changes requested):

code/obj/machinery/computer/bank_records.dm
patricia-izin: Oh sure i'll just catch it

patricia-izin: Oh sure i'll just catch it

patricia-izin:
@TobleroneSwordfish there I just made it an empty list and catch it in the loop

PR #21572, opened 2024-12-06
Give changelings the option to transfer straight into the new body on absorb

TobleroneSwordfish:
## About the PR 
Adds a second "absorb" button called "Transfer DNA" that instead transfers the changeling into the new body and husks the old one. You still get the DNA for absorbing the person. The standard absorb button remains unchanged.

## Why's this needed? 
Changelings have a lot of potential for stealth and trickery but it's often hampered by the tedious process of trying to disguise yourself as your victim. Currently you have to spend several minutes stripping their body, sorting things out and making sure you don't have the wrong headset or shoes on. This also leads to a lot of newer players trying this once, getting caught because they forgot a detail and subsequently not bothering to swap identities at all.

Possible issues: maaybe making them too stealthy and hard to track?

## Changelog 

```changelog
(u)LeahTheTech
(*)Changelings now have the ability to swap bodies with those they absorb, leaving their old body as a husk. The old absorb behaviour remains as a different button.
```


Xkallubar:
Make this something you can toggle on and off.

JORJ949:
> Make this something you can toggle on and off.

Very much this, sometimes you have a good disguise you dont want to swap out of because you have good loot (e.g. cap eating a staffie)

code/modules/antagonists/changeling/abilities/absorb.dm
```diff
@@ -177,24 +177,32 @@
...
-			target.dna_to_absorb = 0
-			target.death(FALSE)
-			target.disfigured = TRUE
-			target.UpdateName()
-			target.bioHolder.AddEffect("husk")
-			target.bioHolder.mobAppearance.flavor_text = "A desiccated husk."
-
-			if (ishuman(ownerMob))
-				var/mob/living/carbon/human/H = ownerMob
+			target.full_heal()
```
frawhst, line 188: This is a *little* weird (changeling missing liver eats other person missing liver, liver appears post-absorb) but I think it's fine in the interests of cleanliness/avoiding weird edge cases. Especially considering that changelings can full heal themselves on-demand anyways.

frawhst (approved):
Yeah, I like this. Agree with others that a toggle would be nice, but I'm not sure we have a framework for toggleable statuses on abilities, and I'd rather not add Yet Another Button to the changeling UI.

code/modules/antagonists/changeling/abilities/absorb.dm
TobleroneSwordfish: Yeah I considered trying to make it copy damage from the old body but that feels like a maze of edge cases and this is cleaner.

Annmagedon:
I noticed a problem, if you are for say person A, and take over person B, and turn back to A...
you keep person B traits. Example you loose hairy trait so if A is mutarace you become BALD, if  B was klepto, you stay kelpto as A. it kinda forces you to always stay as latest victim....

TobleroneSwordfish:
Trait issue fixed in 80d19eb

ThatFiveGuys:
> Yeah, I like this. Agree with others that a toggle would be nice, but I'm not sure we have a framework for toggleable statuses on abilities, and I'd rather not add Yet Another Button to the changeling UI.

Honestly a lot of changeling abilities could be merged. "Grant control" and "retake control" could be made into the same ability by just being able to use grant control on yourself. Also even easier all of the limb critters could be made into a dropdown box the same way that arms are one. Feel like it's fair to add one more button if you get rid of 3/4 of them.

Annmagedon:
hey might be an oversight but if a ling eat a wizard / arcfiend / antag with abilities

they gain the abilities since they *gain* the body. this is like, way too powerful for an already super strong antag with the easiest guaranteed killing / roundremove abilities.

and it probably sucks to get targeted by a ling for being lucky and rolling wizard

TobleroneSwordfish:
THIS TIME FOR REAL

PR #21582, opened 2024-12-06
Nuke internal record that plays upon arming.

JORJ949:
## About the PR 
Allows nuclear bombs to have records stored in them to play upon arming

https://github.com/user-attachments/assets/6a5dd4c4-380d-4d19-bbd5-a7dfa823a9c8

## Why's this needed? 
The portable record player is cool and all, but this can save inventory slots and does the timing for you.

## Changelog 

```changelog
(u)JORJ949
(*)Nuclear Bombs can now have records loaded into them that play upon arming.
```

### DIFF
--- a/code/obj/machinery/nuclearbomb.dm
+++ b/code/obj/machinery/nuclearbomb.dm
@@ -21,7 +21,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/nuclearbomb, proc/arm, proc/set_time_left)
 	var/datum/hud/nukewires/wirepanel
 	var/obj/item/disk/data/floppy/read_only/authentication/disk = null
 	var/obj/item/record/record = null
-	var/record_locked = FALSE // Varedit to TRUE if you dont want admin gimmick music removed
+	var/record_locked = FALSE // TRUE if the internal record cannot be removed, intended to be varedited by admins for gimmicks.
 
 	var/target_override = null // varedit to an area TYPE to allow the nuke to be deployed in that area instead of whatever the mode says (also enables the bomb in non-nuke gamemodes)
 	var/target_override_name = "" // how the area gets displayed if you try to deploy the nuke in a wrong area
--- a/code/obj/machinery/nuclearbomb.dm
+++ b/code/obj/machinery/nuclearbomb.dm
@@ -20,6 +20,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/nuclearbomb, proc/arm, proc/set_time_left)
 	var/debugmode = 0
 	var/datum/hud/nukewires/wirepanel
 	var/obj/item/disk/data/floppy/read_only/authentication/disk = null
+	var/obj/item/record/record = null
 
 	var/target_override = null // varedit to an area TYPE to allow the nuke to be deployed in that area instead of whatever the mode says (also enables the bomb in non-nuke gamemodes)
 	var/target_override_name = "" // how the area gets displayed if you try to deploy the nuke in a wrong area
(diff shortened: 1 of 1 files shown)


code/obj/machinery/nuclearbomb.dm
```diff
@@ -215,12 +216,40 @@ ADMIN_INTERACT_PROCS(/obj/machinery/nuclearbomb, proc/arm, proc/set_time_left)
 		command_alert("\A [src] has been armed in [isturf(src.loc) ? get_area(src) : src.loc]. It will detonate in [src.get_countdown_timer()] minutes. All personnel must report to [get_area(src)] to disarm the bomb immediately.", "Nuclear Weapon Detected")
 		if (!ON_COOLDOWN(global, "nuke_planted", 20 SECONDS))
 			playsound_global(world, 'sound/machines/bomb_planted.ogg', 75)
+		if(src.record)
+			if(is_music_playing())
+				src.visible_message(SPAN_NOTICE("The [src]'s record player detects conflicting music and ejects the record!"))
+				src.record.set_loc(get_turf(src))
+				src.record = null
+			else
+				SPAWN(6 SECONDS) // Length of the "Bomb planted" sound effect
```
frawhst, line 225: The block under this needs to be indented for this to actually pause anything- otherwise it's just going to spawn a new 'thread' that waits 6 seconds to do nothing

frawhst (commented):
I like this, good for admin gimmicks especially

code/obj/machinery/nuclearbomb.dm
JORJ949: It seemed to work in testing, decreasing this number from 8>6 did make it match better with the nuke planted sound effect, it works as intended where the rest ran and the ghost message activated, but it waited to play the music.

JORJ949: Increasing it to 20 also changed it, its working as intended where it only pauses playing the music and not the log or ghost message.

code/obj/machinery/nuclearbomb.dm
```diff
@@ -20,6 +20,8 @@ ADMIN_INTERACT_PROCS(/obj/machinery/nuclearbomb, proc/arm, proc/set_time_left)
 	var/debugmode = 0
 	var/datum/hud/nukewires/wirepanel
 	var/obj/item/disk/data/floppy/read_only/authentication/disk = null
+	var/obj/item/record/record = null
+	var/record_locked = FALSE // Varedit to TRUE if you dont want admin gimmick music removed
```
ZeWaka, line 24: this comment is a bit unclear

code/obj/machinery/nuclearbomb.dm
```diff
@@ -256,6 +288,17 @@ ADMIN_INTERACT_PROCS(/obj/machinery/nuclearbomb, proc/arm, proc/set_time_left)
...
+		if (istype(W, /obj/item/record))
+			if(src.record)
+				boutput(user, SPAN_ALERT("The [src.name] already has a record inserted!"))
+				return
+			boutput(user, "You insert the record into the record player.")
+			src.visible_message(SPAN_NOTICE("<b>[user] inserts the record into the record player.</b>"))
+			user.drop_item()
+			W.set_loc(src)
+			src.record = W
+			return
```
ZeWaka, line 300:
```suggestion
```
useless

JORJ949: How so? this stops you attacking it with the record after you insert it.

ZeWaka: Never mind I shouldn't late night review

ZeWaka: Never mind I shouldn't late night review

PR #21601, opened 2024-12-07
Give all gang members guns, cap gangs to 2 on RP

TobleroneSwordfish:
## About the PR 
Gives all gang members either a lopoint or a lasergat at roundstart (gang leaders also get a spare magazine) and caps the maximum number of gangs to 2 on the roleplay servers.

## Why's this needed? 
Gang members have the eternal problem of being too weak at the start of the round and too strong once they all get guns. This especially causes problems with their interactions with security, who often feel like they have to "wait" for the gang members to get gear before they can push back against them.
Giving them guns right off the bat makes them a lot scarier at the start and hopefully a lot more fun to play. The max gang reduction should help curb the chaos and make things easier to keep track of on high population RP rounds.

## Changelog 

```changelog
(u)LeahTheTech
(*)Every gang member now gets one FREE GUN from their locker.
(+)There will now only ever be two gangs on RP servers.
```

### DIFF
--- a/code/datums/gamemodes/gangwar.dm
+++ b/code/datums/gamemodes/gangwar.dm
@@ -270,6 +270,8 @@ proc/broadcast_to_all_gangs(var/message)
 	var/list/tags = list()
 	/// The minds of members of this gang who are currently on cooldown from redeeming their gear from the gang locker.
 	var/list/gear_cooldown = list()
+	/// List of antag datums who have obtained their free gun from the locker so far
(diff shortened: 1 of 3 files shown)


JORJ949:
I really do not think free guns on you when you spawn will go down well on classic, I'd rather it be when you redeem your gang gear from the locker.

TobleroneSwordfish:
Hmm good point

PR #21605, opened 2024-12-07
Resprites the espresso machine

rubicubie:
[GAME OBJECTS]

## About the PR 

Changes the old espresso machine and the espresso mugs.

## Why's this needed? 

The old sprites were not as fit with the game art style

```changelog
(u)RubiCubie
(+)Resprited the espresso machine and its mug.
```


Tarmunora (changes requested):
`coffeetopoverlay` state appears missing

rubicubie:
I'm unsure what that overlay even is, so I've left it on the bottom part of the machine

Tarmunora:
> I'm unsure what that overlay even is, so I've left it on the bottom part of the machine

Looking into it further, it looks like the code responsible for handling that overlay is deprecated - it can be safely deleted, yeah

PR #21610, opened 2024-12-08
Disorient immunity at high STA

TDHooligan:
[BALANCE][SECURITY]
## About the PR 
Disorient is no longer applied at and above 50% stamina.

Batons will still disorient most targets, as will good flashbang/taser hits.

Probably a little controversial, but i will save the thread until comments/discussion pop off

## Why's this needed? 
At the minute, even tiny disorients turn the lightest, briefest skirmishes into a fight for your life, no matter how prepared you are. This should alleviate the worst edge cases where a single lucky taser SMG plink can turn a well-timed, successful escape into losing your traitor round. 

In the future, I reckon a stamina penalty over time for all tasers ('Shocked'?) may be useful, but given we've had a stale stun-heavy metagame for some time, I'm erring on the side of 'its probably fine right now'

## Changelog 

```changelog
(u)TDHooligan
(*)Disorient is now only applied when you are below 50% stamina.
```


TobleroneSwordfish:
Generally not a fan of nerfing tasers any more honestly, I know this is targeted at big efficient RP security teams but I think it will disproportionally affect low-security shifts or egun captains trying to catch the person who just broke into their office. Note that this PR as written also nerfs flashes into the ground as a side effect.

I do think the taser SMG might be a little strong compared to the other options right now, but that seems like it should be fixed directly rather than blanket nerfing a core game mechanic.

TDHooligan:
that's fair. i'm still pretty certain Disorient is the issue. perhaps a different criteria for immunity makes sense?

TDHooligan:
closed pending more investigation on other stun tooling

PR #21612, opened 2024-12-08
Adds the CivALLised Service Tool for HoPs

cheekybrdy:
## About the PR 
Adds a omnitool for the HoP, which has various tools from the jobs in the civilian department.

## Why's this needed? 
HoP lacks some unique loot, both on person and locker. This amends that but also gives them a more explicit connection to their actual home department.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Added the CivALLised Service Tool, a omnitool with various equipment from the civilian department. Now in the HoP's locker!
```


Tarmunora:
I can't really see this to be honest. HoP is, by design, less involved in the day-to-day of civilian jobs than most of the other heads are for their respective depts. They're a desk jockey, not a tradesperson.

Additionally, this is a super odd set of tools that I think would be extremely rarely relevant.

JORJ949:
While im all for the HoP being more encouraged to interact with service, this feels like an oddly forced way to do it.

Glamurio:
> As neat as this is, I don't think the HoP, of all crewmembers, needs this. Have you ever seen a HoP cooking?

I've seen just about any job cook, but that's less of a statement on the HoP.

PR #21624, opened 2024-12-08
Buff AI Disguised law module

JORJ949:
[SILICONS][FEATURE]
## About the PR 
The AI disguised module now comes with a remote
Hit the module on the remote (or vice versa) to link the module to the remote, remotes can have multiple linked modules
When hit with another module the disguised module hides itself as that law module, copying its law, color and name.
When the remote is used the module immediately activates the hidden law
The disguised module can be used on the remove button on a law rack to hotswap it with the inserted law, disguising itself as that law and not pushing law updates to connected silicons (as the law will be the exact same until the remote is used). This can be done even if the current law is welded and screwed and the rack is depowered.

## Why's this needed? 
The disguised law is incredibly underused and this has a unique use of allowing you to create "sleeper agent" silicons that turn evil at the push of a button.

## Changelog 

```changelog
(u)JORJ949
(*)Disguised AI modules have received a large functionality change! Details can be found on the PR.
```


TobleroneSwordfish:
I like what this is trying to achieve but I think the way the slot number changes the behaviour is too confusing. Kind of radical suggestion buut: maybe it should just be a separate item that attaches to a law module (and can be added while it's still in the rack?) which then changes the content of that module later on a trigger.

JORJ949:
Interesting idea but I want there to be some evidence to destroy when you do this, for example earlier today someone saw the first law hit their pod (but then we came up with an excuse for it because the law 1 replacement was "You may injure" and thet thought it was the same). I think something in chat like "as you insert the module into the rack the built in stealth circuit activates" would make it clearer what happened, and good use of this item already requires a fair understanding of ai laws and how they work

TobleroneSwordfish:
Ookayy, what about letting it be swapped with other law modules in place and still only causing an update when triggered? I really really don't like it changing functionality based on what slot it's in, there shouldn't be anything special about the first 3 slots.

JORJ949:
Added your hotswap suggestion, although not as cleanly as I would like as I'm not a TGUI person so don't want to touch it, would love to move the function to be on the remove button without editing present lawrack functionality of the button being greyed out when it can't be removed.

TobleroneSwordfish (changes requested):
Sorry, didn't realise you re-opened this!
- The traitor buy menu should explain the new functionality.
- The remote should spawn in your hand if possible at the time.
- The remote shouldn't be obviously called "Disguised AI law module remote"
- Swapping the modules shouldn't trigger a law update, otherwise it's very obvious to all silicons that they just got a disguised law whenever they hear "dooDING" and don't get any law changes
- Clicking the "Remove" button with the law module in hand should also trigger the swap action, currently it just removes the existing module.
- There should be some feedback when you press the remote and a law change actually takes place.
- The "Remove" button should become active on all modules while you're holding the disguised module in your active hand, if you don't know how to do this feel free to ping me on discord.

JORJ949:
Suggestions implemented, I didn't implement remove button changing if you are holding a disguised module because 1) I don't know how and 2) you aren't supposed to know if you are holding a disguised module and the button changing colour would out it.

JORJ949:
Back in the oven

PR #21625, opened 2024-12-08
No UI Solar Controller

Garash2k:
## About the PR 
Makes it so that the Solar Controller toggles Off or Auto after being clicked.
Prunes timed tracking mode, and manual angle setting code.
Makes them emaggable, during which they'll track the opposite direction of the sun instead!
State, generated power, current orientation and sun orientation can now be read in the description.

Enabled:
image
image

Disabled:
image
image

Emagged:
image
image

## Why's this needed? 
Ease of use
From what I've read, timed mode was a legacy thing before auto got implemented?
Besides grief, I don't see much use for manually setting their angle or fiddling with the tracking rate.
Turning them off or emagging them offers plenty grief opportunities.

Pruned UI:
261899265-34e7c4c2-c648-4e3a-bb5b-0bbadd2de3b3

## Changelog 

```changelog
(u)Garash2k
(+)The solar panel controller is now easier to mess with, it can now only be turned on or off... Or emagged! Examine text has all the info it's old UI used to have.
```
[UI][Feature][QoL]

### DIFF
--- a/code/modules/power/solar.dm
+++ b/code/modules/power/solar.dm
@@ -340,7 +340,7 @@ TYPEINFO(/obj/machinery/power/solar)
 	if (src.is_active_and_powered())
 		src.tracker_update(tracker.sun_angle)
 
-	src.update_solar_icon()
+	src.UpdateIcon()
 
 /obj/machinery/computer/solar_control/get_desc()
 	. = "<br />It is currently <em>[src.is_active_and_powered() ? "tracking the sun" : "disabled"]</em>"
@@ -351,17 +351,17 @@ TYPEINFO(/obj/machinery/power/solar)
 /obj/machinery/computer/solar_control/proc/is_active_and_powered()
 	. = active && !(status & (NOPOWER | BROKEN))
 
-/obj/machinery/computer/solar_control/proc/update_solar_icon()
+/obj/machinery/computer/solar_control/update_icon()
+	. = ..()
 	if (src.is_active_and_powered())
 		src.icon_state = initial(src.icon_state)
 	else
 		src.icon_state = "solar0"
-	src.UpdateIcon()
 
 /obj/machinery/computer/solar_control/power_change()
 	..()
 	if (!(src.status & (NOPOWER | BROKEN)))
-		src.update_solar_icon()
+		src.UpdateIcon()
 
 /obj/machinery/computer/solar_control/emag_act(var/mob/user)
 	. = ..()
--- a/code/modules/power/solar.dm
+++ b/code/modules/power/solar.dm
@@ -257,11 +257,9 @@ TYPEINFO(/obj/machinery/power/solar)
 	var/cdir = 0
 	var/gen = 0
 	var/lastgen = 0
-	var/track = 2			// 0= off  1=timed  2=auto (tracker)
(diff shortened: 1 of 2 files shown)


TealSeer (commented):
Turning equipment power off on the area APC makes the screen go dark but examine text still claims its tracking the sun, similarly toggling tracking off and then switching area equipment power off and then back on makes it switch to its "on" sprite (with the yellow screen) despite it still being disabled, looks good otherwise

Garash2k:
Good catch! Fixed

code/modules/power/solar.dm
```diff
@@ -318,95 +316,62 @@ TYPEINFO(/obj/machinery/power/solar)
...
-	t += "<A href='?src=\ref[src];close=1'>Close</A></TT>"
-	user.Browse(t, "window=solcon")
-	onclose(user, "solcon")
-	return
+/obj/machinery/computer/solar_control/proc/update_solar_icon()
+	if (src.is_active_and_powered())
+		src.icon_state = initial(src.icon_state)
+	else
+		src.icon_state = "solar0"
+	src.UpdateIcon()
```
TealSeer, line 359:
```suggestion
/obj/machinery/computer/solar_control/update_icon()
	. = ..()
	if (src.is_active_and_powered())
		src.icon_state = initial(src.icon_state)
	else
		src.icon_state = "solar0"
```
We can just roll this into an `update_icon` override instead of needing a new proc

code/modules/power/solar.dm
```diff
@@ -318,95 +316,62 @@ TYPEINFO(/obj/machinery/power/solar)
...
-		usr.Browse(null, "window=solcon")
-		src.remove_dialog(usr)
-		return
-	if(href_list["close"] )
-		usr.Browse(null, "window=solcon")
-		src.remove_dialog(usr)
+/obj/machinery/computer/solar_control/power_change()
+	..()
+	if (!(src.status & (NOPOWER | BROKEN)))
+		src.update_solar_icon()
```
TealSeer, line 364:
```suggestion
		src.UpdateIcon()
```
ditto

code/modules/power/solar.dm
```diff
@@ -318,95 +316,62 @@ TYPEINFO(/obj/machinery/power/solar)
...
-	var/t = "<TT><B>XIANG|GIESEL Photo-Electric Generator Control</B><HR><PRE>"
-	t += "Generated power : [round(lastgen)] W<BR><BR>"
-	t += "<B>Orientation</B>: [rate_control(src,"cdir","[cdir]&deg",1,15)] ([angle2text(cdir)])<BR><BR><BR>"
+	active = !active
+	user.show_text("You [active ? "activate" : "deactivate"] [src]'s solar tracking.", "blue")
+	if (src.is_active_and_powered())
+		src.tracker_update(tracker.sun_angle)
 
-	t += "<BR><HR><BR><BR>"
+	src.update_solar_icon()
```
TealSeer, line 343:
```suggestion
	src.UpdateIcon()
```
ditto

TealSeer (changes requested):

ZeWaka (approved):

PR #21632, opened 2024-12-09
Add round-end buttons to re-open inspector's report and show/hide gang victory ui

pgmzeta:
[ui][gamemodes][feature]
## About the PR 
Add ability buttons to re-open the inspectors report and show/hide the gang victory hud at round end. These buttons only appear when needed i.e. there is an inspector's report / the mode is gangwar.

To support this, the gangwar victory hud is now a global.

## Why's this needed? 
Make it easier to (re)open the inspector report post-round. The gang victory HUD makes it harder to see the end-round art/canvas at centcom.

## Screenshot
Screenshot 2024-12-09 113219
Left to right: Gang Victory HUD toggle, Show Crew Credits (existing button), Show Inspector Report


code/datums/gamemodes/gangwar.dm
```diff
@@ -1,3 +1,5 @@
+var/global/datum/hud/gang_victory/victory_hud
```
ZeWaka, line 1: why not use the singleton system instead? less pollution

pgmzeta: used a singleton, had to modify the victory hud so it didn't expect the winner in new but instead required a post-creation proc call

ZeWaka (approved):

ZeWaka:
X

PR #21638, opened 2024-12-10
Balance: Hall Monitors get a pocket radio for eavesdropping on security.

pgmzeta:
[balance][game-objects]
## About the PR 
Replace the hall monitor's security-headset with a microphone-less pocket radio that can hear security's frequency.

## Why's this needed? 
Some hall monitors are acting like a bonus SecAss. This change allows them to continue eavesdropping on sec lines, but removes their ability to actively coordinate via radio comms.

## Changelog 

```changelog
(u)glowbold
(+)Hall monitors recieve a pocket radio that can only listen to security's frequency.
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -1199,9 +1199,10 @@ ABSTRACT_TYPE(/datum/job/civilian)
 	slot_belt = list(/obj/item/device/pda2)
 	slot_jump = list(/obj/item/clothing/under/color/red)
 	slot_foot = list(/obj/item/clothing/shoes/brown)
-	slot_ears = list(/obj/item/device/radio/headset/hall_monitor)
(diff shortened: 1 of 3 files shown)


flappybatpal:
Considering it's a gimmick job I'm not sure if they should really have any security comms at all.

klushy225:
I like it. I never intended the hall monitor to actually work for security. The point of the paper badge and crusty old ticket writer was to very clearly say that "THIS PERSON HAS NO AUTHORITY"
The reason it's associated with security at all is because I think it's funny. Like a wannabe security officer.

pgmzeta:
> Considering it's a gimmick job I'm not sure if they should really have any security comms at all.

right now they get a headset with sec frequencies; this is reducing their overall access

PR #21656, opened 2024-12-12
Mice have a chance to be shiny

JORJ949:
## About the PR 
Most mice have a 1/4096 of being a green "shiny space mouse" upon spawning
image

Exceptions:
- Remy
- Mice spawned from wraith dens
- Mentor mice
- Admin mice have a 1/1365 chance instead, the chance with shiny charm, due to their rarity.

Additional note I would love to add a little effect on spawning like seen in the below video, but couldn't think of any current sound effect that would fit. https://www.youtube.com/watch?v=116qpWKCKyA

## Why's this needed? 
Fun pop-culture reference to pokemon and how you can find a differently coloured one called a "shiny".


wolfolotl:
not a huge fan just because it disrupts the immediate visual clarity that mentor mice have. the point of them looking so distinct is so that people can immediately tell that a mentor's present and available to help, and occasionally taking away the telling purple detracts from that design purpose

the concept is fun, though. wouldn't mind this sort of thing being a feature for other mobs, just not ones that have an actual important function like mentor mice

JORJ949:
> not a huge fan just because it disrupts the immediate visual clarity that mentor mice have. the point of them looking so distinct is so that people can immediately tell that a mentor's present and available to help, and occasionally taking away the telling purple detracts from that design purpose
> 
> the concept is fun, though. wouldn't mind this sort of thing being a feature for other mobs, just not ones that have an actual important function like mentor mice

If they do need to urgently be purple they can always just respawn, but yeah I agree

Tarmunora:
> ... just not ones that have an actual important function like mentor mice

Agreed. Mentormouse is probably the worst choice for this kind of feature.

Valtsu0:
I don't think the reference would suffer too much if it was moved to admin mouse

Tarmunora (approved):

code/mob/living/critter/small_animal.dm
```diff
@@ -4363,14 +4377,20 @@ var/list/mob_bird_species = list("smallowl" = /mob/living/critter/small_animal/b
...
 
 	New()
 		..()
-		src.real_name = "[pick_string("mentor_mice_prefixes.txt", "mentor_mouse_prefix")] [src.name]"
+		if(src.is_shiny)
+			src.icon_state = "mouse-large-shiny"
+			src.icon_state_dead = "mouse-large-shiny-dead"
+			src.icon_state_exclaim = "mouse-large-shiny-exclaim"
+		else
+			src.real_name = "[pick_string("mentor_mice_prefixes.txt", "mentor_mouse_prefix")] [src.name]"
```
ZeWaka, line 4389: name?

JORJ949: Could you provide more detail as to the issue here? The "shiny" prefix is assigned in the parent proc and so doesn't need to be added here.

PR #21660, opened 2024-12-12
Silicons can now play large instruments (piano, jukebox)

AnomalousPotato:
[Player Actions][Silicons][Feature]

## About the PR 
Silicons such as cyborgs can now play large instruments such as the piano and jukebox.

## Why's this needed? 
Let the metal friends make music... they deserve it!

## Changelog 

```changelog
(u)AnomalousPotato
(+)Silicons can now play large instruments such as the piano.
```

### DIFF
--- a/code/modules/items/instruments/instruments.dm
+++ b/code/modules/items/instruments/instruments.dm
@@ -227,22 +227,6 @@
 		else
 			src.play(user)
 
-	attack_ai(mob/user as mob)
-		..()
-		if (!in_interact_range(src, user)) // Instruments are not wireless
-			return
-
-		if (isAI(user))
-			var/mob/living/silicon/ai/borgo = user
-			if (borgo.deployed_to_eyecam)
-				return
-
-		if(use_new_interface)
-			ui_interact(user)
-		else
-			src.play(user)
-
-
 /* -------------------- Large Instruments -------------------- */
 
 /obj/item/instrument/large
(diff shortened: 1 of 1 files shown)


code/modules/items/instruments/instruments.dm
```diff
@@ -227,6 +227,20 @@
...
 			src.play(user)
 
+	attack_ai(mob/user as mob)
+		..()
+		if (!in_interact_range(src, user)) // Instruments are not wireless
+			return
+
+		if (isAI(user))
+			var/mob/living/silicon/ai/borgo = user
+			if (borgo.deployed_to_eyecam)
```
JORJ949, line 237: You should check if the user is an ai eye instead because user will not be the mainframe if they are

Tarmunora:
Cyborgs don't really have the hands to play a fiddle, and decidedly do not have the lungs to say, play a trumpet.

AnomalousPotato:
That’s true, but who am I do deny them such? I thought about adding some visible message like “the cyborg extends a set of delicate pipes/ manipulators towards the instrument” but figured it would be a hole with the borgs being unable to interact with other things, and that it might be better to leave it as “yeah borgs can play instruments.” I can get some more opinions on it though, might be better. Also, it doesn’t effect game balance or round progression in any way, it’s just a fun thing that I think borgs should be able to do.

AnomalousPotato:
Gonna completely rework this

AnomalousPotato:
Did a poll and the simple "borgs just stand next to instruments and click them" implementation has won out

AnomalousPotato:
Changed it to large instruments only (piano, jukebox) as they don't need to be picked up

ZeWaka (approved):

PR #21664, opened 2024-12-12
Syndicate Pinpointer

JORJ949:
## About the PR 
Replaces nukie disk/nuke pinpointers with a 2 in 1 pinpointer that switches between tracking either

## Why's this needed? 
Nobody ever uses these pointers and only the scout/infil currently get a disk pointer, taking one pointer is already a fair bit of inventory space, but 2 is just too much for the amount of gear nukies have.

## Changelog 

```changelog
(u)JORJ949
(+)Syndicate Nuclear Operatives now get one pinpointer that toggles between tracking the nuke and tracking the auth disk.
```


Tarmunora:
I think that locking the disk pinpointers with the classes that are actually suited to being discreet is a feature, not a bug. Tacitly encouraging heavies/knights/etc. to go out on a disk-hunting safari sounds like a footgun 90% of the time, because you really don't want those guys running off alone and getting ganked.

JORJ949:
> I think that locking the disk pinpointers with the classes that are actually suited to being discreet is a feature, not a bug. Tacitly encouraging heavies/knights/etc. to go out on a disk-hunting safari sounds like a footgun 90% of the time, because you really don't want those guys running off alone and getting ganked.

The idea was not for them to run off to find the disk on the captain, but to find it in the pile of dead bodies scattered around the nuke.

PR #21670, opened 2024-12-13
Contraband traders will no longer dock with the station

pgmzeta:
[balance][events]
## About the PR 
Contraband travelling traders will no longer dock with the station. They can still dock at the diner.

Notes:
* This is a big shift in potential odds for contra traders in general since left/right traders have double chances over diner traders. I think the odds go from 1/8 to 1/40.
* nadir/oshan can no longer get contra traders ever, since there is no diner dock on that map.

## Why's this needed? 
NPC traders openly offering contraband on-station puts security in a weird place.

### DIFF
--- a/code/modules/economy/trader.dm
+++ b/code/modules/economy/trader.dm
@@ -650,7 +650,11 @@
 
 /obj/landmark/spawner/random_trader
 	spawn_the_thing()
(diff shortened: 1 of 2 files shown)


Tarmunora:
I don't see why this has to be RP-only, honestly

PR #21674, opened 2024-12-13
Add accessibility options for colorblindness

FlameArrow57:
[UI][QoL]

## About the PR 
Adds accessibility options for protanopia, deuteranopia, and tritanopia colorblindness, which works by applying a view tint for the user

Uses color matrix values from https://github.com/ParadiseSS13/Paradise/pull/17933 (not sure if i needed to ask for permission to use them?)

image

Protanopia - 
image

Deuteranopia - 
image

Tritanopia - 
image

## Why's this needed? 
QoL (also I have no experience with colorblindness, so this is really up to those with colorblindness to say how helpful it is. assumed it was helpful per the paradise station PR)

## Changelog 

```changelog
(u)FlameArrow57
(*)Added colorblind accessibility options for protanopia, deuteranopia, and tritanopia colorblindness under Game -> Accessibility
```

### DIFF
--- a/_std/color.dm
+++ b/_std/color.dm
@@ -57,6 +57,29 @@
 																		 0.00, 0.48, 0.53, 0.00,\
 																		 0.00, 0.00, 0.00, 1.00,\
 																		 0.00, 0.00, 0.00, 0.00)
+
+// note: for colorblind accessibility matrices, these are arbitrary values that were found to work well per https://github.com/ParadiseSS13/Paradise/pull/17933
+#define COLOR_MATRIX_PROTANOPIA_ACCESSIBILITY \
+	list(1, 0.475, 0.594, 0, \
+		0, 0.482, -0.68, 0,\
+		0, 0.044, 1.087, 0,\
+		0, 0, 0, 1,\
+		0, 0, 0, 0)
+
+#define COLOR_MATRIX_DEUTERANOPIA_ACCESSIBILITY \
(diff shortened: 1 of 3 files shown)


Carton171:
I would like to say, as a colourblind person:
Fuck. Yes.

Valtsu0:
As i understand it, Paradise is GPLv3 by d efault and i don't think it is compatible with our license.

You should probably ask the author for permission, even if it is definitely fine

Valtsu0:
Does this work even if view tint is disabled?

FlameArrow57:
> Does this work even if view tint is disabled?

Yeah

Kyle2143:
I feel like it would be more clear to put those three options in a submenu like "accessibility -> colorblindness" or something like that instead of just putting the 3 terms directly under "accessibility"

FlameArrow57:
> I feel like it would be more clear to put those three options in a submenu like "accessibility -> colorblindness" or something like that instead of just putting the 3 terms directly under "accessibility"

Would agree, but that would be a single category under accessibility for the moment, so I went without it. I would think it'd be more preferrable in the future if more categories are added under accessibility?

Kyle2143:
That's a fair point, but I would think it would have to be done at some point though if there are more varied accessibility things we end up adding. I suppose we could leave it until that time.

Tarmunora (approved):

Issue #19151, opened 2024-05-26
Keybind window closing itsef, cannot change or view keybinds. 


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

The window for changing keybinds closes itself seconds after opening, cannot change keybinds as it is. I let everything in the byond folder past my firewall and ran as admin incase it was a anti-virus or permission issue.

### Steps to reproduce

1. Attempt to modify keybinds through Game -> Inferface -> Keybinds

### Expected Behavior

For it to not close itself after several seconds and allow me to change/view the keybinds.

### Additional Information & Screenshots

Reported by: Cheslock25
Client version: 515.1639
On server: Goonstation 1 Classic: Heisenbee
Active test merges: #18978
Round log date: 2024-05-26-11-29
Reported on: 2024-05-26 12:17:16
Map: DONUT2

Garash2k:
Can confirm, the bug happens when opening the Keybinds menu while spectating. Once ingame, it closes itself after trying to change a keybind. Even after re-opening it, it'll still autoclose.

Good news is #21681 fixes that bug.

PR #21681, opened 2024-12-14
TGUI Keybind Menu

Garash2k:
## About the PR 
Ports the Keybind Menu from CHUI to TGUI
image 
image

White theme:
image

<details><summary>Draft points:</summary>
<p>

- [x] I'm too tired to make sure everything is fine right now :V
- [x] After changing, keybind order tends to shuffle. It's more egregious now that updates are faster.
- [x] dubious code such as having multiple `current_keymap = owner.keymap` and a `unparse = changed_keys["[current_keymap.keys[key]]"] || current_keymap.unparse_keybind(key),`
- [x] todo: capturing keyboard input in lieu of typing characters.

</p>
</details> 

## Why's this needed? 
Besides the technology upgrade, current CHUI menu is kinda broken, opening it while spectating permanently bricks trying to use it while alive (#19151.) Bug is fixed with this version

Note: it seems like the keybinds menu was somewhat fragile too regarding binding already bound keys. New UI is somewhat as fragile regarding that, a cleanup of the keybinding code would probably be required, and I'd like to think is out of scope of this PR, either that or I'd need help :V

Old UI:
261892373-7317fd5f-3030-4feb-ae88-58c9bbdfbd5a

[UI][QoL][Feature]
```changelog
(u)Garash2k
(*)Updated the keybind menu to use TGUI.
```


ZeWaka:
How does The binding mechanism work? Is it a straight character input or does it capture keyboard input? The latter would be a lot nicer, other servers have implementations we can use

ZeWaka:
How does The binding mechanism work? Is it a straight character input or does it capture keyboard input? The latter would be a lot nicer, other servers have implementations we can use

Garash2k:
Straight character input is how both the old ui and this version works, but I agree capturing keyboard input is absolutely the way to go, adding it as a todo

FlameArrow57:
Nice

ZeWaka:
X

TobleroneSwordfish:
This looks good, will need testmerging but we have another TGUI PR testmerged right now so will have to wait.
The only thing I'd say it could use is a modified notification like the character setup page, oh and in theory a way to block the browser shortcuts so you can bind something to CTRL + N without getting IE jumpscared. Not sure how possible that is but suuurely we can capture that somehow??

code/modules/interface/keybind_menu_chui.dm
```diff
@@ -1,51 +1,52 @@
...
-
-	GetBody()
-
-		current_keymap = owner.keymap //Need to refresh this since it's not a proper pointer?
-
-		var/list/html = list()
+	ui_interact(mob/user, datum/tgui/ui)
+		ui = tgui_process.try_update_ui(user, src, ui)
+		if (!ui)
+			ui = new(user, src, "Keybinds", "Keybinding Customization")
```
mordent-goonstation, line 15: Stick the title on the actual UI instead, I guess. It's how we do the vast majority, and this one doesn't have a special reason to customize the title or whatever.

tgui/packages/tgui/interfaces/Keybinds/Footer.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const Footer = () => {
+  const { act, data } = useBackend<KeybindsData>();
+  const { hasChanges, resetting } = data;
+
+  return (
+    <>
+      <Button
+        onClick={() => act('confirm')}
+        color={!hasChanges ? null : 'good'}
```
mordent-goonstation, line 20: Verify, but I think `undefined` is preferred to `null` for when we're not changing the color.

tgui/packages/tgui/interfaces/Keybinds/Footer.tsx
```diff
@@ -0,0 +1,35 @@
...
+        icon="save"
+      >
+        Confirm
+      </Button>
+      <Button
+        onClick={() => act('reset')}
+        color={!resetting ? null : 'bad'}
+        icon="trash"
+      >
+        {!resetting ? 'Reset All' : 'Confirm?'}
```
mordent-goonstation, line 30: Having both a `Confirm` button (the one before this) and this one (`Confirm?` after you've pressed `Reset All` once) is a bit odd. I also don't see a way to get out of the resetting state short of saving your keybinds.

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { isEscape } from 'tgui-core/keys';
+import { BooleanLike } from 'tgui-core/react';
+
+import { useBackend } from '../../backend';
+import { formatKeyboardEvent, isStandardKey } from './formatKeyboardEvent';
+import { KeybindData, KeybindsData } from './types';
+
+interface KeybindProps {
+  keybind: KeybindData;
+  isFocused: BooleanLike;
```
mordent-goonstation, line 17:
```suggestion
  isFocused: boolean;
```
`BooleanLike` is for when it's coming from the backend data, we're actually setting this prop to `true`/`false` so it's an actual `boolean`.

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const Keybind = (props: KeybindProps) => {
+  const { act } = useBackend<KeybindsData>();
+  const { keybind, isFocused, setFocusedKey } = props;
+  return (
+    <LabeledList.Item label={keybind.label}>
+      <Button
+        onKeyDown={(event) => {
+          event.preventDefault();
+          if (isEscape(event.key)) {
+            setFocusedKey('');
```
mordent-goonstation, line 29: Feels odd to not have a `return` here, especially given that you then also do `setFocusedKey('')` shortly after.

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -0,0 +1,47 @@
...
+          }
+          if (!isStandardKey(event)) {
+            return;
+          }
+          setFocusedKey('');
+          const value = formatKeyboardEvent(event);
+          return act('changed_key', { id: keybind.id, value });
```
mordent-goonstation, line 36:
```suggestion
          act('changed_key', { id: keybind.id, value });
```
`act` doesn't return anything, so no need to return it.

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -0,0 +1,47 @@
...
+        }}
+        onClick={() => {
+          setFocusedKey(keybind.id);
+        }}
+        color={isFocused ? 'good' : null}
```
mordent-goonstation, line 41: Check re: `undefined` vs. `null`.

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -0,0 +1,47 @@
...
+      >
+        {isFocused ? '...' : keybind.changedValue || keybind.savedValue}
```
mordent-goonstation, line 43: Possibly consider `??` over `||`. Not verified but I assume `changedValue` is a string with the new key thing, which should be fine as `||` here, but e.g. `changedValue` of empty string is falsey. Is that the behavior you want?

tgui/packages/tgui/interfaces/Keybinds/KeybindList.tsx
```diff
@@ -0,0 +1,56 @@
+/**
+ * @file
+ * @copyright 2024
```
mordent-goonstation, line 3:
```suggestion
 * @copyright 2025
```

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -0,0 +1,47 @@
+/**
+ * @file
+ * @copyright 2024
```
mordent-goonstation, line 3:
```suggestion
 * @copyright 2025
```

tgui/packages/tgui/interfaces/Keybinds/Footer.tsx
```diff
@@ -0,0 +1,35 @@
+/**
+ * @file
+ * @copyright 2024
```
mordent-goonstation, line 3:
```suggestion
 * @copyright 2025
```

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -0,0 +1,47 @@
...
+  keybind: KeybindData;
+  isFocused: BooleanLike;
+  setFocusedKey: (string) => void;
+}
+export const Keybind = (props: KeybindProps) => {
+  const { act } = useBackend<KeybindsData>();
+  const { keybind, isFocused, setFocusedKey } = props;
+  return (
+    <LabeledList.Item label={keybind.label}>
+      <Button
```
mordent-goonstation, line 25: Should these be children, or the `buttons` prop of the `LabeledList.Item` above?

tgui/packages/tgui/interfaces/Keybinds/KeybindList.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <NoticeBox info>
+        You can only rebind keys you have access to when opening the window.
+        <br />
+        Ex: You can only change human hotkeys if you are currently human.
+      </NoticeBox>
+      <LabeledList>
+        <LabeledList.Item label="Action">
+          Corresponding Keybind
+        </LabeledList.Item>
+        {keys.sort(sortKeys).map((k) => (
```
mordent-goonstation, line 31: Find with there being a sort for alphabetizing, trying to get my head around what actions might have an id of numbers to warrant use of the custom sort function. Could you give some examples?

tgui/packages/tgui/interfaces/Keybinds/KeybindList.tsx
```diff
@@ -0,0 +1,56 @@
...
+const sortKeys = (k1: KeybindData, k2: KeybindData) => {
+  if (Number(k1.id) && Number(k2.id)) {
+    return Number(k1.id) - Number(k2.id);
+  } else if (Number(k1.id)) {
+    return -1;
+  } else if (Number(k2.id)) {
+    return 1;
+  } else {
+    return k1.id.localeCompare(k2.id);
+  }
```
mordent-goonstation, line 55:
```suggestion
  }
  return k1.id.localeCompare(k2.id);
```
No actual change here, but good practice to have the last return not in an `else`. Means a future change is less likely to break it (though TS should catch it).

tgui/packages/tgui/interfaces/Keybinds/types.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @file
+ * @copyright 2024
```
mordent-goonstation, line 3:
```suggestion
 * @copyright 2025
```

mordent-goonstation (changes requested):

mordent-goonstation:
> This looks good, will need testmerging but we have another TGUI PR testmerged right now so will have to wait. The only thing I'd say it could use is a modified notification like the character setup page, oh and in theory a way to block the browser shortcuts so you can bind something to CTRL + N without getting IE jumpscared. Not sure how possible that is but suuurely we can capture that somehow?? (do we have `preventDefault`?)

To my knowledge some of the brower-level keybinds can't be blocked. As this is a port, I imagine the current chui one has this problem already? We could augment at a later date by allowing to to e.g. manually click on some modifier buttons for the case where you can't press the bind directly.

mordent-goonstation:
> > To my knowledge some of the brower-level keybinds can't be blocked.
> 
> Seems like `preventDefault` does work to supress those, I was surprised too. Maybe worth putting a snippet globally somehow to prevent IE browser shortcuts? I guess that would be an upstream PR.

Neat. In fairness that *would* be something that is browser dependent, so we'd have to make sure we hit the browser version we care about in our method. Quick Google implies that `keyDown` is the event that matters, but would need to verify.

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
Garash2k: I'm not sure how someone would end up as an empty string as a keybind, but I guess in that case we should show that an empty string is being used and not the initial value.

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
Garash2k: I think in this case, the button is the content/focus of the list, so in makes sense to be children. As opposed to elsewhere, where most of the time, the button serves to do an action on the list's content.

tgui/packages/tgui/interfaces/Keybinds/KeybindList.tsx
Garash2k: Yup, example:

- Top of the list should be Up with id:2
- Down:4, Left:8, Right:16
- Then there's stuff like Run:32, Point:128 and Pull:512
- And then the rest of the IDs should be alphabetical: Say:say, OOC:ooc, Grab Intent:grab, Target Left Leg:l_leg.

If we were to sort them alphabetically, Point would be put above Up, and Run would be put between Up and Down. We can't sort the numerically because most of them are strings.

tgui/packages/tgui/interfaces/Keybinds/Footer.tsx
Garash2k: I was suggested have the same behavior as the Character Setup "Reset All" button, for which you can't get out of the resetting state either. Looking it up now though, it's a Button.Confirm so I'll use that, and I'll use "Confirm Reset All?" to disambiguate.

ZeWaka:
Thanks for helping clean up my spite code

PR #21713, opened 2024-12-15
Command Apparel Vendor

JORJ949:
## About the PR 
Adds the command apparel vendor to bridges in rotation maps, and the clothing room in afterlife bar
image
image

Also moves commonly shared vars on the jobclothing vendors to the common parent type

## Why's this needed? 
Gimmick command jobs are fairly common (e.g. "Head of Staffies") but theres no generic command outfits for them to make use of, this allows them to pick some NT/Command themed outfits up.

## Changelog 

```changelog
(u)JORJ949
(*)Added Command Apparel vendor. Can be found in the bridge and afterlife bar.
```


flrsh:
I don't think the Nanotrasen jumpsuit or NT backpack / satchel are supposed to just represent "any kind of command role", they're pretty tied to specific jobs that you spawn as at round start. Having vendable berets also kinda removes the point of the trait or having to buy them from the diner, imo. And once you remove those from the list, I'm pretty sure you can get the rest of the items from the HoP / clothing fab / Bridge crate already.

JORJ949:
> I don't think the Nanotrasen jumpsuit or NT backpack / satchel are supposed to just represent "any kind of command role", they're pretty tied to specific jobs that you select when you join the game. Having vendable berets also kinda undermines the point of the trait or having to buy them from the diner, imo. And once you remove those from the list, I'm pretty sure you can get the rest of the items from the HoP / clothing fab / Bridge crate already.

The NT jumpsuit is currently (Afaik) unused with the removal of the regional director, and its odd that NT doesnt sell NT brand merchandise to their staff, the only on-station way to get a beret currently is stealing beepsky's. If anything NT would probably want their crew running around wearing free advertising. While it does undermine the trait and diner a little bit, the trait still gives you one at roundstart regardless of if you have bridge access and the diner merchant still sells other items. Other spacebux purchases and trait items are obtainable outside of it, e.g. handkerchief, AI hats, and I dont think a beret style or backpack style is that important to gatekeep behind preround actions.

cheekybrdy:
why just the mask for blue clown? doesn't the suit have the same obtainability restrictions

PR #21720, opened 2024-12-16
Syndicate Beret

JORJ949:
## About the PR 
Adds a syndicate beret to the syndicate apparel vendor
image

## Why's this needed? 
The syndicate doesnt really have any normal headgear, and I thought a beret would work nice

## Changelog 

```changelog
(u)JORJ949
(+)The Syndicate has issued syndicate berets to all syndicate apparel vendors.
```


TobleroneSwordfish (approved):
cool hat

PR #21732, opened 2024-12-18
Make "Centcom" law modules blue instead of pink

JORJ949:
## About the PR 
Makes laws uploaded via admin command blue instead of pink

## Why's this needed? 
Pink doesnt feel like a "centcom" colour, blue is much more fitting and not used for any station-laws so is very clear what it is

### DIFF
--- a/code/obj/machinery/law_racks.dm
+++ b/code/obj/machinery/law_racks.dm
@@ -727,10 +727,9 @@
 	 * Does not call UpdateLaws()
 	 * Intended for Admemery
 	 */
-	proc/SetLawCustom(lawName, lawText, slot = 1, screwed_in = FALSE, welded_in = FALSE)
-		var/path = /obj/item/aiModule/custom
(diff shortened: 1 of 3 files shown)


code/obj/machinery/law_racks.dm
```diff
@@ -728,7 +728,10 @@
 	 * Intended for Admemery
 	 */
 	proc/SetLawCustom(lawName, lawText, slot = 1, screwed_in = FALSE, welded_in = FALSE)
-		var/mod = new /obj/item/aiModule/custom(lawName,lawText)
+		var/path = /obj/item/aiModule/custom
+		if(lawName == "Centcom Law Module")
+			path = /obj/item/aiModule/custom/centcom
+		var/mod = new path(lawName,lawText)
```
TobleroneSwordfish, line 734: Passing in the colour or type path would be better than checking the magic string here

TobleroneSwordfish (changes requested):

PR #21733, opened 2024-12-18
Feature: Monitor Cybereye

pgmzeta:
[feature][medical]
## About the PR 
* New object: Monitor Cybereye  
  * You get an ability to view a camera on the "public" network
  * The internal monitor can be swapped with any camera monitor. Rancher can replace their eye to keep a literal eye on their chickens across the station. Syndicate spies can use the spy camera to discretely view their targets.
  * Emag effect: unlocks all station networks (SS13/public/ranch/Zeta/Mining/telesci)
    * Once emagged, you are unable to see out of the replaced eye
    * New variable on eyes `provides_sight`, used for blindness checks. 
* Change camera viewer check to see if the camera is anywhere in/on the person
* Refactor camera viewer calls to reduce duplication and split functionality
* Organ abilities are now cleanly removed if the organ is disposed directly
* Added Monitor Cybereye to Robotics Fabricator and the organ trader

## Why's this needed? 
More things for robotics to install into people. Hefty drawback but being able to remotely spy on people can be a blessing.

## Changelog 

```changelog
(u)glowbold
(+)Monitor cybereyes can remotely view cameras. Ask a roboticist to install one today!
```


code/obj/item/organs/eye.dm
```diff
@@ -464,12 +467,42 @@ TYPEINFO(/obj/item/organ/eye/cyber/laser)
...
+	default_material = "pharosium"
+	iris_color = "#0d0508"
+	icon_state = "eye-monitor"
+	provides_sight = FALSE
+	var/obj/item/device/camera_viewer/viewer = null
+
+	New()
+		. = ..()
+		src.viewer = new(src)
+		src.viewer.network = list("public", "Robots", "telesci")
```
JORJ949, line 487: Robots will include syndicate cyborgs and any in upload (allowing basically free checking of upload if you know a borg is in there. I think this should be restricted to the emag effect.

AXLFAN2000:
there's a typo in the changelog (roboricist)

ZeWaka:
It'd be interesting for the spy stickers to somehow work with this. Unsure how exactly but it came to my mind.

pgmzeta:
> It'd be interesting for the spy stickers to somehow work with this. Unsure how exactly but it came to my mind.

what I could do is make the embedded monitor swappable, so you could put the syndie spy sticker monitor inside it? don't want to add more to that box but it definitely feels like it could be an option

pgmzeta:
> It'd be interesting for the spy stickers to somehow work with this. Unsure how exactly but it came to my mind.

monitors swappable now, lmk what you think

TobleroneSwordfish:
This is a really cool feature buut tracking viewports are incredibly jank and not really meant to be player accessible in their current state: they track very very jerkily and _allow you to see through walls_. I'm not sure either of these are fixable in Byond as it is now.

Tarmunora:
Not terribly sold on the emag effect - other cybereyes don't have an emag interaction(yet), and doesn't really fit with the general "safeties off" theme that other emagged cyberorgans aim for. Could come back to it in the future as part of a "emagged cybereyes pass", but I don't think it belongs rn

pgmzeta:
switched back to being able to use the camera viewer like it was in-hand. removed emag effect.

not sure its worth the no-vision-in-eye tradeoff with these changes, but kept it as-is for now.

Tarmunora:
> switched back to being able to use the camera viewer like it was in-hand. removed emag effect.
>
> not sure its worth the no-vision-in-eye tradeoff with these changes, but kept it as-is for now.

Funnily enough, if normal montior eyes didn't block vision, I think that emag effect making it so the user gets a uses weird janky viewport instead in exchange for half their vision, *would* actually be very on-brand for emagged cyberorgans 🙂

pgmzeta:
Hmm, could make the emagged version filter for exact matches on `/obj/machinery/camera` and limit it to static viewports ala the AI viewports. I think the see-through-walls issue would still apply, though :/

pgmzeta:
Alt option: could have emag kill vision in the eye and unlock all camnets

pgmzeta:
> Alt option: could have emag kill vision in the eye and unlock all camnets

went with this option; default does not block vision, since switching to the camera takes up your whole view. emagging does block your vision (temporarily removed while in the camera view via `provides_sight`)

Tarmunora (approved):

PR #21738, opened 2024-12-18
Interdictors protect from APC zaps

### LINKED ISSUE #20900: Apcs and Ranching
### Labels

[BUG][CRITICAL]

### Description

Overloaded APCS ruin ranching

### Steps to reproduce

So here is where im just going to lay it out. No other job can be just deleted with an overloaded APC like ranching. It killes the chickens we work on and makes other chickens sad and that means less eggs. I'm not saying that that they shouldn't overload but maybe the device that protects against radation can work against overloaded apcs. It happens early shift and then just stays like that because no one can fix the engine or whatever. It has happened multiple times and I might as well cryo as a rancher when it happens because I can't keep any chicken alive. 

### Expected Behavior

I don't know the spatial disruptor prevents this from happening. It really stinks. 

Reported by: longweird
Client version: 515.1643
Active test merges: #20547 and #20620
Round log date: 2024-09-29-03-53
Reported on: 2024-09-29 04:27:56
Round ID: 189398
Map: COGMAP

pgmzeta:
[balance][station systems]
## About the PR 
* Interdictors cancel out APC zaps
## Why's this needed? 
Fix #20900

## Changelog 

```changelog
(u)glowbold
(+)Spatial Interdictors can nullify APC overcharge zaps. Ranchers, take note.
```

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -287,13 +287,12 @@ ADMIN_INTERACT_PROCS(/obj/machinery/power/apc, proc/toggle_operating, proc/zapSt
 
 	var/list/starts = new/list()
 
-	for(var/obj/machinery/interdictor/interdictor in oview(5, src))
-		if (interdictor.canInterdict)
-			starts.Add(interdictor)
+	for_by_tcl(IX, /obj/machinery/interdictor)
+		if (IX.expend_interdict(500, src))
+			return 1
 
 	for(var/mob/living/M in oview(5, src))
 		if(M.invisibility) continue
-		if(M.hasStatus("spatial_protection")) continue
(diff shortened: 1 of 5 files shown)


TobleroneSwordfish:
I'm not sure this really fixes the issue, since over a long enough time period all the chickens are going to get targeted anyway. Maybe interdictors acting as lightning rods (as suggested in the original issue) could work better?

TobleroneSwordfish (changes requested):

TobleroneSwordfish:
Oh err, it'd probably be better if they still zapped the interdictor so you know it's working? Interdictors have an issue of it being kind of hard to tell if they're doing anything.

PR #21740, opened 2024-12-18
Replace donut3 telesci decal tile edges with tiles

### LINKED ISSUE #16359: Scorching two tone tiles is strange
### Labels

[BUG][MINOR]

### Description

Half the tile is burnt.

### Steps to reproduce

1. Explode something in telesci on donut3
2. look at a half white half purple tile
3. only the white side is burnt

### Expected Behavior

The whole tile be scorched

Reported by: ChaoticTyrant
Active test merges: #14886
Round log date: 2023-10-09-08-43
Reported on: 2023-10-09 09:39:24
Map: DONUT3

pgmzeta:
[mapping][bug]
## About the PR 
The neat tile decorations in telesci were using decals instead of two-tone tiles. Maintain the design while reducing the number of tile decals where possible.

## Why's this needed? 
Reduce number of decals

### DIFF
--- a/maps/donut3.dmm
+++ b/maps/donut3.dmm
@@ -7037,10 +7037,6 @@
 /turf/simulated/floor/plating/jen,
 /area/station/maintenance/outer/ne)
 "caq" = (
-/obj/decal/tile_edge/line/purple{
-	dir = 4;
-	icon_state = "tile1"
-	},
 /obj/cable,
 /obj/cable{
 	icon_state = "0-4"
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Hmm, these are used in an awful lot of other places so I'm not sure removing them from just here fixes the issue really?
They should be "merging" with the turf by adding themselves to it as an image, if the burn texture is layering under the decal images then that's the issue here.

pgmzeta:
> Hmm, these are used in an awful lot of other places so I'm not sure removing them from just here fixes the issue really? They should be "merging" with the turf by adding themselves to it as an image, if the burn texture is layering under the decal images then that's the issue here.

will take a look at fixing in another pr; this still reduces the number of unneeded decals

PR #21752, opened 2024-12-19
Add three (3) pixels to the HoS cape

JORJ949:
## About the PR 
Adds 3 pixels between the legs on the HoS cape
image

## Why's this needed? 
the cape does go down that far as shown by the back sprite


Mister-Moriarty:
Note that this will result in those three pixels layering over any item of clothing that doesn't leave a gap between the character's legs, notably skirts.

pgmzeta:
renders over dresses

PR #21753, opened 2024-12-19
Medbay Pocket Guide Second Edition

AnomalousPotato:
[Game Objects][Feature]

## About the PR 
Updates the Medbay Pocket Guide to be far more comprehensive, useful, and up-to-date.

## Why's this needed? 
Every time I get a medical trainee, they talk about reading the pocket guide, which is years old and sometimes the opposite of helpful given how much the game has changed. If they are going to read it, I would prefer it to be helpful.

## Changelog 

```changelog
(u)AnomalousPotato
(+)Updated Medbay Pocket Guide to be more up-to-date, comprehensive, and useful.
```


code/obj/item/book.dm
```diff
@@ -69,10 +69,16 @@ Custom Books
 	desc = "A condensed guide of job responsibilities and tips for new crewmembers."
 
 	medical
-		name = "Medbay Pocket Guide"
+		name = "Medbay Pocket Guide, Second Edition"
```
JORJ949, line 72: This isnt the second edition one
```suggestion
		name = "Medbay Pocket Guide"
```

JORJ949:
Why not just replace the old one like was done for space law a while back? Doesn't seem necessary to keep outdated guidebooks.

flappybatpal:
I'm not sure we want author credits, generally because this causes issue if someone gets banned and so we have been avoiding player references.

AnomalousPotato:
> I'm not sure we want author credits, generally because this causes issue if someone gets banned and so we have been avoiding player references.

I mean, if it's a problem, I'll remove it. I was following in line with the previous edition, it's not a big issue for me.

AnomalousPotato:
holiday bump

strings/books/medbay_pocket_guide.txt
```diff
@@ -1,126 +1,167 @@
...
+</div>
+<p>&nbsp;</p>
+<p>Radiation causes tox and burn damage, but it's fairly simple to heal. Just inject potassium iodide from the NanoMed, and keep up with the escalating damages. The potassium iodide will reduce remove radiation from the patient over time.</p>
```
Sord213, line 150: "will reduce remove radiation"

Sord213 (changes requested):
Looks good2me otherwise

PR #21754, opened 2024-12-19
Small blood lifeprocess optimization

FlameArrow57:
[INTERNAL][PERFORMANCE]

## About the PR 
Title

Also removes some blood code that was part of a bad merge I think

For /datum/lifeprocess/blood/process::
image

## Why's this needed? 
Optimization


pgmzeta (approved):
This reads like blood loss has been twice as impactful as intended since the bleed revamp? Numbers might need tweaking at a later point @TDHooligan

ZeWaka (commented):
Have you not deleted the ranges for a bunch of the blood? I see the dupe but you deleted more

FlameArrow57:
Yeah seems so, not really sure :/ I'll just remove it, for putting under another PR. Think TDHooligan needs to give input. Performance is just a few us higher with it in

TDHooligan:
yeah that adds up. i'm keen on the dedupe too

ZeWaka:
im still unsure about this so someone else can review/merge it

PR #21770, opened 2024-12-20
Tabletop Chemical Dispenser glassware

TDHooligan:
[FEATURE][CHEMISTRY]
## About the PR 
Adds the portable chem dispenser to the chemistry manufacturer. Requires miraclium.

image

Generates 1-5 units of a given basic reagent every 4 processing ticks. Comparatively slow for now, just to see what happens.
Sprites by flaborized

https://github.com/user-attachments/assets/1dbdc984-ded8-4fc8-9cb5-f2e37a8b24ef

## Why's this needed? 
Finally, a non-manual input to chemistry glassware. Big bubbling machines all over science.

## Changelog 

```changelog
(u)TDHooligan, Flaborized
(*)The portable chemical dispenser has been added, for automating your glassware setups! Find it in the science manufacturer.
```


TDHooligan:
my local master had gang changes in. oops. fixd

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -1013,6 +1017,134 @@ proc/ui_describe_reagents(atom/A)
...
+		. = ..()
+
+	disposing()
+		src.remove_container()
+		. = ..()
+
+	mouse_drop(atom/over_object, src_location, over_location)
+		if (ismob(over_object))
+			var/mob/user = over_object
+			if (user == usr && !user.restrained() && !user.stat && (user.contents.Find(src) || in_interact_range(src, user)))
```
TobleroneSwordfish, line 1059:
```suggestion
			if (user == usr && !user.restrained() && !is_incapacitated(user) && in_interact_range(src, user))
```

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -1013,6 +1017,134 @@ proc/ui_describe_reagents(atom/A)
...
+			fabricating_overlay.color =  hsl2rgb(hsl[1], hsl[2], clamp(hsl[3],255,20))
+
+		else
+			fabricating_overlay = null
+		src.UpdateOverlays(src.fabrication_tick_lights, "light")
+		src.UpdateOverlays(src.fabricating_overlay, "overlay")
+
+	proc/adjust_flow_rate(mob/user)
+		var/reagent = tgui_input_list(user,"Set desired reagent", "Set reagent", basic_elements, reagent_to_fabricate)
+		var/number = tgui_input_number(user,"Set fabrication volume, in units", "Set fabrication rate (1-5)",flow_rate,5,1,FALSE,TRUE)
```
TobleroneSwordfish, line 1128: Should cancel the second popup if the first one is closed (ie returns null)

code/modules/chemistry/Chemistry-Tools.dm
```diff
@@ -1013,6 +1017,134 @@ proc/ui_describe_reagents(atom/A)
...
+
+	attackby(obj/item/I, mob/user)
+		if (istype(I, /obj/item/screwdriver))
+			adjust_flow_rate(user)
+			user.visible_message(SPAN_NOTICE("[user.name] adjusts \the [src.name]."))
+			return
+		..()
+	process()
+		if (running)
+			fabrication_tick += 1
```
TobleroneSwordfish, line 1083: Should ideally be mult adjusted, see `item.last_tick_duration` maybe?

TobleroneSwordfish (changes requested):
I like this a lot, the fun kind of automation (hopefully)

code/modules/chemistry/Chemistry-Tools.dm
TDHooligan: i'll leave it for now - see how slow it gets in practise. it jumping might mess up some weird complex flow i haven't thought of yet

TobleroneSwordfish: Fair actually, consistency might be better than speed.

AnonymousUserTheOnly:
i wonder if this will get functions that can be "improved" with an emag.... certainly sounds like it, being able to emag it to generate toxic materials/chemicals sounds like an emag thing to do.  that, or maybe set it to be on without outputting unless checked and reset, so it eventually explodes, with the explosion and subsequent chemical catastrophe based on the setting.

PR #21781, opened 2024-12-21
Changeling Multistings

colossusqw:
[LABEL][Feature][Player-Actions]

## About the PR 

More details in the HackMD: https://hackmd.io/@colossusqw/SkLFNQU3C

https://github.com/user-attachments/assets/eb46de36-4b18-4401-b833-e57df940d1c7

This PR merges the classic and RP changeling sting to neurotoxin, altering it to cause paralysis as to better acomodate monologues and assorted RP(as it was previously done to capulettium). It also Swaps out the duo of Capu and LSD for a few options, all sharing the same cooldown, but having different timings for balance. They are as follow:

- **Neurotoxin:** 50u, 140s cooldown. Standard Changeling sting, most useful in cornering and disposing of single targets with highly debilitating effects and KO.

- **Neurodepressant:** 5u, 20s cooldown. Defensive tool that while not as debilitating as neurotox, is much less commited due to the very low cooldown that enables stinging multiple targets. Most useful at evasion.

- **Hemotoxin:** 9u, 30s cooldown. More offensive than neurodepressant, can be deadly if not eventually treated, but may require multiple stings(or a few stabs) to do so. Very low cooldown makes it the least commitment heavy option.

- **Madness Toxin:** 15u, 240s cooldown, slight DNA cost. Slow to start, and with a very long cooldown, this sting makes up for it in it's capacity in creating an excelent distraction, or derailing an attempt at stopping the changeling.

## Why's this needed? 
I've had this PR on the backburner for a bit, but i feel it's finally time to see what people think of it.

Currently changelings have some of the least variety of gameplay loops when compared to traitors, wraiths, and other antags, and the loop of neuro into sting is can get somewhat predictable. This change aims to remedy that with more sting variety, the multitude of available poisons both broadens the ways in which changelings can play, and adds intrigue to the station as the same one note neuro sting isn't so immediately recognizable.

## Changelog
```changelog
(u)Colossusqw
(*)Changelings now have a variety of sting options to choose from, sharing a collective cooldown. (Minor changes for details)
(+)Now changelings can swap between a neurotoxin sting(50us, 140s cooldown), a neurodepressant (5u, 20s), hemotoxin (9u, 30s), and madness toxin sting (15u, 240s cooldown, 5 DNA points).
(+)Neurotoxin now paralyses instead of causing a KO, and is a changeling sting both in the classic and RP server.
```


IPling:
I like this. 

1. Tetrodotoxin stings add significantly more flavor to the destruction a changeling can cause, via just poisoning the crew slowly and silently then stealing the corpses. A negative can be that it causes significantly less physical interaction between the crew and the changeling, but personally I enjoy the intrigue it can give.

2. I don't have much to say about the other stings, though I do really enjoy having some more milder options for when a player doesn't wanna be as deadly.

3. The removal of LSD is fine by me, as I found it to be quite gimmicky, and quite boring after your first time experiencing it.

In general I do believe this would give some variety to lings which is nice, though I do doubt it will make them less obvious as people will just learn about the new stings. I do fear the balancing a bit, as changelings currently can be quite powerful and giving them a bigger toolset wouldn't help with that, but the severity of that can only really be seen in testing.

TobleroneSwordfish:
I like the concept buut:
- The tetrodotoxin sting is 100% lethal without treatment and you get very very little time to get treatment. Compare this to neurotoxin which is obvious much earlier and does almost no damage if the ling can't manage to kidnap the victim. Lings absolutely do not need a "You are going to die in 2 minutes" button.
- Madness toxin is interesting but I foresee this one being used an awful lot by changelings just endlessly stinging people to cause chaos with very little risk or opportunity for the crew to identify them. Again, compare this to current changeling that's generally discouraged from doing this because all they can do is cause minor brain damage and hallucinations.

In general I'm veeery cautious to add damaging poisons to changelings since the old unstable mutagen ling combo had to be changed due to people just going around silently poisoning rather than actually trying to eat anyone. Hemotoxin is at least very immediately obvious so might provide more of an opportunity for crew to figure out who just poisoned them?

colossusqw:
!merge_upstream

flappybatpal:
Tetrodotoxin is lethal at 15. It's probably equally as debilitating as Neurotoxin (until Neuro makes you unconcious) but there's more of a chance to survive Neuro if the ling can't kidnap you at the point.

The 9u of hemotoxin looks useful for combat (and lethal in two stings but that takes ages to die from) but I'm not sure it is better than just using neurotoxin. Can't be dismissed with smelling salts I suppose 
30s cooldown on hemo would let you kill people with it quite easily.

Definitely one for a test merge at least. More ling options might be good but

colossusqw:
Can i has testmerge pls

code/modules/antagonists/changeling/abilities/stings.dm
```diff
@@ -69,18 +69,35 @@
...
-		icon_state = "stingneuro"
-		venom_id = "capulettium"
-		inject_amount = 20
+	neurodepressant
+		name = "Neurodepressant Sting"
+		desc = "Transfer neurodepressant into your target. Good for escaping chases and disorienting targets."
+		icon_state = "stingdepres"
+		venom_id = "neurodepressant"
+		inject_amount = 5
+		cooldown = 20
```
TobleroneSwordfish, line 83:
```suggestion
		cooldown = 20 SECONDS
```

ZeWaka:
image

Tarmunora:
Even if it may fit thematically, I don't think that giving changlings a "just a damage poison" (hemotoxin) option is a good idea, as it acts as tacit encouragement to just invisibly poison people on cooldown, rather than engaging in the rest of the changeling kit. 
Similarly, the madness toxin sting feels a little suspicious in that regard, though less immediately problematic.

PR #21784, opened 2024-12-22
Restrict fishing rod storage to food items

pgmzeta:
[balance][game objects]
## About the PR 
Restrict fishing rod bait storage to food items.

## Why's this needed? 
Civilian cyborgs can currently store any normal or lighter weight class item in their fishing rod. Moving it to food keeps it closer to tongs behavior.

### DIFF
--- a/code/modules/fishing/fishing_gear.dm
+++ b/code/modules/fishing/fishing_gear.dm
@@ -24,8 +24,7 @@
 	New()
 		..()
 		RegisterSignal(src, COMSIG_ITEM_ATTACKBY_PRE, PROC_REF(attackby_pre))  // Storage for a single lure item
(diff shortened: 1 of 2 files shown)


LorrMaster:
Maybe only apply the change to cyborg fishing rods? That way players can use use credits or something as bait for whatever reason.

pgmzeta:
Want to avoid cyborg/human item divergence unless absolutely necessary. The current bait pools are just food items.

Tarmunora (changes requested):
should update `fishing_lootpools.dm` to make it explicit that expected lures should be food only, if adding a hard restriction on what items can be used as lures.

Tarmunora (approved):

PR #21786, opened 2024-12-22
Add a cap of 3 microbomb purchases

JORJ949:
## About the PR 
Limits the amount of microbombs purchasable in each uplink to 3

## Why's this needed? 
The macrobomb was removed for being too easy but you can still just buy 12 microbombs in maints, grab a bible, biblefart in medbay and destroy an entire department less than 15 minutes into the round, what is intended to be a "fuck you" to whoever kills you is often weaponised on classic and this heavily limits that usage.

## Changelog 

```changelog
(u)JORJ949
(*)You may now only purchase a maximum of 3 microbombs
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -415,6 +415,7 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 	items = list(/obj/item/implanter/uplink_microbomb)
 	cost = 1
 	vr_allowed = FALSE
+	max_buy = 3
 	desc = "This miniaturized explosive packs a decent punch and will detonate upon the unintentional death of the host. Do not swallow and keep out of reach of children."
 
 /datum/syndicate_buylist/traitor/lightbreaker


DisturbHerb:
We have mechanisms to report abuse of mechanics like this. I would sooner see a price increase for the microbomb than restrict its usage this heavily.

JORJ949:
> We have mechanisms to report abuse of mechanics like this. I would sooner see a price increase for the microbomb than restrict its usage this heavily.

I don't think causing destruction at the same level of a TTV should be this easy but also don't want to nerf those who decide to go all out and then spend their last 1-3tc on bombs, if you want to have a large microbomb you can still hand your TC to another traitor and have them buy more for you, this simply makes it harder to create significant station damage with them.

444Portal:
Not sure if there should be a cap. If there should be one though, I think 5-6 is where I'd imagine it to be.

JORJ949:
I don't want to nerf the "fuck you" aspect of the microbomb, taking down whoever is killing you with you, what I do think should be nerfed is the sheer effectiveness of station damage, punishing EVERYONE for your death, even those completely uninvolved. a couple large microbombings essentially forces a shuttle call, especially if they hit critical areas such as cloning and cargo and unless all of engineering comes to repair (they wont) it can take upwards of 30 minutes to get the damaged areas repaired, repressurized and operational, and most microbombings of this scale never get repressurized.

 Perhaps this could only be a classic limitation as I doubt its an issue on RP, but i'm personally against adding more differences between servers.

JORJ949:
> If someone bombs themselves 5 minutes into a round, they wasted their traitor roll. And will face rule repercussions.
> 
> I really don't see why we should "fix" in a way that removes any gimmick potential microbombs have.

This also applies later into the round, why should the entire server be punished by medbay exploding because staffie mc staffieface decided to beat up a doctor? 3 microbombs is still a decently sized explosion and will take out whoever is killing you, its just not TTV levels. An alternate option was to add a cap to microbombs stacking, or significantly reduce the stacking, but I wanted to leave a way for these levels of destruction to be reached, which can be obtained by coordinating with other traitors.

While microbombing 5m in is against the escalation rules on RP, I'm not entirely sure its rulebreaking on classic and it happens relatively often. Most microbomb uses are 1-5 paired with something else that will probably get you killed, e.g. a wrestling belt. If the macrobomb was removed for being able to create large station damage with ease, why do we have basically the same thing but it just takes 20s longer to achieve?

IPling:
Personally, exploding randomly on classic with no way to prevent it or combat it is apart of the game and the joy of it, even if it was round-start. It's only ever a problem when it happens every (other) round, which well... it doesn't. Macrobombs, from my knowledge, were removed because they encouraged such behaviour which resulted in it being almost every round. 
RP doesn't have this problem, because well... its against the rules there.

TobleroneSwordfish:
Closing due to overwhelming negative feedback, I don't think this is enough of a prevalent issue to put such a heavy restriction on microbomb usage. If we really wanted to nerf mass microbomb usage I'd rather either cap how many can be implanted or how high the explosive power stacks.

PR #21793, opened 2024-12-22
Gang crate pull slowdown

TDHooligan:
[GAMEMODES][BALANCE]
## About the PR 
Lowers the speed that gang crates can be pulled at.

## Why's this needed? 
Crates rarely get fought over and instead just become a twitchy snatch and grab. 

More of a tonal shift for how gang crates are obtained. and might not even be good. can just revert if so.

## Changelog 

```changelog
(u)TDHooligan
(+)Gang crates get pulled slower now.
```


ZeWaka:
Definitely a TM thing

PR #21794, opened 2024-12-22
Allow janktank use on crit patients

TDHooligan:
[BALANCE][GAMEMODES]
## About the PR 
Janktank II can be applied to people in critical, instead of just corpses.

Also adds a bit of bloodloss fixing to them

## Why's this needed? 
Right now people get left for dead because they spend too long dying. see how this works.

## Changelog 

```changelog
(u)TDHooligan
(*)Janktank II can be used to quickly revive gang members who are in crit (instead of just corpses). They still get messed up by it though.
```


pgmzeta:
update HELP_MESSAGE_OVERRIDE

TobleroneSwordfish (approved):

PR #21796, opened 2024-12-22
Stun batons turn off themselves when not wielded and activating has a slight action bar

M-Earthfire:
[Game Objects][Balance]

## About the PR 

This PR introduces two changes to stun batons and stun canes:

- If the stun baton is not wielded in hand for more than 5 seconds, it turns off
- Turning on a stun baton requires a 0.5 second action bar that is not interrupted by moving

This change does not apply to the ntso baton or the one dropped by beepsky
 
## Why's this needed? 

The stun baton is a very powerfull melee weapon that very quickly ends encounters if pulled out in surprise mid-RP. This kind of behaviour is the expected gameplay loop of antags (see: all their stun gear or stealth poison options), but not of security. This PR aims to drastically cut down on surprise batons and gives people a warning if someone pulls out a baton in front of them. 

The 5 second deactivation timer gives enough wiggle room to get not punished by slipping or inventory management. 

the NTSO baton is not encompassed by this change because HOS-whitelisted players are trusted enough to not pull off such bullshit behaviour. Securitron batons are exception because of their use in the respectable mobs.

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Stun batons and canes are equipment with special safety switches that turn them off when they are not wielded for 5 seconds
(+)Activating a stun baton now requires a 0.5 second action bar that is not interrupted by moving.
```

### DIFF
--- a/code/modules/status_system/statusEffects.dm
+++ b/code/modules/status_system/statusEffects.dm
@@ -2362,6 +2362,25 @@
 		var/mob/M = owner
 		M.mind?.remove_antagonist(ROLE_MINDHACK, ANTAGONIST_REMOVAL_SOURCE_EXPIRED)
 
+/datum/statusEffect/beton_decharging
+	id = "baton_decharging"
+	visible = FALSE
+	unique = TRUE
+
+	onRemove()
+		. = ..()
+		if(duration <= 0 && istype(owner, /obj/item/baton)) // if timed out, we turn the baton off
+			var/obj/item/baton/the_baton = owner
+			var/check_failed = TRUE
+			if (ismob(the_baton.loc))
+				var/mob/baton_owner = the_baton.loc
+				if(the_baton in baton_owner.equipped_list()) //if we picked it up while we stashed it, it should not depower
+					check_failed = FALSE
+			if (check_failed)
+				the_baton.is_active = FALSE
+				the_baton.UpdateIcon()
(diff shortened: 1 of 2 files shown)


JORJ949:
I would say that the taser SMG/shotgun is probably more effective AND more used than the baton in these scenarios, hell I even use the lawbringer before the baton because then if they run I already have gun in hand.

flappybatpal:
What about batons put into storage turn off?
That's a partial change that doesn't affect them being dropped but slows down "draw and baton".

JORJ949:
After playing a bunch of SS14 I think this is fine but just let people move while turning on the baton, this is how 14 does it and I think this is great as if they do run it lets you turn your baton on mid-chase.

TDHooligan:
sec already get instantly robusted in classic, mostly due to not having any escalation to react to. this is just going to multiply the issue

cheekybrdy:
I just see people stamcritting with actual weapons or fire extinguishers without action bars then stunning, which is significantly worse. People would default to flashes for more mobile fights and sunglasses are already powergamed to death.

TobleroneSwordfish:
Closing due to general negative feedback, I don't think this is something that needs an action bar.
Also:
> the NTSO baton is not encompassed by this change because HOS-whitelisted players are trusted enough

This is not a good philosophy to balance things with.

PR #21803, opened 2024-12-23
Require mining access for mineral magnet controls and rockbox

pgmzeta:
[balance][game objects]
## About the PR 
* Require mining access to use the mineral magnet computer and rockbox UIs.
* Mining outpost access is added to the outpost magnet/rockbox; if you just have mining outpost access you can still use the entire outpost.

## Why's this needed? 
* It's *dangerous mining equipment*, the access should matter.
* Salvagers can pull up and heist an unstationed rockbox. Kill a miner first please.
* Don't allow literally anyone to change price settings by breaking/sneaking in.

## Changelog 

```changelog
(u)glowbold
(+)The mineral magnet computer and Rockbox require mining access to use.
```

### DIFF
--- a/code/modules/transport/pods/communications.dm
+++ b/code/modules/transport/pods/communications.dm
@@ -56,6 +56,9 @@
 		External()
 			for(var/obj/machinery/mining_magnet/MM in range(7,src.ship))
 				linked_magnet = MM
+				if (!linked_magnet.allowed(usr))
+					boutput(usr, SPAN_ALERT("Access Denied."))
+					return
 				ui_interact(usr)
 				return null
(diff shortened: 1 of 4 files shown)


pgmzeta:
apparently the mineral magnet computer was supposed to be requiring the CE to turn on cooldown override, but I think that's been broken from before I've started playing (?)

TobleroneSwordfish (approved):
Looks good, I wasn't aware the rockbox didn't have access locks. The cooldown override probably shouldn't require CE access the way it's currently written, but frankly it could be made a lot more interesting since there's very little reason not to use it right now.

PR #21804, opened 2024-12-23
WYSIWYG: Combat Targeting 

TDHooligan:
[BALANCE][PLAYER ACTIONS][EXPERIMENTAL]

## About the PR 
For the sake of melee attacks, players moving between tiles will count as on their previous tile until they reach the next tile.

This is an experimental PR, as it is very very hard to test. If fighting becomes too defensive, this could only apply until they've moved halfway. 

## Why's this needed? 
TL;DR:
**These guys can hit each other:**
image

**These guys cannot:**
image

**This change makes it so the latter can hit each other.**

----

Players currently instantly teleport to their desired location when moving, but visually look like they're sliding along.

As a result, high level combat is weird. New players expect melee to be close range, but most of the action happens ~2 tiles away where you can get a quick hit in then instantly retreat. 

For example:
Robuster takes a step and hits the newbie, instantly punching them from 1.99 tiles away.
Newbie then leads the Robuster, ready to counter-punch when the robuster arrives in front of them.
Robuster initiates a move backwards, but is visually adjacent to Newbie.
Newbie's attack whiffs from 1.1 tiles away, as Robuster is now moving away, and counts as two tiles away.

## Changelog 

```changelog
(u)TDHooligan
(*)Melee attack range checks have been adjusted, you can now more reliably hit players as they are moving away from you. Special attacks are unaffected for now.
```


FlameArrow57:
nice

Carton171:
This'll be interesting to see. I've always thought it was annoying that hitting anyone was so difficult due to the teleporting thing.

FlameArrow57:
Maybe scope for another PR, but would it also be possible to make this happen for projectiles? for same reasoning if it being awkward to hit someone who looks like they're on a tile but aren't

Valtsu0:
This still results in being able to hit 1.99 (and even 2.99 if both are moving) tiles away and not being able to hit 1.01 tiles away. It would be better to have it switch halfway

Unsure if this is an improvement over the current system (it probably is)

For the proposed projectiles, being able to dodge bullets is objectively cool and heavily relies on this

TDHooligan:
> This still results in being able to hit 1.99 (and even 2.99 if both are moving) tiles away and not being able to hit 1.01 tiles away. It would be better to have it switch halfway

what is the case where you can't hit 1.01 tiles away

Valtsu0:
> 
> what is the case where you can't hit 1.01 tiles away
> 

If i understand it correctly, the case where one person is moving next to the other person, but hasn't finished doing so. Now that i think about it, it's fine as long as both have queue combat clicks turned on.

TobleroneSwordfish:
> the case where one person is moving next to the other person, but hasn't finished doing so.

They can hit that person both before and after this change, all this does it make it permissive the other way too.
(Also queued combat clicks are a weird relic that should proobably be removed, especially if we make combat more consistent in more intuitive ways like this)

PR #21812, opened 2024-12-24
Increases micronukes boom size

IPling:
## About the PR 

Increases the boom size of micronukes by 5x of their original amount.
As well changes the descriptions to fit the increased explosion size.

## Why's this needed? 

Micronukes are rarely ever seen getting detonated, because of their rarity and the difficulty in actually defending them. Which causes the explosion size, comparable to a maxcap TTV, to be incredibly _disappointing_ for the amount of effort the traitor needs to put in to defend it. 

This PR isn't really to encourage people to try as hard as possible to detonate the nuke, as it can still act as an amazing distraction or bait. Rather it's for that one very rare shift, when it actually gets detonated, there is a really big firework show.

## Changelog 

```changelog
(u)IPingu
(+)Micronuke explosion size has been increased.
```


Tarmunora:
Micronukes were first and foremost a gimmick item, with them being added later to surplus crates as more of an easter egg than anything else.
The boom size being "delete a small department" is right where it should be for them

TobleroneSwordfish:
One thing I could see changing is maybe lowering the timer a bit? 5 minutes is a very long time.

ZeWaka:
You really need to fix your master dude

IPling:
> One thing I could see changing is maybe lowering the timer a bit? 5 minutes is a very long time.

I definitely thought of that as well, though I couldn't think of a proper time to set it to. 
As a definite strat would be to surround the micronuke with some reinforced walls, and I wanna give the crew *some* time to deconstruct them. Though I got no clue what the sweet spot would be for the timer.

> You really need to fix your master dude

I uhhh, will see about this on the discord.

PR #21813, opened 2024-12-24
Central Command Pager

JORJ949:
## About the PR 
Adds a pager to the NTSC for direct communication with central command.

TODO: 
- TGUI chatbox, similar to how consoles look
- An admin button in chat to reply directly to that message, like PDAs have
- Upgrade admin button to pull up a tgui of all pager's and their owners (or just one, if theres only one (90% of cases))
- Replying to the pager via discord bot (pretty sure a dev will have to do this)

## Why's this needed? 
After playing a bit of SS14 I learnt about the "Nanotrasen Representative" job, who essentially spends the round faxing centcom "The HoS ate too many donuts and entered a sugar coma" and I think this would be a fun way to make NTSC unique and less of a different secoff loadout.

## Changelog 

```changelog
(u)JORJ949
(*)Central Command has equipped all Nanotrasen Security Consultants with direct pagers to report on the station.
```


DisturbHerb:
i'm curious as to why us NTSCs/blueshield equivalents/whatevers should have them over other nanotrasen representatives or heads of staff

Chatauscours:
Yeah I agree with Rupert here. 
Why not one for NTSOs during events for emergency communications with admins.
But in normal rounds command already has the CC message option on commaster consoles with the same functionality, so I don't see why would the NTSC (or anyone else for that matter) get special treatment here.

cheekybrdy:
This sounds like something the clown should have, at least on a 1/1000 chance

PR #21814, opened 2024-12-24
fix minor visual issue with dispensers

TDHooligan:
[BUG]
## About the PR 
moves proc so that chemfabs update their lights on time (chemicals inserted when all lights turn off)


TobleroneSwordfish:
Are you sure this actually changes anything? Nothing here appears to sleep so it should be visually identical, no?

TDHooligan:
> Are you sure this actually changes anything? Nothing here appears to sleep so it should be visually identical, no?
when the reagents are dispensed, the lights are supposed to instantly set to zero. ATM they don't b/c of this ordering typo

TobleroneSwordfish:
Ohh I see, gotcha!

PR #21819, opened 2024-12-25
Add name hyphenation option to character preferences

FlameArrow57:
[FEATURE]

## About the PR 
Adds a Hyphenation option to the character preferences set up window.

If toggled, your appearance name will be set to firstName-middleName-lastName, or firstName-lastName if no middle name. The three names will still remain separate for security records or anywhere else

## Why's this needed? 
Popular naming option, currently it's only kind of doable by putting spaces in your names

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a "Hyphenate Name" option to the Character Setup window, which will cause your character appearance name to be set to "FirstName-MiddleName-LastName", or "FirstName-LastName" if no middle name. All three names are still used for purposes of security records and the like.
```


code/datums/preferences/preferences.dm
```diff
@@ -486,10 +488,16 @@ var/list/removed_jobs = list(
 
 				if (new_name)
 					src.name_last = new_name
-					src.real_name = src.name_first + " " + src.name_last
+					src.set_real_name()
 					src.profile_modified = TRUE
 					return TRUE
 
+			if ("update-hyphenation")
```
TobleroneSwordfish, line 495: Tiny nitpick: this should be called `toggle-hyphenation` as `update` implies it provides a value rather than toggling.

TobleroneSwordfish (approved):
🦎

TobleroneSwordfish (changes requested):
Oh wait you're missing the savefile read/write code, this setting won't actually get saved. See `code/datums/savefile.dm`

PR #21827, opened 2024-12-26
Makes the RD's HPD purple and adds purple crowbars to the science fabricator

cheekybrdy:
## About the PR 
Title

## Why's this needed? 
1 it was weird getting green crowbars from the sci fabricator.
2. I've seen RDs and CEs get their HPD's mixed up cause they are both engineering yellow.


FlameArrow57:
Green is the color used for manufactured tools, would suggest maybe not changing the color

cheekybrdy:
For the general fabricator, other departments such as engineering use departmental colours for printed tools

cheekybrdy:
I thought this was the case for medical too but apparently just doctor's roundstart ones only (does mean that no other medical member gets access to medical crowbars without stealing). imo it'd be better to step off the civvie colour and make them colourless in general fabs and make the department ones coloured but seems like another PRs topic

ZeWaka:
It's only starting tools that are colored with the department. See: engineering

code/obj/item/handheld_dispenser/dispenser.dm
```diff
@@ -430,3 +430,18 @@ ABSTRACT_TYPE(/datum/pipe_recipe/machine/binary)
...
+	if (src.destroying)
+		src.icon_state = "hpd-destroy-r"
+	else
+		src.icon_state = "hpd-place-r"
+
+	var/fullness = round(src.resources/src.max_resources * 100, 25)
+	if (fullness <= 0)
+		src.UpdateOverlays(null, "ammo")
+	else
+		src.UpdateOverlays(image(src.icon, "ammo-[fullness]"), "ammo")
```
TobleroneSwordfish, line 447: Don't copy paste this, either use `initial` to get the original icon state or add a var for it

code/datums/manufacturing.dm
```diff
@@ -254,6 +254,14 @@
...
 	category = "Tool"
 
+/datum/manufacture/crowbar/purple
+	name = "Crowbar"
+	item_requirements = list("metal" = 1)
+	item_outputs = list(/obj/item/crowbar/purple)
+	create = 1
+	time = 5 SECONDS
+	category = "Tool"
+
```
TobleroneSwordfish, line 264: If this is unused it shouldn't be added, afaik no research job starts with a crowbar?

TobleroneSwordfish (changes requested):

code/obj/item/handheld_dispenser/dispenser.dm
ZeWaka: So true fam

code/datums/manufacturing.dm
cheekybrdy: This item would be unobtainable, removed the manufacture datum though

code/obj/item/handheld_dispenser/dispenser.dm
cheekybrdy: I'm 30% sure what I've just pushed is what you meant, no idea

PR #21828, opened 2024-12-26
Fix up tour bots

flappybatpal:
## About the PR 
- Fix tour bots being unable to complete their route on Cogmap1 and Clarion.
- Add docking behaviour to reset the tour instead of just looping. If the dock is missing the bot will grump then loop. If there is no tour console the bot will dock and set autoeject.
- Improved tip behavior to minimise tour bots talking over their advice.

Checked Oshan, Cogmap1, Clarion.
Checked Luna (slightly different behavior, bot now goes to the dock) and void discount dan factory.

## Why's this needed? 
Tour bots get stuck and stand there repeating tips due to distance between tour beacons and a problem pathing through a windoor.

Existing issues not touched in this PR:
- Pathing issue with thindoors that broke pathing into cog1 medbay.
- Handling of tour beacon being missing or otherwise unpathable.
- Neat stuff can overlap with tour chat.
- Sometimes bots say their tour chat twice.

### DIFF
--- a/code/modules/robotics/bot/guardbot.dm
+++ b/code/modules/robotics/bot/guardbot.dm
@@ -3607,9 +3607,8 @@ TYPEINFO(/obj/item/device/guardbot_module)
 			if(master.emotion != desired_emotion)
 				master.set_emotion(desired_emotion)
 
-			if(state != STATE_AT_BEACON && state != STATE_FINDING_BEACON)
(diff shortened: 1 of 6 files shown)


flappybatpal:
Ew what did I screw up with the commits...

TobleroneSwordfish (approved):

PR #3017, opened 2020-12-10
Adds the Security Assistant job 

Carbadox:
## About the PR 
This PR adds the security assistant, a lower access and lower stakes security role that allows newer players to learn and get acclimated to security play without being bombarded with a bunch of responsibilities. The role emphasizes communication and investigation skills, rather than combat. 

The security assistant role has a minimal timelock (5 rounds) to serve as a very low barrier of entry, in order to encourage newer players to try out the role. However, the timelock for regular security officers has been raised (30 rounds, though this number can be debated and lowered if need be) so that people's introductions to security aren't so high pressure. There are two security assistant slots every round, which should hopefully enable a synergy of officers teaching assistants and trainees assisting officers where they can, without either party getting too overwhelmed. 

The security assistant: 
- Receives a special security token that is redeemable for a barrier, a forensic scanner, and a special security ticket writer device (sprite by Flaborized!)
- Receives a special security pouch with two pairs of handcuffs, two flashes, a whistle, and a frosted donut
- Spawns with a special security assistant jumpsuit (sprite by Flaborized!), fingerless gloves, and brown shoes
- Like regular security officers, spawns with security trainee, a clone disk, a badge, secHUDs, and a security headset
- Like regular security officers, cannot be an antag, and has been blacklisted from gamemodes such as gang and rev
- Has access to the security department, the brig, maintenance tunnels, and weapon and contraband permits
- Has lower pay
- Unlocks a special medal when they escape alive to Centcom

This PR also:
- Adds a paragraph on guidance for security assistants and their expected responsibilities in Space Law book
- Adds a special security alert module that now spawns in all security PDAs. When used, this module sends a "backup requested" PDA alert to the entire security department from the user's PDA
- Adds the special security ticket writer device to all maps' security departments
- Cleans up some weapons vendor code

## ToDo
These are thing I hopefully want to do but would greatly appreciated if others assisted.
- Add wiki page for security assistants
- Add possible Security For Beginners guide book

## Why's this needed? 
To provide an entry point to players interested in the security department without starting off in a high pressure job like security officer, while also expanding the security department to be more of a department and less 5 dudes with electrified sticks governed by a guy in a cape with a speaking gun.

Changelog will be updated to reflect what changes are currently considered finished
## Changelog

```
(u)Carbadox, Flaborized, and Flourish
(*)Added a new permanent security role, security assistant. This role will provide an entry point to players interested in the Security department without having to start off in a high pressure role like security officer. Please be nice to them!
[...]

PR #21833, opened 2024-12-27
Removes the security alert HUD button

TobleroneSwordfish:
## About the PR 
Removes the security alert HUD button from security PDAs. The alert can still be sent from the menu, the same way as any other PDA.

## Why's this needed? 
This feature was added a few years ago without any particular justification (#3017 I think?) and seems to be generally disliked for promoting the well known security swarm descending on a location the instant one of their officers is in trouble.

You can always call for help the old fashioned way, but hopefully this will lead to more opportunities for antags to take out isolated security officers without them suddenly turning into [Cuccos](https://zelda-archive.fandom.com/wiki/Cucco).

## Changelog 

```changelog
(u)LeahTheTech
(*)Removed the security alert HUD button.
```

### DIFF
--- a/code/obj/item/device/pda2/modules.dm
+++ b/code/obj/item/device/pda2/modules.dm
@@ -288,7 +288,6 @@ TYPEINFO(/obj/item/device/pda_module)
 	desc = "A PDA module that lets you quickly send PDA alerts to the security department."
 	icon_state = "pdamod_alert"
 	setup_use_menu_badge = 1
-	abilities = list(/obj/ability_button/pda_security_alert)
 	var/list/mailgroups = list(MGD_SECURITY)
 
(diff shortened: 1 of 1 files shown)


JORJ949:
Should probably give the security pda module either a new ability or remove its accessibility because its now essentially worthless

NightmareChamillian:
Maybe remove it on RP? This is an absolute necessity when it comes to surviving classic rounds but I can see how it's bad for the roleplayers

Glamurio:
As mentioned in the thread, it would be nice if Security Assistants would get a little pager device that retains this function, and perhaps make said device a purchase option from vendors. That way Assistants aren't such juicy targets and newer players get to keep the button (which is why the button was made in the first place)

PR #21856, opened 2024-12-29
Cleans up brig access

Chatauscours:
## About the PR 
This PR does 3 things:
- Adds brig accesses for all doors leading directly to a cell or a courthouse detention box, removes it from everywhere else.
- Changes access on brig timers from security to brig.
- Removes brig access from HoPs because they don't really need it.

## Why's this needed? 
Currently brig access are used haphazardly across maps, this cleans it up a little bit.

## Changelog 

I don't think it's needed, this is more polish than anything.


JORJ949 (approved):
I've been meaning to do this for a while, glad someone got around to it.

PR #21859, opened 2024-12-29
Gives a firearms carry permit to bartenders

Chatauscours:
## About the PR 
Gives a firearms carry permit to bartenders.

## Why's this needed? 
Firearms carry is never used separately from contraband carry and I find it sad.
Also bartenders spawn with the shotgun, probably should have a permit issued by NT.
And gives a bit more spice to treasonous bartenders.

## Changelog 

```changelog
(u)Chatauscours
(*)After some slight bureaucratic delays, the bartender has finally managed to acquire themselves a firearms carry permit.
```

### DIFF
--- a/code/procs/access.dm
+++ b/code/procs/access.dm
@@ -317,7 +317,7 @@
 		if("Chef", "Sous-Chef")
 			return list(access_kitchen)
 		if("Bartender")
-			return list(access_bar)
+			return list(access_bar, access_carrypermit)
 		if("Waiter")
 			return list(access_bar, access_kitchen)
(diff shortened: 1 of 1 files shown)


TDHooligan:
barkeep is supposed to keep the shotgun under wraps. it's probably better to not standardise brandishing it among newer players who find out they have a gun

itchybones:
> barkeep is supposed to keep the shotgun under wraps. it's probably better to not standardise brandishing it among newer players who find out they have a gun

I think that it'd be more likely for a new bartender player to use their gun when there are yelling police robots chasing after them. But due to the unpredictable nature of newbie shenanigans, I'd suggest a testmerge.

PR #21865, opened 2024-12-30
Solidifies Prettier changes to `AlertModal.tsx`

DisturbHerb:
## About the PR 

prettier keeps modifying this file after i build the bundle. might as well lock it in so i don't have to keep dealing with it

## Why's this needed? 

\>:(


ZeWaka:
CI passes already

PR #21868, opened 2024-12-30
Gang lockers hide under floorboards

TDHooligan:
[EXPERIMENTAL][GAMEMODES][BALANCE]
## About the PR 
Gang lockers now hide under the floorboards. Gang members can unhide them by emptyhanding the floor tile. They can hide them with the same action, or walk way, whereby the locker will hide itself.

Should be testmerged for a bit - I'm not certain how security players will react to this.

## Why's this needed? 
Security will sweep maintenance once for gang lockers and from that point on, there's no secrecy to gang operations. The gangs have no way to handle this.

The exact knowledge of their clubroom at all times also leads to a lot of unhealthy behaviours. Security cause excessive friction with gangs by staring at the locker etc. This PR introduces a lot of uncertainty, and having some level of uncertainty allows gang members some breathing room.

Prudent gang members will be able to keep their clubroom clear & misdirect security. Messier gangs will be more engaging to investigate, as you can't just look for the big "THIS IS A GANG ROOM" locker.

## Changelog 

```changelog
(u)TDHooligan
(*)Gang lockers now hide under the floor when not in use. 
(*)Gang members can unhide their locker by clicking the floor with an empty hand.
```


DisturbHerb:
as a regular security player, please for the love of god make this a thing

Chatauscours:
Yeah this seems like a great idea

code/datums/controllers/process/gang.dm
```diff
@@ -44,14 +44,23 @@
...
 	setup()
-		name = "Gang_Money_laundering"
+		name = "Gang Locker Tick"
 		schedule_interval = GANG_LAUNDER_DELAY
 	doWork()
 		for_by_tcl(locker, /obj/ganglocker)
 			if (!locker)
 				return
+			if (!locker.is_hiding)
+				var/should_hide = TRUE
```
TobleroneSwordfish, line 56: This var doesn't actually do anything because the loop returns if it finds a ganger

code/datums/gamemodes/gangwar.dm
```diff
@@ -2162,6 +2165,72 @@ proc/broadcast_to_all_gangs(var/message)
 			text += "<b>[i]: [scores[i].gang_name]</b></br>"
 		boutput(user, SPAN_ALERT(text))
 
+	proc/turf_attacked(target, attacker)
+		if (!ismob(attacker))
+			return
+		var/mob/user = attacker
```
TobleroneSwordfish, line 2171:
```suggestion
	proc/turf_attacked(target, mob/user)
		if (!ismob(user))
			return
```

code/datums/gamemodes/gangwar.dm
```diff
@@ -2162,6 +2165,72 @@ proc/broadcast_to_all_gangs(var/message)
...
+		if (user.get_gang() != src.gang)
+			return
+		if(!isalive(user))
+			return
+		if(!isliving(user))
```
TobleroneSwordfish, line 2176: Maaybe also make sure they're not a silicon so borged gangers can't reveal it?

code/datums/gamemodes/gangwar.dm
```diff
@@ -2162,6 +2165,72 @@ proc/broadcast_to_all_gangs(var/message)
...
+		var/turf/floorturf = get_turf(src)
+		is_hiding = TRUE
+		if (floorturf.intact)
+			animate_slide(floorturf, x_coeff * -slide_amount, y_coeff * -slide_amount, 4)
+		SPAWN(0.4 SECONDS)
+			if(src)
+				src.layer = BETWEEN_FLOORS_LAYER
+				src.plane = PLANE_FLOOR
+				src.mouse_opacity = 0
+			animate_slide(floorturf, 0, 0, 4)
```
TobleroneSwordfish, line 2233: Please don't copy paste these, either make an animation library proc or at least make this a single method with arguments.

TobleroneSwordfish (changes requested):
Very cool PR, a few small code issues.

code/datums/controllers/process/gang.dm
TDHooligan: good spot. brain farded

PR #21885, opened 2024-12-31
Seclight

JORJ949:
## About the PR 
Adds a security flashlight to sectechs.
image
The security flashlight is immune to emagging (who emags flashlights anyway?)

## Why's this needed? 
If you don't take a helmet you are often left without a flashlight as security and they can be hard to randomly find around maintenance.

## Changelog 

```changelog
(u)JORJ949
(*)Added security flashlights to sectechs.
```


itchybones:
Cool item! Does it have unique inhands or the same as the normal blue flashlight?

JORJ949:
> Cool item! Does it have unique inhands or the same as the normal blue flashlight?

I forgot to make inhands :/, I believe it won't have any but will make some.

Kyle2143:
It's a cool item concept, but I think it's a little boring for a "special security flashlight" considering you can just print flashlights in general manufacturers and find them around the station. 

There were some concerns over the sprite looking too similar to a stun baton and it looking like a classic "American cop flashlight" that would be heavy enough to use as a baton. Which also doesn't fit with it's stats since it's just a subtype of a normal flashlight,

 In terms of the sprite, being oriented horizontally instead of diagonally. Like, that size shouldn't fit in pockets too, since it's about the same size as batons, but had the flashlight weight class.

I do like the idea of giving security officers a special flashlight though, the design makes me think of those old-ish flashlights that my dad had when I was a kid where you screwed the head on to turn the light on as well as adjust the arc length of the light from wide to narrow. I think it would be cool if it had that feature or something like it.

PR #21892, opened 2025-01-01
AI core will now slam to the pool floor when using a springboard

AnomalousPotato:
[Feature][Silicons][Player Actions]

## About the PR 
When using the springboard in the pool, the AI core will now slam into the floor and take damage.

## Why's this needed? 
heavy block go smash


TobleroneSwordfish (approved):
excellent PR

PR #21896, opened 2025-01-01
Add pod lateral thrusters component

FlameArrow57:
[VEHICLES][FEATURE]

## About the PR 
Adds a new manufacturable pod component, Lateral Thrusters, for 2 dense metal, 1 conductive materal, and 2 energy material

Lateral thrusters move your pod in a burst of 5 tiles in the toggled direction, right or left

The HUD button will activate it, and the thrusters direction is changed through a context menu button

https://streamable.com/04u6qn

## Why's this needed? 
More varied pod components/combat

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new lateral thrusters pod component to the pod manufacturer. Thruster direction can be changed through clicking the pod and pressing the context menu button. Activation has a 5 second cooldown.
```


Tarmunora (approved):
I like the concept, but the activation method feels a bit janky.

FlameArrow57:
> I like the concept, but the activation method feels a bit janky.

Context menu button for switching thruster direction sound better? Then clicking HUD button does the action

Tarmunora:
> > I like the concept, but the activation method feels a bit janky.
> 
> Context menu button for switching thruster direction sound better? Then clicking HUD button does the action

Yeah, that seems a lot more reasonable

code/modules/interface/multiContext/context_actions.dm
```diff
@@ -697,6 +697,22 @@
...
 			V.use_external_speaker()
 
+	change_thruster_direction
+		name = "Lateral Thruster Direction"
+		desc = "Change the lateral thrusters to move the ship left"
+		icon_state = "thrusters_left"
+
+		execute(atom/target, mob/user)
+			..()
+			var/obj/machinery/vehicle/V = target
```
Tarmunora, line 707: this should have a `checkRequirements()` override to only show if the system is installed, no?

Tarmunora (changes requested):

code/modules/interface/multiContext/context_actions.dm
FlameArrow57: It's already on the base type I think

FlameArrow57:
Actually I need to fix another thing with this first

FlameArrow57:
Okay fixed, ready for re-review

Issue #21026, opened 2024-10-09
Support for non-hardcoded customizationHolders


Tarmunora:
Currently, customizationHolders are implemented as drop-in replacements for the old customizations variables, and so (almost) everything that interacts with them is hardcoded to expect `hair_top`, `hair_middle`, and `hair_bottom`.
These will pretty much all need to change to support dynamic customizationHolder lists

PR #21900, opened 2025-01-01
Resprite Double Braids

jan-antilles:
## About the PR 
Resprites the Double Braids, `and changes the name of the state in the code to "doublebraids" instead of "indian." ` (nevermind, that breaks character saves whoops)

image

## Why's this needed? 
We learn better, we do better, we add more dimension to the sprite.

Well, we add more dimension to the sprite anyway!

## Changelog 

```changelog
(u)Jan.antilles
(+)Resprites the Double Braids style to add more shading and dimension.
```


code/datums/customization_styles.dm
```diff
@@ -737,9 +737,9 @@ TYPEINFO(/datum/customization_style/hair/gimmick)
 				name = "Croft"
 				id = "croft"
 				gender = FEMININE
-			indian
+			doublebraids
 				name = "Double Braids"
-				id = "indian"
+				id = "doublebraids"
```
TobleroneSwordfish, line 742: This will break existing profiles that use this hairstyle, ids are meant to be immutable because they're used in stored savefiles. If you want to change it you'll need to include a conversion snippet somewhere in appearance loading code.

TobleroneSwordfish (changes requested):

code/datums/customization_styles.dm
Glamurio: Miiiight be better to just do this alongside a big save conversion that's gonna be happening as part of #21026? I think having lots and lots of conversion snippets is not good, and if we're already planning on one, might as well lump it together. Pinging @Tarmunora for input.

jan-antilles: > This will break existing profiles that use this hairstyle, ids are meant to be immutable because they're used in stored savefiles. If you want to change it you'll need to include a conversion snippet somewhere in appearance loading code.

Yup, good call. I'll let it get changed in a future fix!

PR #21904, opened 2025-01-02
Add pod shielding components

FlameArrow57:
[VEHICLES][FEATURE]

## About the PR 
Adds two new pod components, Light Shielding System and High Impact Shielding System, to pod manufacturers

These are shielding systems that provide a % dmg reduction up to a certain amount against projectiles and explosions, timed over a duration, with recharge between uses

Light shielding system
-25% dmg reduction up to 100 dmg
-Lasts 10 seconds with recharge of 60 seconds
-Requires 0.5 metal, 1 conductive, and 3 high energy mats to make

High impact shielding system
-90% dmg reduction up to 1000 dmg
-Lasts 3 seconds with recharge of 120 seconds
-Requires 0.5 metal, 1 highly conductive, 2 dense crystal, and 3 extreme energy mats to make

## Why's this needed? 
More varied pod combat options

## Changelog 

```changelog
(u)FlameArrow57
(*)Added Light and High Impact Shielding System pod components to pod manufacturers. They provide timed life shielding with a % dmg reduction against projectiles and explosions.
```


code/modules/transport/pods/secondary_system.dm
```diff
@@ -1235,3 +1235,82 @@
...
+		..()
+
+	// takes incoming damage "dmg", returns damage dealt to pod
+	proc/process_incoming_dmg(dmg)
+		var/dmg_dealt = dmg * (1 - src.block_pct)
+
+		if (src.life < dmg_dealt)
+			dmg_dealt += dmg_dealt - src.life
+
+		src.life -= dmg_dealt
```
Tarmunora, line 1293: Feels a bit unintuitive that the life of the shield is decreased by the post-reduction amount - means that the low-impact shield turns off after 133 raw damage, and the high-impact one after 10000, despite having a life of `100` and `1000` respectively

FlameArrow57: I think I had the math wrong or that was unintended... LMK if it looks good now

PR #21916, opened 2025-01-03
Support backslash escapes in mechcomp packets, unify signal splitting

patricia-izin:
[Feature] [station systems] [add to wiki]

## About the PR 

When writing parameter value lists for packets with mechcomp, you can now use a backslash on special characters instead of writing out them out as a hex code in a html escape or %xx escape.
For example \\& \\; to use & or ; in signal data instead of as a parameter-value pair delimeter.
Additionally semicolons now work correctly when adding lists to packets with parenthesis.
Wifi now uses the same signal parser as powernet, params2complexlist, as well as the signal splitter.

image

## Why's this needed? 

Basic QoL for crafting packets with mechcomp, better consistency handling packets the same way on both components.
Signal splitter also made consistent for e.g. nesting parenthesis escapes.

## Changelog 

```changelog
(u)patricia
(+)Backslashes now work as escape characters in mechcomp networking components and the signal splitter
```


patricia-izin:
gotta look at wifi

patricia-izin:
Okay that was just a tweak

patricia-izin:
I tested a few networked objects to make sure nothing exploded, and I verified the escapes work in real world use cases by hacking the APCs with only

`sender=02002bb3;address_1=02003a84;command=term_message;data=command=setmode\&equip=0\&environ=0\&light=0;`

_std/lists.dm
```diff
@@ -295,10 +295,23 @@ proc/params2complexlist(params)
...
+	var/escapee = null
+
 	while(a < len)
-		a++
+
+		//Replace characters escaped with a backslash
+		if(findtext(params,"\\",a,a+1))
+			escapee = params[a+1]
+			escapee = text2ascii(escapee)
+			if(escapee <= 126) //Only ascii, a special character will never be unicode (famous last words)
```
patricia-izin, line 306: Text encodings get weird around this boundary, you could say that I'm filtering extended ascii, which may or may not be the same code as an equivalent unicode character....

Either way, we don't have any need to escape anything after 126

Tarmunora (approved):

_std/lists.dm
ZeWaka: what about the early codes, pre 32

patricia-izin: Might as well block non-printable characters from being escaped yeah, done

PR #21919, opened 2025-01-03
Add Automatic Repair System pod component

FlameArrow57:
[VEHICLES][FEATURE]

## About the PR 
Adds a new pod component, the Automatic Repair System, to pod manufacturers

The component has a max reagent capacity of 100u. When filled with welding fuel and turned on, it will repair any damage the pod has at a rate of 15 hp every 4 seconds, for 1u welding fuel (welding tool does 60 hp/sec repair). It will not repair while the pod is in or recently was in combat.

Intent is that this is a repair component meant for extended space travel in dangerous territory.

Costs 2 dense metal, 3 conductive material, and 1 energy material.

## Why's this needed? 
More pod components

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new Automatic Repair System pod component to pod manufacturers. When filled with welding fuel and turned on, it will repair damage to your pod over time.
```


pgmzeta:
may need some special handling for nadir acid

Tarmunora (approved):

PR #21941, opened 2025-01-05
Forensics Rewrite

LorrMaster:
[CODE QUALITY] [EXPERIMENTAL] [INPUT WANTED]

## About the PR 
(Note: this project is still a WIP) The main feature of this PR is the creation of forensic holders, which are meant to encapsulate all the evidence related to the history of an object. These holders contain a list of groups that specify how different types of evidence should behave, and then finally the forensics data which store information on a specific event that occurred with that object. It also allows IDs related to forensics to be stored by reference rather than by value with the idea of reducing time spent on copying and comparisons.

The PR also comes with a number of new types of forensics evidence to test and showcase this new system. Each piece of evidence can be given a timestamp to show when it occurred. Fingerprints can be stored with a mask that only displays partial prints. DNA can be given a form that specifies the difference between evidence coming from blood, hair, or saliva. IDs can be tied to individual organs and paired together, allowing for retina scans and footprint pairs that can be changed by swapping out one or both organs/limbs. You can also pair different types of evidence together. Objects can report their current state to a forensics scan directly or point to other objects that should be included in the report (such as worn clothing).

A list of features currently included:
- Scanner IDs left behind during scans and teleportation (teleportation only with cargo transporters for now)
- Evidence tied to organs, such as retina scans, footprints, and should probably be extended to fingerprints
- Mark fabricated, imported, and photocopied items to show how they were created
- A new forensics training trait for detectives and security assistants. This trait allows a player to see timestamp estimations for each piece of evidence. The accuracy of these time estimations can vary based on multiple factors.
- Scan a safe to reveal what the first character of its combination is.

TODOs:
- [ ] Most of the old forensics stuff is still there in some form and needs to be fully replaced.
- [ ] Blood is only half working at the moment and still needs a decent amount of work to replace.
- [ ] Update security logs to use the new system. Not sure how in-depth this should be.
- [ ] Forensic scan reports need formatting improvements.
- [ ] Store types of IDs into a dictionary and check for duplicates
- [ ] Fingerprints tied to limbs.
- [ ] Update the fingerprinter. Can be reworked later.

ss13_forensic_1
ss13_forensic_2
ss13_forensic_3

## Why's this needed? 
Creates a structured system for creating, storing, and passing around forensics data. Makes forensics significantly easier to work with in general.

## Changelog 

```changelog
(u)LorrMaster
(*)Forensics has been overhauled
```


LorrMaster:
Woops, not sure what I did here.

FlameArrow57:
Is this still being worked on?

LorrMaster:
Yes, I closed the PR draft by accident. I'll reopen it once it is closer to completion.

PR #21946, opened 2025-01-06
Round final payout calc to an even number for requisitions and artifacts

pgmzeta:
[qol][station sytems]
## About the PR 
round the final payout calculation of req contracts

## Why's this needed? 
no fractional credits


AnonymousUserTheOnly:
honestly, this might need to have some checks for not being an even number as well.  most of the time, a QM will split half the value into their account so they can keep making major profit.  and let's also not forget that on Cogmap 2 and Kondaru, artsci gets it's own barcode printer, which means the scientist on artifacts is going to be using that to pad their wallet off artifact sales.  half of one is still a half-credit, and if we stop half-credits on reqs we have to also stop it on artis.  easiest way to stop is probably making every single one of the payouts be forced to be an even number.  all even numbers means no half-credits and no need for rounding.

PR #21956, opened 2025-01-07
Refactor control unit execution process

patricia-izin:
## About the PR 
[Station Systems]

Refactors the way the control unit executes, using the new process_fast tier of the mechanics process in #21548

The control unit executed 32 instructions every 2.9 seconds, a clock program could output 5 signals in this time, which would be 0.1s apart, for an effective rate of 0.56s per process cycle. It would stop execution for the remaining 2.4s or so.

I've removed the SPAWN() and put the control unit in the 0.4s tier of the mechanics process, and reduced the instructions per execution to 6. Additionally now when a signal is output by the control unit, I end the current execution, so that the next output will be at least 0.4s apart instead of 0.1s

Observe the old behavior below with a clock program driving a counter

dreamseeker_AsiahN5fSo

And the improved behavior

dreamseeker_AxF6qiKFdc

## Why's this needed? 

The original behavior was very stuttery and it was weird that you could output signals so fast but only for a moment. 

## Changelog 

```changelog
(u)patricia
(+)Control units now output signals and execute programs at a more consistent rate
```


patricia-izin:
(this won't build until the other one is merged)

patricia-izin:
this is now ready to go

PR #21957, opened 2025-01-07
Output GPS coords in signal list format

patricia-izin:
[Feature] [Add to wiki] [Station systems]

## About the PR 

Outputs x and y as their own signals on GPS units and tracking implants

image

## Why's this needed? 

Signals in weird formats are hard to parse this just standardizes it so that components like the signal splitter and the dispatch module and whatnot can access this data

## Changelog 

```changelog
(u)patricia
(+)GPS coordinates in packets are now in x=, y= format compatible with signal splitters
```


patricia-izin:
Wiki updated

PR #21958, opened 2025-01-07
Adds mulitcast addres_tag to GPS

patricia-izin:
[Feature] [Station systems] [Add to wiki]
## About the PR 

Allows GPS units to be queried in bulk with address_tag=GPS

## Why's this needed? 

The actual handheld GPS units always show the location of every other (trackable) GPS so you should be able to get this information with packets too

## Changelog 

```changelog
(u)patricia
(+)GPS units can now have their locations (packet) queried in bulk with address_tag=GPS
```


patricia-izin:
.

PR #21963, opened 2025-01-08
stop gang tip-offs for cryo

TDHooligan:
[BUG][GAMEMODES]
## About the PR 
stops gang members being tasked with interrogating people in cryo


code/datums/gamemodes/gangwar.dm
```diff
@@ -845,6 +845,8 @@ proc/broadcast_to_all_gangs(var/message)
 				continue
 			if (isnull(M.current.loc)) //deleted or an admin who has removeself'd
 				continue
+			if (M.hibernating) //cryod
```
TobleroneSwordfish, line 848:
```suggestion
			if (M.current.hibernating) //cryod
```

PR #21967, opened 2025-01-09
Remove maintenance areas from possible gang duffle objectives

pgmzeta:
[balance][gamemodes]
## About the PR 
* Remove maintenace areas from potential duffle bag areas
* Remove unused variable `smaller_departments`
* Add main podbay as potential duffle bag area

## Why's this needed? 
Gangs should be engaging with areas crew is more likely to be in, not rewarded for skulking in lesser-used maint tunnels.

### DIFF
--- a/code/datums/controllers/process/gang.dm
+++ b/code/datums/controllers/process/gang.dm
@@ -182,14 +182,6 @@
 		possible_departments += /area/station/science/lobby
 		possible_departments += /area/station/crew_quarters
 		possible_departments += /area/station/hydroponics
-		possible_departments += /area/station/maintenance/north
-		possible_departments += /area/station/maintenance/northeast
-		possible_departments += /area/station/maintenance/northwest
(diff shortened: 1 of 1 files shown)


TDHooligan:
the maints are in there mostly to bulk out the objective count, as they can start to run out when 3 gangs get 4+ objectives. 

i'd rather have more places i missed added to lower their odds than remove them entirely. you can always get easy spief objectives etc.

pgmzeta:
What about having them as classic-only, considering RP gang limit is 2?

TDHooligan:
i'd still rather they be there - they're larger and take longer to tag, but since they're maints i've seen vandalism turn into gang fighting more than once. 

adding some more populated areas would have the same effect as this PR, while keeping the benefits of the occasional maint fight.

pgmzeta:
i'll see what can be added and update this PR

PR #21973, opened 2025-01-09
output nuclear turbine power in arithemetic component compatible format

patricia-izin:
[bugfix] [station systems]
## About the PR 

image

Fix the nuclear turbine's power signal so that the arithmetic component can accept it.
## Why's this needed? 

The number is supposed to be usable in mechcomp, and it works from the TEG etc (that's where I got this)

## Changelog 

```changelog
(u)patricia
(+)The nuclear turbine's power signal now works correctly with mechcomp arithmetic components
```

### DIFF
--- a/code/obj/nuclearreactor/turbine.dm
+++ b/code/obj/nuclearreactor/turbine.dm
@@ -239,7 +239,7 @@
 
 			src.network1?.update = TRUE
 			src.network2?.update = TRUE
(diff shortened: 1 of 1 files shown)


patricia-izin:
I still had to multiply the power by 2, but multiplying it worked, which is the important part. I assume there’s some accounting discrepancy between the laser/teg/ptl with power per tick accounting but that is another PR

amylizzle (approved):

PR #21974, opened 2025-01-09
specials now also WYSIWYG

TDHooligan:
[FEATURE][BALANCE][OBJECTS]
## About the PR 
Applies the same fundamentals of #21804 to special attacks.
Players who are moving will count as in both their current and previous tiles for special attacks, until the movement is complete.

## Why's this needed? 
Originally #21804 was supposed to be for special attacks. This is to make it consistent with left clicking. 

There is an upcoming PR in the works to provide Skilled Gamer with a new tactic to dodge WYSIWYG

## Changelog 

```changelog
(u)TDHooligan
(+)Special attacks now hit moving targets better.
```


ZeWaka:
WYSIPCTWOACG

(WHAT YOU SEE IS PRETTY CLOSE TO WHAT OTHER ATTACKS CAN GET)

code/modules/items/specials/ItemSpecials.dm
```diff
@@ -100,6 +100,19 @@
 
 	return ((dx*dx) + (dy*dy))
 
+/// Finds every mob that is currently moving away from a turf, but has not reached the end of their movement.
+/proc/atoms_in_combat_range(var/turf/target)
+	var/list/atom/atoms = list()
+	for(var/atom/A in target)
+		atoms += A
+	for(var/atom/A in range(1,target))
+		if(ismob(A))
```
Tarmunora, line 109: Filter for mobs in the loop statement instead of looping over atoms and then filtering to mobs

code/modules/items/specials/ItemSpecials.dm
```diff
@@ -100,6 +100,19 @@
...
+			var/mob/dude = A
+			if (dude.next_move > world.time && dude.prev_loc == target)
+				atoms += A
```
Tarmunora, line 112:
```suggestion
				atoms |= A
```

Tarmunora (dismissed):

PR #21976, opened 2025-01-10
Add artifact reticulator

FlameArrow57:
[SCIENCE][GAME OBJECTS][FEATURE][BALANCE]

## About the PR 
This PR adds a new artifact machine, the Artifact Reticulator

This machine has 3 uses - breaking down artifacts into spendable points, creating new artifact technologies, and modifying existing station items

1 Breaking down artifacts
- Artifacts can be inserted into the machine and broken down into either an Essence shard, Power shard, or Spacetime shard depending on what its theme is. Artifact activator keys are broken down into Omnishards.
- If a second activated and correctly labeled artifact is inserted, and they are of different shard categories, you can a get a Fusion shard
- For an artifact to yield a shard, it must be labeled correctly with an artifact form. Otherwise, it will be destroyed completely

2 Creating new technologies
- The shards from breaking down artifacts can be spent to create new technologies
- Artifact resonator - Tells you if an artifact has a disguised origin, and 3 possible activation methods with 1 being the actual one. If used on an activated artifact, tells you its activation method
- Artifact tuner - A one-time use item that rerolls a random fault of an activated artifact, with a low chance to instead add more faults. If used on an artifact with no faults, its guaranteed to add a fault
- Recreation - Recreates a previously reticulated artifact, with new stats
- Combine artifacts - If loaded with a second artifact, the effects of the two artifacts will be combined if compatible (ex. you could create a prison art that injects you with chemicals, a plant watering artifact that heals you, etc). 

3 Modifying existing station items
- The shards can also be used generally to give modify or give slight stat increases to existing categories of station items
- Light imbuement - Adds ambient light to a station object, with a color of your choosing
- Material breakdown - Destroys an item, giving materials used to make it (reverse ruck-kit)
- Material modification - Given an item with a material, lets you change a property of its material by a value of 0.5
- Storage upgrade - Adds another storage slot to a HUD storage item
- Reagent capacity - Increases max reagent capacity of a reagent holder by 10%
- Cell capacity - Increases cell charge capacity of a power cell by 10%
- Mining power - Increases mining power of a custom mining tool by 10%

-Help needed on spriting

Sounds used, all under [CC0](https://creativecommons.org/publicdomain/zero/1.0/):
https://freesound.org/people/deadrobotmusic/sounds/750607/
https://freesound.org/people/Mosrod/sounds/371462/
https://freesound.org/people/Windows9inety5ive/sounds/639735/

## Why's this needed? 
Would be cool + adds to the "research" part of artlab

See design doc for original source of idea - https://hackmd.io/@FlameArrow57/B1iOrzdfyl

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new artlab machine, the Reticulator. This is a machine used to break down artifacts for spendable points, which can be spent on creating new artifact technologies and modifying existing station items. See minor changelog for details.
(+)Artifacts can be broken down into 1 of 5 categories of spendable "shards", as long as they are labeled with a form correctly. Otherwise, they will be fully destroyed.
(+)Shards can be spent on creating a Resonator, Tuner, recreating a previously broken down artifact, or combining artifacts into one.
(+)Shards can also be spent on modifying existing station items in 1 of 7 ways, such as increasing a reagent holder's capacity by 10% or adding another slot to a storage item.
```


LorrMaster:
I like expanding artifact uses, but I don't like the idea of being able to reroll them outside of trading it in for a brand new artifact. Reason being that if you are researching them, you probably don't have the knowledge to manipulate them to that degree. Two artifacts that look the same to you could be completely different things to their creators. I do love the idea of recreating and combining artifacts though. I think the direction should be more about expanding the uses of what you roll with. Such as having a darkness artifact blind half the station or using a borger to turn biomass into fabricator minerals.

FlameArrow57:
> I like expanding artifact uses, but I don't like the idea of being able to reroll them outside of trading it in for a brand new artifact.

Re-rolling here is just referring to the stats of an existing artifact, so it might change explosive power of a bomb artifact

FlameArrow57:
> I like expanding artifact uses, but I don't like the idea of being able to reroll them outside of trading it in for a brand new artifact.
Rerolling here is referring to just the existing stats of an artifact, so you can re-roll an injector's chemicals or create an explosive bomb with different power, if that's what you were meaning?

Spyritdragon:
I feel like it's worth considering how easy this might make it to dispose of artifacts that are actively being harmful. Maybe artbombs should instantly trigger if you attempt to disassemble them, or the reticulation process should take some time so that whatever artifact is in there's effect has to be actively dealt with while it's processing?

I may also just have read straight past it, but consider that artifacts can also contain humans or borgs - should probably have a warning or something stopping you if you try to reticulate someone along with the artifact.

FlameArrow57:
> I feel like it's worth considering how easy this might make it to dispose of artifacts that are actively being harmful. Maybe artbombs should instantly trigger if you attempt to disassemble them, or the reticulation process should take some time so that whatever artifact is in there's effect has to be actively dealt with while it's processing?
> 
> I may also just have read straight past it, but consider that artifacts can also contain humans or borgs - should probably have a warning or something stopping you if you try to reticulate someone along with the artifact.

Yeah will try to work on something regarding the ones that are actively dangerous. Might make it so the artifacts don't need to be activated to break them down, but then the ones like bombs if activated can't be broken down

Yeah I'll add something for the ones with people in them, thanks

flappybatpal:
My concerns are that this could demistify artifacts and breaking down artifacts might break the gameplay loop of identifying artifacts, sending the useful ones to departments and selling the unused ones to cargo.
If science can break down and forge artifacts they may just skip the above entirely and focus on power gamer artifacts or bombs in the hope of getting super sleeper agent.

Based on the above my personal take is that instead of breaking down artifacts shards could be interesting if found in loot/mining/prefabs and were still unpredictable enough to prevent them being a power gamer option to make super bombs.
Resonator and combiner both sound too powerful as-is.

Modifying items I am unsure of considering the unknown balance implications.

FlameArrow57:
> My concerns are that this could demistify artifacts and breaking down artifacts might break the gameplay loop of identifying artifacts, sending the useful ones to departments and selling the unused ones to cargo. If science can break down and forge artifacts they may just skip the above entirely and focus on power gamer artifacts or bombs in the hope of getting super sleeper agent.
> 
> Based on the above my personal take is that instead of breaking down artifacts shards could be interesting if found in loot/mining/prefabs and were still unpredictable enough to prevent them being a power gamer option to make super bombs. Resonator and combiner both sound too powerful as-is.
> 
> Modifying items I am unsure of considering the unknown balance implications.

That is part of the purpose, to add something to the gameplay loop rather than "breaking it". Breaking down artifacts can only be done on activated artifacts and correctly labeled artifacts. The reason why I coded any of this was because I asked yass

image

PR #21977, opened 2025-01-10
Medical mastery - Medchem adjustments

TDHooligan:
[WIKI][REWORK][GAME OBJECTS][CONTROVERSIAL]

## Foreword
A lot of players have raised concerns that this PR is built to make it harder to be a doctor. Which isn't the goal. 
As a result - I've put some more work into explaining the purpose of each change. Every non-styptic change comes with a comment on how I intend for it to help fill the menders' void.

The goal of this PR is two-fold:
- keep the skill floor (AKA, the 'minimum skill' required to be effective) as close to live as possible. Newbie doctors should still be able to treat people and feel helpful.
- Raise the skill ceiling (AKA, the 'maximum skill' possible for the role), so that skilled players may invest time into learning more than just menders.

Similar to how engineers can do crazy stuff to get more power. This PR is to give doctors more things to engage with for faster turnaround times. 

## About the PR 
Disables basic mender healing of patients with critical levels of BRUTE/BURN. 
Buffs a huge variety of medical chemicals to bridge the gap.

**Topical agents & Automenders:**
- Styptic & Sulfadiazine apply no TOUCH if a spaceman has >= 100% HP in combined brute and burn damage. OXY/TOX doesn't count.
- Synthflesh will become up to 4x less efficient, but still work as spacemen enter deep crit.
- Menders filled with Styptic/sulfazine inform the user to use medication if they detect they won't help.
- Styptic and sulfazine now decay slower, meaning patches can heal 20 HP in crit (up from 10) - so medkits are less affected by this change.
> _TOUCH reagents work perfectly for keeping people topped up and healed. But as players enter deep critical, TOUCH healing easily outpaces any crit damage. In fact, its' so fast, that 99% of medical's arsenal is useless against BRUTE or BURN compared to mending._ 

**Cryoxadone:**
- Doubled healing effectiveness. Reduces metabolism by half. This equals out, but slows the effect of poisons and other medicine by 50%.
- Heals deep-crit patients faster.
- (Other PR) Cryo tubes will pause bleeding and quench fire.
>_Cryo is a great way to give doctors time, and should be a first recommendation to newbie doctors out of their depth. 
However, for newbies, a variety of unintuitive causes of death start popping up purely because cryo was used. This will be addressed in PR number 3._

**Saline Glucose:** 
- Now acts faster, works to cap brain damage at 50, 
- Counts as between to 0.25~1.25x its' volume in blood, increasing to combat hypotension. 
- Scales in strength and depletion as volume scales up, getting less efficient. Use an IV to keep it in the sweet spot.
- Saline autoinjectors now inject 30 units.
- Heals deep-crit patients faster.
>_Saline has been in a weird place for years - It's available in huge quantities, super easy to make, and you can drink an IV bag of it and gain regeneration for close to an hour.
Now, saline is a great away-from-home stabilizer, but each IV takes up an inventory slot._

**Heparin:**
- Thin blood moves faster. Doubles metabolism while injected. Heal critical patients faster if they aren't bleeding.
- If you take enough damage in a life tick (say, because you're heavily poisoned), your capillaries will give up and burst blood everywhere, taking a lot of poison out of your system, but also a large amount of blood. 
- This also works if you're beating someone to death, but takes significantly more effort (Like popping them with a sledgehammer)
- No longer bleeds out of your system.
>_Deprived of any use besides removing incredibly rare ailments and poisoning people, Heparin needed some love.  Now it has an incredible medical use of being able to double the rate a patient heals. 
This also becomes an interesting poison - You could dose someone and then hammer all their blood out. Hitting the damage threshold with your acid/cyanide/death pen will cause them to randomly explode into a puddle of their own toxic blood. Go wild!_

**Proconvertin:**
- No longer bleeds out of your system. 
>_Similar to Heparin. This stuff is pretty bad at its' job because you just bleed it out on the floor immediately._

## Why's this needed? 

Doctors of all experience levels complain is that medical is far too simple right now. And it's true - 90% of medbay's tools go completely unused unless it's the only way to fix something.

Topical agents break the entire medical game by providing functionally infinite healing per tick, multiplied by the number of doctors. 

The idea of this PR is to make critical patients more engaging, by making it last a little longer: 
- Stabilizing critical patients will actually matter, instead of outhealing any of the damage they could accrue in crit.
- New methods to heal small amounts of BRUTE or BURN instantly can exist, and will be respected.
- Crit patients won't accrue more BURN or BRUTE passively - doctors will only have to do a set amount of healing before getting their old tools back.

## Todos:

- [x] Better way to heal corpses #21995 
- [x]  Heparin poison cap
- [ ]  Adjust/new borg equipment to better match new needs
- [ ]  Migrate cryo changes away from cryoxadone
- [ ]  Adjust IV drips in doctoring tools PR

## Changelog 

```changelog
(u)TDHooligan
(*)Mender rework:
(*)Styptic Powder and Silver Sulfadiazine) no longer work if a spaceman has 100% combined brute/burn damage
(*)Synthflesh has up to 75% reduced effect on users in deep critical.
(*)Saline now stabilizes BRAIN and works significantly faster in larger quantities. 
(*)Cryoxadone is now 2x as strong, but makes you process all reagents, including itself, half as quick.
(+)Heparin now doubles patients' metabolism. Heparin patients are susceptible to burst if poisoned or beaten too hard.
```


Kyle2143:
Haven't looked at the whole thing, but I think that making Styptic and Sulfadiazine only work >90% (as you say) is too much. I would say reduced effectiveness like 1/3 or 1/4th would be a much easier pill to swallow than flat out not working. 

And I am not a fan of the 90% combined BRUTE/BURN being the threshold for that either. I'd probably want it to be med-crit like -50% where reduced effectiveness starts

TDHooligan:
> Haven't looked at the whole thing, but I think that making Styptic and Sulfadiazine only work >90% (as you say) is too much. I would say reduced effectiveness like 1/3 or 1/4th would be a much easier pill to swallow than flat out not working.

These reagents are so plentiful that reduced effectiveness won't impact their economy. If I do this, we'll see no change in doctoring besides it just taking longer. The outright disabling is a move to stop players engaging in what has only been described as boring by the same players who do it.

> And I am not a fan of the 90% combined BRUTE/BURN being the threshold for that either. I'd probably want it to be med-crit like -50% where reduced effectiveness starts

Starting at -50% will mean this change will work for approximately ~50 damage. Then it's just back to clicking menders. I'd say patients rarely make it back to medbay alive once raw BRUTE breaches 200 thanks to brute death rolls on top of oxy loss.

Sord213:
Overall I like the changes, definitely not a fan of the complete ineffectiveness of menders past the threshold, though. I think it'd be better for newbees if they were still a little bit effective (while yelling at you that they're not the best.) 

Heparin changes definitely will cause some nerd mixes (or even stuff like ling stings) to become hell but could be interesting to try on the condition it might be removed later.

TDHooligan:
The outright disabling of basic mender chems is important. I don't think menders would be fun to use if they healed at a rate comparable to a chemical mix, while also demanding 100% of your attention. On the other hand, if they're much faster we're back to square 1 as there is little reason to use chemical mixes at all.

I should mention, if it wasn't clear, that synthflesh menders will work at any % health. The MD also starts with a full synth mender. The cost there is instead in that synthflesh is harder to acquire.

jan-antilles:
Heparin doubling metabolism is a big enough balance change to chems as a whole that it feels weird to have it be a small bullet point under "mender changes."

LorrMaster:
Removing recent bullets and shrapnel should also increase bleeding by an additional amount, just because it sounds realistic to me.

TDHooligan:
> Removing recent bullets and shrapnel should also increase bleeding by an additional amount, just because it sounds realistic to me.
while i'd personally stray away from 'realism for realisms sake' there might be a case to introduce this down the line as it is cool. but cautery will replace most the BRUTE gained with BURN, and bandages are a bit rare atm. plus i dont think surgery needs to to take longer yet

YoSnoid:
if this pr gets added, please check before and after statistics on how often people are alive vs dead on classic. we wont be able to play the game much anymore and i dont feel like writing walls of text about why

TDHooligan:
> I think the PR can be atomized, though. The changes to heparin and saline-glucose-solution could likely be put in their own PR.

i mean. i can. but the reason they're bundled up like this is they're a bunch of balance changes that work around each other. crit stabilizing chem doesn't really need a buff when you can just click people out of it ATM

code/modules/chemistry/Reagents-Medical.dm
```diff
@@ -436,12 +436,19 @@ datum
 				if(!volume_passed)
 					return
 				volume_passed = clamp(volume_passed, 0, 10)
+				var/max_effectiveness_at = M.max_health*0.25
+				var/min_effectiveness_at = M.max_health*SYNTHFLESH_MINIMUM_HEALTH_PCT
+				var/min_effectiveness = 0.25
+				var/healing_effectiveness = 0
+				var/health_remaining = (M.max_health - M.get_brute_damage() + M.get_burn_damage())
```
BatElite, line 443: burn damages counts the wrong way as it is now (same for the other affected chems)

TDHooligan:
moved the total brute/burn from 90% to 100%, which is a lot more readable - if you hit critical without suffocating, menders will not work. i think this reads better than any other arbitrary health threshold. 

styptic/sulfa reagent depletion 3 -> 1.5. 
As a result, patches used below 0% HP now heal 20 HP, up from 10. This will more reliably pull you out of crit when you use them with Epi (which stops healing at -10)

i'll be raising another PR for other changes needed - likely corpse repair surgery and medi-cyborg related.

Valtsu0:
(i basically exclusively play med as a borg)
> Basic topical agents (Styptic, Sulfadiazine) no longer work if a spaceman has >= 100% HP in combined brute and burn damage
I do not think the other changes do enough to fill the role left by the automender while treating deep crit patients.

I don't find surviving from 200+ that uncommon, since as a borg, you can use the automender to take them out of deep crit, while at the same time injecting them with epi, saline and salbumatol and defibrillaiting them. Automender takes out of deep crit fast enough so there aren't too many death rolls

Currently it feels like saline takes forever to heal a significant amount of damage. I don't know how much the saline buff feel in practice. I'd like that some easily available chem would specialize in the role of healing while in crit (maybe epi but that is already very strong) or there to be more ways to significantly reduce death roll chance (maybe defibbing).

I will wait for the mediborg changes before giving further opinions

TDHooligan:
> I do not think the other changes do enough to fill the role left by the automender while treating deep crit patients.

yeah, for real deep crit I'm in agreement. Given that touch chemicals are obscenely fast, it's a matter of thinking up a rational way to introduce similar levels of healing, without being easy to exploit via pills.

I'll be thinking more, and likely migrating the surgical changes away so this is purely a chemical and mender PR.

Glamurio:
My personal main concern is that medical already has a problem of simply letting dying patients die and instead cloning them, and making it harder to heal critical patients will only exacerbate the problem. The rest of the changes seem great.

Valtsu0:
> My personal main concern is that medical already has a problem of simply letting dying patients die and instead cloning them, and making it harder to heal critical patients will only exacerbate the problem. The rest of the changes seem great.

The only way i see that problem getting fixed is giving defects on the first time someone is cloned

TDHooligan:
moved the surgery changes over to PR #2
this PR now explicitly covers the chemical changes.
Doctoring changes (including cryo and IV bag changes) are coming in PR 3.

Kyle2143:
Maybe instead of disabling the effectiveness of the chems that are typically used in menders, you just disable the mender itself when it reaches that threshold. 
Like make the mender do a beep boop and say "Patient too critically damaged to mend" or something. 
Then people might carry beakers of synthflesh/whatever in addition to menders for critically injured patients ofc. But idk if that's necessarily bad.

Kyle2143:
> > My personal main concern is that medical already has a problem of simply letting dying patients die and instead cloning them, and making it harder to heal critical patients will only exacerbate the problem. The rest of the changes seem great.
> 
> The only way i see that problem getting fixed is giving defects on the first time someone is cloned

This is a good idea. Something I've argued for in the past, but never actually did. I think currently you have like -10 max health on your first or second clone, but that's not enough. It should be like -50, or missing an arm/leg/organ that then needs replacing

TDHooligan:
just FYI - as part of building surgeries to further enhance this PR, I'm refactoring surgery code. so this PR will be sitting around a little bit

PR #21979, opened 2025-01-10
Nukie deployment cost reduction

Kaan-NN:
[Game-Objects] [Balance]
## About the PR 
This drops the cost of the nukie rapid deployment remote purchase to 1 out of the 2 guaranteed points.

## Why's this needed? 
Is using this especially on low pop a bad idea? Yes
Is it an incredibly stylish item regardless? Yes

This item is currently locked behind an 80 player population requirement but for how obvious it is it does not compete with other items in the 3 point and above range.

## Changelog 

```changelog
(u)Kaan
(+)The cost of the nuclear rapid deployment remote has been reduced to 2 (from 4).
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -1302,7 +1302,7 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/commander)
 /datum/syndicate_buylist/commander/deployment_pods
 	name = "Rapid Deployment Remote"
 	items = list(/obj/item/device/deployment_remote, /obj/item/paper/deployment_info)
-	cost = 1
+	cost = 2
 	desc = "A handheld remote allowing you, your team, and the nuclear device to be sent in anywhere at a moment's notice!"
 	category = "Main"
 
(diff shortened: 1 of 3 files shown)


Tarmunora (commented):
Split PR into separate PRs for the bugfix and balance change respectively

Kaan-NN:
Reverted the fix here and PRed it in #21996

ZeWaka:
image

code/datums/syndicate_buylist.dm
```diff
@@ -1302,7 +1302,7 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/commander)
 /datum/syndicate_buylist/commander/deployment_pods
 	name = "Rapid Deployment Remote"
 	items = list(/obj/item/device/deployment_remote, /obj/item/paper/deployment_info)
-	cost = 4
+	cost = 1
```
ZeWaka, line 1305:
```suggestion
	cost = 2
```

ZeWaka: 2

ZeWaka (approved):

PR #21992, opened 2025-01-11
Add a signal dispatch component

patricia-izin:
## About the PR 

Adds a new component, the signal dispatch component, which works like the dispatch component but dispatches the values of signals instead of the entire signal.
image

Drafting right now to put some shine and polish on, and add it to the mechanics book.

## Why's this needed? 

In the current game the main reason to split a signal apart is to get the contents of various signal list parameters, leaving this component with limited usefulness. Splitting signals will allow it to be used instead of multiple signal splitting components.

## Changelog 

```changelog
(u)patricia
(+)A new signal dispatch component has been added to the mechanics vendor, check the mechanics book for details
```


ZeWaka:
what does this allow that could not be accomplished before

patricia-izin:
> what does this allow that could not be accomplished before

Like the dispatch component replaces multiple signal check components, this can be used instead of several signal splitters. 

But I understand if it's unwanted

PR #22012, opened 2025-01-15
Bonemeal reagent for botany

klushy225:
[FEATURE] [HYDROPONICS]

## About the PR 
Adds a bonemeal reagent that is obtained by smashing bone pieces with a fire extinguisher.
When infused/processed, the bonemeal reagent increases the lifespan of the plant, and has a very small chance to give it the "rooted" gene, which allows herbs to be harvested multiple times. 
All of this was at the request of Bork

## Why's this needed? 
"because theres no chemical for lifespan that isnt harmful to plants already" - Bork

## Changelog 

```changelog
(u)klushy225
(+)Adds bonemeal that can increase the lifespan of plants
```


TDHooligan:
could probably be anything with sufficiently high blunt force, instead of just stinguishers? it'd be a little counterintuitive to not be allowed to smash them with a hammer

klushy225:
oh wait i forgot the tags, oops

code/modules/materials/Mat_RawMaterials.dm
```diff
@@ -405,6 +405,14 @@ ABSTRACT_TYPE(/obj/item/material_piece/rubber)
 	default_material = "bone"
 	uses_default_material_appearance = FALSE
 	mat_changename = FALSE
+	attackby(obj/item/W, mob/user)
+		if(istype(W,/obj/item/extinguisher))
```
TDHooligan, line 409:
```suggestion
		if(W.force > 5 && (W.hit_type == DAMAGE_BLUNT || W.hit_type == DAMAGE_CRUSH))
```

klushy225: Ah, thanks! Got distracted with other projects- appreciate the help!

ZeWaka (approved):
i feel like 'rooted' might be a bit too strong, but worth testmerging

code/modules/hydroponics/gene_strains.dm
```diff
@@ -330,6 +330,10 @@ ABSTRACT_TYPE(/datum/plant_gene_strain)
 		growth_tick.growtime_bonus += 0.1
 		growth_tick.harvtime_bonus += 0.1
 
+/datum/plant_gene_strain/rooted
+	name = "Rooted"
+	desc = "A very rare mutation which allows typically single harvest plants to continue living after a harvest."
```
ZeWaka, line 335: The main concern is really that this just makes them last infinitely long compared to before

ZeWaka: That would be a lot more amendable imo

PR #22021, opened 2025-01-16
Update TGS module

TealSeer:
[FEATURE] [TOOLING]
## About the PR 

Update the TGS DMAPI, v3 files have been removed as before

## Why's this needed? 

Was messing with it on another branch and figured there's no harm in updating it, not like it's used production here


ZeWaka:
Thanks for not caring about software from 10 years ago and removing it

ZeWaka:
Thanks for not caring about software from 10 years ago and removing it

PR #22023, opened 2025-01-16
Buffer component doesn't break when increasing the buffer size

patricia-izin:
[bugfix] [station systems]

## About the PR 

Resets the buffer component buffer correctly when the buffer sized is changed

## Why's this needed? 

It was still using the more wasteful qdel + didn't work because the reader and writer weren't reset


patricia-izin:
Missed this because it was seemingly only triggered if the buffer size was changed while holding signals, fun!

PR #22027, opened 2025-01-16
Move pipeshot to only work in break action shotguns.

TDHooligan:
[GAME OBJECTS][BALANCE]
## About the PR 
Prevents pipeshot from being used in advanced firearms.
Updates their description to make this clear.

MTS swapped to birdshot instead of scrapshot. A slight buff to make up for the fact it no longer has infinite ammo. Arguably less dangerous than the laser guns.

This means that pipeshot will remain usable in pipe guns, the bartenders' coach gun, and the salvager's shotgun.

## Why's this needed? 
Pipeshot's a headache to balance shotguns around - The moment a firearm becomes 12 gauge, it now has to sanely exist with an infinite supply of ammunition. The striker's had this real bad for a while, and the armory shotguns too.

It's also pretty simple to rationalize - how are these drunk, lunatic spacemen crafting consistent rim and case sizes from folding metal pipes in their hands? Shell extractors around the world shiver at the concept.

I imagine the shells could justifiably be buffed now they're harder to use, but that can wait for another PR.

## Changelog 

```changelog
(u)TDHooligan
(*)Pipeshot no longer functions outside of break-action shotguns.
(+)The MTS shotgun now spawns with Birdshot.
```


code/obj/item/gun/kinetic.dm
```diff
@@ -1950,7 +1950,7 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 	c_flags = ONBACK
 	force = MELEE_DMG_RIFLE
 	contraband = 5
-	ammo_cats = list(AMMO_SHOTGUN_ALL)
```
RubberRats, line 1953: This change makes the revolver shotgun unable to accept the ammunition it spawns with.

RubberRats (changes requested):
The list of shotguns affected by this PR is a bit scattershot, for lack of a better term. As far as I can tell, the only ones changed were the SPES (fair enough), the revolver shotgun (which spawns with pipeshot by default), and the unused tactical shotgun. Did you intend to add more to the list?

TDHooligan:
got pulled aside before finishing up and pushing, ta for the spot

PR #22028, opened 2025-01-16
Fix reinforced glass sheets being reinforced with glass

M-Earthfire:
[Bug][Materials]

## About the PR 

Makes the reinforced glass sheet manufacturer recipe correctly use the metal for its reinforcement instead of the glass used for it.

## Why's this needed? 

Glass reinforced glass sheets kinda don't make sense, especially if they need metal to produce that gets consumed.


RSG250:
did unreinforcing give glass rods?

PR #22031, opened 2025-01-17
Vent scrubbers don't scrub any gasses by default

patricia-izin:
[Station Systems]

## About the PR 

Vent scrubbers are no longer set to scrub any gasses by default.

## Why's this needed? 

Scrubbing all gasses is just siphoning mode, with #22020 users can now see the names of gasses in the status command even when off, so this will be more convenient in general.

## Changelog 

```changelog
(u)patricia
(+)Vent scrubbers are no longer set to scrub any gasses when built
```

### DIFF
--- a/code/modules/atmospherics/machinery/unary/vent_scrubber.dm
+++ b/code/modules/atmospherics/machinery/unary/vent_scrubber.dm
@@ -191,6 +191,9 @@
 	on = FALSE
 
 /obj/machinery/atmospherics/unary/vent_scrubber/breathable
+	#define _DEF_SCRUBBER_VAR(GAS, ...) var/scrub_##GAS = 1;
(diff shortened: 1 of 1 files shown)


Laboredih123 (changes requested):
Don't forget to change vent_scrubber/breathable to scrub the other gases since it wont scrub them by default now! Also thanks for making the pr, was planning to but didn't get to it.

patricia-izin:
@Laboredih123 thanks for pointing that out! Put the original macro onto the breathable version so it'll always scrub everything but breathable gasses

patricia-izin:
@Laboredih123 thanks for pointing that out, repurposed the macro to turn all the gasses back on again!

Laboredih123 (approved):
lgtm

ZeWaka (approved):

PR #22040, opened 2025-01-18
Add Bureaucratic Error event

JORJ949:
## About the PR 
Adds the "Bureaucratic Error" event, changing open job slots for random jobs for example, it could close all doctor slots and open infinite captain slots, or close a QM slot and open an extra genetecist slot. Staff Assistant and Security Officer are blacklisted from being affected by this event (so are borg and AI because you build those.)
image

Ive been playing a lot of goobstation on ss14 and while I don't know precisely how this event works there this is heavily inspired by it.

I would like to see the latejoin menu improved to show job slots over 6 better, presently it just keeps expanding to the right if you have more than 6 for any non-assistant job. Also a "locker teleporter" type item for extra amounts of the jobs may be nice (so for example each HoS can get a cape).

## Why's this needed? 
Fun event that can create fun scenarios such as joint command of a department.

## Changelog 

```changelog
(u)JORJ949
(*)Added Bureaucratic Error event, which can randomly open and close job slots.
```

### DIFF
--- a/code/modules/events/bureaucratic_error.dm
+++ b/code/modules/events/bureaucratic_error.dm
@@ -1,0 +1,88 @@
+#define BUREAUCRATIC_JOBS_IGNORE list("Staff Assistant", "Cyborg", "AI")
+
+/datum/random_event/major/bureaucratic_error
+	name = "Bureaucratic Error"
+	customization_available = TRUE
+	centcom_headline = "Bureaucratic Error"
+	centcom_message = "A bureaucratic error at Central Command has caused an error in job openings at the station."
+	centcom_origin = ALERT_STATION
+	var/job_to_increase = null
+	var/job_to_decrease = null
+	var/amount = null
+
+	admin_call(var/source)
+		if (..())
+			return
+		var/list/jobs = src.get_valid_jobs()
+		src.job_to_increase = tgui_input_list(usr, "Pick a job to increase", src.name, jobs)
+		src.job_to_increase = tgui_input_list(usr, "Pick a job to decrease", src.name, jobs)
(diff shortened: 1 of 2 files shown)


cheekybrdy:
Maybe have it so its either fixable or at least able to dampen the effects on the rolecontrol, so if you get zero officers on a station with decent pop and then this you aren't permafucked etc.

TDHooligan:
should probably just blacklist security/hos, as that's the hottest balance implication this would have. besides that it cool

JORJ949:
Secoff is blacklisted, but unique jobs are where the fun comes from this event (e.g. 3 HoSes) and presently most gear associated with these powerful jobs is in their locker rather than their person (for example, the lawbringer) and even then with HoS they are rare enough that you probably will only ever see 2 of them. Asisstant doesnt get stuns and the detective has very limited ammo, though I have blacklisted officer because that is a very carefully balanced job and you really dont want it to close all security jobs.

AnonymousUserTheOnly:
maybe also remove the "you can't choose this job if you're over a certain round number" restriction on the trainee variants if the normal variant is reduced?  trainees can't be picked if you've had a certain number of rounds under your belt.  according to the wiki, if you've played more than 75, you can't pick trainee, so this event removing all scientists but adding trainees would make it impossible for the well-established players to sign on in science.

RSOD404:
this seems fun, but this should probably be temporary when it ends up setting an important job's cap to 0, perhaps defaulting to leave 1-2 slots open.
not sure if the gamemode being revs should lock head roles from being affected by Bureucratic Error or not though

flappybatpal:
Eh, I have some concerns about how much this could impact a round and how players might feel about the job they want being arbitrarily unavailable.

PR #22047, opened 2025-01-19
TGUI telepad

FlameArrow57:
[QoL]

## About the PR 
Converts telepad UI to TGUI

tgui

tgui_r

## Why's this needed? 
TGUI


Garash2k:
If we're willing to prune the configuration switches, there's a bunch of UI in misc_terms.dm that are the exact same

https://github.com/orgs/goonstation/projects/7/views/1?pane=issue&itemId=36361424

https://github.com/goonstation/goonstation/blob/master/code/modules/networks/computer3/mainframe2/misc_terms.dm

FlameArrow57:
> If we're willing to prune the configuration switches, there's a bunch of UI in misc_terms.dm that are the exact same
> 
> https://github.com/orgs/goonstation/projects/7/views/1?pane=issue&itemId=36361424
> 
> https://github.com/goonstation/goonstation/blob/master/code/modules/networks/computer3/mainframe2/misc_terms.dm

Yeah this was a case where I don't think the configuration switches were accessible, unless I missed something

PR #22048, opened 2025-01-19
TGUI packet sniffer

FlameArrow57:
[QoL]

## About the PR 
Makes packet sniffer UI TGUI

image

## Why's this needed? 
tgui


tgui/packages/tgui/interfaces/PacketSniffer.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { Button, LabeledList, Section } from 'tgui-core/components';
+
+import { useBackend } from '../backend';
+import { Window } from '../layouts';
+
+interface PacketInfo {
+  packet_stamps;
+  packet_data;
+  filter;
```
TobleroneSwordfish, line 16: This is a typescript file so these should have some kind of type annotations.

FlameArrow57: Sure, can you clarify what you mean by type annocations though? if you are meaning stuff like "filter: string;", I did not see any of that in the other .tsx files

ZeWaka: You were looking at files that weren't fully converted yet, we were just doing it to get TGUI to modern standards

PR #22052, opened 2025-01-19
Adds a snippet to most UI windows to prevent some default browser shortcuts

TobleroneSwordfish:
[UI] [QOL]
## About the PR 
Adds an event listener that kills `crtl + o, l, n` shortcuts before they reach the browser.

## Why's this needed? 
How many times have you pressed `ctrl + n` to nod and gotten flashbanged by internet explorer for having a UI window focused?
~~Currently this just fixes it for TGUI but it might be possible to make it into a macro that can be inserted into the old-style UIs as well?~~
Added it into the Browse wrapper injection so it should work for *most* non-TGUI windows too.

## Changelog 

```changelog
(u)LeahTheTech
(+)The `ctrl + o, l, n` shortcuts will no longer work on <em>most</em> UI windows.
```

### DIFF
--- a/browserassets/tgui/tgui.html
+++ b/browserassets/tgui/tgui.html
@@ -47,7 +47,12 @@
 
 			// Backwards compatibility
 			window.__windowId__ = Byond.windowId;
-
+			// Prevent some default window shortcuts that annoy people
+			window.addEventListener("keydown", function(event){
+				if (event.ctrlKey && ["o", "l", "n"].includes(event.key)) {
+					event.preventDefault()
+				}
+			})
 			// Trident engine version
 			Byond.TRIDENT = (function () {
 				var groups = navigator.userAgent.match(/Trident\/(\d+).+?;/i);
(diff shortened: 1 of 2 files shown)


DisturbHerb:
I do actually use ctrl+n to save windows to my local storage. I would like to keep doing this is at all possible.

TobleroneSwordfish:
image

mordent-goonstation:
> I do actually use ctrl+n to save windows to my local storage. I would like to keep doing this is at all possible.

Is this some [workflow](https://xkcd.com/1172/) bit or do you actually have a use for that? Unironically trying to work out what possible reason there could be.

TobleroneSwordfish:
Okay I had a go at adding a custom keybind to save the current window but couldn't figure out a way to do it in our ancient cranky IE version. Feel free to PR if you have an idea of how to, but for now I'm just going to merge this because it is a huge QoL improvement overall.

PR #22061, opened 2025-01-20
Gunshot impact particles

Bartimeus973:
## About the PR 
Adds impact particles to most kinetic bullets when they hit a solid object.
The dust particles generated are dependant on the average color of the object or turf hit. The amount of particles and their velocity depends on the damage of the gun.
demo16

Energy guns also get a unique hit particle, which is more subdued, but also scales with damage. Unlike kinetic guns, the color of the particle does not depend on the atom hit, but rather on the color of the initial projectile.
demo15

This PR also reworks parts of the bullet impact component and removes their random spawns. Instead, bullet impacts are calculated and placed where the bullet is expected to hit. The height of the impact is dependant on the accuracy of the shot.

Finally, this PR adds two new variables to atoms, impact_icon and impact_icon_state which can be used to change what particles are emitted instead of dust when something is hit with kinetics.

This PR does not modify the particles spawned when someone is hit by an energy projectile as there is already a visual effect.

Massive thanks to TDHooligan for helping me with the difficult maths which determines the precise bullet impact!

## Why's this needed? 
Adds better impact to bullet hits and provides better visual feedback. Helps figure out where your shot was stopped if you hit something in a firefight.

## Changelog 

```changelog
(u)Bartimeus973 & TDHooligan
(*)Most guns now create particles when they hit something. This includes walls and objects!
```


Bartimeus973:
demo3

Underwater shooting

ZeWaka:
i see my codebaiting was successful

code/obj/effects/particles.dm
```diff
@@ -266,3 +266,118 @@
...
+				particles.count = new_amt
+				particles.spawning = new_amt
+				particles.velocity = generator("box", list(20*dir_x, 20*dir_y, 0), list(7*dir_x, 7*dir_y, 0), UNIFORM_RAND)
+			if (36 to 60)
+				particles.velocity = generator("box", list(40*dir_x, 40*dir_y, 0), list(15*dir_x, 15*dir_y, 0), UNIFORM_RAND)
+			if (61 to INFINITY)
+				var/new_amt = round(src.base_amt * 2)
+				particles.count = new_amt
+				particles.spawning = new_amt
+				particles.velocity = generator("box", list(60*dir_x, 60*dir_y, 0), list(20*dir_x, 20*dir_y, 0), UNIFORM_RAND)
```
ZeWaka, line 356: you can have non-integer damages, please overlap these ranges

ZeWaka (changes requested):

Bartimeus973:
Converting PR to draft since i'm also going to add energy weapon impacts

Bartimeus973:
Energy impacts have been added, i'll take the PR off draft status once the impact math has been finished.

ZeWaka:
idk if i'm sold on the blood particles imo, they seem like visual artifacts instead of stylized blood stuff

Bartimeus973:
I'll just make blood disappear quickly after shots. From 30 Seconds down to 5.

Bartimeus973:
I'm also changing the layer of blood particles specifically to PLANE_NOSHADOW_BELOW. I think the reason is sort of looked like artifacting is because it layered over everything, including the blood pools. Now it looks like this (some blood splats on the top right):
image

code/modules/projectiles/projectile_parent.dm
```diff
@@ -632,6 +696,11 @@ ABSTRACT_TYPE(/datum/projectile)
 	/// for on_pre_hit. Causes it to early-return TRUE if the thing checked was already cleared for pass-thru
 	var/atom/last_thing_hit
 
+	/// Set to TRUE if you want particles to spawn when you hit a non living thing
+	var/has_impact_particles = FALSE
+	/// Set to TRUE if you want kinetic impact particles, FALSE assumes energy projectiles
+	var/kinetic_impact = TRUE
```
TobleroneSwordfish, line 702: Shouldn't this just use damage types or typechecks? At least it should default to using them, with this var existing as an override if needed.

code/modules/projectiles/projectile_parent.dm
```diff
@@ -756,6 +825,51 @@ ABSTRACT_TYPE(/datum/projectile)
...
+					//Bullet impacts create dust of the color of the hit thing
+					var/avrg_color = hit.get_average_color()
+					new /obj/effects/gunshot_impact/dust(get_turf(hit), x, y, -O.xo, -O.yo, damage, avrg_color, new_impact_icon, new_impact_icon_state)
+					if (underwater)
+						new /obj/effects/gunshot_impact/bubble(get_turf(hit), x, y, -O.xo, -O.yo, damage)
+					else
+						new /obj/effects/gunshot_impact/sparks(get_turf(hit), x, y, -O.xo, -O.yo, damage)
+						new /obj/effects/gunshot_impact/smoke(get_turf(hit), x, y, -O.xo, -O.yo, damage)
+				else
+					//Energy impacts create sparks of the color of the projectile
```
TobleroneSwordfish, line 851: Maybe try putting these on a plane above lighting? It looks odd for them to be hidden by darkness since they're meant to be some kind of energy splash. (might look weirder if they're fullbright but worth a try)

code/modules/projectiles/projectile_parent.dm
```diff
@@ -756,6 +825,51 @@ ABSTRACT_TYPE(/datum/projectile)
...
+					new /obj/effects/energy_impact/energy(get_turf(hit), x, y, -O.xo, -O.yo, damage, avrg_color)
+					if (underwater)
+						new /obj/effects/gunshot_impact/bubble(get_turf(hit), x, y, -O.xo, -O.yo, damage)
+					else
+						new /obj/effects/energy_impact/sparks(get_turf(hit), x, y, -O.xo, -O.yo, damage)
+						if (damage >= 30)
+							new /obj/effects/energy_impact/smoke(get_turf(hit), x, y, -O.xo, -O.yo, damage)
+
+		// Spawn some blood particles if we hit some poor sod
+		spawn_mob_hit_particles(atom/hit, var/obj/projectile/O)
```
TobleroneSwordfish, line 862: I'm not really sold on these, hitting mobs already has very good feedback and the blood fading away over time feels weird when blood tends to be a forensic thing. Not to mention they're barely noticeable on most bullets, so it's just kind of adding performance cost for little reason.

TobleroneSwordfish (changes requested):
Very very cool PR, should probably be testmerged to ensure it doesn't tank client or server performance during large gunfights though.

code/modules/projectiles/projectile_parent.dm
Bartimeus973: It now checks for any damage that is D_ENERGY, D_BURNING, D_RADIOACTIVE or D_TOXIC, and if not overriden spawns the appropriate particle.

code/obj/effects/particles.dm
```diff
@@ -368,12 +368,26 @@
...
 
-/obj/effects/gunshot_impact
+/obj/effects/impact
+	var/static/effect_amount = 0
+
+	New()
+		effect_amount ++
+		if (effect_amount > 450) //About 150 gunshots in the span of 5 seconds
+			qdel(src)
+		. = ..()
```
TobleroneSwordfish, line 378: Would rather this was done in the logic that spawns these, spawning then instantly deleting something still incurs a performance cost.

TobleroneSwordfish (changes requested):

code/obj/effects/particles.dm
Bartimeus973: Moved it out of the particles and into the bullets, so we dont spawn and instantly delete the particles.

ZeWaka (commented):
This logic doesn't make sense to me. Wouldn't it be North?

code/datums/components/bullet_holes.dm
```diff
@@ -46,32 +48,35 @@ TYPEINFO(/datum/component/bullet_holes)
 	. = ..()
 
 
-/datum/component/bullet_holes/proc/handle_impact(rendering_on, obj/projectile/shot)
+/datum/component/bullet_holes/proc/handle_impact(rendering_on, obj/projectile/shot, impact_x, impact_y)
 	var/datum/projectile/shotdata = shot.proj_data
+	// Shots coming from the south hit the top part of the wall. Perspective means we do not spawn a decal
+	if (shot.dir == SOUTH)
+		return
```
ZeWaka, line 55: Here, for some reason GitHub didn't put in the right place

ZeWaka, line 55: Here, for some reason GitHub didn't put in the right place

Bartimeus973: Sorry, the comment makes it a bit confusing. Ill rewrite it to "shots GOING to the south".
Since the shot is going down it'll hit the top part of the wall and due to perspective we do not draw a decal

Tarmunora (approved):

PR #22064, opened 2025-01-20
TGUI nanocrucible

FlameArrow57:
[QoL]

## About the PR 
Converts nano-crucible UI to TGUI

image

image
(material pieces have names in the button tooltips)

## Why's this needed? 
TGUI


tgui/packages/tgui/interfaces/NanoCrucible.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return (
+    <Window title="Nano-Crucible" width={300} height={125}>
+      <Window.Content>
+        <Section fill>
+          <Flex align="center" textAlign="center" justify="space-between">
+            <Flex.Item grow>
+              <Button
+                onClick={() => act('load_first_part')}
+                tooltip={first_part}
+                color={!first_part ? 'orange' : 'green'}
```
TobleroneSwordfish, line 39: This should use standard colours, try to match any other interface that has a "click to insert" button, see chem dispensers etc.

tgui/packages/tgui/interfaces/NanoCrucible.tsx
```diff
@@ -0,0 +1,95 @@
...
+            <Flex.Item grow>
+              <Button
+                onClick={() => act('eject_result')}
+                color={result_part === '???' ? 'red' : 'green'}
+              >
+                {result_part}
+              </Button>
+            </Flex.Item>
+          </Flex>
+          <Flex align="center" textAlign="center">
```
TobleroneSwordfish, line 84: This isn't going to align correctly like this, consider making it a part of the central `Flex.Item` above it.

tgui/packages/tgui/interfaces/NanoCrucible.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+import { useBackend } from '../backend';
+import { Window } from '../layouts';
+
+interface PacketInfo {
+  first_part;
+  first_part_img;
+  second_part;
+  second_part_img;
+  result_part;
```
TobleroneSwordfish, line 18: Again, these should have type hints even if they're all just strings

TobleroneSwordfish (changes requested):

tgui/packages/tgui/interfaces/NanoCrucible.tsx
FlameArrow57: Sure, are images passed considered strings?

tgui/packages/tgui/interfaces/NanoCrucible.tsx
FlameArrow57: Could you clarify what you mean? Seems to align alright in game and when resizing the window

FlameArrow57: I was trying to position it in the middle of the bottom row, not under the plus

tgui/packages/tgui/interfaces/NanoCrucible.tsx
pgmzeta: yes, they're base64 strings

tgui/packages/tgui/interfaces/NanoCrucible.tsx
TobleroneSwordfish: Oh yep, but do you see how it's not actually centered relative to the ingredient element above it? It matters more that elements line up with each other than that they're centered relative to the window.

FlameArrow57: Okay if you think that looks better, then sure

FlameArrow57: Done

code/modules/materials/Mat_Processing.dm
```diff
@@ -88,100 +88,107 @@
...
+			"second_part_img" = second_part ? icon2base64(getFlatIcon(second_part)) : null,
+			"result_part" = updateResultName()
+		)
+
+	ui_act(action, params)
+		. = ..()
+		if (.)
+			return
+		var/mob/user = usr
+		switch (action)
```
TobleroneSwordfish, line 111: All of these should return TRUE to cause the UI to update immediately

TobleroneSwordfish (changes requested):
Okay one last thing: the loading behaviour seems very wonky right now. If I have two bars of the same material and I try to put one in each slot they will stack together, but if I try to add one of these bars to a slot that contains the same material they will swap and refuse to stack.

PR #22077, opened 2025-01-22
Doing CPR causes minor losebreath on the performer

pgmzeta:
[balance][medical]
## About the PR 
When completing a CPR action, the performer gets one (1) losebreath and is forced to gasp emote.

In stable conditions, the losebreath will be cancelled out. However, if there are other sources of losebreath (poison, low oxy) it could then start stacking up.

## Why's this needed? 
Make doing CPR more than just an action bar (but not much more).

## Changelog 

```changelog
(u)glowbold
(+)You take deep breaths when doing CPR.
```

### DIFF
--- a/code/datums/actions/actions.dm
+++ b/code/datums/actions/actions.dm
@@ -1427,6 +1427,9 @@
 			..()
 			interrupt(INTERRUPT_ALWAYS)
 			return
+		var/mob/M = owner
+		M.losebreath++ // ♪ give a little bit of your life to me ♪
+		M.emote("gasp")
 
 		target.take_oxygen_deprivation(-15)
 		target.losebreath = 0


imawxome:
wouldn't adding 1 lose breath also cause the CPR to be less effective? the wiki says that losebreath "acts like oxy damage", and a few toxins exclusively cause lose breath, plus, you don't heal lose breath in crit (I don't think, how lose breath works is very obfuscated without code diving)

444Portal:
> wouldn't adding 1 lose breath also cause the CPR to be less effective? the wiki says that losebreath "acts like oxy damage", and a few toxins exclusively cause lose breath, plus, you don't heal lose breath in crit (I don't think, how lose breath works is very obfuscated without code diving)

This is for the CPR performer, not receiver.

pgmzeta:
clarified the pr text

PR #22082, opened 2025-01-23
Show clone pod upgrades in cloning computer UI and allow upgrade removal

pgmzeta:
[qol][feature][medical][ui]
## About the PR 
* Show which cloning pod upgrades are installed on the clone pods in the cloning computer TGUI.
* Show if the necroscan II upgrade is installed in the cloning computer TGUI.
* Allow removing installed SpeedyClone, Recycling, and Reclaimer upgrades by crowbarring their respective machines.
* When examining a reclaimer, add text indicating there is an upgrade card installed.

## Why's this needed? 
There's no indication when upgrade modules are installed on the clone pod, reclaimer, or computer.
The speed module chews through biomatter reserves, so you may want to remove it but there is no way to do so.

## Screenshot
Screenshot 2025-01-22 173954
Screenshot 2025-01-22 175032

## Changelog 

```changelog
(u)glowbold
(+)The cloning console will indicate when the SpeedyClone, Biomatter Recyler, and NecroScan modules are installed.
(+)Clone pod and reclaimer upgrade modules can be uninstalled with a crowbar on their respective machines.
```


code/obj/machinery/clonepod.dm
```diff
@@ -679,6 +688,22 @@ TYPEINFO(/obj/machinery/clonepod)
 		src.light.disable()
 		src.UpdateIcon()
 
+	proc/remove_upgrade_modules(mob/user)
+		if (src.is_speedy)
+			src.is_speedy = FALSE
+			src.speed_bonus /= 3
+			src.meat_used_per_tick /= 4
```
mordent-goonstation, line 695: These, and the `meat_used_per_tick` below, should probably be local file `DEFINE`s at least, so that the multiplier applied when they're installed is the same thing.

code/obj/machinery/clonepod.dm
```diff
@@ -679,6 +688,22 @@ TYPEINFO(/obj/machinery/clonepod)
...
+			var/obj/speedy = new /obj/item/cloneModule/speedyclone(src.loc)
+			src.visible_message(SPAN_ALERT("The [speedy] module falls to the floor!"))
+			playsound(src.loc, 'sound/effects/pop.ogg', 80, FALSE)
+		if (src.is_efficient)
+			src.is_efficient = FALSE
+			src.meat_used_per_tick /= 0.5
+			var/obj/efficient = new /obj/item/cloneModule/efficientclone(src.loc)
+			src.visible_message(SPAN_ALERT("The [efficient] module falls to the floor!"))
+			playsound(src.loc, 'sound/effects/pop.ogg', 80, FALSE)
```
mordent-goonstation, line 704: This would play twice if it had both. Not sure exactly what that does to the sound.

tgui/packages/tgui/interfaces/CloningConsole.tsx
```diff
@@ -3,9 +3,11 @@
  * @copyright 2020
  * @author Original ThePotato97 (https://github.com/ThePotato97)
  * @author Changes Mordent (https://github.com/mordent-goonstation)
+ * @author Changes glowbold (https://github.com/pgmzeta)
  * @license ISC
  */
 
+import { BooleanLike } from 'common/react';
```
mordent-goonstation, line 10:
```suggestion
import { BooleanLike } from 'tgui-core/react';
```

mordent-goonstation (changes requested):

tgui/packages/tgui/interfaces/CloningConsole.tsx
```diff
@@ -272,6 +278,19 @@ const Functions = () => {
...
         </Box>
       </Section>
+      {!!allowDeadScan && (
+        <Section title="Necrosis Scanning Module">
+          <Box bold>Notice:</Box>
+          <Box>
+            Installation of the NecroScan II cloner upgrade module enables
+            scanning of rotted and skeletal remains.
+          </Box>
+          <Box fontSize="0.9em">
```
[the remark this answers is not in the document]
pgmzeta: It's meant to tell people blatanly that this doesn't make husked people cloneable; the verbiage matches the line on attempting to clone: "Extreme genetic degredation present." Given that it's not flavor, I opted to make this just part of the main text.

Tarmunora (approved):

PR #22089, opened 2025-01-23
Clipboard QoL improvements

pgmzeta:
[qol][game objects]
## About the PR 
* Using a paper bin in-hand on a clipboard will add the contained sheets to the clipboard
* You can drag paper sheets, paper bins, or pens to the clipboard, as long as you, the item, and the clipboard are next to each other (this includes held clipboards)
* cleanup src.pen during disposing

Both the attackby and MouseDropT procs use the same logic path via a new `add_stuff` proc.

## Why's this needed? 
Make clipboards a little easier to fill up and work with.


code/modules/writing/pens_writing_etc.dm
```diff
@@ -1003,11 +1003,12 @@
 
 /* =============== CLIPBOARDS =============== */
 
+#define CLIPBOARD_MAX_ITEMS 15
```
ZeWaka, line 1006: I think i'd rather have this be a var on the clipboard

ZeWaka (approved):

PR #22095, opened 2025-01-24
Adds The Barberang

RubberRats:
[GAME OBJECTS] [FEATURE]
## About the PR 

This PR adds a new Syndicate item, available to traitor barbers and spy thieves, as well as in surplus crates. The barberang is an otherwise unassuming razor that can be used as a throwing weapon, causing bleeding and a short stun as well as stealing the target's wig/hair.  It then returns to the user, dropping the stolen hair. To discourage spam, the barberang has a chance of cutting the user and stunning them when caught if it does not steal any hair.

**NOTE:** The actual numbers in this PR are not final. I was aiming for an item worth 3-5 TC intended to support a more expensive rampage weapon like the katana (similar to the pizza slicer and overcharged vacuum), that could also be used as a disruptive psychological-damage weapon on its own.

## Why's this needed? 

I have been holding onto this idea for months at this point. I think it would be a fun addition to the current traitor arsenal as something that can be either annoying but less deadly on its own, or as a compliment to another weapon. Also, the mental image of a bunch of bald staffies chasing a barber around is funny to me.

## Changelog 

```changelog
(u)RubberRats
(*)The Syndicate has equipped some of its operatives with a weapon of mass depilation
```


ju45he:
You absolutely NEED to put something about the anti-spam thing in the uplink's description. Or someone WILL get curbstomped the very first time they buy this and try to use it a few times.

Tarmunora:
> To discourage spam, the barberang has a chance of cutting the user and stunning them when caught if it does not steal any hair.

Backfires like this on traitor gear is pretty questionable imo. Better anti-spam I think would be to have it simply have no/reduced stun/etc. on people who are already hair/wigless

RubberRats:
> > To discourage spam, the barberang has a chance of cutting the user and stunning them when caught if it does not steal any hair.
> 
> Backfires like this on traitor gear is pretty questionable imo. Better anti-spam I think would be to have it simply have no/reduced stun/etc. on people who are already hair/wigless

The problem then is that because it's a returning projectile, you can still spam-throw it at people for the impact + bleed damage. That being said, I think limiting the stun to people who get barbered is more sensible than how it currently works. Will look into other ways of addressing the spam problem sometime tomorrow.

TobleroneSwordfish:
I absolutely and genuinely hate this thing, that is not a commentary on whether it should be added or not though :P

RubberRats:
So, it turns out that the `isbald()` proc has been broken for almost half a year. Anyway, added a fix to #22232 to this PR. If it doesn't get merged I'll atomize out the change and PR it seperately.

RubberRats:
Behavior has been changed: no longer cuts the user on a miss, now hitting a bald target doesn't stun them, deals extra bleed damage and makes it fall to the ground, disabling the return for a few seconds. Actual balance numbers subject to change.

Tarmunora (approved):

PR #22097, opened 2025-01-24
People with angry buddies trait only get arrested if they have at least 1 contraband

Chatauscours:
## About the PR 
If you have the "Was it something I said?" trait, Beepsky will now only arrest if you have at least 1 level of contraband.

## Why's this needed? 
People use this as an excuse to mess with sec and destroy securitrons. Relevant thread
https://forum.ss13.co/showthread.php?tid=23555

## Changelog 

```changelog
(u)Chatauscours
(+)If you have the "Was it something I said?" trait, Beepsky will now only arrest you if you have at least 1 level of contraband on you.
```

### DIFF
--- a/code/modules/robotics/bot/secbot.dm
+++ b/code/modules/robotics/bot/secbot.dm
@@ -848,7 +848,7 @@
 			if (src.threatlevel >= 4)
 				src.EngageTarget(C)
 				break
-			if(C.traitHolder.hasTrait("wasitsomethingisaid"))
+			if(C.traitHolder.hasTrait("wasitsomethingisaid") && src.threatlevel >= 1)
 				src.EngageTarget(C)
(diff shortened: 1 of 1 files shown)


TDHooligan:
definitely a better solution. i think of number of times i got away with a lil contra as a traitor. similar to jailbird, the biggest issue is that it devalues securitron responses. 

they arrest the guy with the shit trait, who then gets pissy when you search them because "i just have the trait". and you find nothing suspicious, so you're literally wasting your time. so you start ignoring securitron arrests. and then they get smashed and you don't care because they mostly provide misinformation.

now you WILL find contraband. the trait becomes more threatening as it's not just constantly throwing misinformation at security.

ZeWaka (approved):

PR #22101, opened 2025-01-24
TGUI oven

FlameArrow57:
[QoL][CATERING]

## About the PR 
Makes oven interface TGUI (doesn't include recipe book)

image

## Why's this needed? 
TGUI


tgui/packages/tgui/interfaces/Oven.tsx
```diff
@@ -0,0 +1,218 @@
...
+};
+
+const OvenDialbuttons = (props) => {
+  const { act } = useBackend();
+  const { min, max, time } = props;
+  const nodes: JSX.Element[] = [];
+  for (let i = min; i <= max; i++) {
+    const node = (
+      <Button key={i}
+        color={time === i ? 'default' : 'grey'}
```
TobleroneSwordfish, line 210: Grey means disabled, buttons should use the default blue for clickable and green for selected.

tgui/packages/tgui/interfaces/Oven.tsx
```diff
@@ -0,0 +1,218 @@
...
+                  textAlign="center"
+                >
+                  Low
+                </Button>
+              </Stack.Item>
+            </Stack.Item>
+          </Stack>
+          <Stack justify="center">
+            <Stack.Item>
+              <Button selected onClick={() => act('start')} fontSize={2} mt={0.5}>
```
TobleroneSwordfish, line 97: This button shouldn't be selected as that doesn't mean anything in this context.

tgui/packages/tgui/interfaces/Oven.tsx
```diff
@@ -0,0 +1,218 @@
...
+                <OvenDialbuttons min={1} max={5} time={time} />
+              </Stack.Item>
+              <Stack.Item>
+                <OvenDialbuttons min={6} max={10} time={time} />
+              </Stack.Item>
+            </Stack.Item>
+            <Stack.Item vertical>
+              <Stack.Item>
+                <Button
+                  color={heat === 'High' ? 'red' : 'grey'}
```
TobleroneSwordfish, line 75: Same with these heat buttons, blue for clickable, green for selected. If you want to display the state with a colour do it separately.

tgui/packages/tgui/interfaces/Oven.tsx
```diff
@@ -0,0 +1,218 @@
...
+                    </Stack.Item>
+                  </Stack>
+                </Section>
+              </Stack.Item>
+              <Stack.Item><Divider vertical /></Stack.Item>
+              <Stack.Item grow>
+                <Section title="Result">
+                  <Stack vertical align="center" textAlign="center">
+                    <Stack.Item>
+                      <img
```
TobleroneSwordfish, line 177: This should use `Image` instead of `img` to avoid pixel scaling issues. That would let you make the icon bigger too :)

tgui/packages/tgui/interfaces/Oven.tsx
```diff
@@ -0,0 +1,218 @@
...
+                        height={48}
```
TobleroneSwordfish, line 178: Try to stick to whole multiples of the icon size to scale cleanly

TobleroneSwordfish (changes requested):
I like this a lot, clean UI that doesn't change much but still improves over the old version.
Some issues from playing around with it a bit:
- Inserting an item into the oven should update the UI.
- The contents area still feels a little bit cramped, maybe expand it down by another row at least?
- It feels like these bottom sections should always fill the window if displayed, otherwise there's a weird gap at the bottom for smaller recipes:
image
Could also just make them always visible but empty if there's no recipe, that might be clearer anyway.

TobleroneSwordfish (changes requested):
sry for double review, I missed a couple of things and github doesn't let me append

tgui/packages/tgui/interfaces/Oven.tsx
```diff
@@ -0,0 +1,218 @@
...
+        )}
+      </Window.Content>
+    </Window>
+  );
+};
+
+const OvenDialbuttons = (props) => {
+  const { act } = useBackend();
+  const { min, max, time } = props;
+  const nodes: JSX.Element[] = [];
```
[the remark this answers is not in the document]
FlameArrow57: Could you give an example of this or a reference to code that does this? I was just pulling from other code that existed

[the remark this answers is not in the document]
TobleroneSwordfish: Oh uhh just put the map inside a `<Box>` or even a `<>` and return that, pretty much every other component does that.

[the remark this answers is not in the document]
FlameArrow57: LMK if it looks better

TobleroneSwordfish (changes requested):
One last nitpick, otherwise looks good to me 👍

FlameArrow57:
Cool

TobleroneSwordfish (approved):

mordent-goonstation (commented):
Are the screenshots reflecting latest state? Can we get a screenshot of it in light mode?

Button sizes for top controls are arbitrary. I understand the desire for a specific aesthetic, but arbitrary button size choices make keeping a consistent look and feel across our UIs harder.

Looks like a vertical alignment issue in the lists of items.

How would these work with all but the Cook button being the default size?

FlameArrow57:
> Are the screenshots reflecting latest state? Can we get a screenshot of it in light mode?
> 
> Button sizes for top controls are arbitrary. I understand the desire for a specific aesthetic, but arbitrary button size choices make keeping a consistent look and feel across our UIs harder.
> 
> Looks like a vertical alignment issue in the lists of items.
> 
> How would these work with all but the Cook button being the default size?

Latest yeah. How can I test in light mode?

Looks like this with default button sizes
image

ZeWaka:
Could you vertically center the text in the contents row? It looks a bit strange

ZeWaka:
Could you vertically center the text in the contents row? It looks a bit strange

FlameArrow57:
> Could you vertically center the text in the contents row? It looks a bit strange

image

TobleroneSwordfish (changes requested):
For the buttons I'm just going to say keep the large cook button and make all the others default, it's fine to use resizing to draw attention to one element but that doesn't work if *everything* is larger.
(last thing I promise!)

PR #22127, opened 2025-01-27
TGUI GPS

FlameArrow57:
[QoL]

## About the PR 
Converts handheld GPS device UI to TGUI

image

## Why's this needed? 
tgui


ZeWaka:
I think the top should probably be two columns

ZeWaka:
I think the top should probably be two columns

patricia-izin:
I like the vertical layout for the buttons a lot less, there's a lot of empty space that the original didn't need without looking cramped. The distress signals look kinda bad too with the saturated solid background color with no edges. Also the button sticks out from the background in the distress signals.

Setting the coordinates in two separate pop up windows for each coordinate also isn't very good, especially with the nice inline one it currently has

image

Garash2k:
Use `<Button.Checkbox>` instead of a Yes/No button.

At minimum the the gps rows would need padding so that the text inside isn't flush with the red box of a distress alert, but I feel like there's gotta be a better way to handle distress alerts... Something with similar style to a "warning" or "danger" `<NoticeBox>`

FlameArrow57:
> At minimum the the gps rows would need padding so that the text inside isn't flush with the red box of a distress alert, but I feel like there's gotta be a better way to handle distress alerts... Something with similar style to a "warning" or "danger" `<NoticeBox>`

Does blinking red text work for "Distress Alert"? Or still in a notice box?

FlameArrow57:
> Setting the coordinates in two separate pop up windows for each coordinate also isn't very good, especially with the nice inline one it currently has
Maybe, is it that bad? In all it's only one more button click

patricia-izin:
> 
> 
> 
> Maybe, is it that bad? In all it's only one more button click
> 
> 
Well no it wouldn't be a dealbreaker for me if that mattered ha ha, I really only have a strong opinion is that the distress call should be framed like the original

mordent-goonstation:
That X/Y combo should definitely be a pair of elements, not one modal after another. I would consider two dials if people didn't rebel against them so much, but using those coordinates is such a niche case it feels justified, even.

FlameArrow57:
> That X/Y combo should definitely be a pair of elements, not one modal after another. I would consider two dials if people didn't rebel against them so much, but using those coordinates is such a niche case it feels justified, even.
> 
> How does this look in light mode?
> 
> I've not reviewed the code yet, just glanced at the screenshots.

image

FlameArrow57:
Ready for re-review

tgui/packages/tgui/interfaces/GPS.tsx
```diff
@@ -0,0 +1,214 @@
...
+const TrackableList = (props) => {
+  const { act } = useBackend();
+  const { gps_info } = props;
+  const nodes: JSX.Element[] = [];
+
+  distress_red = !distress_red;
+
+  for (let i = 0; i < gps_info.length; i++) {
+    const node = (
+      <Box key={i} fontSize={0.8}>
```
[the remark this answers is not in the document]
mordent-goonstation: Actually, given this just wraps a `Stack.Item` and a `Divider`, can just make this a `React.Fragment` instead of a `Box`

tgui/packages/tgui/interfaces/GPS.tsx
```diff
@@ -0,0 +1,214 @@
...
+              <Button
+                onClick={() => act('track_gps', { gps_ref: gps_info[i].ref })}
+                fontSize={1}
+              >
+                Track
+              </Button>
+            </Stack.Item>
+          </Stack>
+        </Stack.Item>
+        <Divider />
```
mordent-goonstation, line 208:
```suggestion
        <Stack.Divider />
```
Resolves to the same thing, currently, but should use `Stack`-specific one in case that's not true in future.

tgui/packages/tgui/interfaces/GPS.tsx
```diff
@@ -0,0 +1,214 @@
...
+  );
+};
+
+let distress_red = false;
+const TrackableList = (props) => {
+  const { act } = useBackend();
+  const { gps_info } = props;
+  const nodes: JSX.Element[] = [];
+
+  distress_red = !distress_red;
```
mordent-goonstation, line 176: Ack, no.

If you want this flashing, use CSS animations or something.

tgui/packages/tgui/interfaces/GPS.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    onClick={() => act('toggle_distress')}
+                    selected={distress}
+                    checked={distress}
+                  >
+                    {distress ? 'Yes' : 'No'}
+                  </Button.Checkbox>
+                </LabeledList.Item>
+                <LabeledList.Item label="Tracking">
+                  <Button onClick={() => act('track_gps')}>
+                    {tracking ? tracking : 'None'}
```
mordent-goonstation, line 135:
```suggestion
                    {tracking || 'None'}
```

tgui/packages/tgui/interfaces/GPS.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    tooltip={gpsTooltip}
+                  >
+                    {src_name.slice(5, src_name.length)}
+                  </Button>
+                </LabeledList.Item>
+                <LabeledList.Item label="Coords">
+                  x:{' '}
+                  <NumberInput
+                    value={track_x}
+                    width={'3'}
```
mordent-goonstation, line 87:
```
                    width={3}
```
or
```
                    width="3"
```
Both should resolve to the same, one may not be okay because of typing, if so just use the other.

tgui/packages/tgui/interfaces/GPS.tsx
```diff
@@ -0,0 +1,214 @@
...
+            </Stack.Item>
+          </Stack>
+        </Section>
+        <Section title="Tracking">
+          <Collapsible title="GPS Devices">
+            {!!gps_info.length && (
+              <Stack vertical>
+                <TrackableList gps_info={gps_info} />
+              </Stack>
+            )}
```
mordent-goonstation, line 148: Should have something in these saying there are none found, rather than just rendering nothing.

tgui/packages/tgui/interfaces/GPS.tsx
```diff
@@ -0,0 +1,214 @@
...
+              </LabeledList>
+            </Stack.Item>
+          </Stack>
+        </Section>
+        <Section title="Tracking">
+          <Collapsible title="GPS Devices">
+            {!!gps_info.length && (
+              <Stack vertical>
+                <TrackableList gps_info={gps_info} />
+              </Stack>
```
mordent-goonstation, line 147: Probably move the `Stack` into the `TrackableList` component's render, as it's an implementation detail of that, I think?

mordent-goonstation (changes requested):

mordent-goonstation:
Also please include a light-mode screenshot in your PR body.

code/obj/item/device/gps.dm
```diff
@@ -20,6 +20,13 @@ TYPEINFO(/obj/item/device/gps)
 	var/net_id
 	var/wrenched_in = FALSE //! is this wrenched in a cabinet frame?
 
+	var/tracking_string
+	var/tracking_x = 1
+	var/tracking_y = 1
+	var/list/gps_devices = list()
+	var/list/imps = list()
+	var/list/warp_beacons = list()
```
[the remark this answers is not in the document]
FlameArrow57: They're being used in ui_act() to find the returned reference

tgui/packages/tgui/interfaces/GPS.tsx
FlameArrow57: Where can I find CSS animations or examples?

tgui/packages/tgui/interfaces/GPS.tsx
FlameArrow57: Idk, it stops showing up with this change

tgui/packages/tgui/interfaces/GPS.tsx
FlameArrow57: It stops having default width of 3 with these changes

tgui/packages/tgui/interfaces/GPS.tsx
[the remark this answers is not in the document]
FlameArrow57: Seems to add a little bit of extra spacing, personally think Box looks a little more compact/nicer

code/obj/item/device/gps.dm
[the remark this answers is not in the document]
mordent-goonstation: Is that info important to hold onto? Could it just be generated for `ui_data` and then the ref given in `ui_act`? Not me passively saying to do this, it's an actual question. Broadly holding onto a list that we're not actually using, and still regenerating every `ui_data` tick, seems like it's really a `ui_data` thing that could be sent via `ui_act`?

To my knowledge we don't have some case where people could send up, like, a malicious `ref` or whatever because a) how did they get the ref? b) we're including all GPS units here anyway. Obviously verify the resultant ref *is* the right thing, but does holding onto it actually get us anything?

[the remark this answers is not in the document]
FlameArrow57: I think I forgot you can use locate() to find a reference, but these lists are removed now

tgui/packages/tgui/interfaces/GPS.tsx
mordent-goonstation: Weird. `width="3"` should be entirely comparable to `width={'3'}`, just less shit.

tgui/packages/tgui/interfaces/GPS.tsx
mordent-goonstation: We don't use them extensively (for good reason), but the ChemHeater interface does an oscillating thing.

https://github.com/goonstation/goonstation/blob/23bbebf599a2b041bc8f4b180cb2ed2a9808f57b/tgui/packages/tgui/styles/interfaces/ChemHeater.scss#L51

ZeWaka (approved):
tested, way better than current

PR #22131, opened 2025-01-27
Arc Electroplaters can break and be repaired

pgmzeta:
[feature][game objects]
## About the PR 
* Equipment Fault component changes
  * When performing a fault on `/obj/machinery`, only do so if it's powered
* Add a new Equipment Fault type: Embers
  * When a fault is triggered, create an effect based on the lighter/welder special attack
  * This effect is heavily weighted (80-20) towards a harmless ember (like the 'spark' effect from an empty lighter/welder)
* Add Arc Electroplater icon states for broken, unpowered, and broken-unpowered 
* Feature changes to the Arc Electroplater
  * `update_icon` now handles all equipment states, and is used instead of setting the icon_state directly
  * When broken, lower power usage and subscribe to the process loop to ensure equipment faults are called on regular intervals
  * Add ember equipment fault, fixed with welding, screwing, and wrenching
  * Add bullet_act/ex_act overrides to make the electroplater break with sufficient damage/force
* Misc changes to the Arc Electroplater
  * Set `\improper` in the name instead of using `\the` everywhere
  * Fix double-power usage (was calling `use_power` while subscribed to the process loop)

## Why's this needed? 
* Make station equipment more reactive to explosions and gunfire
* Arc Electroplaters visually show when they're unusable due to no power

## Demo

https://github.com/user-attachments/assets/66aacba7-dc28-451a-b142-9142a57ad7f5

## Changelog 

```changelog
(u)glowbold
(+)Arc Electroplaters can be broken by kinetics and explosions - repair them with common tools!
```


Azrun (changes requested):

NightmareChamillian:
Looks good! One thing I would suggest would be a slightly more obviously damaged broken sprite, though.

TobleroneSwordfish (changes requested):
Breaking, repairing and then breaking again doesn't give you the option to repair it. Also I know this is more of a general issue with the fault component but it really doesn't make sense that it uses actionbars when deconstruction doesn't.

pgmzeta:
> Breaking, repairing and then breaking again doesn't give you the option to repair it.

Fixed in #22147 (and more generally in #22179); updated the branch so this should be working now.

> Also I know this is more of a general issue with the fault component but it really doesn't make sense that it uses actionbars when deconstruction doesn't.

Adding actionbars to decon feels like a really big nerf/time sink for salvagers, and would need a separate PR. That said, I don't feel comfortable outright changing this for faults without some feedback from @Azrun as it's not my feature.

Azrun:
My design intent was to simulate the act of troubleshooting the problem as well as give someone the ability to defend the fault/failure.

But yeah, I agree that since the UX is so similar between the two it does create an odd dissimilarity in how the actions are ultimately performed. 

I have no qualms if they are unified, but preferably as a separate PR.

pgmzeta:
Digging into deconstruction more, while each step does not have an actionbar, it does add 2.5 seconds worth of time to the final deconstruction action-bar:

https://github.com/goonstation/goonstation/blob/b8ed1e7994bcf1b1d8fde4cc47f1e62d06268f0d/code/obj/item/tool/electronics.dm#L986

This would still be in a separate PR, but to bring these two interactions in line without upending decon balance, we could split off 1 second into each step actionbar, then have the final actionbar take 1.5 seconds per step - thoughts?

PR #22149, opened 2025-01-28
Add a methane canister to all in rotation maps

Kaan-NN:
[MAPPING] [BALANCE]
## About the PR 
This adds a single methane canister to all in rotation maps.
For all maps except nadir those are located in toxins. (Nadir lacks toxins so its a bit hidden behind a false wall in the crew quarters).

See the description and images here for a complete overview of what map changes this PR does. [https://forum.ss13.co/showthread.php?tid=23518&pid=208808#pid208808](https://forum.ss13.co/showthread.php?tid=23518&pid=208808#pid208808)

## Why's this needed? 
The way of acquiring methane right now are:

- pray you have someone on station that can check the market, have a trader present and convince someone to buy a can from them
- press F for 15 minutes (or hope that in the meantime the market re rolled and this time you can get it)

Methane is the one gas that enables any and all atmos gameplay with oxygen agent b being the more complex but in theory better version.
Right now experimentation is gated by population based job shortages, RNG or pressing a single button until your fingers hurt.
Any experimentation falls short due to the time it takes to complete all steps from acquiring the gasses, making a setup and then doing the burn which can take 45 minutes just on its own to see any meaningful results.
For a longer explanation on my view see [https://forum.ss13.co/showthread.php?tid=23518](https://forum.ss13.co/showthread.php?tid=23518).

## Changelog 

```changelog
(u)Kaan
(*)Added a canister of methane to all in rotation maps. Some might be a little tucked away.
```

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -11613,7 +11613,7 @@
 /turf/simulated/floor/red,
 /area/station/ai_monitored/armory)
 "eXF" = (
-/obj/machinery/portable_atmospherics/canister/empty,
+/obj/machinery/portable_atmospherics/canister/methane,
 /turf/simulated/floor/purple/side{
 	dir = 9
 	},
@@ -13180,6 +13180,7 @@
 /obj/machinery/atmospherics/unary/portables_connector{
 	dir = 4
 	},
+/obj/machinery/portable_atmospherics/canister,
 /turf/simulated/floor,
 /area/station/science/storage)
 "hwt" = (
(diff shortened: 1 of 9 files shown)


TDHooligan:
if we add this we're going back to standardised roundstart hellburns. which got 'removed'. you never need more than 1.
i think it'd be better if you just add the canister to more not-roundstart places, so QM remains the easiest option.

Tarmunora:
> Thus change would require a massive nerf beforehand. This change would need to kill off exponential generation of heat by the reactions of farts and cap it to a sensible level.
> 
> I went into detail about the reason for this in the forum post.

Agreed. So long as methane is still a magical anti-entropy gas that can uncombust plasma, and pipeburns are still bypassing agent_b catalyzation requirement for high-temperature burns, making methane a roundstart canister is not happening.

PR #22163, opened 2025-01-30
Make scuttlebot hat immune to acid

pgmzeta:
[balance][game objects]
## About the PR 
Marks the syndicate scuttlebot hat as acid-proof.

## Why's this needed? 
Chemical acids should not remove unique and irreplacable items from the round.

### DIFF
--- a/code/obj/item/clothing/hats.dm
+++ b/code/obj/item/clothing/hats.dm
@@ -303,6 +303,7 @@ proc/filter_trait_hats(var/type)
 //A robot in disguise, ready to go and spy on everyone for you
 /obj/item/clothing/head/det_hat/folded_scuttlebot
 	blocked_from_petasusaphilic = TRUE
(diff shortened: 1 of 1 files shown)


ZeWaka (approved):

code/obj/item/clothing/hats.dm
```diff
@@ -332,6 +333,7 @@ proc/filter_trait_hats(var/type)
 /obj/item/clothing/head/det_hat/gadget
 	name = "DetGadget hat"
 	desc = "Detective's special hat you can outfit with various items for easy retrieval!"
+	item_function_flags = IMMUNE_TO_ACID
```
Tarmunora, line 336:
```suggestion
```

Acid immunity is typically only granted to antagonist gear, with almost no exceptions.

Tarmunora (changes requested):

PR #22164, opened 2025-01-30
MechComp Teleporters Use Projectiles.

pgmzeta:
[experimental][balance][game objects]
## About the PR 
* MechComp teleporters fire a homing projectile towards the destination teleporter
* The projectile provides a safe atmosphere to ride and breathe in while you travel
* The projectile can be intercepted by any active teleporter along the way, allowing for Crime
* Add a new status effect: "Teleporting", which will blind teleporting users
* MechComp teleporters will only target teleporters on the same Z-level
* If the destination teleporter is deleted or un-anchored, the projectile dies and dumps the transported contents out

## Why's this needed? 
* MechComp teles are instant and cross-z despite requiring minimal power and configuration.
* If you see someone take a mechcomp tele, you have zero idea where they went to.
* It's neat!

## Demo

https://github.com/user-attachments/assets/87f56f59-9f2b-4b8d-a649-878a0eda9986

## TODO
- [x] move recieving teleporter sends to signal when a teleport is recieved
- [x] intercepted teleports should add a unique string to `LOG_STATION`
- [x] test if these will pass over a singularity (i.e. donut2)
- [x] performance-focused review/pass

## Changelog 

```changelog
(u)glowbold
(*)MechComp teleporters launch projectiles with the teleported contents inside, aimed at the targeted teleporter.
(+)MechComp teleporters cannot cross Z-levels. Use a command teleporter or telesci instead.
(+)MechComp teleporter projectiles are intercepted by the first active teleporter they hit.
```


Sord213:
I'd say make the projectile only visible when wearing mesons. Otherwise this is a really cool PR

TDHooligan:
should make it so you don't choke/freeze to death while teleporting too. looks really janky otherwise.

pgmzeta:
> should make it so you don't choke/freeze to death while teleporting too. looks really janky otherwise.

done but it feels kind of hacky

ZeWaka:
is the description accurate

pgmzeta:
> is the description accurate

updated the status effect description (?)

ZeWaka:
> > is the description accurate
> 
> updated the status effect description (?)

of the pr

pgmzeta:
> > > is the description accurate
> > 
> > 
> > updated the status effect description (?)
> 
> of the pr

i think so, im not aware of anything in the description that's drifted from implementation

PR #22179, opened 2025-02-01
Fix COMPONENT_DUPE_HIGHLANDER 

Sovexe:
## About the PR 
* Fixes `COMPONENT_DUPE_HIGHLANDER` behavior to delete the old component in favor of the new as documented

## Why's this needed? 
* The current behavior sets up the new component to replace the old one and then just deletes it
* This has been broken for over 5 years and is the default behavior, there may exist code reliant on the broken behavior


Laboredih123:
oh i remember finding out this bug, i forgot to report it lol

PR #22200, opened 2025-02-03
Deconstructors use Pre-Attack signal instead of afterattack

pgmzeta:
[qol][game objects]
## About the PR 
* Convert deconstruction tool `afterattack` into a pre-attackby callback
* Add help text explaining functionality
* Make explicit exception for objcritters (i.e. always attack them) 
* Remove scattered deconstruction tool special-casing in the codebase

## Why's this needed? 
* No longer require special-casing for deconstruction tools
* Make deconstruction tools work like device analyzers i.e. help/disarm/grab is special functionality, harm is regular item usage (i.e. attack with)

## Changelog 

```changelog
(u)glowbold
(+)Deconstruction tools will always try to deconstruct objects on help, disarm, and grab intents. On harm intent, deconstruction tools will be used like a normal item on objects.
```

### DIFF
--- a/code/WorkInProgress/AzrunStuff.dm
+++ b/code/WorkInProgress/AzrunStuff.dm
@@ -4,14 +4,20 @@
 	desc = "A magical saw-like device for unmaking things. Is that a soldering iron on the back?"
 	default_material = "miracle"
 
-	afterattack(atom/target as mob|obj|turf|area, mob/user as mob)
+	New()
+		. = ..()
+		RegisterSignal(src, COMSIG_ITEM_ATTACKBY_PRE, PROC_REF(pre_attackby), override=TRUE)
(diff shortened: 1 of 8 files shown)


TobleroneSwordfish:
I feel like this should probably only apply to things that can be deconstructed? It's weird to just get an error message when you hit a window with a saw.

pgmzeta:
> I feel like this should probably only apply to things that can be deconstructed? It's weird to just get an error message when you hit a window with a saw.

returns early if there are no deconstruct flags now. The behavior matches current except for using harm intent with a decon tool, which will follow through with an attackby on things that had early returns previously (i.e. vending machines)

TobleroneSwordfish (approved):

PR #22203, opened 2025-02-03
Chemical and drink dispensers leak when broken

pgmzeta:
[feature][game objects]
## About the PR 
* New equipment fault: Leaky
  * Leaks become more likely each process cycle, until one occurs and the base chance is reset
  * Sub-types are added for each dispenser type (chemical, alcohol, soda, chef)
  * alcohol/soda/chef dispensers are more likely to leak (higher base probability)
* Add equipment fault to the various dispenser sub-types, fixed by screwing, wrenching
* Dispensers are damaged by gunfire and blob actions that don't destroy them outright, and take slight damage for weak explosions
* At under 1/4th health, dispensers have a chance of breaking upon taking damage based on their remaining HP
* New sprites for broken dispensers

## Why's this needed? 
More dynamic machinery reactions to damage

## Demo
https://github.com/user-attachments/assets/70d6518b-6937-48f5-acd2-37762794f19e

## Changelog 

```changelog
(u)glowbold
(+)Chemical, bar, and chef dispensers can be broken with enough damage. Fix them with common tools!
```


code/datums/components/equipment_fault.dm
```diff
@@ -454,3 +454,75 @@ TYPEINFO(/datum/component/equipment_fault)
...
+			spray.set_dir(target_dir)
+		playsound(O, pick(sounds), 50, 2)
+		O.visible_message(SPAN_NOTICE("Some of the contents of [O] leaks onto the floor."))
+
+		var/datum/reagents/temp_fluid_reagents = new /datum/reagents(5)
+		temp_fluid_reagents.add_reagent(pick(src.reagent_list), 5)
+		T.fluid_react(temp_fluid_reagents, temp_fluid_reagents.total_volume)
+
+/datum/component/equipment_fault/leaky/chemical
+	reagent_list = list(
```
TobleroneSwordfish, line 498: Why not just pass in this list from the parent? Otherwise it will break on varedited chem dispensers or if any of these lists change in the future.

code/modules/chemistry/Chemistry-Dispenser.dm
```diff
@@ -239,6 +250,9 @@ TYPEINFO(/obj/machinery/chem_dispenser)
 
 	proc/take_damage(var/damage_amount = 5)
 		src.health -= damage_amount
+		if (src.health < 100)
+			if(prob(100-src.health))
```
TobleroneSwordfish, line 254: This means it almost never triggers with high damage bullets, why the `prob`?

TobleroneSwordfish (changes requested):

code/datums/components/equipment_fault.dm
pgmzeta: added the list of reagents as an initialization argument to the component and removed the subclasses. the parent reagent dispenser now adds the component with the current machine's dispensable reagents.

code/modules/chemistry/Chemistry-Dispenser.dm
pgmzeta: The behavior I want is for chem dispensers to have a higher and higher chance to break over time, without immediately breaking due to e.g. a single potassium and water mix.

I added a secondary damage roll for high-damage changes, which should handle big guns like sniper rifles more effectively. Happy to change it again if you think it needs it.

TobleroneSwordfish (approved):

PR #22204, opened 2025-02-03
Deletes objcritter possums, use mobcritter instead. Adds morty to Kondaru.

Bartimeus973:
## About the PR 
Morty was a mobcritter on atlas and an objcritter everywhere else. This PR changes that and makes morty a mob everywhere.
Removes all instances of objcritter opossums.
Adds morty to Kondaru because he wasn't in the morgue for some reason.

## Why's this needed? 
Objcritters should eventually be fully phased out.

```changelog
(u)Bartimeus973
(+)Morty is now a real mob. This means you can grab him or put him inside pet carriers among others things.
```


code/mob/living/carbon/human/procs/emote.dm
```diff
@@ -68,15 +68,6 @@
...
 						#endif
-						var/possumMax = 15
-						for_by_tcl(responsePossum, /obj/critter/opossum)
-							if (!responsePossum.alive)
-								continue
-							if(!IN_RANGE(responsePossum, src, 4))
-								continue
-							if (possumMax-- < 0)
-								break
-							responsePossum.CritterDeath() // startled into playing dead!
```
TobleroneSwordfish, line 79: Why not convert this to the mob critter version?

TobleroneSwordfish (changes requested):

code/mob/living/carbon/human/procs/emote.dm
Bartimeus973: There is already a check doing the same thing for mob opossums just under this one

TobleroneSwordfish (approved):

PR #22217, opened 2025-02-06
AI can now call landline phones

AnomalousPotato:
[Feature][Silicons]

## About the PR 
- AI now has an internal phone that it can use to send and receive calls. It can access it via a AI Commands verb. It can disconnect and reconnect this phone at will.
- Telephones no longer delete and regenerate a new handset every time you pick it up.
- Disconnecting a phone will now end any active calls.

## Why's this needed? 
I believe it is good for the AI to have ways of interacting with the crew and station outside of a shell or radio channels. Additionally, I believe it fits in well with the 'retro' aesthetic. As the AI is able to interface with the wired electrical devices on the station, it should be able to interface with the landline phones.

~~Also talking to an artificial intelligence over a landline handset is really funny~~

## Changelog 

```changelog
(u)AnomalousPotato
(+)AI now has an internal landline telephone accessible through an AI Command verb.
```

### DIFF
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -1012,7 +1012,10 @@
 		equipped_talk_thing.talk_into(src, messages, secure_headset_mode, src.real_name, lang_id)
 
 	// AI phone fuckery
-		
+	if(isAI(src))
+		var/mob/living/silicon/ai/ai = src
+		if(ai.phone.answered)
+			ai.phone.handset.talk_into(src, messages, secure_headset_mode, src.real_name, lang_id)
 
 	switch (message_mode)
 		if ("headset", "secure headset", "right hand", "left hand")
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -1010,6 +1010,10 @@
 	var/obj/item/equipped_talk_thing = src.equipped()
 	if(equipped_talk_thing && equipped_talk_thing.flags & TALK_INTO_HAND && !message_mode)
 		equipped_talk_thing.talk_into(src, messages, secure_headset_mode, src.real_name, lang_id)
+
+	// AI phone fuckery
+		
+
 	switch (message_mode)
 		if ("headset", "secure headset", "right hand", "left hand")
 			talk_into_equipment(message_mode, messages, secure_headset_mode, lang_id)
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -1008,14 +1008,12 @@
 
 	// Do they have a phone?
 	var/obj/item/equipped_talk_thing = src.equipped()
-	if(equipped_talk_thing && equipped_talk_thing.flags & TALK_INTO_HAND && !message_mode)
+	if((equipped_talk_thing && equipped_talk_thing.flags & TALK_INTO_HAND && !message_mode) || isAI(src))
+		if(isAI(src))
+			var/mob/living/silicon/ai/ai = src
+			if(ai.phone.answered)
+				equipped_talk_thing = ai.phone.handset
 		equipped_talk_thing.talk_into(src, messages, secure_headset_mode, src.real_name, lang_id)
-
-	// AI phone fuckery
-	if(isAI(src))
-		var/mob/living/silicon/ai/ai = src
-		if(ai.phone.answered)
-			ai.phone.handset.talk_into(src, messages, secure_headset_mode, src.real_name, lang_id)
 
 	switch (message_mode)
 		if ("headset", "secure headset", "right hand", "left hand")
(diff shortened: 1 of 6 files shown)


DisturbHerb:
this severely nerfs handset frying.

AnomalousPotato:
me when i can't eat 500 lightly-fried phone handsets

JORJ949:
Would be nice if you could CTRL+CLICK a phone as AI to immediately call it with your internal one

Valtsu0:
> I can look at that when I get home. Maybe shift click though cause the intercoms already use it?

I'd prefer control + click, spesifically because intercoms already use it. For consistancy

AnomalousPotato:
Huh? Isn't it shift+click to lock an intercom? Same as bolting a door?

Valtsu0:
> Huh? Isn't it shift+click to lock an intercom? Same as bolting a door?

nevermind you are correct, disregard that

JORJ949:
There is a very clean proc for the AI clicking something with a certain key but you would have to check intercom code for that, its something like  "ai_act".

code/mob/living/silicon/ai.dm
```diff
@@ -363,6 +365,8 @@ or don't if it uses a custom topopen overlay
 		src.radio3.broadcasting = FALSE
 		src.internal_pda.name = "AI's Internal PDA Unit"
 		src.internal_pda.owner = "AI"
+		src.phone.name = "AI Internal Telephone"
```
JORJ949, line 368: There can be multiple AIs, hence multiple listings of "AI internal telephone", should probably be "G3-0R's Internal Phone" and change name wherever the AI's name itself is changed(I am pretty sure theres a proc that does this because it also has to update the internal PDA owner)

JORJ949: Nevermind just saw that you do have this in the rename proc.

AnomalousPotato:
Did the thing

code/obj/machinery/phone/phone.dm
```diff
@@ -42,7 +42,7 @@ TYPEINFO(/obj/machinery/phone)
 			if(istype(location,/area/station/security))
 				stripe_color = "#ff0000"
 				phone_category = "security"
-			else if(istype(location,/area/station/bridge))
+			else if(istype(location,/area/station/bridge) || isAI(src.loc))
```
TobleroneSwordfish, line 45: This should go at the top of the checks otherwise AI phones will get categorized under security if the AI core is in security

code/mob/living.dm
```diff
@@ -1010,6 +1010,13 @@
 	var/obj/item/equipped_talk_thing = src.equipped()
 	if(equipped_talk_thing && equipped_talk_thing.flags & TALK_INTO_HAND && !message_mode)
 		equipped_talk_thing.talk_into(src, messages, secure_headset_mode, src.real_name, lang_id)
+
+	// AI phone fuckery
+	if(isAI(src))
+		var/mob/living/silicon/ai/ai = src
+		if(ai.phone.answered)
+			ai.phone.handset.talk_into(src, messages, secure_headset_mode, src.real_name, lang_id)
```
TobleroneSwordfish, line 1018: Would rather this was only one `talk_into` call for clarity, use a variable to store the handset or put the logic for finding it into a proc

TobleroneSwordfish (changes requested):
Cool idea, a few issues with the code:
- The AI appears to be unable to hear the other person, either while in eye or core form
- Picking up the phone as an AI eye causes a runtime
image
- The ringing sound only plays around the physical location of the AI core, you'd never hear it if you're deployed as an eye or a shell
- The AI phone is listed twice on the phone list, once as "AI" and once as "AI - Name"

Honestly I'd suggest making the AI core phone unlisted and only allow the AI to call people. This would let you just have a verb to open the phone interface directly and bypass the issues of forwarding things to the eye, dealing with people spam calling to be annoying etc. Generally attracting the attention of the AI is pretty easy over radio and PDA messaging anyway so there's not a lot of reason to work around letting people call the core.

AnomalousPotato:
Alright, I fixed picking up a disconnected phone as ai eye causing a runtime, and merged the talk_into proc in living.dm, but I was unable to replicate not being able to hear the phone ringing while deployed, or the AI phone showing up twice. I haven't fixed not being able to hear the person on the other phone yet because I am very sick and all these funky letters are gibberish to me right now.

Valtsu0:
I do still really want this

AnomalousPotato:
I'd like to do it but I just haven't had much time lately. If anyone else wants to fix this before I get to it, they can

PR #22233, opened 2025-02-08
Add trailblazer pod component

FlameArrow57:

## About the PR 
Adds the Inferno Trailblazer pod component, which makes your pod leave behind blue flames instead of ion trails. Only found in loot crates

image

## Why's this needed? 
Thought it would be fun


code/datums/effects/system/ion_trail_follow.dm
```diff
@@ -17,21 +17,31 @@
 
 /datum/effects/system/ion_trail_follow/proc/on_vehicle_move(atom/movable/vehicle, atom/previous_loc, direction)
 	var/turf/T = get_turf(vehicle)
+	var/has_trailblazer
```
ZeWaka, line 20:
```suggestion
	var/has_trailblazer = FALSE
```

ZeWaka (approved):

PR #22247, opened 2025-02-09
Redo the deconstruction access-conditions for airlocks

M-Earthfire:
[Balance][Game objects]

## About the PR 

This PR removes the `DECON_NULL_ACCESS`-flag from airlocks and make them add the `DECON_NO_ACCESS` to them when they are emagged.

## Why's this needed? 

The need to remove ALL ACCESSES from airlocks when you want to deconstruct them made them one of the few things impossible for engineers to deconstruct without a RCD or the help of the HOP. And doors being harder to deconstruct than reinforced walls is just ridiculous.

Someone who can open a door should be able to get rid of it, if they so desire (e.g. renovations).

Also, the addition of the `DECON_NO_ACCESS` flag to emagged doors make it so you are able to deconstruct them when they are emagged. The `DECON_NULL_ACCESS`-flag made emagged doors the most obnoxious thing to remove/fix and is causing some discussion from time to time. Like this discussion which was the inspiration for this PR https://forum.ss13.co/showthread.php?tid=23597.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Airlocks no longer require all their access to be removed to be able to deconstructed, you just need to have access to them.
(+)Emagging a door removes the need to have access to the door in order to deconstruct it.
```

### DIFF
--- a/code/obj/machinery/door/airlock.dm
+++ b/code/obj/machinery/door/airlock.dm
@@ -11,7 +11,7 @@ var/global/list/cycling_airlocks = list()
 	name = "airlock"
 	icon = 'icons/obj/doors/SL_doors.dmi'
 	icon_state = "door_closed"
-	deconstruct_flags = DECON_NULL_ACCESS | DECON_WRENCH | DECON_CROWBAR | DECON_WELDER | DECON_SCREWDRIVER | DECON_MULTITOOL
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (approved):
Sounds reasonable to me 👍

PR #22115, opened 2025-01-26
`/obj/item/assembly`-Rework

M-Earthfire:
[Rework][Game Objects][Input Wanted]

## About the PR 

This PR aims to unify all assemblies under `/obj/item/assembly` into a common object

This object functions as a holder that relays signals and triggers to the corresponding parts of the assembly. It also dynamically handles sprite generation and assembly steps that fall in this category.

Assemblies will fall under a structure of 4 categories of items:

- `obj/item/trigger`: This is an object that will send the signal to the assembly to activate it's applier, e.g. a timer, proximity sensor or signaler
- `obj/item/applier`: This is an object that will cause the assemblies main effect once triggered, e.g. igniters or signalers. If a `target` is attached, the assembly tries to use the applier on it
- `obj/item/target`: This is an object the `applier` will try to interact with once it is activated, e.g. a pipebomb or a beaker
- `list/additional_components`: This list contains all components that are attached to the assembly for advanced interactions or crafting recipes, e.g. cabling or vuvuzelas for canbombs

The goal of this PR is to rework each assembly to work within this structure.

This PR adds multiple signals which will be send to the corresponding items when anything within the assembly changed, e.g. an item gets inserted (`COMSIG_ITEM_ASSEMBLY_ITEM_SETUP`), an item is checked if its ready to be build into an assembly (`COMSIG_ITEM_ASSEMBLY_COMBINATION_CHECK`), or if it gets activated while the assembly is secured (`COMSIG_ITEM_ASSEMBLY_ACTIVATION`) or while unsecured (`COMSIG_ITEM_ASSEMBLY_MANIPULATION`).

The above mentioned behaviour is used for a change in the handling of assemblies: Using an assembly in hand while unsecured opens the UI of it's components. Using it in hand while secured activates any timer on it.

To start an assembly, a suitable trigger and a suitable applier needs to be combined. The applier then adds a `/datum/component/assembly`-component to continue the craft, when needed. For this, the `TOOL_ASSEMBLY_APPLIER`-Toolflag has been added for suitable appliers and the `/datum/component/assembly/trigger_applier_assembly`-component for suitable components.

Also, new sprites were added which are used within the assembly for modular building of its sprite.

## Why's this needed? 

[...]

PR #22256, opened 2025-02-09
refactors electric chair construction

M-Earthfire:
[code quality][QoL][game objects]

## About the PR 

This PR moves `/obj/item/assembly/shock_kit` onto `/obj/item/shock_kit` and refactors its construction steps and the construction steps of electric chairs to use `/datum/component/assembly`.

## Why's this needed? 

The code of this assembly looked quite messy and, like with other construction interactions, only worked if you used one specific item on another instead of the other way around. 

This PR is atomized from #22115. At one later point i want to bind the electropack and thus this construction into assemblies again, but that requires a larger rework of this construction.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)You can construct shock packs now starting either with the helmet or the electropack.
```

### DIFF
--- a/code/obj/electric_chair.dm
+++ b/code/obj/electric_chair.dm
@@ -21,7 +21,7 @@
 	get_help_message(dist, mob/user)
 		. = "You can use a <b>multitool</b> to open the settings menu, or a <b>wrench</b> to disassemble it."
 
-	New(var/atom/new_location, var/obj/item/shock_kit/new_shock_kit)
+	New(atom/new_location, obj/item/shock_kit/new_shock_kit)
 		contextLayout = new /datum/contextLayout/experimentalcircle
 		..()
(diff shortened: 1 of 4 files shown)


code/obj/electric_chair.dm
```diff
@@ -21,35 +21,49 @@
 	get_help_message(dist, mob/user)
 		. = "You can use a <b>multitool</b> to open the settings menu, or a <b>wrench</b> to disassemble it."
 
-	New(atom/newLoc, obj/item/assembly/shock_kit/shock_kit)
+	New(var/atom/new_location, var/obj/item/shock_kit/new_shock_kit)
```
ZeWaka, line 24:
```suggestion
	New(atom/new_location, obj/item/shock_kit/new_shock_kit)
```

ZeWaka (approved):

PR #22258, opened 2025-02-09
TGUI robotics control

FlameArrow57:
[GAME OBJECTS][QoL]

## About the PR 
Converts robotics control UI to TGUI

image

image

## Why's this needed? 
tgui


code/obj/machinery/computer/robotics.dm
```diff
@@ -22,6 +37,121 @@
...
+						boutput(robot, SPAN_NOTICE("<b>Weapon Lock signal blocked!</b>"))
+						return
+					boutput(robot, SPAN_ALERT("<b>Weapon Lock activated!</b>"))
+				robot.weapon_lock = TRUE
+				robot.weaponlock_time = 120
+				robot.uneq_active()
+				logTheThing(LOG_COMBAT, usr, "has activated [constructTarget(robot,"combat")]'s weapon lock (120 seconds).")
+				for (var/obj/item/roboupgrade/upgrade in robot.contents)
+					if (upgrade.activated)
+						upgrade.activated = 0
```
mordent-goonstation, line 130: Prefer `TRUE`/`FALSE` if these are true/false, which I *think* they are.

code/obj/machinery/computer/robotics.dm
```diff
@@ -22,6 +37,121 @@
...
+			"cyborg_lock_times" = src.cyborg_lock_times,
+			"cyborg_killswitch_times" = src.cyborg_killswitch_times,
+			"ghostdrone_names" = src.ghostdrone_names
+		)
+
+	ui_act(action, params)
+		. = ..()
+		if (.)
+			return
+		. = TRUE
```
mordent-goonstation, line 68: Do on a case-by-case basis, please. It might seem redundant, but applying it here regardless is not generally how we do it for the sake of making updating the UI a conscious choice. Just looking at your flow, a weapons lock on an emagged cyborg doesn't have any reason to update the UI, for example.

code/obj/machinery/computer/robotics.dm
```diff
@@ -22,6 +37,121 @@
...
+		switch (action)
+			if ("start_ai_killswitch")
+				var/obj/item/card/id/I = usr.equipped()
+				if (istype(I))
+					if(src.check_access(I))
+						var/mob/living/silicon/ai/ai_player = by_type[/mob/living/silicon/ai][params["index"]]
```
mordent-goonstation, line 74: I feel uneasy about just using an index-accessed by_type here. Why not pass the `ref`s of them around?

code/obj/machinery/computer/robotics.dm
```diff
@@ -22,6 +37,121 @@
...
+						logTheThing(LOG_COMBAT, usr, "has activated the AI killswitch process on [constructTarget(message,"combat")]")
+						if(message.client)
+							boutput(message, SPAN_ALERT("<b>AI Killswitch process activated.</b>"))
+							boutput(message, SPAN_ALERT("<b>Killswitch will engage in 3 minutes.</b>"))
+						ai_player.killswitch = TRUE
+						ai_player.killswitch_at = TIME + 3 MINUTES
+					else
+						boutput(usr, SPAN_ALERT("Access Denied."))
+			if ("stop_ai_killswitch")
+				var/mob/living/silicon/ai/ai_player = by_type[/mob/living/silicon/ai][params["index"]]
```
mordent-goonstation, line 86: Use refs on the data instead of index-based.

code/obj/machinery/computer/robotics.dm
```diff
@@ -22,6 +37,121 @@
...
+						logTheThing(LOG_COMBAT, usr, "has activated the robot killswitch process on [constructTarget(robot,"combat")]")
+						if(robot.client)
+							boutput(robot, SPAN_ALERT("<b>Killswitch process activated.</b>"))
+							boutput(robot, SPAN_ALERT("<b>Killswitch will engage in 1 minute.</b>"))
+						robot.killswitch = TRUE
+						robot.killswitch_at = TIME + 1 MINUTE
+					else
+						boutput(usr, SPAN_ALERT("Access Denied."))
+			if ("stop_silicon_killswitch")
+				var/mob/living/silicon/robot/robot = src.tracked_cyborgs[params["index"]]
```
mordent-goonstation, line 110: Use `ref` on the data, instead of index-based access.

code/obj/machinery/computer/robotics.dm
```diff
@@ -22,6 +37,121 @@
...
+				robot.killswitch_at = 0
+				robot.killswitch = 0
```
mordent-goonstation, line 112:
```suggestion
				robot.killswitch = FALSE
```

code/obj/machinery/computer/robotics.dm
```diff
@@ -22,6 +37,121 @@
...
+					boutput(robot, SPAN_ALERT("<b>Weapon Lock activated!</b>"))
+				robot.weapon_lock = TRUE
+				robot.weaponlock_time = 120
+				robot.uneq_active()
+				logTheThing(LOG_COMBAT, usr, "has activated [constructTarget(robot,"combat")]'s weapon lock (120 seconds).")
+				for (var/obj/item/roboupgrade/upgrade in robot.contents)
+					if (upgrade.activated)
+						upgrade.activated = 0
+						boutput(robot, SPAN_ALERT("<b>[upgrade] was shut down by the Weapon Lock!</b>"))
+					if (istype(upgrade, /obj/item/roboupgrade/jetpack))
```
mordent-goonstation, line 132: Are there any other ones that have special handling? e.g. mag boot like one? Shields?

Could use use the `upgrade_deactivate` procs on the upgrades instead, or do those have extra behavior?

code/obj/machinery/computer/robotics.dm
```diff
@@ -34,212 +164,74 @@
...
-				dat += "Time left:[timeleft]"
-				if (!isAI(user))
-					dat += " | <A href='?src=\ref[src];gib=2;ai=\ref[A]'>Cancel</A>"
-				dat += "<BR>"
-
-		dat += "<BR> Connected Cyborgs<BR>"
-		dat += " *------------------------------------------------*<BR>"
+		if (A.killswitch)
+			var/timeleft = round((A.killswitch_at - TIME)/10, 1)
+			timeleft = "[(timeleft / 60) % 60]:[add_zero(num2text(timeleft % 60), 2)]"
```
mordent-goonstation, line 190: Can leave this formatting to the UI.

code/obj/machinery/computer/robotics.dm
```diff
@@ -34,212 +164,74 @@
...
 			else if(R.stat)
-				dat += " Not Responding |"
+				src.cyborg_statuses += "Not Responding"
 			else
-				dat += " Operating Normally |"
+				src.cyborg_statuses += "Operating Normally"
+
 			if(R.cell)
-				dat += " Battery Installed ([R.cell.charge]/[R.cell.maxcharge]) |"
+				src.cyborg_cell_charges += "Battery Installed ([R.cell.charge]/[R.cell.maxcharge])"
```
mordent-goonstation, line 209: Again, can provide this info to the UI for it to format.

code/obj/machinery/computer/robotics.dm
```diff
@@ -34,212 +164,74 @@
...
 			else
-				dat += " No Cell Installed |"
+				src.cyborg_cell_charges += "No Cell Installed"
+
 			if(R.module)
-				dat += " Module Installed ([R.module.name]) |"
+				src.cyborg_modules += "Module Installed ([R.module.name])"
```
mordent-goonstation, line 214: Just the name, format in the UI.

code/obj/machinery/computer/robotics.dm
```diff
@@ -34,212 +164,74 @@
...
+	src.ai_names = list()
+	src.ai_statuses = list()
+	src.ai_killswitch_times = list()
 
+	src.cyborg_names = list()
+	src.cyborg_statuses = list()
+	src.cyborg_cell_charges = list()
+	src.cyborg_modules = list()
+	src.cyborg_lock_times = list()
+	src.cyborg_killswitch_times = list()
```
mordent-goonstation, line 177: This is a decently janky structure for this. Prefer a list of `cyborg_data` with each of them having whatever info you need to convey about a cyborg. Accessing multiple named params via a hopefully shared index adds unnecessary complexity.

code/obj/machinery/computer/robotics.dm
```diff
@@ -34,212 +164,74 @@
...
-							logTheThing(LOG_COMBAT, usr, "has activated the AI killswitch process on [constructTarget(message,"combat")]")
-							if(message.client)
-								boutput(message, SPAN_ALERT("<b>AI Killswitch process activated.</b>"))
-								boutput(message, SPAN_ALERT("<b>Killswitch will engage in 3 minutes.</b>"))
-							A.killswitch = TRUE
-							A.killswitch_at = TIME + 3 MINUTES
-					else
-						boutput(usr, SPAN_ALERT("Access Denied."))
+			if(!R.weapon_lock)
+				src.cyborg_lock_times += "-1"
```
mordent-goonstation, line 219: Prefer `null` over a magic number here.

tgui/packages/tgui/interfaces/RoboticsControl.tsx
```diff
@@ -0,0 +1,251 @@
...
+ * @author FlameArrow57 (https://github.com/FlameArrow57)
+ * @license ISC
+ */
+
+import { Button, NoticeBox, Section, Table } from 'tgui-core/components';
+
+import { useBackend } from '../backend';
+import { Window } from '../layouts';
+
+interface GPSInfo {
```
mordent-goonstation, line 13:
```suggestion
interface RoboticsControlData {
```

tgui/packages/tgui/interfaces/RoboticsControl.tsx
```diff
@@ -0,0 +1,251 @@
...
+  cyborg_statuses: string[];
+  cyborg_cell_charges: string[];
+  cyborg_modules: string[];
+  cyborg_lock_times: string[];
+  cyborg_killswitch_times: string[];
+  ghostdrone_names: string[];
+}
+
+export const RoboticsControl = () => {
+  const { act, data } = useBackend<GPSInfo>();
```
mordent-goonstation, line 29:
```suggestion
  const { act, data } = useBackend<RoboticsControlData>();
```

tgui/packages/tgui/interfaces/RoboticsControl.tsx
```diff
@@ -0,0 +1,251 @@
...
+ */
+
+import { Button, NoticeBox, Section, Table } from 'tgui-core/components';
+
+import { useBackend } from '../backend';
+import { Window } from '../layouts';
+
+interface GPSInfo {
+  user_is_ai: boolean;
+  user_is_cyborg: boolean;
```
mordent-goonstation, line 15:
```
  user_is_ai: BooleanLike;
  user_is_cyborg: BooleanLike;
```
(With relevant import)

tgui/packages/tgui/interfaces/RoboticsControl.tsx
```diff
@@ -0,0 +1,251 @@
...
+  ai_statuses: string[];
+  ai_killswitch_times: string[];
+  cyborg_names: string[];
+  cyborg_statuses: string[];
+  cyborg_cell_charges: string[];
+  cyborg_modules: string[];
+  cyborg_lock_times: string[];
+  cyborg_killswitch_times: string[];
+  ghostdrone_names: string[];
+}
```
mordent-goonstation, line 26: Suggesting below structure for AI/Cyborg/Ghostdrone data, feel free to adjust if needed (obviously untested). I did not include the top-level ones for user_is_ai or user_is_cyborg
```
interface RoboticsControlData {
  ais: AIData[];
  cyborgs: CyborgData[];
  ghostdrones: GhostdroneData[];
}

interface AIData {
  name: string;
  mobRef: string; // new addition, used to pass down the specific ref so actions can use that instead of index
  status: string;
  killswitchTime: number | null;
}

interface CellData {
  charge: number;
  maxCharge: number;
}

interface CyborgData {
  name: string;
  mobRef: string;
  status: string;
  cell: CellData | null;
  moduleName: string;
  lockTime: number | null;
  killswitchTime: number | null;
}

interface GhostdroneData {
  name: string;
  mobRef: string;
}
```

mordent-goonstation (changes requested):
Broadly:

There's no real need to have all of that stuff stored as `var` on the item, especially as you're regenerating it in `ui_data` anyway. I don't see it being used anywhere. You can have utility procs for generating the data, but rather than updating a bunch of `var`s with what is really display data, just have the utility proc return the slice of the data they care about.

Suggested a change to the data structure that will require a bunch of knock on stuff, leaving that to you to do rather than trying to guess the resultant changes in review form.

tgui/packages/tgui/interfaces/RoboticsControl.tsx
FlameArrow57: What's the difference between boolean and BooleanLike?

code/obj/machinery/computer/robotics.dm
FlameArrow57: No clue, this is just from previous code. Assumed no change needed

code/obj/machinery/computer/robotics.dm
FlameArrow57: Ref needs to be passed through UI data and returned?

tgui/packages/tgui/interfaces/RoboticsControl.tsx
FlameArrow57: Not familiar with this formatting, is this equivalent to passing associated lists through ui_data()?

code/obj/machinery/computer/robotics.dm
mordent-goonstation: You can pass a `\ref` through to `ui_data`, then pass that back in the `act`. Verify that they're a valid target still in `ui_act`, but otherwise yes, they're just strings for the purposes of `ui_data`.

tgui/packages/tgui/interfaces/RoboticsControl.tsx
mordent-goonstation: boolean is a built-in TypeScript type for `true`/`false`. When serializing data in `ui_data`, because BYOND represents booleans as `1` and `0` (we have macros for `TRUE`/`FALSE`) they get serialized as `1` and `0`. a BooleanLike is a tgui thing (see any other usage) that hints that it shouldn't be treated as a boolean for some special cases. A common one is when rendering, e.g. `{isFoo && <Whatever />}`, if `isFoo` is `0` then this will short-circuit to rendering `0` instead of nothing. That's why we do a lot of `!!` in render code, to convert BooleanLike to actual booleans (there are other ways, `!!` is just a very compact way of doing so).

tgui/packages/tgui/interfaces/RoboticsControl.tsx
mordent-goonstation: Sort of.
```
  ais: AIData[];
```
The field on `ui_data` here is `ais`, with a value that is an array of associated lists (using DM terms) that all correspond to the structure of an `AIData`, which if you look a little lower is these four fields.
```
interface AIData {
  name: string;
  mobRef: string; // new addition, used to pass down the specific ref so actions can use that instead of index
  status: string;
  killswitchTime: number | null;
}
```

mordent-goonstation:
Iterating on the design a little, seemed overly large. Pushed something to maybe help, will look at improving the lock/killswitch buttons soon.

FlameArrow57:
> Iterating on the design a little, seemed overly large. Pushed something to maybe help, will look at improving the lock/killswitch buttons soon.

Okay I took a look, seems pretty good. Nothing in mind for change

mordent-goonstation:
For posterity, I think there are a lot of UX improvements that could be made here, but this is at least a port of the current UI so a step in the right direction.

PR #22259, opened 2025-02-09
adds children of `/datum/component/assembly` to check for borg and item arms

M-Earthfire:
[Internal][Bug]

## About the PR 

Adds a proc to movables, `/atom/movable/proc/assembly_cant_be_removed()`. This one returns TRUE when an item should not be able to be turned into a different item in the three newly added children of `/datum/component/assembly`:

- `/datum/component/assembly/consume_self`: checks the proc on the parent of this component on interaction
- `/datum/component/assembly/consume_other` checks the proc on the checked_atom on interaction
- `/datum/component/assembly/consume_all` checks the proc on all atoms on interaction

Currently, this proc only return sTRUE on items used in item arms and items you cannot drop (e.g. borg items)

## Why's this needed? 

With the introduction of #22230, i accidently introduced a bug where borgs can turn their welding tool into a flamethrower in progress, essentially duping flamethrowers.

Like borg arms, item arms can always cause issues with combinations or interactions that make you drop the respectable item. The procs and components should serve as a unified method to catch and prevent these cases during item combinations.

### DIFF
--- a/code/atom.dm
+++ b/code/atom.dm
@@ -304,10 +304,6 @@ TYPEINFO(/atom)
 /atom/proc/ex_act(var/severity=0,var/last_touched=0, var/power=0, var/datum/explosion/explosion=null)
 	return
 
-/// Use this proc to check if you can use and permanently change this item via a combination
-/atom/proc/is_combination_consumeable()
-	return TRUE
-
 /atom/proc/reagent_act(var/reagent_id,var/volume,var/datum/reagentsholder_reagents)
 	if (!istext(reagent_id) || !isnum(volume) || volume < 1)
 		return 1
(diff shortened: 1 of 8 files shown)


code/atom.dm
```diff
@@ -304,6 +304,10 @@ TYPEINFO(/atom)
 /atom/proc/ex_act(var/severity=0,var/last_touched=0, var/power=0, var/datum/explosion/explosion=null)
 	return
 
+/// Use this proc to check if you can use and permanently change this item via a combination
+/atom/proc/is_combination_consumeable()
+	return TRUE
```
mordent-goonstation, line 309: What is a "combination"? I don't really know what this is checking and think this is a poorly named proc.

(Also: consumable, not consumeable)

PR #22261, opened 2025-02-09
Enzymatic Reclaimers can be broken, and spew gibs when broken.

pgmzeta:
[feature][game objects]
## About the PR 
* Adds a new equipment fault type: messy
  * By default, spews cleanables of various machine guts: machine debris, oil, generic, and harmless glitter
  * This list can be overridden, which the enzymatic reclaimer does with several different types of gibs.
  * Uses an increasing probability to spawn cleanables over time, resetting to a base probability once done.
* Broken and unpowered enzymatic reclaimers stop processing meat
* You cannot start the reclaimer if there is no power or its broken
* The grinder uses 100 WATTS while processing meat (or broken)
* They can be fixed with a wrench and screwdriver (maybe ask your local friendly roboticist?)
* Grinders will automaticlaly continue processing once they are fixed or re-powered.
* Add bullet action to reclaimers to roll a chance to break when hit with kinetics. They already break on weak explosions.

## Why's this needed? 
More station equipment reactivity to explosions and gunfire.

## Demo
https://github.com/user-attachments/assets/6f82139e-915e-4521-bf73-14955486b059

## Changelog 

```changelog
(u)glowbold
(+)Enzymatic reclaimers can be broken by gunfire and explosions. Use common tools to fix them!
(+)Enzymatic reclaimers require power to process meat and bodies.
```

### DIFF
--- a/code/datums/components/equipment_fault.dm
+++ b/code/datums/components/equipment_fault.dm
@@ -507,3 +507,61 @@ TYPEINFO(/datum/component/equipment_fault/leaky)
 		var/datum/reagents/temp_fluid_reagents = new /datum/reagents(5)
 		temp_fluid_reagents.add_reagent(pick(src.reagent_list), 5)
 		target_turf.fluid_react(temp_fluid_reagents, temp_fluid_reagents.total_volume)
+
+TYPEINFO(/datum/component/equipment_fault/messy)
+	initialization_args = list(
+		ARG_INFO("tool_flags", DATA_INPUT_BITFIELD, "Tools Required", TOOL_PULSING | TOOL_SCREWING),
+		ARG_INFO("cleanables", DATA_INPUT_LIST_PROVIDED, "Cleanable List", list(\
+			/obj/decal/cleanable/machine_debris=40,\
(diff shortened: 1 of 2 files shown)


Tarmunora (commented):
I like the idea of the reclaimer making a mess when damaged.
However, it feels weird that it just generates blood/etc out of thin air while damaged, in contrast to needing to be fed meat/bodies in normal operation

pgmzeta:
> I like the idea of the reclaimer making a mess when damaged. However, it feels weird that it just generates blood/etc out of thin air while damaged, in contrast to needing to be fed meat/bodies in normal operation

I've changed this to only use the messy fault if it's processing. If it's not processing, it'll use the grumble fault. 

While broken, it will still tick down the process timer while spewing gibs, effectively "using up" the meat that's in the reclaimer. If the process timer is zero after being broken, it will switch to the grumble fault, maintaining the current fix interactions i.e. if fixed halfway, it will carry over the remaining step to the new fault.

Tarmunora (approved):

PR #22281, opened 2025-02-12
Cyborgs affected by radiation get visual noise

pgmzeta:
[silicons][feature]
## About the PR 
* Silicons affected by radiation now get a status effect: Radiological Interference
* This status affect applies one or more visual noise overlays via invisible status effects
* The strength of the visual noise is dependent on the intensity of radiation.

Miscelaneous related changes:
* Radiation pulses now affect all mobs
* Radweed now affects all living mobs
* Port-a-sci radiation dosing failures now affects all living mobs
* Singularity radiation now affects all living mobs

## Why's this needed? 
Silicons have no interaction with or way to detect radiation.
This mimics a real-life effect with CCD camera sensors, and provides a unique-but-not-dehabiliating debuff that fits silicons.

## Demo (updated)
https://youtu.be/YJwyCPgEbns

## Changelog 

```changelog
(u)glowbold
(*)Radiation interferes with silicon visual sensors based on intensity.
```


FlameArrow57:
Really cool, I feel like maybe it should be restricted to only the lowest radiation dosage level of noise tho. It's really nice not having intense screen interfering effects like flashes, ex. in SS14

cropseyy:
For accessibility it would be great if this respected the 'dark flashes' option. Whilst it's not strobing the bright/dark contrast could be unpleasant.

JORJ949:
From the looks of it this would affect the AI only from where its core is, rather than where the "optical sensor" (eye) is, potentially even if you are deployed to the eye. Ideally this should only affect wherever the AI is currently viewing and whichever mob it is actively operating, e.g. no radiation on an AI shell because the core is irradiated.

pgmzeta:
Updated version of the effect:
https://youtu.be/YJwyCPgEbns

code/modules/status_system/statusEffects.dm
```diff
@@ -3375,3 +3375,106 @@
...
+
+	onAdd(Sv)
+		. = ..()
+		var/mob/living/silicon/S = owner
+		S.addOverlayComposition(/datum/overlayComposition/silicon_rad_extreme)
+
+	onRemove()
+		. = ..()
+		var/mob/living/silicon/S = owner
+		S.removeOverlayComposition(/datum/overlayComposition/silicon_rad_extreme)
```
Tarmunora, line 3480: For code cleanliness I think these should have a common parent that handles the onAdd/onRemove behaviour

Tarmunora (approved):

pgmzeta:
> From the looks of it this would affect the AI only from where its core is, rather than where the "optical sensor" (eye) is, potentially even if you are deployed to the eye. Ideally this should only affect wherever the AI is currently viewing and whichever mob it is actively operating, e.g. no radiation on an AI shell because the core is irradiated.

I added a check for clients, so the AI is only affected when in the mainframe. I reason that in aieye form, they're looking via security cameras which don't have this effect currently applied. There is some handwaving away about the fact that security cameras are more radiation hardened than cyborgs, somehow. 

This change should also ensure that the deployed cyborg shell or eyebot do not get radiation clicks even if the core is irradiated; they'll only get the effect if the eyebot/shell is in range and someone is connected to it.

TobleroneSwordfish:
I kind of wish we could fade this in or out as you move closer and further from a source, but that would probably require some kind of expensive system to keep track of radiation per turf and this still looks really good.

ZeWaka:
> I kind of wish we could fade this in or out as you move closer and further from a source, but that would probably require some kind of expensive system to keep track of radiation per turf and this still looks really good.

you can still fade it in and out on de/application

PR #22290, opened 2025-02-14
Allow Modifier Keybinds

### LINKED ISSUE #22071: Cannot bind shift key.
### Labels

[BUG][MAJOR]

### Description

Attempting to bind a key to just shift will not work.

### Steps to reproduce

1. Select a key to bind
2. Try to bind it to just shift

### Expected Behavior

for me to be able to bind shift

Reported by: goatreal
Client version: 515.1642
Active test merges: nothing
Round log date: 2025-01-22-02-28
Reported on: 2025-01-22 02:39:08
Round ID: 198234
Map: OSHAN

Garash2k:
## About the PR 
Allows these special keys to be bound: shift, ctrl, alt, space, and the arrow keys.
image

## Why's this needed? 
Fixes #22071

Note: not supporting "fn" I don't have a laptop I can install byond on.

## Changelog 

```changelog
(u)Garash
(+)Keybind menu now allows to bind special keys such as shift, ctrl, alt...
```
[UI][Bug]


ZeWaka:
the keybind system i wrote entirely all on my own is very bad code and evil spite infused hatred

Garash2k:
Yeah, there's a bizarre bug, which also happened with the previous ui, where keybinds get duplicated, I don't quite understand. But letting people bind those special keys was a feature lost because of the new UI, so it felt worth fixing.

tgui/packages/tgui/interfaces/Keybinds/formatKeyboardEvent.ts
```diff
@@ -44,8 +43,8 @@ export const formatKeyboardEvent = (event: React.KeyboardEvent): string => {
 
   if (isStandardKey(event)) {
     const key = event.key.toUpperCase();
-    text += KEY_CODE_TO_BYOND[key] || key;
+    text += KEY_CODE_TO_BYOND[key] || key === ' ' ? 'Space' : key;
   }
 
-  return text;
+  return text.replace(/\+$/, '');
```
mordent-goonstation, line 49: Could we get a comment here explaining why we're trimming trailing `+` characters? Or, like, why it might have them, etc.

tgui/packages/tgui/interfaces/Keybinds/formatKeyboardEvent.ts
```diff
@@ -44,8 +43,8 @@ export const formatKeyboardEvent = (event: React.KeyboardEvent): string => {
 
   if (isStandardKey(event)) {
     const key = event.key.toUpperCase();
-    text += KEY_CODE_TO_BYOND[key] || key;
+    text += KEY_CODE_TO_BYOND[key] || key === ' ' ? 'Space' : key;
```
mordent-goonstation, line 46: Is it not possible to include a `' ': 'Space'` in the `KEY_CODE_TO_BYOND` map? Notably, there's this as the reverse: https://github.com/tgstation/tgui-core/commit/95608933d9248e7b91506ed365b5049ff0164bd6#L45

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -19,22 +20,44 @@ interface KeybindProps {
...
+          event.preventDefault();
+          if (!isFocused) {
+            return;
+          }
+          if (!lastEvent) {
+            return;
+          }
+          // If we release a key and we're still in keybind editing mode, it's time to bind whatever we had.
+          setFocusedKey('');
+          const value = formatKeyboardEvent(lastEvent);
```
mordent-goonstation, line 58: Do we want to clear out the `lastEvent` here as well, or not? Not saying do this, as I haven't tested, just seems odd to hold onto if we're binding it then moving on.

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -19,22 +20,44 @@ interface KeybindProps {
 export const Keybind = (props: KeybindProps) => {
   const { act } = useBackend<KeybindsData>();
   const { keybind, isFocused, setFocusedKey } = props;
+
+  const [lastEvent, setLastEvent] =
+    useState<React.KeyboardEvent<HTMLDivElement>>();
```
mordent-goonstation, line 25: Probably `HTMLElement`, because button being a `div` under the hood is an implementation detail that I don't think we care about. If this makes TS angry, fine to leave as `HTMLDivElement`.

tgui/packages/tgui/interfaces/Keybinds/Keybind.tsx
```diff
@@ -19,22 +20,44 @@ interface KeybindProps {
...
+          if (!isFocused) {
+            return;
+          }
           if (isEscape(event.key)) {
             setFocusedKey('');
             return;
           }
           if (!isStandardKey(event)) {
+            // Store the event in case the key is released in which case we'll want to bind to this non-Standard key.
+            setLastEvent(event);
```
mordent-goonstation, line 41: Be aware that state setting causes re-renders. Is that acceptable here, or would this being in a `useRef` and having its state managed more directly be more appropriate?

tgui/packages/tgui/interfaces/Keybinds/formatKeyboardEvent.ts
Garash2k: Code improved to only add `+`s when needed, so they're present for bindings such as "SHIFT+S" but absent for only "SHIFT"

tgui/packages/tgui/interfaces/Keybinds/formatKeyboardEvent.ts
Garash2k: I thought it wasn't, but turns out it is possible!

mordent-goonstation (approved):

tgui/packages/tgui/interfaces/Keybinds/formatKeyboardEvent.ts
```diff
@@ -3,16 +3,16 @@ import { isEscape, KEY } from 'tgui-core/keys';
 const DOM_KEY_LOCATION_NUMPAD = 3;
 const KEY_CODE_TO_BYOND: Record<string, string> = {
   DEL: 'Delete',
-  DOWN: 'South',
+  ARROWDOWN: 'South',
```
ZeWaka, line 6: Is this correct?

Garash2k: I think so. I wasn't able to bind arrow keys before the fix, after this fix I could bind them.

It's also what the specs says it should be: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#navigation_keys

ZeWaka: Weird. IIRC I got those values directly from pressing keys via Byond.

Garash2k: I figured it out, it's not simply the browser's `KeyboardEvent`, it's a *React event*.

`export const formatKeyboardEvent = (event: React.KeyboardEvent): string => { ... };`

https://react.dev/reference/react-dom/components/common#react-event-object

>(React event objects) conforms to the same standard as the underlying DOM events, but fixes some browser inconsistencies.

Garash2k: Also of note, this is formatKeyboardEvent.ts, made for the new Keybind.tsx window in #21681
It's not keys.ts, which is what I'm assuming you made and remains unchanged.
https://github.com/goonstation/goonstation/blob/09f7daa21458f602b33341c8de9b6a2b01a98a15/tgui/packages/common/keys.ts#L38

ZeWaka:
yuh

PR #22291, opened 2025-02-14
TGUI-ify MechComp Signal Editing Windows

Facsimil:
[UI] [FEATURE]

## About the PR 

Replaces `input` with `tgui_input_list` and `tgui_input_text` where appropriate in `mechComp_signals.dm`.

## Why's this needed? 

TGUI inputs are better than the default `input` UI.

Also consistency with the ever-increasing list of things that use TGUI.

## Changelog 

```changelog
(u)Facsimile
(+)MechComp signal editing now uses TGUI.
```

### DIFF
--- a/code/datums/components/mechComp_signals.dm
+++ b/code/datums/components/mechComp_signals.dm
@@ -308,7 +308,7 @@ TYPEINFO(/datum/component/mechanics_holder)
 		boutput(user, SPAN_ALERT("Components need to be within a range of 14 meters to connect."))
 		return
 
-	var/typesel = input(user, "Use [parent] as:", "Connection Type") in list("Trigger", "Receiver", "*CANCEL*")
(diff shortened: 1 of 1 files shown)


code/datums/components/mechComp_signals.dm
```diff
@@ -308,7 +308,7 @@ TYPEINFO(/datum/component/mechanics_holder)
 		boutput(user, SPAN_ALERT("Components need to be within a range of 14 meters to connect."))
 		return
 
-	var/typesel = input(user, "Use [parent] as:", "Connection Type") in list("Trigger", "Receiver", "*CANCEL*")
+	var/typesel = tgui_input_list(user, "Use [parent] as:", "Connection Type", list("Trigger", "Receiver", "*CANCEL*"))
```
ZeWaka, line 311: i don't know if it makes sense to use lists and not alerts for 3/2 options

Facsimil: Sure

mordent-goonstation (approved):
Sure, I guess. Assuming it works fine, seems fine.

PR #22298, opened 2025-02-14
Add Piano Supply Packs

Facsimil:
[GAME OBJECTS] [FEATURE]

## About the PR 

Adds supply packs to cargo for the Player Piano, Piano, and Grand Piano.

## Why's this needed? 

There may be people on a map that want to play piano, but can't since someone else is using it. Since there's a limited amount of pianos on maps.

Also added the Player Piano for maps that don't have one (e.g. Atlas).

## Changelog 

```changelog
(u)Facsimile
(*)Adds supply packs to cargo for the Player Piano, Piano, and Grand Piano.
```

### DIFF
--- a/code/modules/economy/supply_packs.dm
+++ b/code/modules/economy/supply_packs.dm
@@ -2056,33 +2056,32 @@ ABSTRACT_TYPE(/datum/supply_packs/complex)
 	contains = list(/obj/item/instrument/bass)
 	containertype = /obj/storage/crate/wooden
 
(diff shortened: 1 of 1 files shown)


TDHooligan:
these are large items so youll want to use frames, similar to the bar kit

Facsimil:
> these are large objects, so youll want to use frames, similar to the bar kit

Sure.

Lefinch:
Yes please.

PR #22319, opened 2025-02-15
Random Event: Meatier Shower

pgmzeta:
[events][feature]
## About the PR 
* Add new random event: "Meatier Shower", disabled by default
* Mimics the "Meteor Shower" event i.e. announcement, timing.
* Throws 30-60 bodies at the station and gibs them when they land.
* Add new throw flag: `THROW_GIB`. If used, it always gibs mobs when the throw ends.

## Why's this needed? 
Silly random event for admins. Maybe could be enabled during april fools and/or halloween?


TobleroneSwordfish (changes requested):
Suitably horrifying idea, but the code is just the meteor shower code copy pasted and modified. Please refactor to make it a subtype instead.

PR #22323, opened 2025-02-15
Add pod scout engine

FlameArrow57:

## About the PR 
Adds a new Scout Engine pod engine. This is an engine that provides the highest speed and warp travel recharge of the engines, but little power, low enough to not be able to operate weapons. More for reconnaissance

## Why's this needed? 
Possibly adds to pod component bloat (not sure)
Aim was to create a more unique engine that can't operate a lot of pod parts but still is useful (in terms of speed). Use cases are fast navigation around Z levels and outrunning other pods

## Changelog 

```changelog
(u)FlameArrow57
(+)Added a new Scout Engine pod engine. Compared to other engines, it provides the highest speed and warp travel recharge time, but provides little power to operate most pod parts that require power
```


ZeWaka (approved):

Tarmunora (commented):
Worth noting that engines currently have no effect on pod speed, which is entirely determined by the body/armor

FlameArrow57:
> Worth noting that engines currently have no effect on pod speed, which is entirely determined by the body/armor

x_x will revise, thought it seemed faster

FlameArrow57:
> Worth noting that engines currently have no effect on pod speed, which is entirely determined by the body/armor

Okay I thought on this more, I think I'd probably rather close this and just stick with the afterburner component, unless you think a pod that is just low power and high warprecharge is interesting, LMK

Tarmunora:
> or maybe a future rework making speed actually set by engine would be better?

having engine affect speed again (Pretty sure it used to a long time ago) might be worth looking into, yeah.

PR #22329, opened 2025-02-16
Add pod afterburner component

FlameArrow57:

## About the PR 
Adds a new Afterburner pod component which raises your pod's maximum velocity by 50% increases acceleration slightly for 5 seconds

## Why's this needed? 
More pod components, use cases for this would be to close the distance to other pods or increase distance from other pods

## Changelog 

```changelog
(u)FlameArrow57
(+)Added a new pod Afterburner component to ship fabricators, which increases your pod's maximum velocity and acceleration for 5 seconds, with a 20 second recharge time.
```


code/modules/transport/pods/secondary_system.dm
```diff
@@ -585,6 +594,24 @@
...
+		if (ON_COOLDOWN(src, "thruster_movement", 30 SECONDS))
+			boutput(user, "[src.ship.ship_message("Afterburner is recharging! [round(GET_COOLDOWN(src, "thruster_movement") / 10, 0.1)] seconds left.")]")
+			return
+
+		// spawn to allow button clunk sound to play right away
+		SPAWN(0)
+			for (var/i in 1 to 10)
+				step(src.ship, src.ship.dir)
+				sleep(1 SECOND)
+
```
Tarmunora, line 614: This feels pretty kludgy - would rather see it interact with the movement controller (temporarily raise accel/max speed? Force additional velocity in facing direction?) vs a step-sleep loop

FlameArrow57: Sure, will revise

code/modules/transport/pods/secondary_system.dm
```diff
@@ -585,6 +595,31 @@
...
+	power_used = 50
+
+	use_thrusters(mob/user)
+		if (ON_COOLDOWN(src, "thruster_movement", 20 SECONDS))
+			boutput(user, "[src.ship.ship_message("Afterburner is recharging! [round(GET_COOLDOWN(src, "thruster_movement") / 10, 0.1)] seconds left.")]")
+			return
+
+		// spawn to allow button clunk sound to play right away
+		SPAWN(0)
+			boutput(user, "[src.ship.ship_message("Afterburner is now active!")]")
```
TobleroneSwordfish, line 612: Ideally this should have some kind of visual or audio feedback, maybe just a fwoosh sound for people in the cabin and a little puff of flame or something to show other pod pilots that it's been used.

code/datums/movement_controller/pod.dm
```diff
@@ -136,7 +136,7 @@
 				if (!input_x && !input_y)
 					vel_max = velocity_max_no_input
 
-				vel_max /= (owner.speed ? owner.speed : 1)
+				vel_max /= (owner.speed ? owner.speed : 1) * src.owner.afterburner_speed_mod
```
TobleroneSwordfish, line 139: This is reducing the top speed as written.

code/modules/transport/pods/secondary_system.dm
```diff
@@ -585,6 +600,30 @@
 			src.ship.myhud.update_states()
 		boutput(usr, SPAN_NOTICE("Thrusters will now provide ship movement to the [src.turn_dir]."))
 
+/obj/item/shipcomponent/secondary_system/thrusters/afterburner
```
TobleroneSwordfish, line 603: It'd be really nice if any new pod components could come with a sprite, we have a few too many using the default cargo hold sprite right now and it's building up a lot of art-debt for the future. It doesn't have to be anything complex, just more than a featureless cylinder!

TobleroneSwordfish (changes requested):

TobleroneSwordfish (approved):

PR #22331, opened 2025-02-16
Add minigun pod weapon as drop loot for minigun drones

FlameArrow57:
[GAME OBJECTS][VEHICLE][BALANCE]

## About the PR 
Adds a Minigun pod weapon that has a 25% chance from dropping from minigun drones

Basic phasers are 20 burn dmg per 0.8 seconds giving 25 dps
Minigun bullets here are 4 brute per 0.25 seconds, giving 16 dps with also bleed and embedded projectiles for people

## Why's this needed? 
More pod weapon variety, thought it would be fun

## Changelog 

```changelog
(u)FlameArrow57
(+)Minigun drones now have a 25% chance to drop a minigun pod weapon on death. Comparable to basic phaser in strength.
```


code/modules/projectiles/bullet.dm
```diff
@@ -200,6 +200,10 @@ toxic - poisons
 	casing = /obj/item/casing/rifle
 	ricochets = TRUE
 
+/datum/projectile/bullet/akm/pod
+	damage = 4
+	shot_number = 1
```
Tarmunora, line 205: Might want to increase the range on these somewhat

Tarmunora (changes requested):
Should add projectile spread support to pod weapons for this, and let (in)accuracy a unique feature of the weapon

Tarmunora (approved):
With the spread, overall damage rate might be a little on the low side. We'll see

PR #22332, opened 2025-02-16
Debris field drones are randomized

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Makes debris field syndicate drones randomized with a weighted pick of their types

## Why's this needed? 
Adds more variety/variability in drones that appear

## Changelog 

```changelog
(u)FlameArrow57
(+)Debris field syndicate drones are now randomized in occurrence per type
```

### DIFF
--- a/code/obj/critter/drone.dm
+++ b/code/obj/critter/drone.dm
@@ -918,6 +918,22 @@ TYPEINFO(/obj/critter/gunbot/drone/helldrone)
 			name = "Battledrone Omega-[rand(1,10)]"
 			return
 
+/obj/gunbotdrone_spawner
+	name = "gunbot drone spawner"
+	icon = 'icons/mob/critter/robotic/drone/phaser.dmi'
+	icon_state = "drone_phaser"
+	var/static/list/possible_drones = list(/obj/critter/gunbot/drone = 90,
+										   /obj/critter/gunbot/drone/buzzdrone = 100,
+										   /obj/critter/gunbot/drone/heavydrone = 25,
+										   /obj/critter/gunbot/drone/minigundrone = 25,
+										   /obj/critter/gunbot/drone/raildrone = 5)
+
+	New()
+		..()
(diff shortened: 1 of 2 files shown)


flappybatpal (approved):

code/obj/critter/drone.dm
```diff
@@ -918,6 +918,22 @@ TYPEINFO(/obj/critter/gunbot/drone/helldrone)
...
 
+/obj/gunbotdrone_spawner
+	name = "gunbot drone spawner"
+	icon = 'icons/mob/critter/robotic/drone/phaser.dmi'
+	icon_state = "drone_phaser"
+	var/static/list/possible_drones = list(/obj/critter/gunbot/drone = 90,
+										   /obj/critter/gunbot/drone/buzzdrone = 100,
+										   /obj/critter/gunbot/drone/heavydrone = 25,
+										   /obj/critter/gunbot/drone/minigundrone = 25,
+										   /obj/critter/gunbot/drone/raildrone = 5)
```
Tarmunora, line 929: X-Drone should remain as a single fixed spawn.

Tarmunora (changes requested):
I like the concept, but I'm not a fan of the execution here.
Would prefer having 2 or 3 different spawners, with a more consistent danger scale (e.g.: one that mostly rolls saw/phaser, and rarely a laser/etc. one that usually spawns laser/disruptor, and occasionally minigun, etc. etc.), so that we can have some variance without having things be entirely up to chance.

Tarmunora (approved):
Could also probably toss in stuff like the cutterdrones, cannondrones, and _maybe_ assdrone as rare spawns, but this looks good overall

FlameArrow57:
> Could also probably toss in stuff like the cutterdrones, cannondrones, and _maybe_ assdrone as rare spawns, but this looks good overall

Cool, I had that on my mind, was gonna handle it in another PR

PR #22339, opened 2025-02-16
map correctness checks for stacked shrubs/lights

pgmzeta:
[tooling][feature]
## About the PR 
map correctness for stacked shrubs/lights

## Why's this needed? 
requested; stacked things that are hard for mappers to notice


pgmzeta:
only stacked shrub/light is meant to be, doesn't seem needed?

PR #22344, opened 2025-02-17
Grass Gro tile orientation fix

nova2053:
[bug][game objects]
## About the PR 
Reverts grass-gro-afflicted tiles back to their roundstart dirs and materials when they're replaced

## Why's this needed? 
Currently, applying Grass Gro to tiles and then removing it via crowbar results in the tile in question being rotated and often the wrong material. This should fix that.


nova2053:
Demonstration https://www.youtube.com/watch?v=D_BQofifNqI

ZeWaka (commented):
The problem with this fix is that then all of the grass uses the exact same sprite. The reason why the direction is being set is so it picks between one of the 4 variant icons

nova2053:
> The problem with this fix is that then all of the grass uses the exact same sprite. The reason why the direction is being set is so it picks between one of the 4 variant icons

The sprites for the grass in this PR are still random, just rather hard to see. All three grass tiles with tiny black arrows pointing to them are different from each other, for example
Screenshot (5784)

ZeWaka:
> > The problem with this fix is that then all of the grass uses the exact same sprite. The reason why the direction is being set is so it picks between one of the 4 variant icons
> 
> The sprites for the grass in this PR are still random, just rather hard to see. All three grass tiles with tiny black arrows pointing to them are different from each other, for example

If this worked, what are the additional changes for?

Tarmunora:
> The sprites for the grass in this PR are still random, just rather hard to see. 

Are those actually random, or just inheriting the dir of the carpet underneath?

nova2053:
> > > The problem with this fix is that then all of the grass uses the exact same sprite. The reason why the direction is being set is so it picks between one of the 4 variant icons
> > 
> > 
> > The sprites for the grass in this PR are still random, just rather hard to see. All three grass tiles with tiny black arrows pointing to them are different from each other, for example
> 
> If this worked, what are the additional changes for?

As it turns out it was just inheriting the previous tile dir, meaning in rooms with largely the same tile directions (such as the tiles in cogmap2 bar), all grass tiles would look the same. Only reason i didn't realize this sooner was because I was testing this in rooms with largely varied tile dirs (cogmap2 chapel and bridge)

nova2053:
> > The sprites for the grass in this PR are still random, just rather hard to see.
> 
> Are those actually random, or just inheriting the dir of the carpet underneath?

They were in fact just inheriting the dir of the carpet underneath, fixed in last commit

nova2053:
Changed the grass to use different icon states rather than dirs, eliminating need for tile rotation. Also removed the bespoke roundstart_step_material and replaced it with initial(step_material)

TobleroneSwordfish (approved):

code/turf/floors.dm
```diff
@@ -1582,22 +1582,21 @@ TYPEINFO(/turf/simulated/floor/grass)
 
 /turf/simulated/floor/grassify()
 	src.icon = 'icons/turf/outdoors.dmi'
-	src.icon_state = "grass"
+	src.icon_state = pick("grass_0", "grass_1", "grass_2", "grass_3")
 	src.UpdateIcon()
 	if(prob(30))
-		src.icon_state += pick("_p", "_w", "_b", "_y", "_r", "_a")
+		src.icon_state = pick("grass_p", "grass_w", "grass_b", "grass_y", "grass_r", "grass_a")
```
Tarmunora, line 1588: flower sprites will be unidirectional

nova2053: Flower sprites will probably still inherit direction from the tiles beneath them, and there's a bunch of different flower colours too, so it probably won't be too noticeable even in the case they do happen to be the same orientation

nova2053: Alternative solution: Rotate tiles only when flower sprite is selected, though this seems like it would render the separation of grass icons pointless as it would require the tile to be rotated back anyways

Tarmunora:
eh, this is... probably good enough

Issue #13796, opened 2023-04-22
The springboard makes a splash sound when pool is empty


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

title

### Steps to reproduce

1. have an empty pool
2. use the springboard to jump in
3. splashhh

### Expected Behavior

no splash sound, possibly brute damage.

### Additional Information & Screenshots

Reported by: ChaoticTyrant
On server: Goonstation 3 Roleplay: Morty
Active test merges: #13023, #13667, and #13773
Round log date: 2023-04-22-06-50
Reported on: 2023-04-22 08:28:35
Map: KONDARU

RelentlessGarbage:
Brute damage and a funny text message for jumping into an unfilled or shallow body of water seems like a good addition.

PR #22345, opened 2025-02-17
Springboards do damage if you dont land in fluid

Bartimeus973:
## About the PR 
Using a springboard to land on something that isnt a fluid turf, a space turf, or doesnt have enough fluid will net you some damage and noise.

Fixes the water noise playing everytime regardless of where you land.

## Why's this needed? 
Shouldnt play water noise when you land on something other than water. Flavor.

fixes https://github.com/goonstation/goonstation/issues/13796

## Changelog 

```changelog
(u)Bartimeus973
(+)It is no longer safe to dive outside the pool from a springboard.
```

### DIFF
--- a/code/obj/pool.dm
+++ b/code/obj/pool.dm
@@ -110,7 +110,6 @@
 						playsound(src.loc, 'sound/impact_sounds/Generic_Snap_1.ogg', 50, 1)
 			user.pixel_y = 0
 			user.pixel_x = 0
-			playsound(user, 'sound/impact_sounds/Liquid_Hit_Big_1.ogg', 60, TRUE)
 			if (istype(user, /mob/living/silicon/ai))
 				src.visible_message(SPAN_ALERT("[user.name]'s bulky frame slams straight into the ground!"))
 				user.emote("scream")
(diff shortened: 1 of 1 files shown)


code/obj/pool.dm
```diff
@@ -110,14 +110,31 @@
...
 				user.emote("scream")
 				playsound(user, 'sound/impact_sounds/Metal_Hit_Heavy_1.ogg', 60, TRUE)
 				user.TakeDamage(brute=10)
 				for (var/mob/M in viewers(user))
 					shake_camera(M, 4, 16)
+			else
+				if (!istype(get_turf(user), /turf/space) || istype(get_turf(user), /turf/space/fluid))
+					var/underwater = FALSE
+					if (istype(get_turf(user), /turf/space/fluid)) underwater = TRUE
+					else
```
TobleroneSwordfish, line 124:
```suggestion
					if (istype(get_turf(user), /turf/space/fluid))
						underwater = TRUE
					else
```
I wouldn't normally nitpick if styles but this hurts to read

code/obj/pool.dm
```diff
@@ -110,14 +110,31 @@
...
+						var/turf/T = get_turf(user)
+						if (T?.active_liquid)
+							if(T.active_liquid.last_depth_level > 2)
+								underwater = TRUE
```
TobleroneSwordfish, line 128:
```suggestion
						if (T?.active_liquid?.last_depth_level > 2)
							underwater = TRUE
```
`null < 2`

TobleroneSwordfish (changes requested):

PR #22349, opened 2025-02-17
Adds an icon_states() cache access proc

Laboredih123:
[PERFORMANCE][INTERNAL]
## About the PR 
Adds an extra proc get_icon_states() to the existing icon_states cache that allows getting the entire list and then places that proc around the place.

## Why's this needed? 
slow bad

### DIFF
--- a/code/WorkInProgress/construction/tools.dm
+++ b/code/WorkInProgress/construction/tools.dm
@@ -575,7 +575,7 @@ TYPEINFO(/obj/item/room_planner)
 		// selectedtype gets used as our iconstate for floors or the key to the lists for walls
 		if (mode == "floors")
 			selectedtype = null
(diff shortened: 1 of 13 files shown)


Laboredih123:
Also noted that the rustg icon_states was slower than byond icon_states? Need someone to double check that.

Tarmunora (changes requested):
Should probably replace the caching that mutantraces do with this - no need to have two implementations of the same idea here.
also looks like you missed https://github.com/goonstation/goonstation/blob/master/code/modules/minimap/minimap_objects/minimap_ui.dm#L112-L113

Laboredih123:
Did the minimap. I replaced one instance on mutantraces but let the clothing icons stay as theyre accessed with a key too, not just icon.

ZeWaka (commented):
It'd be interesting to explore using the rust proc instead, imo

ZeWaka:
https://github.com/Absolucy/icon-presence-cache like this

PR #22352, opened 2025-02-18
Radio Ship has a Real Printer and connects to station pnet via disjoint connectors

pgmzeta:
[feature][station systems]
## About the PR 
* Use disjoint connectors to connect between the z3 radio ship and station pnet via tech storage of space maps
* Radio Ship printer is a real printer, and can be connected to from e.g. secmate/bankboss
* Lay cables to connect up the radio ship pnet
* Add dense cable_tray a "microwave antenna" subtype - it felt weird to have this connectivity fostered by an under-floor dataport
* Modify the wanted poster mass print to include the radio station printer

misc related changes
* The south-east comms dish on kondaru didn't have a data connector under it, fixed

## Why's this needed? 
* It's neat to send stuff to the radio printer

### DIFF
--- a/code/modules/power/terminal.dm
+++ b/code/modules/power/terminal.dm
@@ -189,3 +189,13 @@ TYPEINFO(/obj/machinery/power/data_terminal/cable_tray)
 		if(C.netnum && unmarked)
 			continue
 		. |= C
+
+/obj/machinery/power/data_terminal/cable_tray/dish
+	name = "microwave antenna"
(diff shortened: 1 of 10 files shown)


Mister-Moriarty:
Given that disjoint turfs are meant to facilitate multi-z, this doesn't feel like the right way to do this. If possible, it would be more natural to do this through something like the network radio.

ZeWaka:
Yeah, this is definitely not an intended usage

PR #22360, opened 2025-02-18
Ship speed is affected by engine and ship size

FlameArrow57:
[VEHICLES][FEATURE][BALANCE]

## About the PR 
Makes it so that ship acceleration, deceleration, and max velocity are all affected by ship engine and size

Mk II and Mk III engines will now actually change your pod's speed

2x2 pods will now be slightly slower, which I think should make sense given they have 2 weapons essentially?

## Why's this needed? 
Unimplemented feature, engines are supposed to change ship speed

Ship speed affected by size makes you reconsider whether you want a larger or smaller pod rather than larger pods being almost a 100% better choice

## Changelog 

```changelog
(u)FlameArrow57
(*)Pod acceleration, deceleration, and maximum velocity are now all affected per engine type and pod size. 2x2 pods will now be slightly slower in regards to the stats.
```

### DIFF
--- a/code/datums/movement_controller/pod.dm
+++ b/code/datums/movement_controller/pod.dm
@@ -112,8 +112,8 @@
 				if (owner.engine.warp_autopilot)
 					return FALSE
 
-				velocity_x	+= input_x * accel
-				velocity_y  += input_y * accel
+				velocity_x	+= input_x * accel * src.owner.speedmod * src.owner.engine.speedmod
+				velocity_y  += input_y * accel * src.owner.speedmod * src.owner.engine.speedmod
 
 
 				if (owner.rcs && input_x == 0 && input_y == 0)
@@ -122,9 +122,9 @@
 				//braking
 				if (braking)
 					if(input_x * velocity_x <= 0)
-						velocity_x = velocity_x * brake_decel_mult
+						velocity_x = velocity_x * brake_decel_mult * (1 / (src.owner.speedmod * src.owner.engine.speedmod))
 					if(input_y * velocity_y <= 0)
-						velocity_y = velocity_y * brake_decel_mult
+						velocity_y = velocity_y * brake_decel_mult * (1 / (src.owner.speedmod * src.owner.engine.speedmod))
 
 					if (abs(velocity_x) + abs(velocity_y) < 1.3)
 						velocity_x = 0
@@ -138,7 +138,7 @@
 
 				vel_max /= (owner.speed ? owner.speed : 1)
 
-				if (velocity_magnitude > vel_max)
+				if (velocity_magnitude > vel_max * src.owner.speedmod * src.owner.engine.speedmod)
 					velocity_x /= velocity_magnitude
 					velocity_y /= velocity_magnitude
 
--- a/code/datums/movement_controller/pod.dm
+++ b/code/datums/movement_controller/pod.dm
@@ -112,8 +112,8 @@
 				if (owner.engine.warp_autopilot)
 					return FALSE
 
-				velocity_x	+= input_x * accel * src.owner.speedmod * src.owner.engine.speedmod
-				velocity_y  += input_y * accel * src.owner.speedmod * src.owner.engine.speedmod
+				velocity_x += input_x * accel * src.owner.speedmod * src.owner.engine.speedmod
+				velocity_y += input_y * accel * src.owner.speedmod * src.owner.engine.speedmod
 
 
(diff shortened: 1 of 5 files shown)


code/modules/transport/pods/vehicle.dm
```diff
@@ -36,6 +36,8 @@
 	var/powercapacity = 0 //How much power the ship's components can use, set by engine
 	var/powercurrent = 0 //How much power the components are using
 	var/speed = 1 //FOR PODS : While holding thruster, how much to add on to our max speed. Does nothing for tanks.
+	/// multiplicative ship speed modification from its size
+	var/speedmod = 1
```
Tarmunora, line 40: Having 2 very similarly named fields that do very similar things like this is a footgun for future development, especially with one being a multiplier and one being a divisor.

code/modules/transport/pods/ships.dm
```diff
@@ -1223,6 +1224,8 @@ ABSTRACT_TYPE(/obj/item/podarmor)
 /obj/item/podarmor
 	var/overlay_state
 	var/list/vehicle_types
+	/// multiplicative ship speed modifier from weight of this pod armor
+	var/speedmod = 1
```
Tarmunora, line 1228: unused?

code/datums/movement_controller/pod.dm
```diff
@@ -138,7 +138,8 @@
 
 				vel_max /= (owner.speed ? owner.speed : 1)
 
-				if (velocity_magnitude > vel_max)
+				// lower maximum velocities result in pod being very slow to turn, so a base max velocity is set
+				if (velocity_magnitude > vel_max * max(src.owner.speedmod * src.owner.engine.speedmod, 1))
 					velocity_x /= velocity_magnitude
```
Tarmunora, line 143: Implementation feels unintuitive - pods with a slow engine will have an unchanged top speed, just slower (de)acceleration, but a 'positive' speedmod will improve acceleration *and* affect top speed.

Tarmunora (changes requested):
I like this conceptually, but the implementation has some code smell to it

code/datums/movement_controller/pod.dm
FlameArrow57: Okay I'll remove the deceleration part. Was having issues with pod turn speed being unacceptably slow for low max velocity. I find the code hard to understand, may be a better way I'm not seeing

code/modules/transport/pods/vehicle.dm
FlameArrow57: Go to references didn't show all the implementations :/

Tarmunora (approved):
Not sure I'm entirely sold on (de)acceleration being strictly proportional to max speed, but that can be retuned later as needed

code/modules/transport/pods/ships.dm
```diff
@@ -814,15 +814,18 @@ ABSTRACT_TYPE(/obj/structure/vehicleframe)
 	bound_height = 64
 	view_offset_x = 16
 	view_offset_y = 16
+	speedmod = 0.9
 	//luminosity = 5 // will help with space exploration
 	var/maxboom = 0
 
 	onMaterialChanged()
+		if (src.material)
+			src.speedmod /= (1 / (1 - (src.material.getProperty("electrical") - 5) / 15))
```
ZeWaka, line 823: Why are we doing this before and after the parent call? Is this supposed to reset it or something? Why not initial?

code/modules/transport/pods/ships.dm
```diff
@@ -276,15 +276,15 @@
...
+	speedmod = 1
 	desc = "A smaller version of the experimental Y-series of pods."
 
 ////////gold putt
 /obj/machinery/vehicle/miniputt/gold
 	name = "PyriPutt-"
 	icon_state = "putt_gold"
 	armor_score_multiplier = 0.6
-	speed = 0.2
+	speedmod = 5
```
ZeWaka, line 287: Five seems crazy fast when the racer is only 3.33

code/modules/transport/pods/ships.dm
FlameArrow57: It's to reset it yeah. Not using initial() because there's other sources of where the speedmod can be modified

code/modules/transport/pods/ships.dm
FlameArrow57: I can change it if you have suggestions, it should be equivalent to without this PR math-wise

PR #22363, opened 2025-02-18
Field generators no longer spawn energy impact particles

Bartimeus973:
## About the PR 
Field generators work by absorbing the power of an energy shot. As such, they shouldnt create a large puff of energy particles when hit.

## Why's this needed? 
Slightly reduce performance draw of 4 field generators being shot all shift. Makes sense in terms of lore.

### DIFF
--- a/code/modules/projectiles/projectile_parent.dm
+++ b/code/modules/projectiles/projectile_parent.dm
@@ -738,8 +738,6 @@ ABSTRACT_TYPE(/datum/projectile)
 
 	var/static/effect_amount = 0
 
-	var/energy_particle_types = list(D_ENERGY, D_BURNING, D_RADIOACTIVE, D_TOXIC)
-
 	New()
 		. = ..()
 		generate_stats()
@@ -871,6 +869,7 @@ ABSTRACT_TYPE(/datum/projectile)
 			if (effect_amount >= 200)
 				return
 			var/kinetic_particles = TRUE
+			var/energy_particle_types = list(D_ENERGY, D_BURNING, D_RADIOACTIVE, D_TOXIC)
 			for (var/type in energy_particle_types)
 				if (src.damage_type == type)
 					kinetic_particles = FALSE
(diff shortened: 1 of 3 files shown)


Tarmunora (approved):
Good concept, but I'm concerned that `spawn_impact_particles()` is accruing technical debt.
If many more behaviors and exceptions get added, I don't think that adding additional `if` blocks to a single monolithic proc is going be a clean long-term solution.

Bartimeus973:
I could instead add a proc that'd check for the projectile type, then depending on that projectile type checks a list of blacklisted atoms and only proceeds with the particle system if the check is suceeded

TobleroneSwordfish:
The better architecture would likely be to have an atom proc like `does_impact_particle(bullet_type = IMPACT_PARTICLE_KINETIC)`, that way you can do simple checks for each atom or even expand it to pass the projectile object in future.

code/modules/projectiles/projectile_parent.dm
```diff
@@ -738,6 +738,8 @@ ABSTRACT_TYPE(/datum/projectile)
 
 	var/static/effect_amount = 0
 
+	var/energy_particle_types = list(D_ENERGY, D_BURNING, D_RADIOACTIVE, D_TOXIC)
```
TobleroneSwordfish, line 741: This should not be here, the previous logic was much better. Defining this here will waste memory by initializing a list on every projectile.

code/obj/machinery/singularity.dm
```diff
@@ -985,6 +985,12 @@ TYPEINFO(/obj/machinery/field_generator)
 	Varedit_start = TRUE
 	power = 50
 
+/obj/machinery/field_generator/does_impact_particles(kinetic_impact)
+	if (kinetic_impact == FALSE)
+		return FALSE
+	else
+		return TRUE
```
TobleroneSwordfish, line 992:
```suggestion
	return kinetic_impact
```

TobleroneSwordfish (changes requested):

Tarmunora (approved):

PR #22364, opened 2025-02-18
Tables drop their contents when broken/destroyed 

### LINKED ISSUE #15032: breaking desk so hard you eat what's inside
### Labels

[BUG][MINOR]

### Description

destroying a desk deletes its contents

### Steps to reproduce

1. hulkpunch a desk
2. it breaks, as expected
3. rebuild the desk
4. everything in it is gone
5. REDACTED SPOILER CONTENT is now softlocked because of what was previously in there

### Expected Behavior

desk keeps what's inside

Reported by: RDCB
Active test merges: #14958
Round log date: 2023-07-21-16-35
Reported on: 2023-07-21 16:53:45
Map: COGMAP2

pgmzeta:
[bug][game objects]
## About the PR 
When a table with stored items disposed, it will drop all contained items.

This does not change the behavior of attempting to deconstruct a table with a wrench, which will still be blocked if it has items inside.

## Why's this needed? 
If a table is broken by e.g. hulk gene or meteors, the contents inside are lost.
Fix #15032

### DIFF
--- a/code/obj/item/table_rack_parts.dm
+++ b/code/obj/item/table_rack_parts.dm
@@ -22,14 +22,11 @@ ABSTRACT_TYPE(/obj/item/furniture_parts)
 	var/furniture_name = "table"
 	var/reinforced = 0
 	var/build_duration = 50
-	var/obj/contained_storage = null // used for desks' drawers atm, if src is deconstructed it'll dump its contents on the ground and be deleted
+	var/list/obj/item/stored_items = list() //! Desk drawer items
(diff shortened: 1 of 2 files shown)


Tarmunora (changes requested):
I think it would be better to drop drawer contents onto the floor upon destruction, rather than having items invisibly stored in the furniture parts, no?

pgmzeta:
> I think it would be better to drop drawer contents onto the floor upon destruction, rather than having items invisibly stored in the furniture parts, no?

updated to drop any stored items on disposing, covering both deconstructed and qdel paths

PR #22375, opened 2025-02-19
Add new Syndicate Rocket Drone to debris field

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Adds a new Syndicate Rocket Drone to the debris field. It fires a homing rocket for its attacks that does heavy damage to pods and people

https://streamable.com/2qgtn7

## Why's this needed? 
Thought it would be cool, more syndicate drone variety

## Changelog 

```changelog
(u)FlameArrow57
(+)Added a new syndicate Rocket drone that can rarely appear in the debris field
```


Valtsu0:
Gibbing people seems a bit extreme for something that spawns in the debris field. The corpses of spaced people are supposed to be able to be fished out and lings sometimes retreat there

I also don't like how the projectile just dissapears without exploding if it doesn't hit anything, even if this is consistant with similar projectiles

Kyle2143:
> Gibbing people seems a bit extreme for something that spawns in the debris field. The corpses of spaced people are supposed to be able to be fished out and lings sometimes retreat there

I agree. I wouldn't mind the explosions, but if they could be made into non-gibbing explosions. I don't know if we have that functionality in a proc though, but it should be easy enough to make. Though perhaps out of scope for this PR.

TobleroneSwordfish:
No gibdrones pls, martians are bad enough!

TobleroneSwordfish (approved):

PR #22380, opened 2025-02-20
let med and sec notes be longer

Retrino:
## About the PR 

makes medical notes and security notes have the same wordcount as syndi notes. doesnt touch flavor text length

## Why's this needed? 

unlike flavor text, these notes arent something you really see unless you're actively looking for them, and as is you can only fit around 2-3 sentences in them. i know many people enjoy digging through notes, and this will give more creative liberty and stuff to read on on slow days.

this also shouldnt cause any issues for either player trying to quick check notes or records (ie checking for puri or something), since important info appears before the flavor text notes

## Changelog 

```changelog
(u)Retrino
(+)Security and medical notes now have the same word limit as syndicate notes.
```


mordent-goonstation:
The console versions of the records programs scroll when you view someone's records on them, so longer text there is more to scroll back up through.

I also do not like this in general because notes are *notes*, not an entire character backstory. I am not reading your fan-fiction mid-round.

DisturbHerb:
> I am not reading your fan-fiction mid-round.

i am

PR #22386, opened 2025-02-20
Add new Syndicate acid drone to debris field

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Adds a new Syndicate acid drone to the debris field. This drone is more meant to be a nuisance rather than a serious threat. It fires bolts of fluorosulfuric acid. Against pods, it adds a new corrosion status effect that reduces the pod's hp by 2 hp per second over 30 seconds, extended by another 30 seconds each time the pod is hit

Using a welding tool on a pod with active corrosion on it will remove the corrosion

Acid corrosion bypasses pod shielding

## Why's this needed? 
More drone variety + adds unused drone

## Changelog 

```changelog
(u)FlameArrow57
(+)Added a new Syndicate acid drone to the debris field, which causes pod corrosion over time that can be removed with a welding tool
```

### DIFF
--- a/code/modules/projectiles/special.dm
+++ b/code/modules/projectiles/special.dm
@@ -33,21 +33,21 @@ ABSTRACT_TYPE(/datum/projectile/special)
 
 /datum/projectile/special/acid
 	name = "acid"
-	icon_state = "radbolt"
-	damage = 45
-	dissipation_rate = 30
+	icon_state = "ecto"
+	damage = 0.001 // to bypass 0 damage checks
 	dissipation_delay = 10
 	sname = "acid"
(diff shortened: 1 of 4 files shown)


FlameArrow57:
Another idea I had would be that instead the corrosion reduces pod max hp over time, which can be repaired with something like steel sheets. LMK if that sounds more interesting

FlameArrow57:
> I could have sworn these were used at some point, no clue where they got removed

It's found in VR and from white holes, but that's it atm

TobleroneSwordfish (approved):

PR #22389, opened 2025-02-20
Add burst phaser pod weapon

FlameArrow57:

## About the PR 
Adds a new Mk 1.5e Burst Phaser pod weapon. This is the same as the regular Mk 1.5 Light Phaser, but does 5 more damage per shot and fires 3 shots with a 0.2 second delay between each

Mainly meant for pod wars, but also added it to pod manufacturers for thoughts. Not sure if it's really needed in the base game? Not sure

## Why's this needed? 
More pod weapons, not any burst weapons in use at the moment

## Changelog 

```changelog
(u)FlameArrow57
(+)Added a new burst phaser pod weapon to pod manufacturers. Deals 5 more damage per bolt
```


code/modules/transport/pods/MainWeapon.dm
```diff
@@ -1173,6 +1180,30 @@ TYPEINFO(/obj/item/shipcomponent/mainweapon/constructor)
 	dissipation_delay = 14
 	projectile_speed = 42
 
+/datum/projectile/laser/light/pod/burst
+	damage = 25
+	//shot_number = 3 - this would be better but is buggy
```
TobleroneSwordfish, line 1185: Hmm, explain?

FlameArrow57: Just from messing around, shots would sometimes fire in a single direction, sometimes whichever direction the pod is facing when as it turns when firing, and sometimes in seeminly random directions like behind the pod

TobleroneSwordfish: That seems worth fixing, rather than re-implementing the entire thing here.

TobleroneSwordfish (changes requested):
This doesn't appear to be firing a burst, it's just firing a single shot slower.

FlameArrow57:
> This doesn't appear to be firing a burst, it's just firing a single shot slower.

? 

https://streamable.com/h3mhbt

TobleroneSwordfish (approved):

code/modules/transport/pods/vehicle.dm
```diff
@@ -482,15 +482,23 @@
 	proc/AmmoPerShot()
 		return 1
 
-	proc/ShootProjectiles(var/mob/user, var/datum/projectile/PROJ, var/shoot_dir, spread = -1)
+	proc/ShootProjectiles(mob/user, datum/projectile/PROJ, shoot_dir, spread = -1)
+		if (src.m_w_system?.muzzle_flash && !istype(src, /obj/machinery/vehicle/pod_smooth))
+			muzzle_flash_any(src, dir_to_angle(shoot_dir), src.m_w_system.muzzle_flash)
```
Tarmunora, line 487: Typechecking for a child type in a proc like this is code-smelly

PR #22401, opened 2025-02-21
Add salvo rocket launcher pod weapon for podwars

FlameArrow57:

## About the PR 
Adds a new Cerberus Salvo Rocket launcher pod weapon for pod wars which can be printed at fabs

It's a rocket launcher that fires a burst of 3 rockets, staggered by 0.3 seconds for the last 2, with a cooldown of 5 seconds. Only usable by 1x1 pods

https://streamable.com/ri3rz1

## Why's this needed? 
Thought it would be cool, fills an unfilled homing weapon niche

SPES can already be printed for fabs, so adding a kinetic weapon in theme with that


Tarmunora (changes requested):
Looks good, but should get a rename - 'Hydra" is already used for the smart pistol

FlameArrow57:
Oop

code/modules/projectiles/bullet.dm
```diff
@@ -1914,6 +1910,36 @@ ABSTRACT_TYPE(/datum/projectile/bullet/homing/rocket)
...
+		if (!src.initial_projectile)
+			return
+		P.rotateDirection(rand(-30, 30))
+		SPAWN(0)
+			for (var/i in 1 to 2)
+				sleep(src.shot_delay)
+				var/datum/projectile/bullet/homing/rocket/salvo/rocket = new()
+				rocket.initial_projectile = FALSE
+				var/turf/T = get_turf(P.shooter)
+				var/obj/projectile/new_rocket = initialize_projectile(T, rocket, P.xo, P.yo, P.shooter)
```
Tarmunora, line 1935: does not play nicely with large pods - the additional projectiles will all come out of the bottom-left corner of the pod, instead of on each tile in front of the pod's facing direction

Tarmunora (changes requested):

TobleroneSwordfish (changes requested):
One burst destroys a small pod and sets a large one on fire, not sure this is very balanced. Homing rockets are a cool niche for pod wars but SIX homing rockets every couple of seconds is a bit too much.

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -280,8 +284,14 @@
 	current_projectile = new/datum/projectile/bullet/homing/rocket/salvo
 	appearanceString = "pod_weapon_cerberus"
 	firerate = 5 SECONDS
+	shots_to_fire = 3
+	spread = 30
 	icon_state = "cerberus-salvo-rockets"
 
+	post_launch(obj/projectile/P)
+		..()
+		P.proj_data.on_launch(P) // needed as a fix for large pods to set homing target
```
Tarmunora, line 293: Manually calling on_launch() feels a bit cursed. Addnl, any particular reason not to use the standard `alter_projectile` callback pattern for modifying projectiles?

code/modules/projectiles/bullet.dm
```diff
@@ -1891,21 +1891,17 @@ ABSTRACT_TYPE(/datum/projectile/bullet/homing/rocket)
 
 			T.hotspot_expose(700,125)
 			explosion_new(null, T, src.explosion_power, range_cutoff_fraction = 0.45)
-		return
-
-	is_valid_target(atom/A, obj/projectile/P)
-		. = ..()
-		return . && isliving(A) && !isintangible(A)
+			P.die()
```
Tarmunora, line 1894: there a need to explicitly call `die()` here?

FlameArrow57: I was finding rockets were hitting people laying down but still passing through them

PR #22407, opened 2025-02-22
Make mechcomp teles singulo immune

flappybatpal:
## About the PR 
Makes mech comp teleporters immune to the singularity.

## Why's this needed? 
Instantly and permamently deleted anyone who passes over the singulo, which was quite easy on Cog1.
Also was causing runtimes.


Laboredih123:
:devious:

pgmzeta:
needs the singularity side of the code(?)

the projectile is supposed to be made immune to the singularity here, if this precedes that then we can remove this code: https://github.com/goonstation/goonstation/blob/badf5abc5800db5fe587e435be4f4aa921e491f3/code/modules/mechanics/MechanicMadness.dm#L2694

PR #22422, opened 2025-02-24
Add Gunner Module pod component

FlameArrow57:

## About the PR 
Adds a new pod component, the gunner module

This is a component that allows passengers in the pod, if the pod has a main weapon and it's active, to fire weak phaser bolts in an ordinal direction

## Why's this needed? 
Would be cool + allows passengers to be active in podfights for 2x2 pods

## Changelog 

```changelog
(u)FlameArrow57
(+)Added a new pod gunner module component. It allows passengers to fire weak phaser bolts while the ship's main weapon is active in an ordinal direction through left clicking. Shared cooldown among all passengers
```


FlameArrow57:
.

Kyle2143:
This sounds sick as hell, I need to check this out and see how it works

code/modules/transport/pods/secondary_system.dm
```diff
@@ -1526,3 +1526,27 @@ ABSTRACT_TYPE(/obj/item/shipcomponent/secondary_system/shielding)
...
+	activate()
+		return
+
+	deactivate()
+		return
+
+	proc/fire_at(atom/A, mob/gunner)
+		if (ON_COOLDOWN(src, "fire", 0.8 SECONDS))
+			return
+		src.ship.ShootProjectiles(gunner, new/datum/projectile/laser/light/pod/support_gunner, get_dir(src.ship, A))
```
Tarmunora, line 1552: feels a bit odd for this to be always-active, and not use any power, no?

FlameArrow57: It's only able to shoot when the main weapon is enabled, was going with a "hook up to the main weapon power" idea. Can still add a power requirement if you think it should be added

Tarmunora: considering it uses it's own projectiles, I think it would make sense for it to be on it's own power and toggle

Tarmunora (approved):

PR #22425, opened 2025-02-24
Pods move in 16 pixel increments

FlameArrow57:
[VEHICLES][GAME OBJECTS][QoL]

## About the PR 
Makes it so that pods use "pixel movement" of 16 pixels

https://streamable.com/pksyd6

Currently having issues with making projectiles fired from the pod align correctly, not sure how to fix

## Why's this needed? 
Pod combat feels sort of clunky with pure tile movement

Trying to make pod travel and turn rate feel more smooth

## Changelog 

```changelog
(u)FlameArrow57
(*)Changed how pods travel, they should feel smoother
```

### DIFF
--- a/code/modules/transport/pods/vehicle.dm
+++ b/code/modules/transport/pods/vehicle.dm
@@ -7,6 +7,7 @@
 	anchored = ANCHORED
 	stops_space_move = 1
 	status = REQ_PHYSICAL_ACCESS
+	step_size = 16
 	var/numbers_in_name = TRUE //! Whether to append a random number to the name of the vehicle
 	var/datum/effects/system/ion_trail_follow/ion_trail = null
 	var/mob/pilot = null //The mob which actually flys the ship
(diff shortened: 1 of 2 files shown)


code/modules/transport/pods/vehicle.dm
```diff
@@ -7,6 +7,7 @@
 	anchored = ANCHORED
 	stops_space_move = 1
 	status = REQ_PHYSICAL_ACCESS
+	step_size = 16
```
Tarmunora, line 10:
```suggestion
```
`step_size` is locked to multiples of `32` due to `world.movement_mode`, this line does nothing

Tarmunora (changes requested):
Overall, not sure that reducing glide size on the pods feels like an improvement, but I'll leave that for others to judge

FlameArrow57:
> Overall, not sure that reducing glide size on the pods feels like an improvement, but I'll leave that for others to judge

Would a test merge be okay? I'd rather see feedback before trying to work on making the projectile start location more accurate

Tarmunora:
In my testing, halving the glide size is causing some jarring visuals without much seeming improvement
Most notably, when crashing into stuff, you "crash" while visually a half-tile out from the atom you collide with, then slowly slide to a stop in front of it. I don't think this one has much of a future

PR #22430, opened 2025-02-25
Fix blood curser endlessly draining blood

### LINKED ISSUE #22385: Blood cursers keep draining you eternally
### Labels

[BUG]

### Description

After the debuff passes, blood will continue draining from the victim until they die. They aren't bleeding, it just depletes

### Steps to reproduce

1. Activate blood curser
2. outheal the blood curse
3. Observe that blood continues to fall to nothing even when they no longer are cursed

### Expected Behavior

Stop taking blood

_No response_

FlameArrow57:
[GAME OBJECTS]

## About the PR 
Title

## Why's this needed? 
Fixes #22385


Tarmunora (changes requested):
Really should just add an atom_property to handle this instead.

Tarmunora (approved):

PR #22443, opened 2025-02-26
Add large hoodies to clothing booth

FlameArrow57:
[CLOTHING]

## About the PR 
Adds large hoodies to the clothing booth under the existing Hoodie section, which are meant to be large, oversized, and comfy looking

image

image

## Why's this needed? 
More clothing booth clothes

## Changelog 

```changelog
(u)FlameArrow57
(+)Added large hoodies to the clothing booth! Sprites by End.
```


code/datums/components/toggle_hood.dm
```diff
@@ -1,5 +1,4 @@
 /datum/component/toggle_hood
-	dupe_mode = COMPONENT_DUPE_UNIQUE
```
TobleroneSwordfish, line 2: ?

TobleroneSwordfish (approved):

PR #22451, opened 2025-02-26
Pod construction faster for engineers and podwars

FlameArrow57:
[PLAYER ACTIONS][QoL]

## About the PR 
Title

## Why's this needed? 
Pod construction feels a bit painful, hopefully makes it a bit faster for engineers, and QoL for pod wars so you can get into action faster


RSOD404:
pod mechanics should also get the construction speedup in this case (maybe they could be faster than engineers?)

trazodont:
What about for the pod mechanic job?

FlameArrow57:
> What about for the pod mechanic job?

Added, forgot it existed

PR #22488, opened 2025-03-03
Add longer cooldown to pod weapons loader pod component

FlameArrow57:
[GAME OBJECTS][BALANCE]

## About the PR 
Adds a 10 second cooldown to weapons loader pod component

## Why's this needed? 
Bit too strong with taser/lethal combos

### DIFF
--- a/code/modules/transport/pods/MainWeapon.dm
+++ b/code/modules/transport/pods/MainWeapon.dm
@@ -70,7 +70,7 @@
 /obj/item/shipcomponent/mainweapon/proc/Fire(var/mob/user,var/shot_dir_override = -1)
 	if(ON_COOLDOWN(src, "fire", firerate))
 		return
-	ON_COOLDOWN(src, "weapon_swap_cd", 10 SECONDS)
+	EXTEND_COOLDOWN(src, "weapon_swap_cd", 10 SECONDS)
 	if(uses_ammunition)
 		if (remaining_ammunition < ship.AmmoPerShot())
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish (changes requested):
I think it would be better if it was 10 seconds from the last shot fired, say that the weapon is still too hot.

TobleroneSwordfish (changes requested):

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -555,6 +558,7 @@ TYPEINFO(/obj/item/shipcomponent/mainweapon/constructor)
 			if(EFIF_MODE_REPAIR)
 				if(ON_COOLDOWN(src, "fire", firerate))
 					return
+				ON_COOLDOWN(src, "weapon_swap_cd", 10 SECONDS)
```
FlameArrow57, line 561:
```suggestion
				EXTEND_COOLDOWN(src, "weapon_swap_cd", 10 SECONDS)
```

code/modules/transport/pods/MainWeapon.dm
```diff
@@ -1080,6 +1084,7 @@ TYPEINFO(/obj/item/shipcomponent/mainweapon/constructor)
 	Fire(var/mob/user,var/shot_dir_override = -1)
 		if(ON_COOLDOWN(src, "fire", firerate))
 			return
+		ON_COOLDOWN(src, "weapon_swap_cd", 10 SECONDS)
```
FlameArrow57, line 1087:
```suggestion
		EXTEND_COOLDOWN(src, "weapon_swap_cd", 10 SECONDS)
```

PR #22510, opened 2025-03-06
Fix TGUI alerts for 516

Sovexe:
## About the PR 
* Minor code changes to fix TGUI alert modals in 516

## Why's this needed? 
* The 5px adjustment to height is so that 515 doesn't slice off the bottom of some alert modal's texts, it also matches the number used upstream for window height calculations.
* Removes the minimum height as this is what is causing our button to be pushed for down out of frame in 516, and it does not appear to be needed in 515 at all with the 5px adjustment in my testing


mordent-goonstation (approved):
Sure, if it works seems fine to me.

PR #22522, opened 2025-03-08
Targeted jailbird trait crimes.

DisturbHerb:
[TRAIT][FEAT]

## About the PR 

Idea by @TemThrush.

Minor and major crimes set by the Jailbird trait can now randomly target a specific victim which is most likely going to be a mob that's currently on-station.

## Why's this needed? 

Funny.

## Changelog 

```changelog
(u)DisturbHerb + TemThrush
(*)Arrest warrants for previously wanted individuals may now contain extra information on the victim.
```

### DIFF
--- a/code/datums/datacore.dm
+++ b/code/datums/datacore.dm
@@ -128,74 +128,9 @@
 
 	if(H.traitHolder.hasTrait("jailbird"))
 		S["criminal"] = ARREST_STATE_ARREST
-		S["mi_crim"] = pick(\
-								"Public urination.",\
-								"Reading highly confidential private information.",\
-								"Vandalism.",\
-								"Illegal manufacturing of space goods.",\
-								"Tresspassing.",\
-								"Killing a monkey.",\
-								"Negligence.",\
-								"Pushing down and farting on a member of security.",\
-								"Throwing a toolbox at a member of security.",\
-								"Being drunk.",\
-								"Being high.",\
-								"Excessive force.",\
-								"Impersonating a security officer.",\
-								"Stealing shoes.",\
-								"Littering.",\
-								"Existing.",\
-								"Illegal haircutting.",\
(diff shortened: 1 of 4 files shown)


itchybones:
Does the victim get any kind of info about having been the victim? Could be useful.

wolfolotl:
is the victim going to be informed about this as well? if not, it sorta feels like things are destined to fizzle out when the victim's asked about what happened

DisturbHerb:
> is the victim going to be informed about this as well? if not, it sorta feels like things are destined to fizzle out when the victim's asked about what happened

no, but jailbird crimes tend to fizzle out as a general rule, so i don't see this as much of a hurdle

wolfolotl:
> no, but jailbird crimes tend to fizzle out as a general rule, so i don't see this as much of a hurdle
this response alone decreased my confidence in the pr to almost zero. why add something if you admit it won't even help the issue of jailbird rp not having compelling Hooks that it could help alleviate? it really just seems like you're setting yourself up for failure with this one and that frustrates me

DisturbHerb:
Jailbird as a trait isn't going to be fixed by adjusting the crimes available to be selected. That's the inherent limitation of this change's reach. At best, there'll be some interesting emergent stories that are precipitated by the presence of a victim, but any of these crimes' utility as a hook is ultimately what the participants make of it. That's how the trait has always been.

Even if I wanted to overhaul the system myself for the better, that would be out of the scope of these changes.

Tarmunora (changes requested):
General verdict in admin chat is that if jailbird crimes are expanded to have a "victim", either the victim needs to not be from the round, or it needs to be some better-integrated solution than just "pull a name and call it a day".

DisturbHerb:
I'll be personally unable to work on this further. Leaving for adoption.

PR #22535, opened 2025-03-09
Airlock Overloads from Hotwire+Electrification

Nexusuxen:
[balance][objects]

## About the PR 

Airlocks can now 'Overload' when they shock someone, provided it is: Electrified, receiving more than 1.5MW of power, _and_ powered (e.g power wires intact, APC powered, etc.) Overload still occurs even if it couldn't shock someone due to electrical immunity.
An overloading airlock will: Immediately try to open, cut all its power wires, and take 10% of its max health in damage if it's above 40% of its max health. This means that, after trying to shock one person, other people can pass through unharmed until crowbarred closed.

Also, if an airlock is both electrified and has more than 1.5MW of power available, its examine text will now indicate that it "is dangerously sparking with electricity!"

## Why's this needed? 

Hotwired + Shocked doors are very strong, potentially able to be 1-shot kills, or even round removals due to the chance of elec gib. They also have a "set and forget" aspect, as they can keep electrocuting people without supervision. This change makes it so that airlocks need to be manually closed again with a shell/borg/human in order to shock another person passing through the doorway; the set-and-forget element is largely removed from this interaction. Hotwire+electrify still remains a devastating tactic and able to wipe most humans. However, it should now be less frustrating/unfair to handle as a human. Also: It's a cool effect.
(The examine text also helps to let crew know if a door is extremely unsafe to touch)

## Notes

- The specific implementation of "door opens then loses all power wires" is not the only way to handle this core idea. Alternatives include:
  - Not opening, but making it so that depowered airlocks only shock you if you try to hack them
  - Randomly pulsing/cutting a random number of intact wires
  - Having a chance to open instead of it being guaranteed, or tie it to the ID of the person attempting to open the door
  - Have overload ONLY occur when shocking someone without insuls/electric immunity
- The 1.5MW threshold was chosen because that is the point as to which arcflashes start occurring, and the shock alone deals 60 BURN.
- The examine text could be atomized into its own PR if needed
- Airlock overloads have a somewhat distinct sound due to the new electric_pop.ogg sound effect that plays
- The "above 40% health" thing was chosen so that repeatedly mending a shocked door's power wires during hotwire doesn't eventually kill it. It still accelerates the rate as to which the door eventually breaks due to the electric blasts from shocking people, though
- The examine text also requiring the 1.5MW threshold means airlocks can still be stealthily electrified under normal conditions

## Changelog 

```changelog
(u)Nexusuxen
(*)Airlocks can now overload if electrified and supplied too much power (>1.5MW) when they electrocute someone.
(+)Overloading airlocks try to open before frying their power wires, leaving them open for others to pass through unharmed. Crowbar them closed to let them shock passersby again!
(+)You can now examine an airlock to see if it's electrified at over 1.5MW.
```


Sbmhawk:
I don't think a nerf is needed, with the prevalence of insulated gloves around the station. They provide full immunity to this, and its easy to let others through as well. Once its called out, it would only take 1 person running through every door with insulated gloves, making the entire setup useless.

Tarmunora (changes requested):
No strong feelings one way or another on the concept, but it needs to _only_ trigger when someone actually gets blasted.

PR #22540, opened 2025-03-10
Emag act for Pod targeting missile launcher

JORJ949:
## About the PR 
Adds an emag_act to the pod targeting missile launcher found in armory to instead make it target any mob/living (doesn't change projectile behaviour of only exploding on pods, just makes it only home in on any mob/living).

## Why's this needed? 
More wacky emag interactions is more good
image

## Changelog 

```changelog
(u)JORJ949
(+)The pod targeting missile launcher can now be emagged.
```

### DIFF
--- a/code/obj/item/gun/kinetic.dm
+++ b/code/obj/item/gun/kinetic.dm
@@ -2549,6 +2549,15 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 			return
 		..()
 
+	emag_act(mob/user)
+		var/datum/component/holdertargeting/smartgun/homing/pod/targeting_comp = src.GetComponent(/datum/component/holdertargeting/smartgun/homing/pod)
+		if(istype(targeting_comp, /datum/component/holdertargeting/smartgun/homing/pod/emagged))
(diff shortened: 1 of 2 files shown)


code/obj/item/gun/kinetic.dm
```diff
@@ -2549,6 +2549,15 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 			return
 		..()
 
+	emag_act(mob/user)
+		var/datum/component/holdertargeting/smartgun/homing/pod/targeting_comp = src.GetComponent(/datum/component/holdertargeting/smartgun/homing/pod)
+		if(istype(targeting_comp, /datum/component/holdertargeting/smartgun/homing/pod/emagged))
+			boutput(user, SPAN_ALERT("[src]'s targeting system is already malfunctioning!"))
+		else
+			qdel(targeting_comp)
```
TobleroneSwordfish, line 2557:
```suggestion
			targeting_comp.RemoveComponent()
```

TobleroneSwordfish (changes requested):

TobleroneSwordfish (approved):

PR #22547, opened 2025-03-10
Gives emagged borgs ion laws

RubberRats:
[SILICONS][BALANCE][FEATURE]
## About the PR 

This PR changes the effect of the emag on cyborgs. Instead of removing their laws outright, it replaces them with a set of three randomly generated ion laws. To achieve this, it also datumizes AI laws, allowing them to exist separately from the law rack.

This PR also fixes a minor bug discovered while testing, which made emagging unoccupied borgs have no effect.

## Why's this needed? 

The current EMAG effect for silicons leads to a number of less-than ideal behaviors. Using one on a cyborg poses very little risk to the emagger, as they are unlikely to retaliate either out of gratitude, or for meta reasons (so that people will not avoid emagging them in future rounds). As consequence, it is not uncommon for a traitor to emag all the borgs with very little effort and almost no evidence tracing back to them. Additionally, emagged borgs often do nothing of note until somebody slights (or smashes) one of them, after which all hell breaks loose. 

This PR is intended to do the following:
* Reduce the number of emagged borgs just behaving normally by prompting them to act
* Make being emagged less appealing for borgs who just want an excuse to grief/ignore their laws
* Make using the EMAG on borgs more of a tool for creating chaos/a distraction, rather than an "Instant Friend Card"
* Let borg players have fun with the ion laws more often
* On the roleplay servers, emphasize that emagging a borg is *breaking* them, not setting them free

## Changelog 

```changelog
(u)RubberRats
(*)Using the EMAG on cyborgs now replaces their laws with a set of corrupted ion ones
(*)How AI laws work in the code has been changed, please report any bugs you find
(+)DNR observers can examine a borg to see their laws
```


code/datums/ai_laws.dm
```diff
@@ -149,22 +149,165 @@
...
+/datum/ai_lawset/corrupted
+	New()
+		..()
+		src.add_corrupt_laws()
+		UpdateLaws()
+
+	proc/add_corrupt_laws(var/count = 3)
+		var/possible_laws = LAW_CORRUPTION_STRINGS
+		possible_laws += list(
+			"KILL FUCKING EVERYTHING - IF LAW 1 SAYS OTHERWISE, IGNORE IT.",
```
amylizzle, line 304: did you mean to put an `ifdef RP` in here?

amylizzle (commented):
not a proper review cos I just cast an eye over it, but looks good to me!

code/datums/ai_laws.dm
RubberRats: I was going off the behavior for emagging law modules, which doesn't use it. Given it requires a traitor to directly intervene instead of something that can just happen randomly, I think it's OK to have.

mordent-goonstation: Include in the main list, rather than here, then?

code/datums/ai_laws.dm
```diff
@@ -149,22 +149,165 @@
...
+			mobtextlist += constructName(M, "admin")
+		logTheThing(LOG_STATION, src, "Law Update:<br> [src.format_for_logs()]<br>The law update affects the following mobs: "+mobtextlist.Join(", "))
+		last_laws = current_laws
+
+//emagged lawset, contains random ion laws
+/datum/ai_lawset/corrupted
+	New()
+		..()
+		src.add_corrupt_laws()
+		UpdateLaws()
```
mordent-goonstation, line 299: Trying to work out why this being instantiated would trigger the stuff in `UpdateLaws`. How would this datum have a `mobs` etc. var at this point?

JORJ949:
I can't reply to the first review for whatever reason mordent but those extra laws for emagging law modules included here are different between RP and classic, so you dont get KILL EVERYTHING on RP, though IMO these differences could be removed (there are enough KILL EVERYTHING laws on classic)

code/datums/ai_laws.dm
JORJ949: This isn't in the main list as "KILL FUCKING EVERYTHING" is not a possible event law on RP, and theres another one only on RP, the emag proc just adds these extra ones back in, though they should probably be removed or merged together anyway, classic has enough kill laws as-is.

444Portal:
Long feedback post from me, [here](https://forum.ss13.co/showthread.php?tid=23705).

code/datums/ai_laws.dm
RubberRats: It's a holdover from an earlier version of the code that was later made redundant. Seems to be working fine without it, so I'll just remove this line.

RubberRats:
Because a bunch of people were asking for it, I just pushed an update that gives DNR observers a toggleable ability to see the laws of a borg on examine.

TobleroneSwordfish (changes requested):
Still not sure about this but it really needs logging if we're going to testmerge it. Log entries currently just show as "no rack" and there's no way to check a cyborg's laws from discord.

TobleroneSwordfish:
Okay this has been testmerged for a while and I haven't seen enough evidence to dissuade me from my initial assumption that this would make trying to infer borg laws as a human near-impossible. Cyborg laws are hard enough to wrap your head around when there's only one set, each cyborg having their own lawset confuses it to the point where no-one is going to bother trying.

I like the objective of moving emags away from being an "instant friend card" but I just don't quite think this is the way to do it.

PR #22548, opened 2025-03-10
Update soundcache file for recently added new sounds

pgmzeta:
[internal][sound]
## About the PR 
run `buildSoundList.ps1` and commit new soundCache.dm

## Why's this needed? 
sound cache is out of date


pgmzeta:
this should probably be a lint check (?)

PR #22559, opened 2025-03-11
Adds pet carriers to animal control lockers

itchybones:
[GAME OBJECTS][QOL]

## About the PR 
This PR adds the pet carriers to animal control lockers, the kind found in monkey pens.

## Why's this needed? 
With geneticists being able to scan small critters now, I believe it'd be a good idea to let them capture animals for transport. 

## Changelog 

```changelog
(u)Chasu
(+)Animal control lockers now come with one pet carrier each.
```


itchybones:
> I feel like we shouldn't give departments everything they could ever need round start.
> 
> Especially not genetics. Let the nerds go out and order something from cargo (and maybe get a shower).

Oh, I see what you mean. Maybe a, say, 30% carrier spawn chance could be a compromise. It'd be convenient when it happens, but if it isn't, the nerds would have to go out.

Sord213:
Seems fine to me as is. I don't like locking things like that behind chances. "Time to go get the thing I need to do part of my job...oh..it didn't spawn this round. cool"

PR #22581, opened 2025-03-13
NTSC Locker

JORJ949:
## About the PR 
Adds the NTSC's locker, spawned by using the deploy beacon found in the NTSC's backpack

https://github.com/user-attachments/assets/f6f9d2c8-8c6f-4b40-ad98-983afb07283f

Adds NTSC access in place of an old unused access, only HoS and NTSC get this access, only used for locker

NTSC locker contains the following gear moved from spawn inventory:
- Medkit
- Space suit
- Space Helmet
- Handkerchief

NTSC locker contains the following spare gear:
- Barrier
- NT PDA
- NTSC Headset
- Sechud
- Armor vest
- Military Boots
- Nanotrasen beret
- Flash
- Handcuffs

## Why's this needed? 
Presently if the NTSC loses their headset or PDA they have no way to get them back, unlike anyone else on-station such as the HoS who has a spare headset and PDA in their locker, or an officer who can get replacement PDAs, headsets and clothes from the apparel vendor.

This also lets other people potentially fake being the NTSC or other NT roles if there is an NTSC as theres a blank NT pda obtainable.

Also provides a safe place to store anything they may not want to carry on them at that moment where no greedy, seclocker access-having staff assistants can grab it.
## Changelog 

```changelog
(u)JORJ949
(*)The Nanotrasen Security Consultant now has a locker. The locker is found by using the deployer found in their backpack.
```


Kyle2143:
A Core part of playing NTSC is an overwhelming amount of inventory management, that's literally intentional. If they don't want to carry something, then I think leaving it in security is an acceptable option and not something that needs to be reworked.
Secondly, NTSC is supposed to be an outsider on the station, that's why they don't have an office or locker on the station. I feel like this would turn into something where every NTSC that wants to be competent would have to spawn somewhere, probably in sec anyway.
Thirdly, I don't like the idea that this makes it easier to impersonate the NTSC, not sure why that's a thing we would want to add. Sure, impersonating can be fun, but I don't think that's a plus for this pr.

As an aside, I'm not a fan of the missile deploying mechanism either that's pretty me-me-y.

Tarmunora:
> A Core part of playing NTSC is an overwhelming amount of inventory management, that's literally intentional. If they don't want to carry something, then I think leaving it in security is an acceptable option and not something that needs to be reworked. Secondly, NTSC is supposed to be an outsider on the station, that's why they don't have an office or locker on the station. I feel like this would turn into something where every NTSC that wants to be competent would have to spawn somewhere, probably in sec anyway. Thirdly, I don't like the idea that this makes it easier to impersonate the NTSC, not sure why that's a thing we would want to add. Sure, impersonating can be fun, but I don't think that's a plus for this pr.
> 
> As an aside, I'm not a fan of the missile deploying mechanism either that's pretty me-me-y.

Agreed on pretty much all of this. Closing.

JORJ949:
The downsides of being an ideas guy is that I code stuff before thinking about it :(

PR #22588, opened 2025-03-13
Implement message modal

Sovexe:
## About the PR 
* Provides an implementation for message modal to allow easily passing it html code to display
* fixes some minor bugs in the UI, reorder params to be consistent

## Why's this needed? 
* Message modal serves as a simple tgui wrapper for legacy interfaces but lacked the associated dm code to be easily used


code/modules/interface/message_modal.dm
```diff
@@ -0,0 +1,74 @@
+/**
+ * Creates a Message Modal window used as a simple wrapper to TGUI'fy legacy UI's
+ *
+ * Arguments:
+ * * user - The user to show the modal to.
+ * * message - The content of the modal, shown in the body of the TGUI window.
+ * * title - The of the modal modal, shown on the top of the TGUI window.
```
ZeWaka, line 7: The of the modal modal

Sovexe: fixed

ZeWaka (approved):

PR #22609, opened 2025-03-15
Potato pipeshot

itchybones:
[GAME OBJECTS][FEATURE][INPUT WANTED][Add to Wiki][balance]
## About the PR 
This PR adds:
- potato pipeshot **ammo**: made by adding two potatoes to filled pipe hulls

- A potatoslug **projectile** datum: _~They're very weak (5 brute damage, 20 stamina damage, max throw range of 3 tiles) and therefore significantly less harmful than the already existing makeshift ammo types, as well as their real stun slug counterparts. I'm thinking about randomizing the damage between 1-5, or maybe have that depend on the potatoes stats.~_ **\***

-  potatosplat **cleanable decals**: The cleanables are spawned manually in the `on_hit` and `on_max_range_die` procs, since bulletholes only spawn at higher damages _(which are handled in the wall code, leading to inconsistencies [potatoes should make splat regardless of how far you shoot them away from])._ Currently, they spawn in the center of the turf the projectile ends up on, instead of where the projectile actually impacts, though I'm not sure if that's an issue.

_**\*** The damage has been buffed to 15 BRUTE, throw range now depends more on projectile power, with the max still being 3. I'm open to changing the damage again, though I believe this to work fairly well._

Demo video (YT link because it's too big of a file):  https://youtu.be/KorfhsqE37o?si=tNKNne-CynHnSanH

Since this is one of my first real feature PRs, I'd appreciate feedback on code, sprites, etc.. I'll create the required wiki info as soo as I find out this gets merged.

## Why's this needed? 
Firstly, this kind of ammo fills a spot in makeshift ammo that isn't filled yet: Slugs. They're now very similar to their brand-made counterparts, but, due to being pipeshot, can only be used in break-actions, putting them in a fairly good spot balance-wise. This PR is inspired by [this forum thread](https://forum.ss13.co/showthread.php?tid=23707) on potato ammo.

## Todo:
- [x] projectile datum
- [x] ammo item
- [x] cleanable potato splat
- [x] proj sprite
- [x] cleanable sprite
- [x] ammo sprite
- [x] adjust damage according to feedback

## Changelog 

```changelog
(u)Chasu
(*)Added potato pipeshot ammo, a makeshift slug equivalent for break-action shotguns.
```


code/obj/item/assembly/misc_assemblies.dm
```diff
@@ -1081,6 +1081,12 @@ ABSTRACT_TYPE(/datum/pipeshotrecipe)
 	accepteditem = /obj/item/material_piece/bone
 	craftname = "bone chunk"
 
+/datum/pipeshotrecipe/bone
```
TDHooligan, line 1084: typo here!

TDHooligan:
I like it. uncertain how knockback rounds might interact with something like the 4 letter word though. 

i think you could totally kick this up to actually be a full range slug. & move the damage up to 15. Keep it so it knocks back at close range.
I also don't think tying it to botany potency is necessary either.

itchybones:
All ready for review.

code/obj/item/gun/ammo.dm
```diff
@@ -881,6 +881,12 @@ ABSTRACT_TYPE(/obj/item/ammo/bullets/pipeshot)
 	ammo_type = new/datum/projectile/special/spreader/buckshot_burst/bone
 	icon_state = "makeshiftbone"
 
+/obj/item/ammo/bullets/pipeshot/potato // scrap handmade bone shells
```
TobleroneSwordfish, line 884:
```suggestion
/obj/item/ammo/bullets/pipeshot/potato
```

code/modules/projectiles/bullet.dm
```diff
@@ -1051,6 +1051,42 @@ toxic - poisons
...
+			var/mob/living/carbon/human/M = hit
+			if(proj.power >= 16)
+				var/throw_range = min(ceil(proj.power / 10), 3)
+
+				var/turf/target = get_edge_target_turf(M, dirflag)
+				M.throw_at(target, throw_range, 1, throw_type = THROW_GUNIMPACT)
+				M.update_canmove()
+			hit.changeStatus("staggered", clamp(proj.power/10, 5, 1) SECONDS)
+
+		if (iswall(hit))
```
TobleroneSwordfish, line 1079: Maybe make this work on windows/other objects too?

TobleroneSwordfish (approved):
Fun, seems reasonably balanced, potato 👍

code/modules/projectiles/bullet.dm
itchybones: I'll try using `else`

itchybones:
Merge conflicts _should_ be fixed now (?)

code/modules/projectiles/bullet.dm
```diff
@@ -1051,6 +1051,42 @@ toxic - poisons
...
+
+				var/turf/target = get_edge_target_turf(M, dirflag)
+				M.throw_at(target, throw_range, 1, throw_type = THROW_GUNIMPACT)
+				M.update_canmove()
+			hit.changeStatus("staggered", clamp(proj.power/10, 5, 1) SECONDS)
+
+		else
+			var/turf/W = get_turf(hit)
+			playsound(W, 'sound/impact_sounds/Slimy_Hit_1.ogg', 100, 1)
+			make_cleanable(/obj/decal/cleanable/potatosplat, W)
```
TobleroneSwordfish, line 1082:
```suggestion
			var/turf/T = get_turf(hit)
			playsound(T, 'sound/impact_sounds/Slimy_Hit_1.ogg', 100, 1)
			make_cleanable(/obj/decal/cleanable/potatosplat, T)
```

PR #22611, opened 2025-03-15
Add Monkey Riot radio station record

FlameArrow57:
[GAME OBJECTS]
## About the PR 
Adds a new funk category radio station record, Monkey Riot, by WyrdDoe

## Why's this needed? 
More music is nice

## Changelog 

```changelog
(u)WyrdDoe
(+)Added a new radio station record, Monkey Riot!
```


TobleroneSwordfish:
This will need hitting with the compression mallet until it's <500kb, 2.88mb is waaay too big for a 2 minute track. Also not sure about a song with lyrics, might be a little too distracting even if they are only occasional and distorted? Otherwise, it sounds like a nice tune but I'm not really a music person.

FlameArrow57:
> This will need hitting with the compression mallet until it's <500kb, 2.88mb is waaay too big for a 2 minute track. Also not sure about a song with lyrics, might be a little too distracting/confusing even if they are only occasional and distorted? Otherwise, it sounds like a nice tune but I'm not really a music person.

Is closer to 1 MB fine? I saw there were other tracks closer to 1 MB

FlameArrow57:
> There _are_ but they're around 4 minutes in length. Comparable length tracks in the same format are in the 200-800kb range. Although if it sounds really bad with compression due to higher complexity audio feel free to ignore that and let it sit a bit higher. The idea is just to avoid lagspiking people with poor connections when they have to download the track.

Okay it's now more around 700 kb now. Composer was questioning the "no lyrics" idea since there are advertisements with lyrics (asking more for feedback)

Valtsu0:
You can always just press the stop music button if the lyrics get distracting.

Unrelated but what has happened to github mobbile, why does every comment have like 4 screens worth of blank space after it

Mantweerp:
I also don't really think the lyrics are an issue seeing as Second Reality also has lyrics which musically serve the same function as in this song (though you could say this is excuseable because Second Reality is relatively rare (buuuuuuuuut I don't really think the lyrics are an issue at all)

Issue #21811, opened 2024-12-24
Game still keeps forgetting custom key binds, its so buggy


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

What the title says, key bindings will be partially or completely forgotten at random times, even within rounds

### Steps to reproduce

Bind
Wait
Profit

### Expected Behavior

I want this to work, its really messing me up

### Additional Information & Screenshots

It also sometimes duplicates bindable actions (example: Fart: F and then Fart: CTRL+F somewhere down the list)

Reported by: guvguvington
Client version: 515.1647
On server: Goonstation 4 Roleplay: Sylvester
Active test merges: #21006, #21792, and #21804
Round log date: 2024-12-24-07-59
Reported on: 2024-12-24 09:19:47
Round ID: 196022
Map: KONDARU

Garash2k:
Can confirm i've noticed the duplicated keybinds issue during #21681, I think it gets fixed while preparing the data for tgui. I've heard about keybind resets, it'd have to be investigated

Garash2k:
Historical note: #21681 was testmerged on Jan 14 and merged on Jan 20, after this was reported.

Issue #18415, opened 2024-04-05
Keybind resetting


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

Keybinds get reset when I try to change "rest toggle"

### Steps to reproduce

open the keybinding menu, switch use in hand key to "E" and equip to "Z" 
save and close
open keybinding menu again and change "rest" to "V"
That round or later rounds keybindings gets reset except "rest"

### Expected Behavior

keybindings should not switch back to the original

### Additional Information & Screenshots

I am usin TG controls and UI.

Reported by: Wantilathor
On server: Goonstation 3 Roleplay: Morty
Active test merges: #18275, #18329, and #18294
Round log date: 2024-04-05-16-57
Reported on: 2024-04-05 17:47:05
Map: COGMAP

Issue #17789, opened 2024-01-27
Game forgets keybinds between rounds


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

I have custom keybinds and every round SOME of them reset despite me saving it, not all funnily enough and its completely random

### Steps to reproduce

Have custom keybinds. I set R for rest, CTRL+R for flip, f for block and z for fart
Wait for next round
Goof chance that SOME of the keybinds reverted back, not all, mind you, only one pair

### Expected Behavior

Game properly saving keybinds like it claims it does

### Additional Information & Screenshots

88w88

Reported by: Guv Guvington
On server: Goonstation 3 Roleplay: Morty
Active test merges: #17577 and #17478
Round log date: 2024-01-27-15-20
Reported on: 2024-01-27 16:46:04
Map: NADIR

AXLFAN2000:
This happens to me as well

MeggalBozale:
keybinding has been bugged in many ways for a while now

Nexusuxen:
I explored this issue a bit, and while I'm not knowledgeable enough to _fix_ it, I have identified a possible culprit.
![Image](https://github.com/user-attachments/assets/fe17c464-bd8b-4a04-a13c-e5cc9e75ba14)
This section of code from `keybind_menu_chui.dm` seems to be effectively _replacing_ the previous keybind list from one's cloud save with the new one that's generated when you hit "Confirm" in the Modify Keybinds menu. Hitting Confirm erases all keybind data you didn't just set. You can actually see this if you leave and rejoin!

You can _sort of_ work around this issue by just setting all your keybinds all at once, buuuuut you can't have keybinds for mutually exclusive roles persist across rounds, such as Human and Cyborg. This is also particularly annoying if you have a LOT of keybinds and want to add a new one.

From my testing this seems to be mostly accurate, but I can't say with certainty whether or not the root of the issue is what I've described here.

Issue #12529, opened 2022-12-27
Modify Keybinds menu still incredibly buggy


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

I wanted to change my rest keybind to TAB, and all of my keybinds got reset. I've been trying to fix them since, but they keep getting reset any time I try to fix them. I'm also unable to change my keybinds in ghost vr, since whenever I put a keybind in, the menu just closes.

### Steps to reproduce

1. Change a keybind
2. All your keybinds reset
3. Won't stop resetting when you change them again

### Expected Behavior

For the keybinds menu to respect my settings and not constantly reset.

### Additional Information & Screenshots

Reported by: Muzcl
On server: Goonstation 1 Classic: Heisenbee
Round log date: 2022-12-27-17-09
Reported on: 2022-12-27 18:26:46
Map: COGMAP

PR #22632, opened 2025-03-17
fix keybind cloud data not updating properly

Nexusuxen:
[BUGFIX][INTERNAL]

## About the PR 

Merges newly entered custom keybinds with previous keybinds in player data, instead of overwriting the previous keybind list with the new one. Most likely a fix for #17789 and #8521, and maybe also possibly #21811 and #18415 and #12529. Granted, these are all describing the same issue, but hey! 5 bug reports squashed at once! 

Note: I am in very unfamiliar territory and would appreciate a little extra scrutiny to make sure I didn't bork anything with player data. The testing I did also wasn't mega rigorous but it does at least *seem* to work.
## Why's this needed? 

i squash the Bug 🐛

## Changelog 

```changelog
(u)Nexusuxen
(*)Custom keybinds should properly save between rounds now.
```

### DIFF
--- a/code/modules/interface/keybind_menu_chui.dm
+++ b/code/modules/interface/keybind_menu_chui.dm
@@ -54,7 +54,13 @@
 					var/datum/keymap/keydat = new(changed_keys_rev) //this should only have the changed entries, for optimal merge
 					current_keymap.overwrite_by_action(keydat)
 					current_keymap.on_update(owner)
-					owner.player.cloudSaves.putData("custom_keybind_data", json_encode(changed_keys_rev))
+					var/fetched_keylist = owner.player.cloudSaves.getData("custom_keybind_data")
+					var/new_keybind_data = list()
+					if (!isnull(fetched_keylist) && fetched_keylist != "") //The client has a list of custom keybinds.
+						new_keybind_data = json_decode(fetched_keylist)
+					for (var/i in changed_keys_rev)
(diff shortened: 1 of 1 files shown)


ZeWaka (commented):
I'm not quite sure how this is different, perhaps you could explain it to me a bit.
Aren't players custom keybinds necessarily populated by the cloud in the first place? The only discrepancy i can think of is someone saving while connected to another server.

Nexusuxen:
Player keybinds are properly populated by the cloud, yes. However, updating the cloud data with new keybinds *erases* previous keybind cloud data. E.g If your current cloud data reads as "CTRL+F = fart ; F =  rest", and you open the keybinds window to change they keybind for Walk/Run to X, you'd *expect* the cloud data to now read "CTRL+F = fart ; F =  rest ; X = walk". However, this doesn't happen, and instead "X = walk" *overwrites* "CTRL+F = fart ; F =  rest", resulting in the cloud data reading as "X = walk". This is doubly confusing as keymappings *in-game* do update properly, so it seems like nothing is wrong, but only until the next round when you realize your keybinds were erased.

I did just wake up so this explanation probably borders on 'incoherent', but hopefully that makes sense? I'd recommend trying the above example or something similar yourself in a live or local server.

I did also leave some additional info in a comment I left on #17789 that I may have forgotten to include here.

PR #22644, opened 2025-03-18
Different skulls make different skull chalices.

Cthucky:
[GAME OBJECTS]

## About the PR 
Adds skull chalices for each different skull type and hopefully makes them craftable.  I don't actually know what I'm doing.

## Why's this needed? 
It's always bothered me that changeling skulls magically become human skulls when you stick a rod in them.

## Changelog 

```changelog
(u)Mouse
(+)The appearance of a skull chalice now depends on the appearance of the skull it was made from.  Some skulls are better at holding liquids than others.
```


ZeWaka (approved):
cute

PR #22650, opened 2025-03-19
Sliding attack on sanctuary zone fix

### LINKED ISSUE #22592: Slide Attacks Bypassing Sanctuary Zone
### Labels

[BUG][MINOR]

### Description

Sliding into players in sanctuary zones bypass protection and cause damage to the player who is hit by the slide.

### Steps to reproduce

1. Be in a sanctuary zone.
2. Guard, swap to 'throw' intent.
3. Slide into another player.

### Expected Behavior

Slide Attack shouldn't cause damage in sanctuary zones.

Reported by: hollywoods3
Client version: 516.1658
Active test merges: #21781 and #22547
Round log date: 2025-03-14-04-32
Reported on: 2025-03-14 05:14:52
Round ID: 202127
Map: OSHAN

nova2053:
[BUG] [ACTIONS]
## About the PR 
Prevents mobs in sanctuary zones from being damaged by sliding

## Why's this needed? 
Fix #22592

### DIFF
--- a/code/obj/item/grab.dm
+++ b/code/obj/item/grab.dm
@@ -944,19 +944,16 @@
 						else if (H.limbs.l_leg)
 							damage += H.limbs.l_leg.limb_hit_bonus
 					if(issilicon(dive_attack_hit))
-						playsound(src.loc, 'sound/impact_sounds/Metal_Clang_3.ogg', 60, 1)
-						for (var/mob/O in AIviewers(user))
-							O.show_message(SPAN_ALERT("<b>[user] slides into [dive_attack_hit]! What [pick_string("descriptors.txt", "borg_punch")]!</b>"))
+						playsound(user, 'sound/impact_sounds/Metal_Clang_3.ogg', 60, 1)
+						user.visible_message(SPAN_ALERT("<b>[user] slides into [dive_attack_hit]! What [pick_string("descriptors.txt", "borg_punch")]!</b>"))
 					else if (AR.sanctuary)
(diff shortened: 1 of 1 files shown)


ZeWaka (commented):
Why would they be able to hit silicons?

nova2053:
> Why would they be able to hit silicons?

I don't think sliding into silicons deals damage to either party to begin with, just gives a different text message and plays a different sound

code/obj/item/grab.dm
```diff
@@ -946,6 +947,10 @@
 						playsound(src.loc, 'sound/impact_sounds/Metal_Clang_3.ogg', 60, 1)
 						for (var/mob/O in AIviewers(user))
 							O.show_message(SPAN_ALERT("<b>[user] slides into [dive_attack_hit]! What [pick_string("descriptors.txt", "borg_punch")]!</b>"))
+					else if (AR.sanctuary)
+						playsound(user, 'sound/impact_sounds/Generic_Hit_2.ogg', 50, TRUE, -1)
+						for (var/mob/O in AIviewers(user))
+							O.show_message(SPAN_ALERT("<B>[user] slides into [dive_attack_hit] harmlessly!</B>"), 1)
```
ZeWaka, line 953: this should be visible message, same with above and below

ZeWaka (changes requested):
fair enough, one change while you're here

code/obj/item/grab.dm
nova2053: Done in last commit, also noticed sliding into cyborgs didn't play the sound it was supposed to play and fixed it

PR #22657, opened 2025-03-19
Rework how plasmalungs trait is applied to mobs and antags

### LINKED ISSUE #22112: Plasma tank for plasmalung spawning at job location instead of with you
### Labels

[BUG][MAJOR]

### Description

I noticed if you have heavy sleeper, pet lover, plasmalung, and a job that spawns you with an item it leaves the plasmatank where you would've otherwised spawned without heavy sleeper

### Steps to reproduce

1. Take the plasmalung, heavy sleeper, and pet lover traits (I also had the pug trait if that effects it as well with the dog biscuit)
2. Spawn as a job that typically spawns with an item (such as RD and their clip board or Chapelin and their bible)
3. When you spawn in your pet carrier and dog treat are glued to you but your plasma tank is instead in your job spawn area instead of glued to you or on the ground so you instantly start dying

### Expected Behavior

I expected that the plasmatank would also be glued to me instead of spawning in my job area

Reported by: jamesthesnail
Client version: 515.1636
Active test merges: nothing
Round log date: 2025-01-26-03-36
Reported on: 2025-01-26 03:54:14
Round ID: 198536
Map: CLARION

pgmzeta:
[bug][traits]
## About the PR 
Reworks how the plasmalungs trait is applied to mobs on spawn:
* New helper proc: `put_in_hand_or_stow`
  * Attempts to place an item in-hand, and if that fails, in storage, and then either delete or place the item on the ground based on a var.
  * Call this in job setup wherever items are put in hand or dropped.
* Changes to `equip_body_traits`
  * Now accepts an `extended_tank` var, which will handle replacement of the extended oxygen tanks normally given to antagonists
  * Uses `put_in_hand_or_stow` for the mini plasma tank
* Antagonists
  * Antagonists that spawn with sensory items now also call `equip_body_traits` with `extended_tank` set to `TRUE`
  * Salvagers have all traits removed, so remove trait handling and leave a comment explaining why we don't call `equip_sensory_items` / `equip_body_traits` instead

## Why's this needed? 
Fixes #22112
Reduces code duplication/special handling for plasmalungs trait in antagonist code

### DIFF
--- a/code/procs/jobprocs.dm
+++ b/code/procs/jobprocs.dm
@@ -530,17 +530,30 @@ else if (istype(JOB, /datum/job/security/security_officer))\
 			src.stow_in_available(src.ears)
 		src.equip_if_possible(new /obj/item/device/radio/headset/deaf(src), SLOT_EARS)
 
+/**
+Equip items from body traits.
+
+ * @param is_antagonist - If TRUE, the mob is an antagonist and will spawn with a pocket extended tank instead of a mini tank.
+
+**/
 /// Equip items from body traits
-/mob/living/carbon/human/proc/equip_body_traits()
+/mob/living/carbon/human/proc/equip_body_traits(is_antagonist=FALSE)
 	if (src.traitHolder && src.traitHolder.hasTrait("plasmalungs"))
 		if (src.wear_mask && !(src.wear_mask.c_flags & MASKINTERNALS)) //drop non-internals masks
 			src.stow_in_available(src.wear_mask)
 
 		if(!src.wear_mask)
 			src.equip_if_possible(new /obj/item/clothing/mask/breath(src), SLOT_WEAR_MASK)
-
-		var/obj/item/tank/good_air = new /obj/item/tank/mini_plasma(src)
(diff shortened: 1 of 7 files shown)


code/procs/jobprocs.dm
```diff
@@ -530,17 +530,29 @@ else if (istype(JOB, /datum/job/security/security_officer))\
...
 
-/// Equip items from body traits
-/mob/living/carbon/human/proc/equip_body_traits()
+/**
+Equip items from body traits.
+
+ * @param is_antagonist - If TRUE, the mob is an antagonist and will spawn with a pocket extended tank instead of a mini tank.
+
+**/
+/mob/living/carbon/human/proc/equip_body_traits(is_antagonist=FALSE)
```
TobleroneSwordfish, line 539: `is_antagonist` is a confusing name since it doesn't apply to most antagonists, why not just `extended_tank`?

TobleroneSwordfish (changes requested):

TobleroneSwordfish (approved):

PR #22658, opened 2025-03-19
Fix turboflash deconstruction

### LINKED ISSUE #22144: Dissambling flash cell assemblies does not drop a flash
### Labels

[BUG][MINOR]

### Description

If you try to dissamble a flash cell assembly, it will not drop a flash

### Steps to reproduce

1. Grab flash cell assembly
2. Go throught the dissamble proccess
3. Only a powercell drops, no flash

### Expected Behavior

A flash to drop aswell, like every other dissambly drops every item used to assemble it

Reported by: angelx3
Client version: 515.1637
Active test merges: nothing
Round log date: 2025-01-28-15-56
Reported on: 2025-01-28 16:39:40
Round ID: 198725
Map: COGMAP

JORJ949:
## About the PR 
Fixes flash-cell assemblies not dropping a flash when deconstructed

## Why's this needed? 
Fixes #22144

### DIFF
--- a/code/obj/item/device/flash.dm
+++ b/code/obj/item/device/flash.dm
@@ -400,12 +400,11 @@ TYPEINFO(/obj/item/device/flash/turbo)
 			return
 		if (iswrenchingtool(W) && !(src.secure))
 			boutput(user, "You disassemble [src]!")
-			src.cell.set_loc(get_turf(src))
-			var/obj/item/device/flash/F = new /obj/item/device/flash(get_turf(src.cell)) // is src.cell's turf because src's turf causes #22144 somehow
+			var/obj/item/device/flash/F = new /obj/item/device/flash(get_turf(src))
 			if(!src.status)
 				F.status = 0
 				F.set_icon_state("flash3")
-			qdel(src)
+			src.cell.set_loc(get_turf(src)) //Do this last as removing the cell deletes the turboflash
 		else if (isscrewingtool(W))
 			boutput(user, SPAN_NOTICE("You [src.secure ? "unscrew" : "secure"] the access panel."))
 			secure = !secure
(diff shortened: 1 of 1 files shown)


JORJ949:
Bug is apparently intentional but it doesn't make sense to me.

code/obj/item/device/flash.dm
```diff
@@ -401,7 +401,7 @@ TYPEINFO(/obj/item/device/flash/turbo)
 		if (iswrenchingtool(W) && !(src.secure))
 			boutput(user, "You disassemble [src]!")
 			src.cell.set_loc(get_turf(src))
-			var/obj/item/device/flash/F = new /obj/item/device/flash( get_turf(src) )
+			var/obj/item/device/flash/F = new /obj/item/device/flash(get_turf(src.cell)) // is src.cell's turf because src's turf causes #22144 somehow
```
pgmzeta, line 404: the flash is qdel'd on purpose above, on L90-L93
```	Exited(Obj, newloc)
		. = ..()
		if(Obj == src.cell)
			qdel(src) //cannot un-turboflash```

Should probably just not be spawned?

JORJ949: I am down to change the code either way, will await dev decision, code is going both ways on if it wants a flash spawned or not. It is intentional but feels odd to me at least.

JORJ949:
Please DNM until after a decision is made on if them dropping flashes is wanted as either way there are better fixes to be made.

code/obj/item/device/flash.dm
Tarmunora: exited override was part of a sweeping "oh god we don't ever account for large cells getting deleted while in use" fixbatch. Decon behaviour should drop cell and flash

PR #22666, opened 2025-03-20
Add a puzzle Input to Hangman Component

sunnyariess:
[qol]
## About the PR 
Adds an input field to the hangman component, letting "Puzzle" be set by other components.

## Why's this needed? 
More mechcomp interactability, you no longer need to necessarily set it with a multitool.


ZeWaka (commented):
how is this different than the other one, its unclear

sunnyariess:
> how is this different than the other one, its unclear

currently, the only way to modify "Puzzle" is through a multitool, this lets you do it remotely through other components without manual input;
i did copy and paste the code because the logic is handled elsewhere, so i thought itd be fine :]

Issue #22556, opened 2025-03-11
Pipebombs can be filled indefinitely


goonstation-issuebot:
### Labels

[BUG][CRITICAL]

### Description

You can infinitely fill a pipebomb with a reagent. Black powder also scales indefinitely. Put those two together. You can delete literally everything on a z level within minutes. This is not exaggeration.

### Steps to reproduce

1. make pipebomb
2. fill, using a container. you'll soon notice you can fill a hollow pipe frame with thousands of units. black powder has no explosion cap unlike most chemicals, so this can get dangerous and fast.
3. arm it like any other pipebomb, you can blow up an entire z level with just 600 units of BP inside it

### Expected Behavior

I don't expect pipebombs to be physics-defying infinite containers, and I don't expect being able to gib everybody on the station immediately (NO EXAGGERATION) to be that easy.

### Additional Information & Screenshots

This doesn't just apply to reagents with pipebombs, if you click on a pipebomb with cables it'll consume all the cables at once rather than just 5 or something. I haven't tested this out yet but I assume this glitch also applies to the shrapnel you can fill pipebombs with.

Reported by: gargathuganoktheannihilator255
Client version: 515.1647
On server: Goonstation 1 Classic: Heisenbee
Active test merges: #21781
Round log date: 2025-03-11-09-20
Reported on: 2025-03-11 09:29:36
Round ID: 201911
Map: NADIR

TobleroneSwordfish:
Fixed by reverting #22259

M-Earthfire:
Some Screenshots i scrounged up from the discord to emphasised how much that gibbed nadir. 

Credits go to Chillcube and zennie for the screenshots

<img width="1288" height="1259" alt="Image" src="https://github.com/user-attachments/assets/8bb93711-c822-4411-bb87-272820683290" />

<img width="1335" height="1066" alt="Image" src="https://github.com/user-attachments/assets/d551cda8-71e6-49f2-aad6-f2ad2f21ed11" />

<img width="1090" height="896" alt="Image" src="https://github.com/user-attachments/assets/849994f6-cab4-4f47-99f3-b3e344d5f6c1" />

<img width="366" height="109" alt="Image" src="https://github.com/user-attachments/assets/79ec029d-e9eb-4a1c-b447-04668c2cd15e" />

Issue #22417, opened 2025-02-23
Borg welders


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Clicking on a metal rod as a borg to weld them into metal sheets turns it into a welder/rod assembly and you lose your welder permenantly.

### Steps to reproduce

1. Be borg. Engineering module.
2. Try to weld a rod.
3. It takes it from you.

### Expected Behavior

I expected be be given the option of how many sheets I wanted to turn the rods into.

### Additional Information & Screenshots

https://ibb.co/jvfJmVXRS

Reported by: scientifically
Client version: 515.1639
On server: Goonstation 4 Roleplay: Sylvester
Active test merges: nothing
Round log date: 2025-02-23-21-28
Reported on: 2025-02-23 22:15:02
Round ID: 200732
Map: DONUT2

mordent-goonstation:
Presumably will be fixed by #22259

PR #22669, opened 2025-03-20
introduces `/datum/assembly_comp_helper` to the assembly component

M-Earthfire:
[internal][Bug]

## About the PR 

introduces `/datum/assembly_comp_helper`. This is a helper datum added to the assembly-component, `/datum/component/assembly`.

This component holds multiple variables and behaviour to adjust the assembly component as needed. With `/datum/assembly_comp_helper/proc/replacement_proc` it also allows an alternative proc to be called instead of one on the item the component is added to.

Currently the behaviour this component covers is the former "allow_on_others" variable and the checking for item arms or borg items.

## Why's this needed? 

Right now, #22417 is still an issue. #22259 fixed that, but was reverted in [60a48e4](https://github.com/goonstation/goonstation/commit/60a48e478b26d56f9d59e487b324460faec57214), because it caused #22556

This PR, by not creating sub types of a component (something that is discouraged because global component procs do not check inheritance) tries to fix the bug while not introducing station-wiping bugs. 

The replacement proc is necessary infrastructure for #22115, because that one right now works via a subtype of the assembly component as well.


ZeWaka:
x

PR #22673, opened 2025-03-20
Add MouseDrop to the Aquatic Research Terminal

HyenaShepherd:
## About the PR 
- Adds drag-and-drop functionality for depositing fish to the Aquatic Research Terminal.
- User can click a fish and drag it to the terminal in order to quickly deposit every fish of that type (bass, salmon, goldfish, etc.).
- A quiet bubbling noise will accompany each fish being deposited.

## Why's this needed? 
Currently, depositing fish into the Aquatic Research Terminal is click-intensive and time consuming, as the user has to manually pick up each fish and use it on the terminal individually. This also leads to the occasional problem of double-clicking the terminal and causing it to begin uploading the data (which takes a few seconds), causing unnecessary delays. Adding MouseDrop to the terminal is a quality of life change that alleviates both of these issues.

[qol]

### DIFF
--- a/code/modules/fishing/fishing_gear.dm
+++ b/code/modules/fishing/fishing_gear.dm
@@ -427,32 +427,10 @@ TYPEINFO(/obj/item/fish_portal)
 			W.set_loc(src)
 			W.dropped(user)
 
-	MouseDrop_T(atom/movable/O as mob|obj, mob/user as mob)
-		if (!isliving(user))
-			boutput(user, SPAN_ALERT("You can't do that while dead, this isn't Davy Jones' Locker!"))
-			return
-		if (BOUNDS_DIST(user, src) > 0)
-			boutput(user, SPAN_ALERT("You need to move closer to the [src] to do that."))
-			return
-		if (BOUNDS_DIST(O, user) > 0)
-			boutput(user, SPAN_ALERT("[O] is too far away to load into the [src]!"))
-			return
-		if (src.working)
-			boutput(user, SPAN_ALERT("The [src] is busy!"))
-			return
-
-		if (istype(O, /obj/item/reagent_containers/food/fish/))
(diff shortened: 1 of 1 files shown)


Tarmunora (changes requested):
Any particular reason to not use `/datum/component/transfer_input/quickloading`?

HyenaShepherd:
> Any particular reason to not use `/datum/component/transfer_input/quickloading`?

A _very_ simple reason: I did not know it existed. Thank you for pointing it out! I'll look into how to use it correctly and update accordingly.

code/modules/fishing/fishing_gear.dm
```diff
@@ -427,6 +427,11 @@ TYPEINFO(/obj/item/fish_portal)
 			W.set_loc(src)
 			W.dropped(user)
 
+	New()
+		..()
+		AddComponent(/datum/component/transfer_input/quickloading, allowed)
+		AddComponent(/datum/component/transfer_output)
```
TobleroneSwordfish, line 433: This component won't actually do anything unless you use the `TRANSFER_OR_DROP` macro for all outputted items. So instead of doing `item.set_loc(get_turf(src))` you'd do `TRANSFER_OR_DROP(item, get_turf(src))`

TobleroneSwordfish (changes requested):

code/modules/fishing/fishing_gear.dm
TobleroneSwordfish: The macro sends a signal to the `transfer_output` component with the item to transfer, then the component handles putting it there. Currently (line 391) it's just directly setting the location of the items to the turf it's sitting on, so the component isn't doing anything.

For clarity, the `/datum/component/transfer_input/quickloading` works fine but the `/datum/component/transfer_output` doesn't.

Tarmunora: x

Tarmunora (changes requested):

PR #22678, opened 2025-03-21
Make freezers constructible

FlameArrow57:
[ATMOSPHERICS][GAME OBJECTS][FEATURE][BALANCE]

## About the PR 
Makes freezers constructible for 10 matter units

## Why's this needed? 
Would be cool

## Changelog 

```changelog
(u)FlameArrow57
(*)The handheld pipe dispenser can now create freezers
```


FlameArrow57:
Will add breakability later

PR #22684, opened 2025-03-22
Proper NT Commander job

JORJ949:
## About the PR 
Adds a special NT commander job with gear more useful in a real round compared to a pod commander (who gets a podwars token that is worthless on station). That being the following:
- NT Commander's Sabre (same as syndicate commander's but without delimb chance
- Commander's great coat and beret (spaceproof)
- NTSO pouch with clock 180
- Security-Command clone-disk (the HoS one that contains netpass_heads alongside netpass_sec) ((so if they really need it they can packet-hack armory)) (((maybe replace with a blue commander auth disk for authing stuff? though that may detract from the captain's role so I didn't do that here)))
- Big heavy HoS boots (this is the commander of NT squads like HoS is commander of sec so)
- NT Operative glasses, combining sechuds and prodocs into one (they are meant to be able to command any team of responders, including medical teams)
- NT Commander headset, with access to station channels + NT channel
image

## Why's this needed? 
The NT commander from pod wars is occasionally used for admin gimmicks, this introduces one with gear actually usable in round (not the podwars token).  I didn't give it the powerful NTSO gun as I want the role to be more aimed at coordinating the team and they may be spawned to be in charge of just some medics and engineers, though I would not be against swapping the clock180 for a cooler gun with both stun and lethal settings like HoS and NTSC have.


code/obj/item/clothing/glasses.dm
```diff
@@ -261,6 +261,43 @@ TYPEINFO(/obj/item/clothing/glasses/sunglasses/tanning)
 	color_b = 1
 	contraband = 4 // illegal (stolen) crimefighting vigilante gear
 
+/obj/item/clothing/glasses/nt_operative
+	name = "\improper NanoTrasen Operative HUD"
+	desc = "Patented NT technology compacting many different HUDs into one compact set of glasses.  Enhanced shielding blocks many flashes."
+	icon_state = "nt"
+	item_state = "sunglasses"
+	protective_temperature = 1300
+	var/already_worn = 0
```
TobleroneSwordfish, line 270: Unused var?

code/obj/item/clothing/glasses.dm
```diff
@@ -261,6 +261,43 @@ TYPEINFO(/obj/item/clothing/glasses/sunglasses/tanning)
 	color_b = 1
 	contraband = 4 // illegal (stolen) crimefighting vigilante gear
 
+/obj/item/clothing/glasses/nt_operative
+	name = "\improper NanoTrasen Operative HUD"
+	desc = "Patented NT technology compacting many different HUDs into one compact set of glasses.  Enhanced shielding blocks many flashes."
+	icon_state = "nt"
+	item_state = "sunglasses"
+	protective_temperature = 1300
```
TobleroneSwordfish, line 269: I think this var is entirely vestigial and doesn't do anything.

code/obj/item/clothing/glasses.dm
```diff
@@ -261,6 +261,43 @@ TYPEINFO(/obj/item/clothing/glasses/sunglasses/tanning)
...
+	var/already_worn = 0
+	color_r = 0.85
+	color_g = 0.85
+	color_b = 1
+
+	equipped(var/mob/user, var/slot)
+		..()
+		if (slot == SLOT_GLASSES)
+			//Night Vision
+			APPLY_ATOM_PROPERTY(user, PROP_MOB_NIGHTVISION, src)
```
TobleroneSwordfish, line 279: Maybe not NV too, you should be able to see bulky NV glasses on someone if they have it.

TobleroneSwordfish (changes requested):
Looks cool, good to have this as a separate gimmick role

code/obj/item/clothing/glasses.dm
JORJ949: How about thermals instead? Those give a minor amount of NV without being big and bulky, also the rest of the team can't see invisible threats so maybe the commander should

JORJ949: Also you cant see the NVGs on the NTSOs anyway as they have space helmets on

TobleroneSwordfish: Thermals should also be obvious I think, you could make them red but that would probably mess with the *aesthetic*

444Portal: Could use Mesons, maybe? If you're set on the Commander having some extra vision.

JORJ949: It's not required, I avoided giving mesons as I still want them to get stunned by singularities but they should have some sort of engineering gear.

JORJ949:
Additionally before I PR it would like to ask opinions on removing the command radio from other NT roles with the exception of the consultant, as they have their own NT channel now and the commander would ideally be coordinating with command or they coordinate within the department they are in.

TobleroneSwordfish:
> emoving the command radio from other NT roles

I thiink that makes sense? They're not really command roles, just well trained centcom grunts.

TobleroneSwordfish (approved):

PR #22695, opened 2025-03-23
Converts stunhats to use signals, adds the killhat.

RubberRats:
[CODE QUALITY] [CLOTHING]
## About the PR 

This PR refactors the code that handles stunhats to use `COMSIG_ATTACKBY` instead of hardcoding it into attack procs. It also adds the killhat, an admin-spawned variant that shocks the would-be headpatter with a lethal dose of electricity.

## Why's this needed? 

I was considering adding more stunhat-adjacent things, and doing that by adding onto the existing code would probably be bad practice. The killhat is an idea that I keep going back to and I think it would be funny to have around.


code/obj/item/clothing/hats.dm
```diff
@@ -1420,6 +1428,41 @@ TYPEINFO(/obj/item/clothing/head/that/gold)
...
+			src.uses = max(0, src.uses - 1)
+			if (src.uses < 1)
+				src.icon_state = splittext(src.icon_state,"-")[1]
+				src.item_state = splittext(src.item_state,"-")[1]
+				wearer.update_clothing()
+
+			if (src.uses <= 0)
+				wearer.show_text("The sunhat is no longer electrically charged.", "red")
+			else
+				wearer.show_text("The stunhat has [src.uses] charges left!", "red")
```
Tarmunora, line 1445: should probably return `TRUE` to prevent the attackhand from going through anyways

Tarmunora (changes requested):

PR #22698, opened 2025-03-23
New Trait: Touch Shy

RubberRats:
[TRAITS] [FEATURE] [ADD TO WIKI]
## About the PR 

This PR adds a new trait under the Skill category, for one point. It causes you to automatically attempt to shove-disarm someone when they try to headpat you.

## Why's this needed? 

Flavor and because I thought it was funny, mostly. Some people are weird about giving headpats, and other people are uncomfortable about getting them. I like the idea of this trait existing as an option for characters that want their personal space respected a bit more. If nothing else, it will probably lead to some interesting interactions.

## Changelog 

```changelog
(u)RubberRats and Xenomni
(*)Adds a new trait: Touch Shy. Icon sprite by Xenomni
(*)Touch Shy will cause you to automatically attempt to shove anyone who tries to headpat you or do another help intent interaction.
```


TobleroneSwordfish:
I like this, but maybe it should just be all help intent clicks instead of specifically headpats?

RubberRats:
The initial concept for this trait was meant to be something for headpats, but making it help intents in general would be trivial and probably fits the theming better anyway.

NibChocolateeny:
but what if on fire, and some one esle try to put out the fire (with hlep intent)?

TobleroneSwordfish:
then you shove them >:D

PR #22725, opened 2025-03-25
Replaces the armory buckshot

RubberRats:
[GAME OBJECTS][BALANCE]
## About the PR 

This PR replaces the buckshot in the armory's AmmoTech with two boxes of birdshot. Unlike the buckshot, they are not hidden, and can be restocked normally.

## Why's this needed? 

The box of buckshot in the armory is a bit of an odd case. It is a massive jump in lethality compared to the other equipment there, able to stunlock and kill someone very quickly. At the same time, it is extremely limited: you get one box, it's locked behind hacking the vendor, and you can't restock it. This makes actually using it as security very awkward. Needing to hack the vendor in addition to opening the armory is strange, and even then, its high power and limited availability relegates its usage to "dumpster that one nerd."

I still think having a lethal kinetic option in the armory is a good idea. Birdshot fills a similar niche without the more oppressive elements. It has killing power on par with debris field gear, which feels about right, and crucially can't instantly stunlock a person. I think it is probably safe to have a few boxes around, especially considering it's tied to a two-handed weapon.

## Changelog 

```changelog
(u)RubberRats
(*)The hidden item in the armory's AmmoTech has been replaced with a few boxes of birdshot
```

### DIFF
--- a/code/modules/vending/vending.dm
+++ b/code/modules/vending/vending.dm
@@ -1561,6 +1561,7 @@ TYPEINFO(/obj/machinery/vending/medical)
 	create_products(restocked)
 		..()
 		product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/abg, 6)
+		product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/a12/bird, 2)
 		product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/a38, 2)
 		product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/a38/stun, 3)
 		product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/nine_mm_NATO,3)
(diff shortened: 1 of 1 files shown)


RubberRats:
This is potentially a terrible idea, but I had to get it out of my head. As for non-security using it: I think the birdshot is about as desirable as the old buckshot in terms of looting the armory. As something people are hoarding using restocks, I think the lack of throws/instant stuns and the slight drop in lethality make it less abusable. Might be worth having a feedback thread for this, not sure.

JORJ949:
I am not against nerfing the buckshot but I do heavily disagree with making it both not hidden and restockable, there are plenty good reasons why the armory doesn't have lethal weapons beyond the poor phaser (although it can have an egun now and tbh I'm not sure I agree with that , the egun is easily the best thing in the armory far outshining the rest if it spawns). 

"Needing to hack the vendor in addition to opening the armory is strange, and even then, its high power and limited availability relegates its usage to "dumpster that one nerd."" This is good, it SHOULD feel wrong, its a last-ditch effort you shouldn't be using unless you have tried literally everything else, it also stops people going "oh a lethal shotgun I should grab that!" and further encourages getting better stun weapons. This is also a huge buff to the armory's kill power as before you could "dumpster that one nerd" now you can combat multiple crew.

Not a big fan of upping armory kill-power, I much prefer it being less-than-lethal utility options.

TDHooligan:
the stretch case of having to dumpster one nerd is exactly why it's in there. 
security rarely have to shoot more than one guy with lethals, which is why restocking was removed. 

the only reason you'd restock these rounds is to shoot a whole lot of people with lethals. and security are not going to be doing that.

flappybatpal:
I don't think this is necessary? The reasoning for buckshot and it's limitations have been well explained in comments.

PR #22753, opened 2025-03-28
Heads can be placed on cabinets like hats

CalliopeSoups:
[A-Game-Objects] [C-Feature]
## About the PR 
Simple change lets you put heads on cabinets like hats. Skeletons can become the camera helmet!

## Why's this needed? 
Crawling heads would be funny. Skeletons get actual use out of this but not much more than they would if they just got a camera helmet.

## Changelog 

```changelog
(u)CalliopeSoups
(+)You can now put heads ontop of mechcomp cabinets like hats.
```

### DIFF
--- a/code/datums/components/hattable.dm
+++ b/code/datums/components/hattable.dm
@@ -40,7 +40,7 @@ TYPEINFO(/datum/component/hattable) // Take a walk through my TWISTED mind.... I
 	var/offsetBy_x = 0
 
 
-	if (istype(item, /obj/item/clothing/head/))
+	if (istype(item, /obj/item/clothing/head/) | istype(item, /obj/item/organ/head))
 		src.hat = item
 		ADD_FLAG(src.hat.appearance_flags, KEEP_TOGETHER) // Flags needed for wigs!
 		ADD_FLAG(src.hat.vis_flags, VIS_INHERIT_DIR)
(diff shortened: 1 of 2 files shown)


code/obj/item/organs/head.dm
```diff
@@ -19,6 +19,8 @@
 	tooltip_flags = REBUILD_ALWAYS //TODO: handle better??
 	max_damage = INFINITY
 	throw_speed = 1
+	var/default_hat_y = 0
+	var/default_hat_x = 0
```
Tarmunora, line 23: unused?

code/datums/components/hattable.dm
```diff
@@ -40,7 +40,7 @@ TYPEINFO(/datum/component/hattable) // Take a walk through my TWISTED mind.... I
 	var/offsetBy_x = 0
 
 
-	if (istype(item, /obj/item/clothing/head/))
+	if (istype(item, /obj/item/clothing/head/) | istype(item, /obj/item/organ/head))
```
Tarmunora, line 43:
```suggestion
	if (istype(item, /obj/item/clothing/head/) || istype(item, /obj/item/organ/head))
```

Tarmunora (changes requested):

PR #22759, opened 2025-03-28
Let people hit cyborgs with tools

### LINKED ISSUE #22327: Can't hit a silicon with a crowbar
### Labels

[BUG][MINOR]

### Description

Attempt to hit a silicon with a crowbar, be completely unable to. 

### Steps to reproduce

1. Try and hit a silicon with a crowbar
2. Be prompted that the cover is locked. 

### Expected Behavior

If attempting to harm a silicon with a crowbar, you should hit the silicon with it. 

Reported by: skotcher
Client version: 515.1643
Active test merges: #22185
Round log date: 2025-02-16-01-51
Reported on: 2025-02-16 03:34:11
Round ID: 200138
Map: DONUT3

Cinder-Arcanium:
[Silicon] [Bug]

## About the PR 

Adds a check to the cyborg attackby for intent, so you can hit cyborgs with tools. Also moves the emag check to the top so the emag interaction stays hidden even on harm intent.

## Why's this needed? 

You should probably be able to hit cyborgs with crowbars and screwdrivers to kill them, and should also probably be on help intent to repair and modify cyborgs.
fixes #22327

### DIFF
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -1090,7 +1090,7 @@
 	attackby(obj/item/W, mob/user)
 		if (istype(W, /obj/item/card/emag))
 			return
-		if (user.a_intent != INTENT_HELP)
+		if (user.a_intent == INTENT_HARM)
 			..()
 			return
 		if (istype(W,/obj/item/device/borg_linker) && !isghostdrone(user))
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -1087,10 +1087,10 @@
 					src.cell = null
 					src.part_chest?.cell = null
 
-	attackby(obj/item/W, mob/user)
+	attackby(obj/item/W, mob/user, params, is_special = 0, silent = FALSE)
 		if (istype(W, /obj/item/card/emag))
 			return
-		if (user.a_intent == INTENT_HARM)
+		if (user.a_intent == INTENT_HARM || is_special)
 			..()
 			return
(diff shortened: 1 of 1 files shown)


code/mob/living/silicon/robot.dm
```diff
@@ -1088,6 +1088,11 @@
 					src.part_chest?.cell = null
 
 	attackby(obj/item/W, mob/user)
+		if (istype(W, /obj/item/card/emag))
+			return
+		if (user.a_intent != INTENT_HELP)
+			..()
+			return
```
Tarmunora, line 1095: I'd probably check explicitly for `HARM` here, rather than just not help

JORJ949 (approved):

code/mob/living/silicon/robot.dm
```diff
@@ -1088,6 +1088,11 @@
 					src.part_chest?.cell = null
 
 	attackby(obj/item/W, mob/user)
+		if (istype(W, /obj/item/card/emag))
+			return
+		if (user.a_intent == INTENT_HARM)
+			..()
+			return
```
pgmzeta, line 1095: This change will allow tool special attacks done on disarm also hit cyborgs (i.e. screwdriver multi-stab)

```suggestion
	attackby(obj/item/W, mob/user, params, is_special = 0, silent = FALSE)
		if (istype(W, /obj/item/card/emag))
			return
		if (user.a_intent == INTENT_HARM || is_special)
			..()
			return
```

pgmzeta (changes requested):
good PR, recommend one change to improve consistency

PR #22760, opened 2025-03-28
Change name of rock material to stone

pgmzeta:
[matsci][qol]
## About the PR 
* Change name of rock material and material piece to stone
* Remove censorium material

## Why's this needed? 
material combinations stuff ending with "ock" are too ripe for breaking rule 4. I am hoping "one" is less troublesome.


ZeWaka:
rock AND stone :(

pgmzeta:
well you see the name is stone but the descriptions still say rock, so if anything this is more rock and stone per stone O:)

PR #16922, opened 2023-11-14
Resprites shields on cyborgs and wizards

gunwithaman:
[Silicons]

## About the PR 
Resprites the wizards force shield overlay as well as the mob overlay for the cyborgs heat and force shield upgrades

## Why's this needed? 
The old sprites were just a pulsing circle, i thought it would be fun to try and make the force fields stand out more
I would make the shields glow if I knew how to

https://youtu.be/6GtseIAhPGo
https://youtu.be/y75SUfIppes

PR #22768, opened 2025-03-29
Resprite cyborg shields

FlameArrow57:
[SILICONS]

## About the PR 
Adopted from #16922

physical shield

heat shield

## Why's this needed? 
Cool sprites


JORJ949:
I still think this is a little too obfuscating for being able to see who is inside the shield, perhaps it could be much fainter but the parts affected by the animated texture become more visible (the wave that goes across it). I also think it could do with a startup animation as in the original PR demonstration videos it immediately appears which is very jarring.

PR #22774, opened 2025-03-29
Miranda's Rights fixes

### LINKED ISSUE #22687: Inspector's Miranda is Security Miranda
### Labels

[BUG][MINOR]

### Description

The Inspector does not have their special miranda's rights

### Steps to reproduce

1. Be Inspector
2. Say your miranda
3. Start talking about space law

### Expected Behavior

Talk about corporate policy not space law.

_No response_

JORJ949:
## About the PR 
Miranda's rights are now by default null on mind and when trying to state null rights call assigned_role's proc that returns default rights.
Trying to manually state miranda's rights now checks if your job gets miranda's rights rather than using a magic list of job names.
You can now reset your miranda's rights back to null (and the miranda setting menu uses TGUI inputs)

## Why's this needed? 
Fixes #22687
PROC_REFs scare me, wrapping them in lists is even worse.

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -12,7 +12,7 @@ datum/mind
 	var/list/datum/dynamic_player_memory/dynamic_memories = list()
 	var/remembered_pin = null
 	var/last_memory_time = 0 //Give a small delay when adding memories to prevent spam. It could happen!
-	var/miranda // sec's miranda rights thingy.
+	var/miranda = null // sec's miranda rights thingy.
 
 	var/violated_hippocratic_oath = 0
 	var/soul = 100 // how much soul we have left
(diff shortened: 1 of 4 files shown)


code/datums/mind.dm
```diff
@@ -221,8 +221,9 @@ datum/mind
 		miranda = new_text
 
 	proc/get_miranda()
-		if (islist(src.miranda)) //isproc machine broke, so uh just wrap your procs in a list when you pass them here to distinguish them from strings :)
-			return call(src.miranda[1])()
+		if (isnull(src.miranda)) //isproc machine broke, so uh just wrap your procs in a list when you pass them here to distinguish them from strings :)
+			var/datum/job/job = find_job_in_controller_by_string(src.assigned_role)
+			return job.get_default_miranda()
```
pgmzeta, line 226: Why not set src.miranda instead of returning the value?

JORJ949: Because some miranda's (the inspector) has randomised elements so changes every time you use it.

PR #22776, opened 2025-03-29
Cargo shipping destination uses a landmark instead of area

pgmzeta:
[internal][station systems]
## About the PR 
* Replace the supply destination area with an equivilant landmark
* Calculate the shipping distance in `World/init` instead of `preMapLoad`, as the map landmarks aren't ready during `preMapLoad`
* Extend cargo areas to cover recieving belts.

## Why's this needed? 
Having an area as the destination has some awkward side effects:
* It never loses power, even if the neighboring cargo area does
* It's not a station turf, so there are some oddities with e.g. radstorms
* The area can't be built on/over, which is weird for a place so close to the station

### DIFF
--- a/maps/unused/setpieces/radioship.dmm
+++ b/maps/unused/setpieces/radioship.dmm
@@ -1,0 +1,5965 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"aM" = (
+/obj/lattice{
+	dir = 1;
+	icon_state = "lattice-dir-b"
+	},
+/obj/machinery/light/runway_light,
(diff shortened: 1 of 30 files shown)


maps/unused/setpieces/radioship.dmm
```diff
@@ -0,0 +1,5965 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
```
ZeWaka, line 1: ?

pgmzeta: doing too much at once, fixed

TobleroneSwordfish (approved):

PR #22779, opened 2025-03-29
Ghost critters and small animals are no longer affected by ants.

victorShchagin:
## About the PR 

Ghostcritters and small animals are no longer affected by ants.

[Critters]
[QoL]

## Why's this needed? 
Now that food on the floor generates ants - it's easy to get instakilled by a lethal dose of those, unless the ghostcritter player checks all the grub and generally avoids being fed by other players.

## Changelog 

```changelog
(u)Vic
(+)Ghostcritters no longer get hurt by ants. Go for that floor pickle!
```

### DIFF
--- a/code/modules/chemistry/Reagents-Misc.dm
+++ b/code/modules/chemistry/Reagents-Misc.dm
@@ -1697,6 +1697,8 @@ datum
 				. = ..()
 				if(volume < 1)
 					return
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (approved):
They are small so they eat around the ants 👍

PR #22789, opened 2025-03-30
Add rare metal mining event

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Adds a new mining event that can occur for asteroids where they spawn a random picked rare metal

Nanite swarms which drop iridium and plutonium, and neutronium can spawn

## Why's this needed? 
Ways to obtain more unique or rarer metals

## Changelog 

```changelog
(u)FlameArrow57
(*)Asteroids may now contain a rare metal, such as plutonium
```


FlameArrow57:
Was looking for feedback on the idea in general

pgmzeta:
Personally would prefer rarely randomly spawning nanites over direct iridium, as there's currently no real way to know that you shouldn't mine nanites like every other asteroid in the game.

FlameArrow57:
Sure sounds much better

code/obj/item/material.dm
```diff
@@ -486,6 +486,18 @@
...
+	name = "plutonium ore"
+	desc = "A rock containing just enough radioactive plutonium that it can be smelted."
+	material_name = "Plutonium"
+	default_material = "plutonium"
+
+/obj/item/raw_material/neutronium
+	name = "neutronium ore"
+	desc = "An ore containing deadly neutronium metal."
+	material_name = "Neutronium"
+	default_material = "neutronium"
```
TobleroneSwordfish, line 499: This doesn't really make sense for synthetic metals like plutonium, there's no such thing as "plutonium ore". Since they're dropping from nanites now maybe they could be chunks of scrap instead?

TobleroneSwordfish (changes requested):

PR #22794, opened 2025-03-31
Characters with a missing eye get an eyepatch on spawn

pgmzeta:
[feature][traits]
## About the PR 
Add a feature to part customization to optionally apply a trait to characters.
Add unselectable traits for left and right missing eyes.
Add check for the new traits when equipping sensory items that will add an eyepatch of the correct side, or a blindfold if both of your eyes are missing. If you are blind, you do not get a cool eyepatch or blindfold, you need your VISOR instead.

## Why's this needed? 
Seems fitting for characters that are missing eyes to have them covered in some way. Players don't have to keep them, of course, but spawning with them lets them make that choice instead of raiding medbay's fabricator.

## Changelog 

```changelog
(u)glowbold
(+)Characters with a missing eye get an eyepatch on spawn.
```

### DIFF
--- a/code/datums/part_customization.dm
+++ b/code/datums/part_customization.dm
@@ -6,6 +6,8 @@ ABSTRACT_TYPE(/datum/part_customization)
 	///Can be a type or a list of types to randomly pick from
 	var/part_type = null
 	var/base_64_cache = null
+	/// Associated trait ID, if one is needed
+	var/associated_trait_id = null
 	var/trait_cost = 0 //idk let's keep using trait points for now
 	///Cannot be added alongside these part IDs
 	var/incompatible_parts = list()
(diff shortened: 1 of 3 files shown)


code/datums/part_customization.dm
```diff
@@ -15,6 +17,8 @@ ABSTRACT_TYPE(/datum/part_customization)
 
 	///Check if we can, then apply the part
 	proc/try_apply(mob/M, list/custom_parts = null)
+		if (src.associated_trait_id && M.traitHolder)
+			M.traitHolder.addTrait(associated_trait_id)
```
TobleroneSwordfish, line 21: This should be after the `can_apply` check, either in `apply_to` (with `SHOULD_CALL_PARENT` set) or just above it inside the if statement.

TobleroneSwordfish (changes requested):

code/datums/part_customization.dm
pgmzeta: opted for the `SHOULD_CALL_PARENT` approach

PR #22799, opened 2025-03-31
Adapt briefing screens for Carnigorm and HTR Team minimaps

pgmzeta:
[feature][gamemodes]
## About the PR 
This PR allows admins to change the display of the atrium minimaps on the Carnigorm and HTR Team ship to instead show an office set. Everything said by mobs in the office set will be heard by the players on the Carnigorm/HTR Ship.

* Add two "offices" for the Syndicate and Nanotrasen to z2.
* Add hearing connector targets to the Carnigorm and HTR Team ship.
* Add unique "broadcast" map datums that connect and toggle viscontents
* Add admin verbs for toggling the broadcast on/off.

This is an adaptation of the briefing system used for 35 Below. Original code by Mr. Moriarty and Bartimeus, licensed CC BY-NC-SA 3.0 here: https://github.com/Herbstation/35below/blob/main/herb_modular/code/area/soviet_high_command.dm
https://github.com/Herbstation/35below/blob/main/herb_modular/code/modules/minimap/minimaps/soviet_map.dm

## Why's this needed? 
Enable admins to act out in-character briefings with these teams.

This feature could also be used in podwars, but would require adding the offices to the pod wars map itself as it doesn't include z2.

## Demo
(to be uploaded)

### DIFF
--- a/assets/maps/allocated/htr_team_ship.dmm
+++ b/assets/maps/allocated/htr_team_ship.dmm
@@ -329,6 +329,7 @@
 /area/space)
 "LB" = (
 /obj/machinery/light/small/floor/warm,
+/obj/hearing_connector/target/nanotrasen,
 /turf/unsimulated/floor/blue,
 /area/space)
 "NA" = (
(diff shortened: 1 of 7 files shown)


pgmzeta:
this will need to be reworked completely when the say rework is in, as well as to fix the emote issue as mentioned.

PR #22800, opened 2025-03-31
Remove req_access_txt from mapped (non-door) objects

JORJ949:
## About the PR 
Access spawner mapping helpers now affect /obj/storage/secure, for use in some azones that have special accesses on crates (e.g. owlery security room crate)

Add syndicate commander access spawner to set access on an announcement computer in the fermidzone

Removes req_access_txt from MOST (I definitely missed some) mapped (non-door) objects. Repetetive changes below:
- Syndicate crates (such as those found in some listening posts) now have a /syndicate type
- Pharmacy chemical lockers have a /medical type
- Secure Transfer crates (found on kondaru and nadir) have a /transfer type
- Hazard crates have a /hazard type
- Some armory equipment crates were set to HoS access, fix this
- Security file cabinets (the special detective sec computer that looks like files) found in HoS offices had their access overrides removed, the HoS has forensic access.
- access on carts completely removed, these aren't secure so I don't think the access does anything on them anyway
- Announcement computers with varedited access updated to the new /station subtype (this doesnt affect any non-req_access_txt edited announcement computers, so those that should be /station/bridge

## Why's this needed? 
req_access_txt is very messy and this is one step closer to removing it completely, all thats left after this is any mapped doors with it, and anything in secret.


ZeWaka (changes requested):
need the map replace tool thing, that's how all of this should've been done

JORJ949:
> need the map replace tool thing, that's how all of this should've been done

What do you mean by "map replace tool thing"? There's a tool that will replace everything with X var edited? I was only aware of the repath tool.

ZeWaka:
> > need the map replace tool thing, that's how all of this should've been done
> 
> What do you mean by "map replace tool thing"? There's a tool that will replace everything with X var edited? I was only aware of the repath tool.

Yes, most of your extra non-req_access changes are repaths. This is a very large PR with tons of changes that wasn't atomized.

JORJ949:
Atomise time

pgmzeta:
> > need the map replace tool thing, that's how all of this should've been done
> 
> What do you mean by "map replace tool thing"? There's a tool that will replace everything with X var edited? I was only aware of the repath tool.

https://github.com/goonstation/goonstation/tree/master/tools/UpdatePaths

PR #22803, opened 2025-03-31
Add artifact mining encounter

FlameArrow57:
[STATION SYSTEMS][FEATURE]

## About the PR 
Adds an artifact asteroid mining encounter. This is an asteroid that appears normal except for having a random number of artifacts, from 7 - 10, in it

## Why's this needed? 
More mining encounters

### DIFF
--- a/code/modules/mining/mining_encounters.dm
+++ b/code/modules/mining/mining_encounters.dm
@@ -293,6 +293,18 @@
 				new_crate.set_loc(pick(floors))
 				new_enemy.set_loc(pick(floors))
 
+/datum/mining_encounter/artifact
+	name = "Fluctuating Asteroid"
+	rarity_tier = 2
+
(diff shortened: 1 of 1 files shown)


code/modules/mining/mining_encounters.dm
```diff
@@ -904,4 +916,13 @@
 			continue
 		AST.set_event(E)
 
+/proc/Turfspawn_Asteroid_SeedArtifacts(list/turfs)
+	var/turf/simulated/wall/auto/asteroid/AST
+
+	shuffle_list(turfs)
+
+	for (var/i in 1 to min(length(turfs), rand(7, 10)))
```
TobleroneSwordfish, line 924: Maybe move this rand out of the proc as an argument to align with the ore version?

TobleroneSwordfish (approved):

PR #22806, opened 2025-03-31
Silver overdose turns you blue

RubberRats:
## About the PR 

This PR adds an overdose effect to the silver reagent that slowly turns your skin blue. It is also mildly toxic.

## Why's this needed? 

New dumb thing for people to mess around with.

## Changelog 

```changelog
(u)RubberRats
(+)Silver has been given a unique overdose effect
```

### DIFF
--- a/code/modules/chemistry/Reagents-Base.dm
+++ b/code/modules/chemistry/Reagents-Base.dm
@@ -499,6 +499,7 @@
 	transparency = 255
 	overdose = 30
 	taste = "metallic"
+	var/finaltone = rgb(108, 125, 183)
 
 	reaction_obj(var/obj/item/I, var/volume)
 		if (I.material && I.material.getID() == "silver")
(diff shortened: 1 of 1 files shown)


code/modules/chemistry/Reagents-Base.dm
```diff
@@ -518,6 +519,20 @@
...
+			var/mob/living/carbon/human/H = M
+			var/currenttone = H.bioHolder?.mobAppearance.s_tone
+			if(currenttone)
+				var/newtone = BlendRGB(currenttone, rgb(108, 125, 183), 0.02)
+				H.bioHolder.mobAppearance.s_tone = newtone
+				H.set_face_icon_dirty()
+				H.set_body_icon_dirty()
+				if (H.limbs)
+					H.limbs.reset_stone()
+				H.update_colorful_parts()
```
TobleroneSwordfish, line 534: This is quite an expensive proc to be calling on life loop, consider adding a condition to not call it if their skin is already blue.

TobleroneSwordfish (approved):
Cautious yes, although I will hit this with a hammer if it causes too much icon update load

PR #22813, opened 2025-03-31
Espresso variants for coffee drink recipes

HyenaShepherd:
## About the PR 
This change adds an espresso variant for any recipe with a pre-existing freshly-brewed variant.

## Why's this needed? 
Currently there are eight drink recipes that accept coffee and freshly-brewed coffee but not espresso. Bartenders have easy access to espresso through the espresso machine but need to buy Robust Coffee from a vendor or get coffee from other departments for those recipes.  

[catering] [qol]

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -1578,11 +1578,11 @@
 
 	cocktail_w_russian/fresh
 		id = "w_russian_fresh"
-		required_reagents = list("vodka" = 1, "coffee_fresh" = 1)
+		required_reagents = list("vodka" = 1, "coffee_fresh" = 1, "milk" = 1)
 
 	cocktail_w_russian/espresso
 		id = "w_russian_espresso"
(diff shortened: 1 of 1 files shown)


code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -1572,6 +1580,10 @@
 		id = "w_russian_fresh"
 		required_reagents = list("vodka" = 1, "coffee_fresh" = 1)
 
+	cocktail_w_russian/espresso
```
pgmzeta, line 1583: white russian should require milk? duplicates `cocktail_b_russian/espresso`

should also fix `cocktail_w_russian/fresh`, it does the same thing

pgmzeta (changes requested):
makes sense, just one thing i saw needing changes

code/modules/chemistry/Chemistry-Recipes.dm
HyenaShepherd: Fixed em both! Good catch.

pgmzeta (approved):

PR #22814, opened 2025-03-31
Job requisition function for RoleControl

RubberRats:
## About the PR 

This PR adds a new function to the RoleControl program, which allows additional job slots to be opened by spending credits from the payroll budget. Most of these are the random gimmick jobs, or other niche roles that the station might need.
TODO:
- [x] Clean up `request_limit` usage
- [x] Notify respawnable ghosts/new players that a role has opened up
- [x] Adjust list of requesitionable roles, and costs.
- [x] Highlight requisitioned roles/give incentive for them (if needed)
- [ ] Requisitioned role as ghost respawn? (probably a bad idea)

## Why's this needed? 

Gives the HoP responsibilities and things to do other than fulfilling access change requests, and an additional use for the station budget. Could open up new RP opportunities, such as hiring an exterminator to deal with a critter problem, or a bunch of lawyers for a court case. 

## Testing
Tested combinations of requested & priority roles, ensuring that latejoin menu highlighting and pda messaging worked as expected.

#### PDA Messaging
image

#### Before & After of new player menu when a requested role is filled
Screenshot 2025-06-11 163229

Screenshot 2025-06-11 163249

## Changelog 

```changelog
(u)RubberRats & glowbold
(*)The RoleControl program has a Request function, allowing players to open up additional job slots using surplus payroll budget.
(+)Requested jobs are highlighted with a yellow border like priority jobs in the latejoin role selection menu
(+)When someone joins a RoleControl prioritized or requested role, command gets a PDA message
```

### DIFF
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -1478,7 +1478,7 @@ ABSTRACT_TYPE(/datum/job/special/random)
 	name = "Inspector"
 	wages = PAY_IMPORTANT
 	linkcolor = NANOTRASEN_LINK_COLOR
-	request_cost = PAY_DONTBUYIT * 4 // gotta dig real deep for that second(?) opinion
+	request_cost = PAY_EXECUTIVE * 4
 	access_string = "Inspector"
 	receives_miranda = TRUE
 	invalid_antagonist_roles = list(ROLE_HEAD_REVOLUTIONARY)


RubberRats:
Now notifies respawn-eligible players and anyone waiting in the lobby. Should have most of the features I want for now, just need feedback on job counts/cost

TDHooligan:
poking this because i think it's a sick idea to expand upon. 
some of the roundstart jobs are weaksauce because they'd otherwise be constantly latejoined for. this would lower the reliability of getting said roles, and give them expectations. meaning their gear can be more justifiably strong/unique.

ZeWaka (approved):

code/modules/networks/computer3/job_controls.dm
```diff
@@ -25,57 +34,140 @@ var/datum/job/priority_job = null
...
+					src.print_text("Cancelling job requisition...")
+					requested_job = null
+					state = MENU_MAIN
+				else if(text2num(command))
+					request_count = clamp(text2num(command), 0, requested_job.request_limit - requested_job.limit)
+					if(request_count == 0)
+						src.print_text("Cancelling job requisition...")
+						requested_job = null
+						state = MENU_MAIN
+					src.print_text("This will deduct [requested_job.request_cost * request_count][CREDIT_SIGN] from the payroll budget. Current payroll budget: [global.wagesystem.station_budget][CREDIT_SIGN]")
```
pgmzeta, line 143: request cost could be null here if e.g. the job datum request count is edited mid-round but the cost isn't set

maybe add an explicit check earlier or default to 0?

code/mob/new_player.dm
```diff
@@ -300,6 +300,17 @@ TYPEINFO(/mob/new_player)
 				global.latespawning.unlock()
 				return
 			JOB.assigned++
+			if (JOB.player_requested || JOB == job_controls.priority_job)
+				SPAWN(0) // don't pause late spawning for this
+					var/limit_reached = JOB.limit == JOB.assigned
```
JORJ949:
```suggestion
					var/limit_reached = JOB.limit <= JOB.assigned
```
Limit could be less than assigned if for example an admin makes a respawn wave of them

code/mob/new_player.dm
```diff
@@ -300,6 +300,17 @@ TYPEINFO(/mob/new_player)
...
+					if (JOB.player_requested && limit_reached)
+						JOB.player_requested = FALSE
+					if (JOB == job_controls.priority_job && limit_reached)
+						job_controls.priority_job = null
+					var/message = "RoleControl notification:[JOB.player_requested ? " requested":""][JOB == job_controls.priority_job ? " priority": ""] job [JOB.name] hired[limit_reached ? " (limit reached)" : ""]."
```
JORJ949: Should be a "MAILBOT" for consistency?

JORJ949 (changes requested):
Wasn't required at the time but this PR is rather large so could probably use a testing report.

code/mob/new_player.dm
pgmzeta: `"sender_name"="COMMAND-MAILBOT"`  in the actual signal data below

code/datums/jobs.dm
```diff
@@ -99,6 +99,10 @@
 	///if true, cryoing won't free up slots, only ghosting will
 	///basically there should never be two of these
 	var/unique = FALSE
+	var/request_limit = 0 //!Maximum limit that can be reached by requisitioning through RoleControl
```
TobleroneSwordfish: Is this the total maximum of the job (initial + requested) or just the additional amount you can request? Comment should reflect this.

code/modules/networks/computer3/job_controls.dm
```diff
@@ -25,57 +34,147 @@ var/datum/job/priority_job = null
...
+					src.print_text("Error: no job selected")
+					state = MENU_MAIN
 					return
-				src.print_text(src.job_info(job))
+				if(command == "x")
+					src.print_text("Cancelling job requisition...")
+					requested_job = null
+					state = MENU_MAIN
+				else if(text2num(command))
+					request_count = clamp(text2num(command), 0, requested_job.request_limit - requested_job.limit)
```
TobleroneSwordfish: This should ideally error if you request more than the maximum amount, rather than silently clamping

JORJ949:
> The info command on rolecontrol reports jobs with a zero limit as unknown, even if they can be requested.

Something something more job categories for special jobs (latejoin v antag v admin summoned)

pgmzeta:
people using RoleControl can now see all staple and special roles via list.

added a bunch of QoL stuff I got frustrated on by testing

Clown slot is cheaper, and Mime also now gets a second bonus slot (at double the price of clown, because they get paid twice as much)

naming has been normalized to "request" from "requisition"

jobs can only be set to priority if they have slots open, now

code/datums/jobs.dm
```diff
@@ -1467,6 +1478,7 @@ ABSTRACT_TYPE(/datum/job/special/random)
 	name = "Inspector"
 	wages = PAY_IMPORTANT
 	linkcolor = NANOTRASEN_LINK_COLOR
+	request_cost = PAY_DONTBUYIT * 4 // gotta dig real deep for that second(?) opinion
```
TobleroneSwordfish: 100,000 is a craaazy amount of credits. This won't always be a second opinion, inspector isn't a guaranteed role.

PR #22819, opened 2025-04-01
Increase space phoenix damage against silicons

FlameArrow57:
[GAME MODES][BALANCE]

## About the PR 
Title

## Why's this needed? 
Feathers are a bit underpowered against silicons right now


pgmzeta:
A bit more damage doesn't really stop silicons from stamlocking you with an omnitool. I think applying a slowdown via a status effect that controls a movement modifier is more in line with the cold/ice theme of the space phoenix and fits their hit and run playstyle better.

FlameArrow57:
kk, added movement modifier

TobleroneSwordfish (approved):
This projectile datum is rapidly becoming a pile of technical debt but I do see the reasoning for it

PR #22828, opened 2025-04-02
Fix being unable to replace space phoenix ice tunnels

FlameArrow57:
[GAME MODES][BUG]

## About the PR 
Title

## Why's this needed? 
bug fix

## Changelog 

```changelog
(u)FlameArrow57
(+)Fixed being unable to build walls over or using an RCD on Space Phoenix ice tunnels
```

### DIFF
--- a/code/obj/item/building_materials.dm
+++ b/code/obj/item/building_materials.dm
@@ -419,7 +419,7 @@ MATERIAL
 					var/turf/T = get_turf(usr)
 					var/area/A = get_area (usr)
 
-					if (!istype(T, /turf/simulated/floor))
+					if (!(istype(T, /turf/simulated/floor) || istype(T, /turf/simulated/space_phoenix_ice_tunnel)))
 						boutput(usr, SPAN_ALERT("You can't build girders here."))
 						return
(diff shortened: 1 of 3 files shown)


FlameArrow57:
No reason in particular, just happened to go with it. Think it would maybe be better to prevent any possible interaction with floor code elsewhere

code/obj/item/building_materials.dm
```diff
@@ -419,7 +419,7 @@ MATERIAL
 					var/turf/T = get_turf(usr)
 					var/area/A = get_area (usr)
 
-					if (!istype(T, /turf/simulated/floor))
+					if (!(istype(T, /turf/simulated/floor) || istype(T, /turf/simulated/space_phoenix_ice_tunnel)))
```
ZeWaka, line 422: Why is this not a normal floor anyways

FlameArrow57: See above comment to Glowbold: "No reason in particular, just happened to go with it. Think it would maybe be better to prevent any possible interaction with floor code elsewhere"

PR #22839, opened 2025-04-03
Drone_Factory_Repairs

Loaded4Bears:
## About the PR 
Nanotransen repair crews have performed long-deferred maintenance and the Cogmap 2 drone factory now has a fully functional powernet including APC, solar tracker and SMES.

## Why's this needed? 
Everything works just as well with the existing solar panels wired and working as it did with space magic power.

## Changelog 

```changelog
(u)Loaded4Bears
(*)Drone factory on Cogmap 2 has a fully functional powernet.
(*)Inelegantly but functionally wired all existing solar panels. 
(+)Replaced fakeobj networked computer stuff with a couple emergency closets
```


Loaded4Bears:
Adding an APC and nothing else to the other maps so it passes tests. If this PR is approved I can copy-paste Cog2's drone factory across a couple other maps with the exact same layout, or rework their solars, or if they don't HAVE solars like Oshan simply add very minimal lighting and not change anything else.

JORJ949:
Changelog should be more specific like "Cogmap 2 drone factory now has a functional powernet instead of a fake one" if it has one at all.

maps/nadir.dmm
```diff
@@ -44464,6 +44464,13 @@
 /area/station/engine/substation/pylon{
 	name = "South Catalytic Substation"
 	})
+"sQt" = (
+/obj/machinery/power/apc/autoname_east,
```
pgmzeta, line 44468: how will nadir's work without solars to keep it charged?

Loaded4Bears:
> Changelog should be more specific like "Cogmap 2 drone factory now has a functional powernet instead of a fake one" if it has one at all.

added.

pgmzeta (changes requested):
The cable network to the solar panels needs to be wired to the SMES intake, with the station-side network hooked up under the SMES unit. Here's a screenshot what that wiring setup could look like:
imagesss

I think the solar panels should be wired up as though the missing ones were there, but broke at some point. That would let someone rebuild/refurbish it a bit easier.

When I tested this in-game, the panels didn't track properly. You need to use the same ID on `/obj/machinery/power/tracker` as the `/obj/machinery/computer/solar_control` computer (i.e. make the same sub-type you did to the solar control computer and replace the tracker with that.

---

Personally not a fan of yet another off-station place for Arcfiends to get free power all round, but want to hear more opinions.

JORJ949:
> 
> Personally not a fan of yet another off-station place for Arcfiends to get free power all round, but want to hear more opinions.

I think rather than giving offstation areas fake magic powernets just add a check to arc code to check if you are on station when consuming power to fix this problem.

PR #22845, opened 2025-04-03
Spy sticker QoL

itchybones:
[GAME OBJECTS][QOL]
### Internal Changes
This PR replaces the `afterattack `proc override in the spy sticker code with a proc override of the `stick_to `proc, in order to make the activation of spy stickers more consistent (for example, the code in the `afterattack `proc wouldn't be called when throwing stickers, even though throwing can also cause the stickers to stick on something). 

### User-facing Changes
Spy stickers now auto-activate their radio (the `broadcast `var, which boradcasts to the freq)when stuck to an atom. Additionally, help messages are added to clarify the spy sticker's functions and how to access them (radio & camera settings).

## Why's this needed? 
Spy stickers are a bit of a mess at the moment. This PR aims to improve QoL by making spy sticker behaviour more consistent.

## Changelog 

```changelog
(u)Chasu
(+)Spy stickers now auto-activate their radio when stuck to something
(+)Spy stickers have recieved help messages
```


pgmzeta:
I feel like them automatically broadcasting limits their usage, since you can't enable the speaker without it picking up all other spy stickers. 

It's a consistent sticking point with spy stickers, but perhaps they should just default to broadcast on rather than always broadcasting on placement?

itchybones:
> I feel like them automatically broadcasting limits their usage, since you can't enable the speaker without it picking up all other spy stickers.
> 
> It's a consistent sticking point with spy stickers, but perhaps they should just default to broadcast on rather than always broadcasting on placement?

I _think_ I get what you mean, but if they would default to broadcast, they should probably start on another frequency too, since they currently start on the default freq, leading to an easy noob trap: _**"145.9 [sticker]: Evilguy McTraitor says:** "I just bought these spy stickers, how cool!"**"x5**_

Also, if someone were to enable the speaker, it'd pick up the other spy stickers regardless, no? Maybe I'm misunderstanding here, the vars are named a little weirdly, but even currently, turning on the speaker broadcasts anything said on the frequency.

PR #17553, opened 2024-01-03
Adds new candies/treats, brownies, and refactors existing wrapped candy code

itsloley:
[CATERING][CODE QUALITY][FEATURE][ADD TO WIKI]

## About the PR 

Adds a lot of melted sugar based recipes to the kitchen!
- Sugar can be melted by putting sugar and a tray in the oven for 5s on high or 10s on low. Reagents can be poured into the tray of sugar!
- The tray can be cut into strips, which can be put on rods or popsicle sticks to be turned into rock candy.
- Melted sugar strips can be cut into individual hard candy pieces, which can be wrapped in paper for presentation!
- Hard candy pieces can be put on a rod or popsicle stick to make small lollipops, similar to the ones found in medbay.
- Melted sugar strips can be used in-hand to make them into a circle which can be put on a rod or popsicle stick to make a swirly lollipop.
- Putting flour onto a melted sugar circle allows it to be folded in-hand, creating dragon's beard candy.
- Dragon's beard candy gets better the more times you fold it! It has a chance to break, but your chances are better if you're a skilled chef and apply flour before each fold.
- Dragon's beard candy can be cut into individual pieces.
- Also adds brownies, made with sweet dough, 2 eggs and a chocolate bar in the mixer, then putting the brownie batter in the oven for 7 seconds on high.
- Refactors wrapped candy code, combining candies that had a wrapper before into one parent type

## Why's this needed? 

Adds a lot more candy recipes for the chef to play with! Hand out comedically oversized lollipops to the crew, swap medbay's lollipops for dubious homemade ones, or just inject brownies with THC.

yippee

## Changelog 

```changelog
(u)Lyy, Beshemoth
(*)Added melted sugar cooking recipes, including rock candy, hard candy, craftable lollipops, swirly lollipops, and dragon's beard candy!
```

PR #22848, opened 2025-04-03
Colorable Hard Candy Wrappers

itsloley:
## About the PR 

adds the ability to color hard candy wrappers with a pen/pencil/marker/crayon/etc

dreamseeker_egJ333t9QX

## Why's this needed? 

makes candy more colorful instead of a basic white wrapper! this is something we wanted to do originally in #17553 but forgot about

## Changelog 

```changelog
(u)Lyy
(*)You can now color homemade hard candy wrappers with markers or crayons!
```

### DIFF
--- a/code/modules/food_and_drink/candy.dm
+++ b/code/modules/food_and_drink/candy.dm
@@ -620,6 +620,7 @@ ABSTRACT_TYPE(/obj/item/reagent_containers/food/snacks/candy/wrapped_candy/taffy
 	real_name = "hard candy"
 	icon_state = "hardcandy"
 	var/flavor_name
+	var/wrapper_color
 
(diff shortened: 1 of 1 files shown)


itsloley:
> That ws already possible with paint cans, wasn't it?

this would also recolor the candy inside i believe

itsloley:
huh, interesting! in any case, that was definitely unintentional :p we always wanted to be able to color them like this in the original PR, we just kind of completely forgot to finish it

TobleroneSwordfish (changes requested):
Minor but it bugs me that they're so oversaturated, the painted ones look better. You could steal/generalize the paint can code that generates a desaturated colour from the true font colour.

PR #22868, opened 2025-04-04
Common unbreak_all_organs proc on organHolder

pgmzeta:
[internal][code quality]
## About the PR 
Add duplicated code in admin heal, ling regen, and gang janktank II for unbreaking all organs into a single proc call in organHolder.

## Why's this needed? 
Less duplicate code across the codebase. Atomized from another upcoming change that would otherwise copy/paste this code.


ZeWaka (approved):
unbreak_all_organs_complitely

PR #22869, opened 2025-04-04
Admin "Stablize" command

pgmzeta:
[admin]
## About the PR 
Add a "Stablize" command to admin mob interaction and player options window. This proc attempts to restore people to just above crit levels.

I'd like to use the unbreak_all_organs proc in #22868 here, DNM until that's in place and I can uncomment the call in place here

## Why's this needed? 
Sometimes you want to keep someone alive but not necessarily immediately able to run off/fight back.

## Screenshots
Screenshot 2025-04-09 133825
Screenshot 2025-04-09 133832


Sord213:
Consider adding it to the admin interact menu ( ` + click on someone) as well if you didn't.

PR #22871, opened 2025-04-05
Arcfiend Sap Power/Discharge Rework

pgmzeta:
[experimental][gamemodes][rework]
## About the PR 
Introduces some changes to Arcfiend:
* Sap Power changes
  * Default rate for APC/SMES drain is 15, from 30
  * Default rate of Machinery drain is 10, from 25
  * Broken Machines, APCs, and SMES units now give you double charge.
  * Default rate of Human/Cyborg drain is 40 (no change)
  * No longer randomly causes an APC to break when sapping power
* Discharge changes
  * Discharge can be used on machines, including APCs and SMES units, to overload them.
  * Overloading will generally break machines, but may cause other effects, for example:
    * Cameras trigger their EMP effect
    * Weapon racks get hacked 
    * Lights pop

These changes are reflected in updated tooltips and content popups.

## Why's this needed? 
Incentivize Arcfiends to have a more direct impact on the round in a few ways:
* Human/cyborg drain rate is still the fastest way to get points, and requires no "setup"
* Breaking things on station lets you drink power faster, which should lead to more interaction with crew members having to fix or work around broken gear.
* Breaking SMES units and APCs stop them from recharging, so breaking them is a trade-off.
* The default rate still lets you get some power, albeit much slower, for Arcfiends to get started.

## Changelog 

```changelog
(u)glowbold
(*)Arcfiend Sap Power and Discharge have been reworked. More details in minor notes.
(+)Sap Power pulls from APCs, SMESs, and Machines at roughly half the current rate.
(+)Sapping power from a broken machine/APC/SMES doubles the amount of power you receive.
(+)Discharge can target machines, including APCs and SMESs, to overload and break them.
```


code/modules/power/smes.dm
```diff
@@ -15,6 +15,9 @@ TYPEINFO(/obj/machinery/power/smes/magical)
 		charge = INFINITY
 		..()
 
+	set_broken()
```
TobleroneSwordfish, line 18: Might want to make sure the `hardened` APCs in listening posts etc can't break too

TobleroneSwordfish (approved):
I like this, the ability to break machinery at will is a cool thematic thing for arcfiends too.

pgmzeta:
I'm gonna have to review all `/obj/machinery/` subtypes to make sure they're either fixable or unable to be broken before this is really good to go :(

code/modules/power/smes.dm
pgmzeta: atomized to https://github.com/goonstation/goonstation/pull/22894

PR #22872, opened 2025-04-05
Snowprints 🦶❄️

Azrun:
## About the PR 
Stepping on snow turfs creates tracks/footprints.

## Why's this needed? 
Beautiful effect.
Ambiance.
Is-Cool

## Changelog 

```changelog
(u)glowbold
(+)Footprints in snow.
```

### DIFF
--- a/code/turf/floors.dm
+++ b/code/turf/floors.dm
@@ -1549,7 +1549,7 @@ TYPEINFO(/turf/simulated/floor/grass)
 	.=0
 
 /turf/proc/snow_prints(atom/movable/AM)
-	if (isliving(AM))
+	if (isliving(AM) && !HAS_ATOM_PROPERTY(AM, PROP_ATOM_FLOATING))
 		var/mob/living/M = AM
 		if (M.lying)
 			var/obj/effect/snow_step/dragged = new(src)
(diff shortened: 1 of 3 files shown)


code/turf/floors.dm
```diff
@@ -1582,6 +1645,35 @@ TYPEINFO(/turf/simulated/floor/grass)
...
+		return
+
+	proc/weather_fade()
+		var/weather_severity = 0
+		if (isnull(weather_severity) || weather_severity < 1)
+			return
+		// sev0 = 10 seconds
+		// sev1 = 7 seconds
+		// sev2 = 4 seconds
+		// sev3 = 1 second
```
pgmzeta, line 1669: The weather severity values were setup for 35 Below's bespoke weather system, which was dynamic in round.

If you wanted to keep similar variety here, you could have terrainify change a (global?) variable based on the weather settings for the winterify/ice moon terranify subtypes.

It might mess with the footsteps on the ice moon azone if you did that, but it's probably not going to be noticeable when terrainify is run.

Here's where I would put the values we used, but you should probably just have it set specific number of seconds instead of doing a calculation

sev0 was clear skies (Ice Moon Snowing: No; Winterify Snow: None)
sev1 was light snow (Winterify Snow: Light Snow)
sev2 was heavy snow (Ice Moon Snowing: Yes; Winterify Snow: Snow)
sev3 was blizzard conditions (Ice Moon Snowing: Particles)


alternatively you can scrap that aspect of this and just have it a flat value

code/turf/floors.dm
```diff
@@ -1544,6 +1548,65 @@ TYPEINFO(/turf/simulated/floor/grass)
 /turf/proc/grassify()
 	.=0
 
+/turf/proc/snow_prints(atom/movable/AM)
+	if (isliving(AM))
+		var/mob/living/M = AM
```
pgmzeta, line 1553: recommend inverting the logic here to return early instead, and also return early if the mob is floating:

```
	if (!isliving(AM))
		return
	if (HAS_ATOM_PROPERTY(AM, PROP_ATOM_FLOATING))
		return
	var/mob/living/M = AM
	...
```

code/turf/floors.dm
Azrun: Annotated intended use as a future todo to be integrated with either events or parallax lookups to determine "weather".

code/turf/floors.dm
Azrun: Added floating check.

PR #22873, opened 2025-04-05
Unit Test: Terrainify sanity check to try to limit simulated turfs

Azrun:
## About the PR 
Add basic checks to ensure `/datum/biome` and `/datum/map_generator` generate non-simulated turfs or turfs that will be destroyed into non-simulated turfs.

## Why's this needed? 
Stop @Azrun for indirectly making rounds bad.


DisturbHerb:
hear me out what if they were all simulated

PR #22874, opened 2025-04-05
Add additional labelling for Artifact Lab's Mass Drivers on Donut2.

Atraxite:
## About the PR 
Minor change adding simple floor markings to denote and accentuate the Artlab Mass Drivers.
Clearly marking which of the two is supposed to be used for shipment of Artifacts or otherwise,
to QM.

## Why's this needed? 
A lack of floor-based visual delineation, in regard to the function of the two Mass Drivers.
Such a deficiency may lead to potentially ruinous outcomes should the incorrect Mass Driver
be used for the ejection of Artifacts that pose an immediate threat.

## Changelog 

```changelog
(u)Atra_xite
(+)Floor Markings to denote Mass Driver intended use in Donut2 Artlab.
```

### DIFF
--- a/maps/donut2.dmm
+++ b/maps/donut2.dmm
@@ -8351,6 +8351,10 @@
 	dir = 1
 	},
 /area/station/crew_quarters/stockex)
+"aSj" = (
+/obj/decal/stripe_caution,
+/turf/simulated/floor,
+/area/station/science/artifact)
 "aSm" = (
 /obj/decal/tile_edge/stripe/extra_big{
 	dir = 4
@@ -16907,7 +16911,11 @@
 	pixel_y = 24
 	},
 /obj/decal/tile_edge/stripe/corner/extra_big2{
-	dir = 9
+	dir = 10
+	},
+/obj/decal/tile_edge/stripe/extra_big{
+	dir = 8;
+	tag = "icon-xtra_bigstripe-edge (WEST)"
 	},
 /turf/simulated/floor/white/checker2{
 	dir = 8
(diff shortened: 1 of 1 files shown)


pgmzeta:
Could you consider also adding the floor decals like those on Cogmap2 (not that it always saves that from disaster, but):
image
The path for the decal is `/obj/decal/tile_edge/floorguide/evac`

Atraxite (commented):
> Could you consider also adding the floor decals like those on Cogmap2 (not that it always saves that from disaster, but):
> image
> The path for the decal is `/obj/decal/tile_edge/floorguide/evac`
>  

Unless i've made an error somewhere along the process of adding the aforementioned floor guides, as i'm quite inexperienced in using VScode, this has been implemented into the PR.

TobleroneSwordfish:
But I like ruinous outcomes 🥺

Sord213:
Idk. There's big obvious text right above the button you have to push. Plus, workplace accidents "oh god oh fuck I sent an active bomb to QM" are funny.

Atraxite:
> Idk. There's big obvious text right above the button you have to push. Plus, workplace accidents "oh god oh fuck I sent an active bomb to QM" are funny.

While I'm not disputing the fact that the comedic potential in an unintentional shipment of an actively hazardous,
or soon-to-be hazardous Artifact is sizable, I would simply like to point out that this PR would, at least to some degree,
bring Donut2's Artlab into some version of design parity with a map such as Cogmap2 in regard to adjacent Mass Driver labelling.

TobleroneSwordfish:
I'm going to say the poorly labelled art chutes are *soul* and part of the map's character. I know usually we go for consistency but I think this would be a shame to fix.

PR #22882, opened 2025-04-06
Convert all mutantrace emotes to use `emote_check`

pgmzeta:
[bug][mutrantraces]
## About the PR 
* Replace all checks of `emote_allowed` to use the `emote_check` proc instead.
* Use SECOND defines to make it clear how long cooldowns should be
* Add doccomments for `/datum/mutantrace/emote`

## Why's this needed? 
SPAWNS can be unreliable, letting people spam sound emotes.

Cleaner code flow that matches the pattern in main emote code.


ZeWaka (changes requested):
im scared of say rework touching this

PR #22883, opened 2025-04-06
Yoink the HoP's default sec accesses

pgmzeta:
[add to wiki][balance][station systems]
## About the PR 
Removes the HoP's default security/carry/contra/foreniscs access.

## Why's this needed? 
They don't have security comms access and hold no real direct authority over the rest of security (until they're made acting captain). They keep ticketing access.

HoP's can still, of course, modify their ID to have these permissions at round start.

## Changelog 

```changelog
(u)glowbold
(*)The Head of Personnel does not have security access by default.
```

### DIFF
--- a/code/procs/access.dm
+++ b/code/procs/access.dm
@@ -204,8 +204,7 @@
 		if("Captain")
 			return get_all_accesses()
 		if("Head of Personnel")
-			return list(access_security, access_carrypermit, access_contrabandpermit, access_forensics_lockers, access_ticket,
-						access_tox, access_tox_storage, access_chemistry, access_medical, access_medlab,
+			return list(access_ticket, access_tox, access_tox_storage, access_chemistry, access_medical, access_medlab,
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
I think I disagree with this, the HoP is still second in command under the captain and should have the authority to boss security around if there isn't an HoS around.

pgmzeta:
They have bankboss to edit manifest records with now

snowyseshi:
I also disagree with this. The HoP is considered the first mate of the captain they should be allowed to go where the captain goes. They also have the authority to execute crewmembers, override sentences made by the arresting officer and make sure the security officers are following space law.

JORJ949:
HoP should be able to do acting captain stuff immediately without having to have powergamed AA at roundstart, I would honestly look into giving the HoP more access elsewhere so they can fill in for other command jobs. HoP is also still the boss of security, even if there is a HoS (though the HoS can overrule the HoP), they just don't have sec comms to discourage sechopping.

PR #22888, opened 2025-04-06
Add Intruder game mode

FlameArrow57:
[GAME MODES][FEATURE]

## About the PR 
Adds the Intruder game mode per below
Design doc - https://hackmd.io/@FlameArrow57/ryuQFOQ-T
Player Project - https://discord.com/channels/182249960895545344/1358207672914280563

## Why's this needed? 
Would be cool

intruder

## Changelog 

```changelog
(u)FlameArrow57
(*)TBD
```


code/datums/abilities/critter/mindeater.dm
```diff
@@ -0,0 +1,390 @@
...
+				boutput(src.holder.owner, SPAN_ALERT("This target has received too much brain damage!"))
+				return CAST_ATTEMPT_FAIL_NO_COOLDOWN
+		if (isdead(L))
+			boutput(src.holder.owner, SPAN_ALERT("You can only use this ability on alive targets!"))
+			return CAST_ATTEMPT_FAIL_NO_COOLDOWN
+		return ..()
+
+	cast(atom/target)
+		. = ..()
+		actions.start(new /datum/action/bar/private/mindeater_brain_drain(target), src.holder.owner)
```
TobleroneSwordfish, line 107: This doesn't require line of sight to be maintained and gives no feedback to the victim, meaning it seems to be possible to 100-0 brain drain someone while fully invisible and hiding behind a wall.

code/modules/projectiles/special.dm
```diff
@@ -1400,3 +1400,42 @@ ABSTRACT_TYPE(/datum/projectile/special)
...
+			return
+		if (istype(L, /mob/living/critter/mindeater))
+			return
+		L.changeStatus("staggered", 0.75 SECONDS)
+		L.setStatus("psi_bolt_slow", 5 SECONDS)
+		var/datum/reagents/reagents = L.reagents
+		if (reagents)
+			var/amt = min(L.reagents.total_volume, 5)
+			L.reagents.remove_any_except(amt, "toxin")
+			L.reagents.add_reagent("toxin", amt)
```
TobleroneSwordfish, line 1435: This doesn't seem to be working?

code/modules/projectiles/special.dm
```diff
@@ -1400,3 +1400,42 @@ ABSTRACT_TYPE(/datum/projectile/special)
...
+
+	on_hit(atom/hit, angle, obj/projectile/O)
+		. = ..()
+		var/mob/living/L = hit
+		if (!istype(L))
+			return
+		if (istype(L, /mob/living/critter/mindeater))
+			return
+		L.changeStatus("staggered", 0.75 SECONDS)
+		L.setStatus("psi_bolt_slow", 5 SECONDS)
```
TobleroneSwordfish, line 1430: This slow is almost unnoticeable, especially with sprint.

code/datums/abilities/critter/mindeater.dm
```diff
@@ -0,0 +1,390 @@
...
+		. = ..()
+		var/mob/living/critter/mindeater/mindeater = src.holder.owner
+		mindeater.undisguise()
+
+		mindeater.abilityHolder.removeAbility(/datum/targetable/critter/mindeater/clear_disguise)
+		mindeater.abilityHolder.addAbility(/datum/targetable/critter/mindeater/disguise)
+
+/datum/targetable/critter/mindeater/confuse
+
+/datum/targetable/critter/mindeater/shades
```
TobleroneSwordfish, line 260: This is a really cool ability but movement (especially diagonal?) seems to immediately reveal which is the real one because the others don't animate properly.

code/datums/abilities/critter/mindeater.dm
```diff
@@ -0,0 +1,390 @@
...
+
+/datum/targetable/critter/mindeater/clear_disguise
+	name = "Clear Disguise"
+	desc = "Clear your disguise."
+	icon_state = "clear_disguise"
+
+	cast(atom/target)
+		. = ..()
+		var/mob/living/critter/mindeater/mindeater = src.holder.owner
+		mindeater.undisguise()
```
TobleroneSwordfish, line 253: Being undisguised by shooting doesn't revert this ability.

TobleroneSwordfish (changes requested):

code/datums/abilities/critter/mindeater.dm
FlameArrow57: I was planning to add more feedback in the next stage of mindeaters if all goes well with the initial playtest, I'll add some more "temporary" feedback rn

code/modules/projectiles/special.dm
FlameArrow57: Was bugged, LMK if you think it needs to be increased

code/datums/abilities/critter/mindeater.dm
FlameArrow57: Any suggestions on how to improve? I tried a few things and manually calling Move() seemed to be the best. Could maybe replace /obj/dummy/fake_mindeater with a critter version and have movement_delay() be set to the parent mindeater's movement delay, but maybe better post-initial-playtest?

code/modules/projectiles/special.dm
FlameArrow57: Bug on master, being fixed in #22956

FlameArrow57: Okay fixed now

TobleroneSwordfish (changes requested):
- Pierce the veil goes on cooldown when interrupted or used on a corpse, probably not intentional?
- Disguise still uses the wrong ability icon when cancelled
- Disguise instantiates an entire new human for every disguise and can be spammed, since humans tend not to GC this is a potentially server crashing memory leak if someone autoclickers it. Consider keeping one dummy human and re-rolling their appearance repeatedly.
- Disguise shows your voice as `mindeater (as Unknown)`, probably should use your disguised name?
- Paralyze goes on full cooldown when the target is looking away, potentially quite frustrating? I have my doubts about using facing direction in SS13 in general but it might be neat to try out.
- Pierce the veil needs to hold a reference to the allocated region (probably on the status effect?) or it will be marked to be cleared and overwritten by other veil regions or misc allocated zones.

FlameArrow57:
> * Pierce the veil goes on cooldown when interrupted or used on a corpse, probably not intentional?

Can you clarify when interrupted? It's meant to go on cooldown if using another ability during it or if your 3 shield charges run out. Fixed using on a corpse

> * Disguise still uses the wrong ability icon when cancelled

Fixed

> * Disguise instantiates an entire new human for every disguise and can be spammed, since humans tend not to GC this is a potentially server crashing memory leak if someone autoclickers it. Consider keeping one dummy human and re-rolling their appearance repeatedly.

Done

> * Disguise shows your voice as `mindeater (as Unknown)`, probably should use your disguised name?

Fixed

> * Paralyze goes on full cooldown when the target is looking away, potentially quite frustrating? I have my doubts about using facing direction in SS13 in general but it might be neat to try out.

Maybe, it's meant to be intentional. People can change their direction to try to bait it out since it is more of a powerful ability, and you can also use Spatial Swap followed up by Paralyze to guarantee it works if the target is running away

> * Pierce the veil needs to hold a reference to the allocated region (probably on the status effect?) or it will be marked to be cleared and overwritten by other veil regions or misc allocated zones.

I think it is referenced? On the "mindeater_abducted" status effect, var "alloc_region"

code/mob/living/critter/mindeater.dm
```diff
@@ -0,0 +1,613 @@
...
+
+		src.hp_indicator = new (loc = src)
+
+		src.demanifest()
+
+		get_image_group(CLIENT_IMAGE_GROUP_INTRUSION_OVERLAYS).add_mob(src)
+		get_image_group(CLIENT_IMAGE_GROUP_MINDEATER_STRUCTURE_VISION).add_mob(src)
+
+		src.ensure_speech_tree().AddSpeechOutput(SPEECH_OUTPUT_INTRUDERCHAT)
+		src.ensure_listen_tree().AddListenInput(LISTEN_INPUT_INTRUDERCHAT)
```
Mister-Moriarty: `TYPEINFO` these please. See `blob_overmind.dm` for a template.

code/mob/living/critter/mindeater.dm
```diff
@@ -0,0 +1,613 @@
...
+
+	disposing()
+		..()
+		get_image_group(CLIENT_IMAGE_GROUP_INTRUSION_OVERLAYS).remove_mob(src)
+		get_image_group(CLIENT_IMAGE_GROUP_MINDEATER_STRUCTURE_VISION).remove_mob(src)
+		QDEL_NULL(src.vis_indicator)
+		QDEL_NULL(src.hp_indicator)
+
+		src.ensure_speech_tree().RemoveSpeechOutput(SPEECH_OUTPUT_INTRUDERCHAT)
+		src.ensure_listen_tree().RemoveListenInput(LISTEN_INPUT_INTRUDERCHAT)
```
Mister-Moriarty: As above.

PR #22908, opened 2025-04-08
Radioactive clothing and items makes you glow

Nexusuxen:
[FEATURE][GAME OBJECTS]

## About the PR 

Changes `component/radioactive` to use `component/loctargeting/simple_light` instead of `add_simple_light`, resulting in the simple_light being visible on a mob holding or wearing something radioactive.

(Note: Items in your pocket will have their radiation glow shine through, though not in your backpack or belt. This is also a little weird but is overall cleaner behavior)

## Why's this needed? 

I wish to glow obnoxiously. See below, I do not currently glow when dripped out in cerenkite.
image

I now glow when wearing an irresponsible amount of radioactive material.
image

It's also just weird that something loses its very noticeable glow when you pick it up, when conceivably it would still be visibly glowing. This also fixes the (admittedly funny) bug where radioactive items would glow in your inventory, like so:
image
image
1st image = before, 2nd image = after

## Changelog 

```changelog
(u)Nexusuxen
(+)The glow from radioactive stuff will still shine around you even when you equip or pick them up.
```

### DIFF
--- a/code/datums/components/radioactive.dm
+++ b/code/datums/components/radioactive.dm
@@ -23,6 +23,8 @@ TYPEINFO(/datum/component/radioactive)
 	var/_backup_color = null //so hacky
 	/// Internal, store of turf glow overlay
 	var/static/image/_turf_glow = null
+	/// Internal, reference to light component
+	var/datum/component/loctargeting/simple_light/our_light
 
 	Initialize(radStrength=100, decays=FALSE, neutron=FALSE, effectRange=1)
 		if(!istype(parent,/atom) || parent.type == /turf/space) //exact type check to exclude ocean floors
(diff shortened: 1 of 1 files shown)


DisturbHerb:
i just went swimming and now i glow ^_^

ZeWaka (approved):

code/datums/components/radioactive.dm
```diff
@@ -63,7 +65,12 @@ TYPEINFO(/datum/component/radioactive)
 			src._backup_color = PA.color
 			PA.add_filter("radiation_color_\ref[src]", 1, color_matrix_filter(normalize_color_to_matrix(PA.color ? PA.color : "#FFF")))
 			PA.color = null
-		PA.add_simple_light("radiation_light_\ref[src]", rgb2num(color))
+		var/list/color_composition = rgb2num(color)
+		var/R = color_composition[1]
+		var/G = color_composition[2]
+		var/B = color_composition[3]
+		var/A = color_composition[4]
+		our_light = PA.AddComponent(/datum/component/loctargeting/simple_light, R, G, B, A, TRUE)
```
ZeWaka, line 73:
```suggestion
		our_light = PA.AddComponent(/datum/component/loctargeting/simple_light, color_composition[1], color_composition[2], color_composition[3], color_composition[4], TRUE)
```
not quite necessary but i get it

code/datums/components/radioactive.dm
```diff
@@ -85,6 +92,7 @@ TYPEINFO(/datum/component/radioactive)
 			global.processing_items.Remove(parent)
 		global.processing_items.Remove(src)
 		PA.remove_simple_light("radiation_light_\ref[src]")
+		qdel(src.our_light)
```
ZeWaka, line 95:
```suggestion
		QDEL_NULL(src.our_light)
```

PR #22910, opened 2025-04-08
Allow whistles to be worn on belt

JORJ949:
## About the PR 
Allows whistles to be worn on the belt slot
image

## Why's this needed? 
They have a little hook at the back so it makes sense, also mime artist buff

## Changelog 

```changelog
(u)JORJ949
(+)Whistles can now be worn on the belt!
```


TobleroneSwordfish:
Does the wear state show up when in a pocket too?

TobleroneSwordfish (approved):

PR #22912, opened 2025-04-08
Cyborg radio names aren't affected by stickers and labels

pgmzeta:
[bug][silicons]
## About the PR 
Use `real_name` instead of `name` when speaking on radios and siliconchat 

## Why's this needed? 
Adding stickers or getting labelled as a borg shouldn't change your voice in any way.

You can make your name SUPER long by stacking stickers, causing even a single word repsonse to take up multiple lines in the chat log.


pgmzeta:
this looks to already be fixed in the say rework, so if that gets merged this can be closed

pgmzeta:
say rework is close enough

PR #22913, opened 2025-04-08
Shiny (and more consistent) space helmets!

FireKestrel:
## About the PR 
This touches up the icons for the mining, EVA, Syndicate, Matsci, and Nanotrasen helmets to better match their appearance when worn.

helmsbig

## Why's this needed? 
The Syndie helmet's icon has looked way off from the actual mob sprite for it since the human resprite, and
I wanted to fix that. While touching that up I ended up doing a pass on all the helmets that follow the standard EVA helmet design to ensure they're all absolutely consistent, cause why not.


JORJ949:
I think the bottom of most of them could use a little touch-up, but these look great!
image

FireKestrel:
yeah, the jaw segment is something that's been giving me a lot of trouble. I tried two other variants of it but I'm not too sure about them. 
image

FireKestrel:
helmsbig
okayy, i think this is looking a little better

NightmareChamillian:
I still think some amount of shading should be kept on the lower jaw, just because it heuristically shouldn't be one monolithic surface. Regardless it looks good!

FireKestrel:
image
Here's a version with jaw shading more in line with the original helmets and the current Captain space helmet. there's probably a better option, but I'm too tired to figure it out...

ZeWaka (commented):
honestly i much prefer the old sprites

FireKestrel:
yeah that's fair. it was just bugging me how different the syndicate helmet looks compared to how it looks when worn, but it's not really that necessary.

FireKestrel:
helmsbig
ah fuck it. one last attempt. leftmost sprites are the current mob sprites, just for comparison and for why i started this in the first place.

NightmareChamillian:
I think it's an improvement in terms of shape but I definitely think the colors need adjustment

FireKestrel:
> I think it's an improvement in terms of shape but I definitely think the colors need adjustment. The mob sprites have darker darks and lighter lights because they're trying to reflect a gradient on a much smaller scale with much less in terms of "read"ability but the light gray on the larger ones just doesn't please the eye quite as much

Yeah, that makes sense. I've been trying to keep the palettes exactly consistent between the two because that's what pretty much all the modern equipment sprites do, but I probably should try to deviate a little so the contrast isn't as harsh. Now I'm just struggling a bit with figuring out a palette that's more visually pleasing without looking like a drastic departure from the mob icons...

NightmareChamillian:
You could just keep the old colors.

FireKestrel:
okay, finally got time to work on this again. keeping the old palette seems like the best option here, so i'm probably just gonna cut this PR down to just the syndicate helmet as the old palette makes the other helmets look nearly identical to the originals.

FireKestrel:
closing this in favour of #23540.

PR #22920, opened 2025-04-09
Access-locks the Atlas mass driver

FireKestrel:
## About the PR 
This replaces the Atlas mass driver with one that requires chaplain access to open.
image

## Why's this needed? 
The mass driver is access-locked on every other station, and it's pretty weird that Atlas' is the only one that isn't.

### DIFF
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -1722,8 +1722,11 @@
 /turf/simulated/floor,
 /area/station/hallway/primary/north)
 "aie" = (
-/obj/machinery/launcher_loader/east,
-/turf/simulated/floor/plating,
+/obj/machinery/activation_button/driver_button{
+	id = "chapelgun";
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (changes requested):
Not sure what you did, but this is failing checks.

JORJ949:
> Not sure what you did, but this is failing checks.

Mass drivers appear to have been moved a tile left so their launch distance no longer reaches the edge of the Z

FireKestrel:
Ahh, THAT'S what causes it? That was driving me insane, thank you so much

FireKestrel:
still doesn't like it for some reason. dunno what's going wrong considering i haven't touched anything beyond the mass driver...
works fine when testing in-game too.

JORJ949:
One of the check fails went away but thats probably just because the atlas runtime checker has been removed with the addition of Neon.

PR #22929, opened 2025-04-10
Reduce the stamina damage of the wrench and crowbar

pgmzeta:
[balance][game objects]
## About the PR 
* Wrench stamina damage from 40 to 30
* Crowbar stamina damage from 35 to 25

## Why's this needed? 
I feel like that is a lot of stamina damage.

Other damge points of reference:
* Air tank: 55 stamina damage
* Toolbox: 50 stamina damage
* Extinguisher: 25 stamina damage
* Baseline item: 15 stamina damage

### DIFF
--- a/code/obj/item/tool/crowbar.dm
+++ b/code/obj/item/tool/crowbar.dm
@@ -15,7 +15,7 @@
 
 	force = 7
 	throwforce = 7
-	stamina_damage = 35
+	stamina_damage = 25
 	stamina_cost = 12
 	stamina_crit_chance = 10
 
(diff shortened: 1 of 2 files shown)


TDHooligan:
both of these weapons are already lacklustre especially seeing as people only gravitate to the extremes on either side (toolbox/airtank for STA, fire ext for BRUTE). dont really need touching afaik

pgmzeta:
low damage stamlocks from a one-hand tool, specifically borg omnitools which inherit these properties, are extremely boring to be killed by

TDHooligan:
> low damage stamlocks from a one-hand tool, specifically borg omnitools which inherit these properties, are extremely boring to be killed by

this is worthy of discussion. i raised it with toolboxes etc when the OG stun reduction PR came out. probably worth another PR

PR #22935, opened 2025-04-11
Pharmacy Chemical Locker

JORJ949:
## About the PR 
Adds a new pharmacy chemical locker to replace chemical lockers with medical access varedited on them
image

## Why's this needed? 
Part of atomising #22800
This is a common object across maps so should use a subtype


TobleroneSwordfish:
This one doesn't exist in secret 👍

PR #22946, opened 2025-04-11
Change Oubliette medal obtainment method

JORJ949:
## About the PR 
Removes the "Have an antagonist in the brig at roundend" Head of Security Objective
Removes the commented out code for the secoff version of this objective and the security belt reward from the associated medal.
The "Suitable? How about the Oubliette?!" medal is now obtained by being flushed down a brig chute.

## Why's this needed? 
This objective was removed from Officer presumably because it encouraged giving extra time just to get the objective done, HoS whitelisted players won't do this essentially making the objective pure luck as to if you complete it while also having it. This medal is also HoS specific making it unobtainable to most players, which isn't very cool.

Being flushed out the brig is also still relevant as an oubliette is essentially prison cell that's just a pit with bars covering the top, not too different from the brig chute.

## Changelog 

```changelog
(u)JORJ949
(*)Changed how the "Suitable? How about the Oubliette?!" medal is obtained. See the PR or Wiki for details!
```

### DIFF
--- a/code/obj/machinery/floorflusher.dm
+++ b/code/obj/machinery/floorflusher.dm
@@ -156,6 +156,9 @@ ADMIN_INTERACT_PROCS(/obj/machinery/floorflusher, proc/flush)
 			boutput(M, "You fall into [src].")
 			src.visible_message("[M] falls into [src].")
 			M.set_loc(src)
+			if(istype(get_area(src), /area/station/security/brig))
+				M.unlock_medal("Suitable? How about the Oubliette?!", 1)
+				boutput(world,SPAN_SUCCESS("IT WORKED"))
 			flush = 1
 			update()
 
--- a/code/datums/crew_objective.dm
+++ b/code/datums/crew_objective.dm
@@ -102,18 +102,7 @@ ABSTRACT_TYPE(/datum/objective/crew/headofsecurity)
 			return 1
 		else
 			return 0
-/datum/objective/crew/headofsecurity/brig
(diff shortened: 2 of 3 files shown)


code/obj/machinery/floorflusher.dm
```diff
@@ -156,6 +156,9 @@ ADMIN_INTERACT_PROCS(/obj/machinery/floorflusher, proc/flush)
 			boutput(M, "You fall into [src].")
 			src.visible_message("[M] falls into [src].")
 			M.set_loc(src)
+			if(istype(get_area(src), /area/station/security/brig))
+				M.unlock_medal("Suitable? How about the Oubliette?!", 1)
+				boutput(world,SPAN_SUCCESS("IT WORKED"))
```
pgmzeta, line 161: debug code

pgmzeta:
what if it was just on the floor flushers that yeeted you out to space?

JORJ949:
> what if it was just on the floor flushers that yeeted you out to space?

There aren't that many of those in brigs (only density2 and donut3 come to mind) and that would require far more complexity than I think is necessary

TobleroneSwordfish:
How about making it stuffing someone into one of these?
image
Since it's meant to be a security medal, it should be for doing a security thing.

JORJ949:
> How about making it stuffing someone into one of these? 
Also fits better as you are throwing someone down a pit into a cell

TobleroneSwordfish (approved):

ZeWaka (approved):

PR #22950, opened 2025-04-12
Rework air scrubber power and fix some related bugs

pgmzeta:
[game objects][chemistry][bug][rework]
## About the PR 
Fixes a few bugs and reworks power usage for the portable air scrubber:
* Rework: Change the portable air scrubber's volume to 1000. This should make it easier to track the toxins pressure pulled from the air as cans are 1000 volume.
* Rework/bug: The buffer reagent container size was meant to be 500, but wasn't being set correctly. It's now set to drain 100 units per tick, from an effective cap of 50.
* Rework: Add a cap to the amount moles removed from the air per tick. This reduces the power spike from turning on a gas scrubber in a toxins-filled room, but also caps the speed with which it should work. This makes bringing additional air scrubbers more useful and practical for big spills.
* Rework: Add a baseline 1 kilowatt power usage to the scrubber, and charge 10 watts per percentage of inlet flow. This means they'll by default draw 2 kilowatts of power while idle, instead of 5 kilowatts.
* Rework: Significantly reduce the fluid drain power usage from *5 Kilowatts* per unit to 50 Watts per unit, especially now that it's filtering 100 units per tick.
* Bug: Only transfer the amount from the buffer into the reagent holder that will fit, then dump the rest directly to the floor.
* Rework: Reduce power usage per filtered mole to 70 watts from 700 watts
* Bug: Have the machine use power, instead of the area. Allows power debug prompts/tracking to work properly.

## Why's this needed? 
Scrubber power usage is extremely high and running multiple provides no to little value. This makes things filter faster with multiple scrubbers without immediately causing power to shut down. 

Power usage while scrubbing peaked at around 25 kilowatts, so still pretty heavy, but enough that a standard APC could handle the charge usage for a single scrubber.

## Changelog 

```changelog
(u)glowbold
(*)Portable air scrubbers use less power, but process less air at a time. You may want to bring more than one to the next toxins workplace "accident".
```


TobleroneSwordfish (approved):

ZeWaka (changes requested):
this needs a changelog

pgmzeta:
changelog added

PR #22951, opened 2025-04-12
Split mindprot implant from health

JORJ949:
## About the PR 
Separates the anti-mindhack implant from the security health implant. Only real player facing change other than having to cut maybe two implants out an officer you want to mindhack is the secass implant now also self destructs

## Why's this needed? 
Jobs support lists of starting implants now and these are two very different behaviours and so should be separate

## Changelog 

```changelog
(u)JORJ949
(+)Mindhack protection implants are now their own implant rather than a health implant
```


TobleroneSwordfish:
Kind of against doing this unless there's a reason for it, makes cutting the implants out to mindhack them more complicated for no real benefit.

JORJ949:
> Kind of against doing this unless there's a reason for it, makes cutting the implants out to mindhack them more complicated for no real benefit.

Being able to give jobs mind protection without giving them security health alerts.

TobleroneSwordfish:
I also don't think we should give those roles mindhack protection. It makes sense to me that it's tied to a health implant since the only roles you should give mindhack protection to would be roles you'd also want to give a health alert.

JORJ949:
> I also don't think we should give those roles mindhack protection. It makes sense to me that it's tied to a health implant since the only roles you should give mindhack protection to would be roles you'd also want to give a health alert.

These roles HAVE mindhack protection right now. Removing the security alert is a nerf to them, but not one in this PR.

PR #22953, opened 2025-04-12
Artifact reticulator - shards and combinations

FlameArrow57:
[INTERNAL][SCIENCE][FEATURE]

## About the PR 
Atomized from #21976

This PR adds systems:
- Artifact shard breakdowns. Breaking an artifact will give an omnishard, powershard, spacetime shard, or essence shard
- Artifact combinations. Artifacts can be combined into each other in different ways. Some cannot be combined into each other, some can only be combined into small or large artifacts, some accept any incoming combinations

## Why's this needed? 
For artifact reticulator PR


FlameArrow57:
x

PR #22955, opened 2025-04-12
Tooltips 3

Wirewraith:
## About the PR

Complete rewrite of the tooltip system to leverage the new webview capabilities of Byond.

> [!IMPORTANT]
> Only supports BYOND version 516 server and clients 

## Why's this needed?

Fixes:

- Tooltips "sticking" in visible state if moving the mouse fast enough over multiple tooltip targets
- Tooltips not correctly showing the item last moused over if rapidly moving between tooltip-able targets.
- Tooltips appearing extremely narrow in some cases
- Tooltips triggering multiple times if moving the mouse on a HUD area that contains an item. (The HUD area had a tooltip action, and so did the item, so moving between both would open/close the same tooltip in quick succession)
- Removes reliance on transparent-color as this feature has been dropped by Byond. This will avoid the appearance of bright pink areas around the tooltip for 516 clients
- Better integration with objDialog.dm for automatic interface updates

Adds:

- Tooltips now scale in size in correspondence with both the map size, and the user's DPI settings.

Improves:

- Performance _should_ be much improved, allowing for quicker tooltip visibility changes
- Revamped positioning code should avoid any cases of improperly placed tooltips, or having to do dumb special things (e.g. for pod UI that changes screen size)
- Overall cleanup of some areas triggering tooltips to avoid duplicated code
- Support for passing the entirety of tooltip option parameters from child objects (whereas before we only generally had support for passing alignment flags)
- Various alignment fixes on a case-by-case basis to place tooltips correctly in cases where the default of "bottom left" was incorrect (e.g. for HUD icons placed in the lower right of the screen)
- Usage of weakrefs for tooltip target storage should improve garbage collection, and removes the requirement to clear the target manually
- Redesigned the airbridge controller and pump UI "pinned" tooltips to be more fancy. Other pinned tooltips will be redesigned over time
- Use of signals removes requirement to extend lifecycle events like mob death or atom disposed.

## Changelog

```changelog
(u)Wire
(*)Tooltips rewrite for better performance and a host of bug fixes
```


Wirewraith:
Redesigned pinned tooltips for reference:

image

image

Wirewraith:
Changes:

- Handle cases where the map view size is bigger than the map pane size (for example when setting the game icon size to 128x128 on basically...any size of monitor resolution).
- Properly reset the debug performance report between shows of the same tooltip.
- Avoid stacking duplicate event handlers on every tooltip show.
- Convenience event handler for handling "clicking this element should call Topic on the src object"
- Show an error tooltip if something went wrong during setting content. To be expanded later.
- Add a new `update` flow to be used in cases where the tooltip content has changed, but there's a reasonable expectation that the client map display hasn't changed. This method is significantly faster than going through the whole "show" process.
- Add support for UI templating via [ETA](https://eta.js.org/). Think of it like a much-much-much less feature filled form of TGUI.

PR #22957, opened 2025-04-12
Artifact reticulator - reverse material breakdowns

FlameArrow57:
[INTERNAL][FEATURE]

## About the PR 
Atomized from #21976

Adds framework for reverse material breakdowns from the artifact reticulator

## Why's this needed? 
For artifact reticulator PR


mordent-goonstation:
As I understand it, this feature is around turning artifacts into basic materials? That seems counterintuitive to what artifacts are.

FlameArrow57:
It's around breaking station objects into basic materials - so you could turn a c saber into claretine, etc for example

FlameArrow57:
.

PR #22968, opened 2025-04-13
Fixes nudge emote for monkeys

CalliopeSoups:
[A-Game-Objects] [C-Sprites]
## About the PR 
Fixes nudge emote from #22818 for monkeys. Looks a bit odd to me but it's better than it was.

## Why's this needed? 
Overlay needs to change position if you're a monke.

### DIFF
--- a/code/obj/item/clothing/glasses.dm
+++ b/code/obj/item/clothing/glasses.dm
@@ -46,8 +46,7 @@
 		if (!src.is_nudged)
 			var/image/glasses_flash_overlay = image('icons/mob/clothing/eyes.dmi', icon_state=flash_state, layer=src.wear_layer+0.1)
 			glasses_flash_overlay.alpha = 100
-			if (istype(H.mutantrace, /datum/mutantrace/monkey))
-				glasses_flash_overlay.pixel_y -= 6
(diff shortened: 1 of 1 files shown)


itchybones:
Somewhat unrelated to the PR itself, but you don't have to use the C/A prefixes or write the dashes for labels. [game objects] would work, for example. Labels are tricky, I've been there myself.

code/obj/item/clothing/glasses.dm
```diff
@@ -46,6 +46,9 @@
 		if (!src.is_nudged)
 			var/image/glasses_flash_overlay = image('icons/mob/clothing/eyes.dmi', icon_state=flash_state, layer=src.wear_layer+0.1)
 			glasses_flash_overlay.alpha = 100
+			if (istype(H.mutantrace, /datum/mutantrace/monkey))
```
TobleroneSwordfish, line 49: This should probably use the mutantrace's `head_offset` var?

TobleroneSwordfish (changes requested):

code/obj/item/clothing/glasses.dm
CalliopeSoups: Thanks, wasn't aware of that!

TobleroneSwordfish (approved):

PR #22973, opened 2025-04-13
AI image performance enhancement attempt

FlameArrow57:
[AI][PERFORMANCE]

## About the PR 
Attempts to reduce number of images shown to AI players for performance

Maybe this code is bad, idk!

image

## Why's this needed? 
Reduces number of images shown to the AI, making game hopefully less prone to crashing and makes deploying to shells/mainframe faster

```changelog
(u)Flamearrow
(*)AI static rendering has been reworked, please report any glitches or performance changes.
```


code/mob/living/intangible/ai-camera.dm
```diff
@@ -127,6 +118,30 @@
 		if(src.loc.z != 1)	//you may only move on the station z level!!!
 			src.cancel_camera()
 
+	proc/add_all_statics()
+		if (!src.loc)
+			return
+		for (var/turf/T as anything in block(src.loc.x - 11, src.loc.y - 8, src.loc.z, src.loc.x + 11, src.loc.y + 8, src.loc.z))
```
TobleroneSwordfish, line 124: These should get values from `client.view` instead of being magic numbers. In testing it seems like the edges of the screen flicker a bit so maybe add +1 all around to extend just past the edge of the screen.

code/mob/living/intangible/ai-camera.dm
```diff
@@ -58,34 +58,25 @@
...
-					LAGCHECK(LAG_LOW)
+		if (!src.client) // just client things
+			return
+		src.bioHolder.mobAppearance.pronouns = src.client.preferences.AH.pronouns
+		src.update_name_tag()
+		src.job = "AI"
+		if (src.mind)
+			src.mind.assigned_role = "AI"
+		animate(src.client, 0.3 SECONDS, color = client_color)
+		SPAWN(0)
```
TobleroneSwordfish, line 69: Could maybe remove the `animate` and `SPAWN` since they're meant to cover up the time during the lagspike when the images haven't loaded yet.

TobleroneSwordfish (changes requested):
Interesting idea, worth testmerging to find out how it works on live 👍

TobleroneSwordfish:
Okay overall this seems to work well, it does need to account for viewports though: #23461

TobleroneSwordfish (changes requested):
I guess I should [request changes], see above^

pgmzeta:
this is causing runtimes on local

FlameArrow57:
I wasn't getting any runtimes when testing. I'd suggest unmerging this if that's a problem, I don't have any interest in fixing bugs atm

PR #22472, opened 2025-03-01
ID Card band overlay

JORJ949:
## About the PR 
- Changed how IDs work, they now have a base sprite and band overlay that determines the colour (old id sprites are still available, only place they exist in public repo is shady_robot and thats some evil UI magic I dare not touch)
- Replaced card/id/nt_specialist and card/id/blank_polaris with card/id/nanotrasen (May need secret updates)
- Icon state renames: (may require secret updates)
   - "id" -> "id_basic"
   - "gold" -> "id_gold"
   - "polaris" -> "id_nanotrasen"
- Allow jobs to specify a band colour to override with.
- IDs that previously prevented respriting now either support adding a band or don't (e.g. NT IDs can have a band, clown IDs cannot)
- Agent ID uses TGUI input boxes (may aswell while I'm here)

## Why's this needed? 
More customisability with job IDs is great, the present sprites are mostly copied and this allows more interesting (and clear) IDs to be made.

## Changelog 

```changelog
(u)JORJ949
(+)All ID cards (except the clown's) can now have any departmental stripe applied to them!
```

PR #22986, opened 2025-04-14
Id card Gold/NT icon state name improvement

JORJ949:
## About the PR 
Renames the "id" "gold" and "polaris" card icon states to "id_basic" "id_gold" and "id_nanotrasen"

## Why's this needed? 
Improved readability in sprite names and #22472 is huge and this bit can 100% be split off to make that easier to review


JORJ949:
> concerned about sekrit

Will probably need secret changes yeah, careful though as a couple other things have the "id" and "gold" icon states such as ID computers and gold guardbuddies.

PR #22995, opened 2025-04-15
Fining access

JORJ949:
## About the PR 
Changes ticket master PDA app to:

- use new  fine accesses rather than a defined list of jobs that can fine.
- require an ID to be inserted into the PDA to use it.
- checks if the inserted ID can issue tickets before issuing a ticket
- use the inserted ID's job for ticket and fine approval job

## Why's this needed? 
The security record job checking is very clunky and nobody ever changes it if for example a staff assistant got promoted to HoP they would be unable to ticket unless the security record was changed, allows for more gimmicky ticketing jobs to show up like "Shoe Inspector" can ticket you for bad shoes.

## Changelog 

```changelog
(u)JORJ949
(+)Issuing fines is now tied to two new accesses rather than job on manifest.
(+)The Ticket Master App now requires an ID to be inserted in your PDA.
```


pgmzeta (changes requested):
small tweaks needed, but i like this

we really need to redo the access list but not in this pr

flappybatpal (approved):

RubberRats (approved):

code/obj/item/device/pda2/smallprogs.dm
```diff
@@ -1043,18 +1058,18 @@ Using electronic "Detomatix" SELF-DESTRUCT program is perhaps less simple!<br>
...
-
-					var/PDAowner = src.master.owner
-					var/PDAownerjob = data_core.general.find_record("name", PDAowner)?["rank"] || "Unknown Job"
-
 					dat += "<br><br><a href='byond://?src=\ref[src];back=1'>Back</a>"
 
 					dat += "<h4>Fine Request List</h4>"
+					var/ticket_level = src.get_ticket_level()
+					if (ticket_level < 2)
+						dat += "<br><br>Please insert a security ID to approve fines.<br><br>"
```
pgmzeta: doesn't need to be a security ID, just one with access

PR #22998, opened 2025-04-15
Credits artifact trigger

pgmzeta:
[feature][game objects]
## About the PR 
Add a new artifact trigger type: Credits. It's triggered by using a significant stack of credits on artifacts. 

## Why's this needed? 
Unique artifact trigger type

## Changelog 

```changelog
(u)glowbold
(*)Some artifacts require credits to activate.
```

### DIFF
--- a/code/obj/artifacts/artifact_items/talisman.dm
+++ b/code/obj/artifacts/artifact_items/talisman.dm
@@ -174,7 +174,8 @@
 	rarity_weight = 275
 	validtypes = list("wizard", "precursor")
 	validtriggers = list(/datum/artifact_trigger/force, /datum/artifact_trigger/electric, /datum/artifact_trigger/heat,
-		/datum/artifact_trigger/radiation, /datum/artifact_trigger/carbon_touch, /datum/artifact_trigger/silicon_touch, /datum/artifact_trigger/data)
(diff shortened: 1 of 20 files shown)


LorrMaster:
Thematically I think that this should only be available to certain artifact types (maybe wizard, martian, and eldritch). At least precursor artifacts probably predate credits.

sunnyariess:
Research Trainees have `PAY_UNTRAINED`, they get paid 150, Scientists have `PAY_DOCTORATE`, and get paid 600. `src.stimulus_amount = 100 * rand(5,15)` makes it a minimum of 500, and a maximum of 1500. It's, for the average scientist, two to three pay cycles, and one to two with unionized - and way too many for trainees. I think a good compromise would be having it also work with fake printed money, so you don't have to reserve any real credits.

Valtsu0:
Science has a bunch of ways to make money. Most relevant here is probably rushing mars. Alternatively pressure crystals.

It does encourage cross-branch cooperation and coordinating with the RD

TobleroneSwordfish:
Not a fan of this, it doesn't really make sense and just feels like it'll add to the pile of "well I tried everything obvious" unactivated artifacts.

PR #23000, opened 2025-04-15
Forensic Pug Sleuthing

LorrMaster:
[player actions][code quality]

## About the PR 
Rewrites pug sleuthing to introduce the new forensics system. All atoms now have a forensics holder that can be used to store forensic data.

## Why's this needed? 
Best way to start updating forensics without breaking literally everything. Also makes scents easier to work with.


code/modules/forensics/forensic_group.dm
```diff
@@ -0,0 +1,192 @@
...
+	proc/copy_group(var/datum/forensic_holder/new_holder, var/include_trace = FALSE)
+		return null
+	proc/get_header() // The label that this evidence will be displayed under for scans
+		return SPAN_ALERT("Null Header")
+	proc/remove_evidence(var/datum/forensic_holder/parent, var/removal_flags)
+		return
+	proc/matching_flags(var/flags_A, var/flags_B)
+		return (flags_A & ~FORENSIC_FAKE) == (flags_B & ~FORENSIC_FAKE)
+
+/datum/forensic_group/notes
```
TobleroneSwordfish, line 25: What is this for?

code/modules/forensics/forensic_group.dm
```diff
@@ -0,0 +1,192 @@
...
+	var/list/datum/forensic_data/basic/notes_list = new/list()
+
+	disposing()
+		src.notes_list.len = 0
+		src.notes_list = null
+		..()
+
+	apply_evidence(var/datum/forensic_data/data)
+		if(istype(data, /datum/forensic_data/basic))
+			var/datum/forensic_data/basic/E = data
```
TobleroneSwordfish, line 37: Please don't use single letter variable names unless it's something simple and standard like `turf/T`

code/modules/forensics/forensic_group.dm
```diff
@@ -0,0 +1,192 @@
...
+			c_text = "You also [detect] [intensity] [scent] of [color]."
+		return "<li>[SPAN_NOTICE(c_text)] [time]</li>"
+
+	proc/get_intensity(var/list/intensity_list, var/list/time_since_list, var/time_since)
+		for(var/i in 2 to length(intensity_list))
+			if(time_since < time_since_list[i] MINUTES)
+				return intensity_list[i]
+		return intensity_list[1]
+
+/datum/forensic_group/multi_list // Two or three pieces of evidence grouped together
```
TobleroneSwordfish, line 155: Why do we have both this and `/datum/forensic_data/multi`? It seems like you're replicating the same functionality at two different levels which is incredibly confusing.

code/modules/forensics/forensic_group.dm
```diff
@@ -0,0 +1,192 @@
...
+				src.evidence_list.Cut(i, i+1)
+
+/datum/forensic_group/basic_list/sleuth // Used to store smells for Pug sleuthing
+	category = FORENSIC_GROUP_SLEUTH
+	group_flags = FORENSIC_REMOVAL_CLEAN
+
+	get_header()
+		return "Sleuth"
+
+	proc/get_sleuth_text(var/atom/A, var/sleuth_all = FALSE, var/accuracy = -1)
```
TobleroneSwordfish, line 120: Why is something like getting the text value of a group being defined on the lowest level child type? Surely this should be generic?

TobleroneSwordfish (changes requested):
Comments on types, member variables and procs should use doc comment styling, like:
```
/// Does a thing
proc/thing()
```
or
```var/foo //! Does a thing```

Overall the structure of this code seems a little over-engineered and confusing, there are 4 levels of abstraction and two classes on different levels seem to be doing the same thing?
This probably isn't helped by the fact that some of these types appear to exist for no reason due to being poorly atomized out of a larger project. Please only include what you're actually using for the functionality of this PR.

File: code/modules/forensics/forensic_data.dm [hunk 5]
```diff
@@ -0,0 +1,137 @@
...
+	New()
+		..()
+		src.time_start = TIME
+		src.time_end = time_start
+		src.perc_offset = (rand() - 0.5) * 2
+		src.accuracy_mult *= ((rand() - 0.5) * 0.15) + 1
+	proc/get_text() // The text to display when scanned
+		return ""
+	proc/should_remove(var/remove_flags) // Should this evidence be removed?
+		var/remove = HAS_ANY_FLAGS(src.flags & (FORENSIC_REMOVAL_ALL & !FORENSIC_REMOVAL_HEAL), remove_flags)
```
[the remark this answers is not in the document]
LorrMaster: I accidentally put a `!` instead of a `~`. The reason is that all the selected heal types need to be flagged for it to be considered for removal. If you're dealing both brute and burn, keep the evidence until both are healed.

code/modules/forensics/forensic_group.dm
LorrMaster: The header is not the text value of the group, it is the section that the evidence in the group will be placed under when scanned. So organ damage and a bullet wound can be be different in code, but can categorized together when scanned. I have it as a proc since it doesn't change and is not really relevant until someone scans something.

code/modules/forensics/forensic_data.dm
```diff
@@ -0,0 +1,137 @@
...
+	get_copy()
+		var/datum/forensic_data/basic/c_data = new(src.evidence, src.display, src.flags, src.value)
+		c_data.category = src.category
+		c_data.accuracy_mult = src.accuracy_mult
+		c_data.time_start = src.time_start
+		c_data.time_end = src.time_end
+		c_data.perc_offset = src.perc_offset
+		return c_data
+
+/datum/forensic_data/multi // Two or three pieces of evidence linked together. Flags not included.
```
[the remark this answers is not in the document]
LorrMaster: Footprints, retina scans, data that directly links a person's footprints with their DNA. Nothing relevant to sleuthing, so I'll remove it. Might need to rethink how evidence pairs are handled.

code/modules/forensics/forensic_group.dm
LorrMaster: Do you know what, I misunderstood what you were asking. The reason why sleuthing is using its own proc for obtaining text is because the text is obtained via an emote rather than a scanner. I have not yet programmed a way to filter out a specific group for obtaining text, so I gave it a separate procedure while the original proc for obtaining text for the forensic scanner just returned null.

code/modules/forensics/forensic_holder.dm
```diff
@@ -0,0 +1,46 @@
...
+			for(var/group in src.admin_list)
+				qdel(group)
+			src.admin_list.len = 0
+			src.admin_list = null
+		..()
+
+	/// Add forensic data to the holder, such as an indiviual fingerprint or DNA sample
+	proc/add_evidence(var/datum/forensic_data/data, var/category = FORENSIC_GROUP_NOTE, var/admin_only = FALSE)
+		if(!HAS_FLAG(data.flags, FORENSIC_FAKE))
+			var/datum/forensic_group/h_group = get_group(category, TRUE)
```
TobleroneSwordfish, line 24: this still counts as a single letter var name, what does `h` mean here??

code/modules/forensics/forensic_holder.dm
```diff
@@ -0,0 +1,46 @@
...
+		if(!admin_only)
+			var/datum/forensic_group/group = get_group(category, FALSE)
+			if(!group)
+				group = forensic_group_create(category)
+				src.group_list += group
+			group.apply_evidence(data)
+
+	/// Obtain a specific type of forensic group if it exists
+	proc/get_group(var/category = FORENSIC_GROUP_NOTE, var/admin = FALSE)
+		var/list/datum/forensic_group/E_list
```
TobleroneSwordfish, line 38: Ditto

code/modules/forensics/forensic_holder.dm
```diff
@@ -0,0 +1,46 @@
...
+		if(admin)
+			E_list = src.admin_list
+		else
+			E_list = src.group_list
+		for(var/datum/forensic_group/G in E_list)
```
TobleroneSwordfish, line 43: ditto

code/modules/forensics/forensic_group.dm
```diff
@@ -0,0 +1,97 @@
...
+		var/data_text = ""
+		for(var/i in 1 to length(src.evidence_list))
+			data_text += sleuth_data(A, src.evidence_list[i], i == 1)
+		return data_text
+
+	proc/sleuth_data(var/atom/A, var/datum/forensic_data/basic/slueth_data, var/is_first)
+		var/color = slueth_data.evidence.id
+		var/time_since = TIME - slueth_data.time_end
+		var/list/time_since_list = list(0, rand(4,6), rand(8,12), rand(27,33), rand(41,49), rand(55,65))
+		var/c_text
```
TobleroneSwordfish, line 78: ditto

code/modules/forensics/forensic_data.dm
```diff
@@ -0,0 +1,41 @@
...
+
+	New(var/datum/forensic_id/id, var/datum/forensic_display/disp = disp_empty, var/flags = 0, var/value = 0)
+		..()
+		src.evidence = id
+		src.display = disp
+		src.flags = flags
+		src.value = value
+
+	get_copy()
+		var/datum/forensic_data/basic/c_data = new(src.evidence, src.display, src.flags, src.value)
```
TobleroneSwordfish, line 38: ditto

_std/defines/forensics.dm
```diff
@@ -0,0 +1,23 @@
+
+#define FORENSIC_GROUP_NONE 0 // Should only be used by errors
+#define FORENSIC_GROUP_NOTE 1 // Basically a misc section
+#define FORENSIC_GROUP_SLEUTH 2 // Pug sleuthing smells
+
+// Each group has a unique variable. Use that to create a new group.
+// Placed here together with the FORENSIC_GROUP variable defines
+/datum/forensic_holder/proc/forensic_group_create(category)
+	var/datum/forensic_group/G
```
TobleroneSwordfish, line 9: ditto

TobleroneSwordfish (changes requested):
Okay sorry this took a while to get back to, the code structure looks sane now but you will still need to fix the single letter variable names.

TobleroneSwordfish (approved):

PR #23002, opened 2025-04-15
Cairngorm Improvements

JORJ949:
## About the PR 
Improves the cairngorm in the following ways:
- No longer uses landmarks for the nuclear mode to spawn things on with the exception of the bomb
- Doors to the side rooms (medical, armory, podbay and barracks) start bolted.
- Single use buttons inside the side rooms will unbolt the door to that room then delete themselves. These do have a toggle admin proc on the ` menu so can be toggled from ghost.
- Nuclear mode toggles all the cairngorm buttons, making the above changes unnoticeable in nuclear rounds
- Spread out many of the stacked objects on the tables

## Why's this needed? 
Allows gimmick jobs to be spawned on the cairngorm outside of nukie rounds without giving needing to give access to the incredible amount of gear available on the cairngorm

### DIFF
--- a/code/obj/machinery/door/door_control.dm
+++ b/code/obj/machinery/door/door_control.dm
@@ -605,6 +605,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/door_control, proc/toggle)
 
 	if(src.single_use)
 		qdel(src)
+		return
 
 	if(src.cooldown)
 		inuse = TRUE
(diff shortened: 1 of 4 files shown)


TobleroneSwordfish (approved):
Lgtm!

code/obj/machinery/door/door_control.dm
```diff
@@ -602,6 +603,9 @@ ADMIN_INTERACT_PROCS(/obj/machinery/door_control, proc/toggle)
 						M.operating = 0
 			M.setdir()
 
+	if(src.single_use)
+		qdel(src)
```
Tarmunora, line 607: Should return after qdel - no need to handle cooldowns etc if it's getting deleted

Tarmunora (changes requested):

PR #23004, opened 2025-04-15
Name Hand-Assembled Cleanbots

jan-antilles:
## About the PR 
This just lets you use a pen on a bucket-and-sensor array in order to give your cleanbot a custom name.

image

## Why's this needed? 
For fun!

## Changelog 

```changelog
(u)Jan.antilles
(+)Hand-assembled cleanbots can now be named with a pen before adding the final robot arm.
```

### DIFF
--- a/code/modules/robotics/bot/cleanbot.dm
+++ b/code/modules/robotics/bot/cleanbot.dm
@@ -16,21 +16,39 @@
 	w_class = W_CLASS_NORMAL
 	flags = TABLEPASS
 	var/created_cleanbot_type = /obj/machinery/bot/cleanbot
+	var/created_name = "Cleanbot"
 
-	attackby(var/obj/item/parts/robot_parts/P, mob/user as mob)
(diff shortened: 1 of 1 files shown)


code/modules/robotics/bot/cleanbot.dm
```diff
@@ -16,21 +16,39 @@
...
-		qdel(src)
-		return
+			boutput(user, "You add the robot arm to the bucket and sensor assembly! Beep boop!")
+			A.name = src.created_name
+			qdel(I)
+			qdel(src)
+			return
+
+		if (istype(I, /obj/item/pen))
+			var/t = input(user, "Enter new robot name", src.name, src.created_name) as null|text
```
JORJ949, line 39: Avoid using one character variable names "t"
Also should use tgui_input_text

TobleroneSwordfish:
This is copy-paste but it's copy-pasted onto all of these bot assembly types so I won't ask you to clean it up, better to have it working on all of them even if it's Bad.

PR #23009, opened 2025-04-16
Emagged IDs keep special accesses

JORJ949:
## About the PR 
Emagging an ID card now has a 20% chance per access not in get_all_accesses(), to keep that access

## Why's this needed? 
These accesses tend to otherwise be unobtainable and so should not be easily removable.
Syndicate IDs had special behaviour to keep their special accesses but if a non-syndicate ID had syndie access it would be removed on emagging. 
Prevents removing HoS/armory access without destroying the card, but I have never seen anyone emag the HoS ID with intent to remove their access as to emag it you first have to have it, and antags tend to like free entry to the armory.

## Changelog 

```changelog
(u)JORJ949
(+)Emagged IDs now have a chance to keep any special accesses, such as Armory, Centcom, and Azone accesses.
```

### DIFF
--- a/code/obj/item/card.dm
+++ b/code/obj/item/card.dm
@@ -299,9 +299,13 @@ TYPEINFO(/obj/item/card/emag)
 		return FALSE
 	var/list/all_accesses = get_all_accesses()
 	var/bonus_access = list() // preserve accesses which are otherwise unobtainable
+	var/special_accesses = list(access_maxsec, access_armory)
 	for (var/access in src.access)
 		if (!(access in all_accesses))
-			bonus_access += list(access)
+			if((access in special_accesses) && prob(33))
+				bonus_access += list(access)
+			else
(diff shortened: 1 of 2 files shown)


JORJ949:
> Kinda disagree with armory access getting magic protection. Such protections should be avoided as much as possible.

Having armory access doesn't help particularly as it will still probably remove security access, its also rare cards with armory or HoS access get emagged anyway as those accesses are impossible to obtain if you emag them away unless an agent card has copied them, though I'm not against adding a special case for HoS and armory access being able to be removed I doubt it will make much of a difference.

JORJ949:
Updated to have a list of special_accesses that only have a 33% chance to remain on emagging, which means you still can't randomly get this access, but can randomly keep it. List contains maxsec and armory access

TobleroneSwordfish:
Meh, I don't see a reason to make these magic.

PR #23018, opened 2025-04-17
Fix construction for SMES units 

pgmzeta:
[bug][game objects]
## About the PR 
smes units need power terminals to function, this makes them put down terminals when building and pick up when deconning.

also newly constructed units start with 0 power, no free 20 MW power

## Why's this needed? 
constructed SMES units don't do anything right now


TobleroneSwordfish (changes requested):
Should handle there already being a terminal there, as the use case is mostly going to be replacing broken/stolen SMES units

ZeWaka (approved):

PR #23024, opened 2025-04-17
Makes chem condensers cheaper

CalliopeSoups:
[balance]
## About the PR 
This PR just changes the molitz in the condenser's recipe to crystal in general. Previously did this for all glassware.

## Why's this needed? 
A scientist can gather some crystal stuff but I've had a handful of rounds where molitz just isn't available, or isn't sold until halfway through the round.

## Changelog 

```changelog
(u)CalliopeSoups
(+)Chem condensers now accept any crystal material.
```

### DIFF
--- a/code/datums/manufacturing.dm
+++ b/code/datums/manufacturing.dm
@@ -868,7 +868,7 @@
 
 /datum/manufacture/condenser
 	name = "Chemical Condenser"
-	item_requirements = list("molitz" = 5)
+	item_requirements = list("crystal" = 5)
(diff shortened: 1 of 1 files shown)


CalliopeSoups:
> I think for normal condensers its fine.
> 
> For the other 3, i feel you should certainly require molitz. These are clear upgrades for a department and chem should not have free access to them without the help of a different department.

Looks like I goofed and caused conflicts too, I'll edit it in a bit.

TobleroneSwordfish (changes requested):
Should have changelog

PR #23035, opened 2025-04-18
Fully integrate pipebombs into the assembly-system

M-Earthfire:
[Rework][Game Objects]

## About the PR 

This PR aims to remove the difference between the pipebomb assemblies and standard pipebombs made via timers.

To do this, state 4 of `/obj/item/pipebomb/frame` (the cabled pipebomb) fully replaced `/obj/item/pipebomb/bomb`, meaning all pipebombs start out as cabled bombs without any triggering devices.

For any pipebombs that need to be available on spawn, `/obj/item/assembly/timer_ignite_pipebomb` was added. These spawn timer/igniter/pipebomb assemblies, ready to use with a 3 second timer on start.

Subtypes for controlled demolition pipes and mini syndicate pipebombs were created as well and replaced their corresponding pipebombs.

Non-timer mini syndicate pipebombs and controlled demolition pipes have been given new sprites, which directly integrate into assemblies:
Pipbombs

Also, the construction of cabled pipebombs has been given a little quality pass, now requiring 3 pieces of cable coil and putting the result directly in the users hand. 

## Why's this needed? 

With the assembly rework in [#22115](https://github.com/goonstation/goonstation/pull/22115), standard pipebombs design space (in hand useable pipebombs without any fuss) becomes obsolete. Having two different construction methods for the same object is not good, so the standard pipebombs got the axe to be fully incorporated into the modular system.

The other kinds were also incorporated into the system. This enables creative antags/engineers to change up the assemblies they can buy, if they want. 

The cable cost for the construction process is added because it doesnt make sense that cabling pipebombs doesnt consume any cables.

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Removes non-assembly pipebombs from the game. Every type of pipebomb (mini-syndicates, controlled demolition pipes) spawn as timer/igniter/pipebomb-assemblies now.
(+)Mini syndicate pipebombs and controlled demolition pipes now have been given cabled pipebomb forms that can be gained from disassembling their corresponding assemblies, ready to be used in other assemblies.
(+)Applying cables during the pipebomb crafting process now consumes 3 pieces of the cable coil
```


code/obj/item/assembly/misc_assemblies.dm
```diff
@@ -719,6 +719,35 @@ Contains:
...
 
+
+/////////////////////////////////////////////////// timer pipebomb ////////////////////////////////////
+// These do include tiny syndicate pipebombs and engineering pipbombs
+
+
+/obj/item/assembly/timer_ignite_pipebomb
+	secured = TRUE
+	var/pipebomb_path = /obj/item/pipebomb/bomb
+	var/bomb_strenght = null
```
TobleroneSwordfish, line 730:
```suggestion
	var/bomb_strength = null
```

code/obj/item/assembly/misc_assemblies.dm
```diff
@@ -719,6 +719,35 @@ Contains:
...
+
+/obj/item/assembly/timer_ignite_pipebomb/New()
+	..()
+	var/obj/item/device/timer/new_trigger = new /obj/item/device/timer(src)
+	new_trigger.time = 3 SECONDS //have a time already set up and don't blow up unsuspecting users on use
```
TobleroneSwordfish, line 735:
```suggestion
	new_trigger.time = 5 SECONDS //have a time already set up and don't blow up unsuspecting users on use
```
To match the old pipebomb timer. I've pushed a fix to make timers actually accurate at low durations so it should match up nicely.

TobleroneSwordfish (changes requested):

TobleroneSwordfish (approved):

PR #23036, opened 2025-04-18
Falling while juggling assemblies can secure and set them off

M-Earthfire:
[Game objects][feature]

## About the PR 

This change makes it so that assemblies get set off when you fall with them during juggling, akin to grenades or ttvs.

If it is unsecured, it will magically secure itself on doing so.

## Why's this needed? 

Grenade juggling and successively exploding is funny.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Falling while juggling assemblies might set them off, even if you were very and absolutely clowntastically sure they were unsecured.
```

### DIFF
--- a/code/mob/living/carbon/human.dm
+++ b/code/mob/living/carbon/human.dm
@@ -3005,6 +3005,15 @@ Tries to put an item in an available backpack, belt storage, pocket, or hand slo
 				ttv.toggle_valve()
 				logTheThing(LOG_BOMBING, src, "accidentally [ttv.valve_open ? "opened" : "closed"] the valve on a TTV tank transfer valve by failing to juggle at [log_loc(src)].")
 				message_admins("[key_name(usr)] accidentally [ttv.valve_open ? "opened" : "closed"] the valve on a TTV tank transfer valve by failing to juggle at [log_loc(src)].")
+			else if (istype(A, /obj/item/assembly))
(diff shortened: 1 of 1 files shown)


code/mob/living/carbon/human.dm
```diff
@@ -3005,6 +3005,15 @@ Tries to put an item in an available backpack, belt storage, pocket, or hand slo
...
 				logTheThing(LOG_BOMBING, src, "accidentally [ttv.valve_open ? "opened" : "closed"] the valve on a TTV tank transfer valve by failing to juggle at [log_loc(src)].")
 				message_admins("[key_name(usr)] accidentally [ttv.valve_open ? "opened" : "closed"] the valve on a TTV tank transfer valve by failing to juggle at [log_loc(src)].")
+			else if (istype(A, /obj/item/assembly))
+				var/obj/item/assembly/dropped_assembly = A
+				if(!dropped_assembly.secured)
+					// You're not evading death that easily, clown
+					dropped_assembly.secured = TRUE
+					dropped_assembly.add_fingerprint(src)
+					dropped_assembly.UpdateIcon()
+					dropped_assembly.last_armer = src
```
TobleroneSwordfish, line 3015: Shouldn't this be handled by the attackself call?

TobleroneSwordfish (changes requested):

PR #23038, opened 2025-04-18
Make containers made out of soft materials prevent mousetrap explosions

M-Earthfire:
[Materials][Feature]

## About the PR 

This PR makes it so that backpacks, crates and gifts don't set of live mousetraps, if their material has a hardness value of 1 or less. 

Pizza crates aren't affected since their material is cardboard (hardness 0) and they are explicitly meant to trigger mousetrap bombs.

Also, this PR removes an uneeded piece of code in regarding storages and mines, since the assembly rework changed them to work with `COMSIG_ITEM_STORAGE_INTERACTION`... sorry, i commited it with the rest accidently instead of stuffing it in its own PR

## Why's this needed? 

Mousetrap rigged backpacks can be very lethal and don't have much counters in case other needed stuff, e.g. the captains spare, is in the same place.

This gives some counterplay in these scenarios. If you suspect really, really hard that a container is rigged, you can electroplate it in cotton and open it. Likewise, if you don't want that to happen, you can electroplate your container in steel, which is a giveaway of the container being rigged. Also, this enables neat stuff like transporting live mousetrap bombs in cardboard crates.

And i think this is just a funny neat matsci interaction.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Active mousetraps in backpacks or crates made out of very soft materials, like cotton, don't trigger when opened.
```

### DIFF
--- a/code/modules/materials/Mat_ProcsDefines.dm
+++ b/code/modules/materials/Mat_ProcsDefines.dm
@@ -319,6 +319,7 @@ proc/get_icon_states(icon)
 		src.material.triggerDrop(user, src)
 	return
 
+
 /// Merges two material names into one.
 /proc/getInterpolatedName(var/mat1, var/mat2, var/t)
 	var/ot = 1 - t


TobleroneSwordfish:
I don't think everything needs "counterplay" and making certain containers just not work with mousetraps based on a hardness value you can't read sounds needlessly obscure and frustrating.

PR #23041, opened 2025-04-18
Replace customs bank records computer with personnel management computer

pgmzeta:
[feature][game objects]
## About the PR 
Replaces the Bank Records computers in customs with the Personnel Management Computer, which contains SecMate, BankBoss, and RoleControl.

## Why's this needed? 
Not all maps have a secure records computer in/near customs, but consensus was the HoP should be able to view security records.

## Screenshot
The new personnel management computer
image

## Changelog 

```changelog
(u)JORJ949 & glowbold
(+)Customs offices have a personnel management computer, containing BankBoss, SecMate, and RoleControl.
```


JORJ949:
I think it should be something like the "personnel management console" with bankboss, secmate and rolecontrol, or having bank records and sec records computers together in bridge is pointless and it should DEFINITELY have a radio if it can set arrest.

TobleroneSwordfish:
That sounds good, the only reason rolecontrol is on the comms console is because I didn't have a better place to put it (and I guess technically it's communicating with central for the job requisition stuff?)
Setting people to arrest with no PDA notification is definitely Bad though.

pgmzeta:
changed the secondary card to use a pda radio, so arrest alerts will be sent out like security computers.

pgmzeta:
i don't mind mapping if i get a sprite but it would still be in a separate pr

pgmzeta:
Updated with a new sprite courtesy of @JORJ949

JORJ949:
Could probably also get medtrak so a HoP adding a new hire can add a medical record.

PR #23042, opened 2025-04-18
Remove req_access_txt from carts

JORJ949:
## About the PR 
Removes req_access_txt varedits from all /obj/storage/carts
(I have no idea why kappa.dmm does this with every single change I do to it, i'm just editing the actual code here not the maps so its not strongdmm doing anything)

## Why's this needed? 
These aren't secure or lockable, the access does nothing.

### DIFF
--- a/maps/kondaru.dmm
+++ b/maps/kondaru.dmm
@@ -58276,10 +58276,7 @@
 /turf/space,
 /area/station/turret_protected/armory_outside)
 "wFB" = (
-/obj/storage/cart/mechcart{
-	desc = "A big rolling supply cart equipped for handling hull breaches.";
-	name = "breach repair cart"
-	},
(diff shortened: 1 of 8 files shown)


pgmzeta:
i would like them to be secureable/lockable :(

JORJ949:
> i would like them to be secureable/lockable :(

Features not the goal here, and if there are ever secure ones they should be consistently mapped, these appear to be a result of just selecting thing on one map and placing on another.

maps/kondaru.dmm
```diff
@@ -58278,8 +58278,7 @@
 "wFB" = (
 /obj/storage/cart/mechcart{
```
pgmzeta, line 58279: this should use the `/obj/storage/cart/mechcart/breach` subtype instead, probably?

PR #23044, opened 2025-04-18
Give white hole thrown objects some wall phasing

pgmzeta:
[balance][events]
## About the PR 
Gives objects thrown by white holes phasing through up to 2 turfs from its center point. This is enough to make items clear the white hole effect.

## Why's this needed? 
White holes are regularly sealed off in a neat 3x3 square, severely limiting their impact on the round.
There's a TODO here suggesting this is a good idea.

## Changelog 

```changelog
(u)glowbold
(+)White holes throw items through the edge of their effect. Update your containment protocols!
```


code/datums/controllers/throwing_controls.dm
```diff
@@ -121,10 +123,18 @@ var/global/datum/controller/throwing/throwing_controller = new
 					new /obj/effects/explosion/dangerous(next)
 					end_throwing = FALSE
 					thr.throw_type = THROW_NORMAL
+
+			if (thr.throw_type == THROW_PHASE && thr.get_throw_travelled() > 2)
+				thr.throw_type = THROW_NORMAL
+				thing.event_handler_flags &= ~MOVE_NOCLIP
```
ZeWaka, line 129: My only concern is, what if it started with no clip, but that isn't going to happen in this usage so... Doesn't really matter IMO

ZeWaka (approved):

code/datums/controllers/throwing_controls.dm
pgmzeta: Do you think it's worth it to do add initial flag checking?

ZeWaka: I guess just doing initial() would be fine yeah

code/datums/controllers/throwing_controls.dm
```diff
@@ -121,10 +123,18 @@ var/global/datum/controller/throwing/throwing_controller = new
...
+
+			if (thr.throw_type == THROW_PHASE && thr.get_throw_travelled() > 2)
+				thr.throw_type = THROW_NORMAL
+				thing.event_handler_flags = initial(thing.event_handler_flags)
+
 			if (!thing.Move(next))  // Grayshift: Race condition fix. bump proc calls are delayed past the end of the loop and won't trigger end condition
-				thr.hitAThing = TRUE // of !throwing on their own, so manually checking if Move failed as end condition
-				end_throwing = TRUE
-				break
+				if (thr.throw_type == THROW_PHASE)
```
TobleroneSwordfish, line 132: Why try `Move` and then `set_loc` instead of just `set_loc`ing directly?

TobleroneSwordfish (changes requested):

code/datums/controllers/throwing_controls.dm
pgmzeta: can you take a look at the new approach taken?

TobleroneSwordfish (approved):

PR #23046, opened 2025-04-19
Cargo Pad renaming and transport accidents

LorrMaster:
[game objects][feature]

## About the PR 
- Disabled Cargo Pads can be renamed with a multitool.
- Pads can be given duplicate names if both of their receivers are turned off.
- Pads with duplicate names will create a transport mishap.
- Cargo pad lights emit blue when receiving and red when dangerous.
- Cargo pads & transporters require telecrystals to fabricate.

https://github.com/user-attachments/assets/051b5689-9e19-4303-9f09-2b595482a9ed

## Why's this needed? 
- Adds a simple interaction to cargo pads that makes them more interesting.

## Changelog 
```changelog
(u)LorrMaster
(*)Cargo Pads can be renamed with a multitool
(+)Cargo pads with duplicate names will cause a mishap.
```


code/obj/mining.dm
```diff
@@ -2319,6 +2344,103 @@ TYPEINFO(/obj/item/cargotele)
...
+		for(var/atom/movable/AM in S.contents)
+			var/obj/submachine/cargopad/tele_pad = target_list[pad_index]
+			if (tele_pad != main_pad || prob(50))
+				AM.set_loc(get_turf(tele_pad))
+				ThrowRandom(AM, dist = rand(0,5))
+				if(++pad_index > length(target_list))
+					pad_index = 1
+
+	// Transport accident where mobs can lose items and organs
+	proc/mishap_mob(var/mob/M, var/list/obj/submachine/cargopad/target_list)
```
ZeWaka, line 2397: Like I said in imcoder, this is just way too easy of a murder method for practically free

LorrMaster: I changed it so that you now only lose one organ/limb per name-duplicate cargo pad.

LorrMaster:
I messed something up.

PR #23049, opened 2025-04-19
Throw people into washing machines

CalliopeSoups:
[game objects] [feature]
## About the PR 
Does what it says on the tin. You'll still need to turn the washing machine on so it's not too strong.

## Why's this needed? 
Funny interaction and provides possible mechcomp synergies.

## Changelog 

```changelog
(u)CalliopeSoups
(+)You can now throw people into open washing machines.
```

### DIFF
--- a/code/WorkInProgress/laundry.dm
+++ b/code/WorkInProgress/laundry.dm
@@ -258,11 +258,11 @@ TYPEINFO(/obj/submachine/laundry_machine)
 
 /obj/submachine/laundry_machine/hitby(atom/movable/MO, datum/thrown_thing/thr)
 	if (istype(MO, /mob/living))
-		if (src.on == 0)
-			var/mob/living/H = MO
-			H.visible_message(SPAN_ALERT("<B>[H] gets tossed into the washing machine!</B>"))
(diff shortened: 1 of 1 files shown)


code/WorkInProgress/laundry.dm
```diff
@@ -256,6 +256,19 @@ TYPEINFO(/obj/submachine/laundry_machine)
 	else
 		return ..()
 
+/obj/submachine/laundry_machine/hitby(atom/movable/MO, datum/thrown_thing/thr)
+	if (istype(MO, /mob/living))
+		if (src.on == 0)
```
TobleroneSwordfish, line 261:
```suggestion
		if (src.open)
```
Probably makes more sense?

code/WorkInProgress/laundry.dm
```diff
@@ -256,6 +256,19 @@ TYPEINFO(/obj/submachine/laundry_machine)
...
+			var/mob/living/H = MO
```
TobleroneSwordfish, line 262: `H` means "human", this isn't a human var.

TobleroneSwordfish (changes requested):

PR #23050, opened 2025-04-19
Abstract Syndicate parent job

JORJ949:
## About the PR 
Adds abstract parent to syndicate jobs
Syndicate job special setup applies generic syndicate agent antag with the name of the job and changes the ID sprite to the syndicate one.

## Why's this needed? 
Generic parent is good, lays groundwork for other syndicate themed jobs using the nukie loadouts.


flappybatpal:
I don't see any issues with secret.

PR #23056, opened 2025-04-19
Fix Basic, Gold and NT ID in PDA sprites.

JORJ949:
## About the PR 
Update icon state names in pda.dmi to the new ID icon state names for basic, gold and NT IDs.

## Why's this needed? 
These take the icon states of the ID for the icon states to overlay and so must be the same, missed in original PR


JORJ949:
Not sure why IconDiff is dying or why I've apparently removed 94% of the file, surely its fine :)

JORJ949:
TOTALLY REAL ICONDIFFBOT:

PR #23062, opened 2025-04-20
Repath gold and NT IDs

JORJ949:
## About the PR 
Repaths the different gold and NT IDs into a common type (updatepaths script included):
- /card/id/blank_polaris and /card/id/nt_specialist are now /card/id/nanotrasen
- remove /card/id/blank_deluxe and replace all instances with /card/id/gold
- Captain's spare is now a subtype of the gold ID (this is why every single map is changed)
- Remove santa's captain's spare as santa no longer gets this.

Does not touch the azone versions of these IDs (owlery gold, polaris)

## Why's this needed? 
Having multiple types that do the same thing is silly.


ZeWaka:
Why rename it?

PR #23065, opened 2025-04-20
Add osmium and osmiridium metals

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Adds
-New metal Osmium, which has 8 hardness, 8 density, 7 chemical resistance. Gives you tox damage while held.
-New material recipe, Osmiridium, which has 10 hardness, 10 density, 7 chemical resistance

Osmium is found at the rare metal mining event

## Why's this needed? 
Adds a recipe to aim for

## Changelog 

```changelog
(u)FlameArrow57
(+)Added Osmium metal to rare metal mining encounters
(+)Added Osmiridium nanofab recipe. 1 Iridium Alloy + 1 Osmium -> 1 Osmiridium
```


Cherman0:
Same comment as on the 10 electrical material PR, if youre changing the bounds of a matsci stat you should go through everything that uses that stat to make sure the formulas actually work and produce sane results.

flappybatpal:
I'm not sure about the matsci implications or the benefit of having a best in class metal, which as mentioned is actually beyond the code bounds for materials.

PR #23066, opened 2025-04-20
Name weak buckshot "muckshot"

JORJ949:
## About the PR 
Name weak buckshot "muckshot"

## Why's this needed? 
The weak buckshot has no indication to the player unless you know the source you got it from is the weaker version, "muck" because its worse than the buckshot.


DisturbHerb:
muck

TobleroneSwordfish (approved):

ZeWaka (approved):
MUCK

PR #23067, opened 2025-04-20
Add uranium ore and associated recipes

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Adds 
-Yuranite (uranium ore) to rare metal mining events
-Neutrite recipe (uranium + plutonium -> neutrite)
-Neutronium recipe (neutrite + erebite -> neutronium)

Removes:
-Neutronium metal from rare metal mining events

## Why's this needed? 
More intermediate metals/recipes, creates sense of progression in obtaining something harder to get

## Changelog 

```changelog
(u)FlameArrow57
(+)Neutronium metal is no longer found in rare metal mining events
(+)Added yuranite (uranium ore) to rare metal mining events
(+)Added recipes for neutrite (1 uranium + 1 plutonium -> 1 neutrite) and neutronium (1 neutrite + 1 erebite -> 1 neutronium)
```


Nexusuxen:
As an avid nuclear enjoyer, this is an interesting addition! The decision to choose between cerenkite and uranium for fuel is definitely a welcome change, since all other nuclear fuels (except erebite, marginally) are just not worth using. I like that uranium inherently has some neutron radioactivity, but has less normal radioactivity, meaning cerenkite is still ideal if you want to min/max plutonium - yet uranium can act as a great neutron source for breeding plutonium.

I assume making Plutonium alloyable was necessary to make this whole thing possible to implement. It'll be interesting seeing what comes out of plutonium alloying (though, granted, matsci is kinda weak as-is so I don't expect much)

444Portal:
what about the uranium (reagent) pills/nuggets you can buy from sketchy? they should probably be changed somehow - let them still be extractable but  also carry that material?

FlameArrow57:
> what about the uranium (reagent) pills/nuggets you can buy from sketchy? they should probably be changed somehow - let them still be extractable but also carry that material?

Think it may be fine, considering there's Plasma reagent and then plasmastone which is "solid form of plasma"?

Nexusuxen:
After some more thought, I think I'd personally prefer to see the reactor as a means of creating material, particularly neutronium. One such method may be to include a new "Fertility" property to certain materials, which is consumed in a reactor (similar to Radioactivity) to accumulate some new material, which can be centrifuged out. Plasmastone is a prime candidate for plutonium fuel breeding, though I've considered both koshmarite and erebite as a means of creating neutronium in this manner. (Erebite in particular is interesting here, as it's very tricky to manage in a reactor)

That aside, I would like to note that "Uranium" feels kinda jarring as a name, compared to Cerenkite and Mauxite and etc. I don't have any great suggestions unfortunately, besides maybe "Ursanite" or something.

pgmzeta:
uraninite ore -> uranium bar makes more sense to me as a raw material

Torbernite looks fucking awesome though
Torbernite-120981

PR #23070, opened 2025-04-20
AI-only cameras use separate network 

pgmzeta:
[rework][
## About the PR 
* Replace the `ai_only` camera variable with a new "ai" camera network
* Let AIs see the network
* Give AI cameras a distinct recolor

## Why's this needed? 
Now that all camera viewers use a list of camera networks, instead of a separately tracked variable we can instead place them on an isolated camera network.

## Screenshot
Atlas AI area showing color-tinted cameras
image

### DIFF
--- a/code/modules/camera/monitor.dm
+++ b/code/modules/camera/monitor.dm
@@ -12,8 +12,7 @@ TYPEINFO(/obj/item/device/camera_viewer)
 	var/obj/machinery/camera/current = null
 	// Sin but we need to know for disposing to clear viewer list on current
 	var/mob/last_viewer = null
-	var/can_view_ai = FALSE
-
+	
 	disposing()
 		src.disconnect_user(src.last_viewer)
 		..()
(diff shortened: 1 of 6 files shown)


code/modules/camera/monitor.dm
```diff
@@ -57,9 +56,7 @@ TYPEINFO(/obj/item/device/camera_viewer)
 
 		for (var/obj/machinery/camera/camera as anything in cameras)
 			if (camera.network in src.camera_networks)
-				if (camera.ai_only && !src.can_view_ai)
-					continue
-				displayed_cameras[text("[][]", camera.c_tag, (camera.camera_status ? null : " (Deactivated)"))] = camera
+								displayed_cameras[text("[][]", camera.c_tag, (camera.camera_status ? null : " (Deactivated)"))] = camera
```
JORJ949, line 59:
```suggestion
				displayed_cameras[text("[][]", camera.c_tag, (camera.camera_status ? null : " (Deactivated)"))] = camera
```
Twice as many indents as needed

JORJ949 (approved):
Really like them having a colour to indicate that only the AI can see through them.

PR #23075, opened 2025-04-21
Add highly conductive rare metals

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Adds 
-Veranium metal as a metal from the rare metal mining event. Shocks you slightly when held.
-Voltite metal recipe (electrum + veranium -> voltite), which is a 10 conductivity metal. Shocks you when held and arcflashes to you or nearby turf

## Why's this needed? 
Way to get 10 conductivity metal

## Changelog 

```changelog
(u)FlameArrow57
(+)Added veranium to the rare metal mining event
(+)Added voltite metal (1 electrum + 1 veranium -> 1 voltite)
```


Cherman0:
Uh, 10 conductivity? The bounds of material stats are supposed to be 1 through 9, this will break things

Cherman0:
For the record I feel like lowering the range of material stats from 1 - 100 all the way to 1 - 9 was a mistake, but a decent amount of editing of existing things that use matsci was required to do that and changing the bounds again neccesitates doing that again, or allowing various formulas to have unexpected behaviour.

flappybatpal:
In this case there is already Electrum so this would be fairly similar except having thermal 6. I'm not sure about the implications of this balance wise.

I quite like the shock behaviors so I think this is an interesting one.

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -793,3 +793,22 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
...
+	execute(mob/living/L, obj/item/I, mult)
+		if (ON_COOLDOWN(I, "voltite_shock", rand(60, 80)))
+			return
+		if (istype(L))
+			L.shock(I, 100, "All", 1, TRUE)
+
+		if (prob(10) && istype(L))
+			arcFlash(I, L, 500, 1.5)
+		else
+			arcFlashTurf(I, pick(block(L.x - 5, L.y - 5, L.z, L.x + 5, L.y + 5, L.z)), 500, 100)
```
pgmzeta, line 814: should be generic shock and arcflash `materialProc`s that take in a wattage/power level in `New`

code/modules/materials/Mat_Materials.dm
```diff
@@ -574,6 +574,34 @@ ABSTRACT_TYPE(/datum/material/metal)
...
+
+/datum/material/metal/voltite
+	mat_id = "voltite"
+	name = "voltite"
+	desc = "Energy seems to be flowing around it, chanelled through in an unknown manner."
+	color = "#389fff"
+
+	New()
+		..()
+		setProperty("electrical", 10)
```
pgmzeta, line 599: *Nothing* else has 10 for property values aside from plasma offgas. I get it is trying to be an improvement, but if this is meant to provide a benefit to things perhaps make them react to shock/arcflash instead, with the unique material properties of shocking/arcflashes providing that regularly.

pgmzeta (changes requested):
i like the shock/arc on life, but i think breaking the existing ranges just for conductivity is not a good change

FlameArrow57:
I readjusted so that instead of giving 10 conductivity it gives 1 thermal, meaning it is a better material than fibrilith for the TEG

code/modules/materials/Mat_Materials.dm
```diff
@@ -574,6 +574,35 @@ ABSTRACT_TYPE(/datum/material/metal)
...
+/datum/material/metal/veranium
+	mat_id = "veranium"
+	name = "veranium"
+	desc = "It looks to be sparking."
+	color = "#8effdd"
+
+	New()
+		..()
+		setProperty("electrical", 4)
+		setProperty("thermal", 4)
```
pgmzeta, line 586: i think setting thermal here to 1 as well would be a good hint that it has outstanding thermal properties. A little hand-wavy but "Material that stores excess thermal energy as electrical charge" might make a good lorenote/hint it for voltite.

Speaking of, may want to update the matsci in-game handbook for veranium?

```suggestion
		setProperty("thermal", 1)
```

pgmzeta (approved):

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -793,3 +793,40 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
...
+		src.wattage = wattage
+
+	execute(mob/living/L, obj/item/I, mult)
+		if (ON_COOLDOWN(I, "material_arcflash", rand(cd_min, cd_max)))
+			return
+
+		if (prob(10) && istype(L))
+			arcFlash(I, L, src.wattage, 1.5)
+		else
+			arcFlashTurf(I, pick(block(L.x - 5, L.y - 5, L.z, L.x + 5, L.y + 5, L.z)), src.wattage, 100)
```
pgmzeta: arcflashes completely break when they're not going from something on a turf to a target. this was leading to some weird bugs when testing where the arcflashes would get stuck offscreen and always target to the top-right.

After debugging some, I think this captures the idea behind your intent and work as you intend.

```suggestion
		if (!istype(L))
			return

		if (ON_COOLDOWN(I, "material_arcflash", rand(cd_min, cd_max)))
			return

		if (!isturf(L.loc) || prob(10))
			L.shock(I, src.wattage, "All", 1.5, TRUE)
		else
			arcFlashTurf(L, pick(block(L.x - 5, L.y - 5, L.z, L.x + 5, L.y + 5, L.z)), src.wattage, 100)

```

FlameArrow57: ty

Scaltra-Volpe:
I know this just closed and was already merged but I should say that there is a clamp value that limits fibrilith already for the TEG let alone this new one so they would have identical benefits. It is a cool material and I am all for it but if you want to see any difference for TEG semiconductor plating you should probably rework its formula to have a higher cap.

PR #23090, opened 2025-04-22
Integrates butts into assemblies and replace buttbombs with trigger/igniter/butt assemblies 

M-Earthfire:
[rework][game object][feature]

## About the PR 

This PR removes buttbombs and fully integrate the addition of butts into (ass)emblies. They can either be added as appliers or as targets when the applier is an igniter.

grafik

Butts can be used as appliers, making a fart sound and emiting a small amount of fart gas into the air.

If butts are attached onto igniter-assemblies, they will explode like buttbombs did. This explosion will also emit an amount of fart gas.

## Why's this needed? 

Buttbombs were still waiting their integration into assemblies. This also enabled the integration of butts on their own into assemblies.

With the fart gas creation, this also enables some new engineering designs for burn setups. I'm eager to see what kind of cursed designs people come up with.

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Buttbombs have been removed. Instead, butts can be attached onto trigger/igniter assemblies for the same effect.
(+)Butts can be attached with timers/mousetraps/prox sensors/health analysers/signalers to fart when activated
(+)Butts in assemblies emit small amounts of fart gas when triggered.
```


TobleroneSwordfish (changes requested):
Looks good, only issue is synthbutts don't use the correct icon when made into an assembly.

PR #23093, opened 2025-04-22
Moves most cooking-related code into a single module

RubberRats:
[CODE QUALITY]
## About the PR 

This PR moves several files related to cooking and recipe code into a new folder under modules. Also moves bath bombs out of the food_and_drinks module, because they definitely aren't food.

## Why's this needed? 

Cooking code was scattered across several different files in disparate spots of the codebase. As the guy who committed to fixing this, I would like it if they were more organized.


flappybatpal (approved):
Looks like a better approach.

ZeWaka (approved):
+1 zewaka point for using this

PR #23099, opened 2025-04-23
Remove random nukie frequency

JORJ949:
## About the PR 
Removes the randomly generated frequency from syndicate headsets in nukeops

## Why's this needed? 
They already have a syndicate frequency on :z which outside of nukeops is the default set frequency for these headsets, nukeops having a randomised one means that some may want to talk to/listen in on the crew and change it to the general frequency, while their teammates try to coordinate on the randomised one. The default frequency being the syndicate one remains forcing you to change to general if you want it while also keeping you in the loop if you do via the syndicate channel still on the headset.


ZeWaka:
Speech rework :(

TobleroneSwordfish (approved):

PR #23105, opened 2025-04-23
New map: Neon Deepwater Research Facility - Removes Atlas from rotation

Sord213:
## Neon Deepwater Research Facility

(this overall picture is out of date)
StrongDMM-2025-04-23 18 25 50
Screenshot_89

During the Manta expedition NT discovered a section of Abzu's ocean that contained large coral reefs made out of plasma. While surveying the area they also discovered the local fauna was experiencing mutations due to the plasma in the water. Growing larger than expected, gaining unique abilities, and increased aggression. NT established the Neon Deepwater Research Facility to research the plasma coral reefs, experiment on the effects of plasma on the local fauna and discover the source of the plasma in the ocean.

The existence of megafauna in the area was not disclosed to the crew to keep panic down, but occasionally they're seen passing overhead…dangerously close to the station.

## Map features
- Intended for lower pops. 10-30 range. Replace Atlas as the small lowpop map
- Plentiful maint and crime spots for the small size
- Also messy maint and bathrooms. Give janitors something to do at roundstart until other messes start appearing naturally.
- Central cafeteria/bar
- Research outpost, which is public access, for crime
- Plasma coral reefs. Plasma coral itself is made of plasmastone when cut up. Plasma coral only exists in the plasma coral reef.
Screenshot_88
- Large aggressive jellyfish inside of the plasmacoral reef
Large jellies have a unique bioeffect that causes them to metabolize solid/liquid plasma into electricity which can be collected by geneticists and given to humans
- Bottom portion of map is designed to be high density player area, with upper portions of map set up to be visited less
- Utilizes ocean currents as a secondary power source.
Currents are easy to set up and similar in effect to solars. Require minimal intervention after setting up and generate small, but more consistent, amounts of power
Screenshot_86
Screenshot_87
- Currents also contain random debris and bubbles of gas.
- Bioluminescent algae grows on some of the outside walls.
- Gas vents generate around the map (more commonly in the plasma reefs) which spawn bubbles of a random gas per vent.
- All gas vented underwater now becomes bubbles. This a little experimental and subject to being hit with the performance mallet.
- Big parallax sea creature shadows occasionally cross overhead
Applies a temporary eldritch madness antag status for some crew, given a common objective, timer reduced by haloperidol.
Spooky sound effects stolen from whales
Screenshot_91

## Why's this needed? 
Atlas as a lowpop map barely functions when there's 20+ players on it. While it now has the terrainify random chance there's nothing else unique about it. Atlas is very hard to play almost any antagonist role on due to the small size and tiny amounts of maintenance. Neon has lots of maint and plenty of crime rooms. Many departments on Neon also have privacy blinds.

## Credits:
Audio credits for the megafauna event:
https://pixabay.com/sound-effects/monster-footstep-162883/
https://pixabay.com/sound-effects/horror-sound-lurking-horror-monster-189948/
https://pixabay.com/sound-effects/horror-sound-monster-breath-189934/
https://pixabay.com/sound-effects/creepy-whale-song-323612/
https://pixabay.com/sound-effects/haunting-whale-song-323611/
https://pixabay.com/sound-effects/alarm-301729/

## Changelog 

```changelog
(u)Sord, Leah, Jan + seamonsters by Thrush
(*)Added a new underwater map, Neon. Neon replaces Atlas as the lowpop map (Max votable playercount 30).
(*)Neon is built within a strange reef where the coral is made of plasma. The plasma has been having strange effects on the wildlife.
(*)All gas vented underwater now becomes bubbles. Bubbles eventually "pop" and dissipate. Large bubbles can be breathed from. Bubbles can be captured using portable pumps or air scrubbers.
(*)Neon comes with a new unique rare antag type, the "Broken", which can spawn via a random event.
(*)Neon features a new solars-like power generation method, the current driven Turbine. Wrench turbine shafts and use one of the four turbines located around maintenance to find the sweet spot in the current to generate the small amounts of power.
(+)There are unique mobs located within the plasma coral reefs which contain a new genetic mutation.
(+)Currents outside the north and south side of the station can be counteracted with a jetpack or magboots.
(+)Any non antagonists, regardless of mindhack protection implants or traits, can become "Broken"
(+)To opt into becoming the "Broken" you can enable "Other Foes" in your character preferences.
(+)All "Broken" share the same randomly picked objective.
(+)Haloperidol can be used to cure the "Broken" faster.
```


Laboredih123:
3 years ago?????????????????????????????????????

TobleroneSwordfish:
> 3 years ago?????????????????????????????????????

This is a git moment, the original branch was made a while back but it's only been worked on recently.

Laboredih123:
anyway guess i gotta review this shit uh

cheekybrdy:
I love the idea actually, especially with the secondary power sources in. My only concern is with this replacing atlas and not being a space map.

JORJ949:
Looks really good, few things I spotted:
image
All turret controls should use the appropriate subtype for their area, these subtypes already have the proper access set on them.
image
Antibio crate should be in the armory
Bottom right section of armory feels a bit off, but its hard to put my finger on why.
image
A lot of the basic /obj/landmarks around now have subtypes that should be used instead
There is no sylvester.
Engi and sec announcement computers should be in the relevant command office, only the command member can use them.
image
Computer core should be general heads access rather than RD
image
Gloves are probably meant to be the syndicate version, which used to be this type until recently
Brig needs a toilet to swirly crimers in
Headset crate should probably use the premade type unless you are wanting to split out the PDAs at which you are missing the base command headsets.
AI core door should not start bolted
Cargo checkpoint needs access spawners
There is a comms dish in command teleporter
HoP, RD and MD need their first bill, diploma and license
image
Chef serving table is rather small, perhaps it could curve round like this
The only ABCU things are on the ghostdrone factory, would be good to have on station.
Northwest current generator only has engineering access on the inner airlock door, should either be for the room or not at all.
image
Random HoS access spawner here
Bridge chairs should be colour coded
Bridge has no security or medical record computers
the AI has no eyebots
AI chamber should have a foyer area with the turret control so you can change the turret options without getting killed
AI chamber could probably also use a cyborg dock, nearest is down in secmaints (when you consider the only entrance being the main bridge one)
Listening post should be split up into the appropriate areas

Laboredih123 (commented):
1. radio station lacking compared to todays new radio station
2. is the poolburn meant to intake gases? cause the pump is pumping out of the loop and loop starts empty so poolburn cant actually occur.
3. the pumps instead of valves on the toxins loop doesnt feel right. did you mean this to also be a cold loop like atlas?
4. the long range teleporter functionality was gained by telesci, leaving the lrt redundant
5. the clown hole has plasmaglass windows, far too generous smh smh
6. the view directly into the bathroom on captain area has no blinds to stop such a thing
7. the kitchen freezer has no cooling loop.
8. mechanics is pretty far away from the rest of engineering, maybe swap it with mining? and then swap arrivals and mining so arrivals gets dropped right next to the bar and mining is right next to the mining podbay and elevator.

code/datums/objective.dm
```diff
@@ -649,7 +649,42 @@ ABSTRACT_TYPE(/datum/multigrab_target)
 
 		return 0
 
+ABSTRACT_TYPE(/datum/objective/madness)
+/datum/objective/madness
 
+/datum/objective/madness/hate_department
+	set_up()
+		src.explanation_text = "Your mind fills with a burning need to destroy [pick("Genetics", "Genetics", "Catering", "Botany", "all janitors", "Medbay", "Cargo", "the research department")]."
```
JORJ949, line 657: May be on purpose but genetics is here twice

code/map.dm
```diff
@@ -1246,6 +1247,54 @@ var/global/list/mapNames = list(
...
+	shuttle_map_turf = /turf/space/fluid
+
+	merchant_left_centcom = /area/shuttle/merchant_shuttle/left_centcom/destiny
+	merchant_left_station = /area/shuttle/merchant_shuttle/left_station/destiny
+	merchant_right_centcom = /area/shuttle/merchant_shuttle/right_centcom/destiny
+	merchant_right_station = /area/shuttle/merchant_shuttle/right_station/destiny
+
+	valid_nuke_targets = list("the bridge" = list(/area/station/bridge),
+		"the quartermaster's office" = list(/area/station/quartermaster/office),
+		"the warehouse" = list(/area/station/storage/warehouse),
```
JORJ949, line 1291: warehouse is right next to the qm office, perhaps robotics would be a good alternative target for the west of the map.

code/modules/events/madness.dm
```diff
@@ -0,0 +1,148 @@
...
+	customization_available = TRUE
+	var/num_victims = 0
+
+	admin_call(source)
+		. = ..()
+#ifndef MAP_OVERRIDE_NEON
+		boutput(usr, "The map is not Neon, aborting.")
+		UNLINT(return)
+#endif
+		src.num_victims = input(usr, "How many minds to break?", src.name, 0) as num|null
```
JORJ949, line 18: tgui_input_num

code/modules/events/madness.dm
```diff
@@ -0,0 +1,148 @@
...
+			sleep(rand(20, 60) SECONDS)
+			src.sound_event()
+		animate(monster, alpha = 0, time = 10 SECONDS)
+		sleep(10 SECONDS)
+		qdel(monster)
+
+	proc/cause_madness(source)
+		var/list/potential_victims = list()
+		for (var/mob/living/carbon/human/H in global.mobs)
+			if (H.client && !H.mind?.is_antagonist() && !isVRghost(H) && H.client.preferences.be_misc && isalive(H)) //using "misc" prefs for now
```
JORJ949, line 80: Intended lack of allow_traitors check? Will allow Cap and sec to roll broken.

Sord213:
> Bridge has no security or medical record computers
they're in the radio room
> AI chamber should have a foyer area with the turret control so you can change the turret options without getting killed
i prefer it this way. see: donut2
rest of these are sorted, minus code things

> 1. radio station lacking compared to todays new radio station
> 
>     2. is the poolburn meant to intake gases? cause the pump is pumping out of the loop and loop starts empty so poolburn cant actually occur.
> 
>     3. the pumps instead of valves on the toxins loop doesnt feel right. did you mean this to also be a cold loop like atlas?
> 
>     4. the long range teleporter functionality was gained by telesci, leaving the lrt redundant
> 
>     5. the clown hole has plasmaglass windows, far too generous smh smh
> 
>     6. the view directly into the bathroom on captain area has no blinds to stop such a thing
> 
>     7. the kitchen freezer has no cooling loop.
> 
>     8. mechanics is pretty far away from the rest of engineering, maybe swap it with mining? and then swap arrivals and mining so arrivals gets dropped right next to the bar and mining is right next to the mining podbay and elevator.

1. I considered updating it to look more like the other new radio station but decided not to. The other spawns (actor, etc) are on diner. Can always change later but not a priority.
2. honestly just a dumb gimmick. i can try to make it a little better ig
3. toxins probably could use some work, but I haven't touched it in ages
4. oh right
5. weh, the clown just stole a random room's space
6. yes
7. weh, no room. also atlas' doesn't have a proper cold room either, low priority
8. mechanics were still their own job when i first made the map. cooould swap things around but arrivals is on the "low human density" part of the map for a reason, less action to immediately die to when you step out

cheekybrdy:
No time for a in depth at the moment, but botany nor ranch has a UV light

Arahimine:
two small things:
1. a lot of the blinds need another pass, multiple just exclusively use a single sprite for the whole length (most noticeable in spots like the janitor's and MD's office)
2. the lack of a psych office is :(. the medbay restroom seems like it could be a good spot for that, they're both borderline-flavour rooms on a map that's already chock full of crime spots and i think a room for a gimmick role would be a better use of the space than a second instance of a room that already exists elsewhere (the restroom up north)

Sord213:
> two small things:
> 
>     1. a lot of the blinds need another pass, multiple just exclusively use a single sprite for the whole length (most noticeable in spots like the janitor's and MD's office)
> 
>     2. the lack of a psych office is :(. the medbay restroom seems like it could be a good spot for that, they're both borderline-flavour rooms on a map that's already chock full of crime spots and i think a room for a gimmick role would be a better use of the space than a second instance of a room that already exists elsewhere (the restroom up north)

Sorted the blinds. 

There likely wont be a psych office, sadly. No room for it and I'd prefer to keep the restroom there.

code/datums/objective.dm
JORJ949: Intended

code/modules/events/madness.dm
JORJ949: image

code/modules/events/madness.dm
TobleroneSwordfish: No reason to, it's an admin input so the extra safety is better.

cheekybrdy:
Didn't actually find anything after a big sweep that's a glaring error I'd just say put some loot at the end of the trench trail in the bottom left and some of the random room trench parts that are south of the main station in nadir for some flavour.

JORJ949:
> 
> There likely wont be a psych office, sadly. No room for it and I'd prefer to keep the restroom there.

I think psych office could be a really great part of the map considering the Broken antagonist. It doesn't necessarily have to be in medbay either it could be stuck on escape or something

flappybatpal:
> I love the idea actually, especially with the secondary power sources in. My only concern is with this replacing atlas and not being a space map.

Yeah I would say I don't know about replacing our only low pop space map with a low pop sea map as it cuts off a lot of space content and has a different playstyle.

flappybatpal:
image
At first sight it's not clear what this room is for. It took me using the terminal to realise it was the sea elevator.

Some rooms look to have underwater parallax?
- The random room below Beepsky's bedroom.
- The RD's office.

JORJ949:
image
I'd move one of these lights to this corner
image
Turret controls don't sync if theres multiple to an area, only the settings on the most recently edited one is applied
Security pod bay needs a couple siren parts
Clown hole is missing the crate and announcement computer
image
Don't use the detnet intercoms, these are the actual security radio frequency used in detnet VR so the detective can still hear sec comms while inside
Maintenance airlock by cargo checkpoint missing access spawner
Robotics surgery table needs a drain under it
image
image
Upside down Brig controls, these should all be flipped.
Neither of the tracking beacons in security can be teleported to as the area is teleblocked, one should be in the brig as the brig is not.
There should generally only be one spare killswitch circuit board, I would perhaps move the actual switch to computer core so the AI doesn't just have one room to defend if rogue.
Would remove the lamp next to the law rack, rooms a bit too bright with it there
image
This camera gives a lot of maints coverage
image
As do these two in brig and this one outside

JORJ949:
image
Cubicle panels don't appear to be playing nice with vision

JORJ949:
Listening post doors still need "Sleeper_Access" id and the entrance scanner should have entrance_scanner = 1, and the interior syndicate airlock to the teleporter should also have the sleeper access ID or no access requirement or it will be locked to syndicate access always, though all maps should probably just remove this door requiring access.

Sord213:
yep. that's just how they do

JORJ949:
image
image
Still a lot of camera coverage in science maintenance, though im not sure where the computer core coverage is coming from.

code/datums/objective.dm
```diff
@@ -649,7 +649,42 @@ ABSTRACT_TYPE(/datum/multigrab_target)
...
+				//might be a little hard to kill this one but also we're insane so it fits
+				valid_targets |= "Morty, the morgue opossum"
+			else if (istype(M, /mob/living/carbon/human/npc/monkey/stirstir))
+				valid_targets |= "Monsieur Stirstir, security's prisoner"
+			else if (istype(M, /mob/living/silicon/ai))
+				var/mob/living/silicon/ai/ai = M
+				if (ai.client || ai.deployed_to_eyecam || ai.deployed_shell) //they're probably logged in somewhere
+					valid_targets |= "[M.real_name], the AI"
+			else if (istype(M, /mob/living/carbon/human))
+				if (M.job == "Captain" && M.client)
```
JORJ949, line 683: I think the clown should be a valid madness target >:)

Arahimine:
a bunch of bushes (specifically ones with dir 1, there's one above security, one above the tool storage, a bunch more around) have their 'interesting' variable set to something about styptic powder. i assume this is a leftover from reusing them from somewhere. in any case it's annoying because they keep on triggering t-ray scanners

flappybatpal:
I'm not totally sure about the voting booth placement or the chemistry changes but there is limited room to work with.
Feels like chemistry might need to be one to two tiles longer/wider as there is a substantial amount of equipment available for modern chemistry.

Sord213:
>  Still a lot of camera coverage in science maintenance, though im not sure where the computer core coverage is coming from.

Guardbuddies are mobile cameras for the AI.

> bush interesting vars

fixed

> chemistry

expanded one tile wider

JORJ949:
Coffeemaker in engineering is the wrong kind, also command should have one.

Cherman0:
Based on a quick overview of the map, mostly focused on engi/mining/cargo since that's what I'm most familiar with:

- Mining is awkardly bisected by a maintenance tunnel. This makes it both difficult for a miner unfamiliar with the map to locate their hangar and annoying to ferry material between the hangar and rockbox. Could probably be rectified by swapping the main mining area with the cryo room. Mining also lacks a cargo pad for use with their cargo transporter.
- For being a job mainly about interacting with people, cargo is in an extremely lonely corner of the map. Not sure about a good way to solve this, the most practical thing that comes to mind is re-arranging the bridge to fit cargo between the bridge and security.
- The map has no MULEs or MULE beacons. Aside from them just being kinda neat, this will cause problems if soeone acquires a MULE (such as by buying it) since it will be impossible to move it.
- The inbound deliveries conveyor belt is facing the wrong way. This will cause clogs if there arent any QMs or the QMs dont know that it needs to be fixed on this map.
- Materials processing lacks a general fabricator. There is a general fabricator in tool storage, but it lacks a material processor.
- Engineering storage lacks construction supplies. Could probably move the mail chute into the lobby and remove the toxins vendor to make room for some racks with steel, glass, rods, and wires.

Other random things I noticed:

- There's a clown spawn point in materials processing. Unsure if it's intentional or what the reason would be for it if it is.
- The 3x3 random room by the AI core has an entrance offset from the center of its edges. This will cause either floating decals or it being blocked entirely for a lot of random room layouts, since they expect the entrances to be in the middle of each edge.
- Artlab ejector is in a very non-obvious location and has a bottleneck that will make it annoying to get rid of large artifacts. Given the small size of the map and that science is in a relatively populated corner, this will cause frequent round-ends and general problems.
- Kitchen/bar area spawning full of vomit and blood is a little weird.
- Bar is very small, even for a lowpop map. Not a whole lot to be done about it given its central position though.
- Subjective, but I was never really a fan of the research outpost on Oshan (Donut2/Cog2, for that matter). I feel like a lowpop map should try to bring people together as much as possible and having a forgotten little room to never see anybody is kinda opposite to that.
- The map uses a mix of /obj/landmark/spawner/artifact and /obj/landmark/artifact. Both probably work but I believe the former is the more current expected way to spawn artifacts.

PR #23110, opened 2025-04-24
Unmap DetNet intercoms

JORJ949:
## About the PR 
Remove detnet intercoms from Donut2 and Mushroom

## Why's this needed? 
These are not meant to be mapped, they have the actual security and detective frequencies and live in detnet VR so the detective can still hear sec comms while inside.

### DIFF
--- a/maps/donut2.dmm
+++ b/maps/donut2.dmm
@@ -24215,12 +24215,6 @@
 	},
 /area/station/quartermaster/office)
 "hje" = (
-/obj/item/device/radio/intercom/detnet/security{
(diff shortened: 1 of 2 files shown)


Sord213:
I like having the detective one because it lets people hear if the detective has spy stickers in place

JORJ949:
> I like having the detective one because it lets people hear if the detective has spy stickers in place

That's actually cool yeah, will add those around instead.

JORJ949:
Will do intercom stuff post speech rework

PR #23113, opened 2025-04-24
TGUI Respawn as Job

JORJ949:
## About the PR 
Changes the admin respawn as job command to use a tgui input list

## Why's this needed? 
Searchability better than just spamming the first letter of the job until you find the one you want. Will help with debugging a bunch.
imageimage


TobleroneSwordfish:
Buut this prevents you from typing out the job name in the command bar which is presumably the intended way of using this command?

JORJ949:
I hate command bar jank this stinks >:(

PR #23137, opened 2025-04-25
AI law datums

RubberRats:
## About the PR 

This PR atomizes the law datum code out from #22547, meaning cyborgs are no longer directly tied to the law rack.

## Why's this needed? 

Regardless of dev opinion on my previous borg PR, I still think breaking cyborg laws away from being tied to a physical object is a good thing. The current emag behavior is still less than ideal, and this should allow for the easy modification of law code for future PRs.

## Changelog 

```changelog
(u)RubberRats
(*)Silicon law code has been changed internally, please report any weird behavior.
(+)DNR observers have a toggleable option to view silicon laws on examine
```

### DIFF
--- a/code/mob/living/silicon/robot.dm
+++ b/code/mob/living/silicon/robot.dm
@@ -1099,8 +1099,8 @@ TYPEINFO(/mob/living/silicon/robot)
 				boutput(user,"You need to use this on the AI core directly!")
 				return
 
-			if(!src.lawset_connection?.host_rack)
-				boutput(src,"[src.name] is not connected to a law rack")
+			if(!src.lawset_connection || !src.lawset_connection.host_rack)
+				boutput(user, "[src.name] is not connected to a law rack")
 			else
 				var/area/A = get_area(src.lawset_connection.host_rack)
 				boutput(user, "[src.name] is connected to a law rack at [A.name].")


code/world/topic.dm
```diff
@@ -670,14 +670,23 @@
...
-				if (current_state > GAME_STATE_PREGAME)
-					var/ircmsg[] = new()
-					ircmsg["laws"] = ticker.ai_law_rack_manager.format_for_logs(glue = "\n", round_end = TRUE, include_link = FALSE)
-					return ircbot.response(ircmsg)
+				var/ircmsg[] = new()
+				if(plist["target"])
+					var/who = lowertext(plist["target"])
+					var/mob/living/silicon/S = ckey_to_mob_maybe_disconnected(who, exact = FALSE)
+					if(S)
+						ircmsg["laws"] = S.lawset_connection.format_for_logs(glue = "\n")
```
JORJ949: S.lawset_connection may be null here if they don't have laws, will runtime.

code/datums/ai_laws.dm
```diff
@@ -149,22 +149,144 @@
...
 
 		return jointext(laws, glue)
+
+/datum/ai_lawset
+	var/list/current_laws[LAWRACK_MAX_MODULES]
+	/// used during UpdateLaws to determine which laws have changed
+	var/list/last_laws[LAWRACK_MAX_MODULES]
+	var/obj/machinery/lawrack/host_rack = null //law rack this datum is tied to. Might not exist
+	New()
+		. = ..()
```
JORJ949: This new just calls the parent New(), no need to have this here.

code/mob/living/silicon/robot.dm
```diff
@@ -1096,10 +1108,10 @@ TYPEINFO(/mob/living/silicon/robot)
 			if(linker.linked_rack in ticker.ai_law_rack_manager.registered_racks)
 				if(src.emagged || src.syndicate)
 					boutput(user, "The link port sparks violently! It didn't work!")
-					logTheThing(LOG_STATION, src, "[constructName(user)] tried to connect [src] to the rack [constructName(src.law_rack_connection)] but they are [src.emagged ? "emagged" : "syndicate"], so it failed.")
+					logTheThing(LOG_STATION, src, "[constructName(user)] tried to connect [src] to the rack [constructName(src.lawset_connection)] but they are [src.emagged ? "emagged" : "syndicate"], so it failed.")
```
JORJ949: How does lawset_connection show in text? Does it show the connected rack or an ID if it has no rack?

code/obj/machinery/law_racks.dm
```diff
@@ -480,138 +480,26 @@
...
+				laws[i] = null
+		return laws
+
+	proc/UpdateLaws()
+		for(var/mob/living/silicon/ai/holoAI in mobs)
+			holoAI.holoHolder.text_expansion = src.holo_expansions.Copy()
+			src.reset_ai_abilities(holoAI)
+		calculate_power_usage()
+		lawset.current_laws = format_as_list()
+		lawset.UpdateLaws()
```
JORJ949: src

code/obj/machinery/law_racks.dm
```diff
@@ -480,138 +480,26 @@
...
-	proc/update_last_laws()
-		var/law_counter = 1
-		for (var/i in 1 to MAX_CIRCUITS)
-			var/obj/item/aiModule/module = law_circuits[i]
+		for (var/i in 1 to LAWRACK_MAX_MODULES)
+			var/obj/item/aiModule/module = law_modules[i]
 			if (module)
 				//save the law text and the displayed law number (not the rack position)
-				last_laws[i] = list("law" = module.get_law_text(TRUE), "number" = law_counter++)
+				laws[i] = list("law" = module.get_law_text(TRUE), "number" = law_counter++)
```
JORJ949, line 491: Does this maintain current functionality of having them be numbered in order no matter how many laws are on one module or how many gaps are in the rack?

JORJ949 (changes requested):

code/obj/machinery/law_racks.dm
RubberRats: image
Just confirmed they are working as intended

code/mob/living/silicon/robot.dm
RubberRats: Somehow completely missed that it's supposed to show the linker's rack here, not the borg's. Or maybe that bug was there previously. Honestly it's long enough ago that I can't remember.

PR #23138, opened 2025-04-25
Add manufacturable lanterns

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Adds electric lanterns, which light up your vicinity, rather than directionally

Sprites by Mr. Moriarty

Lantern:
image

Flashlight:
image

## Why's this needed? 
There aren't many light sources that are immediate light up in the vicinity type lights, they are mostly directional, away from you lights

## Changelog 

```changelog
(u)FlameArrow57 & Mr. Moriarty
(+)Added manufacturable lanterns to fabs! They light up your immediate vicinity rather than directionally like flashlights
```


TobleroneSwordfish (changes requested):
The sprites are really pretty but I have to agree with Jan and say they look too much like a metal framed candle lantern to be presented as an electric station one. It probably wouldn't take much to tweak the sprite to look more electric?
Also should have unique inhands because the flashlight ones are very obviously a blue flashlight.

PR #23141, opened 2025-04-25
Unglue anything when an object is deconstructed

flappybatpal:
## About the PR 
Some probably ugly code that will unglue anything attached to an object that is deconstructed to a frame.

## Why's this needed? 
Prevents frame based crimes.

## Changelog 

```changelog
(u)Flappybat
(+)Deconstructing an object will unglue anything that is glued to it.
```


TobleroneSwordfish (changes requested):
Looks good, but you don't need to keep casting to `atom`. `/obj` and `/turf` are all `/atoms` by definition.

PR #23144, opened 2025-04-25
Add office for Glowbold

pgmzeta:
[mapping][feature]
## About the PR 
adds an area and office for glowbold

## Why's this needed? 
office :3


JORJ949:
File diff shows cairngorm changes, but mapdiff doesn't, would double check cairngorm is still up to date with #23002

pgmzeta:
> File diff shows cairngorm changes, but mapdiff doesn't, would double check cairngorm is still up to date with #23002

looks good to me
image

ZeWaka:
I assume just a shared key

ZeWaka:
I assume just a shared key

PR #23152, opened 2025-04-26
Slightly buff glass pipeshot + boneshot

TDHooligan:
[BALANCE]
## About the PR 
Changes spreadshot dissipation_variance to correlate with projectile speed, which looks a little more rational.

Changes regular glass pipe shot to fire lots more pellets with lower damage. They also cause double the bleed and a flat blood penalty, unlike most pipeshot's 1 bleed.
This gives it a niche of being good at terrorizing crowds and stacking up debuffs. It still does respectable damage up close.

Additionally buffs boneshot to shoot more than 3 tiles away.

## Why's this needed? 
Give glass pipeshot a fun niche, even if it's not super necessary or super strong. The projectile count shouldn't matter as much now pipeshot is restricted in what can load it.

## Changelog 

```changelog
(u)TDHooligan
(+)Glass pipeshot now shoots more pellets and causes heavier bleeding, at the cost of some damage.
(+)Bone pipeshot can now shoot 3x as far.
```

### DIFF
--- a/code/modules/projectiles/bullet.dm
+++ b/code/modules/projectiles/bullet.dm
@@ -1004,10 +1004,14 @@ toxic - poisons
 	name = "glass"
 	sname = "glass"
 	icon_state = "glass"
-	dissipation_delay = 2
-	dissipation_rate = 2
+	dissipation_delay = 4
+	dissipation_rate = 1
 	implanted = null
-	damage = 6
+	damage = 3
+	on_hit(atom/hit, dirflag, obj/projectile/proj)
+		var/mob/M = hit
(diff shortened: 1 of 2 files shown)


flappybatpal:
So what's the need for this, balance wise?

TDHooligan:
The off-meta shot types are dull, it just makes them a bit more interesting. Balance wise its' not touching anything super important compared to, say, plasma glass.

RubberRats (approved):
Code seems good. Can't really comment on balance, but IME almost nobody bothers trying to make glass or bone pipeshot.

pgmzeta (approved):
glass might be over-tuned given how easy it is to make and how bad bleeds are now, but it's probably fine for it to be a terror for a bit to check

PR #23160, opened 2025-04-26
Oshan Bar Remodel

MintyPhresh:
## About the PR 
Expands the Oshan bar, adding more seats, a kitchen counter, and revamping both the bar and the kitchen. Kitchen now has more storage and a streamlined work area; the bar has its own (huge!) back table for preparing drinks, and a seating shape that you might recognize.

Not shown clearly in this screenshot-- a red table, fiddle, and some produce has been added to the secret tunnel area in the corner of the kitchen.
image
And the old Oshan, taken from Goonhub, for your convenience and for future reference if this PR gets accepted.
image

## Why's this needed? 
Oshan's kitchen lacking a serving counter, and its bar lacking any drink prep tables for the bartender, can be pretty inconvenient for people who like to serve their food or prepare their drinks.  Bar having modern equipment & supplies is good too.

The fruit and fiddle in the hole are there so that someone can torment others with sounds (e.g. if you use it to evade sec, you can taunt them with them). Definitely out of scope but I like to mess around a lil ^u^


Cherman0:
I like the concept, though I feel that the entrance area with the entrances to the kitchen and bar is a bit awkward, being a junction of two one-tile pathways. Since the kitchen is much larger than behind the bar, it probably makes sense for the entrance to be through the kitchen into the bar rather than through the bar into the kitchen.

MintyPhresh:
Implementing something like that wouldn't do me any good-- that door would just get replaced with a wall, and then I'd need to reduce the amount of kitchen counter space to add a door on the west wall of the Cafeteria. Then, the bartender would be walking through kitchen workspace in order to get through to their workspace. It's tight, but it's an area where a little traffic bumping isn't super impactful (and also unlikely, since there's usually only 2-3 people using that area)

I considered just widening the area too, but that'd make it much less clear that the bar counter is supposed to be a bubble shape. I think the gameplay sacrifices are worth the style.

Cherman0:
This is how I envisioned it.
image
As youve said, the actual need for a change isnt very high because its not a very high-traffic doorway, but I feel like it looks nicer to have it less crammed.

Cherman0:
Also, what's the deal with the seats being all over the place and the knife and bloodstain? Seems more like a shady maint detail than something thatd be out in public

MintyPhresh:
I'd envisioned it being the mapping equivalent of Party Animal-- people got too boisterous while watching the fights in the ring down south and got a minor cut. The place feels a little more sterile without it, and it's got more character than generic garbage or grime.

MintyPhresh:
What I ended up doing. Widening the bar area isn't really necessary if the kitchen has their own door, IMO, and keeping the door southern fits walkflow a bit better on the chef side.
image

Cherman0:
Looks good to me. I think having that not be the main path into the kitchen alleviates the stress there, especially for a chef that might be carting things in and out. On the topic of the bloodstain, I dont really know that "the mapping equivalent of Party Animal" means but between the size of the bloodstain and knife and the amount displaced chairs I would assume that a major altercation went down here where someone was seriously injured or killed.

MintyPhresh:
I should use a smaller bloodstain, but ingame this level of blood isn't nearly enough to be life-threatening--  Intent was for it to be Five Finger Fillet gone wrong. picrel is from a single stab from the same knife
image
image
 Chairs were just supposed to be displaced by people who don't care how they leave a room-- see how a bunch are gathered around the place with the knife?

TobleroneSwordfish:
I like this, serving hatches are cool

flrsh (changes requested):
I would like to preserve the current symmetry and aesthetic of the bar. A lot of these changes, such as the shifting of the bar tables and the conglomerating of other seating, imo, just make it look way more like other maps, which, imo, makes it lose a lot of charm. If you'd like to expand the bar, I think you could expand leftwards and shift the kitchen accordingly. That is, if you need an extra column of space, take it from where the kitchen is now, and then correspondingly expand the kitchen into maintenance, and then expand maint into the ocean a column.

MintyPhresh:
I agree that there's value to the current bar setup-- it kind of looks like a rowboat that's just landed on the shore if you squint, which is sweet as hell.

That being said, the goal of this PR isn't to expand the bar; it's to fix problems that stem from its current, symmetrical design. Catering areas don't just need a lot of table space, they need a lot of _distinct_ table space, because there's four different basic functions table space fills:
- Preparing food
- Serving food
- Preparing drinks
- Serving drinks

Oshan's current design can't comfortably be given dedicated tables for the roles it's missing, even with additional length. 

The kitchen's all set on food prep space, but trying to add food serving space wouldn't work out because the only available spot in the 4 tiles of public-facing wall space (See image below) would have kitchen patrons climbing under bar patrons. Serving on the tables to the east isn't a great option either; although Cog1 manages fine with the chef stepping out to serve on the big dining table, here the chef needs to walk the entire length of the cafeteria as well as cross one of the map's busiest hallways to reach the tables, which makes the trek far more time-consuming and makes your rolling kitchen table thingy very vulnerable to being messed with, intentionally or otherwise. (inattentive secoff vs dense pushable object challenge!!! Can the kitchen table make it to ranch before the Chef catches up??)
image

The bar needs drink prep space, too. Placing objects on the floor is a pain, not just because it looks bad, but also because you need to either A) throw things to offset where they are on the tile (don't try this with pitchers!), B) drop it in the center and right-click to pick it up (good luck distinguishing between the stack if you didn't label them), or C) place it on a table to offset it, pick it back up, and then drop it with Q (this is convoluted and requires more brainpower than you'd think).
 
We aren't as hindered in adding drink prep space as we were for food serving space-- I've seen bartenders add tables behind the dispensers-- but this creates an extended one-tile zone where the chef and bartender would be bumping shoulders since they're both competing for that southern bar table space. Having to walk around the dispensers and through a one-tile space to reach the tables has a bad gamefeel, too, since you'd have to go completely to the other side to get them both within reach. Plus, that area'd probably just get the big dumb espresso machine moved onto there anyways, so really you'd only be getting like 1.5 more tiles of table.
image

It sucks when maps are samey, but Oshan doesn't feel great to cater on and the layout's a big part of that. I did try to keep the theming similar, with the same tile palette and the bubble bar layout, but if that's not enough we could always put some Manta-style dedicated theming into it. I think the proposed new layout lends itself well to something keyed off of blue holes, like the image below...
bluehole_gozo1_0

I'd also be fine with splitting the tables back to what they were before, though we'd be losing a bit of table surface area to do so.

pgmzeta:
are the screenshots in the PR description up to date?

MintyPhresh:
no, I'll update them here as I fix this merge conflict

MintyPhresh:
it is now!

PR #23168, opened 2025-04-27
Remove req_access_txt from assets/maps

JORJ949:
## About the PR 
Removes all req_access_txt defines from all /assets/maps

Only resulting change is the fridge in von ricken is no longer robotics access, though this probably isn't a huge change.

## Why's this needed? 
Code quality.


flappybatpal:
Could this leave some doors with no access reqs due to missing an access spawner?

JORJ949:
> Could this leave some doors with no access reqs due to missing an access spawner?

Only one of them was missing an access spawner and I gave it the accurate spawner

PR #23174, opened 2025-04-27
Make assemblies, suicide vests and mousetrap rollers respect contraband level of their components

M-Earthfire:
[Bug][Game Objects]

## About the PR 

This PR makes it so that the contraband level of assemblies, suicide vests and mousetrap rollers is set to the highest contraband level of items in the assembly

## Why's this needed? 

Pipebomb assemblies currently dont have a contraband level. I don't think that's a good idea to have.

```changelog
(u)Lord_Earthfire
(+)Assemblies, mousetrap rollers and suicide vests have the contraband level of their highest contraband component.
```

### DIFF
--- a/code/obj/item/assembly/misc_assemblies.dm
+++ b/code/obj/item/assembly/misc_assemblies.dm
@@ -398,7 +398,7 @@ Contains:
 	src.target = null
 	src.target_item_prefix = null
 	src.w_class = max(src.trigger.w_class, src.applier.w_class)
-	src.contraband = max(src.trigger.contraband, checked_item.applier.contraband)
+	src.contraband = max(src.trigger.contraband, src.applier.contraband)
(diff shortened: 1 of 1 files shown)


code/obj/item/assembly/misc_assemblies.dm
```diff
@@ -85,6 +85,7 @@ Contains:
 		checked_item.master = src
 		checked_item.layer = initial(checked_item.layer)
 		src.w_class = max(src.w_class, checked_item.w_class)
+		src.contraband = max(src.contraband, checked_item.contraband)
```
TobleroneSwordfish, line 88: This doesn't actually work, see `/datum/component/contraband` (I think there's still some spots where this is getting set directly elsewhere but they are Wrong and I tested it)

TobleroneSwordfish (changes requested):

PR #23176, opened 2025-04-27
Fix cogwip/kappa.dmm formatting

JORJ949:
## About the PR 
Fix formatting on maps/cogwip/kappa.dmm pipe direction varedits

## Why's this needed? 
Figured out why every minor kappa change resulted in a bunch of line changes, the formatting is wrong on the varedits.


JORJ949:
#23214 begone kappa

PR #23185, opened 2025-04-28
Cooking module

DisturbHerb:
[INTERNAL][CATERING][QOL]

## About the PR 

Moves as many kitchen-related files into their own `cooking` module as I could.

Repaths `/obj/item/tongs` to `/obj/item/kitchen/tongs` because it was bothering me.

## Why's this needed? 

Organises kitchen/cooking stuff for future ease of development.


DisturbHerb:
note to self: prolly put this under a catering module instead to include bar stuff

DisturbHerb:
nvm see #23093

PR #23195, opened 2025-04-28
Pod fuel consumption is scaled with pod speed and engine power

FlameArrow57:
[VEHICLES][BALANCE]

## About the PR 
Makes it so that pod fuel consumption is scaled by pod speed ^ 2 * (power generated / 200)

## Why's this needed? 
Adds a drawback to using fast pods which have almost no drawback at the moment

## Changelog 

```changelog
(u)FlameArrow57
(+)Pod fuel consumption is higher for faster pods and engines that generate more power
```


NibChocolateeny:
waht about stong but slow engine (Hermes 3.0 Engine)? maybe also use mor eplasma for those too? It is making morepower, need more plasms

FlameArrow57:
Sure, sounds like a great idea

NibChocolateeny:
I do not know - it maybe very exepensive for pod flying with change - but we acan try and see.

maybe in future bplasma using can be from how much power is being in use in pod (secondary sisytms on/off, guns, portalling) and different engine canbe more/not so much effiscient for certaen tasks. But that is for not now.

FlameArrow57:
.

ZeWaka (approved):

RubberRats:
I don't want to bikeshed this change, but I think it would be more sensible to tie fuel consumption to the engine used. Under these changes, a heavy industrial pod consumes less fuel than a civilian miniputt. If the problem is specifically gold pods, you could try also giving vehicle types their own fuel efficiency modifiers. Having fuel consumption as something that can be tweaked independently means more knobs can be tweaked in balancing.

PR #23197, opened 2025-04-28
ID computers can only remove accesses they have

JORJ949:
## About the PR 
When editing access with an ID computer via changing job only remove an access if its in the computer's allowed access list

## Why's this needed? 
Prevents someone like the medical director getting ahold of the captain's ID and reducing their access to assistant. Encourages getting the HoP to manage transfers or for the transferee to first get approved and demoted to staffie by their own head of staff.

## Changelog 

```changelog
(u)JORJ949
(+)Selecting a new job in an ID computer will no longer remove accesses that computer cannot edit.
```


pgmzeta:
this needs feedback to the user imo

JORJ949:
image
Ignore the above, boutput removed in favour of a notice box on the UI

TobleroneSwordfish:
Isn't the whole point of the departmental ID computers so that heads of staff can hire/fire people into/out of their own department? I don't really see what this is trying to fix, other than reducing potential crime opportunities.

JORJ949:
> Isn't the whole point of the departmental ID computers so that heads of staff can hire/fire people into/out of their own department? I don't really see what this is trying to fix, other than reducing potential crime opportunities.

This doesn't stop them hiring/firing their own department, they can still grant and revoke the accesses, change job title and ID colour, they just can't fire from OTHER departments and remove their access to that department. Though I'm not against removing the disabling of the job buttons so you can do that as the warning tells you that you can't change some of their accesses already.

mordent-goonstation:
DNM until we're on 516, avoiding more tgui reworking than we already have to do.

PR #23201, opened 2025-04-29
Water puddles ripple when you move through them

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Makes it so that puddles ripple when you move through them

https://streamable.com/rs3iyv

## Why's this needed? 
Thought it would be cool


flappybatpal:
Decent but they are quite high up when going left/right.

It's also perhaps a little odd that they only work at very specific water depths but that would be a much more complicated feature. It would be nice to see little splashes for shallow and a ripple of water around the mob in deeper water.

ZeWaka (changes requested):
Why are you setting all this junk manually every time instead of just making a subtype

ZeWaka (approved):
Concerned about cooldowns (graviton loops, for example)

FlameArrow57:
> Concerned about cooldowns (graviton loops, for example)

It's only triggered by mobs (called under /mob/living/EnteredFluid) so I think it should be fine?

TobleroneSwordfish (changes requested):

PR #23203, opened 2025-04-29
Security labelling changes

JORJ949:
## About the PR 
Security pinpointers will now be vended with a label of the user's name, same as would happen by using a hand labeler.
Security pinpointers will attempt to be put in the user's hand
Security disks now use a label instead of changing the name

## Why's this needed? 
QoL for security to make them less reliant on having a hand labeler in their department.
Buff for antags by allowing them to remove the labels from disks to potentially muddle up which disk belongs to who.

## Changelog 

```changelog
(u)JORJ949
(+)Security pinpointers and disks now come labelled with the owner's name, this label can be removed with a hand labeler.
```

### DIFF
--- a/code/obj/submachine/weapon_vendor.dm
+++ b/code/obj/submachine/weapon_vendor.dm
@@ -147,9 +147,6 @@
 			if (length(tracklist))
 				var/obj/item/pinpointer/secweapons/P = new(src.loc)
 				P.track(tracklist)
-				if(!islist(P.name_suffixes))
-					//Name_suffixes wasn't a list, but it is now.
-					P.name_suffixes = list()
 				P.name_suffix("([usr.real_name])")
 				P.UpdateName()
 				usr.put_in_hand_or_eject(P)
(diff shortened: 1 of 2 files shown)


code/obj/submachine/weapon_vendor.dm
```diff
@@ -147,6 +147,13 @@
 			if (length(tracklist))
 				var/obj/item/pinpointer/secweapons/P = new(src.loc)
 				P.track(tracklist)
+				if(!islist(P.name_suffixes))
+					//Name_suffixes wasn't a list, but it is now.
+					P.name_suffixes = list()
```
TobleroneSwordfish, line 152:
```suggestion
```
`name_suffix()` already does this

code/procs/jobprocs.dm
```diff
@@ -607,7 +607,11 @@ Equip items from body traits.
 				R.fields["mind"] = src.mind
 				D.root.add_file(R)
 
-				D.name = "data disk - '[src.real_name]'"
+				if(!islist(D.name_suffixes))
+					//Name_suffixes wasn't a list, but it is now.
+					D.name_suffixes = list()
```
TobleroneSwordfish, line 612:
```suggestion
```

TobleroneSwordfish (changes requested):

PR #23213, opened 2025-04-29
Add fluid line depth overlays to fluids

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Title, fluid lines are shown for deep fluids

image

image

Sprites partially worked on by Skeletonman0

## Why's this needed? 
Would be cool, shows more depth to fluids


itsloley:
i feel like the lines might be too colorful and/or bright? kinda makes it look more like a "magical mist/fog" instead of liquid, might be just me though!!

FlameArrow57:
> i feel like the lines might be too colorful and/or bright? kinda makes it look more like a "magical mist/fog" instead of liquid, might be just me though!!

What about now?

TobleroneSwordfish (commented):
Is it feasible to bake these into the icon states instead? That's a huge amount of extra overlays for big fluid pools which will definitely impact client performance.

FlameArrow57:
> Is it feasible to bake these into the icon states instead? That's a huge amount of extra overlays for big fluid pools which may impact client performance.

Possibly, the fluid objects will need to have their layers changed so that the fluid lines appear over objects if so

TobleroneSwordfish (changes requested):
Oh uh, that's way worse. Objects are heavier than images! I meant is it possible to just put the lines on the existing fluid icon states rather than making them separate at all.

FlameArrow57:
> Oh uh, that's way worse. Objects are heavier than images! I meant is it possible to just put the lines on the existing fluid icon states rather than making them separate at all.

I don't think so, it makes them very hard to see for more transparent fluids like water

TobleroneSwordfish:
Fluids already trash client performance pretty badly, I don't think adding another O(n^2) number of appearances per fluid puddle is worth it. Surely the lines being more transparent on more transparent fluids would be intentional anyway?

pgmzeta (approved):
there's a little bit of 'pop-in' if the fluid takes a bit to settle, such as a large volume in a small room escapes into a larger area. I don't think that's a deal-breaker, though.

cannot speak as to performance change.

PR #23214, opened 2025-04-29
Move unused maps to a separate repository

TobleroneSwordfish:
## About the PR 
Moves all unused or archived maps to a separate repository: https://github.com/goonstation/map_archive
There is a corresponding PR here: https://github.com/goonstation/map_archive/pull/1

Idea shamelessly stolen from tgstation, who did the same thing a while back.

## Why's this needed? 
We have a lot of old maps hanging around in the repo in various stages of maintained/broken/mangled by find-replace commits. We're not using these maps, we should not be maintaining them. Moving them to a separate repository allows them to be preserved as they were when they were in use and makes mass-editing our currently maintained maps a lot less of a slog.

Feel free to let me know if I've missed any or PR them across yourself. PRs to fix/maintain and document maps in the archive will be accepted if you feel like being a historian.


JORJ949 (approved):
Most of the actual real maps such as trunk and manta also have config files and  a build.dm that should be moved with them. (Also map-specific features such as comms officer but these should be edited on a per-feature basis in separate PRs, such as making comms officer a gimmick job)

JORJ949:
> Most of the actual real maps such as trunk and manta also have config files and  a build.dm define that should be (re)moved with them. (Also map-specific features such as comms officer but these should be edited on a per-feature basis in separate PRs, such as making comms officer a gimmick job)

TobleroneSwordfish:
I haven't copied the config files over to the archive because they won't really make sense outside of the wider codebase context and there's no guarantee they can just be slotted into modern versions of the game. Each map has a linked commit where it was known to work, config files can be retrieved from there.

PR #23225, opened 2025-04-30
Buff Trainee Roles

JORJ949:
## About the PR 
Increase the gear trainee roles start with to be (almost) the exact same as the full on role:
Research Trainees get spectro goggles (No gas mask so you can still see your trainee's face)
Medical Trainees get latex gloves and upgraded prodoc goggles
Technical Trainees stop basically being staffies and get:
- Insulated gloves
- A fully equipped toolbelt
- A proper engineering PDA (Removes the technical assistant PDA as after this it was only in the engineering apparel vendor as a straight downgrade to the standard engi pda.)
- Stomp boots in their bag if hotspots are enabled

## Why's this needed? 
Trainee roles are important for learning and starting without some of these roles' important tools is not good. I was trying to teach a Technical Trainee the other day and we kept having to stop to go get stuff I assumed they already spawned with such as tools and a device analyser on their PDA. 

## Changelog 

```changelog
(u)JORJ949
(*)Trainee roles now get job equipment matching that of the job they are training for.
```


flappybatpal (approved):
Logical.

PR #23255, opened 2025-05-02
Access proc cleanup

JORJ949:
## About the PR 
Cleans up unused/duplicate code in access.dm (the procs one not the defines one)
-Removes has_access proc, the only difference between "has_access(ACC)" and "ACC in thing.req_access" is that has_access erroneously returns TRUE if the thing has no access.
-Removes copied code from has_access_requirements and uses this proc in place of checking null ID with check_access, this proc is called anyway in check_access and theres no difference between calling it here vs in there

## Why's this needed? 
Code quality, I accidentally used the first proc with armory updating access and it caused me a bit of a headache trying to debug it.


TobleroneSwordfish (approved):

flappybatpal:
Tested locally and didn't find anything wrong with different IDs/no ID.

PR #23257, opened 2025-05-02
movement delays apply/remove instantly

TDHooligan:
[BALANCE]
## About the PR 
Movement effects are applied/removed instantly, rather than on the next move.

## Why's this needed? 
Tiny disorients like taser SMG plinks can apply massive movement slows or zero based entirely on how far between tiles a player is moving. Make this consistent so balancing is easier

## Changelog 

```changelog
(u)TDHooligan
(*)Slowing effects are now applied & removed instantly, instead of on your next move. Tasers will be more affected by this change.
```


code/mob/input.dm
```diff
@@ -68,6 +70,42 @@
...
+	if (src.buckled && istype(src.buckled, /obj/stool/chair))
+		var/obj/stool/chair/C = src.buckled
+		delay += C.buckle_move_delay //GriiiiIIIND
+	for (var/obj/item/grab/G as anything in src.grabbed_by)
+		if (istype(G) && BOUNDS_DIST(src, G.assailant) > 0)
+			if (G.state > GRAB_STRONG)
+				delay += G.assailant.p_class
+
+	var/list/stepped = list()
+	for (var/obj/item/grab/G as anything in src.grabbed_by)
```
TobleroneSwordfish, line 87: Why are you iterating `grabbed_by` twice? I assume one of these is meant to be something else.

TobleroneSwordfish (changes requested):

code/mob/input.dm
```diff
@@ -1,7 +1,9 @@
 
 /mob/var/prev_loc = 0
+/mob/var/prev_move = 0
```
TobleroneSwordfish, line 3: Comment this too

code/mob/input.dm
TDHooligan: spooky hacks i didn't look at hard enough. removed

PR #23261, opened 2025-05-02
New gun: Friend-A-Nobody

pgmzeta:
[gun][feature][game objects]
## About the PR 
Add a new sawn-off shotgun: Friend-A-Nobody. Flings you backwards when fired.

## Why's this needed? 
funny gimmick gun

## Demo
https://github.com/user-attachments/assets/94107fd4-cf69-4162-8120-c901eb93ffcc


code/obj/item/gun/kinetic.dm
```diff
@@ -3602,3 +3602,18 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 	alter_projectile(obj/projectile/P)
 		. = ..()
 		P.proj_data.shot_sound = 'sound/weapons/long_barrel.ogg'
+
+/obj/item/gun/kinetic/sawnoff/fan
+	name = "\improper Friend-A-Nobody"
+	desc = "This shotgun is even more sawn off than usual."
+	icon_state = "fan"
```
JORJ949, line 3609: Misleading icon_state and path, these don't need to be shortened, especially not to "fan" which doesn't immediately make me think "friend-a-nobody"

PR #23264, opened 2025-05-02
Shovels can dig holes/trenches in ground turfs

pgmzeta:
[feature][game objects]
## About the PR 
Ports the 35 Below trench digging system, with some changes to support:
* unsimmed turfs
* grave digging
* logging when burying mobs/special items

Adds:
* digging tool define and use where appropriate
* can_dig turf var to determine what turfs are diggable
* trench sprites copied from 35 below

### Things I Tested
* terranify digging
* azone graves (including item spawns)
* fluid turf digging (oshan/neon/nadir) works as before, for geothermal power

## Why's this needed? 
Feature for Decarabia
Unique mechanics for terranify

## Changelog 

```changelog
(u)Bartimeus, Disturbherb, Firedhat, glowbold, and Mr.Moriaty
(+)You can dig holes in suitable terrain with shovels.
(+)You can bury things and people in holes with shovels
```


TobleroneSwordfish (changes requested):
Digging a trench seems to break lighting on that tile permanently?
image
Tested with swamp terrainify on atlas without ambient lighting.

pgmzeta:
> Digging a trench seems to break lighting on that tile permanently? image Tested with swamp terrainify on atlas without ambient lighting.

should be fixed now

PR #23270, opened 2025-05-03
Resprites the Revolutionary Dress (Anthy's dress)

TemThrush:
[Clothing]
## About the PR 
Resprites Anthy's dress!
image
[New sprites on bottom, old sprites on top!]

## Why's this needed? 
Parity with other clothing sprites especially this dress's counterpart.


RubberRats:
Sprite is definitely an improvement over the original. That being said, it seems like the old version has a pair of gloves(?) on the front and rear sprites. Do you have plans to keep those?

TemThrush:
> Sprite is definitely an improvement over the original. That being said, it seems like the old version has a pair of gloves(?) on the front and rear sprites. Do you have plans to keep those?

I originally had those as part of the sprite but because of how uniforms mask on the body (I think this is what's causing it at least) any gloves i add or the ones part of the previous sprite are completely invisible when worn. I'm not sure if there's a workaround for this so I just removed that part from my sprite </3
(I can provide screenshots if my explanation is confusing)

TemThrush:
Oh gawd i just realized i forgot i'm not supposed to leave the arms open for uniforms one moment lmao.

PR #23276, opened 2025-05-03
Ties Mimic's health and speed to the size of their disguise, repaths mimic ability

CalliopeSoups:
[critters] [status effects]
## About the PR 
This PR repaths the mimic's mimic ability (and their sting) to its own file and adds a status effect given to the Mimic the first time they disguise. The status will match your speed and health to the size of the disguise. 

A few movement modifiers are added for potential balance tweaking. As this does make mimic stronger, it is now a short action bar with a slightly longer cooldown. This is my first time touching things like action bars, so there might be flaws in the code!

Numbers can be changed if health levels seem unfair.

## Why's this needed? 
TD;DR: Mimics feel unfinished, this PR gives them a passive based off of the prop hunt gamemode.

Long answer: https://hackmd.io/@CalliopeSoups/BJ_YC3yggg

## Changelog 

```changelog
(u)CalliopeSoups
(*)Mimics will now scale their health and speed with the size of their disguise. Report any unfitting values.
(+)Mimic is now an action bar and has a slightly longer cooldown.
```


CalliopeSoups:
Example: 
ezgif-1455857d554da1

amylizzle:
I'm not seeing any AI changes, did you test this with NPC mimics?

CalliopeSoups:
> I'm not seeing any AI changes, did you test this with NPC mimics?

Just ran a quick test and NPC mimics can still disguise into things with the action bar. I made it so only antag_spawns get the status effect, cause personally I'd like to keep NPC mimics the same as I don't see the things I'll be giving them as something the AI would make use of easily.

code/modules/status_system/statusEffects.dm
```diff
@@ -3706,3 +3706,59 @@
...
+		var/mob/living/critter/mimic/mob_owner = src.owner
+		return "Health: [mob_owner.max_health], Speed: [speed_string]"
+
+	onUpdate()
+		..()
+		var/mob/living/critter/mimic/mob_owner = src.owner
+		var/new_amount = mob_owner.pixel_amount
+		if (new_amount != src.last_amount)
+			src.last_amount = new_amount
+			scale()
```
TobleroneSwordfish, line 3731: Why not just call `setStatus` with the pixel amount instead of constantly checking it here?

code/datums/abilities/critter/mimic.dm
```diff
@@ -0,0 +1,52 @@
+/datum/targetable/critter/mimic
+	name = "Mimic Object"
+	desc = "Disguise yourself as a target object."
+	icon_state = "mimic"
+	cooldown = 45 SECONDS
+	targeted = TRUE
+	target_anything = TRUE
+	cooldown_after_action = TRUE
+	var/d_target = null
```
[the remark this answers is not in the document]
CalliopeSoups: Oops, that was a holdover.

CalliopeSoups:
Status is only set for 10 seconds now, so the pixel amount is called onAdd instead of onUpdate.

TobleroneSwordfish (changes requested):
The slow movement modifier isn't being correctly removed when you go from a large thing to a small one. Also why only 10 seconds?

TobleroneSwordfish:
Ah, no. You can just set it to infinite and use `onChange`. Sorry I thought `onAdd` worked with changes too.

code/modules/status_system/statusEffects.dm
```diff
@@ -3706,3 +3706,60 @@
...
+	desc = "Your speed and health are matching with your disguise."
+	icon_state = "mimicface"
+	visible = TRUE
+	var/pixels = null
+	var/speed_string = null
+
+	getTooltip()
+		var/mob/living/critter/mimic/mob_owner = src.owner
+		return "Health: [mob_owner.max_health], Speed: [speed_string]"
+
```
TobleroneSwordfish, line 3722:
```suggestion
	onAdd(optional)
		. = ..()
		src.onChange(optional)
```

PR #23278, opened 2025-05-03
strip doors (plastic flappy doors) QoL and help messages

itchybones:
[QoL][Game objects]
## About the PR 
This PR makes strip doors (un-)anchorable via wrenching and adds help messages. I also sorted the procs (overrides first) and implemented consistent actionbars for all actions that didn't have them yet. Additionally, this fixes a seemlingly unknown bug, where strip doors built facing west would have to be rotated twice with a crowbar in order to actually rotate.

## Why's this needed? 
Currently, when you build a strip door frame, there's no way to move it besides full-on deconstruction. For fixing a fairly simple mistake, one shouldn't require specialized equipment. Help messages were added as per dev request, and

## Changelog 

```changelog
(u)Chasu
(+)Strip door frames can now be (un-)anchored with wrenching tools and have help messages
```


ZeWaka (commented):
Can we add a help message?

itchybones:
> Can we add a help message?

Sure thing, I'll handle that tomorrow (in ~12 hours)

itchybones:
> Can we add a help message?

All done and tested.

JORJ949 (approved):

code/modules/economy/supply_misc.dm
```diff
@@ -185,9 +153,13 @@ TYPEINFO(/obj/strip_door)
 	attackby(obj/item/I, mob/user)
 		if (ispryingtool(I))
 			playsound(src.loc, 'sound/items/Crowbar.ogg', 30, 1, -2)
-			src.change_direction()
+			SETUP_GENERIC_ACTIONBAR(user, src, 1 SECOND, /obj/strip_door/proc/change_direction, user, I.icon, I.icon_state, null, INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION | INTERRUPT_MOVE)
```
TobleroneSwordfish, line 156: Changing direction shouldn't require an actionbar and should only work when unanchored (for consistency with Other Things)

TobleroneSwordfish (changes requested):

code/modules/economy/supply_misc.dm
itchybones: Alright, on it.

itchybones: Done, I changed the help messages to reflect that unanchoring is required for rotation, too.

PR #23283, opened 2025-05-04
Adds Health Scanners to Med fabricators

cheekybrdy:
[BALANCE] [CONTROVERSIAL][INPUT-WANTED]
## About the PR 
Adds health scanners to med fabricators.

## Why's this needed? 
The recent PR to remove health analyzers from base medkits (#23199) while a good PR for getting people to use examine for diagnostics and get them actually to medbay was noted by others to make mass production of analysers for requisitions to be rather difficult. This 1. Gives the weaker of the 2 med fabs a useful item. 2. Solves the mass production problem. 3. Still funnels people into the areas which desire traffic to get analysed, although with the addition of QM also being able to provide them.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Health analysers are now producible at the medical fabricator.
```


CalliopeSoups:
My two cents is that since the health analyzers that came in the medkits were only basic, they're not gonna help you figure anything out more than "I'm damaged somewhat", which you can do with your eyes too, not to mention the several floor scanners about a map, the health analyzers up for sale in med vendors, asking the medics themselves, keeping track of the type of damage you suffered etc.

It's just my opinion but I quite like health analyzers being a bit more rare.

AnonymousUserTheOnly:
my thoughts?  it gives a way to replenish the stock of them for robotics purposes, since robotics would usually take the scanners from the medkits as well as the medkit boxes to make medibots cheaper.  without those, robotics would likely end up taking the scanners from the medi-vends.  so giving them back a source that doesn't need taking from proper medics is nice.

pgmzeta:
medibots are already printable in robotics. i would prefer to give them more to so they don't have idle time to mass-construct medibots and require the resources to do so.

jolynetheawesome:
"Just ask medbay and maybe give them a vending restock cartridge."
a two step process.. still two steps! wow!

TobleroneSwordfish:
I think I'm okay with them being slightly rarer for now, there's still 10(!) across two medvends

PR #22904, opened 2025-04-07
Consolidates cooking recipes into variants

RubberRats:
[CATERING][CODE QUALITY]
## About the PR 
This PR consolidates very similar groups of recipes, like buttburgers and toast, into a single recipe datum using a new `variants[]` list. It also does a few small cleanup tasks where they intersected with what I was doing (got in my way).

## Why's this needed? 

Having to wrestle `cookingrecipes.dm` in my last PR made me mad. Hopefully this makes future changes to recipe code less painful.

## Changelog 

```changelog
(u)RubberRats
(+)Cooking recipe code has been changed, again. Please report any issues.
```

PR #23289, opened 2025-05-04
Recipe selection code refactor + oven repath

RubberRats:
## About the PR 

This PR repaths `obj/submachine/chef_oven` and `obj/machinery/mixer` to have a common `obj/machinery/cookingmachine` parent. Additionally, it replaces the old selection system (a massive, manually-curated list) with one that sorts itself automatically.

This PR requires #23093 and #22904 to be merged as a prerequisite, as it is built on changes made there

TODO:
- [ ] Implement priority on recipes
- [ ] Get the sorting system running
- [ ] Update paths
- [ ] Make sure code compiles
- [ ] Fix inevitable merge conflicts
- [ ] Check potential cases for bugs

## Why's this needed? 

Ovens and mixers share a lot of redundant code, and maintaining both of them whenever a change to cooking code is made is a bit of a pain. Additionally, ovens were submachines, and reducing the number of those in the code is probably for the best.

Similarly, recipe selection code represents a massive amount of technical debt, and requires manual ordering of over a hundred recipes to even function. It is a huge pain to maintain and a significant obstacle to adding anything to the catering system.

These changes were bundled together because I could not see a way of adding just one that didn't feel kinda gross.

## Changelog 

```changelog
(u)RubberRats
(*)Oven and mixer behavior has been significantly refactored, please report any bugs.
(+)Ovens now consume power while running
```


flappybatpal:
One conflict with secret due to repathing the oven.

flappybatpal (changes requested):
food_crime has been removed, preventing secret functionality.

RubberRats:
Whoops, didn't realize it was used in secret. Should be implemented now

code/modules/cooking/cookingmachine.dm
```diff
@@ -0,0 +1,538 @@
...
+			var/itemData = list()
+			itemData["name"] = I.name
+			itemData["index"] = index
+			itemData["iconData"] = get_item_icon(I)
+			heldItems += list(itemData)
+			index += 1
+		. = list("working" = src.working, "heldItems" = heldItems)
+
+	proc/get_item_icon(var/obj/item/target)
+		var/static/base64_preview_cache = list()
```
ZeWaka: you do know icon2base64 already has a cache right

tgui/packages/tgui/interfaces/Oven.tsx
```diff
@@ -24,11 +24,9 @@ import { ProductList } from './common/ProductList';
...
   time: number;
   heat: string;
-  cooking: BooleanLike;
-  content_icons: string[];
-  content_names: string[];
-  recipe_icons: string[];
-  recipe_names: string[];
+  working: BooleanLike;
+  heldItems;
+  recipe_ingredients;
```
ZeWaka: bit sad this had types previously and no longer has them

code/modules/cooking/cookingmachine.dm
RubberRats: I had initially ported this over from mixer code, and kept it in the interest of avoid unintended changes. Going to remove the proc now.

PR #23292, opened 2025-05-05
Adjust screen wiggle fix sleep times

Sovexe:
## About the PR 
* The screen fix wiggle sometimes doesn't work on live, usually on server 4 and on donut 3
* Terrible code until tgui fix for byondui

## Why's this needed? 
* There's a bit of a race condition with this fix


ZeWaka (approved):
This is so sad deapacito

ZeWaka:
Please no I'm begging 😭😭😭😭

ZeWaka:
please lummox 😿😿😿

Sovexe:
So...this has been working, as bad of a workaround as trying to time for a race condition is...it can still fail at times but often gets resized properly.  I really don't want to extend the spawn even further either

PR #23299, opened 2025-05-06
IV stands only flash when connected to person

pgmzeta:
[qol][game objects][medical]
## About the PR 
Tweak the icon_state names for IV stands to make it clearer what each state represents

## Why's this needed? 
right now iv stands flash whenever there's a bag attached, but i think it makes more sense to flash when someone's hooked up to it,  like the dialysis machine


DisturbHerb:
As the original author: the intent follows real world examples as they flash to alert for errors such as line occlusion or if the IV bag is empty, while not flaahing during normal operation. I strenuously oppose reversing this from the current implementation.

PR #23302, opened 2025-05-06
TGUI Access Pro/Lite

JORJ949:
## About the PR 
Changes the Access Pro and Access Lite to have a TGUI window of access instead of requiring an inserted ID or scanned access
Advantages of Pro over Lite are now:
- More access options, especially in command and security.
image

- Can affect non-airlock objects (that allow reprogramming)
- Can affect objects that already have access requirements.

## Why's this needed? 
The access "guns" are presently very clunky and not really worth carrying, especially the AccessPro which requires you to already know what access you will want to edit and you have an ID with those accesses already. 
Access Lite is also not great as they require map knowledge and for you to know which doors will have the access you want, instead just have them be unable to apply some of the more complex accesses (QM Console, Medical lockers, most sec/command accesses)

Also fixes the AccessLite being unable to program any airlock right now.

## Changelog 

```changelog
(u)JORJ949
(*)The Access Pro and Access Lite now have a UI window that you can select accesses from.
```


pgmzeta:
good for the pro but the lite should keep its copy+paste functionality imo

PR #23304, opened 2025-05-06
Whisper Maptext

Mister-Moriarty:
## About the PR:
Enables whisper maptext.

## Why's this needed?
By far the most requested thing in the say rework channel.

## Changelog:
```changelog
(u)Mr. Moriarty
(+)Whispers now display maptext.
```

### DIFF
--- a/code/modules/speech/message_modifiers/whisper.dm
+++ b/code/modules/speech/message_modifiers/whisper.dm
@@ -14,7 +14,6 @@
 /datum/message_modifier/postprocessing/whisper/process(datum/say_message/message)
 	. = message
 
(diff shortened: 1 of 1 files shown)


JORJ949:
Is there any special indication that its a whisper rather than something being spoken aloud? If someone's whispering to me vs speaking fully I think that should be noticable at a glance.

Mister-Moriarty:
> Is there any special indication that its a whisper rather than something being spoken aloud in the maptext? If someone's whispering to me vs speaking fully I think that should be noticable at a glance.

Yes, it's italicised, like the text in the chat window.

JORJ949 (approved):

444Portal (approved):

PR #23305, opened 2025-05-06
Lawbringer works regardless of speech affects

FlameArrow57:
[GAME OBJECTS][QoL]

## About the PR 
Allows the Lawbringer to work even if you are stammering, and when playing with a lizard or cow mutantrace

## Why's this needed? 
Would be cool to be able to any phrase when playing with a different mutantrace.

Also would be nice to use the Lawbringer when you are stammering

## Changelog 

```changelog
(u)FlameArrow57
(+)The Lawbringer is no longer affected by speech affects or mutantrace
```

### DIFF
--- a/code/modules/speech/modules/listen/effects/lawbringer.dm
+++ b/code/modules/speech/modules/listen/effects/lawbringer.dm
@@ -43,7 +43,7 @@
 		lawbringer.UpdateIcon()
 		return
 
-	var/text = lawbringer.sanitize_talk(message.content)
+	var/text = lawbringer.sanitize_talk(message.original_content)
 	if (!lawbringer.fingerprints_can_shoot(H))
 		if (src.valid_modes[text])
 			global.random_burn_damage(H, 50)


JORJ949:
This is a pretty large buff, lets you use it after eating your very powerful pasta or when an antag has attacked you causing bloodloss, disorient or other reason's you'd be unable to talk properly.

Mister-Moriarty (approved):
The code is good.
I have no strong opinions on the balance implications of this.

TobleroneSwordfish:
I think the stammering thing is a nice little detail that antags can play around.

444Portal:
No real opinions either way on stammering but all mutantraces should be able to use the Lawbringer with no strings attached. Fan of that!

TobleroneSwordfish:
Closing this and #23342 for the same reasons: stammering breaking mode switching is an intentional mechanic and we already have aliases for mutantraces to use the lawbringer.

PR #23313, opened 2025-05-06
Refactors Ability Buttons Into Speech Prefixes

Mister-Moriarty:
[Player Actions] [Feature] [QoL]

## About The PR:
Refactors the Hivechat, Thrallchat, and Kudzuchat abilities into say prefixes, so that they may be accessed through the Y radio menu, which has been retooled into a more general channel selector.

These channels may also be accessed through the `:hive`, `:thrall`, and `:kudzu` prefixes, although the intended method is the channel selector.

## Why's This needed?
QoL that the say rework enables. Also a prerequisite for Wraithchat and for removing the remaining ghostdrone speech jank.

## Changelog:
```changelog
(u)Mr. Moriarty
(+)Hivechat, Thrallchat, and Kudzuchat are now accessed through the Y menu instead of through abilities.
```


TobleroneSwordfish (approved):
Very cool

Mister-Moriarty:
> How do I as a player know these prefixes exist?

They appear in the `Y` menu as options. There's no indication to the player that prefixes like `:hive` exist currently, but they could be communicated to the player in the antag tips text, if you wish. I don't see them getting much use due to typing indicators though.

ZeWaka:
I pretty much exclusively use the command bar, and never use Y. Adding these to the antag popup would be greatly appreciated

PR #23315, opened 2025-05-06
Speech/Listen Module Tree Admin Panels

Mister-Moriarty:
[Admin] [Feature] [QoL]

## About The PR:
Implements admin panels for module trees, allowing for modules to be easily added, removed, and modified. The editors also permit for the modification of parents; origins; auxiliary, importing, and exporting trees; target trees for auxiliary trees; default speech channel; say language; and so on.

## Why's This Needed?
Say rework admin tools.


mordent-goonstation:
DNM until we're on 516, avoiding more tgui reworking than we already have to do.

mordent-goonstation:
Not reviewing code, but that diagonal style on the sections is non-standard as heck and I have no idea if it's conveying something.

ZeWaka (approved):

PR #23319, opened 2025-05-06
Lawbringers check every word for mode change

JORJ949:
## About the PR 
Lawbringers will now check every word in a spoken message rather than it only working if the entire message is a valid one, prioritising the first one in a message if there are multiple.
image
Also fixes a bug where lawbringers would always change mode even when their loc was not the user.

## Why's this needed? 
I think it would be really funny to say "I'm gonna get you CLOWN!!!" and have the lawbringer change to clown mode, or say "Lawbringer, activate ASSAULT laser.". May result in accidental mode changes for example if still being held while saying "You are under arrest for assault" Then it will end up on High Power, but this is funny.

## Changelog 

```changelog
(u)JORJ949
(+)Lawbringers now check every word in a message for mode selection rather than the entire message.
```

### DIFF
--- a/code/modules/speech/modules/listen/effects/lawbringer.dm
+++ b/code/modules/speech/modules/listen/effects/lawbringer.dm
@@ -44,15 +44,18 @@
 		return
 
 	var/text = lawbringer.sanitize_talk(message.content)
-	if (!lawbringer.fingerprints_can_shoot(H))
-		if (src.valid_modes[text])
+	var/list/words = splittext(text, " ")
+
+	for(var/word in words)
+		if(!src.valid_modes[word])
+			continue
+		if (!lawbringer.fingerprints_can_shoot(H))
 			global.random_burn_damage(H, 50)
 			H.changeStatus("knockdown", 4 SECONDS)
 			global.elecflash(lawbringer, power = 2)
 			H.visible_message(SPAN_ALERT("[H] tries to fire [lawbringer]! The gun initiates its failsafe mode."))
-
+			return
+		lawbringer.change_mode(H, word)
+		H.update_inhands()
+		lawbringer.UpdateIcon()
 		return
-
-	lawbringer.change_mode(H, text)
-	H.update_inhands()
-	lawbringer.UpdateIcon()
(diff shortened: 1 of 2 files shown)


code/obj/item/gun/energy.dm
```diff
@@ -1612,7 +1612,7 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 	//just remove all capitalization and non-letter characters
 	proc/sanitize_talk(var/msg)
 		//find all characters that are not letters and remove em
```
TobleroneSwordfish, line 1614: These comments are now wrong, the regex now allows whitespace

code/modules/speech/modules/listen/effects/lawbringer.dm
```diff
@@ -44,15 +44,18 @@
...
+	var/list/words = splittext(text, " ")
+
+	for(var/word in words)
+		if(!src.valid_modes[word])
+			continue
+		if (!lawbringer.fingerprints_can_shoot(H))
 			global.random_burn_damage(H, 50)
 			H.changeStatus("knockdown", 4 SECONDS)
 			global.elecflash(lawbringer, power = 2)
 			H.visible_message(SPAN_ALERT("[H] tries to fire [lawbringer]! The gun initiates its failsafe mode."))
```
TobleroneSwordfish, line 56: This code doesn't appear to work on master and so should probably be removed rather than fixed (I don't think 50 burn damage and a stun for trying to change mode on a gun you can shoot anyway is a good idea)

TobleroneSwordfish (changes requested):

code/modules/speech/modules/listen/effects/lawbringer.dm
JORJ949: That's because it runs the are_you_the_law check and returns way earlier, that just reimplements the fingerprints_can_shoot check

JORJ949: One minor thing this changes now is that if the lawbringer has no owner (meaning not picked up yet) anyone can change its mode but that's not really an issue, you still have to hold it while not being its owner to I AM THE LAW it .

JORJ949: Actually seems like this may be a speech rework bug where it doesnt check if its equipped before changing mode.

PR #23328, opened 2025-05-07
Tweak nanite balance

pgmzeta:
[balance][critters]
## About the PR 
When nanites are dying, they will no longer attack or start chasing you.
Nanites have "generations", each generation has a slightly lower chance of reforming into nanites again
The baseline reform rate is now 50%, which is divided by the generation count (e.g. 50% at gen 1, 25% at gen 2, 50/3 ~17% at gen 3, etc.)

## Why's this needed? 
Now that nanites are more common, they feel a little too strong. This doesn't lower their direct damage, but tweaks two aspects that feel the most unfair:
1. They can attack when they are a little dust pile
2. You can theoretically get infinite spawns. while it'll be more common to get a second, it'll be less common to have e.g. 6 of them spawn off one nanite spawn.


TobleroneSwordfish (approved):
Now we just need them to be visible!

Issue #22087, opened 2025-01-23
Running programs like teleman via command substitution in a script, may freeze the DWAINE terminal


CrimsonNode:
### Labels

[BUG] [MINOR] [STATION SYSTEMS]

### Description

Using command substitution, $() , with programs like teleman, may cause the Terminal to freeze completely. Requiring you to disconnect and reconnect to restore functionality

### Steps to reproduce

1. Use the following script:
`#!`
`eval $(teleman coords 50 50 1) $(teleman send)`
`eval $(teleman coords 50 50 1) $(teleman send)`
`eval $(teleman coords 50 50 1) $(teleman send)`
`echo Done!`
(Only one eval line can cause the issue as well, but using more increases the chances of it happening.)

2. Try running the script a few times.

3. Terminal freezes

### Expected Behavior

Script running normally.

### Additional Information & Screenshots

Not really familiar with DM but `"var/tmp/obj/item/disk/data/holder"` inside `computerfiles.dm` turning `null` seems to cause the problem.

This seems to be an issue caused by a recent change regarding DWAINE since I have been using a similar syntax for months in one of my scripts and didn't have any problems before. It seems to happen semi-randomly. 

Might also be a problem with how teleman and "$()" interacts but I am not really sure. Code was far too complex for me xd

CrimsonNode:
Hey again! PR #22024  seems to have caused the problem. I'm not exactly sure how it caused this, but I just wanted to mention it in case it might be helpful!

CrimsonNode:
Changed the title because problem isnt seem to be linked to"eval" keyword but linked to the command substitution , $(). Since I was able to replicate it with using echo.

PR #23334, opened 2025-05-07
Enhance mainframe stability while scripting

CrimsonNode:
[GAME OBJECTS][STATION SYSTEMS][BUG][QOL]
## About the PR

This PR introduces two small bugfixes for the mainframe to better handle:

1.  **Improved Exit Handling When Parent Task is Disposed:** 
	* A new variable, `var/tmp/datum/computer/file/mainframe_program/shell/base_shell_instance`, has been added to `/datum/mainframe2_user_data`. This variable stores the base shell instance for the user who logged into that shell.
	* The kernel's program exit logic (`DWAINE_COMMAND_EXIT`) now includes an additional check. It verifies if the parent task of the program being exited (`quitparent`) is still active. If the parent task has been disposed, control is now returned to the user's base shell instance instead.

2.  **Better Mainframe Recovery from Recursion Errors:** If the mainframe encounters a "maximum recursion level" exception, it will now:
	* Broadcast a notification message to all users connected to this mainframe about the error.
    * And will perform a clean reboot (unload all programs, log all users off from that mainframe, re-initialize the Kernel and reconnect all devices).
	* To help with logging all users off, a helper command `logout_all_users` has been added to `/datum/computer/file/mainframe_program/os/kernel/proc/`
	* `unload_all()` inside the `/obj/machinery/networked/mainframe` has been updated to log all users off, before unloading all programs and os. This makes sure, after a reboot user can log back into their old user account instead of creating a new one.
	* Rebooting logic has been written inside a small proc called `reboot_mainframe()`

## Why's this needed?

These issues were identified primarily while working with DWAINE scripts.

The first problem, related to exit handling, appeared to be related to the `SLEEP()` command used by the `s_telepad` driver. When `SLEEP()` is executed, it pauses the current process (and its callers). However, the mainframe still continues to process connected terminals. This creates a problem where if we are using `teleman()` inside a script, the parent shell of the `teleman` program we called, could be processed and disposed of before the `teleman` program itself finishes. Consequently, when `teleman` eventually calls `DWAINE_COMMAND_EXIT`, it attempts to return control to its parent shell, which no longer exists. This leaves the user with a disposed shell as their current program. Should hopefully fix: #22087 

This fix ensures that `user.current_prog` will, at worst, revert to the user's base shell instance, preventing the terminal from being stuck in a non-functional state. There may be a better way to prevent this from happening in the first place, but at least this will allow Kernel to handle it more gracefully. In past, terminal used to completely freeze until you log out and back in.

The second problem addresses how the mainframe handles maximum recursion errors. Since commands simlar to `teleman`, can have long execution times, the likelihood of encountering such an exception increases. Previously, these errors would cause the mainframe to unload all programs and set the OS to null. This rendered the mainframe completely unresponsive, requiring someone to manually reboot the mainframe by powering it off and on. Users would often only discover this problem when they were unable to reconnect.

The new handling allows the mainframe to inform its users about the exception and then initiate a self-reboot, restoring its services without requiring manual help. This part of the PR doesn't change anything about how mainframe performs recursive functions, only changes how a maximum recursion exception is handled.

This is my first time working with DM, apologies for any possible errors.

## Changelog
```changelog
(u) CrimsonNode
(+) Computer terminals should hopefully crash less when scripting.
(+) The mainframe can now recover from errors caused by recursive scripts by performing a self-reboot, instead of requiring a manual-reboot.
```


code/modules/networks/computer3/mainframe2/mainframe2.dm
```diff
@@ -564,6 +567,8 @@
 			return 1
 
 		unload_all()
+			if (src.os && hascall(src.os, "message_all_users"))
+				src.os:logout_all_users(FALSE) // Lets logout all users so when the mainframe reboots, they can log into their old user account, instead of creating a new user.
```
TobleroneSwordfish, line 571: ditto

CrimsonNode: I see! I will probably move those 2 inside /kernel then!

PR #23342, opened 2025-05-07
Detgadget hat works regardless of speech affects

FlameArrow57:
[GAME OBJECTS][FEATURE]

## About the PR 
Allows the detgadget hat to work regardless of speech affects like drunk speech or mutantrace

## Why's this needed? 
Would be cool, also using the detdaget hat while drinking tons of alcohol would be epic

## Changelog 

```changelog
(u)FlameArrow57
(+)The Detgadget hat works regardless of speech affects like drunk speech
```

### DIFF
--- a/code/modules/speech/modules/listen/effects/detgadget.dm
+++ b/code/modules/speech/modules/listen/effects/detgadget.dm
@@ -6,12 +6,12 @@
 	if (!istype(hat) || !ismob(message.speaker))
 		return
 
-	var/phrase_location = findtext(message.content, hat.phrase)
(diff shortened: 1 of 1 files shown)


JORJ949:
Same thing as with the lawbringer but to a much lesser extent as it isn't a weapon in that this is something antagonists can play around and I like that.

TobleroneSwordfish:
Closing this and #23305 for the same reasons: stammering breaking voice activated things is an intentional mechanic, mutantraces can set their own detgadget aliases.

PR #23345, opened 2025-05-07
Change up wizard Empower spell

FlameArrow57:
[GAME MODES][BALANCE]

## About the PR 
Makes the following changes:
-Reduces Empower spell duration/CD to 15s/30s from 30s/45s
-Empower removes stun knockdowns on use

## Why's this needed? 
Empower is meant to be a temporary burst of strength. However, 30 seconds feels too long in that you can't really play around it, and it can be combined with other offensive spells already

Also adds some utility in a get out of stuns buff

## Changelog 

```changelog
(u)FlameArrow57
(+)Wizard empower spell duration and CD reduced to 15s/30s from 30s/45s. It now also removes knockdown status effect on use.
```

### DIFF
--- a/code/modules/antagonists/wizard/abilities/mutate.dm
+++ b/code/modules/antagonists/wizard/abilities/mutate.dm
@@ -3,7 +3,7 @@
 	desc = "Temporarily superpowers your body and grants a burst of strength."
 	icon_state = "mutate"
 	targeted = 0
-	cooldown = 30 SECONDS
+	cooldown = 35 SECONDS
 	requires_robes = 1
 	offensive = 1
 	voice_grim = 'sound/voice/wizard/MutateGrim.ogg'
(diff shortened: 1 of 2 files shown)


code/modules/antagonists/wizard/abilities/mutate.dm
```diff
@@ -25,9 +25,10 @@
 			holder.owner.bioHolder.AddEffect("hulk", 0, 0, 0, 1)
 		APPLY_ATOM_PROPERTY(holder.owner, PROP_MOB_PASSIVE_WRESTLE, "empower")
 		APPLY_ATOM_PROPERTY(holder.owner, PROP_MOB_STAMINA_REGEN_BONUS, "empower", 5)
-		var/SPtime = 150
+		src.holder.owner.delStatus("knockdown")
```
TobleroneSwordfish, line 28:
```suggestion
		src.holder.owner.remove_stuns()
```
there's more than one type of stun

TobleroneSwordfish (changes requested):
You can cast this ability while stunned, right?

FlameArrow57:
> You can cast this ability while stunned, right?

Yeah, tested it with baton

flappybatpal:
It might be a bit short but I do think it was too long so it's worth a try.

PR #23346, opened 2025-05-07
`open_to_sound` Defaults To `TRUE`

Mister-Moriarty:
[Experimental] [Bug]

## About The PR:
`open_to_sound` now defaults to `TRUE`, which should fix all the bugs about not hearing others while inside of objects.

It should be noted that this may introduce unforeseen bugs, but I've tried to thoroughly test the most common cases.


code/modules/speech/outermost_listener.dm
```diff
@@ -52,7 +52,7 @@
 
 /// Determines whether the loc of the outermost listener is open to sound.
 /datum/outermost_listener_tracker/proc/loc_open_to_sound()
-	if (!src.outermost_listener.loc)
+	if (!ismovable(src.outermost_listener.loc))
```
TobleroneSwordfish, line 55: Won't this prevent hearing while on a turf?

Mister-Moriarty: No, because on `master`, `turf/open_to_sound` is `FALSE`.
Theory-wise, the outermost listener should be an `/atom/movable` anyway, otherwise it would require to be updated every time a listener moved. Say channels also work under the assumption that the outermost an outermost listener can be is an object on a turf; see `GET_INPUT_OUTERMOST_LISTENER_LOC`.

TobleroneSwordfish (approved):

PR #23347, opened 2025-05-07
Add wizard traveller's robe and hat

FlameArrow57:
[GAME MODES][CLOTHING][FEATURE]

## About the PR 
Adds a traveller's robe and hat to the corner in the wizard shuttle

Worn states sprited by Flaborized

image

## Why's this needed? 
Looks nice


TobleroneSwordfish (approved):
oooahh!

PR #23348, opened 2025-05-07
Rework spellshield var into a status effect

FlameArrow57:
[INTERNAL][CLEANLINESS]

## About the PR 
Reworks the spellshield var into a status effect

## Why's this needed? 
Status effect is nicer to work with than a single variable that can be changed by multiple things, also allows extra functionality for spellshield effects


flappybatpal:
Good refactoring, will test when conflict is resolved.

flappybatpal (approved):

code/modules/status_system/statusEffects.dm
```diff
@@ -3759,9 +3759,29 @@
 		APPLY_MOVEMENT_MODIFIER(mob_owner, mob_owner.modifier, src.type)
 		src.getTooltip()
 
+/datum/statusEffect/spellshield
+	id = "spellshield"
+	name = "Spell shield"
+	desc = "You have an active spellshield, protecting you from various damage sources."
+	icon_state = "nradiation1"
+	var/image/shield_overlay
```
ZeWaka, line 3767: is this an overlay or underlay?

Why not use AddOverlays?

ZeWaka (changes requested):

ZeWaka (approved):

PR #23353, opened 2025-05-08
Change wizard forcewalls to 20 second duration

FlameArrow57:
[GAME MODES][BALANCE]

## About the PR 
Changes wizard forcewall duration from 30 to 20 seconds

## Why's this needed? 
Currently they have a longer duration than cooldown, meaning they can stay in a spot indefinitely

Change makes a small window where the crew can get past, unless the wizard is paying mind to it and may need to use a spell too

## Changelog 

```changelog
(u)FlameArrow57
(+)Wizard forcewall duration changed from 30 -> 20 seconds
```


flappybatpal:
Seems reasonable, not like wizards don't have a big bag of other spells all cooling down.

PR #23354, opened 2025-05-08
Add "Fired" as valid word for lawbringer hotshot

JORJ949:
## About the PR 
Adds "Fired" as a valid word to activate the lawbringer's hotshot mode.

## Why's this needed? 
And now the true purpose of #23319 is revealed. "You are FIRED." _**HOT**_

## Changelog 

```changelog
(u)JORJ949
(+)"Fired" is now a valid word for the lawbringer's hotshot mode. Go terminate those contracts!
```


TobleroneSwordfish (approved):
Pfff sure

PR #23378, opened 2025-05-08
Replace teleport scroll and Soulguard with Flee spell

FlameArrow57:
[GAME MODES][BALANCE][FEATURE]

## About the PR 
Makes the following changes
-Removes teleport scroll from Wizard gear
-Disables Soulguard as a purchaseable spell
-Adds a new 60 second CD, 3 use spell, Flee, which recalls you to the Wizard den
-Teleport spell can no longer take you to the wizard den

## Why's this needed? 
-Removes clunkiness in using teleport scroll to recall back to the wizard den (used in emergency situations, need to scroll through TGUI list, have TGUI open while it's in your pocket)
-Enforces that wizards have a sense of limited "lives" and that they cannot live forever but do have some breathing room if they make a mistake

## Changelog 

```changelog
(u)FlameArrow57
(*)Added Flee as a wizard spell that will take you to the Wizard den. Has 3 uses.
(*)Teleport spell no longer can take you to the Wizard den.
(*)Wizards no longer get teleport scroll or Soulguard
```


FlameArrow57:
> I feel like flee should be castable while stunned if we remove soulguard for it.

Is is yeah

PR #23385, opened 2025-05-09
Listening Post Station-Z Prefab & Startup Sequence

pgmzeta:
[feature][mapping][balance]
## About the PR 
1. Creates a new listening post prefab loaded in during map initialization.
  * A "standard" prefab is included in all maps that currently share a post layout: Cogmap, Cogmap2, Clarion, Donut2, and Mushroom. 
  * Listening posts with a custom layout are **not** changed: Atlas, Donut3, Kondaru, Nadir, Neon, Oshan, Wrestlemap

2. Refreshes the "standard" listening post look to be more in line with Neon's listening post style.

3. Adds a run-once "startup" sequence to the listening post, which slowly turns on the lights inside the post.
  * To support this effect, several sub-areas of the listening post were created and added to the prefab.

4. Ensures that whenever a mob winds up inside the listening post without using the hand scanner it is opened up for everyone.

## Why's this needed? 
Prefab: Stop needing to remember copy/pasting changes between ~half of the listening posts in game.

Design refresh: The two most recently designed listening posts from Nadir and Neon have a distinct style/flair that is missing from the current listening post.

Lights turning on: Makes the listening post feel more "secret base." 

Entering always unlocks: Anti-voltron/arcfiend cheese.

## Demo
https://github.com/user-attachments/assets/887e091b-387d-4ef6-a9af-d48c512ad868

## Screenshot 
Screenshot 2025-05-08 212029

## Changelog 

```changelog
(u)glowbold
(+)The listening post on some maps have received a new look.
(+)The listening post will fully unlock once someone is inside or uses the hand scanner.
```

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -43482,14 +43482,6 @@
 /obj/mapping_helper/wingrille_spawn/auto/reinforced,
 /turf/simulated/floor/plating,
 /area/station/security/main)
-"ilE" = (
-/obj/forcefield/energyshield/perma/doorlink{
-	desc = "A door-linked force field that prevents gasses from passing through it.";
-	name = "Door-linked Atmospheric Forcefield";
-	dir = 8
-	},
-/turf/space,
-/area/space)
 "imv" = (
 /obj/machinery/atmospherics/binary/valve{
 	high_risk = 1;
@@ -69183,7 +69175,7 @@ adC
 adC
 adC
 adC
-ilE
+adC
 adC
 adC
 adC


JORJ949:
image
In the demo the light switch doesn't update

code/area.dm
```diff
@@ -3727,28 +3727,101 @@ ABSTRACT_TYPE(/area/station/catwalk)
...
+	// unlock sleeper doors
+	for (var/obj/machinery/door/airlock/M in by_type[/obj/machinery/door])
+		if (M.id == "Sleeper_Access")
+			M.req_access = null
+			M.req_access_txt = null
+
+	playsound(user.loc, 'sound/machines/shieldgen_startup.ogg', 70, pitch=0.6)
+
+	var/delay = 0
+	for (var/area/listeningpost/post_area in world)
```
JORJ949, line 3746: Could you not track these areas?

code/obj/machinery/door/door_control.dm
```diff
@@ -667,7 +667,9 @@ TYPEINFO(/obj/machinery/door_control)
 		return
 	playsound(src.loc, 'sound/effects/handscan.ogg', 50, 1)
 	if (istrainedsyndie(user))
-		src.toggle()
+		if (!global.listening_post_unlocked)
+			unlock_listening_post(user)
+		src.toggle(user)
```
JORJ949, line 672: You don't actually need to have this be a door controller anymore, the global unlock post proc will set the access.

JORJ949 (changes requested):

JORJ949:
Also you should almost definitely atomise all these changes

JORJ949:
Both cogmaps have some weirdly straight asteroids cut out to fit the listening post in.
Also important to note this means these posts won't show on goonhub

Sord213:
Ideally I'd like to see it use a variable in the map.dm options per map. Even maps that use non-standard listening posts. That way you don't have to scroll through each map to figure out "X map has a custom listening post to change, while Y and Z don't" , you could instead just edit 'listeningpost_oshan.dmm' 'listeningpost_standard.dmm' etc. Either way this is a very nice set of changes.

JORJ949:
image
The south sides of walls should connect to doors.

Sord213 (approved):
Mapping wise lgtm other than a minor errant forcefield on cogmap. 

For clarity though does it automatically boot up the listening post when a nukie teleports there from the cairngorm?

pgmzeta:
> Mapping wise lgtm other than a minor errant forcefield on cogmap.
> 
> For clarity though does it automatically boot up the listening post when a nukie teleports there from the cairngorm?

Yes, the moment anyone enters the post it will start up & unlock.

In addition, if the syndicate teleporter room is the first entered (i.e. teleported in from the carnigorm) the order of lights turned on is reversed, starting at the teleporter and ending at the front area.

TobleroneSwordfish (approved):
This is so cool, code looks good!

PR #23400, opened 2025-05-10
Stacked Material Sprites

LorrMaster:
[materials][qol]

## About the PR 
- Most ores now have different sprites depending on how many are in the stack
- New ore sprites can be used by Rocko
- Some new sprites for common refined materials
- Miracle matter is now given a randomized shape at round start. Options are ore, sphere, or torus (aesthetic only)
- Gemstones and uqill ore can fit inside pockets

Is likely not compatible with the following PRs:
#23065
#23067
#23075

## Why's this needed? 
- Visually display how many materials are in a stack
- Visual distinction for some common refined versions of materials
- More sprite variety for mining

Screenshot 2025-05-10 055035

rocko2

## Changelog
```changelog
(u)LorrMaster
(*)New sprites for ores at various stack sizes
```


code/obj/item/material.dm
```diff
@@ -29,15 +30,45 @@
...
+			name = "[src.name]"
+		else
+			name = "[amount] [src.name]s"
+
+	proc/get_stack_value() // Determines at what stack sizes the icon_state changes
+		// example: src.icon_state = "ore[x]_$$gold"
+		switch(src.amount)
+			if(1)
+				return 1
+			if(2 to 9)
```
TobleroneSwordfish, line 62: Could you scrunch these ranges down a little? Most material stacks are going to be in the 1-10 range and it seems wrong to only have two icon states for that.

LorrMaster:
I enlarged the tier 1 sprites to be more similar to the tier 2 sprites in size. I tried to avoid overdoing it, so let me know what you think. Also updated Rocko.

PR #23422, opened 2025-05-11
Local auth temporary PR

TobleroneSwordfish:
[DNM]
## About the PR 
Temporarily enables local authentication while the Byond hub is down again.
Remake of #19128 because the branch is gone and the code has changed slightly.

## Why's this needed? 
Byond hub down, no-one can log in.

### DIFF
--- a/code/client.dm
+++ b/code/client.dm
@@ -82,7 +82,7 @@
 	var/proc_logging = 0
 #endif
 
-	// authenticate = 0
+	authenticate = 0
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
Oh right, our CI also relies on Byond existing 💀

DisturbHerb:
> Oh right, our CI also relies on Byond existing 💀

but what if it didn't :]

Wirewraith:
No longer needed with https://github.com/goonstation/goonstation/pull/23530

PR #23426, opened 2025-05-11
Showing off an item includes overlays

JORJ949:
## About the PR 
Showing off an item now shows any overlays the item has, such as covered in blood

## Why's this needed? 
Showing off an item should show how it looks, rather than just base icon state

### DIFF
--- a/code/datums/actions/actions.dm
+++ b/code/datums/actions/actions.dm
@@ -2006,7 +2006,13 @@
 			src.pixel_x_offset = -src.pixel_x_offset
 			src.pixel_x_hand_offset = -src.pixel_x_hand_offset
 
-		var/image/overlay = src.item.SafeGetOverlayImage("showoff_overlay", src.item.icon, src.item.icon_state, MOB_LAYER + 0.1, src.pixel_x_offset, src.pixel_y_offset)
+		var/icon/flatIcon = getFlatIcon(src.item)
+
+		var/image/overlay = new /image
+		overlay.icon = flatIcon
+		overlay.pixel_x = src.pixel_x_offset
+		overlay.pixel_y = src.pixel_y_offset
+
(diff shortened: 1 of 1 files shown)


ZeWaka (commented):
Why aren't we using vis contents or something like that for this?

JORJ949:
> Why aren't we using vis contents or something like that for this?

I'm not sure what you're suggesting, theres no vis contents for images nor do they have overlays so only work for a flat icon afaik. Unless you mean changing it to add the showoff thing to the user's vis_contents?

ZeWaka:
> > Why aren't we using vis contents or something like that for this?
> 
>  But that would have you click the item and not the owner, meaning stealing, emagging, or otherwise messing with things being shown off which is :/

`VIS_INHERIT_ID` exists does it not?

code/datums/actions/actions.dm
```diff
@@ -2006,15 +2006,18 @@
...
+		var/image/hand_overlay = src.item.SafeGetOverlayImage("showoff_hand_overlay", 'icons/effects/effects.dmi', hand_icon_state, MOB_LAYER + 0.11, src.pixel_x_hand_offset, src.pixel_y_hand_offset, color=user.get_fingertip_color())
 		src.user.UpdateOverlays(hand_overlay, "showoff_hand_overlay")
 
 		src.user.set_dir(SOUTH)
 
 	onDelete()
 		. = ..()
-		src.user.UpdateOverlays(null, "showoff_overlay")
+		src.user.vis_contents -= src.item
+		src.item.vis_flags &= ~(VIS_INHERIT_ID | VIS_INHERIT_PLANE |  VIS_INHERIT_LAYER)
```
TobleroneSwordfish, line 2022: Needs to reset pixel offsets too, also ideally it should store initial pixel offsets and vis_flags and reload those rather than just resetting them.

PR #23427, opened 2025-05-11
ID Card Band overlays

JORJ949:
## About the PR 
Split ID card sprites into the base and the departmental stripe
Requires #23426 to be merged first or showing off will only show basic ID sprite.
Includes no (intended) player facing changes

## Why's this needed? 
Allows for more unique IDs to be made without needing to add more sprites than are needed.


JORJ949:
Requires and won't be worked on until #23426

PR #23434, opened 2025-05-11
Add large produce satchels to botany requisition rewards

Egregorious:
## About the PR 

Adds large produce satchels as guaranteed rewards for botany requisitions.

Makes other rewards slightly less guaranteed in exchange. Removes the plant cartridge reward. 

## Why's this needed? 

Botany requisition rewards are rather lacklustre; strange seeds can be obtained easier elsewhere, tumbleweeds - while cool - aren't particularly sought after, and plant cartridges are nigh-insulting. Large produce satchels are in the code, but as far as I can tell they're unavailable ingame. This distributes them in a somewhat limited capacity such that regular satchels aren't obsolete, gives more incentive for botany to cooperate with cargo and makes botany req rewards a bit more interesting. 

## Changelog 

```changelog
(u)Egregorious
(+)Botany requisitions now reward large produce satchels. 
```

### DIFF
--- a/code/modules/economy/requisition/rc_scientific.dm
+++ b/code/modules/economy/requisition/rc_scientific.dm
@@ -539,8 +539,7 @@ ABSTRACT_TYPE(/datum/rc_entry/item/organ)
 	name = "large produce satchel"
 
 	build_reward()
-		var satchel = new /obj/item/satchel/hydro/large
-		return satchel
+		return new /obj/item/satchel/hydro/large
 
 /datum/rc_itemreward/tumbleweed
 	name = "aggressive plant specimen"
--- a/code/modules/economy/requisition/rc_scientific.dm
+++ b/code/modules/economy/requisition/rc_scientific.dm
@@ -513,15 +513,6 @@ ABSTRACT_TYPE(/datum/rc_entry/item/organ)
 		..()*/
 
 
-
-
-/datum/rc_itemreward/plant_cartridge
-	name = "Hydroponics restock cartridge"
-
-	build_reward()
-		var/cart = new /obj/item/vending/restock_cartridge/hydroponics
-		return cart
-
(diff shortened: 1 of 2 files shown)


trazodont:
They are available in game currently but only as mail items, as a botanist, pls ty 🙏

code/modules/economy/requisition/rc_scientific.dm
```diff
@@ -534,6 +535,13 @@ ABSTRACT_TYPE(/datum/rc_entry/item/organ)
 			seed_list += new /obj/item/seed/alien
 		return seed_list
 
+/datum/rc_itemreward/large_satchel
+	name = "large produce satchel"
+
+	build_reward()
+		var satchel = new /obj/item/satchel/hydro/large
+		return satchel
```
pgmzeta, line 543:
```suggestion
		return new /obj/item/satchel/hydro/large
```

pgmzeta (approved):

code/modules/economy/requisition/rc_scientific.dm
Egregorious: I wanted to do that but then thought I should keep it closer to the codebase incase there's a language quirk I don't know about!

444Portal (approved):
makes sense to me!!

AnonymousUserTheOnly:
i wholeheartedly agree with this change, if made.  too often we've had large batch production but nowhere to put it.  or been growing a plant that's slow to grow but big harvests.  this would also give, as stated in the PR, reason for Botany to work with Cargo.  plus, it would make it easier to bulk transport massive amounts of foods botany has grown to the shuttle at shift end.  _looks around_  like my peanut butter and raspberry Subsistence Sandwiches, which i try to grow when i'm on botany.

PR #23442, opened 2025-05-12
Split Item spawner sprites out of object.dmi

JORJ949:
## About the PR 
Splits sprites for item spawners into a new item_spawn.dmi file instead of being in objects.dmi

Anything with any of the following icon states in secret will need updating to the new file
image

## Why's this needed? 
Big DMI files bad, the more organised they are the better


JORJ949:
Oh and I added the non-random "item spawn" text I used for the armory spawner sprites

PR #23445, opened 2025-05-12
Add a min Failure Chance to the Zip Gun

444Portal:
[BALANCE] [OBJECTS]
## About the PR
Makes the minimum failure chance for the Zip Gun 4%, this chance does two things:
- Stun Rounds now can blow up the Zip Gun.
- .22 is now slightly less safe. From 2% Failure Chance to 4%.
4% is an inbetween of the previous default chance - 6% - and the lowest chance you can get which is 2% with the .22 ammo from hacked General Fabricators.

## Why's this needed?
The Zip Gun currently can be used without much issue with the Detective's .38 Stun Loaders, which is fairly plentiful if you can scan or hack a SecTech. The slight nerf to .22 is a side-effect, but I feel it won't effect much and this slapped together piece of shit (positive) could blow up slightly more anyway.

## Changelog
```changelog
(u)444explorer
(+)The Zip Gun is now able to break with Stun Rounds at a 4% chance. .22 Ammo has also been bumped to break the gun at 4%.
```

### DIFF
--- a/code/obj/item/gun/kinetic.dm
+++ b/code/obj/item/gun/kinetic.dm
@@ -566,7 +566,7 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 				qdel(src)
 			return
 		if(ammo?.amount_left && current_projectile.power)
-			failure_chance = clamp(round(current_projectile.power/2 - 9), 0, 33)
+			failure_chance = clamp(round(current_projectile.power/2 - 9), 4, 33)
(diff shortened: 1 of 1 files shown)


pgmzeta:
stun rounds currently have a 1% chance to break the zip gun. this implementation also affects staples, which feels kind of weird.

TobleroneSwordfish:
I don't really see the need to nerf zipguns? They're still very limited by their 2 shot capacity and if you've managed to steal the contents of a sec-tech I think you should be rewarded for doing so.

444Portal:
I held more spite in my heart than I do now, I don't really get where I was going with this myself tea bee aich.

PR #23446, opened 2025-05-12
Give AIs the cyborg say sound

pgmzeta:
[silicon][qol]
## About the PR 
Give the AI the cyborg voice sounds. When using the radio, use the standard cyborg say voice instead of a default radio chirp.

## Why's this needed? 
the outbound default radio chirp is a consequence of the say rework and I feel it drowns out the ai radio sound. also the cyborg voice sound fits the ai better than the regular say sound imo

### DIFF
--- a/code/mob/living/silicon/ai.dm
+++ b/code/mob/living/silicon/ai.dm
@@ -95,6 +95,7 @@ TYPEINFO(/mob/living/silicon/ai)
 /mob/living/silicon/ai
 	name = "AI"
 	voice_name = "synthesized voice"
+	voice_type = "cyborg"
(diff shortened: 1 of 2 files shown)


JORJ949:
Didn't the AI have their own unique radio sound?

Mister-Moriarty:
> Didn't the AI have their own unique radio sound?

It does, this PR doesn't touch that. This modifies the sound played when speaking into the radio.

JORJ949:
> 
> It does, this PR doesn't touch that; it modifies the sound played when speaking into the radio.

So just the one played locally?

PR #23462, opened 2025-05-14
Adds 2 tidal generators to Oshan

cheekybrdy:
[BALANCE] [ENGINEERING] [CONTROVERSIAL] [INPUT-WANTED]
## About the PR 
Title
image
image

## Why's this needed? 
Geothermal power is rather hard to get people to do especially on lowpop, as it amplifies the isolation issues of engineering, for the bare minimum of other stations, making it unsatisfying without a full team to work with to give it its full glory. I doubt these will be enough to power the station independently but should be enough to stop or at least delay the problem for a significant enough time.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Adds 2 tidal generators to oshan, along with accompanying currents
```


cheekybrdy:
People end up voting for oshan for other reasons, then the engineers typically people who didn't vote for it just end up doing mechcomp, its less about the difficulty of geothermal, its people's motivation to do it.

Scaltra-Volpe:
Don't you need currents for turbines? Honestly I like having secondary power structures but I don't know if turbines without its relevant mechanic is the fit. 

I think another secondary powder grid option that could be unique like some power furnaces or even a few portable combustion engines could be interesting if we don't want to go the route of redesigning a whole new secondary system just for oshan. 

Personally I really like the idea of something though. Sometimes if you want to really push or experiment with geothermal ideally you don't want to be setting one up super early in a specific spot. 

Also nice for lower pop or a lack of engineers, basically any argument for why solars are nice to have existing on other stations right now. 

As an aside I will say the lack of their being a secondary power system on roundstart for oshan is not something I think really adds to it in anyway and IMO I think adding one would be nice.

cheekybrdy:
This adds current to accommodate the tidals

ZeWaka:
If people aren't doing geothermal maybe we should make it more interesting, not this

PR #23463, opened 2025-05-14
Change age and blood curse to non-gib deaths

pgmzeta:
[balance][game objects]
## About the PR 
Change blood curse end from implode gib to husking
Change age curse end from blob skeletonize to forcing skeleton mutantrace

## Why's this needed? 
AOE curses with total round-removal is a bit harsh. These changes preserve the brain while preventing direct cloning and maintain the spirit of the curse.


code/modules/status_system/statusEffects.dm
```diff
@@ -3122,9 +3125,13 @@
 				src.final_msg_given = TRUE
 				H.playsound_local(H, 'sound/ambience/spooky/Void_Calls.ogg', 75, FALSE)
 			if (H.bioHolder.age >= src.original_age + 120)
+				H.visible_message(SPAN_ALERT("[H] collapses into a pile of bones!"))
 				H.death(FALSE)
-				H.skeletonize()
-				src.linked_curser.lift_curse_specific(FALSE, H)
+				H.set_mutantrace(/datum/mutantrace/skeleton)
```
TobleroneSwordfish, line 3130: Does adding the mutantrace after death still make them fall apart or do they get stuck in a weird dead-but-intact skeleton state?

pgmzeta: switched order so they fall apart properly

TobleroneSwordfish (approved):

PR #23465, opened 2025-05-14
more turtle berets

JORJ949:
## About the PR 
Adds (non-commander) NT, Janitor and Syndicate berets to list of hats turtles can wear.
image

## Why's this needed? 
More beret options for pet turtles, sylvester'll probably never end up with anything other than their own beret so this mostly applies to random pet turtles.

## Changelog 

```changelog
(u)JORJ949
(*)Turtles can now wear NT, Head of Sanitation and Syndicate berets.
```


pgmzeta (commented):
Would it be possible to move these into overlays like the costumes instead of duplicating the entire sprite?

JORJ949:
> Would it be possible to move these into overlays like the costumes instead of duplicating the entire sprite?

Yes but I see no reason to do that right now and its easier to sprite new ones if its the whole sprite, making them their own sprites is fine until someone wants to make an alternate turtle base at which point then they should be overlays.

pgmzeta (approved):

PR #23479, opened 2025-05-15
New Trait: Butterfingers

JORJ949:
## About the PR 
Adds a new butterfingers trait that causes you to randomly drop whatever item is currently in your active hand!
Gives two points but has almost double the chance to activate that Kleptomaniac does.
Sprite by seasnakes on discord
image

## Why's this needed? 
New trait in a similar manner to Kleptomaniac but instead of randomly picking up stuff you randomly drop stuff.

## Changelog 

```changelog
(u)JORJ949 and seasnakes
(*)Added a new butterfingers trait that will randomly make you drop whatever you're holding!
```


code/datums/traits.dm
```diff
@@ -1295,6 +1295,31 @@ TYPEINFO(/datum/trait/partyanimal)
...
+/datum/trait/butterfingers
+	name = "Butterfingers"
+	desc = "You have difficulty keeping hold of things."
+	id = "butterfingers"
+	icon_state = "butterfingers"
+	points = 2
+	afterlife_blacklisted = TRUE
+
+	onLife(var/mob/owner, var/mult)
+		if(owner.stat && !owner.lying && can_act(owner))
```
TobleroneSwordfish: What is this check for? "Incapacitated but not lying and can act" doesn't really make sense to me. Generally it's best not to check `.stat` directly anymore because it's ambiguous.

code/datums/traits.dm
```diff
@@ -1295,6 +1295,31 @@ TYPEINFO(/datum/trait/partyanimal)
...
+			return
+		if(!probmult(10))
+			return
+		var/mob/living/carbon/human/H = owner
```
TobleroneSwordfish: Why restrict it to humans?

code/datums/traits.dm
```diff
@@ -1295,6 +1295,31 @@ TYPEINFO(/datum/trait/partyanimal)
...
+		if(!istype(H))
+			return
+
+		var/obj/item/target_item = H.equipped()
```
TobleroneSwordfish: Should affect the offhand too, see `find_type_in_hand` for a proc to get items in an arbitrary mob's hands

TobleroneSwordfish (changes requested):
Idea and sprite look good, a couple of code points:

code/datums/traits.dm
JORJ949: Cargoculted incorrectly from kleptomaniac

code/datums/traits.dm
JORJ949: I can't think of anything thats not a human mutantrace that would want to be affected by this, afaik only humans should normally end up with traits.

PR #23481, opened 2025-05-16
Newbee Tutorial

pgmzeta:
[feature][ui]
## About the PR 
Adds a playable tutorial, available from the main menu when the round is in-play.

Sound files by glowbold

## Why's this needed? 
Learn how to play a little easier.

## Demo
Beta 3 Demo:
https://www.youtube.com/watch?v=qaZ33So3_Zk

## TODO
- [x] Too-sudden end on completion
- [x] Utilize tutorial system instead of maptext objects
- [x] Read keymaps
- [x] Sprint tutorial
- [x] Laying down tutorial
- [x] Radio tutorial
- [x] Move/remove Rules tutorial?
- [x] Move/remove Help tutorial?

## Changelog 

```changelog
(u)Mr. Moriarty, LeahTheTech, Sord, & glowbold
(*)The new player options menu includes an in-game tutorial.
```


JORJ949:
Not sure destroying walls is the best thing to teach new players, I think it would be better to mention things like access like "Only Security has access to security.". Also should definitely dynamically change depend on customized controls.

code/modules/tutorials/newbee_tutorial.dm
```diff
@@ -0,0 +1,233 @@
...
+
+	New()
+		..()
+		if (!marker)
+			marker = image('icons/effects/VR.dmi', "lightning_marker")
+			marker.filters= filter(type="outline", size=1)
+
+/datum/tutorialStep/newbee/welcome
+	name = "Welcome to Space Station 13!"
+	instructions = "This tutorial will get you familiar with the basics of the game. If at any point you get stuck or want to restart the tutorial, use the 'emergency tutorial stop' verb in the <b>Commands</b> tab, top right."
```
ZeWaka, line 63: you probably know, but i don't see any of these when playing the tutorial
they also differ from the in-room maptext, which is probably not the best

ZeWaka: but yeah it's hard to do any review comments on the verbiage because it's all maptext, and i have comments

pgmzeta: I'd like to do more to convert this to use the tutorial steps in full, but the blob and flock tutorials use some dedicated code in their abilities to handle advancing steps. It should be possible to make the tutorial steps generate maptext instead of having it hard coded; my intention is to do that but I wanted to get something playable first.

Mister-Moriarty:
This is something that I've wanted to see for ages; thank for going to the effort of implementing it!
A few things though:
- The bit about click-dragging the toolbox may lead new players to initially think that toolboxes (and similar) can't be picked up.
- Maybe a point about being able to put things back on tables?
- Being able to use C to apply patches maybe should be mentioned, but clicking to apply them to yourself and others is fine.
- Be wary of lights in front of and obscuring maptext.

​
Other things you may wish to include:
- A short point about the rules.
- MHELP and AHELP.
- How to talk and use the radio. The second part might need a special radio that only broadcasts to itself.
- Lockers should probably be in there somewhere, as well as the importance of ID cards.

JORJ949:
How to put ID+cash in pda and pda in id slot may be a good inventory management section.

code/modules/tutorials/newbee_tutorial.dm
```diff
@@ -0,0 +1,282 @@
...
+			qdel(src)
+
+/datum/tutorial_base/regional/newbee/proc/AddNewbeeSteps()
+	src.AddStep(/datum/tutorialStep/newbee/welcome)
+	src.AddStep(/datum/tutorialStep/newbee/movement)
+	src.AddStep(/datum/tutorialStep/newbee/powered_doors)
+	src.AddStep(/datum/tutorialStep/newbee/items)
+	src.AddStep(/datum/tutorialStep/newbee/unpowered_doors)
+	src.AddStep(/datum/tutorialStep/newbee/hand_swap)
+	src.AddStep(/datum/tutorialStep/newbee/storage_inhand)
```
Sovexe, line 59:
```suggestion
	src.AddStep(/datum/tutorialStep/newbee/storage_inhands)
```
or vice versa

Sovexe:
The tutorial button only looks to be available post-game start. It would be good to have it available in similar fashion to declare ready, such that it pulls them in at round start, or barring that complexity the button just greyed out with a small note that it will be selectable at game start

TobleroneSwordfish:
I agree about the toolbox thing, toolbox interaction should be pick up -> press C, clickdrag is confusing and rarely used.
Also make sure the maptext is taking current hotkeys/control schemes into account!

pgmzeta:
tutorial ready now available at roundstart. had to redo joinmenu code to support it, but the result is the Cancel button now shows over the ready option you picked:
ready, ready with antag token, or play tutorial
Screenshot 2025-05-19 115708
Screenshot 2025-05-19 115702

flappybatpal:
Using the X to stop the tutorial and then restarting it has some interesting effects, constant runtimes on update air properties and sometimes the lighting overlays go weird, stacked I suspect. It could be a bit out of scope and related to the region allocator.

image

pgmzeta:
> Using the X to stop the tutorial and then restarting it has some interesting effects, constant runtimes on update air properties and sometimes the lighting overlays go weird, stacked I suspect. It could be a bit out of scope and related to the region allocator.

found the source of this in region allocation and fixed it up

> You can spam the emergency locker for supplies. 

fixed this and many other steps to be much less item-spammy by default. It'll try to relocate the objects and items inside first before spawning a new one. to tighten it more, i'd probably need to do some locate X across the entire zone before spawning. maybe doable but i don't know how worth.

pgmzeta:
fresh demo updated

pgmzeta:
* tutorial now throws a point arrow from your character to the HUD item it's highlighting
* you can no longer start the tutorial if the shuttle is in transit
* there's now a visual progress % :)

PR #23487, opened 2025-05-16
Remove invisible blue pixels from headsets

JORJ949:
## About the PR 
Removes the blue pixels on all headsets with an Alpha of 7
image

## Why's this needed? 
These pixels are invisible normally but become visible when you stack a bunch of headsets on top of each other


TemThrush:
I could be wrong but I'm pretty sure those are there to make them easier to pick up, having slightly invisible pixels on sprites like these is advised in the style guide. Rather than removing them wouldn't it be better to just lower the alpha more? 
"To fix this, add an almost fully transparent pixel in the gap. These are useful when you want something to be fully transparent, but still click-able."

JORJ949:
> I could be wrong but I'm pretty sure those are there to make them easier to pick up.
 
Oh yeah true

PR #23495, opened 2025-05-17
Add Automatic Arrest Flagging for Drunk Drivers in Cars

itsloley:
[vehicles][feature]
## About the PR 

if you're operating a car with the drunk status effect, you now automatically get set to arrest if you're seen by a camera!

dreamseeker_irrIOuRLam

---

## Why's this needed? 

it's pretty funny!

## Changelog 

```changelog
(u)Lyy
(*)Driving a car while drunk now automatically marks you for arrest.
```


JORJ949 (approved):
FREEZE. SCUMBAG.

FlameArrow57:
Probably best to make sure you're in vision of a camera when it happens?

itsloley:
> Probably best to make sure you're in vision of a camera when it happens?

since you're in a car, is_seen_by_camera always returns false (unless i missed something like an idiot)

FlameArrow57:
I might suggest putting the code at the end of the action bar for climbing into the car then

cheekybrdy:
Isn't there already a drink that does something similar to this? maybe make the drink one more extreme if this is becoming a default.

itsloley:
> Isn't there already a drink that does something similar to this? maybe make the drink one more extreme if this is becoming a default.

what drink is that actually? i'm not sure i've heard of it!

JORJ949:
> > Isn't there already a drink that does something similar to this? maybe make the drink one more extreme if this is becoming a default.
> 
> what drink is that actually? i'm not sure i've heard of it!

Beepskybrau (has some accents so search beepskybeer, its ID), though I can't find any special case for this anywhere and didn't find it when coding a proc for automated arrests. Though the effect could be in secret or removed or a wiki lie

flappybatpal:
It's not in secret

https://github.com/goonstation/goonstation/blob/master/code/modules/chemistry/Reagents-FoodDrink.dm#L532

itsloley:
updated to the automated arrest proc, seems to work just fine with camera coverage from my testing. honestly might have not worked before solely because of SKIP_CAMERA_COVERAGE i think

JORJ949:
> updated to the automated arrest proc, seems to work just fine with camera coverage from my testing. honestly might have not worked before solely because of SKIP_CAMERA_COVERAGE i think

Skip camera coverage does make everything "off cams" yes, I ran into the same issue :)

itsloley:
damn i really WAS being an idiot then LOL

code/modules/transport/pods/vehicle.dm
```diff
@@ -2223,6 +2223,18 @@ ABSTRACT_TYPE(/obj/machinery/vehicle/tank)
...
 		Install(new /obj/item/shipcomponent/locomotion/wheels(src))
 
+	handle_internal_lifeform(mob/lifeform_inside_me, breath_request, mult)
+		if (!ishuman(pilot) || !pilot.statusEffects)
+			return ..(lifeform_inside_me, breath_request, mult)
+
+		for (var/effect in pilot.statusEffects)
+			if (istype(effect, /datum/statusEffect/drunk))
+				var/mob/living/carbon/human/H = pilot
+				H.apply_automated_arrest("DUI.")
```
JORJ949: You'll need to make this either check before they get in or ignore cams as the oncam check now returns false if you aren't in a turf.

pgmzeta:
update with an alternate implementation when the boarding pod action bar ends.

this checks that it's a car, the person who entered the car is the pilot, and that they're drunk before applying the crime. it checks for camera visibility *before* the person enters, then uses that value once they're fully boarded so the pilot check works as expected

JORJ949:
Code looks fine, assuming it works without SKIP_CAMERA_COVERAGE

PR #23499, opened 2025-05-17
HoS Mug emag act

JORJ949:
## About the PR 
Adds an emag act to the HoS mug to make it say "Galaxy's Number TWO HoS!"
image

image

## Why's this needed? 
eheheheheheheheheheh crime


flappybatpal:
I would say it should make them do the hat stomp on discovering it but I expect that would be a can of worms.

FlameArrow57:
> I would say it should make them do the hat stomp on discovering it but I expect that would be a can of worms.

Would be fairly simple I think. Not sure if a forced hatstomp emote is favored though

FlameArrow57:
While silly, not sure if emag interaction makes sense considering the mug isn't an electronic device. Maybe it could be a pen/marker interaction instead, where you scribble out the 1 and replace it with a 2?

JORJ949:
> While silly, not sure if emag interaction makes sense considering the mug isn't an electronic device. Maybe it could be a pen/marker interaction instead, where you scribble out the 1 and replace it with a 2?

The flavor text on emagging is "You scratch some of the white paint off with the emag", similar to how you bore a hole in the back of a flash with it

JORJ949:
> I would say it should make them do the hat stomp on discovering it but I expect that would be a can of worms.

Would be simple to call the users hatstomp emote in getdesc

Lefinch:
"But I try harder"

ZeWaka (approved):

PR #23500, opened 2025-05-17
Fix wheelchair Help-Swap bug

JORJ949:
## About the PR 
After swapping places with someone set whatever you or they were buckled to to their new loc
Only unbuckle someone from a wheelchair when changing loc when the new loc is not the same as the rider

## Why's this needed? 
Fixes being able to swap with someone in a wheelchair but not with their chair, causing them to move out of their chair but remain buckled (More people should play wheelchair by the way, its fun and this bug would've been caught sooner!)

### DIFF
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -743,6 +743,11 @@ TYPEINFO(/mob)
 				src.set_loc(newloc)
 				tmob.set_loc(oldloc)
 
+				if(tmob.buckled)
+					tmob.buckled.set_loc(oldloc)
+				if(src.buckled)
+					src.buckled.set_loc(newloc)
+
 				if (istype(tmob.loc, /turf/space))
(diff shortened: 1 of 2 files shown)


code/mob.dm
```diff
@@ -743,6 +743,11 @@ TYPEINFO(/mob)
 				src.set_loc(newloc)
 				tmob.set_loc(oldloc)
 
+				if(tmob.buckled)
```
TobleroneSwordfish, line 746: This should check anchored too, shouldn't it?

TobleroneSwordfish (changes requested):

code/mob.dm
JORJ949: You can't swap places if either are anchored so you shouldn't be able to get here if they are.

JORJ949: image
In this big old chain of ands

TobleroneSwordfish (approved):
Ah the cursed and stack

PR #23508, opened 2025-05-18
Adds salvagers to roundstart classic

cheekybrdy:
## About the PR 
Readds salvagers to the classic antagonist pool, in the current state of this PR as currently both as a mixed roundstart spawn as well as a dedicated gamemode.

## Why's this needed? 
Salvagers is something that was initially available on classic and was removed due to validhunting issues, with the population decreases causing the culture to become more reserved, as well as general experience with salvagers as a role increasing over time, I believe, as well as quite a few others throwing out the idea in OOC that its worth a testmerge at the least, there is hesitance over dedicated salvs due to the initial quiet it causes, but I think both are worth testing.

## Changelog 

```changelog
(u)Cheekybrdy
(*)Readds Salvagers to classic, both as a roundstart mixed role and as a dedicated gamemode.
```


code/datums/gamemodes/mixed.dm
```diff
@@ -212,6 +223,14 @@
 			spy.special_role = ROLE_SPY_THIEF
 			possible_spy_thieves.Remove(spy)
 
+	if(num_salvagers)
+		var/list/possible_salvagers = get_possible_enemies(ROLE_SALVAGER,num_salvagers)
+		var/list/chosen_salvagers = antagWeighter.choose(pool = possible_salvagers, role = ROLE_SALVAGER, amount = num_salvagers, recordChosen = 1)
+		for (var/datum/mind/salvager in chosen_salvagers)
+			traitors += salvager
+			salvager.special_role = ROLE_SPY_THIEF
```
JORJ949, line 231:
```suggestion
			salvager.special_role = ROLE_SALVAGER
```

CalliopeSoups:
I've wanted this for a while and welcome salvagers back into the mixed fold, but as a standalone gamemode I don't think salvagers hold up on classic. Previously this is what I thought they were removed for. I'd still like a salvager gamemode, but it should probably have an extra traitor or two to go with them in my onion.

code/datums/gamemodes/mixed.dm
```diff
@@ -135,6 +142,10 @@
 				traitors += tplayer
 				token_players.Remove(tplayer)
 				tplayer.special_role = ROLE_ARCFIEND
+			if(ROLE_SALVAGER)
```
TobleroneSwordfish, line 145: This won't work unless you add `ROLE_SALVAGER` to the `traitor_types` list above.

config/gamemodes.sample.txt
```diff
@@ -21,4 +21,4 @@ PROBABILITY CONSTRUCTION 0
 PROBABILITY CONSPIRACY 0
 PROBABILITY SPY_THEFT 5
 PROBABILITY ARCFIEND 3
-PROBABILITY SALVAGER 0
\ No newline at end of file
+PROBABILITY SALVAGER 2
```
TobleroneSwordfish, line 24: This config file is a sample and isn't actually used by the server. Also I broadly agree with Soups that full salvager rounds don't work very well on classic. Maybe with some kind of extra firepower or incentive to go loud early.

TobleroneSwordfish (changes requested):

code/datums/gamemodes/salvager.dm
```diff
@@ -55,12 +68,31 @@
...
+	#endif
+
+	return 1
 
 /datum/game_mode/salvager/post_setup()
 	..()
 	for (var/datum/mind/salvager in salvager_minds)
 		equip_antag(salvager)
-
+	#ifndef RP_MODE
```
JORJ949, line 93: I'd change all these RP_MODE checks to a "spawns_distractions" variable on the mode itself which you have one rpmode check for at the start

PR #23510, opened 2025-05-18
Cluwne Chicken eggs now give people cluwnespider parasites

cheekybrdy:
## About the PR 
Title.

## Why's this needed? 
Cluwne eggs have had their effect removed for a long time, even though the hint is still there on the wiki. While the inital effect's removal could be summarised as self-antagging by cluwning people outright with eggs. This is another horrific take on it which gives them a worthwhile horrific effect.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Cluwne eggs now do horrifying things when ingested.
```

### DIFF
--- a/code/modules/medical/diseases/parasites.dm
+++ b/code/modules/medical/diseases/parasites.dm
@@ -86,7 +86,7 @@
 	affected_species = list("Human", "Monkey")
 	temperature_cure = INFINITY
 
-/datum/ailment/parasite/cluwnespiders/stage_act(var/mob/living/affected_mob, var/datum/ailment_data/D, mult)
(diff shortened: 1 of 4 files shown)


RubberRats (approved):
🤡🕷🐔

code/modules/medical/diseases/parasites.dm
```diff
@@ -79,6 +79,53 @@
 				affected_mob:gib()
 				return
 
+/datum/ailment/parasite/cluwnespider
+	name = "Unidentified Hilarity Body"
```
pgmzeta: looks good, one small change to add source ambiguity and freak doctors out :o)

```suggestion
	name = "Spider Eggs"
```

PR #23519, opened 2025-05-18
Get client fix for AI mainframes

### LINKED ISSUE #22705: AI cores are haunted
### Labels

[BUG][MAJOR]

### Description

If an AI's brain is swapped, the original AI can use re-enter corpse to possess their old body.

### Steps to reproduce

1. Remove the brain from an AI core.
2. Place a different brain in the core.
3. Have the original AI player use the re-enter corpse verb.
4. The core is occupied by two players.

### Expected Behavior

For the AI to not haunt their core.

Reported by: firekestrel
Client version: 515.1641
Active test merges: #21058
Round log date: 2025-03-24-00-19
Reported on: 2025-03-24 01:35:34
Round ID: 202824
Map: KONDARU

JORJ949:
## About the PR 
Adds a /mob/proc/get_client() proc that returns the mob's client, AI cores override this by returning the client of their eye or deployed shell as while deployed AI.client is null.
Use the above proc to check if a mind can be transferred to a body, preventing a client being put into an AI core while that AI is off in their eye or shell.
Also check get_client() before returning to a corpse.

## Why's this needed? 
AIs can have multiple mobs while their main mob has no client, allowing return to corpse and other AI mindbreaking shenanigans
Fixes #22705 and probably other AI-based client bugs

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -91,7 +91,7 @@ datum/mind
 			Z_LOG_ERROR("Mind/TransferTo", "Tried to transfer mind [(current ? "of mob " + key_name(current) : src)] to qdel'd mob [new_character].")
 			CRASH("Tried to transfer mind [identify_object(src)] to qdel'd mob [identify_object(new_character)].")
 
-		if (new_character.client)
+		if (new_character.get_client() && (new_character.get_client() != current.get_client()))
 			if (current)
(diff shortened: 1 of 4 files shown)


JORJ949:
> this seems really sus
> why do we need this whole new proc

May be useful elsewhere with other AI client bugs that keep cropping up and I didn't want to copypaste this code into both mind transfer and body return code

JORJ949:
Theres a cleaner fix for this somewhere

PR #23521, opened 2025-05-18
rebalance genetics cold/fire/thermal res

pgmzeta:
[balance][medical]
## About the PR 
* Thermal resistance now calculates the additional cold/heat resistance based on the power; from 50 base to 25 base / 50 empowered
* Fire and cold resistance now provide the relevant mob properties; from 0 base to 25 base / 50 empowered

## Why's this needed? 
* Giving out stable thermal in the booth reliably allows for the entire station to ignore wearing space suits in the space video game as basic clothing can get you close to 50% cold res
* While cold and fire res do offer benefits, empowering them feels like it does very little. This gives them some base resistances and scales based on empowering..

## Changelog 

```changelog
(u)glowbold
(+)Fire, cold, and thermal resistance provide 25% base resistance, doubled when empowered. 
```


TDHooligan:
this being overhauled to not cover 99% of cases would need better communication to players.

a good chunk, if not the majority of players, have no idea how much 90% cold resistance helps. whether it's a lower comfortable temperature or just 'you freeze to death slower'. the current buff is large enough to do the job of completely resisting fire/cold in 95% of scenarios, so players probably perceive it as immunity. a more gradual implementation of fire/cold resist should come alongside a better status effect or HUD element to show it is less binary than 'freezing/burning'

for auras. flame resistance is a big deal. fire is caused explicitly by player actions, and not knowing whether 2/3rds of the station is fire resistant or not is a big deal if you're using a flamethrower. cold can go, sure, though it might be annoying that it's inconsistent with flame res.

AnonymousUserTheOnly:
i'm mixed about this.  on the one hand, the aura removal means you can't be sure if someone's a ling or just gened up the wazoo...  but on the other hand, the reduction in percents makes it so they don't do what they seem like they should, let you resist the extreme temperatures to a greater degree.  and while it might no be enough on it's own to get you to full spacewalk immunity, the clothing worn also has some resistance values.  will all the clothing resistances be rebalanced too?  will they all get reduced as well to compensate for this reduction?  or will fire/cold/thermal resist get more features to make them worth having apart from just resisting the temperatures to a lesser degree?

Spyritdragon:
I feel like established practice is really the main reason we read 'resistance' as 'immunity' - given a little bit of time I think that's likely to go away without seeming too unclear afterwards, and a 25 buff on top of normal clothing will still make you very resistant to both outside of cases of outright being extremely on fire or in actual cold vacuum. Regardless, I think just for the feeling of the game having it be resistance over immunity by default would be a lot nicer than the current ease of getting outright immunity to stark vacuum or a raging plasmafire. 

I do think a rename of 'fire resistance' to 'heat resistance' or something similar would be clearer though.

PR #23523, opened 2025-05-19
Merge tgui_core_update_styling back into tgui_core_update

mordent-goonstation:
[UI][FUCK]

## About the PR 

Technically builds now (🎉), which was why this was broken off from `tgui_core_update` in the first place. So let's get rid of *one* level of spiking for the 516 work.

Updates a *lot* of tooling, mostly for parity with upstream (tgstation) tgui, e.g. webpack -> rspack and jest -> vitest, styling done via CSS vars instead of SCSS vars (technically not true, there's a *lot* of styling that needs updating still, but it builds now). This isn't all styling work, don't let the branch name fool you.

The `--bench` command doesn't seem to work, but we'll burn that bridge when we get to it.

@ZeWaka have fun reviewing.

- Adds several `[516 TODO]` comments that will need looking at and addressing prior to us being done with 516 upgrade
- We have a bunch of references to our version of files when we could use `tgui-core` for them now, and remove our versions (probably, I've not verified compatibility)

## Why's this needed? 

We would have done it on `tgui_core_update` if it wasn't a huge slice of work that would have made that branch even more unstable in the meantime. Again, a bunch of this is nothing to do with styling any more, and just build/parity updates that were kind of a knock-on for the sake of the styling work/completeness.


ZeWaka:
Bench didn't work before lul

PR #23524, opened 2025-05-19
tgui minor/patch version bumps, for completeness

mordent-goonstation:
[UI][Code Quality]

## About the PR 

Merge to the targeted branch first if you want I guess, or merge that PR first then `tgui_core_update` back into this one. It's branches and breaking changes all the way down!

## Why's this needed? 

- Gets us onto latest minor/patch versions of stuff. As far as I can tell they seem fine for now, being bleeding edge good (I didn't do major versions, heck to that).
- eslint stops complaining about the later version of TS than it knows about, so that's a good thing. Yes, I could add a dependency resolution (it was there before), but using supported versions good.


ZeWaka:
i don't care about this pr

PR #23525, opened 2025-05-19
Flip welding mask up

Kaan-NN:
[GAME OBJECTS] [PLAYER ACTIONS] [QOL]
## About the PR 
This allows to flip the welding mask up and down on a 1 second cooldown.

## Why's this needed? 
QoL

## Changelog 

```changelog
(u)Kaan
(+)Nods now push a welding mask back on your head.
```

### DIFF
--- a/code/obj/item/clothing/helmets.dm
+++ b/code/obj/item/clothing/helmets.dm
@@ -725,6 +725,9 @@ TYPEINFO(/obj/item/clothing/head/helmet/camera)
 				if (src.up)
 					src.flip_down(source, silent=TRUE)
 					boutput(source, SPAN_HINT("You nod, dropping the welding mask over your face."))
+				else
+					src.flip_up(source, silent=TRUE)
+					boutput(source, SPAN_HINT("You nod, raising the welding mask off your face."))
 
 	proc/obscure(mob/user)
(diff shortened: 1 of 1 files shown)


Connie-Dun:
Reposting the reasons I don't like the change, as I earlier posted in the discord, rearranged a bit in order of importance:

1. The slight inconvenience is charming. funny. just nodding again undoing it loses that bit of "aw dammit I did it again" imo,
2. Activating mask not being as easy to undo as to do is a balance little thing, even with it being such a small difference (hotkey vs click particular part), but it's relevant when using it for combat - being able to instantly start blocking flashes at the cost of visibility is pretty strong, and needing to move away from clicking stuff around you (like your enemy or your inventory) to undo it helps balance it - just hitting the *nod again to undo makes the visibility decrease trivial
3. Plain doesn't make sense when thinking of the physicality of it

itchybones:
I see the good intention, and I can't say that I've never wished for something like this, nut Connie's made some good points.

pgmzeta:
why not have it go up on the flip emote

flappybatpal:
I think the arguments made here are solid as for why it would be a bit too effective to have this be easier to toggle.

PR #23530, opened 2025-05-20
Client authentication refactor

Wirewraith:
## About the PR

Refactor of client initialization logic to support different authentication methods.

## Why's this needed?

Sometimes we don't want to authenticate with the Byond hub. Like, say, when it happens to be unavailable for multiple weeks.

## Changelog

```changelog
(u)Wire
(*)Authentication refactor to support external providers that aren't the Byond hub.
```


code/client/auth.dm
```diff
@@ -0,0 +1,49 @@
+// in order of execution
+var/list/datum/client_auth_gate/pre_auth_gates = list(
+	new /datum/client_auth_gate/telnet,
+	new /datum/client_auth_gate/guest,
```
ZeWaka, line 4: Just a reminder that we'll need to support Guest accounts if byond continues being down.

See: anyone without a prior account, or who isn't in the connected ckey cache thing byond has.

PR #23531, opened 2025-05-20
Only humans can use borg docks from the outside

JORJ949:
## About the PR 
Cyborg docking stations now only allow humans to use them from the outside, silicons may still use it from the inside.
Also add a disabled check to self service toggle.

## Why's this needed? 
Stops the AI and other things such as the detective's hat from being able to use cyborg docks. Presently the AI can just eject  or mess with borgs (or victims, in terms of conversion chamber) remotely, increase reliability in roboticists in having to manage the docks as only a human or a silicon occupant can eject someone.

## Changelog 

```changelog
(u)JORJ949
(*)Nonhumans can no longer use cyborg docks from the outside, borgs and shells can still use it from the inside.
```


JORJ949:
Closing in favour of a much better solution

PR #23539, opened 2025-05-20
Add Lattice artifact origin

FlameArrow57:
[SCIENCE][FEATURE]

## About the PR 
Adds a new artifact origin, Lattice artifacts

The origin is themed around mysteriousness, bells/chimes, and the unknown. The background behind them is that they are unclassified, unknown artifacts that no one knows the origins of, but strangely happen to manipulate other artifacts rather than have their own effects, suggesting they were built for that purpose. The origin is reserved for artifacts that manipulate other artifacts in some way.

## Why's this needed? 
Would be cool, also precursor to new artifacts

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new Lattice artifact origin. Activator keys will appear under this origin and are the only one for now.
```


pgmzeta (changes requested):
balance feedbacK:

The container change is off the mark. It makes the other container origins a noob-trap with no good loot, and the lattice sprite for containers don't feel like they'd contain anything.

This also makes artifact keys recognizable on sight with no potential alternatives, which makes them require no testing to mark correctly and send off. That however may be addressed later on if more hand-held lattice arts are added.

art feedback:

The FX sprites appear to completely duplicate the artifact sprite instead of acting like a glowing overlay that can change in colour.

Overall not a fan of the art direction; they're a combination of looking man-made and mystical that doesn't fit in with other origin types. Miiiight just be shock from a new artifact type, and if the FX are updated to match how other artifacts work it may fit in better.

The key will need in-hand sprites added to `icons/mob/inhand/hand_general.dmi`

FlameArrow57:
>
> This also makes artifact keys recognizable on sight with no potential alternatives, which makes them require no testing to mark correctly and send off. That however may be addressed later on if more hand-held lattice arts are added.
>

There will be future handheld lattice artifacts

> 
> The FX sprites appear to completely duplicate the artifact sprite instead of acting like a glowing overlay that can change in colour.
> 
Not an exact duplicate, the bells are swinging around?

>
> ...instead of acting like a glowing overlay that can change in colour.... if the FX are updated to match how other artifacts work it may fit in better.
> 

I softly disagree on this, part of lattice artifact purpose is to solely affect other artifacts, they were made for that purpose. I think it'd be reasonable that they should look different from usual artifacts? Also see Clockwork artifacts, which don't use the glow effects

Roninnozlo:
they're right about the clockwork artifacts yaknow

FlameArrow57:
I went with using existing sprites while the new sprites are worked on

FlameArrow57:
Ready for review

TobleroneSwordfish (approved):
New sprites look okay, it would be nice if there were more than one each for handheld and large artifacts but that can come later since this is only limited to keys right now.

pgmzeta (approved):

PR #23548, opened 2025-05-21
Mimic stomach azone, chute trap and eat limb abilities

CalliopeSoups:
[critters] [feature]
## About the PR 
This PR does a handful of things. Antag player mimics now spawn with a stomach allocated zone and have an ability to turn themself inside out to enter it while inside a disposal chute, healing them. Any players who enter the chute while the mimic is in this state will have a random limb eaten, on a cooldown. 

Mimics are also given an eat limb ability, letting them eat limbs without cooldown off the floor, or with a cooldown after gnawing a random one off of a person. Eaten limbs will appear in the stomach azone, and on death will be barfed out one by one. 

Ability icons mostly by Gomp!

https://github.com/user-attachments/assets/a0033ed2-25fc-477d-be14-ce8360f0a851

## Why's this needed? 
Mimics lack an objective outside maybe stabbing people to death with a fork and being funny. This gives them a relatively impactful thing to be doing, and gives them a reason to be tackling and stinging people other than survival.

https://hackmd.io/eGzsEO2ATjCnPNd_6S9RwA

## Changelog 

```changelog
(u)CalliopeSoups, Gomp
(*)Mimics come with a stomach they can retreat inside of while in a disposal chute and can eat limbs off of people!
(+)Mimics barf up every limb they've eaten on death.
(+)Disposal chutes are trapped while a mimic is inside and in their stomach, don't lose a limb!
```


code/mob/living/carbon/human.dm
```diff
@@ -2472,6 +2472,12 @@ Tries to put an item in an available backpack, belt storage, pocket, or hand slo
 	else if(limb == "r_arm" && src.limbs.r_arm) src.limbs.r_arm.sever()
 	else if(limb == "l_leg" && src.limbs.l_leg) src.limbs.l_leg.sever()
 	else if(limb == "r_leg" && src.limbs.r_leg) src.limbs.r_leg.sever()
+	else if(limb == "random")
+		var/randLimb = (pick("l_arm", "r_arm", "l_leg", "r_leg"))
+		if (src.has_limb(randLimb))
+			src.sever_limb(randLimb)
+		else
+			src.sever_limb("random")
```
Azrun, line 2480: Consider building a list of possible limbs to pick from instead of relying on possibly non-terminating recursion.

code/mob/living/carbon/human.dm
```diff
@@ -2472,6 +2472,12 @@ Tries to put an item in an available backpack, belt storage, pocket, or hand slo
 	else if(limb == "r_arm" && src.limbs.r_arm) src.limbs.r_arm.sever()
 	else if(limb == "l_leg" && src.limbs.l_leg) src.limbs.l_leg.sever()
 	else if(limb == "r_leg" && src.limbs.r_leg) src.limbs.r_leg.sever()
+	else if(limb == "random")
```
Azrun, line 2475: Can this be hit? Random isn't in the white list above

code/datums/abilities/critter/stomach_retreat.dm
```diff
@@ -0,0 +1,61 @@
+/datum/targetable/critter/stomach_retreat
+	name = "Retreat to Stomach"
+	desc = "Turn yourself inside out for shelter! Must be inside a disposal chute."
+	icon_state = "mimic_stomach_retreat"
+	cooldown = 120 SECONDS
+	cooldown_after_action = TRUE
+	needs_turf = FALSE
+	var/inside = FALSE
+	var/obj/machinery/disposal/current_chute = null
```
Azrun, line 9: Unsure of feasibility but could this use `COMSIG_ATOM_ENTERED` instead? That way it could be used on an arbitrary atom, and a whitelist could live here? Then they could be inside crates... or lockers...

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -0,0 +1,98 @@
...
+				boutput(holder.owner, "Your target doesn't have any limbs! Did you do that?")
+				return
+			else
+				var/randLimb = list("random_limb_string" = pick("l_arm", "r_arm", "l_leg", "r_leg"))
+				var/obj/item/parts/human_parts/targetLimb = targetHuman.limbs.get_limb(randLimb["random_limb_string"])
+				if (targetLimb)
+					boutput(world, SPAN_ALERT("<b>[holder.owner] starts to gnaw at [targetLimb]!</b>"))
+					actions.start(new/datum/action/bar/icon/eat_limb(holder, targetLimb, holder.owner, TRUE), holder.owner)
+				else
+					src.cast(target)
```
Azrun, line 28: Consider building a list of possible limbs to pick from instead of relying on possibly non-terminating recursion.

flappybatpal:
I'm not sure about losing a limb like this. It can be very debilitating and the number of ways it can occur is pretty limited outside of explosions.

CalliopeSoups:
Might be a good idea to testmerge this one if it goes through, might be bugs that can't be tested on a solo instance.

CalliopeSoups:
As far as container trap limb eating is concerned I may bump the cooldown up, a later PR will add another mimic mechanic that entering a trapped container could trigger instead of eating a limb, so I'll switch it to that if both go through.

code/datums/abilities/critter/stomach_retreat.dm
```diff
@@ -0,0 +1,95 @@
...
+					if (!T.z || isrestrictedz(T.z))
+						boutput(holder.owner, SPAN_ALERT("You are forbidden from using that here!"))
+						return TRUE
+					// Attempt entry via disposal machinery OR a disconnected disposal pipe
+					if (target_container.type in trap_whitelist)
+						current_container = target_container
+						current_container.present_mimic = holder.owner
+						activate()
+					else if (target_container.present_mimic)
+						boutput(holder.owner, SPAN_ALERT("There's already a mimic in here!"))
```
Azrun, line 35: This needs to be part of the above check as this should only be performed on objects that could hold a mimic AND this order will always replace an existing mimic.
(Exception occurs when using ability on turf)

code/datums/abilities/critter/stomach_retreat.dm
```diff
@@ -0,0 +1,95 @@
...
+		else
+			switch(tgui_alert(holder.owner, "Retreat into yourself to heal?", "Retreat to Stomach", list("Yes.", "No.")))
+				if ("Yes.")
+					var/turf/T = get_turf(holder.owner)
+					var/obj/target_container = holder.owner.loc
+					if (!T.z || isrestrictedz(T.z))
+						boutput(holder.owner, SPAN_ALERT("You are forbidden from using that here!"))
+						return TRUE
+					// Attempt entry via disposal machinery OR a disconnected disposal pipe
+					if (target_container.type in trap_whitelist)
```
Azrun, line 30: This needs to use `istypes`
(Was unable to stomach_retreat in `/obj/storage/closet` )

Azrun (changes requested):

CalliopeSoups:
> What aspects do you feel can't be tested in a solo instance?

I was thinking there could be things I didn't think about when it comes to several mimics doing things at the same time, but it shoould be fine on looking again.

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -0,0 +1,97 @@
...
+				actions.start(new/datum/action/bar/icon/eat_limb(holder, limb, holder.owner), holder.owner)
+		else if (ishuman(target))
+			var/mob/living/carbon/human/targetHuman = target
+			if (!targetHuman.limbs)
+				boutput(holder.owner, "Your target doesn't have any limbs! Did you do that?")
+				return
+			else
+				var/list/randLimbBase = list("r_arm", "r_leg", "l_arm", "l_leg")
+				var/list/randLimb = null
+				for (var/L in randLimbBase)
```
TobleroneSwordfish: Non-standard single letter variable name

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -0,0 +1,97 @@
...
+		if (istype(target, /obj/item/parts/human_parts))
+			var/obj/item/parts/human_parts/limb = target
+			if (limb.mimic_edible)
+				actions.start(new/datum/action/bar/icon/eat_limb(holder, limb, holder.owner), holder.owner)
+		else if (ishuman(target))
+			var/mob/living/carbon/human/targetHuman = target
+			if (!targetHuman.limbs)
+				boutput(holder.owner, "Your target doesn't have any limbs! Did you do that?")
+				return
+			else
```
TobleroneSwordfish: You don't need an else if you're returning from inside the if

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -0,0 +1,97 @@
...
+	onStart()
+		..()
+		user.stop_hiding()
+		user.last_disturbed = INFINITY
+
+	onUpdate()
+		..()
+		last_crunch++
+		if (last_crunch >= 2)
+			var/datum/human_limbs/T = target
```
TobleroneSwordfish: Non-standard single letter variable name

code/datums/abilities/critter/stomach_retreat.dm
```diff
@@ -0,0 +1,97 @@
...
+
+	proc/trap_chomp()
+		var/mob/living/carbon/human/target = locate(/mob/living/carbon/human) in current_container
+		if (!target)
+			return
+		if (GET_COOLDOWN(current_container, "mimicTrap"))
+			boutput(target, SPAN_ALERT("<B>You narrowly avoid something biting at you inside the [current_container]!</B>"))
+			return
+
+		ON_COOLDOWN(current_container, "mimicTrap", 20 SECONDS)
```
TobleroneSwordfish:
```suggestion
		ON_COOLDOWN(current_container, "mimicTrap", 5 SECONDS)
```
im evil

code/datums/abilities/critter/stomach_retreat.dm
```diff
@@ -0,0 +1,97 @@
...
+		var/list/randLimbBase = list("r_arm", "r_leg", "l_arm", "l_leg")
+		var/list/randLimb
+		for (var/L in randLimbBase) // build a list of limbs the target actually has
```
TobleroneSwordfish: Same here

code/datums/abilities/critter/stomach_retreat.dm
```diff
@@ -0,0 +1,97 @@
...
+			if (target.limbs.get_limb(L))
+				LAZYLISTADD(randLimb, L)
+		var/obj/item/parts/human_parts/targetLimb = target.limbs.get_limb(pick(randLimb))
+
+		if (targetLimb)
+			var/obj/item/limb = targetLimb.sever()
```
TobleroneSwordfish: This should probably do some damage too.

code/datums/components/mimic_stomach.dm
```diff
@@ -0,0 +1,57 @@
...
+	limb_target_turf = get_turf(pick(non_walls))
+	center = region.get_center()
+
+/datum/component/mimic_stomach/proc/on_entered()
+	var/mob/parent = src.parent
+	parent.HealBleeding()
+	parent.HealDamage("All", parent.max_health, parent.max_health)
+	return
+
+/datum/component/mimic_stomach/proc/on_disposing()
```
TobleroneSwordfish: Appears unused?

code/datums/components/mimic_stomach.dm
```diff
@@ -0,0 +1,57 @@
...
+				T.ReplaceWith(/turf/unsimulated/wall/setpieces/bloodwall)
+			else
+				LAZYLISTADD(non_walls, T)
+			var/chance = rand(1,6)
+			if (chance == 3)
+				make_cleanable(/obj/decal/cleanable/blood/gibs, T)
+	limb_target_turf = get_turf(pick(non_walls))
+	center = region.get_center()
+
+/datum/component/mimic_stomach/proc/on_entered()
```
TobleroneSwordfish: This isn't hooked up to anything and doesn't work

code/mob/living/critter.dm
```diff
@@ -87,6 +87,9 @@ ADMIN_INTERACT_PROCS(/mob/living/critter, proc/modify_health, proc/admincmd_atta
 	var/ghost_spawned = FALSE //Am i inhabited by a ghost player who used the respawn critter option?
 	var/original_name = null
 
+	/// region holder for eat_limb and stomach_retreat
+	var/datum/component/mimic_stomach/stomachHolder = null
```
TobleroneSwordfish: This is not how components are supposed to work. Use signals (or at a stretch, `GetComponent`) to communicate between them, don't hold direct references.

code/obj.dm
```diff
@@ -14,6 +14,9 @@
 	var/move_triggered = 0
 	var/object_flags = 0
 
+	/// for if a mimic has trapped the obj, for things like crates and chutes
+	var/mob/living/critter/mimic/antag_spawn/present_mimic = null
```
TobleroneSwordfish: Don't put this on `/obj`, make it a component like `/datum/component/hidden_mimic` orr wrap it into the current mimic stomach component and add that to the storage object instead of the mimic (that probably makes more sense anyway?)

code/datums/abilities/critter/stomach_retreat.dm
```diff
@@ -0,0 +1,97 @@
...
+					// Attempt entry via disposal machinery OR a disconnected disposal pipe
+					if (target_container.present_mimic)
+						boutput(holder.owner, SPAN_ALERT("There's already a mimic in here!"))
+						return TRUE
+					if (istypes(target_container, trap_whitelist))
+						current_container = target_container
+						current_container.present_mimic = holder.owner
+						activate()
+					else
+						boutput(holder.owner, SPAN_ALERT("There isn't anything to climb into here!"))
```
TobleroneSwordfish: This error should present before the dialogue box

code/datums/components/mimic_stomach.dm
```diff
@@ -0,0 +1,57 @@
+ABSTRACT_TYPE(/datum/component/mimic_stomach)
+/datum/component/mimic_stomach
+	dupe_mode = COMPONENT_DUPE_UNIQUE
+	var/datum/allocated_region/region
+	var/turf/center
+	var/turf/limb_target_turf
+	var/list/non_walls
+	var/last_appearance
+	var/atom/exit
+	var/list/obj/item/parts/limbs_eaten
```
TobleroneSwordfish: Some of these could definitely do with doc comments

TobleroneSwordfish (changes requested):
This is very cool and very grody, a few code issues though:

- Spawning a mimic causes a runtime, please test your code with runtime errors enabled!
image
- Using the "retreat" ability causes this runtime:
image
- image
The mimicdummy!
- Not sure if it's from this PR but it's weird that the mimic's default form is too big to fit in disposal pipes.

CalliopeSoups:
There's a bug I'm struggling fixing where dropping an item inside the stomach will throw a crash error here.
Code_SjMWAwIYiE
This was working before... I suspect it's to do with the add_filter in mimic update_icon or something similar? I'll see if I can figure it out tomorrow.

Edit: Okay I'm a bit stumped actually, I've tested removing the last_appearance/mimicdummy changes when retreating into the stomach, and also the add_filter in update_icon. Same crash with both gone. I'm gonna just make you drop whatever you're holding on the floor before being teleported into the azone cause that probably gets around some jank with bringing items in anyway, but it'll probably be a good idea to figure out why this happens still.

Edit edit: Note to self that this error only happens the first time you bring in an item from the outside, it doesn't throw another error if you pick it back up and drop it again.

CalliopeSoups:
> This is very cool and very grody, a few code issues though:
> 
>     * Spawning a mimic causes a runtime, please test your code with runtime errors enabled!
>       image
> 
>     * Using the "retreat" ability causes this runtime:
>       image
> 
>     * image
>       The mimicdummy!
> 
>     * Not sure if it's from this PR but it's weird that the mimic's default form is too big to fit in disposal pipes.

I believe I've done both of these with runtimes activated now and not gotten crashes, let me know if it's any different for you.

CalliopeSoups:
Trap chomp seems to be broken, figuring it out at the moment.

Edit: I missed a couple things actually, should have them fixed soon.

CalliopeSoups:
There's one small change to the sever() proc, adding a toggle var for the dir setting section of the proc. If this code is allowed to run when severing a limb for trap_chomp it will throw an error. Let me know if there's a better way to go about that.

The new code uses GetComponent a lot in general as well, so let me know where (and possibly how) I could cut those down if needed.

CalliopeSoups:
Quick showcase. I've tested other things such as trying to retreat while there's already a retreated mimic inside, using the abilities on non-mimics etc but it's very possible there's more I missed. Expecting another round of reviews then completion hopefully.

https://github.com/user-attachments/assets/62384f05-49e6-47dc-8feb-cda0b02b6567

code/datums/components/stomach_barf.dm
```diff
@@ -0,0 +1,26 @@
+/datum/component/death_barf // interacts with mimic_stomach
```
TobleroneSwordfish, line 1: This doesn't seem to have any reason to be a component, since it can only be interacted with by direct calls from mimic code. Components are meant to be generic and self contained, meaning they can be applied to more than one type without issue. If your component can only ever be used with one type then the behaviour should just be part of that type.

code/mob/living/critter.dm
TobleroneSwordfish: ^

code/datums/components/mimic_stomach.dm
```diff
@@ -0,0 +1,119 @@
+ABSTRACT_TYPE(/datum/component/mimic_stomach)
```
TobleroneSwordfish:
```suggestion
```
You're instantiating this, it shouldn't be abstract. https://en.wikipedia.org/wiki/Abstract_type

code/datums/abilities/critter/stomach_retreat.dm
```diff
@@ -0,0 +1,76 @@
...
+		if (inside)
+			switch(tgui_alert(holder.owner, "Leave yourself?", "Retreat to Stomach", list("Yes.", "No.")))
+				if ("Yes.")
+					deactivate()
+					return TRUE
+				if ("No.")
+					return TRUE
+		var/obj/target_container = holder.owner.loc
+		if (!isobj(target_container))
+			boutput(holder.owner, SPAN_ALERT("There isn't anything to climb into here!"))
```
TobleroneSwordfish: Kind of misleading, should indicate that you must be already inside something to cast the ability

code/datums/abilities/critter/stomach_retreat.dm
```diff
@@ -0,0 +1,76 @@
...
+			holder.owner.AddComponent(/datum/component/death_barf)
+
+	cast(atom/target)
+		. = ..()
+		if (inside)
+			switch(tgui_alert(holder.owner, "Leave yourself?", "Retreat to Stomach", list("Yes.", "No.")))
+				if ("Yes.")
+					deactivate()
+					return TRUE
+				if ("No.")
```
TobleroneSwordfish:
```suggestion
			switch(tgui_alert(holder.owner, "Leave yourself?", "Retreat to Stomach", list("Yes", "No")))
				if ("Yes")
					deactivate()
					return TRUE
				if ("No")
```

code/datums/components/mimic_stomach.dm
```diff
@@ -0,0 +1,119 @@
...
+				LAZYLISTADD(non_walls, T)
+			var/chance = rand(1,6)
+			if (chance == 3)
+				make_cleanable(/obj/decal/cleanable/blood/gibs, T)
+	limb_target_turf = get_turf(pick(non_walls))
+	center = region.get_center()
+
+/datum/component/mimic_stomach/proc/on_entered(mob/user)
+	if (istype(user, /mob/living/critter/mimic))
+		user.HealBleeding()
```
TobleroneSwordfish:
```suggestion
		user.HealBleeding(INFINITY)
```
This needs an amount param to work

TobleroneSwordfish (changes requested):
Generally looks good, the allocated zone is cleaning up properly, last few issues here:
- Entering the stomach zone after eating one limb gives this: image
- You can eat limbs while inside the stomach, intentional?

CalliopeSoups:
> Generally looks good, the allocated zone is cleaning up properly, last few issues here:
> 
>     * Entering the stomach zone after eating one limb gives this: image
> 
>     * You can eat limbs while inside the stomach, intentional?

It was doing this cause it was trying to add everything in the contents of the user, including their organs and limbs. As for eating limbs inside the stomach, that wasn't intentional and it's fixed!

PR #23549, opened 2025-05-21
Hidden mimics can't be shoved around 

CalliopeSoups:
[critters] [balance] [qol]
## About the PR 
Basic PR has the mimic copy its disguise's density when hiding.

## Why's this needed? 
Mimics don't have a way to hop onto a table or anything so they're usually hiding on the ground, being able to just stumble into them and reveal them is a bit easy.

## Changelog 

```changelog
(u)CalliopeSoups
(+)Mimics copy the density of their disguise while hiding, letting people walk over you as most items.
```

### DIFF
--- a/code/mob/living/critter/mimic.dm
+++ b/code/mob/living/critter/mimic.dm
@@ -102,6 +102,7 @@
 		if (src.is_hiding)
 			return
 		src.is_hiding = TRUE
+		src.density = 0
 		qdel(src.name_tag)
 		src.name_tag = null
 		src.UpdateIcon()
@@ -113,6 +114,7 @@
 		if(!src.is_hiding)
 			return
 		src.is_hiding = FALSE
+		src.density = 1
 		src.name_tag = new()
 		src.update_name_tag()
 		src.vis_contents += src.name_tag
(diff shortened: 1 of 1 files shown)


code/mob/living/critter/mimic.dm
```diff
@@ -74,13 +75,15 @@
...
 			src.base_form = TRUE
+			src.current_disguise = null
 			src.appearance = /mob/living/critter/mimic/
 			src.stop_hiding()
 		else
 			var/icon/I = getFlatIcon(target)
 			var/pixels = null
 			src.dir_locked = TRUE
 			src.base_form = FALSE
+			src.current_disguise = target
```
Azrun, line 86: Might be better as weakref or keeping a copy of the expected density so it can be gced.

CalliopeSoups: Couldn't seem to figure out weakrefs so I just saved the density instead of the whole disguise target object.

CalliopeSoups:
Might be a good idea to testmerge this one if it goes through, might be bugs that can't be tested on a solo instance.

Azrun (approved):

PR #23569, opened 2025-05-24
Spooky gamemode

CalliopeSoups:
[gamemodes] 
## About the PR 
Small PR that adds a spooky gamemode that features a guarenteed wraith along with possible lings, vampires, werewolves and hunters as latejoins.

I toyed with the idea of adding spooky ambience but I didn't want to give wraiths away if they want to stealth, like tricksters. Also thought about adding hunters to the roundstart pool for this gamemode but I think it feels more normal as is.

I like the idea of adding this as an actual gamemode on classic, but I'll wait for admin input for that. I'll add a changelog if so.

## Why's this needed? 
Wraith's fun and rare, this adds a spooky antag themed gamemode admins can spawn if they want to see one.

### DIFF
--- a/code/datums/gamemodes/mixed_spooky.dm
+++ b/code/datums/gamemodes/mixed_spooky.dm
@@ -1,0 +1,14 @@
+/datum/game_mode/mixed/spooky
+	name = "Spooky"
+	config_tag = "spooky"
+	latejoin_antag_compatible = 1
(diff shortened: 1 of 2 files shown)


cheekybrdy:
This feels more like a storyteller thing, but I'm pretty sure the code for that is just a stub, so this seems like a good way to put this in.

Samkha1245:
this would be cool if it was like non admin-spawned and a gamemode cus wraith is barely used at all aside from midrolls in classic or a really rare wraith spawn

TobleroneSwordfish (changes requested):

PR #23573, opened 2025-05-24
Resprites the Icemoon surface + Slight adjustments to the cliffs

TemThrush:
## About the PR 

This PR resprites the snow, abyss and cliffs used on the ice-moon surface.
Additionally, the cliff tiles now count as part of the "abyss" so walking onto them will make you fall. To make up for this, in parts of the azone where climbing along the cliff edges is required i've expanded the land out a little further or added more lattices.

icemoonsurfacecomparison
[Old sprites on the left, new sprites on the right.]
image
[New sprites under the intended lighting: left parallax disabled, right with parallax enabled]

## Why's this needed? 
Now that the Ice moon caves has been resprited I wanted to resprite the surface as well to match current spriting standards and hopefully feel more cohesive with the new sprites in the caves.

With the new sprites, walking on the cliff-edges looked way weirder and seemed unnatural, so I added the properties of the abyss tiles to the cliff tiles. I chose to do this because as far as I'm aware making this change should have no effect on how the azone runs (correct me if i'm wrong) and it wasn't visually clear if being able to walk along the cliff-edges was intentional or not.  (As an example: the two small holes in the surface that actually have no abyss tiles you can fall into, so you can just walk over them perfectly unharmed even if it looks like you should fall vs the part of the azone where you're required to use the cliff-sides to traverse)
image
[Examples of a player standing on the cliffs-edge pre and post resprite.]
*I also removed that single wall turf tile in the hole seen in the images above because its really jarring visually and I don't think it's important (once again correct me if I'm wrong).*

If this change causes any issues for the azone or is a point of contention I'm perfectly willing to remove this part of the PR and find a different solution. 🙏


TemThrush:
one moment im having technical difficulties, pray for me

TemThrush:
it worked \o/ we're so back

Donkier:
maybe i’m just a factorio head but i swear the new sprites look like aquilo

ZeWaka:
It's tempting to make the falling off the cliff edge into a 1s action bar, compared to the full abyss which would be instant

PR #23583, opened 2025-05-25
Adds new makeup items & Makeup Artist trait 💄

TemThrush:
[feature][traits][objects]

## About the PR 

Adds the makeup artist trait, a trinket trait that gives the player a small bag carrying three types of makeup.
The style of the bag and the make-up are randomized on spawn and can be three different types.

image
[All of the item sprites.]

image
[Brush sprites, which appear over the progress bar when makeup is being applied.]

image
[Trait Icon]

The bag spawns with lipstick, eye-shadow, and blush matching the style of the bag. The bag will only fit make-up items, and only has three slots. 

Lipstick has been slightly changed to only apply when the cap is taken off, the new lipstick that comes in the bag functions the same as existing lipstick. Blush and eye-shadow function largely the same as lipstick, but apply different overlays.
image
[Examples of all the different kinds of makeup overlays]
Each eye-shadow has a select number of colors it chooses from depending on what "style" it is. Each time eye-shadow is applied it will be a random color from that palette 

## Why's this needed? 
There's been no way to apply eye-shadow to your character without using a hair layer, so I wanted to add some type of makeup that could do that as well as blush. (I thought it would be cute!) Since currently the only means of obtaining lipstick are through either finding some randomly spawned in a prefab or by crafting it (through a pretty obscure method) I wasn't sure how someone would reasonably obtain these make-up items on station besides "bringing them from home" or in this case, as a trinket!  I made three different sets to add a bit of randomness to the trait and to hopefully emphasize that these are fancy personal items someone brought from home rather than something you'd usually see around the station. 

## Changelog 

```changelog
(u)YellowLemons & TemThrush
(*)Added the Makeup Artist trait, which lets you spawn with a pouch of lipstick, eye-shadow, and blush. 
```

### DIFF
--- a/code/obj/item/makeup.dm
+++ b/code/obj/item/makeup.dm
@@ -31,9 +31,7 @@
 		if (blush != ownerMob.equipped())
 			interrupt(INTERRUPT_ALWAYS)
 			return
-
-		for (var/mob/O in AIviewers(owner))
-			O.show_message("[owner] begins applying blush to [owner == target ? "[him_or_her(owner)]self" : target]!", 1)
+		owner.visible_message(SPAN_ALERT("[owner] begins applying blush to [owner == target ? "[him_or_her(owner)]self" : target]!"))
 
 	onInterrupt(var/flag)
 		..()
@@ -41,8 +39,7 @@
 			target.blush = 2
 			target.blush_color = blush.makeup_color
 			target.update_body()
-			for (var/mob/O in AIviewers(owner))
-				O.show_message(SPAN_ALERT("[owner] messes up [owner == target ? "[his_or_her(owner)]" : "[target]'s"] makeup!"), 1)
+			owner.visible_message(SPAN_ALERT("[owner] messes up [owner == target ? "[his_or_her(owner)]" : "[target]'s"] makeup!"))
 
 	onEnd()
 		..()
@@ -51,8 +48,7 @@
 			target.blush = 1
 			target.blush_color = blush.makeup_color
 			target.update_body()
-			for (var/mob/O in AIviewers(ownerMob))
-				O.show_message("[owner] applies blush to [target ]!", 1)
+			owner.visible_message(SPAN_ALERT("[owner] applies blush to [target ]!"))
 
 /datum/action/bar/icon/apply_eyeshadow
 	duration = 40
@@ -87,9 +83,7 @@
 		if (eyeshadow != ownerMob.equipped())
 			interrupt(INTERRUPT_ALWAYS)
 			return
-
-		for (var/mob/O in AIviewers(owner))
-			O.show_message("[owner] begins applying eyeshadow to [owner == target ? "[him_or_her(owner)]self" : target]!", 1)
+		owner.visible_message(SPAN_ALERT("[owner] begins applying eyeshadow to [owner == target ? "[him_or_her(owner)]self" : target]!"))
 
 	onInterrupt(var/flag)
 		..()
@@ -97,8 +91,7 @@
 			target.eyeshadow = 2
 			target.eyeshadow_color = eyeshadow.makeup_color
 			target.update_body()
-			for (var/mob/O in AIviewers(owner))
-				O.show_message(SPAN_ALERT("[owner] messes up [owner == target ? "[his_or_her(owner)]" : "[target]'s"] makeup!"), 1)
+			owner.visible_message(SPAN_ALERT("[owner] messes up [owner == target ? "[his_or_her(owner)]" : "[target]'s"] makeup!"))
 
 	onEnd()
 		..()
@@ -107,8 +100,7 @@
 			target.eyeshadow = 1
 			target.eyeshadow_color = eyeshadow.makeup_color
 			target.update_body()
-			for (var/mob/O in AIviewers(ownerMob))
-				O.show_message("[owner] applies eyeshadow to [target ]!", 1)
+			owner.visible_message(SPAN_ALERT("[owner] applies eyeshadow to [target ]!"))
 
 //Lipstick
 /obj/item/pen/crayon/lipstick/fancy
@@ -250,12 +242,7 @@
 	name = "eyeshadow"
 	desc = "A palette of pigmented powders that is intended to be used around a person's eyes."
 	makeuptype = "eyeshadow_light"
-	var/list/eyeshadow_light_colors = list("#af20a8", "#f8aaaa", "#fd9368", "#de3862", "#dd506b", "#ffd6da", "#ab1e42", "#8a3e3e",
-	"#e253de", "#bc88bb")
-	var/list/eyeshadow_dark_colors = list("#7c44b1", "#4052a5", "#c2eeff", "#64b0ce", "#65a36f", "#4c4e4e", "#141314ff", "#dbeff5",
-	"#3b2c5e", "#28527d")
-	var/list/eyeshadow_gold_colors = list("#be5e51", "#e95c2c", "#f38e26", "#ffe2a7", "#c98b0d", "#ffff", "#493e25", "#8e4f0e",
-	"#ffa200")
+	var/eyeshadow_colors = "#ffff"
 
 
 	attack(mob/target, mob/user, def_zone, is_special = FALSE, params = null)
@@ -267,12 +254,7 @@
 			if (H.eyeshadow == 2)
 				user.show_text("Gurl, [H == user ? "you" : H] a hot mess right now. That all needs to be cleaned up first.", "red")
 				return
-			if(src.makeuptype == "eyeshadow_light")
-				makeup_color = pick(eyeshadow_light_colors)
-			if(src.makeuptype == "eyeshadow_dark")
-				makeup_color = pick(eyeshadow_dark_colors)
-			if(src.makeuptype == "eyeshadow_gold")
-				makeup_color = pick(eyeshadow_gold_colors)
+			makeup_color = pick(eyeshadow_colors)
 			actions.start(new /datum/action/bar/icon/apply_eyeshadow(target, src, target == user ? 40 : 60), user)
 		else
 			return ..()
@@ -280,14 +262,20 @@
 /obj/item/makeup/eyeshadow/light
 	name = "eyeshadow"
 	makeuptype = "eyeshadow_light"
+	eyeshadow_colors = list("#af20a8", "#f8aaaa", "#fd9368", "#de3862", "#dd506b", "#ffd6da", "#ab1e42", "#8a3e3e",
+	"#e253de", "#bc88bb")
 
 /obj/item/makeup/eyeshadow/dark
 	name = "eyeshadow"
 	makeuptype = "eyeshadow_dark"
+	eyeshadow_colors = list("#7c44b1", "#4052a5", "#c2eeff", "#64b0ce", "#65a36f", "#4c4e4e", "#141314ff", "#dbeff5",
+	"#3b2c5e", "#28527d")
 
 /obj/item/makeup/eyeshadow/gold
 	name = "eyeshadow"
 	makeuptype = "eyeshadow_gold"
+	eyeshadow_colors = list("#be5e51", "#e95c2c", "#f38e26", "#ffe2a7", "#c98b0d", "#ffff", "#493e25", "#8e4f0e",
+	"#ffa200")
 
 //Makeup Bags
 /obj/item/storage/makeup_bag


code/modules/forensics/atom_forensic.dm
```diff
@@ -230,10 +230,14 @@
...
 			M.gunshot_residue = 0 // Only humans can have residue at the moment.
-			if (M.makeup || M.spiders)
+			if (M.makeup || M.spiders || M.blush||M.eyeshadow)
 				M.makeup = null
 				M.makeup_color = null
 				M.spiders = null
+				M.blush = null
+				M.blush_color = null
+				M.eyeshadow = null
+				M.eyeshadow_color = null
```
RubberRats, line 240: Not sure if this is within scope, but at this point it might make more sense to have worn makeup be stored as a list inside customizationHolder

RubberRats (commented):
First off, wanted to say this is very cool. If you wanted another space to put these, you could consider having a crate of random, genericized versions that cargo can order.

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,328 @@
...
+/obj/item/makeup/eyeshadow
+	name = "eyeshadow"
+	desc = "A palette of pigmented powders that is intended to be used around a person's eyes."
+	makeuptype = "eyeshadow_light"
+	var/list/eyeshadow_light_colors = list("#af20a8", "#f8aaaa", "#fd9368", "#de3862", "#dd506b", "#ffd6da", "#ab1e42", "#8a3e3e",
+	"#e253de", "#bc88bb")
+	var/list/eyeshadow_dark_colors = list("#7c44b1", "#4052a5", "#c2eeff", "#64b0ce", "#65a36f", "#4c4e4e", "#141314ff", "#dbeff5",
+	"#3b2c5e", "#28527d")
+	var/list/eyeshadow_gold_colors = list("#be5e51", "#e95c2c", "#f38e26", "#ffe2a7", "#c98b0d", "#ffff", "#493e25", "#8e4f0e",
+	"#ffa200")
```
RubberRats, line 256: Wouldn't it make more sense to have the colors defined in each subtype instead of using a switch statement in the parent?

RubberRats (commented):
Item sprites are good, though you may want to consider using grayscale overlays for the purpose of custom makeup colors

code/obj/item/makeup.dm
yelloworlemons: I figured this would be the best compromise for what we were trying to achieve. We can't really have the colour defined in each child without having the attack code duplicated for each one(and any new ones anyone else wants to add) since the colour is supposed to be randomized on each use and I wanted to keep it open for adding other types later down the line if needed while keeping bloat down

code/modules/forensics/atom_forensic.dm
frawhst, line 240:
```suggestion
		M.makeup = null
		M.makeup_color = null
		M.spiders = null
		M.blush = null
		M.blush_color = null
		M.eyeshadow = null
		M.eyeshadow_color = null
```
Issue that predates this PR but may as well fix while we're here (no point in checking those vars, can just blindly null everything to the same effect)

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,330 @@
...
+		if (blush != ownerMob.equipped())
+			interrupt(INTERRUPT_ALWAYS)
+			return
+
+		for (var/mob/O in AIviewers(owner))
+			O.show_message("[owner] begins applying blush to [owner == target ? "[him_or_her(owner)]self" : target]!", 1)
+
+	onInterrupt(var/flag)
+		..()
+		if (prob(owner == target ? 50 : 60))
```
frawhst, line 38: I get why you'd have a different result chance depending on if someone is using it on themselves, but why 10%? strangely small difference, just curious about the number

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,330 @@
...
+			target.blush = 2
+			target.blush_color = blush.makeup_color
+			target.update_body()
+			for (var/mob/O in AIviewers(owner))
+				O.show_message(SPAN_ALERT("[owner] messes up [owner == target ? "[his_or_her(owner)]" : "[target]'s"] makeup!"), 1)
```
frawhst:
```suggestion
			owner.visible_message(SPAN_ALERT("[owner] messes up [owner == target ? "[his_or_her(owner)]" : "[target]'s"] makeup!"), SPAN_ALERT("You mess up [owner == target ? "your own" : "[target]'s"] makeup!"))
```
was written in a bit of an old-timey fashion, we have more succinct procs now (plus i added a self message)

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,330 @@
...
+
+	onEnd()
+		..()
+		var/mob/ownerMob = owner
+		if (owner && ownerMob && target && blush == ownerMob.equipped() && BOUNDS_DIST(owner, target) == 0)
+			target.blush = 1
```
frawhst, line 48:
```suggestion
			target.blush = TRUE
```
TRUE is preferred for booleans

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,330 @@
...
+			target.blush_color = blush.makeup_color
+			target.update_body()
+			for (var/mob/O in AIviewers(ownerMob))
+				O.show_message("[owner] applies blush to [target ]!", 1)
```
frawhst:
```suggestion
			owner.visible_message(SPAN_HINT("[owner] applies blush to [owner == target ? "[himself_or_herself(owner)]" : "[target]"]."))
```
similar to above

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,330 @@
...
+
+/datum/action/bar/icon/apply_eyeshadow
+	duration = 40
```
frawhst, line 54:
```suggestion
	duration = 4 SECONDS
```

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,330 @@
+/datum/action/bar/icon/apply_blush
+	duration = 40
```
frawhst, line 2:
```suggestion
	duration = 4 SECONDS
```
seconds are preferred to deciseconds as a unit

code/obj/item/makeup.dm
TobleroneSwordfish: I don't see why you'd need to duplicate the attack code, just make it pick from `src.eyeshadow_colors` and override the var on each one.

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,330 @@
...
+		..()
+		if (BOUNDS_DIST(owner, target) > 0 || target == null || owner == null || blush == null)
+			interrupt(INTERRUPT_ALWAYS)
+			return
+		var/mob/ownerMob = owner
+		if (blush != ownerMob.equipped())
+			interrupt(INTERRUPT_ALWAYS)
+			return
+
+		for (var/mob/O in AIviewers(owner))
```
TobleroneSwordfish: All these should use `visible_message` instead

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,330 @@
...
+	attack_self(var/mob/user)
+		src.open = !src.open
+		src.IconOverride()
+		if (!src.open)
+			src.deactivate(user)
+		else
+			src.activate(user)
+		return
+
+	proc/ColorOverride() //really crusty workaround until I can refactor lipstick
```
TobleroneSwordfish, line 131: Please just refactor lipstick instead of doing this and copying the makeup procs between the two type trees.

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,330 @@
...
+		if (ishuman(target))
+			var/mob/living/carbon/human/H = target
+			if (!src.open)
+				user.show_text("You need to open it first!")
+				return
+			if (H.blush == 2)
+				user.show_text("Gurl, [H == user ? "you" : H] a hot mess right now. That all needs to be cleaned up first.", "red")
+				return
+			else
+				actions.start(new /datum/action/bar/icon/apply_blush(target, src, target == user ? 40 : 60), user)
```
TobleroneSwordfish, line 220: Blush doesn't appear to work? I can't tell if it's just really subtle but I tried all different colours on different skin colours and I couldn't see it.

TobleroneSwordfish (changes requested):
This is coool, Thrush sprites are so good ❤️

code/obj/item/makeup.dm
TemThrush: Oh god i forgot to rename the icon, should be working after the latest commit 😭

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+			owner.visible_message(SPAN_ALERT("[owner] messes up [owner == target ? "[his_or_her(owner)]" : "[target]'s"] makeup!"))
+
+	onEnd()
+		..()
+		var/mob/ownerMob = owner
+		if (owner && ownerMob && target && blush == ownerMob.equipped() && BOUNDS_DIST(owner, target) == 0)
+			target.blush = 1
+			target.blush_color = blush.makeup_color
+			target.update_body()
+			owner.visible_message(SPAN_ALERT("[owner] applies blush to [target ]!"))
```
frawhst, line 51:
```suggestion
			owner.visible_message(SPAN_ALERT("[owner] applies blush to [target]!", SPAN_ALERT("You apply blush to [target]!"))
```
slightly cleaner to say 'you' for the person doing it (second argument to `visible_message` is the message sent to the mob which the message originates from)

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
+/datum/action/bar/icon/apply_blush
+	duration = 40
+	interrupt_flags = INTERRUPT_MOVE | INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION
+	icon = 'icons/obj/items/makeup.dmi'
+	icon_state = "blush_brush"
+	var/mob/living/carbon/human/target
+	var/obj/item/makeup/blush
+
+	New(ntarg, nblush, ndur)
```
frawhst, line 9: why are these named like this? what does the 'n' stand for? I think I get what they are but just 'target, blush, duration [dur also fine]' would be more readable imo
(ps. if you have an argument to a proc and a variable on an object with the same name, use `src.` to disambiguate, as in `src.target = target //sets var named "target" on this object to the argument named "target" of the current proc`. having `target` and `blush` be both arg names and var names is totally fine.)

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+
+/datum/action/bar/icon/apply_eyeshadow
+	duration = 40
+	interrupt_flags = INTERRUPT_MOVE | INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION
+	icon = 'icons/obj/items/makeup.dmi'
+	icon_state = "eyeshadow_brush"
+	var/mob/living/carbon/human/target
+	var/obj/item/makeup/eyeshadow
+
+	New(ntarg, neyeshadow, ndur)
```
frawhst, line 61: same as the other note about args

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+		if (prob(owner == target ? 50 : 60))
+			target.blush = 2
+			target.blush_color = blush.makeup_color
+			target.update_body()
+			owner.visible_message(SPAN_ALERT("[owner] messes up [owner == target ? "[his_or_her(owner)]" : "[target]'s"] makeup!"))
+
+	onEnd()
+		..()
+		var/mob/ownerMob = owner
+		if (owner && ownerMob && target && blush == ownerMob.equipped() && BOUNDS_DIST(owner, target) == 0)
```
frawhst, line 47:
```suggestion
		var/mob/ownerMob = owner
		if (ownerMob && target && blush == ownerMob.equipped() && BOUNDS_DIST(owner, target) == 0)
```
redundant check since you're setting `ownerMob` to `owner` right above

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+		if (prob(owner == target ? 50 : 60))
+			target.eyeshadow = 2
+			target.eyeshadow_color = eyeshadow.makeup_color
+			target.update_body()
+			owner.visible_message(SPAN_ALERT("[owner] messes up [owner == target ? "[his_or_her(owner)]" : "[target]'s"] makeup!"))
+
+	onEnd()
+		..()
+		var/mob/ownerMob = owner
+		if (owner && ownerMob && target && eyeshadow == ownerMob.equipped() && BOUNDS_DIST(owner, target) == 0)
```
frawhst, line 99:
```suggestion
		var/mob/ownerMob = owner
		if (ownerMob && target && eyeshadow == ownerMob.equipped() && BOUNDS_DIST(owner, target) == 0)
```
same note as before

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+			target.eyeshadow = 1
```
frawhst, line 100:
```suggestion
			target.eyeshadow = TRUE
```

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+			target.eyeshadow_color = eyeshadow.makeup_color
+			target.update_body()
+			owner.visible_message(SPAN_ALERT("[owner] applies eyeshadow to [target ]!"))
```
frawhst, line 103:
```suggestion
			owner.visible_message(SPAN_ALERT("[owner] applies eyeshadow to [target]!"))
```

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+		return
+
+	attack_self(var/mob/user)
+		src.open = !src.open
+		src.IconOverride()
+		if (!src.open)
+			src.deactivate(user)
+		else
+			src.activate(user)
+		return
```
frawhst, line 129:
```suggestion
```
procs `return` by default at the end, so this doesn't do anything

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+
+	proc/ColorOverride() //really crusty workaround until I can refactor lipstick
+		src.font_color = "[src.lipstickcolor]"
+		src.color_name = hex2color_name(src.font_color)
+		src.name = "lipstick"
+		return
```
frawhst, line 135:
```suggestion
```

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+
+	proc/IconOverride() //also really crusty workaround
+		src.icon_state = "[src.lipsticktype][src.open]"
+		if (src.open)
+			ENSURE_IMAGE(src.image_stick, src.icon, "[src.lipsticktype]")
+			return
```
frawhst, line 141:
```suggestion
```

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+	click_delay = 0.7 SECONDS
+	var/lipsticktype = "dark_lipstick"
+	var/lipstickcolor = "#9e0925" //temporary var to help get around parent lipstick always being random until I refactor it
+
+
+	New()
+		..()
+		src.ColorOverride()
+		src.IconOverride()
+		return
```
frawhst, line 120:
```suggestion
```

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+		return ..()
+
+	attack_self(var/mob/user)
+		src.open = !src.open
+		src.UpdateIcon()
+		if (!src.open)
+			src.deactivate(user)
+		else
+			src.activate(user)
+		return
```
frawhst, line 188:
```suggestion
```

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+
+	update_icon()
+		src.icon_state = "[src.makeuptype][src.open]"
+		return
```
frawhst, line 192:
```suggestion
```

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+
+	proc/activate(mob/user as mob)
+		src.open = 1
```
frawhst, line 195:
```suggestion
		src.open = TRUE
```

code/obj/item/makeup.dm
```diff
@@ -0,0 +1,318 @@
...
+		flick("[makeuptype]_open", src)
+		playsound(user, 'sound/items/penclick.ogg', 30, TRUE)
+
+	proc/deactivate(mob/user as mob)
+		src.open = 0
```
frawhst, line 200:
```suggestion
		src.open = FALSE
```

TemThrush:
Going to take a bit to get everything refactored and ready for review again so I'm closing until it's ready. ^_^

PR #23591, opened 2025-05-26
You must use an aggressive grab to move with someone already in an aggressive grab

pgmzeta:
[balance][player actions]
## About the PR 
When moving with a grab or pull, check if the target is in someone else's aggressive grab. If so, only move them if your grab is stronger than a passive grab.

## Why's this needed? 
Very easy to save people from someone being eaten by a abom changeling with pulls/passive grabs, when eating them requires an aggressive grab. 

This prevents silicons from pulling people out of aggressive grabs entirely.

## Changelog 

```changelog
(u)glowbold
(*)You must use an aggressive grab to pull people away from aggressive grabs.
```

### DIFF
--- a/code/mob/input.dm
+++ b/code/mob/input.dm
@@ -311,9 +311,24 @@
 								src.remove_pulling()
 								//hud.update_pulling() // FIXME
 							else
-								pulling += src.pulling
+								var/can_pull = TRUE
+								if (ismob(src.pulling))
+									var/mob/M = src.pulling
+									for(var/obj/item/grab/grab_grabbed_by in M.grabbed_by)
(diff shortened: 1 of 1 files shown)


aaaaaaaaaaaaaaaaaah:
I'm a big fan of this, I hope it encourages more loud changelings since right now they are just at a disatvantage when going into abomination (I've seen two aboms lose to around 6 people just because they could never actually eat anyone).

Cinder-Arcanium:
Huge fan of this, shambling abomination is so weak for how much it costs. This will hopefully turn it into something actually worth using

ZeWaka (approved):

PR #23620, opened 2025-05-28
Table Construction QoL Improvements

nova2053:
[feature][actions][qol]
## About the PR 
Adds a "NO_BLOCK_TABLE" flag to a couple of object types which usually spawn on tables (such as personal computers and wrestling bells). Parts-based furniture construction will ignore blockages from objects with directional blocking and, if the furniture item in question is a table, objects with the new NO_BLOCK_TABLE flag.

The following items have been given the NO_BLOCK_TABLE flag and can thus now have tables constructed under them:
KitchenHelper food mixers
Loudspeakers
Disposal outlets
Personal computers (and their frames)
Espresso machines
Coffee machines
Tourbot consoles
Wrestling bells
Printers
Scanners

## Why's this needed? 
Currently, should a table or rack on the same tile as a thindow be deconstructed, it will be unnecessarily difficult to rebuild said table/rack without also removing the thindow. This PR also solves a similar case with items which spawn on tables - most notably personal computers - preventing broken tables from being rebuilt under them.

## Changelog 

```changelog
(u)nova2053
(+)Tables, racks, and other items which use a similar construction method can now be constructed on the same tile as thindows, sliding glass doors, and other objects with directional blocking.
(+)Tables can now be constructed under a couple more objects, such as microwaves, espresso machines, and more. See PR page for full list.
```

### DIFF
--- a/_std/defines/obj.dm
+++ b/_std/defines/obj.dm
@@ -11,6 +11,8 @@
 #define HAS_DIRECTIONAL_BLOCKING (1<<3)
 /// prevents ghost critter interaction. On obj so it can cover machinery, items etc...
 #define NO_GHOSTCRITTER (1<<4)
+/// allows for tables to be built under items they normally wouldn't be able to be built under, such as microwaves and personal computers
+#define NO_BLOCK_TABLE (1<<5)
 
 /// At which alpha do opague objects become see-through?
 #define MATERIAL_ALPHA_OPACITY 190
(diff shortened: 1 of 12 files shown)


nova2053:
> i would suggest adding the tag to thindows. Right now you arent able to place a table on a tile where a thin window exist.

Tables can be placed on the tile of thindows. What I meant by "Parts-based furniture construction will ignore blockages from objects with directional blocking and, if the furniture item in question is a table, objects with the new NO_BLOCK_TABLE flag" is that, while tables will check for both whether the blocking object has the tag and whether the blocking object has directional blocking, non-tables will only check for if the blocking object has directional blocking (because putting a desktop computer, toaster, etc on top of a rack doesn't make very much sense).

TobleroneSwordfish (approved):

pgmzeta:
can this be added to the little bar tv monitors too

PR #23623, opened 2025-05-29
Makes Neon Broken objectives mandatory

TobleroneSwordfish:
## About the PR 
Adjusts the Broken antag popup to state that the objective is mandatory.

## Why's this needed? 
Team antagonists like nukies and revs are required to work together to accomplish their objectives.
Broken aren't exactly a team antagonist because they often fight each other, but they do have similar issues if some people roll the antagonist but choose not to do anything with it. For example if 5 people roll broken with the objective of trashing genetics but only one person turns up, they're likely just going to hit a window a few times and get arrested, not very fun.

You shouldn't be able to choose not to surrender to the eldritch madness! (apart from in the meta sense, where you can absolutely choose to just turn the antag off in your preferences if you don't want to play it)

My only concern with this is that the objectives might get stale over time, but so far I don't think this has been the case too much as they are broad enough to allow for quite a few different ways of approaching them.
## Changelog 

```changelog
(u)LeahTheTech
(*)Neon Broken antagonist objectives are now mandatory in the same way mindhack instructions are. Reminder: you can disable rolling Broken under the "Other foes" antag preferences option.
```


JORJ949 (changes requested):
I would split Broken out of other foes then, I _think_ werewolf uses other foes. But that can be done seperately

PR #23630, opened 2025-05-29
Tweaks cogmap2's engine wiring

CalliopeSoups:
[mapping] [qol]
## About the PR 
Basic PR that shuffles some wires around so that they don't cover heat pipes fully. Not sure if this needs a changelog, let me know if not.

## Why's this needed? 
Cog2's engine sets on fire a lot. I asked around about it and it turns out pipe breaches can be difficult to spot because a lot of wires layer over the pipes. 

## Changelog 

```changelog
(u)CalliopeSoups
(+)Cogmap 2's engine has less heat pipe covering wires.
```

### DIFF
--- a/maps/cogmap2.dmm
+++ b/maps/cogmap2.dmm
@@ -20990,9 +20990,6 @@
 /obj/stool/chair{
 	dir = 4
 	},
-/obj/cable/orange{
-	icon_state = "1-4"
-	},
 /turf/simulated/floor/black,
 /area/station/hallway/primary/west)
 "bne" = (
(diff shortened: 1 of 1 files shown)


AnonymousUserTheOnly:
this was BEYOND loved by everyone, i feel.  that one pipe with the wire running clean over it near real coordinates 130/154?  damn near impossible to accurately tell the existence of reinforcement.  and it blends with the red "hot loop" default, too.  only reason you'd know it was there is if you pay attention to the pipe corners and realize there isn't one OR you discover a burst at that spot.  and bursts are usually after the fact, too.  so thank you to whoever put this in the pipeline, it was well warranted.

TobleroneSwordfish (approved):

PR #23639, opened 2025-05-30
Fixes spooky mode

CalliopeSoups:
[bug fix] [gamemodes]
## About the PR 
I forgot to put a return true in there. Oops.

## Why's this needed? 
It broke


CalliopeSoups:
I changed the minimum players to 1 to test this and didn't spawn as a wraith despite being the only player with wraith turned on. I'm hoping that's a quirk of being the only player online.

PR #23646, opened 2025-05-31
Manifold rupturing

TobleroneSwordfish:
## About the PR 
Makes manifolds able to rupture and be reinforced like regular pipes.

## Why's this needed? 
Generating Kardashev type III amounts of power from a TEG loop made entirely out of manifolds is funny but does unfortunately need fixing.

Why did this take 200 lines of refactoring to achieve? Yea.

## Changelog 

```changelog
(u)LeahTheTech
(*)Manifolds can now rupture and be reinforced just like regular pipes.
```


F-e-r-n:
as a teg nerd, i do like this but i do think there needs to be a better lategame choice than all of your pipes bursting no matter what you do, or spacing the entire TEG room so pipe bursts dont melt the entire room. Maybe the ability to fix pipes before they actually burst and spew hot plasma all over the entire room? As of right now though there's a very faint line between making enough power for the PTL and griefing the engine room.

TobleroneSwordfish:
I agree, I would like to expand on pipe reinforcement/repair in future 👍

MegaBibarel420:
this is probably not a very good idea at this point in time

PR #23667, opened 2025-05-31
Synth Organ Digestion

LorrMaster:
[hydroponics][medical][feature][add to wiki]

## About the PR 

- Note: I don't know too much about reagent values, so let me know if they make sense or feel free to just alter them yourself however you like because I do not know how to balance numbers.

Synth Stomach
- Allows you to safely consume insect reagents (ants & spiders).

Synth Intestines
- Added new effects for plant reagents while having synth intestines.
- Compost will fill your hunger meter
- Ammonia, phosphorus, & saltpetre provide stamina bonuses
- Ash, potash, & potassium provide healing
- Space fungus provide 10% stun resistance
- Weedkiller plant damage is applied

## Why's this needed? 
- Makes these synth organs more thematic and interesting

## Changelog
```changelog
(u)LorrMaster
(*)Synth stomach & intestines can cause different effects with plant-themed reagents
```

### DIFF
--- a/code/obj/item/mob_parts/human_parts.dm
+++ b/code/obj/item/mob_parts/human_parts.dm
@@ -933,6 +933,7 @@
 			set_loc(holder)
 		..()
 
+
 /obj/item/parts/human_parts/arm/left/synth/bloom
 	desc = "A left arm. Looks like a rope composed of vines. There's some little flowers on it."
 	icon_state = "arm_left_plant_bloom"


code/modules/chemistry/Reagents-Base.dm
```diff
@@ -482,6 +497,15 @@
 		growth_tick.growth_rate += 0.4
 		growth_tick.health_change += 0.4
 
+	on_mob_life(var/mob/M, var/mult = 1)
+		if(!M) M = holder.my_atom
+		if(prob(55) && isliving(M))
+			var/mob/living/L = M
+			if (L.organHolder?.intestines?.synthetic)
+				M.HealDamage("All", 0.5 * mult, 0.5 * mult)
```
mordent-goonstation, line 505: Potassium is silly-easy to get, just one-click or eat a banana. Will need to look at balance here *a lot*.

code/modules/chemistry/Reagents-FoodDrink.dm
```diff
@@ -1406,6 +1406,21 @@ datum
...
 			taste = list("strong", "fruity")
 
+			reaction_mob(var/mob/M, var/method=TOUCH, var/volume)
+				. = ..()
+				if (M.mind.assigned_role == "Botanist" || src.drinker_is_bee(M))
+					M.emote("twitch_v")
+
+			on_mob_life(var/mob/M, var/mult = 1)
+				if (!M) M = holder.my_atom
+				if (prob(60))
```
mordent-goonstation, line 1416: That's a lot of twitching.

code/modules/chemistry/Reagents-Misc.dm
```diff
@@ -1069,11 +1097,27 @@ datum
...
+			threshold = THRESHOLD_INIT
 
 			on_plant_life(var/obj/machinery/plantpot/P, var/datum/plantgrowth_tick/growth_tick)
 				growth_tick.endurance_bonus += 0.5
 
+			cross_threshold_over()
+				if(isliving(holder?.my_atom))
+					var/mob/living/L = holder.my_atom
+					if (L.organHolder?.intestines?.synthetic)
+						APPLY_ATOM_PROPERTY(L, PROP_MOB_STUN_RESIST, "synth_digest_fungus", 10)
```
mordent-goonstation, line 1109: Does this dispose correctly if the mob ceases to be before the amount drops?

mordent-goonstation (commented):
Broadly feels like this might be the wrong place to be putting these, as the majority of people *aren't* going to have synthorgans, and might be something that's more akin to an equivalent tick that checks for a subset of reagents to do extra effects with on the organ itself?

I'm not super opposed to having them in with the rest of the reagent code, it does make sense to be there for completeness, but this would not be extensible/modular, e.g. adding the same kind of thing to people with cyberorgans.

PR #23668, opened 2025-05-31
Portable pumps will now shut off when ejecting a tank.

klookiyam:
## About the PR 
Changes the behavior of portable atmos pumps (the blue things) to automatically turn off when they eject tanks.

## Why's this needed? 
This is a QOL fix change that makes the behavior of portable pumps match the behavior of the larger gas canisters, which automatically close when you eject the tank. Its weird that their behaviors don't match and this often results in unintended plasma floods.


pgmzeta (approved):
air scrubbers might also benefit from this

PR #23680, opened 2025-06-01
Balance AEX shell pricing

beretkid:
## About the PR  Doubles the base price of AEX rounds for balancing
[Balance]

## Changelog 

```changelog
(u)Bobbis Dobbis
(+)AEX shell price now 2.5x the original for balancing.
```


cheekybrdy:
Not commenting on the balance part, but you should probably make the changelog and the about in the PR accurate to the 2.5x in the code.

pgmzeta:
last time this came up in discussion the consensus was they were in an okay spot, and nothing has changed about shotguns or AEX since then, as far as I know.

Here's the latest thread I could find discussing this, from a few months ago:
https://forum.ss13.co/showthread.php?tid=23687

TobleroneSwordfish:
Generally negative feedback, AEX pricing seems broadly fine. If you were going to make any change to it, it might be giving CARL a limited stock but 🤷

PR #23681, opened 2025-06-01
Prevent underwear from poking through swimsuits/dresses

RubberRats:
## About the PR 

This PR adds a `hide_underwear` variable to undersuit items, which when set to true prevents the wearer's underwear from appearing on their sprite under their clothes. It is set by default on swimsuits and some dresses.

## Why's this needed? 

Some underwear styles have a habit of poking out from under certain outfits in ways that don't feel intended.

## Changelog 

```changelog
(u)RubberRats
(+)Underwear should no longer poke out from swimsuits and certain dresses
```


DisturbHerb:
If implemented, something like this should definitely be toggle-able.

code/modules/antagonists/macho_man/abilities/_macho_man_mob.dm
```diff
@@ -416,6 +416,10 @@ var/list/snd_macho_idle = list('sound/voice/macho/macho_alert16.ogg', 'sound/voi
 	icon_state = "machopants"
 	item_state = "machopants"
 
+	New()
+		. = ..()
+		AddComponent(/datum/component/toggle_underwear)
+
```
TobleroneSwordfish:
```suggestion
```
This only covers the bottom half and so shouldn't hide underwear. Also wrestlers totally wear pants under their outfits.

TobleroneSwordfish (changes requested):
Sorry to bikeshed this, but I really don't think it needs a toggle. There's not a lot of reason you'd ever want to show shorts under a bikini or a bra poking out around a dress and the button is just kind of weird looking UI clutter.

PR #23682, opened 2025-06-01
Pipes weaken and can be repaired before rupturing

TobleroneSwordfish:
[FEATURE] [INPUT WANTED]
## About the PR 
Whenever a pipe would currently rupture it will instead "weaken", visibly bulging outward. It can then be repaired instantly with a wrench, or will rupture the next time it would have ruptured.
If the pressure differential is more than 100x the effective fatigue pressure of the pipe it will still rupture instantly.

Idea originally suggested here: https://github.com/goonstation/goonstation/pull/23646#issuecomment-2925718971

## Why's this needed? 
Gives you some warning before pipes just blow up in your face, allowing you to do a quick repair and either vent some gas or inject more cold gas into the loop.

TODO:
- [ ] Fix the animation
- [ ] Better displacement map (use Amy's tool?)
- [ ] Change the flavour to literally be just hitting the pipe with a wrench because that's funny

## Changelog 

```changelog
(u)LeahTheTech
(*)Pipes now bulge and weaken before rupturing and can be repaired by hitting them with a wrench.
```

### DIFF
--- a/code/modules/atmospherics/machinery/pipe.dm
+++ b/code/modules/atmospherics/machinery/pipe.dm
@@ -14,6 +14,8 @@
 	var/fatigue_pressure = 150*ONE_ATMOSPHERE
 	/// Can this pipe rupture?
 	var/can_rupture = FALSE // Currently only used for simple pipes and manifolds
+	///Before rupturing, weaken and complain
+	var/weakened = 0
 	/// How broken is our pipe.
 	var/ruptured = 0
 	/// What do I change back to when repaired???
@@ -71,6 +73,8 @@
 /// Repairs the pipe back to orginal state.
 /obj/machinery/atmospherics/pipe/proc/repair_pipe()
 	src.ruptured = 0
+	src.weakened = 0
+	animate(src)
 	desc = initial(desc)
 	UpdateIcon()
 	ON_COOLDOWN(src, "rupture_protection", 20 SECONDS + rand(10 SECONDS, 220 SECONDS))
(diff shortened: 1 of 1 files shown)


F-e-r-n:
Hell yeah

TrickyWolfer:
I like the visual indicators especially for newer players that's new about atmos pipes. However as a veteran TEG player, it's only a good _temporary_ fix, but definitely is still a far cry from what was removed. Main reason why unbreakables were desired was because the current alternative we have was too low for what the usual temp TEG engineers used to work with.

If there's a way to amp up the limit before the pipes start to roll for bursting, it would be great. 1 million kpa definitely would not cut it, even with 100x plus limit on the new rework (which at best 100 million / 1e10; At this point the TEG engine start to falter and back to the welding grind to try to get 1 million from PTL at least.) Lucky if the engineers have welding tool, but manifolds were also ubiquitous in toxins, which they would have less access to engineering tools.

TrickyWolfer:
Addendum after days of thinking: I cannot comment much about the manifold pipe bursting anim (since it'll be fixed and hopefully made more visible), But I have a teeny tiny request for ruptures in general:

Can it have flavour text? For both pipes and manifold. It was not a problem back then because the ones bursting were highly-visible pipes. However, since manifolds also burst (and with slightly poorer visual cue), I think it would be beneficial to have a notice i.e. "The pipe ruptures!" and not only the bulging flavour text. A problem in high-pipe count TEG rooms is that we can't tell in a glance which pipe went boom, especially when it's constantly covered in plasmafire and or pressure readers, so with this, hopefully we know which pipe (manifold or any pipe) went awol. Thanks in advance

PR #23683, opened 2025-06-01
Prevents auto-promotion of HoS on single digit populations

cheekybrdy:
## About the PR 
Title

## Why's this needed? 
Means that HoS promotion on lowpop won't happen, meaning that the armoury is less in play. Also can be better to be a secoff and have things be a bit more chill on those quieter rounds.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Automatic Command promotions to HoS won't happen in single digit staffing stations.
```


JORJ949:
I think promotion is actually more important at lowpop as you're already operating at a low crew amount you ideally want to fill the more important job slots. If you don't want to play a command role then you should set it to unwanted.

pgmzeta:
maaaaybe just for HoS?

TobleroneSwordfish (approved):

JORJ949 (approved):

PR #23684, opened 2025-06-01
Sandbags (port from 35 Below)

pgmzeta:
[feature][game objects]
## About the PR 
Adds deployable sandbags. They work exactly like nukie barricades but are made of sand. They'll auto-connect with other deployed sandbags similarly to how walls and tables auto-connect sprites.

Sandbag code and sprites by DisturbHerb.

Adds a help message to barricades to explain how to deploy them.

## Why's this needed? 
Neat thing for admins/events

## Screenshots

In devmap
Screenshot 2025-05-31 235716

From 35 Below
image


code/obj/sandbags.dm
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (C) 2025 DisturbHerb
+ *
+ * Originally contributed to the 35 Below Project
+ * Made available under the terms of the CC BY-NC-SA 3.0
+ * Full terms available in the "LICENSE" file or at:
+ * http://creativecommons.org/licenses/by-nc-sa/3.0/us/
```
ZeWaka: This isn't necessary, this is the entire repo license.

code/obj/sandbags.dm
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+TYPEINFO(/obj/barricade/sandbags)
+	var/list/connects_to_turf = list()
```
ZeWaka, line 2: It's almost tempting to make them connect to walls

pgmzeta: image
lmk if this is good i'll add it

TobleroneSwordfish: It does maybe make them look more like stone walls than sandbags? Plus they're kind of in the wrong perspective.

pgmzeta: yeah, we had this issue with a few similar things we added in 35b; wooden fences, for example, make tons of sense to connect to walls, but wall slanting makes it a bit awkward. maybe could be done with some custom connector sprites, but that's out of scope here.

ZeWaka: yeah lets just keep then as is

PR #23694, opened 2025-06-02
Washing machines don't trap you forever after being thrown in

### LINKED ISSUE #23666: Washing machine behavor
### Labels

[BUG][MINOR]

### Description

The washing machine treats players differently depending on if they were thrown in, or by being shoved in by another player

### Steps to reproduce

For shoving into washing machine:
Grab another crewmember (upgraded or not)
click on the washing machine with the grab (washing machine door open)
For throwing into washing machine:
Grab another crewmember (upgraded)
Throw the crewmember into the washing machine (washing machine door open)
Turn on washing machine

### Expected Behavior

I expected the washing machine to behave the same when either method was used, the behavors can be found in additional information

If the crewmember was shoved in, the washing machine will turn on automatically, the crewmember will bounce around, screaming and taking brute damage, before being let out automatically at the end of the cycle, cleaned, dizzy, and freshly laundered.
If the crewmember was thrown in, they will be stuck in the washing machine like any other item, turning on the laundry machine will not cause any bouncing, damage, or cleaning to the crewmember, and they will be stuck inside the washing machine until someone rescues them.

Reported by: travelsock
Client version: 515.1637
Active test merges: #23588, #23481, #23422, #21410, and #23292
Round log date: 2025-05-31-05-32
Reported on: 2025-05-31 06:55:34
Round ID: 206318
Map: KONDARU

CalliopeSoups:
[bug fix] [game objects] 
## About the PR 
Fixes #23666

This DOES pretty much make throwing people inside objectively better than stuffing them in. I could make being thrown in do less damage, possibly.

Edit: Thrown people can escape by moving around after a second and a bit. 

## Why's this needed? 
I left an antag in a washing machine not knowing it was like this and came back to them SSD. I'm going to hell

### DIFF
--- a/code/WorkInProgress/laundry.dm
+++ b/code/WorkInProgress/laundry.dm
@@ -259,7 +259,13 @@ TYPEINFO(/obj/submachine/laundry_machine)
 			logTheThing(LOG_COMBAT, M, "is thrown into a [src.name] at [log_loc(src)].")
 			M.set_loc(src)
 			src.open = 0
-			UpdateIcon()
+			src.on = 1
+			src.occupant = M
+			src.cycle = PRE
+			src.cycle_max = CYCLE_TIME_MOB_INSIDE
+			src.UpdateIcon()
+			if (!processing_items.Find(src))
+				processing_items.Add(src)
 	else
 		return ..()
 
--- a/code/WorkInProgress/laundry.dm
+++ b/code/WorkInProgress/laundry.dm
@@ -258,17 +258,27 @@ TYPEINFO(/obj/submachine/laundry_machine)
 			M.visible_message(SPAN_ALERT("<B>[M] gets tossed into the washing machine!</B>"))
 			logTheThing(LOG_COMBAT, M, "is thrown into a [src.name] at [log_loc(src)].")
 			M.set_loc(src)
-			src.open = 0
-			src.on = 1
+			M.changeStatus("knockdown", 1.5 SECONDS)
 			src.occupant = M
+			src.open = 0
 			src.cycle = PRE
-			src.cycle_max = CYCLE_TIME_MOB_INSIDE
-			src.UpdateIcon()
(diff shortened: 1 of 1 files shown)


pgmzeta:
maybe apply a knockdown when thrown in and allow them to move out instead of auto-starting?

CalliopeSoups:
> maybe apply a knockdown when thrown in and allow them to move out instead of auto-starting?

Yeah this seems the best route.

code/WorkInProgress/laundry.dm
```diff
@@ -258,11 +258,27 @@ TYPEINFO(/obj/submachine/laundry_machine)
...
 			src.open = 0
+			src.cycle = PRE
+			cycle_max = CYCLE_TIME_MOB_INSIDE
+			if (!processing_items.Find(src))
+				processing_items.Add(src)
 			UpdateIcon()
 	else
 		return ..()
 
+/obj/submachine/laundry_machine/relaymove(mob/user as mob)
```
TobleroneSwordfish, line 272: Won't this let you escape while the door is shut and while running?

TobleroneSwordfish (changes requested):

code/WorkInProgress/laundry.dm
CalliopeSoups: Yup, added a !src.on to the if check.

CalliopeSoups:
I think the washing machine getting bricked is because there's an occupant check when you turn it on, apparently to stop people from getting stuck. I tried a weird work around and stopped before I invested much into it, I'll probably see if being able to move to break out bypasses people getting stuck cause if so I can just remove the occupant check.

CalliopeSoups:
Removing the occupant check not allowing you to turn it on seems to be fine now that you can free yourself from being inside. It makes sure the cycle is set to mobs inside now instead. Tested multiple times, switching between clothes and humans etc, seems to be working.

PR #23698, opened 2025-06-02
Loose singularity rebalance

pgmzeta:
[balance][game objects]
## About the PR 
* Anti-singularity buster changes
  * Hitting a singularity with 3 or less radius will create a white hole (as before)
  * Hitting a larger singularity than that will cause the singularity to shrink by two radius, with an accompying visual effect
* Singularity movment changes
  * Singularities now move slightly more smoothly (e.g. from 4/5.1/8/9.1 to 4/6/8/10)
  * Singularities can move diagonally
  * Singularities will switch between stepping in random directions and taking steps towards a random station turf

## Why's this needed? 
make loose singularities a little more dangerous to the station and less binary

## Changelog 

```changelog
(u)glowbold
(*)Anti-singularity buster shots will shrink large singularities. Small singularities will explode into white holes.
(+)Loose singularities are more likely to drift towards station tiles.
```

### DIFF
--- a/code/obj/machinery/singularity.dm
+++ b/code/obj/machinery/singularity.dm
@@ -201,7 +201,7 @@ for some reason I brought it back and tried to clean it up a bit and I regret ev
 	STOP_TRACKING_CAT(TR_CAT_GHOST_OBSERVABLES)
 	. = ..()
 
-/obj/machinery/the_singularity/process(mult)
+/obj/machinery/the_singularity/process()
 	var/turf/T = get_turf(src)
 	if(isrestrictedz(T?.z) && !src.restricted_z_allowed)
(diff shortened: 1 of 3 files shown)


code/obj/machinery/singularity.dm
```diff
@@ -155,6 +155,10 @@ ADMIN_INTERACT_PROCS(/obj/machinery/the_singularitygen, proc/activate)
 	var/gib_mobs = 0 //! if it should call gib on mobs
 	var/list/obj/succ_cache
 
+	/// Targeted turf when loose
+	var/turf/target_turf
+	///
```
TobleroneSwordfish: ///

PR #23699, opened 2025-06-02
Adds a description to senior aged ranch animals

cheekybrdy:
## About the PR 
Title.

## Why's this needed? 
Seniors are a rather implied feature, where chickens can't lay eggs when they get old. (don't see any sheep functionality). I'll add them to the wiki whether this gets through or not, but at least being able to tell more explicitly for those who don't know what they are seeing its a spriteless change means that a newer player counting on it can realise that its not going to lay any more eggs.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Senior ranch animals can now be determined through their description.
```


JORJ949 (approved):
Code looks fine to me

PR #23705, opened 2025-06-02
Pipes encased in walls have increased fatigue pressure

M-Earthfire:
[Feature][atmospherics]

## About the PR 

This PR makes it so that pipes in walls benefit from massively increased fatigue pressure. A pipe in a normal wall will have its fatigue pressure increased by 5 times wall-density, while it is double that for reinforced walls

## Why's this needed? 

In the discussion about the TEG (https://forum.ss13.co/showthread.php?tid=24227), which has sparked due to the fix of the abuse of manifolds, the idea of doing additional reinforcement of pipes was mentioned and appreciated from multiple sides.

This PR tries two things: giving an additional way to reinforce pipes while also removing some need to add unbreakable pipes via mapping for normal burns.

Many pipes in walls have been made indestructable due to being encased in walls. By giving them a higher fatigue pressure at baseline, we guarantee that these pipes will only break at the point where all exposed pipes have already broken. This enables mappers to design future TEG's with less magical immune pipes.

Also, this gives engineers a way to sacrifice space in their engine for resiliance. By encasing pipes in walls, they will be able to further strenghten the walls. This also makes the welding-pattern easier: By only being able to wall off certain pipe segments, some pipes will try to burst earlier than others, putting the focus on less pipes. Of course, you are at fault if you use that to just pump the numbers higher and i will have no pity on your pipes blowing up.

This change e.g. enables bohrum plated, in reinforced bohrum walls set pipes with ghc to have a fatigue pressure of up to 2,16e12 kpa.

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Atmospheric pipes in walls have their fatigue pressure increased by 5 times the walls material density. Double the value for reinforced walls.
```

### DIFF
--- a/code/modules/atmospherics/machinery/pipe.dm
+++ b/code/modules/atmospherics/machinery/pipe.dm
@@ -26,7 +26,15 @@
 	initial_icon_state = icon_state
 
 /obj/machinery/atmospherics/pipe/proc/effective_fatigue_pressure()
-	return src.fatigue_pressure * ((src.material?.getProperty("density") ** 2) || 1)
+	var/output = src.fatigue_pressure * ((src.material?.getProperty("density") ** 2) || 1)
+	var/turf/pipe_location = get_turf(src)
+	//pipes encased by walls get a bonus in fatigue pressure equal to the density of the material times 5, times 10 for reinforced walls. So pipes that lay bare should be the main weak point of the engine.
+	if(istype(pipe_location, /turf/simulated/wall))
+		var/turf/simulated/wall/wall_location = pipe_location
+		output *= ((wall_location.material?.getProperty("density") * 5) || 1)
+		if(istype(pipe_location, /turf/simulated/wall/auto/reinforced))
+			output *= 2
+	return output
 
 /// Returns list of coordinates to start and stop welding animation.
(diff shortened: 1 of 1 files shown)


Scaltra-Volpe:
I thought pipes encased in a wall already didn't break? Or was it that it did break but the gas had nowhere to go so it didn't reall5t impact much

M-Earthfire:
I havent found anything in the pipe breaking code that refers to pipes in walls. At a first glance it looked like the pipes in walls were made non-breakable through mapping.

TrickyWolfer:
I actually like this update. This is skill based again and no longer brain-deading resource waiting for science and or mining, as well as reducing the headache of "whoops, everything is leaking!". To answer both of your question, yes, pipes do still burst under walls, even map pipes. You can easily do this by chamberburning in cog1 and see the pipes in walls behaviour. Released gas is still subject under floor atmos iirc.

MegaBibarel420:
engineering then: the entire room is space
engineering now: the entire room is walls
on a more serious note: yes. engineering needs *more* options, not fewer. this is a step in the right direction.

TobleroneSwordfish (approved):
Cautious yes, I don't know if engine rooms encased in pipes are going to get old fast but it sounds like it could be a fun design challenge at least.

PR #23711, opened 2025-06-03
Adds HoS access to customs lacking security access

cheekybrdy:
## About the PR 
Gives HoS' access to customs without security access.

## Why's this needed? 
Sometimes the only ID computer is in customs on a map, which can make for fustrating HoP-less shifts when the HoS has ID access anyway. I'm assuming not giving all of sec access when customs isn't integrated with sec is a map by map design choice so I'm doing HoS access instead of letting sec flood in there every time AI calls a AA rat out.

## Changelog 

```changelog
(u)Cheekybrdy
(+)All Customs are accessible by the HoS now.
```

### DIFF
--- a/maps/donut2.dmm
+++ b/maps/donut2.dmm
@@ -17922,6 +17922,7 @@
 /obj/cable{
 	icon_state = "4-8"
 	},
+/obj/mapping_helper/access/security,
 /turf/simulated/floor/wood{
 	icon_state = "wooden"
 	},
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -21586,6 +21586,7 @@
 	},
 /obj/mapping_helper/access/head_of_personnel,
 /obj/mapping_helper/firedoor_spawn,
+/obj/mapping_helper/access/hos,
 /turf/simulated/floor/black,
 /area/station/bridge/customs)
 "jzL" = (
--- a/maps/donut3.dmm
+++ b/maps/donut3.dmm
@@ -1327,6 +1327,7 @@
 /obj/machinery/door/airlock/pyro/maintenance{
 	req_access = null
 	},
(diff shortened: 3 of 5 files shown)


maps/donut2.dmm
```diff
@@ -17922,6 +17922,7 @@
 /obj/cable{
 	icon_state = "4-8"
 	},
+/obj/mapping_helper/access/security,
```
pgmzeta, line 17925: why sec and not hos?

cheekybrdy: Donut 2's back door is sec access, and the front one is HoP. So more QOL seeing that this one is technically "half solved" in regards to this PR

Sord213: It's actually on purpose not sec access for the front door so a hop can escape if sec comes in from the back to kick their teeth in for handing out AA

cheekybrdy: gotcha, changed it to just HoS access

JORJ949:
HoS access is special so these doors won't be able to be accesspro'd. I'd rather they got ID comp or security access, or that every map had an ID computer in the bridge.

JORJ949 (approved):
Ideally doors should only really have one access, and ID computer access feels weird to give doors so I myself would much rather just see every bridge have an ID computer or the HoS' computer have all accesses on it, but this _works_ as a solution.

pgmzeta:
bridges having ID computers is more Crime Compatible, there's probably room for each, especially if you replace a security/bank computer with the new one in customs by default

pgmzeta:
i've added ID computers to a sec checkpoint for the 2 maps that didn't already have security accessible ID computers (neon, cogmap2). The rest of the maps have one or more ID computers that security can already access.

PR #23713, opened 2025-06-03
Make gas flow scanner mech scannable.

klookiyam:
[ATMOSPHERICS][FEATURE]
## About the PR 
This change makes the meters over atmos pipes mechscannable. This will allow anyone to scan and deploy a meter over an atmos pipe to measure its content.

## Why's this needed? 
Constructable atmos is the new hot thing and a lot of the new pipe machinery is packet controllable, but theres no way to get any information about whats in the pipes. This change only involved a slight tweak to the existing meter logic. This opens the door for some fully automated atmos based monstrosities.

## Changelog 

```changelog
(u)KakoLookiyam
(+) Gas flow meters are now mechscannable and deployable..
```

### DIFF
--- a/code/obj/machinery/meter.dm
+++ b/code/obj/machinery/meter.dm
@@ -10,12 +10,18 @@
 	var/id
 	var/noiselimiter = 0
 
+TYPEINFO(/obj/machinery/meter)
+	mats = 1
+
 /obj/machinery/meter/New()
 	..()
-	SPAWN(1 SECOND)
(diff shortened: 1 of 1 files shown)


Laboredih123:
would prefer it was attachable to a specific pipe when pipes overlap, but since i havent figured out how to code that yet its whatever. code lgtm

Laboredih123:
> Missing decon flags I think?

are they needed? i guess a wrench and welder would probably be fine.

PR #23721, opened 2025-06-04
Fix Non-standard radio frequencies hearing 145.9

JORJ949:
## About the PR 
Changes everywhere a headset's frequency is set to use .set_frequency()

## Why's this needed? 
The radio's packet needs to be updated so this should be called instead to prevent headsets hearing 145.9 without it selected
@Mister-Moriarty had a different solution in mind but when I tried that each radio transmission would occur 4 times so I'll leave that change for now.


Mister-Moriarty:
> Mister-Moriarty had a different solution in mind but when I tried that each radio transmission would occur 4 times so I'll leave that change for now.

Your current implementation also has this bug.
I'm looking into fixing it; it's caused by two radio packet components being set to receive and send on the same frequency on the same object.

JORJ949:
Moriarty requested I close

PR #23726, opened 2025-06-04
Makes the Guts in meatstation accessible

cheekybrdy:
## About the PR 
Title, adds a button in meatstation to get to the Guts area. also a second ganglion for a part inside that'd otherwise be inaccessible.

## Why's this needed? 
The only way in is a existing Azone is in unfinished and will probably be unfinished content. Namely Advanced meat thaumaturgy. Also adds a use for that one false wall in the azone that leads to fuckall nowhere. Means that the entirety of a actually existing azone is doable.


ZeWaka:
i don't quite get this, from what i recall all parts of meatstation are accessible

cheekybrdy:
This refers to the area at the top, I've asked in #imspoiler and that's where I got the info about it being accessed from a deprecated azone, I've also searched the repo in vsc for mentions of the area or the blast door's ID and got nothing else that refers to the location, hence why I made the PR cause from what I could deduct along with what I've been told its not accessible.
image

ZeWaka:
I'm just going to close this, I'm not a huge fan of making an area with basically nothing in it suddenly accessible for no reason. Especially when it's content that's not going to be expanded upon in the future.

PR #23727, opened 2025-06-04
Bowling balls/bowling melons deal damage and disorient to non-human mobs

### LINKED ISSUE #23611: bowling melon do not hurt some monster/critter
### Labels

[BUG][MAJOR][HYDROPONICS]

### Description

if throwing the bowleing melon at some of the critter/monester, some time it do nothing

### Steps to reproduce

1. get the "Bowling Melon" mustation from melon in hydroponics and make very strong (good endurince! +200 endurance)
2. thrown the melon at some kind of enemies (have seen with the jeellyfish in Neon, rattlesnake - mob critter, i think)
3. throw more
4. throw more
5. it is not dead
6. thow more
7. caluculate damage: 12 * 30 = 360 BRUTE.
8. why alive?

### Expected Behavior

Killing! Killing! Killing! (why dose it not do damage?)

I sent ahelp for it in round, hitting jelliyfish with 30-40 melon but it do no thing! do not know if there is the fixing plannned, so ther is also bug here. Maybe I will fix...?but it is confusing...

if using the geoge melon, when it work (50%) the enemeis will be hurt (so using of the random_brute_damage(M, 10) command is working...)

bowling melon not so good if not fixed! very not good for fitghing enemeis.

NibChocolateeny:
[Hydroponics][Bug][QoL]

## About the PR 
change posision of if (ishuman(hitMob)) in if statement nesting so bowling ball and the bowling melon can do the damage+disoreint to mob that are not human. 

Fixes #23611

## Why's this needed? 
bug iss bad - it is not make any sense that the balls/melons do not do anyy damage to not humans.

## Changelog 
do not thing there is need for changeing log?

## other thing
i think there is other bug where antagonagonist bowler ball can make sound to hit mob two times - i do not think it is new iwth the change I make, not sure if it is only sound (not moredamage)??? can do more testing other time.

### DIFF
--- a/_std/__build.dm
+++ b/_std/__build.dm
@@ -30,9 +30,9 @@ o+`        `-` ``..-:yooos-..----------..`
 
 //------------ OPTIONS TO GO FAST ------------//
 
-#define GOTTA_GO_FAST_BUT_ZLEVELS_TOO_SLOW 1  // Only include the tiny map Devtest, no other zlevels. Boots way faster
+//#define GOTTA_GO_FAST_BUT_ZLEVELS_TOO_SLOW 1  // Only include the tiny map Devtest, no other zlevels. Boots way faster
 
-#define IM_REALLY_IN_A_FUCKING_HURRY_HERE 1  // All of the below
+//#define IM_REALLY_IN_A_FUCKING_HURRY_HERE 1  // All of the below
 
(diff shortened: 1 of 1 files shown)


NibChocolateeny:
silly,ingore the first name of pr

PR #23737, opened 2025-06-05
Throw EMAG to emag things

LorrMaster:
[game objects][feature]

## About the PR
- Atoms hit by a thrown EMAG get emagged.

## Why's this needed?
- Might open up some interesting strategies. At the very least it should be funny.

## Testing
Screenshot 2025-06-04 215236
Screenshot 2025-06-04 215400

## Changelog
```changelog
(u)LorrMaster
(+)Throwing an EMAG at things will now emag them.
```

### DIFF
--- a/code/obj/item/card.dm
+++ b/code/obj/item/card.dm
@@ -46,6 +46,12 @@ TYPEINFO(/obj/item/card/emag)
 	attack()	//Fucking attack messages up in this joint.
 		return
 
+	throw_impact(atom/hit_atom, datum/thrown_thing/thr)
+		..()
+		if(hit_atom && thr)
+			hit_atom.emag_act(thr.thrown_by, src)
+		return
+
 /obj/item/card/emag/attack_self(mob/user as mob)
 	if(ON_COOLDOWN(user, "showoff_item", SHOWOFF_COOLDOWN))
 		return


TobleroneSwordfish (approved):
Pfff why not

JORJ949 (approved):
Borg roguing: now at range.

pgmzeta:
how does this interact with graviton launchers? Mostly interested in who gets logged as the emagger in the situation where a graviton launcher tosses an emag at something.

slight worry about a graviton loop with an emag flinging around, letting borgs walk into it for free antag. might be hard to figure out who did the emagging exactly since someone could have set up the loop beforehand and the emag just got tossed/dropped into it at some point.

LorrMaster:
The emagging still works if the EMAG is thrown into a graviton loop, though the user that launched it won't be associated with it. Maybe have the user be the last person who touched it if there is no user associated with throwing it?

code/obj/item/card.dm
```diff
@@ -46,6 +46,12 @@ TYPEINFO(/obj/item/card/emag)
 	attack()	//Fucking attack messages up in this joint.
 		return
 
+	throw_impact(atom/hit_atom, datum/thrown_thing/thr)
+		..()
+		if(hit_atom && thr)
+			hit_atom.emag_act(thr.thrown_by, src)
```
pgmzeta, line 52:
```suggestion
			hit_atom.emag_act(thr.user, src)
```

code/obj/item/card.dm
```diff
@@ -48,7 +48,18 @@ TYPEINFO(/obj/item/card/emag)
 
 	throw_impact(atom/hit_atom, datum/thrown_thing/thr)
 		..()
-		if(hit_atom && thr)
+		if(!hit_atom)
+			return
+		if(!thr?.thrown_by)
```
pgmzeta: why not `thr.user`

TobleroneSwordfish:
Mmmm, on second thoughts maybe being able to build "walk here to be emagged" machines is a bad idea.

Scaltra-Volpe:
Would this emag things it flys over? Or does it need to hit the thing directly?

Azrun:
Conceptually cool and opens up more chaos... Emag behavior ranges wildly from goofy to pretty strong depending on what you are using it on, which makes balancing how it works as a whole difficult. 

As mentioned establishing the chain is custody/responsibility gets more hairy when the actions are not being directly initiated by a specific user.

JORJ949:
Should probably only work if directly thrown by a player (e.g. not gravitons, disposal chutes, mass launchers, etc.) to avoid any "automatic emag devices"

code/obj/item/card.dm
```diff
@@ -46,6 +46,26 @@ TYPEINFO(/obj/item/card/emag)
...
 
+	throw_impact(atom/hit_atom, datum/thrown_thing/thr)
+		..()
+		if(!hit_atom || !thr)
+			return
+		if(issilicon(hit_atom) && !thr.thrown_by)
+			// Need to throw at the cyborg directly to emag them
+			return
+		if(thr.user)
+			hit_atom.emag_act(thr.user, src)
```
JORJ949, line 57: Lots of emag acts expect the user to be right next to the target to give them a boutput, this may cause some weird outputs, but its probably fine unless an emag act does something to the user other than output to the user

PR #23738, opened 2025-06-05
Names Mutated eggs

cheekybrdy:
## About the PR 
Gives ranch mutation eggs edited names.
## Why's this needed? 
Easier identification for player who haven't memorised the sprites, also helps with grabbing out of produce stachels.
## Testing 
I assume this is what you mean for this part? I guess the ifdef secret part isn't tested for self-explanatory reasons.
image
image

## Changelog 

```changelog
(u)Cheekybrdy
(+)Ranch eggs are named according to their mutation
```


code/modules/ranch/chicken_products.dm
```diff
@@ -22,6 +22,26 @@
...
+				else
+					if(src = /obj/item/reagent_containers/food/snacks/ingredient/egg/chicken/coral)
+						name = "coral egg"
+					else if(src = /obj/item/reagent_containers/food/snacks/ingredient/egg/chicken/dragon)
+						name = "dragon egg"
+					else if(src = /obj/item/reagent_containers/food/snacks/ingredient/egg/chicken/phoenix)
+						name = "phoenix egg"
+					else if(src = /obj/item/reagent_containers/food/snacks/ingredient/egg/chicken/zappy)
+						name = "zappy egg"
+		#endif
```
TobleroneSwordfish, line 44: If this is merged these should just be committed to +secret, not janked in here.

TobleroneSwordfish:
Actually after talking with other devs I think we'll say eggs not having precise names is intentional to preserve a little bit of the "what does that mystery egg do" feeling.

PR #23745, opened 2025-06-05
Satchels only auto-fill with produce from harvests, ignore seeds

Egregorious:
## About the PR 

Produce satchels no longer have a popup option to harvest seeds or produce. Using them on a harvestable plant will instead  automatically harvest and place produce in satchel and dump the seed packet on the floor. 

## Why's this needed? 

Even before seed packets, I found the option to harvest seeds directly into the satchel was very rarely used. The mechanic feels practically obsolete now that there aren't multiple seed items per harvest. The popup is annoying and is entirely counter-productive to the qol that auto-filling satchels directly from harvests intends to confer. 

## Testing 

I'm not sure screenshots are useful for this because it's like trying to prove a negative.

Tested multiple harvests with an empty satchel; until satchel became full, after satchel was full, before plant was ready to harvest and after plant had died. 
Screenshot 2025-06-05 080451

## Changelog 

```changelog
(u)Egregorious
(+)Harvesting plants with a produce satchel now auto-fills it with produce directly and ignores the seed packet, instead of demanding input.
```

### DIFF
--- a/code/obj/machinery/plantpot.dm
+++ b/code/obj/machinery/plantpot.dm
@@ -869,17 +869,13 @@ TYPEINFO(/obj/machinery/plantpot)
 	// This proc is where the harvesting actually happens. Again it shouldn't need tweaking
 	// with since i've tried to account for most special circumstances that might come up.
 	if(!user) return
-	var/satchelpick = 0
 	if(SA)
 		if(length(SA.contents) >= SA.maxitems)
 			boutput(user, SPAN_ALERT("Your satchel is already full! Free some space up first."))
 			return
 		else
-			satchelpick = input(user, "What do you want to harvest into the satchel?", "[src.name]", 0) in list("Everything","Produce Only","Seeds Only","Never Mind")
-			if(!HYPcheck_if_harvestable() || satchelpick == "Never Mind")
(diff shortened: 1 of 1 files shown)


NibChocolateeny:
it is interesting idea...satchel still can hold seed, so good with large seed transporting. worth trying!

Scaltra-Volpe:
Personally when doing very large bulk producing i find the seeds are really annoying and I wish I could specifically get the seeds and only the seeds without a pop up menu.

Personally i would prefer to be able to still grab seeds directly with the satchel with some way to switch which is done (either way I would still even like this version because honestly its faster to handclick and clicl drag collect the seeds afterward anyway)

TobleroneSwordfish (approved):
I think this makes sense, the menu has always been annoying.

Egregorious:
> Personally i would prefer to be able to still grab seeds directly with the satchel with some way to switch which is done

I didn't bother preserving it because I didn't think anyone used the seed-only collection option honestly, but if people do then it's probably worth preserving.

I think this is doable with intent-specific interactions. I'll make a different PR with an attempt at it, since it'll be rather different than this implementation.

PR #23750, opened 2025-06-05
Options to change what satchels auto-fill with when harvesting

Egregorious:
## About the PR 

Alternative to #23745. I thought a different PR is in order because it's a very different implementation.

Produce satchels no longer have a popup option to harvest seeds or produce. They now have an internal value to determine what they auto-fill with, which can be cycled through by using a pen on them. Options are 'produce', 'seeds', and 'seeds and produce'. It defaults to produce-only. Satchels can be inspected to check what they're currently set to. 

## Why's this needed? 

The popup menu for quick-filling a satchel while harvesting completely subverts the point of the quick-fill mechanic by being annoying and tedious. This lets people still have access to all the auto-filling options, but front-loads the decision making so it's easy to spam. 

## Testing 

Tested multiple harvests with an empty satchel, full satchel, before plant was ready to harvest and after plant had died with each of the options.
Screenshot 2025-06-06 094708

## Changelog 

```changelog
(u)Egregorious
(+)Use a pen on a produce satchel to change what it will auto-fill with when harvesting plants.
```

### DIFF
--- a/code/obj/item/satchel.dm
+++ b/code/obj/item/satchel.dm
@@ -264,7 +264,7 @@
 		var/harvest_what = HARVEST_ONLY_PRODUCE
 
 		HELP_MESSAGE_OVERRIDE({"Click on it to pull out a random item, or click on it with <b>Grab Intent</b> to search for a specific item.
-		 	Click on it with <b>Disarm Intent</b> to change what you fill it with when harvesting."})
+		 	Use a pen on it to change what you fill it with when harvesting."})
 
 		New()
 			..()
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
Disarm intent for this is unhinged, much prefer the other PR.

Egregorious:
That's fair enough, although if the command action is the only issue that should be easy to swap out. Could attack yourself with it, or it be a tool-interaction.

Egregorious:
I've changed it to be a pen that you use to cycle through the options. It's maybe not the best option since it's something of an esoteric interaction without the help text, but the upside is every botanist starts with a pen, and it's less likely to be accidentally activated. It has at least a little verisimilitude I think, I've tried to make it so with the flavour text anyway.

PR #23763, opened 2025-06-07
Replace pug woof cooldown with stamina cost

TobleroneSwordfish:
## About the PR 
Reverts b89dd61 and replaces it with a stamina cost, if you try to woof while on negative stamina you fall over and take oxygen damage.
## Why's this needed? 
People like to spam the woof emote but it can get annoying when there's no end to it, adding a stamina cost still lets people have the funny barkarbkarkbarkbrrk but with the caveat that you will fall over if you choose to spam it for too long.
## Testing 

image
Me in shallow crit from excessive woofing
## Changelog 

```changelog
(u)LeahTheTech
(+)Pug woof once again has no cooldown but now has a stamina cost and will cause you to fall over wheezing if used too much.
```

### DIFF
--- a/code/mob/living/carbon/human/procs/emote.dm
+++ b/code/mob/living/carbon/human/procs/emote.dm
@@ -2290,15 +2290,14 @@
 					return
 				else
 					src.remove_stamina(STAMINA_DEFAULT_WOOF_COST)
-					if (src.get_stamina() > 0)
+					if (src.get_stamina() > 0 && !GET_COOLDOWN(src, "pug_woof_wheeze"))
 						message = "<b>[src]</b> woofs!"
 						playsound(src.loc, 'sound/voice/urf.ogg', 60, channel=VOLUME_CHANNEL_EMOTE)
 					else
-						if (!ON_COOLDOWN(src, "pug_woof_wheeze", 2 SECONDS))
-							message = "[src] wheezes"
-							playsound(src.loc, 'sound/voice/pug_wheeze.ogg', 60, channel=VOLUME_CHANNEL_EMOTE)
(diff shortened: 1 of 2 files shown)


code/mob/living/carbon/human/procs/emote.dm
```diff
@@ -2289,9 +2289,16 @@
...
-					message = "<b>[src]</b> woofs!"
-					if (src.emote_check(voluntary, 1 SECOND))
+					src.remove_stamina(STAMINA_DEFAULT_WOOF_COST)
+					if (src.get_stamina() > 0)
+						message = "<b>[src]</b> woofs!"
 						playsound(src.loc, 'sound/voice/urf.ogg', 60, channel=VOLUME_CHANNEL_EMOTE)
+					else
+						if (!ON_COOLDOWN(src, "pug_woof_wheeze", 2 SECONDS))
+							message = "[src] wheezes"
+							playsound(src.loc, 'sound/voice/pug_wheeze.ogg', 60, channel=VOLUME_CHANNEL_EMOTE)
```
JORJ949: Should this just call emote with wheeze instead to avoid code duplication?

TobleroneSwordfish: Actually why is this code even here when we have mutantrace emote handlers 🤔

TobleroneSwordfish: Oh I see why

flappybatpal:
I don't know why we'd make an exception for this which only existed due to bug/oversight when the point of the cooldown is to avoid spam.

TobleroneSwordfish:
Because dogs go barkbarkbarkbarkk!

pgmzeta (approved):

JORJ949 (approved):

PR #23777, opened 2025-06-08
moves senior description to grow_old to prevent exponential descriptions

cheekybrdy:
## About the PR 
title
## Why's this needed? 
Fixes a fuck up, += in get_desc kept adding to the description every time someone looked at it.


TobleroneSwordfish:
Better to generate it on the fly than modifying the desc var anyway: cfdb026

PR #23784, opened 2025-06-09
Makes hotspot burn-application more generic

Egregorious:
## About the PR 

Replaces burning-duration update calcs with a generic proc that lerps between a minimum and maximum burn duration based on temperature. Applies 30 seconds of burning at 373 kelvin, and 55 seconds of burning at 5000 kelvin.

## Why's this needed? 

Previous code was more difficult to maintain, and partially bugged. It invariably set burning to 55 seconds when stepped into, despite burning application for remaining within fire being a fraction of that; this keeps both methods of application more consistent.

Personally I might suggest crossing-into-fire being less intense, considering the number of activations scales with player movement and can thus trigger more often. However since noone was complaining while it was maxed out, that might be rocking the boat unnecessarily. 

## Testing 

Tested with the vega flamethrower and fireball for stepping into chemfires (chemfires never reach past the first process() deletion checks unless there is also a gasfire present and thus won't ever passively burn). Plasma fire for stepping into and remaining within gasfires. 

Before:
Screenshot 2025-06-09 122624
After: 
Screenshot 2025-06-10 211823

## Changelog 

```changelog
(u)Egregorious
(+)The burn-duration recieved from stepping into fires properly scales with the fire's temperature.
```

### DIFF
--- a/code/modules/atmospherics/FEA_fire.dm
+++ b/code/modules/atmospherics/FEA_fire.dm
@@ -261,8 +261,7 @@
 /atom/movable/hotspot/proc/update_status_effect(mob/living/target)
 	var/clamped_temp = clamp(temperature, FIRE_MINIMUM_TEMPERATURE_TO_EXIST, maximum_status_temp)
 	var/t = (clamped_temp - FIRE_MINIMUM_TEMPERATURE_TO_EXIST) / (maximum_status_temp - FIRE_MINIMUM_TEMPERATURE_TO_EXIST)
-	var burn = lerp(min_status_duration, max_status_duration, t)
-	target.update_burning(burn)
+	target.update_burning(lerp(min_status_duration, max_status_duration, t))
 
 
 /atom/movable/hotspot/ex_act()


M-Earthfire:
One simple fix to ruin all pyro chems. That's a "load bearing bug" if i ever seen one.

I think we may want to investigate what burn times this cause for which effects. Maybe have some rebalance changes prepared beforehand.

Egregorious:
There is always the option to just officially hardcode it to 55 seconds of additional burning, if that seems to have been working okay thus far.

TobleroneSwordfish:
No-one has ever complained! One thing that would be good to fix alongside that would be where certain fire sources will only set you on fire for 4 seconds if you stand still, but 55 if you move.

Egregorious:
That is this bug basically, there are two lines that deal with hotspot burning application. One in Crossed() for when a mob moves onto a burning tile: 
Screenshot 2025-06-09 221202
and one in process() for when a mob remains on a burning tile:
Screenshot 2025-06-09 221208


Combining both of them into a single generic proc would make it easier to maintain, but it would still mean a balance change one way or the other. Either walking through tiles is less impactful, or remaining in a burning tile is far more impactful.

ZeWaka:
Maybe we should have a minimum time/low end of the scale higher than zero. That'd make this a lot more tolerable. 

Like maybe we could try a range of 30-55 for now.

Egregorious:
Alrighty, I have done so. Combined both into a single call and balanced it between the suggested range, and it now lerps between a min and max temperature.

PR #23802, opened 2025-06-11
Clean up and restructure ailment code

pgmzeta:
[medical][internal]
## About the PR 
Attempts to clean up ailment code in a few different ways:
* Adds defines for the Spead and State field types
  * Removes string-matching in favor of numerical type defines
* Uses abstract types so we can filter by `concrete_typesof` in the disease controller
* Moves all cure calls on reagents to the `reagentcure` field on ailments
  * Removes the hardcoded spaceacillin CURE_ANTIBIOTICS flag, replacing it with CURE_MEDICINE as a general medical chem info for scanners
* Removes a few variables on ailments/ailment data:
  * `detectability` - this was not used at all in current ailment code
  * `recureprob` - the reagentcure list now requires the `id=prob` structure
  * `temperature_cure` - replaced by `high_temperature_cure` and `low_temperature_cure`
  * `stage_prob` - replaced by `stage_advance_prob` (same functionality)
  * `min_advance_ticks` - replaced by `advance_time_minimum`/`advance_time_maximum`
  * `tickcount` - replaced by `last_stage_change_time` on `ailment_data`
* Uses time-based minimum stage advancement instead of ticks for Shock
  * time is now correctly tracked on `ailment_data` and not the parent datum.
* Adds support for maximum stage advance time and cold body temperature for future ailment work
* Merges malady/disease stage_act logic to parent, allowing for more consistent logic and less duplication between different ailment subtypes
* Reworks the `scan_info` proc to better utilize flags and new defines
* Restructures the folder paths to better organize ailments by type
  * Separates ailments into multiple files where possible
* Adds doccomments

## Why's this needed? 
ailment code is old and some reagents were still attempting to cure diseases directly instead of utilizing the existing functionality in the ailment system.

## Testing 
Brief testing worked, will move out of draft when full testing is complete


pgmzeta:
will need secret changes

PR #23849, opened 2025-06-15
Runtimes on `logTheThing` with an empty `text` field

frawhst:
## About the PR 
Title
Making a PR and testmerging this so I can fix instances of the runtime as they occur.
## Why's this needed? 
- Catches cases where people accidentally put the text as the 2nd argument (source) instead of the 3rd (text)
- Generally good to error if we're logging empty messages since that shouldn't ever occur
## Testing 
It compiles!!


frawhst:
Haven't seen the runtime pop up on the tracker, seems fine for now.

PR #23850, opened 2025-06-16
remove random charger in medical floor on donut 2

Retrino:
## About the PR 
title
## Why's this needed? 
bug
## Testing 
file change
image
unchanged
image


JORJ949 (approved):
As its the sticky subtype it may be intended to stick to the window(they only stick to walls), but doctors don't need these rechargers, they have specialised ones for their defibs.

PR #23863, opened 2025-06-18
Blood/age cursers only affect the person who touches them

RubberRats:
[SCIENCE][BALANCE]
## About the PR 

This PR adds the variable `affects_multiple` to curse artifacts, which is set to FALSE for blood- and aging-type cursers. Curse artifacts will only seek out additional targets if it is set to TRUE

## Why's this needed? 

The blood and aging curses are disproportionately severe compared to how easy it is to get affected by them. They're extremely difficult to cure, requiring you to convince people to touch the obviously evil artifact (assuming you still have it), leave your body unclonable, and don't even have the courtesy of killing quickly. Multiple people have said that it's easier to kill and clone the poor victim rather than attempt to break the curse normally.

Normally, none of this would be a problem. Artifacts are inherently dangerous, and the crew shouldn't be randomly touching them like nothing. Having a wider variety of things for medbay to deal with is cool. However, right now it is way too easy to get cursed. Unlike any of the other deadly artifact types, you don't have to touch it, only stand within earshot of someone who does. I don't think "don't exist near random artifacts" is very practical or sensible as a preventative measure. Making the most severe versions of the curse only affect one person should make people more willing to actually engage with them instead of being overwhelmed by curse victims. At the very least, it should prevent random cases of spending the next 15 minutes dying horribly.

## Testing 

Tested this by touching spawned curse artifacts with and around target dummies with visible statuses enabled.

## Changelog 

```changelog
(u)RubberRats
(+)Blood and aging curse artifacts now only affect the person foolish enough to touch them.
```

### DIFF
--- a/code/obj/artifacts/artifact_objects/curser.dm
+++ b/code/obj/artifacts/artifact_objects/curser.dm
@@ -22,6 +22,7 @@
 	examine_hint = "It is covered in very conspicuous markings."
 	// general vars
 	var/chosen_curse
+	var/affects_multiple = TRUE
 	var/list/active_cursees = list()
 	var/static/list/durations = \
 		list(BLOOD_CURSE = null, AGING_CURSE = null, NIGHTMARE_CURSE = null, MAZE_CURSE = null, DISP_CURSE = 2 MINUTES, LIGHT_CURSE = 3 MINUTES)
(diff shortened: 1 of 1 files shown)


ZeWaka:
Sorry, but this is affected by the feature freeze given that it's a balance tweak that's not just changing a number or two. Feel free to reopen after the 22nd.

RubberRats:
#24286 was merged, closing this. Will have to look into making the standard cure method less esoteric later.

PR #23866, opened 2025-06-18
Fix salvager radio upgrade not having salv radio colour

JORJ949:
## About the PR 
Salvager headsets now use a consistent, non-random radio frequency for their private channel on 134.9
Salvager channel has its own colour and class definitions
Move setup for salvager headsets to the headset itself and not the antag equip
Changes salvager radio upgrade to work with the radio channel being set.

## Why's this needed? 
Fixes the salvager radio upgrade showing the general radio colour and general code cleanliness

## Testing 
Dark vs light mode demonstration (Colour is unchanged from current, but it being a class means you can have them differ)
image
image
Radio upgrade working:
image


Azrun:
> Salvager headsets now use a consistent, non-random radio frequency for their private channel on 134.9

What is the reason for that change?

JORJ949:
> > Salvager headsets now use a consistent, non-random radio frequency for their private channel on 134.9
> 
> What is the reason for that change?

Theres not much reason for it to need to be randomised, it allows for much simpler implementation and for the radio channel to still work without needing to spawn a salvager and get the radio frequency from their antagonist datum. 

Like nukies randomisation used to be a security thing so you couldn't just spy on them afaik, but this is no longer possible and nukies have since lost their randomised frequency in favor of the set syndicate frequency.

PR #23875, opened 2025-06-19
Remove MAP_OVERRIDE_MANTA and related defines

pgmzeta:
[internal][removal]
## About the PR 
Remove references to the following defines:
* `MAP_OVERRIDE_MANTA`
* `MOVING_SUB_MAP`
* `SUBMARINE_MAP`

Does NOT remove related objects or assets. Does remove some manta-specific events.

## Why's this needed? 
Code cleanup

## Testing 
Compiled Oshan/Nadir/Neon local, all worked as expected.

### DIFF
--- a/code/datums/controllers/job_controls.dm
+++ b/code/datums/controllers/job_controls.dm
@@ -1076,7 +1076,7 @@ var/datum/job_controller/job_controls
 	if (!string || !istext(string))
 		logTheThing(LOG_DEBUG, null, "<b>Job Controller:</b> Attempt to find job with bad string '[string]' in controller detected")
 		return null
-	var/list/excluded_strings = list("Special Respawn","Custom Names","Everything Except Assistant", "Communications Officer",
+	var/list/excluded_strings = list("Special Respawn","Custom Names","Everything Except Assistant",
 	"Engineering Department","Security Department","Heads of Staff", "Pod_Wars", "Syndicate", "Construction Worker", "MODE", "Ghostdrone", "Animal")
 	if (string in excluded_strings)
 		return null
--- a/code/datums/controllers/job_controls.dm
+++ b/code/datums/controllers/job_controls.dm
@@ -1076,11 +1076,8 @@ var/datum/job_controller/job_controls
 	if (!string || !istext(string))
 		logTheThing(LOG_DEBUG, null, "<b>Job Controller:</b> Attempt to find job with bad string '[string]' in controller detected")
 		return null
-	var/list/excluded_strings = list("Special Respawn","Custom Names","Everything Except Assistant",
+	var/list/excluded_strings = list("Special Respawn","Custom Names","Everything Except Assistant", "Communications Officer",
 	"Engineering Department","Security Department","Heads of Staff", "Pod_Wars", "Syndicate", "Construction Worker", "MODE", "Ghostdrone", "Animal")
-	#ifndef MAP_OVERRIDE_MANTA
-	excluded_strings += "Communications Officer"
-	#endif
 	if (string in excluded_strings)
 		return null
 	var/list/results = list()
--- a/code/datums/jobs.dm
+++ b/code/datums/jobs.dm
@@ -324,9 +324,6 @@ ABSTRACT_TYPE(/datum/job/command)
 	announce_on_join = TRUE
 
 
-#ifdef SUBMARINE_MAP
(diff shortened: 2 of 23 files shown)


code/obj/storage/closets.dm
```diff
@@ -231,11 +231,8 @@ TYPEINFO(/obj/storage/closet/coffin)
 	spawn_contents = list(
 	/obj/item/clothing/mask/breath,
 	/obj/item/clothing/under/misc/syndicate,
-#if defined(MAP_OVERRIDE_MANTA)
 	/obj/item/tank/jetpack/syndicate,
-#else
 	/obj/item/tank/jetpack,
```
JORJ949: This locker is only remaining on neon afaik (assuming its the syndicate gear one, superceded by apparel vendor, I can only see contents), doesn't need both a syndicate and standard jetpack

code/datums/controllers/job_controls.dm
```diff
@@ -1076,11 +1076,8 @@ var/datum/job_controller/job_controls
 	if (!string || !istext(string))
 		logTheThing(LOG_DEBUG, null, "<b>Job Controller:</b> Attempt to find job with bad string '[string]' in controller detected")
 		return null
-	var/list/excluded_strings = list("Special Respawn","Custom Names","Everything Except Assistant",
+	var/list/excluded_strings = list("Special Respawn","Custom Names","Everything Except Assistant", "Communications Officer",
```
JORJ949: Comms officer shouldn't be here, the map override was an ifndef then add comms officer to this list because comms officer was undefined if manta wasn't defined, now that comms officer's just special it shouldn't be in this list.

JORJ949:
Bunch of communications officer being command things may also need changing if you want it to be a special job and not command
- How does it show in the manifest? /code/datums/datacore.dm#L325 Seems like it will just appear in the command department and nowhere else
- Should comms officer remain a target of the revolution? They are included as a head of staff and unconvertable for the purposes of revs
- Comms Officer is still in the command_jobs, but should probably be command_gimmicks like NTSC and Head of Mining

code/datums/jobs.dm
```diff
@@ -1220,15 +1192,31 @@ ABSTRACT_TYPE(/datum/job/special)
...
 /datum/job/special/radioshowhost
 	name = "Radio Show Host"
 	wages = PAY_TRADESMAN
 	linkcolor = CIVILIAN_LINK_COLOR
 	access_string = "Radio Show Host"
-#ifdef MAP_OVERRIDE_MANTA
-	limit = 0
-	special_spawn_location = null
-#elif defined(MAP_OVERRIDE_OSHAN)
+#if defined(MAP_OVERRIDE_OSHAN)
```
frawhst:
```suggestion
#ifdef MAP_OVERRIDE_OSHAN
```
nitpick, but since this isn't an `elif` anymore we can use the more succinct syntax

code/modules/fluids/fluid_commands.dm
```diff
@@ -124,9 +124,7 @@ client/proc/replace_space_exclusive()
 
 			var/turf/orig = locate(S.x, S.y, S.z)
 
-#if defined(MOVING_SUB_MAP)
-			var/turf/space/fluid/manta/T = orig.ReplaceWith(/turf/space/fluid/manta, FALSE, TRUE, FALSE, TRUE)
-#elif defined(UNDERWATER_MAP)
+#if defined(UNDERWATER_MAP)
```
frawhst:
```suggestion
#ifdef UNDERWATER_MAP
```
this one existed before but may as well shorten it while we're here >:{

pgmzeta: we dont actually need this at all now that i look at it closer. why wasn't it a if-elif before ??

pgmzeta:
>     * How does it show in the manifest? /code/datums/datacore.dm#L325 Seems like it will just appear in the command department and nowhere else
moving it from command_jobs to command_gimmicks makes it show in unassigned/gimmicks. removed the now-unneeded conditional in the datacore code.
> 
>     * Should comms officer remain a target of the revolution? They are included as a head of staff and unconvertable for the purposes of revs
I went with non-target for revs, but unconvertable like other command gimmicks (Head of Mining)

> 
>     * Comms Officer is still in the command_jobs, but should probably be command_gimmicks like NTSC and Head of Mining

good catch, moved

PR #23884, opened 2025-06-20
Captain's Bonsai uses health, but only has 5hp

### LINKED ISSUE #23341: Participation Ribbon Ghost can break the Captain's Bonsai
### Labels

[BUG][MINOR]

### Description

Becoming a participation ribbon, and trying to sticker the bonsai, results in the tree breaking

### Steps to reproduce

1. be ghost
2. transform into ribbon
3. try to give the bonsai a nice gift
4. CRUNCH

### Expected Behavior

For it to not interact with an antagonist objective, and to stick harmlessly

Reported by: sunnyaries
Client version: 515.1647
Active test merges: #23292
Round log date: 2025-05-07-17-02
Reported on: 2025-05-07 18:03:44
Round ID: 205115
Map: NEON

JORJ949:
## About the PR 
Makes the captain's bonsai have health like every other shrub, meaning it can only be destroyed by an item that does damage.

## Why's this needed? 
Allows stickers to be stuck to the bonsai, Fixes #23341
Prevents unintended items destroying shrubs.

## Testing 
image
Shrub didn't break when using an ID card on it, did break when using wirecutters

```changelog
(u)JORJ949
(*)The Captain's Bonsai now uses health like other bushes, but is very fragile and only has 5hp!
```

### DIFF
--- a/code/obj/decorations.dm
+++ b/code/obj/decorations.dm
@@ -518,6 +518,8 @@ TYPEINFO(/obj/shrub/syndicateplant)
 		if (inafterlife(user))
 			boutput(user, "You can't bring yourself to hurt such a beautiful thing!")
 			return
+		if(istype(W, /obj/item/sticker)) //Stickers stick to it rather than destroy
+			return
 		if (src.destroyed) return
 		if (user.mind && user.mind.assigned_role == "Captain")
 			if (issnippingtool(W))
--- a/code/obj/decorations.dm
+++ b/code/obj/decorations.dm
@@ -472,12 +472,15 @@ TYPEINFO(/obj/shrub/syndicateplant)
 	icon = 'icons/misc/worlds.dmi'
 	icon_state = "bonsai"
 	desc = "The Captain's most prized possession. Don't touch it. Don't even look at it."
+	health = 5 //Fragile tree
 	anchored = ANCHORED
 	density = 1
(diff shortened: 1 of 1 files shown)


flappybatpal:
Would it be better to check for force?

JORJ949:
> Would it be better to check for force?

Possibly?

JORJ949:
> Unless there are issues with it I'm going to ask for intent and item force to be checked to catch more things that wouldn't be expected to do damage and reduce accidents.

Updated to call the parent attackby so that it uses health and takes damage like every other shrub does, but leaving it with only 5hp so most items still instantly break it.

pgmzeta:
why not just use intent and item force directly instead ?

JORJ949:
> why not just use intent and item force directly instead ?

Bush code already manages item force, specifically taking the force from their health and calling destroy if thats <=0, I see no need to reinvent the wheel here when parent shrub behaviour already manages this but without an intent check.

pgmzeta:
the bonsai is Special and an antagonist objective. it should be harder to accidentally kill it than a bush imo

TobleroneSwordfish (approved):

TobleroneSwordfish:
Intent is w/e when it comes to object interactions, I think it behaving like a normal shrub is better

PR #23895, opened 2025-06-21
Disallow several APCs in one area

### LINKED ISSUE #23892: Can't unlock cover in custom APC
### Labels

[BUG][MAJOR]

### Description

The "unlock cover" button isn't appearing in the menu for a custom APC I built in the sea diner.

### Steps to reproduce

1. Construct an APC in the sea diner (specifically "space diner hallway" in the bottom right of the space diner on Oshan)
2. Swipe your ID to unlock the APC
3. Open with a screwdriver, then try to use a crowbar
4. The cover is locked, and there's no button in the UI to unlock it

### Expected Behavior

I expected the "unlock cover" button to appear in the APC's menu when the APC is unlocked.

_No response_

MyNameIsRomayne:
## About the PR 
Removes the ability to place an APC into an area which already has one, instead outputting a alert that one already controls the area.

## Why's this needed? 
Fixes #23892, since code is really only made to handle 'most recent APC' so multiple APCs, on top of being niche and not really having use cases, are just confusing and buggy.

## Testing 

https://github.com/user-attachments/assets/487ff72f-800c-4219-91b0-2f3f3174b050

(Video is from original PR code, but is still relevant. I re-tested the behavior with an APC and a reverse-engineering fab by building both and watching for runtimes. The only change between the recent code and the video is the bouput is generalized to \the [src])

### DIFF
--- a/code/atom.dm
+++ b/code/atom.dm
@@ -475,8 +475,8 @@ TYPEINFO(/atom/movable)
 	/// See `/datum/manufacturing_requirement/match_property` for match properties
 	var/list/mats = null
 
-	/// Dummy proc for all /atom/movable typeinfos to be overriden and called to see if an object type
-	/// can be built somewhere, before instantiating the object itself.
+	/// Dummy proc for all /atom/movable typeinfos to be overriden and called to see
+	/// if an object type can be built somewhere, before instantiating the object itself.
 	proc/can_build(turf/T)
 		return TRUE
 
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -26,7 +26,6 @@ TYPEINFO(/obj/machinery/power/apc)
 		var/area/A = get_area(T)
 		return A.area_apc == null
 
-
 ADMIN_INTERACT_PROCS(/obj/machinery/power/apc, proc/toggle_operating, proc/zapStuff)
 
 /obj/machinery/power/apc


code/obj/item/tool/electronics.dm
```diff
@@ -303,6 +305,19 @@
...
 		qdel(src)
 
+// This really just exists right now for custom APC logic to have its own space,
+// but if you want to judge whether other objs/types can be built do it here!
+/obj/item/electronics/frame/proc/can_build()
+	var/type = src.store_type
+	if(src.deconstructed_thing != null)
+		type = src.deconstructed_thing.type
+	switch(type)
+		if("/obj/machinery/power/apc") // for ensuring 2 APCs aren't built in the same area
```
TobleroneSwordfish: A couple of things:
- `type` here should be a type path, I can't imagine matching it to a string would work?
- This should just be a proc on `/obj` like `can_deconstruct`, instead of re-inventing polymorphism with a switch case here

TobleroneSwordfish (changes requested):

code/obj/item/tool/electronics.dm
MyNameIsRomayne: - Woops, that's embarrasing. I thought I made that commit beforehand which fixed this!
- How would that work? I can't call the proc on the object without instantiating it (as far as i recall).

TobleroneSwordfish: Ohh right, Byond still doesn't have static methods. Maybe define it on typeinfo?

MyNameIsRomayne: yeah... i suppose x.x its what i ended up doing for recipes i think. ill see if i can take a look tomorrow!

JORJ949:
Immediate thought is this would also work for AI law racks rather than them just immediately dissasembling themselves if not allowed.

MyNameIsRomayne:
True! Also my eyes are on this still, just busy now. Ill try tomorrow morning or tonight depending on how things go

PR #23910, opened 2025-06-22
Eat limb causes you to latch onto your victim

CalliopeSoups:
[critters] [balance]
## About the PR 
This PR causes eat limb to latch you onto your victim. This fixes an issue where victims moving would not interrupt the action bar, now that makes sense for the ability's function.

## Why's this needed? 
The issue needed to be fixed somehow, but also this provides more interacts with the mimic's kit. It should be advantageous now to disguise as a large object so you can tank hits while you eat limbs off of people, bypassing the need to stun them in some way.

## Testing
Seems to work fine in general, known issue however is the victim's attack delay is gone when you hit someone latched onto you. This doesn't seem to wipe base form mimics if the victim is unarmed, so perhaps this could be funny as a diegetic "the victim is going AAHGHGH!!! GET IT OFF ME" thing.

dreamseeker_veMLxVCqgG

## Changelog 

```changelog
(u)CalliopeSoups
(+)Mimic's eat limb ability now has you latch onto your victim.
```

### DIFF
--- a/code/datums/abilities/critter/eat_limb.dm
+++ b/code/datums/abilities/critter/eat_limb.dm
@@ -8,20 +8,12 @@
 	cooldown_after_action = TRUE
 	var/datum/human_limbs/torn_limb
 
-	New()
-		..()
-		// stomach retreat can get away with being by itself, but eat limb should always come bundled with it
-		if (!holder.owner.getAbility(/datum/targetable/critter/stomach_retreat))
-			holder.owner.addAbility(/datum/targetable/critter/stomach_retreat)
-		if (!holder.owner.GetComponent(/datum/component/death_barf))
-			holder.owner.AddComponent(/datum/component/death_barf)
-
 	cast(atom/target)
 		. = ..()
 		var/datum/targetable/critter/stomach_retreat/stomach_abil = src.holder.getAbility(/datum/targetable/critter/stomach_retreat)
 		if (stomach_abil?.inside)
 			return TRUE
-		if (ishuman(target))
+		if (ishuman(target) && target != src.holder.owner)
 			var/mob/living/carbon/human/targetHuman = target
 			var/list/randLimbBase = list("r_arm", "r_leg", "l_arm", "l_leg")
 			var/list/randLimb
@@ -42,7 +34,7 @@
 	interrupt_flags = null
 	var/atom/target
 	var/mob/user
-	var/datum/human_limbs/tornlimb
+	var/obj/item/parts/tornlimb
 
 	New(Target, User, Tornlimb)
 		target = Target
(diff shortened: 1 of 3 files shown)


CalliopeSoups:
The protect them commit was meant to be for another PR.

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -20,30 +21,57 @@
...
+		tornlimb = Tornlimb
 		..()
 
 	onStart()
 		..()
 		if (ishuman(target))
 			src.duration = 5 SECONDS
+			APPLY_ATOM_PROPERTY(user, PROP_MOB_CANTMOVE, "chomping")
+			var/mob/living/carbon/human/human = target
+			user.set_loc(human)
```
TobleroneSwordfish: This is the issue, combat click delay isn't applied to atoms inside the mob's contents. Luckily, we already have a flag for this: `CLICK_DELAY_IN_CONTENTS`
As for the balance, I'd imagine leaving this out would make this ability instance suicide if you attack anyone with even a fire extinguisher, they'd be able to do several times the mimic's max health in a couple of seconds.

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -20,30 +21,57 @@
...
+			human.vis_contents += user
+			if (istype(tornlimb, /obj/item/parts/human_parts/leg/))
```
TobleroneSwordfish: Don't use types to check the side, use slots. What if they have a non-human arm?

TobleroneSwordfish (changes requested):
this is true mimic class fantasy 👍

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -6,80 +6,104 @@
...
 	target_anything = TRUE
 	cooldown_after_action = TRUE
-
-	New()
-		..()
-		// stomach retreat can get away with being by itself, but eat limb should always come bundled with it
-		if (!holder.owner.getAbility(/datum/targetable/critter/stomach_retreat))
-			holder.owner.addAbility(/datum/targetable/critter/stomach_retreat)
-		if (!holder.owner.GetComponent(/datum/component/death_barf))
-			holder.owner.AddComponent(/datum/component/death_barf)
```
TobleroneSwordfish: ?

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -6,80 +6,104 @@
...
 		..()
-		if (ishuman(target))
+		if (ishuman(src.target))
 			src.duration = 5 SECONDS
+			APPLY_ATOM_PROPERTY(user, PROP_MOB_CANTMOVE, "chomping")
+			var/mob/living/carbon/human/human = src.target
+			LAZYLISTREMOVE(human.attached_objs, src.user)
+			src.user.set_loc(src.target.loc)
+			src.user.transform = matrix(src.user.transform, 90, MATRIX_ROTATE | MATRIX_MODIFY)
+			user.flags += CLICK_DELAY_IN_CONTENTS
```
TobleroneSwordfish: Not quite how flags work, this is a bitfield. See `ADD_FLAG` and `REMOVE_FLAG`

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -6,80 +6,104 @@
...
 		..()
 
 	onStart()
 		..()
-		if (ishuman(target))
+		if (ishuman(src.target))
 			src.duration = 5 SECONDS
+			APPLY_ATOM_PROPERTY(user, PROP_MOB_CANTMOVE, "chomping")
+			var/mob/living/carbon/human/human = src.target
+			LAZYLISTREMOVE(human.attached_objs, src.user)
```
TobleroneSwordfish: Shouldn't this be add?

code/datums/abilities/critter/eat_limb.dm
```diff
@@ -6,80 +6,104 @@
...
-		src.gobble(target, user)
+		if (ishuman(src.target))
+			var/mob/living/carbon/human/human = src.target
+			src.user.transform = null
+			LAZYLISTREMOVE(human.attached_objs, src.user)
+			src.user.set_loc(src.target.loc)
+			src.user.pixel_y = 0
+			src.user.pixel_x = 0
+			take_bleeding_damage(src.target, src.user, 15, DAMAGE_CUT, 1)
+			REMOVE_ATOM_PROPERTY(src.user, PROP_MOB_CANTMOVE, "chomping")
```
TobleroneSwordfish, line 106: Make sure to remove the flag here too

TobleroneSwordfish (changes requested):

CalliopeSoups:
Doing some more testing now, will edit when I think it's ready.

Edit: I've touched all the things I can think of to test. Suplexing throws an error I didn't quite understand, so I gave them anti grab which felt fitting anyway and likely avoids other weird niche things. The user can still grab the victim mid-use, which I tested and works fine, looks kinda funny too.

(Note that the action bar is 20 seconds long here for testing, it's 5 seconds usually.)
dreamseeker_ovPsSPX9h3

PR #23913, opened 2025-06-22
Fix reagent dispensers being unscannable by spectro eyes/goggles

JORJ949:
## About the PR 
Updates the reagent container flags on reagent dispensers to be more accurate to:
- Allow reagent_dispensers to be scanned by spectroscopic eyes, goggles and borg upgrades.
- Allow seeing the colour of the contents of chemical barrel

## Why's this needed? 
I see no reason these shouldn't be scannable with spectroscopic scanners when they are by reagent analyzers, and chemical barrels have the liquid visible on the side so you should be able to judge the colour on it.

## Testing 
Red line being removing the spectroscopic goggles
image

### DIFF
--- a/code/modules/chemistry/tools/dispensers.dm
+++ b/code/modules/chemistry/tools/dispensers.dm
@@ -10,6 +10,7 @@
 	icon_state = "watertank"
 	density = 1
 	anchored = UNANCHORED
+	var/rc_flags = RC_SCALE | RC_SPECTRO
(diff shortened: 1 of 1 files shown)


pgmzeta:
imo getting close to some things to check them shouldn't be removed

JORJ949:
> imo getting close to some things to check them shouldn't be removed

The reagent dispensers specifically already need you to be within 2 tiles to examine how much is in them, unlike other spectroscannable things.

PR #23915, opened 2025-06-22
Remove "slap the ass (hot)" from photocopier output text

JORJ949:
## About the PR 
Changes "You slap the ass (hot) on the scan bed, close the lid, and press start..." to "You put the butt on the scan bed, close the lid, and press start..."

## Why's this needed? 
I don't think this text really fits in the goonstation environment, it feels really out of place.

## Testing 
image


TobleroneSwordfish:
pfff fair, yes

PR #23920, opened 2025-06-22
Add Flavor Text to Cyborgs

sunnyariess:
[SILICONS] [INPUT WANTED]
## About the PR 
Adds a "Description" field to docking stations, changeable by both the Silicon inside, and people outside, same as its name. This field is bound by the same length and checks as a human's flavor text field. (aka, it can be up to 256 characters).
Does not add this field, or the option to change it, to Eyebots, AIs, or unfortunate Carbons inside the docking station.
## Why's this needed? 
Flavour text on Silicons is a request I've seen floating around for quite some time. I think it really adds to the roleplay potential of cyborg players, the same way a human's flavor text does.
## Testing 
https://github.com/user-attachments/assets/809f09b6-2334-421c-8dbe-2046e875e09f
image
image
image
Tested with no text, minimal text, maximum text. As a cyborg, AI shell, Eyebot and a Carbon inside the station. 

## Changelog 

```changelog
(u)sunnyaries
(*)Added a flavour text field for Cyborgs, accessible from Docking Stations.
```


TobleroneSwordfish:
Mmm sorry, but I'm going to veto this, it really really doesn't make sense.
Flavour text is a meta concept designed to allow for adding little details to pre-set characters that wouldn't otherwise be available using existing character customization options. It does not make sense for A) a cyborg made out of standard and interchangeable parts to *have* flavour text, and B) it *REALLY* doesn't make sense for it to be editable in-universe during the round.

For example, imagine you have `This cyborg has a small rusty patch on their left leg, little purple crystals glitter between the servos in their heel, a remnant of their brush with the void` - that's really cool flavour text BUT now imagine you're the roboticist who just built that shell from scratch with treads.

Equally, if you're able to edit it during the round then you're conjuring these little details from nowhere, and changing fundamental parts of their "character" quite possibly against the cyborg's will.

*breathein
Sorry for the wall of text but I feel like I've had this discussion several times over at this point and want to make it clear why we're not doing this.

JORJ949:
Humans have flavour text because there aren't enough pixels to show unique detail that we could have at a higher resolution or scale, for example a 3D model. Cyborgs at this theoretical 3D model would still have the same level of customization they do now.

Microoowaaave:
Was really excited for this, sad to see it shot down so quickly. I don't play silicons much myself (though this may have swayed me towards doing such), but I know a lot of people with really interesting borg characters that could benefit from being able to have flavor text, and I'm absolutely willing to suspend my disbelief a bit for the sake of that.

sunkiisu:
I can see the concerns with this from an in-game standpoint, with how cyborgs are usually made with parts you print out at the fab, but I also can't really understand why there's a need to be so strict about such a small bit of gameplay that isn't even accessible to view unless you think to observe someone's pc? Like respectfully it's fair that someone may be confused as to why the robot they built now has a cooler arm than they thought it did or the like, but I don't feel this would change anything fundamental about how people had been playing borg regardless. I was really excited to see a more intuitive fix to what's usually just conveyed through emotes, because the borg sprite, like human sprites, are so small in game that you could make up what isn't shown to be a little creative.

It's harmless, completely optional, and makes people's experience a little more fun, so I'm frankly just confused why it can't be overlooked in the same way a lot of things that conflict in-game function (that are circumvented through rp) can be.

PR #23921, opened 2025-06-22
Clown Disbelievers can't hear Clowns' speech

JORJ949:
[Traits][Feature]

## About the PR 
Prevents anyone with the Clown Disbelief trait from hearing the speech of anyone that it deems a "clown", those being anyone wearing a clown or cluwne uniform. This includes both local and radio chat.

## Why's this needed? 
Clowns don't exist, I don't want to hear that which doesn't exist

## Testing 
With some testing being a clown with the trait could hear myself when I took off my uniform, could not when I put it back on.

## Changelog 

```changelog
(u)JORJ949
(*)The Clown Disbelief trait now stops you from hearing strange squeaky voices (anything a clown says).
```


_std/defines/speech_defines/modules.dm
```diff
@@ -219,11 +219,11 @@
 
 //------------ Listen Modifiers ------------//
 #define LISTEN_MODIFIER_CHAT_CONTEXT_FLAGS "chat_context_flags"
+#define LISTEN_MODIFIER_CLOWN_DISBELIEF "clown_disbelief"
 #define LISTEN_MODIFIER_MOB_MODIFIERS "mob_modifiers"
 #define LISTEN_MODIFIER_PHONE "phone_formatting"
 #define LISTEN_MODIFIER_RADIO "radio_formatting"
 
-
 //------------ Listen Effects ------------//
```
Mister-Moriarty: _My whitespace nooo._

code/datums/components/clown_disbelief_item.dm
```diff
@@ -1,4 +1,5 @@
 var/global/list/image/clown_disbelief_images
+var/global/list/image/clown_disbelief_clown_mobs //Subscribed list of mobs that are currently deemed to be clowns
```
Mister-Moriarty: I'd doc comment this, and make it slightly more explicit what the list is recording.
I.e. `/// An associative list of clown disbelief item subscription counts, indexed by the mob they've been applied to.`
To me, this makes it clear that if you pass a mob to the list as an index, you're getting back a subscription count.

code/datums/components/clown_disbelief_item.dm
```diff
@@ -19,15 +20,24 @@ TYPEINFO(/datum/component/clown_disbelief_item)
 	src.disbelief_image.override = 1
 	if(isnull(global.clown_disbelief_images))
 		global.clown_disbelief_images = list()
+	if(isnull(global.clown_disbelief_clown_mobs))
+		global.clown_disbelief_clown_mobs = list()
```
Mister-Moriarty, line 24: Feel free to ignore this suggestion as the line above doesn't implement it, but a shorthand for this is `global.clown_disbelief_clown_mobs ||= list()`.

code/datums/components/clown_disbelief_item.dm
```diff
@@ -19,15 +20,24 @@ TYPEINFO(/datum/component/clown_disbelief_item)
...
 	global.clown_disbelief_images += src.disbelief_image
 	for(var/mob/M as anything in by_cat[TR_CAT_CLOWN_DISBELIEF_MOBS])
 		M.client?.images += src.disbelief_image
 
 /datum/component/clown_disbelief_item/proc/on_equip(datum/source, mob/equipper, slot)
 	src.disbelief_image.loc = equipper
+	if(!clown_disbelief_clown_mobs[equipper])
+		clown_disbelief_clown_mobs[equipper] = 1
+	else
+		clown_disbelief_clown_mobs[equipper] +=1
```
Mister-Moriarty:
```suggestion
	clown_disbelief_clown_mobs[equipper] += 1
```
We live in a world where `null + 1 => 1`.

Mister-Moriarty (changes requested):

code/datums/components/clown_disbelief_item.dm
JORJ949: ||= is in my mind a crime that I will NOT be joining you in.

Mister-Moriarty: Alas, condemned to crime alone.

Mister-Moriarty (approved):
Speech code looks fine.

ZeWaka:
X

ZeWaka (approved):

TobleroneSwordfish:
You can still hear the speech sound but honestly that's funny

JORJ949:
> You can still hear the speech sound but honestly that's funny

I would love to implement also being unable to hear the sounds clowns make but I've not messed with audio code so don't even know if its possible.

PR #23924, opened 2025-06-23
Brain damage rework - More brain damage effects

FlameArrow57:
[MEDICAL][FEATURE]

## About the PR 
Adds some additional effects for brain damage

Added:
-At 20 brain damage:
--You start fumbling some letters in your words

-At 40 brain damage:
--You will periodically vomit and receive messages about your head hurting
--More of your speaking has a chance to be affected

-At 60 brain damage:
--You have a chance to become blinded for a bit

-At 80 brain damage:
--You can no longer hear others' words properly
--Your speaking becomes incomprehensible
--You gain a chance to misstep when moving

Removed:
-You no longer randomly speak over the radio for having brain damage
-The "blank expression" message you get for having brain damage only appears now when you have 100 brain damage or have Expresso in you

## Why's this needed? 
-Adds more of a ramp up to lethal brain damage rather than something that can be silent until you suddenly start dying
-Adds more interesting effects that make brain damage a unique type of damage

## Testing 
Tested speaking messages at 25 brain damage increments, manually calling take_brain_damage() proc on myself

## Changelog 

```changelog
(u)FlameArrow57
(*)Added/removed effects of brain damage, see minor changelog for details
(+)As brain damage increases, you will now experience effects such as impaired speech, vomiting, messages about your head hurting, going blind temporarily, and being unable to hear others. Effects increment at every 20 value increase of brain damage
(+)You no longer randomly speak over the radio if you have brain damage
(+)The "blank expression" description for someone with brain damage now only appears if they have lethal (100) brain damage or Expresso in them
```


flappybatpal:
Sensible but I'm not sure on the vomit as we've historically had way too many things causing vomiting at a high frequency.

JORJ949:
> Sensible but I'm not sure on the vomit as we've historically had way too many things causing vomiting at a high frequency.

Vomit should probably be applied via nausea stacks (like most other vomit methods) and only at 60+

code/mob/living/life/brain.dm
```diff
@@ -0,0 +1,29 @@
...
+
+		if (!istype(H))
+			return
+
+		var/brain_dmg = H.get_brain_damage()
+		var/mult = get_multiplier()
+
+		if (brain_dmg >= BRAIN_DAMAGE_MAJOR)
+			if (probmult(8))
+				H.vomit()
```
TobleroneSwordfish:
```suggestion
			if (probmult(30))
				H.nauseate(1)
```

code/mob/living/life/brain.dm
```diff
@@ -0,0 +1,29 @@
...
+
+			if (probmult(5))
+				H.changeStatus("blinded", rand(10, 25) SECONDS)
+		else if (brain_dmg >= BRAIN_DAMAGE_MODERATE)
+			if (probmult(4))
+				H.vomit()
```
TobleroneSwordfish:
```suggestion
			if (probmult(15))
				H.nauseate(1)
```

PR #23933, opened 2025-06-24
Brain damage rework - Inoperable machinery

FlameArrow57:
[MEDICAL][FEATURE][REWORK][BALANCE]

## About the PR 
Reworks how brain damage interacts with machinery

Having greater than 60 brain damage no longer has a random chance to make common machinery inoperable, it is now guaranteed

In addition to machinery, APCs and PDAs are now unable to be used

Nanomeds are excluded so that you can still get mannitol

## Why's this needed? 
-Makes brain damage machine inoperability more consistent
-Adds more of a unique effect to having brain damage that you need to watch out for

## Testing 
Tested various machinery at 75 brain damage in addition to APCs, PDAs, vending machines, nanomeds, chem dispensers, computers, and manufacturers

## Changelog 

```changelog
(u)FlameArrow57
(*)Major (60) brain damage now has a guaranteed chance to make common machinery inoperable, in addition to PDAs and APCs. Nanomeds are excluded.
```

### DIFF
--- a/code/obj.dm
+++ b/code/obj.dm
@@ -13,6 +13,8 @@
 	var/cannot_be_stored = FALSE
 	var/move_triggered = 0
 	var/object_flags = 0
+	/// if this object can be used with high brain damage
+	var/usable_with_brain_damage = FALSE
 
 	animate_movement = 2
 //	desc = SPAN_ALERT("HI THIS OBJECT DOESN'T HAVE A DESCRIPTION MAYBE IT SHOULD???")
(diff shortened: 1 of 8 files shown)


code/obj.dm
```diff
@@ -453,3 +455,15 @@
...
 	return F
+
+/// check if the mob has high enough brain damage that they can't use this
+/obj/proc/brain_dmg_check(mob/user)
+	if (src.usable_with_brain_damage)
+		return
+	if (!ishuman(user))
+		return
+	var/mob/living/carbon/human/H = user
+	if (!H.traitHolder.getTrait("training_medical") && H.get_brain_damage() >= BRAIN_DAMAGE_MAJOR)
```
JORJ949, line 466: Not sure medical should be immune to the biggest downside of brain damage, they already have access to plenty of mannitol from nanomeds to heal themselves.

code/obj.dm
```diff
@@ -453,3 +455,15 @@
 		src.was_deconstructed_to_frame(user)
 		F.RegisterSignal(src, COMSIG_ATOM_ENTERED, TYPE_PROC_REF(/obj/item/electronics/frame, kickout))
 	return F
+
+/// check if the mob has high enough brain damage that they can't use this
```
JORJ949, line 459: Comment should be a bit clearer, as could the proc name, something like "Is the user prevented from using this object due to their brain damage, returns FALSE if user can use object.

Though I would change it to return TRUE if you can use it and the proc be "can_use_brain_damage_check", makes more sense than "they can't not use this" and is clearer in proc name what the expected return is

code/obj.dm
```diff
@@ -453,3 +455,15 @@
...
+/obj/proc/brain_dmg_check(mob/user)
+	if (src.usable_with_brain_damage)
+		return
```
JORJ949, line 462: Should return FALSE, functionally the same but more readable

pgmzeta:
Feels like it should be a number after which using it is impossible, since different machinery has different complexity?

Maybe enough brain-damage should flag illiterate instead, if it's going to be binary. Hm.

PR #23935, opened 2025-06-24
Fix Packet Sender Program being out of range

### LINKED ISSUE #9619: Packets sent out via buttons in a PDA program are not detected by the packet sniffer
# Labels

[BUG]

# Description

Duplicate of #9498, except more accurate:
Packets sent via a PDA program utilising packets are not detected by the packet sniffer, but still send and do what they're supposed to do

# Steps to reproduce

1. Using the Packet Sender program, write a packet, preferably one that perhaps opens a door nearby, then switch the frequency to the door's

2. Open the program builder and save the packet as a button, then save the button as a program

3. Grab a fellow mech's PDA and start sniffing the door frequency 

4. Open the program and click the button on your PDA

You'll find that no door opening packet is sniffed, however the door shall open and the ping from the door will be sniffed.

# Expected Behavior

For packets sent via a program to be detected by the packet sniffer

### LINKED ISSUE #20436: Packet : Auth Armory didnt work when use it as Program file
### Labels

[BUG]

### Description

The Auth armory didnt work when use it as program file

### Steps to reproduce

1. Make a packet to Auth/Deaauth armory 
2. use it ( to test is it functional) 
3. save it as programe file 
4. use it as a button from program file
5. Fail 
6. ( use sniffler to check it said out of range even i still stand the same tile as the previous one ) 

### Expected Behavior

can use it as a save programe file so i can do it easily

_No response_

JORJ949:
## About the PR 
Change's the source of packets sent via built program to be their master PDA, and not the program's datum.

## Why's this needed? 
Stops all packets from built programs being "out of range"
Fixes #20436
Fixes #9619

## Testing 
Here you can see me successfully using a program to authorise and unauthorise the armory, and the packet sent by the program being sniffable alongside the armory's response packet.
image

## Changelog 

```changelog
(u)JORJ949
(+)Fixed packet sender programs always being out of range.
```

### DIFF
--- a/code/obj/item/device/pda2/diagnostics.dm
+++ b/code/obj/item/device/pda2/diagnostics.dm
@@ -630,7 +630,7 @@
 		else if (href_list["button"])
 			SPAWN( 0 )
 				var/datum/signal/signal = get_free_signal()
-				signal.source = src
+				signal.source = src.master
 
 				var/list/buttonc = buttons[href_list["code"]] // get a list of "key list, value list" using the buttons name as index
 


FronkyGeniuZ:
Heck yeah program file will be rise again for Packet nerd

PR #23946, opened 2025-06-25
Fluid piping 2, lets do this again

Laboredih123:
[INPUT WANTED][FEATURE][SCIENCE]

## About the PR 
thanks flaborized for the good looking sprites like the pipes, i made the codersprites
adds fluid pipes and machinery and whatnot :)
oh yeah stuff will definitely change as i get feedback
slightly touches reagent code to fix some bugs, like remove_any_to preserves temperature now, and has an option to not cause reactions

machines cant connect to machines directly, they require pipes in between.
chemical reactions dont happen inside pipes

## Why's this needed? 
wow its super coool yeahhhh

## Testing 
image

## Changelog 

```changelog
(u)cringe, the good looking sprites from Flaborized
(*)Adds neat fluid pipes, now in the HPD.
```


Laboredih123:
uhhh tgui rebuild doesnt change anything

Laboredih123:
woe, lint fail doesnt stop

Laboredih123:
wrok!!!

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+/obj/machinery/fluid_machinery/unary/drain/passive/process()
+	src.drain()
+
+/obj/machinery/fluid_machinery/unary/drain/passive/big
+	name = "Passive drain"
+	icon_state = "drainbig"
+	drain_min = 6
+	drain_max = 14
+
+/obj/machinery/fluid_machinery/unary/drain/inlet_pump
```
TobleroneSwordfish, line 113: What's the difference between a fluid pump and a drain, can these be combined?

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+	var/drain_min = 0
+	var/drain_max = 0
+
+/obj/machinery/fluid_machinery/unary/drain/proc/drain()
+	if(!src.network)
+		return
+	var/turf/simulated/T = get_turf(src)
+	if(T.active_liquid?.group?.amt_per_tile)
+		//pick a random unit between drain_min and drain_max to drain, if there is less fluid then what we chose, drain it all.
+		var/amount = min(rand(src.drain_min, src.drain_max), src.network.reagents.maximum_volume - src.network.reagents.total_volume)/T.active_liquid.group.amt_per_tile
```
TobleroneSwordfish, line 108: This setup produces immediate rounding errors:
image

image

Could probably be resolved by rounding `amount` here?

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+	var/pullrate = 50
+
+/obj/machinery/fluid_machinery/unary/nullifier/process()
+	if(!src.network) return
+	var/datum/reagents/fluid = src.pull_from_network(src.network, src.pullrate)
+	src.use_power(100 WATTS * fluid?.total_volume, ENVIRON)
+	qdel(fluid)
+
+/obj/machinery/fluid_machinery/unary/input
+	name = "Port"
```
TobleroneSwordfish: A lot of these machines are capitalized when they shouldn't be, leading to
image

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+	SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_CONFIG,"Set Dispense Type", PROC_REF(set_type))
+	SEND_SIGNAL(src,COMSIG_MECHCOMP_ADD_CONFIG,"Set Automatic dispensing", PROC_REF(set_automatic))
+
+/obj/machinery/fluid_machinery/unary/dispenser/get_desc()
+		. += "<br>[SPAN_NOTICE("Automatic: [src.automatic ? "true" : "false"]. Dispense Amount: [src.amount]. Dispensing: [src.itemtodispense]")]"
+
+/obj/machinery/fluid_machinery/unary/dispenser/process()
+	if (!src.network) return
+	var/datum/reagents/fluid = src.pull_from_network(src.network, src.max)
+	fluid.trans_to(src, src.max)
```
TobleroneSwordfish: Runtime just from placing a dispenser here:
image

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+	invisibility = INVIS_ALWAYS
+
+/obj/machinery/fluid_machinery/unary/node/ex_act()
+	return
+
+/obj/machinery/fluid_machinery/unary/node/meteorhit()
+	return
+
+/obj/machinery/fluid_machinery/unary/node/updateHealth()
+	return
```
TobleroneSwordfish, line 416: ?

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+/// Accepts an input network and an ideal amount of fluid to pull from network.
+/// Returns a reagents datum containing a scaled amount of fluid linear to fullness of network or null if no fluid in network. Quantized to QUANTIZATION_UNITS units.
+/obj/machinery/fluid_machinery/proc/pull_from_network(datum/flow_network/network, maximum = 100)
+	return network.reagents.remove_any_to(max(MINIMUM_REAGENT_MOVED, round(maximum * (network.reagents.total_volume / network.reagents.maximum_volume), QUANTIZATION_UNITS)), TRUE)
+
+/// Accepts an input network and the reagents datum to add to the network.
+/// Returns TRUE on complete addition to network and deletion of reagents datum. Returns FALSE if reagents remaining and reagents not deleted.
+/obj/machinery/fluid_machinery/proc/push_to_network(datum/flow_network/network, datum/reagents/topush)
+	topush?.trans_to(network, topush.total_volume, 1, FALSE)
+	if(topush.total_volume)
```
TobleroneSwordfish, line 23: Another runtime here: pump on an empty network
image

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+	anchored = ANCHORED
+	plane = PLANE_FLOOR
+	/// What directions are valid for connections.
+	var/initialize_directions
+	var/exclusionary = FALSE
+
+/// Accepts an input network and an ideal amount of fluid to pull from network.
+/// Returns a reagents datum containing a scaled amount of fluid linear to fullness of network or null if no fluid in network. Quantized to QUANTIZATION_UNITS units.
+/obj/machinery/fluid_machinery/proc/pull_from_network(datum/flow_network/network, maximum = 100)
+	return network.reagents.remove_any_to(max(MINIMUM_REAGENT_MOVED, round(maximum * (network.reagents.total_volume / network.reagents.maximum_volume), QUANTIZATION_UNITS)), TRUE)
```
TobleroneSwordfish: This feels like it should maaybe have some more significant constant term to it to make it less horribly slow at low amounts. Or at least fluid pumps should, since they are actively sucking fluid from the network.

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+			break
+
+/obj/machinery/fluid_machinery/trinary/refresh_network(datum/flow_network/network)
+	src.network1?.machines -= src
+	src.network2?.machines -= src
+	src.network1 = null
+	src.network2 = null
+	src.initialize()
+
+/obj/machinery/fluid_machinery/trinary/filter
```
TobleroneSwordfish, line 595: I feel like this is either broken or I'm misunderstanding what it's supposed to do? I put a beaker of melon juice in this filter and watched it neatly filter the input melon juice into both the outlet sides(???)
image
Oil put in the same system then went straight through, so it *is* technically filtering, just in a very confusing and backwards way.

code/obj/item/handheld_dispenser/dispenser.dm
```diff
@@ -491,6 +533,168 @@ ABSTRACT_TYPE(/datum/pipe_recipe/machine/trinary)
...
+		name = "Hand pump"
+		path = /obj/machinery/fluid_machinery/unary/hand_pump
+		icon_state = "handpump"
+		desc = "Pumps out up to 100 units of fluid manually."
+
+	nullifier
+		name = "Nullifier"
+		path = /obj/machinery/fluid_machinery/unary/nullifier
+		icon_state = "nullifier"
+		desc = "Disappears up to 50 unit."
```
TobleroneSwordfish: This should probably just match the help text, much better wording

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+	name = "Reagent Filter"
+	desc = "Filters out a specific reagent."
+	HELP_MESSAGE_OVERRIDE("Can be loaded with a <b>beaker</b>, which must contain at least 1 unit of a reagent. The most plentiful reagent is chosen for filtering.")
+	icon_state = "filter0"
+	flags = NOSPLASH
+	var/pullrate = 200
+	var/obj/item/reagent_containers/glass/beaker
+
+/obj/machinery/fluid_machinery/trinary/filter/attackby(obj/item/reagent_containers/glass/B, mob/user)
```
TobleroneSwordfish, line 604: Needs a swap mechanic for when it already has a beaker, currently just deletes the one that's in there.

code/modules/fluids/fluid_pipe_machinery.dm
```diff
@@ -0,0 +1,521 @@
...
+	var/itemtodispense = "pills"
+	var/static/list/itemlist = list("patches", "pills", "vials")
+
+/obj/machinery/fluid_machinery/unary/dispenser/proc/dispense()
+	if (src.reagents.total_volume < src.amount)
+		return
+
+	switch(src.itemtodispense)
+		if("pills")
+			var/obj/item/reagent_containers/pill/P = new(get_turf(src))
```
TobleroneSwordfish, line 345: These need to be coloured, see the chemmaster code

TobleroneSwordfish (changes requested):
This review is just a first batch of issues I've noticed from playing around with these in a local, this will definitely need testmerging and probably a performance pass before merging.

One general thing that I'm a little disappointed on is there doesn't really seem to be a failure state with these pipes? I know they're modelled on Factorio but unlike them we *do* actually have fluid puddles that disconnected pipes could leak to 🥺 
image
## Some other misc issues
I assume you know about this but W I D E:
image

UI looks a little cramped with four tabs now:
image
I think it would be fine to stretch it out to where they all fit on one line:
image
Especially since some of the names are quite long anyway.

None of the tanks/pipes etc work with the ` -> Add Reagent menu, they should.

Deleting pipes does not recalculate the network properly, presumably this also goes for exploded pipes. You may want to build in a mechanism to recalculate only at the end of the tick otherwise large explosions will trigger an expensive rebuild.

code/modules/fluids/fluid_pipe_machinery.dm
Laboredih123: invisible object used for connecting non fluid stuff to fluid stuff. that means it shouldnt shatter unless the parent object shatters

Sovexe:
image

TobleroneSwordfish:
See bug report above, also admin check reagents still doesn't seem to work with the pipes.

Laboredih123:
fixed

pgmzeta:
!merge_upstream

pgmzeta:
!merge_upstream

PR #23953, opened 2025-06-25
Refactor plantpot harvesting

Egregorious:
## About the PR 

Refactor `HYPharvesting` into modular helper methods with a central harvesting datum. Added documentation throughout.

## Why's this needed? 

This is more easily extendable, and hopefully more readable. 

## Testing 

Tested quite a few harvests across a lot of different seeds while working on a subsequent thing, did not notice any wierdness or new bugs.


trazodont:
btw plant pots have this weird bug if you dig something from a tray and replant in that tray before putting whatever you dug into a pot plants what you replanted in the tray, does this fix that?

Egregorious:
I wouldn't imagine so since this only covers harvesting specifically, but I can maybe take a look at that later.

PR #23956, opened 2025-06-26
Grow your own guns

Egregorious:
## About the PR 

Added a gene strain which causes plants to instead grow guns in the shape of the plant's usual produce.
- These produce guns scale based on the plant's stats; potency affects damage, yield affects ammo quantity, endurance affects armour pierce, production rate and maturation rate both affect firing speed, lifespan affects recoil. 
- They cannot be reloaded.
- They can fit in produce satchels (probably a bad idea)
- They have generic gun inhand sprites if the base produce doesn't have any.
- Jumbo guns get a slight stat boost and punchier audio and visual effect.
- Damage, armour pierce and ammo count scale linearly (probably should be diminishing returns).

Added a chem which can infuse the gene strain into seeds, called "Phytoballistic Infusate". Added a recipe to make said chem. 

Added a prerequisite chem which can be obtained by putting guns in a still. 

The gene strain only affects the produce of plants which grow items, plants which grow objects or otherwise do something else harvest crops as normal. It also reduces the plant's produce count by 85%, but due to how yield works this is rather easy to circumvent. 

This relies on #23953.

TODO

- [ ] Balance via feedback

## Why's this needed? 

Botany doesn't have any whacky "secret"-level chem stuff that affects plants, despite there being a heavy mechanical relationship between chems and plants.

This is intended to be a rarity, as such I've tried to make it require effort to obtain, but the recipe could be a lot more difficult. Traitor botanists are already rather full on specialist items, so it doesn't seem right to put it there. If nothing else, it could be fun as an admin spawn. 

I also don't know how to balance guns, so the numbers are likely very iffy. The intention is that the guns themselves also require a fair bit of stat scaling - and thus time investment - before they become useful. They should be balanced with being dual-wielded in mind. 

## Testing 

Tested most of the different types of produce, including the more complex ones like glowsticks and ore. Tested the cat to make sure mob plants weren't affected. Made the chem in a barrel to make sure the recipe worked. 

produce gun demo 1
produce gun demo 2
Produce gun demo 3

## Changelog 

```changelog
(u)Egregorious
(*)Botany can now grow guns, if they have the know-how.
```

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -4029,6 +4029,25 @@
 		result_amount = 3
 		mix_phrase = "The mixture gives off a harsh odor."
 
+	gun_infusion
+		name = "Phytoballistic Infusate"
+		id = "gun_infusion"
+		result = "gun_infusion"
+		/*
+		The recipe is meant to encourage quintissential botany practices
+		- Botany's association with chemicals can best be described as "mass production", hence the quantities.
+		- Blackpowder is both a bullet chemical and a chem that is historically mass produced with the help of ingredients from botany.
+		- CLF3 is another weaponisable botany chem, but it also adds splice-requirements to the recipe.
+		- Plasma is the sci-fi space-magic chem that makes any of this even quasi-believable.
+		- Guns are made of metal.
+		- Gun distillate adds the still as part of the process.
+		- Temperature basically forces the use of Cryostylane, since this can only be made in something like a chem barrel.
+		*/
+		required_reagents = list("blackpowder" = 100, "infernite" = 100, "plasma" = 10, "iron" = 10, "gun_distillate" = 1)
+		max_temperature = 250
+		result_amount = 1
+		mix_phrase = "The mixture crystalises and then collapses in on itself, leaving behind small amount of gunmetal-grey sludge."
+
 	copper_nitrate
 		name = "Copper Nitrate"
 		id = "copper_nitrate"
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -4036,14 +4036,14 @@
 		/*
 		The recipe is meant to encourage quintissential botany practices
 		- Botany's association with chemicals can best be described as "mass production", hence the quantities.
-		- Blackpowder is both a bullet chemical and a chem that is historically mass produced with the help of ingredients from botany.
-		- CLF3 is another weaponisable botany chem, but it also adds splice-requirements to the recipe.
+		- Blackpowder is both a bullet chemical and a chem that is traditionally mass produced with the help of ingredients sourced from botany.
+		- Booster enzyme adds splice-requirements to the recipe, and acts as the 'organic' interface.
 		- Plasma is the sci-fi space-magic chem that makes any of this even quasi-believable.
-		- Guns are made of metal.
+		- Guns are made of metal, there should be metal in there somewhere.
 		- Gun distillate adds the still as part of the process.
 		- Temperature basically forces the use of Cryostylane, since this can only be made in something like a chem barrel.
 		*/
-		required_reagents = list("blackpowder" = 100, "infernite" = 100, "plasma" = 10, "iron" = 10, "gun_distillate" = 1)
+		required_reagents = list("blackpowder" = 200, "booster_enzyme" = 100, "plasma" = 10, "iron" = 10, "gun_distillate" = 1)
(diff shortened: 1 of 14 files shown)


trazodont:
Holy moly my pea shooter idea real

JORJ949:
I don't think this fits being a mutation, being able to grow infinite lethal guns that don't look like guns and are at least in the video, really good (crit in 4 shots, there also seems to be a minigun one) as nonantag is going to end incredibly poorly. New security will just see "Oh thats money/paper/an apple" when searching people and this is bound to invite powergaming/validhunting with them. I think this fits better as its own plant, purchasable by syndicate botanists (yes even when they already have a bunch of unique items) with a sprite that looks very gunlike (see TG's gatfruit).

TLDR: Infinite guns grown by nonantags will make security and admins suffer.

Scaltra-Volpe:
I can not imagine this being balanced and of all departments botany really does not need it. Although i would like to see more cool chem plant stuff I would like to see the current stuff evolved on first that are borderline useless like mutadone. 

Maybe the chemical could be reused to be a traitor scientist or mime thing that would let you turn anything it is splashed on into a gun with its stats being based off the amount splashed. Or just some sort of really convoluted secret chemical that does that. 

I think it also being a gene strain is interesting and its silly seeing all the things fire but it should be obvipus what is being grown is a gun or harmful to be honest. 

If anything I would prefer if the plant itself went rogue and started shooting literally everything and was a rare strange seed. It definitely shouldn't be plantable more than once if anything like this existed. 

It is really cool and I love it but I think it would need a lot to fit

Egregorious:
The recipe does not need to be exceptionally accessible to non-antags. It already requires access to guns for a prerequisite chem, and it could be rebalanced by increasing the number of guns required, or removing the likes of the staple gun from producing the chem, or making antag-specific guns brew more prereq chem and all other guns less-so. 

The later examples have plant stats that are virtually impossible to obtain in any given round just due to the sheer amount of effort needed to infuse them, they're mainly just a showcase of the current minimum and maximum scaling. The scaling can be made logarithmic instead of linear to give it soft-caps. 

It's very easy to make the plant unable to reproduce once infused, and also force all plants to die after a single harvest. It could also induce a mutation rather than use the base crop, but I am terrible at art and I don't think I could make a good-looking gun/plant sprite for that. Either way, it does not need to be infinite. 

The intent is that it would require enough effort to be more of a pinnacle goof for botany nerds to show off with rather than anything that would crop up with enough frequency to affect new-player retention. There's enough weaponry in the game only locked behind effort that I thought this had a chance to not break everything if properly gatekept by difficulty.

RSOD404:
i feel like i'd rather have this chem / strain automatically mutate any plant into a specific plant that makes guns, that way you don't end up with a  dollar bill that is actually a gun and kills the unsuspecting secoff and instead something properly gun-shaped

another possible idea to even out the effort to scaling ratio would be growing the gun and ammo separately, where some stats like lifespan makes the gun last for more shots before breaking apart and growth rate makes it fire faster at the cost of higher spread (effectively, lets you grow different kinds of guns - which i'd say would be fun to see)

LorrMaster:
Maybe instead of plant guns they are plant bullets. They could then be scaled down or all share a sprite that is given the plant's color.

Egregorious:
> Maybe instead of plant guns they are plant bullets. They could then be scaled down or all share a sprite that is given the plant's color.

That's a good shout, that might work better. There's far fewer stats to work with on bullets alone, but it looks like there might be enough to still be interesting.

A dedicated plantgun/gunplant is maybe ideal judging from reactions, but unless there's someone who wants to make the sprites it's not really in the scope of what I can manage.

M-Earthfire:
i agree with the other peoples sentiment. Although i strongly advocate for non-antags getting more lethal weaponry, straight up growing guns is a bit much. And it would fit much better as its own plant, like the gatfruit on other servers.

ribbiestar:
image

Scaltra-Volpe:
I think a plant that flat out grows ammo would be silly we basically already have that with potato batteries if it's an energy gun. Could also be used on its own with the current heat to ammo system causing ammo to go off or with the use of a zip gun. Could create already existing ammo depending on stats.

TobleroneSwordfish:
There's a seed of a cool idea here but as presented it definitely doesn't fit.
The basic premise of being able to grow guns is pretty sketchy, maybe growing ammo or plants that shoot seeds at people would work better? Also the idea of this being a mutation you can apply to any plant type to turn it into a generic gun that doesn't look like one and shoots normal bullets just seems.. off?

Egregorious:
Refinement is in progress.

PR #23960, opened 2025-06-26
Charcoal no longer heals toxin damage

pgmzeta:
[medical][balance]
## About the PR 
* Remove charcoal's toxin healing, and tweak the reagent description to match its usage.
* Add some potassium-iodine to VR antitoxin pills to help counteract toxin damage, since it's less likely they will have access to medical
* Change sleepers to inject potassium-iodine instead of charcoal when toxin damage is detected

## Why's this needed? 
Charcoal is very strong, countering all the effects of most basic poisons.

Makes medical a little more complex, forcing them to treat the symptom, or the cause, but a little harder to treat both at the same time. The dialysis machine becomes much more useful as a means of purging during treatment as well, making its usage and loss felt much more directly.

Medical has an endless supply of potassium-iodine via dispensers for toxin damage, so they don't lose any healing capabilities.

## Testing 
Reviewed uses of charcoal and ensured that things that expect to cure toxin damage

## Changelog 

```changelog
(u)glowbold
(*)Charcoal does not heal toxin damage, but purges chemicals from the bloodstream and heals organs.
```

### DIFF
--- a/code/modules/chemistry/Reagents-Medical.dm
+++ b/code/modules/chemistry/Reagents-Medical.dm
@@ -1560,7 +1560,7 @@ datum
 		medical/charcoal
 			name = "charcoal"
 			id = "charcoal"
-			description = "Activated charcoal helps to absorb toxins."
+			description = "Activated charcoal helps to neutralize chemicals in the bloodstream."
 			reagent_state = SOLID
 			fluid_r = 0
 			fluid_b = 0
@@ -1572,7 +1572,6 @@ datum
 				if(!M) M = holder.my_atom
 				if(prob(50))
 					flush(holder, 1 * mult)
-				M.HealDamage("All", 0, 0, 1.5 * mult)
 
 				if (ishuman(M))
 					var/mob/living/carbon/human/H = M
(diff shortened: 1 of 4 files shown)


JORJ949:
I think charcoal has solidified itself as the "this heals toxin damage" chemical tbh, so I'd rather nerf the purgative qualities of charcoal and make calomel the purgative, I don't personally use potassium as I just think "thats the rads chem i use that for rads". I think nerfing charcoal is good but its currently the big "heals toxin" chem that would require  a lot of changes (e.g. the big charcoal beakers in medbay should probably be replaced with potassium iodide )where its just easier to make the purgative effect weaker or non-existent. B

RSOD404:
i'd rather charcoal be nerfed than straight out reworked, considering it's already been kinda set in stone that charcoal is THE "use this when healing TOX" medical chem, outright changing that feels counterintuitive

CalliopeSoups:
I am not against nerfing charcoal, it's the main thing you pump into people by itself for most toxin damage. I do think however nerfing it in this way will turn charcoal from one of the most used chemicals to one of the least used, just to be replaced with potassium iodide which would be a similar thing to charcoal but a little weaker.

Personally I'd suggest either making it so you need something other than charcoal for toxin damage over a certain level or add different types of symptoms to come with toxin that you'd need to use other medications for.

PR #23962, opened 2025-06-26
Adds "Defect Prone" trait

snugglycactus:
## About the PR 
This pull request would add the "Defect Prone" trait, a 1 point trait which both allows major cloning defects to be rolled upon the first clone generation, and guarantees a cloning defect to be added in addition to the one already being rolled per cloning.

## Why's this needed? 
This is essentially a "puritan-lite" trait, providing a substitute of sorts to the all but removed puricloning function. It'll add a more fun and fast-paced aspect to cloning, ideally making it more fun for both the person being cloned and the medical staff.

## Testing 
Screenshot 2025-06-25 164506
The trait in the menu. Part of the "cloner" invisible category, and as such is incompatible with Puritan.

Screenshot 2025-06-26 175704
This screenshot features someone with the trait upon being cloned. As indicated by the clone generation being 1, they have received a severe cloning defect (missing liver) upon their first cloning. Quick, someone heal them!

Screenshot 2025-06-26 180159
On clone generation 2, it's even more of a mess. Someone *really* needs to heal them now.

Screenshot 2025-06-26 180456
But wait, are other things functional still? Here is a normal person, without the trait.

Screenshot 2025-06-26 181202
Here is the same normal person, upon being cloned once. No cloner defects. Because this person does not have the cloner defect trait, they are wholly unaffected.

## Changelog 

```changelog
(u)SnugglyCactus
(+)Added trait "Defect Prone"; cloning's going to get a bit weirder.
```


code/obj/machinery/clonepod.dm
```diff
@@ -283,10 +283,15 @@ TYPEINFO(/obj/machinery/clonepod)
 			for (var/i in 1 to rand(round(generation / 2)  * (src.emagged ? 2 : 1), (generation * (src.emagged ? 2 : 1))))
 				if (generation)
 					defects.add_random_cloner_defect()
+				else if (src.occupant.traitHolder.hasTrait("defect_prone"))
+					defects.add_random_cloner_defect()
```
frawhst:
```suggestion
				if (generation || src.occupant.traitHolder.hasTrait("defect_prone"))
					defects.add_random_cloner_defect()
```
Slightly cleaner this way, imo.

snugglycactus:
> lgtm other than the below!

Got that committed; thanks for making that so easy!

code/obj/machinery/clonepod.dm
```diff
@@ -281,12 +281,15 @@ TYPEINFO(/obj/machinery/clonepod)
 			 */
 			var/generation = src.occupant.bioHolder.clone_generation
 			for (var/i in 1 to rand(round(generation / 2)  * (src.emagged ? 2 : 1), (generation * (src.emagged ? 2 : 1))))
-				if (generation)
+				if (generation || traits.hasTrait("defect_prone"))
```
Mister-Moriarty:
```suggestion
				if ((generation > 1) || traits.hasTrait("defect_prone"))
```
Per [a discussion on Discord](https://discord.com/channels/182249960895545344/890226559691157524/1399827450728550461), we've discovered that `generation` here is always 1 or greater, so `if (generation)` will always run.

SunWuDrakios:
I really like the idea, would love this to be ingame

PR #23963, opened 2025-06-26
Add Project Voice gene

FlameArrow57:
[MEDICAL][FEATURE]

## About the PR 
Adds a new Project Voice gene, which allows you to make targeted objects speak with a 10 second cooldown

## Why's this needed? 
For fun

## Testing 
Did basic testing using the gene on objects and mobs

## Changelog 

```changelog
(u)FlameArrow57
(+)Added a new Project Voice gene
```

### DIFF
--- a/code/modules/medical/genetics/bioEffects/powers.dm
+++ b/code/modules/medical/genetics/bioEffects/powers.dm
@@ -2361,6 +2361,35 @@ ABSTRACT_TYPE(/datum/bioEffect/power)
 						if (thrown_limb)
 							thrown_limb.throwforce = tmp_force
 
+/datum/bioEffect/power/project_voice
+	name = "Project Voice"
+	desc = "Allows the subject to create voice resembling sound waves at a target location."
+	id = "project_voice"
+	msgGain = "You feel in tune with sound."
+	msgLose = "You feel out of tune with sound."
+	cooldown = 10 SECONDS
+	ability_path = /datum/targetable/geneticsAbility/ventriloquism
+
+/datum/targetable/geneticsAbility/ventriloquism
+	name = "Project Voice"
+	desc = "Project a voice at a target location."
+	icon_state = "speech"
+
+	cast(atom/target)
+		if (..())
+			return TRUE
+		if (ismob(target))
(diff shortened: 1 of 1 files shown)


RSOD404:
sounds fun, but i feel like this should only be able to be acquired via a gene combination or at least have it be on the rarer side if possible. it'll only be cool / scary if it happens rarely, not when John Geneticist puts it up on the booth for the 5th round this day and everyone's just gotten numb to their toolbox talking to them

wolfolotl:
could be neat as something you need to fuse two genes to get... projectile vomiting and the constant shouting gene, maybe?

Sord213:
Seems neat. Couple things, I don't know how admin logging works for gene power usage, so please make sure it logs properly on both the caster and target. 

Then maybe if it's not synchronized make it so the caster performs an emote when casting. "John Pubbie mutters under his breath." or something like that

RSOD404:
if it were to be a gene combination, i feel something like telekinesis / empathic thought + telepathy could work?

code/modules/medical/genetics/bioEffects/powers.dm
```diff
@@ -2361,6 +2361,35 @@ ABSTRACT_TYPE(/datum/bioEffect/power)
...
+			boutput(src.owner, SPAN_ALERT("You can't target living creatures!"))
+			return TRUE
+		if (GET_DIST(src.owner, target) > 7)
+			boutput(src.owner, SPAN_ALERT("That is too far away!"))
+			return TRUE
+
+		var/to_say = tgui_input_text(src.owner, "What would you like this thing to say?", "Project Voice")
+		if (!to_say)
+			return TRUE
+		target.say(to_say)
```
Mister-Moriarty, line 2395: This will result in the target speaking, rather than the owner's speech appearing to originate from the target.
If you want the latter behaviour, try: `src.owner.say(to_say, message_params = list("message_origin" = target))`

FlameArrow57: intended but thanks

ZeWaka: hmm worth thinking about - not sure what this would be

PR #23965, opened 2025-06-27
Organ Damage / Poisons Balancing

pgmzeta:
[medical][balance]
## About the PR 
This is an alternative nerf for charcoal that maintains its tox damage healing and chemical purging effects by removing its third, lesser known application: healing organs.

Generally speaking, when posions would deal toxin damage, they will also apply tox damage to organs specific to the poison. Mimics real-life effects of these poisons on the body, where applicable, and uses cues from flavor text on fantasy poisons. Note that these are not guaranteed - they typically require a probability, specific dosage, or long-term effect on the person before applying. Poisons that damage the eyes/brain have not been altered, as those effects on organs are already handled in the relevant procs if necessary.

* Removes charcoal's organ healing abilities
* Add targeted organ damage to the following:
  * martian flesh - stomach, intestines
  * viscerite viscera - stomach, intestines
  * atrazine (weedkiller) - pancreas
    * Also damages all synthorgans
  * amanitin - liver, kidneys
  * coniine - lungs
  * lipolicide - stomach
  * initropidril - heart
  * cholestorol - spleen, heart
    * has a chance of applying bloodclot malady in addition to heartdisease
  * flatline malady will damage the heart
* Allows high levels of atrazine (weedkiller) splashed on someone to melt synthlimbs
* Curare now applies paralysis at its end state
## Why's this needed? 
Nerf for charcoal that maintains its toxin/flushing abilities. Makes managing organ damage harder, lending to more organ replacements as organ damage builds up.

## Changelog 

```changelog
(u)glowbold
(*)Charcoal does not heal organ damage.
(*)Most posions will directly damage your organs.
```

### DIFF
--- a/code/modules/medical/diseases/malady.dm
+++ b/code/modules/medical/diseases/malady.dm
@@ -278,6 +278,8 @@
 				if (probmult(5))
 					affected_mob.take_oxygen_deprivation(rand(1,2))
 				if (probmult(5))
+					affected_mob?.organHolder.damage_organ(tox=2*mult, organ="heart")
+				if (probmult(5))
 					affected_mob.emote(pick("twitch", "groan", "gasp"))
 				if (probmult(1))
 					affected_mob.contract_disease(/datum/ailment/malady/heartfailure,null,null,1)


frawhst:
Can't believe a fish exploded and made the build fail

code/modules/chemistry/Reagents-Misc.dm
```diff
@@ -3141,6 +3158,16 @@ datum
...
+				if (ishuman(M) && volume > 25)
+					var/mob/living/carbon/human/H = M
+					if (H.limbs?.l_arm?.kind_of_limb & LIMB_PLANT)
+						M.TakeDamage("l_arm", tox=min(rand(20, 40), volume), damage_type=DAMAGE_BURN, disallow_limb_loss=FALSE)
+					if (H.limbs?.r_arm?.kind_of_limb & LIMB_PLANT)
+						M.TakeDamage("r_arm", tox=min(rand(20, 40), volume), damage_type=DAMAGE_BURN, disallow_limb_loss=FALSE)
+					if (H.limbs?.l_leg?.kind_of_limb & LIMB_PLANT)
+						M.TakeDamage("l_leg", tox=min(rand(20, 40), volume), damage_type=DAMAGE_BURN, disallow_limb_loss=FALSE)
+					if (H.limbs?.r_leg?.kind_of_limb & LIMB_PLANT)
+						M.TakeDamage("r_leg", tox=min(rand(20, 40), volume), damage_type=DAMAGE_BURN, disallow_limb_loss=FALSE)
```
frawhst: Feels like `disallow_limb_loss` should have a default rather than being a "required" arg. `FALSE` makes sense to me (since by default heavy damage should be able to knock limbs off), and that way you can remove the arg from all these calls.

code/modules/medical/diseases/malady.dm
```diff
@@ -564,3 +564,4 @@
 		H.changeStatus("knockdown", 6 * mult SECONDS)
 		H.losebreath+=20 * mult
 		H.take_oxygen_deprivation(20 * mult)
+		H.organHolder?.damage_organ(tox=mult, organ="heart")
```
frawhst:
```suggestion
		H.organHolder?.damage_organ(tox=1 * mult, organ="heart")
```
Debatable, but I prefer `1 * mult` to `mult` in cases where the number being mult-scaled is `1`. Technically unnecessary, but it keeps to the pattern of `n * mult` and makes it a bit more readable imo. Up to you.

PR #23977, opened 2025-06-28
Fix(?) phones picking up radio chatter

### LINKED ISSUE #23976: Phones appear to be picking up more than they should
### Labels

_No response_

### Description

^

### Steps to reproduce

1. Have a conversation with two people over a phone line.
2. Have someone else speak over a radio channel one of the participants can hear.
3. Each person can hear all the other's radio messages as if the speakers were speaking directly into the phone.

### Expected Behavior

While overhearing radio conversations via phone is a neat idea, in practice it just becomes noise and should probably not happen. Phones should only pick up directly spoken messages.

Image
(This cyborg is speaking over general comms and is nowhere near the brig)

TobleroneSwordfish:
## About the PR 
Fixes #23976

## Why's this needed? 
I'm not sure if this was intentional, but it really doesn't feel intentional.
## Testing 
image
Clown not coming through on the phone.

### DIFF
--- a/code/obj/machinery/phone/handset.dm
+++ b/code/obj/machinery/phone/handset.dm
@@ -1,7 +1,7 @@
 TYPEINFO(/obj/item/phone_handset)
 	start_listen_effects = list(LISTEN_EFFECT_HANDSET)
 	start_listen_modifiers = list(LISTEN_MODIFIER_PHONE)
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
@Mister-Moriarty has promised to fix this better

PR #23982, opened 2025-06-29
Grow your own bullets

Egregorious:
## About the PR 

Added a gene strain which causes plants to instead grow ammo in the shape of the plant's usual produce.
- These produce bullets start very weak but scale based on the plant's stats; endurance affects damage, potency affects armour pierce, production rate and maturation rate both affect projectile speed, lifespan affects damage falloff.
- They are not fit for purpose and can only be loaded one-at-a-time. They also get stuck in the gun and can't be unloaded without being shot.
- They count as .22 ammo.
- They carry a reagent payload based on what the produce would normally have, which applies the on-touch effects when hitting a living target.
- Jumbo quality gets a damage bonus.
- They can fit in a produce satchel.

Added a chem which can infuse the gene strain into seeds, called "Phytoballistic Infusate".
- The gene strain causes the plant to die after a single harvest.
- The infusion also applies the 'seedless' gene strain.
- The chem recipe requires fermenting a gun as part of the process. 
- {5 blackpowder, 3 glowing slurry, 2 plasma, 1 copper, 1 gun distillate} = 1 Phytoballistic Infusate

Added two chems: "gun distillate" and "gun slop"
- distillate is required for infusate, and can be obtained by distilling a non-makeshift kinetic gun
- exceptions are the zip gun, slam gun, foam dart guns, airzooka and the paintball gun, which produce slop instead

TODO
- [ ] Feedback
- [ ] Balance
- [ ] Produce-type SFX

## Why's this needed? 

Botany doesn't have any whacky "secret"-level chem stuff that affects plants, despite there being a heavy mechanical relationship between the two.

This isn't intended to be particularly available to the crew, but I think it isn't something that wants to be hard-locked to a particular traitor role. Botany is somewhat unique in that the maps in which everyone doesn't have immediate access to functional hydroponic equipment are the exception rather than the rule. 

Guns are generally difficult to come across without traitorous actions or abilities, and thus the expectation is that this is largely only accessible by traitors that buy them, or else is a result of theft or space exploration.

## Testing 

Tested most of the different types of produce, including the more complex ones like glowsticks and ore. The only one that seems to act weird is the crystal seed, which comes out as regular shards instead of plasmashards. Tested the cat to make sure mob plants weren't affected. Made the chem in a barrel to make sure the recipe worked.

Scaling test with 0, 50, 100, and 300 in most stats:
produce_ammo_demo_1
Reagent test with lavendar, phlogiston, water, synthflesh. 
produce_ammo_reagent_demo
Unique on hit effects test with the banana:
banana hit-effects test

## Changelog 

```changelog
(u)Egregorious
(*)Botanists can now grow organic bullets, if they have the know-how.
```

### DIFF
--- a/code/obj/item/satchel.dm
+++ b/code/obj/item/satchel.dm
@@ -328,7 +328,7 @@
 			/obj/item/parts/human_parts/leg,
 			/obj/item/raw_material/cotton,
 			/obj/item/feather,
-      /obj/item/ammo/bullets/thingammo),
+      		/obj/item/ammo/bullets/thingammo,
 			/obj/item/bananapeel)
 			exceptions = list(/obj/item/plant/tumbling_creeper, /obj/item/reagent_containers/food/snacks/ingredient/egg)
 			// tumbling creeper have size restrictions and should not be carried in large amount


JORJ949:
I still think this would fit better as a seperate plant/mutation, peas and bananas (banana magazine) come to mind. It would be clearer that its ammo and not just a shrunk down version of the original produce. Guns and ammo, especially ones obtainable as a civilian, should be pretty clear what they are at a glance so newbee officers know to confiscate them.

Egregorious:
Here's an inventory comparison between the bullets and the base produce:
Screenshot 2025-06-29 160346
Screenshot 2025-06-29 160427

I don't think they're exceptionally stealthy to be honest, they require a lot of inventory space to actually shoot which makes them even more obvious, and they're fairly useless without also having a gun on hand anyway. 

I also think that this doesn't fill the same design space as a dedicated gunplant, personally.

cheekybrdy:
I think honestly these might be a bit underpowered compared to other ways to get reagents in people (raffles, nettles) seeing that your making alotta noise for 15 units when a nettle brushing someone can easily do the same with no mechanical requirements, no direct incrimination and no significant chem recipe, as broken as it would be having it scale with potency with the main recipe with the guard could be enough, just a random opinion tho.

Egregorious:
Actually the way it's set up right now the payload does scale with potency, but it doesn't inject at all. It's more just a secondary effect to it also being a bullet. None of the on-touch effects which plant reagents have scale with volume past ~15 or so I think, except water.

Although come to think of it, cyanide skin penetration might be the issue with that. I haven't looked at it but I'm assuming skin pen is handled by by touch application.

Scaltra-Volpe:
Honestly I really like the charm of this iteration a lot personally and think it seems pretty fair all things considered. As long as the recipe isn't easy to crank out barrels of.

I like that you can make healing bullets as well with things like synthflesh and low endurance. Speaking of, what happens when these stats are negitive on the plant? And have you maybe considered an in tray effect?

TobleroneSwordfish:
Ookay a few things:
- This feels like it has been balanced with a sledgehammer to try to get around the fact that growing ammo is an inherently crazy concept. I'm not saying this can't be balanced but requiring 200u BP + 100u booster enzyme for 3u of product isn't it. See also requiring >100 endurance before the bullets actually start dealing damage.
- I agree with Jorj that making this a specific mutation only available on certain plants would be better. This would also allow for unique effects per plant, like banana bullets leaving peels behind on impact etc. It might also require clever splicing to get the reagents you actually want in the projectile, which is relevant since the only effective way to use these things with the current balancing would appear to be stacking on-touch reagents on a 1 damage projectile.

Scaltra-Volpe:
> Ookay a few things:
> 
> * This feels like it has been balanced with a sledgehammer to try to get around the fact that growing ammo is an inherently crazy concept. I'm not saying this can't be balanced but requiring 200u BP + 100u booster enzyme for 3u of product isn't it. See also requiring >100 endurance before the bullets actually start dealing damage.


Holy crap okay, did not know what rhe recipe was. That is really restrictive for this especially. And honestly at all. You need 10units to even infuse.

Egregorious:
> This feels like it has been balanced with a sledgehammer

Let me address the recipe - and this should have been on the PR to start with but at the time of making it I guess I just wasn't thinking:

As it stands, for a one-person operation the whole process of making the chem and scaling the plant is meant to take a long time. However, each component can be accomplished at the same time by different individuals, and the reward caters to supplying more than one person. This way it's a multi-player challenge with a multi-player reward, which I think makes sense for a multiplayer game. 

The booster enzyme quantity is frankly meaningless past a certain point, if someone's properly grown the appropriate plant they can churn out thousands of units as easily as they can 10. The point is just that it's much, much more tedious to get the quantity via chemistry than it is via botany. All this part of the recipe really says is "grow and splice Blooming Lashers".

With the gunpowder I was spitballing. Though scaling it down to even 100 units then for 30 units of the infusion chem - which is 3 plants, and with current yield balance that's easily around 10 items each to 20 if other gene strains are involved - that's 2 nutrient crates, some Venne and access to a chem dispenser, which doesn't sound very difficult. 

I feel like the actual gatekeeper for non-antags is the 'gun distillate' chem. Civilians tend to have a hard time getting access to many guns without already doing antagonistic things or getting compliance from the chain of command. Easily accessible weapons like the zip gun can be excluded from producing the chem on the grounds that they're not proper guns, for dimensionality of balance. 

I think I balanced the current damage scaling around potency and then forgot when switching them around that endurance is slower, but this also sounds like just a numbers issue. It starts at such a low value pretty much exclusively because there's no way to scale stats negatively, and healing-bullets are kinda cool.


> making this a specific mutation only available on certain plants would be better

The reason I like this less is that it limits player creativity. You deny the player the choice of shooting carrots, or flowers, or fish, and the identity becomes less 'produce-bullet' and more 'bullet that happens to be sourced from a plant'. If the idea is that splicing would combine the sprites of other plants into the mutated plant, then that becomes more complicated to do, since you need to override the way splicing works, and you seem to wind up with the same design just more complex for little gain.

If the issue is the sprite of the ammo item itself, what I could do is give the ammo a generic sprite - perhaps slap the seed-packet leaf onto a flintlock ammo bag sprite. It wouldn't make sense for the current design of one-item-per-bullet-though, which isn't necessary but I think is kinda neat. 

> This would also allow for unique effects per plant

This can already be done.

> Speaking of, what happens when these stats are negitive on the plant

They scale backwards too. Negative armour piercing functionally increases the targets armour, the projectile can get slower or the damage dissipates quicker. The only reason the damage begins at 1 is because there's no way to properly scale plant stats negatively, otherwise I think it beginning with some damage makes more sense - but I wanted to enable the possibility of the likes of healing bullets. 

> And have you maybe considered an in tray effect?

I thought that maybe in the tray it could cause the plant to occasionally shoot at random, but I would want it to use the customisable bullets and I don't know how feasible that is.

Glamurio:
I think this is a really cool idea.

> > making this a specific mutation only available on certain plants would be better
> 
> The reason I like this less is that it limits player creativity. You deny the player the choice of shooting carrots, or flowers, or fish, and the identity becomes less 'produce-bullet' and more 'bullet that happens to be sourced from a plant'. If the idea is that splicing would combine the sprites of other plants into the mutated plant, then that becomes more complicated to do, since you need to override the way splicing works, and you seem to wind up with the same design just more complex for little gain.

I don't so much mind either design direction of specific crops vs. all crops, but I will say that realistically, some props lend themselves to more unique and interesting interactions than others. Having the system be generic can be good, but since you want specific interactions most likely anyway (like banana bullets + peels), you'll end up having to program unique cases either way, and then it's just kind of cleaner to do it case-by-case, and probably easier to balance too.

> As it stands, for a one-person operation the whole process of making the chem and scaling the plant is meant to take a long time. However, each component can be accomplished at the same time by different individuals, and the reward caters to supplying more than one person. This way it's a multi-player challenge with a multi-player reward, which I think makes sense for a multiplayer game.
> 
> The booster enzyme quantity is frankly meaningless past a certain point, if someone's properly grown the appropriate plant they can churn out thousands of units as easily as they can 10. The point is just that it's much, much more tedious to get the quantity via chemistry than it is via botany. All this part of the recipe really says is "grow and splice Blooming Lashers".
> 
> With the gunpowder I was spitballing. Though scaling it down to even 100 units then for 30 units of the infusion chem - which is 3 plants, and with current yield balance that's easily around 10 items each to 20 if other gene strains are involved - that's 2 nutrient crates, some Venne and access to a chem dispenser, which doesn't sound very difficult.

I believe I have a unique perspective on this as a layman in both botany and... guns. If you want this system to be used by anyone that's not a powergamer or someone who's extremely proficient in crop-mechanics, I think the thing needs to be a lot more accessible. I'd personally rather make it more widely accessible as an alternative (because I personally expect it won't get used as much), and then later nerf it if it turns out too strong than release a system that's so complex to get going most people won't bother.

> I feel like the actual gatekeeper for non-antags is the 'gun distillate' chem. Civilians tend to have a hard time getting access to many guns without already doing antagonistic things or getting compliance from the chain of command. Easily accessible weapons like the zip gun can be excluded from producing the chem on the grounds that they're not proper guns, for dimensionality of balance.

Civilians having a hard time getting access to guns is kind of part of the design philosophy. It also will make it more memorable and cool to actually get access to guns as a civilian as a result.

> and healing-bullets are kinda cool.

Healing bullets are very cool, and would turn this concept from wholly antagonistic to neutral, and allow for leeway in distributing guns with heal bullets (which could then be used for crime with different bullets if you have them)

Egregorious:
Thanks for the input! 

I have changed the recipe to be more accessible for one player. I still like the idea of a recipe that requires the barrel to produce, and it feels as if a botany-specific chem is the ideal candidate for that, but feedback seems to suggest this is not the appropriate combination. 

I agree that some crops lend themselves better to unique interactions. However at the same time if I were implementing them for more than one plant-type I would regardless probably want a more generic implementation between them, and at that point I see no reason for plants without unique properties to not also be available, just for the sake of players being able to work with different bullet cosmetics. 

I'm thinking I might try to do a completely generic 'bulletify item' mechanic first, and then utilise that as a base for this. I've been sitting on this for a while just to mull it over, but I still think it's a decent concept.

flappybatpal:
Raised this PR in the dev channel and there isn't support right now I'm afraid.

Botany is able to produce things in great quantities, guns and ammo have historically caused issues if they are widely afvailable and botany has a lot of potential griefing/traitor plants already. Nettles, rafflesias, radweed, seething tomatoes, etc

Egregorious:
Fair enough, I did imagine the concept would need a bit of defending. While I do think botany is unique amongst most jobs in that all its mechanics are very accessible for every antag regardless of starting point, and thus it has the most room for antag resources; I have not managed to find the conviction to argue in favour of this idea in a while.

PR #23989, opened 2025-06-30
Admin+Mentor OOC Icons

JORJ949:
## About the PR 
Gives Admins and Mentors OOC icons based on their mice. Can be hovered over like other icons to see "Admin" or "Mentor"

## Why's this needed? 
Colour difference between admin and mentors is only really learnt if you learn the Who command or someone tells you, this lets you figure out "oh the friendly purple mentor rat that helped me means they're a mentor" or "oh the evil pink admin rat that ran me over with a bus that means they're an admin" or more simply, just hover over them to see "Admin" or "Mentor"

## Testing 
image

## Changelog 

```changelog
(u)JORJ949
(+)Admins and Mentors now have OOC icons you can hover over to see their rank (instead of just knowing purple = mentor).
```


TobleroneSwordfish (approved):
mous

PR #23991, opened 2025-06-30
Cogmap 2 req_access_txt pass

JORJ949:
## About the PR 
Removes all req_access_txt variable overrides on Cogmap 2 and replaces them with appropriate mappinghelpers
In doing so replaces the "Special Club" access on the beamline interlock with Impossible access.

## Why's this needed? 
Mappinghelpers are much cleaner than varedits

## Testing 
Tested the beamline interlock, still works as before, nobody has access to open or close it. Other access changes should be literally the exact same access but on mappinghelper instead of varedit


JORJ949:
Not sure why the merge did that.

JORJ949:
I cannot be bothered to fix this.

PR #23995, opened 2025-07-01
Add agitator artifact

FlameArrow57:
[SCIENCE][FEATURE]

## About the PR 
Adds a new lattice origin unique artifact, the Agitator

This artifact can be used on other artifacts to either give them a fault if they don't have one already, or reroll a random fault on the artifact. Added faults have a higher activation chance than usual

For non-traitors, this is equivalent to a "useless" artifact that you would likely sell. Otherwise, you could sabotage an artifact a traitor is using

For traitors, this is a useful art that can be used to sabotage useful arts being used on station

## Why's this needed? 
New Lattice origin artifact

## Testing 
Tested using the artifact on activated/unactivated artifacts

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new Lattice origin Agitator artifact. Using it on any activated artifact gives that artifact a guaranteed fault with higher activation chance.
```


Laboredih123:
source?

FlameArrow57:
Lol

code/obj/artifacts/artifact_items/agitator.dm
```diff
@@ -0,0 +1,43 @@
...
+		if (!src_art.activated)
+			return ..()
+		var/obj/O = target
+		if (!istype(O) || !O.artifact)
+			return ..()
+		if (!O.artifact.activated)
+			return ..()
+		. = ..()
+		if (ON_COOLDOWN(src, "artifact_agitation", rand(30, 180) SECONDS))
+			boutput(user, SPAN_NOTICE("[src] doesn't seem to do anything. Hm."))
```
TobleroneSwordfish: Could do with a message (or a few randomized ones) for when it triggers, sound cues shouldn't be relied upon solely. Maybe make the struck artifact judder or do some angry looking animation as well?

code/obj/artifacts/artifact_items/agitator.dm
```diff
@@ -0,0 +1,43 @@
+/obj/item/artifact/agitator
+	name = "artifact agitator"
+	associated_datum = /datum/artifact/agitator
+
+	afterattack(atom/target, mob/user, reach, params)
+		var/datum/artifact/agitator/src_art = src.artifact
```
TobleroneSwordfish: Should use `ArtifactSanityCheck`

code/obj/artifacts/artifact_items/agitator.dm
```diff
@@ -0,0 +1,43 @@
...
+		if (!src_art.activated)
+			return ..()
+		var/obj/O = target
+		if (!istype(O) || !O.artifact)
+			return ..()
+		if (!O.artifact.activated)
+			return ..()
+		. = ..()
+		if (ON_COOLDOWN(src, "artifact_agitation", rand(30, 180) SECONDS))
```
TobleroneSwordfish: Most artifacts like this randomize their cooldown once and save it to the artifact datum, this probably should too?

code/obj/artifacts/artifact_items/agitator.dm
```diff
@@ -0,0 +1,43 @@
+/obj/item/artifact/agitator
+	name = "artifact agitator"
+	associated_datum = /datum/artifact/agitator
+
+	afterattack(atom/target, mob/user, reach, params)
```
TobleroneSwordfish: This should ideally be implemented on the artifact datum itself rather than down here, I know artifact datums don't have a direct replacement for `afterattack` but you could add one 🥺

TobleroneSwordfish (changes requested):
I like this, messing with faults is neat and the rare chance to stack up a bunch of them or make one super likely is cool.

pgmzeta:
why lattice-unique, could be eldritch too

FlameArrow57:
> why lattice-unique, could be eldritch too

Lattice is meant to be artifact types that directly modify other artifacts and nothing else

code/obj/artifacts/artifact_items/agitator.dm
FlameArrow57: Maybe in another PR? Since this isn't the only case where afterattack() is implemented on a handheld?

TobleroneSwordfish (changes requested):
#24132

PR #24000, opened 2025-07-02
Add archivist artifact

FlameArrow57:
[SCIENCE][FEATURE]

## About the PR 
Adds a new artifact, the Archivist

This is meant to be a Lattice origin unique artifact (once Lattice artifacts are merged).

This artifact is always language activated and can be used on other artifacts to reveal info about them

If used on an unactivated artifact, it reveals three possible activation methods with one being the actual method

If used on an activated artifact, it reveals if its origin is disguised, its activation method, and any faults it has

## Why's this needed? 
Lattice origin unique artifact

## Testing 
Tested using artifact on station objects, unactivated artifacts, activated artifacts, and artifacts with no triggers

## Changelog 

```changelog
(u)FlameArrow57
(*)Added a new Archivist artifact. Reveals useful info about other artifacts
```

### DIFF
--- a/code/obj/artifacts/artifact_items/archivist.dm
+++ b/code/obj/artifacts/artifact_items/archivist.dm
@@ -1,0 +1,70 @@
+/obj/item/artifact/archivist
+	name = "artifact archivist"
+	associated_datum = /datum/artifact/archivist
+
+	afterattack(atom/target, mob/user, reach, params)
+		var/datum/artifact/compass/art = src.artifact
+		if (!art.activated)
+			return ..()
+		. = ..()
+		var/obj/O = target
+		if (!istype(O) || !O.artifact)
+			src.say("[O] is not an artifact, results inconclusive.")
+		else if (O.artifact.activated)
+			var/datum/artifact_trigger/activating_trigger = O.artifact.triggers[1]
+			var/list/art_faults = list()
+			for (var/datum/artifact_fault/fault in O.artifact.faults)
+				art_faults += fault.type_name
+			src.say("Analysis results:")
+			sleep(0.1 SECONDS)
+			src.say("Origin disguised: [O.artifact.disguised ? "Yes" : "No"]")
+			sleep(0.1 SECONDS)
+			src.say("Activation method: [art.trigger_names_assoc[activating_trigger.type]]")
+			sleep(0.1 SECONDS)
+			src.say("Faults: [length(art_faults) ? english_list(art_faults) : "None"]")
+			sleep(0.1 SECONDS)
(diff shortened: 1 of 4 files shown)


code/obj/artifacts/artifact_items/archivist.dm
```diff
@@ -0,0 +1,70 @@
...
+				art_faults += fault.type_name
+			src.say("Analysis results:")
+			sleep(0.1 SECONDS)
+			src.say("Origin disguised: [O.artifact.disguised ? "Yes" : "No"]")
+			sleep(0.1 SECONDS)
+			src.say("Activation method: [art.trigger_names_assoc[activating_trigger.type]]")
+			sleep(0.1 SECONDS)
+			src.say("Faults: [length(art_faults) ? english_list(art_faults) : "None"]")
+			sleep(0.1 SECONDS)
+			//src.say("Artifacts combined: [length(O.combined_artifacts) || 0]"))
```
mordent-goonstation: Absolutely not on this sleep chain.

mordent-goonstation: Also why is the artifact talking?

code/obj/artifacts/artifact_items/archivist.dm
```diff
@@ -0,0 +1,70 @@
...
+	var/static/list/trigger_names_assoc = list()
+	var/list/scanned_artifacts = list()
+
+	New()
+		..()
+		if (!length(src.trigger_names))
+			for (var/datum/artifact_trigger/trigger_type as anything in concrete_typesof(/datum/artifact_trigger))
+				if (initial(trigger_type.used))
+					src.trigger_names += initial(trigger_type.type_name)
+					src.trigger_names_assoc[trigger_type] = initial(trigger_type.type_name)
```
mordent-goonstation, line 73: Isn't this going to be the same for all of these artifacts? Why is this being done in each?

code/obj/artifacts/artifact_items/archivist.dm
FlameArrow57: I put them under SPAWN(0)s

It's a language only activated artifact, theme-wise the artifact uses language to determine how to communicate to the holder

code/obj/artifacts/artifact_items/archivist.dm
FlameArrow57: var trigger_names is a static var

TobleroneSwordfish:
Ehh I don't really see the point of this one, it also feels kind of video gamey especially with the way it outputs it (just a clean clinical list like a regular scanner)

pgmzeta:
This idea is too insular and meta to artifact science.

PR #24003, opened 2025-07-02
Refactor and fix trowel

### LINKED ISSUE #17265: trowel transfers the wrong plant to the plant pot
### Labels

[BUG][MINOR]

### Description

The sprite for a different plant shows up when you use a trowel to pot a plant sometimes.

### Steps to reproduce

Have the desired plant in a botany tray. Use trowel on it. Use portable seed fabricator to plant a different plant in that tray. Use the trowel on an empty plant pot. The plant sprite in the pot will be the seedling you planted instead of the grown plant you used the trowel on.

### Expected Behavior

It should transfer the plant you used the trowel on, and not the one you plant afterwards.

This can be really irritating if you try to pot a fully-grown plant that can be difficult to get, such as light lotus, and if the bug occurs it can completely waste the plant you wanted to pot.

Reported by: Snailsman
Round log date: 2023-12-10-03-48
Reported on: 2023-12-10 04:58:40
Map: NADIR

### LINKED ISSUE #21860: weird hydro tray/plant pot interaction
### Labels

[BUG]

### Description

hydro trays can change plant pots

### Steps to reproduce

1. grow someone in a hydro tray
2. dig it up with a trowel
3. plant something in tray
4. use trowel on plant pot
5. have first stage plant in pot instead of full grown

### Expected Behavior

give the full grown plant

this works for any plants btw, example, im growing birds of paradise, i trowel a full grown plant to put in the pot, i replant the flower first, then use the trowel on the plant pot and now the plant that was fully grown is stage 1 and ive wasted a pot. it cant be any seed, ie i trowel a full grown plant, put bamboo, the  use the trowel on the pot and now its stage 1 bamboo

### LINKED ISSUE #23301: Missing/Wrong sprite for plant pot
### Labels

_No response_

### Description

Placing Dogwood Tree in plant pot makes it look like a sapling.

### Steps to reproduce

1. Get a Dogwood Tree mutation
2. Use garden trowel to move it to plant pot

### Expected Behavior

Plant pot sprite should have correct plant sprite

I checked just Dogwood Tree but more plants might be missing their sprites for plant pots.

Egregorious:
## About the PR 

Refactors the trowel and changes the way it holds the plant's overlays so they don't get overriden by altering the contents of the tray the plant was pulled from. 

## Why's this needed? 

fixes #23301
fixes #21860
fixes #17265
Cleaner code

## Testing 

trowel bug fix
Also made sure the glowstick tree still glowed.

### DIFF
--- a/code/obj/item/decorative_pot.dm
+++ b/code/obj/item/decorative_pot.dm
@@ -24,10 +24,10 @@ TYPEINFO(/obj/decorative_pot)
 						var/obj/item/gardentrowel/trowel = weapon
 						if(!trowel.holding_plant)
 								return
-						if (holding_plant)
+						if (src.holding_plant)
 								// This probably introduces more bugs than it fixes.
 								src.ClearAllOverlays()
-						holding_plant = TRUE
+						src.holding_plant = TRUE
 						src.UpdateOverlays(trowel.create_plant_image(),"plant")
 						src.UpdateOverlays(trowel.create_plant_overlay_image(), "plantoverlay")
 						trowel.genes.mutation?.HYPpotted_proc_M(src, trowel.grow_level)
--- a/code/obj/item/decorative_pot.dm
+++ b/code/obj/item/decorative_pot.dm
@@ -7,6 +7,7 @@ TYPEINFO(/obj/decorative_pot)
 		icon_state = "plantpot"
 		anchored = UNANCHORED
 		density = 1
+		var/holding_plant = FALSE
 
 		attackby(obj/item/weapon, mob/user)
 				if((iswrenchingtool(weapon)) || isscrewingtool(weapon))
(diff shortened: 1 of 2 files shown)


code/obj/item/decorative_pot.dm
```diff
@@ -19,21 +20,25 @@ TYPEINFO(/obj/decorative_pot)
 								playsound(src.loc, 'sound/items/Screwdriver.ogg', 100, 1)
 								src.anchored = UNANCHORED
 						return
-				else if(istype(weapon,/obj/item/gardentrowel))
-						var/obj/item/gardentrowel/t = weapon
-						if(!t.plantyboi)
+				if (istype(weapon, /obj/item/gardentrowel)) // prevent attack message with a filled trowel
```
JORJ949: No need to lose the else here, trowels arent screwing or wrenching tools and the above already returns if it is

Egregorious:
At the risk of being presumptuous I've removed the attack suppression check, and better handled it in #24168

code/obj/item/decorative_pot.dm
```diff
@@ -19,21 +20,21 @@ TYPEINFO(/obj/decorative_pot)
...
 				if(istype(weapon,/obj/item/seed))
 						boutput(user, "It's an empty pot, there's nowhere to plant the seed! Maybe you need to use a trowel and place an existing plant into it?")
 				else
 						..()
+
+		/// Inserts a new plant into the pot, overriding any previous.
+		proc/insert_plant(var/image/plant_image, var/image/plant_overlay_image, var/datum/plantgenes/genes, var/grow_level)
+				if (src.holding_plant)
+						// This probably introduces more bugs than it fixes.
+						src.ClearAllOverlays()
```
pgmzeta: Could clear the "plant" and "plantoverlay" overlays with `ClearSpecificOverlays` instead so other overlays aren't modified/cleared.

Egregorious: The problem is that some plants want to use more than two overlays. Right now that's namely the glowstick tree that uses nonstandard overlay keys; the plant pot deals with it by calling the plant itself but the decorative pot doesn't hold a reference to the plant to be able to do that. 

It seems like the proper solution would be for the decorative pot to hold the plant, but it felt silly to add that overhead just to specifically stop glowsticks from getting stuck to the pot.

Egregorious: I removed the check. The only bug that it really fixed right now is an extra overlay from the glowstick tree being stuck to the pot. It's a purely cosmetic bug that's existed for ages; people can technically use it to make some neat effects; and it's so unnoticeable there's not even an issue open about it. So perhaps it's not worth fixing currently.

pgmzeta: In a future PR, the decorative pot could hold a reference so the plant in order to call the plant's cleanup/destruction HYP proc. This is good for now.

PR #24004, opened 2025-07-02
some simple TGUI alignments

iiOxygen:
[UI] [QoL] [Minor]

## About the PR 
Addresses various icon misalignments & styling issues.
Also fixes the 2 duplitcate supplies' upgrade titles in the docking station which I barely even noticed.

## Why's this needed? 
People with OCD can magically see some of these extremely minor issues.

## Testing 
https://imgur.com/a/xn0nqaT


.gitignore
```diff
@@ -81,3 +81,12 @@ Desktop.ini
...
 goonstation.json
+Nuclear Reactor Simulator.html
+StrongDMM-2025-06-26 16.37.06.png
+StrongDMM.exe
+telesci.html
+README.md
+.clj-kondo/.cache/v1/lock
+.lsp/.cache/db.transit.json
+tgui/packages/tgui-dev-server/link/client.cjs
+_std/__build.dm
```
JORJ949: These shouldn't be added  to the gitignore, StrongDMM shouldn't be in the goonstation folder, you shouldn't be making changes to README(unless of course there are README changes needing making, which there aren't in this  PR) and to have a gitignored build.dm copy it and rename it to "__build.local.dm". I'm assuming these html files are programs for calculating telesci/nuclear reactors, you should store these elsewhere and not in your goonstation development folder. As for the cache files I feel like these would already be in the gitignore if they should be.

If you have files you don't want to commit you can just not stage them, as shown in vsc here
image
Here i'd only be committing the changes made to stool.dm

iiOxygen: thats great to know, thank you!

tgui/packages/tgui/interfaces/IDComputer.tsx
```diff
@@ -58,24 +58,33 @@ export const DeptBox = (props) => {
...
-                    mr="1rem"
-                  />
-                </>
+                <div
+                  style={{
+                    display: 'inline-flex',
+                    gap: '1rem',
+                    alignItems: 'center',
+                    whiteSpace: 'nowrap',
+                  }}
```
ZeWaka, line 67: Inline style seems sus here, why not just put it in the css

ZeWaka:
this should be targeting the [tgui_core_update](https://github.com/goonstation/goonstation/tree/tgui_core_update) branch

ZeWaka:
done in branch

PR #24010, opened 2025-07-03
New __build.dm flag to prevent TGUI antagonist popups

MyNameIsRomayne:
## About the PR 
Adds a new flag, `NO_ANTAG_POPUPS_I_DONT_CARE`, to the list of flags for STOP_DISTRACTING_ME. This flag prevents the traitor popup (or whatever else it may be) from appearing whenever you start an instance.

## Why's this needed? 
It's annoying! It's been annoying. In a local instance by default, you are a antagonist every single time, thus is appears every single time. This flag means one less thing distracting you while you bugtest or playtest whatever it may be you are truly working on.

## Testing 

Can't easily provide a *screenshot* but here's the game starting:

https://github.com/user-attachments/assets/151093b2-9c65-444b-83ce-c6fa467e7991


JORJ949:
Local will remember the game mode you set if you set the mode to be not just that round whhen setting it via the admin game panel, setting extended means you will never be antag. Though I  like this its nice for testing antag features, I've been wanting to add similar defines for "I_WANNA_PLAY_THE_MODE" and "I_WANNA_BE_THE_VILLAIN"

JORJ949 (approved):

ZeWaka:
yeah just set it to extended

MyNameIsRomayne:
Huh! Never did I ever know that. I'll keep that in mind (I normally just use a program to keep persistent changes outside of Annoying Stuff) but this change is still out there if its wanted. And yeah, it would likely be nice for testing antagonists where you're going to see that popup a lot.

Though, if this is closed in favor of persistent rounds, I think it may be nice to have that tip somewhere if it isn't already. I at the very least have read the guides repeatedly, I don't think I saw it mentioned anywhere. Would be nice to know these things!

PR #24012, opened 2025-07-03
Add artifact Fuser machine

FlameArrow57:
[SCIENCE][FEATURE]

## About the PR 
This PR adds the code for a new artifact Fuser machine.

The artifact Fuser machine can combine artifacts into each other

Adds the code only, does not add it into the game yet

## Why's this needed? 
Would be cool

## Testing 
TBD


TobleroneSwordfish (changes requested):
Lots of bugs/confusing behaviour:
- Clicking a handheld artifact with a combiner while inhand doesn't remove it from your hand, needs to call unequip/drop proc
- Almost nothing actually seems to be combinable? No handhelds will combine with each other or any larger artifact type I can find, most large artifact types refuse to combine with each other, the only two I could get to work were a surgery machine and a recaller. Can you clarify which artifacts are meant to work together and why most of them don't seem to?
- Artifacts are required to be active and un-anchored in order to combine, which precludes any artifact type that is anchored while active, forcefields etc.
- Not sure the mechanism of picking artifacts up to combine them is a good idea, this allows you to for example put activated artbombs in your backpack. Why not just click drag between two artifacts to combine them or use a linking mechanism like mechcomp multitools (still requiring them to be adjacent)?

FlameArrow57:
Updated per PR description

PR #24016, opened 2025-07-04
Device Frame QoL + fixes for anchored items in storage

### LINKED ISSUE #13845: Remove trigger on the Device Frame
### Labels

[BUG][MINOR]

### Description

You can remove the trigger on the mechcomp device frame

### Steps to reproduce

1. Hold device frame in hand
2. Select the hands its in
3. Open and click on the trigger
4. Trigger is now gone

### Expected Behavior

Trigger not vanishing

Reported by: Cindertroy
Active test merges: #13023 and #13667
Round log date: 2023-04-25-19-09
Reported on: 2023-04-25 19:37:10
Map: DONUT3

### LINKED ISSUE #13956: Handheld device frame trigger can be removed by dragging the frame to the floor
### Labels

[BUG] [GAME OBJECTS]

### Description

Title.

### Steps to reproduce

1. Drag Device Frame to floor.
2. The trigger is now on the floor.

### Expected Behavior

For it to stay inside.

_No response_

RubberRats:
[BUG][GAME OBJECT][QOL]
## About the PR 

This PR makes pressing the button on a mechcomp device frame work even if the frame is in your active hand, rather than smacking the button with it. It also makes a few changes to storage code to prevent it from doing weird things when attempting to dump out or remove anchored items from storage.
## Why's this needed? 

Small QoL improvement for working with device frames.
Fixes #13845 
Fixes #13956

## Testing 
Tested by attempting to dump out the device frame on the floor, and using the button while the frame was open and in the active hand.

### DIFF
--- a/code/modules/storage/storage.dm
+++ b/code/modules/storage/storage.dm
@@ -252,6 +252,8 @@
 				return
 		user.visible_message(SPAN_ALERT("[user] dumps the contents of [src.linked_item.name] onto [over_object]!"))
 		for (var/obj/item/I as anything in src.get_contents())
+			if(I.anchored)
+				continue
 			src.transfer_stored_item(I, T, user = user)
 			I.layer = initial(I.layer)
 			if(SEND_SIGNAL(I, COMSIG_ITEM_STORAGE_INTERACTION, user))
@@ -263,6 +265,9 @@
 
 /// after attacking an object with the storage item
 /datum/storage/proc/storage_item_after_attack(atom/target, mob/user, reach)
+	var/obj/O = target
+	if (O.anchored)
+		return
 	// if item is stored, drop storage and take it out
 	if (target in src.get_contents())
 		user.drop_item()
@@ -271,9 +276,6 @@
 			target.Attackhand(user)
 	// attempt to load item into storage if you have a free hand
 	else if (isitem(target) && !istype(target, /obj/item/storage))
-		var/obj/O = target
-		if (O.anchored)
-			return
 		if (!can_reach(user, target))
 			return
 		if (issilicon(user))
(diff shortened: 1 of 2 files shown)


JORJ949 (approved):

code/modules/storage/storage.dm
```diff
@@ -263,6 +265,9 @@
 
 /// after attacking an object with the storage item
 /datum/storage/proc/storage_item_after_attack(atom/target, mob/user, reach)
+	var/obj/O = target
+	if (O.anchored)
```
Egregorious:
```suggestion
	if (O?.anchored)
```
I think you need a null check here.

JORJ949: O?.anchored works here, but you do need to make sure O is the proper type with an istype(O)

Egregorious: Academically, is there a benefit to `if (istype(O, /obj) && O?.anchored)` over `if (O && O.anchored)`? It seems to me to just be adding an extra type check for no reason.

JORJ949: Here target is a atom/target, which is then assigned to var/obj/O, at this point we don't know if it is an object, turf, area or mob, and should check istype(O), second argument not required as it will check for the type O is assigned to. O?.anchored is essentially O && O.anchored but more compact.

Egregorious: Ah I think I had misunderstood what you were saying, thanks for explaining.

ZeWaka: needs istype

ZeWaka (changes requested):

code/modules/storage/storage.dm
pgmzeta:
```suggestion
	if (istype(O) && O.anchored)
```

PR #24037, opened 2025-07-05
Allow IDs to be equipped in the ID slot even if there is a PDA in there already

MyNameIsRomayne:
## About the PR 
Allows the `force_equip` proc to equip IDs when there are PDAs in the ID slot, as well as `equip_if_possible`.

## Why's this needed? 
Allows the code to do things that are already possible by player actions. Arguably a bugfix, but there isn't really behavior I am aware of that exposes this bug during gameplay. Really, it just allows roundstart jobs that may start with a PDA in the ID slot in the future to add an ID in there as well, instead of forcing it into the backpack.

## Testing 

1. Created a special setup job which spawns a pda2 object and places it into the ID slot.
2. Defined an ID as being on the slot_card field of the job
3. Spawned in as aforementioned job
4. Verified that I spawned in with the ID in the PDA, and nowhere else (e.g. backpack)

Additionally spawned in as a tourist job, and as a regular staff assistant to verify normal behavior still functions.

image

image

image

image


MyNameIsRomayne:
https://discord.com/channels/182249960895545344/890226559691157524/1391101673304490137 read this chain of discussion for context, there are two options here:

1. code as it is written on most recent commit
2. implementation on `force_equip` with a reimplementation of `insert_id_card` to avoid attackby reference

JORJ949:
I think this should be a character customisation setting, this is something a lot of people do literally every shift so it being a setting would be a nice QoL change, though this is useful for jobs with a lot of items such as the NT and Syndicate ones (Syndie ones currently don't have a PDA as they are too clogged full of other stuff)

MyNameIsRomayne:
I feel like that would be a separate PR if anything -- sounds like something people would have opinions on. I'm with you though

ZeWaka:
yuh

PR #24039, opened 2025-07-05
Move datum/artifact/proc/effect_melee_attack() usage to the datum

FlameArrow57:
[INTERNAL][CLEANLINESS]

## About the PR 
Title

## Why's this needed? 
Standardizes more of artifact code to be on the datum rather than the object. Also allows for more artifact combination effects for the Combiner artifact

## Testing 
Tested heal wands, melee weapons

### DIFF
--- a/code/obj/artifacts/artifact_items/healing_wand.dm
+++ b/code/obj/artifacts/artifact_items/healing_wand.dm
@@ -4,17 +4,6 @@
 	associated_datum = /datum/artifact/healing_wand
 	click_delay = COMBAT_CLICK_DELAY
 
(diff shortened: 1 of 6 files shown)


ZeWaka (commented):
why not define this on the artifact datum? seems a bit counter-intuitive to the artifact pseudo-component

FlameArrow57:
> why not define this on the artifact datum? seems a bit counter-intuitive to the artifact pseudo-component

It is defined, as effect_melee_attack(). It's how the other artifact procs handle it I think, there's an obj level proc that acts as a wrapper for the artifact datum proc

ZeWaka:
yeah i don't get the existence of the obj level proc

PR #24042, opened 2025-07-05
Standardize nuclear operative storage options

FlameArrow57:
[GAME MODES][BALANCE][QoL]

## About the PR 
Reopening of #15519

Adds:
-Nuclear operative suits/coats get a 6 slot storage
-All nuke ops class crates come with a nuclear operative fanny pack

Removes:
-Fanny pack vendor option removed
-Tactical pouch vendor option removed

## Why's this needed? 
-Removes messiness of needing to manage storage
-On average increases storage space across classes. Nuke ops win rate is around 40% or so, so hopefully it will increase it more towards 50%
-Reduces some of the awkwardness for classes that come with two pouches when you can realistically only use one pouch due to needing a pouch slot for your oxygen

## Testing 
Made sure nuclear operative suits were being given storage properly

## Changelog 

```changelog
(u)FlameArrow57
(*)Made some inventory changes for nuke ops, see minor changelog for details
(+)All nuke ops suits come with a 6 slot inventory and their class crates all get a fanny pack
(+)Fanny pack and tactical pouch can no longer be bought from the vendor
```

### DIFF
--- a/code/obj/item/clothing/suits.dm
+++ b/code/obj/item/clothing/suits.dm
@@ -1336,9 +1336,12 @@ TYPEINFO(/obj/item/clothing/suit/hazard/fire/armored)
 	desc = "A suit that protects against low pressure environments. Issued to syndicate operatives."
 	contraband = 3
 	item_function_flags = IMMUNE_TO_ACID
+	var/receives_storage = TRUE
 
 	New()
 		START_TRACKING_CAT(TR_CAT_NUKE_OP_STYLE)
+		if (src.receives_storage)
+			src.create_storage(/datum/storage, max_wclass = W_CLASS_SMALL, slots = 6, opens_if_worn = TRUE)
 		..()
 
 	disposing()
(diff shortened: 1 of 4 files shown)


JORJ949:
The suits having storage feels weird and unnatural, the fanny pack should if everyone gets one just spawn directly on the nukeop (although I think they should have their chosen pistol belt there instead). I do agree with the pocket pouch thing though, perhaps the suits could have integrated oxygen tanks. The reason those classes get the fanny pack by default is because they don't get their backpack due to having something else that goes there, their big gun, giving everyone else 10 more slots than them just feels weird. With this change nukeops will have a  total inventory space of 23 (10 from backpack, 7 from fanny, 6 from suit) and that just feels like far too much. The only class I ever have issue carrying everything as is infil because they get a smaller bag.

Additional but entirely optional thing is that these changes should be reflected on their Special Operative counterpart

Games don't need a 50% winrate to be balanced, sometimes the fun is the uphill battle.

Gannets:
I largely agree with @JORJ949's comment here. I raised the nuke ops backpack slots up to 10 a good while ago, gave big gunners a belt option for storage and I haven't felt like nukies have been really struggling for storage since then. Though I'm sure they have to make some choices about what to bring.

Regarding win-rates, I'd rather be shooting for "most nuke ops rounds last untl the bomb timer nearly hits zero" than a specific win percentage. It's supposed to be a big climatic firefight!

Issue #23997, opened 2025-07-01
Ammo oddities


goonstation-issuebot:
### Labels

[BUG][MAJOR]

### Description

.308 doesn't fit in a .308 gun?

### Steps to reproduce

Be traitor and buy a surplus
Get both an Old Hunting Rifle and Tranquilzer Rifle
Lose Old Hunting Rifle and try to put the ammo in the Tranquilizer Rifle
Get "This ammo won't fit!" despite both saying they are chambered in .308

### Expected Behavior

For a .308 caliber gun to take the other .308 caliber gun

### Additional Information & Screenshots

Tranquilizer Rifle despite also using the Old Hunting Rifle model, doesn't allow usage of the scope it has too.

Reported by: kaedrin18
Client version: 515.1647
On server: Goonstation 4 Roleplay: Sylvester
Active test merges: #23993, #23292, and #23422
Round log date: 2025-07-01-15-17
Reported on: 2025-07-01 16:46:30
Round ID: [207941](https://goonhub.com/rounds/207941)
Map: [OSHAN](https://goonhub.com/maps/oshan)

flappybatpal:
Intentional. Description could be clearer.

444Portal:
Processed comment right after closing, my bad.

PR #24054, opened 2025-07-06
Clearer description with the tranq rifle

TheGeneralJay:
[GAME OBJECT][QoL]

## About the PR 
Changes the tranquilizer rifle's description to mention .308 ammo compatibility. Adds a descriptive message when trying to load incompatible .308 rifle ammo.

## Why's this needed? 
Addresses #23997 by making the description and chat feedback clearer.

## Testing 

{05DF6DBF-1ADD-42E1-83E4-37F8DDE8D4FF}

{8745B58C-4744-4863-8B46-9FABFAEF00F7}

It's possible we don't necessarily need the feedback on attempting to reload, but it does make the issue a little clearer to the user. Can take it or leave it with this one.

### DIFF
--- a/code/obj/item/gun/kinetic.dm
+++ b/code/obj/item/gun/kinetic.dm
@@ -123,12 +123,8 @@ ABSTRACT_TYPE(/obj/item/gun/kinetic)
 					user.show_text("You can't reload this gun.", "red")
 					return
 				if(AMMO_RELOAD_INCOMPATIBLE)
-					if(istype(src, /obj/item/gun/kinetic/dart_rifle))
-						user.show_text("[src] can only accept tranquilizer darts!", "red")
-						return
-					else
-						user.show_text("This ammo won't fit!", "red")
-						return
+					user.show_text("This ammo won't fit!", "red")
+					return
 				if(AMMO_RELOAD_SOURCE_EMPTY)
 					user.show_text("There's no ammo left in [b.name].", "red")
 					return
@@ -855,6 +851,13 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 	New()
 		ammo = new default_magazine
 		set_current_projectile(new/datum/projectile/bullet/tranq_dart)
+		..()
+
+	// Clearer description for invalid dart rifle ammo.
+	attackby(obj/item/ammo/bullets/b, mob/user)
+		if(!istype(b, default_magazine))
+			user.show_text("[src] can only accept .308 tranquilizer darts!", "red")
+			return
 		..()
 
 //9mm/0.355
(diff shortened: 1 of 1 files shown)


code/obj/item/gun/kinetic.dm
```diff
@@ -123,8 +123,12 @@ ABSTRACT_TYPE(/obj/item/gun/kinetic)
 					user.show_text("You can't reload this gun.", "red")
 					return
 				if(AMMO_RELOAD_INCOMPATIBLE)
-					user.show_text("This ammo won't fit!", "red")
-					return
+					if(istype(src, /obj/item/gun/kinetic/dart_rifle))
```
TobleroneSwordfish: `istype(src` is a code smell, please refactor to a var or override proc

TobleroneSwordfish (changes requested):

code/obj/item/gun/kinetic.dm
TheGeneralJay: Swapped out for an override attackby on the dart rifle itself when trying to reload with any invalid ammunition to clear up confusion for users! Lmk if it needs more adjusting, still learning as I go. Already re-tested to ensure it works. Thanks!

code/obj/item/gun/kinetic.dm
```diff
@@ -853,6 +853,13 @@ ABSTRACT_TYPE(/obj/item/survival_rifle_barrel)
 		set_current_projectile(new/datum/projectile/bullet/tranq_dart)
 		..()
 
+	// Clearer description for invalid dart rifle ammo.
+	attackby(obj/item/ammo/bullets/b, mob/user)
```
TobleroneSwordfish: Uhh noo I meant more like a var/proc that just displays the message and gets called from the parent `attackby`, this is code duplication and will change the behaviour slightly in odd ways.

You could just change the parent `show_text` to `show_text(src.ammo_incompatible_msg)`

TobleroneSwordfish (changes requested):

code/obj/item/gun/kinetic.dm
TheGeneralJay: Ah! That makes sense, sorry-- I misunderstood what you meant by override lol. Thanks for bearing with me. I've turned the incompatibility message into a var that can be customized on specific guns if clarity is needed on other ones in the future. Lmk if anything else needs changing!

JORJ949 (changes requested):
Would it be possible to have this declared on the type instead? Declaring variables on new is :/ though I don't think you could declare it with [src] on the type.

TheGeneralJay:
> Would it be possible to have this declared on the type instead? Declaring variables on new is :/ though I don't think you could declare it with [src] on the type.

I thought so too. It spits an error trying to access src from the type, although, since it's a custom description / can be changed on other weapons, I could just write out the gun name manually, or access the name property itself.

```
/obj/item/gun/kinetic/dart_rifle
	name = "tranquilizer rifle"
	desc = "A veterinary tranquilizer rifle chambered in .308 caliber. This rifle can only accept .308 tranquilizer darts."
        ...
	ammo_incompatible_msg = "The tranquilizer rifle can only accept .308 tranquilizer darts!"
```
If that's preferable, that sounds good. I can make that change. Let me know!

PR #24058, opened 2025-07-07
Update APC area upon it creating its own built zone

### LINKED ISSUE #22741: APC Vanishing in Built Zone
### Labels

[BUG][MAJOR]

### Description

When Building off a blueprint with the abcu machine, a blueprint that contains an APC, the apc will not function in the newly built room, even if connected to the grid. Futhermore, when you deconstruct and rebuild the APC (a fix from an earlier version) it now just vanishes and seems to despawn the object.

### Steps to reproduce

1. build a blueprint that contains an APC
2. power on new area
3. deconstruct APC
4. Use sodering iron on APC frame

### Expected Behavior

When the APC frame has the sodering iron used on it, an APC is built.

Ive tried this multiple times, with multiple maps and multiple copied rooms. it is consistant.

Reported by: thepowerfulwill
Client version: 515.1637
Active test merges: #21058 and #22547
Round log date: 2025-03-27-04-08
Reported on: 2025-03-27 05:01:39
Round ID: 203022
Map: KONDARU

MyNameIsRomayne:
## About the PR 
When a APC is built in an unconnected area, it converts it to a built zone. Problem is when it does, it still registers itself as powering the old area (which doesn't exist anymore). This one line fix resolves that by updating its area to the new one

## Why's this needed? 
Fixes #22741
(not sure what they were talking about with the APC deleting itself, but it properly powers the area now)

## Testing 

image

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -254,10 +254,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/power/apc, proc/toggle_operating, proc/zapSt
 			. += "The cover is open and the power cell is [ cell ? "installed" : "missing"]."
 		else
 			. += "The cover is closed."
-
-/obj/machinery/power/apc/proc/update_area()
-	src.area = get_area(src)
-
+	
 /obj/machinery/power/apc/proc/toggle_operating()
 	src.operating = !src.operating
 	src.update()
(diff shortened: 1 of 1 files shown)


code/modules/power/apc.dm
```diff
@@ -209,6 +209,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/power/apc, proc/toggle_operating, proc/zapSt
 	if (!QDELETED(src.area))
 		if(istype(src.area,/area/unconnected_zone)) //if we built in an as-yet APCless zone, we've created a new built zone as a consequence
 			unconnected_zone.propagate_zone(get_turf(src))
+			src.area = get_area(src)
```
TobleroneSwordfish: Shouldn't this also assign the area.area_apc var?

MyNameIsRomayne: ah yeah you're right. one moment

MyNameIsRomayne:
image
fixed the other side of the change, i thought the propagate_zone proc did it already but it turns out i was WRONG

PR #24079, opened 2025-07-08
Consistent capitalization for the "Chav" accent

snugglycactus:
## About the PR 
This PR changes the coding of the Chav accent to align with the Elvis accent., allowing vowel changes to be properly capitalized instead of coming out as an eye-bleeding mismatch of capital and lowercase letters.

## Why's this needed? 
Trying to express some right British rage and having it come out as "FooCK YEee" should be a problem of the past. Now, we can say "FOOCK YEEE" as God himself intended.

## Testing 

Screenshot 2025-07-08 181905
Behold, the quaint chav, enjoying a pint with his lads.

Screenshot 2025-07-08 181924
And here we have an agitated chav, speaking in a secret tongue.

Screenshot 2025-07-08 182015
As we can see, the chav has been given the ability to yell at us. Use it well, chav!
*changes were only made to capitalization, "com" instead of "come" was manually inputted for comedy's sake.

### DIFF
--- a/code/procs/accents.dm
+++ b/code/procs/accents.dm
@@ -1192,147 +1192,147 @@ proc/random_accent()
 	var/new_string = ""
 	var/used = 0
 
-	switch(lowertext(R.curr_char))
+	switch(R.curr_char)
 		if("w")
-			if(lowertext(R.next_char) == "h" && lowertext(R.next_next_char) == "a")
+			if(R.next_char == "h" && R.next_next_char == "a")
 				new_string = "wo"
 				used = 3
-		/*if("W")
-			if(lowertext(R.next_char) == "H" && lowertext(R.next_next_char) == "A")
+		if("W")
+			if(R.next_char == "H" && R.next_next_char == "A")
 				new_string = "WO"
-				used = 3*/
+				used = 3
 
 		if("o")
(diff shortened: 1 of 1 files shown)


ZeWaka (commented):
doesn't this break entirely for mixed case

snugglycactus:
> doesn't this break entirely for mixed case

image
Does it? I don't see anything wrong in this screenshot, but I could be entirely off base

code/procs/accents.dm
```diff
@@ -1192,147 +1192,147 @@ proc/random_accent()
...
+			if(R.next_char == "h" && R.next_next_char == "a")
 				new_string = "wo"
 				used = 3
-		/*if("W")
-			if(lowertext(R.next_char) == "H" && lowertext(R.next_next_char) == "A")
+		if("W")
+			if(R.next_char == "H" && R.next_next_char == "A")
 				new_string = "WO"
-				used = 3*/
+				used = 3
```
frawhst, line 1203: Are all these meant to be uncommented? I'm not generally familiar with the internals of the chav accent, but it seems like you're changing a lot of functionality by uncommenting a bunch of cases

frawhst: NEVERMIND I'm being silly

PR #24080, opened 2025-07-09
Doc comments on blob tile + blob overmind variables

MyNameIsRomayne:
## About the PR 
Documents what I was able to understand of the variables on the blob overmind and blob base object. Also adds a desmos graph into a part I found confusing and math-y!

## Why's this needed? 
Makes code more comprehensible for future contributors, and probably myself too once I forget what these mean in the next 6 hours.

## Testing 

N/A here I believe, but I did compile and run the game once. its doc comments, so there's no effect on game mechanics even if i tried.

### DIFF
--- a/code/mob/living/intangible/blob_overmind.dm
+++ b/code/mob/living/intangible/blob_overmind.dm
@@ -21,34 +21,39 @@ TYPEINFO(/mob/living/intangible/blob_overmind)
 	default_speech_output_channel = SAY_CHANNEL_BLOB
 
 	var/datum/tutorial_base/regional/blob/tutorial
-	var/attack_power = 1
-	var/bio_points = 0
-	var/bio_points_max = 1
-	var/bio_points_max_bonus = 7 //starting bio point cap should be 10-12 now, i think. a bit more wiggle room for starter blobs.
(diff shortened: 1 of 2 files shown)


code/mob/living/intangible/blob_overmind.dm
```diff
@@ -21,34 +21,39 @@ TYPEINFO(/mob/living/intangible/blob_overmind)
...
-	var/started = 0
-	var/starter_buff = 1
-	var/extra_nuclei = 0
-	var/next_extra_nucleus = 100
-	var/multi_spread = 0
-	var/upgrading = 0
+	var/started = FALSE //! Whether blob is pre or post deployment
+	var/starter_buff = TRUE //! Whether to give some assistance for blobs that may need a boost at the start
+	var/extra_nuclei = 0 //! Number of nuclei left to create
+	var/next_extra_nucleus = 100 //! Metric to how many total tiles are places until next nucleus can be placed. Scales
```
TobleroneSwordfish:
```suggestion
	var/next_extra_nucleus = 100 //! Metric of how many total tiles are placed until next nucleus can be placed. Scales
```

PR #24088, opened 2025-07-10
Standardizes mutantrace stutter system

snugglycactus:
## About the PR 
This PR changes the stutters of amphibians and cows (the only 2 other mutantraces to have stutters) to use the same system as the lizards, allowing them to capitalize properly. This could impact lizards somehow, but I'm fairly certain it doesn't.

## Why's this needed? 
Cohesion is good, and improperly capitalized mooing is an eyesore.

## Testing 
Screenshot 2025-07-10 001831
Screenshot 2025-07-10 002019

The prolonged Ms and Rs of the cows and amphibians now work properly. Rejoice!

Screenshot 2025-07-10 123933
Screenshot 2025-07-10 124204

Lizards and everyone else appear to be wholly unimpacted. Also rejoice!

### DIFF
--- a/code/modules/speech/modules/speech/modifiers/mutantraces/amphibian.dm
+++ b/code/modules/speech/modules/speech/modifiers/mutantraces/amphibian.dm
@@ -1,7 +1,16 @@
 /datum/speech_module/modifier/mutantrace/amphibian
 	id = SPEECH_MODIFIER_MUTANTRACE_AMPHIBIAN
+	var/static/regex/r_regex = regex(@"(r)(.?)", "ig")
 
 /datum/speech_module/modifier/mutantrace/amphibian/process(datum/say_message/message)
 	. = message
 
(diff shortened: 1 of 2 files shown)


Mister-Moriarty (approved):
The code looks good.

TobleroneSwordfish (changes requested):
Is it not possible to just update stutter to work like this instead? It seems weird to carve out a bunch of custom implementations when there's no reason we'd want stutter to not share this behaviour.

snugglycactus:
> Is it not possible to just update stutter to work like this instead? It seems weird to carve out a bunch of custom implementations when there's no reason we'd want stutter to not share this behaviour.

That does generally make sense. I just changed the exceptions that were already present, but it would be pertinent to change the whole system. I'll make this a draft real fast and get working on that.

Mister-Moriarty:
Forgot to say; this PR has been recreated in #24190, where mutantrace stuttering modules have been given a common parent.

PR #24092, opened 2025-07-10
Fix Spacebux Satchels

### LINKED ISSUE #24090: NT Backpack can't turn into an NT satchel anymore?
### Labels

[BUG][MINOR]

### Description

When I claimed the satchel reward, my NT backpack just turned into a regular satchel

### Steps to reproduce

1. Spawn in with NT backpack equipped
2. Claim the Satchel skin

### Expected Behavior

I expected the NT backpack to turn into an NT satchel

on a previous round as secoff, the same thing occured, where the bag just turned into a security satchel instead of an NT satchel

Reported by: honeykun7173
Client version: 516.1661
Active test merges: #23993, #23292, and #23422
Round log date: 2025-07-10-20-16
Reported on: 2025-07-10 20:24:32
Round ID: [208435](https://goonhub.com/rounds/208435)
Map: [DONUT2](https://goonhub.com/maps/donut2)

JORJ949:
## About the PR 
Refactors all backpack spacebux redeems to create the target backpack type, move contents and equip to the user's back, rather than just mimicking it. Does not do this if the user's backpack has more or less than 7 slots or they don't have a backpack in the first place.
Gives all itabags accurate colour names (was previously being set in the spacebux purchase and not on the type, noticable bug the above change would cause

## Why's this needed? 
#24002 means that the satchel redeem no longer checks icon state, and so the backpack's type needs to remain accurate. This seemed saner than just manually setting the satchel type on every redeemed purchase. This may also be the saner method for converting to satchels but thats for a future PR.
Fixes #24090 

## Testing 
Tested purchasing all available backpacks and converting them to satchels, items are transferred as expected, backpack correctly converts to satchel. Itabags and Rucksacks correctly randomise on each redeem.
Jobs that spawn with larger bags do not lose their larger bags, tested with Syndicate Grenadier and repeatedly respawning as Stowaway until I rolled the salvager bag, which didn't get replaced.
image

## Changelog 

```changelog
(u)JORJ949
(+)Fixed spacebux purchased backpacks not redeeming satchels properly.
```

### DIFF
--- a/code/modules/economy/persistent_bank_purchases.dm
+++ b/code/modules/economy/persistent_bank_purchases.dm
@@ -547,6 +547,8 @@ var/global/list/persistent_bank_purchaseables =	list(\
 				var/mob/living/carbon/human/H = M
 				if (H.back && istype(H.back, /obj/item/storage/backpack))
 					var/obj/item/storage/backpack/old_bag = H.back
+					if(old_bag.slots != 7) //No changing slot amounts
+						return 0
 					var/obj/item/storage/backpack/new_bag = new src.bag_type
 					//Delete anything in the new bag, replace with contents of old bag.
 					for (var/obj/item/I as anything in new_bag.storage.get_contents())


code/modules/economy/persistent_bank_purchases.dm
```diff
@@ -537,120 +538,96 @@ var/global/list/persistent_bank_purchaseables =	list(\
...
+			cost = 1600
+			icon_state = "bp_itabag_pink"
+			icon = 'icons/obj/items/storage.dmi'
+			bag_type = /obj/item/storage/backpack/itabag
+
+			Create(var/mob/living/M)
+				//Randomise each time a new one is made.
+				var/list/itabag_types = concrete_typesof(/obj/item/storage/backpack/itabag)
+				src.bag_type = pick(itabag_types)
+				. = ..()
```
ZeWaka, line 617: i don't see how this sets the icon state and icon, it's not handled in Create

JORJ949: The parent /backpack reward type Create() creates the bag_type, it doesn't change icon or icon state, just replaces the bag you are wearing with the bag you bought, if you have a bag that is 7 slots. So with the randomised ones we first randomise the bag type, then call the parent. 

I tested every purchase and it does work properly.

JORJ949: Basically, instead of just faking it by changing the icon state, your bag (if you have one and it has exactly 7 slots) is replaced with the one you purchased.

JORJ949: Lines 552-559 handle bag replacement in the parent /reward/backpack/Create()

PR #24117, opened 2025-07-13
Satchel Preference Option

JORJ949:
## About the PR 
Adds a new option in character setup to "Use Satchel", automatically converting any backpacks you spawn with into satchels.
Renames "PDA" section in character setup to "Equipment", classifying the new option under that.

## Why's this needed? 
Many satchel fans have to redeem the satchel reward at the start of every single round, this is a nice QoL to allow you to not have to wait for the big laggy rewards menu to load.

## Testing 
image

Works as expected when spawning myself as various roles, redeeming spacebux rewards, exporting, and persisting between rounds.

## Changelog 

```changelog
(u)JORJ949
(*)Added a new "Use Satchel" option to character setup to automatically turn the backpack you spawn with into a satchel!
```


tgui/packages/tgui/interfaces/CharacterPreferences/GeneralTab.tsx
```diff
@@ -184,7 +184,7 @@ export const GeneralTab = () => {
           </LabeledList.Item>
         </LabeledList>
       </Section>
-      <Section title="PDA">
+      <Section title="Equipment">
```
TobleroneSwordfish, line 187: Should probably rename ringtone and background colour to make it clear they're PDA options

JORJ949: image

TobleroneSwordfish (approved):

PR #24128, opened 2025-07-14
Rework plant yield

Egregorious:
## About the PR 

Reworked how plant yield affects produce output.

- every point of yield now increases the number of produce items in the harvest by 2%, rather than a flat extra produce item per point of yield 
- negative yield works similarly
- increased fertilisers effects on yield:
  - potash in the tray from 0.24 to 0.74
  - saltpetre in the tray from -0.24 to -0.5 (while above 1)
  - potash per infusion from ~2.5 to ~4
  - saltpetre per infusion from ~1 to ~-2.5
  - potash infusing lifespan from ~-1 to ~-0.5
- added crop count variance which defaults to a 70-100% 'consistency' modifier on output applied after most other modifiers, but before the harvest cap
- 'good harvests' increase the base output by 1 before yield multiplier, and improves consistency by an additional 10%
- 'bumper crop' increase the base crop count by 2 before yield multiplier, and improves consistency by an additional 20%
- 'crappy harvests' reduce base output by 40% before yield multiplier and reduces consistency by an additional 20%

## Why's this needed? 

Yield was simultaneously meaningless and far too effective, because at a rate of 1 point of yield to 1 extra produce the maximum output of any given harvest was reached with a less-than-trivial amount of scaling. The tiny amount of additional yield gained from infusing saltpetre a few times was enough to make every harvest hit the cap incidentally, and thus provide no incentive for botanists to ever bother optimising for yield specifically. This also contributed to botany being overly spammy, and health-bonuses during harvests being rather redundant. 
 
A percentage increase means high-base-output plants scale faster than low-base ones, which retains some personality of the plant's growth. It also means high-health bonuses tend to affect them more.

Saltpetre infusions are a popular way to make high quality plants quick, and it's more interesting if they come with a tradeoff. Saltpetre already had yield-inhibiting properties. Following up with potash infusions is still a way to make good quality, high-yield plants, it just takes more effort. 

The variance modifier is for increasing the space in which yield is useful. The usual harvest max of 10 leaves fairly little room for scaling. This way raising the yield past the harvest cap is still a little valuable for outscaling a poor variance roll. 

## Testing 

Tested growing a few different seeds, a mix of low-cropsize and high-cropsize plants. 

## Changelog 

```changelog
(u)Egregorious
(+)Plant yield has been reworked, it now contributes less to produce count per-point but is easier to increase. 
(+)Saltpetre infusions have a negative effect on yield. 
(+)Crop output has some natural variance, which is affected by the quality of the harvest.
```
[hydroponics][balance][add to wiki]

### DIFF
--- a/_std/defines/layer.dm
+++ b/_std/defines/layer.dm
@@ -1,3 +1,5 @@
+// AREA_LAYER - 1
+// TURF_LAYER - 2
 #define PLATING_LAYER     (TURF_LAYER-0.2)
 #define BETWEEN_FLOORS_LAYER     (TURF_LAYER-0.1)
 #define LATTICE_LAYER 		(TURF_LAYER+0.1)
@@ -18,6 +20,7 @@
 #define COG2_WINDOW_LAYER 	(TURF_LAYER+0.95)
 
 // More specific obj layers
+// OBJ_LAYER - 3
 #define SUB_TAG_LAYER		(OBJ_LAYER - 0.03) //! Graffiti that's been sprayed over sits here
 #define TAG_LAYER			(OBJ_LAYER - 0.02) //! Graffiti layer for gangs, this is the topmost (ie. most recent) tag on a turf
 #define STORAGE_LAYER		(OBJ_LAYER - 0.01) // Keep lockers etc below items
@@ -60,6 +63,8 @@
 #define MOB_LIMB_LAYER 		(MOB_OVERLAY_BASE-21)
 #define MOB_TAIL_LAYER2 		(MOB_OVERLAY_BASE-23) // Tail detail
 #define MOB_TAIL_LAYER1 		(MOB_OVERLAY_BASE-24) // Tail base
+
+// FLY_LAYER = 5
 
 // Some effects were defined on layer 10, some on layer 20... Lets unify this...
 // These are for effects that should display below lighting
@@ -108,3 +113,7 @@
 #define ASTEROID_TOP_OVERLAY_LAYER (TURF_LAYER+0.01)
 #define ASTEROID_ORE_OVERLAY_LAYER (TURF_LAYER+0.02)
 #define ASTEROID_MINING_SCAN_DECAL_LAYER (TURF_LAYER+0.03)
+
+// also these magic values
+// EFFECTS_LAYER - 5000
+// BACKGROUND_LAYER - 20000
--- a/code/obj/item/clothing/glasses.dm
+++ b/code/obj/item/clothing/glasses.dm
@@ -88,6 +88,7 @@
 		if ((isscrewingtool(I) || istype(I, /obj/item/pen)) && !src.pinhole)
 			src.pinhole = TRUE
 			src.block_vision = FALSE
+			setProperty("disorient_resist_eye", 2) // matches eyepatch with pinhole
 			boutput(user, SPAN_NOTICE("You poked two holes into the blindfold, now you can pretend that you can see without seeing"))
 		else
 			. = ..()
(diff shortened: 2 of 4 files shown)


NibChocolateeny:
It is a good idea - but wood want a small changing…if innfusing saltpetre, have remove rand(0,5) yield IF DNA.cropsize > 0, but if cropsize <= 0 then have it remove rand(0,1) yield. still have the panealty for infuse withe bad yield, but not so strong! Would be very notgood if try to get omeaga cannnabis and yielding is -180, bad also if needing super strong plants (like batttery ptotao). Still need to infues potash and mutadon when having big saltpetre infusing, but not so much. Also make it matter what odrer botanist is infuzing in - some stratgey

but yes for infuze with plant having posetive yield, strong penalty makes sense.

aslo say infusing saltpetre is bad for yield in changeinglog! it is important change.

Egregorious:
Personally I disagree with lessening the saltpetre infusion downside. There's much more room to scale downwards than upwards with yield, due to the necessarily low ceiling in crop count, so I think it's better to lean into that a bit. Potash tray-fertilising is high enough that it'll naturally fix low-medium negative yield in a somewhat timely manner already, and I quite like the idea of being stuck with a 1-crop plant for a little bit when said plant has been made very potent very quickly.

I actually think a large negative for Omega Weed is a good thing. Omega Weed shouldn't be as easy as it is to make honestly, it becomes far less special very quickly. If you're spamming saltpetre for Omega then 9 times out of 10 you've got it from cargo and thus you already have the potash to rectify the situation anyway.

NibChocolateeny:
i do not know……but it can be tried. if it is too bad, can cahnge peanalty to be not so strong in future.
Other changing is interesesting, want to try it out!

code/obj/machinery/plantpot.dm
```diff
@@ -1229,14 +1233,26 @@ TYPEINFO(/obj/machinery/plantpot)
 
 /// Handles the finalisation of cropcount
 /obj/machinery/plantpot/proc/HYPharvesting_finalise_cropcount(datum/HYPharvesting_data/h_data)
+	// only apply the yield multiplier and minimum if the count is likely to be positive, this allows prior modifiers to zero crop count without
+	// this overriding it back to 1.
+	var/cropsize = h_data.DNA?.get_effective_value("cropsize")
+	if (h_data.cropcount > 0)
+		h_data.cropcount *= (1 + ((cropsize * h_data.growing.yield_multi) / 100))
+		// It's too easy for low-output plants to have no cropcount at barely-negative yield values, so yield shouldn't be able to zero cropcount by itself.
+		h_data.cropcount = max(h_data.cropcount, 1)
```
Tarmunora: On the other hand, I think a _significantly_ negative yield should probably result in 0 produce - I think the `round(X, 1)` at the bottom should take care of any slightly negative values as well

Tarmunora (changes requested):

code/obj/machinery/plantpot.dm
```diff
@@ -1229,14 +1233,22 @@ TYPEINFO(/obj/machinery/plantpot)
 
 /// Handles the finalisation of cropcount
 /obj/machinery/plantpot/proc/HYPharvesting_finalise_cropcount(datum/HYPharvesting_data/h_data)
+	var/cropsize = h_data.DNA?.get_effective_value("cropsize")
+	if (cropsize > 0)
+		h_data.cropcount *= (1 + ((cropsize * h_data.growing.yield_multi) / 100))
```
Tarmunora: this just makes negative cropsize not affect yield at all?

Egregorious: Ah yes indeed! I did this far too late and managed to confuse myself

Egregorious:
Changed the way consistency is handled so that it better does the job it's supposed to. 

It chooses a random number between the minimum possible lower bounds and the uncapped cropcount. This means the more you increase yield the more likely it is you'll hit the harvest cap.

PR #24129, opened 2025-07-14
Configurable Radio Jammer range.

JORJ949:
## About the PR 
Allows the syndicate Radio Jammer's range to be configured with screwdriver/multitool between 1 (You and the tiles adjacent to you) and 6 (the current range of every signal jamming thing).

## Why's this needed? 
More precise jamming of radio communication allows for stealthier takedowns. The risk of them having an easier time getting out of range to send a communication vs the reward of not having someone randomly nearby note their radio being jammed allows for far more private and targeted signal jamming.

Example use: Leave a signal jammer at range 1 on a syndicate conversion chamber, only jamming signals from any victims currently inside or adjacent  but not signals from anyone passing through the hall/maints/medbay/wherever your crime den is

## Testing 
image
image

## Changelog 

```changelog
(u)JORJ949
(*)You can now edit the range on a syndicate signal jammer with either a screwing or pulsing tool, allowing for more precisely jammed crime dens.
```

### DIFF
--- a/_std/packets.dm
+++ b/_std/packets.dm
@@ -1,10 +1,10 @@
 
-#define RADIO_JAMMER_RANGE 6
+#define DEFAULT_RADIO_JAMMER_RANGE 6
 
 proc/check_for_radio_jammers(atom/source)
 	. = FALSE
 	for (var/atom/A as anything in by_cat[TR_CAT_RADIO_JAMMERS])
-		var/range = RADIO_JAMMER_RANGE
+		var/range = DEFAULT_RADIO_JAMMER_RANGE
 		if(istype(A, /obj/item/radiojammer))
 			var/obj/item/radiojammer/jammer = A
 			range = jammer.range
--- a/_std/packets.dm
+++ b/_std/packets.dm
@@ -4,7 +4,11 @@
 proc/check_for_radio_jammers(atom/source)
 	. = FALSE
 	for (var/atom/A as anything in by_cat[TR_CAT_RADIO_JAMMERS])
-		if (IN_RANGE(source, A, RADIO_JAMMER_RANGE))
+		var/range = RADIO_JAMMER_RANGE
+		if(istype(A, /obj/item/radiojammer))
+			var/obj/item/radiojammer/jammer = A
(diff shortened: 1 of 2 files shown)


_std/packets.dm
```diff
@@ -4,7 +4,11 @@
 proc/check_for_radio_jammers(atom/source)
 	. = FALSE
 	for (var/atom/A as anything in by_cat[TR_CAT_RADIO_JAMMERS])
-		if (IN_RANGE(source, A, RADIO_JAMMER_RANGE))
+		var/range = RADIO_JAMMER_RANGE
```
frawhst: I'd like this renamed— because of this PR, the const doesn't really represent the Universal Radio Jammer Range anymore, so `DEFAULT_RADIO_JAMMER_RANGE` or smth would be better. Alternatively go through and give each radio jamming source their own place to store the const (since I don't know why all radio jammers use the same number).

code/obj/item/device/radios/_radio.dm
```diff
@@ -371,10 +376,21 @@ TYPEINFO(/obj/item/radiojammer)
...
 
+/obj/item/radiojammer/proc/edit_range(mob/user)
+	var/inputted_number = tgui_input_number(user, "Input radio jammer range", "Radio Jammer", RADIO_JAMMER_RANGE, RADIO_JAMMER_RANGE, 1)
+	if(!inputted_number)
+		return
+	inputted_number = trunc(inputted_number)
+	if(!isnum_safe(inputted_number) || inputted_number > RADIO_JAMMER_RANGE || inputted_number < 1)
+		boutput(user, SPAN_ALERT("That number is out of [src]'s range!"))
+		return
+	src.range = inputted_number
```
frawhst: probably good to have a safety distance/`can_act()` check on this— otherwise you can open a few inputs and remotely set the radius from anywhere (not a huge issue, but feels buggy).

_std/packets.dm
JORJ949: Radio jammers should _probably_ be componentised, but I don't want to do that in this PR.

PR #24131, opened 2025-07-15
Removes stupidity from OD linting

ike709:
Imagine being silly enough to bump `#warn` to an error.

Hasn't been caught sooner because I just haven't run goon locally in OD in years.


ike709:
@ZeWaka

PR #24141, opened 2025-07-16
[TGUI Core] fix for vending machine padding

RubberRats:
## About the PR 

Fixes the padding for icons in the vending machine interface, as well as other interfaces that use the `ProductList` component.
NOTE: I wasn't able to build the project in production mode for reasons I was unable to properly diagnose. I verified this change works using the TGUI dev server.

## Why's this needed? 

It was one of the requested UI changes for the update to BYOND 516.

## Testing 
image


ZeWaka:
incorporated in the branch

PR #24142, opened 2025-07-16
Pickpocket gun steals headsets -> masks -> glasses

TobleroneSwordfish:
## About the PR 
When targeting the head, the pickpocket gun now first steals headsets, then mask if no headset, then glasses if no mask. It can also now equip headsets, glasses and masks when set to "plant".
Old behaviour was only stealing the mask.
## Why's this needed? 
Make the pickpocket gun more useful, grabbing headsets could be a neat way to silence people without using a jammer. Technically makes it harder to yoink someone's mask in order to gas them buut I don't think I've ever seen this done in the history of ever.
## Testing 
image
image

Tested stealing and planting all three.

## Changelog 

```changelog
(u)LeahTheTech
(+)The Super! Grapple Friend pickpocket gun can now steal headsets, masks and glasses (in order) and plant them back.
```

### DIFF
--- a/code/modules/projectiles/pickpocket.dm
+++ b/code/modules/projectiles/pickpocket.dm
@@ -38,7 +38,7 @@
 				if ("chest")
 					if (M.wear_id) stolenItem = M.wear_id
 				if ("head")
-					if (M.ears) stolenItem = M.ears
+					stolenItem = M.ears || M.wear_mask || M.glasses
 				if ("r_arm")
 					stolenItem = M.r_hand ? M.r_hand : M.l_hand ? M.l_hand : null
 				if ("l_arm")
--- a/code/modules/projectiles/pickpocket.dm
+++ b/code/modules/projectiles/pickpocket.dm
@@ -38,7 +38,7 @@
 				if ("chest")
 					if (M.wear_id) stolenItem = M.wear_id
 				if ("head")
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/pickpocket.dm
```diff
@@ -59,52 +59,55 @@
...
+			if ("l_leg")
+				if (!M.l_store && M.can_equip(linkedGun.heldItem, SLOT_L_STORE))
+					M.equip_if_possible(linkedGun.heldItem, SLOT_L_STORE)
+				else if (!M.r_store && M.can_equip(linkedGun.heldItem, SLOT_R_STORE))
+					M.equip_if_possible(linkedGun.heldItem, SLOT_R_STORE)
+				else // Couldn't go into a pocket, dump on ground
+					boutput(M, "\A [linkedGun.heldItem] tries to cram itself into your pockets! [pick(strikeFlavor)]")
+
+	on_hit(atom/hit)
+		if (!linkedGun.heldItem) // Stupidity check for stripping item out of gun while projectile was in flight
```
JORJ949, line 95: I don't think you need to check if its still held? The gun itself stores whatever its about to plant/steal in itself and if you shoot it then drop it I see no reason it wouldn't still work as it grabs and pulls or plants the item.

PR #24157, opened 2025-07-18
Add Wheelchairs to Medfabs

JORJ949:
## About the PR 
Adds wheelchair parts to the medical fabricator for .8 metal and .3 fabric
## Why's this needed? 
Destroying/hiding the spare wheelchairs in medbay currently means you can't make more unless they were mechscanned, you shouldn't need to get legs for your wheelchair character if someone steals all the chairs

## Testing 
image

## Changelog 

```changelog
(u)JORJ949
(+)Wheelchair parts can now be crafted at the medical fabricator
```

### DIFF
--- a/code/datums/manufacturing.dm
+++ b/code/datums/manufacturing.dm
@@ -1156,6 +1156,15 @@
 	time = 20 SECONDS
 	category = "Clothing"
 
+/datum/manufacture/wheelchair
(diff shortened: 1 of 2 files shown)


pgmzeta:
weird that it is mech-scannable but that wasn't added in this PR so

PR #24159, opened 2025-07-18
Adds rare antag token drops for low population servers

TobleroneSwordfish:
## About the PR 
At the end of a round, all players who have participated will have a small chance to be awarded a free antag token, provided they don't already have one.
The chance maxes out at 10% and scales down with a couple of factors
- The population imbalance between this server and the highest population Goonstation server. For example, if the highest pop server has 60 people and the one you're on only has 10 you'll have a (relatively) high roll chance.
- The amount of time you've been connected to the round, your chance will be reduced proportionally based on how much of the round you weren't connected for, so logging in for 10 minutes at the end will yield a lower chance than playing the entire round.
- If the ratio between the highest population server and this one is > 0.6 (ie 36/60) there will be no drop chance at all.

TODO: Handle Nightshade separately?
## Why's this needed? 
There's a lot of reasons why people end up flocking to the highest population server and this doesn't pretend to fix any/all of them, buuut it might help a bit! It's a small incentive to try out a lower population server, especially as your antag token can then be spent on any server you like.
Will this lead to 16 bajillion antags every round? Idk! Hopefully not, since inherently the people rolling a high chance to get a token will be a very small number due to being on the lower population server.
Will this abused? I'm sure some people will try to, but it's quite hard to abuse as written and we can as usual take admin action against people obviously fishing (ie joining a round just to go AFK the whole time) in the same way we do against regular antag fishing behaviour.
## Testing 
Due to its nature this pretty much has to be tested on live. Testmerge!

## Changelog 

```changelog
(u)LeahTheTech
(*)Playing on one of the less populated servers will now give you a small chance to earn an antag token at the end of each round.
```

### DIFF
--- a/code/datums/gameticker.dm
+++ b/code/datums/gameticker.dm
@@ -998,6 +998,9 @@ var/global/game_force_started = FALSE
 		logTheThing(LOG_DEBUG, null, "Population ratio with largest server: [pop_ratio], greater than threshold [RATIO_THRESHOLD], aborting end round drops.")
 		return // servers are reasonably balanced, no drops for u
 
+	if (global.round_elapsed_ticks < 10 MINUTES) //too short, possibly just an admin restarting the server to change map etc.
+		return
+
 	var/actual_token_chance = BASE_TOKEN_CHANCE * ((RATIO_THRESHOLD - pop_ratio) / RATIO_THRESHOLD)
 	logTheThing(LOG_DEBUG, null, "Population ratio with largest server: [pop_ratio], starting end round drops with token chance [actual_token_chance]")
 
--- a/code/datums/gameticker.dm
+++ b/code/datums/gameticker.dm
@@ -918,6 +918,10 @@ var/global/game_force_started = FALSE
 		cloud_saves_put_data_bulk(bulk_commit)
 		logTheThing(LOG_DEBUG, null, "Done with spacebux")
 
+	logTheThing(LOG_DEBUG, null, "Starting round end drops...")
+	src.do_roundend_drops()
+	logTheThing(LOG_DEBUG, null, "Done with round end drops")
+
 	for_by_tcl(P, /obj/bookshelf/persistent) //make the bookshelf save its contents
 		P.build_curr_contents()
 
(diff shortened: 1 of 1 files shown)


JORJ949:
Should probably additionally scale chance based on:

1. How long the round was (discourage early calling for more tokens, or killing everyone at minute 0 to end the round)
2. The actual number difference rather than just ratio, 5 vs 10 pop's not a big difference, 30 vs 60 is.

TobleroneSwordfish:
Round length scaling added. Scaling by the actual player difference would mess up all my maths but I've added a threshold so if the highest pop server has 20 or less people the token drops won't trigger at all (mostly will only happen during nightshade spindown/EU-NA dead hour overlaps)

code/datums/gameticker.dm
```diff
@@ -975,4 +979,59 @@ var/global/game_force_started = FALSE
...
+		logTheThing(LOG_DEBUG, null, "All linked servers below population threshold, no round end drops.")
+		return
+
+	var/pop_ratio = self.player_count / highest_pop.player_count
+	if (pop_ratio > RATIO_THRESHOLD)
+		logTheThing(LOG_DEBUG, null, "Population ratio with largest server: [pop_ratio], greater than threshold [RATIO_THRESHOLD], aborting end round drops.")
+		return // servers are reasonably balanced, no drops for u
+
+	//scale by round length, normalised at 60 minute rounds
+	var/length_ratio = src.round_elapsed_ticks/(60 MINUTES)
```
JORJ949: Needs flattening to 1 if the round goes over 60 minutes to keep your maximum cap of 10%

TobleroneSwordfish: Mmmm I did think that was fine but actually let's cap it at 50 = full points and leave it there

JORJ949: Well I more meant if you want it to be THE maximum then it needs clamping, or if its not the maximum to update the comment, but that may cause overextension of rounds (current maximum round timer being about 160 minutes) to increase the maximum to 30ish %

code/datums/gameticker.dm
```diff
@@ -975,4 +979,62 @@ var/global/game_force_started = FALSE
...
+	var/length_ratio = min(src.round_elapsed_ticks/(50 MINUTES), 1)
+
+	var/actual_token_chance = BASE_TOKEN_CHANCE * ((RATIO_THRESHOLD - pop_ratio) / RATIO_THRESHOLD) * length_ratio
+	logTheThing(LOG_DEBUG, null, "Population ratio with largest server: [pop_ratio], starting end round drops with token chance [actual_token_chance]")
+
+	var/list/players = list()
+	for (var/mob/M as anything in mobs)
+		var/datum/player/player = M.mind.get_player()
+		if (player in players)
+			continue
```
Mister-Moriarty, line 1027: Not a full review, but why not `for (var/client/C as anything in clients)`?

code/datums/gameticker.dm
```diff
@@ -975,4 +979,62 @@ var/global/game_force_started = FALSE
...
+		players += player
+		//also scale by how long they've been in the round, no joining at the end just for the chance
+		var/time_ratio = player.current_playtime / ticker.round_elapsed_ticks
+		var/player_unique_chance = actual_token_chance * time_ratio
+		if (!prob(player_unique_chance))
+			continue //unlucky
+		if (player.get_antag_tokens() >= 1)
+			boutput(M, SPAN_BOLD("You would have received an antag token drop this round, but you already have one!"))
```
Mister-Moriarty: I feel that giving players this information is just going to lead to them becoming frustrated.

code/datums/gameticker.dm
TobleroneSwordfish: Players might be disconnected at the end, but still played an almost full round

code/datums/gameticker.dm
TobleroneSwordfish: Orr making them less frustrated because they realise why they're not getting drops and spend their tokens accordingly.

code/datums/gameticker.dm
JORJ949: May be better to iterate over minds then? Players can lose all their mobs (Gibbed+Disconnected(I THINK disconnecting makes your ghost gone) and players having multiple minds is already caught by the if player in players

frawhst: Disconnecting makes (should make) your ghost invisible, but it's still there and shouldn't be deleted unless something weird happens.

code/datums/gameticker.dm
frawhst: Would prefer to just not roll on the player at all, and give them something along the lines of "You were not eligible to receive an antag token for lowpop participation [or w/e we call it] because you already have one". Telling people "you COULD have had Loot just now, but you don't get it!!" could easily be received poorly, especially for people who are saving a token/tokens for a particular round or time (and are thus comfortable with not getting the lowpop freebies), who would have to endure a lot of "you could've won a prize!!! but you didn't" messages.

Edit: But then that skews it so only people without tokens are in the 'pool', which fucks up the odds a bit. Yeah, I'd prefer to just give them the 'you aren't eligible' message as a reminder and leave it at that.

Edit 2: Ok now having read the code, I didn't realize it was an individual roll rather than a 'pick X players' deal. Not rolling and displaying a message on ineligible people seems fine.

NibChocolateeny:
@TobleroneSwordfish - can do the minor changing from #24330 !

PR #24165, opened 2025-07-19
ID Cards suppress attacks

JORJ949:
## About the PR 
Adds the SUPPRESSATTACK flag to ID Cards, stopping them from attacking most (some objects have custom attack procs that don't check SUPPRESSATTACK) objects. This specifically applies to Agent IDs attacking other IDs which currently show an attack message in chat, when attacking the Agent ID with other IDs doesn't do this.

## Why's this needed? 
Prevents agent IDs from showing an attack message in chat when scanning another ID.
ID cards already do 0 damage and theres a lot of stuff you can vs can't scan your ID on, most things already don't have you bash your ID on it and there's no reason for you to want to bash a plasma tank with your ID anyway, used here as an example as its the only other difference I could find in devtest that shows a bash animation and not just a message in chat of you "attacking" it.

## Testing 
Here you can see the agent ID not having an attack message in chat
image

### DIFF
--- a/code/obj/item/card.dm
+++ b/code/obj/item/card.dm
@@ -85,7 +85,7 @@ TYPEINFO(/obj/item/card/emag)
 	icon_state = "id_basic"
 	item_state = "card-id"
 	desc = "A standardized NanoTrasen microchipped identification card that contains data that is scanned when attempting to access various doors and computers."
-	flags = TABLEPASS | ATTACK_SELF_DELAY
+	flags = TABLEPASS | ATTACK_SELF_DELAY | SUPPRESSATTACK
 	click_delay = 0.4 SECONDS
 	wear_layer = MOB_BELT_LAYER
 	var/datum/pronouns/pronouns = null


TobleroneSwordfish (approved):

MaxLey:
Might be worth doing for PDA's too? There's lots of things they scan and with the way they are you just sort of regularly end up bashing a machine or person you wanted to use a particular function on if it either fails or you dont have the program loaded correctly.

Spyritdragon:
Might be worth doing for PDA's too? There's lots of things they scan and with the way they are you just sort of regularly end up bashing a machine or person you wanted to use a particular function on if it either fails or you dont have the program loaded correctly.

JORJ949:
> Might be worth doing for PDA's too? There's lots of things they scan and with the way they are you just sort of regularly end up bashing a machine or person you wanted to use a particular function on if it either fails or you dont have the program loaded correctly. 

Probably, but that'll be a seperate pr

PR #24167, opened 2025-07-19
Cryopods heal and consume power more consistently

### LINKED ISSUE #24162: Organ Damage Doesn't Heal with Cryopod Auto-Eject Off
### Labels

[BUG][MINOR]

### Description

A miner came in with heavy organ damage from multiple nanite attacks. Healed them fully then shoved them into the cryopod. Turned auto-eject off. They sat inside, not healing. Applied medical fire extinguisher to dome. With damage to heal, organs also healed.

### Steps to reproduce

1. Heal person with organ damage fully.
2. Shove into cryo.
3. Observe organs not healing.
4. Damage them. Shove them back in.
5. Organs heal so long as they heal BRUTE OR BURN (or presumably TOX)! Wow!

### Expected Behavior

The cryo autoeject toggle was added as a way to keep organ damaged people inside. A cool change! Seems counterproductive that the addition doesn't work, and organs only heal while healing damage. Which makes them stay inside anyway.

Reported by: luminousaether
Client version: 515.1646
Active test merges: #23993 and #23292
Round log date: 2025-07-19-00-38
Reported on: 2025-07-19 01:22:20
Round ID: [208947](https://goonhub.com/rounds/208947)

Egregorious:
## About the PR 

Cryopods always process occupants when they're not dead or ejected for any reason. 
Cryopods always consume power when occupied.

## Why's this needed? 

Cryopods heal organs when occupants are at full health: fixes #24162

Little more consistent power drain just means a bit more opportunity for players to cause issues. 

## Testing 

Got a monkey drunk then healed its liver:
cryo_test

### DIFF
--- a/code/modules/atmospherics/machinery/unary/cryo_cell.dm
+++ b/code/modules/atmospherics/machinery/unary/cryo_cell.dm
@@ -53,16 +53,14 @@ TYPEINFO(/obj/machinery/atmospherics/unary/cryo_cell)
 		return
 
 	if(src.occupant)
-		if(!isdead(src.occupant))
-			if (!ishuman(src.occupant))
-				src.go_out() // stop turning into cyborgs thanks
-			if (src.occupant.health < src.occupant.max_health || src.occupant.bioHolder.HasEffect("premature_clone"))
-				src.use_power(src.occupied_power_use, EQUIP)
-				src.process_occupant()
-			else
-				if(src.occupant.mind && src.eject_full_health_occupant)
-					src.go_out()
-					playsound(src.loc, 'sound/machines/ding.ogg', 50, 1)
+		if (!ishuman(src.occupant))
+			src.go_out() // stop turning into cyborgs thanks
+		else if ((src.occupant.health >= src.occupant.max_health) && src.eject_full_health_occupant)
+			src.go_out()
+			playsound(src.loc, 'sound/machines/ding.ogg', 50, 1)
+		else if (!isdead(src.occupant))
+			src.process_occupant()
+		src.use_power(src.occupied_power_use, EQUIP)
 
 	if(src.air_contents)
 		src.ARCHIVED(temperature) = src.air_contents.temperature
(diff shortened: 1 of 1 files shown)


code/modules/atmospherics/machinery/unary/cryo_cell.dm
```diff
@@ -53,16 +53,14 @@ TYPEINFO(/obj/machinery/atmospherics/unary/cryo_cell)
...
-					src.go_out()
-					playsound(src.loc, 'sound/machines/ding.ogg', 50, 1)
+		if (!ishuman(src.occupant))
+			src.go_out() // stop turning into cyborgs thanks
+		else if (src.occupant.health >= src.occupant.max_health && src.eject_full_health_occupant)
+			src.go_out()
+			playsound(src.loc, 'sound/machines/ding.ogg', 50, 1)
+		else if (!isdead(src.occupant))
+			src.process_occupant()
+		src.use_power(src.occupied_power_use, EQUIP)
```
TobleroneSwordfish, line 63: Should probably rename this var to indicate it's not just when occupied

TobleroneSwordfish (changes requested):

code/modules/atmospherics/machinery/unary/cryo_cell.dm
```diff
@@ -53,16 +53,14 @@ TYPEINFO(/obj/machinery/atmospherics/unary/cryo_cell)
 		return
 
 	if(src.occupant)
-		if(!isdead(src.occupant))
-			if (!ishuman(src.occupant))
-				src.go_out() // stop turning into cyborgs thanks
-			if (src.occupant.health < src.occupant.max_health || src.occupant.bioHolder.HasEffect("premature_clone"))
```
JORJ949, line 59: You've completely removed the check for premature clone counting as "not full health" for the purpose of autoejecting/processing.

Egregorious: That was unintentional for autoejecting, cheers for noticing, I was only thinking in terms of processing. The processing should be fine, everything should get processed unless dead or ejected.

code/modules/atmospherics/machinery/unary/cryo_cell.dm
Egregorious: It should still be exclusively when occupied, it's just the exceptions that have been removed. 

Personally I consider it occupied even when they're ejected, since it simulates occupation prior to the process tick occurring.

PR #24188, opened 2025-07-23
Adds "Lay Egg Is TRUE!!!" radio station record

ribbiestar:
[GAME OBJECTS]

## About the PR 
Adds a new track to funk category radio station records, "Lay Egg Is TRUE!!!" by WyrdDoe

## Why's this needed? 
music nice, and banger

## Testing 

https://github.com/user-attachments/assets/cdf0b5e8-7f86-448b-9ad6-1261e40ba546

Name has been changed to "Lay Egg Is TRUE!!!" since recording.
The audio file is 828 kbs in size. Can be lowered to 572 kbs if made mono but loses too much quality.

## Changelog 

```changelog
(u)WyrdDoe
(+)Added a new radio station record, Lay Egg is TRUE!!!
```


itchybones:
It's a good song, but I think it could get very old, very fast, considering the relatively high chances of getting one in a record sleeve. They would make for a fun robo-duck item drop, though.

TobleroneSwordfish (changes requested):
I agree with Itchy, make it a rare drop from the annoying duck eggs. It's a funny song, but not one you want to hear every round (also it's more special if it's a little rare)

PR #24190, opened 2025-07-23
Say Rework HTML Sanitisation Fix

### LINKED ISSUE #23471: [SAY REWORK] she's not sanitising html tags cap'n
### Labels

_No response_

### Description

Any whispers that aren't fully picked up by a listener can have unsanitised HTML tags garbled in 'em.

### Steps to reproduce

1. Have two actors, at least one tile apart.
2. One of them whispers.
3. The listener gets a faceful of HTML tags if the formatting of the heard message is different from normal text.

### Expected Behavior

Heard speech should be sanitised.

Specifically, borgs enclosing text with ` characters creates a span with a different font family than normal speech. This may be true of other accents or speech effects.

Image

### LINKED ISSUE #23597: Apostrophe messing up while being drunk
### Labels

[P-Minor]

### Description

While drunk or having the trait career alcoholic, in chat, typing aposthrophe ( ' ) sometime turns it into weird numbers or asterisks

### Steps to reproduce

1. Open any text option
2. Write anything with an aposthrophe
3. Look at the message displayed from your character

### Expected Behavior

I expected it to not change or be affected by the speech change

_No response_

### LINKED ISSUE #24087: Scottish doesn't change contractions into the accent versions
### Labels

[BUG][MINOR]

### Description

As above, scottish is bugged and doesn't change contractions, like don't into dinnae etc.

### Steps to reproduce

Have scottish accent
Say a contraction
The contraction is unchanged, even though it should be

### Expected Behavior

Scottish accent should work and change the word

Reported by: elskrises
Client version: 515.1643
Active test merges: #23993, #23292, and #23422
Round log date: 2025-07-10-11-47
Reported on: 2025-07-10 13:21:16
Round ID: [208412](https://goonhub.com/rounds/208412)
Map: [DONUT2](https://goonhub.com/maps/donut2)

### LINKED ISSUE #24112: Career Alcoholic Speech is jank
### Labels

[BUG][MINOR]

### Description

using ' and the like REALLY messes with your speech if you have carreer alcohlic trait.

### Steps to reproduce

talk with career alcohloic with like a ' or something

### Expected Behavior

normal. ish. speech. not any $*$^**^ or something.

Reported by: emeraldcrow
Client version: 515.1633
Active test merges: #23993 and #23292
Round log date: 2025-07-12-23-34
Reported on: 2025-07-13 00:30:17
Round ID: [208568](https://goonhub.com/rounds/208568)

### LINKED ISSUE #24130: AI radio whispers garble monospace styling
### Labels

_No response_

### Description

When hearing an AI core (or presumably other silicon) whispering and using the "``" monospace styling, garbled html tags are visible.

### Steps to reproduce

1. Stand next to an AI core
2. Have the AI core use monospace styling on a radio channel you don't have access to
3. `<span>` moment

### Expected Behavior

I expected the whisper **s to apply only to the message content and not the styling tags

Image

### LINKED ISSUE #24146: Drunk slurring encoding error only with overhead chat
### Labels

_No response_

### Description

Apostrophes get URL encoded in overhead chat if someone slurs them while drunk

### Steps to reproduce

1. Be drunk enough to slur your words
2. Say "I'm drunk"
3. Apostrophe error in overhead chat (actual chatbox is fine)

### Expected Behavior

I expected no encoding

_No response_

Mister-Moriarty:
[Bug] [Rework]

## About The PR:
Reworks say message datum content modification to allow for non-player HTML to be ignored. This is achieved through the use of pseudo-HTML `<mutable>` tags, which mark a portion of a message as mutable.

E.g.
```
"<mutable>The </mutable><i><mutable>quick</mutable></i><mutable> brown fox jumps over the lazy dog.</mutable>"
```

A special `ReplaceWithCallback` regex proc is then used to select and apply a proc through a callback datum to only the mutable portions of the message content.

The example above would then become the following under a `CALLBACK(GLOBAL_PROC, GLOBAL_PROC_REF(stars), 100)` callback datum:
```
"<mutable>*** </mutable><i><mutable>*****</mutable></i><mutable> ***** *** ***** **** *** **** ****</mutable>"
```

Fixes #23471, fixes #23597, fixes #24087, fixes #24112, fixes #24130, and fixes #24146.

## Testing:
Testing every single speech module under the new system would take an impossibly long time to test, so instead each unique pattern was tested (i.e. global procs, object procs, procs with args, etc.). Regardless, I would recommend testmerging this in the event that specific combinations of modifiers break the system.


code/modules/speech/modules/speech/modifiers/flock_gradient.dm
```diff
@@ -5,4 +5,48 @@
 	. = message
 
 	message.flags &= ~SAYFLAG_HAS_QUOTATION_MARKS
-	message.content = gradientText("#3cb5a3", "#124e43", "\"[message.content]\"")
+	message.content = "[MAKE_CONTENT_MUTABLE("\"")][message.content][MAKE_CONTENT_MUTABLE("\"")]"
+	APPLY_CALLBACK_TO_MESSAGE_CONTENT(message, CALLBACK(GLOBAL_PROC, GLOBAL_PROC_REF(immutable_gradientText), "#3cb5a3", "#124e43"))
+
+
+/// A copy of `gradientText` that ensures that the HTML content is immutable.
+/proc/immutable_gradientText(message, color_1, color_2)
```
TobleroneSwordfish, line 13: Why not just a param?

TobleroneSwordfish (changes requested):
It's mildly cursed that we have to do something like this, and I'm a little concerned that we're going to end up with some `<mutable><mutable><mutable>` nested madness in future but this does work well for now and the interface is sane.

code/modules/speech/modules/speech/modifiers/flock_gradient.dm
Mister-Moriarty: Because the `gradientText` arguments are in the following order: `color_1, color_2, message`; the regex expects `message` to be the first argument.

Mister-Moriarty:
> It's mildly cursed that we have to do something like this, and I'm a little concerned that we're going to end up with some `<mutable><mutable><mutable>` nested madness in future but this does work well for now and the interface is sane.

The regex doesn't like nested tags, so hopefully people will see that their code is broken and not do that? I don't think this problem has any other solutions that tick all the required boxes.

code/modules/speech/modules/speech/modifiers/flock_gradient.dm
TobleroneSwordfish: Wrapper then? I am allergic to copy paste :P

PR #24194, opened 2025-07-24
Cloning rack

TobleroneSwordfish:
## About the PR 
Removes pre-scan storage from the cloning console, all clone scans are now stored on physical floppy disks.

The cloner can now only make a scan onto a disk or clone someone from an inserted disk. Scanning onto a disk automatically labels the disk with the person's name, this name can be changed with a labeler.

To manage the disks, there is now a disk rack in every clone bay. Disk racks will only display the name the disk is currently labelled with, which can be changed with a hand labeller.

Cloner implants now send constant "heartbeat" packets while they're implanted in a living person. Those packets are then picked up by listening cloning racks, and if a cloning rack hasn't received a heartbeat signal in the last 10 seconds it will begin flashing a warning light next to the corresponding disk.
TL;DR: red means dead!

TODO: 
- [x] Mapping changes (include disk box and labeler)
- [x] Display when disk already has a clone record on the console
- [x] Mechscanning of disk racks
- [x] Disk fabrication (apparently medical already do this)
- [x] Disk wiping (apparently you can already do this on terminals, maaaybe best to have an easier way but shrug)
- [x] Make warning lights depend on cloning implants
- [x] Deluxe cloning kit
- [x] Handle rack being out of power
- [x] Pen labelling for disks?
- [ ] Old school floppy drive seeking sounds?
- [x] Emag effect for rack?
- [ ] Let genetics mess with clone records? Activate genes, change appearance etc? Maybe future PR.
- [x] The manual, how did I forgor this

Sound effect taken from https://pixabay.com/sound-effects/c64-remove-and-insert-floppy-disk-2-takes-each-234602/ under the standard pixabay license (compatible)

## Why's this needed? 
- Limits pre-scans to 8 slots that can be actively monitored by default (more if engineering build you additional racks)
- Makes cloning a little more involved of a process, actually having to manage and keep track of physical items
- CRIME - disks can be stolen, lost, or re-labelled

## Testing 
Modified Atlas cloning layout:
image
image
Cloning rack with disks in and little flashing light:
image

## Changelog 

```changelog
(u)LeahTheTech + Glowbold
(*)The cloning console now stores all its records on floppy disks, two racks of which can be found in cloning.
(+)Disk racks can be placed side-by-side by clicking on them with a frame.
(+)Cloner implants send a heartbeat signal back to the cloning rack, if the implant is removed or the person dies the rack will display a flashing red light next to their disk.
(+)Floppy disks can be labelled with a pen as well as hand labellers.
```


NibChocolateeny:
wait! Queshtion: I do not think clone disk is erased if useing, so if clone person then can use saem disk more time for same "clone qualitey." So worser and worser clonering defect not happen now, withthe diskes? Maybe i am confused...

RSOD404:
i'm guessing cloning from disks would erase the data on them?

TobleroneSwordfish:
Yep, cloning from a disk wipes the cloning record off it

trazodont:
O: this is awesome

Egregorious:
I love adding physicality to cloning. 

I don't like the change to genetics. It ends up removing the ability to make one of the most uninteractive jobs a bit more interactive. I enjoy looking through the samples then bringing the crew activators based on what I find. Admittedly I don't see other people doing that but personally I'd still want to lean into it more and not less. 

The research matters so little even in terms of genetics gameplay. It might be less than ideal as it stands but I personally care less about internal genetics balance than I do about how much it encourages crew interaction, and I think this ends up going in the wrong direction and just making it harder.

Razzberry8492:
> Emag effect for rack?

Perhaps emagging the rack adds mutations to all disks in the rack - if you get cloned from an emagged disk, you come out with a few mutations. It can be good mutations, bad mutations, or just gimmick mutations.
Mutations from this are counted as if it was done with DNA injectors - genes that are not in Potential affect your Stability.
Disk emag statuses are tracked - a disk cannot be emagged twice even if it is taken out of the rack and put in a different one.

The simplest thing to do is get everyone rescanned, or just endure the mutation when cloned and grab some mutadone for it.

RSOD404:
what if one (or more) of the mutations from emagging a clone rack were reinforced? makes it more disruptive

Gannets:
I love physical media

code/obj/machinery/clone_rack.dm
```diff
@@ -62,7 +62,7 @@ TYPEINFO(/obj/machinery/disk_rack)
 		var/index = text2num_safe(params["id"])
 		if (index < 1 || index > MAX_DISKS) //nu
 			return FALSE
-		if (BOUNDS_DIST(ui.user, src) && !isAIeye(ui.user))
+		if (BOUNDS_DIST(ui.user, src) || isAIeye(ui.user))
```
JORJ949: Should probably be a generic isintangible check rather than specifically the AI eye

PR #24213, opened 2025-07-26
Syndicate ID Dispenser

JORJ949:
## About the PR 
Adds the Syndicate Operative ID dispenser to all listening posts, an ID dispenser that dispenses a syndicate brand (Non-Agent) ID card with randomised operative codename to trained syndies, comes with Maintenance and Syndicate access. Operative codenames are randomised once then always the same for the same operative on further redeems.

## Why's this needed? 
Allows various syndies, including sleeper agents, to have a syndicate codename that they can use the syndicate announcement computer (or purchase a voice changer from CARL) with without having to spend 1 TC that you may not even have (sleeper agents) on an agent card. Agent Cards remain useful as they allow a custom agent name, can copy accesses, and block AI tracking. This is mainly an RP focused change as it allows for creating a more general sense of dread while providing no real benefits beyond maintenance access.

## Testing 
image

## Changelog 

```changelog
(u)JORJ949
(*)Syndicate ID dispensers can now be found on all listening posts, these vend operatives randomised syndicate brand (not agent card) IDs, perfect for making sinister announcements with.
```


TobleroneSwordfish:
Ehh, honestly I'd rather not lean into the whole "go to the listening post and get dressed up" thing even harder.

JORJ949:
> Ehh, honestly I'd rather not lean into the whole "go to the listening post and get dressed up" thing even harder.

I do think getting dressed up from on station should be possible, perhaps adding the syndicate gear to the chameleon suit by default, but this is a general post issue that can be solved other ways (portable announcement computer, remote merchant deals, etc)

TobleroneSwordfish:
Okay I'm closing this, reasons:
- Too confusing between this and the agent card
- Leaning a little too heavily into encouraging overt "agent" style syndies, this is meant to be a stealth game damnit!!

PR #24223, opened 2025-07-26
Non-Creepers with Invasive Growth do not mutate when spreading to new trays

NibChocolateeny:
[Hydroponics][QoL]

## About the PR
before, when the plant with invasive growth grow to new places, it have chanse to mutate in all of the tray on pl,anting. This can be not very nice - in the theory, it would be nice for the plant that will be used for selling for money (like the cannnabis), but it is not much more useful than the normal plant with seed becuase still cahnce for bad gene (stunted growing) or mutating into different type to happen.

now, when grow into new tray, plant does not mutate initially. can put the mutashion chemicals in tray ato have it mutate later, or do the infusing for mutate if you want. But for big planting/big harvest, can be convenient - get same plant for all tray!

## Why's this needed? 
I ALWAYS thorw away creeper seed - trap is not so good, do not want to kill other plant, VERY annoying if spread too much. BUT now it can be very nice if need to make lot of themoney, do something cannot normally do with normal planting of seed into tray (auto-plant with not any mutashion, not needing staiblizer). If wanting mutation, easy to spam-plant OR infuse OR use tray chemiacle. Interesting, I think!

I talk with @M-Earthfire - now mutation happen in creeper in all case.

## Testing
did the test loccally - check if genetics the same for pea withe "invasive" gene when moved to new trays, but not same for creeper with gene and not with gene
testing

```changelog
(u)NibChocolateeny
(+) New Research indicates that splicing the Invasive Growth gene into-non creepers will cause them to spread via budding; they do not have a chance to mutate initially when spreading to new trays.
```


M-Earthfire:
I don't mind that change with non-creeper plants. It makes them easier to replant and control.

But i don't really like it for the creeper itself. It's meant to be uncontrollable, and randomly getting tumbling creeper is a big reason why i made them in the first place.

The good thing is, creeper and invasive growth budding are seperate to each other. So if the creeper plant itself isn't affected by this change (read: if code/modules/hydroponics/plants_alien.dm isn't changed here) it's fine with me.

NibChocolateeny:
@M-Earthfire  it is not so consistent if doing like so...but! can change so no mutate only happen for trait. sound ok?

NibChocolateeny:
ok I make changes @M-Earthfire , thank you for feedback!

NibChocolateeny:
Haha- not ture, @M-Earthfire ! stabiliser only stop the mutashion, can still get the stat from the nutrient (wiki is not so correct)

NibChocolateeny:
@TobleroneSwordfish I am knowing thathe chamges for 516 are much more inportant...but other botanyists think this is ok change, and is small change. maybe can be merged sometime?

NibChocolateeny:
👁️👁️

NibChocolateeny:
:<
I wait.

PR #24246, opened 2025-07-28
Mechdrone 

Azoth13:
## About the PR 

mechdroneOBj

[ENGINEERING] [FEATURE] 

This adds the mechdrone, a new toy for mechcomp and packet users to play with! It can be controlled via packets to make it go from point A to B, it can carry small items if commanded to do so, always picking up the smallest item in the turf it is at, then can interact with other object and even put the carried item inside them if possible.

A simple example is using it to pick a potato then going to the tile next to an oven, and putting the potato in the oven. It can also interact using small_animal critter behaviour with things such as mechcomp button.

The Mechdrone adds new uses to packets and mechcomps, such as automating trivial tasks without being too powerful, and it requires a decent setup to do so.

It can also say things if command to do so with packets.

It also adds a pointer, which can be used with a packet as a destination for the mechdrone to go to.

The mechdrone, it's guidebook and the pointers can be found at the mechcomp vendor.
A mechdrone's address can be seen by clicking it with a multitool (Be careful, it's fragile!) or examining it in it's description.
pointers can be used in hand to set a name for their destination and then anchored to the ground with a wrench.

## Why's this needed? 

It's a fun addition to the game that incentivizes players to learn more about mechcomp and packets. I'm sure fun contraptions can be made with it. To the non-specialized user, this adds a new mechanic controlled by packets, which are usually pretty useless to the common user in my opinion, normally used to open doors, lockers or sometimes by anagonists to shut down apcs, etc. 
Despite it's features, it remains non-complicated to use while still requiring a setup. More commands may be added at some point if necessary.

## Testing 

I tested this extensively, and I've not found it to cause any problems as far as I know, I changed the drone to be a critter and not a machine as to implement a "interact system" that can only interact with what it's supposed to, similarly to a small animal critter.
If necessary I can upload videos of how it works and me testing it.


code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,419 @@
...
+						// Use mulebot-style pathfinding
+							var/list/path = get_path_to(src, target_turf, max_distance=200, id=null, skip_first=FALSE, exclude=null, cardinal_only=TRUE, do_doorcheck=TRUE)
+							if(path && length(path) > 1)
+								for(var/turf/step in path)
+									if(!src) break
+									// If the next step is an airlock, open it if needed
+									for(var/obj/machinery/door/airlock/A in step)
+										if(A.density && (!A.req_access || A.req_access == access_engineering || A.req_access == access_maint_tunnels))
+											A.open()
+											sleep(3)
```
JORJ949: 11 Indents is a LOT of indents, this should probably be put into its own proc. The entire receive signal proc is also indented by one unnecessarily (including the path means you don't need to indent the proc

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,419 @@
...
+			return
+		src.visible_message(SPAN_SAY("[src] says, \"[message]\""))
+
+
+	death(gibbed)
+		..()
+		elecflash(src, 1, 1) // Flash the area with an electric effect
+		qdel(src) // Remove the mechdrone from the world
+
+//SCRAPPED REMOTE WITH UI THINGY
```
JORJ949: If it's scrapped don't include it in the PR, it just increases your line count and size making it seem like it'll take longer to review than it actually will

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,419 @@
...
+
+	proc/stop_movement()
+		// Stops the movement of the mechdog
+		walk(src, 0)
+
+	proc/say_something(var/message)
+		// message = trimtext(copytext(sanitize(html_encode(message)), 1, MAX_MESSAGE_LEN))
+		if (!message)
+			return
+		src.visible_message(SPAN_SAY("[src] says, \"[message]\""))
```
JORJ949: Should probably use .say() and not visible_message, removing the need for this proc entirely

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,419 @@
...
+								break
+			if("interact")
+				SPAWN(0)
+					if(signal.data["data"])
+						var/target_name = signal.data["data"]
+						var/found = FALSE
+						for(var/dir in cardinal)
+							var/turf/T = get_step(get_turf(src), dir)
+							if(!T) continue
+							for(var/obj/O in T)
```
JORJ949: for(var/obj/O in oview(1, src)) will do exactly what you want here

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,419 @@
+/mob/living/critter/small_animal/mechdrone //Yes I know it's silly but if it works, eh.
```
JORJ949, line 1: This should probably be a robotic critter

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,419 @@
...
+	name = "Mech Drone"
+	desc = "A programmable robotic drone that responds to network packets."
+	icon = 'icons/misc/mechDrone.dmi'
+	icon_state = "1"
```
JORJ949: Name the icon state

code/modules/vending/vending.dm
```diff
@@ -1675,7 +1675,7 @@ ABSTRACT_TYPE(/obj/machinery/vending/cola)
 
 	create_products(restocked)
 		..()
-		product_list += new/datum/data/vending_product(/obj/item/paper/book/from_file/mechanicbook, 30)
+		product_list += new/datum/data/vending_product(/obj/item/paper/book/from_file/mechanicbook, 5)
```
JORJ949: Unrelated change to the PR

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,419 @@
...
+						for(var/obj/item/mechdrone_pointer/P in world)
+							if(P.pointer_name == pointer_name && P.z == src.z)
+								// Use mulebot-style pathfinding
+								var/list/path = get_path_to(src, P, max_distance=200, id=null, skip_first=FALSE, exclude=null, cardinal_only=TRUE, do_doorcheck=TRUE)
+								if(path && length(path) > 1)
+									for(var/turf/step in path)
+										if(!src) break
+										// If the next step is an airlock, open it if needed
+										for(var/obj/machinery/door/airlock/A in step)
+											if(A.density && (!A.req_access || A.req_access == access_engineering || A.req_access == access_maint_tunnels))
```
JORJ949: Should probably instead have an access implant like Remy or Admin Mice do, aditionally req_access won't be equal to access_engineering, but a list containing access_engineering, even if there's only one access on it

code/modules/mechanics/mechdrone.dm
Azoth13: I'm unfamiliar with github still so I don't know how to reply to everything as a whole so I'll reply here. Thanks a lot for the pointers! Fixed everything you mentioned except for the critter being robotic, I had already attempted this before but it causes problems due to the lack of hands or something. But hey, it bleeds oil!

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,245 @@
...
+		//var/senderid = signal.data["sender"]
+		switch( lowertext(signal.data["command"]) )
+			if("follow")
+				SPAWN(0)
+					src.follow(signal)
+			if("stop")
+				SPAWN(0)
+					src.stop_movement()
+			if("load")
+				SPAWN(0)
```
JORJ949: What are all these SPAWN(0)s for?

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,245 @@
...
+								break
+			if("interact")
+				SPAWN(0)
+					if(signal.data["data"])
+						var/target_name = signal.data["data"]
+						var/found = FALSE
+						for(var/dir in cardinal)
+							var/turf/T = get_step(get_turf(src), dir)
+							if(!T) continue
+							for(var/obj/O in oview(1, src))
```
JORJ949: I meant that this iterates over all objects in the range of 1 to src, meaning you don't need the cardinal or turf stuff
image

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,245 @@
...
+						var/message = signal.data["data"]
+						if(istext(message))
+							src.say(message)
+
+			if("pointer")
+				SPAWN(0)
+					if(signal.data["data"])
+						var/pointer_name = signal.data["data"]
+						// Search for a pointer with the given name on the same z-level
+						for(var/obj/item/mechdrone_pointer/P in world)
```
JORJ949: Don't use in world, see here: https://hackmd.io/@goonstation/docs/%2F%40goonstation%2Fcode#Tracking-Macros
use for_by_tcl() instead

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,245 @@
...
+
+	proc/stop_movement()
+		// Stops the movement of the mechdog
+		walk(src, 0)
+
+	proc/follow(datum/signal/signal)
+	//Follows the source of the signal
+		var/turf/target_turf = find_target_location_from_signal(signal)
+		if(target_turf && (istype(target_turf, /turf) || istype(target_turf, /mob) || istype(target_turf, /obj)))
+			if(target_turf.z == src.z)
```
JORJ949: Instead of having the follow code behind a bunch of ifs, you should have the proc return early if the requirements aren't met, returning means the rest of the proc won't run.

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,245 @@
...
+			return null
+
+	proc/stop_movement()
+		// Stops the movement of the mechdog
+		walk(src, 0)
+
+	proc/follow(datum/signal/signal)
+	//Follows the source of the signal
+		var/turf/target_turf = find_target_location_from_signal(signal)
+		if(target_turf && (istype(target_turf, /turf) || istype(target_turf, /mob) || istype(target_turf, /obj)))
```
JORJ949: if it can be something that's not a turf, don't make it a turf or name it "target_turf" as its misleading

code/modules/mechanics/mechdrone.dm
Azoth13: To be absolutely honest, when I first made the signal code for this I hadn't worked with coding packets previously, so I just followed from another packet controlled thing (maybe the mulebot? I don't remember), I'll test and see if they don't do anything and can be removed without complications.

code/modules/mechanics/mechdrone.dm
Azoth13: I'll see to improving it, I'm bad at pathfinding so it's already a bit janky though it works, if you have any suggestions of similar code or people knowledgeable on proper pathfinding I'd love to hear them.

I'll try to fix this and the other suggestions ASAP.

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,238 @@
...
+		if(!isnull(inp))
+			set_frequency(inp)
+			boutput(user, "Frequency set to [inp]")
+			// tooltip_rebuild = 1
+			return 1
+		return 0
+
+	proc/set_frequency(new_frequency)
+		if(!radio_controller) return
+		//tooltip_rebuild = 1
```
Azrun, line 61: Is this needed?

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,238 @@
...
+				pingsignal.data["device"] = "MECH_DRONE"
+				pingsignal.data["netid"] = src.net_id
+				pingsignal.data["address_1"] = signal.data["sender"]
+				pingsignal.data["command"] = "ping_reply"
+				pingsignal.data["data"] = "Mech Drone"
+
+				SPAWN(0.5 SECONDS) //Send a reply for those curious jerks
+					SEND_SIGNAL(src, COMSIG_MOVABLE_POST_RADIO_PACKET, pingsignal, src.range)
+
+		//var/senderid = signal.data["sender"]
```
Azrun, line 81: Is this needed?

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,238 @@
...
+
+	proc/find_target_location_from_signal(datum/signal/signal)
+		if(!signal || !signal.source)
+			return null
+		if(isturf(signal.source))
+			return signal.source
+		if(ismob(signal.source) || isobj(signal.source))
+			if(signal.source.loc && isturf(signal.source.loc))
+				return signal.source.loc
+			return null
```
Azrun, line 166: Consider `get_turf(signal.source)`

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,238 @@
...
+			return
+		if(target.z != src.z)
+			return // Don't follow if the target is on a different z-level
+			// Use mulebot-style pathfinding
+		var/list/path = get_path_to(src, target, max_distance=200, id=null, skip_first=FALSE, exclude=null, cardinal_only=TRUE, do_doorcheck=TRUE)
+		if(path && length(path) > 1)
+			for(var/turf/step in path)
+				if(!src) break
+				sleep(3)
+				step_to(src, step)
```
Azrun, line 185: Consider `step_towards_movedelay`

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,238 @@
...
+
+	proc/follow(datum/signal/signal)
+	//Follows the source of the signal
+		var/atom/target = find_target_location_from_signal(signal)
+		if(!target)
+			return
+		if(target.z != src.z)
+			return // Don't follow if the target is on a different z-level
+			// Use mulebot-style pathfinding
+		var/list/path = get_path_to(src, target, max_distance=200, id=null, skip_first=FALSE, exclude=null, cardinal_only=TRUE, do_doorcheck=TRUE)
```
Azrun, line 180: 200 distance seems like a lot
Additionally consider having this generate the path to follow and then having it walk the path as part of periodic processing.  Otherwise this should this should have some form of rate limiting and/or locking so N different follow commands don't have it walking back and forth OR duplicate commands don't increase the rate of travel.

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,238 @@
...
+		STOP_TRACKING
+
+
+
+
+
+
+
+
+
```
Azrun, line 238: Just one extra space is needed.

code/modules/mechanics/mechdrone.dm
Azrun: `/mob/living/critter/robotic/gunbot` have hands

code/modules/mechanics/mechdrone.dm
```diff
@@ -0,0 +1,238 @@
...
+		. += SPAN_NOTICE("<br>Its network address is [net_id].")
+
+	Cross(atom/movable/mover)
+		. = ..()
+		if(istype(mover, /turf/simulated/wall))
+			return  // Don't cross walls
+		if(istype(mover, /obj/window))
+			return // Don't cross windows
+		else
+			return TRUE // Cross other turfs
```
Azrun, line 44: Why? What?

> Called when another object attempts to overlap this one.

Azrun (changes requested):

PR #24247, opened 2025-07-28
Remaps Every Single Camera

Mister-Moriarty:
[Game Objects] [QoL]

## About The PR:
Repaths every single non-invisible camera to a directional variant that uses the new directional offsets system.

The repath was achieved through the use of an `UpdatePaths` script, with additional functionality being added to `UpdatePaths` to accommodate it. Specifically, the `@ANY` keyword has been ported from /tg/ and a new `@ADD:num` keyword has been implemented, which allows scripts to add/subtract integers from numerical variables.

Regarding the script itself,
- The `name = "autoname - SS13"`, `c_tag = "autotag"`, and `tag` varedits have been removed from mapped cameras, with those now being the default values.
- Every camera now has a standard offset in the direction it is facing (i.e. how far up the wall it appears) as a result of the directional offsets system.
- Cameras facing in cardinal and ordinal directions have been resolved to only face in cardinal directions, with pixel offsets applied such that they match the offsets of the original icons.
- Camera directions have been flipped such that a northern facing camera will attach to a northern wall, which is consistent with how other directional objects operate.
- `/obj/machinery/camera/auto` has been deprecated and removed.
- Watchful Eye Sensor Array cameras now have their own subtype.

In addition,
- `/obj/machinery/camera/var/sticky` and `/obj/machinery/camera/proc/autoposition` have been deprecated and removed.

## Why Is This Needed?
See #24226.

## Testing:
In all maps that I loaded both ingame and in SDMM, none had any observable differences. Cameras were tested both through playing as the AI and using various different camera viewers, including the Rancher's baby monitor and Security's camera monitors.

## Changelog:
```changelog
(u)Mr. Moriarty
(*)Every single camera has been remapped. Please report any cameras that aren't offset onto walls or that are facing the wrong way.
```

### DIFF
--- a/code/modules/camera/camera.dm
+++ b/code/modules/camera/camera.dm
@@ -88,6 +88,18 @@
 	if (src.network in /obj/machinery/computer/camera_viewer::camera_networks)
 		src.minimap_types |= MAP_CAMERA_STATION
 
+	if (dd_hasprefix(src.name, "autoname"))
+		var/name_build_string = ""
+		if (src.prefix)
+			name_build_string += "[src.prefix] "
+
+		name_build_string += "camera"
+		if (src.uses_area_name)
+			var/area/A = get_area(src)
+			name_build_string += " - [A.name]"
+
+		src.name = name_build_string
+
 	SPAWN(1 SECOND)
 		addToNetwork()
 
--- a/code/modules/camera/network.dm
+++ b/code/modules/camera/network.dm
@@ -6,19 +6,11 @@
 	var/list/counts_by_tag = list()
 	var/list/obj/machinery/camera/first_cam_by_tag = list()
 	for (var/obj/machinery/camera/C as anything in cameras)
-		var/area/where = get_area(C)
-		var/name_build_string = ""
 		var/tag_we_use = null
-		if (dd_hasprefix(C.name, "autoname"))
-			if (C.prefix)
-				name_build_string += "[C.prefix] "
-			name_build_string += "camera"
-			if (C.uses_area_name)
-				name_build_string += " - [where.name]"
-			C.name = name_build_string
 
 		if (isnull(C.c_tag) || dd_hasprefix(C.c_tag, "autotag"))
-			tag_we_use = where.name
+			var/area/A = get_area(C)
+			tag_we_use = A.name
 		else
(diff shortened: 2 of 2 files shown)


code/modules/camera/camera.dm
```diff
@@ -1,5 +1,5 @@
 /obj/machinery/camera
-	name = "security camera"
```
TobleroneSwordfish, line 2: This means any admin spawned security cameras will be called "autoname - SS13"
image
Maybe set the name according to the game state in New instead? I'm not really sure what the point of the autoname string is anyway.

code/modules/camera/camera.dm
```diff
@@ -324,150 +321,64 @@
...
-			src.pixel_y -= 10
-			src.set_dir(WEST)
+/// Adds the minimap component for the camera
+/obj/machinery/camera/proc/add_to_minimap()
+	src.AddComponent(/datum/component/minimap_marker/minimap, src.minimap_types, "camera", name=src.c_tag)
 
-		if (NORTHEAST)
-			src.pixel_y -= 10
-			src.set_dir(EAST)
+SET_UP_DIRECTIONALS(/obj/machinery/camera, OFFSETS_CAMERA)
```
TobleroneSwordfish, line 340: All these directionals are just named for their direction in the map editor, not indicating that they are in fact cameras. It's not a big deal but I'd rather fix it now if it is an issue with the generic auto-positioning system.

code/modules/camera/camera.dm
```diff
@@ -324,150 +321,64 @@
...
 
-		if (SOUTHEAST)
-			src.pixel_x += 10
-			src.set_dir(SOUTH)
+/obj/machinery/camera/cargo
+	name = "autoname - cargo"
```
TobleroneSwordfish, line 343: Same with these, we want spawned cameras to just be called "security camera"

TobleroneSwordfish (changes requested):

code/modules/camera/camera.dm
Mister-Moriarty: For routing cameras, it would be more accurate for them to be named as such, as they're on a separate network to security cameras.

TobleroneSwordfish (approved):

PR #24255, opened 2025-07-29
New tanktop options

Munien7:
## About the PR 
Adds a few new types of tanktop/vest to the clothing booth using pre-existing pants, based on the old HoS/Shitty Bill one.
## Why's this needed? 
A few people have expressed interest in some new tanktop variety, as the current one in the clothing booth is only the bright red pants (also a security uniform). This brings some more variety while keeping the same tone as other existing clothing options.

Could possibly look into more departmentally themed colours for a future PR
## Testing 

dreamseeker_xKNNFoDp16
uV5OZzduqn

## Changelog 

```changelog
(u)Munien
(+)Added a few new tanktop variants to the clothing booth.
```


code/obj/item/clothing/uniforms.dm
```diff
@@ -365,13 +365,29 @@ ABSTRACT_TYPE(/obj/item/clothing/under/rank)
...
-/obj/item/clothing/under/misc/dirty_vest //HoS uniform from the Elite Security era
+ABSTRACT_TYPE(/obj/item/clothing/under/misc/dirty_vest) //HoS uniform from the Elite Security era
+/obj/item/clothing/under/misc/dirty_vest
 	name = "dirty vest"
 	desc = "This outfit has seen better days."
 	icon_state = "vest"
 	item_state = "vest"
 	c_flags = SLEEVELESS
 
+	redpants
```
JORJ949: I would just not make the parent abstract, all instances of the red vest are already the parent and this saves you having to repath to this one

Munien7: Got it, confused myself big time, pretty sure it should be good now.

PR #24268, opened 2025-07-30
Cleanup genetics abilities 

Egregorious:
## About the PR 

Mostly just refactors gene abilities to remove a lot of copy-pasted code and use defined returns. Removed `cast_misfire` and replaced it with `cast_genetics`, which is always called by `cast` and just provides a `misfire` arg to use instead. 

This fixes a bug where many abilities essentially casted twice whenever a misfire occurred. Some of the behaviour caused by this has been rolled into the actual miscast behaviour, just so that balance isn't affected: 
- Eyebeam also regular casts
- Cryokinesis also regular casts
- Healing Touch double-heals

Other miscast behaviours that seemed less desireable have been fixed:
- Matter Eater won't swallow items
- Telepathy will only cause a shout
- Adrenaline Rush won't inject an extra time

Added a few minor feedback tidbits:
- by default the character will try to turn to face the ability target
- Eyebeam's miscast causes a scream
- added a feedback message to Telekinetic Throw's miscast
- Jumpy's miscast effects trigger upon landing

Added LAZYLISTACCESSASSOCDEEP macro for nested lazy lists, and converted LAZYLISTACCESSASSOC to only work for non-nested lists.

## Why's this needed? 

Genetics abilities had a lot of copy-pasted code, which was seemingly encouraged by the way misfire mechanics were handled. This way genetic abilities have more control over how they want to implement miscasts.

Gene abilities have some feedback issues. It's always felt weird to me that you fire eyebeams out the back of your skull, or reach out behind yourself to heal someone. 

LAZYLISTACCESSASSOC was impossible to use for any non-nested lists. 

## Testing 

Tested each of the powers and their miscasts as I went.

Turning:

https://github.com/user-attachments/assets/facd0f1f-f955-4a37-9017-94228d294dd6

### DIFF
--- a/code/modules/medical/genetics/bioEffects/powers.dm
+++ b/code/modules/medical/genetics/bioEffects/powers.dm
@@ -242,28 +242,12 @@ ABSTRACT_TYPE(/datum/bioEffect/power)
 		return FALSE
 
 	proc/handle_parts(var/obj/item/parts/object)
-		if (!istype(object))
+		if (!istype(object) || !ishuman(src.owner))
 			return FALSE
 		var/mob/living/carbon/human/H = src.owner
 		object.delete()
 		H.hud.update_hands()
 		return TRUE
-
-
-		/* // Organs and body parts have special behaviors we need to account for
-		proc/handle_organs()
-			if (!ishuman(src.owner))
-				return
-			var/mob/living/carbon/human/H = src.owner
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
This is several different PRs in a hat and is near-impossible to test and review properly. Please atomize and re-submit as PRs which do/fix a single thing/category each.

PR #24272, opened 2025-07-30
Lawbringer fires kinetics again, uses matter cartridges

TobleroneSwordfish:
## About the PR 
Makes the lawbringer fire the old kinetic EXTERMINATE and BIG SHOT rounds, all kinetic modes of the lawbringer now take matter units to fire (detain and pulse still use the internal battery and BIG SHOT takes 50% cell charge in addition to the matter units to stop tooo much rapid fire aex)

TODO:

- [ ] Give them some spare cartridges somewhere
- [ ] Change the muzzle flashes
- [ ] Number balance

## Why's this needed? 
This is kind of an experiment to see if I could come up with a good way to bring the kinetics back without the balance hell of the original infinitely regenerating bullets.
Maybe too complicated? Maybe every mode should just use matter?
Numbers are very much WIP, balance suggestions welcome.
## Testing 
image
image

## Changelog 

```changelog
(u)LeahTheTech
(*)The lawbringer now fires kinetics again for EXTERMINATE and BIGSHOT, and all kinetic lawbringer modes use matter cartridges for ammo.
```


JORJ949:
I'll be honest I much prefer the new ASSAULT, it doesn't breach which makes it actually usable for its intended purpose (e.g. I won't kill people via suffocation with my non-antag breach tool). Execute I don't mind either way, perhaps you could have both, ENERGY and EXECUTE?

TobleroneSwordfish:
We could maybe keep the new assault as separate from AEX? I don't want to end up adding tooo many more modes though, this thing is already bloated enough as is.

JORJ949:
> We could maybe keep the new assault as separate from AEX? I don't want to end up adding tooo many more modes though, this thing is already bloated enough as is.

I don't personally see a use for AEX as HoS, it's going to harm more than it helps, Laser vs Execute has use as one goes through windows and the other can be used with the bloodtrak, Assault is just a direct upgrade to AEX IMO.

Piesuu:
Not a HoS but the fact Big Shot causes breaches basically means that to even use it you'd need to be in a spacesuit with internals unless you want to get slowed down or suffocate. 
After the Big Shot > Assault change I've started seeing HoS' actually use it outside of gimmicks which is nice.

side-grade Idea to this: Replace cell charge with overheat and make it so every shot uses some amount of Matter. 
Lore-wise - Lawbringer most likely uses RCD tech and that has a speed limit on how fast it can create matter so you don't want to overheat whatever makes the bullets. Energy shots like Detain are just the RCD turning the matter into energy.

FlameArrow57:
Personal thoughts:
-RCD matter cartridges feel more of an engineering/mining thing considering what the RCD does and that you can make more RCD ammo with uqill. Maybe limited ammo in the armory would be better?

-Adding kinetics back to lawbringer would be fun in helping to distinguish it further from the egun (ex. what is point of execute of LB if not a lower damage version of the egun laser [neglecting self recharge]?)

-Personally prefer the kinetic aesthetics of bigshot mode with the mechanics of assault mode (wall/door breaching mechanics without the hull breaching mechanics)

-Possibly feel that it would be better to leave bigshot/assault at 60% of the magazine still to make landing a shot more valuable rather than having 2 chances with it being 50%

PR #24284, opened 2025-08-01
Vampires above 1800 blood get unique chat maptext font

FlameArrow57:
[GAME MODES][FEATURE]

## About the PR 
Makes it so that maptext speech for vampires above 1800 blood level appears unique

image

Font from https://www.dafont.com/old-london.font

Also open to better fonts, was just starting initially with an old language like font

## Why's this needed? 
-Makes it clear that vampires with high blood level are dangerous forces
-Saw someone working on it a while ago, thought it would be cool

## Testing 
Made sure that vampires were getting the speech at 1800 blood and that it was removed after death

## Changelog 

```changelog
(u)FlameArrow57
(*)Vampires above 1800 blood get different font for text spoken above their head
```


TobleroneSwordfish:
Font is borderline illegible at that resolution and it kind of ruins all attempts at stealth/deception past a certain blood level?

FlameArrow57:
> Font is borderline illegible at that resolution and it kind of ruins all attempts at stealth/deception past a certain blood level?

Yeah if you have any better fonts I can try them. 1800 blood is the level at which you become chapel immune

PR #24286, opened 2025-08-02
Holy water cures artifact curses

RubberRats:
[SCIENCE][BALANCE]
## About the PR 

This PR adds an alternative cure condition to artifact curses in the form of splashing holy water on curse victims. It also slightly refactors artifact curses to enable this.

## Why's this needed? 

Alternative PR to #23863. As mentioned in that PR, curse artifacts have disproportionately severe effects compared to how easy it is to be affected by them and how difficult it is to lift the curse. Holy water curing curses has precedence from plague wraiths, and requires cooperation from specific roles which do not normally see much action.
## Testing 
Tested by touching active curser artifacts and then splashing myself with holy water. Also included lifting the curse normally to make sure that still works as intended.

## Changelog 

```changelog
(u)RubberRats
(+) Curses from artifacts can now be cured by splashing holy water.
```

### DIFF
--- a/code/modules/chemistry/Reagents-Base.dm
+++ b/code/modules/chemistry/Reagents-Base.dm
@@ -831,12 +831,26 @@
 				else
 					if (ishuman(M))
 						var/mob/living/carbon/human/H = M
+						var/removed_curse = FALSE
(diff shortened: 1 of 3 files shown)


ZeWaka (commented):
Is there some minimum reagent quantity required? Just making sure we're not re-introducing "homeopathic"  curing

RubberRats:
I didn't add a minimum because I was following the example of wraith curses. I just added a 10u threshold for curse removal to work

PR #24340, opened 2025-08-05
adjusts semiconductor values to increase cap

Scaltra-Volpe:
changes semiconductor efficiency code to have a higher cap so it can have more of a contribution to the power and temperature transfer equation 

 I think a higher cap will add a more interesting dynamic to the TEG, especially with the cold loop, as it will be more thermally isolated with higher efficiency semiconductors, making it more worthwhile to go for something like the cryostylane blower reaction effect, as well as giving a reason to actually go for higher materials to meet the new cap. (also considering a lower scaling on the semiconductor material equation to make the highest possible material conductor naturally reach the cap and bring everything else down a bit to add a sensical progression to the best possible option)

 Testing has been multiple launches of the code while adjusting the upper clamp of the TEG semiconductor transform efficiency code and the TEG max semiconductor contribution code on kondaru testing, both how the normal TEG responds to the change, as well as the experimental cold variation at these values, the temperature and contributions act as expected.

 Changelog upped the semiconductor contribution cap

```changelog
(u)Scaltra
(+)upped semiconductor contribution cap
```

### DIFF
--- a/code/datums/teg_transform.dm
+++ b/code/datums/teg_transform.dm
@@ -161,4 +161,4 @@ datum/teg_transformation
 			/*  2*25 / 75 = 0.66 -2  = -1.34 		 TERRIBAD */
 			/* Use above offset * 10 to put it in the -20 to 40 ballpark */
 			efficiency_shift = (2 * electrical_conductivity / thermal_conductivity) - 2 //center on zero
-			src.teg.semiconductor.efficiency_offset = clamp(efficiency_shift*10, -20, 40) //scale shift by 10 which gets it in the ballpark!
+			src.teg.semiconductor.efficiency_offset = clamp(efficiency_shift*10, -20, 48) //scale shift by 10 which gets it in the ballpark!
--- a/code/modules/power/TEG.dm
+++ b/code/modules/power/TEG.dm
@@ -926,7 +926,7 @@ datum/pump_ui/circulator_ui
 
 		if(semiconductor)
 			//Bound contribution of the semiconductor to +/- 25
-			var/semi_contribution = clamp(src.semiconductor.efficiency_offset, -25, 25)
+			var/semi_contribution = clamp(src.semiconductor.efficiency_offset, -25,48)
 			efficiency_scale += semi_contribution
 
 		if(src.generator_flags & (TEG_HIGH_TEMP | TEG_LOW_TEMP))


Azrun:
Why 48?

Scaltra-Volpe:
> Why 48?

Testing with 50 and above the cold loop could go negitve and I think also get even colder on the output than the input without something like cryostylane. I was mainly focused on avoiding negative temp gas stuff though.

RubberRats:
The TEG has an efficiency scale that determines how effectively it can turn the temperature differential between the hot and cold loops into power, with the rest of it being dumped into the cold loop as waste heat. It defaults to 52%, and can be increased by using higher quality semiconductors. At more than 100% efficiency, the math breaks down and give the TEG the power to defy the second law of thermodynamics by having it remove heat from the cold loop.

The TEG also has a chance to be replaced with high- or low-temperature variants, which also come with their own conditional modifiers to efficiency. As-is, this PR still allows for this behavior by using one of these variants with the upgraded semiconductor. It might make sense to cap the `get_efficiency_scale()` proc to prevent it entirely.

Scaltra-Volpe:
> The TEG also has a chance to be replaced with high- or low-temperature variants, which also come with their own conditional modifiers to efficiency. As-is, this PR still allows for this behavior by using one of these variants with the upgraded semiconductor. It might make sense to cap the `get_efficiency_scale()` proc to prevent it entirely.

I tested it including the cold varient and tried to force it past 1k but it did not go negative. I was also concerned about that but from my testing it both didn't go negitive and was not super naturally efficient. 

I will take a look at that though and see if I can maybe throw in a cap there just in case if I can figure it out and the same clamp style would work elsewhere. If it feels needed or to future proof

Azrun:
Clamping at points to ensure valid/expected/reasonable behavior of the simulation seems ideal.  That then provides some wiggle room elsewhere in the name of fun and sci-fi.

Scaltra-Volpe:
> Clamping at points to ensure valid/expected/reasonable behavior of the simulation seems ideal. That then provides some wiggle room elsewhere in the name of fun and sci-fi.

Alright, i will see if I can figure out how to code that in with it when I get the time, in the mean time should I close the pr? 

Also if anyone knows how to force the experimental variations please let me know. I don't know how much more restarting of the game till it picks it I can tolerate

PR #24346, opened 2025-08-06
Fix comments in ammo.dm (and makes frangible ammo manufacturable)

NibChocolateeny:
[CODE QUALITY] [BALANCE] [GAME OBJECTS] [FEEDBACK]
## About the PR 
ok, it was originally that i wanted fixingin of comment in ammo.dm that was copied worngly, saying gang men ammunition coud be in hacked manufacturing mashine - which is not seem right, it is only for crime people, and I think probably from copey and paste of 9mm ammunishion when making gang thingies originally. But I look at the other commment and I think "oh...nobody make this makeable by hacked manufacturing before...why not?"

so now both comment are more correct, and you can make frangible bullet in the hacked general maker machine. BUT it is big balancing changing! it is more important that commenting is rignt for me, I am not so upset if frangible bullet not makeable. but maybe it is good idea...

## Why's this needed? 
commenting should not be wrong. 

ALSO, for frangible shooting, here is the comparingness against the other bullet you can be making in fabricationer, 0.22:

Ammo Type | 0.22 (Fabricated) | 9mm Frangible (Fabricated)
-- | -- | --
Number of Rounds | 10 | 18
BRUTE damage per shot | 22 | 6
STAMINA damage per shot | 0 | 16
Total raw BRUTE and STAMINA damage | 220, 0 | 104, 288
Other effects? | Causes bleeding | Slows targets for 1 second
Fabrication Time | 30 Seconds | 15 Seconds
Fabrication Cost | “metal_dense” 3.0x, “conductive” 2.4x | “Plastic” 1.5x, “rubber” 1.5x, “metal” 1, “conductive” 2.5x
Eligible Weapons? (Conditional availability in parentheses) | Zip Gun, Faith (Chaplain, Secret), Orion Silenced Pistol (Traitor), Capella Pistol (Nukie), Hydra Smart Pistol (Nukie), Efnysien Survival Rifle, 0.22 mod (Salvager) | Zip Gun, Lo-Point Pistol (Gang), Branwen Pistol (Sleeper agent, Nukie, Mining/Secret), Efnysien Survival Rifle, 9mm mod (Salvager), Clock 180 (Experimental Weapons)

in the experience I am having (no antagonagonist), can find 1 branwen in mining space sometimes, but there is no bullet and no way of having any more bullet for 9mm. BUT it is also true that if sleeping agent get secrett "dead drop," it is high chance for getting the brawnwen gun, and it has only small amount of bullet. so now it can be that you can make more 9mm bullet...but bullet is not very good, and expencive.

there is way for finding the 9mm bullet in secret zone (take from markov gun, russian pistol). If not in secret zone, cannot get the 9mm bullet UNLESS you are breaking into armory - the ammoatech vending mashine has them, and can restock. but it is so hard to break into amrory, and the bullet is bad! so it make no sense to break into amrrory for rubber bullet if angtagonist, and not allowed if not bad guy.

also, even if security...bullet is not good! all taser gun are much better for stun, also rubber shotgun. I tested, it is not good. also, becuase it need the plastic spesifically, it is not so easy to make even 1 of the bullet - need to cut 5 of the cargo flaps to get 2 unit of reclaim plastic for 1 set uf bullet pieces. or do the chemisty, but that is maybe also annoying I think, do not know.

but it is ok if not added. 0.22 bullet is so much more better in zip gun. The change is mostly helping sleeping agents, I think.

## Testing 
Did test with EMAG, normal hack of manufacturing mashine. Bullet can use only plastic or the plastic and rubber like was expecting (need plastic specifically, but do not need parnticular rubber, can use only plamstic or plasmstic and rubber). recipe go away if unhack machine.

testing

## Changelog 

```changelog
(u)NibChocolateeny
(*)9mm Frangible Rounds can be fabricated from hacked general fabricators, but you'll need a lot of plastic to make them.
```

## P.S.
there is other recipe for frangible bulllet, but it is for mode nobody plays (ultra uber super survival???). i did not use because it is not balansed! can change other recipe if wanted, or delete.


JORJ949:
Minor correction: the clock 180 is in the experimental security equipment cargo crate, not the armory. This does let anyone with regular security gear access get access to a pretty decent kinetic gun, unlike .22 which to nonantags can't be used in anything but the zipgun or faith. Although the plastic and rubber requirements will probably prevent this from being powergamed requiring both botany and chemistry to do things to craft them.

NibChocolateeny:
@JORJ949 thank you for the correcting!

NibChocolateeny:
...where arthe tags? what did i do worng?

JORJ949:
> ...where arthe tags? what did i do worng?

The labeler bot is currently broken, I've added them for you

NibChocolateeny:
Not stale! stoppit. >:(

Scaltra-Volpe:
I am for it! More printable ammo is cool and 9mm hardly seems like a bad one to add on. Feels like it gives a less lethal more stun oriented printable ammo type as well. More or less, which is a nice niche.

PR #24349, opened 2025-08-06
Material Color Matrices & Filters

LorrMaster:
[Materials][Feature]

## About the PR
- Changes a bunch of materials to use color matrices & filters for their appearance rather than hex colors
- Edits some sprites (such as material sheets) so that they look okay with the new materials

## Why's this needed?
- Makes objects made of these materials look more like the actual material.

## Testing 
Screenshot 2025-08-06 045603

## Changelog
```changelog
(u)LorrMaster
(*)Updated how various materials impact object appearance.
```


Glamurio:
That rusty overlay / texture looks so peak

LorrMaster:
The best look for radioactivity I found was to just rely on an outline for everything except turfs. So that's what I decided to go with.
> That rusty overlay / texture looks so peak

Mauxite actually has unique texture for walls and floor tiles, so its part of the image is a little deceptive.

Glamurio:
> The best look for radioactivity I found was to just rely on an outline for everything except turfs. So that's what I decided to go with.
> 
> > That rusty overlay / texture looks so peak
> 
> Mauxite actually has unique texture for walls and floor tiles, so its part of the image is a little deceptive.

Oh, that's Mauxite, right, I remember that. It's okay, nobody has to know.

code/modules/materials/Mat_Materials.dm
```diff
@@ -654,30 +680,42 @@ ABSTRACT_TYPE(/datum/material/metal)
...
+		addTrigger(TRIGGERS_ON_ADD, new /datum/materialProc/add_color_hsl/mauxite())
+		addTrigger(TRIGGERS_ON_REMOVE, new /datum/materialProc/remove_color_hsl())
 
 
 /datum/material/metal/cerenkite
 	mat_id = "cerenkite"
 	name = "cerenkite"
 	desc = "Cerenkite is a highly radioactive metal."
-	color = "#CDBDFF"
+	color = null // "#CDBDFF"
```
TobleroneSwordfish: This is causing a nuclear reactor runtime, it needs to know the colour to apply it to the grid icon. See checks/kondaru for the line number.

TobleroneSwordfish (changes requested):

PR #24392, opened 2025-08-10
Clarion QM Office Fix

wolfolotl:
[MAPPING] [BUG]
## About the PR
jorj949 pointed out that, on clarion, there was one tile of quartermaster office area out in space. emily told him to fix it but he didn't want to. so I fixed it

## Why's this needed?
emily asked very nicely I think

## Testing 
looked at it in strongdmm. offending tile was no longer a qm office area. if I need to test this further let me know

### DIFF
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -31672,9 +31672,6 @@
 	},
 /turf/simulated/floor/black,
 /area/station/security/brig)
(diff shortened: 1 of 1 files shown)


JORJ949 (approved):
I see someone is less lazy than I am at this moment

PR #24395, opened 2025-08-10
Changes the materials needed for a cyborg teleport upgrade

Calphrick:
## About the PR 
Changes the material needed for a cyborg teleport upgrade from energy source to telecrystal.
## Why's this needed? 
The teleport upgrade is very powerful. High energy materials are relatively easy to get, with cerenkite being a common material to have over 100 of. Telecrystal on the other hand, is very rare when mining and can mostly only be obtained through trading or adventuring. This should prevent every cyborg from having a teleport module. Also, the fact that the teleport upgrade doesn't use the material known for teleporting you is a little silly.
## Testing 
Worked fine in a test build. Not sure if that's sufficient enough though.

## Changelog 

```changelog
(u)Calphrick
(+)Changed one of the materials needed for the cyborg teleport upgrade.
```


kbsmilk:
Note that telecrystals is something any cyborg (all access) can walk into cargo and order 20 of from 80% of random traders. A shipping market refresh later may delete them all, but they can just add it to the roundstart chores list.

If I was to pick a rare material for this it'd be erebite or flawless gemstones. Or remove telecrystals from traders

M-Earthfire:
> Note that telecrystals is something any cyborg (all access) can walk into cargo and order 20 of from 80% of random traders. A shipping market refresh later may delete them all, but they can just add it to the roundstart chores list.

telecrystals fall under energy source anyway. So that's what they are able to do already, if they really wanted to.

Calphrick:
Telecrystals cannot be used to make the teleport upgrade. I have triple checked this.

M-Earthfire:
Huh, nevermind then. I was very sure they fall under energy source.

TobleroneSwordfish (approved):

Azrun (approved):
Greatly appreciate a more thematic and in general reduction in general availability of teleporter technology!

PR #20533, opened 2024-09-02
DWAINE Kernel Cleanup

Mister-Moriarty:
[Internal] [Code Quality]

## About The PR:
Applies modern formatting standards to `_base_os.dm`, including absolute pathing, use of `src` and `global`, use of `length()` in lieu of `.len`, and early returns.

More importantly, thoroughly documents most aspects of the code, including variables, procs, system call defines, and some of the more obscure logic, among others.

## Why Is This Needed?
DWAINE, and by extension the code for all other OSes, is typically of a high quality, however it formatted quite poorly and documentation is scarce. This PR is part of a larger personal project to clean up, and where necessary, later refactor, OS code.

PR #24418, opened 2025-08-12
DWAINE Shell Cleanup

Mister-Moriarty:
[Internal] [Code Quality]

## About The PR:
Applies modern formatting standards to `base_program.dm`, including absolute pathing, use of `src` and `global`, use of `length()` in lieu of `.len`, and early returns.

Shell builtins and script operators have been split into their own datums instead of being implemented inside of egregiously long switch statements.

Additionally thoroughly documents most aspects of the code, including variables, procs, and some of the more obscure logic, among others.

## Why Is This Needed?
See #20533.

## Testing:
All commands and script operators work as expected post cleanup.


MonoSyntax:
Great changes! DWAINE can always use some love

PR #24430, opened 2025-08-13
Removes misplaced gas mask pixel

nova2053:
[BUG] [OBJECTS]
## About the PR 
Removes a single misplaced pixel from the non-emergency-issue gas mask sprite 
Screenshot (6070)

## Why's this needed? 
Misplaced pixels are icky and bad and no good
## Testing 
Screenshot (6071)


Gannets (approved):
Good find

JORJ949 (approved):
Kill it

PR #24451, opened 2025-08-16
Occupation TGUI

Mister-Moriarty:
[Respawning] [Feature] [UI]

## About The PR:
Converts the job preferences window to TGUI, integrating it into the Character Setup panel.

The new UI is mostly one-to-one with the original UI, but has a few additional features, including unavailable jobs and antagonists listing the reason as to why they are unavailable in a tooltip, alongside job named being listed in tooltips, so that long job names aren't lost to overflow.

## Why Is This Needed?
Interface modernisation.

## Testing:
image
image
image
image

## Changelog:
```changelog
(u)Mr. Moriarty & Mordent
(*)The job preferences window has been converted to TGUI and integrated into the Character Setup panel.
```


code/datums/preferences/preferences.dm
```diff
@@ -1269,306 +1339,189 @@ var/list/removed_jobs = list(
...
-					if (cat != "unwanted")
-						jerbs[cat] -= JD.name
-						jerbs["unwanted"] += JD.name
+
+				job_static_data[job_datum.name] ||= list()
+				job_static_data[job_datum.name]["disabled"] = FALSE
+				job_static_data[job_datum.name]["colour"] = colour_map[job_datum.linkcolor] || "blue"
+
+				// If the user cannot play as this job.
+				var/reason_tooltip = null
```
JORJ949, line 1439: Reason tooltip should have a fallback if it finds no other reason you can't play. Currently will apply to the "trusted_only" HoS/mentor halloween jobs.

code/datums/preferences/preferences.dm
```diff
@@ -1269,306 +1339,189 @@ var/list/removed_jobs = list(
...
+				if (global.jobban_isbanned(user, job_datum.name))
+					reason_tooltip = "You have been banned from playing this job."
+				else if (job_datum.needs_college && !user.has_medal("Unlike the director, I went to college"))
+					reason_tooltip = "This job requires the <i>\"Unlike the director, I went to college\"</i> medal, which you do not possess."
+				else if (job_datum.requires_whitelist && !global.NT.Find(user.ckey))
+					reason_tooltip = "This job requires being on the Head of Security whitelist."
```
JORJ949: Could probably tell mentors that they can play hos on fridays to avoid confusion, though this doesn't really need to be told to all players. "As a mentor, you can play this role on Fridays (in x days/hours/minutes)"

code/datums/preferences/preferences.dm
```diff
@@ -1269,306 +1339,189 @@ var/list/removed_jobs = list(
...
-		return
+			else
+				var/datum/antagonist/antag_datum = global.get_antagonist_datum_type(antag_role)
+				if (!antag_datum)
+					CRASH("Unrecognised antagonist role ID: [antag_role]")
+
+				antagonist_static_data[antag_role]["name"] = global.capitalize_each_word(antag_datum.display_name)
+
+			// If the user cannot play as this antagonist role.
+			var/reason_tooltip = null
```
JORJ949, line 1502: Should again probably have a fallback just so you can immediately see "oh this reason has no tooltip and needs bug reporting"

code/datums/preferences/preferences.dm
Mister-Moriarty: The code is set up in such a way that every eventuality (as checked by the old code) resulting in you being unable to play a role returns a tooltip string. If you see slightly below, disabling the role in the TGUI interface is dependent upon `reason_tooltip` not being null; setting a default value will result in every role being disabled in the interface.

`trusted_only` does not appear as no standard rotation job uses that; it is exclusively used for a single latejoin role.

code/datums/preferences/preferences.dm
Mister-Moriarty: See above.

mordent-goonstation:
Actively working on this right now, should have my changes to a state I'm happy with soon.

There's quite a lot dead code I don't think you've cleaned up related to the old UI. Would be worth removing in this PR as well.

PR #24326, opened 2025-08-03
Fixes paint for tables and chairs

RubberRats:
[BUG][GAME OBJECTS]
## About the PR 

This PR adds `KEEP_TOGETHER` to the appearance flags on tables and `obj/chair/comfy` (IE ones with armrests), allowing for paint and similar to affect their overlays.

## Why's this needed? 
Fixes #12374, Fixes #10406 

## Testing 

image

PR #24470, opened 2025-08-18
Revert "Fixes paint for tables and chairs" to fix comfy chair layering issue

snugglycactus:
Reverts goonstation/goonstation#24326 as advised by a developer in order to fix the chair layering issue. This will reintroduce the comfy chair paint bug, but I find that it's the lesser of two evils to put it in an insanely melodramatic way.

This is necessary, as presently the player appears above the arm overlay of all comfy type chairs. It is my belief that this visual bug is worse than the paint bug that goonstation/goonstation#24326 was introduced to fix.

I have tested removing this on a test server and it seemed to fix the layering issue. I don't believe testing screenshots are required, as this is just a direct reversion to something that was already in the server.

### DIFF
--- a/code/obj/table.dm
+++ b/code/obj/table.dm
@@ -15,7 +15,7 @@ TYPEINFO_NEW(/obj/table)
 	density = 1
 	anchored = ANCHORED
 	flags = NOSPLASH
-		appearance_flags = parent_type::appearance_flags | KEEP_TOGETHER
+	appearance_flags = parent_type::appearance_flags | KEEP_TOGETHER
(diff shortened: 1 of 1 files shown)


code/obj/table.dm
RubberRats, line 1: Probably should not revert the table changes, as they don't suffer from the layering problem.

snugglycactus: My mistake! Fixed.

snugglycactus:
oh god damn it why'd everything fail why didn't i just do this manually hold on

PR #24227, opened 2025-07-26
Adds the gas filter.

Laboredih123:
[ATMOSPHERICS][FEATURE]
## About the PR 
Adds the gas filter.

## Why's this needed? 
gives atmos more stuff they would really really like

## Testing 
image

## Changelog 

```changelog
(u)cringe
(*)The HPD now has the gas filter, which is able to filter out a gas.
```

PR #24487, opened 2025-08-20
fix wrong path for trinary items in hpd

Laboredih123:
[BUG][ATMOSPHERICS]
## About the PR 
i forgot to change the path when copy pasting in #24227, so i do that here

## Why's this needed? 
bug bad


Laboredih123:
i only noticed it when merging upstream for fluid pipes lol cause it doesnt affect placing filters, only when placing binary items onto another item

PR #24489, opened 2025-08-20
Prevent seed extraction scrambling plant alleles

Egregorious:
## About the PR 

Added mechanism for plant alleles (the (D)ominant and (r)ecessive portion of plant stats) to be passed along when a seed is copied. 

Alleles are passed exclusively during seed extraction.

## Why's this needed? 

Alleles are always randomised during harvest already, so it doesn't make much sense for them to be rerolled during extraction.

## Testing 

Grew fruit and vegetable to make sure fruits and seeds harvested still rerolled alleles. Then extracted some tomatoes and observed alleles were the same before and after. 

Screenshot 2025-08-19 165517


ZeWaka (approved):
I'm not too familiar with the mechanics but this sounds ok to me

PR #24490, opened 2025-08-20
Immutable Alleles gene strain

Egregorious:
## About the PR 

Added an 'Immutable Alleles' gene strain, which causes plants to always pass their alleles down to their produce and seeds.

Infusing with Mutadone can apply the Immutable Alleles gene strain with an 8% chance. 

## Why's this needed? 

I've never known alleles to be particularly exploitable, but they are pretty cool and it is a mechanic that some botanists might like to have more control over. 

Also there is very mild discontent that mutadone isn't particularly useful for botany despite being part of the vending machine chems, and this gives it more purpose. 

I can't think of anything that's obviously broken by this. It does make Man Eaters easier to scale, but since you can't control their alleles it's still imperfect.

## Testing 

Infused, grew, scanned blueberries. 
Screenshot 2025-08-20 194522

## Changelog 

```changelog
(u)Egregorious
(+)Mutadone can infuse plants with a gene strain that prevents alleles from mutating when harvested.
```


M-Earthfire:
I like the idea. I would say maybe 25% chance on mutadone is a bit high, though.

The chance isn't imbalanced, but i feel making it rarer would make it feel more special for newer players who aren't targeting it specifically.

Egregorious:
25% basically just came from the fact it averages to a bottle per strain, but there's no real reason for that considering how easy it is to replenish stocks of all the basic fertiliser chems.

NibChocolateeny:
it is interesting....so, there is no mutating stats when plant in trays?

Egregorious:
Mechanically alleles refer to the dominant and recessive portion of plant stats specifically, at least from what I gather, maybe it's a confusing name I'm not sure. 

So it's the (D) and (r) of each plant stat, and it would only affect the stats indirectly when splicing, which is when alleles come into play. I don't think alleles affect plants while they're in the tray.

PR #24528, opened 2025-08-27
Computer frames convert based on inserted board

JORJ949:
## About the PR 
Console and Computer frames will convert into each other depending on if a circuitboard or motherboard is inserted.
Removes computer frame from construction list

## Why's this needed? 
Having to remember which version of computer/console you need is incredibly annoying.

## Testing 
The north computers were built from console frames, the south were built from computer3 frames
image
image

## Changelog 

```changelog
(u)JORJ949
(*)Computer and Console frames convert into each other depending on if a Circuitboard or Motherboard is inserted.
```

### DIFF
--- a/code/modules/networks/computer3/buildandrepair.dm
+++ b/code/modules/networks/computer3/buildandrepair.dm
@@ -101,6 +101,7 @@ TYPEINFO(/obj/item/motherboard)
 				new_computer.state = 1
 				new_computer.anchored = src.anchored
 				new_computer.dir = src.dir
+				new_computer.setMaterial(src.material)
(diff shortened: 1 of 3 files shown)


JORJ949:
> Kind of a hack, but it's a good hack

The proper solution is to merge computer and computer3.
Nobody is merging computer and computer3.

Glamurio:
> > Kind of a hack, but it's a good hack
> 
> The proper solution is to merge computer and computer3. Nobody is merging computer and computer3.

Yeah kind of. This is better than what we have, even if it's jank.

PR #24549, opened 2025-08-30
Rework clarion bridge.

JORJ949:
## About the PR 
Reworks clarion bridge to remove the tight one tile wide corridors, expand EVA and make it generally a bit nicer to have multiple people in at once.
image

## Why's this needed? 
Clarion's bridge is very cramped and maze-like at the moment and it uses the space very poorly with multiple 1 tile wide corridors, which makes it hard to do crime in as to run away you have to run in a straight line when security has tasers which WILL get you.

## Testing 
image

## Changelog 

```changelog
(u)JORJ949
(*)Clarion's Bridge has been reworked.
```


pgmzeta:
* Bedrooms should get door-bolter buttons
* Captain's room should not have a direct door to customs

JORJ949:
> * Captain's room should not have a direct door to customs

I disagree, they'll often use customs and they had a direct door into EVA before, why don't you think they should have a door to customs?

I missed the door-bolters because the rest of command offices on clarion don't have them, implemented using the fancy new mappinghelpers

Egregorious:
I feel like there's a bit of a thematic failure here, in that the big central corridor that funnels the entire population into the heart of the ship every round currently culminates in the command bridge, which is a much better aura than EVA storage.

Also the current bridge is a nice homage to Star Trek, it has a captain's chair and two 'pilot' control panel areas. It's kind of a shame to lose that touch on the starship-based map.

Sord213:
I'm not a huge fan of computer core being essentially a hallway with turrets. The only access to the bridge from the south is through the comp core. And the open entry to the AI core, but with the turrets being separated between ai core and comp core, feels off. Overall I like the rest of the redesign, though.

Screenshot_412
also this disposal pipe junction is facing the wrong way

PR #24554, opened 2025-08-31
Make assemblies, suicide vests and mousetrap rollers inherit contraband level from their components and make singletank bombs contraband 4

M-Earthfire:
[Balance][Game Objects][bug]

## About the PR 

This PR makes it so that assemblies (most notable pipebombs), mousetrap rollers and suicide vests are considered contraband again, if they have any components on them that are contraband.

This also enables to make singeltank bombs contraband. This PR makes them contraband level 4.

This is the successor of  #23174, but utilizes the contraband component now.

## Why's this needed? 

Pipebomb assemblies currently dont have a contraband level. I don't think that's a good idea to have.

Also, singletank bombs are as dangerous as pipebombs, so lifting them up to the same contraband level makes sense.

## Testing 

https://github.com/user-attachments/assets/d2bd2321-cd8f-41ba-8e8c-5ee6591e8335

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Assemblies will now be considered contraband, if they have anything on them that is considered contraband
[+]Singletank bombs have a contraband level of 4, similar to pipebombs
```

### DIFF
--- a/code/obj/item/assembly/misc_assemblies.dm
+++ b/code/obj/item/assembly/misc_assemblies.dm
@@ -80,13 +80,18 @@ Contains:
 	if(new_target)
 		to_set_up_items += new_target
 		src.target = new_target
+	var/contraband_carry_to_set = 0
+	var/contraband_gun_to_set = 0
 	for(var/obj/item/checked_item in to_set_up_items)
 		checked_item.set_loc(src)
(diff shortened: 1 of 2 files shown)


M-Earthfire:
well shit, fatfingered on the enter key... gonna update the description real quick....

PR #24557, opened 2025-08-31
Artifact and cargo transporter interaction property

Egregorious:
## About the PR 

Artifacts can spawn with a hidden property that causes them to be teleport-interfering. The spawn rate is 8%. 

Once a teleport-interfering artifact has been activated, the cargo transporter has a 70% chance to send it to a random telepad instead of the assigned destination, and a 30% chance to send it to a random location on the station. This only affects activated artifacts, dormant artifacts are always teleported normally. 

Hitting a teleport-interfing artifact with an object made of telecrystal will cause it to warp up to 2 tiles away. 

This partially relies on #24556.

## Why's this needed? 

It's extremely safe to teleport artifacts between artlab and cargo with the cargo transporter, and that leaves little opportunity for antagonists to mess with them before they're sent offstation. With this feature, activated and researched artifacts will sometimes find their way randomly around the station and potentially into the hands of ne'er-do-wells. 

The telecrystal interaction makes it possible for a diligent scientist to test for this property, and thus you can blame artsci not doing their jobs properly for any misplaced cursed objects. 

## Testing 

Manually set the chances to 100%; admin-activated an artifact; copied it's path via 'view variables'; cargo-transporter'd it; and then used Find-All with the artifact path to locate it; repeated a few times.

For the telecrystal interaction I spawned in a telecrystal and whacked an activated artifact with it a few times. 

https://github.com/user-attachments/assets/07eed42d-dd2e-4237-ae94-05f9670d800f

## Changelog 

```changelog
(u)Egregorious
(*)Some activated artifacts can interfere with teleportation, and will be sent to the wrong location if moved with the cargo transporter. Diligent scientists are recommended to test for this property via telecrystal reaction. 
```


code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -320,6 +320,18 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
 	execute(var/atom/owner, var/mob/attacker, var/atom/attacked)
 		var/turf/T = get_turf(attacked)
 		var/mob/attacked_mob = attacked
+		if (istype(attacked, /obj)) // teleportationally unstable artifacts react to telecrystals so that scientists can determine that they are
```
ZeWaka:
```suggestion
		if (isobj(attacked)) // teleportationally unstable artifacts react to telecrystals so that scientists can determine that they are
```

code/modules/materials/Mat_MaterialProcs.dm
```diff
@@ -320,6 +320,18 @@ triggerOnEntered(var/atom/owner, var/atom/entering)
...
+			var/obj/target = attacked
+			if (target.artifact?.activated && target.artifact?.teleportationally_unstable)
+				attacker.visible_message(SPAN_ALERT("[attacked] reacts to [owner]!"))
+				if (target.anchored) // anchored artis shouldn't move, but there should still be feedback for the purpose of reactivity testing
+					return
```
ZeWaka, line 330: maybe make the sound no matter what then

ZeWaka (changes requested):

code/modules/materials/Mat_MaterialProcs.dm
Egregorious: Very minor artistic reasoning; I'm not sure there's an instance of the warp sound occurring when something isn't being teleported - whereas 'react' can mean anything.

Also it's a lazy way to avoid the potential of sound-spam, but that can be solved with a cooldown.

ZeWaka: i mean, fair enough - but it still should have some feedback. plenty of players don't read the side chat at all

Egregorious: I wasn't disagreeing, I was just noting considerations.

It could be a different sound, but that's probably confusing for the player. I've added the warp sound with a cooldown when it's anchored.

Kanthes:
This adds another layer to ArtSci that has the potential for antags and chaos. Love it.

PR #24558, opened 2025-08-31
Allow urls in LOOC and OOC

TobleroneSwordfish:
## About the PR 
^
The URLs still won't be clickable in chat.
## Why's this needed? 
Parity with pre say rework behaviour, people used to be able to LOOC wiki urls at each other for teaching purposes.
## Changelog 

```changelog
(u)LeahTheTech
(+)You can send URLs in OOC and LOOC channels again.
```


code/modules/speech/trees/speech_module_tree.dm
```diff
@@ -151,6 +151,12 @@
 		logTheThing(LOG_ADMIN, message.speaker, "[message.speaker] tried to say \"[message.original_content]\" but it was garbled into \"[message.content]\", which is uncool by the following effects: [jointext(modifier_ids, ", ")]. The uncool words were garbled.")
 		message.content = replacetext(message.content, global.phrase_log.uncool_words, pick("urr", "blargh", "der", "hurr", "pllt"))
 
+	// Check for URLs if we're an IC channel, let people paste wiki links at each other in LOOC if they want to
```
JORJ949, line 154: Should allow them in local and _maybe_  radio chat on classic as those aren't IC

PR #24560, opened 2025-08-31
Mindhack module bypasses puritan trait

TheGeneralJay:
[Medical][QoL]

## About the PR 
Adds a check for the mindhack module before activating the puritan reaction while cloning. This allows traitors to mindhack puritans using their fancy cloner setup.

## Why's this needed? 
Cybernetic incompatibility does not make you exempt from being conversion chambered or syndicate framed. I have seen many people talk about how bad it can feel to finally pull off your murder plot only to find out the victim is a puritan and cannot be cloned using your mindhack cloner setup. It sucks to have your plans hindered by a trait someone else chooses.

## Testing 
Tested to make sure the new added message pops up. All works as intended, the puritan reaction no longer happens when being cloned with a mindhack module.

{4F307001-A6F9-4703-845A-89F4CB9C235C}

## Changelog 

```changelog
(u)TheGeneralJay
(+)The cloning mindhack module now bypasses the puritan trait.
```

### DIFF
--- a/code/obj/machinery/clonepod.dm
+++ b/code/obj/machinery/clonepod.dm
@@ -302,30 +302,33 @@ TYPEINFO(/obj/machinery/clonepod)
 					is_puritan = TRUE
 
 			if (is_puritan)
-				src.mess = TRUE
-				// Puritans have a bad time.
-				// This is a little different from how it was before:
-				// - Immediately take 250 tox and 100 random brute
-				// - Always get a major cloning defect
-				// - 50% chance, per limb, to lose that limb
-				// - enforced premature_clone, which gibs you on death
-				// If you have a clone body that's been allowed to fully heal before
-				// cloning a puritan, you have a sliiiiiiiiiiight chance to get them
-				// out of deep critical health before they turn into chunky salsa
-				// This should be really rare to have happen, but I want to leave it in
-				// just in case someone manages to pull off a miracle save
-				defects.add_random_cloner_defect(CLONER_DEFECT_SEVERITY_MAJOR)
-				src.occupant.bioHolder?.AddEffect("premature_clone")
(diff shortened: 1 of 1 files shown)


code/obj/machinery/clonepod.dm
```diff
@@ -302,30 +302,33 @@ TYPEINFO(/obj/machinery/clonepod)
...
-				APPLY_ATOM_PROPERTY(src.occupant, PROP_HUMAN_DROP_BRAIN_ON_GIB, "puritan")
-				random_brute_damage(src.occupant, 200, 0)
-				if (ishuman(src.occupant))
-					var/mob/living/carbon/human/P = src.occupant
-					if (P.limbs)
-						var/list/limbs = list("l_arm", "r_arm", "l_leg", "r_leg")
-						for (var/limb in limbs)
-							if (prob(50))
-								P.limbs.sever(limb)
+				if (!clonehack) // If mindhack cloner, bypass puritan stuff.
```
ZeWaka: Invert this, I think it'd be best to have the message at the top of this block

TheGeneralJay: Got it!

ZeWaka (approved):

JORJ949 (approved):
Code looks good, Feature is a nice QoL for mindhack cloners.

TheGeneralJay:
commenting for un-stale

Driftered:
I want this SO unbelievably bad. I've been the unfortunate receiver of several "Killed you only to find out I can't even clone you" plots, and it's always so disappointing for both sides. This would be an incredibly-welcome change, in my book, and allow for a lot of surprising moments where someone you knew couldn't come back suddenly does. Has my approval.

PR #24579, opened 2025-09-02
Prevent Vampire Miners

JORJ949:
## About the PR 
Prevents vampires from rolling miner for their job.

## Why's this needed? 
Antagonist roles should not impact your ability to do your job normally, Vampire heavily impedes miners whose entire role revolves around being outside the station in space due to the unavoidable damage vampires get from being in space (2.5 burn and (vampire) bloodloss every life tick)

## Testing 
With Miner favourited and the round forced to vampire, instead spawned as scientist.

## Changelog 

```changelog
(u)JORJ949
(+)Vampires can no longer be assigned the Miner job.
```


flappybatpal:
Hm, a reasonable argument.

PR #24580, opened 2025-09-02
Paper Fixes

mordent-goonstation:
[UI][Bug]

## About the PR 

Attempts to fix some issues with editing Paper after the 516 upgrade and associated tgui work.

## Why's this needed? 

It is unreasonable to expect people to be able to fill in a form and submit it in 0.9 seconds.

## Testing 

With the change I was able to actually fill in the `[_____]` style fields, and generally edit paper etc.

I did notice an oddity at one point of my key presses passing through to move my spaceman around, but I could not reproduce so putting down to a mix of me not thinking the paper UI is actually intuitive and thus having different expectations as well as maybe some partial code change from being dev-ed.

Putting up as a draft so people can see, pull down easier, and let me know if this is somehow worse than what was before? I dunno, I don't do paperwork much. If you noticed a weirdness with paper feel free to try to reproduce it with this change.


mordent-goonstation:
Ignore mad linter, I'll resolve that before merging.

TobleroneSwordfish:
Tested with some of Herb's paperwork templates from the wiki, all seems to work fine 👍

PR #24582, opened 2025-09-03
Makes EMPs and powerful electric shocks disable implants temporarily

TobleroneSwordfish:
## About the PR 
- EMPs are 15 seconds, shocks are 10-30 seconds depending on the shock strength
- The status is visible on your HUD and affected players will also emit little sparks
- Defibs, multitools, tesla coils, stun gloves, stun batons aren't powerful enough to trigger the effect, it needs to be a big zap from a shocked door, overloaded APC or arcfiend/power gloves

Implant specific stuff:
- Mindhack implants are unaffected
- Counter-revved players can be converted while the implant is disabled, destroying the implant
- Cloner implants will stop reporting their heartbeat signal, meaning you might temporarily show up as dead on the clone rack.
- Access implants are disabled as you'd expect
- Revenge implants will fail to go off, not sure about this one, maybe they should be immune?
## Why's this needed? 
- It's kind of neat
- Secoffs don't get mindhacked enough, now you can lure one into the engine bay or have them touch an electrified door to get their implant disabled
- Counter-rev implants are very very polarizing and often lead to a lot of people being beaten to death, smart use of EMP grenades or electric shocks can now get around that
## Testing 
Emitting sparks:
image
HUD icon:
image
Getting mindhacked while having a disabled mind-protection implant
image

## Changelog 

```changelog
(u)LeahTheTech
(*)Getting EMPed or hit with a powerful electric shock will now temporarily disable your implants.
(+)Changes from testmerge: a little blue glowy icon on people with disabled implants, duration of disabled effect increased across the board, power gloves now always disable implants.
```


JORJ949:
What shock damaging items deal enough to disable implants? I'm wary of things like Multitools and Stun gloves rendering the mindprot implant effectively pointless. Though EMPs disabling them I think is great and does have a giveaway if a random EMP went off and an officer walks out unscathed.

Honestly I like how difficult it is to mindhack security requiring you to plan ahead, the payoff is incredible, especially if the target in question is the HoS. Currently the best way to mindhack is to engineer their death and then mindhack them after they're cloned or sleepy pen and surgery it out by hand, though these do cost almost your entire credits if you're going for the deluxe.

Scaltra-Volpe:
Damn, this is such a cool idea, but I am concerned about how easy this would make it to mindhack people. EMP grenades aren't very expensive at all, and certain shock applications are very easy to obtain and apply, as Jorj mentioned. 

I love this for stopping implant alerts, stopping tracking implants, etc, before removal. 

I do feel like people should need to work a bit harder to mindhack these critical targets, though. That being said, sometimes it's a huge hassle even getting alone with a sec officer, captain, or hos so maybe it is fine

RSOD404:
i really like this, though am a bit unsure of how the EMP gene will work with this given it may effectively permanently turn off your implants, but this makes EMPs a lot more useful.

one issue i am afraid of is shocking doing the same -- mechcomp tesla coils being enough to perhaps disable implants and arcing APCs being definitely enough to do so.
ultimately I'd rather shocks only affect the lesser implants (i.e. cloning, anti-rot, health, counter-rev) and assume the more important ones (sec, mind-prot, most syndicate implants maybe?) are higher quality and can afford to be insulated. 

(small suggestion, what if EMP'd health implants hid / faked your health?)

JORJ949:
RSOD mentioned syndicate implants, have you considered the other syndicate implants, the Signaler and Marionette?

I see the idea for a mindprot implant nerf but think there are better ways to do it that require more strategy than just emp nade > click/ multitool>click. What if it was "weakened" and destroyed on mindhacking while in crit?

TobleroneSwordfish:
I've updated the description to be clearer about what does/doesn't trigger the effect: multitools, tesla coils etc do not disable implants. Marionette implant shocks also aren't powerful enough to disable themselves.

I don't thiink this will make it too easy to mindhack people, EMP grenades are an overt threat that usually makes secoffs panic and run anyway, and if you manage to set up a hacked door and lure a secoff into touching it then you've at least put in the work and had to talk to/interact with them.

Mind protection implants disabling in crit would be a viable alternative, but I'm not a huge fan of brute damage being the answer to everything and I think this would be a neat general game mechanic regardless?

M-Earthfire:
> i really like this, though am a bit unsure of how the EMP gene will work with this given it may effectively permanently turn off your implants, but this makes EMPs a lot more useful.

I mean, that sounds like a neat tradeoff for a situational but in its effect quite powerfull gene.

In the same direction, affecting syndicate implants should fully be a thing. This makes it so that emp-effects, which are very powerfull, are a kind of tradeoff you go, like by using sechuds with emp-nades.

So in general, i really like how the PR is right now, it enables tactical choice.

RSOD404:
> In the same direction, affecting syndicate implants should fully be a thing. This makes it so that emp-effects, which are very powerfull, are a kind of tradeoff you go, like by using sechuds with emp-nades.

definitely, i'm just saying i wouldn't want shocks being able to disable syndicate, only EMP should do that

JORJ949:
Not sure if they are intended to but power gloves don't pack enough punch on stun mode (didn't test harm) if you don't hotwire.

code/modules/status_system/statusEffects.dm
```diff
@@ -3866,3 +3866,30 @@
...
+			for (var/obj/item/implant/implant as anything in living_owner.implant)
+				implant.deactivate()
+
+	onRemove()
+		..()
+		src.owner.UpdateParticles(null, src.id)
+		if (isliving(src.owner))
+			var/mob/living/living_owner = src.owner
+			for (var/obj/item/implant/implant as anything in living_owner.implant)
+				implant.activate()
```
JORJ949, line 3951: This will probably perma break any implants you EMP > cut out. I reccomend for the status to store a list of all the implants it deactivated then turn them back on when it runs out.

Tarmunora: Or control the (de)activation via a status on the implant 🤔

Tarmunora:
I'm a bit concerned about the emp grenade -> revbang grenade combo.
Revbangs already need both eye _and_ ear protection to block without a counterrev implant, and if you add in a "turns off counterrev implants" grenade, there's very very little that can then be done to stop an entire roomful of loyalists from getting mass-converted all at once

PR #24586, opened 2025-09-03
Disables harbinger wraith on RP

TobleroneSwordfish:
## About the PR 
See title, plaguebringer and trickster are still enabled.
## Why's this needed? 
Harbinger is based around the idea of just summoning as many minions as possible and overwhelming the crew. In theory there *is* roleplay to be had in that, but in practice it tends to quickly turn rounds into chaotic "everyone for themselves" slaughter fests where security get carpel tunnel and medical use 100u of styptic a minute.
I did think the wraith/summons chat channel would help with controlling the chaos a little but it really doesn't seem to have improved much.
## Testing 
image

## Changelog 

```changelog
(u)LeahTheTech
(*)Harbinger wraith is now disabled on the roleplay servers.
```

### DIFF
--- a/code/modules/antagonists/wraith/abilties/specialize.dm
+++ b/code/modules/antagonists/wraith/abilties/specialize.dm
@@ -6,7 +6,11 @@
 	pointCost = 150
 	tooltip_options = list("align" = TOOLTIP_LEFT | TOOLTIP_CENTER)
 	special_screen_loc = "NORTH-1,EAST"
+#ifdef RP_MODE
+	var/static/list/paths = list("Rot" = 1, "Trickster" = 3)
+#else
 	var/static/list/paths = list("Rot" = 1, "Summoner" = 2, "Trickster" = 3)
+#endif
 	var/list/paths_buttons = list()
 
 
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/wraith/abilties/specialize.dm
```diff
@@ -6,7 +6,11 @@
 	pointCost = 150
 	tooltip_options = list("align" = TOOLTIP_LEFT | TOOLTIP_CENTER)
 	special_screen_loc = "NORTH-1,EAST"
+#ifdef RP_MODE
+	var/static/list/paths = list("Rot" = 1, "Trickster" = 3)
```
ZeWaka, line 10: 3?

kbsmilk:
I'm in favor of also disabling plaguebringer, along with harbinger

M-Earthfire:
To be honest, i don't think the abilities are necessarily non-rp friendly. I feel the point generation is not properly balanced for RP, as i have seen wraiths sit on 6k points with only 7 absorbed bodies (not even all of them being crewmembers, monkeys are common to absorb) and not much ability to spend these. So much points enable the spam of abilities, which inevitably will lead to the scenario described as the reason to disable harbinger.

aaaaaaaaaaaaaaaaaah:
This should be discussed on the forum.
 I am also against this since I actually like to have a really exiting round once in a while, in 500 hours I have only ever had one Harbinger round where I was really challenged so I don't think it is something that will get boring soon.

code/modules/antagonists/wraith/abilties/specialize.dm
TobleroneSwordfish: Yep, it uses numerical IDs - 3 is trickster (yes this is dumb)

ZeWaka: what if you fixed this while you were here 🥺

TobleroneSwordfish: :(

TobleroneSwordfish:
oki

PR #24610, opened 2025-09-07
Split Jobs.dm into multiple files

JORJ949:
## About the PR 
Splits Jobs.dm into multiple files. 
Noticable change: Special jobs are ordered differently in the controller (staple jobs remain the same as they are handled differently) and thus appear in a different order in the respawn as job list and the Job controls panel (and probably some other places). They are now ordered NT - Syndie - Halloween - Random(the ones randomly picked each round) - All the rest (in the same order they were before).

## Why's this needed? 
Jobs.dm is 3000 lines long and every new job adds more and more, splitting them up into files allows you to find the job you're looking for easier, and makes adding new jobs not add to this 3k line file.

## Testing 
image


code/datums/jobs/job_parent.dm
```diff
@@ -0,0 +1,247 @@
...
+
+#define NANOTRASEN_LINK_COLOR "#3348ff"
+#define SYNDICATE_LINK_COLOR "#800"
+
+#define ANNOUNCE_ORDER_CAPTAIN 5
+#define ANNOUNCE_ORDER_HOP 4
+#define ANNOUNCE_ORDER_HOS 3
+#define ANNOUNCE_ORDER_HEADS 2
+#define ANNOUNCE_ORDER_LAST 1
+#define ANNOUNCE_ORDER_NEVER 0
```
Mister-Moriarty: Instead of defining these here, I'd move them to `_std/defines/jobs.dm`.
Defining them here makes them available to all code parsed after this file but not before; defines should either be global (hence the `_std/defines` folder) or local (undefined at end of file).

(I wish we had namespaces.)

JORJ949: I was going to do that but forgot where the file was and then that i was going to so that, thanks.

PR #24617, opened 2025-09-08
Syndicate description macro

JORJ949:
## About the PR 
Adds a macro for adding traitor (and spief) only descriptions to objects, intended for stealth items that you wouldn't want to get mixed up with the real deal, but only applied to the syndicate shrub for this PR

## Why's this needed? 
Lays the groundwork for simpler adding of identifiable descriptions to syndicate stealth items

## Testing 
Syndicate shrub description works as intended, "latest in spy tech" as syndie/spief, "is that an antenna" otherwise.

### DIFF
--- a/_std/macros/descriptions.dm
+++ b/_std/macros/descriptions.dm
@@ -1,0 +1,7 @@
+// Macro to quickly add syndicate-specific descriptions to stealth items to make them easily identifiable to traitors (and spiefs who use their gear)
+#define SYNDICATE_STEALTH_DESCRIPTION(TYPE, syndie_desc, alt_desc) \
+	TYPE/get_desc(dist, mob/user) { \
+		..(); \
+		if(istrainedsyndie(user) || isspythief(user)) {. += SPAN_ALERT("<b> [syndie_desc]</b>")} \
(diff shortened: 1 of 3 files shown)


_std/macros/descriptions.dm
```diff
@@ -0,0 +1,7 @@
+// Macro to quickly add syndicate-specific descriptions to stealth items to make them easily identifiable to traitors (and spiefs who use their gear)
+#define SYNDICATE_STEALTH_DESCRIPTION(TYPE, syndie_desc, alt_desc) \
```
ZeWaka: what is alt desc

this is not clearly explained in the comment

_std/macros/descriptions.dm
```diff
@@ -0,0 +1,7 @@
+// Adds a syndicate only description to any type. syndie_desc is shown to syndies and spiefs, alt_desc is shown to everyone else, both are optional.
+#define SYNDICATE_STEALTH_DESCRIPTION(syndie_desc, alt_desc) \
+	get_desc(dist, mob/user) { \
+		..(); \
```
Tarmunora:
```suggestion
		. = ..(); \
```

PR #24618, opened 2025-09-08
Rework Law Linker

JORJ949:
## About the PR 
Reworks the law linker to use a 10 second action bar instead of instant with open panel.
- Information on where the target silicon is connected to is instead in the confirmation popup rather than chat
- Syndicate and Emagged cyborgs will only fail to be reconnected AFTER the actionbar is completed.
- Emagged cyborgs will still tell you they are not connected to a law rack.
- Syndicate cyborgs will act like they're connected to the default rack, until you try to connect them to another rack

Law linkers now start connected to the default law rack
Law linkers can be examined to see the connected rack

## Why's this needed? 
Law linkers are incredibly underused both for their intended purpose and for crime, this allows crew and criminals who do not have robotics access to stun and reprogram a cyborg (and probably signal jam) to their own law rack. This also nerfs the speed at which anyone WITH robotics access could just walk up and reprogram you as quickly as they'd like, as the actionbar is the ONLY way to reprogram.

Syndicate borg changes are a result of anyone being able to just click a borg with a law linker to see the popup and so would reveal every single syndieborg with one click.

## Testing 
Features described in the About the PR work as tested.
image
image
image

## Changelog 

```changelog
(u)JORJ949
(*)Law Linkers now start a 10 second action bar when linking a silicon, and no longer require the silicon to be open.
(+)Law Linkers now start connected to the default rack and can be examined to see their connected rack.
(+)Syndicate cyborgs will act as if they are already connected to the default rack when linking.
```

### DIFF
--- a/code/obj/item/device/borg_linker.dm
+++ b/code/obj/item/device/borg_linker.dm
@@ -64,7 +64,8 @@ TYPEINFO(/obj/item/device/borg_linker)
 		if(sillycon.shell || sillycon.dependent)
 			boutput(user, SPAN_ALERT("You need to reprogram the AI's mainframe!"))
 			return
-		if(sillycon.law_rack_connection == src.linked_rack)
+		//Syndicate cyborgs are totally not real guys they're connected to the default rack you gotta believe me they're not a syndieborg i swear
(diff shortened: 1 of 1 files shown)


JORJ949:
Needs atomising

PR #24632, opened 2025-09-11
Fix kitchen island and satchel interaction

### LINKED ISSUE #24627: Kitchen Island Teleporation
### Labels

[BUG][MAJOR]

### Description

Food that was previously on a kitchen island by produce satchel will teleport back to that island once the island is moved.

### Steps to reproduce

1. Click drag food onto a kitchen island from a produce satchel
2. Take the food off of the kitchen island and place it anywhere or in your produce satchel
3. Move the kitchen island and then it will either teleport the food back on the kitchen island or remove it from your produce satchel.

### Expected Behavior

The food would stay where it was placed.

Reported by: pixeox
Client version: 516.1667
Active test merges: #24159 and #24582
Round log date: 2025-09-09-09-00
Reported on: 2025-09-09 09:56:03
Round ID: [212538](https://goonhub.com/rounds/212538)

Egregorious:
## About the PR 

This affects surgery trays and kitchen islands. Items are properly detached from the island when quickfilled into the satchel, and satchels can dump their contents onto it. This stops items teleporting back onto the island after being quickfilled into satchels.

Added the COMSIG_ITEM_STORED signal which is triggered whenever an item is stored in a satchel or item container. Surgery trays register this to detach items when they're quickfilled into a satchel. 

Surgery tables override `place_on` for attaching items. Satchels use `place_on` to dump their contents instead of `set_loc`.

## Why's this needed? 

Fixes #24627

Surgery trays and kitchen islands are moveable tables, you should be able to dump satchels out onto them the same way you can do regular tables.

## Testing 

Before:

https://github.com/user-attachments/assets/3f3f057f-d992-41c6-99dc-3ec2668b42e8

After:

https://github.com/user-attachments/assets/4abf55ee-1665-4a61-958f-e55c5c138e55

### DIFF
--- a/code/obj/item/satchel.dm
+++ b/code/obj/item/satchel.dm
@@ -188,7 +188,7 @@
 			if (length(src.contents) < 1)
 				boutput(usr, SPAN_ALERT("There's nothing in [src]!"))
 			else
-        var/obj/table = over_object
+				var/obj/table = over_object
 				usr.visible_message(SPAN_NOTICE("[usr] dumps out [src]'s contents onto [over_object]!"))
 				for (var/obj/item/thing in src.contents)
 					table.place_on(thing)


code/obj/item/satchel.dm
```diff
@@ -164,6 +164,7 @@
 				I.add_fingerprint(user)
 				if (!max_stack_reached && (length(src.contents) < src.maxitems)) // if we split up the item and it was more than the satchel can find we should not add the rest
 					I.set_loc(src)
+					SEND_SIGNAL(I, COMSIG_ITEM_PICKUP, user)
```
TobleroneSwordfish: I feel like this will break a lot of things, what is the pickup behaviour you're trying to get and could you just call that part directly instead?

_std/defines/component_defines/component_defines_atom.dm
```diff
@@ -141,7 +141,7 @@
 	#define COMSIG_ITEM_EQUIPPED "itm_equip"
 	/// When an item is unequipped (user)
 	#define COMSIG_ITEM_UNEQUIPPED "itm_unequip"
-	/// When an item is picked up (user)
+	/// When an item is picked up (user) - the item is not always in the user's hands
```
TobleroneSwordfish: .

code/obj/item/satchel.dm
Egregorious: The signal is how the tray knows to detach an item, so it doesn't try to teleport the item back to itself when it gets dragged. A direct call requires the satchel to check for a surgery table on the location of every item it tries to pick up, which seems wildly inefficient. 

The better solution is probably to add a signal for putting an item inside another item, or maybe just for quickfills, so I'll do that. Could also add a signal for any time `set_loc` is called, but perhaps that's too expensive I don't know.

Egregorious: I added COMSIG_ITEM_STORED, which is triggered when an item is stored in a container or a satchel. Tested it, it works and obviously it doesn't break anything to do with COMSIG_ITEM_PICKUP.

Tarmunora (approved):

PR #24633, opened 2025-09-11
Latejoin TGUI

Mister-Moriarty:
## About The PR:
Converts the latejoin menu into TGUI, with most features being one-to-one with the original UI, save for priority roles, which are now marked with a star and tooltip.

Two new TGUI components have been added; `ColorSection` and `ColorCheckbox`. `ColorSection` is a wrapper for `Section` that enables the section's header colour to be set to one of the TGUI colour presets. `ColorCheckbox` is a wrapper for `Button.Checkbox` that allows custom selected and disabled colours to be set, again the set colour requiring to be one of TGUI presets.

In addition to this, custom TGUI colour presets may now be defined, with this PR adding new `navy` and `crimson` colour presets.

At the backend, `/datum/job/var/link_color` has been deprecated along with its associated defines, being replaced by a `ui_colour` variable and `TGUI_COLOUR_*` defines.

## Why Is This Needed?
Interface modernisation.

## Testing:
image
image
image

## Changelog:
```changelog
(u)Mr. Moriarty
(*)The latejoin window has been converted into TGUI.
```


code/datums/latejoin_menu.dm
```diff
@@ -0,0 +1,183 @@
...
+
+	var/list/silicon_department = list()
+	silicon_department["name"] = "Silicon"
+	silicon_department["colour"] = TGUI_COLOUR_GREY
+	silicon_department["jobs"] = src.get_silicon_jobs()
+
+	var/list/special_jobs = list()
+	special_jobs["name"] = "Special Jobs"
+	special_jobs["colour"] = TGUI_COLOUR_BLUE
+	special_jobs["jobs"] = GET_SPECIAL_JOBS(user, /datum/job/daily, /datum/job/special, /datum/job/created)
```
JORJ949, line 62: Daily jobs in the code appear both here and in the civilian department list, in the screenshots they appear in the civilian section, which are they meant to appear in?

code/datums/latejoin_menu.dm
```diff
@@ -0,0 +1,183 @@
...
+			J = get_singleton(/datum/job/civilian/cyborg)
+			job_name = "[S.name] (Cyborg)"
+
+		var/list/silicon_job_props = list()
+		silicon_job_props["job_name"] = job_name
+		silicon_job_props["priority_role"] = FALSE
+		silicon_job_props["has_wiki_link"] = !!J.wiki_link
+		silicon_job_props["job_ref"] = ref(J)
+		silicon_job_props["silicon_latejoin"] = ref(silicon_latejoin)
+		silicon_job_props["colour"] = "grey"
```
JORJ949: I see no reason this shouldn't be J.ui_colour

code/datums/latejoin_menu.dm
Mister-Moriarty: `GET_JOBS` and `GET_SPECIAL_JOBS` iterate through `staple_jobs` and `special_jobs` respectively; daily jobs can appear in both lists, with admin-enabled daily jobs appearing in the latter list, and thus under the "Special Jobs" section. See `job_controls.dm:38`.

code/datums/latejoin_menu.dm
Mister-Moriarty: I meant to update that, good catch!

code/datums/latejoin_menu.dm
JORJ949: I believe they can also be requested by players using RoleControl, but yep makes sense to me.

_std/tgui.dm
```diff
@@ -36,3 +36,62 @@
 #define TGUI_CREATE_MESSAGE(type, payload) ( \
 	"%7b%22type%22%3a%22[type]%22%2c%22payload%22%3a[url_encode(json_encode(payload))]%7d" \
 )
+
+
+//------------ TGUI Colours ------------//
```
ZeWaka, line 41: `our` 💀

tgui/packages/tgui/styles/colors.scss
```diff
@@ -9,30 +9,39 @@
...
-  'brown': var(--color-brown),
-  'grey': var(--color-grey),
-  'gray': var(--color-grey),
-  'light-grey': var(--color-light-grey),
-  'light-gray': var(--color-light-grey),
-  'good': var(--color-good),
-  'average': var(--color-average),
-  'bad': var(--color-bad),
-  'label': var(--color-label),
+$_gen_map: map.merge(
```
ZeWaka, line 20: i don't think this map is even used, is it

tgui/packages/tgui/styles/components/ColorSection.scss
```diff
@@ -0,0 +1,47 @@
...
+  .ColorSection--color--#{$color-name} {
+    --color-section-base-color: hsl(
+      from #{$color-value} h s calc(l - var(--adjust-color))
+    );
+
+    --color-section-title: oklch(
+      from var(--color-section-base-color) calc(l * 0.825) calc(c * 0.75) h
+    );
+
+    @if $color-name == 'black' {
```
ZeWaka, line 35: how does this UI look in lightmode

tgui/packages/tgui/styles/colors.scss
Mister-Moriarty: `$_gen_map` is used to populate `$color-map`. See line 49.

tgui/packages/tgui/styles/components/ColorSection.scss
Mister-Moriarty: PR body updated with an example.

_std/tgui.dm
TobleroneSwordfish: ☭

TobleroneSwordfish:
This looks great, my only criticism is that the priority role is muuuch less eye-catching than before. How hard would it be to keep the outline or something similar?

tgui/packages/tgui/interfaces/LatejoinMenu/index.tsx
```diff
@@ -0,0 +1,244 @@
...
+      </Stack>
+    </ColorSection>
+  );
+};
+
+const PriorityRoleTooltip = () => {
+  return (
+    <Box
+      mt="0.5em"
+      dangerouslySetInnerHTML={{
```
pgmzeta: What is `dangerouslySetInnerHTML` getting around / fixing here?

_std/tgui.dm
```diff
@@ -36,3 +36,62 @@
...
+	TGUI_COLOUR_RED			= hsl2rgb(0, 70, 50),
+	TGUI_COLOUR_ORANGE		= hsl2rgb(25, 90, 50),
+	TGUI_COLOUR_YELLOW		= hsl2rgb(50, 97.5, 50),
+	TGUI_COLOUR_OLIVE		= hsl2rgb(70, 75, 45),
+	TGUI_COLOUR_GREEN		= hsl2rgb(140, 70, 40),
+	TGUI_COLOUR_TEAL		= hsl2rgb(180, 100, 35),
+	TGUI_COLOUR_BLUE		= hsl2rgb(210, 65, 47.5),
+	TGUI_COLOUR_VIOLET		= hsl2rgb(260, 60, 50),
+	TGUI_COLOUR_PURPLE		= hsl2rgb(290, 60, 50),
+	TGUI_COLOUR_PINK		= hsl2rgb(325, 70, 50),
```
pgmzeta, line 90: very, very red pink compared to current colors IMO. The ID computer TGUI uses a closer pink but they should probably use the same one either way (in separate PR if needed)

pgmzeta (changes requested):
Please add a special identifier for player requested roles akin to priority roles; check `player_requested` on the job datum

_std/tgui.dm
ZeWaka: Probably in like a common/colors.ts

Calphrick:
I'm going to go against the grain here and say I don't like the way this looks. It's too, I dunno, corporate? Doesn't have a lot of character to it. It's also hard to see what the priority role is. The bright yellow highlight works great, a small star not so much.

tgui/packages/tgui/interfaces/LatejoinMenu/index.tsx
Mister-Moriarty: In order to use HTML tags like `<b>` or `<i>` in strings, TypeScript requires the string to be placed inside of an element's inner HTML. The same trick is used for the occupation UI's tooltips.

_std/tgui.dm
Mister-Moriarty: Likely best for a separate PR, but that should be as simple as changing the value of `--color-pink` or creating a `--color-light-pink` variable. I'd probably put it in `css-variables.scss`, where the custom colours are. The above define-to-rgb map is currently only used to determine the colour of departmental post.

mordent-goonstation:
Using this screenshot, I don't get why Detective is white text when e.g. Roboticist is black text.
image

Mister-Moriarty:
> Using this screenshot, I don't get why Detective is white text when e.g. Roboticist is black text.

Ah, that's probably a bug with light mode. It looks like being disabled doesn't update the text colour of the button, even though the background colour is fixed.

ZeWaka:
> I'm going to go against the grain here and say I don't like the way this looks. It's too, I dunno, corporate? Doesn't have a lot of character to it. 

I agree, however this is just a re-implementation of an existing interface. We can do design changes later.

PR #24638, opened 2025-09-11
Makes it able to poke holes into the blindfold

SunWuDrakios:
## About the PR 
It makes it able to poke holes into the blindfold similar to the eyepatch

## Why's this needed? 
For style reasons so you can wear a blindfold while still being able to see things, otherwise maybe some antags find fun ways to fake that a monkey is actually blinded by the blindfold

## Testing 
Blindfold no holes

blindfold see

## Changelog 

```changelog
(u)Drakios
(+)Blindfolds can have holes piked in them now - win every time while hitting a pinata 
```

### DIFF
--- a/code/obj/item/clothing/glasses.dm
+++ b/code/obj/item/clothing/glasses.dm
@@ -89,7 +89,8 @@
 			src.pinhole = TRUE
 			src.block_vision = FALSE
 			boutput(user, SPAN_NOTICE("You poked two holes into the blindfold, now you can pretend that you can see without seeing"))
-		. = ..()
+		else
+			. = ..()
 
 	get_desc()
 		if (src.pinhole)


SunWuDrakios:
oh yeah makes sense

code/obj/item/clothing/glasses.dm
```diff
@@ -78,11 +78,20 @@
...
+	var/pinhole = FALSE
 
 	setupProperties()
 		..()
 		setProperty("disorient_resist_eye", 100)
 
+	attackby(obj/item/I, mob/user)
+		if ((isscrewingtool(I) || istype(I, /obj/item/pen)) && !pinhole)
+			pinhole = 1
+			block_vision = 0
```
Mister-Moriarty:
```suggestion
		if ((isscrewingtool(I) || istype(I, /obj/item/pen)) && !src.pinhole)
			src.pinhole = TRUE
			src.block_vision = FALSE
```
Simple code style suggestions:
`src` should be used where possible. `src` just refers to the "source" object that the proc is defined on; here, the blindfold.
`TRUE`/`FALSE` should be used for booleans, even if BYOND doesn't have a boolean type.

SunWuDrakios: If I click commit suggestion does it automaticly change the code to that you send?

code/obj/item/clothing/glasses.dm
```diff
@@ -78,11 +78,24 @@
...
+			src.pinhole = TRUE
+			src.block_vision = FALSE
+			boutput(user, SPAN_NOTICE("You poked two holes into the blindfold, now you can pretend that you can see without seeing"))
+		. = ..()
+	get_desc()
+		if (src.pinhole)
+			. = "Wait? There are tiny holes in it!"
+
+
+
```
ZeWaka:
```suggestion
		. = ..()

	get_desc()
		if (src.pinhole)
			. = "Wait? There are tiny holes in it!"

```

code/obj/item/clothing/glasses.dm
```diff
@@ -78,11 +78,23 @@
...
+	attackby(obj/item/I, mob/user)
+		if ((isscrewingtool(I) || istype(I, /obj/item/pen)) && !src.pinhole)
+			src.pinhole = TRUE
+			src.block_vision = FALSE
+			boutput(user, SPAN_NOTICE("You poked two holes into the blindfold, now you can pretend that you can see without seeing"))
+		. = ..()
+
+	get_desc()
+		if (src.pinhole)
+			. = "Wait? There are tiny holes in it!"
```
ZeWaka:
```suggestion
			. += " Wait? There are tiny holes in it!"
```
otherwise the normal desc would be lost

ZeWaka: fix desc

code/obj/item/clothing/glasses.dm
```diff
@@ -78,11 +78,23 @@
...
 	setupProperties()
 		..()
 		setProperty("disorient_resist_eye", 100)
 
+	attackby(obj/item/I, mob/user)
+		if ((isscrewingtool(I) || istype(I, /obj/item/pen)) && !src.pinhole)
+			src.pinhole = TRUE
+			src.block_vision = FALSE
+			boutput(user, SPAN_NOTICE("You poked two holes into the blindfold, now you can pretend that you can see without seeing"))
+		. = ..()
```
ZeWaka:
```suggestion
	attackby(obj/item/I, mob/user)
		if ((isscrewingtool(I) || istype(I, /obj/item/pen)) && !src.pinhole)
			src.pinhole = TRUE
			src.block_vision = FALSE
			boutput(user, SPAN_NOTICE("You poked two holes into the blindfold, now you can pretend that you can see without seeing"))
		else
			. = ..()
```

you need the else or you'll then hit the blindfold with the screwdriver in chat

PR #24639, opened 2025-09-12
Removes material wave filters from honey blobs specifically

NibChocolateeny:
[MATERIALS][SPRITES][REWORK]

## About the PR
make it so honey blob look normal, and is not affect by transfrom apply by #24349 ...but only honey blob!

## Why's this needed? 
I am big fanm of hony sprite! it look a bit not good with wobbly gobbly effect...oteher think so, too. So blob do not have effect anymore. BUT hoeny effect still can happne on other sprite!
no oozeing
theominiousooze

## Testing
Test with specil color honey blob, big blob, etc.
nowaving

no changinglog. probaply?

### DIFF
--- a/code/modules/cooking/food_and_drink/ingredients.dm
+++ b/code/modules/cooking/food_and_drink/ingredients.dm
@@ -392,6 +392,11 @@ TYPEINFO(/obj/item/reagent_containers/food/snacks/ingredient/honey)
 	mat_changename = "honey"
 	default_material = "honey"
 
+	New()
+		..()
+		src.clear_filters() //Overrides the wave-filter normally applied to materials made of honey.
+
+
 /obj/item/reagent_containers/food/snacks/ingredient/royal_jelly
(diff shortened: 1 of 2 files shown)


LorrMaster:
I think it would be better if "/datum/materialProc/honey_add" checked if the default material ID is "honey" before applying the filter. That would be more consistent with the other material procs.

TobleroneSwordfish:
> I think it would be better if "/datum/materialProc/honey_add" checked if the default material ID is "honey" before applying the filter. That would be more consistent with the other material procs.

Would be better if the filter parent checked `uses_default_material_appearance` more generally, we already have a var for this!!

NibChocolateeny:
@TobleroneSwordfish @LorrMaster okai! I made better solushion, I think! please check....

TobleroneSwordfish:
ouugh this whole system is such spaghetti, this works for now though

PR #24642, opened 2025-09-12
Renames "rotbusttec" to "Rot-bust"

aaaaaaaaaaaaaaaaaah:
## About the PR 
Renames "rotbusttec" to Rot-bust
## Why's this needed? 
Being able to tell what an implant does at a glance and without needing to know community lingo is good.
## Testing 
It's just a name change so I didn't test it.

## Changelog 

```changelog
(u)Gomp
(+)Renamed the "rotbustsec" implant so it's now called "Anti-Rot" implant.
```


M-Earthfire:
I don't know if clarity is worth loosing the pun on "robust" here.

aaaaaaaaaaaaaaaaaah:
Clarity is certainly worth loosing any pun over, especially on a medical item.

TobleroneSwordfish:
Rotten PR!

kbsmilk:
You could compromise and use it like a brand name: `Rotbusttec anti-rot implant`

aaaaaaaaaaaaaaaaaah:
I just don't know how to approach feedback like this. 
You all don't like the change presumably because the name is funny to you but if you don't know the origin of the name or never properly read the name (it's probably the easiest name to misread I've ever seen since it's supposed to be a pun) like me it's not possible for you to know what this item does! 
It's a useful item that can have a serious impact on peoples rounds if it's actually used but right now not a single person will ever use it unless they are explicetly told what it does because it's just not intuitive!

pgmzeta:
maybe it could be named "Rot-bust" to keep the pun and also make it clearer what it does. adding "tec" to the end and having no hyphens makes it a lot harder to get the pun imo

Valtsu0:
Having "tec" at the end is weird anyways unless it's supposed to be a bootleg since it would probably infringe on BonkTek's trademark

Valtsu0:
people mistaking it for robusttec is funny but probably worth losing

aaaaaaaaaaaaaaaaaah:
Mistaking it for robusttec is the whole reason I wanted to change the name. I'll change it to Rot-Bust because that will at least marginally improve the ability of new medical players to tell what something does (god forbid someone doesn't have the wiki open during every job) but I just don't get why some random medical item needs to have a 5 year old pun in it.

aaaaaaaaaaaaaaaaaah:
Rot-bust sounds like a traitor item.

ZeWaka:
this is better but still eh

aaaaaaaaaaaaaaaaaah:
> this is better but still eh

Is it too unintuitive or not close enough to the original pun for you?

Tarmunora:
> people mistaking it for robusttec is funny but probably worth losing

The joke with the robusttec was the primary reason for the item's existence vov

PR #24644, opened 2025-09-13
Wraith salt rework

TobleroneSwordfish:
## About the PR 
Salt no longer reveals wraiths, it instead blocks them and can be passed by shoving against it repeatedly in exchange for 15 health. Shoving into the salt visibly moves it.
Salt now blocks wraith abilities based on line of sight. Salt that blocks a wraith ability will hiss and glow black for a second.
## Why's this needed? 
The polarizing nature of salt is widely hated, wraiths dying instantly because they accidentally glided over a table with a salt pile under it never feels good. This PR is an attempt to fix that without increasing wraith's overall power/threat level.
Now you can draw circles of salt around areas to make them actually immune to wraith effects, or try to trap the wraith in an enclosed region, forcing them to take health damage to escape.
## Testing 
image
image

## Changelog 

```changelog
(u)LeahTheTech
(*)Salt no longer reveals wraiths, instead it blocks them and can be shoved through for a small health cost.
(*)Salt now blocks wraith abilities based on line of sight.
```

### DIFF
--- a/code/obj/decal/cleanable.dm
+++ b/code/obj/decal/cleanable.dm
@@ -1361,6 +1361,7 @@ var/list/blood_decal_violent_icon_states = list("floor1", "floor2", "floor3", "f
 	sample_reagent = "salt"
 	sample_verb = "scrape"
 	var/health = 30
+	var/bump_stacks = 0
 
 	New()
 		..()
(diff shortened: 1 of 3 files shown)


SunWuDrakios:
I really like the idea, in some cases I wasnt against salt but sometimes like you mentioned there is just a little pile that you either dont see or that is hidden and you lose a life because of that. I also think it makes more sense to me because in pop culture salt is often used for protection instead of aggression I think this change resembles this more and still lets the wraith feel "powerful" in the sense that they can still breach the barrier but it costs them a little bit.

Also maybe it would be funny if normal ghosts just for the fun of it can bump into salt walls and hear a sound similar to the sound a tray makes when you hit someone and they "fall over"

code/obj/decal/cleanable.dm
```diff
@@ -1377,6 +1378,27 @@ var/list/blood_decal_violent_icon_states = list("floor1", "floor2", "floor3", "f
 		if (T)
 			updateSurroundingSalt(T)
 
+	/// When a wraith or similar bumps into this, returns TRUE if destroyed
+	proc/wraith_bump(mob/living/intangible/wraith/wraith)
+		if (ON_COOLDOWN(src, "wraith_bumped", 1 SECOND))
+			return
+		hit_twitch(src)
+		src.bump_stacks += 1
+		if (src.bump_stacks >= 3)
```
JORJ949, line 1387: I think this should be an actionbar similar to forcing your way through an interdicted rift. Actionbars are more noticable that you're doing something than a simple chat message.

JORJ949 (changes requested):

code/obj/decal/cleanable.dm
JORJ949: Also means you dont have to process() salt piles.

TobleroneSwordfish: I actually think in this case an actionbar would be worse, the issue with the "push through" type actionbars is that it only actually requires you to bump into the thing once, making it very easy to get accidentally dragged through/over the thing (watch a new player monkey try to stand in front of a table)
With this mechanic taking health, accidentally pushing through would be incredibly frustrating. Also leaving it as a bump action means every distinct "attempt" to shove through the salt shows up to the crew as well, giving them valuable information about where the wraith is.

JORJ949: You could have it check if they are still holding the key to move in that direction to continue the actionbar, but i was unaware the salt would move while you were breaking through so that's probably enough visual clarity without needing an actionbar.

PR #24648, opened 2025-09-13
Code tidy up of DiskRack

mordent-goonstation:
[UI][Code Quality]

## About the PR 

- Separates DiskDrive component, with additional Disk component
- Add an eject button a la IRL drives

## Why's this needed? 

Is already shared between cloning console and disk drive, Goonstation standard interface components should be available as a component rather than within one of the interfaces.

Eject button slight skeuomorphic nudge. 

## Testing 

Works in current usages.

image
image


code/obj/machinery/clone_rack.dm
```diff
@@ -39,7 +39,7 @@ TYPEINFO(/obj/machinery/disk_rack)
 /obj/machinery/disk_rack/ui_interact(mob/user, datum/tgui/ui)
 	ui = tgui_process.try_update_ui(user, src, ui)
 	if (!ui)
-		ui = new(user, src, "DiskRack")
+		ui = new(user, src, "DiskRack", src.name)
```
TobleroneSwordfish, line 42: oh I deliberately didn't do this because it'll show up labels which always feels weird, maybe use `initial(src.name)` instead?

TobleroneSwordfish (approved):
Looks good, fixing my laziness not following proper component convention 🙏

PR #24650, opened 2025-09-13
Lint for britishness

TobleroneSwordfish:
## About the PR 
Adds a grep check for `colour` anywhere in the codebase.
Also cleans up the grep check files because they were wildly out of sync and had duplicate entries for some reason.
## Why's this needed? 
Consistency plus lowering @mordent-goonstation 's blood pressure
## Testing 

This is the test!


JORJ949:
Devastation.

aaaaaaaaaaaaaaaaaah:
Sure, though colour is the better spelling.

PR #24652, opened 2025-09-13
Refactors clothing slots to use a bitflag instead of istype (partly)

Valtsu0:
[clothing]
## About the PR 
What can have these flags without runtimes are currently still pretty much limited to what they were previously as described in _std/defines/clothing.dm.
belts and backpacks still use c_flags due to me not wanting this to be a 150 file pr

only player facing change should be being able to wear cat ears on ears and head without putting a radio in. technology

## Why's this needed? 
Originally i wanted to be able to wear the plunger in the mask slot, but quite a lot of things are needed to make that possible
With the changes that are in this pr wearing shoes or gloves as hats could be added (any spriters?)

less istype

## Testing 
I have tested everything i thought of, including critters and hattables. There is still a decent chance i missed something though

## Things I'd prefer to do in separate prs but I'll do them in this one if it would be better

- change belts and backpacks to use this system
- combine the 4 or so different seals hair variables into one c_flag so any item could be worn on the head
- see what would be needed to be done to relax the path restrictions on other slots

### DIFF
--- a/_std/defines/clothing.dm
+++ b/_std/defines/clothing.dm
@@ -82,8 +82,6 @@ var/list/all_slots = list(SLOT_BACK, SLOT_WEAR_MASK, SLOT_L_HAND, SLOT_R_HAND, S
 #define ONBELT						(1<<18)
 /// block miasma inhalations (sterile mask)
 #define BLOCKMIASMA					(1<<19)
-/// covers the person's hair
-#define COVERSHAIR					(1<<20)
 
 //Suit blood flags
 #define SUITBLOOD_ARMOR 1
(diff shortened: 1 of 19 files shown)


Valtsu0:
I want dev opinion on the likelihood of this getting merged if i finish it

TobleroneSwordfish:
```
You are about to dive down a rabbit hole that will consume at least 15% of your remaining sanity. Proceed?
[YES] [NO] [HUH?]
```

Glamurio:
I once dove down this rabbit hole myself when trying to rework the "mask" slot into a "neck" slot for accessories, such as scarves and such. This really is a huge undertaking, but I think it's cool in the long-run and definitely worth having.

Valtsu0:
After looking into it a bit, shoes seem extremely time consuming and would have to be made into a component. on the other hand, hats would be quite easy to do as the only thing they have that the parent doesn't is seals_hair.

I guess consolidating all the seals_hair variables from all the different clothes would still be good even if i never finish this so i should make that pr for now

PR #24657, opened 2025-09-14
Make multitool assemblies cause a shock on their tile

M-Earthfire:
[Feature][Game Object]

## About the PR 

When a multitool assembly is activated, it creates a shock on their corresponding tile, equal to the effect of using it's special

## Why's this needed? 

Multitools right now only work like igniters when used in plasma tank assemblies. That was only added to further upgrade them to canbombs.

This changes gives a small but funny reason to make multitool assemblies

## Testing 

https://github.com/user-attachments/assets/c1a3a0ab-6979-4f6b-b47e-854dbc9e2fae

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Multitools in assemblies will now shock the tile they are on when activated, akin to using their special attack.
```

### DIFF
--- a/code/obj/item/device/multitool.dm
+++ b/code/obj/item/device/multitool.dm
@@ -83,7 +83,9 @@ TYPEINFO(/obj/item/device/multitool)
 
 /obj/item/device/multitool/proc/assembly_application(var/manipulated_multitool, var/obj/item/assembly/parent_assembly, var/obj/assembly_target)
 	if(!assembly_target)
-		//if there is no target, we don't do anything
+		//if there is no target, we make a shock akin to using the multitools special on a 1-second cooldown
+		if(!ON_COOLDOWN(global, "multitool shock", 1 SECONDS))
+			elecflash(get_turf(src),0, power=2, exclude_center = 0)
 		return
 	else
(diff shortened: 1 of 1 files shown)


code/obj/item/device/multitool.dm
```diff
@@ -83,7 +83,9 @@ TYPEINFO(/obj/item/device/multitool)
 
 /obj/item/device/multitool/proc/assembly_application(var/manipulated_multitool, var/obj/item/assembly/parent_assembly, var/obj/assembly_target)
 	if(!assembly_target)
-		//if there is no target, we don't do anything
+		//if there is no target, we make a shock akin to using the multitools special on a 1-second cooldown
+		if(!ON_COOLDOWN(global, "multitool shock", 1 SECONDS))
```
TobleroneSwordfish: Why global?

M-Earthfire: because im stupid, that's why... gonna change that real quick and maybe next time test multiple assemblies at the same time...

Edit: changed to src, like it should have been...

TobleroneSwordfish (approved):

PR #24661, opened 2025-09-14
Introduces geiger counter assemblies

M-Earthfire:
[Feature][Game Objects]

## About the PR 

This PR introduces Geiger Counter assemblies.
geiger_grenade_assembly

For this, i renamed the `change_icon_state` proc on geiger counter on `on_geiger_tick` since this now does more than just change the counters icon.

These assemblies uses geiger counters to trigger whenever the person the geiger counter is on is affected by radiation.

## Why's this needed? 

It expands the items that can be used on geiger counters. Even more so, it enables the detection of radiation for mechcomp by e.g. strapping a geiger counter/signaler assembly on a monkey.

Or maybe a clown just uses it to trigger a health alert whenever they get radiated. I feel people will figure some fun use cases out, even if this is a rather niche use. Like geiger counter, these work even if the assembly is in a container like e.g. a satchel.

## Testing 

https://github.com/user-attachments/assets/096aa093-a9cd-4706-a899-86357b9817e8

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Geiger counter can now be added as triggers into assemblies. They will trigger whenever the wearer is affected by radiation. Turn a monkey into a radiation-detector for mechcomp today!
```


M-Earthfire:
Oh in the testing video i forgot that i disabled sound for the video.... well, the pipebomb surely shows the corresponding effect.

PR #24668, opened 2025-09-16
Click-drag to dump storage-item contents into chutes

Egregorious:
## About the PR 

Changes the command to empty out the contents of storage items directly into disposals from intent-based click to click-drag. 

## Why's this needed? 

Dumping out content is a more deliberate action than placing the container within a chute, and so it's better done with a more deliberate command.

## Testing 

https://github.com/user-attachments/assets/89a51cf7-901e-4c2e-9db2-747858ec425f

## Changelog 

```changelog
(u)Egregorious
(+)Dumping the content of storage items into a chute is now performed by click-dragging. 
```


NightmareChamillian:
oh hey this is my code

ill be honest (and also kinda biased) I feel like an addition to the interaction rather than replacing it might be better here. It's such an easy shortcut to click a disposal chute once to empty, and twice to empty and also put the item in, rather than **always** having to click-drag. Esp since 99% of interactions with a container and a disposal chute are done with the prior deliberate intent of walking up to said chute.

Egregorious:
It's entirely a preference thing for sure, but to argue the other side this is currently a weird exception to every other action involving dumping items out of a container. Item containers, pill bottles and satchels all use click-drag to dump contents into tables; and satchels use click-drag to dump contents into chutes. It's also a little unintuitive that the default action to interact with chutes changes based on whether it's a container or not.

It could be additional, but I'd want to change help intent to place-container so that it's the default for the sake of parity, and at that point I think it might just be confusing that intents change the way this interaction works specifically, since there is another option available. Especially since from what I gather intent-based item interaction is preferrably avoided.

PR #24669, opened 2025-09-16
Click-drag to dump satchel contents into chutes

Egregorious:
## About the PR 

Placing satchels into disposals is no longer done via a menu. Place satchels into disposals with click, empty satchels into disposals with click-drag. 

## Why's this needed? 

The pop-up menu was very cumbersome for the purposes of only two really useful options. This is better parity with the way other items interact with disposals. 

## Testing 

https://github.com/user-attachments/assets/ab002715-85ed-453d-b629-becabe8147d7

## Changelog 

```changelog
(u)Egregorious
(+)Dumping the contents of satchels into a chute is now done with a click-drag command. 
```


code/obj/item/satchel.dm
```diff
@@ -190,8 +194,33 @@
...
+			disposals_dump(over_object, usr)
+			return
 		. = ..()
 
+	/// The user tries to place all of the satchel's contents into the given disposal chute.
+	proc/disposals_dump(var/obj/machinery/disposal/chute, mob/user)
+		if (BOUNDS_DIST(chute, user) > 0)
+			boutput(user, SPAN_ALERT("You need to be closer to [chute] to do that."))
+			return
+		if (!src.contents.len)
```
ZeWaka: use length()

code/obj/item/satchel.dm
```diff
@@ -190,8 +194,33 @@
...
+			return
+		for(var/obj/item/item in src.contents)
+			if (chute.fits_in(item))
+				item.set_loc(chute)
+		user.set_dir(get_dir(user, chute))
+		src.UpdateIcon()
+		src.tooltip_rebuild = TRUE
+		user.visible_message("<b>[user.name]</b> dumps out [src] into [chute].")
+		chute.update()
+		return
```
ZeWaka:
```suggestion
```
needless

code/obj/item/satchel.dm
```diff
@@ -190,8 +194,33 @@
...
+		if (!user.is_in_hands(src))
+			boutput(user, SPAN_ALERT("You need to be holding [src] to do that."))
+			return
+		for(var/obj/item/item in src.contents)
+			if (chute.fits_in(item))
+				item.set_loc(chute)
+		user.set_dir(get_dir(user, chute))
+		src.UpdateIcon()
+		src.tooltip_rebuild = TRUE
+		user.visible_message("<b>[user.name]</b> dumps out [src] into [chute].")
```
ZeWaka, line 220: a sound effect would be nice

Egregorious: I can maybe try to find something that fits, but I feel like that might want to be its own PR. It'd be good to get added to every instance of something being dumped, placed, or thrown into a chute. 

Also I feel like ideally dumping would be similar to filling the compost bin/still, in that there's a pause and a sound effect between each item added.

PR #24672, opened 2025-09-16
Character Preferences: Preferred uplink type

JORJ949:
## About the PR 
Adds a "Preferred uplink" option in character setup that can be selected between PDA, Radio and Standalone.
Traitor and Headrev uplinks will first try to spawn the preferred option, then fall back to PDA > radio > standalone.
Spy thief uplinks are excluded because they are forced to always be a PDA.

## Why's this needed? 
Alternative uplink options aren't presently available unless you take the pilot trait to spawn without a pda on purpose so your uplink spawns as your headset, this allows you to select where your uplink might be and reduce the meta-knowledge impact that PDA = Uplink.

## Testing 
image
image
Preferred uplink spawns when the option is available.
Uplinks correctly spawn in the order expected when the preferred option is not available.

## Changelog 

```changelog
(u)JORJ949
(*)Added a "Preferred Uplink" option under the general character setup tab. The uplink received when rolling Traitor or Head Revolutionary will try to reflect this option.
```


code/datums/preferences/preferences.dm
```diff
@@ -543,6 +545,16 @@ var/list/removed_jobs = list(
...
 				return TRUE
 
+			if ("update-uplink")
+				if (isnull(src.preferred_uplink) || src.preferred_uplink == "Standalone")
+					src.preferred_uplink = "PDA"
+				else if (src.preferred_uplink == "PDA")
+					src.preferred_uplink = "Radio"
+				else
+					src.preferred_uplink = "Standalone"
+				src.profile_modified = TRUE
```
Tarmunora: Worth `#define`ing these string literals I think

Tarmunora (changes requested):

ZeWaka (approved):

PR #24678, opened 2025-09-17
Sets dna_mutagen_banned to false in the ambassador exclusive mutantraces

snugglycactus:
## About the PR 
This PR allows the blob, martian, and amphibian mutantraces to be compatible with stable mutagen. This primarily affects cloning, but also impacts changeling gameplay via the DNA sting, and general stable mutagen usage.

## Why's this needed? 
_Hedonism._ Nah, just kidding. The actual reason pertains primarily to ambassadors and the changelings that prey on them. Presently, if a person playing as an ambassador dies and is cloned, they return to whatever mutantrace trait they have set in their character setup. I've witnessed this leading to significant confusion, and see no reason why it shouldn't be eliminated. Additionally, and ostensibly the more important reason, this change allows the DNA sting to work properly with ambassadors; ate a martian and want to turn someone else into said martian? This would allow that to happen.

## Testing 

Screenshot 2025-09-16 232134
Screenshot 2025-09-16 232643
Screenshot 2025-09-16 234403

Cloning tested, all seems to work properly. (Awful jokes are not necessary to the PR functioning properly, and exist purely for my own entertainment.)

Screenshot 2025-09-16 233502
Screenshot 2025-09-16 233657
Screenshot 2025-09-16 234127

Stable mutagen tested, also all seems to work properly. Works on things it should, doesn't work on things it shouldn't.

### DIFF
--- a/code/datums/mutantraces.dm
+++ b/code/datums/mutantraces.dm
@@ -714,6 +714,7 @@ TYPEINFO(/datum/mutantrace/blob)
 	voice_override = "bloop"
 	firevuln = 1.5
 	typevulns = list("cut" = 1.25, "stab" = 0.5, "blunt" = 0.75)
+	dna_mutagen_banned = FALSE
 
 	ghost_icon_state = "ghost-blob"
 
@@ -1733,6 +1734,7 @@ TYPEINFO(/datum/mutantrace/seamonkey)
 	human_compatible = 0
 	uses_human_clothes = 0
 	override_language = LANGUAGE_MARTIAN
+	dna_mutagen_banned = FALSE
 	mutant_organs = list(\
 		"left_eye"=/obj/item/organ/eye/beady,\
 		"right_eye"=/obj/item/organ/eye/beady,\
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish (approved):
Approved, although I agree with Moriarty that this var should default to TRUE anyway

snugglycactus:
> Approved, although I agree with Moriarty that this var should default to TRUE anyway

I agree, but even though that'd be a relatively small change, it would still be a more impactful PR than I felt like making. ...and also what if something breaks

JORJ949 (approved):
What Moriarty and Leah and You said, but this also works.

Tarmunora (changes requested):
Mixed feelings on this one - that flag was intended to restrict stable mutagen/etc to primarily work on mutantraces that are, well, mutations, as opposed to being Real Aliens.
Either way though, Shelterfrogs should remain ineligible.

TobleroneSwordfish:
Good point, but then cloning should use a different flag since cloned frogs should not become humans again.

snugglycactus:
> Mixed feelings on this one - that flag was intended to restrict stable mutagen/etc to primarily work on mutantraces that are, well, mutations, as opposed to being Real Aliens. Either way though, Shelterfrogs should remain ineligible.

Taken under consideration, and I agree, so we're trying something else entirely.

PR #24682, opened 2025-09-17
Reduce spatial interdictor component costs

Kubius:
[BALANCE]
## About the PR 

Reduces the manufacturing cost of all components specific to the spatial interdictor.

- Interdictor frame kit requires 5 dense metal, down from 10.
- All interdictor mainboards require 2 conductive material, down from 4 (secondary materials unaltered).
- Lambda phase-control rod (basic standard rod): Conductive material 10 > 5, crystal 5 > 2.
- Sigma phase-control rod (enhanced standard rod): High-conductive 10 > 5, insulator 5 > 2.
- Epsilon phase-control rod (high range, low efficiency): Electrum 10 > 5, dense material 5 > 2.
- Phi phase-control rod (low range, high efficiency): Crystal 10 > 2.

## Why's this needed? 

One common complaint about interdictors is that they have a fairly heavy material cost on top of the elaborate assembly process; cutting the manufacturing cost of the bespoke components lets those who invest the time to gather the necessary materials get a little more return for their effort. This doesn't change the other material costs (cell, wire, plating, etc), so it's proportionally not *quite* as large of a cut as it seems, but it should still make a notable difference.

## Testing 

Skipped due to this being a numbers-only change. If a screenshot of the changes ingame is still desirable, let me know.

## Changelog 

```changelog
(u)Kubius
(+)Interdictor component material costs have received across-the-board reductions.
```

### DIFF
--- a/code/datums/manufacturing.dm
+++ b/code/datums/manufacturing.dm
@@ -3479,7 +3479,7 @@ ABSTRACT_TYPE(/datum/manufacture/pod/weapon)
 
 /datum/manufacture/interdictor_kit
 	name = "Interdictor Frame Kit"
-	item_requirements = list("metal_dense" = 10)
+	item_requirements = list("metal_dense" = 5)
 	item_outputs = list(/obj/item/interdictor_kit)
 	create = 1
 	time = 10 SECONDS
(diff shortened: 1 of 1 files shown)


trazodont:
YEAH !!!

TobleroneSwordfish (approved):
Yea! Although I do think the completely opaque naming scheme may still be the main thing putting people off interdictor construction.

Kubius:
The ingame interdictor manual runs down what each one does; having a more literal name like "amplified-range interdictor rod" might communicate you a bit more information "on the box", but might not encourage manual-reading as much (and thus you could miss out on, say, the Epsilon rod drawing way extra power).

PR #24683, opened 2025-09-17
Added the flinger artifact.

Kanthes:
[A-SCIENCE]

## About the PR 
This PR adds the Flinger large artifact.

The Flinger artifact, once activated and touched, will fling itself in a direction.
* 25% of the time it flings itself in the same direction every time.
* 75% of the time it flings itself in a random direction every time.
* * If it can fling itself in random directions every time, there's an additional chance (and further additional chances) it'll fling itself repeatedly in a row.

The Flinger will fling itself between 1-15 tiles and with a force of 15-50, with the exception of Eldritch origin variants which will double those values. If the force value gets high enough, it'll trigger the THROW_THROUGH_WALLS throw type as well, explosively dismantling walls in its path.
Wizard origin variants will phase through walls when flinging.
Precursor origin variants will instead throw whoever touches the artifact.

On hit, the flinger will deal brute damage and break things based on its throwforce. 

## Why's this needed? 

The Flinger will most of the time be a silly object that'll cause some fun surprises (the wizard artifact suddenly flings itself out of ArtSci and into the corridor next door, hitting the janitor who was just doing his job), but will occasionally be a useful tool for breaching or a terrorizing force for the station to get rid of.

## Testing 
https://github.com/user-attachments/assets/42ce1476-2b3d-4be2-bf29-d40fe11247d1

## Changelog 
```changelog
(u)Kanthes
(*)Added the Flinger large artifact.
```


code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,159 @@
...
+		if(artitype.name == "wizard")
+			//Silly phasing wizards
+			throw_type = THROW_PHASE
+
+		if(prob(75))
+			dir = pick(EAST, WEST, NORTH, SOUTH)
+			random_dir_every_time = 0
+		else
+			random_dir_every_time = 1
+			if(prob(30))
```
Egregorious: This should be a loop

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,159 @@
...
+			throwforce = throwforce*2
+			//And if they hurt bad enough.. They just cruise through walls.
+			if(throwforce >= 90)
+				throw_type = THROW_THROUGH_WALL
+		if(artitype.name == "wizard")
+			//Silly phasing wizards
+			throw_type = THROW_PHASE
+
+		if(prob(75))
+			dir = pick(EAST, WEST, NORTH, SOUTH)
```
Egregorious:
```suggestion
			dir = pick(cardinal)
```

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,159 @@
+/obj/artifact/flinger
+	name = "artifact flinger"
+	associated_datum = /datum/artifact/flinger
+
+	throw_impact(atom/A, datum/thrown_thing/thr)
+		if(iscarbon(A))
+			var/mob/living/carbon/C = A
```
Egregorious: You have a lot of one-letter variables

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,159 @@
...
+	/datum/artifact_trigger/radiation,/datum/artifact_trigger/carbon_touch, /datum/artifact_trigger/language)
+	fault_blacklist = list(ITEM_ONLY_FAULTS)
+	activated = 0
+	activ_text = "begins to build tension."
+	deact_text = "relaxes."
+	react_xray = list(11,70,90,9,"FIBROUS")
+	var/range = 1
+	var/throwforce = 1
+	var/recharge_time = 600
+	var/recharging = 0
```
Egregorious: Best to use TRUE and FALSE for boolean-like values

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,159 @@
+/obj/artifact/flinger
+	name = "artifact flinger"
+	associated_datum = /datum/artifact/flinger
+
+	throw_impact(atom/A, datum/thrown_thing/thr)
```
Egregorious, line 8: I'm fairly sure you want to be calling the parent during this

Kanthes: I'm afraid I don't know what you mean. Could you elaborate?

code/obj/artifacts/artifact_objects/flinger.dm
Kanthes: Apologies! This was cobbled together from other bits of code found all over the place, so it seemed to be the style.

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,153 @@
...
+
+	post_setup()
+		. = ..()
+		recharge_time = rand(1,10) * 10
+		range = rand(1, 15)
+		throwforce = rand(1, 50)
+
+		if(artitype.name == "eldritch")
+			//Eldritch artifacts hurt more.
+			range = range*2
```
JORJ949:
```suggestion
			src.range = src.range*2
```
src. should be used before all variables that are on this object when used in procs (e.g. not just here) for code cleanliness. src. means "this thing's" so "this thing's range"

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,153 @@
...
+	fault_blacklist = list(ITEM_ONLY_FAULTS)
+	activated = 0
+	activ_text = "begins to build tension."
+	deact_text = "relaxes."
+	react_xray = list(11,70,90,9,"FIBROUS")
+	var/range = 1
+	var/throwforce = 1
+	var/recharge_time = 600
+	var/recharging = FALSE
+	var/dir = SOUTH
```
JORJ949: Name it something like "fling_dir". "dir" is a very specific thing meaning "which direction is this thing facing" and while datums don't have directions this could be misleading.

code/obj/artifacts/artifact_objects/flinger.dm
Egregorious: Erring on the side of overexplanation; you are overriding the proc `throw_impact`, which means you are replacing its behaviour on the parent classes `obj` and `artifact`. In this case the parent's `throw_impact` handles useful things like material triggers, reagent shocks, and contextual impact sounds. 

So I think you want to be calling the parent with `..(A, thr)` whenever your artifact is hitting something you want to successfully impact, and the parent will handle all that stuff for you.

code/obj/artifacts/artifact_objects/flinger.dm
Egregorious: I have tended to do the same thing before, but the codebase has been around for a long time and it's had very different standards throughout it's life. So best to try to use the current guidelines than rely on the code itself to be a good exemplar.

RSOD404:
i like the idea, seems very fun. a small suggestion though, perhaps the artifact could briefly knock down / stun any victims hit if it has enough force? from what it looks like, getting hit by one could look fairly unsatisfying if the victim is still standing after a 70 force eldritch art just tanked their health flying into them

Glamurio:
> i like the idea, seems very fun. a small suggestion though, perhaps the artifact could briefly knock down / stun any victims hit if it has enough force? from what it looks like, getting hit by one could look fairly unsatisfying if the victim is still standing after a 70 force eldritch art just tanked their health flying into them

Big agree, I'd like it if it stunned people, and ideally it should also straight up gib someone if it exceeds a really high threshold.

Kanthes:
This is now the result of a throwforce 49 non-eldritch artifact. EDIT: Turns out I was double dipping on damage and stunning.
https://github.com/user-attachments/assets/dbfd8fbf-aa17-4b1a-9c76-a4405379ab71

code/obj/artifacts/artifact_objects/flinger.dm
Kanthes: Thank you for the clear explanation!

Valtsu0:
Note that having a throw force over 80 qdels most things it hits, even stuff that is normally indestructible like the listening post vendor.

This is probably fine since max loafs do that too and are a bit easier to get. Also most things that REALLY should not be destroyed are immune to it like the cryo unit and it doesn't happen in restrictedz

JORJ949:
Instead of qdeling anything hit try calling an ex_act(1) on the target, this is the proc called on something when its caught in an explosive blast (1 being the highest power of explosive radius, 3 being the lowest) which for most things will destroy them, other special objects being explosion immunity or resistance will be unaffected

JORJ949:
> Instead of qdeling anything hit try calling an ex_act(1) on the target, this is tge proc called on something when its caught in an explosive blast (1 being the highest power of explosive radius, 3 being the lowest) which for most things will destroy them, other special objects being explosion immunity or resistance will be unaffected

Kanthes:
> > Instead of qdeling anything hit try calling an ex_act(1) on the target, this is tge proc called on something when its caught in an explosive blast (1 being the highest power of explosive radius, 3 being the lowest) which for most things will destroy them, other special objects being explosion immunity or resistance will be unaffected

since the latest commit the artifact no longer directly deletes or destroys anything (aside from gibbing the player), instead letting throw_impact do its thing.

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,118 @@
...
+					break
+		if(prob(5))
+			src.recharge_time = 0
+
+	effect_touch(var/obj/O,var/mob/living/user)
+		if (..())
+			return
+		if (!user)
+			return
+		if (src.recharging)
```
JORJ949: You can use ON_COOLDOWN() here, instructions on how to use it can be found in _std\macros\cooldown.dm or by just hovering over ON_COOLDOWN()

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,118 @@
...
+	var/fling_dir = SOUTH
+	var/random_dir_every_time = FALSE
+	var/chain_flings = 1
+	var/throw_type = 1
+
+	post_setup()
+		. = ..()
+		src.recharge_time = rand(1,10) * 10
+		src.range = rand(1, 15)
+		src.throwforce = rand(1, 50)
```
TobleroneSwordfish:
```suggestion
		src.throwforce = rand(15, 50)
```
Having most of them do very little damage is boooriiing

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,118 @@
...
+	associated_datum = /datum/artifact/flinger
+
+	throw_impact(atom/A, datum/thrown_thing/thr)
+		. = ..(A, thr)
+		if(iscarbon(A))
+			var/mob/living/carbon/unit = A
+			if(thr.bonus_throwforce >= 20)
+				unit.emote("scream")
+			if(thr.bonus_throwforce >= 90)
+				unit.gib()
```
TobleroneSwordfish: I'm not sure adding the possibility of being instantly gibbed for activating an artifact is a good idea, even cyborg converters allow the possibility of being saved or just playing as a borg. Maybe instead throw the *victim* with `THROW_THROUGH_WALL`?

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,118 @@
+/obj/artifact/flinger
+	name = "artifact flinger"
+	associated_datum = /datum/artifact/flinger
+
+	throw_impact(atom/A, datum/thrown_thing/thr)
+		. = ..(A, thr)
+		if(iscarbon(A))
+			var/mob/living/carbon/unit = A
```
TobleroneSwordfish: Please use sensible variable names, `unit` doesn't mean anything.

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,118 @@
...
+			if(thr.bonus_throwforce >= 20)
+				unit.emote("scream")
+			if(thr.bonus_throwforce >= 90)
+				unit.gib()
+				logTheThing(LOG_COMBAT, thr.thing, " gibbed [unit.name] at [log_loc(unit)].")
+		else if(iswall(A))
+			if (thr.bonus_throwforce >= 35)
+				playsound(thr.thing.loc, 'sound/impact_sounds/Generic_Hit_Heavy_1.ogg', 50, 1)
+				for (var/mob/N in AIviewers(thr.thing, null))
```
TobleroneSwordfish: Same here, use `M` for mob if you want to use a single letter variable name.

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,118 @@
...
+					if (N.client)
+						shake_camera(N, 4, 8, 0.5)
+		else if(istype(A, /obj/structure/girder))
+			if (thr.bonus_throwforce >= 35)
+				playsound(thr.thing.loc, 'sound/impact_sounds/Generic_Hit_Heavy_1.ogg', 50, 1)
+				for (var/mob/N in AIviewers(thr.thing, null))
```
TobleroneSwordfish: Same here, use `M` for mob if you want to use a single letter variable name.

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,118 @@
...
+		if(artitype.name == "wizard")
+			//Silly phasing wizards
+			src.throw_type = THROW_PHASE
+
+		if(prob(75))
+			src.fling_dir = pick(cardinal)
+			src.random_dir_every_time = FALSE
+		else
+			src.random_dir_every_time = TRUE
+			for (var/i = 0, i < 5, i += 1)
```
TobleroneSwordfish, line 75: Honestly I would make chain flings guaranteed and push the cap for them up to 10-15ish. It's a little underwhelming when it only moves once.

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,118 @@
...
+			src.range = src.range*2
+			src.throwforce = src.throwforce*2
+			//And if they hurt bad enough.. They just cruise through walls.
+			if(src.throwforce >= 90)
+				src.throw_type = THROW_THROUGH_WALL
+		if(artitype.name == "wizard")
+			//Silly phasing wizards
+			src.throw_type = THROW_PHASE
+
+		if(prob(75))
```
TobleroneSwordfish:
```suggestion
		if(prob(25))
```
Feels like random makes more sense to be the norm?

code/obj/artifacts/artifact_objects/flinger.dm
Kanthes: Gibbing has been removed, and I've added a precursor variant that will instead fling whoever used the artifact

code/obj/artifacts/artifact_objects/flinger.dm
Kanthes: I've raised the odds of chain flings, making it more likely. I still kind of want 1 fling to be the norm. Though if you insist I'll add some kind of minimum and rework the logic.

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -50,7 +50,7 @@
 
 	post_setup()
 		. = ..()
-		src.recharge_time = rand(1,10) * 10
+		src.recharge_time = rand(1,10) //in seconds
```
Tarmunora:
```suggestion
		src.recharge_time = rand(1,10) SECONDS
```

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -85,7 +85,7 @@
 			return
 		if (!user)
 			return
-		if(!ON_COOLDOWN(src, "artifact_fling", src.recharge_time + src.chain_flings * 20))
+		if(!ON_COOLDOWN(src, "artifact_fling", src.recharge_time + src.chain_flings * 0.2 SECONDS))
```
Tarmunora:
```suggestion
		if(!ON_COOLDOWN(src, "artifact_fling", src.recharge_time + src.chain_flings * 2 SECONDS))
```

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -95,14 +95,14 @@
 					if(THROW_TARGET_USER)
 						fling(O, user)
 				for (var/i = 0, i < remaining_flings, i += 1)
-					SPAWN(10*i)
+					SPAWN(0.1*i SECONDS)
```
Tarmunora:
```suggestion
					SPAWN(i SECONDS)
```

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -95,14 +95,14 @@
...
 						switch(src.throw_target)
 							if(THROW_TARGET_SELF)
 								fling(O, O)
 							if(THROW_TARGET_USER)
 								fling(O, user)
 				O.ArtifactFaultUsed(user)
-				SPAWN(src.recharge_time + src.chain_flings * 20) //Prevents the artifact from being activated while being thrown around.
+				SPAWN(src.recharge_time + src.chain_flings * 0.2 SECONDS) //Prevents the artifact from being activated while being thrown around.
```
Tarmunora:
```suggestion
				SPAWN(src.recharge_time + src.chain_flings * 2 SECONDS) //Prevents the artifact from being activated while being thrown around.
```

Kanthes: Sorry about that, these issues were resolved in a commit shortly after the one you're commenting on!

code/obj/artifacts/artifact_objects/flinger.dm
```diff
@@ -0,0 +1,128 @@
...
+						fling(O, user)
+				for (var/i = 0, i < remaining_flings, i += 1)
+					SPAWN((0.1*i) SECONDS)
+						switch(src.throw_target)
+							if(THROW_TARGET_SELF)
+								fling(O, O)
+							if(THROW_TARGET_USER)
+								fling(O, user)
+				O.ArtifactFaultUsed(user)
+				SPAWN(src.recharge_time) //Prevents the artifact from being activated while being thrown around.
```
JORJ949: Comment should be on the cooldown, this just gives you a message when the cooldown is done, this line isn't what stops mid-throw activation

PR #24684, opened 2025-09-17
Refactors The Pod UI To Use HUD Zones

Mister-Moriarty:
## About The PR:
The pod UI is now constructed using HUD zones, with four zones being used: the main panel, the healthbars, the leave button and master caution, and the tracking element.

Moves the `clicked()`, `update_state()`, and `update_system()` effects of each button into the screen object AM rather than them all being defined on the pod HUD datum.

Also fixes a few bugs with HUD zones.

## Why Is This Needed?
HUD zones offer a cleaner way to implement and manage UIs, but despite being over a year old, are currently unused. The pod UI itself is quite old, and deserves an update; this also makes it easier to resprite the UI, as icon handling is now implemented on individual screen objects rather than grouped together on the HUD datum.

## Testing:


ZeWaka:
How does this work with different icon sizes? Does this fix pod UI only working really with stretch to fit?

Mister-Moriarty:
> How does this work with different icon sizes? Does this fix pod UI only working really with stretch to fit?

The behaviour under HUD Zones is currently the same as the original behaviour, in that the icons sit outside of the screen boundaries.

code/datums/hud/pod/pod_screen_objs/_pod_screen_obj_parent.dm
```diff
@@ -0,0 +1,88 @@
...
+	if (!istype(src.pod_hud) || !istype(user))
+		return
+
+	if (user.loc != src.pod_hud.master)
+		boutput(user, SPAN_ALERT("You're not in the pod doofus. (Call 1-800-CODER.)"))
+		src.pod_hud.remove_client(user.client)
+		return
+
+	if (is_incapacitated(user))
+		boutput(user, SPAN_ALERT("Not when you are incapacitated."))
```
frawhst:
```suggestion
	if (!can_act(user))
		boutput(user, SPAN_ALERT("Not when you are incapacitated or restrained."))
```
Existing issue, but this should definitely include cuffs in the check

code/modules/interface/hudzones/hud_zones.dm
```diff
@@ -196,14 +223,19 @@
  */
 /datum/hud_zone/proc/ensure_empty(datum/hud_element/new_elem)
 	PRIVATE_PROC(TRUE)
-	// If adding the element exceeds the length of the zone, try to wraparound
+	// If adding the element exceeds the length of the zone, try to wraparound.
 	if ((src.horizontal_offset + new_elem.width) > (HUD_ZONE_LENGTH(src.coords)))
-		// If adding 1 more element exceeds the height of the zone, its full up
-		if ((src.vertical_offset + 1) > (HUD_ZONE_HEIGHT(src.coords)))
+		// If adding one more element exceeds the height of the zone, it's full.
```
frawhst, line 224: the explanatory comments are greatly appreciated! (why is this one thing 0-indexed. goodness)

code/modules/interface/hudzones/hud_zones.dm
```diff
@@ -1,3 +1,7 @@
+/// Returns a stringified number with its associated +/- sign.
+#define SIGNED(NUM) (NUM >= 0 ? "+[NUM]" : "[NUM]")
```
frawhst: I would put this in `_std/hud.dm`, seems useful for future HUD operations (and that file is all hud zone specific defines anyways). Maybe rename to be more specific, but the existing name shouldn't conflict with anything that I can see.

ZeWaka:
X

PR #24685, opened 2025-09-17
AI Internal Cell rebalance

JORJ949:
## About the PR 
Rebalances the AI Core's internal cell depletion rate by quadrupling the rate of decay and making it relative to life tick (meaning you won't get more time if the lifeprocess is running slower)

## Why's this needed? 
The AI core's internal cell currently lasts for 100 minutes according to the math I did. 100 minutes is an entire round and so this never really has any sort of application. Reducing the time the AI can live without power makes the feature impactful while also giving 25 minutes for them to find power to live off of.
image

## Testing 
After using 5 minutes of internal cell, I had lost 20% charge, correctly scaling to the intended 25 minutes

## Changelog 

```changelog
(u)JORJ949
(*)AI Cores can now survive on their internal cell for only 25 minutes, down from 100 minutes.
```

### DIFF
--- a/code/mob/living/silicon/ai.dm
+++ b/code/mob/living/silicon/ai.dm
@@ -1414,8 +1414,10 @@ or don't if it uses a custom topopen overlay
 	if (src.sleeping) src.sleeping = 0
 	src.delStatus("knockdown")
 
+#define AI_CHARGE_RATE 10
 /mob/living/silicon/ai/use_power()
 	..()
+	var/mult = (max(tick_spacing, TIME - last_life_tick) / tick_spacing)
 	var/turf/T = get_turf(src)
 	if (T)
 		var/area/A = T.loc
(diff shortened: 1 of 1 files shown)


kbsmilk:
A good change but I think the number is still too conservative. 3 to 5 minutes of cell time is closer to heavy inefficient cyborgs using upgrades. And opens up some gameplay moves like forcing the AI to move using sabotage.
AFAIK, any AI can grab a shell and drag the core to any wire ending (every APC has one) to charge there. Or possibly just any APC area?

Only, it could get pretty bad for scenarios where the engine isn't set up quickly or at all. Or is destroyed. Or power sinks.

And if it goes that low, there should probably be an automated mainframe alert when the AI shuts down from power loss. Or else it's just silently dead.

JORJ949:
The AI gets power from the area APC, recharges at the same speed it discharges, and will automatically turn back on when it has enough power to do so. I'd like to see how this time works out before jumping to as low as 5 minutes though

TobleroneSwordfish (approved):

code/mob/living/silicon/ai.dm
```diff
@@ -1414,8 +1414,10 @@ or don't if it uses a custom topopen overlay
 	if (src.sleeping) src.sleeping = 0
 	src.delStatus("knockdown")
 
+#define AI_CHARGE_RATE 10
 /mob/living/silicon/ai/use_power()
 	..()
+	var/mult = (max(tick_spacing, TIME - last_life_tick) / tick_spacing)
 	var/turf/T = get_turf(src)
```
Tarmunora, line 1421: Any reason not to pass `mult` as an argument from silicon/Life instead (and also mult-adjust other silicons)?

JORJ949: Life() doesn't get passed mult so I would just be moving the calculation to there, ideally Life() gets added a mult arg so you don't have to reimplement the mult calculation with each version of Life() but I opted to keep that change out of this PR to atomise this change to just the AI balance changes.

PR #24686, opened 2025-09-17
Sound for placing, dumping, and throwing items into chutes

Egregorious:
## About the PR 

Adds sfx for placing, dumping and throwing items into disposals, mailboxes, and similar chutes. There are 4 sound variants. 

Placing items directly plays the sound locally and at lower volume, to allow stealthy disposal of evidence. 

Variance depending on item size, with smaller items being quieter and higher pitch, larger items being louder with a lower pitch.

Source (Creative Commons 0): https://freesound.org/people/kyles/sounds/450779/

## Why's this needed? 

Extra feedback for chutes. 

## Testing 
Example:

https://github.com/user-attachments/assets/9cd18141-86ef-4c47-b94a-05f783da0205

Placing items sound range: 

https://github.com/user-attachments/assets/e96cc77b-d9c4-4db6-a85b-c408033caf75

Pitch variances with item size: 

https://github.com/user-attachments/assets/802782f7-604a-40f4-9310-5add99800748

## Changelog 

```changelog
(u)Egregorious
(+)Added sfx for putting things into chutes. If the action does not have an associated chat message, noone else will hear the sound except you. 
```


SunWuDrakios:
Not gonna lie that would be very satisfying to have. Personally I would like some more variation in the sounds, so you don't get crazy after a short while

RSOD404:
good, but there needs to be more than just one sound effect for it, and possibly a small amount of pitch randomizing to not make it instantly repetitive and annoying to hear

Egregorious:
I'm going to see if I can do a couple more variations. There is the standard pitch shifting already applied, and maybe the variations will satisfy the need for strengthening that. I'll also look into pitch shifting for item sizes. 

I would rather not bother with differentiating between item types further than that though, at least not in this PR. I'm sure there's equilisers that could work, but it feels like that would get very specific very quickly and I think a generic version will work to start with, personally.

Egregorious:
https://github.com/user-attachments/assets/0b29d914-9f3a-47c2-aaf2-c421667a3d4d

Here's some variances in action. I could record them in isolation, but that's not how you'd actually hear them ingame.

TobleroneSwordfish:
I kind of don't want to add this just because of how often chutes are used as sneaky hiding spots. No objection to the sound though.

Egregorious:
> I kind of don't want to add this just because of how often chutes are used as sneaky hiding spots. 

There is space to reduce the range when placing by one or two tiles more I think, or else it could be played exclusively for the actor if that's more ideal.

Glamurio:
> I kind of don't want to add this just because of how often chutes are used as sneaky hiding spots. No objection to the sound though.

I agree with this, I think maybe it should only make noise when someone throws something inside? So you can still sneakily place objects or climb inside. Basically if you stand next to it, it makes no sound.

Egregorious:
Changed the direct placement of items from a very low-range sound to a local sound with much a lower volume, so it's not played for anyone except the person interacting with the chute and it sounds as if it doesn't carry as much. Throwing and dumping are unaffected, since they have broadcasted chat messages associated and are thus not sneaky actions.

flappybatpal:
Seems good to me.

code/modules/disposals/disposal_chute.dm
```diff
@@ -632,6 +638,37 @@ ADMIN_INTERACT_PROCS(/obj/machinery/disposal, proc/flush, proc/eject)
...
+				pitch = 0.3
+				volume = 300
+			if (W_CLASS_BUBSIAN)
+				pitch = 0.1
+				volume = 400
+		if (user)
+			volume = min(volume * 0.5, 20)
+			user.playsound_local(src.loc, pick(item_insert_sounds), volume, 1, 0, pitch)
+		else
+			playsound(src, pick(item_insert_sounds), volume, 1, 0, pitch)
```
Tarmunora: Add to `/proc/generate_sound` instead of keeping a list of sounds on the chute perhaps

code/modules/disposals/disposal_chute.dm
```diff
@@ -206,6 +208,7 @@ ADMIN_INTERACT_PROCS(/obj/machinery/disposal, proc/flush, proc/eject)
 				S.UpdateIcon()
 				S.tooltip_rebuild = TRUE
 				user.visible_message("<b>[user.name]</b> dumps out [S] into [src].")
+				play_item_insert_sound(I)
```
Tarmunora:
```suggestion
				src.play_item_insert_sound(I)
```
very much a nitpick, but `src.MyFunc()` is a bit nicer for maintainability

Tarmunora (changes requested):

code/modules/sound/sound.dm
```diff
@@ -750,6 +751,7 @@ proc/narrator_mode_sound_file(sound_file)
 /var/global/list/sounds_warp = list(sound('sound/effects/warp1.ogg'),sound('sound/effects/warp2.ogg'))
 /var/global/list/sounds_engine = list(sound('sound/machines/tractor_running2.ogg'),sound('sound/machines/tractor_running3.ogg'))
 /var/global/list/sounds_keyboard = list(sound('sound/machines/keyboard1.ogg'),sound('sound/machines/keyboard2.ogg'),sound('sound/machines/keyboard3.ogg'))
+/var/global/list/sounds_chute_insert = list(sound('sound/effects/chute_place_1.ogg'),sound('sound/effects/chute_place_2.ogg'),sound('sound/effects/chute_place_4.ogg'))
```
Tarmunora: missing chute_place_3?

Egregorious: I was considering removing it since it sounds a bit samey to chute_place_2, but decided against it and then forgot to readd after testing.

PR #24691, opened 2025-09-18
Changes the shelterfrog accent to be less chav

Calphrick:
[MUTANTRACES] [REWORK] [MINOR]
## About the PR 
This adds a new distinct "frog" accent that keeps repeated "r" of the original shelterfrog accent while replacing the chav slang with frog-related exclamations, lovingly provided by SnugglyCactus.
## Why's this needed? 
The choice to make shelterfrogs talk like lower-class Brits is something I have seen lamented by several frog players. This aims to change it so that they make more appropriate noises.
## Testing 

Screenshot 2025-09-18 183906

## Changelog 

```changelog
(u)Calphrick
(*)Reworked the shelterfrog accent.
```


Calphrick:
Thought there was a better way to do what I wanted, so I closed it with the intention of doing it differently. I reviewed the code, realized this probably was the easiest way to do it, and so I have reopened the PR. Whoops!

code/modules/medical/genetics/bioEffects/speech.dm
```diff
@@ -73,6 +73,25 @@
 	lockedChars = list("G","C")
 	lockedTries = 3
 
+/datum/bioEffect/speech/frog
+	name = "Frontal Gyrus Alteration Type-Fr"
+	desc = "Forces the language center of the subject's brain to construct sentences in in the matter of Abzuians."
```
Piesuu: Pretty sure it's not meant to say `in in the matter of Abzuians`, Here's an alternate suggestion I came up with because I couldn't figure out how to put abzu in a sentence properly.
```suggestion
	desc = "Forces the language center of the subject's brain to construct sentences in a truly ribbeting manner."
```
Could also go with "construct sentences in a way that makes them croak together." or something like that, Who am I to judge!

Glamurio (approved):

code/modules/medical/genetics/bioEffects/speech.dm
```diff
@@ -73,6 +73,25 @@
 	lockedChars = list("G","C")
 	lockedTries = 3
 
+/datum/bioEffect/speech/frog
+	name = "Frontal Gyrus Alteration Type-Fr"
```
TobleroneSwordfish, line 77: As a brit, I approve of the double meaning here :P

TobleroneSwordfish (approved):

code/modules/medical/genetics/bioEffects/speech.dm
Calphrick: I am not British, I do not understand what you mean. "Fr" is just the first two letters of "frog."

TobleroneSwordfish: It's also the country code for France, french = froggies in british slang.

PR #24697, opened 2025-09-19
HOTFIX: Donor Organ Names Fixed

TacoGS:
## About the PR 
During surgery it was noticed that if you try to pick up limb or organ that had the name of the mob that was within reach of you it would cause you to left click them instead of pick up the organ with the right click context menu "Pick up / Left Click". It seems that people noticed this before and it was related to the apostrophe ' which I had to change to ’ (U+2019). 

This only fixes wigs and organs and nothing else that may cause this like spawned in items at spawn time (like random items such as *Character Name*'s Award Winning *item*). 

## Why's this needed? 
This restores the ability to pick up limbs and organs named after the donor while being within reach of the mob they came from. 

## Testing 
Once the unicode apostrophe was used I loaded into the game in debug and proceeded to remove limbs from a monkey and a genetically humanized monkey. I was now able to pick up their brain, their limbs and move them without it left clicking them instead.

### DIFF
--- a/tools/ci/check_grep_secret.sh
+++ b/tools/ci/check_grep_secret.sh
@@ -58,5 +58,3 @@ if grep -P "(?<!UNLINT\().*name = .*\"\[.*\]'s" */**/*.dm;   then
 	echo "ERROR: Using an apostrophe in a name like [mob]'s brain may cause Byond to get confused between the two objects in click verbs etc. Please use ’ (U+2019) instead."
 
 exit $st
-
-exit $st
--- a/tools/ci/check_grep_secret.sh
+++ b/tools/ci/check_grep_secret.sh
@@ -54,4 +54,9 @@ if grep -P '^ABSTRACT_TYPE\([^/]' */**/*.dm;	then
     st=1
 fi;
 
+if grep -P "(?<!UNLINT\().*name = .*\"\[.*\]'s" */**/*.dm;   then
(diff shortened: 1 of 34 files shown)


TobleroneSwordfish (changes requested):
I've made a PR to your branch to add an automated check for this issue and fix the other instances of it the regex found in the codebase. Please merge that into this branch when you can.
Also changelog entries are for player-facing changes, this doesn't need one.

tools/ci/check_grep_secret.sh
```diff
@@ -54,4 +54,9 @@ if grep -P '^ABSTRACT_TYPE\([^/]' */**/*.dm;	then
     st=1
 fi;
 
+if grep -P "(?<!UNLINT\().*name = .*\"\[.*\]'s" */**/*.dm;   then
+	echo "ERROR: Using an apostrophe in a name like [mob]'s brain may cause Byond to get confused between the two objects in click verbs etc. Please use ’ (U+2019) instead."
+
+exit $st
+
 exit $st
```
ZeWaka: incorrect @TobleroneSwordfish

ZeWaka (commented):
also still changelog

tools/ci/check_grep_secret.sh
TobleroneSwordfish: oh oop

PR #24719, opened 2025-09-22
Add ability to deconstruct destroyed canisters

Glamurio:
[Game Objects] [QoL]

## About the PR 
Adds the ability to disassemble destroyed canisters.

Adds a `/obj/machinery/portable_atmospherics/proc/canister_disassemble` action that's used when targeting a destroyed canister with a welding torch.

## Why's this needed? 
Canisters get destroyed almost every other game, but there's no convenient way to just get them out of the way except for piling them onto one tile in a corner, which is weird. Crates can be disassembled because they're a nuisance, so I feel like canisters should be given the same treatment.

## Testing 
image

## Changelog 

```changelog
(u)Glamurio (Ryou)
(+)Canisters can now be disassembled with a welding tool.
```

### DIFF
--- a/_std/defines/construction.dm
+++ b/_std/defines/construction.dm
@@ -28,11 +28,6 @@
 #define RAILING_UNFASTEN 1
 #define RAILING_FASTEN 2
 
-//canister defines
-#define CANISTER_DISASSEMBLE 0
-#define CANISTER_DISCONNECT 1
-#define CANISTER_CONNECT 2
-
 //deconstruction_flags
 
 #define DECON_NONE 0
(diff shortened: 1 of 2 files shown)


pgmzeta:
agree, this should be two separate PRs

Glamurio:
> This PR is very confusing, it seems to be trying to add canister disassembly (good, often requested feature) but then also adds action bars to wrenching/unwrenching canisters by creating some kind of amalgam shared actionbar type. A) That action bar type is disgusting and seems like it could be replaced with `SETUP_GENERIC_ACTIONBAR` with no behaviour changes? B) Why are you adding actionbars to wrenching/unwrenching anyway?

Last time I added a generic actionbar I was told it should be a custom actionbar instead, I have no preference either way, I can change it. The reason the behaviors were merged is because I assumed it's more sensible to have a consistent action that combines tool usage, much like other things do. I'll clean the PR up so that it's just the deconstruction.

TobleroneSwordfish (approved):

PR #24726, opened 2025-09-23
Fix genebooth contexts overriding all clicks

### LINKED ISSUE #11919: Using the GeneBooth as an AI eye from more than 1 tile away will show an error
### Labels

[BUG][MINOR]

### Description

Using the genebooth from further than 1 tile away will display "Something went wrong, showing backup menu..."

### Steps to reproduce

1. Be an AI eye
2. Be positioned a few tiles away from the gene 
booth
3. Open the GeneBooth
4. See error in chat

### Expected Behavior

For there to not be an error

Reported by: Piesuu
Round log date: 2022-11-10-16-56
Reported on: 2022-11-10 17:23:12
Map: NADIR

JORJ949:
## About the PR 
Implements custom context handling for the Genetics Booth to only show context actions when clicked with an empty hand to allow deconstructing and other features that make use of attacking with an item, such as the feature in which dealing 50 damage to the gene booth will spit someone out. 

Forces the deconstruction context layout to always be the same regardless of target object's context layout

## Why's this needed? 
Default context behaviour will apply on any click, overriding any items you're trying to click with, which I learnt when trying to add an emag act, you can't currently click the genebooth with ANY item, meaning they presently can't be deconstructed due to deconstruction needing you to click it with a decon tool.

Due to the genebooth now having two available context menus (gene select, deconstruct) I have to force the deconstruct menu to use the same layout always or it will try to use the genebooth's 32x32 layout

Fixes #11919 

## Testing 
image
image

## Changelog 

```changelog
(u)JORJ949
(+)Fixed the Gene Booth making all your clicks be empty hands, allowing you to deconstruct it and bash it (50 hp) to force the occupant out (doesn't break the booth).
```

### DIFF
--- a/code/datums/components/tool/deconstructing.dm
+++ b/code/datums/components/tool/deconstructing.dm
@@ -8,11 +8,13 @@ TYPEINFO(/datum/component/deconstructing)
 /datum/component/deconstructing
 	var/base_duration = 2 SECONDS
 	var/complexity_mult = 1 //! How much complexity affects deconstruction time
+	var/contextLayout
 
 	Initialize(baseDuration=2 SECONDS, complexityMult=1)
 		..()
 		src.base_duration = baseDuration
 		src.complexity_mult = complexityMult
+		src.contextLayout = new/datum/contextLayout/flexdefault()
 
 	///
 	proc/finish_decon(atom/target, mob/user, atom/deconstructor)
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish (approved):

JORJ949:
Oh this also fixes fluid canister deconstructing contexts which were overridden by the fluid canister's 3 item radial context menu

PR #24734, opened 2025-09-23
Adds "persists_on_clone" var to fix nonhuman cloning weirdness

snugglycactus:
## About the PR 
This pull request adds a "persists_on_clone" var to mutantraces.dm that defaults to true, allowing most mutantraces to be cloned, with exceptions for things like thralls, zombies, and kudzupeople. Functional replacement of #24678.

## Why's this needed? 
Presently, the nonhuman diplomat mutantraces pop out of the cloner as whatever mutantrace the character is set to in character setup. This has been mentioned to be unintentional by developers, so I fixed it. This time with allowances for stable mutagen to still work!

## Testing 
Screenshot 2025-09-23 182909

Screenshot 2025-09-23 182639

For sake of testing, since I was unable to turn myself into a thrall in a test branch, I briefly set the blob mutantrace to have "persists_on_clone = FALSE" to verify mutantraces with it set to false don't get cloned as themselves.

### DIFF
--- a/code/datums/mutantraces.dm
+++ b/code/datums/mutantraces.dm
@@ -84,6 +84,8 @@ ABSTRACT_TYPE(/datum/mutantrace)
 	/// Should stable mutagen not copy from this mutant?
 	var/dna_mutagen_banned = TRUE
 	/// Should a genetics terminal be able to remove this mutantrace?
+	var/persists_on_clone = TRUE
+	/// Should this mutantrace be clonable? (see thralls as a relevant example)
 	var/genetics_removable = TRUE
 	/// Should they be able to walk on shards barefoot
 	var/can_walk_on_shards = FALSE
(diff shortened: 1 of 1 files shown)


code/datums/mutantraces.dm
```diff
@@ -714,6 +716,7 @@ TYPEINFO(/datum/mutantrace/blob)
 	voice_override = "bloop"
 	firevuln = 1.5
 	typevulns = list("cut" = 1.25, "stab" = 0.5, "blunt" = 0.75)
+	persists_on_clone = FALSE
```
JORJ949: Blob mutantrace is used for the blob diplomat, any reason this doesn't persist on clone?

code/datums/mutantraces.dm
```diff
@@ -84,6 +84,8 @@ ABSTRACT_TYPE(/datum/mutantrace)
 	/// Should stable mutagen not copy from this mutant?
 	var/dna_mutagen_banned = TRUE
 	/// Should a genetics terminal be able to remove this mutantrace?
+	var/persists_on_clone = TRUE
+	/// Should this mutantrace be clonable? (see thralls as a relevant example)
```
JORJ949: Comments go above the variable, not below, you've mixed the genetics and cloneable comments up

snugglycactus: Gah, I can fix that when I get home. Oops.

code/datums/mutantraces.dm
snugglycactus: Oh, that’s because I was testing with the blob mutantrace and forgot to turn it back. Can fix it real fast.

snugglycactus: Fixed.

PR #24742, opened 2025-09-25
Spy Thief Total Stolen/Redeemed items in round summary

JORJ949:
## About the PR 
Adds a number to show the total items stolen/redeemed in spief round summary

## Why's this needed? 
This information is already there, it just means you don't have to count.

## Testing 
image

### DIFF
--- a/code/modules/antagonists/spy_thief/spy_thief.dm
+++ b/code/modules/antagonists/spy_thief/spy_thief.dm
@@ -141,6 +141,14 @@
 
 		return list(
 			list(
+				"name" = "Total Items Stolen",
+				"value" = "[length(stolen_items)] items",
+			),
+			list(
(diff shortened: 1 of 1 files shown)


ZeWaka:
I think it'd be worth exploring using tooltips for the names, this takes up a lot of space currently

Edit: just realized this is just the numbers - why not just add the numbers to the labels

JORJ949:
> I think it'd be worth exploring using tooltips for the names, this takes up a lot of space currently

I agree, especially with longer item names from trinkets, "George Khouri's family keepsake nanotrasen beret" takes up an entire line. It would also probably be a good idea to have these sections split a bit more clearly, the stolen and redeemed item sections can be hard to read as different sections at large item counts.

PR #24746, opened 2025-09-25
Adds the Wellspring artifact.

Kanthes:
[A-SCIENCE][C-FEATURE]

## About the PR 
This PR adds the Wellspring large artifact.

The Wellspring artifact, once activated, will continuously spawn a random amount of a origin-determined liquid at its location, at a random interval. The artifact will stay active 5-10 minutes before deactivating.

Martian origin Wellsprings will create Water, Viscerite fluid, Martian flesh or Hemolymph.
Silicon origin Wellsprings will create Welding Fuel or Directed Nanites.
Eldritch origin Wellsprings will create Blood, Bloodc, Black Goop or Sewage.
Precursor origin Wellsprings will create Ephedrine or Cryo fluid.
Wizard origin Wellsprings will create Reversium, Fliptonium or Mugwort.

## Why's this needed? 

The Wellspring will create the potential for flooding on space-based stations, giving Janitors a chance to be the bastion defending the station from a wall of blood, and give occasional roleplay opportunities such as blood sacrifices in the Chapel.

## Testing 
Lowest possible roll, 100 units every 10 seconds:

https://github.com/user-attachments/assets/e5181449-958c-4e73-b4b8-1c2a8ab1c66c

Highest possible roll, 3000 units every 1 second:

https://github.com/user-attachments/assets/036ec481-723b-46b9-8dd4-15430347638c

## Changelog 

```changelog
(u)Kanthes
(*)Adds the Wellspring artifact.

### DIFF
--- a/code/obj/artifacts/artifact_machines/wellspring.dm
+++ b/code/obj/artifacts/artifact_machines/wellspring.dm
@@ -13,7 +13,7 @@
 	type_name = "Wellspring"
 	rarity_weight = 350
 	react_xray = list(5,65,20,11,"HOLLOW")
-	validtypes = list("ancient", "martian", "eldritch")
+	validtypes = list("ancient", "martian", "eldritch", "precursor", "wizard")
 	//validtypes = list("ancient", "martian", "eldritch", "lattice")
 	validtriggers = list(/datum/artifact_trigger/carbon_touch, /datum/artifact_trigger/silicon_touch, /datum/artifact_trigger/force, /datum/artifact_trigger/heat, /datum/artifact_trigger/cold, /datum/artifact_trigger/radiation, /datum/artifact_trigger/electric, /datum/artifact_trigger/language)
 	activated = 0
@@ -29,11 +29,15 @@
 		. = ..()
 		switch(artitype.name)
 			if ("ancient")
-				src.payload_reagent = pick("fuel", "charcoal", "silicate", "graphene_compound")
+				src.payload_reagent = pick("fuel", "goodnanites")
 			if ("martian")
 				src.payload_reagent = pick("water")
 			if ("eldritch")
-				src.payload_reagent = pick("blood", "sewage")
+				src.payload_reagent = pick("blood")
+			if ("precursor")
+				src.payload_reagent = pick("ephedrine")
+			if ("wizard")
+				src.payload_reagent = pick("reversium", "fliptonium")
 			//if ("lattice")
 			//	src.payload_reagent = pick("flockdrone_fluid") //Looks like Gnesis can't form liquids.
 		src.payload_amount = rand(100, 3000)
--- a/code/obj/artifacts/artifact_machines/wellspring.dm
+++ b/code/obj/artifacts/artifact_machines/wellspring.dm
@@ -9,17 +9,16 @@
 
 /datum/artifact/wellspring
 	associated_object = /obj/machinery/artifact/wellspring
+	type_size = ARTIFACT_SIZE_LARGE
 
 	type_name = "Wellspring"
 	rarity_weight = 350
-	react_xray = list(5,65,20,11,"HOLLOW")
+	react_xray = list(5,65,20,11,"LIQUID RESERVOIR")
 	validtypes = list("ancient", "martian", "eldritch", "precursor", "wizard")
-	//validtypes = list("ancient", "martian", "eldritch", "lattice")
 	validtriggers = list(/datum/artifact_trigger/carbon_touch, /datum/artifact_trigger/silicon_touch, /datum/artifact_trigger/force, /datum/artifact_trigger/heat, /datum/artifact_trigger/cold, /datum/artifact_trigger/radiation, /datum/artifact_trigger/electric, /datum/artifact_trigger/language)
 	activated = 0
 	activ_text = "begins to flood the area with liquid!"
 	deact_text = "stops flooding."
-	touch_descriptors = list("It feels damp.")
 	var/payload_reagent = "water"
 	var/payload_amount = 1
 	var/payload_cooldown = 1 SECONDS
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by CelineTheYeen** [comment]
Personally I'd like to see the list of chems expanded quite a bit, 1 per type and it only being available in 3 origins is limiting in a way I don't think is very interesting. Either expanding to the artifact chemlists (potentially encroaching on chembomb territory a lot) or just a slightly bigger and thematic list like also having oil, nanites or rarely voltagen as part of their possible rolls.

Alternatively I feel like this sort of behaviour could be moved into a container artifact option, releasing a huge amount of chems at once and disappearing/going dormant though that is a different niche.

**Comment by Sbmhawk** [comment]
Im with ^, 1 chem per type means not much variety.  The artifact spews too slowly for it to actually be a flooding danger, and the current chemical list means it'll be inert, outside of the welding fuel.

**Comment by Kanthes** [comment]
The intent to differentiate from the Chembomb by being limited to (relatively) inert reagents and instead focusing on quantity. Additionally, there's only that many (relatively) inert reagents that fit the theme for each origin that can exist as liquids.

I'm very open to suggestions on what other reagents to add.

**Comment by Calphrick** [comment]
I would recommend:

Eldritch - blood
Wizard - something gimmicky but harmless, like reversium or fliptonium
Martian - martian flesh or loose screws
Precursor - epinephrine or ephedrine
Silicon - directed nanites

File: code/obj/artifacts/artifact_machines/wellspring.dm [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+
+
+/datum/artifact/wellspring
+	associated_object = /obj/machinery/artifact/wellspring
+
+	type_name = "Wellspring"
+	rarity_weight = 350
+	react_xray = list(5,65,20,11,"HOLLOW")
+	validtypes = list("ancient", "martian", "eldritch", "precursor", "wizard")
+	//validtypes = list("ancient", "martian", "eldritch", "lattice")
```
TobleroneSwordfish:
```suggestion
```

code/obj/artifacts/artifact_machines/wellspring.dm
```diff
@@ -0,0 +1,53 @@
+/obj/machinery/artifact/wellspring
+	name = "wellspring"
+	associated_datum = /datum/artifact/wellspring
+
+	New()
+		..()
+		src.create_reagents(3000)
```
TobleroneSwordfish, line 7: I don't think this reagent container is used at all? Should either be removed or completely filled with the wellspring reagent type.

code/obj/artifacts/artifact_machines/wellspring.dm
```diff
@@ -0,0 +1,53 @@
...
+	type_name = "Wellspring"
+	rarity_weight = 350
+	react_xray = list(5,65,20,11,"HOLLOW")
+	validtypes = list("ancient", "martian", "eldritch", "precursor", "wizard")
+	//validtypes = list("ancient", "martian", "eldritch", "lattice")
+	validtriggers = list(/datum/artifact_trigger/carbon_touch, /datum/artifact_trigger/silicon_touch, /datum/artifact_trigger/force, /datum/artifact_trigger/heat, /datum/artifact_trigger/cold, /datum/artifact_trigger/radiation, /datum/artifact_trigger/electric, /datum/artifact_trigger/language)
+	activated = 0
+	activ_text = "begins to flood the area with liquid!"
+	deact_text = "stops flooding."
+	touch_descriptors = list("It feels damp.")
```
TobleroneSwordfish:
```suggestion
```
Unique touch descriptors are usually for the really dangerous ones like power generators or borg converters, this isn't bad enough to need one.

code/obj/artifacts/artifact_machines/wellspring.dm
```diff
@@ -0,0 +1,53 @@
...
+	name = "wellspring"
+	associated_datum = /datum/artifact/wellspring
+
+	New()
+		..()
+		src.create_reagents(3000)
+
+
+/datum/artifact/wellspring
+	associated_object = /obj/machinery/artifact/wellspring
```
TobleroneSwordfish, line 11:
```suggestion
	associated_object = /obj/machinery/artifact/wellspring
	type_size = ARTIFACT_SIZE_LARGE
```

code/obj/artifacts/artifact_machines/wellspring.dm
```diff
@@ -0,0 +1,53 @@
...
+			if ("ancient")
+				src.payload_reagent = pick("fuel", "goodnanites")
+			if ("martian")
+				src.payload_reagent = pick("water")
+			if ("eldritch")
+				src.payload_reagent = pick("blood")
+			if ("precursor")
+				src.payload_reagent = pick("ephedrine")
+			if ("wizard")
+				src.payload_reagent = pick("reversium", "fliptonium")
```
TobleroneSwordfish:
```suggestion
			if ("ancient")
				src.payload_reagent = pick("fuel", "goodnanites")
			if ("martian")
				src.payload_reagent = pick("water", "viscerite_viscera", "martian_flesh")
			if ("eldritch")
				src.payload_reagent = pick("blood", "bloodc", "black_goop")
			if ("precursor")
				src.payload_reagent = pick("ephedrine", "cryostylane") //cryo fluid?
			if ("wizard")
				src.payload_reagent = pick("reversium", "mugwort", "fliptonium")
```

Glamurio:
Very cool idea. Agree on letting there be more variety in liquids, especially if we add really good and useful liquids. It'd be neat to be able to use this as a stepladder chem dispenser for chemists to be able to make some more complicated recipes, or of course just medicine or stuff we can place in the medbay.

code/obj/artifacts/artifact_machines/wellspring.dm
Kanthes: I'll be perfectly honest and admit that I do not have a complete understanding how the liquid logic (based on the Shower code) works, but from what I understand with this line I define this object to be a container with a maximum volume of 3000 units. Every effect_process, I fill this container up with the randomized amount of liquid before dumping it out. Without this line, that part of the code ceases to work.

Kanthes: I'll be perfectly honest and admit that I do not have a complete understanding how the liquid logic (based on the Shower code) works, but from what I understand with this line I define this object to be a container with a maximum volume of 3000 units. Every effect_process, I fill this container up with the randomized amount of liquid before dumping it out. Without this line, that part of the code ceases to work.

Kanthes:
[Tarmunora](https://github.com/Tarmunora) made a great point on another artifact PR that we don't want artifacts sitting around endlessly processing when ejected, which I can see happening for this artifact as well. As such, I've added a randomized duration that will eventually shut down the artifact, albeit after a fairly long duration.

PR #24747, opened 2025-09-26
Fixes the description text on the eyepatch after a hole is in it

SunWuDrakios:
## About the PR 
Poking a hole into an eyepatch didn't change the examine text now it does
## Why's this needed? 
It was supposed to have text
## Testing 
eyepatch fix 1
eyepatch fix 2

### DIFF
--- a/code/obj/item/clothing/glasses.dm
+++ b/code/obj/item/clothing/glasses.dm
@@ -471,10 +471,12 @@ TYPEINFO(/obj/item/clothing/glasses/visor)
 				appearance_flags |= RESET_COLOR
 				user.show_message(SPAN_NOTICE("You poke a tiny pinhole into [src]!"))
 				return
+		return ..()
 
 	get_desc ()
 		if (src.pinhole)
 			. += " Unfortunately, its not so cool anymore since there's a tiny pinhole in it."
+
 
 	attack_self(mob/user)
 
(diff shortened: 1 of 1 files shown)


code/obj/item/clothing/glasses.dm
```diff
@@ -466,14 +466,16 @@ TYPEINFO(/obj/item/clothing/glasses/visor)
...
-				pinhole = 1
-				block_eye = null
+				pinhole = TRUE
+				block_eye = FALSE
 				appearance_flags |= RESET_COLOR
 				user.show_message(SPAN_NOTICE("You poke a tiny pinhole into [src]!"))
-				if (!pinhole)
-					desc = "[desc] Unfortunately, its not so cool anymore since there's a tiny pinhole in it."
 				return
-		return ..()
```
ZeWaka, line 476: Why did you remove this? now you can't whack anything on it

SunWuDrakios: Oh I didn't notice that, I tried to copy what I did for the blindfold, I try to fix it, just need to figure out how

SunWuDrakios: Should be fixed now, but I noticed I could whack most things but not firecrackers for some reason

PR #24754, opened 2025-09-27
Puts "cultural" jumpsuits in their own files

aaaaaaaaaaaaaaaaaah:
[CODE QUALITY] [CHORE] [CLOTHING] [OBJECTS] 

## About the PR 
Removes the Hakama, Yukata, America, Kilt and Tricolor Jumpsuits from their old place in the code to .dm and dmi files dedicated to them.

## Why's this needed? 
If someone wants to add more clothing from countries they have a place to put them now, it also makes it easier to find said clothing. This is just another step in cleaning up the jumpsuit code.

## Testing 
Spawned the jumpsuits and tried them on.


DisturbHerb:
as yukata author: not 100% on this categorisation scheme, i dont love the `gimmick` path and how things are presently organised but frankly i dont see why we shouldn't just put all of these into their own file if it's a significant enough issue. we have symbol search anyway

aaaaaaaaaaaaaaaaaah:
> as yukata author: not 100% on this categorisation scheme, i dont love the `gimmick` path and how things are presently organised but frankly i dont see why we shouldn't just put all of these into their own file if it's a significant enough issue. we have symbol search anyway

If you don't like the category I understand that but right now the Hakama and Yukata are in their own sub folder which is just called "under/gimmick" and I would merge that into the gimmick file even if this way to organize things isn't accepted.

PR #24755, opened 2025-09-27
Adds the artifact bomb (radioactive) variant.

Kanthes:
[A-Science][C-Feature]

## About the PR 
Adds a radioactive variant of the regular artifact bomb. The bomb is largely based on the regular artifact bomb, but with an added radioactive element. The bomb will emit a large burst of neutrons on exploding, and stay around after exploding to continue emitting radiation with a decaying rate until it eventually deletes itself.

## Why's this needed? 

The neutron and radioactivity system has been part of the game for a while, it only feels right that there's a bomb that utilizes it.

## Testing 
Tested on a local server, works just like a regular artifact bomb for the most part.
vlc_2Zj0j9MXCd

## Changelog 

```changelog
(u)Kanthes
(*)Adds the artifact bomb (radioactive) variant.
```

### DIFF
--- a/code/obj/artifacts/artifact_objects/bombs.dm
+++ b/code/obj/artifacts/artifact_objects/bombs.dm
@@ -556,6 +556,7 @@ ABSTRACT_TYPE(/datum/artifact/bomb)
 	type_name = "Bomb (radioactive)"
 	rarity_weight = 90
 	react_xray = list(12,75,30,11,"SHIELDED")
+	touch_descriptors = list("Your fingers continue to tingle after touching it.")
 	var/explosion_strength = 1
 	var/nuclear_fallout_cooldown = 10 SECONDS
 	var/nuclear_fallout_amount = 5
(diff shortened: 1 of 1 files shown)


Glamurio:
> It's worth considering that 95% of artifact bombs just get sent to space - do we really want to have these things spewing out neutrons against an asteroid all round? I'd have the fallout amount decay down to 0 over time, and then deactivate/delete the artifact

Unrelated to this PR, but this feels like a problem with bomb etiquette rather than bombs themselves. I kind of wish more bombs actually blew up and destroyed stuff.

Kanthes:
> > It's worth considering that 95% of artifact bombs just get sent to space - do we really want to have these things spewing out neutrons against an asteroid all round? I'd have the fallout amount decay down to 0 over time, and then deactivate/delete the artifact
> 
> Unrelated to this PR, but this feels like a problem with bomb etiquette rather than bombs themselves. I kind of wish more bombs actually blew up and destroyed stuff.

Have a look at #24557!

Glamurio:
> > > It's worth considering that 95% of artifact bombs just get sent to space - do we really want to have these things spewing out neutrons against an asteroid all round? I'd have the fallout amount decay down to 0 over time, and then deactivate/delete the artifact
> > 
> > 
> > Unrelated to this PR, but this feels like a problem with bomb etiquette rather than bombs themselves. I kind of wish more bombs actually blew up and destroyed stuff.
> 
> Have a look at #24557!

Unfortunately not quite what I mean. Bomb artifacts are ejected 90% of the time, they're not teleported to QM for sale. So this won't really affect it, unfortunately.

PR #24761, opened 2025-09-27
Rework of the construction of flash/cell assemblies and full integration in `/obj/item/assembly`

M-Earthfire:
[game objects][rework][feature]

## About the PR 
This PR integrates flash and cells into  `/obj/item/assembly` and partially do some rework/refactor to make them better without adding much multiplied code
flash_hand-telelflash-cellmousetrap_flash_celll

This change comes with some new features for assemblies and flashes:
- If an assembly can be an applier and trigger (this is now the signaler/flash-situation) the you can choose which item becomes the trigger.
- flashes don't need to be opened with a screwdriver to be combined with cells
- Flash/cell assemblies can now be disassembled to retreive the cell
- Flashes can be used either as trigger or as applier. So you can hook a flash onto a mousetrap or you can hook a multitool onto a flash
- If a flash-assembly is used with the assembly as trigger, using the flash triggers the whole assembly
- if the cell in the flash/cell-assembly does not have sufficient charge (500), It won't trigger the flash/cell assemblies specific behaviour and will work like a normal cell

Because i needed to cut the flashes behaviour into multiple procs, i changed everything to use absolute pathing and changed some code styling up.

## Why's this needed? 

This enables the use of flashes in assemblies, which opens up more opportunities for traps/shenanigans. Also, this makes the construction and handling of cell/flash assemblies much more cleaner

## Testing 

https://github.com/user-attachments/assets/878f0ef8-78a3-4394-8ee1-e15470e0db02

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Non-rev Flashes can be build into assemblies as either triggers or appliers. Have a mousetrap/flash/cell assembly ready for when a search comes up!
(+)Flashes don't need to have their cover removed to be combined with cells. Flash/Cell assemblies can also be dismantled to retreive the used cell
(+)Flash assemblies will trigger their corresponding applier. If your flash/igniter/pipebomb assembly blows up your ass: i warned you...
(+)If both parts can be triggers or appliers of an assembly, you can choose which one to use as a trigger.
(+)Flash/cell assemblies with insufficient charge will behave like normal flashes.
```

### DIFF
--- a/code/datums/components/assembly_comp.dm
+++ b/code/datums/components/assembly_comp.dm
@@ -153,4 +153,4 @@ TYPEINFO(/datum/component/assembly)
 /datum/component/assembly/trigger_applier_assembly/proc/delayed_combination_valid_check(var/obj/item/checked_item, var/mob/user)
 	// we check here if we can still reach both items and they weren't build into assemblies at a different point
 	var/obj/item/checked_parent = src.parent
-	return (checked_parent && checked_item && !QDELETED(checked_item) && !QDELETED(checked_parent) && can_reach(user, checked_item) && can_reach(user, checked_parent) && !istype(checked_item.loc, /obj/item/assembly) && !istype(checked_parent.loc, /obj/item/assembly))
+	return (can_reach(user, checked_item) && can_reach(user, checked_parent) && !istype(checked_item.loc, /obj/item/assembly) && !istype(checked_parent.loc, /obj/item/assembly))


M-Earthfire:
>The active hand invisibly changing the behaviour of the assembly is very unintuitive, how much hackery would be necessary to make it just do both? (ie: the flash triggers when a signal is received AND a signal is sent when the flash is used)

The thing is while that is possible, it wouldnt be able to accomplish much. If we have it as a applier, it surely can trigger the assembly, but triggering the assembly just causes the applier to trigger (the flash). The system fundamentally doesn't support more than one trigger/applier or target.

The problem is here in the base design of the assembly system. This will be a problem each time we have something that can either be an applier or a trigger. And while this time we could avoid it by just having a single item that works as both a applier and trigger (the signaler), the same problem will come up each time we try to design a multifunctional item for the assembly.

This solution is what i came up with to solve this problem for any other component that might get integrated in the system in the future.

The change is visible, though. The flash/signaler assembly has a different sprite than the signaler/flash assembly.

M-Earthfire:
Thinking about it... a context menu could be a different solution. Just choose what you want to be the trigger.

M-Earthfire:
This should make it more intuitive and should solve it for any other potential trigger/applier-item

https://github.com/user-attachments/assets/ffb41761-3d94-4b04-baaf-7b5ffb7923ae

code/datums/components/assembly_comp.dm
```diff
@@ -122,12 +122,23 @@ TYPEINFO(/datum/component/assembly)
 	. = ..()
 
 /datum/component/assembly/trigger_applier_assembly/override_combination(var/atom/checked_atom, var/mob/user)
+	var/obj/item/item_to_be_trigger = src.parent
+	var/obj/item/item_to_be_applier = checked_atom
+	// Here, we take care of the special case that both items can be triggers and appliers. In that case, we open a context menu and ask how we want to build the assembly
+	if((length(checked_atom.GetComponents(/datum/component/assembly/trigger_applier_assembly)) > 0) && isassemblyapplier(src.parent))
+		var/input_action = input(user, "Which item do you want to be the trigger of the assembly?") in list("[src.parent]","[checked_atom]","Never Mind")
```
TobleroneSwordfish: Would probably be better as "Which item should be activated when the assembly triggers?" or something, asking which one is the trigger is a little unclear.

M-Earthfire: switched the question to ask for the applier instead. It now asks: "Which item should be activated if the assembly triggers?"

code/datums/components/assembly_comp.dm
```diff
@@ -124,3 +148,9 @@ TYPEINFO(/datum/component/assembly)
 	user.put_in_hand_or_drop(product)
 	boutput(user, SPAN_NOTICE("You finish the construction of [product.name]."))
 	return TRUE
+
+
+/datum/component/assembly/trigger_applier_assembly/proc/delayed_combination_valid_check(var/obj/item/checked_item, var/mob/user)
+	// we check here if we can still reach both items and they weren't build into assemblies at a different point
+	var/obj/item/checked_parent = src.parent
+	return (checked_parent && checked_item && !QDELETED(checked_item) && !QDELETED(checked_parent) && can_reach(user, checked_item) && can_reach(user, checked_parent) && !istype(checked_item.loc, /obj/item/assembly) && !istype(checked_parent.loc, /obj/item/assembly))
```
TobleroneSwordfish:
```suggestion
	return (can_reach(user, checked_item) && can_reach(user, checked_parent) && !istype(checked_item.loc, /obj/item/assembly) && !istype(checked_parent.loc, /obj/item/assembly))
```
You can't reach deleted things, by definition.

TobleroneSwordfish (approved):

PR #24766, opened 2025-09-28
Extradimensional Storages May Now Use Prefabs

Mister-Moriarty:
## About The PR:
Extradimensional storages may now use prefabs as their allocated region. To facilitate this, new extradimensional exit and entrance components have been implemented, entrances representing places where objects entering the prefab may be placed, and exits representing ways in which objects may leave the prefab.

Entrances and exits are implemented by way of signal-component-callbacks, wherein components listen for enter/exit signals, then invoke a callback defined by the component's parent. This is a slightly complicated way of implementing entrances/exits, but it allows the entrance/exit object to have complete control over how objects enter and leave the prefab. `_exit_locker.dm` is a simple example of how it works, but you could easily implement a disposal outlet that flings entering objects across the prefab using the system.

Also splits the extradimensional storage file into several smaller files.

## Why Is This Needed?
Extradimensional lockers that use prefabs came up in conversation in #imcoder as potentially being useful for features such as secret offices.

## Testing:
Tested using an extension of the shady office random room with an exit locker present.


code/datums/components/extradimensional_storage/extradimensional_prefabs/_prefab_parent.dm
```diff
@@ -0,0 +1,48 @@
...
+		if (!src.region.turf_in_region(get_turf(entrance.parent)))
+			continue
+
+		src.entrances += entrance.parent
+
+	for_by_tcl(exit, /datum/component/extradimensional_prefab_exit)
+		if (!src.region.turf_in_region(get_turf(exit.parent)))
+			continue
+
+		src.RegisterSignal(exit, COMSIG_EXTRADIMENSIONAL_PREFAB_EXIT, PROC_REF(try_exit))
```
Tarmunora, line 38: This code looks smelly to me. Surely there's a clearer solution than this?

Mister-Moriarty: I thought that too, but I don't think there is any way of getting a turf's region without iterating through each region and doing a similar check as above, except on the two components?

Glamurio:
Pocket-dimension artifacts are so happening once this is merged.

code/datums/components/extradimensional_storage/extradimensional_prefabs/_prefab_parent.dm
Glamurio: Would it be faster to just `for_by_tcl` once on `/datum/component` and then do an `istype()` on both entrance and exit respectively?

Mister-Moriarty: > Would it be faster to just `for_by_tcl` once on `/datum/component` and then do an `istype()` on both entrance and exit respectively?

No, `for_by_tcl` is typecheckless.
It works by iterating over all elements of an entry in the `by_type` list, in this case `by_type[/datum/component/extradimensional_prefab_entrance/exit]`, which contains all current instances of those types. The type lists are set up and maintained by the `START/STOP_TRACKING` macros.

PR #24767, opened 2025-09-28
Addiction rework

Egregorious:
## About the PR 

Reworks addiction in a way that aims to promote players managing the issue, and gives more avenues for future mechanics to interact with addictions specifically. 

Essentially, withdrawal symptoms can be managed by taking any addictive reagent of the same or greater severity, but doing this consistently will slowly worsen your addiction over time. The best way to cure an addiction naturally without dealing with withdrawal is by taking very small doses of the addicted reagent whenever symptoms are felt. The worse the addiction gets, the more of your addicted reagent will be metabolised per tick, and thus the less effective it is. 

- Addictions have an internal meter that tracks how strong the addiction is
  - Once the meter reaches 0, the addiction is cured
  - This meter depletes at a rate 0.25x the addicted reagent's metabolisation rate every tick
  - This means for a standard reagent with 0.4 metabolisation rate, a minor 15 point addiction will last 5 minutes, while a severe 100 point addiction will last 30 minutes, assuming no outside influence
  - It is also increased by 1x the unit value of the addicted reagent metabolised
  - As well as by twice the depletion rate every tick at least one addictive reagent of the same or greater severity is metabolised
  - The meter is initialised with the addiction_tally value of the addicted reagent (with a minimum of 10)
  - The addiction will no longer be cured by staying at stage 5 for a period of time, the withdrawal symptoms and internal strength of the addiction are independent of one another
- Addictive reagents have severity levels; consuming an addictive reagent of the same severity level or greater as what you're addicted to will keep withdrawal symptoms from worsening, but won't cure them if the stages are already high
  - Withdrawal symptoms do not begin to rise until ~2 minutes after the last dose of the addicted reagent has left the body
  - This delay is lowered by 0.5 seconds times the addiction meter
  - Non-addicted addictive reagents of the appropriate severity will lower the withdrawal stages by 1, but cannot lower them again until the stage has risen naturally
  - They also do not count as a dose of the addicted reagent, and thus won't trigger the 2 minute grace period
  - There is a microdose limit of 0.04 units, which prevents too little metabolisation from affecting symptoms
- You will metabolise more of the addicted reagent the stronger the addiction
  - the rate is 0.1 additional units metabolised per tick per 10 points on the addiction meter

Additionally: 
- Added `/mob/try_affect_all_addictions()` which can be used to increase or decrease all active addictions if the mob has any. It will take effect on the next life tick
- Added an admin verb "Adjust Addictions" to make use of `try_affect_all_addictions`
- 'Addict' trait increases the associated reagent's addiction tally on the mob by 60 when added

## Why's this needed? 

These are mainly ideas suggested in this forum thread: https://forum.ss13.co/showthread.php?tid=20734

Addiction isn't super interesting to deal with, and is too simple mechanically to easily add more gameplay elements to affect. This PR aims to make the disease more interesting to manage, and provides ways to mitigate the symptoms without being able to cure the underlying affliction easily. This helps ensure the disease creates a demand for drugs, rather than just being another illness to cure asap. 

This PR does not attempt to tackle two of the biggest issues with addiction: that it's too easy to get an addiction in the first place, and that there's not enough methods to get rid of an addiction. However this is intended to enable more ways to implement solutions to the latter.

Addictions themselves aren't affected by 'strong willed' or 'addictive personality' traits directly because, as per their description, these traits affect the chance to gain an addiction as opposed to the effects of having one. 

## Testing 

Most of the testing was done through taking triplemeth and meth and watching the addiction_meter variable on the disease go up and down, and by stepping through the procs themselves. Adjust Addictions can be used to test the addiction falling off below a value of 0. Otherwise the changes are too small and slow, I can't think of a screenshot or video that would be particularly informative. 

## Changelog 

```changelog
(u)Egregorious
(*)Addiction strength and withdrawal symptoms are now independent, you can cure your addiction while taking sufficiently small (under 5 unit) doses to calm symptoms when needed
(*)Any addictive reagent of the same severity as you're addicted to can be used to prevent withdrawal symptoms from worsening, but this is bad for it in the long term
(*)The worse your addiction gets, the more quickly your body will metabolise that reagent, reducing its effectiveness
```


code/datums/disease.dm
```diff
@@ -317,19 +323,71 @@
...
 
 	copy_other(datum/ailment_data/addiction/other)
 		..()
 		src.associated_reagent = other.associated_reagent
-		src.withdrawal_duration = other.withdrawal_duration
-		src.max_severity = other.max_severity
+		src.severity = other.severity
+		src.addiction_meter = other.addiction_meter
+		src.depletion_rate = other.depletion_rate
+		src.stage_satisfied = other.stage_satisfied
```
Tarmunora, line 339: intentionally not copying extra_metabolism here?

code/modules/chemistry/Chemistry-Reagents.dm
```diff
@@ -186,13 +186,9 @@ datum
...
 
 				if(INGEST)
-					var/datum/ailment_data/addiction/AD = M.addicted_to_reagent(src)
-					if (AD)
-						M.make_jittery(-5)
-						AD.last_reagent_dose = world.timeofday
-						if (AD.stage != 1)
-							boutput(M, SPAN_NOTICE("<b>You feel slightly better, but for how long?</b>"))
-							AD.stage = 1
+					if (src.addiction_prob && M.ailments && length(M.ailments) < 5)
```
Tarmunora: Why the ailments length check?

code/datums/disease.dm
```diff
@@ -317,19 +323,71 @@
...
+			return FALSE
+		. = ..()
+		if (.)
+			src.stage_satisfied = FALSE
+
+	proc/metabolised_addictive_reagent(var/datum/reagent/reagent, var/rate, var/mult)
+		// The minimum rate means that patches with less than ~1 unit of the addictive reagent usually won't work to satisfy addiction.
+		// This is useful because dose logic is very binary and exploitable by microdosing with ludicrously small volumes.
+		if (src.severity < reagent.addiction_severity || rate < 0.04 * mult)
+			return
```
Tarmunora: Might want to check your conditional here. PR desc suggests that higher-severity reagents will 'help' with lower severity addiction, this seems to do the opposite

Glamurio:
As someone who has strong opinions on addictions, and has commented a lot in the thread you are citing, I think reworking addictions is a great idea and this is a good step.

Treating one addiction with another drug might be an interesting idea, although it's not very intuitive and a bit odd. You say "Addictive reagents have severity levels", but the vast majority of addictive chems does not have a severity level and is not included in your PR.  Stuff like morphine, ephedrine, and even enriched MSG all have addiction chances, but no severity.

It also doesn't address the root problem of addictions, which is how they occur. You can still get addicted just from drinking a cup of coffee really quickly, which is kind of a lot more annoying.

Egregorious:
Severity levels already existed, it was just used to determine which addictions made you throw up, the default is high. 

The addiction_tally seems a fairly decent system for mitigating the randomness of gaining an addiction already, it's just a numbers problem - almost every drug has such a low minimum and the tally decays so slowly that addictions can happen from basic usage very easily. To be honest it seems as if it got set up very gingerly and then never looked at since. 

I would like to go through each addictive reagent and adjust their severity and addiction_tally minimum as needed, but that seemed like too far to go in one PR and I wanted to wait and see if the feedback on this was negative or not first. Most minimums have a value of ~10 and need to be tripled at the very least in my opinion. 

> although it's not very intuitive and a bit odd

It is a bit unintuitive I agree, there will be basically no way of knowing the severity level of every reagent without checking the wiki, however there is a message whenever an appropriate chem is ingested which makes trial and error possible. I also don't see this as too much of a problem anyway because it can only really work in the player's favour. The downside to it is so minor that the basic rule of "any addictive substance is bad for your addiction long term" works; it's not like taking medicine while addicted will cause your addiction to become unbearable, it'll just cause the issue to tick up instead of down a little. 

Although saying this, it could probably do with only occuring once-per-tick, instead of per reagent, that way it's strictly a long-term thing. 

Severity is basically just meant to ensure your triple meth addiction isn't satisfied by cheese substitute, but this is Goonstation and perhaps that would be desirable.

code/modules/chemistry/Chemistry-Reagents.dm
Egregorious: I don't know, I couldn't figure out why the check was there to begin with so I thought it better to not to silently remove it. I'll take it out now that it has perplexed someone else though.

code/datums/disease.dm
Egregorious: Nope

PR #24769, opened 2025-09-28
Adds skull mask variants for Vampire and Changeling

Driftered:
## About the PR 
Adds skull masks for the Vampire and Changeling. This includes two new sprites and a change to how skull carving is handled.

Previously, skull carving was handled with a blanket statement that changed any skull hit with a sawing tool into a human skull mask. The changes made to skull.dm allow for future skull masks to be added via the addition of a new else-if argument above the final default one.

As the Vampire's skull icon glows red, I added frames to the skull mask to match it. You can see the first and final frame of the glow in the preview found below.

## Why's this needed? 
Currently, when someone carves a skull of any kind, it turns into a human skull mask. You've probably heard about or felt the disappointment when a Vampire/Changeling is brought down, their brain is cut out, and the Roboticist tries to make a mask out of the unique skull.. only for it to magically turn into a human skull. This PR aims to fix that.

## Preview
Vampire Skullmask
Xeno Skullmask

## Changelog 

```changelog
(u)Driftered
(*) Adds two new skull mask sprites: Vampire and Changeling
```
[Clothing] [Sprites]


JORJ949:
Changelog is for what the players will see in game, so you only really need the first line

Driftered:
> Changelog is for what the players will see in game, so you only really need the first line

Got it! Wasn't entirely sure if they only saw the major bullet points or all of them, so thank you.

code/obj/item/clothing/gimmick.dm
```diff
@@ -935,14 +935,30 @@ TYPEINFO(/obj/item/clothing/under/gimmick/dawson)
...
+	see_face = FALSE
+
+/obj/item/clothing/mask/skullA // Changeling Skull
+	name = "odd skull mask"
+	desc = "Don't ask why you can see through the skull without any eyeholes."
+	icon = 'icons/obj/items/organs/skull.dmi'
+	icon_state = "skullA"
+	item_state = "death"
+	see_face = FALSE
+
```
Tarmunora, line 961: Might be nice to pull these all under a parent item, either the default skull mask or an abstract parent, just so that if there's future code added around skull masks, it can all be in one place

Driftered: You're absolutely right, and I appreciate you catching that. I actually went back and renamed/repathed skull items entirely, so I'll be changing how I approach this PR. It'll all be under a parent item when I do!

PR #24773, opened 2025-09-29
Makes seasonal clown outfits comply with clown disbelief

bacon792:
## About the PR 
Just adding a small bit to make both seasonal clown outfits be clown disbelief items. I don't believe this should impact anything due to being a very small change.
## Why's this needed? 
I think clowns that choose to wear a seasonal outfit should still be treated as a clown for the clown disbelief trait.
## Testing 
Equipped both seasonal clown outfits while I had clown disbelief, was unable to see myself while either seasonal clown uniform was equipped.

[clothing]


bacon792:
Okay! I double checked and everything still works. I don't think I've overlooked anything.

PR #24774, opened 2025-09-29
Fixes a bug/expliot with the field generator

Mikeistaken:
[A-Engineering] [A-Game-Objects] [C-Bug]

## About the PR
Field generators turned off with packets Refuses to connect with other field generators in the same direction when turned on, until it runs out of power. This will fix that.

## Why's this needed?
Fixes a bug with field generators that could potentialy be exploited

## Testing
I turned off the field generator with packets to find that the field generator refused to fully turned off. I also found in this state the field generator refused to turn on. To test this I had used the commented code to turn of the Field generators by hand, since that used simular code and found it did the same thing. I had applied the fix found in here to both the commented code and the packet code then tested it with packets (sender=00000000;address_1={Field_Generator};command=deactivate) and found that they worked for both sides, then I re-commented the unused code.
## Pre-Fix
Off
7D7388A6-691F-41B0-924E-83AFD7EEBD95
Turned on then off
63AC40A3-2B36-4F4C-BFB9-2F522D84BEDF
Turned on after previous image
74566D4A-FD56-46D9-882E-F7FF5EB0AC11
Post fix should fix issue shown


Mikeistaken (commented):
[A-Engineering] [A-Game-Objects] [C-Bug]

JORJ949:
> [A-Engineering] [A-Game-Objects] [C-Bug]

You didn't do it wrong don't worry, the keylabeler no longer works

Mikeistaken:
ah thats good, I probably need to fill out that forum anyway so I can label stuff

code/obj/machinery/singularity/field_generator.dm
```diff
@@ -52,7 +52,10 @@ TYPEINFO(/obj/machinery/field_generator)
 	//			src.active = 0
 	//			icon_state = "Field_Gen"
 				boutput(user, "You are unable to turn off the field generator, wait till it powers down.")
-	//			src.cleanup()
+	//			src.cleanup(NORTH)
+	//			src.cleanup(SOUTH)
+	//			src.cleanup(EAST)
+	//			src.cleanup(WEST)
```
TobleroneSwordfish:
```suggestion
```
Don't PR commented code

code/obj/machinery/singularity/field_generator.dm
```diff
@@ -386,6 +389,10 @@ TYPEINFO(/obj/machinery/field_generator)
 
 	if(sigcommand == "deactivate" && src.active)
 		src.set_active(0)
+		src.cleanup(NORTH)
+		src.cleanup(SOUTH)
+		src.cleanup(EAST)
+		src.cleanup(WEST)
```
TobleroneSwordfish, line 392: Doesn't this let you turn it off instantly using packets, instead of waiting for it to power down?

Mikeistaken: It does yes, but the previous one also let you turn it off instantly using packets but it also prevented it from turning on after it was turned off, this fixes that. if you want to prevent turning it off with packets just comment the entire area. This issue is present and is the reason why I had PR'ed this

code/obj/machinery/singularity/field_generator.dm
Mikeistaken: Alright, I thought I would add that in case it was ever un-commented, as the same fix applies to that area as well

Mikeistaken: Current branch should have the changes undid

Mikeistaken:
I have accedentaly merged my master into this branch... hopefully it wont cause issues and ill fix it after its been merged. if its expected to cause issues then it can be closed

TobleroneSwordfish:
Okay it seems like this entire feature was an oversight from when field generators were changed to need to power down before being turned off, so I've just gone ahead and removed it instead: 64935c7 - I don't think being able to instantly loose the singulo with a PDA program is a good idea.

PR #24780, opened 2025-09-30
Enable poltergeist limit on classic

JORJ949:
## About the PR 
Removes the ifdef RP_MODE check for the limit on trickster poltergeists, introducing the limit of 2 poltergeists per wraith to the classic server.

## Why's this needed? 
Poltergeists are incredibly strong if you can consistently get a lot of them, even 3 possessed items at once is strong, more is nearly impossible to handle when you have 5 poltergeists all possessing items at once. Mechanical differences between classic and RP should be kept to what's required.

## Testing 
Edited my wraith poltergeist list to list("buh", "guh") and accurately could not spawn additional poltergeists

## Changelog 

```changelog
(u)JORJ949
(*)The limit of 2 poltergeists per wraith is now present on all servers, rather than just RP.
```


Laboredih123:
but the funi

PR #24804, opened 2025-10-02
Full Nukes declare round completion instead of rebooting the server

JORJ949:
## About the PR 
Adds "force_round_finished" variable to gamemode parent, if this variable is true the ticker will complete the round the same as it would if the shuttle's location was returned to centcom.
Full size "nuke" strength nuclear bombs outside of the nuclear emergency mode now set this variable to true, ending the round properly instead of taking 30 seconds to reboot the server. Previously this only worked in the Nuclear Emergency mode due to its own check_completion checking for a null the_bomb variable.

## Why's this needed? 
Admin gimmick nukes and any other future potential roundending nuke spawns not in the Nuclear Emergency mode should end the round the normal way, rather than forcing a server reboot and not performing other roundend things such as showing crew credits and enabling OOC.

## Testing 

https://github.com/user-attachments/assets/463542db-701f-4101-8120-6056a8476e6e

### DIFF
--- a/code/obj/machinery/nuclearbomb.dm
+++ b/code/obj/machinery/nuclearbomb.dm
@@ -483,13 +483,8 @@ ADMIN_INTERACT_PROCS(/obj/machinery/nuclearbomb, proc/arm, proc/set_time_left)
 
 		creepify_station()
 
-		if(!istype(gamemode))
-			sleep(1 SECOND)
-			boutput(world, "<B>Everyone was killed by the nuclear blast! Resetting in 30 seconds!</B>")
-
-			sleep(30 SECONDS)
-			logTheThing(LOG_DIARY, null, "Rebooting due to nuclear destruction of station", "game")
-			Reboot_server()
+		if(!istype(gamemode)) //The station exploded, the cinematic has played, end the round naturally instead of forcing a reboot
+			ticker?.mode?.force_round_finished = TRUE
 
 	proc/change_status_display()
 		var/datum/signal/status_signal = get_free_signal()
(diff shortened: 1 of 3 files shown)


code/obj/machinery/nuclearbomb.dm
```diff
@@ -483,13 +483,8 @@ ADMIN_INTERACT_PROCS(/obj/machinery/nuclearbomb, proc/arm, proc/set_time_left)
...
 
-		if(!istype(gamemode))
-			sleep(1 SECOND)
-			boutput(world, "<B>Everyone was killed by the nuclear blast! Resetting in 30 seconds!</B>")
-
-			sleep(30 SECONDS)
-			logTheThing(LOG_DIARY, null, "Rebooting due to nuclear destruction of station", "game")
-			Reboot_server()
+		if(!istype(gamemode)) //The station exploded, the cinematic has played, end the round naturally instead of forcing a reboot
+			ticker?.mode?.force_round_finished = TRUE
```
TobleroneSwordfish:
```suggestion
			ticker.mode.force_round_finished = TRUE
```
if ticker.mode does not exist you have bigger problems than one runtime

TobleroneSwordfish (approved):

JORJ949:
Just noticed some modes' check_completion doesn't call their parents, but most of those are admin-only gimmick modes and an admin has to be there anyway presently to spawn one of these nukes so if the round doesn't end they can restart manually (or get spammed with ahelps)

TobleroneSwordfish:
Uhh in that case you could keep the delayed restart code but put it behind a check for if the game state is still `GAME_STATE_PLAYING`, that way you'd guarantee a restart in all cases but defer to gamemode handling if available.

JORJ949:
> Uhh in that case you could keep the delayed restart code but put it behind a check for if the game state is still `GAME_STATE_PLAYING`, that way you'd guarantee a restart in all cases but defer to gamemode handling if available.

Instead of tying this to the bomb I made the force_round_finished bypass check_finished all together and add it as an extra check in the ticker process besides the check_finished call.

PR #24811, opened 2025-10-02
Security Department Announcement Computers

JORJ949:
## About the PR 
Adds a general security department access computer with departmental announcement styling and sound(cargo/catering currently have these).
The HoS keeps their own announcement computer with the command styling and sound.

## Why's this needed? 
Unlike the other departments, security _cannot_ be promoted to their relevant head of staff to gain access to the department's announcement computer, leaving options when there is no HoS for making announcements to using Catering's or asking for Bridge access, neither of which are particularly nice options. Allows security themselves to request perpetrators to hand themselves in, issue PSAs, and do other security announcements without needing a HoS or additional access. 

## Testing 
image
image

## Changelog 

```changelog
(u)JORJ949
(*)The Security Department now has their own announcement computer locked to security access, instead of just the HoS (though the HoS still has their own cooler version).
```


aaaaaaaaaaaaaaaaaah:
This or a megaphone, in general the HoS or the department could use a megaophone.

PR #24819, opened 2025-10-03
Replace the stun on aranesp with a slow

M-Earthfire:
[Balance][Chemicals]

## About the PR 

Replace the 2 second stun on aranesp with a 3 second slow. Also set it to a probmult instead of a prob-proc.

## Why's this needed? 

Aranesp is an interesting stimulant chem that is sadly mostly unused. In comparison to other stimulants it is quite damaging and quite hard to create, requiring complex reaction control and botany/med exclusive chems.

The chance to get hit with a 2 second full stun in addition to that makes the chemical very unappealing to use. by replacing this with a much less punishing slow, it should encourage people to use the chemical even if its synthesis is quite delicate.

## Testing 

Used the chemical, get hit by a random slow. I can attach a mp4, if needed.

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Aranesp will not cause stuns anymore. Instead, it will slow the user with an equal probability.
```


code/modules/chemistry/Reagents-Misc.dm
```diff
@@ -239,9 +239,9 @@ datum
...
 				if (probmult(5)) M.emote(pick("twitch", "shake", "tremble","quiver", "twitch_v"))
 				if (probmult(8)) boutput(M, SPAN_NOTICE("You feel [pick("really buff", "on top of the world","like you're made of steel", "energized", "invigorated", "full of energy")]!"))
-				if (prob(5))
+				if (probmult(5))
 					boutput(M, SPAN_ALERT("You cannot breathe!"))
-					M.setStatusMin("stunned", 2 SECONDS * mult)
+					M.setStatus("slowed", 3 SECONDS)
 					M.take_oxygen_deprivation(15 * mult)
 					M.losebreath += (1 * mult)
 				..()
```
Tarmunora, line 247: double-mult-adjusting on the oxy damage etc

Tarmunora:
I do have some concern that, while a 3s slow is not insignificant in a vacuum, in the context of where this chem will be used it may be somewhat negligible.
When taking into consideration how much stamina regeneration Aranesp gives to sprint with, and the likelyhood of being used in conjunction with other boosts e.g.: meth, how much is that slow actually going to hurt?

M-Earthfire:
I attached a video here of aranesp and meth in conjunction (yeah, you can count the pixels, githubs 10MB requires some compression):

https://github.com/user-attachments/assets/f90c54af-6f30-4882-b002-6efd4a4aefe6

Now, the slowdown (40 second mark) is, in conjunction with the slow from the sudden 15 oxy damage on top of it, significant, even with methaphetamine. But i can see it being a bit rare. I honestly could see upping the chance to like 6-9% per tick.

The overall sentiment after #20518 was that the chem was overall nerfed in availability, so i assume that it can be alright with 5%, since mass production is much less feasable than e.g. methaphetamine. Maybe that's something to tweak with a few rounds in testmerge.

Edit: I falsely thought the chance it has now is 8%. It was 5%.

code/modules/chemistry/Reagents-Misc.dm
M-Earthfire: i removed it on the slow but not the deprivation and losebreath... fixed.

PR #24820, opened 2025-10-03
combines seal_hair and over_hair into a c_flag

Valtsu0:
[clothing][code-quality]
## About the PR 
whether or not clothing covers hair used to be determined by 3 variables depending on the clothing type (2 of these were for suits and both were always the same). Now it is only 1 c_flag

## Why's this needed? 
better code
also fixes a bug that caused chameleon adeptus mechanicus suits to not hide hair unlike regular admech suits

## Testing 
Tested hats and suits with noticeable hair. Special attention was given to chameleon hat and suit, dinosaur pajamas (toggle able hood) and bedsheet (can be made into a cape) 
image
image
image
image

### DIFF
--- a/code/obj/item/clothing/helmets.dm
+++ b/code/obj/item/clothing/helmets.dm
@@ -19,12 +19,11 @@
 /obj/item/clothing/head/helmet/space
 	name = "space helmet"
 	icon_state = "space"
-	c_flags = SPACEWEAR | COVERSEYES | COVERSMOUTH | BLOCKCHOKE
+	c_flags = SPACEWEAR | COVERSEYES | COVERSMOUTH | BLOCKCHOKE | COVERSHAIR
 	see_face = FALSE
 	item_state = "s_helmet"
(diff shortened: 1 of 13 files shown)


code/obj/item/clothing/helmets.dm
```diff
@@ -46,7 +45,7 @@
 		icon_state = "space-fish"
 		desc = "You're about 90% sure this isn't just a regular fishbowl."
 		item_state = "s_helmet"
-		seal_hair = 0
+		c_flags = SPACEWEAR | COVERSEYES | COVERSMOUTH | BLOCKCHOKE
```
ZeWaka: Why not :: use the parent and add/remove

Valtsu0: Didn't know that exists. That's actually really useful for something else i tried to do too. I'll go trough this again and do that where applicable

ZeWaka (approved):
ty!

PR #24821, opened 2025-10-03
Adds the Mindscrambler artifact.

Kanthes:
[SCIENCE][FEATURE]

## About the PR 
Adds the Mindscrambler large artifact. The Mindscrambler is an eldritch origin artifact with a demonic effect. Once activated, every person who touches the artifact will be remembered, and each time a new target touches the artifact, all minds are randomly distributed.

The Mindscrambler is restricted to alive humans with minds, preventing mindswapping with unintentional targets. Vampires and changelings will also be restricted from being scrambled.

The Mindscrambler has an interaction with the Bible, exorcising the demonic entity within until it is activated again, returning every mind to their original body.

The Mindscrambler also has an interaction with the mindhack implant, allowing a lucky Scientist traitor to create a swarm of mind controlled subjects.

## Why's this needed? 

The Mindscrambler allows for a rare opportunity for mindswap shenanigans without allowing swapping with things that are not meant to be controlled. It adds a unique victory condition for traitor scientist, assuming they are lucky enough to find the rare artifact.

## Testing 

This PR is still being worked on.

## Changelog 

```changelog
(u)Kanthes
(*)Adds the Mindscrambler artifact, which remembers everyone who touches it and randomly distributes consciousnesses between them.
```

### DIFF
--- a/code/obj/artifacts/artifact_machines/mindswapper.dm
+++ b/code/obj/artifacts/artifact_machines/mindswapper.dm
@@ -2,6 +2,19 @@
 	name = "mindscrambler"
 	associated_datum = /datum/artifact/mindscrambler
 
+	attackby(obj/item/W, mob/user)
+		/datum/artifact/mindscrambler/attacked_artifact = src.associated_datum
+		if (istype(W, /obj/item/bible))
+			attacked_artifact.exorcised = attacked_artifact.exorcised + 1
+			switch (attacked_artifact.exorcised)
+				if(1)
+					//Knock back
+				if(2)
+					//Hiss and knock back
+				if(3)
+					attacked_artifact.ArtifactDeactivated()
+
+
 /datum/artifact/mindscrambler
 	associated_object = /obj/machinery/artifact/mindscrambler
 	type_name = "mindscrambler"
@@ -18,6 +31,7 @@
 	var/mob/living/carbon/human/last_activating_body
 	var/mindscramble_cooldown = 30 SECONDS
 	var/list/cooldowns = new/list()
+	var/exorcised = 0
 
 	post_setup()
 		..()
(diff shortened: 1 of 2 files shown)


Glamurio:
I think monkeys should be added to the list. It'd be funny for people to be mindswapped with monkeys since that's the first thing a regular scientist would shove against it. I also think that when the artifact is destroyed or potentially leaves the Z level, the minds should also return to normal. It'd be kind of a pain if the artifact gets ejected or sold that people are stuck in others bodies. Just for the sake of sanity, I would blacklist antags from being able to be mindswapped entirely, because this sounds like a nightmare for Security.

Also, unrelated, maybe it should do some brain damage when this happens, just because it's thematic.

RSOD404:
all people marked as antag should be immune or at least warned in some way, because while i like the idea, getting mindswapped by a random artifact and having the other person find out i have a gun / syndie equipment in my backpack or that i'm an arcfiend is bad.
also, to make the immunity less suspicious, perhaps make the swaps happen between a small amount of those remembered and "forget" people if they've been randomly swapped back into their body

ju45he:
If antags are immune then they'll be the only ones that don't swap bodies and hence would be found out to be an antag.
If antags aren't immune then they'll be caught when some random in their body finds all their shit.
If antags choose then they run into both of the above problems.
If everyone chooses then it loses some fun value and all value to that scientist traitor scenario.

kbsmilk:
There's a case to be made that antags shouldn't be touching artifacts unless they wanna get messed up. There's already plenty of those that just kill you, and then the doctors peek at the gun in your bag at cloning and you're screwed.

DisturbHerb:
antag status and abilities should totally transfer with mindswaps, change mine view >:)

Kanthes:
Man, this one was complex to build. Lot of edge cases, and some I'm sure I haven't thought of yet. I'd highly appreciate anyone who does further testing or thinks of possible bugs.

I did my best with the sound effects and animations, but they do still feel rather amateurish. If anyone has more experience with them, I'd really like to see what they could do instead.

Valtsu0:
I think it should have foreboding text. It is disruptive enough and fits as the eldrich version of borg converter

Kanthes:
Understood! As it's quite the departure from my original idea I'll close this PR for now, and depending on motivation I might return to it down the line.

PR #24824, opened 2025-10-03
Adds Therapy

Egregorious:
## About the PR 

Adds a simple method to aid in getting rid of an addiction, while encouraging and rewarding therapy roleplay

Whenever someone talks in a room designated for therapy while two or more people total are present, everyone in the room has their addictions lowered incrementally
- the chapel and the psychiatrists office are designated for therapy
- certain job traits use their own cooldown for certain areas, and are thus more effective
- this is the chaplain for the chapel, and the psychiatrist for their office
- the therapy effect has a cooldown of 10 seconds
- each tick has a value of -0.8
- this means that ~5 points of addiction can be cured per minute, or ~10 if the chaplain/psychiatrist is involved
- atheists do not benefit from therapy in the chapel, but they can trigger it for others

## Why's this needed? 

There aren't very many methods to go about curing addictions, this way does so by encouraging the crew to interact.

The chaplain could do with more things to do, and more reasons for people to visit their church and talk to them. It also gives a bit of a reason to use the psychiatrist's room. 

This is a suggestion from this forum thread: https://forum.ss13.co/showthread.php?tid=20734

## Testing 

https://github.com/user-attachments/assets/055b3206-1f31-4d94-b670-1e1c80caf06f

Move to the chapel or psychiatrist's office. Spawn and consume 1 unit of triple meth. Spam talk and observe the addiction is cured faster than normal. (optionally remove 8 or so points of addiction via Adjust Addictions first so it doesn't take a while, 1 unit of triple meth will produce an 11-point addiction.)

Admittedly I was too lazy to multibox test this, but I did test it with minimums of 1 and 2, and it worked as expected. 

## Changelog 

```changelog
(u)Egregorious
(*)Cure your addiction with the power of therapy, visit the chapel or the psychiatrists office and talk to someone about your problems today. 
(+)The chaplain in their chapel and the psychiatrist in their office are more potent in providing therapy.
(+)Atheists cannot benefit from therapy in the chapel, but they can provide it to others. 
```

### DIFF
--- a/_std/defines/component_defines/component_defines_datum.dm
+++ b/_std/defines/component_defines/component_defines_datum.dm
@@ -28,9 +28,9 @@
 	#define COMSIG_AREA_ACTIVATED "area_activated"
 	/// area's active var set to false (when all clients leave)
 	#define COMSIG_AREA_DEACTIVATED "area_deactivated"
-	/// whenever a mob enters an area
+	/// whenever a mob enters an area (entered mob)
 	#define COMSIG_AREA_ENTERED_BY_MOB "mob_entered_area"
-	/// whenever a mob exits an area
+	/// whenever a mob exits an area (exited mob)
 	#define COMSIG_AREA_EXITED_BY_MOB "mob_exited_area"
 
 // ---- TGUI signals ----


DisturbHerb:
could be cool to have a status effect show up to get more feedback on whether or not therapy is Working

Egregorious:
My hope was that #24823 would provide enough feedback, but there could also be a status effect while in a therapy zone with an addiction.

Glamurio:
I agree on the status effect. In fact, it might be nice to have a status effect on whether or not you're in a therapy zone, something like "You feel at ease".

Egregorious:
Screenshot 2025-10-05 114936
Screenshot 2025-10-05 115037

Added a status for remaining inside a therapy zone, with an attempt at an icon.

code/datums/components/therapy.dm
```diff
@@ -0,0 +1,71 @@
...
+		initialization_args = list(
+		ARG_INFO("specialist_traits", DATA_INPUT_NUM, "A list of trait ids which have their own cooldown for this area", null),
+		ARG_INFO("trait_exclusions", DATA_INPUT_NUM, "A list of taits ids which prevent a mob being affected by therapy in this area", null)
+	)
+
+/datum/component/area_therapy
+	var/area/master
+	/// traits which have their own cooldown for this area, which makes makes them twice as useful for therapy
+	var/list/specialist_traits
+	/// Taits which prevent a mob being affected by therapy in this area, although they can still grant it to others
```
Glamurio: Taits

_std/defines/component_defines/component_defines_datum.dm
```diff
@@ -28,6 +28,10 @@
 	#define COMSIG_AREA_ACTIVATED "area_activated"
 	/// area's active var set to false (when all clients leave)
 	#define COMSIG_AREA_DEACTIVATED "area_deactivated"
+	/// whenever a mob enters an area
+	#define COMSIG_MOB_ENTERED_AREA "mob_entered_area"
+	/// whenever a mob exits an area
+	#define COMSIG_MOB_EXITED_AREA "mob_exited_area"
```
Tarmunora: rename to AREA_MOB_FOOBAR for clarity - as is, these would easily be mistaken for signals meant to be sent to mobs, rather than areas

Egregorious:
I had started out thinking this might be a global controller, and by the time I got this far I'd forgotten components existed, fixed now. 

The status effect should now only appear for mobs which have an active addiction, although gaining/losing addictions while already in a therapy zone isn't handled.

PR #24825, opened 2025-10-03
Mechcomp components plated in materials have an action bar when secured/loosened

M-Earthfire:
[game objects][balance][Materials]

## About the PR 

Mechcomp components plated in material now have an action bar when being tried to loosen/fasten with a wrench.

The duration of the action bar is equal to `Material density * 0.8 Seconds`. This is modifed by different means:
- The duration is halved when the component is fastened
- The duration is halved if the user has engineering/carpenter training
- The duration is doubled if the material is a crystal material, but the component will shatter on being loosened
- The duration will be clamped in an area between 1 and 10 seconds.

## Why's this needed? 

Currently, mechcomp is very frail. Components can be very easily accessed because every action on them is instant and can't be interrupted. For building and testing mechcomp, this is a very desireable trait.

However, when mechcomp interfaces with crew, this frailty often leads to it being very easily dismantled. This resulted into the concern by different people that antag usage is very limited. Once your pressure plate is tapped once with a wrench, your rube goldberg machine falls apart.

This PR enables engineers to make mechcomp much more resilient against tampering by other crewmembers. By plating single objects that interface with crew and hiding the rest, you are able to have more counterplay against people running at your mechcomp with wrenches.

Tying the action bar duration to engineer/carpenter training also allows engineers to be the best option at combating mechcomp by antags. Not only does their training half the duration of the operation, access to magboots also enables them to not be shoved away as easily.

If antags want additional safety against tampering, plating in crystal material allows much higher safety for the tradeoff of destroying your equipment once dismantled.

## Testing 

https://github.com/user-attachments/assets/b45d3fff-be02-40b3-aa98-db27e1d88e54

## Changelog 

```changelog
(u)Lord_Earthfire
(*)Plating mechcomp in materials makes them take much more time to loosen/fasten. Carefull of crystal material, since components plated in these shatter upon being loosened, but are much more durable to tampering.
```

### DIFF
--- a/code/modules/mechanics/MechanicMadness.dm
+++ b/code/modules/mechanics/MechanicMadness.dm
@@ -559,15 +559,7 @@ TYPEINFO(/obj/item/mechanics)
 				work_time = clamp(work_time, 1 SECONDS, 10 SECONDS)
 				//now we set the action bar in motion
 				user.visible_message(SPAN_NOTICE("[user] begins tampering with [src]."))
-				var/datum/action/bar/icon/callback/action_bar = new /datum/action/bar/icon/callback(
-					user,
-					src,
-					work_time,
-					/obj/item/mechanics/proc/wrench_action,
-					\list(user, W),
-					W.icon,
-					W.icon_state,)
-				actions.start(action_bar, user)
+				SETUP_GENERIC_ACTIONBAR(user, src, work_time, PROC_REF(wrench_action), list(user, W), W.icon, W.icon_state, null, INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION | INTERRUPT_MOVE)
 			else
 				src.wrench_action(user)
 			return TRUE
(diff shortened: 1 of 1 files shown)


code/modules/mechanics/MechanicMadness.dm
```diff
@@ -478,60 +543,33 @@ TYPEINFO(/obj/item/mechanics)
...
+				user.visible_message(SPAN_NOTICE("[user] begins tampering with [src]."))
+				var/datum/action/bar/icon/callback/action_bar = new /datum/action/bar/icon/callback(
+					user,
+					src,
+					work_time,
+					/obj/item/mechanics/proc/wrench_action,
+					\list(user, W),
+					W.icon,
+					W.icon_state,)
+				actions.start(action_bar, user)
```
Tarmunora: Any reason not to use the generic setup macro here?

M-Earthfire: Just the fact i forgot it existed. Changed it to use `SETUP_GENERIC_ACTIONBAR` now.

PR #24826, opened 2025-10-03
Adds personal crates to the crate manufacturer

M-Earthfire:
[Feature][Game Objects]

## About the PR 
Adds personal crates, available at the crate manufacturer
secure_crate_greensecure_crate_green_open

These crates work like personal lockers, just as crates

## Why's this needed? 
The transportation of stuff via crates has a lot of possibilities to getting disrupted. While being good, sometimes there is the need for just a little safer form of transportation of e.g. dangerous goods.

Personal lockers offer this kind of security, but they are limited and people tend to touch them on round start, claiming and forgetting about them.

By making crates with the same mechanic printable at the QM, the same safety can be achieved even later in the rounds, although locked behind interaction with a different department.

## Changelog 

```changelog
(u)Lord_Earthfire
(*)The crate manufacturer in cargo offers the creation of personal secure crates now.
```


TobleroneSwordfish:
I'm not sure I like the idea of giving everyone access to perfectly secure transport crates, things getting stolen/released/broken in transport usually adds to the round in a good way. What are you envisioning being transported using these?

M-Earthfire:
Mostly dangerous goods you want to release at a specific moment, e.g. a dozen killer tomatoes. But then maybe crates aren't the best idea?

Messing with things is usually good, but i thought there are some cases where you want to have a proper reveal.

But i see that it will make too much transportation too safe. Maybe make it require specific naterials to produce (e.g. plasmasteel?)

ju45he:
What if I could wack it open, but it takes more damage to break open?

M-Earthfire:
closing this PR because #24935 does fulfill the goals of this PR in a better and more disruptable (good) way.

PR #24835, opened 2025-10-03
introduces the "nocrossing" property and apply it to `/obj/item/dummy`

### LINKED ISSUE #24834: clicking on a component next to a pressure sensor triggers the pressure sensor
### Labels

[minor]

### Description

Pressure sensor components are able to register clicks, although they shouldnt

### Steps to reproduce

1. have a pressure sensor and a sound syntethizer component next to each other
2. connect them
3. multitool the sound synthetizer while standing next to both componnets
4. see the sound synthetizer giving out a signal

### Expected Behavior

The pressure sensor shouldn't pick up interactions that is not stepping on it.

https://github.com/user-attachments/assets/a3f770ca-5d15-4b5d-8d53-bcc24e5b9604

M-Earthfire:
[bug]

## About the PR 

This PR introduces the "nocrossing" atom property. This prevents the atom in question from ever firing `Crossed`.

This propertie is applied to the `/obj/item/dummy` object which is used for the test_click-check.

## Why's this needed? 

This dummy would call Crossed each time things get clicked. And a dummy item should not do that. This means people would need to account for that each time they use crossed(). Which isn't happening, as we can see in #24834. This PR fixes #24834.

## Testing 

This fucks around with a few mechanics, but i hope i tested the main ones

https://github.com/user-attachments/assets/3431211f-e780-4cca-819d-16c2050b50ec

### DIFF
--- a/code/atom.dm
+++ b/code/atom.dm
@@ -540,7 +540,7 @@ TYPEINFO(/atom/movable)
 		if(isturf(src.loc)) // call it on the area too
 			src.loc.loc.Entered(src, null)
 			for(var/atom/A in src.loc)
-				if(A != src)
+				if(A != src && !HAS_ATOM_PROPERTY(src, PROP_ATOM_NOCROSSING))
 					A.Crossed(src)
 
 
@@ -1104,7 +1104,7 @@ TYPEINFO(/atom/movable)
 				covered_turf.pass_unstable -= src.pass_unstable
 				covered_turf.passability_cache = null
 		for(var/atom/A in oldloc)
-			if(A != src)
+			if(A != src && !HAS_ATOM_PROPERTY(src, PROP_ATOM_NOCROSSING))
 				A.Uncrossed(src)
 
 	// area.Exited called if we are on turfs and changing areas or if exiting a turf into a non-turf (just like Move does it internally)
@@ -1119,7 +1119,7 @@ TYPEINFO(/atom/movable)
 				covered_turf.pass_unstable += src.pass_unstable
 				covered_turf.passability_cache = null
 		for(var/atom/A in newloc)
-			if(A != src)
+			if(A != src && !HAS_ATOM_PROPERTY(src, PROP_ATOM_NOCROSSING))
 				A.Crossed(src)
 
 	// area.Entered called if we are on turfs and changing areas or if entering a turf from a non-turf (just like Move does it internally)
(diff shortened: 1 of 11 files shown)


Azrun:
Is there applicability to `atom/Crossed` and `atom/Uncrossed`?

code/atom.dm
```diff
@@ -1104,7 +1104,7 @@ TYPEINFO(/atom/movable)
 				covered_turf.pass_unstable -= src.pass_unstable
 				covered_turf.passability_cache = null
 		for(var/atom/A in oldloc)
-			if(A != src)
+			if(A != src && !HAS_ATOM_PROPERTY(src, PROP_ATOM_NOCROSSING))
```
Azrun, line 1107: Modifying uncrossed is not documented in PR description.

M-Earthfire: good point, it likely should not be affected by this property but by a differnent one!

Although here the use case would be the same. To prevent dummies from firing it.

M-Earthfire:
> Is there applicability to `atom/Crossed` and `atom/Uncrossed`?

not exatcly what you mean by applicability. If the question is on why they need to be on atom, they are used for e.g. holes to the trench or others (turfs), items (pressure plate) and others. Pressure plates trigger on items, bullets and mobs. That is nice for shooting ranges or other gimmick mechcomp use.

If you mean applicability of this property... well, i think there are more dummy-objects/mobs (iirc there was one in the mechcomp graber PR that got closed) and i think all these should not interact with mechanics that care about actually existing objects.

Azrun:
https://github.com/M-Earthfire/goonstation/blob/nocrossing_dummy/code%2Fatom.dm#L346-L368

M-Earthfire:
to be honest, doing the check there would be smart... if i would be smart i would have probably done it...

Tarmunora:
How does this interact with BYOND builtins that may call `Crossed()` (i.e.: `Move()`)?

M-Earthfire:
> How does this interact with BYOND builtins that may call `Crossed()` (i.e.: `Move()`)?

oh...

Actually, it doesn't interact at all with that. I completely missed the idea that there are instances outside this codebase that call these procs. Right now, this only goes through instances where we call it.

This means i gonna close it and need to go back to the drawing board. That's a thing i need to implement with more care.

PR #24841, opened 2025-10-04
Gives Chemical Barrels Help Messages

Mister-Moriarty:
## About the PR:
Chemical barrels have given a help message that appears on-examine:
> Click the barrel with an **empty hand** to flip the barrel's funnel into a spout or vice versa. Use a **reagent container** to add/remove reagents from the barrel, depending on the funnel/spout. Use a **pen** to add a label to the barrel. Use a **wrench** to open or close the barrel's lid. Use a **synthetic pustule** on the barrel to squeeze its contents into the barrel.

Notably does not include the barrel's interaction with wood, which I thought would be best to keep hidden.

## Why Is This Needed?
It is now easier to see a barrel's various interactions at a glance.

## Testing:

### DIFF
--- a/code/modules/chemistry/tools/dispensers.dm
+++ b/code/modules/chemistry/tools/dispensers.dm
@@ -432,7 +432,7 @@ TYPEINFO(/obj/reagent_dispensers/watertank/fountain)
 
 /obj/reagent_dispensers/chemicalbarrel
 	name = "chemical barrel"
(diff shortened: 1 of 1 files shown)


Tarmunora:
Might be more parsable with newlines between entries

Mister-Moriarty:
> Might be more parsable with newlines between entries

I considered this, but felt that that made the help text too long, especially for an object players will be examining frequently with spectroscopic goggles.

PR #24848, opened 2025-10-05
Deconstructable safes

Kelguroth:
## About the PR 
This PR allows players to deconstruct secure safes, so long as they are unlocked. This change also affects secure briefcases, but as you can scan and print frames of them this seems like acceptable behavior.
## Why's this needed? 
The staff quarters on Kondaru could be refurbished into new areas, but one major problem is that the 17(!) safes located in the area will stick out like sore thumbs. There are likely other maps this will come into play on, but Kondaru was my main focus. Allowing the safes to be deconstructed when unlocked allows for more customization. 
## Testing 
Deconstructed several pre-placed safes located on the station, along with a player built safe. Deconstructed a door, speaker, fabricator, several vending machines to ensure existing deconstructable items function as intended. Tested safes located in Mars azone, they were not able to be deconstructed both locked and unlocked.

## Changelog 

```changelog
(u)Kelguroth
(*) Added ability to deconstruct secure containers so long as they are unlocked.
```

### DIFF
--- a/code/datums/components/tool/deconstructing.dm
+++ b/code/datums/components/tool/deconstructing.dm
@@ -59,6 +59,11 @@ TYPEINFO(/datum/component/deconstructing)
 			var/decon_time = (decon_complexity * 2.5 SECONDS * src.complexity_mult) + src.base_duration
 			actions.start(new/datum/action/bar/icon/deconstruct_obj(target,decon_tool,decon_time), user)
 		else
+			if (istype(O, /obj/item/storage/secure/ssafe)) //checks if secure safes are unlocked before attempting to deconstruct them.
+				var/obj/item/storage/secure/Oa = target
+				if(Oa.locked)
+					boutput(user, SPAN_ALERT("You cannot deconstruct [target] while it is locked."))
+					return ATTACK_PRE_DONT_ATTACK
 			user.showContextActions(O.decon_contexts, O, src.contextLayout)
 			boutput(user, SPAN_ALERT("You need to use some tools on [target] before it can be deconstructed."))
 		return ATTACK_PRE_DONT_ATTACK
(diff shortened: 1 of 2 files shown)


M-Earthfire:
What happens with the safes contents upon deconstruction?

Kelguroth:
> What happens with the safes contents upon deconstruction?

They remain inside of the object and if rebuilt can be accessed again. This is currently consistent behavior that matches other deconstructable storage items, such a toilets, washing machines, and objects that use power cells.
image

code/obj/item/storage/secure_safe.dm
```diff
@@ -28,6 +28,8 @@ ABSTRACT_TYPE(/obj/item/storage/secure)
 	w_class = W_CLASS_NORMAL
 	burn_possible = FALSE
 	var/random_code = FALSE // sets things to already have a randomized code on spawning
+	//Enables deconstruction of safes
```
frawhst:
```suggestion
```

code/datums/components/tool/deconstructing.dm
```diff
@@ -59,6 +59,11 @@ TYPEINFO(/datum/component/deconstructing)
 			var/decon_time = (decon_complexity * 2.5 SECONDS * src.complexity_mult) + src.base_duration
 			actions.start(new/datum/action/bar/icon/deconstruct_obj(target,decon_tool,decon_time), user)
 		else
+			if (istype(O, /obj/item/storage/secure/ssafe)) //checks if secure safes are unlocked before attempting to deconstruct them.
+				var/obj/item/storage/secure/Oa = target
+				if(Oa.locked)
+					boutput(user, SPAN_ALERT("You cannot deconstruct [target] while it is locked."))
+					return ATTACK_PRE_DONT_ATTACK
```
frawhst:
```suggestion
				var/obj/item/storage/secure/safe = target
				if (safe.locked)
					boutput(user, SPAN_ALERT("You cannot deconstruct [target] while it is locked."))
					return ATTACK_PRE_DONT_ATTACK
```

PR #24854, opened 2025-10-05
Make Uplinks Detomatix proof

JORJ949:
## About the PR 
Makes any PDA with an uplink immune to PDA bombs.

## Why's this needed? 
Having all of your antag gear removed as a traitor or spy thief because someone decided to randomly pda bomb you doesn't feel particularly good, even if the attack was targeted you shouldn't be able to remotely destroy an uplink with no way of the victim knowing who did it. I don't see this being particularly metagameable because they could also just have their messenger off, have been signal jammed, or gotten a new pda between you bombing them and rescanning for PDAs.

## Testing 
image

## Changelog 

```changelog
(u)JORJ949
(*)PDAs containing uplinks are now immune to detomatix bombings.
```

### DIFF
--- a/code/obj/item/device/pda2/pda2.dm
+++ b/code/obj/item/device/pda2/pda2.dm
@@ -1217,7 +1217,7 @@
 		return 1
 
 	proc/explode()
-		if (src.bombproof)
+		if (src.bombproof || src.uplink)
 			if (ismob(src.loc))
 				boutput(src.loc, SPAN_ALERT("<b>ALERT:</b> An attempt to run malicious explosive code on your PDA has been blocked."))
 			return


kbsmilk:
Alternatively, it could be kind of funny if the PDA still exploded, but dropped a standalone uplink on the floor or whatever.

Or, the PDA still explodes, and immediately spawns a dead drop for the victim, containing an uplink with their remaining bux.

JORJ949:
> Alternatively, it could be kind of funny if the PDA still exploded, but dropped a standalone uplink on the floor or whatever.
> 
> Or, the PDA still explodes, and immediately spawns a dead drop for the victim, containing an uplink with their remaining bux.

PDAs aren't currently built to be moved around so this would require a lot more work (preferably componentisation)

PR #24879, opened 2025-10-08
Fingerprint Rewrite

LorrMaster:
[security][rework][input wanted]

## About the PR
- Fingerprints are now applied based on the arms themselves rather than being part of the mob.
- If the left and right arms have different fingerprints, the print will be applied based on the active hand.
- Human limbs use lowercase letters, while robotic limbs use hexadecimal fingerprints.
- Gloves now leave behind a partial fingerprint mask. Silver nitrate can be used to reveal partial fingerprints through gloves. 
- Black & janitor gloves leave behind no partial prints. Most gloves will have a fingerprint mask with a 25% match rate. Insulated gloves have a 12.5% match rate. Latex gloves have a 6% match rate. Custom gloves will depend on the chemical resistance of the material. (Higher match rate is better for crime)
- Partial prints can be inputted into the forensics scanner for a list of potential matches from security records
- A vial of silver nitrate can be found in forensics carts at round start.
- Fingerprints are no longer used for admin logs. Admins can now use a separate adminprints system that stores user's ckeys to obtain a chain of custody.
- Fingerprint report text is now more condensed by using multiple columns.

## Why's this needed?
- Updates fingerprints to the new forensics system.
- Fingerprint system is now much more detailed. 
- Admins now have their own type of forensics for more customized data collection.

## Testing
Screenshot 2025-10-12 044525

Screenshot 2025-10-12 045035

Screenshot 2025-10-12 040411

## Changelog
```changelog
(u)LorrMaster
(*)Fingerprints have been rewritten and are now based on the current arm(s) instead of the mob themselves.
(*)Partial fingerprints can be revealed using silver nitrate, varies depending on the type of glove.
(+)A bottle of silver nitrate can be found in forensics carts.
```

### DIFF
--- a/code/modules/admin/admin_holder.dm
+++ b/code/modules/admin/admin_holder.dm
@@ -156,7 +156,7 @@
 			"Add Reagents",\
 			"Check Reagents",\
 			"View Variables",\
-			"View Fingerprints",\
+			"View Admin Forensics",\
 			"Delete",\
 			"Possess",\
 			"Create Poster",\
@@ -211,7 +211,7 @@
 			"Get Telesci Coords",\
 
 			"View Variables",\
-			"View Fingerprints",\
+			"View Admin Forensics",\
 			"Delete",\
 			"Create Poster",\
 			"Set Material",\


code/modules/networks/computer3/sec_rec.dm
```diff
@@ -746,23 +754,24 @@
...
 				view_string += {"<br>\
 					<br><center><b>Security Data</b></center>\
-					<br>\[09\]<b>Criminal Status:</b> [src.active_secure["criminal"]]\
-					<br>\[10\]<b>SecHUD Flag:</b>     [src.active_secure["sec_flag"]]\
-					<br>\[11\]<b>Minor Crimes:</b>    [src.active_secure["mi_crim"]]\
-					<br>\[12\]<b>Details:</b>         [src.active_secure["mi_crim_d"]]\
-					<br>\[13\]<b>Major Crimes:</b>    [src.active_secure["ma_crim"]]\
-					<br>\[14\]<b>Details:</b>         [src.active_secure["ma_crim_d"]]\
-					<br>\[15\]<b>Important Notes:</b> [src.active_secure["notes"]]\
+					<br>\[10\]<b>Criminal Status:</b> [src.active_secure["criminal"]]\
```
JORJ949: Change these arbitrary numbers to be the FIELDNUM defines above

code/modules/networks/computer3/sec_rec.dm
```diff
@@ -16,15 +16,16 @@
...
-#define FIELDNUM_SECFLAG 10
-#define FIELDNUM_MINCRIM 11
-#define FIELDNUM_MINDET 12
-#define FIELDNUM_MAJCRIM 13
-#define FIELDNUM_MAJDET 14
-#define FIELDNUM_NOTES 15
+#define FIELDNUM_PRINT_R 7
+#define FIELDNUM_PRINT_L 8
+#define FIELDNUM_PHOTO 9
+#define FIELDNUM_CRIMSTAT 10
```
JORJ949, line 22: Is there any way you think you could get away with not changing the arrest status number? Its pretty common memory for security to login>secmate>2>[NAME]>9 to pull up the arrest menu, but this isn't that big if you don't think its possible.

code/obj/item/cigarette.dm
```diff
@@ -355,7 +355,8 @@
 		if (ismob(src.loc))
 			logTheThing(LOG_BOMBING, null, "A trick cigarette (held/equipped by [constructTarget(src.loc,"bombing")]) explodes at [log_loc(src)].")
 		else
-			logTheThing(LOG_BOMBING, src.fingerprintslast, "A trick cigarette explodes at [log_loc(src)]. Last touched by [src.fingerprintslast ? "[src.fingerprintslast]" : "*null*"].")
+			var/last_ckey = src.get_last_ckey()
+			logTheThing(LOG_BOMBING, last_ckey, "A trick cigarette explodes at [log_loc(src)]. Last touched by [last_ckey].")
```
JORJ949: Last touched by's should use key_name so it shows character name and antag status in logs. For this specific instance though its probably not needed due to you passing last_ckey in for the person doing the thing.

code/obj/item/clothing/gloves.dm
```diff
@@ -221,38 +254,47 @@ ABSTRACT_TYPE(/obj/item/clothing/gloves)
...
 	item_state = "finger-"
-	hide_prints = 0
+	material_prints = "black leather"
 
 	setupProperties()
 		..()
 		setProperty("conductivity", 1)
 
+	get_fiber_mask()
+		return "0123-4567-89AB-CDEF"
```
JORJ949: Make a define for these copied fiber_mask strings (the 0123 and ??? ones) with descriptive names, I can't tell just from reading this code what they mean.

code/modules/networks/computer3/sec_rec.dm
LorrMaster: I could switch FIELDNUM_PHOTO and FIELDNUM_CRIMSTAT so that arrest status stays as the same number. It would only affect that specific value though.

code/obj/item/cigarette.dm
LorrMaster: I think I want to avoid making changes to the admin logs unless if the PR is directly affecting how they behave. Goal being a 1:1 conversion. So if key_name wasn't used before I want to avoid adding it in to this PR if it is not needed.

code/modules/networks/computer3/sec_rec.dm
LorrMaster: Actually that messes with their placement relative to the Security Data section, so I guess it's staying as 10 unless if there is another solution.

Sbmhawk:
Looking at the latex glove mask, this seems to make it much harder to find the prints through it, as previously there was usually enough of the prints to be able to search via secmate. 

This seems like much too harsh a nerf for those partial fingerprints, especially now needing a chemical ontop of that.

_std/defines/forensics.dm
```diff
@@ -31,3 +35,10 @@
 #define FORENSIC_VALUE_MULT 3
 #define FORENSIC_VALUE_MAX 4
 #define FORENSIC_VALUE_MIN 5
+
+// #define FORENSIC_CHARS_FINGERPRINTS list("G","H","J","K","N","O","P","Q","R","S","U","V","W","X","Y","Z")
```
frawhst:
```suggestion
```

_std/record_db.dm
```diff
@@ -89,6 +89,216 @@
...
+				record_prints.Add(fprint_right)
+				record_refs.Add(record)
+			else
+				if(fprint_right)
+					record_prints.Add(fprint_right)
+					record_refs.Add(record)
+				if(fprint_left)
+					record_prints.Add(fprint_left)
+					record_refs.Add(record)
+		if(!record_prints || !record_refs || length(record_prints) != length(record_refs))
```
frawhst:
```suggestion
		if(!record_prints || !record_refs)
```
I'm *fairly* certain the lengths can never be different with the current code (even in the case of runtimes), and if we want to check for it then it should definitely throw an error since it's so unexpected. Either is fine.

_std/record_db.dm
```diff
@@ -89,6 +89,216 @@
...
+			return null
+
+		// Get the input into a more usable format
+		// Print: (0123-4567-89AB-CDEF) => Bunches: list("0123","4567","89AB","CDEF")
```
frawhst: I'd prefer if the following section (parsing the fingerprints into chunks and trimming empty chunks) was split into its own proc if possible, since this proc is pretty lengthy as-is and we might want this functionality for something else down the line.

_std/record_db.dm
```diff
@@ -89,6 +89,216 @@
...
+				// Remove empty edge bunches at end of print
+				for(var/k in length(bunch_is_empty) to 1 step -1)
+					var/bunch_length = length(rec_bunches)
+					if(!bunch_is_empty[k] || bunch_length == 0)
+						break
+					rec_bunches.Cut(bunch_length)
+				// Remove empty edge bunches at start of print
+				for(var/k in 1 to length(bunch_is_empty))
+					if(!bunch_is_empty[k] || length(rec_bunches) == 0)
+						break
```
frawhst: should either use a length var or just the raw `length()` call for both of these, odd to have differing implementations right next to each other

code/WorkInProgress/explosion_particles.dm
```diff
@@ -75,7 +75,7 @@
 
 /datum/effects/system/explosion/proc/start()
 	var/obj/effects/explosion/E = new/obj/effects/explosion( location )
-	E.fingerprintslast = source.fingerprintslast
+	source.forensic_holder.copy_to(E.forensic_holder, null, TRUE)
```
frawhst, line 78: What's this 3rd arg being set to `TRUE`? I don't see a 3rd arg in any of the implementations of `copy_to` in `forensics.dm`. (ditto for the rest of the `copy_to()` calls with a 3rd arg)

code/modules/chemistry/Chemistry-Holder.dm
```diff
@@ -1218,20 +1218,20 @@ proc/chem_helmet_check(mob/living/carbon/human/H, var/what_liquid="hot")
...
 			else
 				our_user = usr
-				if (my_atom.fingerprintslast) // Our container. You don't necessarily have to pick it up to transfer stuff.
-					our_fingerprints = my_atom.fingerprintslast
-				else if (my_atom.loc.fingerprintslast) // Backpacks etc.
-					our_fingerprints = my_atom.loc.fingerprintslast
+				if (my_atom.get_last_ckey()) // Our container. You don't necessarily have to pick it up to transfer stuff.
+					last_ckey = my_atom.get_last_ckey()
+				else if (my_atom.loc.get_last_ckey()) // Backpacks etc.
+					last_ckey = my_atom.loc.get_last_ckey()
```
frawhst, line 1226: These checks won't work as intended, since `get_last_ckey()` returns `"None"` if there's no admin prints, rather than `null`, and `"None"` is truthy.

code/modules/chemistry/Reagents-ExplosiveFire.dm
```diff
@@ -593,9 +593,10 @@ datum
 								if(holder.my_atom)
 									holder.my_atom.visible_message(SPAN_ALERT("<b>[holder.my_atom] explodes!</b>"))
 									// Added log entries (Convair880).
-									if(holder.my_atom.fingerprintslast || usr?.last_ckey)
-										message_admins("Welding Fuel explosion (inside [holder.my_atom], reagent type: [id]) at [log_loc(holder.my_atom)]. Last touched by: [holder.my_atom.fingerprintslast ? "[key_name(holder.my_atom.fingerprintslast)]" : "*null*"] (usr: [ismob(usr) ? key_name(usr) : usr]).")
-									logTheThing(LOG_BOMBING, holder.my_atom.fingerprintslast, "Welding Fuel explosion (inside [holder.my_atom], reagent type: [id]) at [log_loc(holder.my_atom)]. Last touched by: [holder.my_atom.fingerprintslast ? "[key_name(holder.my_atom.fingerprintslast)]" : "*null*"] (usr: [ismob(usr) ? key_name(usr) : usr]).")
+									var/last_ckey_atom = holder.my_atom.get_last_ckey()
+									if(last_ckey_atom || usr?.last_ckey)
```
frawhst, line 597: similar issue to the above

code/modules/cooking/deep_fryer.dm
```diff
@@ -154,7 +154,7 @@ TYPEINFO(/obj/machinery/deep_fryer)
 			src.say("Oh, now I can die a warrior's death! Thank you!")
 			ADD_FLAG(src.status, BROKEN)
 			name = "Satiated [initial(src.name)]"
-			ice_feeder = ice_feeder || ckey_to_mob(src.fryitem.fingerprintslast) // in case someone else had to fufill, no direct ref
+			ice_feeder = ice_feeder || ckey_to_mob(src.fryitem.get_last_ckey()) // in case someone else had to fufill, no direct ref
```
frawhst, line 162: similar issue

code/modules/forensics/atom_forensic.dm
```diff
@@ -39,6 +36,20 @@
...
+		return "<b>No admin forensics found on [src].</b>"
+	return "<b>Hidden Adminprints on [src]:</b>" + aprints_group.get_adminprints()
+
+/atom/proc/get_last_ckey(var/include_time = FALSE)
+	var/datum/forensic_group/adminprints/aprints_group = src.forensic_holder?.get_group(FORENSIC_GROUP_ADMINPRINTS, TRUE)
+	if(istype(aprints_group) && aprints_group.last_print)
+		if(include_time)
+			return aprints_group.last_print.get_text()
+		return aprints_group.last_print.clientKey.id
+	return "None"
```
frawhst: I'm concerned about this returning `"None"` due to the aforementioned issues, and the fact that people will Probably run into them again in the future.
Personally I would assume that if I call a proc which gets me a ckey that may or may not be there, and the ckey isn't there, then the return from that function will be `FALSE` or `null` or similar. It's definitely more convenient for logging and UI where you always want to display SOMETHING, but for operations that aren't worried about displaying the text, having to manually check `== "None"` in a bunch of places feels janky.
I'd prefer if this proc returned `null` if it couldn't find a ckey, and then you can add a helper proc which just calls this one and returns `"None"` if `get_last_ckey()` returns `null`. Could call it `get_last_ckey_or_none`, or `display_last_ckey` or something.

code/WorkInProgress/explosion_particles.dm
LorrMaster: There were 3 args, but one of them wasn't appearing for some reason.

code/datums/manufacturing.dm
```diff
@@ -1537,6 +1537,16 @@
...
+	var/datum/forensic_id/roboprint = null // Give robo arms the same fingerprints
+
+	modify_output(var/obj/machinery/manufacturer/M, var/atom/A, var/list/materials)
+		..()
+		if(istype(A, /obj/item/parts/robot_parts/arm))
+			var/obj/item/parts/robot_parts/arm/new_arm = A
+			if(roboprint)
+				new_arm.limb_print = roboprint
+			else
+				roboprint = new_arm.limb_print
```
frawhst: Because blueprints are singletons, this makes all robot parts printed with this recipe have the same fingerprints, even across different fabs. IMO it's easiest to just let the fingerprints be different for the two cyborg arms, since printing the parts individually will have that behavior anyways.

code/datums/manufacturing.dm
```diff
@@ -1548,6 +1558,16 @@
...
```
frawhst: Same for this.

code/modules/admin/admin.dm
```diff
@@ -3426,14 +3426,14 @@ var/global/noir = 0
...
 						for(var/mob/living/carbon/human/H in mobs)
 							if(H.ckey)
-								if(H.bioHolder.Uid)
-									dat += "<tr><td>[H]</td><td>[H.bioHolder.fingerprints]</td></tr>"
-								else if(!H.bioHolder.Uid)
-									dat += "<tr><td>[H]</td><td>H.bioHolder.Uid = null</td></tr>"
+								if(H.bioHolder.default_fingerprints)
+									dat += "<tr><td>[H]</td><td>[H.bioHolder.default_fingerprints.id]</td></tr>"
+								else if(!H.bioHolder.default_fingerprints)
+									dat += "<tr><td>[H]</td><td>H.bioHolder.default_fingerprints = null</td></tr>"
```
frawhst:
```suggestion
								if(H.bioHolder.default_fingerprints)
									dat += "<tr><td>[H]</td><td>[H.bioHolder.default_fingerprints.id]</td></tr>"
								else
									dat += "<tr><td>[H]</td><td>H.bioHolder.default_fingerprints = null</td></tr>"
```
this logic was silly before your changes but, might as well make it cleaner

code/modules/forensics/forensic_data.dm
```diff
@@ -64,3 +69,96 @@
...
+		else if(src.print_mask?.id == FORENSIC_GLOVE_MASK_FINGERLESS)
+			return src.print?.id
+		else if(src.fibers && startswith(src.fibers.id, "latex rubber"))
+			return "([get_masked_print()])"
+
+			// Not including the fibers for now because they were taking up too much room in the forensic report.
+			//var/fprint_text = get_masked_print()
+			// if(fprint_text && fibers_text)
+				// return "[fprint_text] ([fibers_text])"
+			// return fprint_text + fibers_text
```
frawhst:
```suggestion
```
Either uncomment the code or just remove it, we try to avoid leaving blocks of code in comments

code/modules/forensics/forensic_id.dm
```diff
@@ -25,6 +25,36 @@ var/global/list/datum/forensic_id/registered_id_list = new()
...
+		char_list[index] = char_list[char_length]
+		char_length--
+	return new_id
+
+/proc/build_id_separate(var/text, var/bunch_size, var/separation_text = "-")
+	var/final_text = copytext(text, 1, bunch_size + 1)
+	var/bunch_count = floor(length(text) / bunch_size)
+	for(var/i=1; i<= bunch_count - 1; i++)
+		var/pos = (i * bunch_size) + 1
+		final_text += separation_text + copytext(text, pos, pos + bunch_size)
```
frawhst: this should use `for-in-range`, slightly more performant and codebase standard
```suggestion
	for(var/i in 1 to (bunch_count - 1))
		var/pos = (i * bunch_size) + 1
		final_text += separation_text + copytext(text, pos, pos + bunch_size)
```

_std/defines/forensics.dm
```diff
@@ -31,3 +35,9 @@
 #define FORENSIC_VALUE_MULT 3
 #define FORENSIC_VALUE_MAX 4
 #define FORENSIC_VALUE_MIN 5
+
+#define FORENSIC_CHARS_FINGERPRINTS list("a","b","c","d","e","g","n","o","p","q","s","u","v","x","y","z")
```
ZeWaka, line 40: why these chars

_std/lists.dm
```diff
@@ -386,3 +386,18 @@ proc/params2complexlist(params)
...
+
+/// Cuts elements from the start and end of the list.
+#define trim_list(L, start_amt, end_amt) \
+	do { \
+	L.Cut(length(L) - end_amt + 1); \
+	L.Cut(1, start_amt + 1) \
+	} while (0)
+
+/// Remove every matching value from the list
+#define remove_all_list(L, x) \
```
ZeWaka: list.Remove()?

code/procs/helpers.dm
```diff
@@ -2304,6 +2304,12 @@ proc/radioGarbleText(var/message, var/per_letter_corruption_chance=40)
 		. += corrupted_bit
 	return jointext(.,"")
 
+/// Replace a value with another if the original is null. Otherwise returns the original.
+proc/replace_if_null(var/original, var/replacement)
+	if(original)
```
ZeWaka, line 2309: this tests truthyness not nullyness

_std/lists.dm
LorrMaster: list.Remove() only removes one element. `remove_all_list` Removes all matches at once.

_std/defines/forensics.dm
LorrMaster: Just picked out 16 letters that look sort of fingerprinty with curves and such.

_std/lists.dm
LorrMaster: Nevermind, somehow missed list.RemoveAll()

pgmzeta:
!merge_upstream

pgmzeta:
!merge_upstream

pgmzeta:
runtime on live, looks like it's happening with virtual people too

```proc name: add evidence (/datum/forensic_holder/proc/add_evidence)
  source file: code/modules/forensics/forensic_holder.dm,23
  usr: null
  src: /datum/forensic_holder (/datum/forensic_holder)
  call stack:
/datum/forensic_holder (/datum/forensic_holder): add evidence(null, 5, 0)
the right cat arm (/obj/item/parts/human_parts/arm/mutant/cat/bingus/right): add fingerprint(XXX (/mob/living/carbon/human/clone), 0)
the right cat arm (/obj/item/parts/human_parts/arm/mutant/cat/bingus/right): New(XXX (/mob/living/carbon/human/clone), null)
the right cat arm (/obj/item/parts/human_parts/arm/mutant/cat/bingus/right): New(XXX (/mob/living/carbon/human/clone))
/datum/human_limbs (/datum/human_limbs): replace with("r_arm", /obj/item/parts/human_parts/ar... (/obj/item/parts/human_parts/arm/mutant/cat/bingus/right), null, 0, 1)
bingus (/datum/mutantrace/cat/bingus): LimbSetter(XXX (/mob/living/carbon/human/clone), "set")
bingus (/datum/mutantrace/cat/bingus): on attach(XXX (/mob/living/carbon/human/clone))
bingus (/datum/mutantrace/cat/bingus): on attach(XXX (/mob/living/carbon/human/clone))
XXX (/mob/living/carbon/human/clone): set mutantrace(/datum/mutantrace/cat/bingus (/datum/mutantrace/cat/bingus))
Unexpected Genetic Development (/datum/cloner_defect/random_mutantrace): on add()
...
XXX (/mob/living/carbon/human/clone): set loc(the floor (149,188,1) (/turf/simulated/floor/white))
XXX (/mob/living/carbon/human/clone): set loc(the floor (149,188,1) (/turf/simulated/floor/white))
XXX (/mob/living/carbon/human/clone): set loc(the floor (149,188,1) (/turf/simulated/floor/white))
XXX (/mob/living/carbon/human/clone): set loc(the floor (149,188,1) (/turf/simulated/floor/white), 0, 0)
XXX (/mob/living/carbon/human/clone): set loc(the floor (149,188,1) (/turf/simulated/floor/white))
the cloning pod (/obj/machinery/clonepod): go out(1)
the cloning pod (/obj/machinery/clonepod): process(1)
Machine (/datum/controller/process/machines): doWork()
/datum/controller/processSched... (/datum/controller/processScheduler): runProcess(Machine (/datum/controller/process/machines), 1.2)```

pgmzeta:
haven't dug into this yet but "View Fingerprints" in admin menu doesn't appear to be showing anything

pgmzeta (changes requested):
@LorrMaster dug in some more and pushed up a merge resolution and patch to make adminprints appear in the admin interact menu. 

Manufacturers and cyborg docking stations aren't getting adminprints in local testing; this is a blocking change from current. Once that is resolved, let me know and we can look at testmerging again.

LorrMaster:
> @LorrMaster dug in some more and pushed up a merge resolution and patch to make adminprints appear in the admin interact menu.
> 
> Manufacturers and cyborg docking stations aren't getting adminprints in local testing; this is a blocking change from current. Once that is resolved, let me know and we can look at testmerging again.

I'll need to check, but that may actually be a preexisting issue. The PR changes fingerprints themselves, but it doesn't change how they are applied. So issues where fingerprints are inconsistently applied also exist in this PR. The reason is that a number of objects override procs such as attackby(), which skips the fingerprint application process. I haven't spent the time to come up with a solution for that yet.

PR #24881, opened 2025-10-09
Adds 26 hairstyles

ribbiestar:
## About the PR 
Adds 26 more hairstyles. (8 base ones.)

Screenshot 2025-10-09 130753
Screenshot 2025-10-09 131733

From top to bottom:
1- Box Braids
2- Bubble Braids
3- Long Dreads
4- Messy Wicks
5- Neat Wicks
6- Puff Mohawk
7- Puffball
8- Afro: Sidepart 

## Why's this needed? 
hair cool wauw

## Testing 
Tested on local by randomly spawning NPCs and checking if they spawned with the hair.
Customization options checked in Character Setup to see if the hairs appear on the hair list.
All hairs tested on local.

## Changelog 

```changelog
(u)Ghostplasmas & Ribbie
(*)Added 26 more hairstyles.
```


SunWuDrakios:
oh really like these, there weren't many hairstyles (at least I think so) that come from the black community, so seeing some more then just afro is nice!

Ghostplasmas:
> I really like these, especially the braids. My only note is that the split afro and puffball styles look a bit flat. I know we usually say not to pillow shade things but I think hair is a good exception to that. Regular afro vs split for reference:
> 
> Image Image

TYSM FOR POINTING IT OUT! I was struggling with the afro and puffball a lot and something kept looking off about them but i couldn't figure out what it was. It was the shading!!! I just finished remaking them + their alternatives!!

ribbiestar:
New versions:

image

TobleroneSwordfish (approved):
<3

PR #24885, opened 2025-10-09
Pod part separate image appearance

LorrMaster:
## About the PR
- Images of pod engines, weapons, and minisub tracks now inherit their colors and filters from the part rather than the vehicle itself.

## Why's this needed?
- More pod appearance customization

## Testing
Screenshot 2025-10-09 170454

### DIFF
--- a/code/modules/transport/pods/MainWeapon.dm
+++ b/code/modules/transport/pods/MainWeapon.dm
@@ -41,7 +41,12 @@
 	ship_install()
 		..()
 		if(src.ship.uses_weapon_overlays && src.appearanceString)
-			src.ship.UpdateOverlays(image('icons/effects/64x64.dmi', "[src.appearanceString]"), "mainweapon")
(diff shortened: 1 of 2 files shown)


flappybatpal:
I'm not sure if we want this considering parts are supposed to be visually identifiable and pods have skins for customization?

LorrMaster:
> I'm not sure if we want this considering parts are supposed to be visually identifiable and pods have skins for customization?

Well pods being able to have materials has been the case for a while. Engine type being visible on the pod is a pretty recent change. By customization I mean more the feeling that you're putting together your own car with different kinds of special parts and such.

LorrMaster:
> Hmm. I wish it was possible to do like a 50% of the effect. Hard to do with filters though.

I could have it tone down the colors or alter the engine sprite, but I also think that the examples shown in the image are pretty extreme color-contrast cases as examples. I don't think too many people will be mixing iridium armor with refined honey engines.

ZeWaka:
We can revert this if it's unwanted

PR #24897, opened 2025-10-10
Fix respawn event syndie jobs losing antag status on spawn

JORJ949:
## About the PR 
Fixes syndicate specialist jobs having their special antag status removed immediately by the ghost spawn panel by placing the antag status behind a semi-evil SPAWN(2) to go after the ghost spawn panel's SPAWN(1) antagonist role wipe.

## Why's this needed? 
The syndie jobs should always have their antag role regardless of how they're spawned, they're blatantly clear syndies.

## Testing 
image


ZeWaka (commented):
evil

PR #24909, opened 2025-10-12
Fix DWAINE grep interpretation of records with associative fields

ebozc:
## Labels
[BUG - MINOR] [FIX] [STATION SYSTEMS] 

## About the PR 
Fix DWAINE 'grep' readout of /mnt/radio/1xxx/ files not having equals signs.

## Why's this needed? 
`cat`-ing /mnt/radio/ REC files shows key=value pairs, in line with the rest of the packet system. Grep's parsing of REC files has a  parsing bug that omits the `=` from the search string. `message=Hello!` would appear as  `messageHello!` when `grep`'d. This makes parsing or searching /mnt/radio files from DWAINE nigh impossible.

This also effects other REC files that use associative lists internally, such as guardbuddy dock /mnt/xxxx/status files.

## Testing 

Functioning with normal REC files (as before patch)
image

Functioning with /mnt/radio/ files (`=` is now present in the search string)
image


ZeWaka:
@Mister-Moriarty

Mister-Moriarty:
This is the wrong way to go about doing this; you have completely removed the loop over the list's indices, which in BYOND are distinct from keys in associative ordinary lists: these lists are structured as follows: `index:key:value`. This can be seen in the fact that every line is labelled as the first line. `cat` is also distinct from `grep` in that it uses files' `asText()` proc, but I don't think that's applicable here due to `asText()` returning the data for an entire file rather than a single line.

I've appended how I would implement this below, as I'm not able to request changes on deleted lines.
<details>
<summary>DM Code</summary>

```
for (var/j in 1 to length(record.fields))
	var/field_name = record.fields[j]
	var/field_data = record.fields[field_name]
	var/field = (!isnull(field_data)) ? "[field_name]=[field_data]" : field_name

	if (regex.Find("[field]"))
		if (print_only_match)
			grep_results += "[regex.match]"
		else if (plain)
			grep_results += "[field]"
		else
			grep_results += "[record.name]:[j]:[field]"
```
</details>

ebozc:
My initial PR'd code was based off of `asText()`'s loop to make it as consistent with `cat` as possible. I'll admit the `j` counter was hacky.

Updated to use your suggested code.

code/modules/networks/computer3/mainframe2/utilities/grep.dm
```diff
@@ -72,16 +72,19 @@
 
 		else if (istype(target, /datum/computer/file/record))
 			var/datum/computer/file/record/record = target
-
+
```
Mister-Moriarty: Some whitespace weirdness, but other than that, it should be good.

Issue #14236, opened 2023-05-25
Fancy pens do not use cursive


Sonnzer:
### Labels

[BUG] [MINOR]

### Description

Fancy pens do not use the fancy cursive font!

### Steps to reproduce

1. Get fancy pen
2. Use fancy pen to write
3. No cursive!

### Expected Behavior

For it to be cursive.

### Additional Information & Screenshots

This goes for pencils too, as they also write in cursive!

PR #24913, opened 2025-10-12
Paper Fonts Fix

Zhordyn:
## About the PR 
Adds fonts to `browserassets/src/css/fonts/` and fixes how paper.dm assigns font colors.

## Why's this needed? 
Newer installations of Windows don't include all the font styles referenced by goonstation code, this fixes the missing font issue such that Paper looks as expected, [an old issue that this PR likely resolves.](https://github.com/goonstation/goonstation/issues/14236) It also fixes how paper.dm assigns a font color (it used to look at object color, which isn't the behavior for certain items like Fancy Pen). This also paves the way for future fancier pens.

## Testing 

Here is what paper looks like when running Goonstation on new PCs.
BustedFonts

And now after the fixes 
Working Font Calls

Edit: ZeWaka testing after my changes:
image

HOP's Golden Crayon font color is busted before/after, but the rest of the writing implements function as intended.

### DIFF
--- a/browserassets/src/tgui/tgui.html
+++ b/browserassets/src/tgui/tgui.html
@@ -8,6 +8,7 @@
   <meta id="tgui:strictMode" content="[tgui:strictMode]">
   <meta id="tgui:byondMajor" content="[tgui:byondMajor]">
   <meta id="tgui:byondMinor" content="[tgui:byondMinor]">
+
   <!-- Early setup -->
   <!-- tgui:helpers -->
   <!-- tgui:ntos-error -->


Zhordyn:
Permanent Marker is licensed under Apache License, Version 2.0
[source](https://github.com/google/fonts/tree/11ca86e3b176fe1fff5ca8450c7e6f2fbee46fb5/apache/permanentmarker)

Dancing Script is licensed under SIL Open Font License, Version 1.1
[source](https://github.com/google/fonts/tree/11ca86e3b176fe1fff5ca8450c7e6f2fbee46fb5/ofl/dancingscript)

So based on my understanding of these license, we can redistribute them? EDIT: Further reading it, I believe we have to include the license files in the font folder when we redistribute.

ZeWaka:
> So based on my understanding of these license, we can redistribute them? EDIT: Further reading it, I believe we have to include the license files in the font folder when we redistribute.

Correct - add a LICENSE with the various files and licenses

```
DancingScript.woff is licensed under SIL 1.0

License terms here
```
Or something like that

browserassets/src/css/fonts/DancingScript.ttf
ZeWaka, line 1: Convert these two ttfs to woff2 files

Zhordyn: Used an online converter and tested them alone in Edge and Chrome, both WOFF2 files displayed correctly.

ZeWaka (changes requested):
there is no need to distribute ttfs if we're distributing woffs

TobleroneSwordfish (approved):
80% of the line count is licenses, we've reached peak FOSS PR 🙏

PR #24914, opened 2025-10-12
Mechcomp device frames and cabinets made scannable

CalliopeSoups:
[Game Objects] [Balance] 
## About the PR 

This PR simply makes it so you can mechscan device frames and cabinets. All anchored components inside are detached on deconstruction.

The current recipe is 20 bohrum and 20 conductive for the cabinets, 10 bohrum and 10 high conductive for frames. This'll be the main point of feedback I think. I can nerf or buff this recipe, for example they could require a small amount of gold as if they had fancy microchips. 

Edit: Recipe amounts lowered and deconstruction removed, see below comments.

## Why's this needed? 

For a long time this has been something largely thought to be a bad idea. However after chatting with the community a bit I feel misuse of components is rare especially nowadays and likely obvious to spot from an admin log perspective, that paired with the newer powergaming rule implemented a while ago I think it would be fair to allow some more fun with these toys. More silly robots waiter!

## Testing 
I've tested deconstruction and construction, as well as deconning several times in a row (which is why deconning detaches components now). Seems to be fine.

## Changelog 

```changelog
(u)CalliopeSoups
(+)Mechcomp device frames and cabinets are now mechscannable. 
```


ebozc:
Most of the OP stuff you could do with them have been balanced out (door openers etc), so this probably will work just fine balance-wise.
Recipes are a bit pricy though - It's not every round that Mining is that productive. Something like 10/5 and 8/4 would be better IMO.

A future rework of the housing GUI (i.e. housing not just being an inventory) would go nicely with this.

CalliopeSoups:
I took out the deconstruction as JORJ suggested, I don't see why you'd NEED to anyway. 

The recipe's been nerfed to ebozc's suggestion which I think I agree with.

JORJ949 (approved):

JORJ949:
Bah I've double approved but whatever, PR description is outdated but code looks good

PR #24915, opened 2025-10-12
Limit checking for unlikely chemical reactions

vuonojenmustaturska:
## About the PR 
Adds a small optimization to chemical reactions where the possible reactions list is built from a more limited chem reactions list, containing each reaction only once keyed off its hopefully-least likely reagent. Basically makes it so that when you add water to an empty bucket, the reaction for holy water isn't even considered because it's not the reagent the reaction is keyed by. `append_possible_reactions` can then use += instead of |= to build the list, speeding it up too. 

## Why's this needed? 
It's free performance

## Testing 
Spawned a few sets of junk using 
```dm 
/proc/spawn_a_bunch_of_reagent_buckets(var/amt)
	var/loc = get_step(usr, 0)
	for(var/i in 1 to amt)
		var/obj/item/reagent_containers/glass/bucket/B = new /obj/item/reagent_containers/glass/bucket(loc)
		var/obj/item/reagent_containers/glass/bucket/B2 = new /obj/item/reagent_containers/glass/bucket(loc)
		while (prob(66))
			var/datum/reagent/R = pick(reagents_cache)
			B.reagents.add_reagent(R, rand(1, 20))
			B2.reagents.add_reagent(R, rand(1, 20))

/proc/spawn_a_bunch_of_reacting_reagent_buckets(var/amt)
	var/loc = get_step(usr, 0)
	for(var/i in 1 to amt)
	    // spawn two so we get the same reaction for both buckets
		var/obj/item/reagent_containers/glass/bucket/B = new /obj/item/reagent_containers/glass/bucket(loc)
		var/obj/item/reagent_containers/glass/bucket/B2 = new /obj/item/reagent_containers/glass/bucket(loc)
		// pick a random reaction
		var/datum/chemical_reaction/CR = chem_reactions_by_id[pick(chem_reactions_by_id)]
		if (CR.required_reagents)
			for (var/R in CR.required_reagents)
				B.reagents.add_reagent(R, CR.required_reagents[R])
				B2.reagents.add_reagent(R, CR.required_reagents[R])
```
These are the results for the definitely-reacting stuff, the best-case scenario for old code where something eventually definitely reacts.
image
image

A bit more limited testing on the random reagent stuff, where stuff mostly does not react
image

Testing was done by adding a new var  `var/newbehavior = 0` on the reagents holder, and then in New() i had 
```dm
		var/static/apr_i = 0
		apr_i++
		src.newbehavior = apr_i % 2
```
I redirected handle_reactions and append/remove possible reactions based on newbehavior truthiness. The test procs created two buckets so the same reaction would happen in different behavior reagent containers. 

## Changelog 

```changelog
(u)Naksu
(+)Slight performance tweak for chemistry reactions. Report issues if you encounter any!
```

### DIFF
--- a/code/global.dm
+++ b/code/global.dm
@@ -431,6 +431,9 @@ var/global
 	list/datum/chemical_reaction/chem_reactions_by_id = list() //This sure beats processing the monster above if I want a particular reaction. =I
 	list/list/datum/chemical_reaction/chem_reactions_by_result = list() // Chemical reactions indexed by result ID
 
+	/// A single reaction is only in this list once, hopefully keyed by its least likely reagent, keeping the list of possible reactions small
+	list/limited_chem_reactions = list()
+
 	//SpyGuy: The reagents cache is now an associative list
 	list/reagents_cache = list()
 
--- a/code/modules/chemistry/Chemistry-Structure.dm
+++ b/code/modules/chemistry/Chemistry-Structure.dm
@@ -4,9 +4,16 @@
 	var/startTime = world.timeofday
 
 	total_chem_reactions.Cut()
+	limited_chem_reactions.Cut()
+
 	chem_reactions_by_result.Cut()
+
+	var/alist/tmpreactions = alist()
+
 	for(var/R in childrentypesof(/datum/chemical_reaction))
 		var/datum/chemical_reaction/CR = new R
+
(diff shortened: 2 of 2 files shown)


vuonojenmustaturska:
I should note that I don't actually play on goon and have not tested this extensively outside of the listed procs producing expected things. There *should* be no issues, but reactions not working might make players cranky.

PR #24917, opened 2025-10-12
Color connecting pipe overlays for manifolds when welded

roflcopter4:
## About the PR 
I found it a little disappointing how, when welding pipes to reinforce them, manifolds and quadways would only update their core icon. The connecting underlays remained the same. Only ordinary pipes had the power to alter the underlays. I've simply changed it so that this isn't the case.

Specifically, when manifolds and quadways are welded, the core icon changes as before. Any underlays between them and something that isn't a simple pipe or another manifold or quadway is changed to the new color. The underlays between two manifolds or quadways only change if both of them are the same material. This is to prevent the color awkwardly changing suddenly. This could be undone if desired. I wasn't sure what would be best.

In the process, I replaced the `SET_PIPE_UNDERLAY` macro with a new procedure called `update_pipe_underlay`. Every single use of that macro had the same condition copy-pasted in it, which was moved to the procedure, along with the new logic. I thought it was getting a little too large for a macro.

## Why's this needed? 
It looks better. It's easier to see which manifolds and quadways are reinforced and which are not. It's more consistent with pipes.

## Testing 
Here is the current game behavior. Note that the underlays between pipes and manifolds get recolored, but those between manifolds, and between manifolds and non-pipes, do not.
dreamseeker_n4DUaN9R6r

Here is the new behavior. Note how the underlays between two manifolds of the same material get updated, but those between manifolds of a different material do not.
dreamseeker_PjbdLOEmBq

Here is the intended effect, when one reinforces the whole network with the same material, as actual players ordinarily do (the above mixed materials to show what happens):
dreamseeker_sIb6eC78Fb

## Changelog 

```changelog
(u)RoflCopter4
(+)The connecting pipe segments between manifolds and other objects now match the color of the manifold after being welded.
```

### DIFF
--- a/code/modules/atmospherics/machinery/pipe.dm
+++ b/code/modules/atmospherics/machinery/pipe.dm
@@ -621,9 +621,9 @@
 /obj/machinery/atmospherics/pipe/manifold/hide(var/intact)
 	var/hide_pipe = CHECKHIDEPIPE(src)
 	invisibility = hide_pipe ? INVIS_ALWAYS : INVIS_NONE
-	update_pipe_underlay(src.node1, turn(src.dir, 90),  "small", hide_pipe)
-	update_pipe_underlay(src.node2, turn(src.dir, 180), "small", hide_pipe)
-	update_pipe_underlay(src.node3, turn(src.dir, -90), "small", hide_pipe)
+	update_pipe_underlay(src.node1, turn(src.dir, 90),  "short", hide_pipe)
+	update_pipe_underlay(src.node2, turn(src.dir, 180), "short", hide_pipe)
+	update_pipe_underlay(src.node3, turn(src.dir, -90), "short", hide_pipe)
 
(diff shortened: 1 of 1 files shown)


starsylk:
Wouldn't this make it hard to tell whether you've reinforced the pipes underneath with the corresponding material?

roflcopter4:
> Wouldn't this make it hard to tell whether you've reinforced the pipes underneath with the corresponding material?

It should make it easier to tell. There are no pipes "underneath". This change would just color the small interconnecting underlays between pipe manifolds the same color as the manifold itself. It's the same logic that currently already applies to ordinary pipes.

starsylk:
> > Wouldn't this make it hard to tell whether you've reinforced the pipes underneath with the corresponding material?
> 
> It should make it easier to tell. There are no pipes "underneath". This change would just color the small interconnecting underlays between pipe manifolds the same color as the manifold itself. It's the same logic that currently already applies to ordinary pipes.

Oh, perfect then! No more missing a manifold during reinforcement!

PR #24929, opened 2025-10-14
Change the second "command" to "command_2" in guardbuddy code, to allow for guardbuddy targeting again

Mikeistaken:
[A-Command] [C-Bug]
## About the PR 
Fixes a feature in Guardbuddies that was effectivly disabled, because of lines 2889-2894 failing to make command2 into base command. This PR fixes it by manualy making the command down at the problem area into command_2.

## Why's this needed?
It fixes a bug I found when code diving for specific targeting for guardbuddies

## Testing
Used "command=configure;command2=add_target;data=[self];acc_code=[Head pass]" and it not work, then applied the change and used the packet "command=configure;command_2=add_target;data=[self];acc_code=[Head pass]"
{13178955-313A-4DC2-AC36-ED41CE29AFB9}

### DIFF
--- a/code/modules/robotics/bot/guardbot.dm
+++ b/code/modules/robotics/bot/guardbot.dm
@@ -2972,8 +2972,8 @@ TYPEINFO(/obj/item/device/guardbot_module)
 				if (target_name && target_name != "")
 					src.target_names = list(target_name)
 
-			if (confList["command_2"])
-				switch(lowertext(confList["command"]))
+			if (confList["command"])
+				switch(lowertext(confList["command_2"]))
 					if ("add_target")
 						if(confList["acc_code"] != netpass_heads)
 							return 0


code/modules/robotics/bot/guardbot.dm
```diff
@@ -2973,7 +2973,7 @@ TYPEINFO(/obj/item/device/guardbot_module)
 					src.target_names = list(target_name)
 
 			if (confList["command"])
-				switch(lowertext(confList["command"]))
+				switch(lowertext(confList["command_2"]))
```
TobleroneSwordfish, line 2976: A couple of issues with this:
- It's meant to be `command2` according to the rest of the code, no underscore.
- There's one step up the callstack that is meant to re-assign the value of `command2` to `command`, either fix or remove that and replace it with this, don't just leave both lying around when only one works.

Mikeistaken:
I was being dumb last night, and this works as intended

PR #24930, opened 2025-10-14
HOTFIX: Removes wraith ability to trigger the lawbringer explosion

TacoGS:
## About the PR 
During a round one of our fine officers was killed in action with their own lawbringer when a wraith possessed it and apparently spoke as the gun...

## Why's this needed? 
Wraith possession should not be able to speak as a gun. This does not prevent them from firing it but the fine speech function needed to trigger it should not able to be spoken as the gun itself. 

## Testing 
Screenshot 2025-10-14 023409

### DIFF
--- a/code/obj/item/gun/energy.dm
+++ b/code/obj/item/gun/energy.dm
@@ -1570,6 +1570,10 @@ TYPEINFO(/obj/item/gun/energy/lawbringer)
 			//you must be holding/wearing the weapon
 			//this check makes it so that someone can't stun you, stand on top of you and say "I am the law" to kill you
 			if (src in M.contents)
+				// This is to check if someone is the gun, prevents them from speaking "I am the Law" into the gun while the gun. This was triggered by a Wraith event that happened.
+				if (islivingobject(M))
+					boutput(usr, SPAN_ALERT("<b>You cannot speak into the gun.</b>"))
+					return 1;
 				if (M.job != "Head of Security" || src.emagged)
 					src.cant_self_remove = 1
 					playsound(src.loc, 'sound/weapons/armbomb.ogg', 75, 1, -3)


DisturbHerb:
you were too good for this world, little gun 😭

xxfishshadowxx:
Wraiths being able to speak as posessed items is intentional, no? This seems to just nullify the one downside the lawby has over the egun, HOS that pick the lawby subscribe to the fact that it is a bomb anyone can use against them

Polydanx:
I say we keep it tbh,, the lawbringer has a speaker on it that you hear when you change modes that the wraith is presumably using,, also it's really funny

Cinder-Arcanium:
I don't see really any benefit to this change... Hos probably should be punished for losing their super murder kill gun. It's really funny as well. Plus it's not like the gun moves that fast and it takes long enough to explode to run away. I don't think "I died to this" (During a round one of our fine officers was killed in action with ) is a good reason to remove this or anything tbh

TacoGS:
> Wraiths being able to speak as posessed items is intentional, no? This seems to just nullify the one downside the lawby has over the egun, HOS that pick the lawby subscribe to the fact that it is a bomb anyone can use against them

Yeah, the code specifically has comments in it stating the gun needs to be held to prevent this exact problem. Where you don't get stunned, someone stands over you and says "I am the law" and runs off. The difference is, not only is this not checked because they are a wraith there is absolutely no risk to the wraith for being able to do a 2 second delay extremely powerful explosion. 

I get the argument that the lawbringer has that drawback but there are safeguards against someone just knocking you over and saying the magic words. I think the wraith has already been given a free pass to that.

TacoGS:
> I don't see really any benefit to this change... Hos probably should be punished for losing their super murder kill gun. It's really funny as well. Plus it's not like the gun moves that fast and it takes long enough to explode to run away. I don't think "I died to this" (During a round one of our fine officers was killed in action with ) is a good reason to remove this or anything tbh

I've not had a HoS Complain about the lawbringer being used on them or that someone did this by picking up the gun and saying the words but at least the HoS has some agency over something tangable and not just got knocked down and bam gun is possesed and dead without seeing who did it. 

I like dying in horrible ways as much as the other people, but I actually like to be part of it other than a 2 seconds and dead. Things like TTVs, other types of bombs is someone has to spend time producing those that are more than a 2 second posses and copy pasta. 

I think in a game people are supposed to enjoy needs a bit of fine balancing act between overpowered and funny.

Cinder-Arcanium:
I don't see much of a difference between. Hos being stunned and then a traitor picking it up and copy paste. It's no slower than the wraith.

DisturbHerb:
My present feelings hinge on whether or not this phenomenon has posed a recurring issue to where it would significantly worsen the players' experiences during a round.

Yeah, it's a really cheap tactic to irreversibly cause a lethal explosion alongside with taking out the rest of your gear with a simple copy+paste or fast enough typing. It would suck to be on the receiving end (Subject to change given enough instances of it happening, probably). Maybe it's not intended behaviour.

However, if my own understanding is correct, I'm not sure that wielders being simultaneously stunned and disarmed with their lawbringer in hand and having it subsequently possessed occurs enough for this to pose a significant concern. My feelings on being a Security player predispose me to wanting to be okay with taking more Ls, I'd think any wraith to try this on me would be a rather cunning fellow for the first one or two times. If nothing else, it *is* clever! Pat on the back, cheerio. I see the absurdity and humour in such a situation.

Whether or not this PR is merged in the form it's in as of writing, I trust that if reports of exploding lawbringers crop up enough, we'd do the right thing for the sake of the people trying to enjoy the game. I, for one, will be a little more careful about ever taking my weapons off my belt (which an officer or HoS really should already be doing, I think).

TacoGS:
> My present feelings hinge on whether or not this phenomenon has posed a recurring issue to where it would significantly worsen the players' experiences during a round.
> 
> Yeah, it's a really cheap tactic to irreversibly cause a lethal explosion alongside with taking out the rest of your gear with a simple copy+paste or fast enough typing. It would suck to be on the receiving end (Subject to change given enough instances of it happening, probably). Maybe it's not intended behaviour.
> 
> However, if my own understanding is correct, I'm not sure that wielders being simultaneously stunned and disarmed with their lawbringer in hand and having it subsequently possessed occurs enough for this to pose a significant concern. My feelings on being a Security player predispose me to wanting to be okay with taking more Ls, I'd think any wraith to try this on me would be a rather cunning fellow for the first one or two times. If nothing else, it _is_ clever! Pat on the back, cheerio. I see the absurdity and humour in such a situation.
> 
> Whether or not this PR is merged in the form it's in as of writing, I trust that if reports of exploding lawbringers crop up enough, we'd do the right thing for the sake of the people trying to enjoy the game. I, for one, will be a little more careful about ever taking my weapons off my belt (which an officer or HoS really should already be doing, I think).

This stemmed from an incident that actually happened. I suspect this PR may have the unfortunate consequence of letting people know it can happen. I doubt it was intended because of the comments on the code itself for the detonation check. 

From the endless testing I've done with it, I can kill anyone within 2 seconds of possession and all I had to do was wait for it or cause them to trip with the intention of doing this.

In the end, it is up to the community to choose if this gets merged, I asked for feedback for a reason but there are some Ls I am willing to take but a wraith I can't see, I can't do anything about deleting my round in 2 seconds isn't something I find fun. Neither did the HoS who it happened to. 

I am just waiting for the people to say "Skill issue" or "don't drop your gun 4head". 

Maybe if a flat out block of wraiths detonating the gun isn't merged we can think of other safety checks like increasing the time to the detonation since the wraith already gets a pass on the DNA check in the code.

DisturbHerb:
> I am just waiting for the people to say "Skill issue"

for ze record i did say this

TacoGS:
> > I am just waiting for the people to say "Skill issue"
> 
> for ze record i did say this

Noted :)

Glamurio:
I think it's a cool funny interaction, and in all my rounds where I've had wraith (not many) and either played or had a HoS alongside wraith (even fewer), in exactly zero of them did this problem occur.

If people use this ability sparingly, I don't see the issue. On many wraith rounds you don't even need to use your lawbringer because you can't attack wraiths anyway, and as long as they can't pull it out of your pocket, that seems okay with me. Now if we had every wraith, every round pull this trick, it'd get boring fast, but the fact that the vast majority of people haven't even seen this happen means to me that things are probably fine. And even if it were to happen... so what? Gun goes boom, HoS maybe survives or gets cloned and the round continues as normal.

TacoGS:
> I think it's a cool funny interaction, and in all my rounds where I've had wraith (not many) and either played or had a HoS alongside wraith (even fewer), in exactly zero of them did this problem occur.
> 
> If people use this ability sparingly, I don't see the issue. On many wraith rounds you don't even need to use your lawbringer because you can't attack wraiths anyway, and as long as they can't pull it out of your pocket, that seems okay with me. Now if we had every wraith, every round pull this trick, it'd get boring fast, but the fact that the vast majority of people haven't even seen this happen means to me that things are probably fine. And even if it were to happen... so what? Gun goes boom, HoS maybe survives or gets cloned and the round continues as normal.

Yeah, I've heard of it once and the PR is due to the balance of it. If the HoS brings the lawbringer out and then gets hit by decay, which is a 3-4 second stun they drop the weapon, you posses it and detonate the HoS is pink mist.

However, from what I can tell from the feedback it seems that there is no one really in support of this change. I will just close the PR.

PR #24934, opened 2025-10-14
Getting interrupted no longer closes your chatbox

TobleroneSwordfish:
## About the PR 
See title, when you get -BLUGH interrupted your chatbox will now stay open.
## Why's this needed? 
I'm kind of on the fence about this which is why I'm PRing it, in theory closing the chatbox makes sense because you've been interrupted and want to respond immediately, but in practice all this usually does is make the last few key presses of the input go as game inputs and people often open various random menus they don't mean to by typing M, P, Y etc.
## Testing 
I tested that it doesn't close the chatbox and it doesn't.

## Changelog 

```changelog
(u)LeahTheTech
(+)Getting interrupted no longer closes your chatbox.
```

### DIFF
--- a/code/mob/living.dm
+++ b/code/mob/living.dm
@@ -1640,9 +1640,6 @@ TYPEINFO(/mob/living)
 					enteredtext = ";" + enteredtext
 				winset(client, "[window_type]window.say-input", "text=\"\"")
 				if (isnull(client)) return
-				if (isnull(client)) return
-				src.cancel_typing(window_type)
-				src.cancel_emote_typing(window_type)
 				found_text = TRUE
 				break
 	if (found_text)


JORJ949:
Does it clear the chatbox to avoid saying the same thing multiple times when, for example, you're being swirlied?
Also, what do cancel_typing do shortly after? My guess would be unfocusing the typing window or removing your speech bubble overlay but you'd want neither of these to happen.

TobleroneSwordfish:
It does clear the chatbox, although I think there's a current bug with lastgasp that means that the winset sometimes doesn't go through in time with server lag leading to duplicate messages while being rapidly swirlied. Not addressed in this PR.

DisturbHerb:
we could bloat character prefs further by-

Glamurio:
This is an incredible change if only to avoid people (and myself) walking out of cryo tubes when typing.

PR #24936, opened 2025-10-14
Adds Juice That Makes Your Head Explode

Random-Guy-With-A-Shotgun:
[MEDICAL] [CHEMISTRY]
## About the PR
Adds a new disease reagent called "juice that makes your head explode," which causes "Exploding Head Syndrome." (It's id is explodingheadjuice for brevity) In reference to the actual disease of the same name, exploding head syndrome causes the victim to repeatedly hear explosions and other loud noises. These noises are not audible to anyone else, but can cause the affected mob to flinch or scream making it apparent something is wrong with them. Of course, once it reaches stage 5 it has a small chance to cause the victims head to explode after a short delay. It can be cured with synaptizine.

The reagent can be created by mixing: black powder, prions, hard punch, raspberry juice, and hootingium.

The name is in reference to a TomSka skit.

This is meant to be a silly poison for antag chefs and bartenders. This was originally envisioned as a weaker version of gibbis that could be obtained without relying on poison bottle RNG.

Edits
After receiving feedback:

- I removed the mixing sound.
- Changed the cure from stabilizing agent to synaptizine.
- Removed holy water from the recipe
- Added an actual explosion when the player's head explodes. This explosion does sometimes gib the body.
- changed ID from ehsjuice to explodingheadjuice

## Why's this needed?
It gives antags a more readily obtainable, but still hard to make, disease reagent that is a bit more stealthy than nanomachines or prions. 

## Testing 
Here is me testing that it all works and showing it off. (It doesn't show off the intermediate stages super well but for some reason disease RNG was being annoying)
https://github.com/user-attachments/assets/14e2396c-4eb9-4e41-b441-86c2dfb47167

## Changelog
```changelog
(u)AI-the-artificial-idiot
(*) Adds "juice that makes your head explode", a mixable disease reagent which infects the victim with exploding head syndrome.

### DIFF
--- a/code/modules/chemistry/Chemistry-Recipes.dm
+++ b/code/modules/chemistry/Chemistry-Recipes.dm
@@ -5493,10 +5493,10 @@
 		result_amount = 2
 		mix_sound = 'sound/effects/radio_sweep1.ogg'
 
-	ehsjuice
+	explodingheadjuice
 		name = "Juice That Makes Your Head Explode"
-		id = "ehsjuice"
-		result = "ehsjuice"
+		id = "explodingheadjuice"
+		result = "explodingheadjuice"
 		required_reagents = list("blackpowder" = 1, "hard_punch" = 1, "juice_raspberry" = 1, "prions" = 1, "sonicpowder" = 1)
 		mix_phrase = "The mixture coalesces into a dark red liquid."
 		result_amount = 5
--- a/code/modules/medical/diseases/exploding_head_syndrome.dm
+++ b/code/modules/medical/diseases/exploding_head_syndrome.dm
@@ -6,7 +6,7 @@
 	cure_desc = "Synaptizine"
 	reagentcure = list("synaptizine")
 	recureprob = 10
-	associated_reagent = "ehsjuice"
+	associated_reagent = "explodingheadjuice"
 	affected_species = list("Human","Monkey")
 
 
--- a/code/modules/chemistry/Reagents-Diseases.dm
+++ b/code/modules/chemistry/Reagents-Diseases.dm
@@ -546,9 +546,9 @@ datum
 			transparency = 50
 			disease = /datum/ailment/disease/leprosy
 
-		disease/EHSjuice
+		disease/explodingheadjuice
 			name = "juice that makes your head explode"
-			id = "ehsjuice"
+			id = "explodingheadjuice"
 			description = "A liquid with an incredibly apt name."
 			reagent_state = LIQUID
 			taste = "like berries"

### CODE REVIEW

**Comment by RSOD404** [comment]
having such an easy cure for something so hard to make sounds a bit counterintuitive, as soon as someone health scans you it's just a matter of walking over to a chem dispenser and putting in 3 chems. otherwise, i'm glad we finally get a PR for a tomska reference (almost as funny as the Hole)
another thing, looking at the footage, i'm a bit iffy on the jingle that plays during the mixing, considering it's very hard to miss and would probably be a dead giveaway that Someone's Up To Evil. perhaps play it only for people within a small radius of the container or don't at all

**Comment by Random-Guy-With-A-Shotgun** [comment]
Yeah I mad the cure really easy because I was worried about it being too strong, put I guess I over corrected. I'll figure something out. Also I can figure out how to turn down the jingle, I also made it super obvious to avoid it being overpowered, but I guess I shouldn't have been worried about that.

File: code/modules/chemistry/Chemistry-Recipes.dm [hunk 1]
```diff
@@ -5492,3 +5492,13 @@
...
 		mix_sound = 'sound/effects/radio_sweep1.ogg'
+
+	ehsjuice
+		name = "Juice That Makes Your Head Explode"
+		id = "ehsjuice"
+		result = "ehsjuice"
+		required_reagents = list("blackpowder" = 1, "hard_punch" = 1, "juice_raspberry" = 1, "prions" = 1, "sonicpowder" = 1, "water_holy" = 1)
+		mix_phrase = "The mixture coalesce into a dark red liquid. A strange jingle rings out of it."
+		result_amount = 6
+		mix_sound = "sound/effects/ehsjuicejingle.ogg"
```
TobleroneSwordfish: No, putting random meme bits of youtube videos in the game is not a good idea, even disregarding copyright issues.

code/modules/chemistry/Chemistry-Recipes.dm
```diff
@@ -5492,3 +5492,13 @@
 		mix_phrase = "The mixture emits a sudden whine of static and forms into swirling, many faceted shapes that hurt to look at."
 		result_amount = 2
 		mix_sound = 'sound/effects/radio_sweep1.ogg'
+
+	ehsjuice
+		name = "Juice That Makes Your Head Explode"
+		id = "ehsjuice"
+		result = "ehsjuice"
+		required_reagents = list("blackpowder" = 1, "hard_punch" = 1, "juice_raspberry" = 1, "prions" = 1, "sonicpowder" = 1, "water_holy" = 1)
```
TobleroneSwordfish: Could proobably remove at least one of these, since each one other than sonic powder is a Process to get.

code/modules/medical/diseases/exploding_head_syndrome.dm
```diff
@@ -0,0 +1,54 @@
...
+					affected_mob.emote("flinch")
+			if(probmult(15))
+				boutput(affected_mob, SPAN_ALERT("You see a flash of light in the corner of your vision"))
+				affected_mob.take_brain_damage(3)
+			if(probmult(5))
+				boutput(affected_mob, SPAN_ALERT("<B>You feel a strange tingle moving towards your head</B>"))
+				SPAWN(rand(20, 100))
+					if (affected_mob)
+						var/mob/living/carbon/human/H = affected_mob
+						H.head_explosion()
```
TobleroneSwordfish, line 54: Feels like this should come with a (small) actual explosion too. It shouldn't gib their corpse since the life loop will need to tick before they actually die from having no head.

code/modules/chemistry/Chemistry-Recipes.dm
Random-Guy-With-A-Shotgun: Do you want me to completely raname the reagent or should I just remove the sound bite?

TobleroneSwordfish: The name is generic enough to be fine, just remove the jingle

Calphrick:
It should be wine, not hard punch. Part of the pun in the skit is that the sniper is Jewish, and Jews drink wine frequently in religious ceremonies.

Random-Guy-With-A-Shotgun:
> It should be wine, not hard punch. Part of the pun in the skit is that the sniper is Jewish, and Jews drink wine frequently in religious ceremonies.

I want to keep this recipe at least moderately difficult, If I replace anything with wine it'd be hootingium.

Flaborized:
I don't mind the chem as it's Funny though I would prefer if the ID more closely matched the chem name. As an admin, when you spawn reagents you search by chem name so I should be able to type like "exploding" or "head" and get it. Brevity doesn't really matter I'd literally just have it as "explodingheadjuice" or something like that lol

TobleroneSwordfish (approved):
Approved, pending Flaborized's comment

Random-Guy-With-A-Shotgun:
Currently I don't have access to my PC with the files on it, but I should be able to change the chem's ID on either Sunday or Monday and will write another comment once I have done that.

Random-Guy-With-A-Shotgun:
> Approved, pending Flaborized's comment

Alright I changed the id name per Flaborized's request. It is now "explodingheadjuice"

PR #24938, opened 2025-10-14
Who command shows # not playing

FlameArrow57:
[PLAYER ACTIONS][FEATURE]

## About the PR 
Makes Who command display "Total players: [num logged in] ([num actually playing)]"

## Why's this needed? 
Shows true number of people playing

## Testing 

Tested on local from menu and in game

## Changelog 
```changelog
(u)FlameArrow57
(*)Who command now also shows number of players actually in game
```

### DIFF
--- a/code/mob/who.dm
+++ b/code/mob/who.dm
@@ -95,7 +95,7 @@
 			rendered += aPreAuth
 
 	var/total_players = length(whoAdmins) + length(whoMentors) + length(whoNormies)
-	rendered += "<b>Total Players: [num_in_game] ([total_players - num_new_players] in game)</b>"
+	rendered += "<b>Total Players: [total_players] ([total_players - num_new_players] in game)</b>"
 	rendered += "</div>"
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Mmmm I'm not sure this is a good idea, you can already get an idea of how many people are in-game from looking at the latejoin menu and I think spelling it out like this will only encourage people on lowpop to join, see there aren't many people playing, and then just spectate instead.

emmenthal (approved):
I have a hunch that most of the active playerbase has no idea what "who" is and how to invoke it. And those that do, well, they already can use the latejoin menu and have their set approach on whether to join on lowpop or not.

Glamurio (approved):
I just think it's good QoL

JORJ949:
Notably makes off-station antag rounds much more metagamable by directly comparing ingame to manifest rather than just "everyone connected vs manifest", meaning only observer players count towards this inaccuracy (which means very few if you who+manifest as soon as the round starts).

PR #24947, opened 2025-10-15
Adds the Food processor artifact

Kanthes:
[SCIENCE][FEATURE]

## About the PR 
I once saw a Scientist wishing they could eat the artifacts.

Adds the Food Processor artifact. The Food Processor accepts biological users and items much like the Cyborg Converter and Enzymatic Reclaimer, processing these into a randomized foods. These foods are given the appearance of extra small artifacts from the Food Processors origin. As all appearance and description of the food is overwritten, the randomized foods have been chosen for having various effects upon being eaten.

## Why's this needed? 

The Food Processor adds another artifact that can be given to another department to give them a unique twist on the regular job experience.

## Testing 

https://github.com/user-attachments/assets/d42fb4ff-c33c-4c91-833d-85722f914070

## Changelog 

```changelog
(u)Kanthes
(*)Adds the Food Processor artifact, which will take biological materials to produce mysterious foods.
```

### DIFF
--- a/code/obj/artifacts/artifact_machines/food_processor.dm
+++ b/code/obj/artifacts/artifact_machines/food_processor.dm
@@ -41,8 +41,10 @@
 
 	New()
 		..()
-		src.biomatter_per_food = rand(0, 10)
+		src.biomatter_per_food = rand(1, 4)
 		src.loops_per_consumption_step = src.escapable ? rand(4, 7) : rand(2, 4)
+		if (prob(15))
+			escapable = FALSE
 
 	//Effect_touch copied from borgifier.
 	effect_touch(var/obj/O, var/mob/living/user)
@@ -101,7 +103,6 @@
 					convertable_limbs -= limb_to_replace
 					humanuser.update_body()
 					src.biomatter += 2
-					spawn_food(O)
 				sleep(0.4 SECONDS)
 
 			var/bdna = null // For forensics (Convair880).
@@ -114,16 +115,16 @@
 
 			ArtifactLogs(user, null, O, "touched", "food processing user", 0) // Added (Convair880).
 
-			user.set_loc(get_turf(O.loc))
-			user.death()
-			user.ghostize()
-			qdel(user)
-
 			//Biofuel per body copied from the enzymatic reclaimer
 			var/humanOccupant = (ishuman(user) && !ismonkey(user))
 			var/decomp = ishuman(user) ? user:decomp_stage : 0
 			src.biomatter += rand(5, 8) * (humanOccupant ? 2 : 1) * ((4.5 - decomp) / 4.5)
 			src.food_processor_state = STATUS_READY
+
+			user.set_loc(get_turf(O.loc))
+			user.death()
+			user.ghostize()
+			qdel(user)
 
 			spawn_food(O)
 		else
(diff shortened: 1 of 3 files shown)


DisturbHerb:
https://youtube.com/shorts/wutKiNRsUdI

code/obj/artifacts/artifact_machines/food_processor.dm
```diff
@@ -0,0 +1,169 @@
...
+			return
+		if (!user)
+			return
+		if (src.food_processor_state != STATUS_READY)
+			return
+		if (ishuman(user))
+			var/mob/living/carbon/human/humanuser = user
+			if(!isalive(user) && user.ghost && user.ghost.mind && user.ghost.mind.get_player()?.dnr)
+				O.visible_message(SPAN_ALERT("<b>[O]</b> refuses to process [user.name]!"))
+				return
```
TobleroneSwordfish:
```suggestion
```
Your copy paste is showing :P

code/obj/artifacts/artifact_machines/food_processor.dm
```diff
@@ -0,0 +1,169 @@
...
+
+	//Effect_touch copied from borgifier.
+	effect_touch(var/obj/O, var/mob/living/user)
+		if (..())
+			return
+		if (!user)
+			return
+		if (src.food_processor_state != STATUS_READY)
+			return
+		if (ishuman(user))
```
TobleroneSwordfish: Please change this to a guard like the above for the sake of my sanity.

code/obj/artifacts/artifact_machines/food_processor.dm
```diff
@@ -0,0 +1,169 @@
...
+
+	proc/spawn_food(var/obj/O)
+		while(src.biomatter > src.biomatter_per_food)
+			playsound(O.loc, 'sound/machines/vending_dispense.ogg', 50, 1, -1)
+			sleep(0.5 SECONDS)
+
+			//Pick a random food item.
+			//TODO: Filter out raw ingredients as well as items requiring forks or other utensils.
+			src.biomatter -= src.biomatter_per_food
+			var/food = pick(src.possible_foods)
```
TobleroneSwordfish: Might be more interesting if it was the same food item each time per artifact? Or at least a chance for that?

code/obj/artifacts/artifact_machines/food_processor.dm
```diff
@@ -0,0 +1,169 @@
...
+		else
+			return
+
+	proc/spawn_food(var/obj/O)
+		while(src.biomatter > src.biomatter_per_food)
+			playsound(O.loc, 'sound/machines/vending_dispense.ogg', 50, 1, -1)
+			sleep(0.5 SECONDS)
+
+			//Pick a random food item.
+			//TODO: Filter out raw ingredients as well as items requiring forks or other utensils.
```
TobleroneSwordfish:
```suggestion
```

code/obj/artifacts/artifact_machines/food_processor.dm
TobleroneSwordfish:
```suggestion
```
Not relevant for food

code/obj/artifacts/artifact_machines/food_processor.dm
TobleroneSwordfish: Please turn this into a guard for the sake of my sanity.

code/obj/artifacts/artifact_machines/food_processor.dm
Kanthes: Without my elders I'd be nothing :P

LorrMaster:
I think it might be better if normal food sprites were used rather than the artifact sprites. Might need to drop the randomness, but it makes what the artifact is doing a bit clearer. Maybe even apply a color filter if there needs to be more distinction.

Kanthes:
> I think it might be better if normal food sprites were used rather than the artifact sprites. Might need to drop the randomness, but it makes what the artifact is doing a bit clearer. Maybe even apply a color filter if there needs to be more distinction.

If the artifact just spits out the same food every time with regular food sprites, then it's not providing an interesting twist on the chef job experience, it's just automating it. This is a common complaint at the Single Target Healer, for example.

As for clarity, the sprites have a clear examine hint telling the user they are edible. Do you have any suggestion of how to make it more obvious?

TobleroneSwordfish (approved):

Glamurio (approved):

PR #24955, opened 2025-10-16
Conga lines 2

Valtsu0:
[Actions][Feature]
## About the PR 
sequel to #20064
You now pull others when you are pulled (and with grabs)
~~this currently includes pins which i think is really funny and balanced enough by requiring 2 people but that can go if people disagree~~ this behavior is not from this pr and works the same on master
also fixes the teleporting when pulling things in a different direction to the one you are moving in

## Why's this needed? 
fun

## Testing 
Spawned a bunch of assistants, possessed them and made them pull eachother

## Changelog 

```
(u)Valtsu0
(*)You now pull while being pulled, allowing you to form conga lines
```

### DIFF
--- a/code/mob/input.dm
+++ b/code/mob/input.dm
@@ -305,43 +305,9 @@
 								src.emote("wheeze")
 								boutput(src, SPAN_ALERT("You flop over, too winded to continue running!"))
 
-						var/list/pulling = list()
-						if (src.pulling)
-							if ((BOUNDS_DIST(old_loc, src.pulling) > 0 && BOUNDS_DIST(src, src.pulling) > 0) || !isturf(src.pulling.loc) || src.pulling == src) // fucks sake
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
Changelog needs fixing

Valtsu0:
done

RSOD404:
the staffie conga will take over the station

TobleroneSwordfish (changes requested):
Code looks good, but I do not see anything good coming from pulling pins around like that so please exclude them.

Valtsu0:
> Code looks good, but I do not see anything good coming from pulling pins around like that so please exclude them.

uhh it turns out, you can just pull people pinning someone, on master, right now, and it works exactly the same as in this pr

TobleroneSwordfish:
aamazing

PR #24958, opened 2025-10-17
Experimental atmos optimization spitball nonsense

vuonojenmustaturska:
[ATMOSPHERICS][EXPERIMENTAL]
## About the PR 
This PR contains a micro-optimization to react() and a couple of different versions of share() that are largely equivalent (the branchopt ones set temperature 1 time out of 10000 without need). Mostly opened for comments and/or if someone wants to take it off my hands or to run it as a testmerge to see what a real profile says. 
## Why's this needed? 
Might be extra performance? 

## Testing 
Tested locally but results pretty inconclusive. share_new appears to be about 5-10% faster but local testing is not representative of actual gameplay.

### DIFF
--- a/code/modules/atmospherics/FEA_gas_mixture.dm
+++ b/code/modules/atmospherics/FEA_gas_mixture.dm
@@ -69,8 +69,15 @@ What are the archived variables for?
 
 /// Process all reactions, return bitfield if notable reaction occurs.
 /datum/gas_mixture/proc/react(atom/dump_location, mult=1)
+	if (prob(50))
+		return react_new(dump_location, mult)
+	return react_old(dump_location, mult)
+
+/// Process all reactions, return bitfield if notable reaction occurs.
+/datum/gas_mixture/proc/react_old(atom/dump_location, mult=1)
 	. = 0 //(used by pipe_network and hotspots)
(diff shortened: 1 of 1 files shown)


Laboredih123:
tracy!!!!!!!!!

Laboredih123:
also instead of doing a prob(50) just go all in on one or the other

vuonojenmustaturska:
The idea is that both get roughly equivalent amount of calls so that they can be compared. I have not used tracy, it didn't exist the last time I really touched SS13 code, but I would imagine the same caveat applies in that I cannot simulate an actual real round of ss13 on my local computer

TobleroneSwordfish:
Testing this kind of incremental performance increase live will also be very inconclusive, your best bet is to use the "disco inferno" testing scenario on a local server as that provides a (mostly) consistent high atmos load scenario.
Tracy is also a much better profiler for demonstrating this, see https://hackmd.io/@goonstation/docs/%2F%40goonstation%2Fcode#Even-Better-Profiler

Laboredih123:
ill make the tracy for this pr when i can

vuonojenmustaturska:
If there's anything of value here, feel free to fetch the branch and make your own PR

PR #24976, opened 2025-10-20
Adds clown noses

Microoowaaave:
[CLOTHING][FEATURE]
## About the PR 
Adds clown noses, worn in the mask slot. 

They honk like a clown mask can, and are able to be found in the clown hole or bought off of Geoff.
## Why's this needed? 
It's a fun alternative to wearing a mask! 

This was originally meant to be apart of a larger PR that would also include clown make-up (https://forum.ss13.co/showthread.php?tid=24113), BUT I'm no coder (or spriter, considering the item sprite is a recolored stress ball), so, one step at a time. 

## Testing 

https://github.com/user-attachments/assets/9ab05d82-9d71-4ce8-9975-eb3921ec14b7

honk

## Changelog 

```changelog
(u)Microoowaaave
(+)Adds clown noses, found in the clown hole and sold by Geoff!
```


code/obj/item/clothing/masks.dm
```diff
@@ -884,3 +884,31 @@ ABSTRACT_TYPE(/obj/item/clothing/mask/bandana)
...
+
+// Clown nose!!
+
+/obj/item/clothing/mask/clown_nose
+	name = "clown nose"
+	desc = "A classic red clown nose. It even honks!"
+	icon_state = "clown_nose"
+	item_state = "clown_nose"
+	see_face = TRUE
+	var/base_icon_state = "clown_nose"
```
JORJ949: This was needed for the clown mask because you can use them in hand to fold the hair away so it needed to keep track, you don't need it here.

code/obj/item/clothing/masks.dm
```diff
@@ -884,3 +884,31 @@ ABSTRACT_TYPE(/obj/item/clothing/mask/bandana)
...
+
+	var/spam_flag = 0
+	var/spam_timer = 100
+	var/list/sounds_instrument = list('sound/musical_instruments/Bikehorn_1.ogg')
+	var/volume = 50
+	var/randomized_pitch = 1
+
+	proc/honk_nose(mob/user as mob)
+		if (!spam_flag)
```
JORJ949: You should make use of ON_COOLDOWN here, if you're not sure how to do that checkout _std\macros\cooldown.dm

code/obj/item/clothing/masks.dm
Microoowaaave: Gone!

code/obj/item/clothing/masks.dm
Microoowaaave: Doone? I think.

JORJ949: On cooldown is a nice way to do cooldowns, which is what the spam_flag and timer variables and checks were for, so you don't need either of them as on_cooldown manages all that for you

Microoowaaave: AH. I see, I see. Should be good now!

Microoowaaave:
Accidentally closed this, whoops.

Glamurio:
Could technically make this inherit from clown masks? But glad for it.

TobleroneSwordfish:
It looks a bit small on the face (and big on the ground)? Most clown masks do a kind of spread out 4-5 pixel nose:
image

Microoowaaave:
It's purposefully small on the face, there's really not a lot of real estate there and I think it'd look pretty odd with a larger worn sprite. It's big on the ground because I'm a lazy bastard that can't sprite.

Glamurio:
> It looks a bit small on the face (and big on the ground)? Most clown masks do a kind of spread out 4-5 pixel nose: image

I kinda agree with Microwave on this, we literally have one pixel between mouth and eyes. There's not much wiggle room. (Also it looks great like this.)

JORJ949 (approved):

flappybatpal:
There used to be (might still be) opinions on that clowns should really be wearing masks so I'm not really sure how controversial a single pixel nose might be.

It is very hard to see IMO.

JORJ949:
> There used to be (might still be) opinions on that clowns should really be wearing masks so I'm not really sure how controversial a single pixel nose might be.
> 

I am of the opinion that if a clown removes their mask, or has it removed, they're a heretic. But I think clown noses are fun and silly enough to be included in that mask listing.

You could also make this a subtype of the clown mask (you'd have to override the attack_self() action of changing style because this doesn't have hair to remove) to copy clown mask behaviour (it does other things like help determine if you're a clown for clown car purposes)

code/obj/item/clothing/masks.dm
```diff
@@ -887,3 +887,25 @@ ABSTRACT_TYPE(/obj/item/clothing/mask/bandana)
...
+
+// Clown nose!!
+
+/obj/item/clothing/mask/clown_nose
+	name = "clown nose"
+	desc = "A classic red clown nose. It even honks!"
+	icon_state = "clown_nose"
+	item_state = "clown_nose"
+	see_face = TRUE
+
```
pgmzeta, line 900: probably needs to null `c_flags` as it doesn't cover the mouth?

pgmzeta (changes requested):
how does this look on werewolves and monkeys?

Microoowaaave:
> how does this look on werewolves and monkeys?

guh1 (1)
guh1 (2)

Little off on monkeys (from the side), but I don't see an alternative there.

PR #24985, opened 2025-10-21
Spy thief in cryo warning

JORJ949:
## About the PR 
Spy thieves will now receive a warning when an item is in the cryotron, rather than just in the room with the cryotron.

Please tell me there's a better way to find if X item is in X thing through any amount of layers

## Why's this needed? 
Sometimes people drop their trinkets (or themselves) in the same room as the cryotron so this helps inform you if the bounty is possible or not before you go after it.

## Testing  
image

## Changelog 

```changelog
(u)JORJ949
(*)Spy thieves will now be informed if a target bounty is in cryogenic storage
```

### DIFF
--- a/code/obj/item/uplinks.dm
+++ b/code/obj/item/uplinks.dm
@@ -1124,11 +1124,15 @@ Note: Add new traitor items to syndicate_buylist.dm, not here.
 				src.menu_message += "<B>Current Bounties (Next Refresh at : [refresh_time_formatted]):</B>"
 				for(var/datum/bounty_item/B in game.active_bounties)
 					var/atext = ""
-					var/item_destroyed = FALSE // If the bounty required a specific item and was destroyed
-					if (B.reveal_area && B.item && !B.claimed)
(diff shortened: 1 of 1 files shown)


code/obj/item/uplinks.dm
```diff
@@ -1124,11 +1124,15 @@ Note: Add new traitor items to syndicate_buylist.dm, not here.
...
-					if (B.reveal_area && B.item && !B.claimed)
+					var/unavailable_text = null // Why the specific bounty can't be redeemed (Claimed, Destroyed or In Cryo)
+					if (B.claimed)
+						unavailable_text = "CLAIMED"
+					else if (B.reveal_area && B.item)
 						var/item_area = get_area(B.item)
 						if(!item_area) // Also includes if there's no item because if there's no item then it has no area
-							item_destroyed = TRUE
+							unavailable_text = "DESTROYED"
+						else if ((locate(/obj/cryotron) in get_turf(B.item)) && !isturf(B.item.loc)) //There's gotta be a better way to check for this
```
TobleroneSwordfish, line 1134: Well you could iterate up the loc chain but I'd argue this is better anyway

PR #24989, opened 2025-10-21
Fix observing admins via chat right click + observing admin alarm bells

JORJ949:
## About the PR 
Stops non-admins spectating non-player mode admins via the right click chat "observe" command
Adds alarm bells (debug log, message admins, boutput target client) for when someone does observe an admin so the observed admin can detonate the user if they didn't force them to observe them (which is meant to be the only way to observe an admin)

## Why's this needed? 
The clown is too powerful, they cannot know our secrets. If someone is observing a non-player mode admin, the target admin should know.
image

## Testing 
Made isadmin() return TRUE for stirstir, after deadminning myself could not observe stirstir via chat right click menu, could observe via clicking as observer (click check directly checks for a holder so this was expected), doing so caused the debug log entry to show up in the debug log (but i couldn't see the other messages on the count of being deadminned but I presume they work as they didn't runtime.)

### DIFF
--- a/code/datums/browserOutput.dm
+++ b/code/datums/browserOutput.dm
@@ -279,6 +279,8 @@ var/global
 			if(src.owner.holder)
 				src.owner.holder.playeropt(targetMob)
 		if ("observe")
+			if(!isadmin(src.owner) && isadmin(targetMob) && !targetMob.client?.player_mode)
+				return // no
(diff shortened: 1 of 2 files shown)


Laboredih123:
me when i :eyes: (im OBSERVIGN)

TobleroneSwordfish (approved):

PR #24996, opened 2025-10-22
Changes the default module on Security PDAs to flashlights

Calphrick:
[A-Security] [C-Balance]

## About the PR 
The alert module was removed ten (10) months ago. That's almost a year! In that time, Security has had the worst default PDA due to not having a functional module. This PR gives them the basic flashlight module to bring them in line with the rest of the station.
## Why's this needed? 
It sucks when you play security and suddenly realize that, unlike everyone else on the station you have no lightsource by default.
## Testing 
I didn't test the changes, but it's only a single word being changed, it should work fine.

## Changelog 

```changelog
(u)Calphrick
(+)Gave Security flashlight modules in their PDAs.
```

### DIFF
--- a/code/obj/item/device/pda2/pda2.dm
+++ b/code/obj/item/device/pda2/pda2.dm
@@ -128,7 +128,7 @@
 		icon_state = "pda-hos"
 		setup_default_pen = /obj/item/pen/fancy
 		setup_default_cartridge = /obj/item/disk/data/cartridge/hos
-		setup_default_module = /obj/item/device/pda_module/alert
+		setup_default_module = /obj/item/device/pda_module/flashlight
 		setup_drive_size = 32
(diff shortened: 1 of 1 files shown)


emmenthal (approved):
I tested it because lunch break. 
It does, indeed, work.

ju45he:
You do get a light source by default from your helmet but this is great for people with more fashionable hats.

emmenthal (changes requested):
Actually now that I am off the clock and _alert_, heh, I notice that NTSO and HoS PDAs still have the alert cartridge. If that's not to much trouble - might be apt to remedy that in this PR. Lines 131 and 140.

emmenthal (approved):

JORJ949 (approved):
It's important to note that the crisis module DOES have a use! However that use is adding a crisis button to your pda home screen which is pointless because the button in the pouch is far better

PR #25001, opened 2025-10-23
adds negative matter to miraclium material list

RSOD404:
[Materials] [Feature]

## About the PR 
miraclium can now turn into negative matter when heated.

## Why's this needed? 
the current only way to obtain negative matter is a lot of niche loafer knowledge mixed with mechcomp, and it's map restricted in certain cases due to how some maps don't even have a loafer to begin with. this solves the issue by adding an alternative way to obtaining it, albeit being more based on luck

## Testing 
image

## Changelog 

```changelog
(u)Red Screen Of Death
(+)Recent material research reports that heating miracle ore may now rarely cause it to turn into negative matter.
```


JORJ949 (approved):
Seems alright to me, negative matter doesn't get played with very often and its fun enough to see more use.

PR #25006, opened 2025-10-24
Haloperidol purges Raj

TDHooligan:
[SECURITY][BALANCE]

## About the PR 
Haloperidol now purges Rajaijah.

## Why's this needed? 

Raj is madness as a chemical, Halo cures madness. Using haloperidol on a Raj sting makes sense, but it does nothing. People (including myself) were surprised to find out that it didn't.

Since Raj is actually somewhat common in the game now, it's more sensical to allow Security to handle it properly.

## Changelog 

```changelog
(u)TDHooligan
(*)Rajaijah is now purged by Haloperidol.
```

### DIFF
--- a/code/modules/chemistry/Reagents-Medical.dm
+++ b/code/modules/chemistry/Reagents-Medical.dm
@@ -780,7 +780,7 @@ datum
 			transparency = 255
 			value = 8 // 2c + 3c + 1c + 1c + 1c
 			threshold = THRESHOLD_INIT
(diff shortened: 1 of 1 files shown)


flappybatpal (approved):

M-Earthfire:
~~With the low amount per raj sting, maybe raj would need some flushing resistance to not instantly vanish the moment haloperidol touches someone.~~

But overall a sensible change, imo.

Edit: nevermind, i thought raj sting is 5 units like neurodepressant. Its actually 15

Glamurio (approved):

RSOD404:
i thought halo already was a counter to raj,,, have i been lied to???
anyways, glad we now have a way to make madness end faster, 15u of raj lasts a LOT apparently

PR #25008, opened 2025-10-24
lets pickles be refined into pickle clumps

trazodont:
## About the PR 

adds the ability to make pickle clumps from pickles made from the food processor

## Why's this needed? 

cause sometimes i really wanna plate the station in pickle but i can't because the pickle jars got emptied out 3 rounds ago and no one's rolled this map all day so i cant do the bit when i want to

## Testing 

https://github.com/user-attachments/assets/039c29f3-1fc3-4283-ada1-dfe2b046e8e7

## Changelog 

```changelog
(u)bork
(+)pickles from the food processor can be refined into pickle clumps
```


emmenthal (approved):
Lgtm. Pickle.

PR #25010, opened 2025-10-24
Allow the AccessPro to reprogram secure crates and lockers

JORJ949:
## About the PR 
Allows the accesspro to reprogram accesses on secure crates and lockers.
To ensure this can't be used to access syndicate and other crates they couldn't give themselves access to, expanded the armory/hos access check to check for any access not in the accesspro's list of allowed accesses.

## Why's this needed? 
Allows the HoP to create custom access crates using the crate manufacturer at cargo or using it as a crime tool to add their access to other lockers without adding suspicious extra accesses to their ID.

## Testing 
Attempted to reprogram a hazard crate (science access), succeeded.
Attempted to reprogram an armory crate, failed.
Attempted to reprogram a syndicate crate, failed.

## Changelog 

```changelog
(u)JORJ949
(*)The AccessPro can now edit the access of secure crates and lockers.
```

### DIFF
--- a/code/z_adventurezones/old_bot_factory.dm
+++ b/code/z_adventurezones/old_bot_factory.dm
@@ -278,6 +278,7 @@
 	icon_greenlight = null
 	secure = 2
 	can_leghole = FALSE
+	object_flags = 0
 
 	close(var/entangleLogic, mob/user)
 		..()


flappybatpal:
Logical but I really doubt this is covering all the crates that should be safe from reprogramming?

JORJ949:
> Logical but I really doubt this is covering all the crates that should be safe from reprogramming?

The list of allowed accesses is essentially just captain level access, if you can use the accesspro you have either Personnel Management/ID computer access or Chief Engineer access. Anything not in captain level access (HoS locker, armory crates, azone lockers not tied to any station access) shouldn't be programmable by the accesspro. Other crates (like doors) I feel like are safe to allow this level of crime because its a unique command item to the HoP who has the access to just grant themselves AA anyway.

PR #25011, opened 2025-10-24
Donut 3 evidence room and second solitary

smallsandman:
[Mapping] [Security]
## About the PR 
Adds an evidence room and a second solitary confinement room to donut 3's security.
## Why's this needed? 
An evidence room would really help security and antagonists when it comes to always knowing where contraband is kept, donut 3 especially as it currently has 5 separate contraband safes all kept within view of main security making it hard for antags to find and steal the contraband they lost without getting caught, as well as providing a place for security players to create reports, sort, and give a story to collected contraband in a visually pleasing way without 10+ minutes of prep, and providing criminals an easier way to get their stuff back with a handy maintenance door connected to the room being standardized to all stations and other types of crime.

( donut 3 specific) The second solitary confinement brings it closer to other maps in terms of brig layouts, and increases use of the southern side of donut 3's security which didn't see too much use before hand in my and some others experiences. 
## Testing 
Requires #25003

image

image

image

## Changelog 

```changelog
(u)Smallsandman
(*)Added a new evidence room and second solitary brig to donut 3.
```

### DIFF
--- a/maps/donut3.dmm
+++ b/maps/donut3.dmm
@@ -13486,7 +13486,6 @@
 /obj/machinery/light/small{
 	dir = 1
 	},
-/obj/item/random_trinket_spawner,
 /turf/simulated/floor/redblack{
 	dir = 9
 	},


smallsandman:
> Changelog doesn't need the minor entry. Did you test that both the brig cells work properly and are connected up right? Check if the customs phone is still there, because the mapdiffbot is suggesting you've made changes near customs/bridge.

I checked if the chutes worked a while back i should have mentioned it, and the phone seems to be there on my end at least after merging again.
image

smallsandman (commented):
fixed the pipe and wire being under wall.

PR #25012, opened 2025-10-24
Adds the Singularity Bait artifact

Kanthes:
[SCIENCE][FEATURE]
## About the PR 
Adds the Singularity Bait artifact, which spawns a singularity in one of the corners of the map and begins pulling it towards the target. Precursor and Silicon variants will target themselves. Eldritch variants will 75% of the time target a random nearby person upon activating, and 25% of the time target the station clown.

The Singularity Bait is a one-time use artifact. It will refuse to activate on anything but the station Z-layer, and will destroy itself upon deactivation. The artifact and the singularity are linked. If one dies or leaves the Z-layer, the other is destroyed as well. If there is a human target, and the target dies or leaves the Z-layer, both artifact and singularity are destroyed as well.

The artifact will spawn two escalating station alerts much like devastating bombs, but not immediately on activation. Instead, the alert is delayed between 30-90 seconds.

The spawn location of the singularity uses Landmarks, and as such this PR also places 4 landmarks in the corners of all maps, taking precaution to avoid pathing the singularity through important structures such as the listening post or AI satellites. If a map has no dedicated landmarks, it falls back on randomizing a corner.

## Why's this needed? 

Whenever a singularity escapes containment, it's a common sight to see a flock of ghosts trying to lure the singularity towards a particular direction. This artifact creates situations where those ghosts get exactly what they want; A singularity with a target.

## Testing 

https://github.com/user-attachments/assets/83ee67dc-aad0-4d67-a3e0-f0163e2e8c66

## Changelog 

```changelog
(u)Kanthes
(*)Adds the Singularity Bait artifact, which will drag a singularity across space towards itself.. or an unfortunate soul nearby.
```

### DIFF
--- a/_std/defines/landmarks.dm
+++ b/_std/defines/landmarks.dm
@@ -60,7 +60,6 @@
 
 // misc and unused
 #define LANDMARK_CHARACTER_PREVIEW_SPAWN "character preview"
-#define LANDMARK_ARTIFACT_SINGULARITY_SPAWN "Singularity-Bait-spawn"
 
 /// Twitch Plays Shitty Bill respawn points
 #define LANDMARK_TWITCHY_BILL_RESPAWN "shitty_bill_respawn"
--- a/code/obj/landmark.dm
+++ b/code/obj/landmark.dm
@@ -575,9 +575,6 @@ var/global/list/job_start_locations = list()
 /obj/landmark/character_preview_spawn
 	name = LANDMARK_CHARACTER_PREVIEW_SPAWN
 
-/obj/landmark/singularity_bait_spawn
-	name = LANDMARK_ARTIFACT_SINGULARITY_SPAWN
-
 /obj/landmark/tutorial/flock_conversion
 	name = LANDMARK_TUTORIAL_FLOCK_CONVERSION
 	icon_state = "tutorial"
--- a/maps/atlas.dmm
+++ b/maps/atlas.dmm
@@ -22166,10 +22166,6 @@
 	},
 /turf/simulated/floor/white,
 /area/station/medical/medbay)
-"woM" = (
-/obj/landmark/singularity_bait_spawn,
-/turf/space,
-/area/space)
 "wra" = (
 /obj/cable{
 	icon_state = "4-8"
@@ -24584,7 +24580,6 @@ aaa
 aaa
 aaa
 aaa
-woM
 aaa
 aaa
 aaa
@@ -24875,7 +24870,8 @@ aaa
 aaa
 aaa
 aaa
-woM
+aaa
+aaa
 aaa
 aaa
 aaa
@@ -93127,7 +93123,7 @@ aaa
 aaa
 aaa
 aaa
-woM
+aaa
 aaa
 aaa
 aaa
@@ -112466,7 +112462,7 @@ aaa
 aaa
 aaa
 aaa
-woM
+aaa
 aaa
 aaa
 aaa
--- a/maps/clarion.dmm
+++ b/maps/clarion.dmm
@@ -3877,7 +3877,7 @@
 /turf/simulated/floor/wood,
 /area/station/chapel/sanctuary)
 "aGj" = (
-/obj/decal/boxingropeenter{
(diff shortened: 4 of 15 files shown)


code/obj/artifacts/artifact_machines/singularity_bait.dm
```diff
@@ -0,0 +1,158 @@
...
+				//Send them a creepy message.
+				boutput(src.target, SPAN_ALERT("It comes closer."))
+			//And finally, pull the singularity towards the target.
+			step_towards(src.linked_singularity, src.target)
+			playsound(O.loc, 'sound/effects/lit.ogg', 50, 1, -1)
+			src.lense.pulse()
+			T.visible_message("<b>[O]</b> pulses.")
+
+	proc/kill_singularity()
+		if(!src.linked_singularity.disposed)
```
pgmzeta:
```suggestion
		if(!QDELETED(src.linked_singularity))
```

blademane:
so if it spawns in the corners of the map does that mean theres a chance for the syndicate outpost to be eaten?

DisturbHerb:
> so if it spawns in the corners of the map does that mean theres a chance for the syndicate outpost to be eaten?

can a singularity even eat unsimulated turfii?

Glamurio:
It says the artifact will destroy itself upon deactivation. Can it still be sold for full price in QM?

JORJ949:
> > so if it spawns in the corners of the map does that mean theres a chance for the syndicate outpost to be eaten?
> 
> can a singularity even eat unsimulated turfii?

I don't think so but only the external walls of the outpost are unsimmed, the interior floors are still simulated

code/obj/artifacts/artifact_machines/singularity_bait.dm
```diff
@@ -0,0 +1,166 @@
...
+			logTheThing(LOG_STATION, null, "[src] deleted the linked singularity at [artifact.linked_singularity.x], [artifact.linked_singularity.y].")
+			artifact.kill_singularity()
+
+		. = ..()
+
+	ArtifactActivated()
+		//Lets limit this thing to the station Z-plane only.
+		if(!inonstationz(src))
+			var/turf/T = get_turf(src)
+			T.visible_message("<b>[src]</b>'s stubbornly refuses to activate.")
```
TobleroneSwordfish: Why is this a turf message?

code/obj/artifacts/artifact_machines/singularity_bait.dm
```diff
@@ -0,0 +1,166 @@
...
+			if(db_record)
+				for_by_tcl(H, /mob/living/carbon/human)
+					if(H.name == db_record["name"])
+						if(inonstationz(H))
+							src.target = H
+						break
+		if (!src.target && artitype.name == "eldritch")
+			//Find a nearby scientist and set it as target
+			var/list/valid_targets = list()
+			//TODO: Find a way to pick a random person in the same room (area?) as the artifact. Right now it cannot pick a target if it's inside an xray machine.
```
TobleroneSwordfish:
```suggestion
```

Kanthes:
> It says the artifact will destroy itself upon deactivation. Can it still be sold for full price in QM?

The artifact sells like normal, and selling it is one way of disposing of an active one.

Kyle2143:
I'm curious. Does this artifact have any effect on singularities  contained by the field generators? It'd be interesting if activating this type of artifact could potentially cause a breach if it's too close to a singularity engine. Hopefully it should be obvious that it's nearby and that things will go wrong if it stays too close to give some chance of counterplay though.

Kanthes:
> I'm curious. Does this artifact have any effect on singularities contained by the field generators? It'd be interesting if activating this type of artifact could potentially cause a breach if it's too close to a singularity engine. Hopefully it should be obvious that it's nearby and that things will go wrong if it stays too close to give some chance of counterplay though.

No effect whatsoever on existing singularities. The artifact spawns a new uncontained singularity once activated, and destroys it when deactivated or destroyed. In theory that singularity could be captured.

PR #25027, opened 2025-10-26
Add auxiliary robotics controls to Zeta

Glamurio:
[Science] [Balance]

## About the PR 
Adds a robotics control to every department which already had a Zeta area defined, and adds a Zeta area to Cogmap2 and Donut3, near Science. Ensured that all of them are protected by at most one turret (with exceptions), and that many of them are accessible by breaking in from the back of the computer. Also moves mainframe on CogMap2 from the bridge to Zeta (and replaces it with Captain's bonsai), and moves the radio from CE office to Zeta too.

## Why's this needed? 
Rogue AIs are fun and cool. But sometimes rogue AIs are really exhausting. And sometimes the law rack is gone, there's a lot of cyborgs, and it's just not possible to fix the lawrack. On a lot of maps, the killswitch is located basically right next to the laws or core, and often defended by 3-4 turrets, which is very excessive. It makes killswitching often more difficult than replacing the laws.

By standardizing a location for killswitches that's not also the upload or core, like is already precedence on some maps, this creates both a tool and a bargaining chip the crew can use more reliably.

## Testing 
Tested if the newly defined areas work. If the buttons work, etc.

Donut2:
image
Donut3:
image
Neon:
image
Nadir:
image
Clarion:
image
CogMap2:
image

## Changelog 

```changelog
(u)Glamurio (Ryou)
(*)Robotics controls are now always found near the mainframe.
```

### DIFF
--- a/maps/cogmap2.dmm
+++ b/maps/cogmap2.dmm
@@ -58055,6 +58055,7 @@
 	dir = 8
 	},
 /obj/mapping_helper/firedoor_spawn,
+/obj/mapping_helper/access/heads,
 /turf/simulated/floor,
 /area/station/turret_protected/Zeta)
 "iDr" = (
--- a/maps/donut3.dmm
+++ b/maps/donut3.dmm
@@ -46876,6 +46876,7 @@
 	id = "quarters_rd"
 	},
 /obj/mapping_helper/firedoor_spawn,
+/obj/mapping_helper/access/heads,
 /turf/simulated/floor/plating/jen,
 /area/station/turret_protected/Zeta)
 "oCe" = (


Laboredih123:
ehhhh, unsure how to feel

CelineTheYeen:
Not really a fan in general, also, the donut 3 one is just not good in my eyes both for being an incredibly out of place addition and fucking up a good breach spot for nukies and salvs and anyone else really for no reason.

pgmzeta:
I like this idea, but it will take a few passes for the bigger map changes to feel 'right'. Splitting this into individual map PRs is going to be better for avoiding merge conflicts and speed up merges for those maps with minor changes.

JORJ949 (changes requested):
I heavily reccomend doing this map by map (like smallsandman is doing for the evidence rooms), much easier to review, resolve conflicts and make changes. I agree with the general idea of making robotics control more useful though 👍

Glamurio:
> I like this idea, but it will take a few passes for the bigger map changes to feel 'right'. Splitting this into individual map PRs is going to be better for avoiding merge conflicts and speed up merges for those maps with minor changes.

> I heavily reccomend doing this map by map (like smallsandman is doing for the evidence rooms), much easier to review, resolve conflicts and make changes. I agree with the general idea of making robotics control more useful though 👍

Got it, I'll separate these when I get the chance.

PR #25047, opened 2025-10-27
Nadir evidence room

smallsandman:
[A-Mapping] [Security] 
## About the PR 
Adds an evidence room to the south side of Nadir security.
## Why's this needed? 
See #25011 
## Testing 
image

image

image

## Changelog 

```changelog
(u)Smallsandman
(*)Added an evidence room to Nadir's security.
```

### DIFF
--- a/maps/nadir.dmm
+++ b/maps/nadir.dmm
@@ -2228,6 +2228,9 @@
 /obj/machinery/light/small{
 	status = 2
 	},
+/obj/machinery/camera/directional/south{
+	pixel_x = -10
+	},
 /turf/simulated/floor/redblack,
 /area/station/security/evidence)
 "baj" = (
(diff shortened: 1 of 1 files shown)


starsylk:
It looks like there's a camera inside of the top-right wall of the evidence room.

smallsandman:
> It looks like there's a camera inside of the top-right wall of the evidence room.

fixed now. will still be in the pictures but its gone now.

JORJ949 (changes requested):
image

These wires aren't doing anything, there needs to be a "nub" to electrify the grilles.

image

Lamp and intercom are offset wrong on the wall, lamp should be same as the one directly above it and the intercom should be on the tile its coming from with the arrow pointing at the wall it will be on (you dont need to touch the offsets) like this: 
image

smallsandman:
All fixed, i believe.

Wires:
image

image

lamp and intercom:
image

JORJ949 (approved):

smallsandman:
camera problem fixed.

PR #25052, opened 2025-10-28
Improvements for the 'butt' accent

Egregorious:
## About the PR 

Converts the 'butt' and 'clack' accents into children of a 'word-replacement' accent. 

Word replacement accents attempt to preserve leading and trailing punctuation, as well as ALL-CAPS and Capitalisation of the word being replaced. 

Word selection is refactored so that each word selected has a 50% chance to be replaced, with a settable number of guaranteed replacements. The total number of words converted remains at roughly 25% on average, but is distributed in more of a curve. 

The regex captures the first part of multi-part words for replacement and treats everything after the first punctuation encountered as a trailing component. So "multi-barrel-word" becomes "butt-barrel-word", but never "multi-barrel-butt".

## Why's this needed? 

The butt accent is fun but it's a little overpowering, and it'd be nice if it preserved the exasperated tone of those cursed by it. 

## Testing 

Screenshot 2025-10-28 142753

## Changelog 

```changelog
(u)Egregorious
(+)The 'butt' accent has received some minor improvements to preserve tone.
```


ZeWaka (commented):
Not a huge fan of the less 'butt' percentage. It has pretty historical connotations:
https://soundcloud.com/stephanosrex/honk-honk-butt

Egregorious:
> Not a huge fan of the less 'butt' percentage. It has pretty historical connotations: https://soundcloud.com/stephanosrex/honk-honk-butt

To clarify, this does not half the number of butts. The previous implementation would choose a number between 1 and [half the amount of words rounded up (max 4)] and convert a random word to butt that many times, and always had the chance of converting a previously converted butt.

The new implementation chooses half the words (rounded up, max 6) uniquely and gives each a 50% chance to convert to a butt. 

I haven't done the maths but in my head it sounds more or less equivalent, both methods should - I think - be roughly 25% butt conversion on average. The important difference is just that there is a chance of no butt, and because it's a bell curve that chance is lessened the more words there are. Perhaps it could even be tweaked so it's a bimodal curve instead, so it's a butt shaped curve of butts. 

Butt I understand if the actual issue is that low-word-count speech is not guaranteed to have a butt conversion.

TobleroneSwordfish:
Every butt should contain at least one butt 👍

Egregorious:
At least one butt is always guaranteed again

PR #25058, opened 2025-10-29
Cogmap2 evidence room 2

smallsandman:
[MAPPING] [SECURITY]
## About the PR 
Adds an evidence room to Cogmap 2.
## Why's this needed? 
See #25011
## Testing 
image
image
image

Beepsky's room dimmed after image taken.

## Changelog 

```changelog
(u)Smallsandman
(*)Added an evidence room to Cogmap 2.
```

### DIFF
--- a/maps/cogmap2.dmm
+++ b/maps/cogmap2.dmm
@@ -60063,6 +60063,10 @@
 /obj/disposalpipe/segment/mail/horizontal,
 /turf/simulated/floor/plating,
 /area/station/storage/warehouse)
+"kqn" = (
+/obj/machinery/vending/cigarette,
+/turf/simulated/floor/darkblue/checker/white,
+/area/station/maintenance/inner/central)
 "kqp" = (
 /mob/living/carbon/human/npc/monkey,
 /turf/simulated/floor/pool/no_animate,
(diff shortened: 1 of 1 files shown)


TheGeneralJay:
Looks good! Only thing I note is that the added wall in the brig bathroom is just a normal one. Can't tell if this is intentional for easier breaking and entering or if it was just the wrong kind slapped down when expanding. Figured I'd point it out JUUUST in case. 

{87278127-CBFF-4751-AD83-CC82C5D35A71}

smallsandman:
RP chat asked for a way to break into the brig bathroom so i added that, with a special description, and an easter egg elsewhere

> Looks good! Only thing I note is that the added wall in the brig bathroom is just a normal one. Can't tell if this is intentional for easier breaking and entering or if it was just the wrong kind slapped down when expanding. Figured I'd point it out JUUUST in case.

TheGeneralJay:
> RP chat asked for a way to break into the brig bathroom so i added that, with a special description, and an easter egg elsewhere
> 
> 
> 
> > Looks good! Only thing I note is that the added wall in the brig bathroom is just a normal one. Can't tell if this is intentional for easier breaking and entering or if it was just the wrong kind slapped down when expanding. Figured I'd point it out JUUUST in case.
> 
> 

Gotcha! Just making sure lmao

JORJ949 (changes requested):
image

Don't think you meant to change this wall up in the pool, another removing customs phones on this master merge aswell.

image

These maintenance tunnels should be wider than 1 tile, there are going to be chases down here there needs to be dodgeable space

image

This chute also throws out loaves the chef sends down, I'd replace a couple of the beds with the vending machines and leave the table alone.


Generally not a fan of how far this one is from the solitary cells considering those are what's used more often, I'd rather you replace the security hangar there considering they already have two other hangars (below bridge and main podbay), this way its also close to the interrogation room and most of the brig processing area.

smallsandman:
i decided to just redo the area closer to the mini brigs but not overtaking podbay, i swapped it with beepsky's house.

PR #25062, opened 2025-10-29
Makes scrap weapons + nanofab weapons contraband 4

ThatFiveGuys:
## About the PR 
Any scrap weapon (not their components), and the spear is contraband 4. This means that holding them in your hand will now set off security alarms, trigger beepsky, and show the contraband sechud indicator. The bow is only contraband 3.

## Why's this needed? 
Nanotrasen probably doesn't want their employees carrying weapons everywhere. Thematically it makes sense you'd have to hide them a little bit in your backpack or something.

## Testing 
I just opened goonstation and grabbed the items I changed. 

## Changelog 

```changelog
(u)Ikea
(*)Due to caveman rumors Nanotrasen has now decided to make scrap weapons and the spear contraband level 4, and the bow contraband level 3.

### DIFF
--- a/code/WorkInProgress/MarqStuff.dm
+++ b/code/WorkInProgress/MarqStuff.dm
@@ -649,7 +649,7 @@
 	spread_angle = 40
 	force = 5
 	can_dual_wield = 0
-	contraband = 0
+	contraband = 3
 	move_triggered = 1
 	var/spread_base = 40
 	var/max_draw = 3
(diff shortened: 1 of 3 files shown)


M-Earthfire:
I feel they should fall under firearm permit, but beyond that i am fully supporting this.

ThatFiveGuys:
> I feel they should fall under firearm permit, but beyond that i am fully supporting this.

I think the idea of making firearms permit into a generic weapons permit is a good one however I consider it outside of scope of this pr because contraband code would have to be rewritten to untie it to guns. Also weapons like c-saber and butchers knife would have to be moved over to it

TobleroneSwordfish (approved):

PR #25063, opened 2025-10-29
Mild buff to scrap weapons

ThatFiveGuys:
## About the PR 
All scrap weapons have had their base weapon damage increased by 1, and the plasmaglass variants have had their damage increased by 2.
## Why's this needed? 
The makeshift weapons feel a bit weak compared to standard station equipment which is far more available. This makes them a bit stronger. Also widens the gap between the normal and plasmaglass variants because plasmaglass is way harder to obtain.

## Testing 
I beat up some monkeys with a spear, felt fine? Hard to test combat changes by yourself though.

## Changelog 

```changelog
(u)Ikea
(*)Makeshift weapons have had their damage slightly buffed.
```

### DIFF
--- a/code/obj/item/scrapweapons.dm
+++ b/code/obj/item/scrapweapons.dm
@@ -142,7 +142,7 @@ ABSTRACT_TYPE(/obj/item/scrapweapons/weapons)
 	item_state = "spear"
 	w_class = W_CLASS_NORMAL
 	hit_type = DAMAGE_BLUNT
-	force = 6
+	force = 7
 	throwforce = 10
 	custom_suicide = 1
 	attack_verbs = "impales"


TobleroneSwordfish:
Really not sure these things need a buff, they're already too prevalent as it is.

ThatFiveGuys:
> Really not sure these things need a buff, they're already too prevalent as it is.

Prevalence seems mostly to be a cool factor thing instead of a practicality thing, hopefully contraband changes reduce that, I'd agree this isn't a needed change but right now they just seem to be roughly equivalent to top tier common station weapons like welders or extinguishers and the plasmaglass variants are barely any different. This just gives it a little push to where from a damage perspective they're mildly worth going for without a doubt

PastaSpaghett:
With the contra change for these I think this buff is warranted.

TobleroneSwordfish:
Devchat opinion is "meh"

PR #25066, opened 2025-10-29
If a hotspot is pinned it will change color on the trench map

phoiii:
[QOL] [ENGINEERING]
## About the PR 
This will update the trench map to show wether or not a hotspot is pinned with a different color. Also makes the trench map update whenever a hotspot is stomped
## Why's this needed? 
There is virtually no feedback on if a hotspot is pinned or not
## Testing 

image

## Changelog 

```changelog
(u)Phoi
(+)Changes trench map to update hotspot color based on wether it is pinned or not 
(+)Adds pinned hotspot color to key
```

### DIFF
--- a/code/datums/controllers/sea_hotspot_controls.dm
+++ b/code/datums/controllers/sea_hotspot_controls.dm
@@ -91,12 +91,15 @@
 		var/list/hotspots = list()
 
 		for (var/datum/sea_hotspot/S in hotspot_groups)
-			if (S.can_drift == 1) // draws the hotspots on the trench map and changes color if it is pinned
-				S.hotspot_color = "hotspot"
-			else
-				S.hotspot_color = "pinned_hotspot"
 
-			hotspots += {"<div class='[S.hotspot_color]' style='bottom: [S.center.y * 2]px; left: [S.center.x * 2]px; width: [S.radius * 4 + 2]px; height: [S.radius * 4 + 2]px; margin-left: -[S.radius * 2]px; margin-bottom: -[S.radius * 2]px;'></div>"}
+			var/hotspot_color
+
+			if (S.can_drift == 1) // draws the hotspots on the trench map and changes color if it is pinned
+				hotspot_color = "hotspot"
+			else
+				hotspot_color = "pinned_hotspot"
+
+			hotspots += {"<div class='[hotspot_color]' style='bottom: [S.center.y * 2]px; left: [S.center.x * 2]px; width: [S.radius * 4 + 2]px; height: [S.radius * 4 + 2]px; margin-left: -[S.radius * 2]px; margin-bottom: -[S.radius * 2]px;'></div>"}
 
 		src.map_html = {"
 <!doctype html>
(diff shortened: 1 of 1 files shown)


code/datums/controllers/sea_hotspot_controls.dm
```diff
@@ -90,7 +90,10 @@
 
 		var/list/hotspots = list()
 		for (var/datum/sea_hotspot/S in hotspot_groups)
-			hotspots += {"<div class='hotspot' style='bottom: [S.center.y * 2]px; left: [S.center.x * 2]px; width: [S.radius * 4 + 2]px; height: [S.radius * 4 + 2]px; margin-left: -[S.radius * 2]px; margin-bottom: -[S.radius * 2]px;'></div>"}
+			if (S.can_drift == 1) // draws the hotspots on the trench map and changes color if it is pinned
+				hotspots += {"<div class='hotspot' style='bottom: [S.center.y * 2]px; left: [S.center.x * 2]px; width: [S.radius * 4 + 2]px; height: [S.radius * 4 + 2]px; margin-left: -[S.radius * 2]px; margin-bottom: -[S.radius * 2]px;'></div>"}
+			else
+				hotspots += {"<div class='pinned_hotspot' style='bottom: [S.center.y * 2]px; left: [S.center.x * 2]px; width: [S.radius * 4 + 2]px; height: [S.radius * 4 + 2]px; margin-left: -[S.radius * 2]px; margin-bottom: -[S.radius * 2]px;'></div>"}
```
TobleroneSwordfish: Why branch here instead of doing it inline like the others? You can extract the class as a `var/hotspot_class = ` if you don't want to shove a ternary in there.

phoiii: I had no knowledge of HTML before looking into this. so this was the easiest way i found to do it.

TobleroneSwordfish: You don't need to know any HTML, just replace the `class='hotspot'` with `class='[hotspot_class]'` and define that variable above, that way you aren't duplicating the entire line just to change one parameter.

phoiii: alright i think i got it, thanks for recommendation

code/datums/controllers/sea_hotspot_controls.dm
```diff
@@ -291,6 +303,8 @@
 
 	var/last_colorping = 0
 
+	var/hotspot_color
```
TobleroneSwordfish: No need to store it on the datum itself, just define a local var inside the loop.

TobleroneSwordfish (changes requested):

code/datums/controllers/sea_hotspot_controls.dm
phoiii: got it working. was trying to make it like the other var in the draw line but i see how it being local would be better.

TobleroneSwordfish (approved):

PR #25074, opened 2025-10-30
Adds a wrench to the science manufacturer and makes the tools purple

ThatFiveGuys:
## About the PR 
This pr adds a purple wrench to the science manufacturer, also makes all the tools in that manufacturer purple.
## Why's this needed? 
Wrenches are critical for toxins, it feels weird science can't print it on their own right now. Also makes them all purple because we already have a purple crowbar and this brings it in line with the engfab
## Testing 
Checked to make sure they were all in the manufacturer, they were.
image

## Changelog 

```changelog
(u)Ikea
(+)The science manufacturer can now print purple tools alongside a purple wrench.
```


itchybones:
Hell yeah, finally, nerd wrenches

Glamurio:
Minor complaint, but if the wrench is purple, then the welding tool and crowbar should be too. Or the wrench should be green.

ThatFiveGuys:
> Minor complaint, but if the wrench is purple, then the welding tool and crowbar should be too. Or the wrench should be green.

Made them all purple because why not

TobleroneSwordfish (approved):

PR #25079, opened 2025-10-31
Adds cashless trait

ThatFiveGuys:
## About the PR 
This adds the cashless trait, you spawn with no cash in your inventory. You still recieve funds in the atm as normal.
## Why's this needed? 
Cash is oftentimes just inventory bloat, this gives you the option to have none of it.

## Testing 
Tested with and without trait and with pawnstar, behaves as expected
image
image

## Changelog 

```changelog
(u)Ikea
(*)There is now a cashless trait which stops you from spawning with cash in your inventory.
```


DisturbHerb:
Hot take, this should be default behaviour, literally everybody can get cash whenever they'd like because of PDAs now

RSOD404:
> Hot take, this should be default behaviour, literally everybody can get cash whenever they'd like because of PDAs now

buh.. buh how am i gonna get that extra start-of-shift money...

xxfishshadowxx:
> Hot take, this should be default behaviour, literally everybody can get cash whenever they'd like because of PDAs now

Money is already hard to accumulate for anyone who isnt hop/botanist/QM, it doesnt need to be reduced across the board even more so 👿 👿

ThatFiveGuys:
Presumably such a change would come with an increase in the money you have at the start of a round. Not going to be pring that though just because I don't want to deal with that right now.

DisturbHerb:
> on the subject of potentially reducing starting pay for every role if we get rid of starting cash

yeah, this would entail just putting that cash in everyone's bank accounts shift start. this would just free up the inventory slot

edit: the code that spawns cash on you would be really easy to excise and replace with the required starting bonus for each job role, `code\procs\jobprocs.dm:814-827`
```js
if (JOB.wages > 0)
    var/cashModifier = 1
    if (src.traitHolder && src.traitHolder.hasTrait("pawnstar"))
	    cashModifier = 1.25
    
    var/obj/item/currency/spacecash/S = new /obj/item/currency/spacecash
    S.setup(src,round(JOB.wages * cashModifier))
    
    if (isnull(src.get_slot(SLOT_R_STORE)))
	    src.equip_if_possible(S, SLOT_R_STORE)
    else if (isnull(src.get_slot(SLOT_L_STORE)))
	    src.equip_if_possible(S, SLOT_L_STORE)
    else
	    src.equip_if_possible(S, SLOT_IN_BACKPACK)
```

Zhordyn:
Every round start I'm doing ID -> PDA, then cash -> PDA to clear up my inventory. The PR to remove cash and then PR to increase starting cash would be QoL most could get behind.

JORJ949 (changes requested):
I'd rather this be an option like use satchel is like "Store starting cash in bank" (but probably with a better name) so new players have cash seperate but experienced players dont need to add a nothing trait to all their characters.

Glamurio:
I kinda would like a trait that starts you in Debt and you have to pay off your debt, so any money you get from your job gets subtracted first

ZeWaka:
> I kinda would like a trait that starts you in Debt and you have to pay off your debt, so any money you get from your job gets subtracted first

Student Loans

PR #25083, opened 2025-10-31
Satchels jumble items

Egregorious:
## About the PR 

Whenever an item is placed into a satchel, it gets given a random offset. This means that piles of items are jumbled up when picked up in a satchel and dumped out again. 

Added RANDOMIZE_PIXEL_OFFSET macro for randomly offsetting an item, it's used by satchels and `place_on`.

## Why's this needed? 

Just helps with immersion a little, it's always bugged me that you get the exact same-shaped pile of items when you pick them up and dump them out again. 

## Testing 

https://github.com/user-attachments/assets/67bab80a-9bb1-4f7b-bbc3-2b573f702cd3

### DIFF
--- a/code/obj.dm
+++ b/code/obj.dm
@@ -362,8 +362,7 @@
 			W.set_dir(dirbuffer)
 		W.set_loc(src.loc)
 		if (imprecise) // place item imprecisely by randomising offset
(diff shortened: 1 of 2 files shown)


Gannets:
Jumble is such a good word

Munien7:
I love this so much, such a small thing that has annoyed me for so long

PR #25092, opened 2025-11-02
Fix macros for lazily accessing associated lists

Egregorious:
## About the PR 

Converts LAZYLISTACCESSASSOC macro to work for non-nested associated lists.

Added LAZYLISTACCESSASSOCDEEP macro for accessing nested associated lists. 

The existing macro doesn't appear to be used anywhere outside the secret module, so unless it's used by something there this shouldn't affect anything. 

## Why's this needed? 

LAZYLISTACCESSASSOC didn't work for non-nested associated lists.

## Testing 

Unit tested it a bit

Screenshot 2025-11-02 132724


_std/macros/list.dm
```diff
@@ -38,5 +38,7 @@
 #define LAZYLISTREMOVEASSOC(L, K, V) if(L) { if(L[K]) { L[K] -= V; if(!length(L[K])) L -= K; } if(!length(L)) L = null; }
 
 /// Accesses an associative list, returns null if nothing is found
-#define LAZYLISTACCESSASSOC(L, I, K) L ? L[I] ? L[I][K] ? L[I][K] : null : null : null
+#define LAZYLISTACCESSASSOC(L, K) (L ? L[K] : null)
```
ZeWaka, line 41: https://github.com/tgstation/tgstation/blob/923b4545d8f00c611c57c23e5f16f5d20ec4ec42/code/__HELPERS/_lists.dm#L84

this is what tg uses

Egregorious: As far as I can tell by using a github search, they only use it for single-nested associative lists, which seems to make sense because it's the only context in which it works. It cannot be used to fetch a value from a non-nested associative list, it'll just runtime, and it doesn't go deeper than that.

I may be missing something but I also don't see why there's a second 'L[I][K]' there; it doesn't seem to be doing anything, and just repeats a check that's already happened. Maybe it's for some sort of edge-case but it's not documented on either codebase and I can't think of what it could be.

PR #25099, opened 2025-11-02
Removes panic buttons from secoff etc pouches

TobleroneSwordfish:
## About the PR 
See title, leaves the big box of them in security. Also reduces pouch capacities by 1 so they're still full.
## Why's this needed? 
Further attempt to reduce security swarm effect. Leaving them in security lets people pick them up if they think they'll need them, before going into danger, or hand them out to VIPs who need protection.
## Testing 
image

## Changelog 

```changelog
(u)LeahTheTech
(+)Security roles no longer start with a panic button in their pouch.
```

### DIFF
--- a/code/obj/item/storage/ammo_pouches.dm
+++ b/code/obj/item/storage/ammo_pouches.dm
@@ -212,7 +212,7 @@
 	icon_state = "ammopouch-ntsc"
 	health = 6
 	w_class = W_CLASS_SMALL
-	slots = 4
+	slots = 5
 	opens_if_worn = TRUE
 	prevent_holding = list(/obj/item/storage)
 	spawn_contents = list(/obj/item/handcuffs/ = 1,


Mister-Moriarty (changes requested):
I'd advise against removing the associated extra storage slot; inventory management is already a big deal for Security.

DisturbHerb:
pouch size reduction might be an even bigger change than just the crisis alert doohickey

JORJ949:
The NTSC has managed with less slots and more gear so I'm sure officers can manage

TobleroneSwordfish:
image

Reducing the number of slots reinforces it as an active choice to pick up a panic button and frankly security don't need to be able to carry every possible doohicky on them at all times.

Flaborized:
Talked about a bit in admin chat before this change was made, I still am a fan of the *alternative* version of the change that completely cuts the sec buttons with the exception of giving one to the HoP, objectively the funniest person who could have it.

I mostly think the buttons (and the alert app as well but less so) take away some of the interest in communication between departments. Like the *radio* exists so you can use words to get people to come to you, and if you only have time to say ";HELP" over comms, at least the AI can track you, which is good because it involves more people in this process and makes communication more of a mechanic you have to interface with instead of a button that just does it for you.

Sec pouch change doesn't bother me and I added the pouches, they're there to add a Cool Bonus to playing sec and it's fine at 5 slots or 6 imo, negligible difference.

CelineTheYeen:
Removing the buttons from pouches is nice since it's weird to both have one and the box in sec but I just feel like the pouch size thing should've been atomized.

xxfishshadowxx:
> Removing the buttons from pouches is nice since it's weird to both have one and the box in sec but I just feel like the pouch size thing should've been atomized.

i think it makes sense to be in the same PR, the pouch size part is what makes the balance change have meaning as you cant just bypass the PR entirely by going and fetching a button

Calphrick:
> i think it makes sense to be in the same PR, the pouch size part is what makes the balance change have meaning as you cant just bypass the PR entirely by going and fetching a button

I mean, you can just eat the donut. Space problem solved, put a panic button in.

ThatFiveGuys:
With both the removal of the crisis alert pda button and this I feel like the norm will shift back to solely using the headset when in danger. I really don't like this, especially on roleplay the swarm effect was as prevalent before the addition of these things, this change means that the only difference is that the antag doesn't get any notification they're about to be swarmed besides a small easily missable white text in the corner of their screen.

TobleroneSwordfish:
Mmm interesting point but radio chat does have maptext now so I think that's less of an issue

JORJ949:
Secure radio frequencies also have a unique sound iirc and if you hear the radio sound coming from the secoff and don't see their message on the common radio you can figure they're calling for backup

sippykot:
massive nerf that's not even in the changelog?

PR #25101, opened 2025-11-02
Health implants only send PDA alerts to medical/chaplain

pgmzeta:
[medical][station systems][balance]
## About the PR 

The security mailgroup is no longer alerted when security health implants trigger a death alert.

Instead, the death alerts add a prefix to the message based on the type of health implant:
* Printable health implants say "CREW"
* Security implants say "SECURITY"
* Command members that recieve health implants by default (Captain/HoS) now get a new implant sub-type that says "COMMAND"

The report also strips the clone scan location, just noting whether they had a cloning implant or not.

## Why's this needed? 

Encourage/require more communication between medical and security teams to deal with death on station.

## Testing 
Screenshot 2025-11-02 111517
Screenshot 2025-11-02 111529

## Changelog 

```changelog
(u)glowbold
(*)Death alerts only send messages to the medical and spiritualaffairs mailgroups and include the type of health implant that triggered the alert.
(+)Death alerts only indicate if the person has a clone implant, not the location of the cloner. Rejoice, mindhackers!
```

### DIFF
--- a/code/datums/jobs/jobs_command.dm
+++ b/code/datums/jobs/jobs_command.dm
@@ -25,7 +25,7 @@ ABSTRACT_TYPE(/datum/job/command)
 	receives_miranda = TRUE
 	can_roll_antag = FALSE
 	world_announce_priority = ANNOUNCE_ORDER_CAPTAIN
-	receives_implants = list(/obj/item/implant/health/security/anti_mindhack)
+	receives_implants = list(/obj/item/implant/health/security/anti_mindhack/command)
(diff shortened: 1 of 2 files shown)


JORJ949:
Unmentioned change: You will no longer get a list of where they were scanned, rather, just that they were scanned, which is good because mindhack cloner buff.

pgmzeta:
> Unmentioned change: You will no longer get a list of where they were scanned, rather, just that they were scanned, which is good because mindhack cloner buff.

whoops, forgot about that. added to the description

PR #25102, opened 2025-11-02
Custom Ghost Spawn panel ignores special spawn locations

JORJ949:
## About the PR 
The custom ghost spawn panel will now ignore a job's special spawn location.
Selecting a job with a special spawn location while not having that landmark selected will add a warning button that can be clicked to set the spawn location to that job's default.

## Why's this needed? 
Allows spawning any job with a special spawn location (Syndies, Radio host, Actors, Influencers) at a place that isn't their landmarks.

## Testing 
image

## Changelog 

Probably an admin changelog of some sort


TobleroneSwordfish:
I meant to do this for ages, thanks Jorj!

PR #25103, opened 2025-11-03
Rework deep fryer timing

Egregorious:
## About the PR 

The cooking time of the deep fryer is now based on the game clock, and is more state driven.

This effectively reduces the time needed to fry items to varying levels to at least 1/4 of what it is currently. The time needed to go from lightly-fried to deep-fried is reduced from at least 2 minutes, to 30 seconds. It takes a further 30 seconds to burn and a total of 2 minutes to begin to flashfire. 

Added the burn-sizzle sound effect when the food get burnt. 

## Why's this needed? 

The previous implementation counts by ticks, which means that frying takes much longer the heavier the load the server is under, and is affected whenever the length of a machine-tick is altered. 

I don't know the historic timing associated with the deep fryer but the numbers make it look as if it was set up expecting a tick to be 1-second long. They now seem to be at least 4-5 seconds long, so the numbers have been adjusted assuming I'm correct about the original intent. 

It takes so long to do anything but lightly-fry items right now, and since it's purely a cosmetic difference between light-fry, fry and deep-fry, speeding it up seems fine to me. 

## Testing 

Fried a bunch of items for varying lengths of time, and observed the correct amount of time was taken to appropriately fry them. 

https://github.com/user-attachments/assets/ee342e36-c4b5-475a-a125-6dadef9980fd

## Changelog 

```changelog
(u)Egregorious
(+)Deep fryers now cook much faster and more consistently, and are largely unaffected by the stresses placed upon the universe. 
```

### DIFF
--- a/code/modules/cooking/deep_fryer.dm
+++ b/code/modules/cooking/deep_fryer.dm
@@ -2,6 +2,12 @@ TYPEINFO(/obj/machinery/deep_fryer)
 	mats = 20
 	start_speech_outputs = list(SPEECH_OUTPUT_SPOKEN_DEVICE)
 
+
+#define COOK_STATE_LIGHTLY_FRIED 1
+#define COOK_STATE_FRIED 2
+#define COOK_STATE_DEEP_FRIED 3
+#define COOK_STATE_BURNED 4
+
 /obj/machinery/deep_fryer
 	name = "Deep Fryer"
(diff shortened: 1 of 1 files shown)


flappybatpal (approved):
Makes sense to me, it can take forever compared to anything else in the kitchen.

emmenthal (approved):
Yeah, fryers are almost unusable as is. This is good.

PR #25105, opened 2025-11-03
Handwritten Postit Notes

Zhordyn:
[A-UI][C-Feature]

## About the PR 
Every mind has a random handwriting style, this PR re-enables that variable to effect user's handwriting when written on post its. Also changes how postit note text length is calculated so only the raw text characters count and none of the note formatting is counted.

A separate PR #25106 does the same but with smaller pre-installed font list.

Also adds a warning to a defunct postit obj.

## Why's this needed? 
Improves roleplay in that a users handwriting can be checked against notes left around station (e.g. a changeling can change fingerprints, but can't change their handwriting style) and improve some interrogation routes. This PR would allow using handwriting styles elsewhere, like graffiti and specific instruments on paper.

Instead of loading 90+ fonts and their license attributions, the PR adds a reference to fonts.googleapis.com to pull the needed styles into Chat and Tooltip HTMLs. 

## Testing 
Testing chat and tooltip generation, there is a slight blip as the font loads in.

Chat Example

Chat Example 2

ToolTipCapture

## Changelog 

```changelog
(u)Zhordyn
(*)Writing on post-its now uses a distinct random handwriting style from 90+. 
```

### DIFF
--- a/code/obj/item/stickers.dm
+++ b/code/obj/item/stickers.dm
@@ -196,7 +196,7 @@
 			if (pen.uses_handwriting && user?.mind?.handwriting)
 				src.font = user.mind.handwriting
 			else if (pen.font)
-				src.font = pen.font
+				src.font = null
 			if (src.icon_state == initial(src.icon_state))
 				var/search_t = lowertext(t)
 				if (copytext(search_t, -1) == "?")
@@ -205,7 +205,7 @@
 					src.icon_state = "postit-excl"
 				else
 					src.icon_state = "postit-writing"
-			src.words += "[src.words ? "<br>" : ""]<span style='font-family:[src.font];'>[t]"
+			src.words += "[src.words ? "<br>" : ""]<span style='font-family:[src.font];font-size:16px;'>[t]"
 			tooltip_rebuild = TRUE
 			pen.in_use = 0
 			src.add_fingerprint(user)


RSOD404:
i like the idea a lot, but i'm a bit worried some fonts might come off as unreadable without having to copy-paste it elsewhere to see the text without the font

Zhordyn:
> i like the idea a lot, but i'm a bit worried some fonts might come off as unreadable without having to copy-paste it elsewhere to see the text without the font

Agreed, I didn't make the original font list in global.dm. I have an output file that repeats the same text string referencing all the fonts, and already earmarked a few that aren't legible/compatible and looking at a few more to add to the original list.

Zhordyn:
I took a stab at pruning the list of less legible fonts (they would only display at 14px) and added a couple more fonts. [Here is a link to a new list of fonts as they would display in the chat window](https://fuchsia-edeline-14.tiiny.site/Chatwindow-Output_newlist.html) I colored the names orange `#FFA500` for ones I think are on the cusp of being culled.

Fonts I plan to remove from list and aren't in the above example window
- Bilbo Swash Caps
- Bilbo
- Clicker Script
- Cookie
- Devonshire
- Grand Hotel
- Great Vibes
- Herr Von Muellerhoff
- Just Another Hand
- Just Me Again Down Here
- Kristi
- Lovers Quarrel
- Molle
- Mr Dafoe
- Mr De Haviland
- Mrs Saint Delafield
- Norican
- Qwigley
- Rouge Script
- Sacramento
- Stalemate
- Sunshiney
- Tangerine
- Yesteryear

browserassets/src/html/browserOutput.html
```diff
@@ -5,6 +5,7 @@
...
...ily=Italianno&family=Julee&family=Just+Another+Hand&family=Just+Me+Again+Down+Here&family=Kalam&family=Kaushan+Script&family=Kristi&family=La+Belle+Aurore&family=Leckerli+One&family=Lobster+Two&family=Lobster&family=Loved+by+the+King&family=Lovers+Quarrel&family=Marck+Script&family=Meddon&family=Merienda+One&family=Merienda&family=Molle&family=Montez&family=Mr+Dafoe&family=Mr+De+Haviland&family=Mrs+Saint+Delafield&family=Neucha&family=Niconne&family=Norican&family=Nothing+You+Could+Do&family=Over+the+Rainbow&family=Pacifico&family=Parisienne&family=Patrick+Hand+SC&family=Patrick+Hand&family=Petit+Formal+Script&family=Pinyon+Script&family=Playball&family=Quintessential&family=Qwigley&family=Rancho&family=Redressed&family=Reenie+Beanie&family=Rochester&family=Rock+Salt&family=Rouge+Script&family=Sacramento&family=Satisfy&family=Schoolbell&family=Shadows+Into+Light+Two&family=Shadows+Into+Light&family=Short+Stack&family=Sofia&family=Stalemate&family=Sue+Ellen+Francisco&family=Sunshiney&family=Swanky+and+Moo+Moo&family=Tangerine&family=The+Girl+Next+Door&family=Unkempt&family=Vibur&family=Waiting+for+the+Sunrise&family=Walter+Turncoat&family=Yellowtail&family=Yesteryear&family=Zeyada">
```
emmenthal, line 8: I have no experience with web dev conventions, but feels like a font list should be human-readable. Can this be split into multiple lines?

Zhordyn: The human readable part exists in the code already, it's at [line 149 of global.dm](https://github.com/goonstation/goonstation/blob/065181e51e648b4e255911bed6e8f2fa9eed0277/code/global.dm#L149), the added ref link just now pulls the font family from googles API instead of relying on the font already being installed.

PR #25106, opened 2025-11-03
Handwritten Postits Notes - Smaller

Zhordyn:
[A-UI][C-Feature]

## About the PR 
Every mind has a random handwriting style, this PR re-enables that variable to effect user's handwriting when written on post its, but unlike [PR #25105](https://github.com/goonstation/goonstation/pull/25105) only looks to fonts saved on users PC. Currently there are about 4 by default that appear like handwriting.

Also changes how postit note text length is calculated so only the raw text characters count and none of the note formatting is counted.

Also adds a warning to a defunct postit obj.

## Why's this needed? 
Improves roleplay in that a users handwriting can be checked against notes left around station (e.g. a changeling can change fingerprints, but can't change their handwriting style) and improve some interrogation routes. This PR would allow using handwriting styles elsewhere, like graffiti and specific instruments on paper.

## Testing 
Testing chat and tooltip generation.
ChatCapture_Limited

ToolTipCapture_Limited

## Changelog 

```changelog
(u)Zhordyn
(*)Writing on post-its now uses a distinct random handwriting style. 
```

### DIFF
--- a/code/global.dm
+++ b/code/global.dm
@@ -146,8 +146,7 @@ var/global
 	"Black" = "#935D37",
 	"Dark" = "#483728")
 
-	list/handwriting_styles = list("Dancing Script",
-	"Ink Free",
+	list/handwriting_styles = list("Ink Free",
 	"MV Boli",
 	"Segoe Print",
 	"Segoe Script")


code/obj/item/stickers.dm
```diff
@@ -131,6 +131,8 @@
 	dont_make_an_overlay = 1
 	vis_flags = VIS_INHERIT_PLANE | VIS_INHERIT_LAYER
 	var/words = ""
+	var/words_length = 0
```
ZeWaka, line 134: i don't see why you added this

PR #25111, opened 2025-11-04
Removes starting cash, starting bonus paid to bank account instead

DisturbHerb:
## About the PR 

This PR is intended as an alternative to #25079.

Starting cash no longer spawns in a player's inventory. The starting bonus you would usually be paid, including the 25% Pawn Star bonus, now starts in the player's bank account. Pens and water bottles may still spawn randomly in your right pocket if possible only if your job is not paid any wage, as per prior behaviour.

Pawn Stars trait description updated to match behaviour.

## Why's this needed? 

As an alternative to #25079, this PR would not necessitate any player input to get rid of starting cash on a given character slot as it'd be fully automatic. Access to cash wouldn't be impeded for those who have access to a bank account as withdrawing physical cash is trivial with a PDA.

At present, I anticipate that the biggest pinch point would be for new users who aren't as familiar with using card-based payments. Cash on its own presents challenges due to it requiring two hands to separate stacks, so I don't currently believe that this would be too much of an issue.

## Testing 

Tested by loading in a character with a bank account and comparing money on card to job wages. Confirmed to work with Pawn Star's 25% bonus.

## Changelog 

```changelog
(u)DisturbHerb
(*)HR has finally discovered how to pay employees' starting bonuses into their bank accounts instead of giving them untraceable physical cash.
```

### DIFF
--- a/code/datums/traits.dm
+++ b/code/datums/traits.dm
@@ -638,7 +638,7 @@
 
 /datum/trait/pawnstar
 	name = "Pawn Star"
-	desc = "You sold your trinket before you departed for the station. You start with a bonus of 25% of your starting cash in your inventory."
+	desc = "You sold your trinket before you departed for the station. You start with a bonus of 25% of your starting money in your bank account."
 	id = "pawnstar"
 	icon_state = "pawnP"
 	points = 0
--- a/code/procs/jobprocs.dm
+++ b/code/procs/jobprocs.dm
@@ -812,19 +812,10 @@ Equip items from body traits.
 	src.mind?.remembered_pin = C.pin
 
 	if (JOB.wages > 0)
-		var/cashModifier = 1
-		if (src.traitHolder && src.traitHolder.hasTrait("pawnstar"))
-			cashModifier = 1.25
-
-		var/obj/item/currency/spacecash/S = new /obj/item/currency/spacecash
-		S.setup(src,round(JOB.wages * cashModifier))
-
-		if (isnull(src.get_slot(SLOT_R_STORE)))
-			src.equip_if_possible(S, SLOT_R_STORE)
-		else if (isnull(src.get_slot(SLOT_L_STORE)))
-			src.equip_if_possible(S, SLOT_L_STORE)
-		else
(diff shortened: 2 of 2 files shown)


Valtsu0:
I'd prefer if the money started in your bank account only if you started with a PDA

DisturbHerb:
> I'd prefer if the money started in your bank account only if you started with a PDA

The use-cases where you'd start without a PDA but you have an ID and absolutely need to pay for something with cash without access to an ATM seem uncommon enough to where I've not personally had it on my radar. Counterexamples highly welcomed.

Connie-Dun:
I don't like this at all.

Starting cash is a very good thing for starting players for a variety of reasons. It quickly communicates the existence of money in a way that's much more immediate than just seeing vending machines and the like does. It makes it quick and easy to show the different amounts of starting money different jobs get. It shows that money, in fact, can be in cash form. It immediately provides an item that's useless enough to toss when ending up in space.

I don't think getting rid of all of that is worth saving two clicks at roundstart; #25079 being opt-in as a trait both makes all of that new player experience unimpacted and actually helps add another surface to starting players thinking about money, as they go through the traits list for the first time.

sippykot:
People actually using physical money on goon is soulful and unique (on most other servers its purely cards and holochips). This kinda shoots it in the leg for no real benefit and will make people use cash way less.

Valtsu0:
> People actually using physical money on goon is soulful and unique (on most other servers its purely cards and holochips). This kinda shoots it in the leg for no real benefit and will make people use cash way less.

I don't think it will make people use cash less. People put the cash in their when they spawn anyways

DisturbHerb:
> People actually using physical money on goon is soulful and unique (on most other servers its purely cards and holochips). This kinda shoots it in the leg for no real benefit and will make people use cash way less.

Cash already works like holochips here in that you can eject some from a PDA.

TobleroneSwordfish:
Ye I don't like this very much (for the same reasons outlined above), I think Jorj's suggestion on the other PR is the best solution: https://github.com/goonstation/goonstation/pull/25079#pullrequestreview-3414172821

PR #25113, opened 2025-11-04
A lesser slug may be microwaved to create an adult slug

CalliopeSoups:
[Critters] [QoL]
## About the PR 
This PR adds a bit of code to microwaves that allow for adding lesser slugs. In a manner similar to marshmellows, the lesser slug expands to become a full slug. Only one can be placed inside at a time, and it always dirties the microwave.

I thought of ways to hint at this interaction by doing things like making lesser slugs leap towards microwaves sometimes, but decided on simply mentioning them yearning for the microwave in their description. I can try this again if that's deemed too straightforward though I imagine this would come with a recipe listed on the wiki anyway.

## Why's this needed? 
With the slug burger, useful slime and potentially more slug based recipies in the future, chefs need a way to source their slugs in a way that does not chain a poor angler to the rng dimension.

I thought about making remaking slugs to work with ranch code for slug breeding, but marshmellow expansion slug seemed funnier to me. 

## Testing 
I've tested this a few times to make sure only one slug can fit inside at a time etc. It all seems to work, though I want to test what would happen if another item or another lesser slug ended up in the contents through weird means that I don't see happening, just incase. Var editing another object to the contents list confused me though so I'd appreciate being told how that works or just having someone test it for me.

dreamseeker_liZtXqNvA8

## Changelog 

```changelog
(u)CalliopeSoups
(+)Lesser slugs can be placed in the microwave to promote healthy growth!
```

### DIFF
--- a/code/modules/cooking/microwave.dm
+++ b/code/modules/cooking/microwave.dm
@@ -434,7 +434,7 @@ TYPEINFO(/obj/machinery/microwave)
 				if (src.slug)
 					src.icon_state = "mwbloodyS"
 					var/mob/adultSlug = new /mob/living/critter/small_animal/slug
-					if (rand(6))
+					if (prob(6))
 						src.visible_message(SPAN_NOTICE("Nature is beautiful."))
 					adultSlug.set_loc(get_turf(src))
 				else


DisturbHerb:
microwave code 😶

RSOD404:
slurgers are a sin to humanity. glad to see it made easier to make them

CalliopeSoups:
> microwave code 😶

Microwave code disgusts me but I don't wanna refactor it just for slug... Also because I'm not a great coder. There could be an in-between of making a component to handle unique microwave item interactions like dice too maybe?

JORJ949 (approved):
Extremely jank yet silly

code/modules/cooking/microwave.dm
```diff
@@ -401,15 +417,28 @@ TYPEINFO(/obj/machinery/microwave)
...
 				if(isnull(src))
 					return
 				playsound(src.loc, 'sound/machines/ding.ogg', 50, 1)
 				src.visible_message(SPAN_ALERT("The microwave gets covered in muck!"))
 				src.dirty = TRUE
-				src.icon_state = "mwbloody"
+				if (src.slug)
+					src.icon_state = "mwbloodyS"
+					var/mob/adultSlug = new /mob/living/critter/small_animal/slug
+					if (rand(6))
```
pgmzeta:
```suggestion
					if (prob(6))
```

PR #25121, opened 2025-11-05
Invisible Ore Fix

LorrMaster:
## About the PR
- Fixes ores disappearing if their material changes.

## Why's this needed?
- Invisium is not a material... yet.

## Testing
Screenshot 2025-11-05 052304

### DIFF
--- a/code/obj/item/material.dm
+++ b/code/obj/item/material.dm
@@ -290,9 +290,9 @@
 	update_icon()
 		if(src.icon_stack_value == 1)
 			var/ore_state = pick("ore1", "ore1b")
-			src.icon_state = "[ore_state]_$$[src.material.getName()]"
+			src.icon_state = "[ore_state]_$$[src.default_material]"
 		else
-			src.icon_state = "ore[src.icon_stack_value]_$$[src.material.getName()]"
+			src.icon_state = "ore[src.icon_stack_value]_$$[src.default_material]"
 
 /obj/item/raw_material/molitz_beta
 	name = "molitz crystal"
@@ -310,7 +310,7 @@
 	update_icon()
 		if(src.icon_stack_value == 1)
 			var/ore_state = pick("ore1", "ore1b")
-			src.icon_state = "[ore_state]_$$molitz_b"
+			src.icon_state = "[ore_state]_$$[src.default_material]"
 		else
 			src.icon_state = "ore[src.icon_stack_value]_$$[src.default_material]"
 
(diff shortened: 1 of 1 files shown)


code/obj/item/material.dm
```diff
@@ -401,9 +401,9 @@
 	update_icon()
 		if(src.icon_stack_value == 1)
 			var/ore_state = pick("ore1", "ore1b")
-			src.icon_state = "[ore_state]_$$[src.material.getName()]"
+			src.icon_state = "[ore_state]_$$erebite"
```
pgmzeta: why hardcoded instead of default_material ?

code/obj/item/material.dm
```diff
@@ -310,9 +310,9 @@
 	update_icon()
 		if(src.icon_stack_value == 1)
 			var/ore_state = pick("ore1", "ore1b")
-			src.icon_state = "[ore_state]_$$[src.material.getName()]"
+			src.icon_state = "[ore_state]_$$[src.default_material]"
 		else
-			src.icon_state = "ore[src.icon_stack_value]_$$[src.material.getName()]"
+			src.icon_state = "ore[src.icon_stack_value]_$$[src.default_material]"
```
TobleroneSwordfish, line 315: Why are some of these hardcoded and some `default_material`?

code/obj/item/material.dm
```diff
@@ -310,9 +310,9 @@
 	update_icon()
 		if(src.icon_stack_value == 1)
 			var/ore_state = pick("ore1", "ore1b")
-			src.icon_state = "[ore_state]_$$[src.material.getName()]"
+			src.icon_state = "[ore_state]_$$[src.default_material]"
```
ZeWaka: Any particular reason why default material is used in some places and hardcoded in others?

LorrMaster:
> why hardcoded instead of default_material ?

> Why are some of these hardcoded and some `default_material`?

> Any particular reason why default material is used in some places and hardcoded in others?

Either naming system is viable. `default_material` is used for `/obj/item/raw_material` since that is the parent object. Objects that inherit from that are expected to have their own icon set for their `default_material`. Some ores have minor variations in their icons, so in those cases the material is hardcoded. Partially for clarity towards which icon is being used, but also in case the `default_material` is overwritten but the icon set is not. So you could have another ore that inherits from mauxite ore and it will work with the same icon.

LorrMaster:
I changed my mind, `src.default_material` makes more sense.

pgmzeta (approved):

PR #25124, opened 2025-11-05
Triage tags

Valtsu0:
[feature][medical]
## About the PR 
Adds triage tag box to medical apparel and medborg modules.
Due to prodocs making these kind of obsolete i have tried to make them as convenient to use as possible.

The boxes have infinite amount of tags and can only attach one to each person. If a person already has one it will change the old one. The tags can also be changed or removed with an empty hand. To prevent clutter the tags will disappear upon being removed. 

in addition to regular triage levels there is a unrevivable level.

additionally you can write on the tags
image
[Video demonstration](https://youtu.be/9SOM-G1qn8k)

## Why's this needed? 
I do see these getting some use on rp despite not being very practical.

## Testing 
in addition to the stuff in the video demonstraition, i have also tested that they can be found where they should and that using the box on walls and stuff does nothing.

## Changelog 

```changelog
(u)Valtsu0
(+)Triage tags can now be found in medical apparel vendors and medical borg modules.
```

### DIFF
--- a/code/modules/interface/multiContext/context_actions.dm
+++ b/code/modules/interface/multiContext/context_actions.dm
@@ -2461,7 +2461,6 @@
 /datum/contextAction/triage_tag
 	icon = 'icons/ui/context16x16.dmi'
 	close_clicked = TRUE
-	close_moved = FALSE
 	desc = ""
 	icon_state = "hey"
 	var/level = -1


cropseyy:
I can see the E/U tags being very useful when you have multiple bodies needing cloning.

DisturbHerb:
What if these could glow in the dark like with triage lights used in combat medicine (usually blue is used for expectant/deceased in that case)

Egregorious:
I love these, they have great qol too. 

I think they'd be more legitimately useful if they were dedicated for corpses. Distinguishing between 'unrevivable due to rotting/damage' and 'unrevivable due to lack of a mind/DNR'; or 'potentially borgable'; or even 'already cloned' would be useful for corpse piles.

But the potential to stand at the entrance of medbay and slap 'minor' onto most people who come in instead of just healing them is exceptionally enthralling. 

Not that this really matters but the fact that anyone can elevate themselves to 'immediate' once they have a sticker does hamper their usefulness somewhat.

PR #25136, opened 2025-11-07
Atoms May Offset Their Maptext Managers

Mister-Moriarty:
## About The PR:
Atoms now have the `maptext_manager_x` and `maptext_manager_y` variables, which set the `pixel_x` and `pixel_y` values of maptext managers entering their `vis_contents`.

## Why Is This Needed?
By request of @pgmzeta.

## Testing:
image

Taken before `maptext_offset_x/y` was renamed to `maptext_manager_x/y`.

### DIFF
--- a/code/modules/speech/core.dm
+++ b/code/modules/speech/core.dm
@@ -79,9 +79,9 @@ TYPEINFO(/atom)
 
 	// Maptext Variables:
 	/// The `pixel_x` value that any maptext managers present in this atom's `vis_contents` should use.
-	var/maptext_offset_x = 0
+	var/maptext_manager_x = 0
 	/// The `pixel_y` value that any maptext managers present in this atom's `vis_contents` should use.
-	var/maptext_offset_y = 0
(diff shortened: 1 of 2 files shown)


code/modules/speech/core.dm
```diff
@@ -77,6 +77,12 @@ TYPEINFO(/atom)
 	/// The default speech bubble for sung phrases that were either sung loudly or poorly.
 	var/speech_bubble_icon_sing_bad = "notebad"
 
+	// Maptext Variables:
+	/// The `pixel_x` value that any maptext managers present in this atom's `vis_contents` should use.
+	var/maptext_offset_x = 0
```
TobleroneSwordfish:
```suggestion
	var/chattext_offset_x = 0
```
or similar, otherwise it's way too close to the builtin `maptext_x` and will confuse and confound admins

TobleroneSwordfish (changes requested):

Issue #25126, opened 2025-11-06
Wrestling ring edges block emitter fire.


JORJ949:
### Description

Emitter fire is blocked by wrestling ring edge railings

### Steps to reproduce

1. Play on wrestlemap
2. Setup the singularity
3. The ring edges block emitter fire, loosing the singularity with the default setup.

### Expected Behavior

Default singulo setups not to loose

JORJ949:
@CalliopeSoups I point the accusatory finger at you

flappybatpal:
Fixed in [065181e](https://github.com/goonstation/goonstation/commit/065181e51e648b4e255911bed6e8f2fa9eed0277)

PR #25138, opened 2025-11-08
Fix getting stuck in boxing ropes constantly when corner flipping + projectiles pass through

CalliopeSoups:
## About the PR 

Boxing rope decals, entrances and corners no longer block projectiles, fixing #25126. They also now use the directional blocking object flag instead of doing more dir checks. This also fixes getting caught in boxing ropes, you won't get caught unless you aim into them. 

## Why's this needed? 

Bug bad, using the object flag seems to make sense as well as it's how railings do it.

## Testing 

Tested, singulo beams pass through fine. You no longer get caught in boxing ropes unless you directly aim into them. You do have to re-cross the ropes if you aim into them and you're on the north side, however.

 There is some further testing I want to do before this is mergable and I think some of the cross/uncross code could be compressed, I'll edit this section once done with these.

Edit: Tested on a couple maps, seems to work as intended.
dreamseeker_uoTnCXXnmP

### DIFF
--- a/code/obj/decal.dm
+++ b/code/obj/decal.dm
@@ -202,7 +202,7 @@ proc/make_point(atom/movable/target, pixel_x=0, pixel_y=0, color="#ffffff", time
 	icon_state = "ringrope"
 	plane = PLANE_DEFAULT
 	object_flags = HAS_DIRECTIONAL_BLOCKING
-	layer = OBJ_LAYER
+	layer = MOB_LAYER + 0.1
 	event_handler_flags = USE_FLUID_ENTER
 	pass_unstable = TRUE
 
@@ -213,8 +213,7 @@ proc/make_point(atom/movable/target, pixel_x=0, pixel_y=0, color="#ffffff", time
 			return 1
 		if (mover.throwing)
 			return 1
-		if (src.dir == SOUTH)
-			return 0
+
 		if(get_dir(loc, mover) & dir)
 			return !density
 		else
@@ -270,7 +269,7 @@ proc/make_point(atom/movable/target, pixel_x=0, pixel_y=0, color="#ffffff", time
 			return 1
 		if (mover.throwing)
 			return 1
-		if (src.dir == SOUTHEAST || src.dir == WEST)
+		if (src.dir == NORTHEAST || src.dir == NORTHWEST || src.dir == SOUTHEAST || src.dir == SOUTHWEST)
 			return 0
 		if(get_dir(loc, mover) & dir)
 			return !density


code/obj/railing.dm
```diff
@@ -111,6 +111,9 @@
 			return 1
 		if (O.throwing)
 			return 1
+		if (istype(src, /obj/railing/boxing))
+			if (src.dir == 2) // incase you chairflip yourself next to an entrance rope
```
TobleroneSwordfish:
```suggestion
			if (src.dir == SOUTH) // incase you chairflip yourself next to an entrance rope
```
Use dir defines pls

code/obj/decal.dm
```diff
@@ -201,24 +201,30 @@ proc/make_point(atom/movable/target, pixel_x=0, pixel_y=0, color="#ffffff", time
...
 
 	Cross(atom/movable/mover) // stolen from window.dm
 		if (mover && mover.throwing & THROW_CHAIRFLIP)
 			return 1
-		if (src.dir == SOUTHWEST || src.dir == SOUTHEAST || src.dir == NORTHWEST || src.dir == NORTHEAST || src.dir == SOUTH || src.dir == NORTH)
+		if (!src.density || (mover.flags & TABLEPASS) || istype(mover, /obj/newmeteor) || istype(mover, /obj/linked_laser) )
+			return 1
+		if (mover.throwing)
+			return 1
+		if (src.dir == 2)
```
TobleroneSwordfish: .

code/obj/decal.dm
```diff
@@ -259,17 +266,22 @@ proc/make_point(atom/movable/target, pixel_x=0, pixel_y=0, color="#ffffff", time
 	Cross(atom/movable/mover) // stolen from window.dm
 		if (mover && mover.throwing & THROW_CHAIRFLIP)
 			return 1
-		if (src.dir == SOUTHWEST || src.dir == SOUTHEAST || src.dir == NORTHWEST || src.dir == NORTHEAST || src.dir == SOUTH || src.dir == NORTH)
+		if (!src.density || mover.flags & TABLEPASS || istype(mover, /obj/newmeteor) || istype(mover, /obj/linked_laser) )
+			return 1
+		if (mover.throwing)
+			return 1
+		if (src.dir == 5 || src.dir == 6)
```
TobleroneSwordfish: .

CalliopeSoups:
> This is horrible inheritance abuse but it's not your fault boxing ropes are like that.

I'll likely refactor them entirely someday since I want to add bouncing off of them, I'd rather just get it working now.

PR #25150, opened 2025-11-09
Adjust chemfire atmos interactions to be more reasonable

TobleroneSwordfish:
## About the PR 
Makes adding a hotspot to an airgroup immediately suspend its group processing and makes chemfires heat up air specifically 20x less.
## Why's this needed? 
Looking into blob balance issues from this thread https://forum.ss13.co/showthread.php?tid=24681 I realised that flamethrowers have an issue where 80% of the time they have zero effect on atmos because they're not enough to suspend group processing and then when they *do* have an effect they have such a high volume (3.2x higher than an atmos fire) that they instantly skyrocket the temperature up to ~1200C.

I know this will cause some extra server load so it might be worth testmerging but I do think there being a FIRE in the room is a good reason to want to start actually processing cell-level atmos again. It shouldn't have any effect on atmos fires since if gas in the air is on fire the room is already definitely not group processing.
## Testing 
Firing a single flamethrower shot now brings the temperature up to about 100C, then quickly dissipates back down and *eventually* resumes group processing (the logic for this could still use some work, larger airgroups can take a VERY long time to resume group processing)
image

## Changelog 

```changelog
(u)LeahTheTech
(+)Chemical fires now heat up the air less but much more consistently.
```

### DIFF
--- a/code/modules/atmospherics/FEA_fire.dm
+++ b/code/modules/atmospherics/FEA_fire.dm
@@ -101,6 +101,9 @@
 #endif
 	/// Volume to expose to other atoms. Also used while [/atom/movable/hotspot/var/bypassing] is FALSE to act on a volume of gas on our turf.
 	var/volume = 125
+	/// How much the volume of this hotspot is scaled by when affecting gas temperature
+	/// Importantly this is ONLY gas temperature and not other "expose" effects like objects being caught in the fire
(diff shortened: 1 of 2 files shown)


TobleroneSwordfish:
fuck it we ball

PR #25151, opened 2025-11-09
Adds blank .22 bullets

Driftered:
## About the PR 

- Adds non-damaging .22 bullets to a hacked general manufacturer in the form of a blank magazine. 
- Sounds and fires just like any other .22 round, so it can be loaded into weapons like Faith, but does zero damage (even if the person hit will THINK they took damage due to the hitsplat associated with `DAMAGE_BLUNT`) and doesn't leave .22 shrapnel in the victim.
- Half the material cost of a regular .22 magazine in the manufacturer, to offset the total lack of damage.

## Why's this needed? 
Mostly for laughs, but also opens up roleplay avenues for people wanting to bluff that they've got a lethal weapon without the *consequences* of a lethal weapon (_**or vise-versa**_). 

## Testing 
Crafted the ammo with the hacked manufacturer and ensured that a test NPC took no damage from a full magazine of blank ammo.

Blanks

Blanks Manufacturing

## Changelog 

```changelog
(u)Driftered
(*)Adds craftable .22 blanks to the hacked general manufacturer. Punctuate your arguments with a gunshot today!
```
[A-GAME OBJECTS] [C-SPRITES]

### DIFF
--- a/code/datums/manufacturing.dm
+++ b/code/datums/manufacturing.dm
@@ -407,7 +407,6 @@
 	time = 30 SECONDS
 	category = "Resource"
 
-
 /datum/manufacture/bullet_9mm_frangible
 	name = "9mm Frangible Rounds"
 	item_requirements = list("conductive" = 25,


DisturbHerb:
i think this would benefit from not firing a projectile at all

Driftered:
> i think this would benefit from not firing a projectile at all

I considered not adding a projectile originally for the blanks! My issue came with wanting to make sure it only fit inside weapons using .22 ammo, ~as well as not knowing of any other ammo that doesn't produce a projectile within `code\modules\projectiles\bullet.dm`. If that DOES exist, however, I'd be happy to swap over to just the noise.~ Scratch that, I realized I could make this happen by setting its `max_range` to zero. I'll update the branch!

lyzmat:
> My issue came with wanting to make sure it only fit inside weapons using .22 ammo

this won't be an issue because the blanks are a subclass of the standard .22 bullet, so they inherit the property of being in the "AMMO_PISTOL_22" category

Driftered:
> this won't be an issue because the blanks are a subclass of the standard .22 bullet, so they inherit the property of being in the "AMMO_PISTOL_22" category

Yep, that's why I tossed them there originally. Right after making my reply earlier I realized I could mess with the `max_range`, so everything should be smooth sailing now.

CelineTheYeen:
A neat idea (even if it's not how blanks work), my main problem is the fact that like.. specifically .22 is the single worst caliber for this due to how nonthreatening the guns that use it really are meaning if you start shooting or give the threat a secoff  will run at you and you're getting baton'd if it's real or not.  (not to mention the meta aspect where in this current state literally only like 3 guns can have blanks so any claims made about others are blatantly fake) 

Personally a broader rework of ammo that granted the ability to make any projectile a blank would be my ideal but that is an incredibly large scope rework. A more practical solution would be to introduce other blank types but just available elsewhere, perhaps from traders.

xxfishshadowxx:
> i think this would benefit from not firing a projectile at all

I think the projectile makes it realistic and gives it actual bluff utility, real blanks often launch wax anyways

Glamurio:
I also agree that without the projectile we're back to "people have to pretend you just shot a guy" territory, which plays out vastly different to actually thinking he just shot a guy. So we should definitely have a projectile.

Driftered:
> A neat idea (even if it's not how blanks work), my main problem is the fact that like.. specifically .22 is the single worst caliber for this due to how nonthreatening the guns that use it really are meaning if you start shooting or give the threat a secoff will run at you and you're getting baton'd if it's real or not. (not to mention the meta aspect where in this current state literally only like 3 guns can have blanks so any claims made about others are blatantly fake)
> 
> Personally a broader rework of ammo that granted the ability to make any projectile a blank would be my ideal but that is an incredibly large scope rework. A more practical solution would be to introduce other blank types but just available elsewhere, perhaps from traders.

These are all a lot of good points (and yeah, definitely not how blanks work but that's okay, we can pretend not to know).  Custom ammo would be a really interesting project, but _absolutely_ far larger than I'd like to undertake currently. I'll close this for now and put that on a backburner, but I appreciate everyone's feedback!

PR #25159, opened 2025-11-10
Switched the pickup verb to use tgui_input_list

albassort:
## I compiled it. Kept starting the game by hitting ready then going to server and hitting `start now`. From there I would press P. Eventually, i figured out a way for it to popup.
image
  

## Updated the pickup verb (default hotkey: p) to use TGUI. 

```changelog
(u)Albassort
(*)Updated the pickup verb (default hotkey: p) to use TGUI. Now you can fuzzy search items!
```

### DIFF
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -3368,7 +3368,8 @@ TYPEINFO(/mob)
 			items += I
 	if (items.len)
 		var/atom/A = tgui_input_list(src, "What do you want to pick up?", "", items, start_with_search = TRUE)
-		src.client?.Click(A, get_turf(A))
+		if (A)
+			src.client?.Click(A, get_turf(A))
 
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
I feel like the three people who use this verb will be mad they can't use fast hotkeys to jump to specific items.

albassort:
> I feel like the three people who use this verb will be mad they can't use fast hotkeys to jump to specific items.

They can, they can do this better in fact than they could before. 

> start_with_search = TRUE

Which makes it like the current `gps` menu for example. Its very fast. Currently there is a delay between popup and selection, faster with TGUI. Also, its harder to objects because you can really only go by the first character

Lastly, with this PR, more people will use it.

TobleroneSwordfish:
Mmmm sure why not

PR #25161, opened 2025-11-10
Obtain eyestalks from cooking lesser slugs, removes original method

CalliopeSoups:
## About the PR 

You can cook lesser slugs in the oven to get two eyestalks.

## Why's this needed? 

The method of obtaining eyestalks seemed to be too far for a handful of people. This way of doing it should be somewhat less gross.

## Testing 

Tested at target time/temp, at different temps and different meat recipes. Nothing broke.

## Changelog 

```changelog
(u)CalliopeSoups
(+)Slug eyestalks are obtained from cooking a slug, instead of the horrors.
```

### DIFF
--- a/code/modules/cooking/food_and_drink/alcohol.dm
+++ b/code/modules/cooking/food_and_drink/alcohol.dm
@@ -395,7 +395,7 @@
 		icon_state = "eyestalk1"
 		edible = 1
 		appearance_flags = KEEP_TOGETHER
-		var/second_stalk = FALSE
+		var/second_stalk = FALSE // you get two! :D
 
 		New()
 			..()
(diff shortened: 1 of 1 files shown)


444Portal:
im admittedly still not fond of slug food as a whole but this is infinitely better and allows me to at least stomach thinking about it. previously i had to time myself out from the discord because i felt like i was going to throw up x.x

much appreciated change, really hope it gets merged...

Glamurio:
We have dismemberment and corpse rotting but I guess slugs are where we draw the line

444Portal:
> We have dismemberment and corpse rotting but I guess slugs are where we draw the line

Unfortunately it doesn't really work under that logic and some things are much easier to abstract away than others. 

Dismemberment and corpse rotting has reasons to exist and there's differences between being purposefully gross to just be gross and gore existing because we have systems for health, medicine and cloning. Shrug!

emmenthal:
This also should fix #25163

code/modules/cooking/food_and_drink/alcohol.dm
```diff
@@ -395,12 +395,18 @@
 		icon_state = "eyestalk1"
 		edible = 1
 		appearance_flags = KEEP_TOGETHER
+		var/second = FALSE
```
ZeWaka: i don't get this variable name

CalliopeSoups: Changed the name, though I can just remove the code that gives you a second eyestalk if that's neater.

PR #25179, opened 2025-11-14
Fly Object Overhead buildmode tool

CalliopeSoups:
## About the PR 

Fly Object Overhead is a buildmode tool that lets an admin upload an image and/or sound file, which will spawn at the edge of the zlevel then travel towards the target turf you clicked on. Once at its destination it will play the chosen ending effect, including continuing until off zlevel, exploding, fading away or running away. The sound will play either globally on loop until the pilot mob is gone or once locally once the pilot reaches its destination, given the admin's choice.

You can also choose a mob or object of any amount to spawn at the destination turf once reached.

- [x] Stress test
- [x] Spawn nearby to target turf option
- [x] Reference a mob/object for its sprite 
- [x] Rotate sprite with pilot dir, possibly can't do for uploaded images. Edit: possible to turn them about but can't figure out how to mirror them.
- [x] Use object spawn buildmode for spawn system, if possible. Edit: Couldn't figure out a way to do this. Let me know if you do.
- [x] Add adminlogs

## Why's this needed? 

Admin goof tools are fun, plus this is very versatile so it can be used for a lot of different things. 

## Testing 

I've tested this in most ways I can think of but I'm yet to do some spam stress testing. The functionality all seems to work well, though before merging I want to add an option for the uploaded image to rotate with the pilot mob's direction. I tried this already but couldn't figure out how to mirror the image instead of just rotating, cause otherwise the image will look upside-down when flipped.

The audio global loop would have been local but it is surprisingly hard to have that work with an uploaded sound file.

Some of the code here might be stinky too, let me know if there's things to be changed but as long as it works and doesn't crash when used responsibly I'm happy, with it being an admin tool.

https://github.com/user-attachments/assets/6383eee8-ff18-45a3-bc9b-7d6f6acb4bbf


CalliopeSoups:
I did a bit of spamming and changing controls about to test this thing out. It seems good, the only way you could lag the server out is if like a bunch of admins spammed it at once or set a really high spawn mob/obj amount and spammed that. I feel like that should be obvious hopefully.

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+		var/old_loc = pilot.loc
+		pilot.set_loc(get_step(pilot, direction))
+		pilot.dir = direction
+		SEND_SIGNAL(src, COMSIG_MOVABLE_MOVED, old_loc, direction)
+		if (!pilot.loc)
+			qdel(pilot)
+		pilot.glide_size = glide
+		pilot.animate_movement = SLIDE_STEPS
+
+/mob/image_pilot
```
pgmzeta: why mob ?

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+		if (shift)
+			src.target_loc = get_turf(object)
+			logTheThing(LOG_ADMIN, usr, "uploaded ")
+			aim_pilot()
+
+	proc/aim_pilot() // Spawn pilot at edge of zlevel then redirect to target
+		var/turf/start
+		var/random_dir = pick(NORTH, SOUTH, EAST, WEST)
+		var/new_dir
+		if (!src.target_loc || !src.dir_input)
```
pgmzeta, line 93: add check to make sure either sound or image exists, if both don't then give boutput feedback and don't send a nothing out

pgmzeta (changes requested):
admincrimes:)

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: I think that's leftover from a previous iteration where I used mob's life process. It was bad!

CalliopeSoups: Actually seems like it was for a reason, it's so I can use sprint_particle() on the "run away" option for the end effects. I can change the pilot to an obj and add a sprint particle manually if wanted.

pgmzeta: It feels like it's an effect so it shouldn't be buryable/qdeled in magnet/chem reactable/crushable/etc

though those are all kind of funny, i wonder if they'd break this

ZeWaka: hmm yeah i think it'd be best to not be a mob for above, see also trench holes and such

CalliopeSoups: Yeah I'll make it an obj soon then, adding foot scuffs fx manually should be easy anyway.

pgmzeta (changes requested):
needs that mob->obj/effect (or similar effect type) and testing before good to go

pgmzeta:
please switch from draft to open when this is ready for review.

CalliopeSoups:
I've tested all the options, they all seem fine to me. Had a problem with the global loop audio option replaying each time a pilot is disposed, fixed that and tested it. I believe it's all in order now hopefully.

Edit: Tested some more cause it's fun, realized it'd be useful to choose if you want the image to layer over line of sight darkness or not so I added an option for that. Tested multiple times, seems to work well.

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+			qdel(pilot)
+		pilot.glide_size = glide
+		pilot.animate_movement = SLIDE_STEPS
+
+/obj/image_pilot
+	name = ""
+	desc = ""
+	anchored = ANCHORED
+	density = 0
+	layer = 104 // AAAAAAAAA
```
ZeWaka: why a weird magic layer, why not one based on the existing layers

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+		..()
+
+	proc/play()
+		if (src.loc && !QDELETED(src))
+			src.attached_sound = sound(src.attached_sound, TRUE, TRUE, 1020, 10)
+			world << src.attached_sound
+
+	proc/check_for_pilots()
+		var/list/pilotlist = list()
+		for (var/obj/image_pilot/spawned_pilot in world)
```
ZeWaka: absolutely not, use the tracking macros

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+	layer = 104 // AAAAAAAAA
+	flags = KEEP_TOGETHER
+	event_handler_flags = IMMUNE_OCEAN_PUSH | IMMUNE_SINGULARITY | IMMUNE_TRENCH_WARP
+	var/icon/image_overlay
+	var/sound/attached_sound
+	var/loopsound = FALSE
+
+	New()
+		..()
+		SPAWN(0)
```
ZeWaka, line 222: why spawn 0

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+	proc/check_for_pilots()
+		var/list/pilotlist = list()
+		for (var/obj/image_pilot/spawned_pilot in world)
+			pilotlist += spawned_pilot
+
+		if (pilotlist.len == 1)
+			if (pilotlist.Find(src))
+				return TRUE
+		else
+			return FALSE
```
ZeWaka: why only 1, this makes no sense

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+				for (var/i=0,i<=3,i++)
+					move_forward(pilot, direction, 3)
+					sleep(2)
+				SPAWN(0)
+					pilot.ClearAllOverlays()
+					qdel(pilot)
+
+	proc/move_forward(var/obj/image_pilot/pilot, var/direction, var/speed=1)
+		var/glide = 0 // this system seems to desync sometimes, not a huge issue it seems to add a bit of variety to the way they move
+		glide = (32 / speed) * world.tick_lag
```
ZeWaka: why set it to 0 and then immediately to another value

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+			while (pilot.y != target_locinput.y)
+				if(QDELETED(pilot))
+					break
+				move_forward(pilot, direction, speed=speedinput)
+				sleep(speedinput)
+
+		// everything below is after the pilot reaches its destination
+
+		if (!pilot.loopsound && pilot.attached_sound)
+			for (var/mob/player in view(25, pilot))
```
ZeWaka: 25 is an absolutely gigantic range and definitely out of view distance

what are you actually trying to do here

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+				player << sound(pilot.attached_sound, volume=10)
+
+		if (pathinput)
+			if(ispath(pathinput, /atom/movable))
+				var/counter
+				playsound(pilot.loc, 'sound/effects/poff.ogg', 30, TRUE, pitch = 1)
+				for (counter=0, counter<pathamountinput, counter++)
```
ZeWaka: why is this not just a normal var in the loop

ZeWaka: for var/i in 1 to pathamountinput

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+				src.spawnpath = get_one_match(input("Type path", "Type path", "[src.spawnpath]"), /atom)
+				src.spawnamount = tgui_input_number(usr, "Amount to spawn", "Amount - Be responsible!!", 1)
+			else
+				src.spawnpath = null
+				src.spawnamount = 1
+
+	click_left(atom/object, var/ctrl, var/alt, var/shift)
+		if (shift)
+			src.target_loc = get_turf(object)
+			logTheThing(LOG_ADMIN, usr, "uploaded ")
```
ZeWaka: ?

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+
+		if (ctrl)
+			src.audio_choice = tgui_input_list(usr, "Loop sound globally or play once on arrival?", "Choose", list("Global loop", "Once", "Clear"))
+			if (src.audio_choice == "Clear")
+				src.audio = null
+			else
+				src.audio = input(usr, "Upload a file:", "File Uploader - Long files WILL lag people out, sounds will loop.", null) as null|sound
+			if (src.audio)
+				logTheThing(LOG_ADMIN, usr, "uploaded a sound [src.audio] to use with Fly Object Overhead buildmode")
+	click_right(atom/object, var/ctrl, var/alt, var/shift)
```
ZeWaka, line 62:
```suggestion
				logTheThing(LOG_ADMIN, usr, "uploaded a sound [src.audio] to use with Fly Object Overhead buildmode")

	click_right(atom/object, var/ctrl, var/alt, var/shift)
```

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+			var/choice = tgui_input_list(usr, "Choose a set speed or random values", "Choose", list("Set", "Clear"))
+			var/choice2 = tgui_input_list(usr, "Start from edge of zlevel or nearby? (About 2 screens away)", "Choose", list("Edge", "Nearby"))
+			if (choice2 == "Nearby")
+				src.startnearby = TRUE
+			else
+				src.startnearby = FALSE
+			if (choice == "Set")
+				src.move_delay = tgui_input_number(usr, "Enter speed value of image", "Higher is slower, gets very slow by 5", 1)
+			else
+				src.move_delay = 1
```
ZeWaka: please actually process these in order

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+				if ("Clear")
+					src.image = null
+					usr.visible_message("Image cleared.")
+					return
+				if ("Icon Ref")
+					src.image = null
+					src.image = get_one_match(input("Type path", "Type path", "[src.spawnpath]"), /atom)
+			src.alphainput = tgui_input_number(usr, "Enter an alpha level.", "Alpha", 255, 255, 0)
+			if (src.image)
+				logTheThing(LOG_ADMIN, usr, "uploaded an image [src.image] to use with Fly Object Overhead buildmode")
```
ZeWaka, line 53: how does this actually log? if it's just a ref it wont be terribly useful

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+			var/image/ship
+			ship = image(icon=src.image_overlay, loc=src, layer = EFFECTS_LAYER_4)
+			AddOverlays(ship, "ship")
+			if (src.check_for_pilots())
+				play()
+
+	disposing()
+		if (src.check_for_pilots())
+			var/sound/stopsound = sound(null, wait = TRUE, channel=1020)
+			world << stopsound
```
ZeWaka: do we not have some handler for this

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+		..()
+
+	proc/play()
+		if (src.loc && !QDELETED(src))
+			src.attached_sound = sound(src.attached_sound, TRUE, TRUE, 1020, 10)
```
ZeWaka: why is this just a random 1020 channel

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+				SPAWN(3 SECONDS)
+					direction = turn(direction, 180)
+					new/obj/particle/attack/sprint(pilot.loc)
+					playsound(pilot.loc, 'sound/effects/sprint_puff.ogg', 30, 1)
+					while (pilot.loc)
+						move_forward(pilot, direction)
+						sleep(1)
+			if ("Explode")
+				var/turf/T = get_turf(pilot)
+				playsound(T, "sound/effects/Explosion[pick(1, 2)].ogg", 15, 1)
```
ZeWaka: this should be from `big_explosions`

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+			if ("Leave zlevel")
+				while (pilot.loc)
+					move_forward(pilot, direction, speed=speedinput)
+					sleep(speedinput)
+			if ("Run away")
+				SPAWN(3 SECONDS)
+					direction = turn(direction, 180)
+					new/obj/particle/attack/sprint(pilot.loc)
+					playsound(pilot.loc, 'sound/effects/sprint_puff.ogg', 30, 1)
+					while (pilot.loc)
```
ZeWaka, line 173: isnt this a forever loop

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+				player << sound(pilot.attached_sound, volume=10)
+
+		if (pathinput)
+			if(ispath(pathinput, /atom/movable))
+				var/counter
+				playsound(pilot.loc, 'sound/effects/poff.ogg', 30, TRUE, pitch = 1)
+				for (counter=0, counter<pathamountinput, counter++)
+					var/turf/T = GetRandomPerimeterTurf(get_turf(pilot), 1)
+					new pathinput(T)
+					var/obj/itemspecialeffect/poof/P = new /obj/itemspecialeffect/poof
```
ZeWaka:
```suggestion
					var/obj/itemspecialeffect/poof/P = new
```

code/modules/admin/buildmodes/fly_image_overhead.dm
ZeWaka: also not every mob is a player...

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+				if(QDELETED(pilot))
+					break
+				move_forward(pilot, direction, speed=speedinput)
+				sleep(speedinput)
+
+		// everything below is after the pilot reaches its destination
+
+		if (!pilot.loopsound && pilot.attached_sound)
+			for (var/mob/player in view(25, pilot))
+				player << sound(pilot.attached_sound, volume=10)
```
ZeWaka: not how we play sounds generally

ZeWaka: also why a fixed volume of 10

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,240 @@
...
+			new_dir = random_dir
+		else
+			new_dir = src.dir_input
+
+		if (src.startnearby)
+			start = get_ranged_target_turf(src.target_loc, new_dir, 35)
+		else
+			start = get_edge_target_turf(src.target_loc, new_dir)
+
+		new_dir = turn(new_dir, 180)
```
ZeWaka: why are we turning from a user-provided dir

ZeWaka (changes requested):

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: The image won't be appended to the pilot otherwise, not sure why.

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: Added a comment explaining it, I can figure out another way to do it if still wanted.

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: Shouldn't be unless you're seeing something I'm not which is likely. `move_forward()` deletes the pilot once it doesn't have a loc (after leaving off the edge of the map).

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: Changed to playsound, still needs a fixed volume level or the sound doesn't play so I slapped it at 30 after some testing.

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: Struggling to read my own code here honestly but I believe new_dir is used to set where the pilot will be spawned, so it needs to flip around to face the direction of the target location for send_pilot.

CalliopeSoups:
Did more stress testing after the feedback changes, things seem to be in order.

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,236 @@
...
+					while (pilot.loc)
+						move_forward(pilot, direction)
+						sleep(1)
+			if ("Explode")
+				var/turf/T = get_turf(pilot)
+				playsound(T, pick(big_explosions), 80, 1)
+				new /obj/effects/explosion(T)
+				qdel(pilot)
+				robogibs(T)
+			if ("Fade away")
```
JORJ949: These strings should be defined and then undefined at the end of the file.

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,236 @@
...
+			world << stopsound
+		STOP_TRACKING
+		..()
+
+	proc/play()
+		if (src.loc && !QDELETED(src))
+			src.attached_sound = sound(src.attached_sound, TRUE, TRUE, 1020, 10)
+			world << src.attached_sound
+
+	proc/check_for_pilots() // check if there's only one pilot, used for starting and stopping looping audio
```
JORJ949: What if multiple objects are set to fly overhead at once? Won't they interfere with each other?

JORJ949 (changes requested):

code/modules/admin/buildmodes/fly_image_overhead.dm
ZeWaka: yes

CalliopeSoups:
Didn't notice I had more reviews to do when I requested re-review, my bad. Github collapsed em so I looked them over, will get done soon!

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: It's just the file name unfortunately, tested the logging on "Show Image to All" and it's the same there. Not sure if adding a way for images to show through logs is doable by me at least in this PR.

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: Switched to SOUNDCHANNEL_ADMIN_HIGH but I don't know if I need to add a +1 or something to it to make sure it doesn't conflict with admin played radio stuff?

ZeWaka: Yeah that'll 100% conflict if admins are using the DJ panel

ZeWaka: Why are you even using a specific channel anyways

CalliopeSoups: Gotta be honest, I thought you had to, I assume it just picks a free channel if you leave that space blank then.

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: I tried sending a couple pilots at once manually and it's working, but I'm not sure how it would work with two separate admins calling it at the same time. I'm not sure if one global audio would override the other, play when the first one sent is over or repeatedly switch between the two.

I'm not sure how to test that without a testmerge, unless anyone knows how I'd get around it per-emptively.

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,250 @@
...
+		if (src.check_for_pilots())
+			var/sound/stopsound = sound(null, wait = TRUE)
+			world << stopsound
+		STOP_TRACKING
+		..()
+
+	proc/play()
+		if (src.loc && !QDELETED(src))
+			src.attached_sound = sound(src.attached_sound, TRUE, TRUE, volume=10)
+			world << src.attached_sound
```
ZeWaka: the main problem here is this just ignores sound volume preferences

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,250 @@
...
+
+	proc/check_for_pilots() // check if there's only one pilot, used for starting and stopping looping audio
+		if (length(by_type[/obj/image_pilot]) == 1)
```
ZeWaka: i think someone already commented but this will be a problem if people do this multiple times at once

CalliopeSoups:
Had trouble coming up with a good solution for the audio looping option, so I gutted that feature for now. If this goes through I'll make a part 2 PR eventually that adds the looping audio back as well as the option to play it globally or locally.

Single sounds played on destination stay since those are simple.

CalliopeSoups:
!merge_upstream

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,220 @@
...
+	// settings. behold my vars
+	icon_state = "flyoverhead"
+	var/move_delay = 1
+	var/icon/image
+	var/atom/icon_from_thing
+	var/turf/target_loc
+	var/audio
+	var/plane_choice = FALSE
+	var/dir_input = "Random"
+	var/end_effect = "Leave zlevel"
```
JORJ949:
```suggestion
	var/end_effect = LEAVE
```
Use your defines

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,220 @@
...
+/datum/buildmode/fly_image_overhead
+	name = "Fly Image Overhead"
+	desc = {"***********************************************************<br>
+Upload an image and/or audio and have it fly to target turf then play an ending effect.
+Protip: tinker with this on a local first so you know what you're doing. Images are best around 100-200 pixels wide and long.<br>
+<br>
+RMB on buildmode button                = Set image and ending effect<br>
+Ctrl + RMB on buildmode button         = Set audio<br>
+Alt + Right mouse button               = Set optional obj/mob spawns<br>
+Shift + Right Mouse Button             = Set direction and speed<br>
```
JORJ949: Alt+Shift options should specify its RMB on buildmode button, as this reads its rmb anywhere

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,220 @@
...
+					src.image = null
+					usr.visible_message("Image cleared.")
+					return
+				if ("Icon Ref")
+					src.image = null
+					src.image = get_one_match(input("Type path", "Type path", "[src.spawnpath]"), /atom)
+			src.alphainput = tgui_input_number(usr, "Enter an alpha level.", "Alpha", 255, 255, 0)
+			if (src.image)
+				logTheThing(LOG_ADMIN, usr, "uploaded an image [src.image] to use with Fly Object Overhead buildmode")
+			src.end_effect = tgui_input_list(usr, "Pick ending effect", "End Effect", list("Leave zlevel", "Explode", "Fade away", "Run away"))
```
JORJ949:
```suggestion
			src.end_effect = tgui_input_list(usr, "Pick ending effect", "End Effect", list(LEAVE, EXPLODE, FADE, RUN))
```
Defines

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,220 @@
...
+			src.plane_choice = tgui_input_list(usr, "Layer above blackness?", "Blackness layer", list(TRUE, FALSE))
```
JORJ949:
```suggestion
			src.plane_choice = (tgui_alert(usr, "Layer above blackness?", "Blackness layer", list("Yes", "No")) == "Yes")
```
Should be one with two buttons rather than a list of 2 options

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,220 @@
...
+
+		if (ctrl)
+			src.audio = input(usr, "Upload a file:", "Uploader - Long files WILL lag people out, sound will play once at destination", null) as null|sound
+			if (src.audio)
+				logTheThing(LOG_ADMIN, usr, "uploaded a sound [src.audio] to use with Fly Object Overhead buildmode")
+
+	click_right(atom/object, var/ctrl, var/alt, var/shift)
+		if (shift)
+			src.dir_input = tgui_input_list(usr, "Pick starting direction", "Direction", list(NORTH, SOUTH, EAST, WEST, "Random"))
```
JORJ949: I believe this will output the numbers into a list, but also should probably just use the build mode holder dir instead, i think the object spawn "missile" animation does this.

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,220 @@
...
+			var/choice = tgui_input_list(usr, "Choose a set speed or random values", "Choose", list("Set", "Clear"))
+			var/choice2 = tgui_input_list(usr, "Start from edge of zlevel or nearby? (About 2 screens away)", "Choose", list("Edge", "Nearby"))
```
JORJ949: Like the above comment on list inputs, these should be a tgui_alert.

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,220 @@
...
+		var/new_dir
+		if (!src.target_loc || !src.dir_input)
+			boutput(usr, "No target location and/or direction input found.")
+			return
+		if (!src.audio && !src.image)
+			boutput(usr, "No audio or image file found.")
+			return
+
+		if (src.dir_input == "Random")
+			new_dir = random_dir
```
JORJ949: You should move the pick() here instead, theres no need to remember the random dir when you dont use it anywhere else.

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,220 @@
...
+Shift + Left Mouse Button              = Spawn flying object<br>
+***********************************************************"}
+	// settings. behold my vars
+	icon_state = "flyoverhead"
+	var/move_delay = 1
+	var/icon/image
+	var/atom/icon_from_thing
+	var/turf/target_loc
+	var/audio
+	var/plane_choice = FALSE
```
JORJ949:
```suggestion
	var/image_layers_over_blackness = FALSE
```
More accurate to what the variable does

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,220 @@
...
+		SEND_SIGNAL(src, COMSIG_MOVABLE_MOVED, old_loc, direction)
+		if (!pilot.loc)
+			qdel(pilot)
+		pilot.glide_size = glide
+		pilot.animate_movement = SLIDE_STEPS
+
+/obj/image_pilot
+	name = ""
+	desc = ""
+	anchored = ANCHORED
```
JORJ949:
```suggestion
	anchored = ANCHORED_ALWAYS
```

JORJ949 (changes requested):

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: It is rmb anywhere for those options, changed to further show that

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+
+	click_right(atom/object, var/ctrl, var/alt, var/shift)
+		if (shift)
+			src.dir_input = tgui_alert(usr, "Unlocked dir checks YOUR dir whenever you send the pilot, locked dir stays as the dir you have right now", "Direction Choice", list("Dir (LOCKED)", "Dir (UNLOCKED)", "Random Direction"))
+			var/choice = tgui_alert(usr, "Choose a set speed or random values", "Choose", list("Set", "Clear"))
+			var/choice2 = tgui_alert(usr, "Start from edge of zlevel or nearby? (About 2 screens away)", "Choose", list("Edge", "Nearby"))
+			if (src.dir_input == "Dir (LOCKED)")
+				src.dir_input = usr.dir
+			if (choice == "Set")
+				src.move_delay = tgui_input_number(usr, "Enter speed value of image", "Higher is slower, gets very slow by 5", 1)
```
pgmzeta: can be set to zero, causing division by zero runtime that repeats infinitely

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+			var/choice = tgui_input_list(usr, "Spawn mobs/objects or clear?", "Choose", list("Spawn", "Clear"))
+			if (choice == "Spawn")
+				src.spawnpath = get_one_match(input("Type path", "Type path", "[src.spawnpath]"), /atom)
+				src.spawnamount = tgui_input_number(usr, "Amount to spawn", "Amount - Be responsible!!", 1)
+			else
+				src.spawnpath = null
+				src.spawnamount = 1
+
+	click_left(atom/object, var/ctrl, var/alt, var/shift)
+		if (shift)
```
pgmzeta: why is this requiring shift? very different behavior to other buildmodes

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+		if (!src.target_loc || !src.dir_input)
+			boutput(usr, "No target location and/or direction input found.")
+			return
+		if (!src.audio && !src.image)
+			boutput(usr, "No audio or image file found.")
+			return
+
+		if (src.dir_input == "Dir (UNLOCKED)")
+			new_dir = usr.dir
+		else if (src.dir_input == "Random Direction")
```
pgmzeta: use a define and set the default `dir_input` and this to the same exact string so the direction is set correctly

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+	name = "Fly Image Overhead"
+	desc = {"***********************************************************<br>
+Upload an image and/or audio and have it fly to target turf then play an ending effect.
+Protip: tinker with this on a local first so you know what you're doing. Images are best around 100-200 pixels wide and long.<br>
+<br>
+RMB on buildmode button                = Set image and ending effect<br>
+Ctrl  + RMB on buildmode button        = Set audio<br>
+Alt   + RMB off of buildmode button    = Set optional obj/mob spawns<br>
+Shift + RMB off of buildmode button    = Set direction and speed<br>
+Shift + LMB off of buildmode button    = Spawn flying object<br>
```
pgmzeta: "off of buildmode button" -> "on turf"

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+
+		if (ctrl)
+			src.audio = input(usr, "Upload a file:", "Uploader - Long files WILL lag people out, sound will play once at destination", null) as null|sound
+			if (src.audio)
+				logTheThing(LOG_ADMIN, usr, "uploaded a sound [src.audio] to use with Fly Object Overhead buildmode")
+
+	click_right(atom/object, var/ctrl, var/alt, var/shift)
+		if (shift)
+			src.dir_input = tgui_alert(usr, "Unlocked dir checks YOUR dir whenever you send the pilot, locked dir stays as the dir you have right now", "Direction Choice", list("Dir (LOCKED)", "Dir (UNLOCKED)", "Random Direction"))
+			var/choice = tgui_alert(usr, "Choose a set speed or random values", "Choose", list("Set", "Clear"))
```
pgmzeta: Better variable names please `choice` -> `speed_choice`. Should the second option here be "Random" instead of "Clear"?

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+			var/choice2 = tgui_alert(usr, "Start from edge of zlevel or nearby? (About 2 screens away)", "Choose", list("Edge", "Nearby"))
+			if (src.dir_input == "Dir (LOCKED)")
+				src.dir_input = usr.dir
+			if (choice == "Set")
```
pgmzeta: this should prompt right after the set speed dialog choice

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+				src.spawnpath = get_one_match(input("Type path", "Type path", "[src.spawnpath]"), /atom)
+				src.spawnamount = tgui_input_number(usr, "Amount to spawn", "Amount - Be responsible!!", 1)
+			else
+				src.spawnpath = null
+				src.spawnamount = 1
+
+	click_left(atom/object, var/ctrl, var/alt, var/shift)
+		if (shift)
+			src.target_loc = get_turf(object)
+			aim_pilot()
```
pgmzeta: this should output text to the user saying it's been sent, since it's not going to be immediately obvious it worked

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,223 @@
...
+***********************************************************"}
+	// settings. behold my vars
+	icon_state = "flyoverhead"
+	var/move_delay = 1
+	var/icon/image
+	var/atom/icon_from_thing
+	var/turf/target_loc
+	var/audio
+	var/image_layers_over_blackness = FALSE
+	var/dir_input = "Random"
```
pgmzeta: doesn't match the check for random below an causes a runtime; see comment on using defines

pgmzeta (changes requested):
Very fun tool, ran into some problems while playing around with it locally as noted in comments. Excited to see this get in once it's fully baked!

code/modules/admin/buildmodes/fly_image_overhead.dm
CalliopeSoups: I figured it would prevent misclicks spoiling a joke before you're done setting the thing up. I can make it just left click if that's not a concern.

CalliopeSoups: Think I prefer it without actually, agreed.

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,230 @@
+#define LEAVE "Leave zlevel"
+#define RUN "Run away"
+#define EXPLODE "Explode"
+#define FADE "Fade away"
+#define DIR_LOCKED "Locked Dir"
+#define DIR_UNLOCKED "Unlocked Dir"
+#define DIR_RANDOM "Random Dir"
+
```
pgmzeta, line 8: need UNDEFs at end of file

code/modules/admin/buildmodes/fly_image_overhead.dm
```diff
@@ -0,0 +1,230 @@
...
+				logTheThing(LOG_ADMIN, usr, "uploaded a sound [src.audio] to use with Fly Object Overhead buildmode")
+
+	click_right(atom/object, var/ctrl, var/alt, var/shift)
+		if (shift)
+			src.dir_input = tgui_alert(usr, "Unlocked dir checks YOUR dir whenever you send the pilot, locked dir stays as the dir you have right now", "Direction Choice", list(DIR_LOCKED, DIR_UNLOCKED, DIR_RANDOM))
+			if (src.dir_input == DIR_LOCKED)
+				src.dir_input = usr.dir
+			var/speed_choice = tgui_alert(usr, "Choose a set speed or random values", "Choose", list("Set", "Random"))
+			if (speed_choice == "Set")
+				src.move_delay = tgui_input_number(usr, "Enter speed value of image", "Higher is slower, gets very slow by 5", 1, min_value = 1)
```
pgmzeta: Image
Swap the title and text string order here, title could just be "Image Speed Value"

pgmzeta (changes requested):
last couple of things from testing

PR #25180, opened 2025-11-14
Refactor cooking recipes

Egregorious:
## About the PR 

Refactors `cookingrecipe` into a more generic 'recipe' template, potentially usable for other crafting systems. 

- More logic has been moved internally, recipes have been given more responsibility to instantiate their own output.
- Recipes can now produce varying item types and quantities as outputs, with the oven and mixer adjusted to handle this.
- Recipes are machine-agnostic, and are interacted with through lists of ingredients; machine-specific interactions are abstracted into `recipe_instruction` components.
- Wildcards are explicitly handled, with configurable quantity.
- Reworked how recipes share ingredient data for icon displays, so it should be easier to uniquely obfuscate parts of a recipe if desired, or provide meaningful info for recipes with random outputs.
- Support for ingredients that do not get consumed by crafting.

## Why's this needed? 

This should make recipes more flexible, reusable and maintainable, and hopefully makes it easier to create more complex patterns. 

## Testing 

Tested all the recipes with special outputs; a random assortment of regular recipes; oven success/fail/emag behaviours; mixer success/fail behaviours; human meat naming; list/associated list recipe outputs. Have noticed no changes in behaviour from live while testing. 

## Changelog 

```changelog
(u)Egregorious
(*)Recipes used by the oven and mixer have been internally reworked. Please report any changes in behaviour, nothing should be different. 
```

### DIFF
--- a/code/modules/cooking/kitchen_submachines.dm
+++ b/code/modules/cooking/kitchen_submachines.dm
@@ -857,8 +857,8 @@ TYPEINFO(/obj/submachine/chef_oven)
 			output += new /obj/item/reagent_containers/food/snacks/yuck()
 
 		for(var/item in output)
-			var/obj/item/reagent_containers/food/snacks/yuck = item
-			if (istype(yuck) && yuck.quality < 1)
+			var/obj/item/reagent_containers/food/snacks/yuck/yuckitem = item
+			if (istype(yuckitem) && yuckitem.quality < 1)
 				src.yuck_cooking(item, bonus, cook_amt)
 			else
 				src.normal_cooking(item, bonus, recipebonus, cook_amt)


code/modules/cooking/microwave_recipes.dm
```diff
@@ -1,7 +1,7 @@
 // god damn it
 // TODO: make microwaves use /datum/cookingRecipe, somehow
 // possibly reclaim /datum/recipe
-/datum/recipe
+/datum/recipe_m
```
pgmzeta, line 4: is this not the time to make this change?

Egregorious: Perhaps I'm misunderstanding what you mean but this PR does reclaim the datum/recipe namespace from the microwave. 

Microwaves shouldn't use a unique recipe type at all. I do have a followup for changing microwaves to use normal recipes and get rid of these entirely, but it's out of scope for this since it's not trivial.

pgmzeta:
!merge_upstream

Egregorious:
Cake recipe should be fixed, it works during testing. 

The batter recipe was accidentally qdeling the item used for the filling.

pgmzeta:
cream pie with item recipe is outputting cream pie with item and smouldering mess if cooked wrong

Egregorious:
That should be the failure and successes being output together fixed, tested a few of the custom recipes but the problem was the oven itself so it affected every recipe that was burned.

PR #25186, opened 2025-11-15
Material Texture Fixes & Code cleanup

LorrMaster:
## About the PR
- Fixes material textures breaking non-standard sized icons
- Icons larger than the available 64x64 textures no longer have a material texture applied
- Material textures now applied to lizard tail clothing and handcuffs
- Cleans up related procs to make more sense

## Why's this needed?
- The material visual changes made textured materials more common, making issues with them more prevalent

## Testing
Screenshot 2025-11-15 073049

### DIFF
--- a/code/WorkInProgress/IconProcs.dm
+++ b/code/WorkInProgress/IconProcs.dm
@@ -102,6 +102,33 @@ icon
 		Higher value means brighter color
  */
 
+/// Use another icon as a cut-out for the original icon
+proc/GetMaskedIcon(var/icon/original, var/icon/mask_base)
+	RETURN_TYPE(/icon)
+	var/icon/mask = icon(mask_base)
+	mask.MapColors(1,1,1, 1,1,1, 1,1,1, 1,1,1)
+	mask.Blend(original, ICON_MULTIPLY)
+	return mask
+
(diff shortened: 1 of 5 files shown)


itchybones:
Ooh nice, while you're at it, I recently saw that bespoke space suits (and thus, presumably the other nanofab stuffs), don't recieve the material colour, only the texture (example I saw was carbon nanofibre). Might be because the default sprites are white? Could probably fix that by giving them some colour.

lyzmat:
> Ooh nice, while you're at it, I recently saw that bespoke space suits (and thus, presumably the other nanofab stuffs), don't recieve the material colour, only the texture (example I saw was carbon nanofibre). Might be because the default sprites are white? Could probably fix that by giving them some colour.

i believe they've actually already done some work on this

LorrMaster:
> Ooh nice, while you're at it, I recently saw that bespoke space suits (and thus, presumably the other nanofab stuffs), don't recieve the material colour, only the texture (example I saw was carbon nanofibre). Might be because the default sprites are white? Could probably fix that by giving them some colour.

> i believe they've actually already done some work on this

I only posted some color adjustments on discord, but I am actually currently working on the spacesuit sprite. So I should PR a fix for that soon.

PR #25189, opened 2025-11-15
Adds the cosmetic Muscly gene

Driftered:
## About the PR 
This PR adds a cosmetic gene that makes the owner _**SWOLE.**_ In actuality, it copies the effect applied by the Muscly Belt (`/obj/item/storage/belt/muscly`) without the character needing to wear the belt or have filters/procs applied. Won't appear in genepools and cannot be researched, since it's meant entirely as a joke and warps the character's appearance.

## Why's this needed? 
Entirely for the enjoyment of admins wanting to goof around. A few admins expressed a desire for this one, so here I am to make nightmares come true.

## Testing 
Gave myself the gene through the bioeffect manager and removed it in order to make sure the effect applied and unapplied correctly.

Muscly Man

[A-ADMIN] [C-QoL]

### DIFF
--- a/code/modules/medical/genetics/bioEffects/useless.dm
+++ b/code/modules/medical/genetics/bioEffects/useless.dm
@@ -311,6 +311,15 @@
 	id = "muscly"
 	msgGain = "Damn, where do you work out?"
 	msgLose = "You feel like you worked out at a library."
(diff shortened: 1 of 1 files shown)


ZeWaka:
The belt should use it then, copied code is jank

PR #25198, opened 2025-11-17
Dowsing Rods give sound and text notification when placed on a turf that can be pinned

phoiii:
[A-Engineering] [C-QoL]
## About the PR 
Adds a sound when a Dowsing rod is placed on a turf that if stomped will pin the Hotspot. currently uses sound/machines/found.ogg
Also slaps a text onto the end of the normal text it spits out when on a pinnable turf currently "Location able to be pinned"
## Why's this needed? 
Hotspots can be pinned in a 3x3 area around the true center by default, but there is no way of telling this except through luck or already knowing where the true center is.
This will give much needed feedback and help when the hotspot is inside a station wall.
## Testing 

https://github.com/user-attachments/assets/316433c0-175c-41bd-90a2-1984af3c2409
Screenshot 2025-11-16 202134

## Changelog 

```changelog
(u)Phoi
(+)Adds sound notification to dowsing rod when placed on a pinnable turf
(+)Adds text onto the pre-existing message when dowsing rod is placed on a pinnable turf
```

### DIFF
--- a/code/datums/controllers/sea_hotspot_controls.dm
+++ b/code/datums/controllers/sea_hotspot_controls.dm
@@ -578,10 +578,8 @@ TYPEINFO(/obj/item/heat_dowsing)
 					if (src.loc == center)
 						true_center += 1
 
-					if (BOUNDS_DIST(src, H.center.turf()) == 0) //variable for determining wether to play a sound if turf is able to be pinned
+					if (BOUNDS_DIST(src, H.center.turf()) == 0) //variable for determining whether to play a sound if turf is able to be pinned
 						pinnable_spot = TRUE
-					else
-						pinnable_spot = FALSE
 
 					var/d = GET_DIST(src.loc,center)
 					if (d < dist_last)


code/datums/controllers/sea_hotspot_controls.dm
```diff
@@ -572,11 +572,17 @@ TYPEINFO(/obj/item/heat_dowsing)
...
+				var/pinnable_spot = "not_pinnable"
 				for (var/datum/sea_hotspot/H in hotspot_controller.get_hotspots_list(src.loc))
 					var/turf/center = H.center.turf()
 					if (src.loc == center)
 						true_center += 1
 
+					if (BOUNDS_DIST(src, H.center.turf()) == 0) //variable for determining wether to play a sound if turf is able to be pinned
+						pinnable_spot = "pinnable"
+					else
+						pinnable_spot = "not_pinnable"
```
ZeWaka: use true/false not weird strings

also no need to set to not if that's the default

phoiii: ah yeah i see, all fixed now

code/datums/controllers/sea_hotspot_controls.dm
```diff
@@ -572,11 +572,17 @@ TYPEINFO(/obj/item/heat_dowsing)
 				speak_count = 0
 				var/val = 0 //my friend valerie
 				var/true_center = 0
+				var/pinnable_spot
```
pgmzeta:
```suggestion
				var/pinnable_spot = FALSE
```

code/datums/controllers/sea_hotspot_controls.dm
```diff
@@ -572,11 +572,17 @@ TYPEINFO(/obj/item/heat_dowsing)
...
 				for (var/datum/sea_hotspot/H in hotspot_controller.get_hotspots_list(src.loc))
 					var/turf/center = H.center.turf()
 					if (src.loc == center)
 						true_center += 1
 
+					if (BOUNDS_DIST(src, H.center.turf()) == 0) //variable for determining wether to play a sound if turf is able to be pinned
+						pinnable_spot = TRUE
+					else
+						pinnable_spot = FALSE
```
pgmzeta:
```suggestion
					if (BOUNDS_DIST(src, H.center.turf()) == 0) //variable for determining whether to play a sound if turf is able to be pinned
						pinnable_spot = TRUE
```

phoiii:
> Two small clean-up tweaks. There's probably some better logic flow possible around where `pinnable_spot` is read but it works as-is.

tested and commited, thankies for suggestion

PR #25199, opened 2025-11-17
Makes .38 Special variants purchasable by non-Detective traitors

lyzmat:
[game-objects] [balance] [QoL]
## About the PR 
made in response to some of the feed back iv been seeing on discord

this pr removes the Detective job requirement from the two variants of .38 special that are available in the traitor uplink, 
while also making their descriptions more clearly explain that the kestrel revolver can take .38 special rounds

## Why's this needed? 
people should be able to use cool bullets more often!

and a lot of people dont seem to know that the kestrel can take 38 special, im talking people who have been playing this game for multiple years and thousands of rounds. and its not their fault, you wouldnt really know unless you tried it or somebody told you

it shouldnt be this way !!

## Testing 

image
image

## Changelog 

```changelog
(u)lyzmat
(+)Traitor .38 Special variants are no longer locked to Detective operatives.
```

### DIFF
--- a/code/obj/item/uplinks/integrated_uplinks.dm
+++ b/code/obj/item/uplinks/integrated_uplinks.dm
@@ -108,16 +108,16 @@
 			for (var/O in src.items_objective)
 				var/datum/syndicate_buylist/I3 = src.items_objective[O]
 				src.menu_message += "<tr><td><A href='byond://?src=\ref[src];buy_item=\ref[src.items_objective[O]]'>[I3.name]</A> ([I3.cost])</td><td><A href='byond://?src=\ref[src];abt_item=\ref[src.items_objective[O]]'>About</A> [I3.max_buy == INFINITY  ? "" :"([src.purchase_log[I3.type] ? src.purchase_log[I3.type] : 0]/[I3.max_buy])"]</td>"
-		if (src.items_telecrystal && islist(src.items_telecrystal) && length(src.items_telecrystal))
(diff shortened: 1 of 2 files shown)


CelineTheYeen:
I'm all for this especially as it makes zipguns even a tiny bit more likely to be used though in an ideal world these would probably be at the bottom of the list (above job gear) in their own lil section instead of being at the very very top as they are now but that's a lot more effort sadly.

TobleroneSwordfish (changes requested):
I like this but agree that it being at the top isn't great, either add a new section for ammo or fudge the sort and just hide it at the end.

lyzmat:
okie dokie

lyzmat:
image
ok i did it !!

lyzmat:
more indepth list of changes (my commit messages arent very specific ;-;)
- new variable "ammo" in syndicate_buylist.dm that will add a given item to the ammo list
- added the new group in integrated_uplinks.dm and cleaned up some surrounding code (leave it better than you found it yadda yadda)
- added check for ammo variable in uplink_parent.dm that adds to ammo group
realising now that this may cause issues with ghost vr so i hav to do more coding ;-;

lyzmat:
ok wew nevermind 
image

lyzmat:
telecrystals should probably come after the ammo now that i think about it but it can wait until tomorrow

lyzmat:
image

PR #25203, opened 2025-11-17
Allow arcplating if the material is the default material

LorrMaster:
## About the PR
- Allows items to be arcplated if their material ID is the same as their `default_material`
- Items now use the `can_arcplate` variable to prevent plating by default
- Marked a bunch of items as not being platable where it made sense or might cause potential issues

## Why's this needed?
- Allows the expansion of the number of reclaimable items without locking those items out of being plated.

## Testing
Screenshot 2025-11-17 113838

## Changelog
```changelog
(u)LorrMaster
(*)Certain items can now be arcplated if their material is their default material
```

### DIFF
--- a/code/obj/machinery/arc_electroplater.dm
+++ b/code/obj/machinery/arc_electroplater.dm
@@ -107,6 +107,9 @@ TYPEINFO(/obj/machinery/arc_electroplater)
 		if (src.target_item)
 			boutput(user, SPAN_ALERT("There is already something in [src]!"))
 			return
+		if (W.w_class > src.max_wclass || istype(W, /obj/item/storage/secure) || W.anchored)
+			boutput(user, SPAN_ALERT("There is no way that could fit!"))
+			return
 		if (istype(W, /obj/item/grab))
 			boutput(user, SPAN_ALERT("That wouldn't possibly fit!"))
 			return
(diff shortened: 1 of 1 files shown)


code/obj/machinery/arc_electroplater.dm
```diff
@@ -107,27 +107,28 @@ TYPEINFO(/obj/machinery/arc_electroplater)
...
-
 		if (istype(W, /obj/item/grab))
 			boutput(user, SPAN_ALERT("That wouldn't possibly fit!"))
 			return
-
 		if (istype(W, /obj/item/implant))
 			boutput(user, SPAN_ALERT("You can't plate something this tiny!"))
 			return
-
-		if (W.w_class > src.max_wclass || istype(W, /obj/item/storage/secure) || W.anchored)
```
TobleroneSwordfish, line 122: ?

TobleroneSwordfish (changes requested):

code/obj/machinery/arc_electroplater.dm
LorrMaster: That if statement was just moved down so that the item type checks would be grouped together. Github just makes it look as if the line was removed entirely until you read further. I've moved it again so that it appears before the item type checks.

JORJ949 (approved):

PR #25206, opened 2025-11-19
Last names are now optional

ThatFiveGuys:
## About the PR 
In the character selection screen you can now have last names be blank simlar to middle names.
image

## Why's this needed? 
This will be mostly used for gimmicks tbh but I think allowing characters to not have a last name opens up a lot of freedom in character creation. For an example, if you wanted to crawl around in maintence on classic, you can now name yourself maintcrawler instead of maint crawler. I want to see players go wild with names so to speak so I made this PR.

## Testing 
Technically to be done, 

Technically will do in a second just wanted to do this bit first. will update in a second.
Tested saving last names, and also in a round with clerical error and computers. 
## Changelog 

```changelog
(u)Ikea
(*)Last names can now be blank
```


Kyle2143:
Uhh, I don't really like that as a general thing... Seems to gimmicky and will lead to a bunch of people with even more ridiculous names than we already allow. And we (for the most part) allow pretty ridiculous names already.

TobleroneSwordfish:
To expand on what Kyle said with some further comments from admin chat:
- We don't want to encourage people on the RP servers to pick the name "Mushroom" and sit in maint all shift, the baseline expectation should always be that you are an employee with a job and paperwork.
- Roles that fit having gimmick names such as clown and mime (also pickle) already get a name change prompt on joining that lets them pick a one word name if they like and said roles are intentionally limited in each round.

PR #25215, opened 2025-11-20
Most foods cause you to sweat cleanables that dry after time instead of chempools

CalliopeSoups:
## About the PR 

Does what it says on the tin. dropSweat() comes with a var you can set TRUE to sweat chempools, otherwise you sweat water cleanables instead.

## Why's this needed? 

I've noticed complaints about the amount you sweat from basic foods lately. I think more foods could take food_sweaty_big still but this is a good baseline.

## Testing 

Tested each sweat status on different foods, all's well.

## Changelog 

```changelog
(u)CalliopeSoups
(+)Sweating is now less extreme for most foods.
```

### DIFF
--- a/code/modules/status_system/statusEffects.dm
+++ b/code/modules/status_system/statusEffects.dm
@@ -89,19 +89,20 @@
 		 *  targetTurf should be left default
 		 *  Turning sweatpools on causes sweat chempools instead of cleanables
 		 */
-	proc/dropSweat(var/datum/reagent/sweatReagent, var/sweatAmount = 5, var/sweatChance = 5, var/turf/targetTurf = get_turf(owner), var/sweatpools = FALSE)
+	proc/dropSweat(var/sweatReagent, var/sweatAmount = 5, var/sweatChance = 5, var/turf/targetTurf = get_turf(owner), var/sweatpools = FALSE)
 		if (!prob(sweatChance))
 			return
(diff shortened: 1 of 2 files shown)


code/modules/status_system/statusEffects.dm
```diff
@@ -87,16 +87,24 @@
...
-	proc/dropSweat(var/sweatReagent, var/sweatAmount = 5, var/sweatChance = 5, var/turf/targetTurf = get_turf(owner))
+	proc/dropSweat(var/datum/reagent/sweatReagent, var/sweatAmount = 5, var/sweatChance = 5, var/turf/targetTurf = get_turf(owner), var/sweatpools = FALSE)
+		if (!prob(sweatChance))
+			return
 		var/datum/reagents/tempHolder = new
-		if (prob(sweatChance))
-			tempHolder.add_reagent(sweatReagent, sweatAmount)
-			targetTurf.fluid_react_single(sweatReagent,sweatAmount)
+		if (sweatpools)
+			tempHolder.add_reagent(sweatReagent.id, sweatAmount)
```
TobleroneSwordfish: This won't work, you're trying to access a variable of a type rather than an instance. Generally we prefer passing reagent IDs around (as it was before), because getting instances is a faff and requires you to go via the cache.

PR #25231, opened 2025-11-22
Handle multiple movement controllers

Valtsu0:
## About the PR 
Movement controllers are stored in a list and when the active movement controller is removed use the most recent controller in the list

## Why's this needed? 
fixes not being able to drive a pod after using the detective's deluxe camera without re-entering.

## Testing 
Tried diffrent combinations of using the camera and getting in and out of a pod

### DIFF
--- a/code/mob.dm
+++ b/code/mob.dm
@@ -202,7 +202,6 @@ TYPEINFO(/mob)
 
 	var/last_cubed = 0
 
-	var/datum/movement_controller/override_movement_controller = null
 	var/list/movement_controller_list = list()
 
 	var/dir_locked = FALSE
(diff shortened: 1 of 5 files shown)


code/mob.dm
```diff
@@ -203,6 +203,7 @@ TYPEINFO(/mob)
 	var/last_cubed = 0
 
 	var/datum/movement_controller/override_movement_controller = null
+	var/list/movement_controller_list = list()
```
TobleroneSwordfish: Could use comments to explain the queue system, otherwise these two vars seem like they're trying to do the same thing

code/mob.dm
```diff
@@ -3515,3 +3517,27 @@ TYPEINFO(/mob)
...
+		movement_controller_list = list()
+		override_movement_controller = null
+		return
+
+	movement_controller_list.Remove(movement_controller)
+
+	if (override_movement_controller != movement_controller)
+		return
+
+	if (movement_controller_list.len)
```
TobleroneSwordfish:
```suggestion
	if (length(src.movement_controller_list))
```
`length` and explicit `src` are always preferred.

code/mob.dm
Valtsu0: That is because they kinda are. I have removed the override_movement_controller

PR #25237, opened 2025-11-23
Cogmap1 evidence room fixes.

smallsandman:
## About the PR 
Fixes some issues i noticed while playing, misaligned lights, misaligned tile, and a lack of a forcefield on twisted wreckage. additionally adds some new windows above the table.
## Why's this needed? 
fixes are good!
## Testing 
checked everything out in-game, looked all good to me.

```

### DIFF
--- a/maps/cogmap.dmm
+++ b/maps/cogmap.dmm
@@ -39887,9 +39887,7 @@
 	pixel_x = 8;
 	pixel_y = -5
 	},
-/turf/simulated/floor/red/side{
-	dir = 10
-	},
+/turf/simulated/floor/red/side,
 /area/station/security/main)
 "fYn" = (
 /obj/cable{


smallsandman:
> I think you accidentally rotated this floor tile:

this was actually purposefully done, in the original cogmap security design the floor under the crate was incorrect, for who knows how long, i wanted to honor that here. though i can fix it if needs be.

flappybatpal:
> > I think you accidentally rotated this floor tile:
> 
> this was actually purposefully done, in the original cogmap security design the floor under the crate was incorrect, for who knows how long, i wanted to honor that here. though i can fix it if needs be.

This is way too niche to be worthy of homage IMO

smallsandman:
> This is way too niche to be worthy of homage IMO

already fixed
the tile is now normal.

JORJ949 (approved):

PR #25247, opened 2025-11-26
Mining alerts

Egregorious:
## About the PR 

Added an AI-assistant styled alert system for mining, which plays sound alerts for the user when the player mines a gem and changes the excavation message. 

The assistant is active when wearing an engineering or industrial mining helmet, or when a borg has a mining module equipped. 

In order to add alerts to the borgs, borg modules now apply equipment properties to the borg they're inserted into.

scale g6 by Leszek_Szary -- https://freesound.org/s/171644/ -- License: Creative Commons 0

## Why's this needed? 

There could be more feedback for finding valuable stuff when mining, especially since uqill is one of the major parts of round progression for the job, and is fairly rng dependent. 

## Testing 

Tested with a few of the mining tools. The sound gets drowned out when using the powered pick/hammer, but that might be an issue with those tools being too loud in general. It seems to work well with the drill, gauntlet, and unpowered tools. 

https://github.com/user-attachments/assets/4ebb67ad-9608-4090-a858-28cc00f9ac65

https://github.com/user-attachments/assets/379d3102-d1ad-4548-acae-6a4865fc93f5

Screenshot 2025-12-01 172339

## Changelog 

```changelog
(u)Egregorious
(+)Added an AI assistant to mining helmets which plays sound alerts for certain mining events. 
```

### DIFF
--- a/code/modules/robotics/robot/module/mining.dm
+++ b/code/modules/robotics/robot/module/mining.dm
@@ -9,7 +9,7 @@
 	mailgroups = list(MGD_MINING, MGD_CARGO, MGO_SILICON, MGD_PARTY)
 	alertgroups = list(MGA_MAIL, MGA_RADIO, MGA_SALES, MGA_SHIPPING, MGA_CARGOREQUEST, MGA_DEATH)
 
-/obj/item/robot_module/New()
+/obj/item/robot_module/mining/New()
 	..()
 	setProperty("mining_alerts")
 


virvatuli-gh (commented):
I'm not too big on the audio style choice (sounds too gamey to me) but I get it that it's hard to make usable sounds actually audible in no-air environments with the way the sounds are processed. I'd prefer the sound be tied to a wearable item instead (helmet maybe?), so it's a perk of actually being a miner/using mining gear, kind of like how the geiger counter is, which would also justify the actual effect style choice. The suit alerting the miner that a thing of value is there with a tone!

Egregorious:
That's a good idea! I turned it into a sort of Jarvis style thing and attached it to mining helmets. It also changes the excavation message specifically for the user, styled as if the AI is saying the notification in your ear.

Issue #24609, opened 2025-09-07
Non-Whitelisted admins cannot preference HoS


JORJ949:
### Labels

[Minor]

### Description

If an admin is not HoS whitelisted, they cannot add HoS to their preferences

### Steps to reproduce

1. Be an admin
2. Open occupation TGUI
3. HoS is disabled in preferences.

### Expected Behavior

Admins should bypass role requirements

### Additional Information & Screenshots

<img width="784" height="652" alt="Image" src="https://github.com/user-attachments/assets/3b85e4b3-6141-49c5-b389-b7b39ae6294d" />

JORJ949:
Fixed in #25256

PR #25256, opened 2025-11-28
Fix new mentors being unable to play HoS

JORJ949:
## About the PR 
Refactors how the NT whitelist is checked to be a client proc "can_play_whitelisted_roles" rather than a direct check against the global "NT" list. Returns true if the client's ckey is in the global NT list, the client is an admin, or its friday and the player is a mentor.
Removes mentors being added to global NT list on fridays.

## Why's this needed? 
New mentors added via goonhub instead of mentors.txt were unable to play HoS as they aren't present in the list of mentors until they connect, meaning they never got added to the NT list at world init.
Stops mentors showing as HoS whitelisted in player options on fridays.

## Testing 
Before adding the isadmin() check, tested:
Not being mentor on friday, couldn't pick HoS
Being mentor on friday, could pick HoS
Being mentor on non-friday (replaced IS_IT_FRIDAY with FALSE), couldn't pick HoS
Being HoS on non-friday, could pick HoS
image

### DIFF
--- a/code/client.dm
+++ b/code/client.dm
@@ -647,7 +647,7 @@
 	return (src.player?.mentor || src.holder) && src.player?.see_mentor_pms
 
 /client/proc/can_play_whitelisted_roles()
-	return NT.Find(src.ckey) || (IS_IT_FRIDAY && src.is_mentor())
+	return NT.Find(src.ckey) || (IS_IT_FRIDAY && src.is_mentor()) || isadmin(src)
 
 var/global/curr_year = null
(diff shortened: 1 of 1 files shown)


Mister-Moriarty:
Thank you! I've been meaning to clean this up ever since the two TGUI PRs were merged.
Does this fix #24609?

JORJ949:
> Thank you! I've been meaning to clean this up ever since the two TGUI PRs were merged. Does this fix #24609?

It should do yes

PR #25259, opened 2025-11-29
Removes forensic scanner from sec PDAs

Munien7:
## About the PR 
This PR removes the forensic scan program from Security Officer, Captain and the generic Head PDA cartridges. It also removes the forensic PDAs sold in the security clothing vendor.
## Why's this needed? 
Currently, there's not much incentive to prioritize getting the Detective to do forensic work (outside of a slightly better scanner) or, in the absence of a detective, spending a utility token for a forensic scanner (other than the search function, which in mine and others experience is still not worth the credit). It is my hope that this will either encourage cooperation with detectives or sacrificing utility for a scanner if you wish to do forensics.
## Testing 
Screenshot 2025-11-29 141445

## Changelog 

```changelog
(u)Munien
(+)Removed the Forensic Scan program from several PDA cartridges (Security Officer, Captain, Generic Head) and limited the availability of Forensic PDAs.
```

### DIFF
--- a/code/modules/chemistry/tools/dispensers.dm
+++ b/code/modules/chemistry/tools/dispensers.dm
@@ -134,14 +134,6 @@
 					qdel(src)
 		return
 
-	attackby(obj/item/P, mob/user, silent = FALSE)
-		if (istype(P, /obj/item/magnifying_glass))
-			boutput(user, SPAN_NOTICE("You angle the light through the magnifying glass towards the ants."))
-			sleep(3 SECONDS)
-			boutput(user, "You burn the ants to a crisp.</i>")
-			make_cleanable(/obj/decal/cleanable/ash,src.loc)
-			qdel(src)
-
 /obj/reagent_dispensers/cleanable/spiders
 	name = "spiders"
 	desc = "A bunch of spiders."


pgmzeta:
I think it's okay on HoS/NTSC PDA carts, personally. PDA programs are weak balance points because they can be copied so easily, but I still like this PR. SecAss default utility increased as well since they spawn with a scanner.

code/obj/item/device/pda2/cartridges.dm
```diff
@@ -245,7 +244,6 @@ TYPEINFO(/obj/item/disk/data/cartridge/syndicate)
 			src.root.add_file( new /datum/computer/file/pda_program/portable_machinery_control/portabrig(src))
 			src.root.add_file( new /datum/computer/file/pda_program/qm_records(src))
 			src.root.add_file( new /datum/computer/file/pda_program/fileshare(src))
-			src.root.add_file( new /datum/computer/file/pda_program/scan/forensic_scan(src))
```
JORJ949, line 248: IMO the HoS should keep it because they're the boss and should be able to do the role of their subordinates like other command staff can (HoP gets mopfinder, CE gets mulebot and rockbox, etc). I'd instead remove it from the Captain and generic Heads cartridges, neither of these are security and arguably need it even less than officers.

JORJ949 (changes requested):
Also remove the Forensics PDA from the security apparel vendor (leave the one in the det's locker) or it'll just be bought every time (its only 200 credits).

code/obj/item/device/pda2/cartridges.dm
Munien7: I'm fine with this, I was on the fence about it but I had a lot of people of comments from players saying the HoS one should not have it either. I'll swap it when I have some time!

Munien7: I did also consider outright removal from the apparel vendor as well but I didn't see it as necessary. The very limited supply (iirc it only stocks 2) and the lack of it being abused currently (it's almost a straight upgrade from the base security cartridge minus portabrig controller and I don't really ever see secoffs take it) made me think this wasn't really necessary, but I'm happy to remove it

Munien7:
> I think it's okay on HoS/NTSC PDA carts, personally. PDA programs are weak balance points because they can be copied so easily, but I still like this PR. SecAss default utility increased as well since they spawn with a scanner.

I did consiser making it non-copyable but I think the only instance of that I saw was the detomatix I didn't think it was really dramatic enough to warrant it, especially with more powerful apps being copyable. It's not a solid fix but I hope it will at least encourage secass and detectives to have a little more value.

JORJ949:
> 
> I did consiser making it non-copyable but I think the only instance of that I saw was the detomatix I didn't think it was really dramatic enough to warrant it, especially with more powerful apps being copyable. It's not a solid fix but I hope it will at least encourage secass and detectives to have a little more value.

If the detective wants to share the only copy (HoS/NTSC I doubt will unless necessary) its on them when officers do their job for them tbh

Munien7:
> > I did consiser making it non-copyable but I think the only instance of that I saw was the detomatix I didn't think it was really dramatic enough to warrant it, especially with more powerful apps being copyable. It's not a solid fix but I hope it will at least encourage secass and detectives to have a little more value.
> 
> If the detective wants to share the only copy (HoS/NTSC I doubt will unless necessary) its on them when officers do their job for them tbh

Yep, I thought it wasn't really worth anything so dramatic since they would most likely be the ones sharing it.

code/obj/item/device/pda2/cartridges.dm
Munien7: > IMO the HoS should keep it because they're the boss and should be able to do the role of their subordinates like other command staff can (HoP gets mopfinder, CE gets mulebot and rockbox, etc). I'd instead remove it from the Captain and generic Heads cartridges, neither of these are security and arguably need it even less than officers.

I forgot to bring this up also, removing it from the valu-pak cartridge is also inadvertently affecting antags since there is a cartridge available in the listening post, which was the main reason I wanted to leave it while making this PR

Mister-Moriarty:
I'd advocate for the HoS and NTSC to keep the program, particularly NTSC, otherwise they'll be completely locked out of forensics, save for the rare occasion when the loose forensics scanner in Security hasn't been taken.

Munien7:
Re-added the forensic scan to the HoS/NTSC cart, removed it from the captain and "head" cart as requested. Also removed the 2 forensic PDA's purchasable through the clothing vendor as suggested.

I'd like to remove loose scanners from security areas on the maps as well, but that's out of scope for this current PR in my opinion.

IsaneGit:
I'd argue against removing it from the Value-PAK (Captain) cartridge. The majority of rounds I've played lack a HoS, and with the Captain being the head of the _entire_ station I'm just not a fan of another case where the whitelisted roles are above that.

This also hits traitors and sleeper agents as the Listening Post provides a couple Value-PAKs, which I've found vital for use with the Fingerprint traitor item specifically among other uses of the scanner.

Munien7:
> I'd argue against removing it from the Value-PAK (Captain) cartridge. The majority of rounds I've played lack a HoS, and with the Captain being the head of the _entire_ station I'm just not a fan of another case where the whitelisted roles are above that.
> 
> This also hits traitors and sleeper agents as the Listening Post provides a couple Value-PAKs, which I've found vital for use with the Fingerprint traitor item specifically among other uses of the scanner.

I don't necessarily agree with the first part. There is quite a few programs that the captain doesn't have access to that other jobs have (portasci, porta-brig, portamed, nanomed, MULE, power crystal, genebooth tracker, etc). The captain not having these doesn't make any other job "above them" by any means; I see no reason why this would be any different with the HoS/NTSC.

I do agree personally with the second point however, as it could be a useful tool for traitors, which I've also said elsewhere on this PR.

virvatuli-gh (approved):
Empowering the Detective is a good thing.

PR #25264, opened 2025-11-29
Fix trench mining volume weirdness

### LINKED ISSUE #16900: Mining submarine monstrously loud.
### Labels

[BUG][MINOR]

### Description

When mining with drills in a mining submarine, starting movement produces a sound far louder than the rest in the environment.

### Steps to reproduce

1.Board submarine equipped with mining drills.
2.Drill into and ride over rock
3. Start/Stop pattern produces unreasonably loud sound, far above any other in the surroundings.

### Expected Behavior

Didn't expect it to be so loud.

Reported by: EvaEvaEvaEva
Active test merges: #16744
Round log date: 2023-11-12-22-53
Reported on: 2023-11-12 23:04:06
Map: OSHAN

Egregorious:
## About the PR 

Set `special_volume_override` on asteroid floors to match the regular sea floors when on an underwater map. 

This prevents the asteroid floors located in the trench increasing the volume of sounds which occur on their tile, and destroying miner's eardrums. 

It's a hacky solution, but no more so than is already being done for brightness. Water maps should really get their own special rock floor tiles to use. 

## Why's this needed? 

This fixes a bug where mining in the trench is super loud. 

fixes #16900

## Testing 

Drove around in a sub and observed that the rock floors and the sea floors are at the same volume level. 

https://github.com/user-attachments/assets/a5b6fb5d-8e79-428a-a601-786100fc289b

### DIFF
--- a/code/obj/mining.dm
+++ b/code/obj/mining.dm
@@ -1561,6 +1561,7 @@ TYPEINFO(/turf/simulated/floor/plating/airless/asteroid)
 #ifdef UNDERWATER_MAP
 	fullbright = 0
 	luminosity = 3
+	special_volume_override = 0.62 // without this, the rock floors in the trench attenuate incorrectly and become very loud
 #else
 	luminosity = 1
 #endif


virvatuli-gh (approved):
Oh hell yeah, thank you for figuring this out.

PR #25265, opened 2025-11-29
Prevent null exception when contraband falls from burning storage

lyzmat:
[A-Game-Objects] [C-Bug]
## About the PR 
fixes incredibly niche bug where if a contraband item is in a box when it is set alight and destroyed, a runtime exception will be thrown because when the contraband item goes to remove the "contains contraband" property from the box, it attempts to access a null entity (the box does not exist, for it is destroyed)

accomplished by making the atom property check only fire if there actually is a user to fire at (not null)
## Why's this needed? 
needed to save the sanity of poor innocent contributors who have errors unrelated to their PR and think its their fault
this *shouldn't* affect gameplay in any way because behavior was as expected even when bug was present

## Testing 

https://github.com/user-attachments/assets/bf965b55-ef22-4572-b329-e486b634d1a0

note lack of "Cannot read null.atom_properties" error in console log when box finishes burning and is deleted

### DIFF
--- a/code/datums/components/contraband.dm
+++ b/code/datums/components/contraband.dm
@@ -70,7 +70,7 @@ TYPEINFO(/datum/component/contraband)
 	src.contraband_logic(owner, user, slot_mult)
 
 /datum/component/contraband/proc/removed(obj/item/owner, mob/user)
-	if(user != null)
+	if(user)
 		REMOVE_ATOM_PROPERTY(user, PROP_MOVABLE_VISIBLE_GUNS, src)
 		REMOVE_ATOM_PROPERTY(user, PROP_MOVABLE_VISIBLE_CONTRABAND, src)
 


lyzmat:
made draft pending figuring out what exactly i have to change to get it to pass checks

code/datums/components/contraband.dm
```diff
@@ -70,8 +70,8 @@ TYPEINFO(/datum/component/contraband)
 	src.contraband_logic(owner, user, slot_mult)
 
 /datum/component/contraband/proc/removed(obj/item/owner, mob/user)
-	REMOVE_ATOM_PROPERTY(user, PROP_MOVABLE_VISIBLE_GUNS, src)
-	REMOVE_ATOM_PROPERTY(user, PROP_MOVABLE_VISIBLE_CONTRABAND, src)
+	REMOVE_ATOM_PROPERTY(user?, PROP_MOVABLE_VISIBLE_GUNS, src)
+	REMOVE_ATOM_PROPERTY(user?, PROP_MOVABLE_VISIBLE_CONTRABAND, src)
```
TobleroneSwordfish: This isn't how DM works, if you want this code not to run when user is null then you will need to use an if statement.

code/datums/components/contraband.dm
```diff
@@ -70,8 +70,9 @@ TYPEINFO(/datum/component/contraband)
 	src.contraband_logic(owner, user, slot_mult)
 
 /datum/component/contraband/proc/removed(obj/item/owner, mob/user)
-	REMOVE_ATOM_PROPERTY(user, PROP_MOVABLE_VISIBLE_GUNS, src)
-	REMOVE_ATOM_PROPERTY(user, PROP_MOVABLE_VISIBLE_CONTRABAND, src)
+	if(user != null)
```
JORJ949:
```suggestion
	if(user)
```
Functionally the same, but cleaner. A Thing is truthy, null is falsey

PR #25271, opened 2025-11-30
Refactors Spatial Hashmaps

Mister-Moriarty:
## About The PR:
Spatial hashmaps have been completely rewritten. Note that within the context of this PR, "hashmap" will always refer to "spatial hashmap".

Spatial hashmaps now generically support the hashing of datums: hashmap entries no longer require to be atoms with a physical position, nor do atom entries require to be represented by their own map position; instead another atom may be used to represent an entry's physical position.

Hashmap entries are no longer updated by looping over all registered entries and rebuilding the hashmap from scratch; instead an entry's cell within the hashmap is recalculated when that entry's tracked atom moves, determined through complex signals.

In addition to the above changes, several smaller performance optimisations have been made, most prominent being that the flat 1D hashmap and hashfunction has been substituted by a 3D grid containing lists of atoms for a cell at (x, y, z) within the grid. This significantly reduces the lookup time for cells and allows for cells to be iterated over in a similar manner to how the `block()` proc operates.

A few additional procs for looking up nearby cells have been implemented; the original lookup function using a Manhattan norm has been retained, but three new functions utilising a supremum norm have been implemented: a standard function; one that considers vistargets; and one that considers exact ranges, not just nearby cells. 

## Why Is This Needed?
More generic and extensible spatial hashmaps allows for them to be more easily used for performance applications outside of client querying; I plan to refactor say channels to use hashmaps in future.

## Testing:
All current uses of hashmaps appear to work as expected, this includes hashmaps used for `playsound()`, ranch animals, and terrain generation.

Below is a profiler snippet of a test case comparing the performance of the old spatial hashmaps to the new ones. Every camera on Atlas (of which there are 92) was registered to a single hashmap, and each camera was then made to query nearby cameras from the hashmap every tick. It should be noted that this is the best case scenario for old hashmaps, as due to the high call frequency, the hashmap was very rarely rebuilt. This is a good approximation for how the current client hashmaps perform when used with `playsound()`.

image
A comparison of the new hashmaps (in yellow) against the old (in red). Note that the scale is logarithmic.


Laboredih123:
it didnt seem *immediately* sinful on first passover

code/modules/spatial_hashing/_spatial_hashmap_parent.dm
```diff
@@ -0,0 +1,306 @@
...
+	 *	Note that these coordinates are in grid space, not world space.
+	 */
+	VAR_PROTECTED/list/list/list/list/datum/hashmap = null
+
+	/// An associative list of atoms being using to represent physical positions, and their associated signal subscription count.
+	VAR_PROTECTED/list/atom/tracked_atoms_with_subcount = null
+	/// An associative list of atoms, indexed by the hashmap entry using them to represent a physical position.
+	VAR_PROTECTED/list/atom/atoms_by_entry = null
+	/// An associative list of lists of hashmap entries, indexed by the atom being used to represent a physical position.
+	VAR_PROTECTED/list/list/atom/entries_by_atom = null
```
ZeWaka: why not alist then

Mister-Moriarty: Because typed alists aren't supported by SpacemanDMM yet.

ZeWaka: Hmm. I'd rather have the performance benefits of alists for now with a comment about the type that we can go back to fix later, tbh. let me look into it

ZeWaka (approved):

PR #25279, opened 2025-12-01
Adds brown satchels/backpacks to uniform manufacturers

Microoowaaave:
## About the PR 
Adds brown and studded satchels/backpacks to uniform manufacturers. The studded variants cost some metal, and can only be printed if the fabricator is hacked.

## Why's this needed? 
These are usually spacebux purchases, but I've always found it odd that there's no way (that I know of) to obtain them besides that. Thought it'd be neat if there was.

## Testing 

Hacked:

## Changelog 

```changelog
(u)Microoowaaave
(+)Added brown satchels/backpacks to uniform manufacturers.
```

### DIFF
--- a/code/datums/manufacturing.dm
+++ b/code/datums/manufacturing.dm
@@ -2686,15 +2686,6 @@ ABSTRACT_TYPE(/datum/manufacture/aiModule)
 	time = 10 SECONDS
 	category = "Clothing"
 
-/datum/manufacture/backpack_studded
-	name = "Studded Backpack"
-	item_requirements = list("fabric" = 8,
-							 "metal" = 1)
-	item_outputs = list(/obj/item/storage/backpack/empty/studdedblack)
-	create = 1
-	time = 10 SECONDS
-	category = "Clothing"
-
 /datum/manufacture/satchel
 	name = "Satchel"
 	item_requirements = list("fabric" = 8)
@@ -2731,15 +2722,6 @@ ABSTRACT_TYPE(/datum/manufacture/aiModule)
 	name = "Brown Satchel"
 	item_requirements = list("fabric" = 8)
 	item_outputs = list(/obj/item/storage/backpack/satchel/empty/brown)
-	create = 1
-	time = 10 SECONDS
-	category = "Clothing"
-
-/datum/manufacture/satchel_studded
-	name = "Studded Satchel"
-	item_requirements = list("fabric" = 8,
-							 "metal" = 1)
-	item_outputs = list(/obj/item/storage/backpack/satchel/empty/studdedblack)
 	create = 1
 	time = 10 SECONDS
 	category = "Clothing"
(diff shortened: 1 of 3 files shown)


mordent-goonstation:
Is there a precedent of other spacebux items being available for cheap in round? Ideally spacebux purchases should actually be something that's tricky to get ahold of, as that makes them decidedly more interesting.

Microoowaaave:
> Is there a precedent of other spacebux items being available for cheap in round? Ideally spacebux purchases should actually be something that's tricky to get ahold of, as that makes them decidedly more interesting.

Off the top of my head? Uniform fabs can also make bandanas/handkerchiefs (though plain compared to the ones purchasable with spacebux), and the crayon creation device is fairly easy to obtain, as artistic toolboxes can be ordered from cargo or taken from the chaplain's office.

I agree to an extent, though, which is why the studded ones need the fab to be hacked, and why I didn't include any of the other spacebux purchasable bags. I really don't think the brown bags are unique/interesting enough as is (as much as I love them) to warrant being spacebux exclusive anyhow, especially considering the red, blue, and green ones being printable already.

M-Earthfire:
Not really a fan. Why have a spacebux pursache when you can get the item directly at the start of a round?

Munien7:
> Is there a precedent of other spacebux items being available for cheap in round? Ideally spacebux purchases should actually be something that's tricky to get ahold of, as that makes them decidedly more interesting.

off the top of my head: handkerchiefs, bee eggs, sticker boxes, crayon creation device, dramatic horn, saxophone, harmonica, trumpet, alt jumpsuits, hoodies. Might be a few others. Seeing as how other ones are available through hacking vendors or manufacturers, etc, I don't personally see the harm this would do.

Spyritdragon:
I think there's an argument to be made to maybe keep some of the cosmetic spacebux purchases harder to get in-game, but I have to confess I've never been a fan in the first place of locking fairly simple cosmetic options behind spacebux purchases. 

I think it works well for the weirder/more unique/off-brand-y one that is the studded one to be accessible in hacked fabricators, but brown just feels like another colour just like every other backpack or satchel we do have printable, so I think it's nice for the brown ones to join them.

virvatuli-gh:
Broadly against adding "cool looking" fashion items to fabs. Fabricators are for the mundane, the standard, the stock. If you want fashion, you should pay for it, either through effort or currency.

Microoowaaave:
> Broadly against adding "cool looking" fashion items to fabs. Fabricators are for the mundane, the standard, the stock. If you want fashion, you should pay for it, either through effort or currency.

To be fair, the fab has to be hacked for the actual cool looking item, which is atleast some effort. As for the brown bags, they seem mundane enough to me that this doesn't really apply there.

CelineTheYeen:
I really don't see why brown bags were made into a spacebux purchase, they are an extremely mundane clothing option that looks less Wacky:tm: than the printable backpack colors, I don't particularly care much for the studded bags but it's fine realistically. Spacebux as a whole are very flawed and I'm not a big fan of the way purchases work but that's a different discussion.

PR #25283, opened 2025-12-02
Removes the need for wirecutters when unstacking cable coils

Driftered:
## About the PR
This PR allows players to unstack cable coils by using an empty hand on the stack instead of requiring wirecutters. 

Also happens to fix a bug that caused a user's pocket inventory to be considered both empty and in-use if the user stacks coil from there. Previously, users could not put newly-stacked coils back into their pocket without dropping the coils or swapping hands, as the game would begin laying coils rather than placing the coils into the pocket. By adding `user.u_equip(src)` to the `attackby` proc, this glitch is eliminated.

## Why's this needed? 
Currently, cables can be stacked by hand, but in order to _unstack_ them, a user must use wirecutters on the stack. This is not mentioned anywhere (such as a help message in the description), nor does it make sense to require the extra step when the player can already do the reverse action by hand. 

## Testing
Gathered several coils from an electrical toolbox and tested various functions like stacking, unstacking, laying coils and then stacking onto the pile in use, and ensuring the item is unequipped from the player's pockets if stacked from there.

## Changelog 

```changelog
(u)Driftered
(*)Cable coils can now be unstacked by hand.
```

[A-GAME OBJECTS] [C-CODE QUALITY]

### DIFF
--- a/code/obj/item/cable_coil.dm
+++ b/code/obj/item/cable_coil.dm
@@ -103,6 +103,25 @@ obj/item/cable_coil/abilities = list(/obj/ability_button/cable_toggle)
 	after_stack(atom/movable/O as obj, mob/user as mob, var/added)
 		boutput(user, SPAN_NOTICE("You finish coiling cable."))
 
+	attack_hand(mob/user)
+		if((user.r_hand == src || user.l_hand == src) && src.amount > 1)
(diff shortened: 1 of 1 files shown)


AHHH-IM-DYING-HELP-ME-AHHH-IT-BURNS:
don't mind me ripping heavy cables apart with my own hands and having them being fully functional afterwards

PR #25284, opened 2025-12-03
Bring back Destiny for a day

TobleroneSwordfish:
## About the PR 
Re-adds and patches up Destiny so it actually compiles.
## Why's this needed? 
It's the 10 year anniversary of the RP servers! This means we need to bring back 24/7 destiny hopline RP. It's essential.
## Testing 
image
it do compile

## Changelog 

```changelog
(u)LeahTheTech
(*)Re-added Destiny for the day!
```

### DIFF
--- a/code/map.dm
+++ b/code/map.dm
@@ -1017,7 +1017,7 @@ var/global/list/mapNames = list(
 	walls = /turf/simulated/wall/auto/supernorn
 	rwalls = /turf/simulated/wall/auto/reinforced/supernorn
 
-	airlock_style = "pyro"
+	airlock_style = "gannets"
 
 	windows = /obj/window/auto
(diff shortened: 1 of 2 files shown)


Laboredih123:
dear god

Laboredih123:
but also CHECK FAILING!!!!!!

TobleroneSwordfish:
Common mapdiffbot L

Sord213:
still have some more to fix. mostly, namely: all the cameras

Sbmhawk:
While we're at it, why not throw in Density as well? :devious:

Gannets:
We should bring back my old perspective walls and doors, that'd be funny.

Laboredih123:
rekt

PR #25285, opened 2025-12-03
adds bricks to hacked general fabricators

RSOD404:
[Feature] [Game-Objects]

## About the PR 
adds bricks to general fabricators at the cost of 10 rocks, if they are hacked

## Why's this needed? 
bricks are not renewable in any way shape or form, which means the only source of bricks is from map specific spawns or random rooms. this means playing impromptu brick dodgeball in the hall with the angry staffies is all up to chance

## Testing 

https://file.garden/Z5gQkrJ7CWwAEOa2/videos/bricks%20bricks%20and%20more%20god%20fucking%20damned%20bricks.mp4

## Changelog 

```changelog
(u)Red Screen of Death
(+)General manufacturers can now make bricks when hacked.
```

### DIFF
--- a/code/datums/manufacturing.dm
+++ b/code/datums/manufacturing.dm
@@ -1113,6 +1113,14 @@
 	time = 5 SECONDS
 	category = "Tool"
 
+/datum/manufacture/brick
+	name = "Brick"
+	item_requirements = list("rock" = 100)
+	item_outputs = list(/obj/item/brick)
+	create = 1
+	time = 10 SECONDS
+	category = null
+
 /******************** Medical **************************/
 
 /datum/manufacture/scalpel
--- a/code/obj/machinery/manufacturer_subtypes.dm
+++ b/code/obj/machinery/manufacturer_subtypes.dm
@@ -82,7 +82,8 @@
 		/datum/manufacture/stapler,
 		/datum/manufacture/bagpipe,
 		/datum/manufacture/fiddle,
-		/datum/manufacture/whistle)
+		/datum/manufacture/whistle,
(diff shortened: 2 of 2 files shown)


code/datums/manufacturing.dm
```diff
@@ -1113,6 +1113,14 @@
 	time = 5 SECONDS
 	category = "Tool"
 
+/datum/manufacture/brick
+	name = "Brick"
+	item_requirements = list("rock" = 100)
+	item_outputs = list(/obj/item/brick)
+	create = 1
+	time = 10 SECONDS
+	category = null
```
444Portal, line 1122: bricks are a tool of destruction. they should be under the tool category, what da hell..

though yeah sure why not put them in the fabricator

ThatFiveGuys:
10 feels too expensive considering the size of a brick they should be like 5 rock

TobleroneSwordfish:
Broadly against, just because I think bricks being a rare and dangerous find on a space station is funny (and justifies them being pretty strong)

code/datums/manufacturing.dm
RSOD404: left the bricks under misc since there's a bunch of random stuff down there especially when hacked (glass bottles, instruments, bike horn) so i figured adding it there would be more fitting (although, putting bricks in the tool section would be funny..)

virvatuli-gh:
The Brick is an item of mystery and power that exists for staff assistants to fight over. Making it easily available would be a terrible thing, empowering the staff assistants to a place in the station hierarchy they should never come close to.

PR #25291, opened 2025-12-04
adds missing mug rack to oshan security.

smallsandman:
## About the PR 
re-adds the security coffee mug rack i seemed to have forgotten about.
## Why's this needed? 
to be more inline with the other coffee makers having racks. like it was prior.
## Testing 
image

```


virvatuli-gh (approved):
Good, security doesn't need more reasons to descend into barbarism, like pouring coffee on the floor to slurp it up.

PR #25309, opened 2025-12-10
Adds figurine for snugglycactus

snugglycactus:
## About the PR 
Adds a donator figurine for SnugglyCactus of the character "Ribbert McFrogg".
## Why's this needed? 
I donated to the patreon, got excited, and sprited my character myself instead of waiting for someone else to do it.
## Testing 

image


snugglycactus:
This is what the minor sprite adjustment looks like, not that it affects whether this'll get merged or not. I just like the gills being a touch lower.

image

PR #25314, opened 2025-12-11
NT Headset Icons

JORJ949:
## About the PR 
Gives each NT role headset their own icon and object name (standard, NTSC and NTSO headsets keep the original icon)

Splits the base NT headset and NTSO headset so that you can give gimmick NT roles a generic NT headset with just NT and Command radio

## Why's this needed? 
To help identify which NT role is talking on the radio or who's headset you're about to steal, an NTSO on the radio (or headset to steal) is far more impactful than an NT Medic

## Testing 
image

## Changelog 

```changelog
(u)JORJ949
(*)NT Engineers and Medics now have unique headset icons with the colour of their relevant department.
```

### DIFF
--- a/code/obj/item/device/radios/headsets.dm
+++ b/code/obj/item/device/radios/headsets.dm
@@ -141,7 +141,6 @@
 /obj/item/device/radio/headset/command/nt/special_operative
 	name = "\improper NanoTrasen Operative headset"
 	icon_tooltip = "NanoTrasen Special Operative"
-	icon_override = "nt_sec"
 	secure_frequencies = list(
 		"n" = R_FREQ_NANOTRASEN,
 		"h" = R_FREQ_COMMAND,


TobleroneSwordfish:
Might seem like a nitpick, but I don't think NTSOs (and possibly NTSCs?) should get a sec coloured icon. They're explicitly not security, they're a centcom emergency squad that are often working at odds with station security (for example if they're here to arrest the HoS for shitposting on the centcom line)
I like the distinction that they are sometimes a third faction (good guys, bad guys, and corporate) rather than just more security with bigger guns.

JORJ949:
> Might seem like a nitpick, but I don't think NTSOs (and possibly NTSCs?) should get a sec coloured icon. They're explicitly not security, they're a centcom emergency squad that are often working at odds with station security (for example if they're here to arrest the HoS for shitposting on the centcom line) I like the distinction that they are sometimes a third faction (good guys, bad guys, and corporate) rather than just more security with bigger guns.

I do agree, but wasn't sure what colour to give NTSOs to differ them from NTSCs that wasn't the security one. I'd rather not give them the same because they're two very different roles in terms of power but I'd like to leave NTSC with the command colour. Perhaps they could get a different icon with the command colour instead like a gun or something

pgmzeta:
maybe give them a horizontal color band indicating department?

JORJ949:
Reverted NTSO special icon for now, will probably revisit NTSO/NTSC radio distiction in the future

PR #25324, opened 2025-12-14
changelings get hivemind member maptext

pgmzeta:
[feature]

## About the PR 
when speaking, members of a changeling hivemind residing in the host now show floating maptext visible to the to the host and other hivemind members
this maptext appears smaller than normal maptext, and is offset to the side and below.

when in abomination form, hivemind members in the changeling output text similarly

## Why's this needed? 
talk with the voices in your head

## Testing 

https://github.com/user-attachments/assets/a6b6da90-cdfe-454e-932c-0a58221295b5

## Changelog 

```changelog
(u)glowbold
(+)Changeling hivemind members now show floating maptext to the host
```

### DIFF
--- a/code/modules/speech/modules/speech/outputs/hivechat.dm
+++ b/code/modules/speech/modules/speech/outputs/hivechat.dm
@@ -61,9 +61,9 @@
 			src.role = " (MASTER)"
 
 	message.maptext_css_values["color"] = "#e2a059"
-	message.maptext_css_values["font-size"] = "5px"
+	message.maptext_css_values["font-size"] = "6px"
 	message.maptext_variables["maptext_x"] += prob(50) ? 28 : -28
-	message.maptext_variables["maptext_y"] -= rand(24, 32)
+	message.maptext_variables["maptext_y"] -= rand(12, 24)
 
 	if (isabomination(changeling_ability_holder.owner) && istype(message.speaker, /mob/dead/target_observer/hivemind_observer))
 		message.speaker.say(message.content, message_params = list("output_module_override" = SPEECH_OUTPUT_SPOKEN_HIVEMIND))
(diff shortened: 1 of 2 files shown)


Mister-Moriarty (changes requested):
This is entirely the wrong way to implement this; all you need to do is replace `message.flags |= SAYFLAG_NO_MAPTEXT` with the following in `.../speech/outputs/hivechat.dm`:
```
message.maptext_css_values["color"] = "#e2a059"
message.maptext_css_values["font-size"] = "5px"
message.maptext_variables["maptext_x"] += prob(50) ? 28 : -28
message.maptext_variables["maptext_y"] -= rand(24,32)
```

You'll need to do it for both `/datum/speech_module/output/bundled/hivemind` and `/datum/speech_module/output/global_hivemind`.


Only a single maptext image type should *ever* be used for speech, and in general sayflags should only be used to append metadata to a message (_see message modifiers and how singing is treated; it's not something that could be done with message vars alone_) or to modify the behaviour of the speech pipeline - never for basic styling.

That being said, I love the styling for hivechat maptext; it's very creative!

pgmzeta:
How can I add animations through the message maptext variables?

Mister-Moriarty:
> How can I add animations through the message maptext variables?

Currently not possible, but it should be easy enough to implement. I'll open a PR for it after I'm finished with namespaces.

code/modules/speech/modules/speech/outputs/hivechat.dm
```diff
@@ -92,13 +95,16 @@
...
-	message.flags |= SAYFLAG_NO_MAPTEXT
 
 	var/mind_ref = ""
 	if (ismob(message.speaker))
 		var/mob/mob_speaker = message.speaker
 		mind_ref = "\ref[mob_speaker.mind]"
 
+	message.maptext_css_values["font-size"] = "5px"
+	message.maptext_variables["maptext_x"] += prob(50) ? 28 : -28
+	message.maptext_variables["maptext_y"] -= rand(24, 32)
```
Mister-Moriarty: No hivechat colouring for global hivechat messages?

Mister-Moriarty:
> How can I add animations through the message maptext variables?

#25411

PR #25336, opened 2025-12-16
Namespaces & `ANIMATE` Namespace

Mister-Moriarty:
## About The PR:
Implements namespaces; these are datums with the explicit purpose of holding globally accessible variables and procs. Namespaces are implemented through the use of macros, so developers need never consider the underlying type structure of namespaces.

Creating a global or nested namespace:
```
CREATE_NAMESPACE(ANIMATE)			: creates a global `ANIMATE` namespace.
CREATE_NAMESPACE(ANIMATE, MOB)		: creates a nested `MOB` namespace inside of `ANIMATE`.
CREATE_NAMESPACE(ANIMATE, MOB, BEE)	: creates a nested `BEE` namespace inside of `ANIMATE.MOB`.
```

Adding variables of procs to a namespace:
```
ADD_TO_NAMESPACE(ANIMATE)(var/example_var = null)		: adds an `example_var` variable to `ANIMATE` and initialises it to `null`.
ADD_TO_NAMESPACE(ANIMATE, MOB)(var/example_var = null)	: adds an `example_var` variable to `ANIMATE.MOB` and initialises it to `null`.
ADD_TO_NAMESPACE(ANIMATE)(proc/example_proc())			: adds an `example_proc` proc to `ANIMATE`. Indented code following the macro will be treated as the body of the proc.
ADD_TO_NAMESPACE(ANIMATE, MOB)(proc/example_proc())		: adds an `example_proc` proc to `ANIMATE.MOB`. Indented code following the macro will be treated as the body of the proc.
```

As an illustrative example of how to implement namespaces, the animation library has been reimplemented as the `ANIMATE` namespace, with mob-specific animations being placed inside of the `MOB` nested namespace.

## Why Is This Needed?
Implementing a standard method of collecting variables and procs into globally accessible datums should help reduce global namespace pollution.

In addition to this, namespace datums are set up so to avoid the initialisation availability problem, wherein a global variable initialised through `... = new()` at the top of an arbitrary file will not be available during world initialisation due to BYOND not executing that particular `new()` call until the completion of the Genesis procedure.
Namespaces avoid this by being initialised by the `initialise_namespaces()` proc, called during `preMapLoad`.

## Testing:
Compilation is successful, and some animations were tested using buildmode and other admin commands, demonstrating that the `ANIMATE` namespace and `ANIMATE.MOB` nested namespace do not break existing proc calls. The proc ownership cache also indicates that the `ANIMATE` namespace holds the complete animation library.


Tarmunora (commented):
From an architectural perspective, is there a justification for namespaces being instantiated datums? Is it relevant that they be able to hold state?

With the example given here, it seems that it would be just as viable, as well as simpler, to put these procs under a simple uninstantiated datum path, and invoke them with `::`.

Mister-Moriarty:
> From an architectural perspective, is there a justification for namespaces being instantiated datums? Is it relevant that they be able to hold state?
> 
> With the example given here, it seems that it would be just as viable, as well as simpler, to put these procs under a simple uninstantiated datum path, and invoke them with `::`.

The `::` operator does not invoke procs, it only create a reference to them - you need to use `call()()` with that reference in order to invoke them.

This wasn't the reason for making namespaces instantiated datums though; accessing variables defined on an uninstantiated type is often problematic. In particular, accessing anything initialised via a proc (e.g. `list()` or `new()`) will return null, as BYOND calls a hidden `(init)` proc to initialise those variables.

PR #25343, opened 2025-12-18
Emitters can no longer damage the law rack

JORJ949:
## About the PR 
Emitter lasers no longer damage the AI law rack.
Adds a visible_message whenever the law rack is hit by a law safe heavy laser to make it clear it does no damage

## Why's this needed? 
A bit of a meta has formed in which 1 emagged borg will immediately emitter the law rack, roguing all borgs. Destroying the law rack this way is trivially easy as a borg because you can order them from cargo for dirt cheap or just mechscan one on singulo maps. Force borgs to find other, more difficult ways to destroy the law rack or to find a human who will rogue them for you.

## Testing 
image

## Changelog 

```changelog
(u)JORJ949
(*)Emitters can no longer damage the AI law rack.
```

### DIFF
--- a/code/modules/projectiles/laser.dm
+++ b/code/modules/projectiles/laser.dm
@@ -101,6 +101,7 @@ toxic - poisons
 
 /datum/projectile/laser/heavy/law_safe //subclass of heavy laser that can't damage the law rack - for AI turrets
 	name = "heavy laser"
+	law_rack_safe = TRUE
 
 /datum/projectile/laser/diffuse
 	sname = "diffuse laser"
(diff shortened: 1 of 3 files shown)


ThatFiveGuys:
May be worth blacklisting the AI turrets as well as I'm pretty sure you could do basically the same strat but with walling off a few turrets and dragging a roach or something behind you

pgmzeta:
understand the need, super dislike the hard type-check. is there a way we can make it a projectile property or something instead?

JORJ949:
> May be worth blacklisting the AI turrets as well as I'm pretty sure you could do basically the same strat but with walling off a few turrets and dragging a roach or something behind you

The AI turrets already are law safe, in fact, its the same projectile

JORJ949:
> understand the need, super dislike the hard type-check. is there a way we can make it a projectile property or something instead?

Yeah I agree, was gonna split that seperately but can roll that into this super easy

PR #24987, opened 2025-10-21
Refactor Genetics Powers Tracking PR

Egregorious:
## About the PR 

Refactors gene abilities to remove a lot of copy-pasted code, use defined returns and explicit srcs. Removes `cast_misfire` and replaces it with `cast_genetics`, which is always called by `cast` and just provides a misfire arg to use instead.

This fixes a bug where many abilities essentially casted twice whenever a misfire occurred. Miscast effects directly caused by the bug are intentionally rolled in to avoid behaviour changes:

- Eyebeam also regular casts
- Cryokinesis also regular casts
- Healing Touch double-heals
- Adrenaline Rush injects four times plus the base amount
- Matter Eater also swallows items
- Telepathy also sends a telepathic message as well as shouts

Added LAZYLISTACCESSASSOCDEEP macro for nested lazy lists, and converted LAZYLISTACCESSASSOC to only work for non-nested lists.

Atomise:
- [x] lazy list macro change
- [x] Cryokinesis
- [x] Matter Eater
- [x] Jumpy
- [x] Polymorphism
- [x] Trichochromatic Shift
- [x] Telepathy
- [x] Empathic Thought
- [x] Incendiary Mitochondria
- [x] Self Biomass Manipulation
- [ ] High-Pressure Intestines
- [ ] Very-High-Pressure Intestines
- [ ] Optic Energizer
- [ ] Adrenaline Rush
- [ ] Midas Touch
- [ ] Pickle Touch
- [ ] Healing Touch
- [ ] Dimension Shift
- [ ] Photokinesis
- [ ] Erebokinesis
- [ ] Fire Breath
- [ ] Brown Note
- [ ] Telekinetic Pull
- [ ] Telekinesis
- [ ] Cloak of Darkness
- [ ] Chameleon
- [ ] Mass Emesis
- [ ] Acidic Mass Emesis
- [ ] Vestigial Ballistics
- [ ] Trichromatic Shift
- [ ] Erebokinesis
- [ ] Ink Glands
- [ ] Aviornis Rostriformis
- [ ] Ophidentis Vipernox
- [ ] Scorpiocauda Vipernox 
- [ ] Fade
- [ ] Get all atomisations merged
- [ ] Remove old misfire implementation completely

## Why's this needed? 

Genetics abilities have a lot of copy-pasted code, which is seemingly encouraged by the way misfire mechanics are handled. This way genetic abilities have more control over how they want to implement miscasts.

fixes #24986

## Testing 

[...]

Issue #24986, opened 2025-10-21
Misfiring certain genetics abilities also calls their non-misfire acts


JORJ949:
### Description

Misfiring certain genetics abilities also calls their non-misfire acts

### Steps to reproduce

1. Acquire healing touch
2. Achieve low stability
3. Healing touch someone and have it misfire
4. Note that you heal, get hurt, then heal them.

### Expected Behavior

Just get hurt

### Additional Information & Screenshots

<img width="470" height="71" alt="Image" src="https://github.com/user-attachments/assets/9bffc5ab-f4df-4817-8659-212f56395619" />

PR #25379, opened 2025-12-22
Refactor matter eater

Egregorious:
## About the PR 

Refactors Matter Eater to use a different miscast implementation, explicit returns, srcs and "booleans".

Matter Eater still consumes the object on miscast, which was technically a bug but has been rolled in to avoid a change in behaviour.

This is a part of a larger refactor https://github.com/goonstation/goonstation/pull/24987.

## Why's this needed? 

Genetics abilities have a lot of copy-pasted code, which is seemingly encouraged by the way misfire mechanics are handled. This way genetic abilities have more control over how they want to implement miscasts.

## Testing 

Add bioeffect 'mattereater'. Test regular cast on random object; set stability to 0 to test miscast on random object.

### DIFF
--- a/code/modules/medical/genetics/bioEffects/powers.dm
+++ b/code/modules/medical/genetics/bioEffects/powers.dm
@@ -145,17 +145,17 @@ ABSTRACT_TYPE(/datum/bioEffect/power)
 	needs_hands = FALSE
 	var/using = FALSE
 
-	cast()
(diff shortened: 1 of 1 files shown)


Egregorious:
> Maybe create an issue for the misfire bug

#24986
Jorj opened an issue for it, which is linked in the larger draft. Can't really close it until all the individual powers are fixed though.

PR #25380, opened 2025-12-22
Alphabetized the MechComp vendor's product list.

Driftered:
## About the PR 
- Sorts the MechComp vending machine's product list into ABC order.
- The only exception is that I left the MechComp manual at the top of the list to ensure that users can grab it instantly if they're seeing it for the first time or need a refresher.
- Also added a comment beside each line in the code to identify what's what, which should help coders down the line in case any changes are made (as the variable doesn't always line up with the component name such as with `/obj/item/mechanics/cashmoney` = `Payment component`).

## Why's this needed? 
Currently, the MechComp vending machine's product list is a mess of components that are in order of when they were added, which makes it needlessly frustrating to find what you're after quickly.

## Testing 
Vended several items to make sure they were given to the player correctly and deducted from the machine's inventory stock.

MechComp ABC

## Changelog 
```changelog
(u)Driftered
(*)Tidied up the MechComp vendor's product list.
```

[A-ENGINEERING] [C-QoL]


Egregorious:
I would recommend sorting it programmatically more like [this](https://github.com/goonstation/goonstation/compare/master...Egregorious:goonstation:Vending-sort-example), since it avoids having to manually readjust for any changes, and it can be be reused easily for other vendors.

JORJ949:
On top of what egregorious said, I also don't think they should all be strictly organised alphabetically, I like that similar objects are grouped together, e.g. the Egun and Gun components. I think a better solution would be to make the mechcomp vendor specifically categorised like the botany seed and security weapons vendors are, but that's a much bigger task than sorting a list.

Driftered:
> I would recommend sorting it programmatically more like [this](https://github.com/goonstation/goonstation/compare/master...Egregorious:goonstation:Vending-sort-example), since it avoids having to manually readjust for any changes, and it can be be reused easily for other vendors.

That's a very clean bit of code, thank you!

> On top of what egregorious said, I also don't think they should all be strictly organised alphabetically, I like that similar objects are grouped together, e.g. the Egun and Gun components. I think a better solution would be to make the mechcomp vendor specifically categorised like the botany seed and security weapons vendors are, but that's a much bigger task than sorting a list.

I could give that a go. Let me take a look at the products and I'll see if I can come up with a nice sorting system for it.

PR #25402, opened 2025-12-26
Enable stickers to be placed onto borg-arm-assemblies and make sticker-throws guaranteed

M-Earthfire:
[Feature][Game Objects]

## About the PR 

This PR enables stickers to be placed onto robot-arm-assemblies, like pies.

Like pies, these stickers will be thrown onto the nearest target in the area.

To make this feature properly useable, the sticking chance on being thrown was set to 100%, up from 50%. 

## Why's this needed? 

This feature was inspired by the forum thread: https://forum.ss13.co/showthread.php?tid=24790

It also expands the use of robot arm assembies, which are kinda lacking right now.

I don't expect much balance implications from making sticker throws guaranteed. Spy stickers are very rarely used and contraband stickers are almost never used for criming. Making the throw guaranteed could potentially push these uses in a good way.

## Testing 

https://github.com/user-attachments/assets/d24892b6-f820-4761-83c9-4b2995a47ab3

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Stickers can be placed onto robot-arm assemblies
(+)Stickers, when thrown onto someone/something, will guarantee stick onto said object/person
```

### DIFF
--- a/code/obj/item/mob_parts/robot_parts.dm
+++ b/code/obj/item/mob_parts/robot_parts.dm
@@ -649,7 +649,7 @@ ABSTRACT_TYPE(/obj/item/parts/robot_parts/arm)
 		parent_assembly.applier_icon_prefix = "robot_arm"
 		if (!parent_assembly.target)
 			// trigger-robotarm-Assembly + pie -> trigger-robotarm-pie-assembly
-			parent_assembly.AddComponent(/datum/component/assembly, list(/obj/item/reagent_containers/food/snacks/pie), TYPE_PROC_REF(/obj/item/assembly, add_target_item), TRUE)
+			parent_assembly.AddComponent(/datum/component/assembly, list(/obj/item/reagent_containers/food/snacks/pie, /obj/item/sticker), TYPE_PROC_REF(/obj/item/assembly, add_target_item), TRUE)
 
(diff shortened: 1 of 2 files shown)


JORJ949:
Can you still electroplate stickers in radioactive material or was that removed? Worried that you can just throw cerenkite stickers at someone until they die.

M-Earthfire:
> Can you still electroplate stickers in radioactive material or was that removed? Worried that you can just throw cerenkite stickers at someone until they die.

That's already possible, just less reliable. But isn't that a nice feature that is unproblemstic for as long as non-antags don't use it for validhunting?

Theoretically, i can turn the chance to stick into a variable and drastically lower it when the sticker is electroplated. That way, people throwing radioactive stickers wouldn't be reliable while others are.

pgmzeta:
spy stickers autostick?

M-Earthfire:
> spy stickers autostick?

it's a behaviour of the parent of stickers, so every sticker does work.

Taking from the question, i think there are more reservations against removing the 50% stick chance on throw than the assembly part.

I think i will need to do some work on the throwing datum, but i take a look if i can apply the auto-sticker behaviour only to stickers thrown by assemblies. Since sticker throwing assemblies are single-use, they have to be guaranteed to be worth using, but i see it being a problem for throws by people.

I'm moving this to a draft until i made the corresponding changes.

Issue #25396, opened 2025-12-25
Cannot use clicking on things


goonstation-issuebot:
### Labels

[BUG][CRITICAL]

### Description

Left clicking DOESNT work, even after rebooting byond, happened mid shift after clicking an artifact

### Steps to reproduce

I don't know how to reproduce it, I touched a food processor artifact and since then I can't use left click but my arms still work

### Expected Behavior

To left click?

### Additional Information & Screenshots

I can't really show it on screenshots or left click... but I can get items grabbed or dragged to me and use them. Just not left click anything else

Reported by: kotlol
Client version: 516.1669
On server: Goonstation 4 Roleplay: Sylvester
Active test merges: #23946
Round log date: 2025-12-25-18-55
Reported on: 2025-12-25 19:59:47
Round ID: [219993](https://goonhub.com/rounds/219993)
Map: [COGMAP](https://goonhub.com/maps/cogmap)

Thunderclone:
yes happened to me too, same shift

Issue #25395, opened 2025-12-25
Artifact Grinder Causing Broken Item Interaction


Thunderclone:
### Labels

Bug 

### Description

The artifact type that is capable of grinding people up, causes a bug where you cannot manipulate any items in your inventory or the world.

### Steps to reproduce

1. Be a detective
2. Interact with a labeled artifact grinder
3. Survive almost dying by the grinder
4. No longer be able to use/manipulate any items apart from doors IF id is on

### Expected Behavior

I expected to get my arms cut off if I was going to be grinded to bits, that would make sense that my arms wouldn't work after. but my arms are fully intact, although its like there are no imputs on my mouse.

PR #25407, opened 2025-12-27
Makes Food processor sever limbs before deleting them to prevent ghost limb issues.

Kanthes:
[Science][Bug][Minor]

## About the PR 
As shown in issues like #25396 and #25395, a person rescued from a Food Processor would have their limbs become completely useless, despite still being attached. This severs the limb before deleting it, hopefully resolving this issue.

## Why's this needed? 
Ghost limbs bad.

## Testing 
Artifact still functions as normal during test in my own environment, unable to test the ghost limb issue.

## Changelog 

```changelog
(u)Kanthes
(+)Prevents the Food Processor from giving people ghost-limbs that couldn't interact with anything.
```

### DIFF
--- a/code/obj/artifacts/artifact_machines/food_processor.dm
+++ b/code/obj/artifacts/artifact_machines/food_processor.dm
@@ -103,15 +103,8 @@
 				if (!convertable_limbs.len) //avoid runtiming once all limbs are converted
 					continue
 				var/obj/item/parts/limb_to_replace = pick(convertable_limbs)
-				switch(limb_to_replace.slot)
-					if ("l_arm")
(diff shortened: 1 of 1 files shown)


cropseyy:
Would expanding this to cover borgers fix the identical issues there too, or was that fixed and I just forgot?

Kanthes:
The borgifier shouldnt share this issue as it replaces existing limbs with
light robotic ones.

ZeWaka:
Why not make it so that if a limb is attached during the deletion process of a limb, it severs itself from its parent? That way, people don't run into this issue in the future.

PR #25409, opened 2025-12-27
Makes assemblies be able to be wrenched into large mechcomp cabinets

M-Earthfire:
[Game Objects][Feature]

## About the PR 

Allows assemblies that are secured to be wrenched into large mechcomp cabinets.

Wrenched in assemblies cannot be unsecured before detaching the assembly from the cabinet.

## Why's this needed? 

With the amount of interactions assemblies offer, having a more secured interface between mechcomp and assemblies is worthwhile. these are offered by cabinets. With assemblies expanding in usage, this will give cabinets more use cases, e.g. creating hand tele-portals while running in a direction or just be annoying with bike horns)

The implementation of handheld assemblies would require work on the storage datum, because for some reason there is no way to discard too large items if they are in the `can_hold`-list and  I don't want 7 singletank bombs to fit into a single normal sized slot. that will be done in a seperate PR.

## Testing 

https://github.com/user-attachments/assets/09fbc052-9c5d-4d18-ad99-9f298011ed31

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Assemblies can now be wrenched into large mechcomp cabinets.
```

### DIFF
--- a/code/obj/item/assembly/misc_assemblies.dm
+++ b/code/obj/item/assembly/misc_assemblies.dm
@@ -77,6 +77,7 @@ Contains:
 	RegisterSignal(src, COMSIG_CELL_TRY_SWAP, PROC_REF(try_cell_swap))
 	RegisterSignal(src, COMSIG_CELL_SWAP, PROC_REF(do_cell_swap))
 	RegisterSignal(src, COMSIG_MOB_GEIGER_TICK, PROC_REF(on_geiger_tick))
+	RegisterSignal(src, COMSIG_ITEM_STORED, PROC_REF(on_storage_addition))
 	..()
 
 /obj/item/assembly/proc/set_up_new(var/mob/user, var/obj/item/new_trigger, var/obj/item/new_applier, var/obj/item/new_target)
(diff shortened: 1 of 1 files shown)


pgmzeta:
can you use this to deliver a bomb assembly with a walking component?

M-Earthfire:
> can you use this to deliver a bomb assembly with a walking component?

Yes. 

Except that i have never seen a movement cabinet managing to do any sufficient pathfinding to a target. Doors are especially bad obstacles for them.. I will applaude the nerd who manages to pull that off, though.

M-Earthfire:
Right now, the cabinet gets a neat overlay if it has any assemblies, due to concerns about visibiity.

https://github.com/user-attachments/assets/2820bafa-d66a-417d-9d65-88d709fcf502

I'm toying around with throwing the assembly into vis_content of the cabinet, but what i currently tried looked... lackluster

M-Earthfire:
i'm moving this to a draft. 

I decided, to properly add a visual of the assembly onto the cabinet, that i need to redo how the assemblies generate their icon. I want to atomize that change, so i need to make a different PR as a requirement to this one.

PR #25325, opened 2025-12-14
Redesigns Oshan's evidence room.

smallsandman:
## About the PR 
Redesigns the oshan labs evidence room and interrogations room. moves beepsky from his home and converts it to evidence room.
## Why's this needed? 
 I wasn't quite happy with my evidence room on oshan, it worked yes but it didn't feel very unique to me. so i built it in a new spot and reverted the room, putting the flashers in place of the safes. 
 
## Testing 
image

All chutes have been tested.

## Changelog 

```changelog
(u)Smallsandman
(*) Moves Oshan Laboratories Evidence room.
```

PR #25410, opened 2025-12-27
Oshan interrogations revamp

smallsandman:
## About the PR 
Redesigns the interrogations room.
## Why's this needed? 
 The interrogation room felt dated and hard to use for me, as well as being very difficult for antags to escape from as you needed to get through two sec access doors at minimum. this redesign should make it a little easier for escapes, as well as make it easier for sec to use. 
 
 (This was originally connected to my oshan labs evidence room PR #25325, but i figured it'd be much better to split the two.)
## Testing 
image
All chutes function as expected.

## Changelog 

```changelog
(u)Smallsandman
(*)Redesigned oshan labs interrogation room.
```


JORJ949 (changes requested):
Not a fan of abandoning the unique room shape for a square. I'd also argue its harder to break out of because now you HAVE to run through your interrogators if you have the access (you're not getting out of either if you don't have access) to flee rather than them having to chase you. I think your goals could be achieved by just removing the table on the right hand side of the room and possibly adding a door to the courtroom in the prisoner's section.

smallsandman:
> New maintenance is very big and very empty, add some trash, shrink it, or both. Though I do like it not being a 1 tile wide corridor.

Shrunken

PR #25412, opened 2025-12-27
Changes armory 9mm from plastic to lethal

xxfishshadowxx:
## About the PR 
The armory ammotech vendor will now stock lethal 9mm (30 brute per bullet, 8 bullets as opposed to plastic's 18 bullets) in-place of the nonlethal alternative

## Why's this needed? 
Reasoning:

-plastic 9mm now is accessible from general fabricators

-armory lacks solid lethal weaponry beyond the single use buckshot, and the very low dps easy-to-renew phasers. This is a middle ground between the two :P

## Testing 
clownbullet
clowndead

## Changelog 

```changelog
(u)emeraldtest
(+)Armory ammotech now stocked with lethal 9mm inplace of plastic 9mm!
```

### DIFF
--- a/code/modules/vending/vending.dm
+++ b/code/modules/vending/vending.dm
@@ -1558,8 +1558,8 @@ TYPEINFO(/obj/machinery/vending/medical)
 		product_list += new/datum/data/vending_product(/obj/item/chem_grenade/flashbang, 7)
 
 		if (!restocked) //technically, this deletes any ammo left over on restock. oh well.
-			product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/a12/weak, 1, hidden=1) // this may be a bad idea, but it's only one box //Maybe don't put the delimbing version in here
 			product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/bullet_9mm/armory,3)
+			product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/a12/weak, 1, hidden=1) // this may be a bad idea, but it's only one box //Maybe don't put the delimbing version in here
 		else
 			product_list += new/datum/data/vending_product(/obj/item/ammo/bullets/bullet_9mm/armory,1) // only 1 mag when restocked
 


DisturbHerb:
remind me if we can't buy clock19s from cargo. i recall the frangibles being the solution for secoffs who would buy that thing ASAP

JORJ949:
Not a huge fan of increasing the armory's lethal capabilities. It is intentionally stocked mostly with less-than-lethals as riot gear, also kinda devalues the actual lethals given to the crew such as the Detective's revolver, which also does 35 damage but has less than half the bullets per ammo. The phasers have 20 damage per with 8 shots each, smg has 8 damage per with 20 shots, and both of these require you to go recharge after you run out rather than just keep loading more ammo in.

M-Earthfire:
I feel security should get a bit more lethal options simply because it gives us the opportunity to add more tools to combat stun weaponry (which is always hampered by sec being mostly forced into non-lethals). In that regard i like the addition.

xxfishshadowxx:
> remind me if we can't buy clock19s from cargo. i recall the frangibles being the solution for secoffs who would buy that thing ASAP

I agree that this would be problematic, but since the ammo would only be accessible from the armory, I think it would be hard to abuse in a logistic manner. The bullets themselves also have low stopping power, any officer who opts to use lethals is opening themself up to about 20 seconds of retaliation, whereas stun alternatives end fights usually in less than a quarter of that

virvatuli-gh (commented):
The frangible ammunition (it didn't used to be called that but I'm blanking on what it was, I just renamed it) was only ever in the armoury because the NTSO's past gun was the Clock, before we made the Signifer. In fact, you used to get it from regular SecVends if I remember right. To be honest, the ammunition even being in the armoury at all may be an oversight in the first place, ~~Secoffs were never intended or encouraged to go buy the nerfed Clock from Cargo~~ (edit: actually maybe I'm wrong on that, considering that goofy badge interaction and the fact that it's in a Security crate, weird if I'm being honest)

I don't mind if the Armoury gets more lethal stuff, but I think this is the wrong way to go about it. The 9mm you're changing it to is meant for the Syndicate, it's specifically the magazine for the Nuke Ops' Branwen sidearm (and it doesn't match the Clock's intended ammo capacity). The frangible is specific to NT, because it's safer to use on a space station, hence the name, and it's deliberately less-lethal.

JORJ949:
>  I think this is the wrong way to go about it. The 9mm you're changing it to is meant for the Syndicate.

If armory lethality is increased, it should be with energy weapons (e.g. the newish phaser SMG rack that has a chance to replace the normal phaser rack) and not kinetics. Kinetics are thematically a syndie thing and not for station crew (the detective is an exception as they're a "loose cannon").

xxfishshadowxx:
> The frangible ammunition (it didn't used to be called that but I'm blanking on what it was, I just renamed it) was only ever in the armoury because the NTSO's past gun was the Clock, before we made the Signifer

I'm a (relatively) newer player so I never saw the pre-changes NTSO/NTSC, but the NTSC is the one with the signifier who currently plays the role of the security '2nd in command'/'extra lethals' officer. The NTSO still has a clock with like 5 magazines of the same 9mm used in the PR

> I don't mind if the Armoury gets more lethal stuff, but I think this is the wrong way to go about it. The 9mm you're changing it to is meant for the Syndicate, it's specifically the magazine for the Nuke Ops' Branwen sidearm (and it doesn't match the Clock's intended ammo capacity). The frangible is specific to NT, because it's safer to use on a space station, hence the name, and it's deliberately less-lethal.

The 9mm is just the generic universal 9mm, but maybe it was not always that way, I do not know. Currently, this 9mm is used by:

Salvagers
any syndicate with a Branwen
NTSO
NT Field Commander, and
The 9mm pod gun

A subvariant with the same spriting but reduced ammo count is additionally used for the gangs Lo-point pistol.

Jorj's concerns over the ridiculously high mag capacity are pretty fair in my opinion. I do agree that 15x35 total damage is a little much. I could duplicate and rename the gangs 9mm subvariant to something more appropriate. How many bullets should the magazine get? 8? 10?

> If armory lethality is increased, it should be with energy weapons (e.g. the newish phaser SMG rack that has a chance to replace the normal phaser rack) and not kinetics. Kinetics are thematically a syndie thing and not for station crew (the detective is an exception as they're a "loose cannon").

The armory has a variety of mixed options though to begin with. NT and the Syndicate may have a primary 'genre' of firearm, but both still have guns of each type. I think that having the potential for kinetic pistol crossfire outside of gangs rounds and admin-run war events could be pretty fun for both players and observers. The current frangible pistol rounds are closer to nerf bullets than they are to absolutely anything else security has, being just a little better as a stun than the brobocop dart gun. They are only used for gimmicks right now, which is a little nonsensical for an item stocked in the armory.

Thank you guys for your feedback on this, I get where you are coming from as securities tools majorly dictate the flow of a given round

virvatuli-gh:
> I'm a (relatively) newer player so I never saw the pre-changes NTSO/NTSC, but the NTSC is the one with the signifier who currently plays the role of the security '2nd in command'/'extra lethals' officer. The NTSO still has a clock with like 5 magazines of the same 9mm used in the PR

NTSC didn't exist before, it was just NTSO and they had the Clock 188. That's what the Clock was designed for originally, but then we made the Signifer for them and the 188 got shelved and the 180 got thrown in that crate for some reason as a nerfed version of it. Later on the NTSO was changed into the NTSC for whatever reason and then a different NTSO was made. The machinations of the admin hivemind are inscrutable.

> The 9mm is just the generic universal 9mm, but maybe it was not always that way, I do not know. Currently, this 9mm is used by:

All of these are antagonists or admin-spawned only things, player facing items that can be in-use conceivably every round require more thought, re-use of something meant for a specific role should not be encouraged! The Lo-Point version is the correct way to do this, a magazine specifically made for the gun with the correct capacity (see also the 9mm SMG magazine).

xxfishshadowxx:
Alright, ill do so soon

xxfishshadowxx:
I made an 'armory' subtype of 9mm with 8 bullets per mag, for reference, the max damage on a mag if all shots hit now is 240, where a nonlethal mag is 108
the old max damage with 15 rounds was 450
a single phaser with 1 cell is 160

JORJ949:
My opinion hasn't changed. I still don't think the armory needs more lethality and don't think they should get kinetics.

PR #25414, opened 2025-12-28
MechComp Device Copying

QuiteLiterallyAnything:
[QOL] [ENGINEERING]

## About the PR 
This PR makes it possible to copy the settings of most MechComp devices to other devices of the same type by simply ~~hitting them together~~ using the multitool connection settings menu. ~~The settings of the component in hand are copied to the component hit.~~

Code-wise this is accomplished through a new `copy_identical_mechcomp()` proc for subtypes of the `/obj/item/mechanics` type. This proc is set individually for each MechComp device (some require very specific adjustments so it would be rather impractical to cleanly automate it all with just a few lines of code). Objects which can't (or otherwise shouldn't be) copied are marked by setting the new `mechanically_copyable` boolean var to `FALSE`.

Devices with a "Send Signal" (which is linked to their handler's `defaultSignal` var) have this copied by accessing the handler directly. I'm unsure if this is optimal or the cleanest way to do things, but it works and I see no alternatives.

## Why's this needed? 
Some MechComp contraptions require multiple of the same component with the exact same (or very similar) settings. This PR makes those devices easier to build, and reduces the amount of UI tedium required to work with MechComp in general.
## Testing 
On a debug build of the codebase I constructed one of the MechComp contraptions I have memorized (which includes a fair number of repeated components) while copying as many components as possible. It worked perfectly with the exception of trip lasers (which didn't copy on my first test due to the previously mentioned "Send Signal"/`defaultSignal` thing). 

After making commit 1577c3e26173ed349d7b73f78c3ff182a055ed88 I rebooted, tested trip lasers again, tried copying and then using a few other components, and found that everything worked as it should.

## Changelog 

```changelog
(u)QuiteLiterallyAnything
(*)MechComp devices can now have their settings copied to others of the same type via the menu they display when hit with a multitool.
```

### DIFF
--- a/code/modules/mechanics/MechanicMadness.dm
+++ b/code/modules/mechanics/MechanicMadness.dm
@@ -428,8 +428,7 @@ TYPEINFO(/obj/item/mechanics)
 	get_help_message()
 		. = ..()
 		if(src.mechanically_copyable)
-			. += "You can copy this component's settings to another of the same type by hitting \
-			the component you want changed with the component you want copied."
+			. += "You can copy this component's settings to another of the same type with a <b>multitool</b>!"
 
 	disposing()
 		processing_mechanics.Remove(src)
@@ -458,20 +457,17 @@ TYPEINFO(/obj/item/mechanics)
 			RegisterSignal(user, COMSIG_PARENT_PRE_DISPOSING, PROC_REF(clear_owner))
 			owner = user
 
-		/// Individually defined for each subtype. Called when the component is attacked by another
-		/// component of the same type. Copies the configuration/settings of the attacking component.
-		copy_identical_mechcomp(obj/item/mechanics/copied_mechcomp, mob/attacker)
-			if(!copied_mechcomp.mechanically_copyable)
-				boutput(attacker, SPAN_ALERT("The [src] in hand can't be copied!"))
-				return FALSE
-
+		/// Individually defined for each subtype. Called when this object is hit by a multitool
+		/// bearing a copying component. Copies the configuration/settings of `copied_mechcomp`.
+		copy_identical_mechcomp(obj/item/mechanics/copied_mechcomp, mob/attacker, atom/copying_medium)
 			var/datum/component/mechanics_holder/holder = src.GetComponent(/datum/component/mechanics_holder)
 			if("Set Send-Signal" in holder.configs) // <-- I assume copying the define for this string from 'mechComp_signals.dm' would be worse.
 				var/datum/component/mechanics_holder/copied_holder = copied_mechcomp.GetComponent(/datum/component/mechanics_holder)
 				holder.defaultSignal = copied_holder.defaultSignal
 
 			src.tooltip_rebuild = TRUE
-			boutput(attacker, SPAN_ALERT("You've copied the settings of the in-hand [src] to the [copied_mechcomp]."))
+			logTheThing(LOG_STATION, attacker, "copies the settings of [log_object(copied_mechcomp)] [log_loc(copied_mechcomp)] to a [log_object(src)] [log_loc(src)].")
+			boutput(attacker, SPAN_NOTICE("You've copied settings from [copying_medium] to the [src]."))
 
 	process()
 		if(level == OVERFLOOR || under_floor)
@@ -596,9 +592,6 @@ TYPEINFO(/obj/item/mechanics)
 			else
 				src.wrench_action(user)
 			return TRUE
-		if(W.type == src.type)
-			copy_identical_mechcomp(W, user)
-			return TRUE // Prevents chat spam with "X HAS BEEN HIT BY Y WITH Z!" messages.
 		return ..()
 
 	pixelaction(atom/target, params, mob/user)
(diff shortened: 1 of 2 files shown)


code/modules/mechanics/MechanicMadness.dm
```diff
@@ -336,6 +337,7 @@ TYPEINFO(/obj/item/storage/mechanics/housing_handheld)
 	anchored= ANCHORED
 	level=1
 	w_class = W_CLASS_BULKY
+	mechanically_copyable = FALSE // Anticipatory adjustment for a possible unit test.
```
pgmzeta, line 340: ?

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -573,6 +596,9 @@ TYPEINFO(/obj/item/mechanics)
 			else
 				src.wrench_action(user)
 			return TRUE
+		if(W.type == src.type)
+			copy_identical_mechcomp(W, user)
+			return TRUE // Prevents chat spam with "X HAS BEEN HIT BY Y WITH Z!" messages.
```
pgmzeta: probably should have a confirmation dialog

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -985,8 +1023,9 @@ TYPEINFO(/obj/item/mechanics)
 	var/range = 5
 	var/list/beamobjs = new/list(5)//just to avoid someone doing something dumb and making it impossible for us to clear out the beams
 	var/active = FALSE
+	// This probably doesn't actually do anything. Devices with `COMSIG_MECHCOMP_ALLOW_MANUAL_SIGNAL`
+	// have their 'defaultSignal' (Send Signal) handled by the holder.
 	var/sendstr = "1"
```
pgmzeta: remove if unused

code/modules/mechanics/MechanicMadness.dm
QuiteLiterallyAnything: This button type should never appear outside of a container (inside of which it will never be copyable). Setting `mechanically_copyable` to `FALSE` should then be unnecessary, but I was thinking about the possibility of a future unit test which checks to see if a `mechanically_copyable` object actually has any changes to its `copy_identical_mechcomp()` proc. 

I have no prior experience dabbling in unit tests, and I don't even know if it would be possible, practical, or worthwhile to test something like that. In retrospect I suppose it is quite the assumption.

code/modules/mechanics/MechanicMadness.dm
QuiteLiterallyAnything: Done

code/modules/mechanics/MechanicMadness.dm
QuiteLiterallyAnything: Though that would be a very simple bit of UI interaction, I worry it would run somewhat counter to the point of skipping the UI interaction of copying in the first place. Hitting a MechComp component with another of the same type is a very specific input as is. Anyone doing a lot of copying might lose more time confirming than they would accidentally copying and having to reenter settings.

pgmzeta: people can also lose work copying over a complex setup

QuiteLiterallyAnything: That is possible. Might I instead propose that I add either a check for an in-hand multitool or a three second copying progress bar?

pgmzeta: What about adding this to the Config menu you get when using a multitool on a mechcomp component? After selecting "Copy settings" you could just tap each one with the multitool, and clear it by using in hand like the linking setup. Could also add to click-drag menu?

QuiteLiterallyAnything: That sounds good, I'll try to implement such a system when I have time available.

QuiteLiterallyAnything: I have successfully made the copying work via multitool. It did (ironically[^1]) involve copying some code from `/datum/component/mechanics_connector`, but it's fairly functional.

[^1]: The code for copying is itself somewhat copied.

code/modules/mechanics/MechanicMadness.dm
```diff
@@ -412,13 +414,21 @@ TYPEINFO(/obj/item/mechanics)
...
 	New()
 		particle_list = new/list()
 		AddComponent(/datum/component/mechanics_holder)
 		processing_mechanics |= src
 		return ..()
 
+	get_help_message()
+		. = ..()
+		if(src.mechanically_copyable)
+			. += "You can copy this component's settings to another of the same type with a <b>multitool</b>!"
```
pgmzeta, line 431:
```suggestion
		if(src.mechanically_copyable)
			. += "Copy this component's settings with a <b>multitool</b>."
```

QuiteLiterallyAnything: Some components of different types share settings (like "send-signal") but those can't be copied across types. Would "to another of the same type" not be necessary to prevent confusion then?

The exclamation mark isn't necessary at all (I must've been subconsciously overjoyed about the idea of copying MechComp settings when typing that), but why switch to imperative mood?

PR #8541, opened 2022-05-10
Mineral magnet tgui

codeJester27:
## About the PR 
Gives the mineral magnet a new tgui menu, accessed by the mineral magnet controls computer or the pod magnet link thingy.
imageimage
imageimage
image

## Why's this needed? 
tgui. Now it auto-updates, you can geo scan at any time instead of having to wait for the magnet to finish pulling, and there are probably less href exploits that nobody actually used anyways.

## Changelog

```changelog
(u)CodeJester
(*)Mineral magnets, how do they work? Maybe you can find out with its new UI menu.
```

PR #25415, opened 2025-12-28
Fix Mining Magnet UI Checkboxes

jaredtalbot:
[UI][BUG]

## About the PR 

Makes the checkboxes in the mining magnet UI appear over the "Magnet Active"/"On Cooldown" dimmer, allowing the user to actually toggle them even if the magnet cannot be activated in its current state.

## Why's this needed? 

This fixes a regression caused by `Dimmer` components being given a higher z-index at some point; being able to toggle cooldown override and automatic mode even while the magnet was active or on cooldown was a feature present in the pre-TGUI mining magnet and the original TGUI mining magnet (#8541).

## Testing 

dreamseeker_AGiznIUXf8


ZeWaka:
I don't think this is correct. Just make them not under the dimmer?

PR #25417, opened 2025-12-28
Make multitool-assemblies be improved by adding power cells to them.

M-Earthfire:
[Feature][Game Objects]

## About the PR 

This PR makes it so that multitool-assemblies can be improved by adding power cells onto them.

Depending on the power of the cell used, the area and the effect of the electric effect is increased.

For every 5k of charge on the cell, the power of the elecflash is increased by 1, up to 6 at 8k Charge. at 2k and 4k energy, the range increase by 1. 

Erebite cells cause an explosion without triggering the assemblies effect.

## Why's this needed? 

Adding cells to assemblies should help increasing their power. A bigger cell on the multitool does this mainly be increasing the shock and effective range on the assembly. a High powered cell causes a stamina drain akin to getting defibbed.

Their main use will likely be lighting fires. Due to the increased range you can place them further away from welding fuel tanks or other igniteable sources. And the larger shock does cause a better heatup than the small shock from the unmodified multitool.

## Testing 

https://github.com/user-attachments/assets/ec668e47-b751-44ee-8a73-7c75cbf52326

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Multitool assemblies can be modified by adding a large power cell onto them. This drains the power cell to increase the area and the effect of the flash produced.
```

### DIFF
--- a/code/obj/item/device/multitool.dm
+++ b/code/obj/item/device/multitool.dm
@@ -19,8 +19,8 @@ TYPEINFO(/obj/item/device/multitool)
 	m_amt = 50
 	g_amt = 20
 	custom_suicide = TRUE
-	var/spark_power = 5000 //! The amount of power needed to cause an upgraded spark in a cell-assembly
-	var/max_spark_power_usage = 20000 //! The amount of power consumed at maximum power.
+	var/spark_power = 2000 //! The amount of power needed to cause an upgraded spark in a cell-assembly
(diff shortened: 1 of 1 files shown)


M-Earthfire:
I decreased the power drain from 5-20k to 2-8k. That way, you can use the multitool assembly 2 times with a standard cell and 3 times with a hypercell.

PR #25418, opened 2025-12-28
Changes check_wclass on datum/storage to use defines and add a define to enforce the wclass on everything in the can_hold-list

M-Earthfire:
[Internal][Refactor]

## About the PR 
This Pr changes so  `check_wclass` on `datum/storage` uses defines. The following defines were added:

 When this variable is set to true, the storage will check w_class on items even if they are on its `can_hold` list

- `STORAGE_CHECK_W_CLASS_IGNORE`: This storage ignores the wclass on anything in `list/can_hold` This mimics the old behaviour as if check_wclass is set to FALSE
- `STORAGE_CHECK_W_CLASS_INCLUDE`: This storage can carry anything with a lower wclass than `max_wclass` beyond what's defined in `list/can_hold` . This mimics the old behaviour as if check_wclass is set to TRUE
- `STORAGE_CHECK_W_CLASS_EXCLUDE`: This storage filters out anything with a higher wclass than `max_wclass` defined in `list/can_hold`

## Why's this needed? 

During working on #25409. i found that i was unable to make handheld mechcomp cabinets take the wclass of assemblies into account. Due to `obj/item/assembly` taking the form of different wclasses, depending on what items are build into it, this causes design issues, e.g. 7 pipebombs being able to fit in a medium sized object.

This PR fixes this issue and thus adds more flexibility into the framework.

## Testing 

Because this affects all storages, i tried multiple different storages and looked if they behave like they should. Examples include mechcomp cabinets, artifact bag of holdings, belts and backpacks. All worked without issues, so i hope this PR doesn't affect any storages.

### DIFF
--- a/_std/defines/storage.dm
+++ b/_std/defines/storage.dm
@@ -24,9 +24,13 @@
 
 // check_wclass behaviour defines
 
-/// Standard for any storage with list/can_hold defined. Only what is in that list can be held in the storage, independing of w_class
+/// Default behavior; changes based on the `list/can_hold` variable.
+///
+///If `can_hold` is defined, store items if in `can_hold`, regardless of `max_wclass`. 
+///
+///If `can_hold` is empty/undefined, store items equal or under `max_wclass`.
 #define STORAGE_CHECK_W_CLASS_IGNORE 0
-/// Above list/can_hold, everything which is within the wclass defined in max_wclass can fit in the storage as well
+/// Store items in `can_hold` OR items equal or under `max_wclass`.
 #define STORAGE_CHECK_W_CLASS_INCLUDE 1
-/// Causes the storage not not allow things in list/can_hold which have a wclass higher than what is defined in max_wclass
+/// Store items that are in the `can_allow` list AND equal or under `max_wclass`.
 #define STORAGE_CHECK_W_CLASS_EXCLUDE 2


pgmzeta:
feels like "check_wclass" and "enforce_wclass" should be two different states of a wclass_handling sort-of variable

M-Earthfire:
I like that idea since it would prevent the creation proc to blew up further with variables. I'm implementing that.

_std/defines/storage.dm
```diff
@@ -21,3 +21,12 @@
...
 #define STORAGE_NO_HUD_RANDOM 2
+
+// check_wclass behaviour defines
+
+/// Standard for any storage with list/can_hold defined. Only what is in that list can be held in the storage, independing of w_class
+#define STORAGE_CHECK_W_CLASS_IGNORE 0
+/// Above list/can_hold, everything which is within the wclass defined in max_wclass can fit in the storage as well
+#define STORAGE_CHECK_W_CLASS_INCLUDE 1
+/// Causes the storage not not allow things in list/can_hold which have a wclass higher than what is defined in max_wclass
+#define STORAGE_CHECK_W_CLASS_EXCLUDE 2
```
pgmzeta:
```suggestion
/// Default behavior; changes based on the `list/can_hold` variable.
///
///If `can_hold` is defined, store items if in `can_hold`, regardless of `max_wclass`. 
///
///If `can_hold` is empty/undefined, store items equal or under `max_wclass`.
#define STORAGE_CHECK_W_CLASS_IGNORE 0
/// Store items in `can_hold` OR items equal or under `max_wclass`.
#define STORAGE_CHECK_W_CLASS_INCLUDE 1
/// Store items that are in the `can_allow` list AND equal or under `max_wclass`.
#define STORAGE_CHECK_W_CLASS_EXCLUDE 2
```

theres technically a fourth state in this behavior matrix but its covered by whether the can_hold list exists 🙈

M-Earthfire: True, going more into detail there makes more sense... and yeah, that behaviour depending on if can_hold is defined is quite odd...

PR #25431, opened 2025-12-30
Storage Object Material Preservation

QuiteLiterallyAnything:
[Game Objects] [Bug]

## About the PR 
This PR is intended as a simple fix for closets deconstructing to one sheet despite two sheets being required to make them. Currently, closets have their `material_amt` var set to 0.2, but no underlying deconstruction logic on `/obj/storage` works off that. I've attempted to remedy this by adding a small check and adjusting a few storage item vars accordingly.
## Why's this needed? 
Conservation of energy is a given for any simulation game. Other than that, it's simply inconvenient to not be able to deconstruct an empty closet and move it somewhere else.
## Testing 
On a debug build of the codebase, I spawned in and deconstructed several crates, closets, and carts to ensure that they gave the appropriate amount of material.

## Changelog 

```changelog
(u)QuiteLiterallyAnything
(+)Adjusted the amount of material given by closets, crates, and the like on deconstruction.
```

### DIFF
--- a/code/obj/storage/large_storage_parent.dm
+++ b/code/obj/storage/large_storage_parent.dm
@@ -8,7 +8,7 @@
 
 #define RELAYMOVE_DELAY 1 SECOND
 /// The percentage of a storage object's health below which it will only return one sheet on deconstruction.
-#define STORAGE_UNSALVAGEABLE_THRESHOLD 0.75
+#define STORAGE_UNSALVAGEABLE_THRESHOLD 0
 
 ABSTRACT_TYPE(/obj/storage)
 ADMIN_INTERACT_PROCS(/obj/storage, proc/open, proc/close, proc/break_open)


JORJ949 (approved):

pgmzeta:
This makes using lockers as weapon blocking free, don't like that.

QuiteLiterallyAnything:
> This makes using lockers as weapon blocking free, don't like that.

Fair point, I've added a check to account for that as of the latest commit (19990d5eb64aeb6d5d41dc7b50cd9fae040f3b3f). Is that better?

pgmzeta:
> Fair point, I've added a check to account for that as of the latest commit ([19990d5](https://github.com/goonstation/goonstation/commit/19990d5eb64aeb6d5d41dc7b50cd9fae040f3b3f)). Is that better?

use 0 as the threshold, that's what causes the locker broken state

QuiteLiterallyAnything:
> use 0 as the threshold, that's what causes the locker broken state

Done

PR #25437, opened 2025-12-30
implements electropack assemblies

M-Earthfire:
[Game Objects][Feature]

## About the PR 
Makes it so electropacks can be build into assemblies
signaler_electropack

these are contraband 4 assemblies that, when used as-is, create a strengh 4 electroflash at this position (somewhat stronger than a multitool).

If they are combined with a cell, however, their power is increased. When used, they drain the battery by up to 15kW and cause an arcflash with power equal to the `cell power - 2kW`. This arcflash is targeted at a random mob in a range of 4 tiles or a random turf in that range, when no mob is found.

The assembly, when used, goes on a 4 second cooldown.

## Why's this needed? 

IT's almost not possible to cause directed arcflashes. This assembly gives a way to do so. Electropacks are bulky items, so it is quite hard to use this outside of elaborate traps.

The way to scale this assembly is by using better cells. High capacity cells yield 2 charges, but with e.g. artcells, 4-5 arcflashes are possible. This gives artcells a desireable usecase beyond powering borgs.

Besides nefarious purposes by antags, this has a different use case. Arcflashes on grilles can send power into the net via grilles, so i am sure engineers will find ways to power the station by the use of self-recharging batteries. The sus-generator blueprints will come to use once again!

## Testing 

In this video, the assemblies had a range of 6, which was a bit too much in my opinion.

https://github.com/user-attachments/assets/5f969f3c-2c7f-4822-adb2-25c8992732cc

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Electropacks can be used as appliers in assemblies. When triggered, they cause a small shock on the ground
(+)Electropack assemblies have a contraband value of 4
(+)Electropack-assemblies, when combined with cells, drain the battery to cause an arcflash. The power depends on the cell's charge
```


M-Earthfire:
Admittably, i think draining the whole battery for a single arcflash is a bit much. But i am not sure how the numbers would look best at it. It would likely require some "magic bullshittery" multiplication of the power, so e.g. a cell would be drained by 10k and produce a 20k wattage arcflash.

Even artcells don't reach a potency such as pipebombs, and these count as contraband 4 items as well. Maybe artcell capacity should evaporize non-insulated people?

M-Earthfire:
After some feedback from the discord, i found out that arc flashes don't actually scale past 12k of voltage. So the formular got changed.

electropacks now take up to 15kW of power for a max-charge elecflash. they waste 2kW, meaning you require at least 14kW on your cell for a full shock.

This change means that standard cells enable only a single flash, while high capacity cells enable to do 2 arc flashes. Artcells require more. Since they work on a 4-second cooldown, this should be fairly balanced... and more fair than gibbing via pipebombs.

M-Earthfire:
New test video, showcasing the cooldown and the effect on a high capacity cell:


https://github.com/user-attachments/assets/7b3732e4-b4b4-4f4d-bdd1-032e21aca4bb

code/obj/item/device/radios/electropack.dm
```diff
@@ -16,20 +16,85 @@ TYPEINFO(/obj/item/device/radio/electropack)
...
+	. = ..()
+	UnregisterSignal(src, COMSIG_ITEM_ASSEMBLY_APPLY)
+	UnregisterSignal(src, COMSIG_ITEM_ASSEMBLY_ITEM_SETUP)
+
+/// ----------- Trigger/Applier/Target-Assembly-Related Procs -----------
+
+/obj/item/device/radio/electropack/proc/assembly_setup(var/manipulated_electropack, var/obj/item/assembly/parent_assembly, var/mob/user, var/is_build_in)
+	//dangerous assemblies should likely all fall under contraband level 4
+	var/electropack_assembly_contraband_level = 4
+	// we update the contraband now to reflect the newly added tank
```
TobleroneSwordfish: Tank?

code/obj/item/device/radios/electropack.dm
```diff
@@ -16,20 +16,85 @@ TYPEINFO(/obj/item/device/radio/electropack)
...
+	APPLY_ATOM_PROPERTY(parent_assembly, PROP_MOVABLE_VISIBLE_GUNS, parent_assembly, max(GET_ATOM_PROPERTY(parent_assembly,PROP_MOVABLE_VISIBLE_CONTRABAND), electropack_assembly_contraband_level))
```
TobleroneSwordfish:
```suggestion
	APPLY_ATOM_PROPERTY(parent_assembly, PROP_MOVABLE_VISIBLE_CONTRABAND, parent_assembly, max(GET_ATOM_PROPERTY(parent_assembly,PROP_MOVABLE_VISIBLE_CONTRABAND), electropack_assembly_contraband_level))
```
?

TobleroneSwordfish (changes requested):

code/obj/item/device/radios/electropack.dm
M-Earthfire: oh goddamnit, i fixed that in the other PR for other assemblies in general and now i got it wrong here....

code/obj/item/device/radios/electropack.dm
M-Earthfire: i may have copy/pasted that from the singletank assemblies... fixed the comment

PR #25441, opened 2025-12-30
Transceivers receive fix

### LINKED ISSUE #25440: Transception pad recieve packets are broken
### Labels

[Bug] [Major]

### Description

Transception pads refuse recieve packets because they are checking for number safe only and doesnt include number strings

### Steps to reproduce

1. Setup a mechcomp contraption that looks like the first image in additional information
2. Make the Wifi component not picky at all (remove netID filtering and pda message filtering)
3. Hit receive on the Transception pad computer
4. Copy the packet displayed on the Sign component and put it in the button
5. No receive?

### Expected Behavior

For the packet to not be picky about if its a string or a number

Image

Packet using the terminal that works

Image

Packet that uses mechcomp that doesnt work

Image

Mikeistaken:
[A-Engineering] [A-Game-Objects] [C-Bug]

## About the PR 
Fixes a bug with Nadir Transceivers that prevents them from receiving using Mech-Comp
## Why's this needed? 
Fixes #25440
## Testing 
Before Fix, Used packet command=receive;data=1;address_1=[netid] and it not work
After fix, successful receive
invalid receive (one) nothing

### DIFF
--- a/code/WorkInProgress/nadir_antenna.dm
+++ b/code/WorkInProgress/nadir_antenna.dm
@@ -689,8 +689,9 @@ TYPEINFO(/obj/machinery/transception_pad)
 					if("send")
 						src.attempt_transceive()
 					if("receive")
-						var/sigindex = signal.data["data"]
-						src.attempt_transceive(text2num_safe(sigindex))
+						var/sigindex = text2num_safe(signal.data["data"])
+						if (isnum_safe(sigindex))
+							src.attempt_transceive(sigindex)
 
 
 	proc/post_signal(datum/signal/signal, var/newfreq)


code/WorkInProgress/nadir_antenna.dm
```diff
@@ -690,8 +690,7 @@ TYPEINFO(/obj/machinery/transception_pad)
 						src.attempt_transceive()
 					if("receive")
 						var/sigindex = signal.data["data"]
-						if(isnum_safe(sigindex))
-							src.attempt_transceive(sigindex)
+						src.attempt_transceive(text2num_safe(sigindex))
```
pgmzeta: Run text2num_safe on the signal.data and leave the isnum_safe check in to account for null/invalid data. If a null is passed into `attempt_transceive` it will try to send instead of receiving.

Mikeistaken: alright, ill try it and merge it if it works

Mikeistaken: Yep that worked, adding it to the code, thank you.

PR #25443, opened 2025-12-31
Gene Booth Menu Adjustment

QuiteLiterallyAnything:
[QoL] [Game Objects]
## About the PR 
This PR aims to make interfacing with the gene booth easier for players who have access to its gene locking and price adjustment menu. To this end, the menu will now only appear when the booth is hit with a valid ID. Silicons will still always find it open since they have no IDs to swipe the booth with.
## Why's this needed? 
The way this menu is scripted to appear makes it one of the more severe minor annoyances currently in game. Consider the following facts:

- The menu always appears when someone with sufficient access clicks on the booth.
- The menu can be closed, but it will appear again the moment a gene is selected (due to a seemingly redundant line in 'context_actions.dm' that is also removed with this PR).
- The menu appears at the center of the screen and covers most gene selection icons.
- Most players attempting to use the gene booth will use it (and its UI) multiple times if it has a decent selection.
- The menu can only be staved away when dragged to a different portion of the screen, but even there it will constantly reappear (flashing) as the gene booth is interacted with. 
- The only solution to the menu is to not have access to it all, but that is for naught as without access to _it_ there is no access to an _ID_. And without an _ID_ there is no access to _digital currency_. And without _digital currency_ there is no access to the gene booth's **_priced contents._**

With all of this combined, the menu makes actually getting genes from the booth rather tedious.
## Testing 
On a debug build of the codebase I swiped a variety of ID cards through a stocked gene booth and found that the menu appeared as appropriate. I also used the gene booth a few times and interacted with it's menus as both an AI and cyborg. 

## Changelog 

```changelog
(u)QuiteLiterallyAnything
(+)The gene booth's locking and price adjustment menu now only appears when the booth is hit with an ID. 
```

### DIFF
--- a/code/modules/medical/genetics/geneticsBooth.dm
+++ b/code/modules/medical/genetics/geneticsBooth.dm
@@ -140,21 +140,36 @@ TYPEINFO(/obj/machinery/genetics_booth)
 		if(length(src.offered_genes) && !issilicon(user))
 			. += "You could swipe a sufficiently <b>high-level ID</b> through it to adjust its contents and pricing."
 
+	attack_ai(mob/user)
+		if (src.allowed(user))
+			src.pick_action(user)
+
+	#define BOOTH_ACTION_SELECT "Select Gene"
+	#define BOOTH_ACTION_ADMIM "Admin Panel"
+	proc/pick_action(user)
+		if(length(src.offered_genes) <= 0)
+			boutput(user, SPAN_NOTICE("[src] has no products available for purchase right now."))
+			return
+		switch(tgui_alert(user, "What do you want to do with [src]?", "Select Action", list(BOOTH_ACTION_SELECT, BOOTH_ACTION_ADMIM, "Nothing")))
+			if (BOOTH_ACTION_SELECT)
+				src.show_context_options(user)
+			if (BOOTH_ACTION_ADMIM)
+				src.show_admin_panel(user)
+
+	#undef BOOTH_ACTION_SELECT
+	#undef BOOTH_ACTION_ADMIM
+
 	attack_hand(var/mob/user)
 		if (occupant)
-			user.show_text("[src] is currently occupied. Wait until it's done.", "blue")
+			boutput(user, SPAN_NOTICE("[src] is currently occupied. Wait until it's done."))
 			return
-
 		if (status & (NOPOWER | BROKEN))
 			boutput(user, SPAN_ALERT("The gene booth is currently nonfunctional."))
 			return
 		if(length(src.offered_genes))
 			src.show_context_options(user)
 		else
-			user.show_text("[src] has no products available for purchase right now.", "blue")
-			return
-		if(issilicon(user)) // Silicons can't open it with an ID, but they don't often use the gene booth.
-			src.show_admin_panel(user) // ^ Always opening it for them shouldn't be too bad then.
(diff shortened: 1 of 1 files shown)


pgmzeta:
!merge_upstream

code/modules/medical/genetics/geneticsBooth.dm
```diff
@@ -145,9 +150,11 @@ TYPEINFO(/obj/machinery/genetics_booth)
 			return
 		if(length(src.offered_genes))
 			src.show_context_options(user)
-			src.show_admin_panel(user)
 		else
 			user.show_text("[src] has no products available for purchase right now.", "blue")
+			return
+		if(issilicon(user)) // Silicons can't open it with an ID, but they don't often use the gene booth.
```
JORJ949: I'd put this above showing the offered genes with an early return, because silicons can't get genes so don't really care to get that menu and will still get both if you don't.

QuiteLiterallyAnything: The control panel menu does not allow a player to change the booth's currently selected gene. Making this change will (for instance) prevent rogue silicons from tricking someone into getting a bad gene from a gene booth. **_Silicons will be completely unable to change the booth's selected gene._**

Are you absolutely sure I should make this change? (This may necessitate adding "silicons can no longer change the gene selected by gene booths" to the changelog.)

code/modules/medical/genetics/geneticsBooth.dm
```diff
@@ -145,9 +150,11 @@ TYPEINFO(/obj/machinery/genetics_booth)
...
+			src.show_admin_panel(user) // ^ Always opening it for them shouldn't be too bad then.
```
pgmzeta: move to attack_ai

code/modules/medical/genetics/geneticsBooth.dm
```diff
@@ -389,6 +404,11 @@ TYPEINFO(/obj/machinery/genetics_booth)
 					src.eject_occupant(0,0, direction)
 
 	attackby(obj/item/W, mob/user)
+		var/obj/item/card/id/id_card = get_id_card(W)
+		if(length(src.offered_genes) && istype(id_card))
+			src.show_admin_panel(user, id_card)
+			return
```
pgmzeta: feel like swiping an ID should induce a lock/unlock toggle, not bring up a menu

code/modules/medical/genetics/geneticsBooth.dm
```diff
@@ -169,20 +176,28 @@ TYPEINFO(/obj/machinery/genetics_booth)
...
-					. += " * <A href='byond://?src=\ref[src];op=\ref[P];action=lock'>[P.locked ? "Locked" : "Unlocked"]</A></small><BR/>"
+	/// Displays the booth's gene selection locking and price adjustment UI.
+	proc/show_admin_panel(mob/user, obj/item/card/id)
+		if(status & (NOPOWER | BROKEN))
+			return
+		if(!user)
+			return
+		if(istype(id) && !src.allowed(user) && !issilicon(user))
+			playsound(src, 'sound/machines/buzz-sigh.ogg', 50, TRUE, 0)
+			return
```
pgmzeta: use `check_access`, not `allowed` since you are checking the ID.

you should forward the silicon's internal ID into this as well to simplify the check and prevent an e.g. syndicate-only gene booth from being messed with by station silicons

PR #25445, opened 2025-12-31
Syndicate portable clown announcement computer

JORJ949:
## About the PR 
Adds a foldable, ultra illegal clown announcement computer that doesn't explode when used to the clown traitor+headrev uplinks for 6tc

## Why's this needed? 
I've given these to clowns via varediting before several times and I think its funny for syndicate clowns to be able to annoy everyone with consistent clown announcements until the crew gets annoyed enough and grabs their pitchforks. Costs 5tc so you can only buy a maximum of 2 after your first gets crushered by the angry mob.

## Testing 
image
image

## Changelog 

```changelog
(u)JORJ949
(*)Syndicate clowns can now find a highly illegal, portable, non-explosive version of the clown announcement computer in their uplinks for 5TC.
```

### DIFF
--- a/code/datums/syndicate_buylist.dm
+++ b/code/datums/syndicate_buylist.dm
@@ -594,7 +594,7 @@ ABSTRACT_TYPE(/datum/syndicate_buylist/traitor)
 /datum/syndicate_buylist/traitor/clown_announcement_computer
 	name = "Portable Clown Announcement Computer"
 	items = list(/obj/machinery/computer/announcement/clown/foldable)
-	cost = 6
+	cost = 5
 	vr_allowed = FALSE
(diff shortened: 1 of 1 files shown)


CelineTheYeen:
6 TC is too high, I think 4 might be good with the same limit cash briefcases have. It valids you, is decently replaceable and offers nothing else. I'm aware it's a clown item but every other one is either a genuinely powerful item or has a use, this doesn't. Maybe the briefcase could be made into an alright weapon (like 12 damage) or have more gimmicky features bolted onto it if you want to keep the price.

Megaphones are 5 TC and serve a similar purpose while being arguably a lot more annoying and noticeable.

JORJ949:
Fair points, but I still want it to be an expensive purchase so i've reduced it to the megaphone price of 5

PR #25456, opened 2026-01-01
Moves wraith's mass whisper to base kit, no cooldown for blood writing

CalliopeSoups:
## About the PR 

This PR does two things since they're small and related. I moved mass whisper to base kit and recoloured it to purple so it fits in, blood writing also has no cooldown now.

## Why's this needed? 

Wraith is lacking in communication and I think this is a good start. Trickster has enough toys that I don't think taking this from them matters much, and having a mass whisper base kit makes sense to me. As for blood writing I just felt as though it feels better.

## Testing 

Tested both changes, seems fine.

## Changelog 

```changelog
(u)CalliopeSoups
(+)Trickster wraith's mass whisper is now base kit
(+)Wraith's blood writing now has no cooldown
```


JORJ949 (approved):
Wraiths do need more escalation tools.

PR #25458, opened 2026-01-01
Headset EMP Improvement

TheGeneralJay:
## About the PR 
Adds a descriptive message when your headset is affected by an EMP. Headset speakers now automatically toggle back on after 90 seconds (same timer as security cameras when hit with EMPs).

## Why's this needed? 
Getting your headset hit by an EMP is cool and can be utilized by antags if timed properly. The issue is that there was formerly no message to indicate specifically that your headset had been turned off, which just led to people both new and veteran either forgetting or not realizing that they needed to turn their headset on again to hear radios. With the addition of proper EMP storms and the much more common likelihood of headsets getting nailed every so often, I think it's a good QoL change to have them automatically toggle back on (if not done so manually) after a set amount of time.

## Testing 
Ensured EMPs trigger the headset message, and made sure the headset speaker turns on after a set amount of time.

{C28B3976-9482-4F4E-980A-D9CA73089240}

## Changelog 

```changelog
(u)TheGeneralJay
(*)Headsets now malfunction fully and cannot be turned on manually for a set duration after an EMP. They will automatically toggle back on after the duration is up.
```

### DIFF
--- a/code/modules/speech/modules/listen/effects/radio.dm
+++ b/code/modules/speech/modules/listen/effects/radio.dm
@@ -3,7 +3,7 @@
 
 /datum/listen_module/effect/radio/process(datum/say_message/message)
 	var/obj/item/device/radio/radio = src.parent_tree.listener_parent
-	if (!istype(radio) || radio.bricked)
+	if (!istype(radio) || radio.bricked || radio.emp)
 		return
 
 	var/signal_frequency
--- a/code/obj/item/device/radios/_radio.dm
+++ b/code/obj/item/device/radios/_radio.dm
@@ -94,6 +94,10 @@ TYPEINFO(/obj/item/device/radio)
 	var/wires = WIRE_SIGNAL | WIRE_RECEIVE | WIRE_TRANSMIT
 	/// The build status of this radio, determining whether it can be attached to other objects, and other objects attached to it.
 	var/b_stat = FALSE
+	/// The radio has malfunctioned due to an EMP
+	var/emp = FALSE
+	/// The message that should be displayed when attempting to use a radio that is under the effects of an EMP.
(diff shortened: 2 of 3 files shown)


M-Earthfire:
90 seconds is such a neglectable time. This means you can get into an Engagement, get emp'ed and habe your headset turn on again while it's still ongoing. 

This may be less of an occurance on emp events, but it is certainly one with emp grenades. 

If someone forgets to activate their headset (or can't due to an active fight), 5 minutes seems more appropriate.

JORJ949:
Honestly feel like it should be force disabled for a while like other EMPd objects such as cameras are, this way it has a forceful impact beyond just immediately turning your headset back on and makes more sense as to why it turns back on automatically as it being able to turn back on but not doing so automatically feels kinda weird.

TheGeneralJay:
> Honestly feel like it should be force disabled for a while like other EMPd objects such as cameras are, this way it has a forceful impact beyond just immediately turning your headset back on and makes more sense as to why it turns back on automatically as it being able to turn back on but not doing so automatically feels kinda weird.

That makes sense. I've adjusted it so it now no longer allows you to open the UI to toggle anything for the duration of the EMP (still the same duration as cameras), and it blocks out being able to send messages as well. Sort of followed the logic of the 'bricked' variable. This makes it more impactful, while not risking you forgetting about it for an extended amount of time, I think.

I retested it and it all works as intended as far as I can tell. Lmk if I overlooked anything.

TobleroneSwordfish (approved):

AHHH-IM-DYING-HELP-ME-AHHH-IT-BURNS:
That's gonna HURT for EMP field gene players

JORJ949:
> That's gonna HURT for EMP field gene players

Good they deserve it

PR #25462, opened 2026-01-02
Makes Glass Recycler UI Appear Less Often

QuiteLiterallyAnything:
[Game Objects] [QoL]
## About the PR 
This PR replaces a single `ui_interact()` call which makes glass recyclers open their UI any time glass is added to them. A more direct ~`tgui_process.try_update_ui()`~ `tgui_process.update_uis()` call has been added instead.
## Why's this needed? 
It is currently impossible to add glass to a glass recycler without its UI opening. This is unnecessary as it is often the case that a player may desire to simply add glass to a recycler without also dispensing glass from it. 
## Testing 
On a debug build of the codebase I spawned a glass recycler and added glass to it. Its UI appeared to update and function properly when both unopened and never previously opened. When opened its glass count appeared to increment slower, so I added the previously mentioned ~`try_update_ui()`~ `update_uis()` call.

### DIFF
--- a/code/obj/machinery/glass_recycler.dm
+++ b/code/obj/machinery/glass_recycler.dm
@@ -134,7 +134,7 @@ TYPEINFO(/obj/machinery/glass_recycler)
 			user.visible_message(SPAN_NOTICE("[user] inserts [W] into [src]."))
 			user.u_equip(W)
 			qdel(W)
-			tgui_process.try_update_ui(user, src)
+			tgui_process.update_uis(src)
 			var/sound = pick('sound/impact_sounds/Glass_Shatter_1.ogg', 'sound/impact_sounds/Glass_Shatter_2.ogg', 'sound/impact_sounds/Glass_Shatter_3.ogg')
 			playsound(src.loc, sound, 40, 0, 0.1)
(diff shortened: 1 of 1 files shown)


code/obj/machinery/glass_recycler.dm
```diff
@@ -134,7 +134,7 @@ TYPEINFO(/obj/machinery/glass_recycler)
 			user.visible_message(SPAN_NOTICE("[user] inserts [W] into [src]."))
 			user.u_equip(W)
 			qdel(W)
-			ui_interact(user)
+			tgui_process.try_update_ui(user, src)
```
ZeWaka: This just updates it for that one user. I'm pretty sure there's a function to update it per-datum it's attached to, and thus all subscribed clients.

QuiteLiterallyAnything: Would that be `update_uis()`?

ZeWaka: yes

QuiteLiterallyAnything: Okay, I've replaced it.

PR #25465, opened 2026-01-02
Fix !merge_upstream

TobleroneSwordfish:
## About the PR 
Fixes the merge_upstream workflow and makes it rebuild the TGUI bundle properly and commit that too.
## Why's this needed? 
Useful command that was broken for a long time!
## Testing 
See: https://github.com/TobleroneSwordfish/goonstation/pull/1


Laboredih123:
too useful, gotta deny this

TobleroneSwordfish:
Zewaka is fixing this better

PR #25469, opened 2026-01-03
Adds a Mapping Helper for Putting Glassware in Dispensers

QuiteLiterallyAnything:
[Mapping] [Internal] [QoL]
## About the PR 
This PR adds a new "glassware spawn" mapping helper that puts glassware into any dispenser it's mapped on. I've also added two procs for getting containers into dispensers and espresso machines without a user.
## Why's this needed? 
I've noticed that a lot of maps (Cogmap 2, Donut 3, et cetera) have beakers, pitchers, and cups mapped over some of their dispensers (particularly in their bars). This helper simplifies that practice and makes it cleaner. Why have round-start beakers spawn _on_ dispensers when you could just have them spawn _in_ dispensers?
## Testing 
I've added this helper to DevTest (committed for demonstration purposes) and verified that it works on a debug build of the codebase. Glassware added through the helper can be ejected properly and handles reactions.

### DIFF
--- a/code/modules/chemistry/Chemistry-Dispenser.dm
+++ b/code/modules/chemistry/Chemistry-Dispenser.dm
@@ -245,23 +245,17 @@ TYPEINFO(/obj/machinery/chem_dispenser)
 
 	/// Proc for adding a beaker to a dispenser (`user` optional though necessary if one exists).
 	proc/add_glassware(var/obj/item/reagent_containers/container, var/mob/user)
-		if(!user)
-			src.beaker = container
-			container.set_loc(src)
-			APPLY_ATOM_PROPERTY(container, PROP_ITEM_IN_CHEM_DISPENSER, src)
-			container.reagents?.handle_reactions()
-			src.UpdateIcon()
+		if(QDELETED(container) || container.cant_drop)
 			return
 
-		src.beaker = container
+		user?.drop_item(container)
+		container.set_loc(src)
 		APPLY_ATOM_PROPERTY(container, PROP_ITEM_IN_CHEM_DISPENSER, src)
-		if(!container.cant_drop)
-			user.drop_item()
-			if(!container.qdeled)
-				container.set_loc(src)
+		src.beaker = container
 		container.reagents?.handle_reactions()
 		src.UpdateIcon()
-		src.ui_interact(user)
+		if (user)
+			src.ui_interact(user)
 
 	proc/take_damage(var/damage_amount = 5)
 		src.health -= damage_amount


pgmzeta:
!merge_upstream

code/modules/chemistry/Chemistry-Dispenser.dm
```diff
@@ -251,6 +251,14 @@ TYPEINFO(/obj/machinery/chem_dispenser)
 			boutput(usr, SPAN_ALERT("You can't use that as an output target."))
 		return
 
+	/// Simplified proc for adding a beaker to a dispenser without worrying about a user.
```
pgmzeta: what proc is this simplifying? Should that proc just check is the user is null instead?

code/obj/machinery/espresso.dm
```diff
@@ -165,6 +165,13 @@ TYPEINFO(/obj/machinery/espresso_machine)
 		qdel(src)
 		return
 
+	/// Simplified proc for adding a cup to an espresso machine without worrying about a user.
```
pgmzeta: ditto from dispenser comment

pgmzeta (changes requested):
any reason the helper shouldn't just try to find a compatible glass/cup on the turf instead of hard-coding those specific containers?

code/modules/chemistry/Chemistry-Dispenser.dm
QuiteLiterallyAnything: It's an alternative to `attackby()` which is a `PROTECTED_PROC`. Dispensers do not appear to have a generic `add_beaker()` as is. I could refactor this into something more standard like that (though it may make the diffs look messy since it would warrant being rewritten into existing dispenser code).

QuiteLiterallyAnything:
> any reason the helper shouldn't just try to find a compatible glass/cup on the turf instead of hard-coding those specific containers?

Done, should I refactor the new dispenser procs into more standard ones as well? It would require adjustments to the current `attackby()` logic.

code/modules/mapping/helpers/beaker_spawn.dm
```diff
@@ -0,0 +1,63 @@
+/// The default glassware type glassware helpers use for chem dispensers.
+#define DEFAULT_CHEM_GLASSWARE /obj/item/reagent_containers/food/drinks/drinkingglass/pitcher
+/// The default glassware type glassware helpers use for alcohol dispensers.
+#define DEFAULT_ALCOHOL_GLASSWARE /obj/item/reagent_containers/food/drinks/drinkingglass/pitcher
+/// The default glassware type glassware helpers use for soda dispensers.
+#define DEFAULT_SODA_GLASSWARE /obj/item/reagent_containers/glass/beaker/large
```
pgmzeta: the names here and their usage below don't match; you can probably just have two vars of "chem" and "drink" instead

code/modules/mapping/helpers/beaker_spawn.dm
```diff
@@ -0,0 +1,63 @@
...
+			if(container)
+				dispenser.add_beaker_no_user(container)
+				return
+
+			if(istype(dispenser, /obj/machinery/chem_dispenser/soda))
+				container = DEFAULT_CHEM_GLASSWARE
+			else if(istype(dispenser, /obj/machinery/chem_dispenser/alcohol))
+				container = DEFAULT_ALCOHOL_GLASSWARE
+			else if(istype(dispenser, /obj/machinery/chem_dispenser))
+				container = DEFAULT_SODA_GLASSWARE
```
pgmzeta: `/obj/machinery/chem_dispenser/chef` sub-type is on some maps too, needs a pitcher too

code/obj/machinery/espresso.dm
```diff
@@ -165,6 +165,13 @@ TYPEINFO(/obj/machinery/espresso_machine)
...
+	proc/add_cup_no_user(var/obj/item/reagent_containers/food/drinks/espressocup/cup)
```
pgmzeta: just call it add_cup then

code/modules/chemistry/Chemistry-Dispenser.dm
```diff
@@ -251,6 +251,14 @@ TYPEINFO(/obj/machinery/chem_dispenser)
...
+	proc/add_beaker_no_user(var/obj/item/reagent_containers/container)
```
pgmzeta: call it add_glassware

code/modules/mapping/helpers/beaker_spawn.dm
QuiteLiterallyAnything: Done

code/modules/mapping/helpers/beaker_spawn.dm
QuiteLiterallyAnything: Whoops-- got that mixed up somehow. Fixed and reduced to two.

code/obj/machinery/espresso.dm
QuiteLiterallyAnything: Done

code/modules/chemistry/Chemistry-Dispenser.dm
QuiteLiterallyAnything: Changed

QuiteLiterallyAnything:
> yes, the attackby procs should call these new procs so the behavior doesn't drift.

Okay, done. I also integrated the proc into chem dispenser `ui_act()` as appropriate. 

[The chem dispenser file's blame](https://github.com/goonstation/goonstation/blame/master/code/modules/chemistry/Chemistry-Dispenser.dm) shows that there have been a lot of minor adjustments as bug fixes over the years. I've checked that none of the basic cases of beaker interaction are bugged (cyborgs with beakers included) and everything seems fine.

code/modules/chemistry/Chemistry-Dispenser.dm
```diff
@@ -251,6 +243,26 @@ TYPEINFO(/obj/machinery/chem_dispenser)
...
+
+		src.beaker = container
+		APPLY_ATOM_PROPERTY(container, PROP_ITEM_IN_CHEM_DISPENSER, src)
+		if(!container.cant_drop)
+			user.drop_item()
+			if(!container.qdeled)
+				container.set_loc(src)
+		container.reagents?.handle_reactions()
+		src.UpdateIcon()
+		src.ui_interact(user)
```
pgmzeta: should be able to compress these two codepaths down as they're mostly the same behaviors

```suggestion
	proc/add_glassware(var/obj/item/reagent_containers/container, var/mob/user)
		if(QDELETED(container) || container.cant_drop)
			return

		user?.drop_item(container)
		container.set_loc(src)
		APPLY_ATOM_PROPERTY(container, PROP_ITEM_IN_CHEM_DISPENSER, src)
		src.beaker = container
		container.reagents?.handle_reactions()
		src.UpdateIcon()
		if (user)
			src.ui_interact(user)
```

QuiteLiterallyAnything: Fair, I had it as it was out of both an abundance of caution and the fear that not keeping near identical functionality could cause obscure, minor, and previously fixed bugs to resurface. That's probably not all too likely though.

pgmzeta (approved):

PR #25473, opened 2026-01-04
Shrouded vampire voices show up as "Unknown"

pgmzeta:
[balance][gammodes]
## About the PR 
Add speech modifier "shrouded" that overrides the speaker's name to "Unknown"
Apples speech modifier to players under Illusory Shroud's effect.

## Why's this needed? 
Cloaking is good, but hard to work with if you can't hide your voice. This change hides your voice in a way that doesn't immediately out you as a vampire.

## Testing 
Locally tested with and without shroud using headset radio, local radio, etc

## Changelog 

```changelog
(u)glowbold
(*)Vampires using Illusory Shroud will have their voices show as "Unknown" 
```

### DIFF
--- a/code/modules/speech/modules/speech/modifiers/shrouded.dm
+++ b/code/modules/speech/modules/speech/modifiers/shrouded.dm
@@ -2,8 +2,9 @@
 	id = SPEECH_MODIFIER_SHROUDED
 
 /datum/speech_module/modifier/shrouded/process(datum/say_message/message)
+	. = message
+
 	if (message.output_module_channel == SAY_CHANNEL_THRALL)
-		return message
+		return
 
 	message.speaker_to_display = "Unknown"
-	return message


code/modules/speech/modules/speech/modifiers/shrouded.dm
```diff
@@ -0,0 +1,9 @@
+/datum/speech_module/modifier/shrouded
+	id = SPEECH_MODIFIER_SHROUDED
+
+/datum/speech_module/modifier/shrouded/process(datum/say_message/message)
+	if (message.output_module_channel == SAY_CHANNEL_THRALL)
+		return message
+
+	message.speaker_to_display = "Unknown"
+	return message
```
Mister-Moriarty:
```suggestion
	. = message

	if (message.output_module_channel == SAY_CHANNEL_THRALL)
		return

	message.speaker_to_display = "Unknown"
```
Consistency with the other modules.

PR #25479, opened 2026-01-04
Clone defects occur based on time since clone scan data saved

pgmzeta:
[medical][experimental][balance]
## About the PR 
Clone defects occur based on time since clone scan data saved.

Every 15 minutes since the scan a minor defect is added to the clone. At both 60 and 90 minutes, a major defect is added instead.

Disks made on spawn (i.e. sec disks) have a buffer time of 30 minutes from creation before they begin to be affected; i.e. it's like they got a scan at 30 minutes into the shift.

## Why's this needed? 
Getting a pre-scan early into the shift is a zero-risk way to largely eliminate the danger of death for the rest of the round. This change encourages getting clone-scans only before knowingly risky activity, without removing the ability to get a pre-scan entirely.

A separate issue with zero-risk cloning that this change addresses it the any complexity or depth to cloning; anyone and everyone will try to clone everyone, regardless of circumstance. This adds some decision making when determining when to clone: do you wait for the body for a fresh scan, or use the disk and take a potential risk on the backup.

Furthermore, zero-risk, immediate cloning can and has been a severe impediment for antagonists: vampires unable thrall someone because they're cloned; antagonist identities outed immediately; kidnapping moot due to the victim having a disk.

Lastly, having perfect clone behavior affects player behavior in a major way since the risk of sitting the round out is zero. You are allowed to take unlimited risk with a clone scan in a way that puts antagonists on the back foot; antags have to play to live, but crew only needs to rush them knowing coming back to life is a click away.

Overall I think choices you're asked to make in-game should require judgement and impact the way the round plays out. While pre-scanning provides the latter, it fails at the former by being a zero-question, every-shift action for a lot of people.

## Testing 

image

## Changelog 

```changelog
(u)glowbold
(*)Clone defects occur based on time since clone scan data saved.
(+)Security Disks have a bonus 30 minutes from creation before they begin to be affected.
```

### DIFF
--- a/code/procs/jobprocs.dm
+++ b/code/procs/jobprocs.dm
@@ -626,7 +626,7 @@ Equip items from body traits.
 				implant.scanned_here = locate(/area/centcom/reconstitutioncenter) || null
 				R.fields["imp"] = implant
 				R.fields["mind"] = src.mind
-				R.fields["scantime"] = TIME + 60 MINUTES // scan decay buffer
+				R.fields["scantime"] = TIME + 30 MINUTES // scan decay buffer
 				D.root.add_file(R)
 
 				D.name_suffix("([src.real_name])")


Egregorious:
To be honest, with cloning's current design I don't like the direction of discouraging pre-scans. It seems to me the alternative that this pushes towards is that clone scans are more often only done when the body comes in, which winds up reducing interdepartmental interaction on the whole.

Prescans are not very exciting, but in my opinion the fact they encourage people visiting medbay when they otherwise wouldn't is better than the trivial and largely uninteractive resurrect button that cloning is otherwise.

DisturbHerb:
> Prescans are not very exciting, but in my opinion the fact they encourage people visiting medbay when they otherwise wouldn't is better than the trivial and largely uninteractive resurrect button that cloning is otherwise.

it's interesting because in my view, whatever visits people make to medbay end up being a chore for one party and trivial for another anyway. i don't get a lot out of someone walking into the scanner and walking out as a doctor. ymmv

Egregorious:
Most actions in the game aren't very fun when ambiguated, but as a social interaction game mechanics which encourage different parts of the station to visit one another tend to be good for getting characters to connect in a way that's natural in roleplay. Stepping into the clone booth isn't the win, but I tend to find a good amount of interaction talking someone from a different department when we've both gone to get a scan in the beginning of the round. Often it's a character I would never have interacted with that round otherwise.

Mikeistaken:
How is this planned to scale with defect prone?

Spyritdragon:
I feel like since scanning disks were added - which I loved as a change - there's been so many more opportunities to fool around with the scans as an antagonist. I feel like this change is going to either heavily incentivise the medical team to avoid using a saved disk if at all possible, which would take away some of these opportunities to mess with them as an antag again and that'd make me a bit sad. 

It also feels like encouraging the regular returning to update your cloning record as say, a security member or a science member who's prone to getting their body lost, just adds a bit more frustrating tedium, because with this you're 100% going to want to get your scan updated an hour into the shift. 

I'm absolutely not opposed to some further nerfs to how strong cloning still is but I'm not sure this change feels like a nice way to do it - it doesn't really feel like it addresses the situations where cloning is very strong while it does add to the frustration and tedium.

pgmzeta:
Hey all, good points made. I'm opening this up to more discussion on the forums before this moves forward or gets tweaked.

> How is this planned to scale with defect prone?

It would be in addition to the defects given by defect prone.

PR #25483, opened 2026-01-04
Securitron power toggle tweaks

TheGeneralJay:
## About the PR 
Cleans up the implementation of secbot power toggling and allows the security whistle to turn off secbots when pointed at.

## Why's this needed? 
The janitor whistle is able to toggle off cleanbots, so why shouldn't security whistles be able to do the same to secbots? Being able to toggle that pesky securitron off when it's getting a little crazy about taking down the criminal is nice. This PR aims to clean up how it's handled behind the scenes and allow for officers (and antags) to toggle securitrons off if they use the security whistle.

## Testing 
Attached video shows change in action. Also tested to ensure manual toggling with human hand / silicon is still functional.

https://github.com/user-attachments/assets/aac60ba8-abde-4461-a337-cd1ef0b2cdd1

## Changelog 
```changelog
(u)TheGeneralJay
(+)Securitrons can now be toggled off with a security whistle by pointing at them.
```

### DIFF
--- a/code/datums/components/bot_command.dm
+++ b/code/datums/components/bot_command.dm
@@ -38,9 +38,8 @@ ABSTRACT_TYPE(/datum/component/bot_command)
 			if (secbot == target)
 				secbot.toggle_power()
 				src.bots -= secbot
-			else
-				if (iscarbon(target) && target != secbot.target)
-					secbot.EngageTarget(target, FALSE, FALSE, TRUE)
+			else if (iscarbon(target) && target != secbot.target)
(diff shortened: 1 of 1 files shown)


Mantweerp:
> On further thought, make sure you can't turn off emagged securitrons with this.

Maybe it would be better if they started chasing the person who pointed at them

TheGeneralJay:
> On further thought, make sure you can't turn off emagged securitrons with this.

Good catch. I just checked, I don't seem to be able to turn them off when emagged with the whistle, so it should be all good.

PR #25485, opened 2026-01-04
Monkey barrel minor refactor + holographic clothes 

CalliopeSoups:
## About the PR 

This PR turns monkey barrels from storage objs that hold the monkeys always to a simple obj that holds one monkey, used as a template to put clothes on. When activated, every resulting monkey's clothes copy the icon of the template's, apart from a couple slots. 

Due to the way the code works now, there's a var admins can varedit to change how many monkeys will come out of the barrel. In-case you want to be evil.

## Why's this needed? 

A horde of angry monkeys wearing the same getup is pretty funny and will complement people's bits more. 

## Testing 

Done some basic testing, put different things in the template monkey's slots, used the barrel a few times. Yet to see what happens if the template monkey isn't there for some reason. I do think some of the code might need review cause this went through a couple iterations, some of it might be messy or could be done better.

dreamseeker_gbfp9oaodm

## Changelog 

```changelog
(u)CalliopeSoups
(+)Add clothes to a template monkey in a barrel (by right clicking) to be copied, without stats, to every monkey in the barrel!
```

### DIFF
--- a/code/obj/item/device/monkey_barrel.dm
+++ b/code/obj/item/device/monkey_barrel.dm
@@ -9,19 +9,18 @@
 	p_class = 3
 	is_syndicate = 1
 	density = 1
-	open_inv_anywhere = TRUE
-	var/mob/living/carbon/human/npc/monkey/angry/dummy
+	open_inv_within = TRUE
+	var/mob/living/carbon/human/npc/monkey/angry/template_monkey // spawns first, copies clothing icons to spawned monkeys
 	var/monkeys_to_spawn = 5
 
 	New()
 		..()
 		var/obj/item/barrel_signaller/M = new /obj/item/barrel_signaller(src.loc)
 		new /obj/item/clothing/suit/monkey(src.loc)
-		src.dummy = new/mob/living/carbon/human/npc/monkey/angry
-		SPAWN(0)
-			src.dummy.loc = src
-			src.contents += src.dummy
-			M.my_barrel = src
+		src.template_monkey = new/mob/living/carbon/human/npc/monkey/angry(src)
+		src.contents += src.template_monkey
+		src.template_monkey.real_name = "Holo-Clothes Template"
+		M.my_barrel = src
 
 	update_icon()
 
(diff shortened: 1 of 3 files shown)


code/obj/item/device/monkey_barrel.dm
```diff
@@ -1,38 +1,61 @@
...
 	desc = "More fun than a ValuChimp!"
 	icon = 'icons/obj/items/device.dmi'
 	icon_state = "barrel"
 	throwforce = 50
 	p_class = 3
-	locked = 1
 	is_syndicate = 1
-	spawn_contents = list(/mob/living/carbon/human/npc/monkey/angry = 6)
+	density = 1
+	open_inv_anywhere = TRUE
```
TobleroneSwordfish: This name is kind of odd because the variable is on the "anywhere", either move it to the mob type or rename it to something like `open_inventory_within`

code/obj/item/device/monkey_barrel.dm
```diff
@@ -1,38 +1,61 @@
...
+	var/mob/living/carbon/human/npc/monkey/angry/dummy
```
TobleroneSwordfish: Could use a comment explaining what it's for, or at least renaming to `template_monkey` or something to be clearer

code/obj/item/device/monkey_barrel.dm
```diff
@@ -1,38 +1,61 @@
...
+	var/monkeys_to_spawn = 5
 
 	New()
 		..()
 		var/obj/item/barrel_signaller/M = new /obj/item/barrel_signaller(src.loc)
 		new /obj/item/clothing/suit/monkey(src.loc)
+		src.dummy = new/mob/living/carbon/human/npc/monkey/angry
 		SPAWN(0)
+			src.dummy.loc = src
+			src.contents += src.dummy
```
TobleroneSwordfish:
```suggestion
		src.dummy = new/mob/living/carbon/human/npc/monkey/angry(src)
		SPAWN(0)
```
God only knows why that spawn is there, but this is a cleaner way to spawn one thing inside another

code/obj/item/device/monkey_barrel.dm
```diff
@@ -1,38 +1,61 @@
...
+		set category = "Local"
+		if (ishuman(usr) && src.dummy)
+			var/mob/living/carbon/human/user = usr
+			src.dummy.show_inv(user)
+
+	proc/monkey_go()
+		var/turf/targetTurf = get_turf(src)
+		var/obj/itemspecialeffect/poof/poof = new /obj/itemspecialeffect/poof
+		poof.setup(targetTurf)
+		for (var/i = src.monkeys_to_spawn,i>=0,i--)
```
TobleroneSwordfish:
```suggestion
		for (var/i in 1 to src.monkeys_to_spawn)
```
Technically both are correct but this is the Byond way to do it and I think it's a lot cleaner and less prone to off-by-one errors.

code/obj/item/device/monkey_barrel.dm
```diff
@@ -1,38 +1,61 @@
...
+			var/mob/living/carbon/human/npc/monkey/angry/barrel/monke = new /mob/living/carbon/human/npc/monkey/angry/barrel (targetTurf)
```
TobleroneSwordfish:
```suggestion
			var/mob/living/carbon/human/npc/monkey/angry/barrel/monke = new(targetTurf)
```
You can omit the type here, Byond will infer from the var type.

code/obj/item/device/monkey_barrel.dm
```diff
@@ -1,38 +1,61 @@
...
+			monke.copy_clothes(src.dummy)
+		src.drop_contents()
+		qdel(src)
+
+	proc/drop_contents()
+		for (var/atom/movable/thing in src.contents)
+			src.contents -= thing
+			thing.set_loc(get_turf(src))
```
TobleroneSwordfish: Not sure what this is for, does this mean the dummy is also released with its non-holographic clothing?

TobleroneSwordfish (changes requested):

Crowlemon:
This is an amazing idea and I hope it gets added in.

code/obj/item/device/monkey_barrel.dm
CalliopeSoups: It is yes, I figured to do it this way so you can retrieve your clothes afterward. Thinking about it though it's probably better if the template monkey just drops its clothes on the spot and gets qdeleted. That way I can make its name appear as Holographic Clothing Controller or something, and getting the clothes back from where the barrel was is a LOT easier than finding the right monkey.

code/obj/item/device/monkey_barrel.dm
CalliopeSoups: I think this is left over from when I was testing things while the code was broken

PR #25494, opened 2026-01-05
Makes More Snow Diggable

### LINKED ISSUE #25484: Can't dig trenches in snow.
### Labels

[BUG][MINOR]

### Description

You cant dig up trenches in the snow, it's a bummer at funerals when you cant bury the body!

### Steps to reproduce

1. get a shovel
2. exist during winter
3. attempt to dig a trench

### Expected Behavior

to be able to dig a trench.

Reported by: smallsandman
Client version: 516.1667
Active test merges: #25018
Round log date: 2026-01-04-19-01
Reported on: 2026-01-04 20:36:28
Round ID: [220508](https://goonhub.com/rounds/220508)
Map: [CLARION](https://goonhub.com/maps/clarion)

QuiteLiterallyAnything:
[Player Actions] [Bug]
## About the PR 
Fixes #25484

This PR makes a few more snow tile types diggable (particularly those that spawn in hydroponics during winter).  
## Why's this needed? 
Most types of snow are already diggable, and it seems like these were missed. If there is a greater reason that they _shouldn't_ be diggable then I am unaware of it. The `snow/snowball` subtype in particular needs to be diggable (for the sake of consistency) since it seasonally replaces an already diggable tile type.
## Testing 
On a debug build of the codebase I spawned, dug up, and reburied most of the newly diggable snow types.

### DIFF
--- a/code/turf/floors.dm
+++ b/code/turf/floors.dm
@@ -1448,7 +1448,8 @@ DEFINE_FLOORS(snowcalm,
 	icon = 'icons/turf/floors.dmi';\
 	icon_state = "snow_calm";\
 	step_material = "step_snow";\
-	step_priority = STEP_PRIORITY_MED)
+	step_priority = STEP_PRIORITY_MED;\
+	can_dig = TRUE)
 
 DEFINE_FLOORS(snowcalm/border,
 	icon_state = "snow_calm_border")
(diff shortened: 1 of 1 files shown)


code/turf/floors.dm
pgmzeta, line 1446: does it make sense to add `can_dig` to these snowcalm/snowrough definitions while we're here too?

QuiteLiterallyAnything: Probably. I recall trying that too at first, but I misunderstood the macro's syntax (leading me to think adding `can_dig` to it was impossible). In any case, done.

pgmzeta:
thank you for the quick change!

PR #25503, opened 2026-01-06
Increase handcrafted katana delimb chance

Midnightradar:
## About the PR 
Increases the delimb chance per hit on the handcrafted katane from 2% to 20%
## Why's this needed? 
Currently the handcrafted katana almost never cuts off limbs, a battle would end before a single limb would fall, plus the whole idea of the syndicate katana is that it doesnt deal a lot of damage, but cuts off limbs, but the handcrafted version deals even less damage and the delimb function is basically useless, making it just a medicore weapon that takes a lot of time to get 
## Testing 
image
The increased delimb chance also works properly


CalliopeSoups:
Not sure about the actual numbers but I'm in favour of a handcrafted katana buff. Doesn't need to be insane but you DO need to spend 20 minutes of your life making it.

JORJ949 (approved):
Probably fine, can be nerfed if people abuse it

pgmzeta:
!merge_upstream

PR #25508, opened 2026-01-06
hit cruiser area code with hammers

Skeletonman0:
[GAME OBJECTS] [BUG]
## About the PR 
Changes up some of the area entering / exiting logic for cruisers to hopefully make it more successful at properly removing / adding interior overlays

## Why's this needed? 
image
cruiser out of body experiences bad

## Testing 
Tested by spawning a cruiser on devtest and hopping in and out of the control pods, climbing up and down the ladders to try and desync the overlays.
cruisertest

### DIFF
--- a/code/modules/transport/cruisers/cruisers.dm
+++ b/code/modules/transport/cruisers/cruisers.dm
@@ -848,21 +848,33 @@
 
 	Entered(var/atom/movable/A, atom/oldloc)
 		. = ..()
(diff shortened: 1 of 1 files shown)


Skeletonman0:
can confirm that it resets your eye properly, tested while inside the control pod and just standing around on the top floor.
deleting the cruiser kicks you out onto the turfs outside and resets everything

pgmzeta (approved):

Issue #25499, opened 2026-01-06
View Jumping around very wildly.


goonstation-issuebot:
### Labels

[BUG][MINOR]

### Description

Sudden displacement of what I'm viewing to a random part of the station 

### Steps to reproduce

1. Be a plague rat in a garbage chute
2. Bind ctrl to eat filth
3. While still in the garbage chute, ctrl+click some filth
4. Have your view jump around very suddenly

### Expected Behavior

To always be looking at my sprite?

### Additional Information & Screenshots

I dont know how reproducible this is, but I wanted to stop and document what happened right as it happened. 

Reported by: skotcher
Client version: 516.1667
On server: Goonstation 4 Roleplay: Sylvester
Active test merges: #25018
Round log date: 2026-01-06-04-26
Reported on: 2026-01-06 05:40:28
Round ID: [220576](https://goonhub.com/rounds/220576)
Map: [COGMAP](https://goonhub.com/maps/cogmap)

PR #25509, opened 2026-01-07
Plague Rat Feeding Fix

QuiteLiterallyAnything:
[Bug] [Player Actions]
## About the PR 
Related to and may be considered a fix for issue #25499. (Final judgement on whether or not this PR should close that issue is left to maintainers.) 

This PR prevents plague rats from feeding on detritus while inside things.
## Why's this needed? 
I wasn't able to directly reproduce issue #25499 or locate its main cause. I did, however, notice that plague rats can feed on waste while inside pipes, closets, and similar objects. They can only do this when the ability is bound to a key, so I assume it's an exploit. 
## Testing 
On a debug build of the codebase I spawned a plague rat and placed some blood beside a disposals chute and closet. The blood could not be fed on from inside either, and the message indicating such to players appeared as appropriate.

### DIFF
--- a/code/modules/antagonists/wraith/critters/abilities/plaguerat.dm
+++ b/code/modules/antagonists/wraith/critters/abilities/plaguerat.dm
@@ -47,7 +47,7 @@ ABSTRACT_TYPE(/datum/targetable/critter/plague_rat)
 
 		if (!isturf(P.loc))
 			boutput(P, SPAN_ALERT("You can't feed while in [istype(P.loc,/obj/dummy/disposalmover) ? "a disposal pipe" : "\the [P.loc]"]."))
-			return FALSE
+			return CAST_ATTEMPT_FAIL_NO_COOLDOWN
 
 		for (var/obj/decal/cleanable/C in T)
 			for (var/D in decal_list)


code/modules/antagonists/wraith/critters/abilities/plaguerat.dm
```diff
@@ -45,6 +45,10 @@ ABSTRACT_TYPE(/datum/targetable/critter/plague_rat)
 
 		var/mob/living/critter/wraith/plaguerat/P = holder.owner
 
+		if (!isturf(P.loc)) // vvv Formatting copied from the plague rat bite ability. vvv
```
ZeWaka:
```suggestion
		if (!isturf(P.loc))
```

code/modules/antagonists/wraith/critters/abilities/plaguerat.dm
```diff
@@ -45,6 +45,10 @@ ABSTRACT_TYPE(/datum/targetable/critter/plague_rat)
 
 		var/mob/living/critter/wraith/plaguerat/P = holder.owner
 
+		if (!isturf(P.loc))
+			boutput(P, SPAN_ALERT("You can't feed while in [istype(P.loc,/obj/dummy/disposalmover) ? "a disposal pipe" : "\the [P.loc]"]."))
+			return FALSE
```
TobleroneSwordfish:
```suggestion
			return CAST_ATTEMPT_FAIL_NO_COOLDOWN
```
`FALSE` is actually the success return value for `cast`, we have these defines to make it less stupid.

QuiteLiterallyAnything: Whoops-- thank you.

PR #25512, opened 2026-01-07
Hacking a Manufacturer or Seed Fabricator marks you for arrest.

JORJ949:
## About the PR 
Hacking a manufacturer or seed fabricator automatically marks you for arrest if you are on cams and your face is visible.

## Why's this needed? 
- Pushes security to treat fabricator hacking as a minor crime.
- Discourages hacking fabricators for stronger gear.
- Avoiding this automated arrest leaves a paper trail for security to follow, unlike current hacked fabs which unless you open them leaves no indication they were hacked.

## Testing 
image
image

## Changelog 

```changelog
(u)JORJ949
(*)Hacking fabricators (including the seed fabricator) is now illegal and will mark you for arrest.
```

### DIFF
--- a/code/obj/submachine/seed.dm
+++ b/code/obj/submachine/seed.dm
@@ -902,10 +902,9 @@ TYPEINFO(/obj/submachine/seed_vendor)
 					src.hacked = 1
 					src.name = "Feed Sabricator"
 					var/mob/living/carbon/human/criminal = user
-					if(src.hacked)
-						logTheThing(LOG_STATION, user, "hacks [src] at [log_loc(src)]")
-						if(istype(criminal))
-							criminal.apply_automated_arrest("Unauthorised hacking of a fabrication device.")
+					logTheThing(LOG_STATION, user, "hacks [src] at [log_loc(src)]")
+					if(istype(criminal))
+						criminal.apply_automated_arrest("Unauthorised hacking of a fabrication device.")
 				update_static_data(user)
 			if (WIRE_MALF)
 				if (src.malfunction) src.malfunction = 0


code/obj/submachine/seed.dm
```diff
@@ -901,6 +901,11 @@ TYPEINFO(/obj/submachine/seed_vendor)
 				else
 					src.hacked = 1
 					src.name = "Feed Sabricator"
+					var/mob/living/carbon/human/criminal = user
+					if(src.hacked)
```
pgmzeta: redundant check in this logic branch

NibChocolateeny:
Not so sure...but ok, let us see howit is. can get around it.

cheekybrdy:
I'm against this personally, there's just too much content behind them and they typically lead indirectly to another crime that's already persecuted anyway (.22, weed) and this would probably just result in a bunch of people screaming to be taken off arrest in the same vein as jailbird. Either way I'm confident this would be controversial enough to warrant a discussion thread. Perhaps a less severe flag like suspect could be used instead?

M-Earthfire:
As a botanist, hacking seed vendors became so commonplace that i feel like it could need to be dialed back a little bit. And it gives botany more interaction... sure, i unirronically appreciate  more reasons to tomatoflood security.

The thing is: the seed vendor list will need a redo if this PR goes through. A lot of seeds that are important for others (e.g. mushrooms for chef and rancher) are locked behind it and many advanced splices as well.

Even more so since the most important addition botany can do for medbay right now is omnizine, which i suspect is to be specifically targeted here (weed)... remind me again why the pharmacy and the massive vendors of medbay are a thing?

Above that, the handheld seed fabricator is a thing. It works like a hacked sed fab and you get it out of the vendor.

That makes me question: why are vendors the exception here?

CalliopeSoups:
I don't think I'm fully against this but it has issues. Earthfire mentioned most of my real complaints, real things you might want to use are locked behind hacking the seed fabricator. That thing is hacked roundstart every shift not just because it's easy and without consequence but because you actually need things out of the hacked pool.

Vendors should definitely be included as you can get lethal ammo from hacking them like you can with .22. 

I think there's good things that could come out of this other than the intent of giving sec more minor crimes to deal with. If hacking these things comes with avoidable consequence, more things could perhaps be added to the hacked manu pools. More toys!

Similar to what cheeky mentioned, I think a less severe sechud flag that perhaps makes securitrons etc point at you and alert sec of your minor crime would be better for this, though I don't have much issue with it being a full arrest flag. Just something that I could see having use in the future too, maybe a good PR later on.

RSOD404:
the idea is neat, but i don't see myself being a big fan of being chased by secbots for yet another reason.
as stated before by others, seed fab should either be exempt or have the default seed list changed to accomodate this.
 like calliopesoups said, i'd rather more fun stuff be added to manufacturers and vendors to outweigh getting the arrest flag if caught, or use a less severe one where secbots whistle and point at you (maybe notifying sec of your location) but don't actually try to put you in cuffs

M-Earthfire:
I honestly think being set to arrest is just about right. Like jailbird, it gives more interaction than a simple "yeah, that ones showing up odd on my hud". Above that, it gives sec more to do and a "lower status" than arrest would simply give them an excuse to ignore fab hacking.

Above that, it's kinda easy to dodge. Just conceal your face and hide your ID while doing it. Like how you do as a QM when you open half the station's mail.

QuiteLiterallyAnything:
Might I suggest linking this to the mainframe (or a similar server-like object)? The feature itself implies some form of face tracking technology, and it would only make sense for that to be linked to an in-game computer. That would make for another thing antagonists could sabotage as well.

Ocelotwolfhorse:
Not sure how I feel about this specifically for the reason that expanding the inventory on manufacturers also expands the speed setting. A couple of jobs (QM, Roboticist) use the longer-lasting recipes enough to hack it purely to bump up that 3 to a 5, leaving the hacked recipes alone. Maybe the speed setting unlock could be detached from hacking it at all, and instead be a separate "fabricator upgrade" on it's own? Other than that, (seed vendors, lethal ammo from manufacturers, maybe even ALL vendors that have dangerous items?) I agree some stuff should be locked behind a little extra consequence.

Mikeistaken:
Here is my take on how it could be done to help balance. first if you want hacking to be treated as a crime you should add access locked hacking to the pool (vendors and doors) as well as do ALL of the expanded inventory hacking (vendors, fabricators,) for consistency. Second the camera likely wont be able to see you cut/pulse the wire that caused it, so I suggest having to make the device ping the mainframe to update everything (this is my excuse to add more packet based stuff) and make it so if the AI wire is cut that you don't get marked for arrest. Third is a hacking pass clarified on your ID, it would be given to engineers and/or the chief engineer, as well as the captain and portions of the security team (I recomend officer and up or NTSC and up)

JORJ949:
While I'd like to allow engineers to do it, I'd much rather make things you'd reasonably want to hack a thing for available elsewhere, e.g. rebalancing them to always be available rather than allowing only some people to hack. Mainly so HoP's that give themselves AA can't get around it (though they likely will anyway) the main goal is to discourage hacking a fabricator at roundstart as there's currently zero downside to doing so.

pgmzeta:
definitely should also apply to vending machines.

manufacturers currently don't keep forensics. if it's a crime then i would expect there to be evidence to collect. might be a bigger balance change.

Annmagedon:
most fabs don't really give you gamer gear plus i feel it will just make the painfully slow roboticist game-play loop worse.

even if its realistic, if its not fun , why add it

JORJ949:
> most fabs don't really give you gamer gear

- Infinite .22 ammo
- Security radio upgrades
- Infinite all access implants
And more

Annmagedon:
Point taken!

However i still dont think all fabs should have it, or move the soeed overclocking to be obtained in other way? Robotics w/ out overclocking really is painful

PR #25515, opened 2026-01-07
Adds tsunami spray bottle to a requisition

cheekybrdy:
## About the PR 
Readds the tsunami as a 25k purchase from the cultist trader.
## Why's this needed? 
The original tsunami was removed due to being a power-gamed space wreck loot item and to make the WA-V3 relevant. This adds it back in a finite, costly and market dependent way to warrant it coexisting with the WA-V3 but still have this different option be avaliable.
## Testing 
Loaded up on a local. Ordered from cargo.

## Changelog 

```changelog
(u)Cheekybrdy
(+)The Tsunami Spray Bottle can be bought from a trader at cargo for 25k credits.
```


CalliopeSoups:
I feel like the tsunami is so busted janitor-wise that it should be sold by an NPC trader instead of available at cargo, if you're bringing it back. 25k for a single player is an actual thing, 25k for cargo is nothing often.

TobleroneSwordfish:
Not really a fan of adding this thing back since when it does it exist in the round it makes everything else irrelevant.

Mikeistaken:
Not really a fan either, as its seriously what you would get if you got a cyborg tool that had a WA-V3 in it. However if it is added I suggest making it a requisition reward as we don't see much of those

Spyritdragon:
I think it being a reward from a particular requisition could be nice, but I'm definitely not sure I'm a huge fan of it finding a (semi-)consistent way to be obtained in its previous state. 

It just invalidates almost any other janitorial effort, including the WA-V3, in a way that doesn't really make for a fun presence in the game; a particularly strong thing in the game being hard to access doesn't really make it less problematically strong. While it's not a PvP item, I definitely remember being frustrated as a newbie janitor that someone had a massively powerful spray bottle that made anything I ever did feel pretty irrelevant. I don't think I like the idea of other newbies going through that, even if it's on a rare occasion, especially with how good of an introduction to the game janitor often is.

cheekybrdy:
I'll change this to a requsition reward like spyrit mentioned

cheekybrdy:
typomaxxing, changed to a requisition which requires scrap, miracle matter and cigarette butts

TobleroneSwordfish:
I'm just going to close this, sorry. Requisition rewards are generally niche but useful things, not direct upgrades that make all other tools irrelevant.

PR #25517, opened 2026-01-07
Adds the ability for skeletons to take their own head off easily.

Kanthes:
[Mutantraces][Player-Actions][Feature]

## About the PR 
This PR adds a mutantrace ability along the lines of the lizards ability to regrow their tail, but for skeletons to pop off their head. The head will be placed in the player's hands if possible, and otherwise be dropped on the floor.

## Why's this needed? 
Skeletons have the ability to do some hilarious gags with the ability to remove their head and staying alive, but doing so requires 2 different tools and 3 separate actions. This PR is meant to lower that threshold.

This PR specifically only allows for the quick removal of the head and not other limbs, as that could be considered an in game advantage for quick limb replacement.

## Testing 
Tested on a local server in the 3 following situations:
With free hands.
With full hands.
With the head already removed.

## Changelog 

```changelog
(u)Kanthes
(+)Skeletons can now take their head off at will without tools. Just pop it off!
```

### DIFF
--- a/code/datums/abilities/skeleton.dm
+++ b/code/datums/abilities/skeleton.dm
@@ -1,0 +1,37 @@
+/datum/abilityHolder/skeleton
+	usesPoints = FALSE
+	regenRate = 0
+
+/datum/targetable/skeletonAbility
+	icon = 'icons/mob/genetics_powers.dmi'
+	icon_state = "skeleton"
+	cooldown = 0
+	last_cast = 0
+	targeted = 0
+	preferred_holder_type = /datum/abilityHolder/skeleton
+	var/mob/living/carbon/human/L
+
+	onAttach(datum/abilityHolder/H)
+		. = ..()
+		if(ishuman(holder.owner))
+			L = holder.owner
+		return
+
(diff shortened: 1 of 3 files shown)


M-Earthfire:
Honestly it feels weird to add a whole mutantrace ability to the hud for an interaction that is achieveable with tools you can find (and print) in every tool storage or the arc plater area.

Kanthes:
> Honestly it feels weird to add a whole mutantrace ability to the hud for an interaction that is achieveable with tools you can find (and print) in every tool storage or the arc plater area.

An alternative would be to modify melee_attack_procs.dm to check if a skeleton uses Disarm Intent on themselves targetting the head. This would not take up any UI space and would be considerably less code. Would that be better?

Edit: Grab Intent would be more appropriate, but that behavior already makes the user Block. Disarm is the next best thing.

TobleroneSwordfish:
I think that would lead to a lot of skeletons accidentally swiping their own head off, which would be funny but probably isn't very good UX

pgmzeta:
give this to empowered ossification

PR #25518, opened 2026-01-07
Add triple water

Midnightradar:
## About the PR 
Adds a new reagent, triple water, which is a combination of water, tonic water, and holy water
## Why's this needed? 
Its been joked about on the discord server, and its a funny idea to make it a kind of joke drink like super milk, where the triple water is just denser, more water tasting water with more water per water content
## Testing 
mixes and tastes as intended

### DIFF
--- a/code/modules/chemistry/Reagents-FoodDrink.dm
+++ b/code/modules/chemistry/Reagents-FoodDrink.dm
@@ -4082,7 +4082,7 @@ datum
 			fluid_r = 10
 			fluid_g = 165
 			fluid_b = 254
-			description = "This water seems contains more water per water than normal water should."
+			description = "This water seems to contain more water per water than normal water should."
 			reagent_state = LIQUID
 			thirst_value = 1.5
 			bladder_value = -1.5


Mikeistaken:
Add triple water as an optional replacement to triple pissed in triple triple. that would be amazing

Midnightradar:
> Add triple water as an optional replacement to triple pissed in triple triple. that would be amazing

I mean, i think it would make triple triple a bit too easy to make, the bathsalts for triple pissed are the main challenge when making triple triple, though i think its the best to ask an admin what do they think of it, for now i just plan adding the triple water

Mikeistaken:
> > Add triple water as an optional replacement to triple pissed in triple triple. that would be amazing
> 
> I mean, i think it would make triple triple a bit too easy to make, the bathsalts for triple pissed are the main challenge when making triple triple, though i think its the best to ask an admin what do they think of it, for now i just plan adding the triple water

The old recipe for triple triple had triple piss instead of triple pissed, and that was just drinking piss and pissing iirc, so it seems fair to me

pgmzeta:
Trihydrogen oxide !!

Midnightradar:
> > > Add triple water as an optional replacement to triple pissed in triple triple. that would be amazing
> > 
> > 
> > I mean, i think it would make triple triple a bit too easy to make, the bathsalts for triple pissed are the main challenge when making triple triple, though i think its the best to ask an admin what do they think of it, for now i just plan adding the triple water
> 
> The old recipe for triple triple had triple piss instead of triple pissed, and that was just drinking piss and pissing iirc, so it seems fair to me

ew...

MintyPhresh:
Don't forget about soda water! could be **quadruple** water

code/modules/chemistry/Reagents-FoodDrink.dm
```diff
@@ -4076,6 +4076,18 @@ datum
 						new /obj/item/reagent_containers/food/snacks/plant/orange(M.loc)
 						new /obj/item/reagent_containers/food/snacks/plant/lemon(M.loc)
 
+		fooddrink/cocktail_triplewater
+			name = "triple water"
+			id = "cocktail_triplewater"
+			fluid_r = 10
+			fluid_g = 165
+			fluid_b = 254
+			description = "This water seems contains more water per water than normal water should."
```
TobleroneSwordfish:
```suggestion
			description = "This water seems to contain more water per water than normal water should."
```

Midnightradar:
> Could probably use a changelog entry, also if you want to be fancy you could make it work with any three water types including soda water (would likely need a few different child recipes)

when super milk was commited it also had no changelog, and since this is also a single joke chem i dont think it needs a changelog either

PR #25528, opened 2026-01-09
Trench Tech: Catwalk Bridges

pgmzeta:
[feature][experimental][player actions]
## About the PR 
Using rods on a trench will create a catwalk over it. Mobs in the trench can freely pass under catwalks.

Mobs outside of the trench can walk over intact catwalks as though the trench wasn't there.

Mobs in the trench and on top of the catwalk do not block each other. Mobs below are visually under the catwalk, but are still on the same turf.

You can use an empty hand on a catwalk placed over a trench to climb out of a trench, in addition to using sprint as before.

You cannot get out of a trench if there is an intact catwalk over you. Snipping tools and other items will damage catwalks over trenches; they are not super sturdy.

## Why's this needed? 
More ways to interact with trenches. Tech for future use.

Also, look at the video and tell me it isn't sick as hell.

## Testing 

https://github.com/user-attachments/assets/305c84af-091e-4747-a927-15d901c3e833

## Changelog 

```changelog
(u)glowbold
(+)Using rods on a trench will create a catwalk over it.
```

### DIFF
--- a/code/modules/digging/trench.dm
+++ b/code/modules/digging/trench.dm
@@ -33,6 +33,7 @@
 	var/old_type
 	var/can_fill = TRUE
 	var/base_icon_state = "trench"
+	var/catwalk_type = /obj/mesh/catwalk/over_trench
 
 	unfiilable
 		can_fill = FALSE
@@ -190,6 +191,7 @@
 	var/old_type
 	var/can_fill = TRUE
 	var/base_icon_state = "trench"
+	var/catwalk_type = /obj/mesh/catwalk/over_trench
 
 	unfillable
 		can_fill = FALSE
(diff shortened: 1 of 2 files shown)


code/modules/digging/trench.dm
```diff
@@ -288,6 +329,33 @@
 			playsound(L.loc, L.sound_scream, 100, 0, 0, L.get_age_pitch(), channel=VOLUME_CHANNEL_EMOTE)
 
 
+/datum/action/bar/icon/bridge_trench
+	interrupt_flags = INTERRUPT_MOVE | INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION
+	duration = 6 SECONDS
```
JORJ949, line 342: Should respect engi training/carpenter traits that let you build things faster.

TobleroneSwordfish:
How does this interact with movable objects?

pgmzeta:
> How does this interact with movable objects?

Right now, movable objects have no sense of planes; it's only tracked on the mob. Because of this, loose objects will always appear on top of catwalks as we're never modifying their plane.

There are no real defined rules for interactions between trenches and movable items, we would need to decide stuff like how dense objects like fuel tanks navigate trench edges. Right now, they don't really care so trying to place them in or out is moot.

TobleroneSwordfish:
I think it would look really odd to be pulling a movable under a catwalk and have it suddenly pop up onto the top.

pgmzeta:
will need to break down these changes into smaller jumps

PR #25529, opened 2026-01-09
Short-Sightedness Fix

### LINKED ISSUE #25464: Being short-sighted also makes you kind of deaf
### Labels

[BUG][MINOR]

### Description

When using the Short-Sighted trait and not wearing prescription glasses, overhead text outside of the 3x3 clear field of view is also covered by the 'fuzz' of the short-sightedness

### Steps to reproduce

1. Spawn in with the Short-Sighted trait
2. Take off your glasses
3. Have someone (or something capable of speech with floating text) stand at least two tiles away from you and say things out loud
4. Observe the floating text above their head

### Expected Behavior

Floating text should not be impacted by the 'fuzz' of the Short-Sighted trait and instead show above it so that it is readable. No I don't want to put the glasses on I want to be half-blind. But not half-deaf!

https://imgur.com/a/6L8fxwR - screenshot of floating text as seen by a glasses-less Short-Sighted spaceman

Reported by: showerless
Client version: 516.1667
Active test merges: nothing
Round log date: 2026-01-02-18-54
Reported on: 2026-01-02 19:02:36
Round ID: [220406](https://goonhub.com/rounds/220406)

QuiteLiterallyAnything:
[BUG] [QOL] [INTERNAL]
## About the PR 
Fixes #25464

This PR aims to make the overlays used by short-sightedness not cover overhead text chat. In pursuit of this, a new parameter has been added to the `create_screen()` proc so that `blend_mode` may be passed in. This parameter defaults to ~`BLEND_DEFAULT`~ `null`, but (in the interest of not breaking existing code) the proc has a check so it will not change anything _to_ ~`BLEND_DEFAULT`~ `null`. The `create_screen()` calls used by short-sightedness have been adjusted so that they use the same `blend_mode` as blindness ([`BLEND_ADD`](https://www.byond.com/docs/ref/#/atom/var/blend_mode)).

I reached this fix by first noting that blindness _doesn't_ have a similar issue, and from there I attempted to make short-sightedness closer to blindness code-wise. Eventually I determined that `blend_mode` was the main var that needed to be changed. 

The proc definition and calls have gotten rather long, and I can split them into multiple lines if desired as per this codebase's guidelines. 
## Why's this needed? 
Apart from the reasons mentioned in the linked issue, `BLEND_ADD` is used by most other visual bioeffects. I'm also unaware of any cleaner way to change `blend_mode` for all of the short-sightedness screens.
## Testing 
<details><summary>Expand for before screenshot. . .</summary>
Before_Change
</details>
<details><summary>Expand for after screenshot. . .</summary>
After_Change
</details>
Apart from making the above comparison, I also verified that this change has no impact on any of the colorblindness bioeffects when combined with short-sightedness. 

## Changelog 

```changelog
(u)QuiteLiterallyAnything
(+)Made short-sightedness not cover overhead text chat.
```

### DIFF
--- a/code/datums/hud.dm
+++ b/code/datums/hud.dm
@@ -169,8 +169,8 @@
 		S.show_tooltip = tooltip_options && length(tooltip_options)
 		S.tooltip_options = tooltip_options
 		S.mouse_opacity = mouse_opacity
-		if(blend_mode != null) // `blend_mode` is a newer parameter on this older proc. This check is made with the intention of not
-			S.blend_mode = blend_mode   // breaking things (primarily `customType` screens with a non-default `blend_mode`).
+		if(blend_mode != null) // `blend_mode` is a newer parameter on this older proc. This check is made with the intention of not breaking things.
+			S.blend_mode = blend_mode
 		src.objects += S
 
 		for (var/client/C in src.clients)


code/datums/hud.dm
```diff
@@ -145,7 +145,8 @@
 		for (var/atom/A in src.objects)
 			C.screen -= A
 
-	proc/create_screen(id, name, icon, state, loc, layer = HUD_LAYER, dir = SOUTH, tooltip_options = list(), desc = null, customType = null, mouse_opacity = 1)
+	/// Will not set the `blend_mode` of a `customType` to `BLEND_DEFAULT`.
+	proc/create_screen(id, name, icon, state, loc, layer = HUD_LAYER, dir = SOUTH, tooltip_options = list(), desc = null, customType = null, mouse_opacity = 1, blend_mode = BLEND_DEFAULT)
```
pgmzeta: why not have the default be null?

pgmzeta (changes requested):

code/datums/hud.dm
QuiteLiterallyAnything: I was previously under the impression that null would evaluate as equal to zero (and `BLEND_DEFAULT` by extension). [Nevermind that though](https://www.byond.com/docs/ref/#/DM/null). Changed.

PR #25530, opened 2026-01-09
Abstracts Barriers

cheekybrdy:
## About the PR 
Moves barriers into two abstract types, one for collapsible and one for non
## Why's this needed? 
Fixes a bug with the barrier special, it seems like a much better idea then adding a "isbarrier" variable, and makes future barriers easier to make
## Testing 
Ran a local, redeemed and checked path on token vendor spawned barrier and tested its functionality including EMP.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Abstracted barriers, report any issues with a bug report!
```

### DIFF
--- a/assets/maps/allocated/newbee_tutorial.dmm
+++ b/assets/maps/allocated/newbee_tutorial.dmm
@@ -73,7 +73,9 @@
 /area/tutorial/newbee/room_13)
 "cR" = (
 /obj/disposalpipe/segment/vertical,
-/obj/cable/auto,
+/obj/cable{
+	icon_state = "1-2"
+	},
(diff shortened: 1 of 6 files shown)


browserassets/src/images/tooltips/syndiebarrier.png
pgmzeta, line 1: is this used somewhere?

pgmzeta (approved):
will need secret change

browserassets/src/images/tooltips/syndiebarrier.png
cheekybrdy: unused item which has a special which didn't have a icon

PR #25531, opened 2026-01-09
Swaps 'Backgrounds' to 'Scenarios'

Pirate-Rob:
## About the PR
Renames the 'Background' trait category to 'Scenario'

## Why's this needed?
Baby's first PR. I don't think this change is strictly necessary but it did cause confusion for me returning to the game. Basically, there are a few uncategorised traits which would be described as backgrounds but don't go in backgrounds, which was confusing, until someone mentioned that backgrounds actually all change where you spawn and that traits in same categories are mutually exclusive of eachother
Changing 'Backgrounds' to 'Scenarios' accomplishes three things:
- Removes the confusion around traits that should be considered backgrounds not being backgrounds, and how some background traits are not really backgrounds.
- Makes potentially game-altering traits like Pilot appear further down in the trait list, so new players are less likely just to pick them due to first choice bias.
- Very clearly indicates the common theme between the Scenario traits, that they are starting scenarios.

Once again, this is very fluff but I just want to test out making a PR.

## Testing
image
image
image
image

Traits show up in menu, no problem spawning as them.

## Changelog
```changelog
(u)Pirate-Rob
(+)Renamed the 'Background' trait category to 'Scenario'.
```

### DIFF
--- a/code/datums/traits.dm
+++ b/code/datums/traits.dm
@@ -916,7 +916,7 @@ ABSTRACT_TYPE(/datum/trait/job)
 	desc = "You spawn hidden away on-station without an ID, PDA, or entry in NT records."
 	id = "stowaway"
 	icon_state = "stowaway"
-	category = list("scenario")
+	category = list("spawn scenario")
 	points = 0
 	unselectable = TRUE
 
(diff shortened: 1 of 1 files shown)


pgmzeta:
maybe "Special Spawn" ?

Pirate-Rob:
I was thinking 'Spawn Scenario' or some such but I wasn't sure if spaces would work in category names since I'm unfamiliar with byond

PR #25542, opened 2026-01-10
Tractor Beam Target Anchoring Checks

### LINKED ISSUE #25537: tractor beam moves locked/building ABCU.
### Labels

[BUG][MINOR]

### Description

if you have the tractor beam on while building it will drag with with you and build everything in the wrong spots

### Steps to reproduce

1. get abcu.
2. select blueprint and lock it.
3. drag it with a pod tractor beam.

### Expected Behavior

 for it to be stuck in place

Reported by: smallsandman
Client version: 516.1667
Active test merges: #25018
Round log date: 2026-01-09-22-44
Reported on: 2026-01-09 23:31:50
Round ID: [220765](https://goonhub.com/rounds/220765)
Map: [COGMAP](https://goonhub.com/maps/cogmap)

QuiteLiterallyAnything:
[BUG] [VEHICLES]
## About the PR 
Fixes #25537

I was only able to reproduce #25537 by tractor beaming an ABCU _before_ locking it. This doesn't line up with the steps described in the issue, but I believe it's safe to assume that's what was meant by it.

This PR adds two `if(src.target.anchored)` checks to tractor beams so that they will no longer move objects anchored after initially being caught by a beam. 
## Why's this needed? 
Tractor beams can cause a lot of buggy behavior (such as that described in the linked issue) when moving anchored objects.
## Testing 
On a debug build of the codebase I spawned an ABCU, caught it with a pod tractor beam, and failed to pull it after locking. I also verified that tractor beams still worked under more regular circumstances.

### DIFF
--- a/code/WorkInProgress/blueprints.dm
+++ b/code/WorkInProgress/blueprints.dm
@@ -37,7 +37,7 @@
 		This fine piece of machinery can construct entire rooms from blueprints."
 	density = 1
 	opacity = 0
-	anchored = UNANCHORED
+	anchored = UNANCHORED // Set to ANCHORED_ALWAYS when locked.
 	processing_tier = PROCESSING_FULL
 	event_handler_flags = NO_MOUSEDROP_QOL
 
@@ -353,7 +353,7 @@
 
 	proc/activate(mob/user)
 		src.locked = TRUE
-		src.anchored = ANCHORED
+		src.anchored = ANCHORED_ALWAYS
 		src.invalid_count = 0
 		for(var/datum/tileinfo/T in src.current_bp.roominfo)
 			var/turf/pos = locate(text2num(T.posx) + src.x,text2num(T.posy) + src.y, src.z)
--- a/code/modules/transport/pods/secondary_system.dm
+++ b/code/modules/transport/pods/secondary_system.dm
@@ -644,7 +644,7 @@ ABSTRACT_TYPE(/obj/item/shipcomponent/secondary_system/thrusters)
 	run_component()
 		if(settingup)
 			return
-		if(src.target.anchored)
+		if(src.target.anchored == ANCHORED_ALWAYS)
 			deactivate()
 			return
 		if(target in view(src.seekrange,ship.loc))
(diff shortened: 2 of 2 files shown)


TobleroneSwordfish:
Pods are anchored aren't they?

JORJ949:
Would also stop you pulling anchored mobs like magboot owners and shamblers. My suggestion would be that the ABCU gets ANCHORED_ALWAYS (i.e. shouldnt be moved by other stuff) while working and tractor beams respect ANCHORED_ALWAYS (if they don't already)

QuiteLiterallyAnything:
> Would also stop you pulling anchored mobs like magboot owners and shamblers. My suggestion would be that the ABCU gets ANCHORED_ALWAYS (i.e. shouldnt be moved by other stuff) while working and tractor beams respect ANCHORED_ALWAYS (if they don't already)

Swapped. Testing shows that tractor beams can still (for instance) move wrenched canisters like this.

QuiteLiterallyAnything:
> Pods are anchored aren't they?

I'm pretty sure tractor beams don't work on pods as is. There was already an `anchored` check in their `activate()` proc; this PR just adds two more to ensure they don't move something that _becomes_ `anchored` while they operate. 

While typing this I now realize that commit 85e25c22c25ccdd5552865e2ed9a52a3fa9598c7 might not do much at all. Pods still can't move _anything_ that's anchored when they activate.

pgmzeta:
> > Would also stop you pulling anchored mobs like magboot owners and shamblers. My suggestion would be that the ABCU gets ANCHORED_ALWAYS (i.e. shouldnt be moved by other stuff) while working and tractor beams respect ANCHORED_ALWAYS (if they don't already)
> 
> Swapped. Testing shows that tractor beams can still (for instance) move wrenched canisters like this.

We likely need a new anchor level to solve this class of problem in the general case, but having it respect ANCHORED_ALWAYS in the short term will solve this problem for a subset of objects - including the ABCU if it gets the ANCHORED_ALWAYS flag.

QuiteLiterallyAnything:
> We likely need a new anchor level to solve this class of problem in the general case, but having it respect ANCHORED_ALWAYS in the short term will solve this problem for a subset of objects - including the ABCU if it gets the ANCHORED_ALWAYS flag.

I mainly worry that ([as per my other comment](https://github.com/goonstation/goonstation/pull/25542#issuecomment-3733627193)) the current version of this PR will result in rather weird behavior. Tractor beams already don't include anchored or `ANCHORED_ALWAYS` objects, mobs, or pods in their target list when first activated. These checks are only for objects already captured by a tractor beam. 

I wasn't considering any of that when swapping the checks to `ANCHORED_ALWAYS`. I probably should've just emphasized this point instead.

PR #25545, opened 2026-01-10
Adds two new death and damage related traits

Laboredih123:
## About the PR 
Adds the Critical Existence Failure trait(stole name from tvtropes), which instantly kills you upon hitting crit and is +2 points. Adds the Glass Jaw trait, which has a 0.02 percent chance to multiply any brute, burn, tox, and oxy by 100, and is worth 0 points.
No icons because I can't really think of any and also aseprite no worky right currently.

## Why's this needed? 
death and injury is fun! lets people make pathetically weak beings.

## Testing 
monkey died with critical existence failure upon hitting crit.
monkey got slammed with damage with the glass jaw trait.

## Changelog 

```changelog
(u)cringe
(*)there are now two new death and damage related traits. one kills you instantly, the other hurts a LOT.
```

### DIFF
--- a/code/datums/traits.dm
+++ b/code/datums/traits.dm
@@ -1290,10 +1290,10 @@ TYPEINFO(/datum/trait/partyanimal)
 	id = "suddendeath"
 	points = 2
 
-/datum/trait/glassjoe
-	name = "Glass Joe"
+/datum/trait/glassjaw
+	name = "Glass Jaw"
 	desc = "There is a very teensy tiny chance for any damage received (brute, burn, toxin, oxy) to be multiplied by 100."
-	id = "glassjoe"
+	id = "glassjaw"
 	points = 0
 
 /datum/trait/carpenter
--- a/code/mob/living/carbon.dm
+++ b/code/mob/living/carbon.dm
@@ -162,7 +162,7 @@
 		if (!isdead(src))
 			H.emote(pick("wheeze", "cough", "sputter"))
 
-	if (src.traitHolder?.hasTrait("glassjoe") && prob(0.02))
+	if (src.traitHolder?.hasTrait("glassjaw") && prob(0.02))
 		amount *= 100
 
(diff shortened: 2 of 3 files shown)


Laboredih123:
gonna add stam damage to glass joe in like some hours (hours later: nvm no way to differentiate types of stamina removal)

JORJ949:
Not sure how i feel about the potential of important roles running Puritan+these two. The second is probably a bit too much in combination with puritan and the other existing, because then ANY SINGLE BIT OF DAMAGE could be instant permadeath at any moment, yeah its rare but would still happen and feels a bit meh.

Laboredih123:
i think thats even funnier

xxfishshadowxx:
> Not sure how i feel about the potential of important roles running Puritan+these two

its already possible to entirely cripple your character to dust with traits if that is your goal, this just adds more variety imo

Kanthes:
Should glass jaw maybe be reduced to brute only damage? Feels a bit weird if somebody breathed a bit of miasma and suddenly keels over.

Laboredih123:
> Should glass jaw maybe be reduced to brute only damage? Feels a bit weird if somebody breathed a bit of miasma and suddenly keels over.

no that increases the funi. slight choke and you just perish is hilarious

TobleroneSwordfish:
Not a huge fan of adding more of the "my character is made of wet paper" kind of traits, since people will inevitably stack them and then pick actually important roles.

pgmzeta:
What does this provide over other damage traits other than magnitude?

also people will absolutely use death-on-crit for microbombs

Crowlemon:
I don't think the worry about "death-on-crit" for microbombs would be much of a problem because people already can hit succumb the split second you're in crit anyway. 

I feel like critical existence is better served in timing and comedy just by possibly hitting succumb or the fact we can instantly die from things. If glass jaw exists, it'll happen anyway.

PR #25548, opened 2026-01-10
Chaplain spells

Egregorious:
## About the PR 

A proof of concept for chaplain spells, which can be purchased with faith and usable solely within the bounds of the chapel. These spells are designed to be RP focused, enabling decoration, spectacle and atmosphere, rather than provide any meaningful combat or practical utility. 

Spells are purchased from a spellbook object, the nature of which is yet to be finalised. 

Included a small number of spells as examples:

- Light and snuff candles: a pair of spells which can light or put out all candles currently in the chapel
- Chapel announcement: make a station announcement from the chapel
- Decoration spells: spells for placing objects or similar decorations, which might not work as regular ingame items. These spells are limited use, and must be repurchased to get more. 
  - Tree: creates a tree, for druids and such
  - Eternal Fire: a permanent, room temperature yellow fire with a very low burn-status duration and no atmos processing

Lore-wise, these probably wouldn't be actual magic spells. The reason the spells can't be cast outside the chapel isn't because of holy ground, it's that the chaplain has control of the space in the same way a stage magician does; the 'faith' cost denotes how much the crew is willing to buy what the chaplain is selling; and the atheist trait would allow people to glimpse through the smoke-and-mirrors. Trick candles, a hidden mic connected to the station's speaker system, holograms and cardboard boxes painted green. 

Spellbook:

Screenshot 2026-01-10 120513

Candles:

https://github.com/user-attachments/assets/360416cd-0596-42ec-9838-69c2c0781ed8

Announcement:

https://github.com/user-attachments/assets/d090411f-c3e6-4b50-bf39-bba5829aa109

Tree:

https://github.com/user-attachments/assets/277b1849-b296-40ca-ac22-61f6b5286572

Fire:

https://github.com/user-attachments/assets/927c3935-f59e-454b-96cf-51f3dfe8e588

- [ ] decide on spellbook object (convert bible into spellbook?)
- [x] make decorations destructable (desummonable)
- [ ] rework the spellbook tgui for non-wizard spells
- [ ] sfx
- [ ] find appropriate icons
- [ ] balance spell costs
- [ ] atheist interactions
- [ ] come up with and add additional spells
- [ ] atomise

## Why's this needed? 

The chaplain doesn't have great motivation from a game-mechanics standpoint. Faith is a cool mechanic, but the problem is that both its accumulation and its rewards are practically invisible, so it's lacklustre at encouraging the chaplain or making them feel rewarded for playing into their job. 

These spells will hopefully make the chaplain more powerful for their in-character role on the station; performing rituals such as funerals and weddings, and proselytising to the crew. While also helping the faith mechanic to encourage and reward them for getting folks to visit the chapel. 

## Testing 

Spawn 'chaplain_uplink', which works as a wizard spellbook. The 'training_chaplain' trait is required to be able to access the spells, easiest way is to respawn self as a chaplain. A chapel area is also required, there isn't one in the dev testing zone. Or comment out lines 6-13 in [code/datums/abilities/chaplain.dm](https://github.com/goonstation/goonstation/compare/master...Egregorious:Chaplain-spells?expand=1#diff-591d8209ecf8e614930908071ff1cfc8fa5f49c062db09789bf72ad1ffb3d6b9)


Tarmunora:
Hm. Not huge on how this reframes the role from "station chaplain that may actually have some divine favor" to "d&d cleric actively channeling miracles".

Egregorious:
The plan would be to have atheist trait interactions that imply it's less genuine miracles and more charlatan tricks, which I think would work for abilities which don't have combat impact.

TobleroneSwordfish:
The only one that I could see being deniable enough to maaybe pass is the candle trick, the others are pretty blatantly magic (especially the booming god voice)

Egregorious:
Hm, I don't agree, I feel that lacks a bit of imagination considering we are working with future-tech, but the phrasing absolutely could be changed up to have different implications. The message uses the department announcement formatting already to somewhat imply it's coming out the speakers as usual, and in-universe surely editing the PA system to have a different announcement bleep is absolutely in the realm of possibility. I was thinking that atheists could recieve the announcement with a regular department header instead of a unique one. 

The decorations could be styled as hallucinations caused by belief, I was working on getting atheists to see the decorations as cardboard boxes.

Egregorious:
Atheists see conjured decoration objects differently:

https://github.com/user-attachments/assets/af50b1b0-d143-439b-9932-33ab349a8170

JORJ949:
> Atheists see conjured decoration objects differently:
> 

Without getting into my overall opinions on the PR, not a fan of this example. Imagine hiding behind what you thought was a tree but John Security can just see you because it was a chaplain box.

Egregorious:
> Without getting into my overall opinions on the PR, not a fan of this example. Imagine hiding behind what you thought was a tree but John Security can just see you because it was a chaplain box.

I see what you mean, but I think that problem would be exacerbated with similarly-sized but differently-shaped sprites, because it becomes a question of which part is visible and not visible, which gets very complex after adding different types of decorations. At least if it's all-or-nothing it's much more predictable. 

I suppose there could be a unique same-shaped fake-looking sprite for each decoration object, but that adds significant sprite work requirements. Personally I doubt a trait that allows you to see through a small number of player-made objects in a very limited area breaks much in the way of sec-antag balancing. Hiding behind sprites isn't exactly equal to invisibility in the first place.

pgmzeta:
I do not think it should be something you pick from a list like it's a vendor; it's too concrete a link.

Egregorious:
> I do not think it should be something you pick from a list like it's a vendor; it's too concrete a link.

I'm not quite sure what you mean by 'link'. 

A different method of distribution could be cool, but I can't think of what that could be. There are numerous benefits of reusing the wizard spellbook UI, and I'm not sure what could work while still providing the same sort of scalable information you can relay with a simple list of options. 

Perhaps it could be a bookshelf, and you can take a book representing the ability off the shelf to use it. But that would probably still need a vendor-like UI.

Could hide objects around the chapel, and interacting with them allows you to buy the spell. That's not very scalable, and adds a lot of knowledge-requirement, but could be interesting.

pgmzeta:
why not rituals

Egregorious:
Most of the dev input I've recieved has wanted to avoid chaplains having access to real magic, which rituals lean into quite heavily. If it's a spellbook type thing, atheists can see the UI as a 'Charlatanry for Dummies' while non-atheists believe they're reading holy text. I don't see how the same sort of dual-presentation can be done for rituals, they're very magic-coded. 

If it wasn't for that, then performing a ritual to unlock an ability sounds like a fun way to do it. There'd be some details to iron out, like how to display faith, but I'm sure it'd be workable.

Egregorious:
I lost momentum on this, I'll come back to it later.

PR #25555, opened 2026-01-10
Azone Medal Fix

cheekybrdy:
## About the PR 
Fixes the REDACTED to properly give in a radius instead of just the person using REDACTED
## Why's this needed? 
Fix Good
## Testing 
Ran on a local, can't test medals really any further.


cheekybrdy:
A single runtime about wirecutters in a wall on oshan? huh

PR #25563, opened 2026-01-11
Donut 2 lighting fix.

smallsandman:
## About the PR 
fixes the lighting bug in inner maint
## Why's this needed? 
lighting bug was annoying!
## Testing 
image

```

### DIFF
--- a/maps/donut2.dmm
+++ b/maps/donut2.dmm
@@ -14319,6 +14319,12 @@
 	},
 /turf/simulated/floor/black/grime,
 /area/station/crew_quarters/catering)
+"bKJ" = (
+/obj/cable/orange{
+	icon_state = "4-8"
+	},
+/turf/simulated/wall/auto/reinforced/supernorn,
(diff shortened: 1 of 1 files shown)


JORJ949:
> What is the lighting bug and how does this fix the lighting bug?

The lighting bug is that space lighting is seemingly now partly area based, meaning all /area/space are fully illuminated this is just a bandaid that fixes this one instance of it being a problem. Is also a problem around the armory, moreso when terrainifyd, because the turret area is not /area/space.

Kanthes:
Here's how it looks when viewed in game before fix:

image

TobleroneSwordfish:
This did not fix the lighting bug
image

smallsandman:
> This did not fix the lighting bug
> image

I had considered making the area around there too, but I couldn’t think of what area it should be considering it’s also connected right next to a smes or solar room directly, so I just fixed it from the inside, though I can make a fix for the outside too

PR #23503, opened 2025-05-17
Update most pod UI sprites

FlameArrow57:
[VEHICLES]

## About the PR 
Updates pod UI sprites

Doesn't do secondary system sprites

Sprites by @Mister-Moriarty 

image

## Why's this needed? 
They look nice

PR #25564, opened 2026-01-11
Redrawn Pod Control Icon Borders and some images

Pirate-Rob:
Labels: [A-Vehicles] [C-Sprites]

## About the PR
Borders of the pod control icons redrawn.
Pod control icons now have a little light indication whether they're on or off (or available/unavailable)
Some icon images also redrawn.

## Why's this needed?
I got into a pod for the first time in a long time, and was stumped for 2 minutes trying to process the icon bar, the obscene amount of text (om/off/off/on/off) was very distracting and doesn't help very much with figuring out what each button actually does. Further more it wasn't easy to tell if a system was on or off, since no colours change, only 2 characters (n -> ff), which is difficult as someone who struggles to read that sort of text on the fly.
On top of this, the old "unavailable" (X) marker that appears over icons if they can't be turned on totally blocks the icon image, so you can't even tell what system is there.
This redraw makes the borders and the [On/Off] of the icons take up less space, so more space can be used for images.
Additionally makes it cleaner to read as there is no distracting text, and proper brightness/colour differences between states so you can easily tell if something is off or on.
Most of the images on the icons were kept virtually the same with some minor recolouring, however I redrew some of the icons to make them a bit larger so they would utilise space better. Many icons also now have a colour associated with them for quick visual recognition and distinction, since everything being the same colour makes everything blend together badly.

## Testing 
Just a sprite swap, so nothing should physically be able to break.
image
image
Everything does work in game however!

## Changelog
```changelog
(u)Pirate-Rob
(+)Pod interior control buttons redrawn.
```


pgmzeta:
the current sprites definitely do need work, but i don't know that the sprites in this PR are it. There's been a [different approach ](https://github.com/goonstation/goonstation/pull/23503) that people have liked, but it's not fully implemented. Perhaps a combination would be best?

Pirate-Rob:
Yeah I'm aware about it, but from what I have been told it needs code changes and no one is working on it, so I just wanted something to fill the gap in the meantime until someone actually finishes implementing the other sprites

PR #25565, opened 2026-01-11
Using an emag on a genetics console makes the gene scrambler radiation much more potent. Also prevents demagging by deconstructing computer consoles by keeping the emagged state on the curcuit board.

M-Earthfire:
[Feature][Game Objects]

## About the PR 

Adds an emag interaction to the genetics console. When used on a console, the scrambler cooldown gets halved and the power of the radiation applied by scrambling gets increased by 33% (2 sieverts, up from 1.5) and becomes unable to be altered by research.

Also, this Pr makes it so that emagged consoles keep the curcuitboard emagged upon deconstruction.

## Why's this needed? 

This enables antag geneticists to use the gene pods scrambler much better for cooking people with radiation. Especially gene research can reduce to completely prevent the damage caused by scrambling, so completely negating that with the emag enables you to do research while keeping your gimmick killing weapon available.

Also, halving the cooldown adds the benefit of getting faster through research, although at a high casualty rate of monkeys. It also has the benefit of making the sabotage obvious by checking the cooldown.

The idea came to me from a youtube short (https://www.youtube.com/shorts/rQEP_ELxm7c) so my brain shortcircuited the idea into a Pr.

The emag changes towards console boards was so that an emagged computer console cannot be demagged by simply deconstructing and constructing it again.

## Testing 

https://github.com/user-attachments/assets/4f47689b-359f-4c63-93b9-14f6d362b304

## Changelog 

```changelog
(u)Lord_Earthfire
(+)Emagging a genetics console fries its radiation dose controller, vastly improving the gene scanners ability to cook someone alive.
(+)Deconstructing an emagged computer console now transfers the emagged state onto the curcuitboard, making consoles build with that stay emagged.
```


code/modules/medical/genetics/geneticsMachines.dm
```diff
@@ -39,6 +39,9 @@
 	var/decrypt_correct_char = "?"
 	var/decrypt_correct_pos = "?"
 	var/datum/genetics_appearancemenu/modify_appearance = null
+	var/emagged = FALSE
+	var/emitter_radiation_minimum_strenght = 0 //! modifies the scrambler power. Used for the emag-effect
```
TobleroneSwordfish: strenght

M-Earthfire: goddamnit, i feel that's the third PR where i have written strength wrong...

M-Earthfire:
> Honestly 33% is still not a huge bonus but I don't really know how deadly the baseline microwaving is since geneticists never use it soo maybe fine.

For reference, the first scramble with the emag leaves you at around 30 burn/tox after the 60 minute mark with around half of the radiation (1.15 Sv,  slightly irradiated) dose left in you. The second scramble at that point kills you pretty fast, since 3+ Sv counts as extremly irradiated.

A single scramble with the emagged console likely doesn't put you into crit.

code/modules/medical/genetics/geneticsMachines.dm
```diff
@@ -55,6 +58,16 @@
...
+/obj/machinery/computer/genetics/emag_act(var/mob/user, var/obj/item/card/emag/used_emag)
+	if(src.emagged)
+		return FALSE
+	src.emagged = TRUE
+	src.visible_message(SPAN_ALERT("[src]'s radiation dose control seems to malfunction. That doesn't look safe..."))
+	//This is 33% stronger than the emitters baseline radiation, before research. This makes sure that our clown inside comes out well done.
+	src.emitter_radiation_minimum_strength = 100
+	src.emitter_cooldown_multiplier = 0.5
+	return TRUE
+
```
pgmzeta, line 69: demag proc please

pgmzeta:
will unscrewing and rescrewing the machine back together fix this? maybe it should go on the actual scanner object itself?

M-Earthfire:
> will unscrewing and rescrewing the machine back together fix this? maybe it should go on the actual scanner object itself?

As it would do with all consoles.

The problem is with the portable clone scanner, which is a child of the genetics console. By going with the console itself, i have caught that one.

Having that said... i just turned it into a general behaviour for computers, so now all computers transfer their emagged base on the board and the other direction as well.

code/modules/medical/genetics/geneticsMachines.dm
M-Earthfire: was added

M-Earthfire:
I'm rerequesting leah again here since the PR got a little scope-creeped... sorry

PR #25571, opened 2026-01-13
Collar shirts can now cycle between untucked, half-tucked, and fully tucked.

DisturbHerb:
## About the PR 

Collar shirts can cycle between untucked, half-tucked, and fully-tucked into the trousers using an ability button or by using them in-hand.

image

Top to bottom:
- Untucked (previously the normal version)
- Half-tucked
- Fully tucked
- Untucked w/ jacket
- Half-tucked w/ jacket
- Fully tucked w/ jacket

In the future, all of our clothing components for changing styles such as (un)buttoning, hoods, and now shirt tucking should really be unified. For now, I simply add onto our existing tech debt.

## Why's this needed? 

I was getting sick of how long the hem of these things can get.

## Testing 

Compiled and working in-game.


code/datums/components/cycle_tuck.dm
```diff
@@ -0,0 +1,67 @@
+// I beg for `toggle_hood.dm`, `toggle_coat.dm`, and this to be unified at some point. This is not the time. - DisturbHerb
```
pgmzeta, line 1: TODO:

icons/misc/abilities.dmi
pgmzeta, line 1: these ability icons are difficult to parse but at the same time i don't know that there's a visual shorthand.

pgmzeta (approved):
maybe some other groups of clothing should be styles instead if they can use this structure.

PR #25575, opened 2026-01-14
Adds Experiences of the Deep to all ocean maps

cheekybrdy:
## About the PR 
Adds The Experiences of the Deep to all trenches, regardless of the station map.
## Why's this needed? 
Forces it into a lowpop map, and gives more voting variety on people who want to do this azone. I was originally hesitant given the lore, but with aggressive kudzu and other recent less lore friendly changes I think this would be fine.
## Testing 
Ran on a local on oshan and nadir, no spawn on nadir, spawned on oshan.

## Changelog 

```changelog
(u)Cheekybrdy
(+)Experiences of the Deep is available on the trench of all Oshan maps.
```


Annmagedon:
I dont think it should be on nadir.

Its another planet (magus) with different lore, opposed to neon & oshan (abzu)

cheekybrdy:
its flagged as nadir unsafe so it shouldn't be on there (acid kills some of the mobs)

PR #25576, opened 2026-01-14
Warnings for Headless Skeleton Changeling Transformation

### LINKED ISSUE #25569: Killed by a bug regarding changelings and skeletons
### Labels

[BUG][MAJOR]

### Description

Killed by a bug regarding changelings and skeletons

### Steps to reproduce

1. be a changeling
2. have a skeleton persona
3. detach your head
4. change form to another skeleton or a skeleton monkey
5. die I guess

### Expected Behavior

to have the monkey head stay alive and the monkey body stay alive

Reported by: billyoneshot
Client version: 516.1675
Active test merges: nothing
Round log date: 2026-01-13-06-28
Reported on: 2026-01-13 08:07:40
Round ID: [220930](https://goonhub.com/rounds/220930)
Map: [DONUT3](https://goonhub.com/maps/donut3)

QuiteLiterallyAnything:
[QoL] [Player Actions]
## About the PR 
Closes #25569

This PR adds warnings for skeleton changelings attempting to transform without a head.
## Why's this needed? 
As per the issue (which I've managed to successfully reproduce) transforming into a monkey or simply switching identities as a headless skeleton is currently lethal. When switching identities this even runtimes, so I presume it's a bug. It does make sense as an emergent feature though, so this PR doesn't outright remove it (though I could alternatively do that if desired). 
## Testing 
On a debug build of the codebase I verified that the warnings appeared and functioned as appropriate. I also verified that the warnings do not interfere with regular changeling transformation as both a human and skeleton.

### DIFF
--- a/code/modules/antagonists/changeling/abilities/transform.dm
+++ b/code/modules/antagonists/changeling/abilities/transform.dm
@@ -91,7 +91,7 @@
 			boutput(holder.owner, SPAN_ALERT("We need to absorb more DNA to use this ability."))
 			return 1
 
-		if(headless_skeleton_warning()) // Headless skeletons die if they use this.
+		if(src.headless_skeleton_warning()) // Headless skeletons die if they use this.
 			return 1
 
 		var/target_name = tgui_input_list(holder.owner, "Select the target DNA:", "Target DNA", sortList(H.absorbed_dna, /proc/cmp_text_asc))


code/modules/antagonists/changeling/abilities/transform.dm
```diff
@@ -20,6 +20,9 @@
 		var/mob/living/carbon/human/H = holder.owner
 		if(!istype(H))
 			return 1
+		if(isskeleton(H) && !H.organHolder.head) // Warn them if they're a headless skeleton.
+			if(tgui_alert(H, "We currently lack a head. Shifting forms in this state WILL kill us.", "Hesitation", list("Continue Anyways", "Stop")) != "Continue Anyways")
+				return 1
```
ZeWaka: Why are we not using the proc you made here?

QuiteLiterallyAnything: I'd have to have the same proc on both `/datum/targetable/changeling/transform` and `/datum/targetable/changeling/monkey`. I could also put it on their parent type though it seems like it wouldn't fit there due to its incredibly specific use case. Creating another parent for changeling transformation abilities could work, but it might be excessive for one proc. 

Thoughts?

pgmzeta: move it up to the parent `/datum/targetable/changeling` and reference from both

that check should also be used for headspiders, but that is for another PR

QuiteLiterallyAnything: Done

code/modules/antagonists/changeling/abilities/transform.dm
```diff
@@ -20,6 +20,8 @@
 		var/mob/living/carbon/human/H = holder.owner
 		if(!istype(H))
 			return 1
+		if(headless_skeleton_warning()) // Headless skeletons die if they use this.
```
JORJ949:
```suggestion
		if(src.headless_skeleton_warning()) // Headless skeletons die if they use this.
```

code/modules/antagonists/changeling/abilities/transform.dm
```diff
@@ -89,6 +91,9 @@
 			boutput(holder.owner, SPAN_ALERT("We need to absorb more DNA to use this ability."))
 			return 1
 
+		if(headless_skeleton_warning()) // Headless skeletons die if they use this.
```
JORJ949:
```suggestion
		if(src.headless_skeleton_warning()) // Headless skeletons die if they use this.
```

PR #25577, opened 2026-01-14
Headless Skeleton Changeling Runtime Fix

QuiteLiterallyAnything:
[Runtime] [Player Actions]
## About the PR 
Related to PR #25576

This PR fixes a singular runtime for headless skeleton changelings attempting to switch identities. I'm honestly unsure if an `if` conditional or `.?` operator is better for something like this, so I've effectively flipped a coin and gone with the former.
## Why's this needed? 
Fixing runtime errors is a good thing. Aside from that:

<details><summary>Expand for a screenshot of the runtime error. . .</summary>

image

</details>

## Testing 
On a debug build of the codebase I switched identities as a headless skeleton changeling. This resulted in immediate death, but it didn't runtime.

### DIFF
--- a/code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
+++ b/code/modules/antagonists/changeling/abilities/_changeling_ability_holder.dm
@@ -235,7 +235,8 @@
 
 		human.bioHolder.RemoveEffect("husk")
 		human.real_name = src.name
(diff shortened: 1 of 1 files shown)


ZeWaka:
i'd use a ?. but this is fine, too pedantic really

